PR #2946, opened 2022-03-09
[mini-PR] Fix incomplete docstring

lucafedeli88:
This PR fixes an incomplete docstring comment


ax3l:
Jup, saw that as well in #2953.

Will use your exact wording there :)

PR #2948, opened 2022-03-09
Mesh Refinement: Add possibility to remove the finest level at a given step

tmsclark:
Co-authored-by: Luca Fedeli <luca.fedeli@for.unipi.it>

*This pull request add the possibility to remove the finest level during mesh refinement simulations.

    -Add an input parameter: `end_fine_patch_step`, step at which the finest level is removed

    -Multiply by two the timestep after the removal of the finest level when do_subcycling=0

The finest level is removed in `WarpX::Evolve`, in `OneStep_nosub`  and `OneStep_sub1` with `regrid(coarse_lev, cur_time)` at the `end_fine_patch_step`, followed  by `warpx.ComputeDt()` in the case of `do_subcycling=0`.

Finally, max_level is replaced at several places by finest level since `max_level` corresponds to the value given at the start. The purpose of WarpX::FinelevInit_flag is to use `max_level` until `finestLevel()` is initialised.

~~⚠️ Does not work on GPUs !! Help needed 🚨~~


Docs/source/usage/parameters.rst
```diff
@@ -166,6 +166,11 @@ Setting up the field mesh
 
     Note: in development; currently allowed value: ``2 2 2``.
 
+* ``warpx.end_fine_patch_step`` (`integer`, default: ``-1``)
+    When using mesh refinement, this can be used to remove the finest level at a given timestep.
+
+    Note: in development; currently, only working on CPU. Moreover, the max_step is modified and we encourage you to use stop_time instead.
```
lucafedeli88: I would add few more details concerning the max_step
```suggestion
    Note: in development; currently, only working on CPU. Moreover, as soon as the fine patch is removed, ``max_step`` is modified so that the final step corresponds to the physical time calculated with the initial timestep duration. Therefore, users are advised to use ``stop_time`` instead.
```

Docs/source/usage/parameters.rst
```diff
@@ -166,6 +166,11 @@ Setting up the field mesh
 
     Note: in development; currently allowed value: ``2 2 2``.
 
+* ``warpx.end_fine_patch_step`` (`integer`, default: ``-1``)
+    When using mesh refinement, this can be used to remove the finest level at a given timestep.
```
lucafedeli88, line 246: I would also add a couple of lines to explain what happens to the timestep duration.

Source/Diagnostics/Diagnostics.cpp
```diff
@@ -362,6 +361,14 @@ Diagnostics::FilterComputePackFlush (int step, bool force_flush)
     WARPX_PROFILE("Diagnostics::FilterComputePackFlush()");
     MovingWindowAndGalileanDomainShift (step);
 
+    const auto& warpx = WarpX::GetInstance();
+    const auto finest_level = warpx.finestLevel();
+
+    for (auto& buf : m_mf_output){
+        if(buf.size() > finest_level+1) buf.pop_back();
+    }
```
lucafedeli88: I would suggest a `resize` here. Moreover, I think that adding a comment before the `for` loop could help developers working on this part of the code in the future.

Source/Evolve/WarpXComputeDt.cpp
```diff
@@ -22,17 +22,19 @@
...
+#include <AMReX_AmrMesh.H>
 #include <algorithm>
 #include <memory>
 
 /**
  * Determine the timestep of the simulation. */
 void
 WarpX::ComputeDt ()
-{
+{   int Finelev;
```
lucafedeli88:
```suggestion
{
    int Finelev;
```

Source/Evolve/WarpXComputeDt.cpp
```diff
@@ -22,17 +22,19 @@
...
+    if (!FinelevInit_flag) Finelev = max_level;
```
lucafedeli88: I would suggest adding a comment to clarify why this is needed.

Source/Evolve/WarpXComputeDt.cpp
```diff
@@ -84,8 +86,10 @@ WarpX::ComputeDt ()
 
 void
 WarpX::PrintDtDxDyDz ()
-{
-    for (int lev=0; lev <= max_level; lev++) {
+{   int Finelev;
```
lucafedeli88:
```suggestion
{
    int Finelev;
```

Source/Evolve/WarpXComputeDt.cpp
```diff
@@ -84,8 +86,10 @@ WarpX::ComputeDt ()
...
+    if (!FinelevInit_flag) Finelev = max_level;
```
lucafedeli88: I would suggest adding a comment to explain why this is needed.

Source/Evolve/WarpXComputeDt.cpp
```diff
@@ -84,8 +86,10 @@ WarpX::ComputeDt ()
...
+    else Finelev = AmrMesh::finestLevel();
+    for (int lev=0; lev <= Finelev; lev++) {
         const amrex::Real* dx_lev = geom[lev].CellSize();
         amrex::Print() << "Level " << lev << ": dt = " << dt[lev]
```
lucafedeli88: You can use the new `Utils::TextMsg::Info(msg)` text formatter ;-)

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -45,7 +45,6 @@
 #include <AMReX_REAL.H>
 #include <AMReX_Utility.H>
 #include <AMReX_Vector.H>
-
```
lucafedeli88, line 50: I would actually keep the space here

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -376,6 +375,17 @@ WarpX::OneStep_nosub (Real cur_time)
     // Deposit current j^{n+1/2}
     // Deposit charge density rho^{n}
 
+    if(istep[0]+1==end_fine_patch_step){
+        auto& warpx = WarpX::GetInstance();
+        SyncCurrent();
+        SyncRho();
+        const int coarse_lev = 0;
+        regrid(coarse_lev, cur_time);
+        Print() << "Remove the patch" << '\n';
```
lucafedeli88: You could use the new `Utils::TextMsg::Info` text formatter ;-)

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -787,6 +798,14 @@ WarpX::OneStep_sub1 (Real curtime)
     }
     if ( safe_guard_cells )
         FillBoundaryB(coarse_lev, PatchType::fine, guard_cells.ng_FieldSolver);
+
+    if(istep[0]+1==end_fine_patch_step){
+        SyncCurrent();
+        SyncRho();
+        regrid(coarse_lev, curtime);
+        Print() << "Remove the patch" << '\n';
```
lucafedeli88: You could use the new `Utils::TextMsg::Info` text formatter ;-)

Source/WarpX.H
```diff
@@ -763,6 +763,11 @@ public:
     static amrex::Real moving_window_v;
     static bool fft_do_time_averaging;
 
+    //Parameters to remove the fine patch at a given time step
```
lucafedeli88:
```suggestion
    //Parameter to remove the fine patch at a given time step
```

Source/WarpX.H
```diff
@@ -763,6 +763,11 @@ public:
...
+    static int end_fine_patch_step;
+    //Flag which said when the finest_level element is initialised
```
lucafedeli88:
```suggestion
    //This flag is set to true as soon as the finest_level element is initialized
```

Source/WarpX.cpp
```diff
@@ -633,6 +636,18 @@ WarpX::ReadParameters ()
             moving_window_v *= PhysConst::c;
         }
 
+
+        if(max_level>0){
+            const auto end_fine_patch_defined=queryWithParser(pp_warpx,"end_fine_patch_step", end_fine_patch_step);
+            if(has_max_step && end_fine_patch_defined && do_subcycling==0){
+                auto ss = std::stringstream{};
+                if(max_step>=end_fine_patch_step){
+                    ss << "When the fine patch is removed, the max step will be " << end_fine_patch_step + std::floor(0.5*(max_step-end_fine_patch_step)) ;
```
lucafedeli88: You could add to the message that the advice is to use the final time

lucafedeli88 (changes requested):
Thanks a lot for this PR @tmsclark  ! Great feature ! I left few comments, but nothing very substantial...

Source/Evolve/WarpXComputeDt.cpp
WeiqunZhang: I don't think this function needs to be modified.  It should be OK to keep the size of dt fixed at `max_level+1`.

Source/Evolve/WarpXComputeDt.cpp
WeiqunZhang: Again, I don't think this needs to be modified.  We could also always print from level 0 to the finest level if in InitData we move the call to WarpX::PrintDtDxDyDz after `InitFromScratch` is called.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -376,6 +375,17 @@ WarpX::OneStep_nosub (Real cur_time)
     // Deposit current j^{n+1/2}
     // Deposit charge density rho^{n}
 
+    if(istep[0]+1==end_fine_patch_step){
+        auto& warpx = WarpX::GetInstance();
+        SyncCurrent();
+        SyncRho();
+        const int coarse_lev = 0;
+        regrid(coarse_lev, cur_time);
```
WeiqunZhang, line 427:
```suggestion
        regrid(coarse_lev, cur_time);
        mypc->Redistribute();
```

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -787,6 +798,14 @@ WarpX::OneStep_sub1 (Real curtime)
     }
     if ( safe_guard_cells )
         FillBoundaryB(coarse_lev, PatchType::fine, guard_cells.ng_FieldSolver);
+
+    if(istep[0]+1==end_fine_patch_step){
+        SyncCurrent();
+        SyncRho();
+        regrid(coarse_lev, curtime);
```
WeiqunZhang, line 948:
```suggestion
        regrid(coarse_lev, curtime);
        mypc->Redistribute();
```

Source/Diagnostics/Diagnostics.cpp
WeiqunZhang: Agreed.

ax3l:
When we remove the patch, do target ions and electrons get moved to level 0?
In diags of `rho_tgt_ele`/`_ion` in one of our tests, it looks to me like in a step after patch removal they are still on level 1 (because I cannot find them in `rho_..` of level 0).

I think with the current logic: regrid then redistribute, the particles might remain on level 1?
https://github.com/AMReX-Codes/amrex/blob/development/Src/Particle/AMReX_ParticleContainerI.H#L1358-L1373

update: according to a test case I did, it looks like particles move to level zero already.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -377,6 +376,18 @@ WarpX::OneStep_nosub (Real cur_time)
     // Deposit current j^{n+1/2}
     // Deposit charge density rho^{n}
 
+    if(istep[0]+1==end_fine_patch_step){
+        auto& warpx = WarpX::GetInstance();
+        SyncCurrent();
+        SyncRho();
+        const int coarse_lev = 0;
+        regrid(coarse_lev, cur_time);
+        mypc->Redistribute();
```
ax3l, line 428: see comment here: https://github.com/ECP-WarpX/WarpX/pull/2948#issuecomment-1084139026

I think particles are not moved from level 1 to level 0 here:
I think with the current logic: regrid then redistribute, the particles might remain on level 1?
https://github.com/AMReX-Codes/amrex/blob/development/Src/Particle/AMReX_ParticleContainerI.H#L1358-L1373

ax3l: update: according to a test case I did, it looks like particles move to level zero already.

ax3l: otherwise I would have had tested this diff:
```diff
diff --git a/Source/Evolve/WarpXEvolve.cpp b/Source/Evolve/WarpXEvolve.cpp
index c244fc1683..58ec676633 100644
--- a/Source/Evolve/WarpXEvolve.cpp
+++ b/Source/Evolve/WarpXEvolve.cpp
@@ -382,6 +382,7 @@ WarpX::OneStep_nosub (Real cur_time)
         SyncRho();
         const int coarse_lev = 0;
         regrid(coarse_lev, cur_time);
+        mypc->moveParticlesToLevel0();
         mypc->Redistribute();
         Print() << "Remove the patch" << '\n';
         warpx.ComputeDt()  ;
diff --git a/Source/Particles/MultiParticleContainer.H b/Source/Particles/MultiParticleContainer.H
index 56f1b42ff3..840baac762 100644
--- a/Source/Particles/MultiParticleContainer.H
+++ b/Source/Particles/MultiParticleContainer.H
@@ -213,6 +213,8 @@ public:
 
     void Redistribute ();
 
+    void moveParticlesToLevel0 ();
+
     void defineAllParticleTiles ();
 
     void RedistributeLocal (const int num_ghost);
diff --git a/Source/Particles/MultiParticleContainer.cpp b/Source/Particles/MultiParticleContainer.cpp
index 156f90905d..252da4327a 100644
--- a/Source/Particles/MultiParticleContainer.cpp
+++ b/Source/Particles/MultiParticleContainer.cpp
@@ -571,6 +571,28 @@ MultiParticleContainer::Redistribute ()
 {
     for (auto& pc : allcontainers) {
         pc->Redistribute();
+        std::cout << "number of particles at level 0: " << pc->NumberOfParticlesAtLevel(0) << std::endl;
+        std::cout << "number of particles at level 1: " << pc->NumberOfParticlesAtLevel(1) << std::endl;
+    }
+}
+
+void
+MultiParticleContainer::moveParticlesToLevel0 ()
+{
+    int const old_finest_lvl = 1;
+    for (auto& pc : allcontainers) {
+        // copy
+        for (WarpXParIter pti(*pc, old_finest_lvl); pti.isValid(); ++pti)
+        {
+            int const new_lvl = 0;
+            int const ngrow = 0;
+
+            auto& tile = pc->ParticlesAt(old_finest_lvl, pti);
+            pc->AddParticlesAtLevel(tile, new_lvl, ngrow);
+        }
+
+        // remove
+        pc->RemoveParticlesAtLevel(old_finest_lvl);
     }
 }
```

Source/Evolve/WarpXComputeDt.cpp
tmsclark: HI, I choose to use Finelev here to print only the value of dt of the coarsen level when the patch is removed. If we keep max_level+1, it will print the value of dt on both levels.

WeiqunZhang: We can still keep all the dts without printing levels no longer exist.

Issue #2936, opened 2022-03-07
picmi: initializing default WarpX data_list if none is provided


RTSandberg:
Error when trying to run the  laser_acceleration_PICMI.py from the PICMI GitHub examples link
```
 % python laser_acceleration_PICMI.py
Warning: WarpX does not support specifying the random number seed in PseudoRandomLayout
Traceback (most recent call last):
  File "/Users/ryansand/Documents/work/LBNL_postdoc_2022_2023/projects/02_warpx_rz/00_setup/00_picmi/laser_acceleration_PICMI.py", line 208, in <module>
    sim.write_input_file(file_name='input_script')
  File "/opt/anaconda3/envs/warpx-dev/lib/python3.10/site-packages/pywarpx/picmi.py", line 1076, in write_input_file
    self.initialize_inputs()
  File "/opt/anaconda3/envs/warpx-dev/lib/python3.10/site-packages/pywarpx/picmi.py", line 1063, in initialize_inputs
    diagnostic.initialize_inputs()
  File "/opt/anaconda3/envs/warpx-dev/lib/python3.10/site-packages/pywarpx/picmi.py", line 1145, in initialize_inputs
    for dataname in self.data_list:
TypeError: 'NoneType' object is not iterable
```

It appears that PICMI was developed presuming that codes supporting PICMI would provide automatic filling-in of data_list if not provided in the PICMI file:
From https://github.com/picmi-standard/picmi/tree/master/PICMI_Python/diagnostics.py :

```py
class PICMI_FieldDiagnostic(_ClassWithInit):
    """
    Defines the electromagnetic field diagnostics in the simulation frame
      - grid: Grid object for the diagnostic
      - period: Period of time steps that the diagnostic is performed
      - **data_list=None: List of quantities to write out. Possible values 'rho', 'E', 'B', 'J', 'Ex' etc.
                        Defaults to the output list of the implementing code.**
      - write_dir='.': Directory where data is to be written
      - step_min=None: Minimum step at which diagnostics could be written (optional)
                       Defaults to step 0.
      - step_max=None: Maximum step at which diagnostics could be written (optional)
                       Defaults to no limit.
      - number_of_cells=None: Number of cells in each dimension (optional)
                              If not given, will be obtained from grid.
      - lower_bound=None: Lower corner of diagnostics box in each direction (optional)
                          If not given, will be obtained from grid.
      - upper_bound=None: Higher corner of diagnostics box in each direction (optional)
[...]

PR #2949, opened 2022-03-09
allow diagnostic data_list to be None

RTSandberg:
Address issue #2936 by allowing picmi file not to specify diagnostic data lists, in which case data_list is set to None.
Now the WarpX picmi file now does not try to initialize from data_list is data_list is None, 
but rather initializes WarpX defaults


dpgrote (approved):
This looks good. Thanks for implementing!

PR #2950, opened 2022-03-09
CMake: pip_install w/o MPI

ax3l:
- Respect `PYINSTALLOPTIONS` also for `requirements.txt`.

- We automatically trigger a dependency to `mpi4py` when MPI is used.
  This makes sure this requirement stays consistent.


dpgrote (approved):
This sounds like a good thing to

PR #2951, opened 2022-03-09
RZ: Use WarpX::ncomps more consistently

ax3l:
We used `2*modes - 1` in many places, which is harder to follow then the variable we have.

### DIFF
--- a/Source/FieldSolver/SpectralSolver/SpectralFieldDataRZ.cpp
+++ b/Source/FieldSolver/SpectralSolver/SpectralFieldDataRZ.cpp
@@ -444,9 +444,9 @@ SpectralFieldDataRZ::ForwardTransform (const int lev,
         if ( !(field_mf[mfi].box().contains(field_mf_copy[mfi].box())) ) {
             // If field_mf[mfi] is smaller than field_mf_copy[mfi], then fill field_mf_copy[mfi] with
             // zeros so that all of it is initialized.
-            field_mf_copy[mfi].setVal<amrex::RunOn::Device>(0._rt, realspace_bx, 0, ncomp);
+            field_mf_copy[mfi].setVal<amrex::RunOn::Device>(0._rt, realspace_bx, 0, m_ncomps);
             }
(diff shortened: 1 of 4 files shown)


Source/FieldSolver/SpectralSolver/SpectralFieldDataRZ.cpp
```diff
@@ -506,8 +505,8 @@ SpectralFieldDataRZ::ForwardTransform (const int lev,
         field_mf_t_copy[mfi].copy<amrex::RunOn::Device>(field_mf_t[mfi], 0, 0, 1); // Real part of mode 0
         field_mf_r_copy[mfi].setVal<amrex::RunOn::Device>(0._rt, realspace_bx, 1, 1); // Imaginary part of mode 0
         field_mf_t_copy[mfi].setVal<amrex::RunOn::Device>(0._rt, realspace_bx, 1, 1); // Imaginary part of mode 0
-        field_mf_r_copy[mfi].copy<amrex::RunOn::Device>(field_mf_r[mfi], 1, 2, 2*n_rz_azimuthal_modes-2);
-        field_mf_t_copy[mfi].copy<amrex::RunOn::Device>(field_mf_t[mfi], 1, 2, 2*n_rz_azimuthal_modes-2);
+        field_mf_r_copy[mfi].copy<amrex::RunOn::Device>(field_mf_r[mfi], 1, 2, 2*n_rz_azimuthal_modes-2); // why -2 not -1 or 0?
+        field_mf_t_copy[mfi].copy<amrex::RunOn::Device>(field_mf_t[mfi], 1, 2, 2*n_rz_azimuthal_modes-2); // why -2 not -1 or 0?
```
dpgrote: The multifab `field_mf_r_copy` includes the imaginary part of mode 0 (even though is it all zeros) and so has a total of `2*n_rz_azimuthal_modes` components. The lines above handle mode 0. These lines handle the rest of the modes, so the number of components copied is `2*(n_rz_azimuthal_modes - 1)`.

dpgrote (approved):
This looks good. Thanks for the clean up.

RTSandberg (commented):
Two additional changes to make in EvolveF.cpp:
* comments in lines 174, 175

PR #2952, opened 2022-03-09
Use uniform error message formatting for all the amrex::Abort() calls in WarpX.cpp

lucafedeli88:
This PR replaces most of the calls to `amrex::Abort()` in `WarpX.cpp` with the new macro `WARPX_ALWAYS_ASSERT_WITH_MESSAGE`. It also adds `Utils::TextMsg::Err` to the few residual calls to `amrex::Abort()`


Source/WarpX.cpp
```diff
@@ -1176,52 +1178,57 @@ WarpX::ReadParameters ()
         // Check whether the default comoving velocity should be used
         bool use_default_v_comoving = false;
         pp_psatd.query("use_default_v_comoving", use_default_v_comoving);
+
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+            !use_default_v_comoving || !boosted_frame,
```
ax3l: Should this read
```suggestion
            !use_default_v_comoving || boosted_frame,
```
?

lucafedeli88: yes, sorry! You are right!!

Source/WarpX.cpp
```diff
@@ -836,17 +837,15 @@ WarpX::ReadParameters ()
...
-                amrex::Abort(ss.str());
-            }
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                do_pml_dive_cleaning == do_pml_divb_cleaning,
+                "warpx.do_pml_dive_cleaning = "
+                + std::to_string(do_pml_dive_cleaning)
+                +" and warpx.do_pml_divb_cleaning = "
+                +std::to_string(do_pml_divb_cleaning)
+                +": this case is not implemented yet,"
+                +" please set both parameters to the same value"
```
ax3l: Readability:
```suggestion
                +" and warpx.do_pml_divb_cleaning = "
                + std::to_string(do_pml_divb_cleaning)
                + ": this case is not implemented yet,"
                + " please set both parameters to the same value"
```

Source/WarpX.cpp
```diff
@@ -993,25 +989,23 @@ WarpX::ReadParameters ()
...
-                amrex::Abort("\nalgo.particle_shape must be set in the input file:"
-                             "\nplease set algo.particle_shape to 1, 2, or 3");
+            if (queryWithParser(pp_algo, "particle_shape", particle_shape)){
+
+                WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                    (particle_shape >= 1) && (particle_shape <=3),
+                    "algo.particle_shape can be only 1, 2, or 3"
+                );
+
+                nox = particle_shape;
```
ax3l, line 1081: At some point we should rebase out these three to a single control variable, since we do not support non-isotropic shapes anymore.

Source/WarpX.cpp
```diff
@@ -1176,52 +1178,57 @@ WarpX::ReadParameters ()
...
+            [](const auto& val){return val == 0.;});
+
+        const auto v_comoving_is_zero =
+            std::all_of(m_v_comoving.begin(), m_v_comoving.end(),
+            [](const auto& val){return val == 0.;});
+
+
+        // Galilean and comoving algorithms should not be used together
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+            v_galilean_is_zero || v_comoving_is_zero,
```
ax3l, line 1285: According to the message, I think this needs to be an `xor` not an `or`.
```suggestion
            v_galilean_is_zero ^ v_comoving_is_zero,
```

Source/WarpX.cpp
```diff
@@ -1176,52 +1178,57 @@ WarpX::ReadParameters ()
...
+        );
 
         // The comoving PSATD algorithm is not implemented nor tested with Esirkepov current deposition
-        if (current_deposition_algo == CurrentDepositionAlgo::Esirkepov) {
-            if (m_v_comoving[0] != 0. || m_v_comoving[1] != 0. || m_v_comoving[2] != 0.) {
-                amrex::Abort("Esirkepov current deposition cannot be used with the comoving PSATD algorithm");
-            }
-        }
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+            (current_deposition_algo != CurrentDepositionAlgo::Esirkepov) ||
```
ax3l: I think this also needs to be an `xor`
```suggestion
            (current_deposition_algo == CurrentDepositionAlgo::Esirkepov) ^
```

Source/WarpX.cpp
```diff
@@ -1176,52 +1178,57 @@ WarpX::ReadParameters ()
...
 
-        if (current_deposition_algo == CurrentDepositionAlgo::Vay) {
-            if (m_v_galilean[0] != 0. || m_v_galilean[1] != 0. || m_v_galilean[2] != 0.) {
-                amrex::Abort("Vay current deposition not implemented for Galilean algorithms");
-            }
-        }
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+            (current_deposition_algo != CurrentDepositionAlgo::Vay) ||
+            v_galilean_is_zero,
+            "Vay current deposition not implemented for Galilean algorithms"
```
ax3l, line 1303: If `current_deposition_algo` is Vay and `v_galilean_is_zero` is True this will not fail. I think it should read:
```suggestion
            (current_deposition_algo == CurrentDepositionAlgo::Vay) ^
            v_galilean_is_zero,
            "Vay current deposition not implemented for Galilean algorithms"
```

Source/WarpX.cpp
```diff
@@ -1176,52 +1178,57 @@ WarpX::ReadParameters ()
...
-        if (current_correction) {
-            if (m_v_galilean[0] != 0. || m_v_galilean[1] != 0. || m_v_galilean[2] != 0.) {
-                if (fft_do_time_averaging) {
-                    amrex::Abort("Current correction not implemented for averaged Galilean algorithm");
-                }
-            }
-        }
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+            (!current_correction) || v_galilean_is_zero || (!fft_do_time_averaging),
+            "Current correction not implemented for averaged Galilean algorithm"
```
ax3l, line 1308: I don't think this assert is correct yet, but too late rn to get the triple if right :sweat_smile: 

memo to myself: check from here to end tomorrow

Source/WarpX.cpp
lucafedeli88: I agree!

Source/WarpX.cpp
lucafedeli88: Uhm... I am not actually sure about this... the nested `if` in the original code is equivalent to:
`if ( !v_galilean_is_zero && !v_comoving_is_zero)`, so the negation of that condition should be:
`v_galilean_is_zero || v_comoving_is_zero` . Moreover, if they are both zero, it means that you are not using neither the galilean algorithm nor the comoving algorithm, which should be fine if I am not mistaken...

Source/WarpX.cpp
lucafedeli88: Uhm, I am not sure.. not (A and B) shouldn't be (not A) or (not B) ? 

Here, we should abort only when Esirkepov is used together with the co-moving algorithm.

Using `assert ( (dep != Esirkepov) || (v_comoving==0) )` the truth  table is:

| deposition   | v_comoving | result |
|--------------|------------|--------|
| Esirkepov    | 0          | OK     |
| Esirkepov    | != 0       | ABORT  |
| != Esirkepov | 0          | OK     |
| != Esirkepov | != 0       | OK     |

while using `assert ( (dep == Esirkepov) XOR (v_comoving==0) )` the truth  table is:

| deposition   | v_comoving | result |
|--------------|------------|--------|
| Esirkepov    | 0          | ABORT |
| Esirkepov    | != 0       | OK  |
| != Esirkepov | 0          | OK     |
| != Esirkepov | != 0       | ABORT     |

which is not what we want...

Source/WarpX.cpp
lucafedeli88: I think that this is what the previous code did, actually.
With 
```
  if (current_deposition_algo == CurrentDepositionAlgo::Vay) {
            if (m_v_galilean[0] != 0. || m_v_galilean[1] != 0. || m_v_galilean[2] != 0.) {
                //ABORT
            }
}
```
if `current_deposition_algo` is `Vay` the first `if` clause is True. But if `v_galilean_is_zero` is also True, this means that  all the components of `m_v_galilean` are zero. Therefore the second `if` clause would be False and we would **not** abort. 
This is exactly what the assert with `(current_deposition_algo != CurrentDepositionAlgo::Vay) || v_galilean_is_zero` does, I think...

Source/WarpX.cpp
lucafedeli88: I actually think it is! `if (A and B and C) {ABORT}` means that it is enough that either not(A) or not(B) or not(C) are True to prevent the abort!

Source/WarpX.cpp
ax3l: You are right, I must have overlooked the change to `_is_zero` variables last night here.

ax3l:
Sorry, little merge conflict in `Source/WarpX.cpp` to be resolved @lucafedeli88

lucafedeli88:
> Sorry, little merge conflict in `Source/WarpX.cpp` to be resolved @lucafedeli88

The conflict should be fixed now!

Source/WarpX.cpp
```diff
@@ -1256,56 +1252,62 @@ WarpX::ReadParameters ()
...
+                "Esirkepov current deposition cannot be used with the comoving PSATD algorithm");
 
-        if (current_correction) {
-            if (m_v_galilean[0] != 0. || m_v_galilean[1] != 0. || m_v_galilean[2] != 0.) {
-                if (fft_do_time_averaging) {
-                    amrex::Abort("Current correction not implemented for averaged Galilean algorithm");
-                }
-            }
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(v_galilean_is_zero,
+                "Esirkepov current deposition cannot be used with the Galilean algorithm."");
```
lucafedeli88:
```suggestion
                "Esirkepov current deposition cannot be used with the Galilean algorithm.");
```

ax3l (approved):

Issue #2821, opened 2022-02-02
Refactor out Distutils


ax3l:
[distutils is deprecated](https://www.python.org/dev/peps/pep-0632/) (PEP 632).

We need to refactor it out for newer modules such as:
- setuptools (already used, use more functions from it)
- [packaging](https://github.com/pypa/packaging)
- [build](https://github.com/pypa/build)

Upstream example from pybind11:
- https://github.com/pybind/cmake_example/blob/master/setup.py

ax3l:
Python 3.12 seems to remove it for good, so we gotta update a couple of Python `setup.py` scripts soon.

ax3l:
Migration guide:
https://peps.python.org/pep-0632/#migration-advice

PR #2954, opened 2022-03-10
Clean Pip Staging Directory

ax3l:
The distutils staging (`build` -> `_tmppythonbuild`) directory that pip uses to collect build artifacts is not by default cleaned between multiple `pip` runs.

This can be confusing when we recompile, because old `libwarpx*` files can be still in it that are not present in our own `build/lib/`.

This cleans that staging directory before builds now. It also sets the whole `build_base` so that no artifact lands in the default, which was `build/`. So far, `bdist.linux-x86_64/` still was out-of-tree.

First reported by @RTSandberg :pray:

### DIFF
--- a/Regression/WarpX-tests.ini
+++ b/Regression/WarpX-tests.ini
@@ -2946,7 +2946,7 @@ runtime_params =
 customRunCmd = python3 PICMI_inputs_2d.py
 dim = 2
 addToCompileString = USE_PYTHON_MAIN=TRUE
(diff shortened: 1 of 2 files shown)


dpgrote (approved):
This looks good (though I haven't tested it).
A comment is that distutils is being deprecated - is there a `clean` tool in setuptools?

ax3l:
Thanks!

Yes, we need to refactor that out at once at some point #2821
I am not sure yet, but I think yes, they duplicate/modernize that functionality :crossed_fingers:

PR #2955, opened 2022-03-10
Fix ambiguity in reference to Parser

RevathiJambunathan:
In assigning external fields to particles,  we have a Parser type.
The compiler sometimes triggers an error with a message that the `reference to "Parser" is ambiguous`

This PR fixes it by explicitly defining the enumerator `ExternalFieldInitType::Parser` to reference to the Parser-type.

Thanks to @atmyers for the offline discussion.


atmyers (approved):

ax3l (approved):
Nice, thanks a lot!

PR #2957, opened 2022-03-10
FAQ: Forward Declarations

ax3l:
Adds a new FAQ entry with links & background/rationale for C++ forward declarations.

### DIFF
--- a/Docs/source/developers/faq.rst
+++ b/Docs/source/developers/faq.rst
@@ -54,6 +54,16 @@ Generally speaking, C++ file endings can be arbitrary, we just keep them consist
 To be explicit and avoid confusion (with C/ObjC), we might change them all to ``.hpp`` and ``.cpp``/``.cxx`` at some point, but for now ``.H`` and ``.cpp`` is what we do (as in AMReX).
 
 
+What are ``#include "..._fwd.H"`` and ``#include <...Fwd.H>`` files?
+--------------------------------------------------------------------
+
+These are `C++ forward declarations <https://en.wikipedia.org/wiki/Forward_declaration>`__.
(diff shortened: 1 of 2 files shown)


Docs/source/developers/faq.rst
```diff
@@ -54,6 +54,16 @@ Generally speaking, C++ file endings can be arbitrary, we just keep them consist
...
 
 
+What are ``#include "..._fwd.H"`` and ``#include <...Fwd.H>`` files?
+--------------------------------------------------------------------
+
+These are `C++ forward declarations <https://en.wikipedia.org/wiki/Forward_declaration>`__.
+In C++, ``#include`` statements copy the referenced header file *literally* into place, which can increase the compile time of a ``.cpp`` file to an object file significantly, especially with transitive header files including each other.
+
+In order to reduce compile time, we define :ref:`forward declarations in WarpX and AMReX <developers-cpp-includes-fwd>` for commonly used, large classes.
+The C++ standard library also uses that concept, e.g., in `iosfwd <https://en.cppreference.com/w/cpp/header/iosfwd>`__.
```
RemiLehe, line 64: Could you add some more details about **what** needs to be added in the `fwd` file? (My understanding is that you need to copy a subset of the code from the `.H` code, but it is unclear to me exactly what needs to be copied.)

Also: could you specify in which cases we can include the forward declaration, and in which cases you need to include the full `.H` file?

ax3l: Oh, both of those questions are explained in detail in the reference, which I link as `forward declarations in WarpX and AMReX <developers-cpp-includes-fwd>` :)
That's part of the regular reference in `Docs/source/developers/repo_organization.rst`

The FAQ list currently is only a 1-2 paragraph answer followed by/including a link to full explanations/reference.

Can you check if the also herein changed file `Docs/source/developers/repo_organization.rst` answers these questions in an understandable manner?

RemiLehe (approved):

PR #2959, opened 2022-03-11
WarpX/PICSAR Version Define: File

ax3l:
Move the `-D` defines that contain the latest WarpX commit to a config file. Using always the latest commit counteracts `ccache` since the signature of the compilation call changes every commit.

- [x] CMake
- [x] GNUmake


WeiqunZhang:
The change to gnu make depends on https://github.com/AMReX-Codes/amrex/pull/2653.

ax3l:
Thank you @WeiqunZhang, updated accordingly :)

WeiqunZhang (approved):

PR #2960, opened 2022-03-11
Langmuir 2D Tests: Better Analysis Plots

EZoni:
While looking at some of the Langmuir 2D tests, I noticed that the plots that we produce with our analysis script could be slightly improved (see comparison below for the test `Langmuir_multi_2d_psatd_Vay_deposition`; notable differences: size of colorbar, axis labels, axis ticks (physical extent instead of cell number)).

Old plot:
<p align="center">
    
</p>

New plot:
<p align="center">
    
</p>

### DIFF
--- a/Examples/Tests/Langmuir/analysis_langmuir_multi_2d.py
+++ b/Examples/Tests/Langmuir/analysis_langmuir_multi_2d.py
@@ -17,7 +17,6 @@ import os
 import re
 import sys
 
-import matplotlib
 import matplotlib.pyplot as plt
 from mpl_toolkits.axes_grid1.axes_divider import make_axes_locatable
 import yt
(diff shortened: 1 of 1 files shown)


Examples/Tests/Langmuir/analysis_langmuir_multi_2d.py
```diff
@@ -18,9 +18,8 @@
 import sys
 
 import matplotlib
```
EZoni: **ltgm-com** says import of 'matplotlib' is not used:
```suggestion
```

ax3l (approved):
LGTM :)

Not sure why we hard-code to use the `Agg` backend everywhere, though.

dpgrote (approved):
This looks good! Thanks - I had considered making similar changes but never got to it.

Something I had considered was adding a third plot, showing `E_sim - E_th` since the individual plots are visually indistinguishable.

PR #2963, opened 2022-03-12
Load-balancing heuristic weights: not negative by default

RemiLehe:
If the user uses the heuristic load-balancing cost, passes only e.g. the weight on the cells (and omits the weight on the particles like this):
```
algo.load_balance_costs_update = heuristic
algo.costs_heuristic_cells_wt = 0.4
# Users says nothing about `algo.costs_heuristic_particles_wt`
```
then the current WarpX default is that the particles are counted **negatively** (!!).
In other words, the load-balancing algorithm will try to stack boxes that have a lot of particles together on the same rank, while spreading out boxes that do not have any particle across the other ranks. (I did encounter this in practice!)

This PR avoids this issue by setting the default weights to 0.

### DIFF
--- a/Source/WarpX.H
+++ b/Source/WarpX.H
@@ -1260,13 +1260,13 @@ private:
      * The problem setup for these tests is an empty (i.e. no particles) domain
      * of size 256 by 256 by 256 cells, from which the average time per iteration
      * per cell is computed. */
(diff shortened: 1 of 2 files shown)


ax3l (approved):
Haha, that's a good bug :)

Glad you caught it!

PR #2958, opened 2022-03-10
Vay Deposition: Add Cumulative Sums

EZoni:
Merge first:
- [x] #2937
- [x] #2961
- [x] #2962
- [x] #2965
- [x] #2991
- [x] #3012

To-do:
- [x] Fix GPU build bugs
- [x] Fix 2D Langmuir tests
- [x] Fix 3D Langmuir tests
- [x] Clean up names (e.g., rename `jx` as `Dx` where needed)
- [x] More testing on NCI with uniform plasmas
- [x] Add CI test(s) for two-particle case

Here's an example of results without (left) v. with (right) cumulative sums in 3D ([inputs_3d.txt](https://github.com/ECP-WarpX/WarpX/files/8773148/inputs_3d.txt)):
<p align="center">
    
    
</p>

PR #2965, opened 2022-03-14
Vay Deposition: Separate Arrays, Correct Index Types w/ FFTs

EZoni:
With the Vay deposition, we now deposit the preliminary current D in separate MultiFabs. Before this PR, D was deposited in the same MultiFab as the actual current J, while D had to have a different index type (always matching that of the charge rho), which had to be passed by hand to the functions performing FFTs. This was going to cause issues if those functions were to be called multiple times (e.g. when the Vay deposition is completed with the calculation of the cumulative sums implemented in #2958).

The main changes in this PR are:
- New MultiFabs `current_fp_vay` with index type given by that of rho;
- Abort when Vay deposition is combined with either current centering or mesh refinement (implementation not yet completed or verified);
- New interfaces for the functions `AddCurrentFromFineLevelandSumBoundary`, `ApplyFilterandSumBoundaryJ` and `NodalSyncJ` (now taking as input arguments the MultiFabs storing the currents on which they perform their operations, instead of having those MultiFabs hard coded inside the functions themselves);
- New interfaces for the functions `PSATDForwardTransformJ` (now taking as input arguments the MultiFabs storing the currents on which they perform their operations, instead of having those MultiFabs hard coded inside the functions themselves), as well as `ForwardTransformVect` and `ForwardTransform` (not taking anymore as input arguments the index type, or staggering, of the current MultiFabs, thus all unused overloads removed).


RemiLehe:
Thanks a lot for this PR!
The changes make the code a lot cleaner.

Btw, in a separate PR, would you mind making the same changes to the function for `rho`, i.e. allowing the user to pass a `MultiFab` for `RestrictRhoFromFineToCoarsePatch`, `ApplyFilterandSumBoundaryRho`, `AddRhoFromFineLevelandSumBoundary` and `NodalSyncRho`? 
There are a few future WarpX features for which this would be useful...

RemiLehe (approved):

Issue #2746, opened 2022-01-14
Backtransformed diagnostics with mesh refinement simulation


prkkumar:
Backtransformed diagnostics with mesh refinement simulations segfaults with the following Backtrace:

```
=== If no file names and line numbers are shown below, one can run
            addr2line -Cpfie my_exefile my_line_address
    to convert `my_line_address` (e.g., 0x4a6b) into file name and line number.
    Or one can use amrex/Tools/Backtrace/parse_bt.py.

=== Please note that the line number reported by addr2line may not be accurate.
    One can use
            readelf -wl my_exefile | grep my_line_address'
    to find out the offset for that line.

 0: ./warpx.3d.MPI.OMP.DP.OPMD.PSATD.QED(+0x50e1e5) [0x55ab69d631e5]
    ?? ??:0

 1: ./warpx.3d.MPI.OMP.DP.OPMD.PSATD.QED(+0x50ff8d) [0x55ab69d64f8d]
    ?? ??:0

 2: /lib/x86_64-linux-gnu/libc.so.6(+0x46210) [0x7fe2cea57210]

 3: /lib/x86_64-linux-gnu/libc.so.6(gsignal+0xcb) [0x7fe2cea5718b]

 4: /lib/x86_64-linux-gnu/libc.so.6(abort+0x12b) [0x7fe2cea36859]

 5: /lib/x86_64-linux-gnu/libstdc++.so.6(+0x9e911) [0x7fe2cee0b911]
    ?? ??:0

 6: /lib/x86_64-linux-gnu/libstdc++.so.6(+0xaa38c) [0x7fe2cee1738c]
    ?? ??:0

 7: /lib/x86_64-linux-gnu/libstdc++.so.6(+0xaa3f7) [0x7fe2cee173f7]
    ?? ??:0

 8: /lib/x86_64-linux-gnu/libstdc++.so.6(+0xaa6a9) [0x7fe2cee176a9]
    ?? ??:0

 9: /lib/x86_64-linux-gnu/libstdc++.so.6(_ZSt19__throw_logic_errorPKc+0x41) [0x7fe2cee0e21e]
    ?? ??:0

10: ./warpx.3d.MPI.OMP.DP.OPMD.PSATD.QED(+0x100f2b) [0x55ab69955f2b]

11: ./warpx.3d.MPI.OMP.DP.OPMD.PSATD.QED(+0x10b991) [0x55ab69960991]
    ?? ??:0

[...]

PR #2967, opened 2022-03-14
Bug Fix : BTD + MR, initialize output buffer only for the level in the output

RevathiJambunathan:
When visualizing BTD output from simulations run with MR, the results looked skewed (see figure below)
Screenshot from 2022-03-14 15-09-44

BTD output is only generated for the coarse grid, i.e., the mother grid.
When running with more than one level, the extent of the buffer multifab, (the multifab that stores the back-transformed data), is updated with level 1 grid, which is inconsistent since the buffer multifab is only generated for the coarsest level in the simulation.

This PR fixes this issue by ensuring that the buffer data is initialized only for the levels in the output, which for BTD is 1.
The input file used is 
[inputs_2d.txt](https://github.com/ECP-WarpX/WarpX/files/8248827/inputs_2d.txt)
This should fix one of the errors observed in Issue #2746. Other bug-fixes for this Issue to follow

On a related note. 
In this line (link below ), I had initialized the nlev_output to be 1 and don't update it (on purpose because we want there to be only one level for BTD output, at-least for now)
Any suggestions on a better approach? if yes, let me know, and I will add that as a separate mini-PR
https://github.com/ECP-WarpX/WarpX/blob/f3fecb3324ed42be8985c0ff495597f70d21f317/Source/Diagnostics/BTDiagnostics.cpp#L61

### DIFF
--- a/Source/Diagnostics/Diagnostics.cpp
+++ b/Source/Diagnostics/Diagnostics.cpp
@@ -261,9 +261,17 @@ Diagnostics::InitData ()
     // loop over all buffers
     for (int i_buffer = 0; i_buffer < m_num_buffers; ++i_buffer) {
         // loop over all levels
+        // This includes full diagnostics and BTD as well as cell-center functors for BTD.
+        // Note that the cell-centered data for BTD is computed for all levels and hence
+        // the corresponding functor is also initialized for all the levels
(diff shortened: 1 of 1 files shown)


ax3l (approved):
Thank you, LGTM :)

PR #2968, opened 2022-03-16
ABLASTR: Full Library

ax3l:
Build ABLASTR as a full library instead as of previously an object collection. This makes linking across project boundaries a bit easier as well as declarations of `PRIVATE` vs. `PUBLIC` dependencies and their propagation. Also will allow to pre-build and install it independently in the future.

This becomes necessary as we start to share more than header files.


EZoni (approved):
Looks good to me!

PR #2970, opened 2022-03-16
openPMD: Only Write up to Finest Level

### LINKED ISSUE #1212: Problem to open bp file when an attribute is removed at a certain timestep
Good morning,
I have a problem opening a file with the backend bp when I removed attributes at a certain timestep (E_lvl1, B_lvl1...). Indeed, from this timestep i cannot open the following files, this problem happens only with bp backend. I try to use h5 backend and there is no more problem.

To obtain this error, i just open the files with:
```py
import openpmd_api as io
series = io.Series("diags/diag1h5/openpmd_00200.bp", io.Access.read_only)
```

and I obtain this error :
image

I give two files for each backend(h5,bp), one step before the removal and at the step of the removal : 
https://www.transfernow.net/dl/20220301qicEOJPJ

Current work-around: split series manually for read back (before and after these records/variables change).

**Software Environment**
 - version of openPMD-api: 0.14.4
 - installed openPMD-api via: pip
 - name and version of Python implementation: Python 3.9
 - version of HDF5: 1.12.0
 - version of ADIOS2:2.7.1
 - name and version of MPI:

ax3l:
With #2948, now that we support "dynamic" levels, we need to differentiate between the maximum allowed refinement level (maxLevel) and the currently maximum used level (finestLevel) of a simulation.

This should fix https://github.com/openPMD/openPMD-api/issues/1212

### DIFF
--- a/Source/Diagnostics/FlushFormats/FlushFormatOpenPMD.H
+++ b/Source/Diagnostics/FlushFormats/FlushFormatOpenPMD.H
@@ -32,7 +32,7 @@ public:
         const amrex::Vector<amrex::MultiFab>& mf,
         amrex::Vector<amrex::Geometry>& geom,
         const amrex::Vector<int> iteration, const double time,
-        const amrex::Vector<ParticleDiag>& particle_diags, int /*nlev*/,
+        const amrex::Vector<ParticleDiag>& particle_diags, int output_levels,
         const std::string prefix, int file_min_digits,
         bool plot_raw_fields,
         bool plot_raw_fields_guards,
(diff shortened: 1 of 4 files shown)


guj (approved):
Thanks for adding this.  So finest levels is a newer addition in the input file? 
Is it necessary to assert  that finest level  is not exceeding max level, just to be on the safe side?

ax3l:
> Thanks for adding this. So finest levels is a newer addition in the input file?

So far we only used the max_level and knew it is always also truly used. AMReX supports that user can say "create up to maxLevel levels, but there can be dynamically be used less". We start to add such features with #2948 in an early stage.

> Is it necessary to assert that finest level is not exceeding max level, just to be on the safe side?

No need, AMReX is checking/maintaining that constrain.

atmyers (approved):

ax3l:
oh, GitHub actions had an incidence today. restarting CI

lucafedeli88 (approved):
Looks good to me! Thanks!

ax3l:
Thanks! Please let me know once you rebase #2948 if that fixed the issue you saw :)

PR #2972, opened 2022-03-16
[WIP] Introduce graph measurements

RemiLehe:
Posted for development purposes.

### DIFF
--- a/Source/BoundaryConditions/PML.cpp
+++ b/Source/BoundaryConditions/PML.cpp
@@ -1185,6 +1185,12 @@ PML::Exchange (MultiFab& pml, MultiFab& reg, const Geometry& geom,
         MultiFab::Add(totpmlmf,pml,2,0,1,0); // Sum the third split component
     }
 
+    // record metrics of this comm op in a graph before and after load balance steps
+    //auto & warpx = WarpX::GetInstance();
+    //auto const & lb_intervals = warpx.load_balance_intervals;
+    //auto const cur_step = warpx.istep[0];
+    //auto & graph = warpx.graph;
+
     // Copy from the sum of PML split field to valid cells of regular grid
     if (do_pml_in_domain){
         // Valid cells of the PML and of the regular grid overlap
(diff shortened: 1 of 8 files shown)


Source/BoundaryConditions/PML.cpp
```diff
@@ -42,6 +42,7 @@
 #include <AMReX_RealVect.H>
 #include <AMReX_SPACE.H>
 #include <AMReX_VisMF.H>
+#include <AMReX_Graph.H> // kngott\graphviz
```
ax3l:
```suggestion
#include <AMReX_Graph.H> // kngott/graphviz
```

Source/BoundaryConditions/PML.cpp
```diff
@@ -1183,6 +1184,10 @@ PML::Exchange (MultiFab& pml, MultiFab& reg, const Geometry& geom,
         if (ngr.max() > 0) {
             MultiFab::Copy(tmpregmf, reg, 0, 0, 1, ngr);
             WarpXCommUtil::ParallelCopy(tmpregmf, totpmlmf, 0, 0, 1, IntVect(0), ngr, period);
+            amrex::Graph graph;
```
ax3l: We could make this a member var maybe and then print the table later on

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -57,6 +57,7 @@ using namespace amrex;
 void
 WarpX::Evolve (int numsteps)
 {
+    WARPX_REGION("WarpX::Evolve()")
```
ax3l: Does not compile for me, probably should be:
```suggestion
    WARPX_PROFILE_REGION("WarpX::Evolve");
```
?

Source/Parallelization/WarpXRegrid.cpp
```diff
@@ -128,6 +124,77 @@ WarpX::LoadBalance ()
...
+        graph.addFab(*costs[lev], name, sizeof(amrex::Real));
+        std::vector<double> costs_local(costs[lev]->size());
+        for (int n=0; n<costs[lev]->size(); ++n)
+        {
+            costs_local[n] = costs[lev]->data()[n];
+        }
+        double const scaling = 1.0;
+        bool const available_locally = false; // with respect to costs_local
+        graph.addNodeWeight("costs", "cost_value", costs_local, scaling, available_locally);
+    }
```
ax3l: @kngott let's double check the costs here

PR #2892, opened 2022-02-25
Add by-cell averages of particle properties

hklion:
Adds a new type of field to `FullDiagnostics`, implemented in `ParticleReductionFunctor`. The diagnostic calculates the by-cell average of a parser function that takes particle position and velocity as arguments. The average is calculated separately for each species specified by the user (`<diag_name>.particle_fields_species`).

One thing I'd particularly like some input on: When there are no particles in a cell, I currently set the average to zero. This is to avoid the zero-divide-by-zero that would otherwise occur. Is there a better way to handle this? I initially had it just do the operation, since a NaN is actually meaningful in this context, but that tripped the `fpe_trap_invalid` used in the test suite.

PR #2975, opened 2022-03-17
add openPMD test for particle_fields_diags to CI

hklion:
Test openPMD output in the particle_fields_diags tests (#2892).

This adds a second check to the particle_fields_diags CI tests that compares the field output in openPMD to the particle averages calculated from the plotfile particle data.

1) As written, the single precision test will fail on the checksum verification. It looks like the fact that I'm outputting an openPMD file in addition to the plotfile is changing the plotfile checksum, but only for the single precision test. Is this expected behavior? If so I'll go ahead an update the checksum for that test. Just wanted to make sure that this wasn't a sign of something weird :)

2) I couldn't figure out how to specify the openPMD file name except by doing it explicitly:
```py
opmd = io.Series('diags/openpmd/openpmd_%T.h5', io.Access.read_only)
```
vs deriving it from a command line argument like the plotfile. I'm not sure there's a better way of doing this (short of adding a separate openPMD test that doesn't also rely on a plotfile), but if there is, let me know.


Examples/Tests/particle_fields_diags/inputs
```diff
@@ -76,3 +76,14 @@ diag1.particle_fields_species = electrons protons photons
 diag1.particle_fields.z(x,y,z,ux,uy,uz) = z
 diag1.particle_fields.uz(x,y,z,ux,uy,uz) = uz
 diag1.particle_fields.zuz(x,y,z,ux,uy,uz) = z * uz
+
+openpmd.format = openpmd
```
ax3l, line 80: If we find ADIOS2, we will default to faster ADIOS `.bp` files.
Since the script required HDF5, we need to write this here to make it stable:
```suggestion
openpmd.format = openpmd
openpmd.openpmd_backend = h5
```

ax3l:
> As written, the single precision test will fail on the checksum verification. It looks like the fact that I'm outputting an openPMD file in addition to the plotfile is changing the plotfile checksum, but only for the single precision test. Is this expected behavior? If so I'll go ahead an update the checksum for that test. Just wanted to make sure that this wasn't a sign of something weird :)

No, that's unusual. Must be something wrong with that test already.
Feel free to set PR as ready and let's take a look at the failing test together.

ax3l:
> I couldn't figure out how to specify the openPMD file name except by doing it explicitly:
> 
> opmd = io.Series('diags/openpmd/openpmd_%T.h5', io.Access.read_only)
>
> vs deriving it from a command line argument like the plotfile. I'm not sure there's a better way of doing this (short of adding a separate openPMD test that doesn't also rely on a plotfile), but if there is, let me know.

Opening the whole series is correct, the "file" that we pass for plotfiles is the "last" time step we created - something we don't need for openPMD.

If you want to play with iteration-access, you can either access it by step number as you currently do or generically take the last iteration found in the series. For a CI test, accessing a concrete step is probably even better than what we currently do for plotfiles: passing a new file if we change steps is often hard to understand why/when it fails.

ax3l:
I think I have an idea on the issue. The `particle_fields_diags_single_precision` test is in single precision.
When we output momentum in both plotfiles and openPMD, we do once a (somewhat urgh) transformation of units from "proper velocity" (gamma * v) to momentum - then dump - then transform back.

That operation is only precise up to single precision machine precision, but the checksum is checking more significant digits.

```
...
ERROR: Benchmark and plotfile checksum have different value for key [photons,particle_momentum_z]
Benchmark: [photons,particle_momentum_z] 1.424658520698910e-18
Plotfile : [photons,particle_momentum_z] 1.424658534915974e-18
Absolute error: 1.42e-26
Relative error: 9.98e-09
...
```

I think the right thing to do is go into `Examples/Tests/particle_fields_diags/analysis_particle_diags_impl.py`, and add in `evaluate_checksum` an appropriate (relative) tolerance instead of the default.
https://github.com/ECP-WarpX/WarpX/blob/4f4bd027801394ea634e2929140138c2e7eef4f7/Examples/Tests/particle_fields_diags/analysis_particle_diags_single.py#L16
https://github.com/ECP-WarpX/WarpX/blob/4f4bd027801394ea634e2929140138c2e7eef4f7/Examples/Tests/particle_fields_diags/analysis_particle_diags_impl.py#L28
https://github.com/ECP-WarpX/WarpX/blob/4f4bd027801394ea634e2929140138c2e7eef4f7/Examples/Tests/particle_fields_diags/analysis_particle_diags_impl.py#L156
https://github.com/ECP-WarpX/WarpX/blob/4f4bd027801394ea634e2929140138c2e7eef4f7/Regression/Checksum/checksumAPI.py#L38-L39

hklion:
Thanks for tracking that down, @ax3l. I adjusted the tolerance on the single precision check, and the test is now passing.

ax3l:
Awesome, hooray!

ax3l (approved):
Thanks a lot, awesome! :)

PR #69, opened 2019-03-19
R zgeometry

dpgrote:
This implements the FDTD RZ mode 0, axisymmetric solver.

PR #2977, opened 2022-03-17
Update to be compatible with PICMI changes

dpgrote:
In PICMI, the PR #69 changed the internals of the grid classes, removing saving the individual attributes and requiring the implementation to use the vector arguments. This PR updates the WarpX implementation replacing its use of the separate arguments for the boundary conditions.

Note that the required PICMI version does not need to be updated since the changes made here are compatible with older versions.


EZoni (approved):
Thank you, Dave! Looks good to me.

PR #2878, opened 2022-02-18
openPMD RZ modes dumped in thetaMode

RTSandberg:
In RZ geometry, only cylindrical field components are allowed and Cartesian components give errors.  
E.g. `<diag_name>.fields_to_plot = Er` is allowed but not `<diag_name>.fields_to_plot = Ex`
In RZ, the default full diagnostics are `Er Et Ez Br Bt Bz jr jt jz`.

If `<diag_name>.format = openpmd`, then all diagnostics but `part_per_cell` and `part_per_grid` will output all theta modes.  Otherwise, all diagnostics default to a 2D slice at `theta=0.`

Changes that affect all full diagnostics (openPMD, plotfiles, Ascent, SENSEI):
- E/B/Jtheta are now called `Et`/`Bt`/`Jt`

PR #2979, opened 2022-03-17
[WIP] Azure: Compile with `-g`

ax3l:
Uses default compilation type (`Release`) which implies `-O3`, but adds `-g` debug symbols (just larger binaries) for more useful backtraces.

First seen in #2878.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -14,6 +14,7 @@ jobs:
     CEI_SUDO: 'sudo'
     CEI_TMP: '/tmp/cei'
     CMAKE_GENERATOR: 'Ninja'
+    CXXFLAGS: '-g'
     FFTW_HOME: '/usr'
     LAPACKPP_HOME: '/usr/local'
     OMP_NUM_THREADS: 1
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -82,7 +82,7 @@ jobs:
         openmpi-bin libopenmpi-dev \
         libfftw3-dev libfftw3-mpi-dev libhdf5-openmpi-dev pkg-config make \
         python3 python3-pandas python3-pip python3-venv python3-setuptools libblas-dev liblapack-dev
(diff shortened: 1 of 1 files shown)


EZoni (approved):
Awesome, looks good to me!

ax3l:
I think with `-g` we get problems with respect to cache size. I checked `du -hs build/` and a single WarpX build with debug symbols `-g` is about 3.2GB in size.

Without `-g`, a default `build/` is just 226 MB. In total, we can only create 10 GB of temporary storage in CI, which also includes all the dependencies and toolchains we need to install...

We can probably optimize a bit by removing `-g` from dependencies:
```console
$ du -hs build/lib/*
1.7M	build/lib/libablastr.a
305M	build/lib/libamrex.a
16K	build/lib/libbuildInfoapp.a
562M	build/lib/libopenPMD.a
```

For our CI, it's probably sufficient to have line numbers on WarpX backtraces and then skip them in AMReX, openPMD-api et al.

.azure-pipelines.yml
```diff
@@ -81,7 +82,7 @@ jobs:
         openmpi-bin libopenmpi-dev \
         libfftw3-dev libfftw3-mpi-dev libhdf5-openmpi-dev pkg-config make \
         python3 python3-pandas python3-pip python3-venv python3-setuptools libblas-dev liblapack-dev
-      ccache --set-config=max_size=10.0G
+      ccache --set-config=max_size=2.0G
```
ax3l: Single build with `-g` is 3.2 GB:
```suggestion
      ccache --set-config=max_size=5.0G
```

EZoni:
@ax3l Do you think we should try a `ccache` max. size between 5GB and 10GB here?

ax3l:
I think the issue we face is that the binary caches are not large enough for this.
We could retry this after the CTest transition, maybe we are lucky and are below 10G.

EZoni:
Implemented in #5443.

PR #1898, opened 2021-04-19
Particle Buffer for Backtransformed Diagnostics

RevathiJambunathan:
This PR implements the capability to transform particle-data from boosted-frame to lab-frame and visualize it using plotfiles and openpmd format using the new BTD interface.

Please note that the output below is not in SI to compare with old BTD. However, the code has been fixed to only output SI units similar to FullDiagnostics.
Also an Abort is added in the code when adios backend is used with openpmd format to write particles in lab-frame. A corresponding Issue is generated #2448 

Below is the 2D input file that was used to verify the simulations using the customized output from old BTD with that from new BTD using plotfile and openpmd format.
[inputs_2d_boost_testBTD.txt](https://github.com/ECP-WarpX/WarpX/files/7334559/inputs_2d_boost_testBTD.txt)

Output from old-BTD for 2nd snapshot
Ez_old_BTD_iteration2

Output from new-BTD with plotfile format

Ez_newBTD_plotfile_iteration2

Output from new BTD with openpmd format : 
I was unable to plot particles and fields in the same slice using openpmd-viewer , but they are plotted separately here
Screenshot from 2021-10-12 20-18-19
Screenshot from 2021-10-12 20-19-36
Screenshot from 2021-10-12 20-19-56

I compared the mean(uz) from old BTD and new BTD (plotfile) and found that both outputs agree exactly.
for completeness : mean(uz) for the driver particles in WarpX units from old and new BTD = 59605413002956.99

PR #2980, opened 2022-03-17
openPMD: Handle Zero Particles Well

### LINKED ISSUE #2900: Particle openPMD IO issues when species not present in the first dump
When using openPMD output, if no particles of a species is written at the first dump, the openPMD-viewer cannot read particle data in the later dumps, even if particles are actually dumped. This happens in a wakefield acceleration simulation with a moving window, where the box z dimension is originally (-100.e-6,0) and the plasma starts at 200.e-6 for some reasons.

A reproducer can be found below (using ADIOS2 here, but I think it's the same for HDF5). For a box with z = (-25.e-6,5.e-6), it runs:
- A simulation with plasma electrons starting **in** the box (at z=0.e-6)
- A simulation with plasma electrons starting **out** of the box (at z=10.e-6)

This is done with the following submission script (a few seconds, here on the Juwels Booster):
```sh
#!/bin/bash -l
#SBATCH -A plasmabbq
#SBATCH --partition=booster
#SBATCH --nodes=1
#SBATCH --ntasks-per-node=4
#SBATCH --gres=gpu:4
#SBATCH --time=00:10:00
#SBATCH --job-name=test

source ../profile.sh

EXEC=$HOME/src/warpx/build/bin/warpx.3d.MPI.CUDA.DP.OPMD

srun -n 4 --cpu_bind=sockets $EXEC inputs \
     xz_slice.file_prefix = diags_in/xz_slice \
     yz_slice.file_prefix = diags_in/yz_slice \
     beams.file_prefix = diags_in/beams \
     electrons.zmin = 0.e-6 \
    &> output_in.txt

srun -n 4 --cpu_bind=sockets $EXEC inputs \
     xz_slice.file_prefix = diags_out/xz_slice \
     yz_slice.file_prefix = diags_out/yz_slice \
     beams.file_prefix = diags_out/beams \
     electrons.zmin = 10.e-6 \
    &> output_out.txt
```

After 500 iterations, the box contains electrons in both cases, but only the first run can be properly read by the openPMD-viewer. This can be seen with the following Notebook:
```py
%matplotlib inline
from openpmd_viewer import OpenPMDTimeSeries

iteration = 500

# Works fine
ts_in = OpenPMDTimeSeries('diags_in/beams')
z, = ts_in.get_particle(species='electrons', iteration=iteration, var_list = ['z'])
print(z.shape)

# Gives warnings and errors
ts_out = OpenPMDTimeSeries('diags_out/beams')
z, = ts_out.get_particle(species='electrons', iteration=iteration, var_list = ['z'])
print(z.shape)
```
In the second case, one gets warnings and errors like:
```
Warning: File 200 has different openPMD parameters than the rest of the time series.
OpenPMDException: No particle data in this time series
```
Here is the [input file](https://github.com/ECP-WarpX/WarpX/files/8160908/inputs.txt) used.

ax3l:
When a time step for output encounters zero particles in a species, then we still want to dump them as "empty" species (for a step) in openPMD. That simplifies post-processing a lot and we have the mechanisms in openPMD for it :)

This also enables ADIOS `.bp` support for BTD :)

Fix #2900
Partly a follow-up to #1898, but mostly a missing implementation for zero particle cases.

## To Do

- [x] finalize coding (dump: `makeEmpty()`)
- [x] testing
- [x] fix BTD CI issues
- [x] check if BTD with ADIOS now works, too

### DIFF
--- a/Source/Diagnostics/WarpXOpenPMD.cpp
+++ b/Source/Diagnostics/WarpXOpenPMD.cpp
@@ -682,16 +682,16 @@ WarpXOpenPMDPlot::DumpToFile (ParticleContainer* pc,
     openPMD::ParticleSpecies currSpecies = currIteration.particles[name];
 
     // prepare data structures the first time BTD has non-zero particles
-    //   we set some of them to zero extent, so we need to time that welll
+    //   we set some of them to zero extent, so we need to time that well
     bool const is_first_flush_with_particles = num_dump_particles > 0 && ParticleFlushOffset == 0;
     // BTD: we flush multiple times to the same lab step and thus need to resize
     //   our declared particle output sizes
     bool const is_resizing_flush = num_dump_particles > 0 && ParticleFlushOffset > 0;
     // write structure & declare particles in this (lab) step empty:
     //   if not BTD, then this is the only (and last) time we flush to this step
-    //   if BTD, then we do this multiple times
+    //   if BTD, then we may do this multiple times until it is the last BTD flush
     bool const is_last_flush_to_step = !isBTD || (isBTD && isLastBTDFlush);
-    // well, even in BTD we have to recognize that some lab stations have no
+    // well, even in BTD we have to recognize that some lab stations may have no
     //   particles - so we mark them empty at the end of station reconstruction
     bool const is_last_flush_and_never_particles =
             is_last_flush_to_step && num_dump_particles == 0 && ParticleFlushOffset == 0;
@@ -709,7 +709,7 @@ WarpXOpenPMDPlot::DumpToFile (ParticleContainer* pc,
         doParticleSetup = is_first_flush_with_particles || is_last_flush_and_never_particles;
 
     // this setup stage also implicitly calls "makeEmpty" if needed (i.e., is_last_flush_and_never_particles)
-    //   for BTD, we call this multiple times as we will resize in subsequent dumps
+    //   for BTD, we call this multiple times as we may resize in subsequent dumps if number of particles in the buffer > 0
     if (doParticleSetup || is_resizing_flush) {
         SetupPos(currSpecies, NewParticleVectorSize, isBTD);
         SetupRealProperties(pc, currSpecies, write_real_comp, real_comp_names, write_int_comp, int_comp_names,
--- a/Source/Diagnostics/WarpXOpenPMD.cpp
+++ b/Source/Diagnostics/WarpXOpenPMD.cpp
@@ -667,166 +667,192 @@ WarpXOpenPMDPlot::DumpToFile (ParticleContainer* pc,
                     amrex::ParticleReal const mass,
                     const bool isBTD,
                     const bool isLastBTDFlush,
-                    int ParticleFlushOffset)
-{
-  WARPX_ALWAYS_ASSERT_WITH_MESSAGE(m_Series != nullptr, "openPMD: series must be initialized");
+                    int ParticleFlushOffset) {
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(m_Series != nullptr, "openPMD: series must be initialized");
 
-  AMREX_ALWAYS_ASSERT(write_real_comp.size() == pc->NumRealComps());
-  AMREX_ALWAYS_ASSERT( write_int_comp.size() == pc->NumIntComps());
-  AMREX_ALWAYS_ASSERT(real_comp_names.size() == pc->NumRealComps());
-  AMREX_ALWAYS_ASSERT( int_comp_names.size() == pc->NumIntComps());
+    AMREX_ALWAYS_ASSERT(write_real_comp.size() == pc->NumRealComps());
+    AMREX_ALWAYS_ASSERT(write_int_comp.size() == pc->NumIntComps());
+    AMREX_ALWAYS_ASSERT(real_comp_names.size() == pc->NumRealComps());
+    AMREX_ALWAYS_ASSERT(int_comp_names.size() == pc->NumIntComps());
 
-  WarpXParticleCounter counter(pc);
-  if (counter.GetTotalNumParticles() == 0) return;
-  openPMD::Iteration currIteration = GetIteration(iteration, isBTD);
+    WarpXParticleCounter counter(pc);
+    auto const total_particles = counter.GetTotalNumParticles();
 
-  openPMD::ParticleSpecies currSpecies = currIteration.particles[name];
-  // meta data for ED-PIC extension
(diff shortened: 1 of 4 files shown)


Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -662,161 +664,195 @@ WarpXOpenPMDPlot::DumpToFile (ParticleContainer* pc,
...
+    // define positions & offset structure
+    const unsigned long long NewParticleVectorSize = total_particles + ParticleFlushOffset;
+    // we will set up empty particles unless it's BTD, where we might add some in a following buffer dump
+    //   during this setup, we mark some particle properties as constant and potentially zero-sized
+    m_doParticleSetUp = true;
+    if (isBTD)
+        m_doParticleSetUp = is_first_flush_with_particles || is_last_flush_and_never_particles;
+
+    // this setup stage also implicitly calls "makeEmpty" if needed (i.e., is_last_flush_and_never_particles)
+    SetupPos(currSpecies, NewParticleVectorSize, charge, mass, isBTD);
```
guj: When m_doParticleSetup is false, should it still go to Setup*() functions?

ax3l: yes, there is handling inside them to skip parts if `m_doParticleSetUp` is set.

guj: Yes I saw that.  If m_doParticelSetup =false, why still need to go through part of the Setup*() functions?

ax3l: Good question, I will check if we can skip that altogether and delay it in BTD until we fulfill `m_doParticleSetUp`.

ax3l: Ah, the challenge here is that we need to call this setup multiple times for BTD, since we use it to resize the output as we add more particles to the same step in subsequent calls.

guj: when there is no particle, nothing to adjust?

ax3l: I have to adjust a bit more next week for BTD. Have to move out the declaration of constant records so they only happen once

ax3l: Done :)

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -662,161 +664,156 @@ WarpXOpenPMDPlot::DumpToFile (ParticleContainer* pc,
...
+    AMREX_ALWAYS_ASSERT(int_comp_names.size() == pc->NumIntComps());
+
+    WarpXParticleCounter counter(pc);
+    auto const num_dump_particles = counter.GetTotalNumParticles();
+
+    openPMD::Iteration currIteration = GetIteration(iteration, isBTD);
+    openPMD::ParticleSpecies currSpecies = currIteration.particles[name];
+
+    // prepare data structures the first time BTD has non-zero particles
+    //   we set some of them to zero extent, so we need to time that welll
```
RevathiJambunathan:
```suggestion
    //   we set some of them to zero extent, so we need to time that well
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -662,161 +664,156 @@ WarpXOpenPMDPlot::DumpToFile (ParticleContainer* pc,
...
+    bool const is_first_flush_with_particles = num_dump_particles > 0 && ParticleFlushOffset == 0;
+    // BTD: we flush multiple times to the same lab step and thus need to resize
+    //   our declared particle output sizes
+    bool const is_resizing_flush = num_dump_particles > 0 && ParticleFlushOffset > 0;
+    // write structure & declare particles in this (lab) step empty:
+    //   if not BTD, then this is the only (and last) time we flush to this step
+    //   if BTD, then we do this multiple times
```
RevathiJambunathan:
```suggestion
    //   if BTD, then we may do this multiple times until it is the last BTD flush
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -662,161 +664,156 @@ WarpXOpenPMDPlot::DumpToFile (ParticleContainer* pc,
...
+    bool const is_last_flush_to_step = !isBTD || (isBTD && isLastBTDFlush);
+    // well, even in BTD we have to recognize that some lab stations have no
```
RevathiJambunathan:
```suggestion
    // well, even in BTD we have to recognize that some lab stations may have no
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -662,161 +664,156 @@ WarpXOpenPMDPlot::DumpToFile (ParticleContainer* pc,
...
+    // define positions & offset structure
+    const unsigned long long NewParticleVectorSize = num_dump_particles + ParticleFlushOffset;
+    // we will set up empty particles unless it's BTD, where we might add some in a following buffer dump
+    //   during this setup, we mark some particle properties as constant and potentially zero-sized
+    bool doParticleSetup = true;
+    if (isBTD)
+        doParticleSetup = is_first_flush_with_particles || is_last_flush_and_never_particles;
+
+    // this setup stage also implicitly calls "makeEmpty" if needed (i.e., is_last_flush_and_never_particles)
+    //   for BTD, we call this multiple times as we will resize in subsequent dumps
```
RevathiJambunathan:
```suggestion
    //   for BTD, we call this multiple times as we may resize in subsequent dumps if number of particles in the buffer > 0
```

RevathiJambunathan (commented):
This looks good to me. 
Thank you for this PR @ax3l !
I have minor suggestions in the comments

RevathiJambunathan (approved):

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -976,51 +970,110 @@ void
...
+    }
+
+    // make constant
+    using namespace amrex::literals;
+    auto const scalar = openPMD::RecordComponent::SCALAR;
+    for( auto const& comp : positionComponents ) {
+        currSpecies["positionOffset"][comp].makeConstant( 0._prt );
+    }
+    currSpecies["charge"][scalar].makeConstant( charge );
+    currSpecies["mass"][scalar].makeConstant( mass );
```
ax3l, line 1010: Oops, 2x forgot `resetDataset`. Fixed in #3451

PR #2982, opened 2022-03-18
LaserAccelection_1d: Use same build as other 1D tests

PhilMiller:
This should cut ~6:30 off the 1D CI build/test, with no negative impact.


ax3l (approved):
Yay! Good catch!

PR #2985, opened 2022-03-21
Measure time taken in the moving window

RemiLehe:
This adds a timer of the function `MoveWindow`. 
By looking at the corresponding **inclusive** time in the Tiny profiler report, one should be able to see the cost associated with moving the window.


EZoni (approved):

ax3l (approved):
Thx!

PR #2986, opened 2022-03-21
BTD : Redistribute Particles in buffer to lab frame box array and copy with (local=true)

RevathiJambunathan:
When copy particles before we flush the data with plotfile format, we set local=true. 
Otherwise before writing the data, the cell-indices of the particles are determined and are mapped to the box array. 
This is not required when we flush the particle data for BTD, similar to the regular full diagnostics.

Without this change, sometimes, a segfault is generated with 
`amrex::Abort::0::ParticleContainer::locateParticle(): invalid particle. !!!`

Here is the input file I used to test this
[inputs_test1.txt](https://github.com/ECP-WarpX/WarpX/files/8337830/inputs_test1.txt)

Update : 
This PR also Redistributes particles in the lab-frame buffer before copying them to tmp particle containers that are flushed out

### DIFF
--- a/Source/Diagnostics/FlushFormats/FlushFormatPlotfile.cpp
+++ b/Source/Diagnostics/FlushFormats/FlushFormatPlotfile.cpp
@@ -367,7 +367,7 @@ FlushFormatPlotfile::WriteParticles(const std::string& dir,
             }, true);
         } else {
             PinnedMemoryParticleContainer* pinned_pc = particle_diags[i].getPinnedParticleContainer();
-            tmp.copyParticles(*pinned_pc);
+            tmp.copyParticles(*pinned_pc, true);
         }
         // real_names contains a list of all particle attributes.
         // real_flags & int_flags are 1 or 0, whether quantity is dumped or not.
--- a/Source/Diagnostics/WarpXOpenPMD.cpp
+++ b/Source/Diagnostics/WarpXOpenPMD.cpp
@@ -721,6 +721,16 @@ WarpXOpenPMDPlot::DumpToFile (ParticleContainer* pc,
         SetConstParticleRecordsEDPIC(currSpecies, NewParticleVectorSize, charge, mass);
     }
 
+
+    amrex::Print() << "isBTD=" << isBTD << std::endl;
+    amrex::Print() << "iteration=" << iteration << std::endl;
+    amrex::Print() << "num_dump_particles=" << num_dump_particles << std::endl;
+    amrex::Print() << "ParticleFlushOffset=" << ParticleFlushOffset << std::endl;
+    amrex::Print() << "doParticleSetup=" << doParticleSetup << std::endl;
+    amrex::Print() << "is_resizing_flush=" << is_resizing_flush << std::endl;
+    amrex::Print() << "is_last_flush_and_never_particles=" << is_last_flush_and_never_particles << std::endl;
+    amrex::Print() << "is_last_flush_to_step=" << is_last_flush_to_step << std::endl;
+
     // open files from all processors, in case some will not contribute below
     m_Series->flush();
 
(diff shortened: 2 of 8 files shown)


atmyers (approved):

Source/Diagnostics/FlushFormats/FlushFormatPlotfile.cpp
```diff
@@ -367,7 +367,7 @@ FlushFormatPlotfile::WriteParticles(const std::string& dir,
             }, true);
         } else {
             PinnedMemoryParticleContainer* pinned_pc = particle_diags[i].getPinnedParticleContainer();
-            tmp.copyParticles(*pinned_pc);
+            tmp.copyParticles(*pinned_pc, true);
```
ax3l, line 370: Does the same change need to go into `Source/Diagnostics/WarpXOpenPMD.cpp`? :)

RevathiJambunathan: thanks for checking this! yes and commit 8dd66f4 adds this

ax3l (approved):
Thanks a lot, Reva :tada:

atmyers:
This needs more thought I think.

RevathiJambunathan:
This PR to AMReX would fix the issue.
https://github.com/AMReX-Codes/amrex/pull/2673

- [x] once merged, we will update AMReX in the current PR to fix this.

RevathiJambunathan:
The PR is failing for BTD with openPMD format. 
Looking into this now

atmyers:
Do we still need this PR, with the fix to AMReX?

ax3l:
I'll also take a look at the crash with @RevathiJambunathan.

> Do we still need this PR, with the fix to AMReX?

@RevathiJambunathan do you have a reproducer inputs file for the original issue that you saw?

RevathiJambunathan:
Good point @atmyers 
I am re-running this case and testing it. WIll let you know shortly

ax3l:
debugged locally: info for tomorrow.

Somehow, the particle container I receive in `WarpXOpenPMDPlot::DumpToFile` contains zero particles even though we did count it to have some particles before passing it over.
Due to that, the `ParticleFlushOffset` argument increases for subsequent `DumpToFile` for a lab station that has not yet actually dumped particles.

We count particles we get passed in here:
https://github.com/ECP-WarpX/WarpX/blob/4dff22e34e68480d3694dffa671b247d3d4328d2/Source/Diagnostics/WarpXOpenPMD.cpp#L659-L680

@RevathiJambunathan estimates this to be a problem in copying before we call `DumpToFile`. This problem is backend independent.

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -721,6 +721,16 @@ WarpXOpenPMDPlot::DumpToFile (ParticleContainer* pc,
...
+
+    amrex::Print() << "isBTD=" << isBTD << std::endl;
+    amrex::Print() << "iteration=" << iteration << std::endl;
+    amrex::Print() << "num_dump_particles=" << num_dump_particles << std::endl;
+    amrex::Print() << "ParticleFlushOffset=" << ParticleFlushOffset << std::endl;
+    amrex::Print() << "doParticleSetup=" << doParticleSetup << std::endl;
+    amrex::Print() << "is_resizing_flush=" << is_resizing_flush << std::endl;
+    amrex::Print() << "is_last_flush_and_never_particles=" << is_last_flush_and_never_particles << std::endl;
+    amrex::Print() << "is_last_flush_to_step=" << is_last_flush_to_step << std::endl;
+
```
atmyers: did you mean to include these?

RevathiJambunathan: uh oh! no

RevathiJambunathan:
```suggestion
```

RevathiJambunathan:
https://github.com/AMReX-Codes/amrex/pull/2673 
The above AMReX PR fixed the segfault / crash that was detected earlier for BTD with plotfile format.

It also exposed a bug in the BTD in WarpX (also causing PR 3003 to crash)
https://github.com/ECP-WarpX/WarpX/pull/2986/commits/edf044fd29b710df2e38f4c8ab406d80cf2f3fdb
This commit ^ fixes the bug by calling Redistribute for particles in the BTD buffer with lab-frame box-array and geometry.

ax3l (approved):
Thank you so much @atmyers and @RevathiJambunathan :rocket: :sparkles:

PR #2988, opened 2022-03-22
Switch test to use common build with openPMD enabled

PhilMiller:
This should cut a couple minutes off the `cartesian2d` CI build, since it's the only one in 2D that specifies ` -DWarpX_OPENPMD=OFF`

### DIFF
--- a/Regression/WarpX-tests.ini
+++ b/Regression/WarpX-tests.ini
@@ -2008,7 +2008,7 @@ customRunCmd = ./analysis.py
 runtime_params = warpx.do_dynamic_scheduling=0
 dim = 2
 addToCompileString = USE_OPENPMD=FALSE
-cmakeSetupOpts = -DWarpX_DIMS=2 -DWarpX_OPENPMD=OFF
+cmakeSetupOpts = -DWarpX_DIMS=2
 restartTest = 0
 useMPI = 1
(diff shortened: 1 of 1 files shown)


PhilMiller:
With this, `cartesian2d` only does two distinct builds - the 'main' one, and one with PSATD enabled.

Regression/WarpX-tests.ini
```diff
@@ -2008,7 +2008,7 @@ customRunCmd = ./analysis.py
 runtime_params = warpx.do_dynamic_scheduling=0
 dim = 2
 addToCompileString = USE_OPENPMD=FALSE
```
ax3l: Just to keep GNUmake and CMake logic in sync:
```suggestion
addToCompileString =
```

ax3l:
> With this, cartesian2d only does two distinct builds - the 'main' one, and one with PSATD enabled.

Good catch. Since I streamlined all possible binary variants to be opt-in and PSATD not to be mutually exclusive anymore, we could turn on PSATD for all cartesian2d tests and reduce to a single build.

We cover "reduced feature build" in separate CI (GH actions) and don't need to spend time on this in Azure Pipelines (runtime tests).

ax3l (approved):

PR #2989, opened 2022-03-22
Clean up handling of `v_galilean`

dpgrote:
While working on some other code, I noticed that the handling of `v_galilean` as used in the routine `LowerCorner` was messy. The `v_galilean` variable was being passed down through the call chain and then passed into `LowerCorner`, even though `LowerCorner` is a WarpX method and WarpX already owns `v_galilean`. Also, `v_galilean` was used to calculate `galilean_shift` along with other quantities pulled directly from WarpX, with basically the same calculation being done in many different places.  And, `PhysicalParticleContainer` was generating its own copy of `v_galilean`, duplicating what is done in WarpX.

This PR cleans all of this up by putting the Galilean calculations in the `LowerCorner` routine, including a flag to turn it off and an extra time shift. The `m_v_galilean` attribute was removed from the `PhysicalParticleContainer` class.

This turned up several places where the Galilean shift is not applied.
- In the LaserParticleContainer, where the `m_v_galilean` is not initialized (and so zero). This is used in `DepositCurrent`. This was a bug and is fixed in this PR (with the benchmarks updated).
- When the mirror is applied, in the routine `NullifyMF`, the Galilean factor was not included (`v_galilean` is set to zero). This is fixed and the benchmarks updated.
- The same for the FieldProbe when doing the field gather. This is also changed to use the shift (but shouldn't require any benchmark updates).
 
Note that now all calls need the shift so it is now always included.


Source/WarpX.cpp
```diff
@@ -2174,12 +2177,23 @@ WarpX::getRealBox(const Box& bx, int lev)
 }
 
 std::array<Real,3>
-WarpX::LowerCorner(const Box& bx, std::array<amrex::Real,3> galilean_shift, int lev)
+WarpX::LowerCorner(const Box& bx, int lev, bool with_galilean_shift, amrex::Real time_shift_delta)
```
```suggestion
WarpX::LowerCorner(const Box& bx, const int lev, const bool with_galilean_shift, const amrex::Real time_shift_delta)
```

Source/WarpX.H
```diff
@@ -724,15 +724,10 @@ public:
 
     static std::array<amrex::Real,3> CellSize (int lev);
     static amrex::RealBox getRealBox(const amrex::Box& bx, int lev);
-    static std::array<amrex::Real,3> LowerCorner (const amrex::Box& bx,
-                                                  std::array<amrex::Real,3> galilean_shift, int lev);
+    static std::array<amrex::Real,3> LowerCorner (const amrex::Box& bx, int lev,
+                                                  bool with_galilean_shift, amrex::Real time_shift_delta);
```
```suggestion
    static std::array<amrex::Real,3> LowerCorner (const amrex::Box& bx, const int lev,
                                                  const bool with_galilean_shift, const amrex::Real time_shift_delta);
```

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -623,24 +613,14 @@ WarpXParticleContainer::DepositCharge (WarpXParIter& pti, RealVector const& wp,
...
-                m_v_galilean[0]*time_shift_rho_old,
-                m_v_galilean[1]*time_shift_rho_old,
-                m_v_galilean[2]*time_shift_rho_old };
+            time_shift_delta = 0;
         } else{
-            galilean_shift = {
-                m_v_galilean[0]*time_shift_rho_new,
-                m_v_galilean[1]*time_shift_rho_new,
-                m_v_galilean[2]*time_shift_rho_new };
+            time_shift_delta = dt;
```
EZoni: Would it work if we simply do the following?
```cpp
const amrex::Real time_shift_delta = (icomp == 0) ? 0._rt : dt;
```

Source/WarpX.cpp
```diff
@@ -1207,6 +1207,9 @@ WarpX::ReadParameters ()
             if (m_v_comoving[0] != 0. || m_v_comoving[1] != 0. || m_v_comoving[2] != 0.) {
                 amrex::Abort("Esirkepov current deposition cannot be used with the comoving PSATD algorithm");
             }
+            if (m_v_galilean[0] != 0. || m_v_galilean[1] != 0. || m_v_galilean[2] != 0.) {
+                amrex::Abort("Esirkepov current deposition  cannot be used with the Galilean algorithm.");
```
```suggestion
                amrex::Abort("Esirkepov current deposition cannot be used with the Galilean algorithm.");
```

Source/Particles/PhotonParticleContainer.cpp
```diff
@@ -136,14 +136,7 @@ PhotonParticleContainer::PushPX (WarpXParIter& pti,
...
     // Lower corner of tile box physical domain (take into account Galilean shift)
-    amrex::Real cur_time = WarpX::GetInstance().gett_new(lev);
-    const auto& time_of_last_gal_shift = WarpX::GetInstance().time_of_last_gal_shift;
-    amrex::Real time_shift = (cur_time - time_of_last_gal_shift);
-    amrex::Array<amrex::Real,3> galilean_shift = {
-        m_v_galilean[0]*time_shift,
-        m_v_galilean[1]*time_shift,
-        m_v_galilean[2]*time_shift };
-    const std::array<Real, 3>& xyzmin = WarpX::LowerCorner(box, galilean_shift, gather_lev);
+    const std::array<Real, 3>& xyzmin = WarpX::LowerCorner(box, gather_lev, true, 0.);
```
EZoni:
```suggestion
    const std::array<amrex::Real, 3>& xyzmin = WarpX::LowerCorner(box, gather_lev, true, 0._rt);
```

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -2223,7 +2211,7 @@ PhysicalParticleContainer::PushP (int lev, Real dt,
 
             const auto getExternalEB = GetExternalEBField(pti);
 
-            const auto& xyzmin = WarpX::GetInstance().LowerCornerWithGalilean(box,m_v_galilean,lev);
+            const auto& xyzmin = WarpX::LowerCorner(box, lev, true, 0.);
```
EZoni:
```suggestion
            const std::array<amrex::Real,3>& xyzmin = WarpX::LowerCorner(box, lev, true, 0._rt);
```

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -2576,14 +2564,7 @@ PhysicalParticleContainer::PushPX (WarpXParIter& pti,
...
     // Lower corner of tile box physical domain (take into account Galilean shift)
-    Real cur_time = WarpX::GetInstance().gett_new(lev);
-    const auto& time_of_last_gal_shift = WarpX::GetInstance().time_of_last_gal_shift;
-    Real time_shift = (cur_time - time_of_last_gal_shift);
-    amrex::Array<amrex::Real,3> galilean_shift ={
-        m_v_galilean[0]*time_shift,
-        m_v_galilean[1]*time_shift,
-        m_v_galilean[2]*time_shift };
-    const std::array<Real, 3>& xyzmin = WarpX::LowerCorner(box, galilean_shift, gather_lev);
+    const std::array<Real, 3>& xyzmin = WarpX::LowerCorner(box, gather_lev, true, 0.);
```
EZoni:
```suggestion
    const std::array<amrex::Real, 3>& xyzmin = WarpX::LowerCorner(box, gather_lev, true, 0._rt);
```

Source/Particles/RigidInjectedParticleContainer.cpp
```diff
@@ -356,7 +356,7 @@ RigidInjectedParticleContainer::PushP (int lev, Real dt,
 
             const auto getExternalEB = GetExternalEBField(pti);
 
-            const auto& xyzmin = WarpX::GetInstance().LowerCornerWithGalilean(box,m_v_galilean,lev);
+            const auto& xyzmin = WarpX::LowerCorner(box, lev, true, 0.);
```
EZoni:
```suggestion
            const std::array<amrex::Real,3>& xyzmin = WarpX::LowerCorner(box, lev, true, 0._rt);
```

Source/Utils/WarpXUtil.cpp
```diff
@@ -220,14 +220,7 @@ void NullifyMF(amrex::MultiFab& mf, int lev, amrex::Real zmin, amrex::Real zmax)
...
         // Get box lower and upper physical z bound, and dz
-#if defined(WARPX_DIM_3D)
-            amrex::Array<amrex::Real,3> galilean_shift = { 0._rt, 0._rt, 0._rt, };
-#elif defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
-            amrex::Array<amrex::Real,3> galilean_shift = { 0._rt, std::numeric_limits<amrex::Real>::quiet_NaN(),  0._rt, } ;
-#elif defined(WARPX_DIM_1D_Z)
-            amrex::Array<amrex::Real,3> galilean_shift = {std::numeric_limits<amrex::Real>::quiet_NaN(), std::numeric_limits<amrex::Real>::quiet_NaN(),  0._rt, } ;
-#endif
-        const amrex::Real zmin_box = WarpX::LowerCorner(bx, galilean_shift, lev)[2];
+        const amrex::Real zmin_box = WarpX::LowerCorner(bx, lev, false, 0.)[2];
```
EZoni:
```suggestion
        const amrex::Real zmin_box = WarpX::LowerCorner(bx, lev, false, 0._rt)[2];
```

Source/Particles/ElementaryProcess/QEDPhotonEmission.cpp
```diff
@@ -57,11 +56,7 @@ PhotonEmissionTransformFunc (QuantumSynchrotronGetOpticalDepth opt_depth_functor
     m_dx_arr = {dx[0], dx[1], dx[2]};
 
     // Lower corner of tile box physical domain (take into account Galilean shift)
-    amrex::Real cur_time = WarpX::GetInstance().gett_new(lev);
-    const auto& time_of_last_gal_shift = WarpX::GetInstance().time_of_last_gal_shift;
-    amrex::Real time_shift = (cur_time - time_of_last_gal_shift);
-    amrex::Array<amrex::Real,3> galilean_shift = { v_galilean[0]*time_shift, v_galilean[1]*time_shift, v_galilean[2]*time_shift };
-    const std::array<amrex::Real, 3>& xyzmin = WarpX::LowerCorner(box, galilean_shift, lev);
+    const std::array<amrex::Real, 3>& xyzmin = WarpX::LowerCorner(box, lev, true, 0.);
```
EZoni:
```suggestion
    const std::array<amrex::Real, 3>& xyzmin = WarpX::LowerCorner(box, lev, true, 0._rt);
```

Source/Particles/ElementaryProcess/QEDPairGeneration.cpp
```diff
@@ -54,11 +53,7 @@ PairGenerationTransformFunc (BreitWheelerGeneratePairs const generate_functor,
     m_dx_arr = {dx[0], dx[1], dx[2]};
 
     // Lower corner of tile box physical domain (take into account Galilean shift)
-    amrex::Real cur_time = WarpX::GetInstance().gett_new(lev);
-    const auto& time_of_last_gal_shift = WarpX::GetInstance().time_of_last_gal_shift;
-    amrex::Real time_shift = (cur_time - time_of_last_gal_shift);
-    amrex::Array<amrex::Real,3> galilean_shift = { v_galilean[0]*time_shift, v_galilean[1]*time_shift, v_galilean[2]*time_shift };
-    const std::array<amrex::Real, 3>& xyzmin = WarpX::LowerCorner(box, galilean_shift, lev);
+    const std::array<amrex::Real, 3>& xyzmin = WarpX::LowerCorner(box, lev, true, 0.);
```
EZoni:
```suggestion
    const std::array<amrex::Real, 3>& xyzmin = WarpX::LowerCorner(box, lev, true, 0._rt);
```

Source/Particles/ElementaryProcess/Ionization.cpp
```diff
@@ -64,11 +63,7 @@ IonizationFilterFunc::IonizationFilterFunc (const WarpXParIter& a_pti, int lev,
     m_dx_arr = {dx[0], dx[1], dx[2]};
 
     // Lower corner of tile box physical domain (take into account Galilean shift)
-    amrex::Real cur_time = WarpX::GetInstance().gett_new(lev);
-    const auto& time_of_last_gal_shift = WarpX::GetInstance().time_of_last_gal_shift;
-    amrex::Real time_shift = (cur_time - time_of_last_gal_shift);
-    amrex::Array<amrex::Real,3> galilean_shift = { v_galilean[0]*time_shift, v_galilean[1]*time_shift, v_galilean[2]*time_shift };
-    const std::array<amrex::Real, 3>& xyzmin = WarpX::LowerCorner(box, galilean_shift, lev);
+    const std::array<amrex::Real, 3>& xyzmin = WarpX::LowerCorner(box, lev, true, 0.);
```
EZoni:
```suggestion
    const std::array<amrex::Real, 3>& xyzmin = WarpX::LowerCorner(box, lev, true, 0._rt);
```

Source/FieldSolver/WarpXPushFieldsEM.cpp
```diff
@@ -1183,8 +1182,7 @@ WarpX::ApplyInverseVolumeScalingToChargeDensity (MultiFab* Rho, int lev)
         // Lower corner of tile box physical domain
         // Note that this is done before the tilebox.grow so that
         // these do not include the guard cells.
-        std::array<amrex::Real,3> galilean_shift = {0,0,0};
-        const std::array<Real, 3>& xyzmin = WarpX::LowerCorner(tilebox, galilean_shift, lev);
+        const std::array<Real, 3>& xyzmin = WarpX::LowerCorner(tilebox, lev, false, 0.);
```
EZoni:
```suggestion
        const std::array<amrex::Real, 3>& xyzmin = WarpX::LowerCorner(tilebox, lev, false, 0._rt);
```

Source/FieldSolver/WarpXPushFieldsEM.cpp
```diff
@@ -1020,8 +1020,7 @@ WarpX::ApplyInverseVolumeScalingToCurrentDensity (MultiFab* Jx, MultiFab* Jy, Mu
...
```
EZoni:
```suggestion
        const std::array<amrex::Real, 3>& xyzmin = WarpX::LowerCorner(tilebox, lev, false, 0._rt);
```

Source/Diagnostics/ReducedDiags/ParticleExtrema.cpp
```diff
@@ -434,7 +429,7 @@ void ParticleExtrema::ComputeDiags (int step)
                     amrex::Box box = pti.tilebox();
                     box.grow(ngEB);
                     const Dim3 lo = amrex::lbound(box);
-                    const std::array<amrex::Real, 3>& xyzmin = WarpX::LowerCorner(box, galilean_shift, lev);
+                    const std::array<amrex::Real, 3>& xyzmin = WarpX::LowerCorner(box, lev, true, 0.);
```
EZoni:
```suggestion
                    const std::array<amrex::Real, 3>& xyzmin = WarpX::LowerCorner(box, lev, true, 0._rt);
```

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
```diff
@@ -466,8 +466,7 @@ void FieldProbe::ComputeDiags (int step)
                 ParticleReal* const AMREX_RESTRICT part_Bz = attribs[FieldProbePIdx::Bz].dataPtr();
                 ParticleReal* const AMREX_RESTRICT part_S = attribs[FieldProbePIdx::S].dataPtr();
 
-                amrex::Vector<amrex::Real> v_galilean{amrex::Vector<amrex::Real>(3, amrex::Real(0.))};
-                const auto &xyzmin = WarpX::GetInstance().LowerCornerWithGalilean(box, v_galilean, lev);
+                const auto &xyzmin = WarpX::LowerCorner(box, lev, false, 0.);
```
EZoni:
```suggestion
                const auto &xyzmin = WarpX::LowerCorner(box, lev, false, 0._rt);
```

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -415,22 +415,12 @@ WarpXParticleContainer::DepositCurrent (WarpXParIter& pti,
...
     // Take into account Galilean shift
-    Real cur_time = warpx.gett_new(lev);
-    const auto& time_of_last_gal_shift = warpx.time_of_last_gal_shift;
-    Real time_shift = (cur_time + 0.5_rt*dt - time_of_last_gal_shift);
-    amrex::Array<amrex::Real,3> galilean_shift = {
-        m_v_galilean[0]* time_shift,
-        m_v_galilean[1]*time_shift,
-        m_v_galilean[2]*time_shift };
-    const std::array<Real, 3>& xyzmin = WarpX::LowerCorner(tilebox, galilean_shift, depos_lev);
+    const std::array<Real, 3>& xyzmin = WarpX::LowerCorner(tilebox, depos_lev, true, 0.5_rt*dt);
```
EZoni:
```suggestion
    const std::array<amrex::Real, 3>& xyzmin = WarpX::LowerCorner(tilebox, depos_lev, true, 0.5_rt*dt);
```

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -623,24 +613,14 @@ WarpXParticleContainer::DepositCharge (WarpXParIter& pti, RealVector const& wp,
...
+            time_shift_delta = 0;
         } else{
-            galilean_shift = {
-                m_v_galilean[0]*time_shift_rho_new,
-                m_v_galilean[1]*time_shift_rho_new,
-                m_v_galilean[2]*time_shift_rho_new };
+            time_shift_delta = dt;
         }
-        const auto& xyzmin = WarpX::LowerCorner(tilebox, galilean_shift, depos_lev);
+        const auto& xyzmin = WarpX::LowerCorner(tilebox, depos_lev, true, time_shift_delta);
```
EZoni:
```suggestion
        const std::array<amrex::Real,3>& xyzmin = WarpX::LowerCorner(tilebox, depos_lev, true, time_shift_delta);
```

EZoni (changes requested):
Thank you for this PR, Dave! The code is much cleaner and less error-prone now. I'll think about your question on whether missing the Galilean shift in the laser particle container is a bug. In the meantime I suggested a few minor changes.

Source/Particles/WarpXParticleContainer.cpp
dpgrote: Thanks @EZoni Yes, that will work. I'll make the change. I've accepted all of your other suggestions as well.

RemiLehe:
@dpgrote Thanks a lot for this PR!!
You are correct: the fact that `m_v_galilean` is zero for the laser particles is indeed a bug! Thanks for fixing it.

RemiLehe (approved):

EZoni:
@dpgrote As you pointed out in the PR description and as confirmed with @RemiLehe on Slack, the Galilean shift for mirrors also needs to be fixed.

EZoni:
@dpgrote The flag for the Galilean shift needs to be removed from this call, too:
https://github.com/dpgrote/WarpX/blob/5170b720ec193f52f8bcd1bb5aedbe0125173423/Source/FieldSolver/SpectralSolver/SpectralFieldDataRZ.cpp#L178-L180
```cpp
        // Create the Hankel transformer for each box.
        std::array<amrex::Real,3> xmax = WarpX::UpperCorner(mfi.tilebox(), lev, false, 0._rt);
        multi_spectral_hankel_transformer[mfi] = SpectralHankelTransformer(grid_size[0], n_rz_azimuthal_modes, xmax[0]);
```
I guess that including the shift (as it will be by default once we remove the flag from this call) is the right thing to do here as well, so this was probably another bug in the main branch?

dpgrote:
Thanks @EZoni - that was an oversight. In this case, the shift doesn't matter since only the transverse (i.e. radial) value is used and a radial Galilean velocity is not supported.

Source/WarpX.H
```diff
@@ -724,9 +724,17 @@ public:
...
+    /**
+     * \brief Return the lower or upper corner of the box in real units.
+     * \param bx The box
+     * \param lev The refinement level of the box
+     * \param time_shift_delta The time relative to the current time at which to calculate the position
+     *                         (when v_galilean is not zero)
+     * \return An array of the position coordinates
+     */
+    static std::array<amrex::Real,3> LowerCorner (const amrex::Box& bx, const int lev, const amrex::Real time_shift_delta);
+    static std::array<amrex::Real,3> UpperCorner (const amrex::Box& bx, const int lev, const amrex::Real time_shift_delta);
```
EZoni: Sorry, just one last request. I think for Doxygen it'd be better to have separate Doxygen strings for separate functions:
```suggestion
    /**
     * \brief Return the lower corner of the box in real units.
     * \param bx The box
     * \param lev The refinement level of the box
     * \param time_shift_delta The time relative to the current time at which to calculate the position
     *                         (when v_galilean is not zero)
     * \return An array of the position coordinates
     */
    static std::array<amrex::Real,3> LowerCorner (const amrex::Box& bx, const int lev, const amrex::Real time_shift_delta);
    /**
     * \brief Return the upper corner of the box in real units.
     * \param bx The box
     * \param lev The refinement level of the box
     * \param time_shift_delta The time relative to the current time at which to calculate the position
     *                         (when v_galilean is not zero)
     * \return An array of the position coordinates
     */
    static std::array<amrex::Real,3> UpperCorner (const amrex::Box& bx, const int lev, const amrex::Real time_shift_delta);
```

EZoni (approved):
Thank you again for this PR, Dave! Ready to merge now.

PR #2995, opened 2022-03-24
Synchronize all nodal points in sub-cycling loop

RemiLehe:
This should avoid instabilities that happen at the boundary of domains.


lucafedeli88 (approved):
Thanks really a lot! it seems to work nicely!

lucafedeli88:
This PR fixes an issue that I've recently experienced with mesh refinement used together with single precision.
This figure shows a corner of a simulation case with mesh refinement in double precision (no apparent issues):
image
however, the same simulation performed in single precision leads to the development of nonphysical instabilities: 
image
with this  PR the issue is fixed and the result is almost identical to that obtained in double precision:
image

PR #2998, opened 2022-03-25
Add Field Probe to Laser Ion Test

ElisaRheaume:
Adds Field Probes to Laser Ion Example

- Z axis instantaneous values. 1 per cell in Z direction
- Instantaneous point for forward-scattered light on Z axis at 15 micrometers
- Integrated line for forward-scattered light normal to z axis at 15 micrometers


ax3l (approved):
Great work, thanks a lot! :)

PR #2999, opened 2022-03-25
Langmuir 3D Tests: Better Analysis Plots

EZoni:
Same as #2960, but for the 3D tests (now plotting the slice at y=0, instead of that at z=0).

Old plot:
<p align="center">
    
</p>

New plot:
<p align="center">
    
</p>


ax3l (approved):
Thank you! :)

PR #3001, opened 2022-03-25
Add particle filter in ParticleReductionFunctor

hklion:
This adds an optional filter function to the particle averaging diagnostic (#2892). Particles are included or not based on a filter function that takes `(x, y, z, ux, uy, uz)`. The `particle_fields_diags` tests also test the filter for both plotfiles and openPMD.

PR #3007, opened 2022-03-28
Change particle filter parser to check it as a boolean instead of comparing to 0.5

hklion:
In discussing #3001, @EZoni pointed out that comparing the diagnostics particle filter functions to 0.5 isn't intuitive, especially because they are described as booleans. I also realized that there was a mismatch between the documentation and the implementation in the particle histograms. We think it would make more sense to check the filter functions as booleans, rather than by comparing to 0.5. For boolean expressions, this will be equivalent. This PR is also to check that this change wouldn't break any tests.

* The filter function in the particle histogram was already checking `if (filter(...))` rather than `if (filter(...) > 0.5)`, so I've updated the documentation to match.
* The `ParserFilter` in `Particles/Filter/FilterFunctors.H` is used to filter the particles written out in `FullDiagnostics` for both plotfiles and OpenPMD. It's currently checking `if (filter(...) > 0.5)` so I've changed it to `if (filter(...))` and updated the documentation.

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -2045,7 +2045,7 @@ In-situ capabilities can be used by turning on Sensei or Ascent (provided they a
     The value provided should be an integer greater than or equal to 0.
 
 * ``<diag_name>.<species_name>.plot_filter_function(t,x,y,z,ux,uy,uz)`` (`string`) optional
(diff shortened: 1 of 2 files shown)


EZoni (approved):
Thank you for cleaning this up, Hannah! Looks good to me, let's see if CI goes well.

ax3l (approved):
Thank you, good call!

PR #3008, opened 2022-03-28
Take into account the cost of 'shiftMF' in load-balancing

RemiLehe:
The cost of shifting the field values (in the moving window) can sometimes be non-negligible, and needs to be taken into account when doing load-balancing.

### DIFF
--- a/Source/Utils/WarpXMovingWindow.cpp
+++ b/Source/Utils/WarpXMovingWindow.cpp
@@ -177,7 +177,7 @@ WarpX::MoveWindow (const int step, bool move_j)
                 if (dim == 2) Efield_parser = Ezfield_parser->compile<3>();
             }
             shiftMF(*Bfield_fp[lev][dim], geom[lev], num_shift, dir, lev, B_external_grid[dim], use_Bparser, Bfield_parser);
-            shiftMF(*Efield_fp[lev][dim], geom[lev], num_shift, dir, lev, E_external_grid[dim], use_Eparser, Efield_parser, lev);
+            shiftMF(*Efield_fp[lev][dim], geom[lev], num_shift, dir, lev, E_external_grid[dim], use_Eparser, Efield_parser);
             if (fft_do_time_averaging) {
                 shiftMF(*Bfield_avg_fp[lev][dim], geom[lev], num_shift, dir, lev, B_external_grid[dim], use_Bparser, Bfield_parser);
                 shiftMF(*Efield_avg_fp[lev][dim], geom[lev], num_shift, dir, lev, E_external_grid[dim], use_Eparser, Efield_parser);
@@ -296,7 +296,7 @@ WarpX::MoveWindow (const int step, bool move_j)
 
 void
 WarpX::shiftMF (amrex::MultiFab& mf, const amrex::Geometry& geom,
-                int num_shift, int dir, int lev,
+                int num_shift, int dir, const int lev,
                 amrex::Real external_field, bool useparser,
                 amrex::ParserExecutor<3> const& field_parser)
 {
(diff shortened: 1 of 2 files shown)


Source/WarpX.H
```diff
@@ -151,7 +151,7 @@ public:
     ParticleBoundaryBuffer& GetParticleBoundaryBuffer () { return *m_particle_boundary_buffer; }
 
     static void shiftMF (amrex::MultiFab& mf, const amrex::Geometry& geom,
-                         int num_shift, int dir, amrex::Real external_field=0.0,
+                         int num_shift, int dir, int lev, amrex::Real external_field=0.0,
```
```suggestion
                         int num_shift, int dir, const int lev, amrex::Real external_field=0.0,
```

Source/Utils/WarpXMovingWindow.cpp
```diff
@@ -355,13 +356,19 @@ WarpX::shiftMF (amrex::MultiFab& mf, const amrex::Geometry& geom, int num_shift,
...
 #endif
 
-
     for (amrex::MFIter mfi(tmpmf); mfi.isValid(); ++mfi )
     {
+        if (cost && WarpX::load_balance_costs_update_algo == LoadBalanceCostsUpdateAlgo::Timers)
+        {
+            amrex::Gpu::synchronize();
+        }
+        Real wt = amrex::second();
```
EZoni:
```suggestion
        amrex::Real wt = amrex::second();
```

Source/Utils/WarpXMovingWindow.cpp
```diff
@@ -295,7 +295,8 @@ WarpX::MoveWindow (const int step, bool move_j)
 }
 
 void
-WarpX::shiftMF (amrex::MultiFab& mf, const amrex::Geometry& geom, int num_shift, int dir,
+WarpX::shiftMF (amrex::MultiFab& mf, const amrex::Geometry& geom,
+                int num_shift, int dir, int lev,
```
EZoni:
```suggestion
                int num_shift, int dir, const int lev,
```

Source/Utils/WarpXMovingWindow.cpp
```diff
@@ -176,78 +176,78 @@ WarpX::MoveWindow (const int step, bool move_j)
                 if (dim == 1) Efield_parser = Eyfield_parser->compile<3>();
                 if (dim == 2) Efield_parser = Ezfield_parser->compile<3>();
             }
-            shiftMF(*Bfield_fp[lev][dim], geom[lev], num_shift, dir, B_external_grid[dim], use_Bparser, Bfield_parser);
-            shiftMF(*Efield_fp[lev][dim], geom[lev], num_shift, dir, E_external_grid[dim], use_Eparser, Efield_parser);
+            shiftMF(*Bfield_fp[lev][dim], geom[lev], num_shift, dir, lev, B_external_grid[dim], use_Bparser, Bfield_parser);
+            shiftMF(*Efield_fp[lev][dim], geom[lev], num_shift, dir, lev, E_external_grid[dim], use_Eparser, Efield_parser, lev);
```
EZoni: Was the last argument `lev` added by mistake here?
```suggestion
            shiftMF(*Efield_fp[lev][dim], geom[lev], num_shift, dir, lev, E_external_grid[dim], use_Eparser, Efield_parser);
```

EZoni (commented):
Thank you, Remi! Looks good to me, I left only a few quick suggestions. Also to double check: in the calls on the coarse patch, the level passed as argument intentionally matches that of the MultiFab rather than the one of the `geom` objects, right? (I'm referring to calls where the first arguments are like `shiftMF(*mf[lev], geom[lev-1], ...)`)

EZoni (approved):

RemiLehe:
Thanks for the review!

Regarding the occurrence of `geom[lev-1]`: this comes up because we call `shiftMF` on the coarse patch (i.e. `_cp` MultiFab) of level `lev`. However, here the cost are used in order to load-balance the DistributionMapping of each level, and the `_cp` of level `lev` are indeed associated to the DistributionMapping for level `lev` (not `lev - 1`), so the associated cost should be attributed to level `lev` (not `lev-1`).

(I am not sure how clear the above is, but we could also discuss it off-line.)

PR #3010, opened 2022-03-30
CI: Fix Small Issue w/ Langmuir Plots

EZoni:
This was introduced by mistake in #2960 and #2999. If one is debugging a PR that by chance produces the wrong electric field `E_sim`, for example with very high min/max values because of some instability, then using the same (global) min/max values for the colorbars of both the plot of `E_sim` and the plot of `E_th` would result in the wrong plot for `E_th` (whose wrong colorbar would make the plot not look like it should according to the theoretical formulas). This makes sure that the colorbar of the theory plot is separate and not affected by potential bugs in the simulation plot, making debugging easier.


ax3l (approved):
Thx! :)

PR #3011, opened 2022-03-30
[WIP] Add Zhang et al. high-field correction to ADK ionization model (Hydrogen only)

MaxThevenet:
ADK is valid for an external field smaller than the critical field, but we sometimes needs to simulate higher fields. Article [Zhang et al., PRA 90, 043410 (2014)](https://journals.aps.org/pra/pdf/10.1103/PhysRevA.90.043410) proposes some correction to ADK  for high fields, which can yield some small difference. This PR proposes to implement this correction, for Hydrogen only for the moment.


MaxThevenet:
This old PR is very out of sync, superseded by #4505.

PR #3012, opened 2022-03-30
Vay Deposition: Fix SyncCurrent, FFTs

EZoni:
This PR fixes a few bugs in the Vay current deposition algorithm:
- The initialization to zero of `current_fp_vay` was missing from Source/Initialization/WarpXInitData.cpp.
- The synchronization of the current was done on the wrong current (D, instead of J). 
- Both forward and backward FFTs must be computed including the guard cells in this case, and the option `psatd.periodic_single_box_fft` cannot be used (especially once the cumulative sums will be added).

Old pseudo-code:
- deposit D
- synchronize D
- FFT of D (with/without guard cells, depending on `psatd.periodic_single_box_fft`)
- "J = D/k"
- FFT^(-1) of J (always without guard cells)
- push fields

New pseudo-code:
- deposit D
- FFT of D (always with guard cells, cannot use `psatd.periodic_single_box_fft`)
- "J = D/k"
- FFT^(-1) of J (always with guard cells)
- synchronize J
- FFT of J (always with guard cells, cannot use `psatd.periodic_single_box_fft`)
- push fields

Note that as a result of not using the periodic single box option for the Langmuir tests with the Vay deposition, we have to increase the tolerance used to check rho v. div(E) in the analysis scripts, from `1e-9` to `1e-3`. Let me know if you think this should not happen.


RemiLehe (approved):
Thanks a lot for this PR!

PR #3013, opened 2022-03-31
Merge `FillBoundary` and `NodalSync` (`OneStep_nosub`, `OneStep_sub1`)

RemiLehe:
Address some of the issues in #3027. In particular, merge calls to `FillBoundary` and `NodalSync` when possible, for the subroutines `OneStep_nosub` and `OneStep_sub1` (`OneStep_multiJ` will be tackled separately, because it seemed to cause some roundoff benchmark changes).

There are no optimizations with respect to avoiding redundant calls to `FillBoundary` yet here: will be tackled in future PRs.

The CI tests `Python_wrappers` and `reduced_diags_single_precision` required a benchmark update (maybe too sensitive).


cmake/dependencies/AMReX.cmake
```diff
@@ -234,10 +234,10 @@ set(WarpX_amrex_src ""
     "Local path to AMReX source directory (preferred if set)")
 
 # Git fetcher
-set(WarpX_amrex_repo "https://github.com/AMReX-Codes/amrex.git"
+set(WarpX_amrex_repo "https://github.com/WeiqunZhang/amrex.git"
     CACHE STRING
     "Repository URI to pull and build AMReX from if(WarpX_amrex_internal)")
-set(WarpX_amrex_branch "3aafa306ad820bbcc50a4b7c14fe912406427d1d"
+set(WarpX_amrex_branch "fbsync"
```
ax3l: TODO: open as PR to mainline AMReX (cc @WeiqunZhang).

@RemiLehe saw the performance is good with this feature (no overhead in `FillBoundary` and saves the extra blocks to sync we saw in traces).

EZoni: I think the PR was already opened and merged in AMReX: https://github.com/AMReX-Codes/amrex/pull/2683.

EZoni:
Cartesian 2D tests that currently fail:
- [ ] `Langmuir_multi_2d_MR_psatd`
- [ ] `comoving_2d_psatd_hybrid`
- [ ] `galilean_2d_psatd`
- [ ] `galilean_2d_psatd_hybrid`
- [ ] `multi_J_2d_psatd`
- [ ] `multi_J_2d_psatd_pml`
- [ ] `pml_x_psatd`

Cartesian 3D tests that currently fail:
- [ ] `divb_cleaning_3d`
- [ ] `galilean_3d_psatd`
- [ ] `pml_psatd_dive_divb_cleaning`
- [ ] `restart_psatd`

Python/PICMI tests that currently fail:
- [ ] `Python_wrappers`

These test failures are caused by moving the synchronization of nodal points to the beginning of the PIC loop. Looking into how to do this step by step, work in progress (see future commits).

EZoni:
Python/PICMI tests that currently fail:
- [ ] `Python_wrappers`

EZoni:
Cartesian 3D tests that currently fail:
- [ ] `Langmuir_multi_2d_MR_psatd`
- [ ] `galilean_3d_psatd`
- [ ] `pml_psatd_dive_divb_cleaning`
- [ ] `restart_psatd`

EZoni:
Note that the CI test failures above are caused by the fact that the synchronization of nodal points was moved at the beginning of the PIC loop.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -469,14 +463,11 @@ WarpX::OneStep_nosub (Real cur_time)
...
-        FillBoundaryE(guard_cells.ng_FieldSolver);
+        FillBoundaryE(guard_cells.ng_FieldSolver, WarpX::sync_nodal_points);
         EvolveF(0.5_rt * dt[0], DtType::SecondHalf);
         EvolveG(0.5_rt * dt[0], DtType::SecondHalf);
         EvolveB(0.5_rt * dt[0], DtType::SecondHalf); // We now have B^{n+1}
-
-        // Synchronize E and B fields on nodal points
-        NodalSync(Efield_fp, Efield_cp);
-        NodalSync(Bfield_fp, Bfield_cp);
+        FillBoundaryB(guard_cells.ng_FieldSolver, WarpX::sync_nodal_points);
```
EZoni: This call to `FillBoundaryB` was added in order to keep the nodal synchronization otherwise removed when deleting the call to `NodalSync` (note that this is in `WarpX::OneStep_nosub`).

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -774,19 +765,21 @@ WarpX::OneStep_sub1 (Real curtime)
...
 
-    FillBoundaryF(fine_lev, PatchType::coarse, guard_cells.ng_FieldSolverF);
+    FillBoundaryF(fine_lev, PatchType::coarse, guard_cells.ng_FieldSolverF, WarpX::sync_nodal_points);
 
     EvolveE(coarse_lev, PatchType::fine, 0.5_rt*dt[coarse_lev]);
-    FillBoundaryE(coarse_lev, PatchType::fine, guard_cells.ng_FieldSolver);
+    FillBoundaryE(coarse_lev, PatchType::fine, guard_cells.ng_FieldSolver, WarpX::sync_nodal_points);
 
     EvolveB(coarse_lev, PatchType::fine, 0.5_rt*dt[coarse_lev], DtType::SecondHalf);
+    FillBoundaryB(coarse_lev, PatchType::fine, guard_cells.ng_FieldSolver, WarpX::sync_nodal_points);
```
EZoni, line 776: This call to `FillBoundaryB` was added in order to keep the nodal synchronization otherwise removed when deleting the call to `NodalSync`.

Source/Evolve/WarpXEvolve.cpp
WeiqunZhang: Note that calling `MultiFab::OverrideSync` is now equivalent to calling `FillBoundaryAndSync` with zero ghost cells.  So if the only purpose here is to synchronize, not to fill ghost cells, `NodalSync` or `FillBoundaryB` with zero ghost cells should be used.

EZoni: OK, thank you, then I think I'll keep the call to `NodalSync` for now. In my opinion it does make sense to also fill the guard cells at this point, as B has been updated, but I think it would make sense to do this while we also maybe remove other redundant calls to `FillBoundarB` (without synchronization) that appear in other parts of the PIC loop. This is part of the optimizations mentioned in #3027, so it's probably better to try them separately in small PRs. Here I'll try to just merge the calls when possible, without changing the number of calls to fill the guard cells or synchronize.

EZoni:
I implemented many of the latest changes in this PR, so I think it would make more sense if the review is led by @WeiqunZhang and @RemiLehe here (@RemiLehe note that I cannot select you as reviewer because you are the author of the original PR).

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -802,8 +795,7 @@ WarpX::OneStep_sub1 (Real curtime)
     if ( safe_guard_cells )
         FillBoundaryB(coarse_lev, PatchType::fine, guard_cells.ng_FieldSolver);
 
-    // Synchronize all nodal points at the end of the timestep
-    NodalSync(Efield_fp, Efield_cp);
+    // Synchronize nodal points at the end of the time step
     NodalSync(Bfield_fp, Bfield_cp);
```
RemiLehe: The `NodalSync` for the B field and F field should be remove too, no?

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -472,13 +466,12 @@ WarpX::OneStep_nosub (Real cur_time)
             amrex::Abort(Utils::TextMsg::Err("Medium for EM is unknown"));
         }
 
-        FillBoundaryE(guard_cells.ng_FieldSolver);
+        FillBoundaryE(guard_cells.ng_FieldSolver, WarpX::sync_nodal_points);
```
RemiLehe, line 473: Could you also add `WarpX::sync_nodal_points` for `FillBoundaryB` above?
In this case, we can also probably remove the `NodalSync` for the Bfield below.

Source/Evolve/WarpXEvolve.cpp
EZoni: I thought this was making some CI tests fail, but I removed all calls in https://github.com/ECP-WarpX/WarpX/pull/3013/commits/54d9d776e90c0606810164dbdb48cb0810506096 and we can see again how CI performs.

Source/Evolve/WarpXEvolve.cpp
EZoni: Same comment as above, I thought this was breaking CI, but I removed all calls in https://github.com/ECP-WarpX/WarpX/pull/3013/commits/263acac6662f99972fe7216625472ddb43113cc8 and we can see again how CI performs. See update in https://github.com/ECP-WarpX/WarpX/pull/3013#discussion_r882824464 (below).

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -472,15 +466,11 @@ WarpX::OneStep_nosub (Real cur_time)
...
         EvolveF(0.5_rt * dt[0], DtType::SecondHalf);
         EvolveG(0.5_rt * dt[0], DtType::SecondHalf);
         EvolveB(0.5_rt * dt[0], DtType::SecondHalf); // We now have B^{n+1}
```
EZoni, line 476: @RemiLehe It seems that we would need to add an extra call to `FillBoundaryB(guard_cells.ng_FieldSolver, WarpX::sync_nodal_points);` here (after the call to `EvolveB`) in order for the CI test `divb_cleaning_3d` to pass without checksum changes. Note that the failure of `divb_cleaning_3d` was caused by the last commit only, https://github.com/ECP-WarpX/WarpX/pull/3013/commits/263acac6662f99972fe7216625472ddb43113cc8. We could have two calls to `FillBoundaryB`, after each call to `EvolveB`, but include the synchronization only in the second call, I believe this works too (but it's probably adding extra communication compared to the previous code base).

ax3l: Discussed offline with @RemiLehe and me: good to reset.

ax3l (approved):
Awesome! :rocket: :sparkles:

Thierry992:
I compiled the new code using nsight system and here are the differences between the old and the new code : 

### Before 
Screen Shot 2022-06-07 at 5 02 26 PM

### After 
Screen Shot 2022-06-07 at 5 02 41 PM

We noticed that the `OverideSync` blocks are no longer here and that the `Evolve::step` is faster with the new code ✨

PR #3015, opened 2022-04-01
Implemnted the BCK fallback for the ECT solver

lgiacome:


lgiacome:
Closed because messed up commit history

PR #2707, opened 2021-12-22
Implement stair-case Yee solver with EB in RZ geometry

RemiLehe:
This implements the Yee solver with EB. The code is similar to the corresponding Cartesian code: in the cells that are covered with EB cells, the E field is simply not updated.

I added a test where a laser is diffracted by a cylindrical object (represented by the EB). The laser is emitted from the bottom, in the plots below:
Figure 80
The test then checks that the position of the first dark ring (i.e. first minimum) corresponds to the theoretical prediction (Airy pattern).

PR #3016, opened 2022-04-01
Implemented the BCK fallback for the ECT solver

### LINKED ISSUE #2258: ECT: Many Irregular Cells
Limitation of the first implementation of the Enlarged Cell Technique (ECT): If many irregular cells intrude the same regular cell, the regular cell might run out of area and the algorithm fails. In this case we fall back to staircasing. (work-around: better resolution)

Follow-up to #1923

cc @lgiacome @RemiLehe

lgiacome:
As discussed offline, we use the [BCK method](https://ieeexplore.ieee.org/document/1638381) to stabilise the few cells which the ECT cannot extend. To test the method, we changed the rotated cube test to cover a case in which the ECT cannot extend some cells by modifying the geometry and the number of cells.

This will fix #2258.


Source/EmbeddedBoundary/WarpXFaceExtensions.cpp
```diff
@@ -218,15 +218,39 @@ WarpX::ComputeFaceExtensions(){
             + std::to_string(N_ext_faces_after_eight_ways(2))
         );
     }
+
+    bool using_bck = false;
+
+    // If any cell could not ne extended we use the BCK method to stabilize them
```
EZoni:
```suggestion
    // If any cell could not be extended we use the BCK method to stabilize them
```

Source/EmbeddedBoundary/WarpXFaceExtensions.cpp
```diff
@@ -658,6 +682,38 @@ WarpX::ComputeEightWaysExtensions() {
 #endif
 }
 
+void
+WarpX::ApplyBCKCorrection(const int idim) {
+#if defined(AMREX_USE_EB) and !defined(WARPX_DIM_RZ)
+    auto const &cell_size = CellSize(maxLevel());
```
EZoni:
```suggestion
    std::array<amrex::Real,3> const &cell_size = CellSize(maxLevel());
```

Source/EmbeddedBoundary/WarpXFaceExtensions.cpp
```diff
@@ -658,6 +682,38 @@ WarpX::ComputeEightWaysExtensions() {
...
+
+    const amrex::Real dx = cell_size[0];
+    const amrex::Real dy = cell_size[1];
+    const amrex::Real dz = cell_size[2];
+
+    for (amrex::MFIter mfi(*Bfield_fp[maxLevel()][idim]); mfi.isValid(); ++mfi) {
+
+        amrex::Box const &box = mfi.validbox();
```
EZoni: Do you know if tiling on CPU would be possible here? I'm talking about adding `amrex::TilingIfNotGPU()` as a second argument when you define `mfi` and also replacing `mfi.validbox()` with `mfi.tilebox()`. I don't know if the algorithm would be compatible with tiling. If so, there could be a speedup on CPU (not sure if this could matter or if it would be negligible anyways). See the AMReX documentation [here](https://amrex-codes.github.io/amrex/docs_html/Basics.html#mfiter-and-tiling) for more info on `MFIter` with/without tiling.

Source/EmbeddedBoundary/WarpXFaceExtensions.cpp
```diff
@@ -658,6 +682,38 @@ WarpX::ComputeEightWaysExtensions() {
...
+        auto const &flag_ext_face = m_flag_ext_face[maxLevel()][idim]->array(mfi);
+        auto const &flag_info_face = m_flag_info_face[maxLevel()][idim]->array(mfi);
+        auto const &S = m_face_areas[maxLevel()][idim]->array(mfi);
+        auto const &lx = m_face_areas[maxLevel()][0]->array(mfi);
+        auto const &ly = m_face_areas[maxLevel()][1]->array(mfi);
+        auto const &lz = m_face_areas[maxLevel()][2]->array(mfi);
```
EZoni: Would it be possible to replace these `auto` as well? Something like `amrex::Array4<amrex::Real>` would probably work.

EZoni (changes requested):
Thank you for this PR, Lorenzo! I jump in as a second reviewer. I edited the PR description to enable [automatic linking](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) to the corresponding issue, and left a few minor suggestions and questions.

Source/EmbeddedBoundary/WarpXFaceExtensions.cpp
lgiacome: Thanks a lot for the explanation. The kernel is completely local so tiling should work here. As you said we probably don't save much time, but it's still better to have it for consistency with other parts of the code.

lgiacome:
Thanks a lot for the review @EZoni! And also thanks for adding the automatic linking. I've implemented the changes you suggested 😄

Source/EmbeddedBoundary/WarpXFaceExtensions.cpp
```diff
@@ -218,15 +218,39 @@ WarpX::ComputeFaceExtensions(){
...
+        using_bck = true;
     }
+
+#if !defined(WARPX_DIM_XZ) && !defined(WARPX_DIM_RZ)
     if (N_ext_faces_after_eight_ways(2) > 0) {
-        amrex::Abort(Utils::TextMsg::Err("Some z faces could not be extended"));
+        ApplyBCKCorrection(2);
+        using_bck = true;
     }
+#endif
```
EZoni: A question here, just to make sure I understand this correctly:
- we may apply the BCK correction along x only in 3D (not in 2D, not in RZ);
- we may apply the BCK correction along y in all geometries (what would this do in 2D or RZ, though?);
- we may apply the BCK correction along z only in 3D (not in 2D, not in RZ).

Did I get this right from the code?

lgiacome: Yes, you are right. So the point here is that in RZ and 2D the x- and z- components of the magnetic field are not updated and for this reason we don't need any face extensions and we don't need to do anything in x and z and we can skip the call to `ApplyBCKCorrection`.
In 2D/RZ the BCK correction works in the same way as in 3D which is why I don't do anything special.

I have to say that the ECT+RZ could work in theory but it's not tested (yet) and in a separate PR I could even insert an abort somewhere to avoid this combination (if it's not already forbidden, not sure).

If you have better ideas to handle this I am happy to modify the current implementation.

RemiLehe: Regarding ECT+RZ: good point. 
In fact, even RZ with stair-case is not yet implemented: https://github.com/ECP-WarpX/WarpX/pull/2707

Source/EmbeddedBoundary/WarpXFaceExtensions.cpp
```diff
@@ -218,15 +218,39 @@ WarpX::ComputeFaceExtensions(){
...
+
+    if(WarpX::verbose and using_bck) {
+        amrex::Print() << Utils::TextMsg::Info(
```
RemiLehe: Could you use the function [WarpX::RecordWarning](https://github.com/ECP-WarpX/WarpX/blob/development/Source/WarpX.cpp#L436) instead for this?

Source/EmbeddedBoundary/WarpXFaceExtensions.cpp
```diff
@@ -658,6 +682,38 @@ WarpX::ComputeEightWaysExtensions() {
 #endif
 }
 
+void
+WarpX::ApplyBCKCorrection(const int idim) {
+#if defined(AMREX_USE_EB) and !defined(WARPX_DIM_RZ)
```
RemiLehe, line 684:
```suggestion
#if defined(AMREX_USE_EB)
```
How about allowing this for RZ for now? (even though RZ does not support EB for the moment)
Otherwise, I am afraid we may forget about this part of the code, once we add RZ support...

Source/EmbeddedBoundary/WarpXFaceExtensions.cpp
RemiLehe: In this case, you could also probably remove the `if` condition on `WarpX::verbose`.

RemiLehe (approved):

RemiLehe (changes requested):
Thanks for this PR @lgiacome.
I added a few inline suggestions.

Source/WarpX.H
```diff
@@ -1427,6 +1427,12 @@ public:
     * \brief Do the eight-ways extension
     */
     void ComputeEightWaysExtensions();
+    /**
+    * \brief Whenever an unstable cell cannot be extended we increase its area to be the minimal for stability.
+    *        This is the method Benkler-Chavannes-Kuster method and it is less accurate than the regular ECT but it
+    *        still works better than staircasing. (see https://ieeexplore.ieee.org/document/1638381)
```
ax3l, line 1433:
```suggestion
    *        still works better than staircasing. (see https://ieeexplore.ieee.org/document/1638381)
    *
    * @param idim (is this a vector component of B or so? pls doc me :) )
```

lgiacome: Ah yeah 😇

Source/EmbeddedBoundary/WarpXFaceExtensions.cpp
lgiacome: Unfortunately the warning makes the CI fail. Is there a way to fix this?

EZoni: Yes, you can add `runtime_params = warpx.abort_on_warning_threshold=medium` to the entry corresponding to the test `embedded_boundary_rotated_cube` in Regression/WarpX-tests.ini. It will overwrite what we set here:
https://github.com/ECP-WarpX/WarpX/blob/5ea23bf77ad1f6e599769baf56b279c953cac01b/Regression/prepare_file_ci.py#L71-L79

lgiacome: Amazing! It works 😄

RemiLehe (approved):

EZoni (approved):

PR #3019, opened 2022-04-04
LoadBalanceCosts: num_cells & num_macro_particles

ax3l:
Add two new columns to the reduced diagnostics `LoadBalanceCosts` (LBC): `num_cells` & `num_macro_particles` for each box.

- [x] test
- [x] [script/example update](https://warpx.readthedocs.io/en/latest/usage/workflows/plot_distribution_mapping.html)

### DIFF
--- a/Source/Diagnostics/ReducedDiags/LoadBalanceCosts.cpp
+++ b/Source/Diagnostics/ReducedDiags/LoadBalanceCosts.cpp
@@ -7,6 +7,7 @@
 #include "LoadBalanceCosts.H"
 
 #include "Diagnostics/ReducedDiags/ReducedDiags.H"
+#include "Particles/MultiParticleContainer.H"
 #include "Utils/IntervalsParser.H"
 #include "Utils/WarpXAlgorithmSelection.H"
 #include "WarpX.H"
@@ -30,8 +31,38 @@
 #include <iomanip>
 #include <istream>
 #include <memory>
+#include <utility>
 
 using namespace amrex;
+
+namespace
+{
+    amrex::Long
+    countBoxMacroParticles (amrex::MFIter const & mfi, int const lev)
+    {
+        int gid = mfi.index();
+        int tid = mfi.LocalTileIndex();
+        auto box_index = std::make_pair(gid, tid);
+
+        auto & warpx = WarpX::GetInstance();
+        MultiParticleContainer const & mpc = warpx.GetPartContainer();
+        int const nSpecies = mpc.nSpecies();
+
+        amrex::Long num_macro_particles = 0;
+        for (int i_s = 0; i_s < nSpecies; ++i_s)
+        {
+            WarpXParticleContainer const & pc = mpc.GetParticleContainer(i_s);
+            auto const & plev  = pc.GetParticles(lev);
+
+            auto const & ptile = plev.at(box_index);
+            auto const & aos   = ptile.GetArrayOfStructs();
+            auto const np = aos.numParticles();
+            num_macro_particles += np;
+        }
+
+        return num_macro_particles;
+    }
+}
 
 // constructor
 LoadBalanceCosts::LoadBalanceCosts (std::string rd_name)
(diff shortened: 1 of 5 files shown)


atmyers (approved):

Source/Diagnostics/ReducedDiags/LoadBalanceCosts.cpp
```diff
@@ -112,9 +143,12 @@ void LoadBalanceCosts::ComputeDiags (int step)
 #else
             m_data[shift_m_data + mfi.index()*m_nDataFields + 5] = 0.;
 #endif
+            m_data[shift_m_data + mfi.index()*m_nDataFields + 6] = tbx.d_numPts(); // note: difference to volume
```
kngott, line 146: Do we want tile box, or grown tile box? I would think the data included in ghost cells could also be relevant, but it's a diagnostic, so that's up to interpretation.

ax3l: I think for load balance, it's good to get the associated memory per box. I picked `numPts` because my understanding is that this is closer to real memory used than volume. Is that correct?

For the points in the entries above, I think using lo without the grown box is less confusing (e.g., plotting), but using real particle/cell memory size in these two new columns is what we want - which would be the grown tile box.

kngott: Yeah, I would agree that you probably want to dump the tile box, not the grown box, as the description. You can grab both at the top and use as desired, as well. 🤷 

I believe num_pts is correct, yes. 

Also, if you want the raw memory of the grid points, you can look up the size of the data in the FAB at each point with something like: `const std::size_t type_size = sizeof(typename FAB::value_type);`, and then multiply. Depends on what you specifically want to output.

ax3l: Thanks for checking & confirming!

Yes, memory as well would be a good call, but I think we need to gather all multifabs that we have currently active for that (with EB, extra buffers for some algorithms, etc.), which would need a few more helpers. Maybe something we can add as well in the future, definitely helpful.

Tools/PostProcessing/plot_distribution_mapping.py
```diff
@@ -69,19 +69,24 @@ def _get_costs_reduced_diagnostics(self, directory, prange):
...
+        #           k_low_box_1, num_cells_1, num_macro_particles_1,
+        #           (, gpu_ID_box_1 if GPU run), hostname_box_1,
+        #           num_macro_particles_n
         #      ...
         #      cost_box_n, proc_box_n, lev_box_n, i_low_box_n, j_low_box_n,
-        #           k_low_box_n(, gpu_ID_box_n if GPU run), hostname_box_n
+        #           k_low_box_n, num_cells_n, num_macro_particles_n,
+        #           (, gpu_ID_box_n if GPU run), hostname_box_n,
+        #           num_macro_particles_n
         i, j, k = (data[0,3::n_data_fields],
```
ax3l, line 90: Even in the original part, I think this code here has an off-by-one error and should start at column `2:`

Tools/PostProcessing/plot_distribution_mapping.py
```diff
@@ -69,19 +69,22 @@ def _get_costs_reduced_diagnostics(self, directory, prange):
             unique_headers=[''.join([l for l in w if not l.isdigit()])
                             for w in h.split()][2::]
 
-        # Either 7 or 8 depending if GPU
-        n_data_fields = 7 if (len(set(unique_headers)) - 2)%7 == 0 else 8
```
ax3l, line 73: Since we remove the first two columns already two lines above, this is likely wrong already in the original code and should not have `- 2`

ax3l: fixed in this PR

Tools/PostProcessing/plot_distribution_mapping.py
```diff
@@ -102,8 +111,6 @@ def _get_costs_reduced_diagnostics(self, directory, prange):
         j_is_int = all([el.is_integer() for el in j/j_blocking_factor])
         k_is_int = all([el.is_integer() for el in k/k_blocking_factor])
```
ax3l, line 112: No-unused vars fixed in de9b317d70770b4257472c1771f57da130aa7da4

PR #3022, opened 2022-04-04
Minor fixes removing warnings from MSVC C++

dpgrote:
While fixing some other things, I noticed the MSVC C++ is showing several warnings. This PR fixes some of the issues. It is also showing several other warnings in the `FlushFormat` code, but I can't tell what it is unhappy about.

As an aside, MSVC C++ is also showing a large number of warnings in the openPMD code and several in AMReX code.


ax3l (approved):
Thank you!

Yes, let's more consequently use named values for our `bool`s, we have them for 23years now :)

PR #3024, opened 2022-04-04
Semicoarsening in electrostatic solvers

WeiqunZhang:
Enable semicoarsening in electrostatic solvers if the problem is anisotropic
due to relativistic beam and/or anisotropic cell spacing.


RemiLehe (approved):

RemiLehe:
Thanks a lot!

PR #3025, opened 2022-04-05
Added PlasmaLens class to PICMI

dpgrote:
The plasma lens lattice can now be defined using PICMI. Note that this class is specific to Warp and is not defined in the PICMI standard.

Other changes were made to clean up issues.
- For the CI tests, the cfl parameter was moved from `Regression/WarpX-tests.ini` to the input files so that they will be consistent with the PICMI input file.
- Made a fix in `ParticleListDistribution` allowing the particle velocity input to be lists.


Regression/WarpX-tests.ini
```diff
@@ -2992,6 +2992,26 @@ compareParticles = 1
...
+restartTest = 0
+useMPI = 1
+numprocs = 2
+useOMP = 1
+numthreads = 1
+compileTest = 0
+doVis = 0
+compareParticles = 1
+particleTypes = electrons
+analysisRoutine = Examples/Tests/plasma_lens/analysis.py
```
ax3l, line 3013: We looked at this (@EZoni @RemiLehe ) and we think we either need to add `outputFile = ...` here or above add:
```
write_dir = '.',
warpx_file_prefix = 'Python_plasma_lens_plt'
```
in the diags of the PICMI file.

dpgrote: Thanks! I added the lines to the PICMI file.

Examples/Tests/plasma_lens/PICMI_inputs_3d.py
```diff
@@ -0,0 +1,89 @@
...
+                                     method = 'Yee',
+                                     cfl = 0.7)
+
+# Diagnostics
+part_diag1 = picmi.ParticleDiagnostic(name = 'diag1',
+                                      period = max_steps,
+                                      species = [electrons],
+                                      data_list = ['ux', 'uy', 'uz'],
+                                      write_dir = '.',
+                                      warpx_file_prefix = 'Python_plasma_lens')
```
EZoni: I think `_plt` needs to be added to the file prefix, otherwise the test still fails:
```suggestion
                                      warpx_file_prefix = 'Python_plasma_lens_plt')
```

EZoni (approved):
Thank you for this PR, Dave! Looks good to me.

PR #3028, opened 2022-04-06
Add flag for whether a terminate signal was received

roelof-groenewald:
Also allow that flag to be read by python. This is useful for us to handle all relevant clean-up within python.

### DIFF
--- a/Python/pywarpx/_libwarpx.py
+++ b/Python/pywarpx/_libwarpx.py
@@ -275,6 +275,7 @@ class LibWarpX():
         self.libwarpx_so.warpx_getdt.restype = c_real
         self.libwarpx_so.warpx_maxStep.restype = ctypes.c_int
         self.libwarpx_so.warpx_stopTime.restype = c_real
+        self.libwarpx_so.warpx_terminateFlag.restype = ctypes.c_bool
         self.libwarpx_so.warpx_finestLevel.restype = ctypes.c_int
         self.libwarpx_so.warpx_getMyProc.restype = ctypes.c_int
         self.libwarpx_so.warpx_getNProcs.restype = ctypes.c_int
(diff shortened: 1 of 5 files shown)


ax3l:
Thx!

Can you elaborate a bit when we might need this as a workflow? Wouldn't we shut down interactive simulations by just not doing more steps - and non-interactive ones with signals?

I wonder if we can just shut down from Python as well by setting max_step (or stop_time) to a lower value.
This could be done also in a helper-function, sparing an additional member variable (an additional state) & logic to check.

Just asking this, because I wonder how many ways to shut down a simulation we need :D

roelof-groenewald:
We sometimes want to run simulations until they reach some form of a steady state - usually with regards to particle fluxes onto boundaries. To do this we have a Python function (lets say `converged()`) that looks at the last `n` simulation steps and returns `True` if it decides the flux changes during those steps were small enough to consider the simulation in steady state. We then drive the simulation from Python as:
```
while not converged():
    simulation.step(n)
```
With the signal handling capabilities in WarpX we now need to check whether the simulation loop was terminated due to reaching `n` steps or because it received an interrupt signal, so the above becomes:
```
while not converged() and not terminate_signal:
    simulation.step(n)
```

Thanks for the push back though. We have decided we will change our approach so that we don't need this PR anymore. Instead we will only call the WarpX loop once (using a large number of total steps) and use a `callback` to check convergence every `n` steps. If convergence is reached we will use the signal handling to terminate the simulation by letting Python send an interrupt signal to it's own process (`os.kill(os.getpid(), signal.SIGTERM)`).

So I'll close this PR.

ax3l:
Sounds good, thanks for the details @roelof-groenewald :sparkling_heart:

PR #2896, opened 2022-02-26
Allow process signals to trigger checkpoint or break behavior

PhilMiller:

PR #3029, opened 2022-04-06
Fix MPI Signal Handling on Perlmutter

ax3l:
Follow-up to #2896. Not part of release `22.04`.
First seen by @RTSandberg and @Yin-YinjianZhao.

Thx to @kngott and Brandon for rapid NERSC consulting support.

## Fix: `MPI_C(XX)_BOOL`

C99 types were aded in MPI-2.2, while Cray's MPICH fork in version 8.1.13 defines `MPI_CXX_BOOL` to `MPI_DATATYPE_NULL` on Perlmutter.

We could use `MPI_C_BOOL`, which is technically a `_Bool` from [<cstdbool>](https://en.cppreference.com/w/cpp/header/cstdbool) (deprecated: C++17; removed: C++20) - or we simply do a static assert on `sizeof(bool)` and communicate as a `MPI_BYTE` or `MPI_CHAR`.

## Do not Broadcast if no action is registered 

Do no MPI Comms if no action is configured.

### DIFF
--- a/Source/ablastr/utils/SignalHandling.cpp
+++ b/Source/ablastr/utils/SignalHandling.cpp
@@ -21,6 +21,7 @@
 
 namespace ablastr::utils {
 
+bool SignalHandling::m_any_signal_action_active = false;
 std::atomic<bool> SignalHandling::signal_received_flags[NUM_SIGNALS];
 bool SignalHandling::signal_conf_requests[SIGNAL_REQUESTS_SIZE][NUM_SIGNALS];
 bool SignalHandling::signal_actions_requested[SIGNAL_REQUESTS_SIZE];
@@ -29,7 +30,7 @@ MPI_Request SignalHandling::signal_mpi_ibcast_request;
 #endif
 
 int
-SignalHandling::parseSignalNameToNumber(const std::string &str)
+SignalHandling::parseSignalNameToNumber (const std::string &str)
 {
     amrex::IParser signals_parser(str);
 
@@ -110,20 +111,8 @@ SignalHandling::parseSignalNameToNumber(const std::string &str)
     return sig;
 }
 
-bool
-SignalHandling::anySignalConfActive ()
-{
-    bool any_signal_active = false;
-    for (int signal_number = 0; signal_number < NUM_SIGNALS; ++signal_number) {
-        for (int signal_request = 0; signal_request < SIGNAL_REQUESTS_SIZE; ++signal_request) {
-            any_signal_active |= signal_conf_requests[signal_request][signal_number];
-        }
-    }
-    return any_signal_active;
-}
-
 void
-SignalHandling::InitSignalHandling()
+SignalHandling::InitSignalHandling ()
 {
 #if defined(__linux__) || defined(__APPLE__)
     struct sigaction sa;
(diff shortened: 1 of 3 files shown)


Source/ablastr/utils/SignalHandling.cpp
```diff
@@ -140,6 +152,10 @@ SignalHandling::InitSignalHandling()
 void
 SignalHandling::CheckSignals()
 {
+    // Do not perform MPI communication if no signal action is configured
```
PhilMiller: The comment here threw me off - I was going to remark that this check should come after the testing of signal flags, lest signal handling be disabled in the single-process case. I had to think to realize that this is itself a check whether any signal handling is enabled. Please reword the comment

Source/ablastr/utils/SignalHandling.cpp
```diff
@@ -162,14 +178,19 @@ SignalHandling::CheckSignals()
 
 #if defined(AMREX_USE_MPI)
     auto comm = amrex::ParallelDescriptor::Communicator();
+    static_assert(sizeof(bool) == 1, "We communicate bools as 1 byte-sized type in MPI");
```
PhilMiller, line 179: Please comment that this is an accommodation for Cray's deficient MPI implementation. They support the MPI 3 asynchronous collective, but not the simple datatype definition.

Source/ablastr/utils/SignalHandling.cpp
```diff
@@ -110,6 +110,18 @@ SignalHandling::parseSignalNameToNumber(const std::string &str)
     return sig;
 }
 
+bool
+SignalHandling::anySignalConfActive ()
+{
+    bool any_signal_active = false;
```
PhilMiller: Should this maybe be a member variable of `SignalHandling`, that can be set just once, rather than looping over everything twice every step?

PhilMiller:
Code and updated documentation look fine to me.

Source/ablastr/utils/SignalHandling.cpp
ax3l: Yeah, it's a pitty!
Will add the same description as in the PR to make clear why we hack back to bytes

Source/ablastr/utils/SignalHandling.cpp
ax3l: that probably makes sense, yes.
I was coding this in a way  that one could register signals later on, e.g., from the Python layer. But not sure if that's truly needed.

We set that variable at the end of the config step, so even if someone configures it later, it would be set consistently.

Source/ablastr/utils/SignalHandling.cpp
ax3l: Thx, rephrased :)

ax3l:
@PhilMiller thank you for the review, updated accordingly :)

Source/ablastr/utils/SignalHandling.cpp
```diff
@@ -134,12 +135,23 @@ SignalHandling::InitSignalHandling()
                                                "Failed to install signal handler for a configured signal");
         }
     }
+
+    for (int signal_number = 0; signal_number < NUM_SIGNALS; ++signal_number) {
+        for (int signal_request = 0; signal_request < SIGNAL_REQUESTS_SIZE; ++signal_request) {
+            m_any_signal_action_active |= signal_conf_requests[signal_request][signal_number];
```
PhilMiller, line 141: You can just set this to true at line 128

PhilMiller (approved):
Up to you whether to do the last suggested simplification.

RemiLehe (approved):

Source/ablastr/utils/SignalHandling.cpp
ax3l: Hm, cannot follow, do you mean line 138?
Even then I am not sure - if init handling is called but no conf request was passed, then I still want this to be false, no?

PhilMiller: I was suggesting inside the `if (signal_active) {` block - i.e. exactly where we're already checking the array entries

ax3l: ah, of course. good idea :)

PR #3030, opened 2022-04-07
RZ + openPMD : don't transpose data, save data with shape (Nm, Nz, Nr)

RTSandberg:
Fixing PR #2878 
--
Code transposition is turned off
Data is stored in the shape (Nm, Nz, Nr) rather than (Nm, Nr, Nz).

In PR #2878, a choice was made to transpose data on CPU to adhere to the openPMD standard requiring data in the shape (Nm, Nr, Nz).  However, the data shape was incorrect for the GPU untransposed case.  Discussing with the OpenPMD developers, the OpenPMD standard will be relaxed to allow data of the shape (Nm, Nz, Nr); see https://github.com/openPMD/openPMD-viewer/pull/337  This means no transposing is necessary


ax3l:
Hi @RTSandberg, thanks a lot.

In order to finalize this, can you please link the PR to openPMD-viewer that handles the data reconstruction depending on the value in `axisLabels` for RZ accordingly? (time sensitive, since @camille12225 depends on this)

We also need to update the openPMD standard together to document this relaxation :)
(less timely, e.g., next week also ok)

RemiLehe:
Thanks for this PR @RTSandberg !
Since we are not doing the transposition anymore, would it make sense to clean up the variable `do_transpose`, and remove the transposition functions (`transposeChunk`) from the code altogether?

RTSandberg:
@RemiLehe Yes, I can remove `do_transpose` and refactor.  I wanted to make sure everyone is ok with this decision before removing the transposing code

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -274,35 +274,46 @@ namespace detail
...
-        // if we are start to write individual modes
-        vs const fieldComponents{"r", "t", "z"};
-        // if we just write reconstructed fields at theta=0
-        //vs const fieldComponents{"x", "y", "z"};
-#else
-        // note: 1D3V and 2D3V simulations still have 3 components for the fields
-        vs const fieldComponents{"x", "y", "z"};
-#endif
-        return fieldComponents;
+// #if defined(WARPX_DIM_RZ)
```
RemiLehe:
```suggestion
```

RemiLehe (approved):

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -1459,20 +1414,8 @@ WarpXOpenPMDPlot::WriteOpenPMDFieldsAll ( //const std::string& filename,
 #endif
                 {
                     amrex::Real const *local_data = fab.dataPtr(icomp);
-#ifdef WARPX_DIM_RZ
```
ax3l, line 1462: Note: this block was only taking the CPU path, and we forgot to transpose on GPU. With this PR (+ https://github.com/openPMD/openPMD-viewer/pull/337), we don't need to anymore anyway :)

ax3l (approved):
LGTM, thank you! :sparkles: 
I added a little clean-up.

PR #3036, opened 2022-04-11
Add a python script to generate timestep duration plots

lucafedeli88:
This PR adds a script in the `Tools` folder to generate timestep duration plots like:
image
and
image
from the standard output of WarpX. The PR also adds a page in the documentation.


Tools/PostProcessing/plot_timestep_duration.py
```diff
@@ -0,0 +1,88 @@
...
+import re
+import sys
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+
+def extract_data(filename):
+
+    regex_step = re.compile(
```
dpgrote, line 12: Using `re` is good here. Just to comment, when I do this kind of thing, I use more basic string manipulation (which I find easier to write than composing an `re` string), like:
```py
line = f.read()
if line.startswith('Evolve time'):
    sline = line.split(' ')
    time_data.append(float(sline[8])
```

Tools/PostProcessing/plot_timestep_duration.py
```diff
@@ -0,0 +1,88 @@
...
+    ax.spines['top'].set_visible(False)
+    ax.spines['left'].set_color('gray')
+    ax.spines['right'].set_visible(False)
+
+    plt.tight_layout()
+
+    plt.savefig(name + "_cumulative_duration.png", transparent=False, dpi=300)
+    print("...done!")
+
+
```
dpgrote, line 84: Perhaps, put these lines inside an `if` block, `if __name__ == "__main__":`. This would allow a user to import this file to use the `extract_data` function to get the time data if they would want to do some other kind of plot or analysis.

dpgrote:
This is a nice idea - thanks! I have some comments.

Docs/source/dataanalysis/plot_timestep_duration.rst
```diff
@@ -0,0 +1,19 @@
+.. _analyze-timestep-duration:
```
ax3l: Thanks for the docs!
To finalize this, please include it in a table of content, e.g., here
- https://warpx.readthedocs.io/en/latest/usage/workflows.html
  - source: `Docs/source/usage/workflows.rst`

Tools/PostProcessing/plot_timestep_duration.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env python3
```
ax3l, line 1: Just double checking: file has `chmod a+x` permissions? :)

Tools/PostProcessing/plot_timestep_duration.py
ax3l: Hm, I think the regrex here is more expressive than this (e.g., shows exact structure of what we expect, no magic number, multiple lines, etc.) :)
I would keep the regex.

Tools/PostProcessing/plot_timestep_duration.py
```diff
@@ -0,0 +1,88 @@
...
+    ax.spines['left'].set_color('gray')
+    ax.spines['right'].set_visible(False)
+
+    plt.tight_layout()
+
+    plt.savefig(name + "_cumulative_duration.png", transparent=False, dpi=300)
+    print("...done!")
+
+
+output_log_file_name = sys.argv[1]
```
ax3l: Please use `argparse`, which will automatically document the positional command-line argument that is required for users on the CLI
https://docs.python.org/3/library/argparse.html

Tools/PostProcessing/plot_timestep_duration.py
lucafedeli88: ah no!

lucafedeli88: I have to add the permissions!

Tools/PostProcessing/plot_timestep_duration.py
lucafedeli88: I will

Tools/PostProcessing/plot_timestep_duration.py
lucafedeli88: done

Tools/PostProcessing/plot_timestep_duration.py
lucafedeli88: done

Tools/PostProcessing/plot_timestep_duration.py
lucafedeli88: done!

Docs/source/dataanalysis/plot_timestep_duration.rst
lucafedeli88: It should be done!

lucafedeli88:
Thanks for the feedback and the advice. I've implemented all the suggested corrections except for the point regex vs string manipulations. I have no strong opinions on this point, both solutions are OK for me.

dpgrote (approved):
Looks good, thanks!

Pondering this, perhaps it would be useful to save this timing data and write it out as part of the plot file. The same thing for the tiny profile data. Is any of this already saved in the plot files?

ax3l:
Probably a question for @WeiqunZhang ? :)

Docs/source/usage/workflows.rst
```diff
@@ -11,6 +11,7 @@ This section collects typical user workflows and best practices for WarpX.
    workflows/parallelization
    workflows/debugging
    workflows/libensemble
+   workflows/plot_timestep_duration.rst
```
ax3l: not needed
```suggestion
   workflows/plot_timestep_duration
```

ax3l (changes requested):
The `workflows/plot_timestep_duration.rst` file is missing in the PR

Docs/source/usage/workflows.rst
WeiqunZhang: @dpgrote https://github.com/ECP-WarpX/WarpX/pull/3036#pullrequestreview-939783903
> Pondering this, perhaps it would be useful to save this timing data and write it out as part of the plot file. The same thing for the tiny profile data. Is any of this already saved in the plot files?

Neither timing data nor tiny profile data is saved in the plot files currently.  For the time step duration data, I assume it won't be hard to do it.  But it might not be easy to do that for tiny profile data, because currently amrex::TinyProfiler only supports printing the data at the end of the run.

lucafedeli88:
> The `workflows/plot_timestep_duration.rst` file is missing in the PR

You are right! Somehow I have deleted it!

lucafedeli88:
> > The `workflows/plot_timestep_duration.rst` file is missing in the PR
> 
> You are right! Somehow I have deleted it!

I've added it back!

dpgrote (approved):
Looks good! Thanks!

Docs/source/usage/workflows/plot_timestep_duration.rst
```diff
@@ -0,0 +1,18 @@
+.. _plot-timestep-duration:
+
+Plot timestep duration
+=========================
```
ax3l: rst is a bit picky rendering this:
```suggestion
Plot timestep duration
======================
```

Docs/source/usage/workflows/plot_timestep_duration.rst
```diff
@@ -0,0 +1,18 @@
...
+We provide a simple python script to generate plots of the timestep duration
+from the stdandard output of WarpX (provided that ``warpx.verbose`` is set to 1):
+`plot_timestep_duration.py<../../../../Tools/PostProcessing/plot_timestep_duration.py>`__ .
```
ax3l: readability:
```suggestion
`plot_timestep_duration.py <../../../../Tools/PostProcessing/plot_timestep_duration.py>`__ .
```

ax3l (approved):
Cool, thanks a lot! :)

Issue #3027, opened 2022-04-06
Optimize Synchronization of Nodal Points


EZoni:
Here's a few points that need to be addressed in order to optimize the synchronization of nodal points in WarpX:
- [x] Merge synchronization of nodal points into `FillBoundary`:
  - [x] `OneStep_nosub`
  PR: #3013
  - [x] `OneStep_sub1`
  PR: #3013
  - [x] `OneStep_multiJ`
  PR: #3169
- [x] Merge and move synchronization of nodal points for PML fields as well
  PR: #3534
- [x] Check if we can remove the synchronization of nodal points in J/rho (historically, we first implemented the synchronization of J/rho, and then that of E/B: https://github.com/ECP-WarpX/WarpX/pull/1817. However, after adding the synchronization of E/B, intuitively it seems that synchronizing J/rho is not needed anymore.)
PR: #3038
- [x] Clean up or remove the functions `NodalSync`
PR: #3169

PR #3038, opened 2022-04-11
Remove Nodal Synchronization of J and Rho

EZoni:
Checking if we can remove the nodal synchronization of J and rho, as suggested in #3027.

The only CI test that seems to need a benchmark reset is `reduced_diags_single_precision`.


RemiLehe (approved):

RemiLehe:
Thanks! This should make the code a bit faster :)

PR #2364, opened 2021-09-30
Python Script to Compute PSATD Stencil

EZoni:
Add a Python script `Stencil.py` to compute the minimum number of guard cells for a given error threshold (default single precision, 1e-07), based on the measurement of the PSATD stencil extent. Currently implemented only in 3D (an older 2D version I had seemed to flip the x and z directions for reasons I have not yet been able to understand).

I added the script to `Tools/DevUtils`, not sure if this is the best place. Any preferences (an alternative could be `Source/Utils`, since this is kind of an algorithmic tool)?

I also flagged this PR with the label "component: post-processing" because that is the closest label I found (even though this is more of a pre-processing tool).

The code is documented with docstrings. Given the number of possible input arguments it is probably better to set those directly within the script and make the main function a callable function (instead of using a parser).

To-do:
- [x] Replace parser with new main callable function;
- [x] Clean up output, make more variables available.

Here's an example of a possible output:
```
$ python Stencil.py

 >> Numerical Setup
    ---------------

    cell size:
    - dx = 1e-06
    - dy = 1e-06
    - dz = 2e-06
    - dz/dx = 2
    - dz/dy = 2

    time step:
    - dt = 1e-14
    - c*dt/dx = 2.99792
    - c*dt/dy = 2.99792
    - c*dt/dz = 1.49896

    spectral order:
    - nox = 8
    - noy = 8
    - noz = 16

    Lorentz boost, Galilean velocity:
    - gamma = 30
    - v_gal = -2.99626e+08

 >> Compute Stencil
    ---------------

    error threshold:
    - ex = 1e-07
    - ey = 1e-07
    - ez = 1e-07

    nodal solver:
    - 21 guard cells along x
    - 21 guard cells along y
    - 19 guard cells along z

    staggered solver:
    - 10 guard cells along x
    - 10 guard cells along y
    - 14 guard cells along z

 >> Output Summary
    --------------

    path to figures: '/home/edoardo/Repo/LBL/WarpX/Tools/DevUtils'

    stencil arrays:
    - sx_nodal
    - sx_stagg
    - sy_nodal
    - sy_stagg
    - sz_nodal
    - sz_stagg
```

PR #3055, opened 2022-04-20
`Stencil.py`: Follow PEP 8 Coding Style

EZoni:
Just making the Python script `Stencil.py` compliant with [PEP 8 style guide for Python code](https://peps.python.org/pep-0008/).

Mainly fixing: indentation, maximum line length, whitespace in expressions and statements.

It's not difficult to follow those coding conventions, especially for newly added Python code (like the script here, added by myself in #2364): seems to me like an easy way to keep setting a good example, so why not?

### DIFF
--- a/Tools/DevUtils/Stencil.py
+++ b/Tools/DevUtils/Stencil.py
@@ -1,12 +1,15 @@
 """
-Python script to compute the minimum number of guard cells for a given error threshold,
-based on the measurement of the PSATD stencil extent (that is, the minimum number of
-guard cells such that the stencil measure is not larger than the error threshold).
+Python script to compute the minimum number of guard cells for a given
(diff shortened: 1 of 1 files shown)


dpgrote (approved):
This looks good - I can approve.
Though my own personal opinion is that the 79 character line length is way too short. I prefer around 110 to 130 maximum. For me, shorter lines are too choppy and are harder to parse as a whole.
Also, in several places you use the backslash to break up lines of code. It is better to wrap the equations in parenthesis to take advantage of Python's implied line continuation.

ax3l:
Agreed, I am also more on the 100 character per line front these days.
If we want to add auto-formatting for python scripts, then we could add/overwrite that detail :)

ax3l (approved):

PR #3056, opened 2022-04-21
Bug fix for BTD - particle BA, and geom, same as field buffer

RevathiJambunathan:
This PR ensures that the BoxArray and geometry object used for particle buffer is same as that for the field buffer. 
This PR solves part of the problem with Issue #2746 i.e., particles are redistributed correctly to the box array for a single-level simulation. 
Here is the plotfile result - I have slightly modified the input file from the Issue to ensure its single level and could be tested with plotfiles, and changed prob-hi in z-direction to be ==0
 (Work is in progress to test single-level OpenPMD output -- this should ideally just work.)

Screenshot from 2022-04-20 16-25-44
input file : 
[inputs.3d.txt](https://github.com/ECP-WarpX/WarpX/files/8526196/inputs.3d.txt)

However, the code still crashes for the same input file and test with mesh-refinement. 
I am discussing this offline with @atmyers to understand how to ensure particle container for BTD does not use finestLevel from the simulation, but from the diagnostics.
**Update : This is fixed after @atmyers merged a PR to amrex https://github.com/AMReX-Codes/amrex/pull/2732** 
**The code now works without and with mesh refinement**

Also, the Prob hi in the ZINDEX direction must be 0.0 for cases with and without mesh-refienement. 
Added an ASSERT in the code to crash with error message. 
Discussing this offline with @RemiLehe and @atmyers to see how to incorporate an offset for Lo,Hi in the lab-frame.
**Update : This is fixed as well.**

### DIFF
--- a/Examples/Modules/RigidInjection/inputs_2d_BoostedFrame
+++ b/Examples/Modules/RigidInjection/inputs_2d_BoostedFrame
@@ -69,6 +69,7 @@ btd_pltfile.dt_snapshots_lab = 1.8679589331096515e-13
 btd_pltfile.fields_to_plot = Ex Ey Ez Bx By Bz jx jy jz rho
 btd_pltfile.format = plotfile
 btd_pltfile.buffer_size = 32
+btd_pltfile.write_species = 0
 
 # old BTD diagnostics
 warpx.do_back_transformed_diagnostics = 1
--- a/Examples/Modules/RigidInjection/inputs_2d_BoostedFrame
+++ b/Examples/Modules/RigidInjection/inputs_2d_BoostedFrame
@@ -58,7 +58,7 @@ btd_openpmd.do_back_transformed_fields = 1
 btd_openpmd.num_snapshots_lab = 2
 btd_openpmd.dt_snapshots_lab = 1.8679589331096515e-13
 btd_openpmd.fields_to_plot = Ex Ey Ez Bx By Bz jx jy jz rho
-btd_openpmd.format = plotfile
+btd_openpmd.format = openpmd
 btd_openpmd.openpmd_backend = h5
 btd_openpmd.buffer_size = 32
 
--- a/Examples/Modules/RigidInjection/inputs_2d_BoostedFrame
+++ b/Examples/Modules/RigidInjection/inputs_2d_BoostedFrame
@@ -49,7 +49,7 @@ beam.zinject_plane = 20.e-6
 beam.rigid_advance = true
 
 # Diagnostics
-diagnostics.diags_names = diag1 btd_openpmd btd_pltfile
+diagnostics.diags_names = diag1 btd_pltfile
 diag1.intervals = 10000
 diag1.diag_type = Full
 
@@ -58,7 +58,7 @@ btd_openpmd.do_back_transformed_fields = 1
 btd_openpmd.num_snapshots_lab = 2
 btd_openpmd.dt_snapshots_lab = 1.8679589331096515e-13
 btd_openpmd.fields_to_plot = Ex Ey Ez Bx By Bz jx jy jz rho
-btd_openpmd.format = openpmd
+btd_openpmd.format = plotfile
 btd_openpmd.openpmd_backend = h5
 btd_openpmd.buffer_size = 32
 
@@ -69,7 +69,7 @@ btd_pltfile.dt_snapshots_lab = 1.8679589331096515e-13
 btd_pltfile.fields_to_plot = Ex Ey Ez Bx By Bz jx jy jz rho
 btd_pltfile.format = plotfile
 btd_pltfile.buffer_size = 32
-btd_pltfile.write_species = 0
+btd_pltfile.write_species = 1
 
 # old BTD diagnostics
 warpx.do_back_transformed_diagnostics = 1
--- a/Examples/Modules/RigidInjection/inputs_2d_BoostedFrame
+++ b/Examples/Modules/RigidInjection/inputs_2d_BoostedFrame
@@ -49,7 +49,7 @@ beam.zinject_plane = 20.e-6
 beam.rigid_advance = true
 
 # Diagnostics
-diagnostics.diags_names = diag1 btd_pltfile
+diagnostics.diags_names = diag1 btd_pltfile btd_openpmd
 diag1.intervals = 10000
 diag1.diag_type = Full
 
--- a/Examples/Modules/boosted_diags/inputs_3d_slice
+++ b/Examples/Modules/boosted_diags/inputs_3d_slice
@@ -10,7 +10,7 @@ my_constants.ymin = -128.e-6
 my_constants.zmin = -40.e-6
 my_constants.xmax = +128.e-6
 my_constants.ymax = +128.e-6
-my_constants.zmax = 0.96e-6
+my_constants.zmax = 0.e-6
 
 geometry.dims = 3
 geometry.prob_lo     = xmin ymin zmin
@@ -133,6 +133,7 @@ btd_pltfile.dz_snapshots_lab = 0.001
 btd_pltfile.fields_to_plot = Ex Ey Ez Bx By Bz jx jy jz rho
 btd_pltfile.format = plotfile
 btd_pltfile.buffer_size = 32
+btd_pltfile.write_species = 0
 
 # old BTD diagnostics
 warpx.do_back_transformed_diagnostics = 1
--- a/Examples/Modules/boosted_diags/inputs_3d_slice
+++ b/Examples/Modules/boosted_diags/inputs_3d_slice
@@ -133,7 +133,7 @@ btd_pltfile.dz_snapshots_lab = 0.001
 btd_pltfile.fields_to_plot = Ex Ey Ez Bx By Bz jx jy jz rho
 btd_pltfile.format = plotfile
 btd_pltfile.buffer_size = 32
-btd_pltfile.write_species = 0
+btd_pltfile.write_species = 1
 
 # old BTD diagnostics
 warpx.do_back_transformed_diagnostics = 1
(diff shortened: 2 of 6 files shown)


Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -141,6 +141,10 @@ BTDiagnostics::ReadParameters ()
         m_crse_ratio == amrex::IntVect(1),
         "Only support for coarsening ratio of 1 in all directions is included for BTD\n"
         );
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+        warpx.Geom(0).ProbHi(WARPX_ZINDEX) <= 0._rt,
+        " The geometry.prob_hi of the boosted-frame domain in the moving window direction must be ==0.\n"
+        );
```
RevathiJambunathan: This ASSERT is to be discussed with @RemiLehe and @atmyers 
This is what works with the way we do BTD as of now, since we don't add an offset when converting from boosted frame to lab-frame co-ordinates.

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -404,7 +408,7 @@ BTDiagnostics::InitializeFieldFunctors (int lev)
     // Fill vector of cell-center functors for all field-components, namely,
     // Ex, Ey, Ez, Bx, By, Bz, jx, jy, jz, and rho are included in the
     // cell-center functors for BackTransform Diags
-    for (int comp=0, n=m_cell_center_functors[lev].size(); comp<n; comp++){
+    for (int comp=0, n=m_cell_center_functors.at(lev).size(); comp<n; comp++){
```
RevathiJambunathan, line 436: this was not part of the bug-fix. but is just cleaner

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -524,6 +528,10 @@ BTDiagnostics::PrepareFieldDataForOutput ()
                         }
                         DefineFieldBufferMultiFab(i_buffer, lev);
                     }
+                    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                        m_current_z_lab[i_buffer] >= m_buffer_domain_lab[i_buffer].lo(m_moving_window_dir) and
+                        m_current_z_lab[i_buffer] <= m_buffer_domain_lab[i_buffer].hi(m_moving_window_dir),
+                        "z-slice in lab-frame is outside the buffer domain physical extent. ");
```
RevathiJambunathan, line 561: Assert added to ensure that the z-slice coordinate is within the bounds of the buffer in the lab-frame

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -1060,33 +1072,43 @@ BTDiagnostics::PrepareParticleDataForOutput()
...
-                        m_particles_buffer[i_buffer][i]->SetParticleGeometry(lev, geom);
+                        if (m_do_back_transformed_fields) {
+                            // use the same Box, BoxArray, and Geometry as fields for particles
+                            amrex::Box particle_buffer_box = m_buffer_box[i_buffer];
+                            amrex::BoxArray buffer_ba( particle_buffer_box );
+                            buffer_ba.maxSize(m_max_box_size);
+                            amrex::DistributionMapping buffer_dmap(buffer_ba);
+                            m_particles_buffer[i_buffer][i]->SetParticleBoxArray(lev, buffer_ba);
+                            m_particles_buffer[i_buffer][i]->SetParticleDistributionMap(lev, buffer_dmap);
+                            m_particles_buffer[i_buffer][i]->SetParticleGeometry(lev, m_geom_snapshot[i_buffer][lev]);
```
RevathiJambunathan, line 1140: if fields are defined, we re-use the BoxArray and Geometry for setting ParticleBoxArray, and Geometry.

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -599,10 +607,14 @@ BTDiagnostics::DefineFieldBufferMultiFab (const int i_buffer, const int lev)
...
-            amrex::Real buffer_hi = m_prob_domain_lab[i_buffer].lo(idim) +
-                                          (buffer_ba.getCellCenteredBox( buffer_ba.size()-1 ).bigEnd(idim) + 1) * cellsize;
+            amrex::Real buffer_lo = m_snapshot_domain_lab[i_buffer].lo(idim)
+                                    + ( buffer_ba.getCellCenteredBox(0).smallEnd(idim)
+                                      - m_snapshot_box[i_buffer].smallEnd(idim)
+                                      ) * cellsize;
+            amrex::Real buffer_hi = m_snapshot_domain_lab[i_buffer].lo(idim)
+                                    + ( buffer_ba.getCellCenteredBox( buffer_ba.size()-1 ).bigEnd(idim)
+                                      - m_snapshot_box[i_buffer].smallEnd(idim)
+                                      + 1 ) * cellsize;
```
RevathiJambunathan, line 645: Here we get the Physical (lo,hi) of the buffer that is filled with slices. The main bug-fix here is to subtract the index offset for the lo-index corresponding to the smallEnd of the entire lab-frame snapshot

RevathiJambunathan:
Update : 
This commit fixes the MR issue 
[8cf9df7](https://github.com/ECP-WarpX/WarpX/pull/3056/commits/8cf9df77cfea2660f10da2d7fbfce3c5b60d8a4d) 
along with @atmyers' PR : https://github.com/AMReX-Codes/amrex/pull/2732 

However, the code crashes with the assert that z_slice_lab is not within the bounds of the buffer domain in lab-frame, but this was not the case for the single-level simulation.

This PR currently fails CI because https://github.com/AMReX-Codes/amrex/pull/2732 PR from @atmyers needs to be merged in amrex, and then updated when compiling+testing the code

Source/Diagnostics/BTDiagnostics.cpp
RevathiJambunathan: update : discussed offline with @RemiLehe 
we are looking into the expressions used to obtain z coordinates in the lab-frame, and checking whether offsets are needed here

RevathiJambunathan:
Update : 
Mesh-refinement works and offset is fixed. This PR is ready for review

ax3l:
@RevathiJambunathan when you have the chance next time, can you please fix the merge issue? :)

RevathiJambunathan:
Temporarily commenting out the assert in [3c1e023](https://github.com/ECP-WarpX/WarpX/pull/3056/commits/3c1e02386ae274e67a58b74d5dd763a90ec8459a) as discussed offline with @RemiLehe 
Locally, the CI test fails with the following error : 
```

amrex::Abort::0::ParticleContainer::locateParticle(): invalid particle. !!!
SIGABRT
--------------------------------------------------------------------------
MPI_ABORT was invoked on rank 0 in communicator MPI COMMUNICATOR 3 DUP FROM 0
with errorcode 6.

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
You may or may not see output from other processes, depending on
exactly when Open MPI kills them.
--------------------------------------------------------------------------
```
A solution I found was to move the MoveWindow function call to the beginning of the PIC loop. 
This ensures that all CI tests pass without and with the assert I commented out.

We are exploring how this would break the current benchmarks in this PR 
https://github.com/ECP-WarpX/WarpX/pull/3202

atmyers:
Looks like it is complaining about unused ParmParse variables in the RigidInjection_BTD 2D test now.

ax3l:
Current crash in `cartesian 2d`:
```
RigidInjection_BTD CRASHED
```

RevathiJambunathan:
> Current crash in `cartesian 2d`:
> 
> ```
> RigidInjection_BTD CRASHED
> ```

Unused ParmParse Variables:
  [TOP]::btd_openpmd.openpmd_backend(nvals = 1)  :: [h5]
It crashed because there was a mistake in the format! Fixed it.

Examples/Modules/RigidInjection/inputs_2d_BoostedFrame
```diff
@@ -69,6 +69,7 @@ btd_pltfile.dt_snapshots_lab = 1.8679589331096515e-13
 btd_pltfile.fields_to_plot = Ex Ey Ez Bx By Bz jx jy jz rho
 btd_pltfile.format = plotfile
 btd_pltfile.buffer_size = 32
+btd_pltfile.write_species = 1
```
ax3l, line 72: Can be a follow-up PR:
Should this file also add `btd_openpmd.write_species = 1`?

Examples/Modules/boosted_diags/inputs_3d_slice
```diff
@@ -133,6 +133,7 @@ btd_pltfile.dz_snapshots_lab = 0.001
 btd_pltfile.fields_to_plot = Ex Ey Ez Bx By Bz jx jy jz rho
 btd_pltfile.format = plotfile
 btd_pltfile.buffer_size = 32
+btd_pltfile.write_species = 1
```
ax3l, line 136: Can be a follow-up PR:
Should this file also add `btd_openpmd.write_species = 1`?

Source/Diagnostics/BTDiagnostics.cpp
ax3l: That looks good. All the offset logic here might warrant adding an inline comment before the `buffer_lo/li` assignments?

ax3l (approved):
Great work, thank you!

I added a few inline comments - ok for me to address them in a follow-up PR :)

Examples/Modules/RigidInjection/inputs_2d_BoostedFrame
RevathiJambunathan: Thank you @ax3l 
`write_species =1 ` is on by default. 
Actually it's not really needed here, since 1 the default.
I was turning it off (0) and on (1) when testing locally and when plotfile passed for particles. 
When plotfile passed with particles, I left it in there.

Examples/Modules/boosted_diags/inputs_3d_slice
RevathiJambunathan: openpmd is already writing particles.

Issue #3041, opened 2022-04-14
SP Nuclear Fusion Warning: floating-point operation result is out of range


ax3l:
I just saw this on perlmutter with SP builds:
```
/global/homes/a/ahuebl/src/warpx/Source/Particles/Collision/BinaryCollision/NuclearFusion/ProtonBoronFusionInitializeMomentum.H(94): warning #222-D: floating-point operation result is out of range

/global/homes/a/ahuebl/src/warpx/Source/Particles/Collision/BinaryCollision/NuclearFusion/ProtonBoronFusionInitializeMomentum.H(95): warning #222-D: floating-point operation result is out of range
```
Build:
```console
cmake -S . -B build -DWarpX_COMPUTE=CUDA -DWarpX_QED=OFF -DWarpX_GPUCLOCK=OFF -DWarpX_PRECISION=SINGLE
cmake --build build -j 16
```
Version: WarpX as of 22.03
Dependencies: https://warpx.readthedocs.io/en/22.03/install/hpc/perlmutter.html

NeilZaim:
Closing the issue following #3058.

PR #3058, opened 2022-04-21
Nuclear fusion: remove underflow warning in single precision

### LINKED ISSUE #2616: Possible issue concerning fusion module in single precision
While compiling WarpX in single precision and with GPU support, I got the following warning messages:
```
~/WarpX/Source/Particles/Collision/BinaryCollision/ProtonBoronFusionInitializeMomentum.H(94): warning #222-D: floating-point operation result is out of range

~/WarpX/Source/Particles/Collision/BinaryCollision/ProtonBoronFusionInitializeMomentum.H(95): warning #222-D: floating-point operation result is out of range
```
These lines are:

```
#ifdef AMREX_SINGLE_PRECISION_PARTICLES
        const amrex::ParticleReal ma_sq = m_alpha*m_alpha;
        const amrex::ParticleReal mBe_sq = m_beryllium*m_beryllium;
#else
//....
```
Can this be an issue? The square of these masses in SI units cannot be represented with a `float` ...

NeilZaim:
Close #2616, #3041 

These values underflow in single precision which can cause a warning during compilation, so I just set them to `std::numeric_limits<amrex::ParticleReal>::min()` (nuclear fusion currently does not support single precision anyways).

### DIFF
--- a/Source/Particles/Collision/BinaryCollision/NuclearFusion/ProtonBoronFusionInitializeMomentum.H
+++ b/Source/Particles/Collision/BinaryCollision/NuclearFusion/ProtonBoronFusionInitializeMomentum.H
@@ -86,13 +86,13 @@ namespace {
         // Rest energy of alpha+beryllium
         constexpr amrex::ParticleReal E_rest_abe = (m_alpha + m_beryllium)*c_sq;
 
-        // These constexprs underflow in single precision because we use SI units and sometimes the
-        // code won't compile. Nuclear fusion module does not currently work with single precision
-        // anyways so these #ifdef are just here to make sure that the code always compiles with
-        // single precision.
+        // These constexprs underflow in single precision because we use SI units, which can cause
+        // compilation to fail or generate a warning. Nuclear fusion module does not currently work
+        // with single precision anyways so this #ifdef is here to make sure that compilation
+        // always succeeds with single precision.
 #ifdef AMREX_SINGLE_PRECISION_PARTICLES
-        const amrex::ParticleReal ma_sq = m_alpha*m_alpha;
-        const amrex::ParticleReal mBe_sq = m_beryllium*m_beryllium;
+        constexpr amrex::ParticleReal ma_sq = std::numeric_limits<amrex::ParticleReal>::min();
+        constexpr amrex::ParticleReal mBe_sq = std::numeric_limits<amrex::ParticleReal>::min();
 #else
         constexpr amrex::ParticleReal ma_sq = m_alpha*m_alpha;
         constexpr amrex::ParticleReal mBe_sq = m_beryllium*m_beryllium;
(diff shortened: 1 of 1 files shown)


lucafedeli88 (approved):
Thanks for this fix, @NeilZaim !

Source/Particles/Collision/BinaryCollision/NuclearFusion/ProtonBoronFusionInitializeMomentum.H
```diff
@@ -86,13 +86,13 @@ namespace {
         // Rest energy of alpha+beryllium
         constexpr amrex::ParticleReal E_rest_abe = (m_alpha + m_beryllium)*c_sq;
 
-        // These constexprs underflow in single precision because we use SI units and sometimes the
-        // code won't compile. Nuclear fusion module does not currently work with single precision
-        // anyways so these #ifdef are just here to make sure that the code always compiles with
-        // single precision.
+        // These constexprs underflow in single precision because we use SI units, which can cause
+        // compilation to fail or generate a warning. Nuclear fusion module does not currently work
+        // with single precision anyways so this #ifdef is here to make sure that compilation
```
ax3l: If this module does not work with SP, can you please add an always-assert with a message in runtime code in case a user calls it anyway? Or is there such an assert already?

Source/Particles/Collision/BinaryCollision/NuclearFusion/ProtonBoronFusionInitializeMomentum.H
```diff
@@ -86,13 +86,13 @@ namespace {
...
+        // always succeeds with single precision.
 #ifdef AMREX_SINGLE_PRECISION_PARTICLES
-        const amrex::ParticleReal ma_sq = m_alpha*m_alpha;
-        const amrex::ParticleReal mBe_sq = m_beryllium*m_beryllium;
+        constexpr amrex::ParticleReal ma_sq = std::numeric_limits<amrex::ParticleReal>::min();
```
ax3l: Note: I would probably just write `constexpr double` here if this is known to be a number that does not fit into float ranges.

We implement a mixed-precision approach for things we cannot (yet) normalize well, when compiling in "single precision"-builds.

Source/Particles/Collision/BinaryCollision/NuclearFusion/ProtonBoronFusionInitializeMomentum.H
NeilZaim: Yes, there is already such an assert.

https://github.com/ECP-WarpX/WarpX/blob/8eb3163d865fbabb22c227b8f1810db079cca8f3/Source/Particles/Collision/BinaryCollision/NuclearFusion/NuclearFusionFunc.H#L61-L63

Source/Particles/Collision/BinaryCollision/NuclearFusion/ProtonBoronFusionInitializeMomentum.H
NeilZaim: Yes, mixed precision makes sense, I've modified the PR as you suggested.

Although I think that if we want to have this module working with single precision in the future it would probably be easier to change its system of units. (I'm expecting these issues to be all over the place since we're using nuclear masses in SI units)

ax3l (approved):
LGTM, thanks a lot

PR #3064, opened 2022-04-26
Boundary scraping diagnostic

### LINKED ISSUE #2855: Dump scraped particles to an openPMD file
Some notes from a discussion on this topic with @RevathiJambunathan @ax3l @atmyers :

- The buffer that contain the particles that touched the boundaries (either domain boundary or EB boundary) are here:
https://github.com/ECP-WarpX/WarpX/blob/development/Source/WarpX.H#L1244
https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/ParticleBoundaries.H

- We would create a new type of diagnostics (in addition to the existing `Full` and `BackTransformed` diagnostics). The new diagnostic type would store `ParticleDiagnostics` object that have pointers to the scraped particle buffer. The new diagnostics would have a `Flush` function that looks very similar to the one for the backtransformed diagnostics: https://github.com/ECP-WarpX/WarpX/blob/development/Source/Diagnostics/BTDiagnostics.cpp#L699

- The particle time should be stored as a runtime variable in the buffer, in order to keep track of the time at which the particles were scraped.

RemiLehe:
This PR adds a new type of diagnostic (`BoundaryScrapingDiagnostic`), in addition to `FullDiagnostic` and `BackTransformDiagnostic` (closes #2855). 

This new type of diagnostic dumps the particles that were scraped at the EB boundary during the simulation. Currently, it dumps all these particles to file only at the very end of the simulation (along with the time at which each of them reached the boundary). A follow-up PR could modify this, so that the code periodically dumps the particles to file, in a way that is similar to the back-transformed diagnostics (#3131). No fields are written to disk, in this diagnostic. Currently, this diagnostic only works with the openPMD output format.

This PR implements this diagnostics by creating a new diagnostic class `BoundaryScrapingDiagnostics` that derives from `Diagnostics`. (The classes `FullDiagnostics` and `BackTransformedDiagnostics` also derive from `Diagnostics`.) As such, there are a number of mandatory member functions that need to be defined (see `BoundaryScrapingDiagnostics.H` for the list of these member functions and there docstring.)

Note that the code that collects and stores particles at the EB boundary throughout the simulation, was already implemented. In particular, these particles are stored in the [`ParticleBoundaryBuffer` class](https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/ParticleBoundaryBuffer.H). Thus, this PR simply connects the `ParticleBoundaryBuffer` object to a `Diagnostic` object.


Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -1321,6 +1320,9 @@ WarpXOpenPMDPlot::WriteOpenPMDFieldsAll ( //const std::string& filename,
         series_iteration.setTime( time );
     }
 
+    // If there are no fields to be written, interrupt the function here
+    if ( varnames.size()==0 ) return;
```
RemiLehe, line 1325: This is needed in order to be able to gracefully handle the case where there is no field.

Docs/source/usage/parameters.rst
```diff
@@ -2231,6 +2232,17 @@ Back-Transformed Diagnostics (legacy output)
     slice diagnostic if there are within the user-defined width from
     the slice region defined by ``slice.dom_lo`` and ``slice.dom_hi``.
 
+Boundary scraping diagnostics
```
ax3l: Spelling of titles:
```suggestion
Boundary Scraping Diagnostics
```

Docs/source/usage/parameters.rst
```diff
@@ -2231,6 +2232,17 @@ Back-Transformed Diagnostics (legacy output)
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``BoundaryScrapingDiagnostics`` are used to collect the particles that are absorbed at the embedded boundary, throughout the simulation.
+(Note that this diagnostics does not save any field ; it only saves particles.)
+Currently, the only supported output format is openPMD, so the user needs to set ``format=openpmd``. In addition, the user needs
+to set ``save_particles_at_eb=1`` for each of the species that are to be saved in this diagnostic.
```
```suggestion
to set ``<species>.save_particles_at_eb=1`` for each of the species that are to be saved in this diagnostic.
```

Docs/source/usage/parameters.rst
```diff
@@ -2231,6 +2232,17 @@ Back-Transformed Diagnostics (legacy output)
     slice diagnostic if there are within the user-defined width from
     the slice region defined by ``slice.dom_lo`` and ``slice.dom_hi``.
 
+Boundary scraping diagnostics
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``BoundaryScrapingDiagnostics`` are used to collect the particles that are absorbed at the embedded boundary, throughout the simulation.
+(Note that this diagnostics does not save any field ; it only saves particles.)
+Currently, the only supported output format is openPMD, so the user needs to set ``format=openpmd``. In addition, the user needs
```
```suggestion
Currently, the only supported output format is openPMD, so the user needs to set ``<diag>.format=openpmd``. In addition, the user needs
```

Source/Diagnostics/BoundaryScrapingDiagnostics.H
```diff
@@ -0,0 +1,56 @@
...
+ * License: BSD-3-Clause-LBNL
+ */
+#ifndef WARPX_BOUNDARYSCRAPINGDIAGNOSTICS_H_
+#define WARPX_BOUNDARYSCRAPINGDIAGNOSTICS_H_
+
+#include "Diagnostics.H"
+
+#include <string>
+
+class
```
ax3l, line 16:
```suggestion
/** collect the particles that are absorbed at the embedded boundary, throughout the simulation
 */
class
```

Source/Diagnostics/BoundaryScrapingDiagnostics.H
```diff
@@ -0,0 +1,56 @@
...
+BoundaryScrapingDiagnostics final : public Diagnostics
+{
+public:
+
+    BoundaryScrapingDiagnostics (int i, std::string name);
```
ax3l, line 26: TODO: the `...` parts
```suggestion
    /** Constructor
     *
     * @param i ...
     * @param name ...
     */
    BoundaryScrapingDiagnostics (int i, std::string name);
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -556,8 +556,15 @@ WarpXOpenPMDPlot::WriteOpenPMDParticles (const amrex::Vector<ParticleDiag>& part
   WARPX_PROFILE("WarpXOpenPMDPlot::WriteOpenPMDParticles()");
 
   for (unsigned i = 0, n = particle_diags.size(); i < n; ++i) {
+
     WarpXParticleContainer* pc = particle_diags[i].getParticleContainer();
-    auto tmp = pc->make_alike<amrex::PinnedArenaAllocator>();
+    PinnedMemoryParticleContainer* pinned_pc = particle_diags[i].getPinnedParticleContainer();
```
ax3l, line 561: Why is this extra `if` needed now again (can you potentially also add an inline comment)?
Did you move this up from below for clarity?

Examples/Tests/scraping/inputs_rz
```diff
@@ -51,3 +52,6 @@ diag2.intervals = 1
 diag2.diag_type = Full
 diag2.fields_to_plot = Er
 diag2.format = openpmd
+
+diag3.diag_type = BoundaryScraping
```
ax3l, line 56: Please add in `Regression/WarpX-tests.ini`
```
addToCompileString = ... USE_OPENPMD=TRUE
cmakeSetupOpts = ... -DWarpX_OPENPMD=ON
```
so we can filter the test out if we build w/o openPMD.

Source/Diagnostics/BoundaryScrapingDiagnostics.cpp
```diff
@@ -0,0 +1,149 @@
...
+        bool save_particles_at_eb;
+        pp.query("save_particles_at_eb", save_particles_at_eb);
+        if (save_particles_at_eb == false) particle_saving_activated = false;
+    }
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE( particle_saving_activated,
+    "You need to set `save_particles_at_eb=1` for each species, in order to use BoundaryScrapingDiagnostic.");
+
+    // Check that the output format is openPMD
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE( m_format == "openpmd",
+    "You need to set `format=openpmd` for the BoundaryScrapingDiagnostic.");
```
ax3l: Let's add the value of `m_diag_name` as prefix before `format` so we can suggest the exact line that is needed:
```suggestion
    std:string("You need to set `") + m_diag_name +
    std::string(".format=openpmd` for the BoundaryScrapingDiagnostic."));
```

Source/Diagnostics/BoundaryScrapingDiagnostics.cpp
```diff
@@ -0,0 +1,149 @@
...
+    // Check that saving at EB has been activated for each requested species
+    bool particle_saving_activated = true;
+    for (auto const& species_name : m_output_species_names){
+        amrex::ParmParse pp(species_name);
+        bool save_particles_at_eb;
+        pp.query("save_particles_at_eb", save_particles_at_eb);
+        if (save_particles_at_eb == false) particle_saving_activated = false;
+    }
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE( particle_saving_activated,
+    "You need to set `save_particles_at_eb=1` for each species, in order to use BoundaryScrapingDiagnostic.");
```
ax3l: Let's make this error message more helpful:
- create little list (or [set](https://en.cppreference.com/w/cpp/container/set)) of species names that do not have `save_particles_at_eb` set
- output them of the list is not empty in the user-facing error message.

Source/Diagnostics/BoundaryScrapingDiagnostics.cpp
```diff
@@ -0,0 +1,149 @@
...
+    m_varnames = {}; // No fields in boundary scraping diagnostics
+
+    // Number of buffers = 1 for BoundaryScrapingDiagnostics.
+    // (buffers are used in BTDiagnostics, and correspond to different snapshots)
+    m_num_buffers = 1;
+
+    // Do a few checks
+#ifndef AMREX_USE_EB
+    amrex::Abort("You need to compile WarpX with EB, in order to use BoundaryScrapingDiagnostic.");
+#endif
```
ax3l, line 48: You could also do an openPMD-was-compiled check here.

Source/Diagnostics/BoundaryScrapingDiagnostics.cpp
```diff
@@ -0,0 +1,150 @@
...
+        bool save_particles_at_eb;
+        pp.query("save_particles_at_eb", save_particles_at_eb);
+        if (save_particles_at_eb == false) particle_saving_activated = false;
+    }
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE( particle_saving_activated,
+    "You need to set `save_particles_at_eb=1` for each species, in order to use BoundaryScrapingDiagnostic.");
+
+    // Check that the output format is openPMD
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE( m_format == "openpmd",
+    std:string("You need to set `") + m_diag_name +
```
ax3l:
```suggestion
    std::string("You need to set `") + m_diag_name +
```

Source/Diagnostics/BoundaryScrapingDiagnostics.H
```diff
@@ -0,0 +1,63 @@
...
+ */
+class
+BoundaryScrapingDiagnostics final : public Diagnostics
+{
+public:
+
+    /** Constructor
+     *
+     * @param i ...
+     * @param name ...
```
ax3l: What's `i` and `name` again, I forgot...
```suggestion
     * @param i ... ?
     * @param name diagnostics name in the inputs file
```

Source/Diagnostics/WarpXOpenPMD.cpp
ax3l: ~~Can you try to move this further up in the function? Or does the particle write not work if we don't do the open ops in the lines above?~~ update: ah I see, yes that is the function that opens the iteration.

ax3l:
PR updated. Only a question left here that might be good to be documented inline:
https://github.com/ECP-WarpX/WarpX/pull/3064#discussion_r885118109

Source/Diagnostics/WarpXOpenPMD.cpp
RemiLehe: In the previous code, we extracted the name of the particle components to be written from `pc`, e.g. by calling `getParticleRuntimeComps` on `pc` (see a few lines below).

However, there are cases where `pc` (which is a `WarpXParticleContainer`) and `pinned_pc` (which, here, is the buffer that stores the particles that were collected at the boundary) don't have the exact same components. For instance, `pinned_pc` stores `timestamp` (the time at which particles reached the boundary), whereas the corresponding `pc` does not.

For this reason, it is better to call `getParticleRuntimeComps` (and similar functions) on `pinned_pc`, in cases where this buffer is actually used. This is why the `if` condition was added.

RemiLehe:
Thanks for fixing the PR!! 
I just answered the question. If that works for you, could you approve the PR so that it can get merged?

ax3l (approved):
Thanks a lot, this is great! :)

PR #2017, opened 2021-06-15
[WIP] CMake: Particle DP with Field SP

ax3l:
Try a mixed precision build and see where we need to fix it :)

Compilation for mixed precision:
```bash
cmake -S . -B build -DWarpX_PRECISION=SINGLE -DWarpX_QED=OFF
cmake --build build -j 4
```

Equivalent to `-DWarpX_PRECISION=SINGLE -DAMReX_PARTICLES_PRECISION=DOUBLE`

cc @atmyers

PR #3065, opened 2022-04-26
Specify particle precision

KZhu-ME:
This PR adds the `WarpX_PARTICLE_PRECISION` build option that allows specifying `ParticleReal` to be different from `Real`. It helps save memory in simulations with many particles but a small domain size. 

If `WarpX_PARTICLE_PRECISION` is not set, it defaults to the value of `WarpX_PRECISION`.

Setting particle precision to single while keeping precision as double will cause build errors if `WarpX_QED` is on due to mismatched types in PICSAR.


Python/pywarpx/_libwarpx.py
```diff
@@ -620,6 +620,17 @@ def add_particles(self, species_name, x=None, y=None, z=None, ux=None, uy=None,
...
+        numpy_ParticleReal = f"f{self.libwarpx_so.warpx_ParticleReal_size()}"
+
+        x = x.astype(numpy_ParticleReal)
+        y = y.astype(numpy_ParticleReal)
+        z = z.astype(numpy_ParticleReal)
+        ux = ux.astype(numpy_ParticleReal)
+        uy = uy.astype(numpy_ParticleReal)
+        uz = uz.astype(numpy_ParticleReal)
+        attr = attr.astype(numpy_ParticleReal)
+
```
peterscherpelz: @KZhu-ME For this file I'd instead take all the changes in https://github.com/ModernElectron/WarpX/pull/165/files  -- except I do like using `astype` rather than `array`, BUT please add `copy=False` to the `.astype` functions - see https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html , without it there will be unnecessary array copies made every time this is called.

* Storing the particle real size is a little nicer than calling it again here.
* Using particlereal size rather than `float` when creating all the x/y/z/ux/uy/uz/w/attr arrays, if they do need to be created, also saves you an extra copy/cast.

Source/Particles/Pusher/UpdatePosition.H
```diff
@@ -15,6 +15,34 @@
 #include <AMReX_REAL.H>
 
 
+/** \brief Push the particle's positions over one timestep,
+ *    given the value of its momenta `ux`, `uy`, `uz` */
+AMREX_GPU_HOST_DEVICE AMREX_INLINE
+void UpdatePosition(amrex::Real& x, amrex::Real& y, amrex::Real& z,
+                    const amrex::Real ux, const amrex::Real uy, const amrex::Real uz,
+                    const amrex::Real dt )
```
peterscherpelz: Is there a reason this isn't *always* `ParticleReal`, rather than needing both `Real` and `ParticleReal` versions?

Source/Utils/WarpXUtil.cpp
```diff
@@ -478,6 +478,23 @@ getArrWithParser (const amrex::ParmParse& a_pp, char const * const str, std::vec
     }
 }
 
+#if defined(BL_USE_FLOAT) != defined(AMREX_SINGLE_PRECISION_PARTICLES)
+void
+getArrWithParser (const amrex::ParmParse& a_pp, char const * const str, std::vector<amrex::ParticleReal>& val,
+                    const int start_ix, const int num_val)
```
peterscherpelz: Granted I'm really bad at modern C++, but is there a way to use a template function instead so that this automatically works with both `Real` and `ParticleReal` versions rather than having to repeat the full code?
Also in this case `val` is `static_cast` to `Real` which seems fishy...

peterscherpelz (commented):
Thanks for tracking these things down! Please see questions/suggestions below.

Source/Utils/WarpXUtil.cpp
KZhu-ME: I redefined the function as it is called multiple times with val being a vector of `Real` or `ParticleReal` values, so having only one definition errors out if `Real` and `ParticleReal` are different. As for why it is `static_cast`to `Real`, I'm guessing it's because the next operations done on it expect it to be `Real`.

Python/pywarpx/_libwarpx.py
```diff
@@ -594,42 +597,43 @@ def add_particles(self, species_name, x=None, y=None, z=None, ux=None, uy=None,
...
-        attr = attr.astype(numpy_ParticleReal)
+        # Iff x/y/z/ux/uy/uz are not numpy arrays of the correct dtype, new
+        # array copies are made with the correct dtype
+        x = x.astype(self._numpy_particlereal_dtype, copy=False)
+        y = y.astype(self._numpy_particlereal_dtype, copy=False)
+        z = z.astype(self._numpy_particlereal_dtype, copy=False)
+        ux = ux.astype(self._numpy_particlereal_dtype, copy=False)
+        uy = uy.astype(self._numpy_particlereal_dtype, copy=False)
+        uz = uz.astype(self._numpy_particlereal_dtype, copy=False)
+        attr = attr.astype(self._numpy_particlereal_dtype, copy=False)
```
peterscherpelz:
```suggestion
```

This line should never trigger.

peterscherpelz (approved):
No further comments from me except the one suggestion, so now the WarpX folks who know the C++ code should look :-)

ax3l:
Oh this might be similar to #2017. Let's discuss tomorrow?

Source/Particles/Pusher/UpdatePosition.H
```diff
@@ -18,7 +18,7 @@
 /** \brief Push the particle's positions over one timestep,
  *    given the value of its momenta `ux`, `uy`, `uz` */
 AMREX_GPU_HOST_DEVICE AMREX_INLINE
-void UpdatePosition(amrex::ParticleReal& x, amrex::ParticleReal& y, amrex::ParticleReal& z,
```
KZhu-ME: This line doesn't compile when setting `WarpX_PRECISION` to DOUBLE and `WarpX_PARTICLE_PRECISION` to SINGLE. Changing the declaration to be `amrex::ParticleReal` instead of `amrex::ParticleReal&` fixes the compile error, but also breaks tests. The compilation works when both precisions are set to double. Below is the error message
```
repos/warpx/Source/Particles/PhysicalParticleContainer.cpp:1509:36: error: cannot bind non-const lvalue reference of type ‘amrex::ParticleReal&’ {aka ‘float&’} to an rvalue of type ‘amrex::ParticleReal’ {aka ‘float’}
 1509 |                 UpdatePosition(pos.x, pos.y, pos.z, u.x, u.y, u.z, t_fract);
      |                                ~~~~^
```

Source/Diagnostics/ReducedDiags/ParticleHistogram.cpp
```diff
@@ -211,18 +211,19 @@ void ParticleHistogram::ComputeDiags (int step)
                 {
                     amrex::ParticleReal x, y, z;
                     GetPosition(i, x, y, z);
-                    auto const w  = d_w[i];
+                    auto const w  = (amrex::Real)d_w[i];
```
ax3l, line 214: The casts here are costly. For all the things below, I would rather stay/cast to `ParticleReal`. But is it possible that all the functor stuff accepts only `Real`?

Source/Initialization/WarpXInitData.cpp
```diff
@@ -119,6 +119,20 @@ WarpX::PrintMainPICparameters ()
     amrex::Print() << "--------------------------- MAIN EM PIC PARAMETERS ----------------------------\n";
     amrex::Print() << "-------------------------------------------------------------------------------\n";
 
+    // print warpx build information
+    if (std::is_same<Real, float>::value) {
```
```suggestion
    if constexpr (std::is_same<Real, float>::value) {
```

Source/Initialization/WarpXInitData.cpp
```diff
@@ -119,6 +119,20 @@ WarpX::PrintMainPICparameters ()
...
+      amrex::Print() << "Precision:            | SINGLE" << "\n";
+    }
+    else {
+      amrex::Print() << "Precision:            | DOUBLE" << "\n";
+    }
+    if (std::is_same<ParticleReal, float>::value) {
```
```suggestion
    if constexpr (std::is_same<ParticleReal, float>::value) {
```

Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
```diff
@@ -106,7 +106,8 @@ void UpdateMomentumPerezElastic (
                ( m1*g1s*m2*g2s/(p1sm*p1sm*inv_c2) + T_R(1.0) );
 
         // Compute the minimal impact parameter
-        T_R bmin = amrex::max(PhysConst::hbar*MathConst::pi/p1sm,b0);
+        T_R hbar_pi = static_cast<T_R>(PhysConst::hbar*MathConst::pi);
+        T_R bmin = amrex::max(hbar_pi/p1sm,b0);
```
ax3l:
```suggestion
        constexpr T_R hbar_pi = static_cast<T_R>(PhysConst::hbar*MathConst::pi);
        const T_R bmin = amrex::max(hbar_pi/p1sm, b0);
```

Source/Diagnostics/ReducedDiags/ParticleHistogram.cpp
PhilMiller: Are you concerned about execution cost, or how the code is written? This could be
```suggestion
                    amrex::Real const w  = d_w[i];
```
If it's about the expression

IIRC, the parsers are generic and could be passed potentially single-precision `ParticleReal`, but there's a comment somewhere saying we should stick with double-precision for calls to parsed functions, maybe by Remi.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1496,17 +1510,19 @@ PhysicalParticleContainer::AddPlasmaFlux (amrex::Real dt)
...
                 u = inj_mom->getMomentum(pos.x, pos.y, pos.z, engine);
+                auto pu = PDim3(u);
 
                 u.x *= PhysConst::c;
                 u.y *= PhysConst::c;
                 u.z *= PhysConst::c;
 
                 const amrex::Real t_fract = amrex::Random(engine)*dt;
-                UpdatePosition(pos.x, pos.y, pos.z, u.x, u.y, u.z, t_fract);
+                UpdatePosition(ppos.x, ppos.y, ppos.z, pu.x, pu.y, pu.z, t_fract);
```
PhilMiller, line 1525: The change to add and use `PDim3` is incomplete. The values in `ppos` and `pu` are going to be modified by this function, but calls below still use `pos`

PhilMiller (changes requested):

Source/Utils/WarpXUtil.H
```diff
@@ -265,13 +339,6 @@ int queryWithParser (const amrex::ParmParse& a_pp, char const * const str, doubl
  * \param[in] num_val number of input values to use (optional with arrays, default is
  * amrex::ParmParse::LAST for reading until the last input value)
  */
```
PhilMiller: This comment block refers to the code that's deleted just below, it seems.

Source/Utils/WarpXUtil.cpp
```diff
@@ -388,135 +388,6 @@ parseStringtoInt(std::string str, std::string name)
     return ival;
 }
 
-// Overloads for float/double instead of amrex::Real to allow makeParser() to query for
-// my_constants as double even in single precision mode
-// Always parsing in double precision avoids potential overflows that may occur when parsing user's
-// expressions because of the limited range of exponentials in single precision
```
PhilMiller, line 394: @ax3l Here's the comment I was thinking of

Source/Diagnostics/ReducedDiags/ParticleHistogram.cpp
KZhu-ME: The `fun_filterparser` and `fun_partparser` stuff below can stay as `ParticleReal`, I'll change it back that way.

Source/Particles/PhysicalParticleContainer.cpp
KZhu-ME: Good catch, I'll update the code below.

Source/Utils/WarpXUtil.H
KZhu-ME: I believe the comment refers to the function above it, so I'll move it up.

Source/Diagnostics/ReducedDiags/ParticleHistogram.cpp
ax3l: @PhilMiller I am concerned about the cost of a cast on GPU, it's a special instruction and "surprisingly" costly.

Source/Particles/Filter/FilterFunctors.H
```diff
@@ -125,7 +125,7 @@ struct ParserFilter
             uz /= m_mass;
         }
         // ux, uy, uz are now in beta*gamma
-        if ( m_function_partparser(m_t,x,y,z,ux,uy,uz) ) return 1;
+        if ( m_function_partparser(m_t,(amrex::Real) x,(amrex::Real) y,(amrex::Real) z,ux,uy,uz)) return 1;
```
ax3l: can we make `m_t` a `ParticleReal` instead of casting the others?

I think `m_mass` should also be of `ParticleReal` type.

Source/Diagnostics/ReducedDiags/ParticleHistogram.cpp
```diff
@@ -211,18 +211,17 @@ void ParticleHistogram::ComputeDiags (int step)
...
                     auto const ux = d_ux[i] / PhysConst::c;
                     auto const uy = d_uy[i] / PhysConst::c;
                     auto const uz = d_uz[i] / PhysConst::c;
 
                     // don't count a particle if it is filtered out
                     if (do_parser_filter)
-                        if (!fun_filterparser(t, x, y, z, ux, uy, uz))
+                         if (!fun_filterparser(t, x, y, z, ux, uy, uz))
```
ax3l:
```suggestion
                        if (!fun_filterparser(t, x, y, z, ux, uy, uz))
```

ax3l (commented):
Very solid PR, nice build & docs.

Comments inline :)

Source/Diagnostics/ReducedDiags/ParticleHistogram.cpp
PhilMiller: Today I learned. Thanks

Regression/WarpX-tests.ini
```diff
@@ -3030,6 +3030,24 @@ compareParticles = 1
 particleTypes = electrons he_ions
 analysisRoutine = Examples/analysis_default_regression.py
 
+[background_mcc_dp_psp]
+buildDir = .
+inputFile = Examples/Physics_applications/capacitive_discharge/inputs_2d
+runtime_params = warpx.abort_on_warning_threshold = high
+dim = 2
+addToCompileString =
```
PhilMiller, line 3038: @ax3l Right now, this will build in the 2d pipeline. Do we want to adjust the test selection in `prepare_file_ci.py` to encompass this mixed-precision test?

More broadly, looking at this, following my recent CI adjustments in #3107 (still to be merged), I think it may be worth amending that script to strictly partition the tests, to ensure tests are neither doubled nor dropped on the floor.

ax3l: That sounds reasonable to me.

Generally speaking, we try to have the number of unique binary builds per Azure entry <=3, IIRC. That way, a cache re-silvering (usually on Mondays after we merge the AMReX weekly update) can pass within an hour.

Source/Particles/Filter/FilterFunctors.H
```diff
@@ -125,7 +125,7 @@ struct ParserFilter
             uz /= m_mass;
         }
         // ux, uy, uz are now in beta*gamma
-        if ( m_function_partparser(m_t,x,y,z,ux,uy,uz) ) return 1;
+        if ( m_function_partparser((amrex::ParticleReal)m_t,x,y,z,ux,uy,uz)) return 1;
```
ax3l: I actually thought about this...
```suggestion
        if ( m_function_partparser(m_t,x,y,z,ux,uy,uz)) return 1;
```

Source/Particles/Filter/FilterFunctors.H
```diff
@@ -137,7 +137,7 @@ public:
     /** Store physical time. */
     amrex::Real m_t;
```
ax3l: ...with this:
```suggestion
    amrex::ParticleReal m_t;
```

Is that possible?

Source/Diagnostics/ReducedDiags/ParticleHistogram.cpp
KZhu-ME: Should I leave this line as is or remove the explicit cast and change the line to be what Phil suggested `amrex::Real const w = d_w[i];`?

PhilMiller: You should probably leave it as is, to emphasize that a cast operation is occurring, which may be a noticeable expense.

Regression/WarpX-tests.ini
ax3l: @KZhu-ME @PhilMiller as a follow-up PR, can we add a test that does the exact opposite: compiling in `SP` for `amrex::Real` but using `DP` for `amrex::ParticleReal`? :)

ax3l (approved):
Great work, thank you so much! :tada:

Regression/WarpX-tests.ini
KZhu-ME: Sounds good, I'm trying to run the `background_mcc` test in that configuration with `SP` for `Real` and `DP` for `ParticleReal`, but I'm running into particle out of bounds errors on `RedistributeCPU` calls.

ax3l: Thanks, ok let's debug in the follow-up PR :+1:

PR #3067, opened 2022-04-27
Add do_average option to the particle fields diagnostic

dpgrote:
This option allows the particle field diagnostics to be sums over the specified function (rather than only averages). The option defaults to take averages.


Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -634,7 +634,7 @@ FullDiagnostics::InitializeFieldFunctors (int lev)
         std::string varname = m_pfield_varnames[pcomp];
         for (int ispec=0; ispec<int(m_pfield_species.size()); ispec++) {
             m_all_field_functors[lev][nvar + pcomp * nspec + ispec] = std::make_unique<ParticleReductionFunctor>(nullptr,
-                    lev, m_crse_ratio, m_pfield_strings[varname], m_pfield_species_index[ispec],
+                    lev, m_crse_ratio, m_pfield_strings[varname], m_pfield_species_index[ispec], m_pfield_do_average[pcomp],
```
hklion: Since `m_pfield_dofilter` and `m_pfield_filter_strings` are both `std::map`, would it make sense for `m_pfield_do_average` to also be, so the access pattern is consistent here?

Source/Diagnostics/ComputeDiagFunctors/ParticleReductionFunctor.H
```diff
@@ -26,12 +26,14 @@ public:
...
      * \param[in] ispec index of the species over which to calculate the average
+     * \param[in] do_average Whether to do an average or a sum of the function
      * \param[in] do_filter Whether to apply a filter function to particles before averaging
      * \param[in] filter_str Parser string for filter function to apply before averaging
      * \param[in] ncomp Number of component of mf_src to cell-center in dst multifab.
      */
     ParticleReductionFunctor(const amrex::MultiFab * const mf_src, const int lev,
-                       const amrex::IntVect crse_ratio, const std::string fn_str, const int ispec,
+                       const amrex::IntVect crse_ratio, const std::string fn_str,
+                       const int ispec, const int do_average,
```
hklion: Is there a particular reason for `do_average` to be an int rather than a bool? If not, it might make sense for it to be a bool in order to parallel `do_filter` (or for `do_filter` to also be an int).

hklion (commented):
Thanks for this, @dpgrote! Looks good. :) I just have a few minor style comments to more closely mirror the surrounding code.

Source/Diagnostics/FullDiagnostics.cpp
dpgrote: Yes, I can make it a map. Is there a reason that maps are used here for the other quantities? It doesn't seem like that is needed since a vector would work just as well. Are there plans for further changes the need maps?

Source/Diagnostics/ComputeDiagFunctors/ParticleReductionFunctor.H
dpgrote: I tend to avoid `vector<bool>` since that type can be problematic. It would work ok here though since only the `[]` operator is used. I can make the change.

Source/Diagnostics/FullDiagnostics.cpp
hklion: Yeah, I agree that maps and vectors are essentially interchangeable here. I think I originally had it as a map to avoid bookkeeping errors, since I initially had a separate vector of functors for these diagnostics, instead of appending them to `m_all_field_functors`. I don't have any plans that would require keeping them as maps, though.

Source/Diagnostics/ComputeDiagFunctors/ParticleReductionFunctor.H
hklion: Gotcha. All things being equal, I think my personal (weak) inclination would be towards towards bool here, especially since the `std::vector<bool>` should be okay, but if there's a convention against it, I'm fine with keeping it as an int. I know there's been some discussion about int vs bool at recent dev meetings, but I forget whether it was just re: avoiding `amrex::Vector<bool>` in device code.

Source/Diagnostics/FullDiagnostics.cpp
dpgrote: Ok. For the sake of code simplification, would it be ok if I change the maps to be vectors?

hklion: Sure, go for it.

dpgrote: Changes done.

BTW, I think that this diagnostic will work with RZ, depositing the data on a r-z grid. Since the code is accessing the `SuperParticleType` directly, the `p.pos(0)` will be `r` and `p.pos(1)` will be `z`. The filtering parser however will work with `x`, `y`, and `z`.

hklion: Awesome. Thanks! You're probably right, and if it doesn't work it would likely be a quick fix. Though some additional code would need to be added in either `Diagnostics.cpp` or `FullDiagnostics.cpp` to create the functors when RZ is turned on. It may work with plotfiles but it definitely doesn't work with RZ+OpenPMD at the moment. I'm also not sure if we'd also want to output modes for these fields, or if that's just for the fields used in evolution.

hklion (approved):

Source/Diagnostics/FullDiagnostics.cpp
dpgrote: Yes, some extra work will be needed, and there are complications. I don't need this now in RZ. This can be done in a future PR.

dpgrote: Thanks for the comments!

Source/Diagnostics/ComputeDiagFunctors/ParticleReductionFunctor.H
EZoni: Sorry, I might have missed something here. It seems like now we do have `vector<bool>` in Source/Diagnostics/Diagnostics.H (the member variable `m_pfield_do_average`), so why would it be problematic to replace the argument `const int do_average` with `const bool do_average` here (and in the other instances below)?
```suggestion
                       const int ispec, const bool do_average,
```

Source/Diagnostics/ComputeDiagFunctors/ParticleReductionFunctor.H
```diff
@@ -45,6 +47,7 @@ public:
 private:
     int const m_lev; /**< level on which mf_src is defined */
     int const m_ispec; /**< index of species to average over */
+    int const m_do_average; /**< Whether to calculate the average of the data */
```
EZoni:
```suggestion
    bool const m_do_average; /**< Whether to calculate the average of the data */
```

Source/Diagnostics/ComputeDiagFunctors/ParticleReductionFunctor.cpp
```diff
@@ -16,9 +16,10 @@
 using namespace amrex::literals;
 
 ParticleReductionFunctor::ParticleReductionFunctor (const amrex::MultiFab* mf_src, const int lev,
-        const amrex::IntVect crse_ratio, const std::string fn_str, const int ispec,
+        const amrex::IntVect crse_ratio, const std::string fn_str,
+        const int ispec, const int do_average,
```
EZoni:
```suggestion
        const int ispec, const bool do_average,
```

EZoni (commented):
Thank you for the PR, @dpgrote! And thank you for the review, @hklion. I started looking into this and left a couple of additional minor comments.

Source/Diagnostics/ComputeDiagFunctors/ParticleReductionFunctor.H
dpgrote: @EZoni Your changes are correct - it was an oversight that the `do_average` scalar was not changed to a bool.

EZoni (approved):
Thank you for updating the PR, @dpgrote! Looks good to me.

PR #3068, opened 2022-04-28
Avoid resizing ParticleBoundaryBuffer allocation to total particle count

### LINKED ISSUE #164: Slow performance and suspicious memory behavior on large runs
When running a few runs with embedded boundaries and ~350 million particles on 1x A100 80GB GPU, I'm seeing very poor performance.

* With 16 grids in the simulation, A100 memory usage is ~62 GB per nvtop. On some steps it is quite fast (< 1 sec/step) whereas other steps take ~25 seconds. Looking at nvtop, it looks like during each 25sec step host memory increases from ~8 GB to ~80 GB; GPU memory is basically unchanged, and at the end of the step host memory drops back to 8 GB.
* With 1 grid in the simulation, A100 memory usage is even lower (30-40 GB per nvtop), but the ~25 sec/step is consistent on every single step, and the same host memory sawtoothing from 8GB to 80 GB on every step is seen.

The profiling is also somewhat weird, as it shows that the `WarpX::Evolve::step` function takes up most of the _exclusive_ time (86.7% in the case of 16 grids), but I was under the impression almost everything that could take some performance time was timed as a separate function under this.

PhilMiller:
Fixes ModernElectron/WarpX#164

Previously, having enough particles to fill ~50% of GPU memory would lead to some time steps spending inordinate amounts of time in the pre-copy `resize()` calls for the pinned-memory buffer, and increasing particle count just a bit further would lead the simulation to fail. The excess time was spent in allocation, not initialization or copying over existing data.

With this change, I was able to run a small test that creates enough particles that it filled essentially 100% of GPU memory, with no crashes or slowdowns.

### DIFF
--- a/Source/Particles/ParticleBoundaryBuffer.cpp
+++ b/Source/Particles/ParticleBoundaryBuffer.cpp
@@ -15,6 +15,7 @@
 
 #include <AMReX_Geometry.H>
 #include <AMReX_ParmParse.H>
+#include <AMReX.H>
 
 struct IsOutsideDomainBoundary {
     amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> m_plo;
@@ -23,7 +24,7 @@ struct IsOutsideDomainBoundary {
     int m_iside;
 
     template <typename SrcData>
-    AMREX_GPU_DEVICE AMREX_FORCE_INLINE
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
     int operator() (const SrcData& src,
                     int ip, const amrex::RandomEngine& /*engine*/) const noexcept
     {
@@ -159,6 +160,9 @@ void ParticleBoundaryBuffer::gatherParticles (MultiParticleContainer& mypc,
     const amrex::Geometry& geom = warpx_instance.Geom(0);
     auto plo = geom.ProbLoArray();
     auto phi = geom.ProbHiArray();
+
+    int *pre_count = static_cast<int*>(amrex::The_Managed_Arena()->alloc(sizeof(size_t)));
+
     for (int idim = 0; idim < AMREX_SPACEDIM; ++idim)
     {
         if (geom.isPeriodic(idim)) continue;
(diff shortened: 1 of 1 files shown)


PhilMiller:
With this change, I was able to run a small test that creates enough particles that it filled essentially 100% of GPU memory, with no crashes or slowdowns.

PhilMiller:
Wow, lots of totally unrelated CI failures, all at once. Not one of those even touches the code.

ax3l:
Ping @atmyers for review :)

KZhu-ME (approved):

KZhu-ME (approved):

PhilMiller:
@KZhu-ME We could maybe run a synthetic test that fills memory with just electrons, to maximally stress the cost of a single species, since the impact would be spread across species. I think in that case, the original code would slow and fail at even lower occupancy, while with this PR it should be completely solid.

Source/Particles/ParticleBoundaryBuffer.cpp
```diff
@@ -159,6 +160,9 @@ void ParticleBoundaryBuffer::gatherParticles (MultiParticleContainer& mypc,
     const amrex::Geometry& geom = warpx_instance.Geom(0);
     auto plo = geom.ProbLoArray();
     auto phi = geom.ProbHiArray();
+
+    int *pre_count = static_cast<int*>(amrex::The_Managed_Arena()->alloc(sizeof(size_t)));
```
atmyers: I suspect it will perform better to use `amrex::ReduceOps` for this, rather than atomically incrementing a managed memory counter, as here for example: 

https://github.com/ECP-WarpX/WarpX/blob/development/Source/Diagnostics/ReducedDiags/ParticleExtrema.cpp#L451

More worried about systems like Crusher where the managed memory performance is still pretty bad than NVIDIA...

PhilMiller: Yeah, this was an admittedly lazy, expedient approach. Managed memory is a problem. Revised to use the reduction approach, and testing now.

atmyers (approved):

PR #3070, opened 2022-04-28
Add velocity in pml as an input in function alpha

amadou38:
In this first tasks, I replaced the constant variable `c` by the velocity `v` (v < c) of the macroparticle in the PML_current.H file.
This is done by adding new variables (as a type of `amrex::Real`) as well as in the WarpX.H, WarpX.cpp, PML.H, PML.cpp and WarpXInitData.cpp in order to be accessible and modifiable.
This velocity can now be given by the user in the input file (as `warpx.v_particle_pml = ...`).

### DIFF
--- a/Source/BoundaryConditions/PML.H
+++ b/Source/BoundaryConditions/PML.H
@@ -208,8 +208,6 @@ private:
     bool m_dive_cleaning;
     bool m_divb_cleaning;
 
-    amrex::Real m_v_sigma_sb;
-    
     const amrex::Geometry* m_geom;
     const amrex::Geometry* m_cgeom;
 
--- a/Source/BoundaryConditions/PML.H
+++ b/Source/BoundaryConditions/PML.H
@@ -43,15 +43,15 @@ struct SigmaBox
 {
     SigmaBox (const amrex::Box& box, const amrex::BoxArray& grids,
               const amrex::Real* dx, const amrex::IntVect& ncell, const amrex::IntVect& delta,
-              const amrex::Box& regdomain, amrex::Real v_sigma);
+              const amrex::Box& regdomain, const amrex::Real v_sigma);
 
     void define_single (const amrex::Box& regdomain, const amrex::IntVect& ncell,
                         const amrex::Array<amrex::Real,AMREX_SPACEDIM>& fac,
-                        amrex::Real v_sigma);
+                        const amrex::Real v_sigma);
     void define_multiple (const amrex::Box& box, const amrex::BoxArray& grids,
                           const amrex::IntVect& ncell,
                           const amrex::Array<amrex::Real,AMREX_SPACEDIM>& fac,
-                          amrex::Real v_sigma);
+                          const amrex::Real v_sigma);
 
     void ComputePMLFactorsB (const amrex::Real* dx, amrex::Real dt);
     void ComputePMLFactorsE (const amrex::Real* dx, amrex::Real dt);
@@ -76,7 +76,7 @@ class SigmaBoxFactory
 public:
     SigmaBoxFactory (const amrex::BoxArray& grid_ba, const amrex::Real* dx,
                      const amrex::IntVect& ncell, const amrex::IntVect& delta,
-                     const amrex::Box& regular_domain, amrex::Real v_sigma_sb)
+                     const amrex::Box& regular_domain, const amrex::Real v_sigma_sb)
         : m_grids(grid_ba), m_dx(dx), m_ncell(ncell), m_delta(delta), m_regdomain(regular_domain), m_v_sigma_sb(v_sigma_sb) {}
     virtual ~SigmaBoxFactory () = default;
 
@@ -112,7 +112,7 @@ public:
     MultiSigmaBox(const amrex::BoxArray& ba, const amrex::DistributionMapping& dm,
                   const amrex::BoxArray& grid_ba, const amrex::Real* dx,
                   const amrex::IntVect& ncell, const amrex::IntVect& delta,
-                  const amrex::Box& regular_domain, amrex::Real v_sigma_sb);
+                  const amrex::Box& regular_domain, const amrex::Real v_sigma_sb);
     void ComputePMLFactorsB (const amrex::Real* dx, amrex::Real dt);
     void ComputePMLFactorsE (const amrex::Real* dx, amrex::Real dt);
 private:
@@ -208,6 +208,8 @@ private:
     bool m_dive_cleaning;
     bool m_divb_cleaning;
 
+    amrex::Real m_v_sigma_sb;
+    
     const amrex::Geometry* m_geom;
     const amrex::Geometry* m_cgeom;
 
--- a/Source/BoundaryConditions/PML.H
+++ b/Source/BoundaryConditions/PML.H
@@ -43,13 +43,15 @@ struct SigmaBox
 {
     SigmaBox (const amrex::Box& box, const amrex::BoxArray& grids,
               const amrex::Real* dx, const amrex::IntVect& ncell, const amrex::IntVect& delta,
-              const amrex::Box& regdomain);
+              const amrex::Box& regdomain, amrex::Real v_sigma_sb);
 
     void define_single (const amrex::Box& regdomain, const amrex::IntVect& ncell,
-                        const amrex::Array<amrex::Real,AMREX_SPACEDIM>& fac);
+                        const amrex::Array<amrex::Real,AMREX_SPACEDIM>& fac,
+                        amrex::Real v_sigma);
     void define_multiple (const amrex::Box& box, const amrex::BoxArray& grids,
                           const amrex::IntVect& ncell,
-                          const amrex::Array<amrex::Real,AMREX_SPACEDIM>& fac);
+                          const amrex::Array<amrex::Real,AMREX_SPACEDIM>& fac,
+                          amrex::Real v_sigma);
 
     void ComputePMLFactorsB (const amrex::Real* dx, amrex::Real dt);
     void ComputePMLFactorsE (const amrex::Real* dx, amrex::Real dt);
@@ -64,6 +66,7 @@ struct SigmaBox
     SigmaVect sigma_cumsum_fac;
     SigmaVect sigma_star_fac;
     SigmaVect sigma_star_cumsum_fac;
+    amrex::Real v_sigma;
 
 };
 
@@ -73,8 +76,8 @@ class SigmaBoxFactory
 public:
     SigmaBoxFactory (const amrex::BoxArray& grid_ba, const amrex::Real* dx,
                      const amrex::IntVect& ncell, const amrex::IntVect& delta,
-                     const amrex::Box& regular_domain)
-        : m_grids(grid_ba), m_dx(dx), m_ncell(ncell), m_delta(delta), m_regdomain(regular_domain) {}
+                     const amrex::Box& regular_domain, amrex::Real v_sigma_sb)
+        : m_grids(grid_ba), m_dx(dx), m_ncell(ncell), m_delta(delta), m_regdomain(regular_domain), m_v_sigma_sb(v_sigma_sb) {}
     virtual ~SigmaBoxFactory () = default;
 
     SigmaBoxFactory (const SigmaBoxFactory&) = default;
@@ -86,7 +89,7 @@ public:
 
     virtual SigmaBox* create (const amrex::Box& box, int /*ncomps*/,
                               const amrex::FabInfo& /*info*/, int /*box_index*/) const final
-        { return new SigmaBox(box, m_grids, m_dx, m_ncell, m_delta, m_regdomain); }
+        { return new SigmaBox(box, m_grids, m_dx, m_ncell, m_delta, m_regdomain, m_v_sigma_sb); }
     virtual void destroy (SigmaBox* fab) const final {
         delete fab;
     }
@@ -99,6 +102,7 @@ private:
     amrex::IntVect m_ncell;
     amrex::IntVect m_delta;
     amrex::Box m_regdomain;
+    amrex::Real m_v_sigma_sb;
 };
 
 class MultiSigmaBox
@@ -108,7 +112,7 @@ public:
     MultiSigmaBox(const amrex::BoxArray& ba, const amrex::DistributionMapping& dm,
                   const amrex::BoxArray& grid_ba, const amrex::Real* dx,
                   const amrex::IntVect& ncell, const amrex::IntVect& delta,
-                  const amrex::Box& regular_domain);
+                  const amrex::Box& regular_domain, amrex::Real v_sigma_sb);
     void ComputePMLFactorsB (const amrex::Real* dx, amrex::Real dt);
     void ComputePMLFactorsE (const amrex::Real* dx, amrex::Real dt);
 private:
@@ -129,7 +133,7 @@ public:
          int do_moving_window, int pml_has_particles, int do_pml_in_domain,
          const bool do_multi_J,
          const bool do_pml_dive_cleaning, const bool do_pml_divb_cleaning,
-         int max_guard_EB,
+         int max_guard_EB, amrex::Real v_sigma_sb,
          const amrex::IntVect do_pml_Lo = amrex::IntVect::TheUnitVector(),
          const amrex::IntVect do_pml_Hi = amrex::IntVect::TheUnitVector());
 
--- a/Source/BoundaryConditions/PML.H
+++ b/Source/BoundaryConditions/PML.H
@@ -43,7 +43,7 @@ struct SigmaBox
 {
     SigmaBox (const amrex::Box& box, const amrex::BoxArray& grids,
               const amrex::Real* dx, const amrex::IntVect& ncell, const amrex::IntVect& delta,
-              const amrex::Box& regdomain, amrex::Real v_sigma_sb);
+              const amrex::Box& regdomain, amrex::Real v_sigma);
 
     void define_single (const amrex::Box& regdomain, const amrex::IntVect& ncell,
                         const amrex::Array<amrex::Real,AMREX_SPACEDIM>& fac,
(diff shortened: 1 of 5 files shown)


Source/BoundaryConditions/PML.H
```diff
@@ -43,13 +43,15 @@ struct SigmaBox
 {
     SigmaBox (const amrex::Box& box, const amrex::BoxArray& grids,
               const amrex::Real* dx, const amrex::IntVect& ncell, const amrex::IntVect& delta,
-              const amrex::Box& regdomain);
+              const amrex::Box& regdomain, amrex::Real v_sigma);
```
EZoni: Since this new function argument is read-only, I think it would be appropriate to add the `const` qualifier (same comment holds for similar instances below):
```suggestion
              const amrex::Box& regdomain, const amrex::Real v_sigma);
```

Source/BoundaryConditions/PML.H
```diff
@@ -43,13 +43,15 @@ struct SigmaBox
...
 
     void define_single (const amrex::Box& regdomain, const amrex::IntVect& ncell,
-                        const amrex::Array<amrex::Real,AMREX_SPACEDIM>& fac);
+                        const amrex::Array<amrex::Real,AMREX_SPACEDIM>& fac,
+                        amrex::Real v_sigma);
```
EZoni:
```suggestion
                        const amrex::Real v_sigma);
```

Source/BoundaryConditions/PML.H
```diff
@@ -43,13 +43,15 @@ struct SigmaBox
...
     void define_multiple (const amrex::Box& box, const amrex::BoxArray& grids,
                           const amrex::IntVect& ncell,
-                          const amrex::Array<amrex::Real,AMREX_SPACEDIM>& fac);
+                          const amrex::Array<amrex::Real,AMREX_SPACEDIM>& fac,
+                          amrex::Real v_sigma);
```
EZoni:
```suggestion
                          const amrex::Real v_sigma);
```

Source/BoundaryConditions/PML.H
```diff
@@ -73,8 +76,8 @@ class SigmaBoxFactory
 public:
     SigmaBoxFactory (const amrex::BoxArray& grid_ba, const amrex::Real* dx,
                      const amrex::IntVect& ncell, const amrex::IntVect& delta,
-                     const amrex::Box& regular_domain)
-        : m_grids(grid_ba), m_dx(dx), m_ncell(ncell), m_delta(delta), m_regdomain(regular_domain) {}
+                     const amrex::Box& regular_domain, amrex::Real v_sigma_sb)
```
EZoni:
```suggestion
                     const amrex::Box& regular_domain, const amrex::Real v_sigma_sb)
```

Source/BoundaryConditions/PML.H
```diff
@@ -108,7 +112,7 @@ public:
     MultiSigmaBox(const amrex::BoxArray& ba, const amrex::DistributionMapping& dm,
                   const amrex::BoxArray& grid_ba, const amrex::Real* dx,
                   const amrex::IntVect& ncell, const amrex::IntVect& delta,
-                  const amrex::Box& regular_domain);
+                  const amrex::Box& regular_domain, amrex::Real v_sigma_sb);
```
EZoni:
```suggestion
                  const amrex::Box& regular_domain, const amrex::Real v_sigma_sb);
```

Source/BoundaryConditions/PML.cpp
```diff
@@ -58,7 +58,8 @@ namespace
 {
     static void FillLo (Sigma& sigma, Sigma& sigma_cumsum,
                         Sigma& sigma_star, Sigma& sigma_star_cumsum,
-                        const int olo, const int ohi, const int glo, Real fac)
+                        const int olo, const int ohi, const int glo, Real fac,
+                        amrex::Real v_sigma)
```
EZoni:
```suggestion
                        const amrex::Real v_sigma)
```

Source/BoundaryConditions/PML.cpp
```diff
@@ -74,19 +75,20 @@ namespace
...
+                p_sigma_star_cumsum[i-sslo] = (fac*(offset*offset*offset)/3._rt)/v_sigma;
             }
         });
     }
 
     static void FillHi (Sigma& sigma, Sigma& sigma_cumsum,
                         Sigma& sigma_star, Sigma& sigma_star_cumsum,
-                        const int olo, const int ohi, const int ghi, Real fac)
+                        const int olo, const int ohi, const int ghi, Real fac,
+                        amrex::Real v_sigma)
```
EZoni:
```suggestion
                        const amrex::Real v_sigma)
```

Source/BoundaryConditions/PML.cpp
```diff
@@ -139,7 +141,7 @@ namespace
 
 
 SigmaBox::SigmaBox (const Box& box, const BoxArray& grids, const Real* dx, const IntVect& ncell,
-                    const IntVect& delta, const amrex::Box& regdomain)
+                    const IntVect& delta, const amrex::Box& regdomain, amrex::Real v_sigma_sb)
```
EZoni:
```suggestion
                    const IntVect& delta, const amrex::Box& regdomain, const amrex::Real v_sigma_sb)
```

Source/BoundaryConditions/PML.cpp
```diff
@@ -182,14 +184,15 @@ SigmaBox::SigmaBox (const Box& box, const BoxArray& grids, const Real* dx, const
...
     } else {
-        define_multiple(box, grids, ncell, fac);
+        define_multiple(box, grids, ncell, fac, v_sigma_sb);
     }
 }
 
 void SigmaBox::define_single (const Box& regdomain, const IntVect& ncell,
-                              const Array<Real,AMREX_SPACEDIM>& fac)
+                              const Array<Real,AMREX_SPACEDIM>& fac,
+                              amrex::Real v_sigma_sb)
```
EZoni:
```suggestion
                              const amrex::Real v_sigma_sb)
```

Source/BoundaryConditions/PML.cpp
```diff
@@ -223,15 +226,15 @@ void SigmaBox::define_single (const Box& regdomain, const IntVect& ncell,
...
+                   olo, ohi, dhi, fac[idim], v_sigma_sb);
         }
     }
 
     amrex::Gpu::streamSynchronize();
 }
 
 void SigmaBox::define_multiple (const Box& box, const BoxArray& grids, const IntVect& ncell,
-                                const Array<Real,AMREX_SPACEDIM>& fac)
+                                const Array<Real,AMREX_SPACEDIM>& fac, amrex::Real v_sigma_sb)
```
EZoni:
```suggestion
                                const Array<Real,AMREX_SPACEDIM>& fac, const amrex::Real v_sigma_sb)
```

Source/BoundaryConditions/PML.cpp
```diff
@@ -492,9 +495,9 @@ SigmaBox::ComputePMLFactorsE (const Real* a_dx, Real dt)
 MultiSigmaBox::MultiSigmaBox (const BoxArray& ba, const DistributionMapping& dm,
                               const BoxArray& grid_ba, const Real* dx,
                               const IntVect& ncell, const IntVect& delta,
-                              const amrex::Box& regular_domain)
+                              const amrex::Box& regular_domain, amrex::Real v_sigma_sb)
```
EZoni:
```suggestion
                              const amrex::Box& regular_domain, const amrex::Real v_sigma_sb)
```

Source/BoundaryConditions/PML.cpp
```diff
@@ -536,7 +539,7 @@ PML::PML (const int lev, const BoxArray& grid_ba, const DistributionMapping& gri
           int do_moving_window, int /*pml_has_particles*/, int do_pml_in_domain,
           const bool do_multi_J,
           const bool do_pml_dive_cleaning, const bool do_pml_divb_cleaning,
-          int max_guard_EB,
+          int max_guard_EB, amrex::Real v_sigma_sb,
```
EZoni:
```suggestion
          int max_guard_EB, const amrex::Real v_sigma_sb,
```

Source/BoundaryConditions/PML.cpp
```diff
@@ -714,9 +717,10 @@ PML::PML (const int lev, const BoxArray& grid_ba, const DistributionMapping& gri
     }
 
     Box single_domain_box = is_single_box_domain ? domain0 : Box();
+    amrex::Real m_v_sigma_sb = v_sigma_sb;
```
EZoni: I think it could be better to declare this variable `m_v_sigma_sb` in PML.H as a member variable of the class `PML` (probably `private` would work, such as for other variables like `m_dive_cleaning`) and then initialize it in the [initializer list](https://en.cppreference.com/w/cpp/language/constructor) of the `PML` constructor in PML.cpp (where `m_dive_cleaning` and other similar variables are initialized). With all of this, the line here would not be needed anymore. Could you try to see if this also works?

Source/WarpX.H
```diff
@@ -1279,7 +1280,7 @@ private:
 #if (defined WARPX_DIM_RZ) && (defined WARPX_USE_PSATD)
     amrex::Vector<std::unique_ptr<PML_RZ> > pml_rz;
 #endif
-
+    amrex::Real v_particle_pml;
```
EZoni, line 1283: Let's add an empty line below `v_particle_pml` here, since the variables declared below that one (`moving_window_x` and `current_injection_position` do not belong do the group of PML variables):
```suggestion
    amrex::Real v_particle_pml;
    
```

Source/WarpX.cpp
```diff
@@ -853,6 +853,8 @@ WarpX::ReadParameters ()
         pp_warpx.query("do_pml_j_damping", do_pml_j_damping);
         pp_warpx.query("do_pml_in_domain", do_pml_in_domain);
         pp_warpx.query("do_similar_dm_pml", do_similar_dm_pml);
+        v_particle_pml = PhysConst::c;
+        pp_warpx.query("v_particle_pml", v_particle_pml);
```
EZoni: Could you add a few lines of code to check that `v_particle_pml` is not zero (so after the variable is queried from the inputs, to make sure that the user did not enter `warpx.v_particle_pml = 0.` by mistake and thus avoid divisions by zero at runtime)? You could use the functionality provided by `WARPX_ALWAYS_ASSERT_WITH_MESSAGE` and find examples of how this is used in WarpX by grepping it within the source code (using `git grep`).

EZoni (changes requested):
Thank you for your first PR, @amadou38! It looks great. I left only some minor suggestions in the comments below.

Source/BoundaryConditions/PML.cpp
```diff
@@ -716,7 +720,7 @@ PML::PML (const int lev, const BoxArray& grid_ba, const DistributionMapping& gri
     Box single_domain_box = is_single_box_domain ? domain0 : Box();
     // Empty box (i.e., Box()) means it's not a single box domain.
     sigba_fp = std::make_unique<MultiSigmaBox>(ba, dm, grid_ba_reduced, geom->CellSize(),
-                                               IntVect(ncell), IntVect(delta), single_domain_box);
+                                               IntVect(ncell), IntVect(delta), single_domain_box, m_v_sigma_sb);
```
RemiLehe:
```suggestion
                                               IntVect(ncell), IntVect(delta), single_domain_box, v_sigma_sb);
```

Source/BoundaryConditions/PML.cpp
```diff
@@ -839,7 +843,7 @@ PML::PML (const int lev, const BoxArray& grid_ba, const DistributionMapping& gri
 
         single_domain_box = is_single_box_domain ? cdomain : Box();
         sigba_cp = std::make_unique<MultiSigmaBox>(cba, cdm, grid_cba_reduced, cgeom->CellSize(),
-                                                   cncells, cdelta, single_domain_box);
+                                                   cncells, cdelta, single_domain_box, m_v_sigma_sb);
```
RemiLehe:
```suggestion
                                                   cncells, cdelta, single_domain_box, v_sigma_sb);
```

Source/BoundaryConditions/PML.cpp
```diff
@@ -536,10 +539,11 @@ PML::PML (const int lev, const BoxArray& grid_ba, const DistributionMapping& gri
           int do_moving_window, int /*pml_has_particles*/, int do_pml_in_domain,
           const bool do_multi_J,
           const bool do_pml_dive_cleaning, const bool do_pml_divb_cleaning,
-          int max_guard_EB,
+          int max_guard_EB, const amrex::Real v_sigma_sb,
           const amrex::IntVect do_pml_Lo, const amrex::IntVect do_pml_Hi)
     : m_dive_cleaning(do_pml_dive_cleaning),
       m_divb_cleaning(do_pml_divb_cleaning),
+      m_v_sigma_sb(v_sigma_sb),
```
RemiLehe: Do we really need to store `v_sigma_sb` in the `PML` object here? 
It seems that we could directly use the input argument (`v_sigma_sb`) in the body of the constructor here (see inline comments below).

Source/WarpX.cpp
```diff
@@ -853,6 +853,10 @@ WarpX::ReadParameters ()
...
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(v_particle_pml > 0.7_rt,
```
RemiLehe: Should this be `0.` instead of `0.7`?
```suggestion
        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(v_particle_pml > 0._rt,
```

Source/BoundaryConditions/PML.cpp
RemiLehe: Not that, if the above is indeed correct, then you could remove `m_v_sigma_sb` from `PML.H`

Source/WarpX.cpp
```diff
@@ -853,6 +853,10 @@ WarpX::ReadParameters ()
         pp_warpx.query("do_pml_j_damping", do_pml_j_damping);
         pp_warpx.query("do_pml_in_domain", do_pml_in_domain);
         pp_warpx.query("do_similar_dm_pml", do_similar_dm_pml);
+        v_particle_pml = PhysConst::c;
+        pp_warpx.query("v_particle_pml", v_particle_pml);
```
RemiLehe: Could you use the following code? This lets the user enter the velocity in units of c, which is more pratical.
```suggestion
        // Read `v_particle_pml` in units of the speed of light
        v_particle_pml = 1;
        pp_warpx.query("v_particle_pml", v_particle_pml);
        // Scale by the speed of light
        v_particle_pml = v_particle_pml * PhysConst::c;
```

RemiLehe (changes requested):
Thanks a lot for this PR! This looks great. 

I had a few minor inline comments. In addition, could you add documentation for the new parameters in [this file](https://github.com/ECP-WarpX/WarpX/blob/development/Docs/source/usage/parameters.rst). See the section on the PML parameters within this file. This could look like:
```
* ``warpx.v_particle_pml`` (`float`; default: 1)
    When `do_pml_j_damping = 1`, the assumed velocity of the particles to be absorbed in the PML, in units of the speed of light `c`.
```

Source/WarpX.cpp
```diff
@@ -853,6 +853,13 @@ WarpX::ReadParameters ()
         pp_warpx.query("do_pml_j_damping", do_pml_j_damping);
         pp_warpx.query("do_pml_in_domain", do_pml_in_domain);
         pp_warpx.query("do_similar_dm_pml", do_similar_dm_pml);
+        // Read `v_particle_pml` in units of the speed of light
+        v_particle_pml = 1;
```
EZoni: Just adding `_rt` for consistency:
```suggestion
        v_particle_pml = 1._rt;
```

Source/WarpX.cpp
```diff
@@ -853,6 +853,13 @@ WarpX::ReadParameters ()
...
+        pp_warpx.query("v_particle_pml", v_particle_pml);
+        // Scale by the speed of light
+        v_particle_pml = v_particle_pml * PhysConst::c;
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(v_particle_pml > 0._rt,
+            "input value for the velocity warpx.v_particle_pml of the macroparticle must be positive.");
```
EZoni: Maybe here it would make sense to check that the user's input `v_particle_pml` is such that both `0._rt < v_particle_pml` and `v_particle_pml <= 1._rt`, since a case `v_particle_pml > c` (after scaling by the speed of light) would not be acceptable.

I think something like this could work, to be done _before_ scaling `v_particle_pml` by the speed of light `c` (note also that we would need to update the error message, as in the following suggestion):
```cpp
        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(0._rt < v_particle_pml && v_particle_pml <= 1._rt,
            "Input value for the velocity warpx.v_particle_pml of the macroparticle must be in (0,1] (in units of c).");
```

EZoni (commented):
Thank you for the latest changes, @amadou38! I left a few more minor comments.

Source/WarpX.cpp
```diff
@@ -853,6 +853,13 @@ WarpX::ReadParameters ()
         pp_warpx.query("do_pml_j_damping", do_pml_j_damping);
         pp_warpx.query("do_pml_in_domain", do_pml_in_domain);
         pp_warpx.query("do_similar_dm_pml", do_similar_dm_pml);
+        // Read `v_particle_pml` in units of the speed of light
+        v_particle_pml = 1._rt;
+        pp_warpx.query("v_particle_pml", v_particle_pml);
+        // Scale by the speed of light
+        v_particle_pml = v_particle_pml * PhysConst::c;
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(0._rt < v_particle_pml && v_particle_pml <= 1._rt,
+            "Input value for the velocity warpx.v_particle_pml of the macroparticle must be in (0,1] (in units of c).");
```
```suggestion
        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(0._rt < v_particle_pml && v_particle_pml <= 1._rt,
            "Input value for the velocity warpx.v_particle_pml of the macroparticle must be in (0,1] (in units of c).");
        v_particle_pml = v_particle_pml * PhysConst::c;
```

amadou38: Yes. I just saw the error here. Thanks

RemiLehe (approved):

EZoni (approved):
Thank you, @amadou38! Looks good to me.

PR #3072, opened 2022-04-29
Use the same types for different pinned-memory particle containers

RemiLehe:
We use different pinned memory containers throughout the code. 

This PR attempts to clean up the code by unifying these different types and use `PinnedMemoryParticleContainer` everywhere.

Note that the actual type of `PinnedMemoryParticleContainer` was switched from `amrex::AmrParticleContainer<0,0, PIdx::nattribs, 0, amrex::PinnedArenaAllocator>` to ` WarpXParticleContainer::ContainerLike<amrex::PinnedArenaAllocator>`. This should however not affect WarpX since these two types are very similar.

### DIFF
--- a/Source/Diagnostics/BTDiagnostics.cpp
+++ b/Source/Diagnostics/BTDiagnostics.cpp
@@ -1036,7 +1036,7 @@ BTDiagnostics::InitializeParticleBuffer ()
         for (int isp = 0; isp < m_particles_buffer[i].size(); ++isp) {
             m_totalParticles_flushed_already[i][isp] = 0;
             m_totalParticles_in_buffer[i][isp] = 0;
-            m_particles_buffer[i][isp] = std::make_unique<PinnedMemoryParticleContainer>(&WarpX::GetInstance());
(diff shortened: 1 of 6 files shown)


atmyers (approved):

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -1036,7 +1036,7 @@ BTDiagnostics::InitializeParticleBuffer ()
         for (int isp = 0; isp < m_particles_buffer[i].size(); ++isp) {
             m_totalParticles_flushed_already[i][isp] = 0;
             m_totalParticles_in_buffer[i][isp] = 0;
-            m_particles_buffer[i][isp] = std::make_unique<PinnedMemoryParticleContainer>(&WarpX::GetInstance());
+            m_particles_buffer[i][isp] = std::make_unique<PinnedMemoryParticleContainer>(WarpX::GetInstance().GetParGDB());
```
RemiLehe, line 1039: Now that `PinnedMemoryParticleContainer` is of type `WarpXParticleContainer::ContainerLike<amrex::PinnedArenaAllocator>` instead of `amrex::AmrParticleContainer`, its constructor cannot take an `AmrCore object as input. 
This is why we are passed `GetParGDB` instead here.

PR #2939, opened 2022-03-08
Make error and info messages visually uniform

lucafedeli88:
The goal of this PR is to make "info" messages and error messages visually uniform (i.e., using the same prefix, the same style...).
There may be better ways to do so, therefore I am completely open to suggestions for improvements!

Anyway, the message styles are implemented in the new files `Utils/TextMsg.H` , and they use a new "automatic text wrapping" method implemented in `WarpXUtil.H` . 

Ideally, an error message would be raised with something like:
`WARPX_ALWAYS_ASSERT_WITH_MESSAGE(condition, msg)` (`AMREX_ALWAYS_ASSERT_WITH_MESSAGE` does not allow for `std::string` for the message). The result would be something like:

```
0::Assertion `dims == dims_compiled' failed, file "/home/luca/Projects/WarpX/Source/Utils/WarpXUtil.cpp", line 545, Msg: 
### ERROR   : The selected WarpX executable was built as '3'-dimensional, but
#             inputs file declares 'geometry.dims = 2'.
#            Please re-compile with a different WarpX_DIMS option or select the
#             executable name.
 !!!
```

"info" messages would be generated with something like
`amrex::Print() << Utils::TextMsg::Info(msg)` and the result would be like:

```
--- INFO    : Writing plotfile diags/diag1000002
```

Warnings are normally collected with the warning logger, but their style in case we ask warpx to warn immediately is implemented in `Utils/TextMsg.H` 

What do you think?

PR #3074, opened 2022-04-29
abort if BTD selected for RZ, and unsupported field/particle varnames

RevathiJambunathan:
Until BTD is added in RZ, we will abort when BackTransformed Diagnsostics is selected for RZ simulations.
Additionally, the code will also abort if unsupported fields for BTD are requested by user.

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -2006,6 +2006,7 @@ In-situ capabilities can be used by turning on Sensei or Ascent (provided they a
    Note that these averages do not respect the particle shape factor, and instead use nearest-grid point interpolation.
    Default is none.
    Parser functions for these field names are specified by ``<diag_name>.particle_fields.<field_name>(x,y,z,ux,uy,uz)``.
+   Also, note that this option is only available for ``<diag_name>.diag_type = Full``
 
 * ``<diag_name>.particle_fields_species`` (list of `strings`, optional)
          Species for which to calculate ``particle_fields_to_plot``.
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -1988,7 +1988,7 @@ In-situ capabilities can be used by turning on Sensei or Ascent (provided they a
 
 * ``<diag_name>.fields_to_plot`` (list of `strings`, optional)
     Fields written to output.
-    Possible scalar fields: ``part_per_cell`` ``rho`` ``phi`` ``F`` ``part_per_grid`` ``divE`` ``divB`` and ``rho_<species_name>``, where ``<species_name>`` must match the name of one of the available particle species. Note that ``phi`` will only be written out when do_electrostatic==labframe.
+    Possible scalar fields: ``part_per_cell`` ``rho`` ``phi`` ``F`` ``part_per_grid`` ``divE`` ``divB`` and ``rho_<species_name>``, where ``<species_name>`` must match the name of one of the available particle species. Note that ``phi`` will only be written out when do_electrostatic==labframe. Also, note that for ``<diag_name>.diag_type = BackTransformed``, the only scalar field currently supported is ``rho``.
     Possible vector field components in Cartesian geometry: ``Ex`` ``Ey`` ``Ez`` ``Bx`` ``By`` ``Bz`` ``jx`` ``jy`` ``jz``.
     Possible vector field components in RZ geometry: ``Er`` ``Et`` ``Ez`` ``Br`` ``Bt`` ``Bz`` ``jr`` ``jt`` ``jz``.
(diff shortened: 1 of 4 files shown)


Source/Diagnostics/BackTransformedDiagnostic.cpp
```diff
@@ -577,7 +577,9 @@ BackTransformedDiagnostic (Real zmin_lab, Real zmax_lab, Real v_window_lab,
       m_particle_slice_width_lab_(particle_slice_width_lab)
 {
 
-
+#ifdef WARPX_DIM_RZ
+    amrex::Abort("BackTransformed diagnostics is currently not supported with RZ");
+#endif
```
EZoni: Do we need the `amrex::Abort` both here and in `MultiDiagnostics.cpp` or can we have just a single one?

EZoni (commented):
Awesome, thank you @RevathiJambunathan! Looks good to me, I just left one question.

Docs/source/usage/parameters.rst
```diff
@@ -1988,7 +1988,7 @@ In-situ capabilities can be used by turning on Sensei or Ascent (provided they a
 
 * ``<diag_name>.fields_to_plot`` (list of `strings`, optional)
     Fields written to output.
-    Possible scalar fields: ``part_per_cell`` ``rho`` ``phi`` ``F`` ``part_per_grid`` ``divE`` ``divB`` and ``rho_<species_name>``, where ``<species_name>`` must match the name of one of the available particle species. Note that ``phi`` will only be written out when do_electrostatic==labframe.
+    Possible scalar fields: ``part_per_cell`` ``rho`` ``phi`` ``F`` ``part_per_grid`` ``divE`` ``divB`` and ``rho_<species_name>``, where ``<species_name>`` must match the name of one of the available particle species. Note that ``phi`` will only be written out when do_electrostatic==labframe. Also, note that for ``<diag_name>.diag_type = BackTransformed``, the only scalar field currently supported is ``rho``.
```
ax3l, line 1991: Just a quick question: why is `rho_<species>` not supported as well? :)

Source/Diagnostics/BackTransformedDiagnostic.cpp
RevathiJambunathan: This is the old BTD :) 
and the MultiDiagnostics.cpp is the new BTD

Docs/source/usage/parameters.rst
RevathiJambunathan: `rho` is obtained by performing Lorentz transform of `j` and `rho`
https://github.com/ECP-WarpX/WarpX/blob/1f3d2f3678f5ae2e934f1531bfcad4a2aae10b3f/Source/Diagnostics/ComputeDiagFunctors/BackTransformFunctor.cpp#L217 
We can add an enhancement once we have particles in the buffers and compute rho per species

But, currently there are some issues with the particle buffer -- fixed here
https://github.com/ECP-WarpX/WarpX/pull/3056

EZoni (commented):
Thanks for the explanations, Reva. I think we also prefer to replace the `amrex::Abort` with the newer `WARPX_ALWAYS_ASSERT_WITH_MESSAGE` introduced in #2939. Would it be possible to do that here, before we merge this PR?

RevathiJambunathan:
@EZoni 
I am not sure how exactly one would do this for an abort within compilation #ifdefs 
It seems to me that the Assert needs a condition followed by message.
From PR #2939 -- it seems to me that  `WARPX_ALWAYS_ASSERT_WITH_MESSAGE` replaces `AMREX_ALWAYS_ASSERT_WITH_MESSAGE` , since the latter did not support a string message.

EZoni:
> I am not sure how exactly one would do this for an abort within compilation #ifdefs
> It seems to me that the Assert needs a condition followed by message.

Maybe it could be done at least for the two `amrex::Abort`s in Source/Diagnostics/BTDiagnostics.cpp, where there are clear conditions? I think replacing `amrex::Abort` with `WARPX_ALWAYS_ASSERT_WITH_MESSAGE` is what @lucafedeli88 is trying to do in #2952, where possible.

RevathiJambunathan:
thanks for pointing me to that PR
`AMREX_ALWAYS_ASSERT_WITH_MESSAGE` is replaced with `WARPX_ALWAYS_ASSERT_WITH_MESSAGE` 
I have updated the PR for the conditionals

`amrex::Abort` is replaced with `amrex::Abort(Utils::TextMsg::Err()`
and I have done that for the RZ abort messages

Docs/source/usage/parameters.rst
ax3l: Ok, thanks for the details :)

EZoni (approved):
Thank you for updating the PR, @RevathiJambunathan! Looks good to me.

PR #3080, opened 2022-05-03
[WIP] Remove variable names that are not used for openPMD

RemiLehe:


Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -184,16 +184,9 @@ FullDiagnostics::InitializeFieldFunctorsRZopenPMD (int lev)
 
     // Species index to loop over species that dump rho per species
     int i = 0;
-    bool convert2cartesian;
-    int ncomp;
-    if (m_format == "openpmd") {
```
RemiLehe, line 189: This function is only called when `m_format` is `openpmd`, so no need for this `if` condition.

PR #3082, opened 2022-05-04
Close #3077 Fix RZ openPMD output with MR

### LINKED ISSUE #3077: openPMD diagnostics fail with RZ + mesh refinement
Observed on [this input script](https://github.com/ECP-WarpX/WarpX/blob/development/Examples/Tests/ElectrostaticSphereEB/inputs_rz_mr), but with the last line replaced by:
```
diag1.fields_to_plot = phi
diag1.format = openpmd
```
So as to activate the `openpmd` format. 

In that case, I am getting the following error:
```
amrex::Abort::0::Error: phi_0_real is not a known field output type in RZ geometry !!!
```

Note that this error does **not** show up if I disable mesh refinement by setting `amr.max_level = 0`.
It also does not show up if I keep mesh refinement but remove the line `diag1.format = openpmd` so as to enable to `plotfile` format.

RemiLehe:
Fix #3077. When writing RZ openPMD output, WarpX needs to internally convert the list of fields requested by the user (e.g. `rho Er`) into a list that looks like this: `rho_0_real rho_1_real rho_1_imag Er_0_real Er_1_real Er_1_imag`.

In the current version of WarpX, this is done by modifying the attribute `m_varnames` **in-place**. However, this causes issues with MR, because the function that modifies `m_varnames` in-place is called twice:

- For `lev=0`: `m_varnames = rho Er` -> `m_varnames = rho_0_real rho_1_real rho_1_imag Er_0_real Er_1_real Er_1_imag`.
- For `lev=1`: `m_varnames = rho_0_real rho_1_real rho_1_imag Er_0_real Er_1_real Er_1_imag` -> ???

However, we can take advantage of the fact that the field required by the user are also actually stored in another variable: `m_varnames_fields` (which is then copied to `m_varnames` ;  see [here](https://github.com/ECP-WarpX/WarpX/blob/development/Source/Diagnostics/Diagnostics.cpp#L166))

The new code does:
- For `lev=0`: `m_varnames_fields = rho Er` -> `m_varnames = rho_0_real rho_1_real rho_1_imag Er_0_real Er_1_real Er_1_imag`.
- For `lev=1`: `m_varnames_fields = rho Er` -> No need to update `m_varnames` (but it's actually not a problem if we do update `m_varnames`.

### DIFF
--- a/Examples/Tests/scraping/inputs_rz
+++ b/Examples/Tests/scraping/inputs_rz
@@ -41,7 +41,13 @@ electron.momentum_function_uy(x,y,z) = "if(x*x+y*y>0.0, -1.0*y/sqrt(x*x+y*y), 0.
 electron.momentum_function_uz(x,y,z) = "0"
 electron.do_not_deposit = 1
 
-diagnostics.diags_names = diag1
+diagnostics.diags_names = diag1 diag2
+
 diag1.intervals = 1
 diag1.diag_type = Full
 diag1.fields_to_plot = Ex
+
+diag2.intervals = 1
+diag2.diag_type = Full
+diag2.fields_to_plot = Er
+diag2.format = openpmd
--- a/Source/Diagnostics/FullDiagnostics.cpp
+++ b/Source/Diagnostics/FullDiagnostics.cpp
@@ -177,144 +177,137 @@ FullDiagnostics::InitializeFieldFunctorsRZopenPMD (int lev)
             warpx.get_pointer_current_fp(lev, dim)->nComp() == ncomp_multimodefab );
     }
 
-    amrex::Vector< std::string > temp_m_varnames(m_varnames);
-    // Clear any pre-existing vector to release stored data.
-    m_varnames.clear();
-    m_all_field_functors[lev].clear();
-
     // Species index to loop over species that dump rho per species
     int i = 0;
-    bool convert2cartesian;
-    int ncomp;
-    if (m_format == "openpmd") {
-        convert2cartesian = false;
-        ncomp = ncomp_multimodefab;
-    } else {
-        convert2cartesian = true;
-        ncomp = 1;
+    int ncomp = ncomp_multimodefab;
+    // This function is called multiple times, for different values of `lev`
+    // but the `varnames` need only be updated once.
+    bool update_varnames = (lev==0);
+    if (update_varnames) {
+        m_varnames.clear();
+        int n_rz = ncomp * m_varnames.size();
+        m_varnames.reserve(n_rz);
     }
-    int n_rz = ncomp * temp_m_varnames.size();
-    m_varnames.reserve(n_rz);
-    m_all_field_functors[lev].resize(temp_m_varnames.size());
+
+    // Reser field functors
+    m_all_field_functors[lev].clear();
+    m_all_field_functors[lev].resize(m_varnames_fields.size());
 
     // Fill vector of functors for all components except individual cylindrical modes.
-    for (int comp=0, n=temp_m_varnames.size(); comp<n; comp++){
-        if        ( temp_m_varnames[comp] == "Er" ){
+    for (int comp=0, n=m_varnames_fields.size(); comp<n; comp++){
+        if        ( m_varnames_fields[comp] == "Er" ){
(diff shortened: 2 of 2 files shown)


Examples/Tests/scraping/inputs_rz
```diff
@@ -41,7 +41,13 @@ electron.momentum_function_uy(x,y,z) = "if(x*x+y*y>0.0, -1.0*y/sqrt(x*x+y*y), 0.
...
+diagnostics.diags_names = diag1 diag2
+
 diag1.intervals = 1
 diag1.diag_type = Full
 diag1.fields_to_plot = Ex
+
+diag2.intervals = 1
+diag2.diag_type = Full
+diag2.fields_to_plot = Er
+diag2.format = openpmd
```
RemiLehe, line 53: This includes openPMD output for an MR RZ case (max_level=1 in this input script), in order to have automated testing for the new bug fix.

Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -177,145 +177,138 @@ FullDiagnostics::InitializeFieldFunctorsRZopenPMD (int lev)
...
-    amrex::Vector< std::string > temp_m_varnames(m_varnames);
-    // Clear any pre-existing vector to release stored data.
-    m_varnames.clear();
-    m_all_field_functors[lev].clear();
-
     // Species index to loop over species that dump rho per species
     int i = 0;
-    bool convert2cartesian;
-    int ncomp;
-    if (m_format == "openpmd") {
```
RemiLehe, line 189: This function is actually only called when `m_format == "openpmd"` (see the actual name of the function: `InitializeFieldFunctorsRZopenPMD`). Therefore, the `else` part of this `if` condition is never used.

RemiLehe: As a consequence, we also explicitly replace `convert2cartesian` by `false` is the rest of the function.

Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -177,145 +177,138 @@ FullDiagnostics::InitializeFieldFunctorsRZopenPMD (int lev)
             warpx.get_pointer_current_fp(lev, dim)->nComp() == ncomp_multimodefab );
     }
 
-    amrex::Vector< std::string > temp_m_varnames(m_varnames);
```
RemiLehe, line 180: The variable `temp_m_varnames` is completely removed in this PR, as it is not needed anymore.

Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -177,145 +177,138 @@ FullDiagnostics::InitializeFieldFunctorsRZopenPMD (int lev)
...
+    }
+
+    // Reser field functors
+    m_all_field_functors[lev].clear();
+    m_all_field_functors[lev].resize(m_varnames_fields.size());
 
     // Fill vector of functors for all components except individual cylindrical modes.
-    for (int comp=0, n=temp_m_varnames.size(); comp<n; comp++){
-        if        ( temp_m_varnames[comp] == "Er" ){
+    for (int comp=0, n=m_varnames_fields.size(); comp<n; comp++){
```
RemiLehe, line 197: Here we loop over `m_varnames_fields` instead of `m_varnames`, in order to avoid issues when this function is called a second time, **after** `m_varnames` was updated.

Examples/Tests/scraping/inputs_rz
```diff
@@ -41,7 +41,13 @@ electron.momentum_function_uy(x,y,z) = "if(x*x+y*y>0.0, -1.0*y/sqrt(x*x+y*y), 0.
...
 electron.do_not_deposit = 1
 
-diagnostics.diags_names = diag1
+diagnostics.diags_names = diag1 diag2
+
 diag1.intervals = 1
 diag1.diag_type = Full
 diag1.fields_to_plot = Ex
+
+diag2.intervals = 1
```
ax3l, line 50: To be technically clean for filtering tests, please add the compile flags here for openPMD:
https://github.com/ECP-WarpX/WarpX/blob/development/Regression/WarpX-tests.ini#L3351-L3352

ax3l (approved):
Thank you so much for the fix! :sparkling_heart:

PR #3085, opened 2022-05-05
Use maps to store physical properties in Source/Particles/SpeciesPhysicalProperties

lucafedeli88:
This PR uses maps to store physical properties in `SpeciesPhysicalProperties` , making it easier to expand the list of supported materials in the future.


lucafedeli88:
See https://github.com/ECP-WarpX/WarpX/pull/3090

PR #3089, opened 2022-05-05
Add STL files support in pywarpx

lgiacome:
With this PR we add to the Python interface the possibility to specify the embedded boundary as an STL file.

AMReX reads the inputs related to STL files from the `eb2` parser, so we add a new corresponding bucket to pywarpx to handle these inputs.

In a follow up PR it would probably make sense to move also the implicit function input to this bucket and to the eb2 parser.

### DIFF
--- a/Python/pywarpx/EB.py
+++ b/Python/pywarpx/EB.py
@@ -1,0 +1,9 @@
+# Copyright 2021 Roelof Groenewald
+#
+# This file is part of WarpX.
+#
+# License: BSD-3-Clause-LBNL
+
+from .Bucket import Bucket
+
+eb = Bucket('eb2')
--- a/Python/pywarpx/EB.py
+++ b/Python/pywarpx/EB.py
@@ -1,4 +1,4 @@
-# Copyright 2021 Roelof Groenewald
+# Copyright 2022 Lorenzo Giacomel
 #
 # This file is part of WarpX.
 #
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -921,15 +921,15 @@ class EmbeddedBoundary(picmistandard.base._ClassWithInit):
     an implicit function or as an STL file (ASCII or binary). In the latter case the
     geometry specified in the STL file can be scaled, translated and inverted.
     - implicit_function: Analytic expression describing the embedded boundary
-    - stl_file: STL file containing the embedded boundary geometry
-    - stl_scale: factor by which the STL geometry is scaled
-    - stl_center: vector by which the STL geometry is translated
+    - stl_file: STL file path (string), file contains the embedded boundary geometry
+    - stl_scale: factor by which the STL geometry is scaled (pure number)
+    - stl_center: vector by which the STL geometry is translated (in meters)
     - stl_reverse_normal: if True inverts the orientation of the STL geometry
     - potential: Analytic expression defining the potential. Can only be specified
                  when the solver is electrostatic. Optional, defaults to 0.
      Parameters used in the expressions should be given as additional keyword arguments.
     """
-    def __init__(self, implicit_function=None, stl_file=None, stl_scale=None, stl_center=None, stl_reverse_normal=None,
+    def __init__(self, implicit_function=None, stl_file=None, stl_scale=None, stl_center=None, stl_reverse_normal=False,
                  potential=None, **kw):
 
         assert stl_file is None or implicit_function is None, Exception('Only one between implicit_function and '
@@ -939,9 +939,9 @@ class EmbeddedBoundary(picmistandard.base._ClassWithInit):
         self.stl_file = stl_file
 
         if stl_file is None:
-            assert stl_scale is None, Exception('EB can be scaled only when using an stl file')
-            assert stl_center is None, Exception('EB can be translated only when using an stl file')
-            assert stl_reverse_normal is None, Exception('EB can be reversed only when using an stl file')
+            assert stl_scale is None, Exception('EB can only be scaled only when using an stl file')
+            assert stl_center is None, Exception('EB can only be translated only when using an stl file')
+            assert stl_reverse_normal is False, Exception('EB can only be reversed only when using an stl file')
 
         self.stl_scale = stl_scale
         self.stl_center = stl_center
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -917,8 +917,14 @@ class EmbeddedBoundary(picmistandard.base._ClassWithInit):
     """
     Custom class to handle set up of embedded boundaries specific to WarpX.
     If embedded boundary initialization is added to picmistandard this can be
-    changed to inherit that functionality.
+    changed to inherit that functionality. The geometry can be specified either as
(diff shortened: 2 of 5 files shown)


Python/pywarpx/picmi.py
```diff
@@ -923,15 +923,27 @@ class EmbeddedBoundary(picmistandard.base._ClassWithInit):
                  when the solver is electrostatic. Optional, defaults to 0.
      Parameters used in the expressions should be given as additional keyword arguments.
     """
-    def __init__(self, implicit_function, potential=None, **kw):
+    def __init__(self, implicit_function=None, stl_file=None, stl_scale=None, stl_center=None, stl_reverse_normal=None,
+                 potential=None, **kw):
         self.implicit_function = implicit_function
+        self.stl_file = stl_file
+
```
dpgrote, line 940: Can you add a check to make sure that both `implicit_function` and `stl_file` are not set?

dpgrote:
Hi @lgiacome, this looks good. I have just the one comment.

Python/pywarpx/picmi.py
lgiacome: Good idea thanks a lot! I've just added an assert

Python/pywarpx/EB.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2022 Lorenzo Giacomel
+#
+# This file is part of WarpX.
+#
+# License: BSD-3-Clause-LBNL
+
+from .Bucket import Bucket
+
+eb = Bucket('eb2')
```
dpgrote: One more comment. For clarity, it would be better if the object had the same name as the input group name. This should be `eb2 = Bucket('eb2')`.

lgiacome: Sure! It's done now :)

dpgrote (approved):
Looks good! Thanks!

Python/pywarpx/picmi.py
```diff
@@ -923,15 +923,31 @@ class EmbeddedBoundary(picmistandard.base._ClassWithInit):
...
     """
-    def __init__(self, implicit_function, potential=None, **kw):
+    def __init__(self, implicit_function=None, stl_file=None, stl_scale=None, stl_center=None, stl_reverse_normal=None,
+                 potential=None, **kw):
+
+        assert stl_file is None or implicit_function is None, Exception('Only one between implicit_function and '
+                                                                            'stl_file can be specified')
+
         self.implicit_function = implicit_function
+        self.stl_file = stl_file
```
ax3l, line 939: Since this is the user-facing interface, please update the doc string of this method:
- list the new option in the list
- let's start to describe each argument (parameter) of the function

lgiacome: Sure, as usual I forgot. Sorry 😓

Python/pywarpx/picmi.py
```diff
@@ -917,21 +917,43 @@ class EmbeddedBoundary(picmistandard.base._ClassWithInit):
...
     Custom class to handle set up of embedded boundaries specific to WarpX.
     If embedded boundary initialization is added to picmistandard this can be
-    changed to inherit that functionality.
+    changed to inherit that functionality. The geometry can be specified either as
+    an implicit function or as an STL file (ASCII or binary). In the latter case the
+    geometry specified in the STL file can be scaled, translated and inverted.
     - implicit_function: Analytic expression describing the embedded boundary
+    - stl_file: STL file containing the embedded boundary geometry
+    - stl_scale: factor by which the STL geometry is scaled
+    - stl_center: vector by which the STL geometry is translated
```
ax3l: in meters?

Python/pywarpx/picmi.py
```diff
@@ -917,21 +917,43 @@ class EmbeddedBoundary(picmistandard.base._ClassWithInit):
     """
     Custom class to handle set up of embedded boundaries specific to WarpX.
     If embedded boundary initialization is added to picmistandard this can be
-    changed to inherit that functionality.
+    changed to inherit that functionality. The geometry can be specified either as
+    an implicit function or as an STL file (ASCII or binary). In the latter case the
+    geometry specified in the STL file can be scaled, translated and inverted.
     - implicit_function: Analytic expression describing the embedded boundary
+    - stl_file: STL file containing the embedded boundary geometry
```
ax3l: Is this an STL _file path_ (string) or an STL _file object_?
Suggested wording for the former:
```
- stl_file: STL file path (string), file contains the embedded boundary geometry
```

Python/pywarpx/picmi.py
```diff
@@ -917,21 +917,43 @@ class EmbeddedBoundary(picmistandard.base._ClassWithInit):
...
+    - stl_scale: factor by which the STL geometry is scaled
+    - stl_center: vector by which the STL geometry is translated
+    - stl_reverse_normal: if True inverts the orientation of the STL geometry
     - potential: Analytic expression defining the potential. Can only be specified
                  when the solver is electrostatic. Optional, defaults to 0.
      Parameters used in the expressions should be given as additional keyword arguments.
     """
-    def __init__(self, implicit_function, potential=None, **kw):
+    def __init__(self, implicit_function=None, stl_file=None, stl_scale=None, stl_center=None, stl_reverse_normal=None,
```
ax3l: Would `stl_reverse_normal=False` be a better default than `None`?
```suggestion
    def __init__(self, implicit_function=None, stl_file=None, stl_scale=None, stl_center=None, stl_reverse_normal=False,
```

Python/pywarpx/picmi.py
```diff
@@ -917,21 +917,43 @@ class EmbeddedBoundary(picmistandard.base._ClassWithInit):
...
+        assert stl_file is None or implicit_function is None, Exception('Only one between implicit_function and '
+                                                                            'stl_file can be specified')
+
         self.implicit_function = implicit_function
+        self.stl_file = stl_file
+
+        if stl_file is None:
+            assert stl_scale is None, Exception('EB can be scaled only when using an stl file')
+            assert stl_center is None, Exception('EB can be translated only when using an stl file')
+            assert stl_reverse_normal is None, Exception('EB can be reversed only when using an stl file')
```
ax3l: See suggestion above:
```suggestion
            assert stl_reverse_normal is False, Exception('EB can be reversed only when using an stl file')
```

Python/pywarpx/picmi.py
```diff
@@ -917,21 +917,43 @@ class EmbeddedBoundary(picmistandard.base._ClassWithInit):
...
+                 potential=None, **kw):
+
+        assert stl_file is None or implicit_function is None, Exception('Only one between implicit_function and '
+                                                                            'stl_file can be specified')
+
         self.implicit_function = implicit_function
+        self.stl_file = stl_file
+
+        if stl_file is None:
+            assert stl_scale is None, Exception('EB can be scaled only when using an stl file')
```
ax3l: In all three: I think it might be more readable to write `... can only be ...`

Python/pywarpx/picmi.py
lgiacome: Yep I fixed this

ax3l (approved):
Nice, LGTM :) :+1:

PR #3091, opened 2022-05-06
Apply shift in BTD for zmax different than 0

RemiLehe:
In the BTD, the lab snapshot are done for `t_lab = 0, dt_snapshot, 2*dt_snapshot, 3*dt_snapshot, ...`

However, when `zmax > 0`, we cannot fully reconstruct the data for `t_lab=0`. This because, in this case, the Lorentz transform implies that we need data from **negative** `t_boost`, i.e. from before the beginning of the simulation.

In this PR, the lab snapshot are instead requested for `t_lab = t0, t0+dt_snapshot, t0+2*dt_snapshot, t0+3*dt_snapshot, ...`, where the offset `t0` is set so as to ensure that the Lorentz transform never requires data from negative time.

### DIFF
--- a/Source/Diagnostics/BTDiagnostics.cpp
+++ b/Source/Diagnostics/BTDiagnostics.cpp
@@ -233,8 +233,9 @@ BTDiagnostics::InitializeBufferData ( int i_buffer , int lev)
 {
     auto & warpx = WarpX::GetInstance();
     // Lab-frame time for the i^th snapshot
-    m_t_lab.at(i_buffer) = i_buffer * m_dt_snapshots_lab;
-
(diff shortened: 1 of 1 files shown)


RevathiJambunathan:
Thank you so much for this PR, Remi
I tested it on the input attached here 
[inputs.3d.txt](https://github.com/ECP-WarpX/WarpX/files/8643314/inputs.3d.txt)

for zmax = 1e-6, and for zmax - -1.e-6 

and both work!!
Screenshot from 2022-05-06 12-57-31

RevathiJambunathan (approved):

Issue #2462, opened 2021-10-21
Problems with Milestone Run in RZ


EZoni:
In preparation of the last milestone report, we ran some simulations of 2 plasma stages in RZ geometry, which did not seem to produce correct results.

Here's the input file that was used: [inputs_rz.txt](https://github.com/ECP-WarpX/WarpX/files/7392075/inputs_rz.txt). The run was performed on Summit using 4 nodes (24 GPUs).

Here's a movie showing the time evolution of the component Ez of the electric field:

https://user-images.githubusercontent.com/59625522/138336562-1299f4cb-a560-4178-8c10-169e17171b79.mp4

Note that this was maybe run on someone's branch instead of the `development` branch. Certainly one would need to rerun this test with the latest version of the code in order to confirm the issue.

RemiLehe:
For comparison, here are the result from a **3D** script, which should in principle simulate the same physics (albeit with different numerical parameters):
![img008500](https://user-images.githubusercontent.com/6685781/138529292-bf44348b-ced2-4f7a-838f-49d37eaaa6de.png)
This snapshot should correspond roughly to iteration 1500 from the above RZ movie. However, the fields significantly different.

Here is the corresponding input script: [input.txt](https://github.com/ECP-WarpX/WarpX/files/7401646/input.txt)

EZoni:
Small update: I re-ran the test with the latest version of WarpX and can confirm that I still see the issue reported here. Note that the RZ test runs quite quickly even on a local desktop computer (e.g. using as little as 8 CPU cores).

EZoni:
@RemiLehe One difference between the 3D input and the RZ input is about the plasma lenses.

In the 3D input we set:
```
particles.B_ext_particle_init_style = repeated_plasma_lens
particles.repeated_plasma_lens_period = 0.35
particles.repeated_plasma_lens_starts = 0.34
particles.repeated_plasma_lens_lengths = 0.002
particles.repeated_plasma_lens_strengths_B = 1369533.545948272
```

While in the RZ input we set:
```
particles.E_ext_particle_init_style = repeated_plasma_lens
particles.B_ext_particle_init_style = repeated_plasma_lens
particles.repeated_plasma_lens_period = 0.35
particles.repeated_plasma_lens_starts = 0.32
particles.repeated_plasma_lens_lengths = 0.03
particles.repeated_plasma_lens_strengths_E = 205291085141280.0
[...]

PR #3092, opened 2022-05-06
Change Default for `psatd.current_correction`

EZoni:
Change default for `psatd.current_correction`: always active, unless either a charge-conserving current deposition scheme (Esirkepov, Vay) or the div(E) scheme are used. 

Also, due to the new default, we do not require anymore that the current correction be used only in combination with the periodic single box option.

To-do:
- [x] Fix CI tests (will deactivate current correction to keep CI status as it was before this change)
- [x] Add warnings (if charge conservation is not guaranteed by either of the available methods)

P.S. An example of a realistic test case where charge conservation plays a significant role is given in #2462.

### DIFF
--- a/Source/WarpX.cpp
+++ b/Source/WarpX.cpp
@@ -1307,11 +1307,6 @@ WarpX::ReadParameters ()
             "Vay current deposition not implemented for Galilean algorithms"
         );
 
-        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
-            (!current_correction) || v_galilean_is_zero || (!fft_do_time_averaging),
-            "Current correction not implemented for averaged Galilean algorithm"
-        );
-
 #   ifdef WARPX_DIM_RZ
         update_with_rho = true;  // Must be true for RZ PSATD
 #   else
--- a/Source/WarpX.cpp
+++ b/Source/WarpX.cpp
@@ -1222,6 +1222,21 @@ WarpX::ReadParameters ()
         }
 
         pp_psatd.query("current_correction", current_correction);
+
+        if (current_correction == false &&
+            current_deposition_algo != CurrentDepositionAlgo::Esirkepov &&
+            current_deposition_algo != CurrentDepositionAlgo::Vay)
+        {
+            RecordWarning(
+                "Algorithms",
+                "The chosen current deposition algorithm does not guarantee"
+                " charge conservation, and no additional current correction"
+                " algorithm is activated in order to compensate for that."
+                " Lack of charge conservation may negatively affect the"
+                " results of the simulation.",
+                WarnPriority::low);
+        }
+
         pp_psatd.query("do_time_averaging", fft_do_time_averaging);
 
         if (WarpX::current_deposition_algo == CurrentDepositionAlgo::Vay)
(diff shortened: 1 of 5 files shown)


Source/WarpX.cpp
```diff
@@ -1303,11 +1306,6 @@ WarpX::ReadParameters ()
             "Vay current deposition not implemented for Galilean algorithms"
         );
 
-        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
-            (!current_correction) || v_galilean_is_zero || (!fft_do_time_averaging),
-            "Current correction not implemented for averaged Galilean algorithm"
-        );
```
EZoni, line 1310: I don't see obvious reasons why we should not be allowed to use the Galilean current correction with time averaging. Maybe to be discussed offline.

Source/WarpX.cpp
```diff
@@ -1204,23 +1204,41 @@ WarpX::ReadParameters ()
...
+        // current deposition (Esirkepov, Vay) or the div(E) cleaning scheme
+        // are used
+        current_correction = true;
+        if (WarpX::current_deposition_algo == CurrentDepositionAlgo::Esirkepov ||
+            WarpX::current_deposition_algo == CurrentDepositionAlgo::Vay ||
+            WarpX::do_dive_cleaning)
+        {
+            current_correction = false;
+        }
+
```
RemiLehe, line 1223: Could you confirm that the current_correction is only activated when we use a spectral (PSATD) Maxwell solver?
Also: could you mention the default value of `current_correction` in the documentation for this parameter?

Source/WarpX.cpp
RemiLehe: Hm, I agree we could remove this.

I seems that in general, the current correction is not implemented for the case where `J` is linear in time:
https://github.com/ECP-WarpX/WarpX/blob/development/Source/FieldSolver/SpectralSolver/SpectralAlgorithms/PsatdAlgorithmJLinearInTime.cpp#L427
But we already have an abort there, so no need to have another check in `WarpX.cpp`.

Discussed offline, let's add an `amrex::Abort` in Source/Evolve/WarpXEvolve.cpp, inside the `OneStep_multiJ` function.

RemiLehe (changes requested):

Source/WarpX.cpp
EZoni: Documentation updated in https://github.com/ECP-WarpX/WarpX/pull/3092/commits/cfd316c91bc66e22fe0dbba948dc91e442b5adae. The parameter itself is set with the `psatd` prefix, FDTD simulations will not parse it.

Source/WarpX.cpp
EZoni: Implemented in https://github.com/ECP-WarpX/WarpX/pull/3092/commits/ce22f93a6dd437a1daaede9dee2afd3d69fdabba (also, default changed to `False` for multi-J).

ax3l (approved):

ax3l:
Let me add bug labels: "bad" defaults giving buggy/subideal physics (just so we don't overlook it).

RemiLehe (approved):

PR #2732, opened 2022-01-11
Sort the argument list that Python writes to the inputs files

dpgrote:
This makes the inputs file that is written out by the Python interface more human readable.

PR #3093, opened 2022-05-06
Add Headers to Input File Written by PICMI

EZoni:
This adds headers such as `# algo` on the top of the block of input parameters that start with `algo.` in the input file written by PICMI with the method `write_input_file`.

I see this PR as a follow-up of #2732, whose purpose was to make such input files more human readable. In my opinion, the new headers make the input file even more human readable. Though maybe it's only because of the color scheme of my editor and/or it's only subjective anyways, not sure.

If you think this does not improve the readability, feel free to close the PR.

Here's a comparison of an example of input file without and with headers:

inputs_PICMI_old

inputs_PICMI_new

### DIFF
--- a/Python/pywarpx/WarpX.py
+++ b/Python/pywarpx/WarpX.py
@@ -5,6 +5,7 @@
 #
 # License: BSD-3-Clause-LBNL
 
+import re
 from . import Particles
 from .Algo import algo
 from .Amr import amr
@@ -103,8 +104,7 @@ class WarpX(Bucket):
 
             prefix_old = ''
             for arg in argv:
-                dot_found = (arg.find('.') >= 0)
-                prefix_new = arg[0:arg.find('.')] if dot_found else arg[0:arg.find('=')].rstrip()
+                prefix_new = re.split(' |\.', arg)[0]
                 if prefix_new != prefix_old:
                     ff.write(f'# {prefix_new}\n')
                     prefix_old = prefix_new
--- a/Python/pywarpx/WarpX.py
+++ b/Python/pywarpx/WarpX.py
@@ -6,6 +6,7 @@
 # License: BSD-3-Clause-LBNL
 
 import re
+
 from . import Particles
 from .Algo import algo
 from .Amr import amr
--- a/Python/pywarpx/WarpX.py
+++ b/Python/pywarpx/WarpX.py
@@ -101,7 +101,13 @@ class WarpX(Bucket):
 
         with open(filename, 'w') as ff:
 
+            prefix_old = ''
             for arg in argv:
+                dot_found = (arg.find('.') >= 0)
+                prefix_new = arg[0:arg.find('.')] if dot_found else arg[0:arg.find('=')].rstrip()
+                if prefix_new != prefix_old:
+                    ff.write(f'# {prefix_new}\n')
+                    prefix_old = prefix_new
                 ff.write(f'{arg}\n')
 
 warpx = WarpX('warpx')
--- a/Python/pywarpx/WarpX.py
+++ b/Python/pywarpx/WarpX.py
@@ -109,6 +109,8 @@ class WarpX(Bucket):
                 # before each group to make the input file more human readable
                 prefix_new = re.split(' |\.', arg)[0]
                 if prefix_new != prefix_old:
+                    if prefix_old != '':
+                        ff.write('\n')
                     ff.write(f'# {prefix_new}\n')
                     prefix_old = prefix_new
                 ff.write(f'{arg}\n')
--- a/Python/pywarpx/WarpX.py
+++ b/Python/pywarpx/WarpX.py
@@ -105,6 +105,8 @@ class WarpX(Bucket):
 
             prefix_old = ''
             for arg in argv:
+                # This prints the name of the input group (prefix) as a header
+                # before each group to make the input file more human readable
                 prefix_new = re.split(' |\.', arg)[0]
                 if prefix_new != prefix_old:
                     ff.write(f'# {prefix_new}\n')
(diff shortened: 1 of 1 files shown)


Python/pywarpx/WarpX.py
```diff
@@ -101,7 +101,13 @@ def write_inputs(self, filename='inputs', **kw):
 
         with open(filename, 'w') as ff:
 
+            prefix_old = ''
             for arg in argv:
+                dot_found = (arg.find('.') >= 0)
+                prefix_new = arg[0:arg.find('.')] if dot_found else arg[0:arg.find('=')].rstrip()
```
EZoni: These two lines are here to make sure that parameters such as `algo.particle_shape = 3` are preceded by the header `# algo`, while parameters that are missing a prefix, such as `max_step = 100`, are preceded by the header `# max_step`. There's probably smarter ways to do this, overall.

dpgrote: Adding the headers is a good idea since it does visually separate the sections.

An alternative is to do `prefix_new = re.split(' |\.', arg)[0]`.

EZoni: Thank you for the suggestion, @dpgrote. That's more compact, I committed the change in https://github.com/ECP-WarpX/WarpX/pull/3093/commits/d913c0b120246cc135b998acbad2cafcc8056e9d.

Python/pywarpx/WarpX.py
```diff
@@ -101,7 +103,12 @@ def write_inputs(self, filename='inputs', **kw):
 
         with open(filename, 'w') as ff:
 
+            prefix_old = ''
             for arg in argv:
+                prefix_new = re.split(' |\.', arg)[0]
+                if prefix_new != prefix_old:
+                    ff.write(f'# {prefix_new}\n')
```
dpgrote, line 114: Can you add a comment here explaining what this is for? Something like

\# This prints the name of the input group as a header before each group to improve the human readability of the input file.

dpgrote (approved):
Looks good! Thanks!

ax3l:
If you like, let's also add a newline before starting a new section. That helps finding sections without colorful syntax highlighting enabled

EZoni:
@ax3l Thank you for the suggestion! That makes it even more readable indeed, see new figure in the PR description.

dpgrote (approved):
Looks good!

ax3l (approved):

ax3l:
Very nice. As a follow-up, we could add a simple lookup map that translates the abbreviations of the base group in the comment to more reabable names or actual little descriptions when matched (`amr`, `algo`, ...)

Python/pywarpx/WarpX.py
```diff
@@ -101,7 +103,16 @@ def write_inputs(self, filename='inputs', **kw):
 
         with open(filename, 'w') as ff:
 
+            prefix_old = ''
             for arg in argv:
+                # This prints the name of the input group (prefix) as a header
+                # before each group to make the input file more human readable
+                prefix_new = re.split(' |\.', arg)[0]
```
ax3l, line 110: Just a note: this usage is slow and rebuilds the regex in the `for` loop, although it is constant. If you re-use the same regex multiple times, better pre-compile it.
https://docs.python.org/3/howto/regex.html#using-regular-expressions

General comments on the regex:
- Why do yo match a single space *or* `.`?
- What happens with [the prefix-free parameters](https://warpx.readthedocs.io/en/latest/usage/parameters.html#overall-simulation-parameters) `authors`, `stop_time` and `max_step`?
- does this handle multi-line AMReX parser strings well? They are allowed in the C++ inputs and technically a user can also express multi-line strings in arguments of the Python API
- I think you look for `re.search` and not the extra cost of `split`:
```
re.search(pattern, string, flags=0)

    Scan through string looking for the first location where the regular expression pattern produces a match, and return a corresponding [match object](https://docs.python.org/3/library/re.html#match-objects). Return None if no position in the string matches the pattern; note that this is different from finding a zero-length match at some point in the string.
```
- I find the overall patch a bit verbose for what it does. I would maybe it a function call, or at least add an empty newline in the code before the final `ff.write(f'{arg}\n')`. I would maybe even add it as a transformation to the `argv` list instead of patching the print loop.

dpgrote: I can respond. I agree that it is standard practice to precompile an expression that is used repeatedly, but "slow" is relative. Doing some timings of the `re.split` command shows that for the few dozen lines of output here, it will take ~30 microseconds. Using `re.compile` does reduce it to ~10 microseconds.

- It is checking for space or `.` to handle both cases, with or without a prefix.
- The interface is already concatenating multiline input into a single line
- Yes, `re.search` would work as well, but the time difference would be negligible
- It could be compressed some by combing the two `if` statements (and I think `prefix_old = None` is better)
```
if prefix_old is not None and prefix_new != prefix_old:
    ff.write('\n')
```

ax3l: - slow: yes, just a note really
- `It is checking for space or . to handle both cases, with or without a prefix.` - I don't get why no-prefix starts with a space, do you introduce that intentionally for the Python wrapper?
- multiline: cool, ok

rest: yup, ok. reads a bit verbose to me, but otherwise ok.

Python/pywarpx/WarpX.py
```diff
@@ -101,7 +103,16 @@ def write_inputs(self, filename='inputs', **kw):
...
+                if prefix_new != prefix_old:
+                    if prefix_old != '':
+                        ff.write('\n')
+                    ff.write(f'# {prefix_new}\n')
+                    prefix_old = prefix_new
                 ff.write(f'{arg}\n')
```
ax3l, line 117: Slight readability update
```suggestion

                ff.write(f'{arg}\n')
```

ax3l (approved):

Python/pywarpx/WarpX.py
dpgrote: Thanks for merging. For the case without a prefix, the text looks like this `max_step = 100`. It is looking for that space before the `=`.

PR #3096, opened 2022-05-09
Use queryWithParser to Parse v_particle_pml

amadou38:
Use queryWithParser, instead of query, to parse v_particle_pml, in order to support functions like sin, cos, sqrt, etc.


EZoni (dismissed):
Thank you, @amadou38! Looks good to me.

EZoni (commented):
Since we did not add any documentation of the new input `v_particle_pml` in #3070, shall we maybe add it with this PR as well? We can discuss offline how to do this.

EZoni (approved):
Thank you for adding the documentation, @amadou38! Looks good to me.

PR #3097, opened 2022-05-10
Stricter Tolerance for 3D Langmuir Tests

EZoni:
While looking into some benchmark resets for #3092, I noticed that we can probably decrease the relative tolerance used to compare the simulation results with the theoretical predictions for the 3D Langmuir tests. I think we can set the relative tolerance to `5e-2`, instead of the previous `15e-2`, thus making the analysis slightly more stringent for these 3D basic tests.


ax3l (approved):
LGTM :+1:

PR #3099, opened 2022-05-10
Doc: Improve Doxygen Links

ax3l:
- download also openPMD-api tag files & link them
- improve developer page: clearer links


EZoni (approved):
Thank you, @ax3l! Looks good to me.

PR #3100, opened 2022-05-11
Add new class `NamedComponentParticleContainer`

RemiLehe:
This PR adds a new class `NamedComponentParticleContainer` from which both `PinnedMemoryParticleContainer` and `WarpXParticleContainer` now derive.

The new class has some members and member functions for accessing particle components with names (strings) instead of integers. These members and member functions used to be in `WarpXParticleContainer` ; the aim of this PR is to make them accessible to `PinnedMemoryParticleContainer` - for diagnostic purposes.


Source/Particles/ParticleBoundaryBuffer.cpp
```diff
@@ -172,7 +172,7 @@ void ParticleBoundaryBuffer::gatherParticles (MultiParticleContainer& mypc,
                 if (!buffer[i].isDefined())
                 {
                     buffer[i] = pc.make_alike<amrex::PinnedArenaAllocator>();
-                    buffer[i].AddIntComp(false);  // for timestamp
+                    buffer[i].AddIntComp("timestamp", false);
```
RemiLehe, line 179: We can now use the function that gives a name to this quantity.

Source/Particles/ParticleBoundaryBuffer.cpp
```diff
@@ -215,7 +215,7 @@ void ParticleBoundaryBuffer::gatherParticles (MultiParticleContainer& mypc,
         if (!buffer[i].isDefined())
         {
             buffer[i] = pc.make_alike<amrex::PinnedArenaAllocator>();
-            buffer[i].AddIntComp(false);  // for timestamp
+            buffer[i].AddIntComp("timestamp", false);
```
RemiLehe, line 240: We can now use the function that gives a name to this quantity.

Source/Particles/ParticleBoundaryBuffer.cpp
```diff
@@ -273,7 +273,7 @@ int ParticleBoundaryBuffer::getNumParticlesInContainer(
     else return 0;
 }
 
-WarpXParticleContainer::ContainerLike<amrex::PinnedArenaAllocator> &
+PinnedMemoryParticleContainer &
```
RemiLehe, line 314: The corresponding `.H` file indicates that this function should return a `PinnedMemoryParticleContainer`.
Now that `PinnedMemoryParticleContainer` is a type of `NamedComponentParticleContainer`, it is not equivalent anymore to use `WarpXParticleContainer::ContainerLike<amrex::PinnedArenaAllocator>` instead of `PinnedMemoryParticleContainer`.

Source/Particles/PinnedMemoryParticleContainer.H
```diff
@@ -1,14 +1,8 @@
...
-#include <AMReX.H>
-#include <AMReX_Vector.H>
-#include <AMReX_AmrParticles.H>
-#include <AMReX_Particles.H>
+#include "NamedComponentParticleContainer.H"
 
-#include <AMReX_ParIter.H>
-
-using PinnedMemoryParticleContainer = WarpXParticleContainer::ContainerLike<amrex::PinnedArenaAllocator>;
+using PinnedMemoryParticleContainer = NamedComponentParticleContainer<amrex::PinnedArenaAllocator>;
```
RemiLehe, line 6: `PinnedMemoryParticleContainer` is now a kind of `NamedComponentParticleContainer`.

Source/Particles/WarpXParticleContainer.H
```diff
@@ -111,7 +111,7 @@ public:
  * derived classes, e.g., Evolve) or actual functions (e.g. CurrentDeposition).
  */
 class WarpXParticleContainer
-    : public amrex::ParticleContainer<0,0,PIdx::nattribs>
+    : public NamedComponentParticleContainer<amrex::DefaultAllocator>
```
RemiLehe, line 114: `WarpXParticleContainer` now derives from `NamedComponentParticleContainer`.

Source/Particles/WarpXParticleContainer.H
```diff
@@ -289,35 +289,6 @@ public:
     // split along diagonals (0) or axes (1)
     int split_type = 0;
```
RemiLehe, line 291: The functions below were moved to `NamedParticleContainer`.

Source/Particles/WarpXParticleContainer.H
```diff
@@ -379,11 +345,6 @@ public:
      void defineAllParticleTiles () noexcept;
 
 protected:
```
RemiLehe, line 347: The member variables below were moved to `NamedComponentParticleContainer`.

Source/Particles/NamedComponentParticleContainer.H
```diff
@@ -0,0 +1,120 @@
...
+#define NamedComponentParticleContainer_H_
+
+#include <AMReX.H>
+#include <AMReX_Vector.H>
+#include <AMReX_AmrParticles.H>
+#include <AMReX_Particles.H>
+#include <AMReX_ParIter.H>
+#include "Utils/TextMsg.H"
+
+struct PIdx
```
RemiLehe, line 21: This `struct` used to be in `WarpXParticleContainer_fwd.H`.

Source/Particles/NamedComponentParticleContainer.H
```diff
@@ -0,0 +1,127 @@
+#ifndef NamedComponentParticleContainer_H_
```
RemiLehe, line 7: @ax3l 
Here, should we split this file into a `_fwd.H` file and a regular `.H` file?

Source/Particles/NamedComponentParticleContainer.H
```diff
@@ -0,0 +1,128 @@
...
+ *  with a name (string) instead of doing so with an integer index.
+ *
+ *  This is done by storing maps that give the index of the component
+ *  that corresponds to a given string.
+ */
+template <template<class> class Allocator>
+class NamedComponentParticleContainer :
+public amrex::ParticleContainer<0,0,PIdx::nattribs,0,Allocator>
+{
+public:
```
RemiLehe, line 48: Most of the functions below where moved over from `WarpXParticleContainer`

Source/Particles/NamedComponentParticleContainer.H
```diff
@@ -0,0 +1,128 @@
+#ifndef NamedComponentParticleContainer_H_
```
ax3l, line 7: Please add copyright header :)

Source/Particles/NamedComponentParticleContainer.H
```diff
@@ -0,0 +1,128 @@
...
+#endif
+        nattribs
+    };
+};
+
+/** Particle Container class that allows to add/access particle components
+ *  with a name (string) instead of doing so with an integer index.
+ *
+ *  This is done by storing maps that give the index of the component
+ *  that corresponds to a given string.
```
ax3l, line 39: please add `@tparam T_Allocator ...`

Source/Particles/NamedComponentParticleContainer.H
```diff
@@ -0,0 +1,128 @@
...
+ */
+template <template<class> class Allocator>
```
ax3l: maybe rename to `T_Allocator` to avoid potential naming collisions.

Source/Particles/NamedComponentParticleContainer.H
ax3l: yes, if you have a use case where you can already use `NamedComponentParticleContainer<T>` without accessing its members?

Being able to use `PIdx` on its own is already worth it, I think.

Source/Particles/NamedComponentParticleContainer.H
```diff
@@ -0,0 +1,128 @@
...
+        // build up the map of string names to particle component numbers
+        particle_comps["w"]  = PIdx::w;
+        particle_comps["ux"] = PIdx::ux;
+        particle_comps["uy"] = PIdx::uy;
+        particle_comps["uz"] = PIdx::uz;
+    #ifdef WARPX_DIM_RZ
+        particle_comps["theta"] = PIdx::theta;
+    #endif
+    }
+    virtual ~NamedComponentParticleContainer() {}
```
ax3l:
```suggestion

    virtual ~NamedComponentParticleContainer() = default;
```

Source/Particles/NamedComponentParticleContainer.H
ax3l: maybe add `= amrex::DefaultAllocator` default?

Source/Particles/NamedComponentParticleContainer.H
```diff
@@ -0,0 +1,128 @@
...
+            particle_icomps[name] = NumIntComps();
+            particle_runtime_icomps[name] = NumIntComps() - 0;
+            AddIntComp(comm);
+        } else {
+            amrex::Print() << Utils::TextMsg::Info(
+                name + " already exists in particle_icomps, not adding.");
+        }
+    }
+
+    std::map<std::string, int> getParticleComps () const noexcept { return particle_comps;}
```
ax3l, line 165: Do you need sorted keys? Otherwise please use `std::unordered_map`

Source/Particles/NamedComponentParticleContainer.H
```diff
@@ -0,0 +1,128 @@
...
+
+        return tmp;
+    }
+
+    using amrex::ParticleContainer<0,0,PIdx::nattribs,0,Allocator>::NumRealComps;
+    using amrex::ParticleContainer<0,0,PIdx::nattribs,0,Allocator>::NumIntComps;
+    using amrex::ParticleContainer<0,0,PIdx::nattribs,0,Allocator>::AddRealComp;
+    using amrex::ParticleContainer<0,0,PIdx::nattribs,0,Allocator>::AddIntComp;
+
+    void AddRealComp (const std::string& name, bool comm=true)
```
ax3l, line 133: Please add doc strings to all methods, including getters :)
Doc strings should explain the abbreviations we use in the names.

Source/Particles/NamedComponentParticleContainer.H
ax3l, line 165: Consider to clarify in the  doc string:
- components means SoA or AoS or both?
- is `Real` implicit?
- can we use `Int` instead if `i`?
- What is different for "Runtime"?

Source/Particles/ParticleBoundaryBuffer.cpp
ax3l: Should we move `ContainerLike` from `WarpXParticleContainer` up to `NamedComponentParticleContainer` now?

Source/Particles/NamedComponentParticleContainer.H
```diff
@@ -0,0 +1,128 @@
...
+    particle_comps(p_comps),
+    particle_icomps(p_icomps),
+    particle_runtime_comps(p_rcomps),
+    particle_runtime_icomps(p_ricomps) {}
+
+    /* Move operators */
+    NamedComponentParticleContainer ( NamedComponentParticleContainer && ) = default;
+    NamedComponentParticleContainer& operator= ( NamedComponentParticleContainer && ) = default;
+
+    /** Create an empty particle container
```
ax3l, line 102: Consider adding the `ContainerLike` trait here as well? (Move up from `WarpXParticleContainer`?)

Source/Particles/NamedComponentParticleContainer.H
```diff
@@ -0,0 +1,128 @@
+#ifndef NamedComponentParticleContainer_H_
+#define NamedComponentParticleContainer_H_
+
+#include <AMReX.H>
+#include <AMReX_AmrParGDB.H>
+#include <AMReX_Particles.H>
+#include "Utils/TextMsg.H"
```
ax3l: - include WarpX includes before AMReX
- Includes missing for classes used from stdlib

Source/Particles/ParticleBoundaryBuffer.cpp
RemiLehe: `ContainerLike` is actually a type/function that is defined in `amrex::ParticleContainer`, I think, and then `WarpXParticleContainer` inherits it.

ax3l: Oh, I added it there, hah. Overlooked, sorry :)

Source/Particles/NamedComponentParticleContainer.H
RemiLehe: Hm, actually, I think that `ContainerLike` is a trait of the parent class, i.e. `amrex::ParticleContainer`

Source/Particles/NamedComponentParticleContainer.H
RemiLehe: OK, I will add the docstrings. 
Regarding the naming of the functions: note that this code was moved from `WarpXParticleContainer`, hence the names are the same as the previous version of these functions. We could potentially change the names, but that would imply changing many places in the code where we are **already** using these functions. This could be done in a second PR.

RemiLehe: OK, I tried that, but this breaks the code. This is because these functions are **already** used in the rest of the code - since this PR is just moving them from `WarpXParticleContainer` to `NamedParticleContainer` - and the part of the code that calls the function expects a `map`.

RemiLehe:
Thanks a lot for your review @ax3l, and sorry for the late update.
I updated the code (and in particular added docstrings), and addressed some of the comments directly on Github.

Source/Particles/NamedComponentParticleContainer.H
ax3l: I forgot I added it upstream, thx!

Source/Particles/NamedComponentParticleContainer.H
ax3l: Generally speaking, I would like to move this into its own file in WarpX.

That way, we could move `NamedComponentParticleContainer` into ABLASTR.

Source/Particles/NamedComponentParticleContainer.H
ax3l: Sounds good, yes that explains it.
Let's refactor in a follow-up.

Source/Particles/NamedComponentParticleContainer.H
```diff
@@ -0,0 +1,178 @@
...
+    std::map<std::string, int> getParticleComps () const noexcept { return particle_comps;}
+    /** Return the name-to-index map for the compile-time and runtime-time integer components */
+    std::map<std::string, int> getParticleiComps () const noexcept { return particle_icomps;}
+    /** Return the name-to-index map for the runtime-time real components */
+    std::map<std::string, int> getParticleRuntimeComps () const noexcept { return particle_runtime_comps;}
+    /** Return the name-to-index map for the runtime-time integer components */
+    std::map<std::string, int> getParticleRuntimeiComps () const noexcept { return particle_runtime_icomps;}
+
+protected:
+    std::map<std::string, int> particle_comps;
```
ax3l, line 174: I think these should all be the faster `std::unordered_map`s, because we don't care about the order of the `key` (`std::string`) values in the container:
https://en.cppreference.com/w/cpp/container/unordered_map

ax3l (approved):
Very good for now, leftover open comments can go into follow-up PRs :)

PR #3101, opened 2022-05-11
Fix mirrors for RZ

dpgrote:
This PR changes `NullifyMF` so that it loops over the components. This change allows it to work for the RZ solver which has multiple complements for each MF.

This helps fix some of the issues reported in #2462.


EZoni (approved):
Thank you, @dpgrote! Looks good to me. This helps fix some of the issues reported in #2462.

PR #3104, opened 2022-05-12
Fix Laser Injection in RZ

EZoni:
Laser injection with continuous injection was not working in RZ geometry, because we were checking that the injection box (strictly) contain the laser antenna both along R and Z, which always failed since the R coordinate of the laser was 0.0, coinciding with the box edge. Now we perform the check only along the z direction in RZ geometry.

This helps fix some of the issues reported in #2462.

### DIFF
--- a/Source/Particles/LaserParticleContainer.cpp
+++ b/Source/Particles/LaserParticleContainer.cpp
@@ -261,7 +261,21 @@ LaserParticleContainer::ContinuousInjection (const RealBox& injection_box)
 #else
     const Real p_pos[1] = {m_updated_position[2]};
 #endif
-    if ( injection_box.contains(p_pos) ){
+#if defined(WARPX_DIM_RZ)
+    // In RZ, check if laser enters the box only along Z. This is needed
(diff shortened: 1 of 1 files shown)


Source/Particles/LaserParticleContainer.cpp
```diff
@@ -261,7 +261,15 @@ LaserParticleContainer::ContinuousInjection (const RealBox& injection_box)
 #else
     const Real p_pos[1] = {m_updated_position[2]};
 #endif
-    if ( injection_box.contains(p_pos) ){
+#if defined(WARPX_DIM_RZ)
+    // In RZ, check if laser enters the box only along z
```
dpgrote: It would make for a long comment, but maybe also say that the special check is needed since "contains" fails in RZ, and explaining how it fails. But otherwise this is good.

EZoni: Thank you, Dave. I changed the comment, let me know if the new comment is clear enough.

dpgrote (approved):
Looks good!

ax3l (approved):
Nice, well done :+1:

PR #3106, opened 2022-05-12
Fix Breaking Pip 22.1 Logic

ax3l:
## No Build Isolation & Build Dependencies

Pip's `--no-build-isolation` says we have to have build-deps pre-installed. But Pip does offer no option to install those from `pyproject.toml` :cry:
https://github.com/pypa/pip/issues/9794

We want isolated builds, because they
- have a deterministic, relative path that we can cache with ccache (can probably also be done by setting `export TMPDIR`)
- can use external modules for our build dependencies, e.g., if pypi.org ships no `cmake` for a host arch (looking at you, PPC64le, until recently)

## Dropped in-tree-build

Pip now made `in-tree-build` the default.
```
option --use-feature: invalid choice: 'in-tree-build' (choose from '2020-resolver', 'fast-deps')
```

Great, if they would silently ignore the old option during the transition :cry: 

Of course, there is no way to force a minimal version of pip:
https://github.com/pypa/pip/issues/4145#issuecomment-266245512
> I can't see why a project would need to require a given version of pip.

:cry: 
https://github.com/pypa/pip/issues/11118

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -37,6 +37,7 @@ jobs:
         cmake --build build --config Debug --parallel 2
 
         python3 -m pip install --upgrade pip setuptools wheel
+        python3 -m pip install --upgrade cmake
         $env:PYWARPX_LIB_DIR="$(Get-Location | Foreach-Object { $_.Path })\build\lib\Debug\"
         python3 -m pip install . -vv --no-build-isolation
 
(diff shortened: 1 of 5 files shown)


EZoni (approved):
Thank you for the fix, @ax3l! I pre-approve so that it can be merged as soon as all builds pass.

PhilMiller (approved):

PR #3107, opened 2022-05-13
Fold Python tests into their corresponding base configurations, to better exploit ccache

PhilMiller:
Rather than building each of 1D, 2D, 3D, and RZ redundantly in a 'Python' CI job, just run the Python tests in the job of the corresponding configuration that will have already done the work of them.

### DIFF
--- a/Regression/WarpX-tests.ini
+++ b/Regression/WarpX-tests.ini
@@ -1044,8 +1044,8 @@ inputFile = Examples/Physics_applications/laser_acceleration/PICMI_inputs_1d.py
 runtime_params =
 customRunCmd = python3 PICMI_inputs_1d.py
 dim = 1
-addToCompileString = USE_PYTHON_MAIN=TRUE
-cmakeSetupOpts = -DWarpX_DIMS=1 -DWarpX_APP=OFF
+addToCompileString = USE_PYTHON_MAIN=TRUE USE_OPENPMD=TRUE QED=FALSE
+cmakeSetupOpts = -DWarpX_DIMS=1 -DWarpX_APP=OFF -DWarpX_OPENPMD=ON -DWarpX_QED=OFF
 target = pip_install
 restartTest = 0
 useMPI = 1
@@ -1782,8 +1782,8 @@ inputFile = Examples/Physics_applications/plasma_acceleration/PICMI_inputs_plasm
 runtime_params =
 customRunCmd = python3 PICMI_inputs_plasma_acceleration_1d.py
 dim = 1
-addToCompileString = USE_PYTHON_MAIN=TRUE
-cmakeSetupOpts = -DWarpX_DIMS=1 -DWarpX_APP=OFF
+addToCompileString = USE_PYTHON_MAIN=TRUE USE_OPENPMD=TRUE QED=FALSE
+cmakeSetupOpts = -DWarpX_DIMS=1 -DWarpX_APP=OFF -DWarpX_OPENPMD=ON -DWarpX_QED=OFF
 target = pip_install
 restartTest = 0
 useMPI = 1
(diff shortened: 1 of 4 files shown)


PhilMiller:
Note that I do expect this to lengthen the critical path of the existing builds a bit. The overall cost and job throughput should be improved, though, so during periods with any contention for runners, this should be a net benefit to productivity.

ax3l:
This looks good to me. I removed the need to re-build the WarpX objects for libs a while ago, so that is now a pure add-on.

I checked the builds in `development` and this PR:
```
1D:  3 -> 6
2D: 54 -> 68
3D: 54 -> 62
Py: 23
SP: 5 -> 5
RZ or noMPI: 16 -> 18
QED: 10 -> 10
EB: 15 -> 15 
```

Somehow I count 27 new builds, not 23. anything doubly tested now?

Regression/WarpX-tests.ini
```diff
@@ -1044,8 +1044,8 @@ inputFile = Examples/Physics_applications/laser_acceleration/PICMI_inputs_1d.py
 runtime_params =
 customRunCmd = python3 PICMI_inputs_1d.py
 dim = 1
-addToCompileString = USE_PYTHON_MAIN=TRUE
-cmakeSetupOpts = -DWarpX_DIMS=1 -DWarpX_LIB=ON -DWarpX_APP=OFF
+addToCompileString = USE_PYTHON_MAIN=TRUE USE_OPENPMD=TRUE QED=FALSE
+cmakeSetupOpts = -DWarpX_DIMS=1 -DWarpX_APP=OFF -DWarpX_OPENPMD=ON -DWarpX_QED=OFF
```
dpgrote, line 1048: Overall, this PR is a good idea and looks good. Just one comment (maybe tangential) - why do all of the 1D test cases have QED=OFF and OPENPMD=ON (even though many of them don't use openpmd)?

PhilMiller: They need configuration commonality across the tests, so that they can use a cached build taking 30 s rather than a fresh build taking 10 minutes.

dpgrote: Ok, sounds reasonable. Presumably at least one 1D test is using openpmd. But why QED=OFF? Is QED not supported in 1D?

PhilMiller: Some existing test was specifying that, and since none of the 1D tests use QED, that's how I got to commonality.

Probably not relevant to this, but QED won't compile for a mixed- (or single-, I think) precision build.

PhilMiller:
> This looks good to me. I removed the need to re-build the WarpX objects for libs a while ago, so that is now a pure add-on.
> 
> I checked the builds in `development` and this PR:
> 
> ```
> 1D:  3 -> 6
> 2D: 54 -> 68
> 3D: 54 -> 62
> Py: 23
> SP: 5 -> 5
> RZ or noMPI: 16 -> 18
> QED: 10 -> 10
> EB: 15 -> 15 
> ```
> 
> Somehow I count 27 new builds, not 23. anything doubly tested now?

I just downloaded the logs, and got the set of tests built as follows:
> `grep configuring ~/Downloads/wx* | awk '{print $3}' | sort | uniq -c | sort -n`

That doesn't show any duplicates (counts greater than 1). That rather suggests there were tests being omitted, slipping through the cracks between the different suites' conditions.

I count 184 tests in `Regression/WarpX-tests.ini`, and 183 distinct tests being run across those jobs. The missing one seems to be `Python_Langmuir`, which is specified as 3d. I'll see why that one is skipped.

PhilMiller:
Ah, there's this in `Regression/prepare_file_ci.py`:

```python
# Remove Python test (does not compile)
text = re.sub( '\[Python_Langmuir\]\n(.+\n)*', '', text)
```

ax3l:
Ah, looks outdated. Let's remove this line?

PhilMiller:
I also removed the `Langmuir_[xyz]` exclusion, since those tests themselves were deleted in 2020.

ax3l (approved):
Thanks a lot - LGTM :)

ax3l (approved):

PR #3109, opened 2022-05-13
[WIP] Improve Tool to Plot NCI Growth Rate

EZoni:
To-do:
- [x] Improve post-processing analysis
- [x] Test and improve based on feedback (Cc @oshapoval)

To be decided whether a Python script or a Jupyter notebook is preferred.

This is an example of the plots produced by the script:
<p align="center">
    
    
</p>

Might need to be improved in the future. For example, looks to me like it should be more robust with respect to the time interval chosen for the calculation of the growth rate (provided that such time interval is chosen within the exponential growth phase - linear growth in log scale).


oshapoval (changes requested):
Thanks Edoardo for this PR! I have  a few minor suggestions.

Tools/PostProcessing/plot_nci_growth_rate.py
```diff
@@ -0,0 +1,185 @@
...
+    spectrum_itmax = np.fft.fftshift(np.fft.fft2(data_itmax))[Nx//2:,Nz//2:]
+
+    # Compute growth rate
+    growth_rate = np.log(np.abs(spectrum_itmax)) - np.log(np.abs(spectrum_itmin))
+    growth_rate /= c*Dt
+
+    # Plot energy over time (based on reduced diagnostics, if available)
+    energy = np.loadtxt(os.path.join(path_diags_reduced, diag_name_reduced), usecols=2)
+    fig, ax = plt.subplots(dpi=100)
+    ax.plot(energy, '-')
```
oshapoval, line 158: Would it be possible here also to indicate time steps used for NCI growth rate calculation? Foe example, one can add something like this:
`ax.scatter([iter_min,iter_max], [energy[iter_min],energy[iter_max]], color='red')`

Tools/PostProcessing/plot_nci_growth_rate.py
```diff
@@ -0,0 +1,185 @@
...
+    cax = make_axes_locatable(ax).append_axes('right', size='5%', pad='5%')
+    # Min and max values for the colorbar of the plot
+    # (possibly change these here to get better plots)
+    vmax = np.max(growth_rate)
+    vmin = np.min(growth_rate)
+    im = ax.imshow(growth_rate, origin='lower', extent=[0,1,0,1],
+                   vmin=vmin, vmax=vmax, aspect='auto', cmap='RdBu_r')
+    cb = fig.colorbar(im, cax=cax)
+    set_axes(ax, iter_min, iter_max)
+    set_colorbar(cb)
```
oshapoval, line 183: One can also add the labels:
`ax.set_xlabel('Normalized $k_x$')`
`ax.set_ylabel('Normalized $k_z$')`

PR #3110, opened 2022-05-16
Fix bug in automatic_text_wrap

lucafedeli88:
Recently @EZoni spotted a bug in the `automatic_text_wrap` function (used to format warning messages). In some cases a word was lost in the formatted message. This PR should fix the bug.


ax3l (approved):

ax3l:
Thx for reporting @EZoni and the fix @lucafedeli88 :tada:

PR #3115, opened 2022-05-17
ABLASTR: localIDtoGlobal

ax3l:
Move the control of `localIDtoGlobal` to ABLASTR for particle IDs.

We want to reuse this in ImpactX et al.


RTSandberg (approved):
This looks good and makes sense to me

PR #3118, opened 2022-05-19
Extend MCC to accurately treat relativistic particles

roelof-groenewald:
This PR incorporates the description of a relativistic MCC routine from the PhD thesis of Chul-Hyun Lim, 2007, Univ. of California, Berkeley, "The interaction of energetic charged particles with gas and boundaries in the particle simulation of plasmas" into the existing MCC routine in WarpX. This is similar to the approach used by OOPD1.

I tested the new MCC implementation against the same drift tube problem presented in Lim's thesis. The results of the simulation comparison to theory is shown below. Note however that the old MCC implementation actually also produced the expected results so this isn't really a good test specifically of the accuracy of the relativistic MCC version versus the non-relativistic version. It does however, provide confidence in the new implementation being (largely) bug free.
new_mcc

~~Furthermore, the [Wentzel-Moliere elastic scattering model](https://www.sciencedirect.com/science/article/abs/pii/0168583X9395827R) was added to sample scattering angles during elastic collisions. This model reduces to the currently used isotropic scattering for low impact energy particles as can be seen in the histogram below:~~
wentzel_model

~~I would therefore propose that the Wentzel model becomes the standard elastic scattering model in WarpX and that we deprecate the purely isotropic scattering eventually.~~
[update: the Wentzel model is for scattering in the stationary frame of the target, we use the COM frame for collisions so we just do isotropic scattering, which naturally gives small angle scattering for energetic particles]

Items still to be done:
- [x] allow user input to specify energy range to calculate `nu_max`
- [x] clean-up back-scattering versus charge exchange collisions
- [x] add new details to documentation
- [ ] ~~add drift-tube problem to CI~~
- [x] confirm that new MCC implementation still reproduces benchmark results from Turner et al.


roelof-groenewald:
My understanding from Lim, 2007 is that they perform the collisions in a frame where the target is stationary using the analytical expression for energy loss as a function of scattering angle to determine the outgoing particle energy. This expression is given below where `W` is the energy loss, theta is the scattering angle and `E` is the incoming energy.
energy_loss
 
It feels more natural to me, however, for the MCC implementation to use the center-of-momentum (COM) frame to do collisions in. In this frame there is no energy loss for elastic collisions so the only calculation necessary (besides the Lorentz transformations) is to sample a scattering angle. I performed a test of the MCC implementation using the attached code snippet to confirm that doing the scattering in the COM frame recovers the same energy loss as a function of scattering angle as derived by Lim, 2007. The attached plot shows the results of this check and also shows the error if a relativistic implementation of MCC is not used for highly energetic particles - the classical MCC gives the low energy scattering result.
mcc_test
[collision_test.txt](https://github.com/ECP-WarpX/WarpX/files/8794703/collision_test.txt)

Examples/Physics_applications/capacitive_discharge/PICMI_inputs_1d.py
```diff
@@ -9,10 +9,132 @@
 
 import numpy as np
 from pywarpx import callbacks, fields, picmi
+from scipy.sparse import csc_matrix
+from scipy.sparse import linalg as sla
 
 constants = picmi.constants
 
 
+class PoissonSolver1D(picmi.ElectrostaticSolver):
```
roelof-groenewald, line 18: This can be removed and the MLMG solver used once https://github.com/ECP-WarpX/WarpX/issues/3123 is addressed.

Docs/source/usage/examples.rst
```diff
@@ -69,9 +69,9 @@ Uniform plasma
...
+The Monte-Carlo collision (MCC) model can be used to simulate electron and ion collisions with a neutral background gas. In particular this can be used to study capacitive discharges between parallel plates. The implementation has been tested against the benchmark results from Turner et al. in `Phys. Plasmas 20, 013507, 2013 <https://aip.scitation.org/doi/abs/10.1063/1.4775084>`_. The figure below shows a comparison of the ion density as calculated in WarpX (in May 2022 with `PR #3118 <https://github.com/ECP-WarpX/WarpX/pull/3118>`_) compared to the literature results (which can be found `here <https://aip.scitation.org/doi/suppl/10.1063/1.4775084>`__).
 
-.. figure:: https://user-images.githubusercontent.com/40245517/123883650-4f614680-d8fe-11eb-9302-92a282191e8f.png
+.. figure:: https://user-images...
```
roelof-groenewald: I'm just waiting for the 3rd test case to complete then I will post the accompanying plot and update this link.

roelof-groenewald:
I will revert this to work-in-progress for now since I found that although the Turner benchmarks are properly reproduced when using double precision for particles, it fails when using single precision now. So I will mark this as WIP until I can sort out that issue.

roelof-groenewald:
The mixed precision runs (single precision for `amrex::ParticleReal` but double precision for `amrex::Real`) now also reproduces the first Turner benchmark calculation (as the full double precision does). I was able to recover this agreement by using `amrex::Real` for the particle masses and energy in the collision energy calculation - it failed when I used `ParticleReal`s for these quantities. **Therefore a full single precision run would also fail**.
Is there a preferred way to force the use of a double for these quantities? I will switch to explicitly use `double` unless I'm directed otherwise.

roelof-groenewald:
The new implementation still gives results that match the Turner et al. 2013 benchmark results:
warpX_benchmark_Turner2013

I also ran the first two tests with mixed precision and they match expectation (after the above mentioned fix to use `double`s for the masses and energies).
warpX_benchmark_Turner2013_mixed_prec

Source/Particles/Collision/BackgroundMCC/ImpactIonization.H
```diff
@@ -123,7 +46,7 @@ public:
     */
     ImpactIonizationFilterFunc(
         MCCProcess const& mcc_process,
-        amrex::Real const mass,
+        double const mass,
```
dpgrote, line 52: Can you add a comment in the code why `double` is used here (explaining that double is needed for good accuracy with single precision particles)?

roelof-groenewald: Sure thing!

Source/Utils/ParticleUtils.H
```diff
@@ -31,6 +32,96 @@ namespace ParticleUtils {
...
+     * \brief Return (relativistic) particle energy given velocity and mass.
+     * Note the use of `double` since this calculation is prone to error with
+     * single precision.
+     *
+     * @param[in] v2 square of particle speed ( v dot v)
+     * @param[in] mass particle mass
+     * @param[out] energy particle energy in eV
+     */
+    AMREX_GPU_HOST_DEVICE AMREX_INLINE
+    void getEnergy ( amrex::ParticleReal const v2, double const mass,
```
dpgrote: I have used the notation that `u = gamma*v`, so the argument should be `u2`. Also, the comment should make it clear that this is `gamma*v`.

Source/Utils/ParticleUtils.H
```diff
@@ -31,6 +32,96 @@ namespace ParticleUtils {
...
+                     double& energy )
+    {
+        using std::sqrt;
+        using namespace amrex::literals;
+
+        // Compute c^2 since it is used repeatedly
+        constexpr auto c2 = PhysConst::c * PhysConst::c;
+        energy = mass * c2 * (sqrt(1.0_prt + v2 / c2) - 1.0_prt) / PhysConst::q_e;
```
dpgrote: This form of the expression is not good when the velocity is non relativistic (since it is subject t round off). The expression can be recast as `energy = mass * u2 / (gamma + 1.) / PhysConst::q_e;`
This might fix the need for double precision mass.

roelof-groenewald: That is very nice. Thanks for pointing that out! I'll change it in a bit. 
There would still be an issue with single precision in `ParticleUtils::getCollisionEnergy` though (which is where I first observed the problem).

Source/Utils/ParticleUtils.H
```diff
@@ -31,6 +32,96 @@ namespace ParticleUtils {
...
+        using std::sqrt;
+        using namespace amrex::literals;
+
+        // Compute c^2 since it is used repeatedly
+        constexpr auto c2 = PhysConst::c * PhysConst::c;
+
+        auto const beta2 = v2 / c2;
+        gamma = sqrt(1.0_prt + beta2);
+        auto const m_tot = M + gamma*m;
+        energy = (sqrt(m_tot*m_tot - m*m*beta2) - m - M) * c2 / PhysConst::q_e;
```
dpgrote: This expression could probably also be rewritten to be more tolerant to round off fro small v2.

Docs/source/developers/collisions.rst
```diff
@@ -0,0 +1,80 @@
+.. _developers-collisions:
```
ax3l: Do you like to add this in the theory chapter instead of the development chapter? :)

Docs/source/developers/collisions.rst
```diff
@@ -0,0 +1,80 @@
...
+
+The MCC implementation assumes that the background neutral particles are **thermal**, and are moving at non-relativistic velocities in the lab frame. For each simulation particle considered for a collision, a velocity vector for a neutral particle is randomly chosen. The particle velocity is then boosted to the stationary frame of the neutral through a Galilean transformation. The energy of the collision is calculated using the particle utility function, ``ParticleUtils::getCollisionEnergy()``, as
+
+    .. math:: E_{coll} = \sqrt{(\gamma mc^2 + Mc^2)^2 - (mv)^2} - (mc^2 + Mc^2)
```
ax3l: Doesn't this have to be on a new line to be parsed properly?
```suggestion
    .. math::

       E_{coll} = \sqrt{(\gamma mc^2 + Mc^2)^2 - (mv)^2} - (mc^2 + Mc^2)
```

Docs/source/developers/collisions.rst
```diff
@@ -0,0 +1,80 @@
...
+                \gamma_v c \\
+                v_x \\
+                v_y \\
+                v_z
+            \end{bmatrix}
+
+where :math:`\gamma` is the Lorentz factor of the relative speed between the lab frame and the COM frame, :math:`\beta_i = v^{COM}_i/c` is the i'th component of the relative velocity between the lab frame and the COM frame with
+
+    .. math::
+        v^{COM}_i = m v_i / (\gamma_v m + M)
```
ax3l: Doesn't this need a newline to be parsed properly?
```suggestion

       v^{COM}_i = m v_i / (\gamma_v m + M)
```

Docs/source/developers/collisions.rst
```diff
@@ -0,0 +1,80 @@
...
+^^^^^^^^^^
+
+The process is also the same as for elastic scattering except the excitation energy cost is subtracted from the particle energy. This is done by reducing the velocity before a scattering angle is chosen.
+
+Particle cooling due to elastic collisions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+It is straight forward to determine the energy a projectile loses during an elastic collision with another body, as a function of scattering angle, through energy and momentum conservation. See for example `Lim (2007) <https://search.library.berkeley.edu/permalink/01UCS_BER/s4lks2/cdi_proquest_miscellaneous_35689087>`_ for a derivation. The result is that given a projectile with mass :math:`m`, a target with mass :math:`M`, a scattering angle :math:`\theta`, and collision energy :math:`E`, the post collision energy of the projectile is given by
+
+    .. math:: E_{final} = E - [(E + mc^2)\sin^2\theta + Mc^2 - \cos(\theta)\sqrt{M^2c^4 - m^2c^4\sin^2\theta}] \\ \times\frac{E(E+2mc^2)}{(E+mc^2+Mc^2)^2 - E(E+2mc^2)\cos^2\theta}
```
ax3l: You could try this to make it more readable:
```suggestion
    .. math::

       \begin{aligned}
       E_{final} &= E - [(E + mc^2)\sin^2\theta + Mc^2 - \cos(\theta)\sqrt{M^2c^4 - m^2c^4\sin^2\theta}] \\
       &\times \frac{E(E+2mc^2)}{(E+mc^2+Mc^2)^2 - E(E+2mc^2)\cos^2\theta}
       \end{aligned}
```

Examples/Physics_applications/capacitive_discharge/PICMI_inputs_1d.py
ax3l: :+1: Let's also note this inline:
```suggestion
class PoissonSolver1D(picmi.ElectrostaticSolver):
    """This can be removed and the MLMG solver used once
       https://github.com/ECP-WarpX/WarpX/issues/3123 is addressed."""
```

ax3l:
> Is there a preferred way to force the use of a double for these quantities? I will switch to explicitly use double unless I'm directed otherwise.

Yes, write `double` and add an inline comment is the preferred way.

Docs/source/developers/collisions.rst
```diff
@@ -0,0 +1,80 @@
...
+
+The MCC implementation assumes that the background neutral particles are **thermal**, and are moving at non-relativistic velocities in the lab frame. For each simulation particle considered for a collision, a velocity vector for a neutral particle is randomly chosen. The particle velocity is then boosted to the stationary frame of the neutral through a Galilean transformation. The energy of the collision is calculated using the particle utility function, ``ParticleUtils::getCollisionEnergy()``, as
+
+    .. math:: E_{coll} = \sqrt{(\gamma mc^2 + Mc^2)^2 - (mv)^2} - (mc^2 + Mc^2)
+
+where :math:`v` is the speed of the particle as tracked in WarpX (really :math:`\gamma` times the particle speed), :math:`m` and :math:`M` are the rest masses of the simulation and background species, respectively. The Lorentz factor is defined in the usual way, :math:`\gamma = \sqrt{1 + v^2/c^2}`. The collision cross-sections for all scattering processes are evaluated at the energy as calculated above.
```
dpgrote: As I mentioned elsewhere, the standard is to use `u` when it is `gamma*v`.

Source/Utils/WarpXConst.H
```diff
@@ -39,6 +39,7 @@ namespace PhysConst
     static constexpr auto m_u   = static_cast<amrex::Real>( 1.66053906660e-27 );
     static constexpr auto hbar  = static_cast<amrex::Real>( 1.054571817e-34 );
     static constexpr auto alpha = static_cast<amrex::Real>( 0.007297352573748943 );//mu0/(4*MathConst::pi)*q_e*q_e*c/hbar;
+    static constexpr auto a0    = static_cast<amrex::Real>( 5.29177210903e-11 );//hbar/(m_e*c*alpha)
```
dpgrote: This, `a0`, doesn't seem to be used anywhere.

Docs/source/developers/collisions.rst
roelof-groenewald: I'm happy to move it to the theory chapter. I intended it to be a guide for how the collisions are implemented, specifically with regards to the equations used to calculate the collision energies and the transformations done to move between reference frames. Naturally having the equations in typeset would be much simpler for future users to get up to speed quickly rather than having to read through the code.
I do see that specifically with the inclusion of the difference in outcome between the classical and relativistic implementations, however that this would be better placed in the theory chapter.

Docs/source/developers/collisions.rst
roelof-groenewald: Building the docs locally it rendered fine but maybe there is some difference with different versions of `sphinx`, so I will add the new line to be safe.

Source/Utils/WarpXConst.H
roelof-groenewald: True. I used it before when I was looking at the Wentzel model but decided to skip that implementation. I'll remove the `a0` addition.

Source/Utils/ParticleUtils.H
roelof-groenewald: Yeap. I can write is as:
```
energy = 2.0 * m * M * u2 / (gamma + 1) / (M + m + sqrt(m*m + M*M + 2*m*M*gamma))
```
I'll quickly test to see if the benchmark calculation is successful when using these expressions and single precision particles (reverting to using `ParticleReal` for the energy and mass).

roelof-groenewald: I found that the calculation still requires `double`s for the energy and mass. But I did swap the expression for the above one since it is clearer how this reduces to the classical collision energy as gamma -> 1.

dpgrote (approved):
Looks good to me!

ax3l (approved):

PR #3119, opened 2022-05-19
Allow forced charge deposition of otherwise non-depositing species

PhilMiller:


Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -572,59 +572,60 @@ WarpXParticleContainer::DepositCharge (WarpXParIter& pti, RealVector const& wp,
                                        const int * const ion_lev,
                                        amrex::MultiFab* rho, int icomp,
                                        const long offset, const long np_to_depose,
-                                       int thread_num, int lev, int depos_lev)
+                                       int thread_num, int lev, int depos_lev,
+                                       bool deposit_on_demand)
 {
-    if (!do_not_deposit) {
-        WarpX& warpx = WarpX::GetInstance();
+    if (do_not_deposit && !deposit_on_demand) return;
```
PhilMiller, line 560: Turn off whitespace in the diff viewer to see this clearly.

roelof-groenewald (approved):
Looks good to me. For context for others, we have an interest to run what we call "non-interacting" simulations in which particles are not fed back to the field solver i.e. particles don't interact. This can be achieved by setting `do_not_deposit` for the particle species. However, currently if you do that you cannot access the particle density at all which we would still like to do for diagnostic purposes or to feed back in an iterative sense to find a steady state solution (people sometimes do this for beam simulations).

Python/pywarpx/_libwarpx.py
```diff
@@ -1062,7 +1062,7 @@ def clearParticleBoundaryBuffer(self):
         '''
         self.libwarpx_so.warpx_clearParticleBoundaryBuffer()
 
-    def depositChargeDensity(self, species_name, level, clear_rho=True, sync_rho=True):
+    def depositChargeDensity(self, species_name, level, clear_rho=True, sync_rho=True, deposit_on_demand=False):
```
ax3l, line 1065: please add `deposit_on_demand` to the docstring of parameters

Source/Particles/WarpXParticleContainer.H
```diff
@@ -214,7 +214,8 @@ public:
                                const long np_to_depose,
                                int thread_num,
                                int lev,
-                               int depos_lev);
+                               int depos_lev,
+                               bool deposit_on_demand = false);
```
ax3l, line 241: Please add a Doxygen string to thus member function. You can copy most from the one above.

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -572,59 +572,60 @@ WarpXParticleContainer::DepositCharge (WarpXParIter& pti, RealVector const& wp,
                                        const int * const ion_lev,
                                        amrex::MultiFab* rho, int icomp,
                                        const long offset, const long np_to_depose,
-                                       int thread_num, int lev, int depos_lev)
+                                       int thread_num, int lev, int depos_lev,
+                                       bool deposit_on_demand)
```
ax3l, line 558: The Doxygen string in this `.cpp` file should be moved to the corresponding `.H` file and the `deposit_on_demand` be added

Source/Particles/WarpXParticleContainer.cpp
ax3l: Looking at this again, it's pretty wild that we patch this here instead on doing the patch at the call side.

PhilMiller: I choose to implement it this way, and not do that refactoring, because there are a variety of call sites, with somewhat different use cases, and I didn't really want to mess about with them. I think a few of them were outside `WarpXParticleContainer` itself and its descendants, so it felt wrong to have them inspecting the `do_not_deposit` member variable to control their calls to the `DepositCharge` member function.

PhilMiller: It may be possible to refactor those call sites to use something more like the higher-level `DepositCharge` member that contains the loop over calls to this one, and then apply the flag there.

PhilMiller:
I'll fill in and clean up the documentation bits shortly.

Source/Particles/WarpXParticleContainer.cpp
ax3l: I have to admit I feel the existing `do_not_deposit` inside the call and the addition in this PR are not great design patterns.

There are 10 files with 1-4 calls each that call this function. I really would prefer refactoring and `if`-ing on the PC's `do_not_deposit` for the call.

PhilMiller:
Replaced by #3225

PR #3120, opened 2022-05-20
Fix unnecessary float to double promotions

lucafedeli88:
This PR fixes few (most likely) unwanted float to double promotions. I've found these issues using `clang-tidy`


ax3l (approved):
Thanks for catching this!

ax3l (approved):

PR #3122, opened 2022-05-20
Clarify error messages when parsing diagnostics input

hklion:
This clarifies a few of the error messages thrown in `Diagnostics.cpp`.


EZoni (approved):
Thank you for the fix, @hklion! Looks good to me.

PR #3125, opened 2022-05-23
Use forward declaration for WarpXFaceInfoBox

lucafedeli88:
This PR adds a forward declaration for WarpXFaceInfoBox, which is then used in the WarpX.H file

### Background

- https://warpx.readthedocs.io/en/latest/developers/faq.html#what-are-include-fwd-h-and-include-fwd-h-files
- https://en.wikipedia.org/wiki/Forward_declaration
- https://warpx.readthedocs.io/en/latest/developers/repo_organization.html#developers-cpp-includes-fwd


ax3l (approved):
Good idea :)

PR #3127, opened 2022-05-23
Docs: New Perlmutter Default Env, Microarch

ax3l:
The default environment on Perlmutter is now the one we already used, `PrgEnv-gnu`. Thus, we don't need to switch anymore.

Build CPU microarch: Tuned for Zen3, which is used on Login and Compute nodes alike: https://docs.nersc.gov/systems/perlmutter/system_details/ - already done since #3079

### DIFF
--- a/Tools/machines/perlmutter-nersc/perlmutter_warpx.profile.example
+++ b/Tools/machines/perlmutter-nersc/perlmutter_warpx.profile.example
@@ -47,6 +47,10 @@ export CRAY_ACCEL_TARGET=nvidia80
 # optimize CUDA compilation for A100
 export AMREX_CUDA_ARCH=8.0
 
+# optimize CPU microarchitecture for AMD EPYC 3rd Gen (Milan/Zen3)
+export CXXFLAGS="-march=znver3"
+export CFLAGS="-march=znver3"
+
 # compiler environment hints
 export CC=cc
 export CXX=CC
--- a/Tools/machines/perlmutter-nersc/perlmutter_warpx.profile.example
+++ b/Tools/machines/perlmutter-nersc/perlmutter_warpx.profile.example
@@ -49,8 +49,9 @@ export CRAY_ACCEL_TARGET=nvidia80
 export AMREX_CUDA_ARCH=8.0
 
 # optimize CPU microarchitecture for AMD EPYC 3rd Gen (Milan/Zen3)
-export CXXFLAGS="-march=znver3"
-export CFLAGS="-march=znver3"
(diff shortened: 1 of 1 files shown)


kngott:
I believe the plan should be to add the arch flag to AMReX, so do we want to add it right now, or should we let that happen in AMReX and not do double-duty?

Tools/machines/perlmutter-nersc/perlmutter_warpx.profile.example
```diff
@@ -3,7 +3,6 @@
 
 # required dependencies
 module load cmake/3.22.0
```
ax3l, line 5: Long term stability, because it does not hurt and Perlmutter uses lmod:
```suggestion
module load cmake/3.22.0
module load PrgEnv-gnu
```

Tools/machines/perlmutter-nersc/perlmutter_warpx.profile.example
```diff
@@ -48,6 +48,10 @@ export CRAY_ACCEL_TARGET=nvidia80
 # optimize CUDA compilation for A100
 export AMREX_CUDA_ARCH=8.0
 
+# optimize CPU microarchitecture for AMD EPYC 3rd Gen (Milan/Zen3)
+export CXXFLAGS="-march=znver3"
+export CFLAGS="-march=znver3"
```
ax3l, line 53: Actually, since #3079 we use the `CC` wrappers, that have these flags.

ax3l: Actually, since #3079 we use the `CC` wrappers, that have these flags.
```suggestion
# optimize CPU microarchitecture for AMD EPYC 3rd Gen (Milan/Zen3)
# note: the cc/CC/ftn wrappers below add those
#export CXXFLAGS="-march=znver3"
#export CFLAGS="-march=znver3"
```

ax3l: Tested via
```
objdump -d ./warpx > warpx.asm 
grep ymm warpx.asm 
```
and it's there :)

ax3l:
> I believe the plan should be to add the arch flag to AMReX, so do we want to add it right now, or should we let that happen in AMReX and not do double-duty?

Discussed: we need to update https://github.com/AMReX-Codes/amrex/blob/5d88558d2aeff93855a5c46ec8be1a7ef8545f70/Tools/GNUMake/sites/Make.nersc#L28

kngott (approved):

PR #3132, opened 2022-05-24
Write Inputs to `warpx_used_inputs`

ax3l:
Write a general file `warpx_used_inputs` that contains all inputs of a simulation.

This is needed, since we add & overwrite the inputs parameters in the inputs file with command-line options regularly. Writing the actually used inputs increases reproducibility of such workflows, especially in unsupervised scripted runs/optimizations.

- [x] tested
- [x] doc as subsection of run https://warpx.readthedocs.io/en/latest/usage/how_to_run.html#run

### DIFF
--- a/Source/Initialization/WarpXInitData.cpp
+++ b/Source/Initialization/WarpXInitData.cpp
@@ -62,12 +62,11 @@
 #include <algorithm>
 #include <array>
 #include <cctype>
+#include <fstream>
 #include <iostream>
 #include <memory>
 #include <string>
 #include <utility>
-#include <vector>
-#include <sstream>
 
 #include "FieldSolver/FiniteDifferenceSolver/FiniteDifferenceSolver.H"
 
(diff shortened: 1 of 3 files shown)


Source/Initialization/WarpXInitData.cpp
```diff
@@ -332,6 +331,17 @@ WarpX::PrintMainPICparameters ()
     }
 }
 
+void
+WarpX::WriteUsedInputsFile (std::string const & filename) const
+{
+    amrex::Print() << "For full input parameters, see the file: warpx_used_inputs\n";
```
dpgrote: This looks good. For the print statement, this should write out the filename instead of the fixed "warpx_used_inputs".

Source/Initialization/WarpXInitData.cpp
```diff
@@ -332,6 +331,17 @@ WarpX::PrintMainPICparameters ()
     }
 }
 
+void
+WarpX::WriteUsedInputsFile (std::string const & filename) const
+{
+    amrex::Print() << "For full input parameters, see the file: warpx_used_inputs\n\n";
```
```suggestion
    amrex::Print() << "For full input parameters, see the file: " << filename << "\n\n";
```

dpgrote (approved):
Looks good to me!

MaxThevenet (approved):

Source/Initialization/WarpXInitData.cpp
```diff
@@ -346,6 +345,17 @@ WarpX::PrintMainPICparameters ()
     }
 }
 
+void
+WarpX::WriteUsedInputsFile (std::string const & filename) const
+{
+    amrex::Print() << "For full input parameters, see the file: " << filename << "\n\n";
+
+    std::ofstream jobInfoFile;
```
ax3l, line 353: Oopsi fixed in #3234: Only the IO Processor should write this, not all MPI ranks.

PR #3144, opened 2022-05-27
remove extraneous WARPX_DIM_RZ

RTSandberg:
In the Gaussian Laser initialization, `WARPX_DIM_RZ` appeared in an`#if` condition and the immediately following `#elif` condition.  The second `WARPX_DIM_RZ`, in the `#elif`,  is vacuous and has been removed


EZoni (approved):
Thanks @RTSandberg, looks good! Not really a bug per se, but I added the corresponding labels, since the old code did not make much sense anyways.

PR #3146, opened 2022-05-30
CMake: Move Install Defaults

ax3l:
Moving setting defaults for install dirs to the install section.
This improves logic when using WarpX as subproject, e.g., in ABLASTR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,9 +51,6 @@ set_ccache()
 # temporary build directories
 set_default_build_dirs()
 
-# install directories
-set_default_install_dirs()
-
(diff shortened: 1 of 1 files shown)


ax3l:
Ok, actually not a good idea for subprojects that we want to control:
```
CMake Error at build_sp/_deps/fetchedopenpmd-src/CMakeLists.txt:1022 (install):
  install EXPORT given no DESTINATION!


CMake Error at build_sp/_deps/fetchedopenpmd-src/CMakeLists.txt:1027 (install):
  install FILES given no DESTINATION!
```

PR #3151, opened 2022-05-31
Add properties for neutron, hydrogen isotopes, helium isotopes

RemiLehe:
This adds the values of the mass and charge for the following species:
- neutron
- deuterium
- tritium
- helium3
- helium4

The mass data is taken from:
- [this NIST page](https://physics.nist.gov/cgi-bin/cuu/Value?mn) for the neutron
- [this NIST page](https://physics.nist.gov/cgi-bin/Compositions/stand_alone.pl?ele=H) for the isotopes of hydrogen
- [this NIST page](https://physics.nist.gov/cgi-bin/Compositions/stand_alone.pl?ele=He) for the isotopes of helium

In the cases of the isotopes, I subtracted the mass of the electron, and converted to the units of proton mass, using:
```
M = (M_nist * m_u - m_e) / m_p
```
where `m_u` is the atomic mass unit, `m_e` is the electron mass, `m_p` is the proton mass.

### DIFF
--- a/Source/Particles/SpeciesPhysicalProperties.H
+++ b/Source/Particles/SpeciesPhysicalProperties.H
@@ -18,7 +18,7 @@
 #include <string>
 
 enum struct PhysicalSpecies{unspecified=0, electron, positron, photon, neutron,
-    hydrogen, deuterium, tritium, helium, helium3, helium4, boron, boron10,
+    hydrogen, hydrogen2, hydrogen3, helium, helium3, helium4, boron, boron10,
     boron11, carbon, nitrogen, oxygen, argon, copper, xenon};
 
 namespace species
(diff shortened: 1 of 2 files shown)


Source/Particles/SpeciesPhysicalProperties.H
```diff
@@ -33,12 +34,20 @@ namespace species
             return PhysicalSpecies::positron;
         if( species=="photon" )
             return PhysicalSpecies::photon;
-        if( species=="hydrogen" )
-            return PhysicalSpecies::hydrogen;
-        if( species=="proton" )
+        if( species=="neutron" )
+            return PhysicalSpecies::neutron;
+        if( (species=="hydrogen") || (species=="proton") )
```
RemiLehe, line 39: Note that this is not very consistent with the way in which we treat "helium". 
For consistency, we should distinguish:
- hydrogen (includes the mass of the electron, averages over isotopes abundance)
- proton/hydrogen1 (does not include the mass of the electron, does not average over isotopes)
However, in order to preserve the code behavior, I did not introduce this change.

ax3l: Agreed. Didn't we discuss this for #3090 that @lucafedeli88 has open? :)

RemiLehe: Ah, yes. Actually the present PR duplicates #3090. I will close it for now.

PR #3154, opened 2022-06-01
Move warning logger in ablastr

lucafedeli88:
Move the warning manager to ABLASTR so we can create warnings in ABLASTR code sections as well :)


Source/FieldSolver/SpectralSolver/SpectralFieldDataRZ.cpp
```diff
@@ -6,9 +6,10 @@
  */
 #include "SpectralFieldDataRZ.H"
 
-#include "WarpX.H"
```
ax3l, line 9: Still needed for `WarpX::getCosts(lev);`:
```suggestion
#include "WarpX.H"
```

ax3l (approved):
That looks great to me, thank you very much!

Do you want to add again the forward declaration `WarnManager_fwd.H`?

PR #3156, opened 2022-06-02
Fix: ABLASTR rm MPIInitHelpers

ax3l:
For now, let's remove these helpers. They are not yet used in ImpactX and the routines need:
- moving in a separate directory
- renaming
- fixing of linking (at least on macOS) due to #3154 not yet being ported


EZoni (approved):
Looks good, thank you, @ax3l!

PR #3159, opened 2022-06-02
Fix: WarnManager

ax3l:
Fix concurrently landing PRs while moving the warning manager in #3154


EZoni (approved):
Thank you for the fix, @ax3l!

PR #2884, opened 2022-02-23
Add background stopping

dpgrote:
This adds background stopping to the types of Collisions. This version directly implements equation 14.12 from Goldston and Rutherford (rather than reading in an energy dependent stopping power table of data).

Todo:

- [x] Add documentation
- [x] Allow background density and temperature to be spatially and temporally variable
- [x] Add stopping on ions

PR #3163, opened 2022-06-06
For background stopping, added error checks of the user input

dpgrote:
This adds checks ensuring that the user input is sensible and doesn't violate the assumptions of the model.

Note that some of the checks (of user expressions) are done inside of the kernels and will only be active when compiled with debugging turned on.

Follow-up to #2884


Source/Particles/Collision/BackgroundStopping/BackgroundStopping.cpp
```diff
@@ -156,6 +159,9 @@ void BackgroundStopping::doBackgroundStoppingOnElectronsWithinTile (WarpXParIter
             amrex::Real const n_e = n_e_func(x, y, z, t);
             amrex::Real const T_e = T_e_func(x, y, z, t)*PhysConst::kb;
 
+            AMREX_ASSERT(n_e > 0.);
+            AMREX_ASSERT(T_e > 0.);
```
ax3l:
```suggestion
            AMREX_ASSERT(n_e > 0._rt);
            AMREX_ASSERT(T_e > 0._rt);
```

Source/Particles/Collision/BackgroundStopping/BackgroundStopping.cpp
```diff
@@ -84,6 +84,9 @@ BackgroundStopping::doCollisions (amrex::Real cur_time, amrex::Real dt, MultiPar
     amrex::Real species_mass = species.getMass();
     amrex::Real species_charge = species.getCharge();
 
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(species_mass > 0., "Error: With background stopping, the species mass must be > 0");
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(species_charge != 0., "Error: With background stopping, the species charge must be nonzero");
```
ax3l:
```C++
#include <cmath>
#include <limits>
```
and
```suggestion
    amrex::Real const species_mass = species.getMass();
    amrex::Real const species_charge = species.getCharge();

    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(species_mass > 0._rt,
        "Error: With background stopping, the species mass must be > 0");
    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(std::fabs(species_charge) > std::numeric_limits<amrex::Real>::epsilon(),
        "Error: With background stopping, the species charge must be nonzero");
```

note: the 2nd one should better be a comparison to an epsilon, since exact `==`/`!=` is flaky for floating point numbers.

Source/Particles/Collision/BackgroundStopping/BackgroundStopping.cpp
```diff
@@ -175,6 +181,8 @@ void BackgroundStopping::doBackgroundStoppingOnElectronsWithinTile (WarpXParIter
             amrex::Real const lambdadb3 = lambdadb*lambdadb*lambdadb;
             amrex::Real const loglambda = log((12._rt*pi/Zb)*(n_e*lambdadb3));
 
+            AMREX_ASSERT(loglambda > 0.);
```
ax3l:
```suggestion
            AMREX_ASSERT(loglambda > 0._rt);
```

Source/Particles/Collision/BackgroundStopping/BackgroundStopping.cpp
```diff
@@ -226,6 +234,9 @@ void BackgroundStopping::doBackgroundStoppingOnIonsWithinTile (WarpXParIter& pti
             amrex::Real const n_i = n_i_func(x, y, z, t);
             amrex::Real const T_i = T_i_func(x, y, z, t)*PhysConst::kb;
 
+            AMREX_ASSERT(n_i > 0.);
+            AMREX_ASSERT(T_i > 0.);
```
ax3l:
```suggestion
            AMREX_ASSERT(n_i > 0._rt);
            AMREX_ASSERT(T_i > 0._rt);
```

Source/Particles/Collision/BackgroundStopping/BackgroundStopping.cpp
```diff
@@ -247,6 +258,8 @@ void BackgroundStopping::doBackgroundStoppingOnIonsWithinTile (WarpXParIter& pti
             amrex::Real const lambdadb3 = lambdadb*lambdadb*lambdadb;
             amrex::Real const loglambda = log((12._rt*pi/Zb)*(n_i*lambdadb3));
 
+            AMREX_ASSERT(loglambda > 0.);
```
ax3l:
```suggestion
            AMREX_ASSERT(loglambda > 0._rt);
```

ax3l (commented):
Thank you! Small suggestions inlie.

I am actually wondering if all those particle operations should not use `amrex::ParticleReal` and `_prt` instead of the currently implemented `amrex::Real` type in all locations. Is that by accident?

Source/Particles/Collision/BackgroundStopping/BackgroundStopping.cpp
dpgrote: Thanks for your comments @ax3l. Regarding use of `amrex::ParticleReal`, there is some vagueness about when it should be used. My understanding is that `ParticleReal` is for the large particle data arrays to save space. But for other variables, does it matter? Is there extra cost converting between `float` and `double`? Does using doubles slow down the GPU? There is a lot of inconsistency through out WarpX with Real and ParticleReal.

Source/Particles/Collision/BackgroundStopping/BackgroundStopping.cpp
dpgrote: I agree that in general, checking for floating point equality is problematic. However, checking for `x != 0.` should be robust since `0` is an exact floating point value.

Source/Particles/Collision/BackgroundStopping/BackgroundStopping.cpp
ax3l: Yes, the idea is to use it for all particle quantities. Where still unsafe, we use explicit double instead, with the goal to be able to run full single precision long term.

Yes, runtime casts are surprisingly expensing, especially on GPU they are a special instruction.

Source/Particles/Collision/BackgroundStopping/BackgroundStopping.cpp
ax3l: Yes, that's true but -0.0 exits as well and maybe the value comes from a calc/parser thingy and is only close to zero.

Source/Particles/Collision/BackgroundStopping/BackgroundStopping.cpp
dpgrote: Ok. I can make everything here `ParticleReal`.

Source/Particles/Collision/BackgroundStopping/BackgroundStopping.cpp
dpgrote: In C++, the expression `-0.0 == 0.0` is true, so that is not an issue. Also, the value of `epsilon` is 2.22045e-16 so this check would fail since `fabs(species_charge)` would likely be 1.6e-19.

ax3l (approved):
Thank you for the PR & updates! :)

ax3l (approved):

PR #3169, opened 2022-06-08
Merge `FillBoundary` and `NodalSync` (`OneStep_multiJ`)

EZoni:
Follow up #3013 by merging `FillBoundary` and `NodalSync` also for the `OneStep_multiJ` method.

Not clear to me why this causes some checksums changes (for the tests `multi_J_2d_psatd` and `multi_J_2d_psatd_pml`). Looks to me like the old and new code should be identical, since the nodal synchronization is done at the same point in the PIC loop as the old code (only exception is the case of PML BCs with div cleaning, where the nodal synchronization of F and G was not done in the old code - see old and new `if` conditions): @WeiqunZhang and @RemiLehe, any opinion on this?

Also, the functions `NodalSync` do not seem to be used/needed anymore, so I removed them in the last commit here.


ax3l:
@EZoni sorry, needs a small rebase now due to #3167

RemiLehe:
@EZoni Hm, I am also not sure why this is causing a change in checksum.
Would you be able to isolate which change in the code introduced this change in checksum, by e.g. incrementally introducing the new `FillBoundary` call for `E`, then `B`, then `F`, then G?

EZoni:
@RemiLehe I checked what you suggested. The merge of `FillBoundary` and `NodalSync` for E, B and G works fine. It is the one for F that causes the checksum changes. Not sure why at this point.

RemiLehe (approved):

PR #3172, opened 2022-06-08
Summit: Fix BLAS++/LAPACK++ Modules

ax3l:
These modules are by default using GPU support, which points to CUDA 11.0. Due to C++17 compiler bugs in that early CUDA release, we cannot use that version and pull in circular dependencies to two different CUDA toolkit versions as soon as we use BLAS++ & LAPACK++: when compiling PSATD+RZ.

Nonetheless, we do not use BLAS++/LAPACK++ on GPU yet anyway. Thus, we can simply use the CPU versions of those modules.


EZoni (approved):
Thank you for looking into this, @ax3l! Just tested successfully on Summit, looks good to me.

RTSandberg:
They work for me, thanks Axel

PR #3174, opened 2022-06-10
Update glossary.rst 

n01r:
Add *Ascent, IPO, LTO, PICMI, PICSAR, and SENSEI* to the WarpX glossary.


EZoni (approved):
@n01r Thank you for updating the glossary and for your first contribution to WarpX! Looks good to me.

PR #3179, opened 2022-06-14
[WIP] Topic picmi reduced diagnostics

RTSandberg:
Expose reduced diagnostics to the PICMI interface


Examples/Tests/reduced_diagnostics_python/PICMI_inputs_reduced.py
```diff
@@ -0,0 +1,163 @@
...
+    lower_boundary_conditions=['periodic', 'periodic','periodic'],
+    upper_boundary_conditions=['periodic', 'periodic','periodic'],
+    lower_boundary_conditions_particles=['periodic', 'periodic','periodic'],
+    upper_boundary_conditions_particles=['periodic', 'periodic','periodic'],
+    warpx_max_grid_size=32)
+
+electrons = picmi.Species(
+        particle_type='electron',
+        name=f'electrons',
+        initial_distribution= picmi.UniformDistribution(
```
dpgrote, line 24: Please use consistent spacing. There should be a space after commas (except at the end of the line). I prefer using spaces around the equals sign, `name = 'electrons',` (and note that the 'f' is not needed). Also, add an empty line after each species setup.

Examples/Tests/reduced_diagnostics_python/PICMI_inputs_reduced.py
```diff
@@ -0,0 +1,163 @@
...
+# reduced diagnostics
+EP = picmi.ReducedDiagnostic(name='EP', type='ParticleEnergy',intervals=200)
+NP = picmi.ReducedDiagnostic(name='NP', type='ParticleNumber',intervals=200)
+EF = picmi.ReducedDiagnostic(name='EF', type='FieldEnergy',intervals=200)
+PP = picmi.ReducedDiagnostic(name='PP', type='ParticleMomentum',intervals=200)
+PF = picmi.ReducedDiagnostic(name='PF', type='FieldMomentum',intervals=200)
+MF = picmi.ReducedDiagnostic(name='MF', type='FieldMaximum',intervals=200)
+MR = picmi.ReducedDiagnostic(name='MR', type='RhoMaximum',intervals=200)
+FP = picmi.ReducedDiagnostic(name='FP',
+                                type='FieldProbe',
```
dpgrote, line 55: My preference is that arguments on separate lines should line up with the start of the first argument since it looks cleaner. This is just a suggestion and is at your own discretion.

Examples/Tests/reduced_diagnostics_python/PICMI_inputs_reduced.py
```diff
@@ -0,0 +1,163 @@
...
+                                intervals=200,
+                                #warpx_probe_geometry='Point',
+                                warpx_x_probe = 0.53125,
```
dpgrote, line 58: The `warpx` prefix is not needed for these arguments since they are essential for the description (and wouldn't be specific to WarpX).

Examples/Tests/reduced_diagnostics_python/PICMI_inputs_reduced.py
```diff
@@ -0,0 +1,163 @@
...
+    sim.add_species(
+        species,
+        layout=picmi.GriddedLayout(grid=grid, n_macroparticle_per_cell=[1,1,1]))
+
+# Add diagnostics
+sim.add_diagnostic(field_diag)
+for rd in reduced_diags:
+    sim.add_reduced_diagnostic(rd)
+
+sim.initialize_inputs()
```
dpgrote, line 157: These two lines are not needed since they will be called during `sim.step`.

Examples/Tests/reduced_diagnostics_python/PICMI_inputs_reduced.py
```diff
@@ -0,0 +1,163 @@
...
+sim.initialize_warpx()
+
+sim.write_input_file('inputs_PICMI_reduced')
```
dpgrote, line 160: This is also not needed since nothing will be done with the `inputs_PICMI_reduced` file.

Python/pywarpx/Particles.py
```diff
@@ -1,12 +1,13 @@
...
+# 2022 Ryan Sandberg
 #
 # This file is part of WarpX.
 #
 # License: BSD-3-Clause-LBNL
 
 from .Bucket import Bucket
 
-particles = Bucket('particles', species_names=[], rigid_injected_species=[])
+particles = Bucket('particles', species_names=[], photon_species=[],rigid_injected_species=[])
```
dpgrote, line 10: Is there a reason that the photon species are separated out? I don't see any difference in their setup.

Python/pywarpx/WarpX.py
```diff
@@ -24,6 +24,7 @@
 from .Particles import particles, particles_list
 from ._libwarpx import libwarpx
 
+reduced_diagnostics = Bucket('reduced_diagnostics', _reduced_diagnostics_dict={})
```
dpgrote, line 27: This should follow the pattern for the other instances and should be defined in a separate file, `ReducedDiagnostics.py`. Use the files `Diagnostics.py` as the starting point.

Python/pywarpx/WarpX.py
```diff
@@ -32,6 +33,7 @@ class WarpX(Bucket):
 
     def create_argv_list(self):
         argv = []
+        warpx.reduced_diags_names = reduced_diagnostics._reduced_diagnostics_dict.keys()
```
dpgrote, line 36: This should follow the pattern of how it is done with diagnostics.

Python/pywarpx/picmi.py
```diff
@@ -121,25 +126,45 @@ def initialize_inputs(self, layout,
...
-                                             save_particles_at_xlo = self.save_particles_at_xlo,
-                                             save_particles_at_xhi = self.save_particles_at_xhi,
-                                             save_particles_at_ylo = self.save_particles_at_ylo,
-                                             save_particles_at_yhi = self.save_particles_at_yhi,
-                                             save_particles_at_zlo = self.save_particles_at_zlo,
-                                             save_particles_at_zhi = self.save_particles_at_zhi,
-                                             save_particles_at_eb = self.save_particles_at_eb,
-                                             save_previous_position = self.save_previous_position,
-                                             do_not_deposit = self.do_not_deposit)
+        if self.species_type == 'photon':
```
dpgrote, line 129: The same question as above - is there a reason that photon species are handled separately?

Python/pywarpx/picmi.py
```diff
@@ -1022,6 +1050,8 @@ class Simulation(picmistandard.PICMI_Simulation):
     # by Simulation objects to permit multiple WarpX runs simultaneously.
     extension = pywarpx.libwarpx
 
+    reduced_diagnostics = []
```
dpgrote, line 1053: This should be defined in the `init` method, as `self.reduced_diagnostics = []`.

Regression/WarpX-tests.ini
```diff
@@ -64,7 +64,7 @@ branch = 18d0a2861d31c52c65752a1d5856f54e08699ce3
 [source]
 dir = /home/regtester/AMReX_RegTesting/warpx
 branch = development
-cmakeSetupOpts = -DAMReX_ASSERTIONS=ON -DAMReX_TESTING=ON -DWarpX_LIB=ON
+cmakeSetupOpts = -DAMReX_ASSERTIONS=ON -DAMReX_TESTING=ON -DWarpX_LIB=ON -DopenPMD_USE_ADIOS1=OFF
```
dpgrote, line 67: Why is this needed? This will affect all of the CI and should probably be in a separate PR.

Examples/Tests/reduced_diagnostics_python/PICMI_inputs_reduced.py
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/env python3
```
dpgrote, line 1: This file should be in the same directory as the other reduced diagnostic tests, `Examples/Tests/reduced_diags`. This separate directory is not needed.

dpgrote:
Thanks for this PR! Overall it looks good, but I do have a number of comments and questions.

Python/pywarpx/Particles.py
RTSandberg: WarpX requires the line `particles.photon_species = photons` and this seemed like an easy way to be sure that line gets into the input deck. I only added photons in the first place because the reduced diagnostics test has photons.

Examples/Tests/reduced_diagnostics_python/PICMI_inputs_reduced.py
RTSandberg: ok, thanks for the style suggestions

Examples/Tests/reduced_diagnostics_python/PICMI_inputs_reduced.py
RTSandberg: it does look cleaner.  Done!

Examples/Tests/reduced_diagnostics_python/PICMI_inputs_reduced.py
RTSandberg: ok, this is an artifact of the development and can be refactored out

Python/pywarpx/WarpX.py
RTSandberg: The reason for the distinction is the way the reduced diagnostics are implemented in WarpX.  As opposed to regular diagnostics, indicated in the deck with `diagnostics.diags_names = <diag 1> <diag 2> ...`, the reduced diagnostics are called by a line of the form`warpx.reduced_diags_names = <reduced diag 1> <reduced diag 2> ...`

Examples/Tests/reduced_diagnostics_python/PICMI_inputs_reduced.py
RTSandberg: On a related note, the reduced diagnostics aren't in the PICMI standard.  In a sense, this whole PR is maybe breaking with the spirit of PICMI and is something I wanted to ask about

RTSandberg: Re: your comment, I was borrowing the `handle_init` function, which requires the warpx prefix.  That said, it is awkward and worth removing.

RTSandberg: actually, this doesn't have a straighforward simplification.  

-  to keep all reduced diagnostics in one PICMI class, some diagnostic specific inputs are passed as keyword arguments
- all keyword arguments have to be prefixed with the code prefix or errors are thrown

- one alternative is to have separate diagnostic classes for more complicated reduced diagnostics
- I chose not to pursue this direction, but it could be revisited if we think it is more annoying to prefix keyword arguments with "warpx_" than to have multiple reduced diagnostics classes

dpgrote: Regarding the reduced diagnostics not being in the PICMI standard, that is OK for now. The interface can be worked out here and then later moved to PICMI.

Having separate classes for the more complicated diagnostics might be a good idea. This would make it easier to do error checking on the input arguments. Each diagnostic type that has different input parameters would get its own class. This would fit the pattern of the PICMI standard better, with all of the arguments explicitly defined in the class constructor.

dpgrote: Thinking about this again, it may be a bit more work now, but would save work in the long term if this was added to the PICMI standard now (instead of only in WarpX). The classes would only have to be written once and the extra stuff needed to define the classes in WarpX would not be needed. This PR would be put on WIP status until that is ready and then this PR would update the required version of the picmi standard.

Python/pywarpx/Particles.py
dpgrote: Ok, this is good. Though see my comment regarding setting up the species. And add a space after the comma.

Python/pywarpx/WarpX.py
dpgrote: Ok, I see. Though, I think that this really needs to be changed and the input argument should be `diagnostics.reduced_diags_names` but that would be a separate PR.

A way to make this cleaner perhaps would be to put this line in the `ReducedDiagnostic.initialize_inputs` method after adding the name to `WarpX.reduced_diagnostics._reduced_diagnostics_dict`.

Python/pywarpx/picmi.py
```diff
@@ -39,6 +39,7 @@ class constants:
 picmistandard.register_constants(constants)
 
 class Species(picmistandard.PICMI_Species):
+    species_type = None
```
dpgrote, line 42: The `species_type` attribute doesn't seem to be needed. You could remove it and equivalently check for `self.particle_type == 'photon'`.

Python/pywarpx/picmi.py
dpgrote: Having the separate `photon_species` list is good, but this separate setup of self.species is not needed. The `self.species.species_type` attribute isn't used anywhere.

dpgrote (commented):
This PR looks good overall and is definitely needed. Don't be put off by my critiques.

Regression/WarpX-tests.ini
RTSandberg: Oops - this is just something I have to do on my local computer and I forget to remove this before submitting

RTSandberg:
I appreciate the feedback; I knew there were some conversations to have before this would be submitted.  I'll change to [WIP] while the other changes are made and we talk about the PICMI standard

ax3l:
maybe good to touch base tomorrow with @dpgrote and @roelof-groenewald since #3475 was merged now

ax3l:
ping @dpgrote @RTSandberg is something in this PR that was not yet merged in @roelof-groenewald 's?

dpgrote:
@ax3l I don't see anything here that isn't already included. This PR can be closed.

EZoni:
Closing this based on https://github.com/ECP-WarpX/WarpX/pull/3179#issuecomment-1440832760. Anyone please feel free to reopen if necessary.

PR #3180, opened 2022-06-14
CMake: Generalize `CMAKE_INSTALL_CMAKEDIR`

ax3l:
Generalize the control of the global `CMAKE_INSTALL_CMAKEDIR` variable. This way, we can control the variable in superbuilds that include ABLASTR, e.g., in ImpactX.


cmake/WarpXFunctions.cmake
```diff
@@ -345,10 +351,7 @@ function(warpx_print_summary)
     message("        bin: ${CMAKE_INSTALL_BINDIR}")
     message("        lib: ${CMAKE_INSTALL_LIBDIR}")
     message("    include: ${CMAKE_INSTALL_INCLUDEDIR}")
-    message("      cmake: ${CMAKE_INSTALL_CMAKEDIR}")
-    if(WarpX_HAVE_PYTHON)
-        message("     python: ${CMAKE_INSTALL_PYTHONDIR}")
-    endif()
```
EZoni, line 351: Just double checking: these lines with `CMAKE_INSTALL_PYTHONDIR` (and not `CMAKE_INSTALL_CMAKEDIR`) were also removed on purpose?

ax3l: yes, as per 2nd commit they were unused

EZoni (approved):

PR #3183, opened 2022-06-15
Do not query redundant `do_back_transformed_particles`

NeilZaim:
This input parameter (which is not in the doc) is redundant with the `write_species` input parameter and is in fact never used since it is overwritten there:

https://github.com/ECP-WarpX/WarpX/blob/95e92ff51599b685af4d09f867af4c49a42f2a72/Source/Diagnostics/BTDiagnostics.cpp#L102-L105

I've also slightly modified line 102 above so that `m_do_back_transformed_particles` is false when `write_species` is false. I think this could happen if a user specifies a list of species to plot for this diag as well as `diag.write_species = 0`. In that case `m_do_back_transformed_particles` would be set to true in the particle containers (which I guess causes some memory allocation), which I think should not be the case because the diagnostic member variable `m_output_species_names` would be cleared soon after:

https://github.com/ECP-WarpX/WarpX/blob/95e92ff51599b685af4d09f867af4c49a42f2a72/Source/Diagnostics/Diagnostics.cpp#L313-L322

### DIFF
--- a/Source/Diagnostics/BTDiagnostics.cpp
+++ b/Source/Diagnostics/BTDiagnostics.cpp
@@ -99,7 +99,7 @@ void BTDiagnostics::DerivedInitData ()
     if (m_output_species_names.size() == 0 and write_species == 1)
         m_output_species_names = mpc.GetSpeciesNames();
 
-    if (m_output_species_names.size() > 0) {
+    if (m_output_species_names.size() > 0 and write_species == 1) {
         m_do_back_transformed_particles = true;
     } else {
         m_do_back_transformed_particles = false;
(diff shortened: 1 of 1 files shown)


RevathiJambunathan (approved):
Thank you @NeilZaim for this PR!! 
Looks good to me. 

Is there a reason to remove` AMREX_ALWAYS_ASSERT(m_do_back_transformed_fields or m_do_back_transformed_particles);`?

EZoni:
> Is there a reason to remove `AMREX_ALWAYS_ASSERT(m_do_back_transformed_fields or m_do_back_transformed_particles);`?

@NeilZaim Do you have an answer to @RevathiJambunathan's question above? Once that's clarified, we can merge this PR.

NeilZaim:
Hello @RevathiJambunathan @EZoni .
Sorry for not replying earlier. This assert can be removed because now that `m_do_back_transformed_particles` is not set here, it is exactly the same as the assert below.

EZoni (approved):

PR #3184, opened 2022-06-15
Fix value of particle container m_do_back_transformed_particles when …

NeilZaim:
…there are multiple BT diagnostics

The value of the member variable `m_do_back_transformed_particles` in the particle containers is directly set from that of the same variable in the `BTDiagnostic` class. However, I think that in principle there can be multiple BT diagnostics but there will only be one e.g. `MultiParticleContainer`.
Looking at the code, it looks like the value of `m_do_back_transformed_particles` in the particle containers will be that of the last BT diagnostic that is initialized. This is potentially incorrect (for instance if the last BTdiagnostic has particles turned off but the one before has particles turned on), and I would think that the correct way would be to set `m_do_back_transformed_particles` to `true` in the particle containers if it is true for at least one of the BT diags. Am I correct?


RevathiJambunathan (approved):

RevathiJambunathan (approved):
Thank you @NeilZaim for this PR

Issue #3188, opened 2022-06-16
Refactor Single-Precision Comms


n01r:
To optimize communication across MPI ranks when running WarpX with the PSATD solver, data is recently cast to single precision-before float before copying which saves bandwidth. However, a temporary multifab is first created which causes additional overhead.  
Since AMReX [#2708](https://github.com/AMReX-Codes/amrex/pull/2708/files), `FillBoundary` allows for specialization with the buffer type.  Depending on the global variable `WarpX::do_single_precision_comms`, the code should now call something like `FillBoundary<float>` which concerns mostly the file `Source/ablastr/utils/Communication.cpp` (see also #3167). 
(Thank you, @WeiqunZhang and @atmyers for clarification!)

Replacing the first is straightforward
<details>
<summary><b>FillBoundary</b> usage</summary>

```C++
void FillBoundary (amrex::MultiFab &mf, bool do_single_precision_comms, const amrex::Periodicity &period)
{
    BL_PROFILE("ablastr::utils::communication::FillBoundary");

    if (do_single_precision_comms)
    {
        amrex::FabArray<amrex::BaseFab<comm_float_type> > mf_tmp(mf.boxArray(),
                                                                 mf.DistributionMap(),
                                                                 mf.nComp(),
                                                                 mf.nGrowVect());

        mixedCopy(mf_tmp, mf, 0, 0, mf.nComp(), mf.nGrowVect());

        mf_tmp.FillBoundary(period);

        mixedCopy(mf, mf_tmp, 0, 0, mf.nComp(), mf.nGrowVect());
    }
    else
    {
        mf.FillBoundary(period);
    }
}
```

</details>

Replacing the second is not immediately as easy since `FillBoundaryAndSync` does not support template specialization with the buffer type, yet. Changes will affect AMReX as well.

<details>
<summary><b>FillBoundaryAndSync</b> usage</summary>

```C++
void FillBoundary(amrex::MultiFab &mf,
                  amrex::IntVect ng,
                  bool do_single_precision_comms,
                  const amrex::Periodicity &period,
                  const bool nodal_sync)
{
    BL_PROFILE("ablastr::utils::communication::FillBoundary");

    if (do_single_precision_comms)
    {
        amrex::FabArray<amrex::BaseFab<comm_float_type> > mf_tmp(mf.boxArray(),
                                                            mf.DistributionMap(),
                                                            mf.nComp(),
                                                            mf.nGrowVect());

[...]

PR #3190, opened 2022-06-17
Remove temporary copy of MultiFab

n01r:
Checks boxes in #3188.

Use template specification of `amrex::FillBoundary` directly with
`comm_float_type` in case of single-precision communication.

Leave out `FillBoundaryAndSync` for now because it requires AMReX changes.

Should this wait for AMReX changes of the other Comm utils like `FillBoundaryAndSync`, `ParallelCopy`, `SumBoundary` etc.?


ax3l:
I am ok to start with this and do more as a follow-up.

ax3l (approved):
This is technically correct. Please feel free to follow up with the performance result (see linked issue).

PR #3196, opened 2022-06-23
CMake: 3.20+

ax3l:
We already use this version in ImpactX. The main reason is a severe simplification of CMake CUDA search logic that we can rely on.

X-Ref.:
https://github.com/AMReX-Codes/amrex/blob/2d931f63cb4d611d0d23d694726889647f8a482d/Tools/CMake/AMReXParallelBackends.cmake#L55


ax3l:
Likely needs https://github.com/AMReX-Codes/amrex/pull/2849

RemiLehe:
@ax3l Would you have time to rebase this PR?

ax3l:
Rebased now :)

EZoni (approved):
Thank you, @ax3l! Looks good to me.

PR #2419, opened 2021-10-13
Added `none` as an option for fields_to_plot

dpgrote:
This option allows turning off all field diagnostics, if for example the user only wants particle output.

This change was causing problems in AMReX, producing divide by zero errors or failed assertions. ~~It is likely that this can't be done (at least for plot files) without changes in AMReX.~~ The problems were avoided by skipping the call to the AMReX routine that would write out the fields, and skipping the creation of the MultiFab that would contain the fields.

PR #3198, opened 2022-06-23
Enable openPMD diagnostics with only particles (no fields)

RemiLehe:
This PR allows the user to only output particles (no fields), by setting:
```
<diag_name>.fields_to_plot = none
```
in the input script.

This currently only works with the openPMD format since, by construction, plotfiles require at least one field to be output.


dpgrote:
I already have PR #2419 submitted and it should be ready for merging now (once the CI tests finish). My PR works with both plot files and OpenPMD.

RemiLehe:
@dpgrote Great! Sorry that I overlooked #2419. I just merged it now.

PR #3202, opened 2022-06-24
[WIP] Move moving window

RevathiJambunathan:
Exploring the consequences of moving MoveWindow to the start of the PIC loop
This is a potential fix so that failing CI tests in PR #3056 pass.

### DIFF
--- a/Source/Diagnostics/BTDiagnostics.H
+++ b/Source/Diagnostics/BTDiagnostics.H
@@ -149,9 +149,6 @@ private:
 
     /** Vector of lab-frame time corresponding to each snapshot */
     amrex::Vector<amrex::Real> m_t_lab;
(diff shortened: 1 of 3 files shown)


RemiLehe:
Thanks a lot for looking into this and opening this PR!

It looks like most tests only require a reset of the checksums.
However, a few tests seem more tricky:
- `Python_PlasmaAcceleration` crashed, apparently in the `GatherAndPush` function
- `subcyclingMR` crashed, again in `GatherAndPush`
- `RigidInjection_BTD` crashed in `FilterComputePackFlush`, but this is maybe related to other issues addressed in other PRs?

RevathiJambunathan:
as discussed offline, we found other solutions and actually moving the MoveWindow to the start of the loop did not make a difference for all BTD tests. So I will close this PR!

PR #3207, opened 2022-06-28
Remove fields from PICMI ParticleDiagnostic

RemiLehe:
- [ ] `ParticleDiagnostic`
- [ ] `LabFrameParticleDiagnostic`
- [ ] `LabFrameFieldDiagnostic`

### DIFF
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -1891,6 +1891,7 @@ class WarpXDiagnosticBase(object):
             self.diagnostic = pywarpx.Diagnostics.Diagnostic(
                 self.name, _species_dict={}
             )
+            self.diagnostic.fields_to_plot = 'none'
             bucket._diagnostics_dict[self.name] = self.diagnostic
 
     def set_write_dir(self):


Python/pywarpx/picmi.py
```diff
@@ -1335,6 +1335,7 @@ def initialize_inputs(self):
             pywarpx.diagnostics._diagnostics_dict[self.name] = self.diagnostic
 
         self.diagnostic.diag_type = 'Full'
+        self.diagnostic.fields_to_plot = 'none'
```
dpgrote: Thanks for posting this PR. This new line should go after line 1334 that creates a new `Diagnostic` instance. If a previously created `Diagnostic` instance is fetched (in line 1332) it may have be made by a `FieldDiagnostic` and so already have `fields_to_plot` defined.

dpgrote (approved):
Thanks! Looks good!

RemiLehe:
@dpgrote It seems that the CI tests fail with the following error: 
```
yt.utilities.exceptions.YTUnidentifiedDataType: Could not determine input format from `'Python_plasma_lens_plt000084'`
```
This is a Python test, with a single diagnostic - which is a `ParticleDiagnostic` (see the [input script](https://github.com/ECP-WarpX/WarpX/blob/development/Examples/Tests/plasma_lens/PICMI_inputs_3d.py)). Here is the plotfile that `yt` was not able to read:
[Python_plasma_lens_plt000084.zip](https://github.com/ECP-WarpX/WarpX/files/9046511/Python_plasma_lens_plt000084.zip)
Do you understand why `yt` was confused?

dpgrote:
I think that Yt is looking for the Header file, but that file wasn't written out since there are no fields (the AMReX routine that would write out that file was not called). This will be a problem for the no fields option since it means that the plot files cannot be processed with Yt. This is something that was not noticed in PR #2419. For the failing test, a simple fix is to not have Yt open the plot file. I'll look at this a bit and see if I can find a way around this issue.

dpgrote:
This should now work as PR #3219 has been merged. Note that the CI benchmark file will need to be updated to remove the references to the fields.

Python/pywarpx/picmi.py
```diff
@@ -1891,6 +1891,7 @@ def add_diagnostic(self):
             self.diagnostic = pywarpx.Diagnostics.Diagnostic(
                 self.name, _species_dict={}
             )
+            self.diagnostic.fields_to_plot = 'none'
```
dpgrote, line 1894: I saw that this PR was revived. I don't think this is the right place for this line since `WarpXDiagnosticBase` is used by all of the diagnostic classes. This setting would mess up `Checkpoint` since `fields_to_plot` shouldn't be set for it. Also, `fields_to_plot` is not used by `ReducedDiagnostic`. The better place for this statement is in `ParticleDiagnostic` after the call to `self.add_diagnostic()`, adding it this way
```
        if 'fields_to_plot' not in self.diagnostic.argvattrs:
            self.diagnostic.fields_to_plot = 'none'
```

EZoni:
@RemiLehe @dpgrote 

I believe this was taken care of in #4317 and #5109. 

Do you agree? If so, I think we should close this PR.

In the current `development` branch, I see the following:

- For `ParticleDiagnostic`:
https://github.com/BLAST-WarpX/warpx/blob/ca0310a13f673ce8c7d9fa9dea2ecdc0412dff3a/Python/pywarpx/picmi.py#L4406-L4407

- For `LabFrameParticleDiagnostic`:
https://github.com/BLAST-WarpX/warpx/blob/ca0310a13f673ce8c7d9fa9dea2ecdc0412dff3a/Python/pywarpx/picmi.py#L4723-L4724

- For `ParticleBoundaryScrapingDiagnostic`:
https://github.com/BLAST-WarpX/warpx/blob/ca0310a13f673ce8c7d9fa9dea2ecdc0412dff3a/Python/pywarpx/picmi.py#L5223-L5224

dpgrote:
> @RemiLehe @dpgrote
> 
> I believe this was taken care of in #4317 and #5109.
> 
> Do you agree? If so, I think we should close this PR.

I agree - this can be closed.

PR #3208, opened 2022-06-29
[WIP] Vay Deposition: Filter D, Sum Guards of J

EZoni:
With the Vay deposition, apply filtering on D, after it is deposited, and sum the guard cells of J, after it is computed from D. Before this PR, both filtering and sum of the guard cells were applied to J at the end (namely, right before pushing the electromagnetic fields in spectral space).

In order to accomplish this, the function `SyncCurrent` takes now a number of arguments, including the MultiFabs on which it operates as well as two flags to select whether filtering is applied and whether the sum of the guard cells is applied. Please let me know if you think that this interface could be confusing and/or have suggestions on how to improve it.

New code:
figure_Jz_new

Old code (spurious signal, light-blue dots, close to the boundaries at Z = +/- 40 m):
figure_Jz_old


EZoni:
Discussed with @RemiLehe offline: we need to understand how to handle the MR case (now the sum among levels is probably performed twice, once on D and once on J, which is probably not correct) and try to separate the filtering from the sum over guard cells into two different functions rather than using additional flags.

EZoni:
> Discussed with @RemiLehe offline: we need to understand how to handle the MR case (now the sum among levels is probably performed twice, once on D and once on J, which is probably not correct) and try to separate the filtering from the sum over guard cells into two different functions rather than using additional flags.

Following up on this, the goal would be to use the new functions implemented in #3222.

PR #3209, opened 2022-06-29
Use amrex::RandomNormal in Maxwell-Boltzmann momentum injection

NeilZaim:
I have been experiencing crashes with Maxwell-Boltzmann momentum injection + Single precision + GPU.

After debugging, I found that the crashes come from random numbers being exactly zero in the Box-Mueller transform used in the initialization of the particles momentum (`std::log(x1)` leads to `infs` then `nans` if `x1` is exactly 0). Using the amrex method to obtain a number from a normal distribution seems to fix the issue.

On a side note, should the fact that I'm reliably getting exactly 0 from random number generation with GPU+Single precision concern us? Or maybe it's not that unlikely?


Source/Initialization/InjectorMomentum.H
```diff
@@ -153,34 +153,30 @@ struct InjectorMomentumBoltzmann
...
         // high for Boltzmann or less than zero
         amrex::Real const theta = temperature(x,y,z);
-        if (theta < 0) {
+        if (theta < 0._rt) {
             amrex::Abort("Negative temperature parameter theta encountered, which is not allowed");
         }
         // Calculate local velocity and abort if |beta|>=1
         amrex::Real const beta = velocity(x,y,z);
-        if (beta <= -1 || beta >= 1) {
+        if (beta <= -1_rt || beta >= 1_rt) {
```
EZoni: Just adding the `.` for consistency (but I think it's fully equivalent):
```suggestion
        if (beta <= -1._rt || beta >= 1._rt) {
```

Source/Initialization/InjectorMomentum.H
```diff
@@ -153,34 +153,30 @@ struct InjectorMomentumBoltzmann
...
             amrex::Abort("beta = v/c magnitude greater than or equal to 1");
         }
         // Calculate the value of vave from the local temperature
-        amrex::Real const vave = std::sqrt(amrex::Real(2) * theta);
+        amrex::Real const vave = std::sqrt(theta);
```
EZoni, line 170: Just double checking: did you remove the `amrex::Real(2)` inside `std::sqrt` in the definition of `vave` because it is included in `amrex::RandomNormal` below?

EZoni (commented):
Thank you for this fix, @NeilZaim! Looks good to me, I left only a couple of minor comments.

Source/Initialization/InjectorMomentum.H
NeilZaim: Yes, exactly.

There is usually a factor sqrt(2) in the Box-Mueller transform in order to get a standard (i.e. std of 1) normal distribution. This factor was not present in the old version of the code so the std was `vave`/sqrt(2). To get an equivalent result, we can either remove the factor sqrt(2) in the definition of `vave` (as I did here) or pass `vave`/sqrt(2) instead of `vave` as the second argument to the `amrex::RandomNormal` calls.

EZoni (approved):

PR #3217, opened 2022-07-07
Parse the quantity `zmax_plasma_to_compute_max_step` in  PICMI

RemiLehe:
This allows PICMI script to use the quantity `zmax_plasma_to_compute_max_step`

### DIFF
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -1066,8 +1066,7 @@ class Simulation(picmistandard.PICMI_Simulation):
             pywarpx.warpx.gamma_boost = self.gamma_boost
             pywarpx.warpx.boost_direction = 'z'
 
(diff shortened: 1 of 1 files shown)


Python/pywarpx/picmi.py
```diff
@@ -1065,6 +1066,9 @@ def initialize_inputs(self):
             pywarpx.warpx.gamma_boost = self.gamma_boost
             pywarpx.warpx.boost_direction = 'z'
 
+        if self.zmax_plasma_to_compute_max_step is not None:
```
dpgrote: This check for None is not needed. Anything set to None is ignored when passing the parameters into WarpX.

dpgrote (approved):
This looks good to me!

EZoni (approved):

PR #3218, opened 2022-07-07
Species variables in diagnostics: deal with runtime components

NeilZaim:
Currently we cannot add extra runtime components to `diag.species.variables`, so if we want them in the diags we need to output all real components. This PR implements this feature.

I've also removed the `ParticleStringNames` namespaces which is no longer used in the code after this PR (and was redundant with the `PIdx` struct).


RemiLehe (approved):

RemiLehe (commented):
Thanks! This also makes the code cleaner.
It looks like some of the tests are failing ; could you have a look at it?

NeilZaim:
Thanks @RemiLehe, all the tests should have been fixed

PR #3219, opened 2022-07-07
Cleanup how fields_to_plot = none is handled

dpgrote:
There has been a change in AMReX that allows MultiFabs to be created with ncomp = 0. (AMReX PR #2873) This allows the handling the `fields_to_plot = none` to be cleaned up, removing the `if` checks around the allocation and call to `WriteToFile`.

This also fixes another issue that arose in PR #2419, that the plot files were not readable by Yt (because the Header file was not written out). With this PR, that is also fixed.

Once the version of AMReX that WarpX depends on is updated, I remove the WIP and this can be merged.

PR #3220, opened 2022-07-07
[WIP]Fix fields_to_plot = none

dpgrote:
With the previous change to allow `fields_to_plot = none`, PR #2419, the resulting plot files could not be read in by Yt since there was no `Header` file, which is required by Yt. This change fixes that with a bit of a hack.

The previous problem was that the MultiFab holding the data to be written out was being allocated but with the number of components zero. This fails, causing a seg fault deep inside of AMReX. The hack is to create the MultiFab with ncomp = 1, but then hack it setting its attribute n_comp = 0. This works locally, and hopefully will pass the CI tests.

With this change, the Header file is written out and the plot file can be read with Yt.


dpgrote:
This PR is not needed since PR #3219 will now work after a fix to AMReX.

PR #3221, opened 2022-07-08
Improve Docs on max_grid_size/blocking_factor for RZ

EZoni:
Just adding a note to clarify that the parameters corresponding to the longitudinal direction in RZ geometry are `amr.max_grid_size_y` and `amr.blocking_factor_y` (and not the `_z` counterparts).

See the code inside the function `CheckGriddingForRZSpectral` for reference:
https://github.com/ECP-WarpX/WarpX/blob/7b165605f1b5e308d7da41db9941ac75793b05aa/Source/Utils/WarpXUtil.cpp#L467-L495


RTSandberg (approved):
I think this is a helpful note for RZ users

dpgrote (approved):
This looks good, thanks!

PR #3222, opened 2022-07-08
`SyncCurrent`: Split Filter and Sum over Guard Cells

EZoni:
To-do:
- [x] Split `ApplyFilterandSumBoundaryJ` into new functions `ApplyFilterJ` and `SumBoundaryJ`
- [x] Use new function(s) `ApplyFilterJ` in MR part of `AddCurrentFromFineLevelandSumBoundary`:
  - [x] fix the periodicity used to call `WarpXSumGuardCells`
  - [x] fix bug(s) (CI tests with MR currently fail)
  - [x] use new functions in remaining code blocks (probably not covered by CI testing)
- [x] Merge #3252 first, to make sure that there won't be checksum changes for those CI tests here


EZoni:
CI tests currently failing:
- `VayDeposition2D` (checksums)
- `bilinear_filter` (error v. tolerance): fixed by 9127c44
- `VayDeposition3D` (checksums)
- `LaserAcceleration_single_precision_comms` (checksums)

Source/Parallelization/WarpXComm.cpp
```diff
@@ -1038,24 +1064,39 @@ void WarpX::AddCurrentFromFineLevelandSumBoundary (
...
+                    *current_buf[lev+1][idim], *J_cp[lev+1][idim],
+                    0, 0, current_buf[lev+1][idim]->nComp(), ng);
+
+                ablastr::utils::communication::ParallelAdd(
+                    mf, *current_buf[lev+1][idim], 0, 0,
+                    current_buf[lev+1][idim]->nComp(),
+                    ng, amrex::IntVect(0),
+                    do_single_precision_comms, period);
+
+                SumBoundaryJ(J_cp, lev+1, period);
```
EZoni: @WeiqunZhang

When you have some time, would you be able to tell why calling the new functions `ApplyFilterJ` and `SumBoundaryJ` as done here is not equivalent to the code commented right above? Details could be discussed on Slack or offline, of course. Thank you.

P.S. That the two codes are not equivalent (and, in fact, that the new code is wrong) is inferred by the fact that the MR tests using filtering and current buffers fail (e.g., the test `Langmuir_multi_2d_MR_psatd`).

Source/Parallelization/WarpXComm.cpp
```diff
@@ -1038,24 +1064,39 @@ void WarpX::AddCurrentFromFineLevelandSumBoundary (
...
+                //bilinear_filter.ApplyStencil(jfb, *current_buf[lev+1][idim], lev+1);
+
+                //MultiFab::Add(jfb, jfc, 0, 0, current_buf[lev+1][idim]->nComp(), ng);
+                //ablastr::utils::communication::ParallelAdd(mf, jfb, 0, 0, current_buf[lev + 1][idim]->nComp(),
+                //                                           ng, IntVect::TheZeroVector(), WarpX::do_single_precision_comms, period);
+
+                //WarpXSumGuardCells(*J_cp[lev+1][idim], jfc, period, ng_depos_J, 0, J_cp[lev+1][idim]->nComp());
+
+                ApplyFilterJ(J_cp, lev+1);
+                ApplyFilterJ(current_buf, lev+1);
```
WeiqunZhang: ApplyFilterJ has a loop over dimensions.  But we are already inside a loop over dimensions.

Source/Parallelization/WarpXComm.cpp
WeiqunZhang: SumBoundaryJ has a loop over dimensions too.

Source/Parallelization/WarpXComm.cpp
EZoni: Thank you for spotting this, definitely a bug. Should be fixed in 16f8aa7.

Source/Parallelization/WarpXComm.cpp
EZoni: Thank you for spotting this, definitely a bug too. Should be fixed in 16f8aa7.

Source/WarpX.cpp
```diff
@@ -1781,6 +1787,11 @@ WarpX::AllocLevelMFs (int lev, const BoxArray& ba, const DistributionMapping& dm
     Efield_fp[lev][1] = std::make_unique<MultiFab>(amrex::convert(ba,Ey_nodal_flag),dm,ncomps,ngEB,tag("Efield_fp[y]"));
     Efield_fp[lev][2] = std::make_unique<MultiFab>(amrex::convert(ba,Ez_nodal_flag),dm,ncomps,ngEB,tag("Efield_fp[z]"));
 
+    if (use_filter)
+    {
+        ngJ += bilinear_filter.stencil_length_each_dir - amrex::IntVect(1);
+    }
```
EZoni: Increasing the guard cells allocated for the current MultiFabs seems to be necessary in order for the CI test `bilinear_filter` to pass (otherwise the filtered current does not match the analytical and theoretical prediction).

Source/Initialization/WarpXInitData.cpp
```diff
@@ -392,10 +392,6 @@ WarpX::InitData ()
         WarpX::InitNCICorrector();
     }
 
-    if (WarpX::use_filter) {
```
RemiLehe, line 395: Could you explain why this call had to be moved to `WarpX::AllocLevelData` ?

Source/Parallelization/WarpXComm.cpp
```diff
@@ -953,43 +953,81 @@ void WarpX::RestrictCurrentFromFineToCoarsePatch (
...
+    amrex::MultiFab& J = *current[lev][idim];
+
+    const int ncomp = J.nComp();
+    const amrex::IntVect ngrow = J.nGrowVect();
+    amrex::MultiFab Jf(J.boxArray(), J.DistributionMap(), ncomp, ngrow);
+    bilinear_filter.ApplyStencil(Jf, J, lev);
+
+    const int srccomp = 0;
+    const int dstcomp = 0;
+    const int numcomp = J.nComp();
```
RemiLehe: Isn't `numcomp` the same as `ncomp` (which was declared a few lines above)?

Source/Parallelization/WarpXComm.cpp
```diff
@@ -953,43 +953,81 @@ void WarpX::RestrictCurrentFromFineToCoarsePatch (
...
+    const amrex::IntVect nghost = J.nGrowVect();
```
RemiLehe: Similarly, this seems to be the same as `ngrow`, which was declared a few lines above.

RemiLehe (changes requested):
Looks good to me. 
I just added a minor comment on cleaning up some variables.

Source/Initialization/WarpXInitData.cpp
EZoni: Yes, I think this is because otherwise these lines added in WarpX.cpp
```cpp
    if (use_filter)
    {
        ngJ += bilinear_filter.stencil_length_each_dir - amrex::IntVect(1);
    }
```
would be ineffective: `bilinear_filter.stencil_length_each_dir` would not be initialized at this point, if `InitFilter` were called from within `InitData` as before.

Source/Parallelization/WarpXComm.cpp
EZoni: Yes, I can change these variable names. I think I was using the same naming conventions as the corresponding AMReX functions to help myself try out various solutions, when I was implementing this. Will do the same for the comment below. See https://github.com/ECP-WarpX/WarpX/pull/3222/commits/9edc3866c83257e41b7ecbc67d228e35d4d7680d.

EZoni:
Setting this as "[WIP]" while investigating with @oshapoval a potential bug that shows up only in DEBUG mode.

EZoni:
@oshapoval @RemiLehe 
The bug briefly mentioned in the last comment above is due to the fact that the real-space box `realspace_ba` used to construct the object `spectral_solver_fp` is grown by `ngEB` guard cells, which are less than `ngJ` when `ngJ` is increased by the length of the filter's stencil:
https://github.com/ECP-WarpX/WarpX/blob/b13e7f53959656a8d47adc63f20d9ed055288be0/Source/WarpX.cpp#L1944-L1953

I guess increasing `ngJ` when we allocate the current MultiFabs, as it is done in this PR, does not go well together with the fact that we would like all MultiFabs to have the same number of guard cells, the one ultimately used to construct the real-space box that is passed to the constructor of the spectral solver. On the other side, in the case of more than one filter pass, we probably do not want to allocate (many) more guard cells for all MultiFabs besides J.

**Update**
The solution we chose is to allocate more guard cells for all MultiFabs in the case of filtering: abbd64b.

EZoni:
@RemiLehe
The checksum changes, all related to CI tests that use mesh refinement, are due to the fact that, by modifying `ng_alloc_J` directly in the guard cells manager (so that all other guard cells parameters related to `ng_alloc_J` are also modified accordingly, in order to fix the PSATD bug reported in the comment above), `n_current_deposition_buffer` is also increased by a number of cells equal to the length of the stencil of the bilinear filter:
https://github.com/ECP-WarpX/WarpX/blob/4876446dff60411cbdd4892a044643f5fcc9dd00/Source/WarpX.cpp#L1660-L1662
I think it is correct that the number of cells for the current deposition buffer are modified similarly to the guard cells of the current MultiFabs (thus resulting in some checksum changes, as the current status of this PR), but let me know if you have a different take on this.

RemiLehe (approved):

PR #3225, opened 2022-07-11
Move handling of do_not_deposit to physics callers

PhilMiller:
This is in support of allowing charge density output for non-interacting (e.g. tracer) particles

Replaces #3119.

### DIFF
--- a/Source/FieldSolver/ElectrostaticSolver.cpp
+++ b/Source/FieldSolver/ElectrostaticSolver.cpp
@@ -221,10 +221,11 @@ WarpX::AddSpaceChargeFieldLabFrame ()
     bool const do_rz_volume_scaling = false;
     for (int ispecies=0; ispecies<mypc->nSpecies(); ispecies++){
         WarpXParticleContainer& species = mypc->GetParticleContainer(ispecies);
-        if (species.do_not_deposit) continue;
-        species.DepositCharge(
-            rho_fp, local, reset, do_rz_volume_scaling, interpolate_across_levels
-        );
+        if (!species.do_not_deposit) {
+            species.DepositCharge( rho_fp,
+                                   local, reset, do_rz_volume_scaling, interpolate_across_levels
+                                  );
+        }
     }
 #ifdef WARPX_DIM_RZ
     for (int lev = 0; lev <= max_level; lev++) {
--- a/Source/FieldSolver/ElectrostaticSolver.cpp
+++ b/Source/FieldSolver/ElectrostaticSolver.cpp
@@ -78,6 +78,7 @@ WarpX::ComputeSpaceChargeField (bool const reset_fields)
         // due to simulation boundary potentials
         for (int ispecies=0; ispecies<mypc->nSpecies(); ispecies++){
             WarpXParticleContainer& species = mypc->GetParticleContainer(ispecies);
+            if (species.do_not_deposit) continue;
             if (species.initialize_self_fields ||
                 (do_electrostatic == ElectrostaticSolverAlgo::Relativistic)) {
                 AddSpaceChargeField(species);
@@ -218,6 +219,7 @@ WarpX::AddSpaceChargeFieldLabFrame ()
     bool const do_rz_volume_scaling = false;
     for (int ispecies=0; ispecies<mypc->nSpecies(); ispecies++){
         WarpXParticleContainer& species = mypc->GetParticleContainer(ispecies);
+        if (species.do_not_deposit) continue;
         species.DepositCharge(
             rho_fp, local, reset, do_rz_volume_scaling, interpolate_across_levels
         );
--- a/Source/FieldSolver/ElectrostaticSolver.cpp
+++ b/Source/FieldSolver/ElectrostaticSolver.cpp
@@ -78,7 +78,6 @@ WarpX::ComputeSpaceChargeField (bool const reset_fields)
         // due to simulation boundary potentials
         for (int ispecies=0; ispecies<mypc->nSpecies(); ispecies++){
             WarpXParticleContainer& species = mypc->GetParticleContainer(ispecies);
-            if (species.do_not_deposit) continue;
             if (species.initialize_self_fields ||
                 (do_electrostatic == ElectrostaticSolverAlgo::Relativistic)) {
                 AddSpaceChargeField(species);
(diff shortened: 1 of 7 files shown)


PhilMiller:
Marking 'ready for review' to run through CI

ax3l:
@PhilMiller can you pls take a look why CI fails?

PhilMiller:
All of the test failures are in RZ builds with PSATD enabled, with the profiler stack
```
===== TinyProfilers ======
main()
WarpX::InitData()
Diagnostics::FilterComputePackFlush()
```

Source/Particles/MultiParticleContainer.cpp
```diff
@@ -486,12 +486,19 @@ MultiParticleContainer::GetZeroChargeDensity (const int lev)
...
-                                               dmap,WarpX::ncomps,ng_rho);
+    bool is_PSATD_RZ = false;
+#ifdef WARPX_DIM_RZ
+    if (WarpX::maxwell_solver_id == MaxwellSolverAlgo::PSATD)
+        is_PSATD_RZ = true;
+#endif
+    if( !is_PSATD_RZ )
+        nba.surroundingNodes();
+
+    auto zero_rho = std::make_unique<MultiFab>(nba, dmap, WarpX::ncomps, ng_rho);
```
PhilMiller, line 501: The inconsistency between the box arrays of the zero rho and the filled rho in the RZ PSATD case was causing the test failures. I think making the zero array behave the same as the regular one is the right thing to do, but that needs someone more knowledgeable to confirm.

Source/FieldSolver/ElectrostaticSolver.cpp
```diff
@@ -165,15 +165,18 @@ WarpX::AddSpaceChargeField (WarpXParticleContainer& pc)
         BoxArray nba = boxArray(lev);
         nba.surroundingNodes();
         rho[lev] = std::make_unique<MultiFab>(nba, DistributionMap(lev), 1, ng);
+        rho[lev]->setVal(0.);
```
ax3l, line 168: @RemiLehe @dpgrote did we leave this non-zero intentionally (so we can add to `rho` multiple times)?

ax3l: oh wait, looks like it's a temporary here.

Source/Particles/MultiParticleContainer.cpp
ax3l: ccing @RemiLehe and @EZoni for PSATD question.

Source/FieldSolver/ElectrostaticSolver.cpp
PhilMiller: Yeah, it's a temporary. I added this call because I can no longer rely on a call below with `reset=true`

Source/Particles/MultiParticleContainer.cpp
PhilMiller: There's a broader or conceptually deeper inconsistency between RZ PSATD and everything else here, in that the diagnostic output will be slightly different depending on the solver used. At first, I was concerned that I'd have to fix *that* to get this working

EZoni: Hi, sorry I missed this. I can let @dpgrote and @RemiLehe comment as well, but I think rho should be always fully nodal (regardless of whether the simulation is nodal or staggered, which is controlled by `warpx.do_nodal`), except for RZ PSATD where all fields are fully cell-centered, including rho.

Now, what we do for the diagnostics is kind of a choice, so I guess it's debatable. But anyways I think I agree with the fact that the "zero" rho array should behave the same as the regular rho array.

Two other small things I noticed are:
- I see that we use `WarpX::ncomps`, which should be fine, but I think extracting the number of components directly from the rho MultiFab (by calling the method `nComp`), whenever possible, would be more robust.
- I see that `ng_rho` is set to `warpx.get_ng_depos_rho().max()` and I'm not sure why we need to take the `max` here, instead of simply taking the `IntVect` itself. I would be in favor of the latter, unless there is a specific reason why the `max` is needed here.

I'm not sure if this answer any of the questions, but I hope it helps. @dpgrote and @RemiLehe can comment more.

dpgrote: I think that this change is the correct thing to do, to make the rho created here consistent with the one that would be created with `GetChargeDensity`. This makes sense since the will be used in the same contexts.

Note that here, there is no place to get `ncomps` from other than `Warpx::ncomps` since there is no multifab available to get the value from. This is consistent with `GetChargeDensity` which also uses `WarpX::ncomps`. Similarly, `GetChargeDensity` also uses `warpx.get_ng_depos_rho().max()`.

Source/FieldSolver/ElectrostaticSolver.cpp
```diff
@@ -218,6 +221,7 @@ WarpX::AddSpaceChargeFieldLabFrame ()
     bool const do_rz_volume_scaling = false;
     for (int ispecies=0; ispecies<mypc->nSpecies(); ispecies++){
         WarpXParticleContainer& species = mypc->GetParticleContainer(ispecies);
+        if (species.do_not_deposit) continue;
```
dpgrote: This is a picky detail, but perhaps use the same construction as above, checking for `(!species.do_not_deposit)`. This would make it easier to compare the two routines if needed in the future. Also, I find it a bit easier to follow code if blocks are wrapped in `if` statements since it can be easy to miss a `continue`. The same for below.

Source/Python/WarpXWrappers.cpp
```diff
@@ -624,6 +624,7 @@ namespace
         {
             const long np = pti.numParticles();
             auto& wp = pti.GetAttribs(PIdx::w);
+            // Do this unconditionally from myspc.do_not_deposit, to support diagnostic uses
```
dpgrote: A suggested change to the comment:

Do this unconditionally, ignoring myspc.do_not_deposit, to support diagnostic use

dpgrote (approved):
This looks good, though I have some suggestions for cleanup that are optional.

Source/FieldSolver/ElectrostaticSolver.cpp
PhilMiller: In general, I prefer to maintain shallower nesting, but given that it's just one statement, I agree with you here. Done.

Source/Particles/MultiParticleContainer.cpp
dpgrote: Thinking about this some more, here is a suggestion for a later PR perhaps to reduce duplicate code. Since almost everything done in this routine is using the `warpx` instance, a better place for this might be in the `WarpX` class, perhaps a routine that returns a ready made rho MultiFab. Then the routine `WarpXParticleContainer::GetChargeDensity` could call that WarpX function to get the rho MultiFab instead of creating one itself.

Source/Particles/LaserParticleContainer.cpp
```diff
@@ -585,7 +585,7 @@ LaserParticleContainer::Evolve (int lev,
             plane_Yp.resize(np);
             amplitude_E.resize(np);
 
-            if (rho && ! skip_deposition) {
+            if (rho && ! skip_deposition && ! do_not_deposit) {
```
ax3l, line 588: Unrelated to the PR that fixes it: Wow, this is some messy logic already in place from us.

ax3l (approved):

PR #3228, opened 2022-07-12
Docs: New OLCF Machine

ax3l:
Document usage of that new OLCF machine.

- [x] configure
- [x] compile & link
- [x] run on head node
- [x] run on interactive node
- [x] adjust/add and test HDF5
- [x] document interactive node usage

Follow-up PRs:
- [ ] document and test ADIOS2 usage
- [ ] document and test Ascent usage

### DIFF
--- a/Tools/machines/frontier-olcf/frontier_warpx.profile.example
+++ b/Tools/machines/frontier-olcf/frontier_warpx.profile.example
@@ -33,6 +33,9 @@ module load cray-python/3.9.12.1
 # fix system defaults: do not escape $ with a \ on tab completion
 shopt -s direxpand
 
+# make output group-readable by default
+umask 0027
+
 # an alias to request an interactive batch node for one hour
 #   for paralle execution, start on the batch node: srun <command>
 alias getNode="salloc -A $proj -J warpx -t 01:00:00 -p batch -N 1 --ntasks-per-node=8 --gpus-per-task=1 --gpu-bind=closest"
--- a/Tools/machines/frontier-olcf/frontier_warpx.profile.example
+++ b/Tools/machines/frontier-olcf/frontier_warpx.profile.example
@@ -35,10 +35,10 @@ shopt -s direxpand
 
 # an alias to request an interactive batch node for one hour
 #   for paralle execution, start on the batch node: srun <command>
-alias getNode="salloc -A $proj -J warpx -t 01:00:00 -p batch -N 1 -c 8 --ntasks-per-node=8"
+alias getNode="salloc -A $proj -J warpx -t 01:00:00 -p batch -N 1 --ntasks-per-node=8 --gpus-per-task=1 --gpu-bind=closest"
 # an alias to run a command on a batch node for up to 30min
 #   usage: runNode <command>
-alias runNode="srun -A $proj -J warpx -t 00:30:00 -p batch -N 1 -c 8 --ntasks-per-node=8"
+alias runNode="srun -A $proj -J warpx -t 00:30:00 -p batch -N 1 --ntasks-per-node=8 --gpus-per-task=1 --gpu-bind=closest"
 
 # GPU-aware MPI
 export MPICH_GPU_SUPPORT_ENABLED=1
--- a/Tools/machines/frontier-olcf/frontier_warpx.profile.example
+++ b/Tools/machines/frontier-olcf/frontier_warpx.profile.example
@@ -51,5 +51,5 @@ export CC=$(which cc)
 export CXX=$(which CC)
 export FC=$(which ftn)
 export CFLAGS="-I${ROCM_PATH}/include"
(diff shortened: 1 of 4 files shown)


Tools/machines/frontier-olcf/submit.sh
```diff
@@ -0,0 +1,28 @@
...
+# "Optimized 3rd Gen EPYC" CPU (with 2 hardware threads per physical core) with
+# access to 512 GB of DDR4 memory.
+# Each node also contains [4x] AMD MI250X, each with 2 Graphics Compute Dies
+# (GCDs) for a total of 8 GCDs per node. The programmer can think of the 8 GCDs
+# as 8 separate GPUs, each having 64 GB of high-bandwidth memory (HBM2E).
+
+# note (5-16-22)
+# this environment setting is currently needed on Frontier to work-around a
+# known issue with Libfabric
+export FI_MR_CACHE_MAX_COUNT=0
```
ax3l: Double check if we need this. We just need this currently on Crusher.

atmyers: My first run without this export hangs, so it looks like it still needed.

Tools/machines/frontier-olcf/frontier_warpx.profile.example
```diff
@@ -0,0 +1,55 @@
+# please set your project account
+#export proj=APH114-frontier
+
+# required dependencies
+module load cmake/3.22.2
+module load craype-accel-amd-gfx90a
+module load rocm/5.1.0
+module load cray-mpich
+#module load cce/14.0.0  # must be loaded after rocm
```
ax3l:
```
Currently Loaded Modules:
  1) craype-x86-trento                      12) DefApps/default
  2) libfabric/1.15.0.0                     13) cmake/3.22.2
  3) craype-network-ofi                     14) craype-accel-amd-gfx90a
  4) perftools-base/22.06.0                 15) rocm/5.1.0
  5) xpmem/2.4.4-2.3_2.12__gff0e1d9.shasta  16) cray-mpich/8.1.17
  6) cray-pmi/6.1.3                         17) ccache/4.5.1
  7) cce/14.0.1                             18) ninja/1.10.2
  8) craype/2.7.16                          19) nano/4.9
  9) cray-dsmml/0.2.2                       20) boost/1.79.0-cxx17
 10) cray-libsci/22.06.1.3                  21) cray-hdf5-parallel/1.12.1.1
 11) PrgEnv-cray/8.3.3                      22) cray-python/3.9.12.1
```

They have adios2/2.8.1 available if we have cce/14.0.0 loaded. There is an issue with the latest CCE 14.0.1 that causes some builds to fail so a full software stack is not available for that compiler toolchain. CCE 14.0.0, however, has the latest versions of most E4S packages.

Tools/machines/frontier-olcf/frontier_warpx.profile.example
```diff
@@ -0,0 +1,55 @@
...
+# optional: for openPMD support
+#module load adios2/2.7.1
+module load cray-hdf5-parallel/1.12.1.1
+
+# optional: for Python bindings or libEnsemble
+module load cray-python/3.9.12.1
+
+# fix system defaults: do not escape $ with a \ on tab completion
+shopt -s direxpand
+
```
ax3l, line 35:
```suggestion

# make output group-readable by default
umask 0027

```

Tools/machines/frontier-olcf/frontier_warpx.profile.example
ax3l: Due to a breaking change in read of operators, we would prefer adios/2.8.2 (or 2.7.1)

Tools/machines/frontier-olcf/submit.sh
```diff
@@ -0,0 +1,28 @@
...
+# access to 512 GB of DDR4 memory.
+# Each node also contains [4x] AMD MI250X, each with 2 Graphics Compute Dies
+# (GCDs) for a total of 8 GCDs per node. The programmer can think of the 8 GCDs
+# as 8 separate GPUs, each having 64 GB of high-bandwidth memory (HBM2E).
+
+# note (5-16-22 and 7-12-22)
+# this environment setting is currently needed on Frontier to work-around a
+# known issue with Libfabric (both in the May and June PE)
+export FI_MR_CACHE_MAX_COUNT=0
+
```
ax3l, line 26: We temporarily needed
```bash
export PMI_MMAP_SYNC_WAIT_TIME=1800
export PMI_SHARED_SECRET=1234
```
on AlpineTDSS, but this will not be universally be needed, so I would not add it here.

(Just leaving this comment for temporary docs.)

Also, we temporarily excluded nodes with
```bash
#SBATCH --exclude=frontier[00642,00319,05930,01145,01092,01171,10215]
```

atmyers (approved):

PR #3153, opened 2022-05-31
D-T fusion

RemiLehe:
This PR implements a Monte-Carlo Deuterium-Tritium fusion module, and builds upon the previously-existing proton-Boron fusion module (See https://github.com/ECP-WarpX/WarpX/pull/2290).

## Overview of the underlying model

The cross-section analytical fits for the D-T cross-section are taken from [this paper](https://iopscience.iop.org/article/10.1088/0029-5515/32/4/I07), which expresses the cross-sections as
Screen Shot 2022-07-12 at 3 03 23 AM
where `S(E)` is the "astrophysical factor", which is expressed as a ratio of two polynomials, whose coefficients have been fitted to experimental data.
The momentum of the two product particles (a helium4 nucleus and a neutron) are found by a simple conservation of energy and momenta, and by assuming isotropic emission in the center-of-mass frame.

As much as possible, the code was written in a way that could be easily extensible to other fusion reactions that lead to 2 product species (e.g. D-D fusion). As such, the mass of the product species is now passed as an argument in the `ParticleCreationFunc` (instead of using hard-coded values).

## Automated tests

The automated tests are very similar to [the proton-Boron fusion tests](https://github.com/ECP-WarpX/WarpX/pull/2450), but with Deuterium and Tritium instead of proton and Boron. For the sake of simplification, I only retained the first two tests:
- Test in the center-of-mass frame
- Test in the frame of one of the reactant, with the other reactant having high-weight macroparticles at rest, and low-weight macroparticles in motion.
We still check the conservation of energy, momentum, and number of particles, and we check that the fusion rate corresponds to the expected analytical rates.

PR #3229, opened 2022-07-12
Refactor code in fusion module to avoid machine-precision issues

RemiLehe:
The formula `p_star_sq` (which should always be a positive number) could occasionally return small negative numbers, due to machine-precision errors. (in the case where energy in the center-of-mass frame is very close to the mass energy of the two particles, i.e. when the particles have negligible kinetic energy in the center-of-mass frame). This is a problem because we later take the square root of `p_star_sq` in the code.

Therefore, this PR refactors this formula, so as to write it as the sum of 2 intrinsically positive numbers:

image

### DIFF
--- a/Source/Particles/Collision/BinaryCollision/NuclearFusion/SingleNuclearFusionEvent.H
+++ b/Source/Particles/Collision/BinaryCollision/NuclearFusion/SingleNuclearFusionEvent.H
@@ -75,7 +75,6 @@ void SingleNuclearFusionEvent (const amrex::ParticleReal& u1x, const amrex::Part
     const amrex::ParticleReal w_max = amrex::max(w1, w2);
 
     constexpr auto one_pr = amrex::ParticleReal(1.);
-    constexpr auto inv_two_pr = amrex::ParticleReal(1./2.);
     constexpr auto inv_four_pr = amrex::ParticleReal(1./4.);
     constexpr amrex::ParticleReal c_sq = PhysConst::c * PhysConst::c;
     constexpr amrex::ParticleReal inv_csq = one_pr / ( c_sq );
@@ -107,7 +106,8 @@ void SingleNuclearFusionEvent (const amrex::ParticleReal& u1x, const amrex::Part
     const amrex::ParticleReal E_star_sq = E_lab*E_lab - c_sq*p_total_sq;
 
     // Kinetic energy in the center of mass frame
-    const amrex::ParticleReal E_kin_star = std::sqrt(E_star_sq) - (m1 + m2)*c_sq;
+    const amrex::ParticleReal E_star = std::sqrt(E_star_sq);
+    const amrex::ParticleReal E_kin_star = E_star - (m1 + m2)*c_sq;
 
     // Compute fusion cross section as a function of kinetic energy in the center of mass frame
     auto fusion_cross_section = amrex::ParticleReal(0.);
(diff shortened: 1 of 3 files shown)


Source/Particles/Collision/BinaryCollision/NuclearFusion/SingleNuclearFusionEvent.H
```diff
@@ -118,14 +118,15 @@ void SingleNuclearFusionEvent (const amrex::ParticleReal& u1x, const amrex::Part
 
     // Square of the norm of the momentum of one of the particles in the center of mass frame
     // Formula obtained by inverting E^2 = p^2*c^2 + m^2*c^4 in the COM frame for each particle
-    auto constexpr pow3 = [](double const x) { return x*x*x; };
-    const amrex::ParticleReal p_star_sq =
-            E_star_sq*inv_four_pr*inv_csq - (m1_sq + m2_sq)*c_sq*inv_two_pr +
-            pow3(c_sq) * inv_four_pr * pow2(m1_sq - m2_sq) / E_star_sq;
+    // The expression below is specifically written in a form that avoids returning
+    // small negative numbers due to machine precision errors, for low-energy particles
+    const amrex::ParticleReal E_ratio = E_star/((m1 + m2)*c_sq);
```
NeilZaim, line 123: Do we have a complete guarantee that this number will never be strictly smaller than 1? E.g. if all the momenta are 0, this corresponds to the ratio between sqrt(E_rest**2) and E_rest so it's not obvious which of the two numbers will be smaller within machine precision.

NeilZaim (commented):
Thanks for this PR. The formula looks good to me (note that there is a square missing at the end of the last parenthesis in the image).

I'm not yet 100% sure that the number can never be negative (see comment below), what do you think?

Also, be aware that the old formula is also currently used here: https://github.com/ECP-WarpX/WarpX/blob/32a5989d1765853a9bdac38d93c1c863790c65cc/Source/Particles/Collision/BinaryCollision/NuclearFusion/TwoProductFusionUtil.H#L109-L114

Source/Particles/Collision/BinaryCollision/NuclearFusion/SingleNuclearFusionEvent.H
RemiLehe: Agreed. This could in principle be negative. 

However, I introduced this formula also in this PR https://github.com/ECP-WarpX/WarpX/pull/3153, where it did suppress an earlier failure due to invalid values.
(See this failure: https://dev.azure.com/ECP-WarpX/WarpX/_build/results?buildId=9168&view=logs&jobId=fc138a38-7e4f-5d28-e3f1-923e4d419807&j=fc138a38-7e4f-5d28-e3f1-923e4d419807&t=7e91da1c-aa6d-5e37-9c36-add2c6c921e8)

So my (very incomplete) understanding at this point is that the effect that you describe is maybe possible, but is maybe not as common as the previous failure.

If we encounter issues in the future with this new formula, we could certainly reconsider its expression. As an alternative, we could also intentionally enforce that `E_star` is larger than `(m1 + m2)*c_sq`, in the code, by e.g. setting `E_star = max( E_star, (m1 + m2)*c_sq )`. What do you think?

RemiLehe:
Thanks for checking the formula. I just updated it in the PR description.

RemiLehe:
Also: thanks for pointing out the other part of the code where this is used. I updated the formula accordingly.

Source/Particles/Collision/BinaryCollision/NuclearFusion/SingleNuclearFusionEvent.H
NeilZaim: That definitely makes sense to me.
I think I'm fine with both leaving it as it is (and reconsider if we encounter issues in the future) and with anticipating possible future issues by adding a line like `E_star = max( E_star, (m1 + m2)*c_sq )`.

Feel free to do what you think is best, I'll approve the PR in the meantime.

NeilZaim (approved):
Thanks a lot for this PR! That's good to me!

EZoni (approved):

PR #2540, opened 2021-11-10
Add 2D tests for proton boron fusion

NeilZaim:
This PR is based on #2450, which is based on #2330, so it should probably be reviewed once these PRs have been merged.

It adds the same automated tests as in #2450, but in 2D instead of 3D. The same analysis script is used to analyze both tests, with only minor modifications.

Putting the [WIP] tag until I have tested this on GPU.
Edit: tested on GPU so I'm removing the WIP tag.

PR #3230, opened 2022-07-12
Update 2D nuclear fusion test input file

NeilZaim:
Looks like separately merging #3090 and #2540 has broken CI.

The input file of the 2D proton-boron fusion test needs to be updated. (and possibly we will need to update some benchmarks as well, we'll see)


RemiLehe (approved):

RemiLehe:
Thanks for fixing my reckless merging! :)

PR #3234, opened 2022-07-12
Fix: Write Used Inputs Once

ax3l:
Follow-up to #3132:
Ooopsi, nearly killed a PFS again :D
Only the IO processor should write this and not every rank on the planet.

- [x] tested

First spotted by @atmyers :sparkles:


lucafedeli88 (approved):
Thanks a lot for fixing this!

PR #3235, opened 2022-07-12
Add option to deposit laser on main grid

RemiLehe:
We recently observed that depositing the laser field on the mesh refinement patch can cause issues, if the antenna is only partially contained in the mesh refinement patch.

This PR implements the option `deposit_on_main_grid` for the `LaserParticleContainer`. The option is similar to the one that already exists for `PhysicalParticleContainer`.


Source/Particles/LaserParticleContainer.cpp
```diff
@@ -578,18 +580,22 @@ LaserParticleContainer::Evolve (int lev,
...
             plane_Xp.resize(np);
             plane_Yp.resize(np);
             amplitude_E.resize(np);
 
+            // Determine which particles deposit/gather in the buffer
+            long nfine_current = np;
+            if (m_deposit_on_main_grid && lev > 0) {
+                nfine_current = 0;
+            }
+            const long np_current = (cjx) ? nfine_current : np;
```
RemiLehe:
```suggestion
            const long np_current = (has_buffer) ? nfine_current : np;
```

RemiLehe:
Note that, for the `PhysicalParticleContainer`, we also check whether particles are within the deposition buffer for each patch (i.e. whether they are within a few cells away from the edge of the patch).

This is not yet implemented for the `LaserParticleContainer`. It would require calling the partition routine for `LaserParticleContainer`, and could be done in a follow-up PR.

Docs/source/usage/parameters.rst
```diff
@@ -1265,6 +1265,11 @@ Laser initialization
     per angular mode. The laser particles are loaded into radial spokes, with
     the number of spokes given by min_particles_per_mode*(warpx.n_rz_azimuthal_modes-1).
 
+* ``<laser_name>.deposit_on_main_grid`` (`int`) optional (default `0`)
+    When using mesh refinement, whether the antenna that emits the laser
+    deposits charge/current only on the main grid (i.e. level 0), or also
+    on the higher mesh-refinement levels.
+
```
EZoni: From the code below, in Source/Particles/MultiParticleContainer.cpp,
```cpp
pp_lasers.queryarr("deposit_on_main_grid", tmp);
```
it seems like the input parameter should be `lasers.deposit_on_main_grid` (array of string, the names of the lasers to be deposited on the main grid), rather than `<laser_name>.deposit_on_main_grid` (int). Is that right or am I misinterpreting the code?

ax3l: the documented/2nd one (option per laser) is better/more common to what we do.

Source/Particles/MultiParticleContainer.cpp
```diff
@@ -347,6 +348,21 @@ MultiParticleContainer::ReadParameters ()
 
         ParmParse pp_lasers("lasers");
         pp_lasers.queryarr("names", lasers_names);
+        auto const nlasers = lasers_names.size();
+        // Get lasers to deposit on main grid
+        m_laser_deposit_on_main_grid.resize(nlasers, false);
+        std::vector<std::string> tmp;
+        pp_lasers.queryarr("deposit_on_main_grid", tmp);
```
ax3l, line 355: loop over `lasers_names` instead and query each laser for `deposit_on_main_grid`? (see above)

Docs/source/usage/parameters.rst
RemiLehe: As discussed today: I will change this in a follow-up PR for both `lasers.deposit_on_main_grid` and `particles.deposit_on_main_grid`.

ax3l (approved):

Issue #3182, opened 2022-06-15
Issue with RZ + Yee solver + momentum-conserving field gathering 


RTSandberg:
The combination of RZ + Yee + momentum-conserving field gathering appears not to work.  If I launch a laser in this configuration, I get the following
<img width="411" alt="Screen Shot 2022-06-14 at 5 13 29 PM" src="https://user-images.githubusercontent.com/10621396/173710070-ff40e80f-8792-4e31-8202-fd412b57e94f.png">
:
whereas if I turn off the momentum-conserving option, I get the expected behavior
<img width="401" alt="Screen Shot 2022-06-14 at 5 14 53 PM" src="https://user-images.githubusercontent.com/10621396/173710194-d2a302af-b514-48ed-804d-8b5689f61f37.png">

Here is the input deck I'm using
[inputs.txt](https://github.com/ECP-WarpX/WarpX/files/8904593/inputs.txt)

Perhaps we want to abort if someone tries to do this?

PR #3236, opened 2022-07-13
[WIP] abort if RZ + Yee (FDTD) + momentum-conserving

RTSandberg:
This aborts with a warning message if a user attempts an RZ simulation with the Yee solver and momentum-conserving field gathering.  These options were seen to be problematic in issue #3182.


EZoni (approved):
Thank you, @RTSandberg! Looks good to me.

RTSandberg:
Some CI tests are failing because they use the combination RZ + Yee + momentum-conserving.  Several of these are electrostatic, which suggests that the problem observed in issue #3182 is possibly with the laser launching or the electromagnetic solver itself

PR #3238, opened 2022-07-15
[WIP] New Function to Allocate MultiFabs w/ Initial Value

EZoni:
Could something like this be useful? A small WarpX function, `WarpX::AllocInitMultiFab`, that allows us to allocate a MultiFab and initialize at the same time its values using `setVal`. For development and debugging purposes, I personally find it quite useful to have this function ready to use (shorter, less error-prone with respect to uninitialized data), but please feel free to give feedback and we can easily close the PR if there isn't agreement.

### DIFF
--- a/Source/BoundaryConditions/PML.cpp
+++ b/Source/BoundaryConditions/PML.cpp
@@ -657,45 +657,58 @@ PML::PML (const int lev, const BoxArray& grid_ba, const DistributionMapping& gri
     const int ncompe = (m_dive_cleaning) ? 3 : 2;
     const int ncompb = (m_divb_cleaning) ? 3 : 2;
 
-    pml_E_fp[0] = std::make_unique<MultiFab>(amrex::convert( ba,
-        WarpX::GetInstance().getEfield_fp(0,0).ixType().toIntVect() ), dm, ncompe, nge );
-    pml_E_fp[1] = std::make_unique<MultiFab>(amrex::convert( ba,
-        WarpX::GetInstance().getEfield_fp(0,1).ixType().toIntVect() ), dm, ncompe, nge );
-    pml_E_fp[2] = std::make_unique<MultiFab>(amrex::convert( ba,
-        WarpX::GetInstance().getEfield_fp(0,2).ixType().toIntVect() ), dm, ncompe, nge );
+    WarpX::GetInstance().AllocInitMultiFab(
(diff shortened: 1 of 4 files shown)


dpgrote:
Something like this is a good idea. Though, it looks like AMReX has a facility for setting the intial value. It looks like there are routines "MultiFab::set_do_initval" and "MultiFab::set_initval" that set it up, setting static variables. The default for "do_initval" is false. Maybe @WeiqunZhang can comment.

WeiqunZhang:
If this is meant for debugging only, you could use ParmParse parameter `fab.init_snan=true` to initialize all fab data to signaling NaNs.  You could also use `fab.do_initval=true` and `fab.initval=A_REAL_VALUE`.

EZoni:
@WeiqunZhang @dpgrote

In my view, this is not meant for debugging only. It is meant to facilitate code development overall.

Here's an example of a simple code development workflow that is simplified by having such a function:
When a new MultiFab `mf` is allocated in WarpX.cpp, in order to avoid uninitialized data, one has to remember to set its initial values (oftentimes to `0.0`) in a different file, namely Initialization/WarpXInitData.cpp; moreover, one has to remember that simply adding `mf->setVal(0.0)` there might still lead to errors, if `mf` is allocated in WarpX.cpp only when certain flags are activated, in which case that logic needs to be duplicated around the call `mf->setVal(0.0)` (or one needs to, at least, add an `if (mf)` check). All of this is avoided if the initial value `0.0` can be set directly when the MultiFab is allocated with the new function call, within a one-line instruction, within one file.

This is meant to be used only if necessary. From previous discussions I remember someone pointing out that one might not always want to initialize the values of a MultiFab, in which case the old allocations using the AMReX constructors work just fine.

ax3l:
I think that is very useful, but I would add one more function.

Goal: have a book/index of all MultiFabs in WarpX that we can query by name (same `std::string` as our MultiFab tag).

We often create an empty MultiFab, but want to delay allocating (`::define()`) it based on user runtime parameters (think: MR, PML, EBs, etc.). For that, we also need a helper function so we can push back the MultiFab refs/pointers (or objects themselves) in a book keeping location such as a `std::map<std::string, MultiFab*>`.

EZoni:
Closing this PR because all changes are included in #3484.

PR #3239, opened 2022-07-15
Fixes to allow mixed precision, ParticleReal float, Real double

dpgrote:
Changes are needed to allow compilation with mixed precision, with ParticleReal float, and Real double. Most of the changes are in the QED code so that it consistently uses ParticleReal, particularly needed when calling picsar code.

Note that when compiling with DEBUG=TRUE, there are still a very large number of warnings about implicit conversions, but at least the code compiles without error.


ax3l (approved):
Awesome, thanks a lot!

PR #3240, opened 2022-07-15
Use proper units for momenta, in BoundaryScrapingDiagnostic

RemiLehe:
In WarpX, the macroparticles' momenta are stored internally in `m.s^-1`, but in openPMD and plotfiles, the momenta need to be written in `kg.m.s^-1`. 

Therefore, inside the code that write particle properties to file, we convert the units of the momenta. This was done for the regular diagnostics, but not yet for the BoundaryScrapingDiagnostics and BackTransformedDiagnostics.


Source/Particles/ParticleIO.H
```diff
@@ -49,7 +50,7 @@ particlesConvertUnits (ConvertDirection convert_direction, T_ParticleContainer*
 #ifdef AMREX_USE_OMP
 #pragma omp parallel if (Gpu::notInLaunchRegion())
 #endif
-        for (WarpXParIter pti(*pc, lev); pti.isValid(); ++pti)
+        for (ParIter pti(*pc, lev); pti.isValid(); ++pti)
         {
             // - momenta are stored as a struct of array, in `attribs`
             auto& attribs = pti.GetAttribs();
```
dpgrote: I looked at this some more. As is, this doesn't compile since the `ParIter` class doesn't have a `GetAttribs` method. This can be made to compile by replacing this line with the direct call instead of GetAttribs,
```
auto& attribs = pti.GetStructOfArrays().GetRealData();
```
This worked and is now added.

dpgrote:
I tried merging in the latest version of development and there are conflicts, due to the PR that allowed writing out of scraped particles at all boundaries. I can fix them if you would like, since I'm familiar with that code.

RemiLehe:
@dpgrote That would be great. Thanks a lot for your help!

dpgrote:
@RemiLehe This PR is ready to go, though I do have one question (before I remove the WIP). The way it is now, in normal diagnostic output, the conversion of the momentum units is done on the particles in the particle container so that the multiplication can be done on the GPU. But with the pinned container for the lost particles, the conversion will presumably be done on the CPU since the data is in main memory. Is that correct? Or is the pinned data copied to the GPU, updated, and copied back in particlesConvertUnits? Either way, then perhaps a different way of doing the unit conversion might be better, to do it at the place where the data is gathered, copied from the pc to pinned memory, avoiding the extra work. I haven't done timing tests so don't know if this would matter.

RemiLehe:
That's a good point. 
Actually, because the new code is now calling a `ParallelFor` (i.e. GPU kernel, when WarpX is compiled for GPU) that touches data which is in pinned memory, I wonder if this will work at all on GPU. I can try to test it.

RemiLehe:
Given today's discussion with @WeiqunZhang, I think that we can merge this.

dpgrote:
Sounds good to me. Did you still want Axel or Edoardo to review it? Otherwise I can approve it.

EZoni (approved):
@RemiLehe @dpgrote Thank you for these changes, the PR looks good to me.

PR #3241, opened 2022-07-15
Specialize Field Gather when external field is None

lucafedeli88:
This PR specializes the Field Gather routine for the simplest case: no external field. This avoids the call to a relatively complex function in one of the most expensive kernels.


dpgrote:
Before you put much more effort in this, I noticed that this version of the `doGatherShapeN` routine isn't used anywhere.

EZoni:
@lucafedeli88 Thanks for opening this PR. Based on @dpgrote's comment, do you think we still want to have this PR? Would you like to comment on this?

lucafedeli88:
Ah, right... well... in this case I will close this PR...

PR #3243, opened 2022-07-16
ABLASTR: Update Poisson Solver API

ax3l:
Update the Poisson Solver API to be more usable, e.g., with non-temporary rho and phi fields.

Needed for ImpactX: https://github.com/ECP-WarpX/impactx/pull/162

Note: I noticed that during `computePhi`, we are manipulating the source `rho`. This is not ideal, since this is a side effect. We probably don't care so far, since we throw `rho` away afterwards. It would be nice if we could express this normalization of the source to AMReX in another way.

### DIFF
--- a/Source/ablastr/fields/PoissonSolver.H
+++ b/Source/ablastr/fields/PoissonSolver.H
@@ -10,6 +10,7 @@
 #include "Utils/WarpXConst.H"
 
 #include <ablastr/utils/Communication.H>
+#include <ablastr/utils/TextMsg.H>
 #include <ablastr/warn_manager/WarnManager.H>
 
 #include <AMReX_MultiFab.H>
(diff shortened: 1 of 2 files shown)


Source/ablastr/fields/PoissonSolver.H
```diff
@@ -125,7 +126,7 @@ computePhi (amrex::Vector<std::unique_ptr<amrex::MultiFab> > const & rho,
     // scale rho appropriately; also determine if rho is zero everywhere
     amrex::Real max_norm_b = 0.0;
     for (int lev=0; lev<=finest_level; lev++) {
-        rho[lev]->mult(-1._rt/PhysConst::ep0);
+        rho[lev]->mult(-1._rt/PhysConst::ep0); // TODO: when do we "un-multiply" this? We need to document this side-effect!
```
dpgrote, line 129: It is assumed that rho won't be used after phi is computed. Though I agree that this should be documented, perhaps in the documentation for the routine. Make rho [in/out] with an explanation. In WarpX, this assumption is correct, rho is not used afterward. For example, the rho diagnostics regenerate rho.

dpgrote (approved):
These changes look Ok by me.

Source/ablastr/fields/PoissonSolver.H
ax3l: In WarpX that's the case, it's temporary and then removed. But it's a bit surprising indeed and maybe we can expreaa source constants in an easier way to amrex

PR #3251, opened 2022-07-21
Add Python callback call when a checkpointing signal is received

roelof-groenewald:
We have some Python specific diagnostics that we checkpoint (timeseries of injected and scraped charges) whenever WarpX checkpoints are created. This callback allows us to trigger writing those diagnostics if the user sends a checkpointing signal to the WarpX process. 
This functionality is particularly useful to us while running simulations on AWS where SPOT instances can be arbitrarily reclaimed by AWS - we can now benefit maximally from the lower cost SPOT instances.


dpgrote (approved):
This looks good to me!

EZoni (approved):

PR #3252, opened 2022-07-21
CI: Add Missing Regression Analysis (NCI corrector)

EZoni:
While debugging #3222, I could not understand the behavior of two CI tests, `nci_corrector` and `nci_correctorMR`, namely why the regression analysis would always pass despite some changes that should have affected the checksums.

It turns out that the regression analysis was not done for these tests and no checksum JSON files were available.

Was the regression analysis avoided intentionally for these two tests, or was this simply overlooked by mistake? Could not find a PR that removed them, seems like they were simply not added since the beginning.


ax3l (approved):
Thank you - this looks indeed like a test bug to me.

PR #3253, opened 2022-07-22
Fix a bug in GPU version of Hankel Transform

WeiqunZhang:
amrex::Array4 is a 4D array that can be accessed with three spatial indices
plus an optional component index.  We must always provide all three spatial
indices even in 2D.


EZoni (approved):
Thank you for the fix, @WeiqunZhang! Looks good to me.

dpgrote:
That's interesting. The code seemed to be working though, producing the correct results when I tested it. Was this noticed by chance or was it causing problems?

WeiqunZhang:
Yes, it actually produces the correct result if we ignore the assertion.  This is because `i+nx*j+nx*ny*k+nx*ny*nz*n` is equal to `i+nx*j+nx*ny*n` when `k=0` and `nz=1`.

EZoni:
@dpgrote It was noticed by @RTSandberg while trying to run some RZ PSATD test in DEBUG mode in order to debug another PR, which could not be done because the runs were crashing anyways (only in DEBUG mode) because of the bug fixed here.

PR #3254, opened 2022-07-24
[WIP] Implement formula for high-energy D-T cross-section

RemiLehe:
This is a follow-up on #3153 . It implements the modified formula for the high-energy cross-section (applicable for E > 550 keV)
Screen Shot 2022-07-24 at 1 47 11 AM


NeilZaim (commented):
Thanks for this PR, the formula looks good to me. I only have a few minor comments:

- Looking at the paper (see screenshots below), it looks like it is slightly cleaner to use an energy transition of 530 keV instead of 550 keV.

image
image

- You can probably use `_prt` literals when defining the coefficients.

- I don't know what's faster between using constexprs (but then we have a branch in the computation of the astrophysical factor) or doing what you did here. But in any case, it's most likely a negligible part of the simulation time so we don't really have to worry about that.

EZoni:
@RemiLehe I saw I'm assigned to this PR, so this is just a reminder to make sure that it does not get lost. There were a couple of minor suggestions by @NeilZaim, and it seems like the PR would be ready to go once they are addressed, unless you have a different opinion on either of those.

EZoni:
Closing this PR without merging for now, after checking with @RemiLehe and @roelof-groenewald that, at this time, we do not have any use case at hand where this correction plays a role.

PR #2510, opened 2021-10-28
Correct collision algorithm in RZ.

Yin-YinjianZhao:
In this PR, we will try to do a particle momentum rotation before calling the collision function, such that different azimuth positions of particles are also taken into account.

PR #3255, opened 2022-07-24
Correct and test fusion module in RZ geometry

RemiLehe:
In RZ geometry, particles need to be temporarily rotated before/after the two-particle fusion event, in order to reflect the cylindrical symmetry of the distribution function for the plasma. (This is similar to what was implemented for the Coulomb collisions: https://github.com/ECP-WarpX/WarpX/pull/2510.)

In this PR we implement this rotation, but **only** for the incident particles (i.e. not yet for the products of the reaction, which will be done in a follow-up PR). As a result the reaction rates and number of particles created will be correct  (because the incident have been properly rotated), but the exact momentum distribution of the product species will not. (At least, not before we implement the rotation of the product, in the follow-up PR.)

This PR also adds an RZ test, where particles are initialized with an **azimuthal** velocity. The test does **not** check for isotropy in the momentum distribution of the products, because of the above-mentioned limitation for the momentum distribution of the products.


Examples/Modules/nuclear_fusion/analysis_deuterium_tritium_fusion.py
```diff
@@ -330,24 +344,25 @@ def check_fusion_yield(data, expected_fusion_number, E_com, reactant0_density, r
         assert(np.all(is_close(product_weight_theory, product_weight_simulation,
                                rtol = 5.*relative_std_weight)))
 
-def specific_check1(data):
-    check_isotropy(data, relative_tolerance = 3.e-2)
+def specific_check1(data, dt):
+    if not is_RZ:
+        check_isotropy(data, relative_tolerance = 3.e-2)
```
RemiLehe, line 354: We don't check for isotropy in the RZ case, since the distribution of momentum for the products is not yet correct.

Examples/Modules/nuclear_fusion/analysis_deuterium_tritium_fusion.py
```diff
@@ -66,15 +66,26 @@
 
 E_fusion = 17.5893*MeV_to_Joule # Energy released during the fusion reaction
 
+## Checks whether this is the 2D or the 3D test
+is_RZ = "RZ" in sys.argv[1]
+
```
EZoni: One other way to do this (which I personally find a little more robust, because it does not rely on the test's name) is to parse the output file `warpx_used_inputs` and determine the geometry from there:
```py
warpx_used_inputs = open('./warpx_used_inputs', 'r').read()
if re.search('geometry.dims = 2', warpx_used_inputs):
    is_RZ = False
elif re.search('geometry.dims = RZ', warpx_used_inputs):
    is_RZ = True
```

Regression/WarpX-tests.ini
```diff
@@ -2216,6 +2216,22 @@ compileTest = 0
...
+inputFile = Examples/Modules/nuclear_fusion/inputs_deuterium_tritium_rz
+runtime_params = warpx.do_dynamic_scheduling=0 warpx.serialize_initial_conditions=1
+dim = 2
+addToCompileString =
+cmakeSetupOpts = -DWarpX_DIMS=RZ
+restartTest = 0
+useMPI = 1
+numprocs = 2
+useOMP = 1
+numthreads = 2
```
EZoni: I think this needs to be maximum 1 if `numprocs = 2`:
```suggestion
numthreads = 1
```

EZoni (commented):
Thank you for this PR, Remi! Looks good to me, I just left a couple of minor comments.

EZoni (approved):

PR #3259, opened 2022-07-25
Starting with the pusher, consistently use ParticleReal

dpgrote:
This PR goes through the particle pusher and makes consistent use of the ParticleReal type, cleaning up mixed type declarations. This ended up touching code elsewhere to use the consistent type. For example, the type of the charge and mass of the species was changed to ParticleReal, requiring various changes.

One problematic area is the use of the PhysConst values, which are declared `Real`, leaving implicit conversions from `Real` to `ParticleReal`. Should these conversions be made explicit with `static_cast` (there are quite a number of them)?

### DIFF
--- a/Regression/Checksum/benchmarks_json/background_mcc_dp_psp.json
+++ b/Regression/Checksum/benchmarks_json/background_mcc_dp_psp.json
@@ -2,11 +2,11 @@
   "electrons": {
     "particle_cpu": 187506.0,
     "particle_id": 58260371529.0,
-    "particle_momentum_x": 1.0114377612798924e-18,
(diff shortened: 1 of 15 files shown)


ax3l (approved):
Looks great, thank you!

ax3l:
> One problematic area is the use of the PhysConst values, which are declared Real, leaving implicit conversions from Real to ParticleReal. Should these conversions be made explicit with static_cast (there are quite a number of them)?

Storing them in a `constexpr amrex::ParticleReal` as you did is the right way to cast and avoid runtime overheads from runtime casts.

PR #3262, opened 2022-07-26
ParticleReal for external fields and collisions

dpgrote:
This PR makes further changes to have consistent use of ParticleReal for the particle code.


ax3l (approved):
Thank you, that's great! :)

PR #3265, opened 2022-07-27
Fix: mpi4py requirement for `--target pip_install`

ax3l:
Installs `mpi4py` if not already found during
```
cmake --build build -j 4 --target pip_install
```


dpgrote (approved):
This looks good to me

PR #3270, opened 2022-07-28
[WIP] Add regions and BLProf only macros.

kngott:
First draft at adding profiling regions to the code.

Adds the Full Profiler only-macros, so this pass doesn't affect the current output, and took my best guess at reasonable regions for the UniformPlasma and KPP tests. First runs seem to work.

Iteration and discussion of where to put regions, and Full Profiler or TinyProfiler regions are welcome.

### DIFF
--- a/Source/Evolve/WarpXEvolve.cpp
+++ b/Source/Evolve/WarpXEvolve.cpp
@@ -80,6 +80,7 @@ WarpX::Evolve (int numsteps)
     const int step_begin = istep[0];
     for (int step = istep[0]; step < numsteps_max && cur_time < stop_time; ++step)
     {
+        WARPX_PROFILE_REGION_START("WarpX::Evolve::step");
         WARPX_PROFILE("WarpX::Evolve::step");
         Real evolve_time_beg_step = amrex::second();
 
@@ -368,6 +369,7 @@ WarpX::Evolve (int numsteps)
             break;
         }
 
+        WARPX_PROFILE_REGION_STOP("WarpX::Evolve::step");
         // End loop on time steps
     }
     multi_diags->FilterComputePackFlushLastTimestep( istep[0] );
@@ -394,7 +396,9 @@ WarpX::OneStep_nosub (Real cur_time)
     ExecutePythonCallback("particlescraper");
     ExecutePythonCallback("beforedeposition");
 
+    WARPX_PROFILE_REGION_START("WarpX::nosub::PushParticles");
     PushParticlesandDepose(cur_time);
+    WARPX_PROFILE_REGION_STOP("WarpX::nosub::PushParticles");
 
     ExecutePythonCallback("afterdeposition");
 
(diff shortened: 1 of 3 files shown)


kngott:
This is associated with the AMReX PR: https://github.com/AMReX-Codes/amrex/pull/2891

So, including that in the build would also show particle comm data.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -80,6 +80,7 @@ WarpX::Evolve (int numsteps)
     const int step_begin = istep[0];
     for (int step = istep[0]; step < numsteps_max && cur_time < stop_time; ++step)
     {
+        WARPX_PROFILE_REGION_START("WarpX::Evolve::step");
```
ax3l, line 83: Wait is `WARPX_PROFILE` not sufficient here?

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -394,7 +396,9 @@ WarpX::OneStep_nosub (Real cur_time)
     ExecutePythonCallback("particlescraper");
     ExecutePythonCallback("beforedeposition");
 
+    WARPX_PROFILE_REGION_START("WarpX::nosub::PushParticles");
     PushParticlesandDepose(cur_time);
```
ax3l, line 400: Do we directly want to put a `WARPX_PROFILE` inside `PushParticlesandDepose` instead?

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -403,11 +407,13 @@ WarpX::OneStep_nosub (Real cur_time)
     // the actual current J. This is computed later in WarpX::PushPSATD, by calling
     // WarpX::PSATDVayDeposition. The function SyncCurrent is called after that,
     // instead of here, so that we synchronize the correct current.
+    WARPX_PROFILE_REGION_START("WarpX::nosub::SyncJ_rho");
```
ax3l, line 410: Directly in `SyncCurrent` and `SyncRho` as `WARPX_PROFILE`?

Source/Evolve/WarpXEvolve.cpp
kngott: It could. Up for discussion. I stuck to wrapped section of Evolve and OneStep for now, as you generally want bigger, all-encompassing blocks of code for the FullProfiler analysis and probably matching these regions with the overall flow of the code rather than specific functions seemed like a better deal.

(After reading through the code, I was picturing all the OneStep functions instrumented in a matching way, meaning the profile gives consistent cuts no matter how you run WarpX.)

Source/Evolve/WarpXEvolve.cpp
kngott: It could. Same as above.

(Apparently, while switching focus between windows, I clicked right on the "resolve" button, so don't take that as any commentary. 😄)

Source/Evolve/WarpXEvolve.cpp
kngott: I specifically added and used PROFILE_REGION_START and REGION_STOP because START and STOP do not add a region to TinyProfiler outputs, only Full Profiler outputs. So, this change would not impact the output of most users runs. (Too many regions can make parsing the output impossible.)

They can be easily changed to a PROFILE_REGION (see line 64 for the only REGION example so far) if you also want an isolated TInyProfiler region out for each unique REGION you instrument.

PR #2115, opened 2021-07-20
Allow extra particle attributes (besides ux, uy, uz and w) to be set at particle creation in AddNParticles()

roelof-groenewald:
This PR follows on PR #2112 and should close out the remaining requirements for issue #1902.
The functionality in `WarpXParticleContainer.cpp:AddNParticles()` has been extended to also set extra particle attributes at creation (currently only weight is used from the `attr` array passed to that function). This in effect also fixes a bug with that function as it would currently set the weight incorrectly if more than 1 attribute value is passed to the `AddNParticles` function which would execute without any warning being displayed.

This PR should be merged after 
1. PR #2109
2. PR #2119 
3. PR #2126

PR #3272, opened 2022-07-29
Initialize particle runtime attributes inside AddNParticles

NeilZaim:
In the code, physical particles are created either through the `AddPlasma` routine or the `AddNParticles` routine. However, runtime attributes (QED optical depths, ionization levels and user defined attributes) are only initialized when the chosen injection style uses `AddPlasma`. This is **very dangerous**, because if the chosen injection style uses `AddNParticles` (as is the case for `gaussian_beam` injection for instance), the runtime attributes are silently zero-initialized, which can lead to incorrect but plausible results (this is particularly true for QED simulations).

In this PR, I thus initialize the runtime attibutes for the injection styles that use `AddNParticles`, building on top of #2115 (which allows `AddNParticles` to initialize runtime components, but this feature was never used except in the python bindings). The proposed solution is maybe not ideal (there's plenty of points that can be discussed) but in any case I think it's important to quickly merge a patch given how dangerous this bug is.


atmyers (approved):

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -622,17 +622,138 @@ PhysicalParticleContainer::AddPlasmaFromFile(ParticleReal q_tot,
...
+                particle_w.dataPtr(), 1, -1, amrex::RandomEngine{});
 #endif // WARPX_USE_OPENPMD
 
     ignore_unused(q_tot, z_shift);
 
     return;
 }
 
+void
+PhysicalParticleContainer::InitializeRuntimeAttributesAndAddNParticles (
```
dpgrote: Could this code setting up the attributes be incorporated into `AddNParticles`? This would avoid having extra copies of the attribute arrays (which could be problematic with large numbers of particles). Also, after these changes, is `AddNParticles` called directly anywhere else?

Examples/Modules/RigidInjection/analysis_rigid_injection_LabFrame.py
```diff
@@ -85,5 +85,17 @@ def remove_rigid_lines(plotfile, nlines_if_rigid):
 
 assert( error_rel < tolerance_rel )
 
+
+### Check that user runtime attributes are correctly initialized
```
dpgrote, line 92: This is a general WarpX comment - something that is missing is documentation about what the CI tests are testing. This is fine that this test is here, but it would be helpful if there was a `README` or some other standardized documentation saying that this rigid injection test is also testing the use of run time attributes. I've noticed this with other tests, that they test multiple things unrelated to the main topic of the test. Though, maybe this comment here is good enough?

Source/Particles/PhysicalParticleContainer.cpp
roelof-groenewald: `AddNParticles` can directly be called from Python - we make heavy use of it for our thermionic injectors. So my vote would be to not absorb too much extra functionality into `AddNParticles`.

Python/pywarpx/_libwarpx.py
```diff
@@ -625,6 +627,9 @@ def add_particles(self, species_name, x=None, y=None, z=None, ux=None, uy=None,
             # --- The -3 is because components 1 to 3 are velocities
             attr[:,self.get_particle_comp_index(species_name, key)-3] = vals
 
+        nattr_int = 0
+        attr_int = np.empty([0], ctypes.c_int)
```
roelof-groenewald, line 631: This is not necessarily needed for this PR but we should at some point improve the logic here to allow both real and integer attributes to be set via `add_particles`. This would require adding a function to `WarpXWrappers` to link to `getParticleiComps()`. This implementation somewhat moves the problem this PR is fixing on to Python specific injection i.e. integer runtime components will silently be set to 0 for newly injected particles. Maybe just creating a corresponding Issue is sufficient for now to remind us of this shortcoming.

roelof-groenewald (approved):
I left one comment above for a possible follow-up PR.

Source/Particles/PhysicalParticleContainer.cpp
dpgrote: Ok, presumably you are setting the needed attributes from Python. But would this have the same issue being addressed here, that some attributes may have incorrect default values? Also, this means that you couldn't use the `addRealAttributes` and `addIntegerAttributes` features. Perhaps the routine `InitializeRuntimeAttributesAndAddNParticles` could be made callable from Python also?

dpgrote: What I was thinking of was modifying AddNParticles to use the default values as being setup here in `InitializeRuntimeAttributesAndAddNParticles` for any attributes that are not passed in. This would not affect your current use from Python.

NeilZaim: Yes, I did not want to directly modify the `AddNParticles` function because I saw that it was used from Python. It is also still used when splitting particles in `PhysicalParticleContainer::SplitParticles`, possibly incorrectly with respect to runtime parameters (I don't know this function well enough but in this case we may need to copy the runtime parameters from the initial particles so the `InitializeRuntimeAttributesAndAddNParticles` may not be good).

Also, this PR indeed does not fix the issue when using Python but I wanted to fix the pure C++ version first because QED (which is the module for which the issue is the most dangerous IMO) does not yet work with Python, so I thought a Python fix was less urgent.

What you suggest in your latest message @dpgrote is that we modify `AddNParticles` so that it accepts a list of runtime parameters that are passed in `attr_real` and "default initializes" (same as I did here) the ones that are not passed in? If so, I think that it's generally a good idea. We'd have to slightly modify the Python interface so that `AddNParticles` can know which attributes are passed and we'd also have to make sure that we don't break the `SplitParticles` routine (but it may already be broken). I think this can be done either here or in a follow-up PR, what do you think?

Examples/Modules/RigidInjection/analysis_rigid_injection_LabFrame.py
NeilZaim: I admit that I'm guilty of frequently high-jacking some CI random tests to add quick tests of minor features and I have no idea if it's considered good practice. I think that if this part of the test breaks, then it's easy to understand that it's in fact the runtime attributes that are broken (so in that sense I reckon adding this new test is better than nothing).

But it's also true that it's very difficult to know if some minor feature is tested or not in CI.

Regarding the `README`, we often explain the test at the beginning of the analysis script, so maybe that's equivalent. I'll add a sentence at the beginning of the script to state that we also check runtime attributes here.

Python/pywarpx/_libwarpx.py
NeilZaim: Yes, this does not fix the integer runtime attributes initialization from Python, but they were already broken, right? (I just want to make sure that I'm not breaking anything in your workflows). I'd agree to say that this can go to a follow-up PR.

roelof-groenewald: That is correct, yes, it was already broken. This PR just made it clear so documenting the issue now would be good so we don't have to learn about it the hard way again later on.

Source/Particles/PhysicalParticleContainer.cpp
dpgrote: Yes, that is what I am suggesting. I make the suggestion since I think it will simplify the code some and make it more robust.

If I understand it correctly, currently `nattr` attributes are passed into `AddNParticles` and any beyond that are set to zero. Perhaps this PR could fix so that a proper default is used instead of only zero. A second PR could update the Python interface to allow more flexibility.

Examples/Modules/RigidInjection/analysis_rigid_injection_LabFrame.py
dpgrote: I think that adding extra tests to existing CI tests is ok, and maybe a good thing since it will reduce the overall number of tests. One caveat though is that if a test fails, it should be clear which thing being tested is actually failing.

Having an explanation in the analysis script sounds good to me.

NeilZaim:
After discussions with @dpgrote, we have decided to directly initialize the runtime attributes inside `AddNParticles`. This saves a bit of memory and is also more robust. The main difficulty is that `AddNParticles` is a `WarpXParticleContainer` method while most runtime attributes are physical and located in `PhysicalParticleContainer`.
Thus, inside `AddNParticles` we call a new method `DefaultInitializeRuntimeAttributes` that does nothing when called from `LaserParticleContainer` but that is overriden by `PhysicalParticleContainer`. This function initializes the runtime attributes that have not already been initialized inside `AddNParticles` by the `attr_real` and `attr_int` vectors.

Python/pywarpx/_libwarpx.py
NeilZaim: Note that with the new implementation, the int attributes should be correctly set with python (but there's no way to initialize them through a numpy array yet).

Source/Particles/PhysicalParticleContainer.H
```diff
@@ -221,6 +221,23 @@ public:
         amrex::Gpu::HostVector<amrex::ParticleReal>& particle_uz,
         amrex::Gpu::HostVector<amrex::ParticleReal>& particle_w);
 
+    /**
+     * \brief Default initialize runtime attributes in a tile.
```
dpgrote: Overall, this looks good. For clarity, perhaps add a comment here that this routine is only setting attributes that have a potentially nonzero default value and that were not set elsewhere.

NeilZaim: Thanks, I've updated this comment to state that this does not initialize attributes that were set elsewhere. (actually I think that this routine does zero initialize attributes that should be zero initialized, like `prev_x`, so I haven't added a comment on that part)

dpgrote (approved):
This looks good to me! Thanks!

PR #3195, opened 2022-06-22
[WIP] Implement scraping diagnostics from all boundaries

RemiLehe:
Addresses  #3130

PR #3274, opened 2022-07-30
Implement output of scraped particles at domain boundaries

dpgrote:
This replaces PR #3195. In the process of updating that PR, I managed to corrupt the commit history. The new commits from that branch were cherry picked into this branch, plus my new changes.

With this PR, the particles lost at the domain boundaries can be saved to output files. To separate the data, the data from each boundary is written to a separate subdirectory (as explained in the documentation). I've also added the option of setting an interval so that the data is written out during the simulation instead of only at the end.

A bit of extensive changes were needed since getting this to work since a new option was needed in the `WriteToFile` routines specifying a pinned particle container that was separate from the `isBTD` option.

The avoids writing out empty data sets (if there are no particles lost) since the empty data confuse the openpmd reader.

### DIFF
--- a/Examples/Tests/embedded_circle/inputs_2d
+++ b/Examples/Tests/embedded_circle/inputs_2d
@@ -37,8 +37,8 @@ electrons.density = 10.2e16
 electrons.profile = constant
 electrons.momentum_distribution_type = maxwell_boltzmann
 electrons.theta = (kb*30000/(m_e*clight^2))
-electrons.save_particles_at_zlo = 1
-electrons.save_particles_at_zhi = 1
+electrons.save_particles_at_xlo = 1
+electrons.save_particles_at_xhi = 1
 electrons.save_particles_at_eb = 1
 
 ar_ions.species_type = argon
@@ -55,8 +55,8 @@ ar_ions.uz_m = 0.0
 ar_ions.ux_th = 2.6285641070031447e-06
 ar_ions.uy_th = 2.6285641070031447e-06
 ar_ions.uz_th = 2.6285641070031447e-06
-ar_ions.save_particles_at_zlo = 1
-ar_ions.save_particles_at_zhi = 1
+ar_ions.save_particles_at_xlo = 1
+ar_ions.save_particles_at_xhi = 1
 ar_ions.save_particles_at_eb = 1
 
 collisions.collision_names = coll_electrons coll_ar_ions
@@ -80,8 +80,12 @@ coll_ar_ions.back_cross_section = ../../../../warpx-data/MCC_cross_sections/Ar/i
 coll_ar_ions.elastic_cross_section = ../../../../warpx-data/MCC_cross_sections/Ar/ion_scattering.dat
 coll_ar_ions.charge_exchange_cross_section = ../../../../warpx-data/MCC_cross_sections/Ar/charge_exchange.dat
 
-diagnostics.diags_names = diag1
+diagnostics.diags_names = diag1 diag3
 diag1.diag_type = Full
 diag1.format = plotfile
 diag1.intervals = 1
 diag1.fields_to_plot = phi rho_electrons rho_ar_ions
+
+diag3.diag_type = BoundaryScraping
+diag3.format = openpmd
+diag3.intervals = 5
--- a/Examples/Tests/scraping/analysis_rz.py
+++ b/Examples/Tests/scraping/analysis_rz.py
@@ -47,7 +47,7 @@ assert(error==tolerance)
 # remaining particles and scraped particles is equal to the
 # original number of particles
 ts_full = OpenPMDTimeSeries('./diags/diag2/')
-ts_scraping = OpenPMDTimeSeries('./diags/diag3/')
+ts_scraping = OpenPMDTimeSeries('./diags/diag3/particles_at_eb')
 
 def n_remaining_particles( iteration ):
     w, = ts_full.get_particle(['w'], iteration=iteration)
--- a/Source/Diagnostics/BoundaryScrapingDiagnostics.cpp
+++ b/Source/Diagnostics/BoundaryScrapingDiagnostics.cpp
@@ -35,18 +35,20 @@ BoundaryScrapingDiagnostics::ReadParameters ()
     m_varnames_fields = {}; // No fields in boundary scraping diagnostics
     m_varnames = {}; // No fields in boundary scraping diagnostics
 
-    // Number of buffers = 1 for BoundaryScrapingDiagnostics.
-    // (buffers are used in BTDiagnostics, and correspond to different snapshots)
-    m_num_buffers = 1;
+    // num_buffers corresponds to the number of boundaries
+    // (upper/lower domain boundary in each dimension)
+    // + the EB boundary if available
+    m_num_buffers = AMREX_SPACEDIM*2;
+#ifdef AMREX_USE_EB
+    m_num_buffers += 1;
+#endif
 
     // Do a few checks
-#ifndef AMREX_USE_EB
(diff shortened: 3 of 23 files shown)


Examples/Tests/embedded_circle/inputs_2d
```diff
@@ -37,8 +37,8 @@ electrons.density = 10.2e16
 electrons.profile = constant
 electrons.momentum_distribution_type = maxwell_boltzmann
 electrons.theta = (kb*30000/(m_e*clight^2))
-electrons.save_particles_at_zlo = 1
-electrons.save_particles_at_zhi = 1
+electrons.save_particles_at_xlo = 1
+electrons.save_particles_at_xhi = 1
```
RemiLehe, line 41: Could you explain why it was necessary to change `save_particles_at_zlo` to `save_particles_at_xlo` here?

Source/Diagnostics/BoundaryScrapingDiagnostics.cpp
```diff
@@ -133,35 +119,47 @@ BoundaryScrapingDiagnostics::DoComputeAndPack (int /*step*/, bool /*force_flush*
...
     const amrex::Geometry& geom = warpx.Geom(0); // For compatibility with `WriteToFile` ; not used
 
+    // The data for each boundary is written out to a separate directory with the boundary name
+    std::string file_prefix = m_file_prefix + "/particles_at_" + particle_buffer.boundaryName(i_buffer);
+
     m_flush_format->WriteToFile(
         m_varnames, m_mf_output[i_buffer], m_geom_output[i_buffer], warpx.getistep(),
-        0., m_output_species[i_buffer], nlev_output, m_file_prefix,
-        m_file_min_digits, false, false, isBTD, i_buffer, geom,
+        warpx.gett_new(0), m_output_species[i_buffer], nlev_output, file_prefix,
```
RemiLehe, line 158: Could you explain why we pass `warpx.gett_new(0)` here?
In principle, the `BoundaryScraping` collects particles throughout the simulation, so my understanding was that it cannot be associated with a specific time.

Source/Diagnostics/FlushFormats/FlushFormatSensei.cpp
```diff
@@ -55,14 +55,15 @@ FlushFormatSensei::WriteToFile (
     const amrex::Vector<ParticleDiag>& particle_diags,
     int nlev, const std::string prefix, int file_min_digits,
     bool plot_raw_fields, bool plot_raw_fields_guards,
+    const bool use_pinned_pc,
```
RemiLehe, line 58: Could we have use `/* use_pinned_pc */` here, as in the case of the other formats?

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -616,7 +618,9 @@ WarpXOpenPMDPlot::WriteOpenPMDParticles (const amrex::Vector<ParticleDiag>& part
       GeometryFilter const geometry_filter(particle_diags[i].m_do_geom_filter,
                                            particle_diags[i].m_diag_domain);
 
-      if (! isBTD) {
+      if (isBTD || use_pinned_pc) {
```
RemiLehe, line 621: Side-note for a potential follow-up PR:
There is some similar code here:
https://github.com/ECP-WarpX/WarpX/blob/development/Source/Diagnostics/FlushFormats/FlushFormatPlotfile.cpp
Should we try to unify these two functions before they diverge too much?

Examples/Tests/scraping/analysis_rz.py
```diff
@@ -47,7 +47,7 @@
 # remaining particles and scraped particles is equal to the
 # original number of particles
 ts_full = OpenPMDTimeSeries('./diags/diag2/')
-ts_scraping = OpenPMDTimeSeries('./diags/diag3/')
+ts_scraping = OpenPMDTimeSeries('./diags/diag3/particles_at_eb')
```
RemiLehe, line 50: For this test, would you mind setting `diag.intervals = 3`, so that as to test that the `intervals` feature is working as expected?

Source/Diagnostics/BoundaryScrapingDiagnostics.cpp
```diff
@@ -133,35 +119,47 @@ BoundaryScrapingDiagnostics::DoComputeAndPack (int /*step*/, bool /*force_flush*
...
-    // of the simulation for BoundaryScrapingDiagnostics, we always set `isLastBTD`.
-    // This tells WarpX to write all the metadata (and not purely the particle data)
-    bool const isLastBTD = true;
+    // If the saving of the particles was not set up for any of the species for this boundary
+    // or if no particles have been lost, then don't write anything out.
+    if (n_particles == 0) return;
+
+    // This is not a backtransform diagnostics
+    bool const isBTD = false;
+    bool const isLastBTD = false;
```
RemiLehe, line 148: I am not sure what happens when writing multiple times to the **same** openPMD file **without** passing `isBTD=true`. Do you have any thoughts on this? 
Is there a possibility e.g. that the current code will erase any previously existing dataset in the file, at each successive write, if we don't pass `isBTD=true`. (Note: I did not have a detailed look at the internals of the function `DumpToFile`.)

RemiLehe (commented):
Thanks for this PR!
I have a few minor comments ; would you have time to address them?

Examples/Tests/embedded_circle/inputs_2d
dpgrote: Thanks for the comments!
I don't know why this test case was setting `electrons.save_particles_at_zlo` and `hi` since it has periodic boundary conditions in z, so no particles are ever lost in z. Particles are lost in x however, so I changed it to x. I can update the analysis.py file in this test to check that the output of the scraped particles is correct.

Source/Diagnostics/BoundaryScrapingDiagnostics.cpp
dpgrote: With the allowing of intervals, the data can be written out into multiple files. Passing in the current time here just makes the files consistent. The time is the time when the particles are written out, so the file will have only particles lost before that time.

Source/Diagnostics/FlushFormats/FlushFormatSensei.cpp
dpgrote: That routine is already using the `ignore_unused` way of ignoring arguments. I was being consistent, using the same way as the original code. Perhaps a separate PR can clean this up and use the commented name arguments instead of `ignore_unused`.

Source/Diagnostics/WarpXOpenPMD.cpp
dpgrote: That was something that I was considering. I'll test this and see if the writing out of the particles will work with plotfiles (instead of only openpmd).

Examples/Tests/scraping/analysis_rz.py
dpgrote: The intervals is tested in the other test case, `embedded_circle`. I'll take that into account when updating the analysis file for that case.

Source/Diagnostics/BoundaryScrapingDiagnostics.cpp
dpgrote: When the intervals is set, the scraped particle data is written out into separate files, a file each time the data is written out. This seems more robust, so for example the scraped particle data will be available as the simulation is running, and will be available if for some reason the simulation doesn't finish (it runs out of time or crashes for some other reason). This makes it natural to read in the data using as an `OpenPMDTimeSeries`.

RemiLehe (approved):

PR #3277, opened 2022-08-01
`SyncCurrent`: Pass References To Current MultiFabs

EZoni:
This is a follow-up PR to #2965. The flexibility of passing to `SyncCurrent` any current MultiFabs as input arguments is needed in a variety of other PRs related to the development and debugging of Vay's current deposition scheme. I also added more detailed comments to explain what happens with current centering, when the hybrid scheme is used.

### DIFF
--- a/Source/WarpX.H
+++ b/Source/WarpX.H
@@ -643,9 +643,21 @@ public:
     void FillBoundaryG   (int lev, amrex::IntVect ng, const bool nodal_sync = false);
     void FillBoundaryAux (int lev, amrex::IntVect ng);
 
+    /**
+     * \brief Apply filter and sum guard cells across MR levels.
+     * If current centering is used, center the current from a nodal grid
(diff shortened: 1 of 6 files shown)


Source/WarpX.H
```diff
@@ -643,7 +643,9 @@ public:
     void FillBoundaryG   (int lev, amrex::IntVect ng, const bool nodal_sync = false);
     void FillBoundaryAux (int lev, amrex::IntVect ng);
 
-    void SyncCurrent ();
+    void SyncCurrent (
```
ax3l, line 657: Just a note: please add doxygen for the new params.

RemiLehe (approved):

Source/WarpX.H
EZoni: Yes, thank you. Done in https://github.com/ECP-WarpX/WarpX/pull/3277/commits/de8942733bc9821681631a27f5f8846992a6af0f.

PR #3279, opened 2022-08-02
Properly check that `save_particles_at_eb=1` when using `BoundaryScraping` diagnostic

### LINKED ISSUE #3248: (Question) An embedded implicit function with boundary scraping gives error messages
Hello,

I am trying to use embedded boundary with BoundaryScraping for tracking beam particles escaping from cylindrical plasma channel.
So I set related parameters as below.

```
my_constants.r0 = 1.e-3
warpx.eb_implicit_function = "( x * x + y * y - r0 * r0 )"

driver.save_particles_at_eb = 1
driver.save_particles_at_zlo = 1
driver.save_particles_at_zhi = 1

diag2.diag_type = BoundaryScraping
diag2.format = openpmd
```

But I got error messages as below.

```
 !!!
SIGABRT
0::Assertion `m_do_boundary_buffer[boundary][index]' failed, file "/home/kmoon/warpx-src/Source/Particles/ParticleBoundaryBuffer.cpp", line 337, Msg: 
### ERROR   : Attempted to get particle buffer for boundary 6, which is not
#            used!
 !!!
```

I am not understanding what this message means..

Can I just save escaping particles in OpenPMD and check later?
Are these functions already working in WarpX 22.07?
I am attaching WarpX input script and its output in any case you would want to see..
 
[input_and_log.zip](https://github.com/ECP-WarpX/WarpX/files/9142813/input_and_log.zip)
 
Could I have any hint of this situation?

Thanks a lot,
Kook-Jin

RemiLehe:
The code for `BoundaryScraping` contains some code that checks that the flag `save_particles_at_eb` has been activated. However, this code relies on the variable `m_output_species` being properly initialized, which is not the case in the current version of WarpX.

I moved this check to the function `InitializeParticleBuffer`, just after `m_output_species` is initialized.

This fixes #3248.

### DIFF
--- a/Source/Diagnostics/BoundaryScrapingDiagnostics.cpp
+++ b/Source/Diagnostics/BoundaryScrapingDiagnostics.cpp
@@ -47,29 +47,8 @@ BoundaryScrapingDiagnostics::ReadParameters ()
     amrex::Abort("You need to compile WarpX with openPMD support, in order to use BoundaryScrapingDiagnostic: -DWarpX_OPENPMD=ON");
 #endif
 
-    // Check that saving at EB has been activated for each requested species
-    std::set<std::string> particle_saving_activated;
-    for (auto const& species_name : m_output_species_names){
-        amrex::ParmParse pp(species_name);
-        bool save_particles_at_eb;
-        pp.query("save_particles_at_eb", save_particles_at_eb);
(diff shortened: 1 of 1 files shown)


dpgrote (approved):
Yes, this looks correct. Thanks for the fix.

RemiLehe:
I think that this PR is actually made obsolete by #3274.

PR #3281, opened 2022-08-02
Allow filtering in z with RZ FDTD

dpgrote:
This PR allows WarpX to be used with filtering along the `z` axis when using the RZ FDTD solver. When using RZ and FDTD, It changes the check to require either `use_filter` is false or `filter_npass_each_dir[0] == 0`, no passes along the radius.


Source/WarpX.cpp
```diff
@@ -730,9 +730,9 @@ WarpX::ReadParameters ()
         }
         else // FDTD
         {
-            // Filter currently not working with FDTD solver in RZ geometry
+            // Filter currently not working with FDTD solver in RZ geometry along R
             // (see https://github.com/ECP-WarpX/WarpX/issues/1943)
-            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(!use_filter,
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(!use_filter || filter_npass_each_dir[0] == 0,
                 "Filter currently not working with FDTD solver in RZ geometry");
```
RemiLehe: Could you modify the error message to say that filtering only works along z for the FDTD solver in RZ?
Could the error message also tell the user how to set filtering along z only in their input script (using `n_pass_each_dir`)?

RemiLehe (changes requested):
Thanks for this PR.

Could you also activate filtering (in z) in one of the RZ FDTD automated tests?

Source/WarpX.cpp
dpgrote: I added filtering in z in the LaserAccelerationRZ test and updated the benchmark.

RemiLehe (approved):

PR #3282, opened 2022-08-02
[WIP] PML with incident angle: FDTD

amadou38:
Adding new extra terms that take into account the incident angle for the case.

NB: for the moment, I explicitely added the angle in EvolveEPML, EvolveBPML in order to check their extra terms separately (0 or not). But the theta_pml from the input file will be accessible only by alpha* for the moment.

### DIFF
--- a/Source/FieldSolver/FiniteDifferenceSolver/EvolveEPML.cpp
+++ b/Source/FieldSolver/FiniteDifferenceSolver/EvolveEPML.cpp
@@ -43,6 +43,7 @@ using namespace amrex;
 /**
  * \brief Update the E field, over one timestep
  */
+
 void FiniteDifferenceSolver::EvolveEPML (
     std::array< amrex::MultiFab*, 3 > Efield,
     std::array< amrex::MultiFab*, 3 > const Bfield,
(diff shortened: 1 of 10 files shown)


Source/FieldSolver/FiniteDifferenceSolver/EvolveEPML.cpp
```diff
@@ -151,10 +190,14 @@ void FiniteDifferenceSolver::EvolveEPMLCartesian (
 
                 Ey(i, j, k, PMLComp::yx) -= c2 * dt * (
                     T_Algo::DownwardDx(Bz, coefs_x, n_coefs_x, i, j, k, PMLComp::zx)
-                  + T_Algo::DownwardDx(Bz, coefs_x, n_coefs_x, i, j, k, PMLComp::zy) );
+                  + T_Algo::DownwardDx(Bz, coefs_x, n_coefs_x, i, j, k, PMLComp::zy)
+                  - sin_th*cos_th*(T_Algo::DownwardDy(By, coefs_y, n_coefs_y, i, j, k, PMLComp::yz)
+                  + T_Algo::DownwardDy(By, coefs_y, n_coefs_y, i, j, k, PMLComp::yx)) );
```
EZoni: One comment discussed offline: for each of the new terms (like dBy/dy here), one needs to check that the correct finite difference function ("downward" v. "upward") is used, depending on whether the field, which the derivative is applied to, is nodal or cell-centered along the direction of that derivative, according to the Yee scheme.

Issue #3283, opened 2022-08-03
BTD crashing with invalid particle error


RevathiJambunathan:
For a 2D psatd test case, the plotfile format for BTD crashed with 
```
amrex::Abort::0::ParticleContainer::locateParticle(): invalid particle. !!!
SIGABRT
See Backtrace.0.3 file for details
--------------------------------------------------------------------------
MPI_ABORT was invoked on rank 0 in communicator MPI COMMUNICATOR 3 DUP FROM 0
with errorcode 6.

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
You may or may not see output from other processes, depending on
exactly when Open MPI kills them.
```

Looking into this further, the particle in question is located at z=1.72299362e-05, and it gets mapped to index `(17,1536) `
However, the particle BoxArray is 
```
 ba : (BoxArray maxbox(1)
       m_ref->m_hash_sig(0)
       ((0,1280) (63,1535) (0,0)) )
```
Thus the particle is out-of-bounds. 
The input file was provided by @RTSandberg

RTSandberg:
Running on Perlmutter in debug mode, the simulation crashes with the following message:
```
STEP 1 ends. TIME = 4.365974052e-16 DT = 4.365974052e-16 
Evolve time = 0.018899698 s; This step = 0.018899698 s; Avg. per step = 0.018899698 s

STEP 2 starts ...
 (64,559,0,0) is out of bound (0:63,512:639,0:0,0:0)
... <lots of similar out of bound messages> ...
 (-1,2116,0,0) is out of bound (0:63,2048:2175,0:0,0:0) 
```
The `WarpX.e` file says
```
/global/u2/r/rsand/src/WarpX/build_perlmutter_lattice/_deps/fetchedamrex-src/Src/Base/AMReX.H:152: void amrex::Abort(const char *): block: [26,0,0], thread: [75,0,0] Assertion `0` failed.
...
/global/u2/r/rsand/src/WarpX/build_perlmutter_lattice/_deps/fetchedamrex-src/Src/Base/AMReX.H:152: void amrex::Abort(const char *): block: [16,0,0], thread: [62,0,0] Assertion `0` failed.
amrex::Abort::0::CUDA error 710 in file /global/homes/r/rsand/src/WarpX/build_perlmutter_lattice/_deps/fetchedamrex-src/Src/Base/AMReX_GpuDevice.cpp line 660: device-side assert triggered !!!
SIGABRT
```

[...]

PR #3285, opened 2022-08-03
Use extended box array in setting par GDB for BTD

RevathiJambunathan:
This PR fixes the error in Issue #3283 

The GPU simulation crashed on perlmutter with 1 GPU after fixing the ParGDB array. 
A synchronize() call fixed this issue.

### DIFF
--- a/Source/Diagnostics/ComputeDiagFunctors/BackTransformParticleFunctor.cpp
+++ b/Source/Diagnostics/ComputeDiagFunctors/BackTransformParticleFunctor.cpp
@@ -145,6 +145,7 @@ BackTransformParticleFunctor::operator () (PinnedMemoryParticleContainer& pc_dst
                    if (Flag[i] == 1) GetParticleLorentzTransform(dst_data, src_data, i,
                                                                  old_size + IndexLocation[i]);
                 });
+                amrex::Gpu::synchronize();
             }
         }
     }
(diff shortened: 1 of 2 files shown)


Source/Diagnostics/ComputeDiagFunctors/BackTransformParticleFunctor.cpp
```diff
@@ -145,6 +145,7 @@ BackTransformParticleFunctor::operator () (PinnedMemoryParticleContainer& pc_dst
                    if (Flag[i] == 1) GetParticleLorentzTransform(dst_data, src_data, i,
                                                                  old_size + IndexLocation[i]);
                 });
+                amrex::Gpu::synchronize();
```
ax3l, line 148: Awesome catch :tada: 

Can you add a comment why we need to synchronize here? Is this because we reuse `FlagForPartCopy` and `IndexForPartCopy` for every particle tile in the loop?

RTSandberg (approved):
Thanks, Reva!

ax3l (approved):
I'll merge for now, happy weekend and congratulations!

Source/Diagnostics/ComputeDiagFunctors/BackTransformParticleFunctor.cpp
RevathiJambunathan: thats right!

RevathiJambunathan: oh but this is merged already. I will add a mini-PR

PR #3286, opened 2022-08-03
PICMI: Fix Bug in Class `Species` Init

EZoni:
I think this is a bug in the `init` method of the PICMI class `Species`.

If `particle_type` is not passed as input, it defaults to `None` and the assignment
```py
m = re.match(r'(?P<iso>#[\d+])*(?P<sym>[A-Za-z]+)', self.particle_type)
```
fails with error "expected string or bytes-like object".

I think `m` needs to be set to `None` by hand if `self.particle_type` is `None`.

### DIFF
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -61,23 +61,26 @@ class Species(picmistandard.PICMI_Species):
                     self.charge = '-q_e'
                 else:
                     self.charge = self.charge_state*constants.q_e
-            # Match a string of the format '#nXx', with the '#n' optional isotope number.
-            m = re.match(r'(?P<iso>#[\d+])*(?P<sym>[A-Za-z]+)', self.particle_type)
-            if m is not None:
-                element = periodictable.elements.symbol(m['sym'])
-                if m['iso'] is not None:
-                    element = element[m['iso'][1:]]
-                if self.charge_state is not None:
-                    assert self.charge_state <= element.number, Exception('%s charge state not valid'%self.particle_type)
-                    try:
-                        element = element.ion[self.charge_state]
-                    except ValueError:
-                        # Note that not all valid charge states are defined in elements,
-                        # so this value error can be ignored.
-                        pass
-                self.element = element
-                if self.mass is None:
-                    self.mass = element.mass*periodictable.constants.atomic_mass_constant
+            if self.particle_type is not None:
+                # Match a string of the format '#nXx', with the '#n' optional isotope number.
+                m = re.match(r'(?P<iso>#[\d+])*(?P<sym>[A-Za-z]+)', self.particle_type)
+                if m is not None:
+                    element = periodictable.elements.symbol(m['sym'])
+                    if m['iso'] is not None:
+                        element = element[m['iso'][1:]]
+                    if self.charge_state is not None:
+                        assert self.charge_state <= element.number, Exception('%s charge state not valid'%self.particle_type)
+                        try:
+                            element = element.ion[self.charge_state]
+                        except ValueError:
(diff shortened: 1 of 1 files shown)


Python/pywarpx/picmi.py
```diff
@@ -62,7 +62,9 @@ def init(self, kw):
                 else:
                     self.charge = self.charge_state*constants.q_e
             # Match a string of the format '#nXx', with the '#n' optional isotope number.
-            m = re.match(r'(?P<iso>#[\d+])*(?P<sym>[A-Za-z]+)', self.particle_type)
+            m = None
+            if self.particle_type is not None:
+                m = re.match(r'(?P<iso>#[\d+])*(?P<sym>[A-Za-z]+)', self.particle_type)
             if m is not None:
```
dpgrote: Thanks for finding this. Instead of setting `m = None`, can you put the lines 65-80 (old line numbers) inside an if statement `if self.particle_type is not None:`?

Python/pywarpx/picmi.py
```diff
@@ -61,9 +61,9 @@ def init(self, kw):
                     self.charge = '-q_e'
                 else:
                     self.charge = self.charge_state*constants.q_e
-            # Match a string of the format '#nXx', with the '#n' optional isotope number.
-            m = re.match(r'(?P<iso>#[\d+])*(?P<sym>[A-Za-z]+)', self.particle_type)
-            if m is not None:
+            if self.particle_type is not None:
+                # Match a string of the format '#nXx', with the '#n' optional isotope number.
+                m = re.match(r'(?P<iso>#[\d+])*(?P<sym>[A-Za-z]+)', self.particle_type)
                 element = periodictable.elements.symbol(m['sym'])
```
dpgrote: Sorry, this block of code, from lines 67 to 80 (new line numbers) still needs to be inside the if test `if m is not None:` (indented one level more).
Also, something that was missing from the code is an error message when the `particle_type` is bad. This could be added here, adding an `else` clause that has the line
```
raise Exception('The species "particle_type" is not known')
```

EZoni: Thank you, Dave. I reinstated the `if m is not None:` condition in the last commit (force-pushed), hope I got it right this time.

Regarding the exception, I'm not sure we need it. I see that when I type in an unknown `particle_type`, say "blabla", I get anyways the error "ValueError: unknown element blabla" after the call `element = periodictable.elements.symbol(m['sym'])`. The `else` clause that you suggested, where did you want it to be exactly (like, corresponding to which `if` clause)?

Python/pywarpx/picmi.py
```diff
@@ -61,23 +61,24 @@ def init(self, kw):
...
+                        assert self.charge_state <= element.number, Exception('%s charge state not valid'%self.particle_type)
+                        try:
+                            element = element.ion[self.charge_state]
+                        except ValueError:
+                            # Note that not all valid charge states are defined in elements,
+                            # so this value error can be ignored.
+                            pass
+                    self.element = element
+                    if self.mass is None:
+                        self.mass = element.mass*periodictable.constants.atomic_mass_constant
```
dpgrote, line 81: Thanks for the changes so far. The `else` would be here, after line 81, as the `else` to the `if m is not None:`. It is possible to set `particle_type` to a string where the `re.match` returns None. For example, try something that starts with a number, `particle_type = '2test'`.

EZoni: Got it, thank you. I tested it with a wrong input starting with a number, as you suggested, and it raises the exception: I think this is implemented correctly now.

dpgrote (approved):
Great! Thanks!

PR #3288, opened 2022-08-04
Coarsen IO : tmp mf to be constructed with ngrowvect

RevathiJambunathan:
This will fix a bug related to Issue #3283 
The cell-centered mf for BTD has 1 guard cell (its required.) 
However, in coarsen IO, the tmp_mf constructed to copy data from src to dst does not have the same number of guard cells as the dst cell-centered mf. Hence the code fails with out of bounds error. 
This bug gets triggered by BTD only, since all other diags has 0 guard cells.


EZoni (approved):
Thank you for this bug fix, @RevathiJambunathan! Looks good to me.

PR #3290, opened 2022-08-05
Fix Bugs w/ Current Correction and Vay Deposition

### LINKED ISSUE #3232: Issue with `current_correction` and domain decomposition
Current correction off, running on 6 GPUs:
Non-laser transverse polarization without domain decomposition
electron density, Ez, and lineouts

Current correction on, 1 GPU and 1 grid:
Non-laser transverse polarization with trivial domain decomposition
electron density, Ez, and lineouts

Current correction on, 6 GPUs and 6 grids
Non-laser transverse polarization with domain decomposition
electron density, Ez, and lineouts

Note that in the preceding images, the Ey label should be Ex.

Input deck for these simulations:
[inputs.txt](https://github.com/ECP-WarpX/WarpX/files/9283741/inputs.txt)

This manifests in all fields and at lower orders.

@n01r also observed this; consider 
Figure 32
Figure 31
[inputs_2d.txt](https://github.com/ECP-WarpX/WarpX/files/9095853/inputs_2d.txt)

See also issue #3166, where an acausal instability was resolved by turning off current_correction

On the other hand, current_correction seems to be in fact beneficial in the boosted frame, see issue #2462.  The prior were lab frame simulations.

EZoni:
Fix #3232. Filtering will also need to be moved to the right locations, though (depends on another open PR).

Currently not supporting (to be addressed in future PRs):
- Vay deposition and current correction at the same time;
- Vay deposition with periodic single box (cumulative sums to be fixed).

To-do:
- [x] Add CI tests to cover case without periodic single box (2D, 3D, RZ), possibly running on 2 MPI processes (merge #3322 first)

Tests:

- Example with 2D Langmuir test with 2x2 grids, with current correction and direct deposition ([inputs_2d.txt](https://github.com/ECP-WarpX/WarpX/files/9273773/inputs_2d.txt)):

<p align="center">
    
    
</p>

- Example with RZ LWFA with 1x6 grids, with current correction and direct deposition (input deck attached as first test case reported in #3232, modified by adding `warpx.numprocs = 1 6` to have one grid per GPU):

<p align="center">
    
    
</p>

<p align="center">
    
    
</p>


Source/FieldSolver/WarpXPushFieldsEM.cpp
```diff
@@ -322,6 +322,8 @@ void WarpX::PSATDForwardTransformRho (
     const amrex::Vector<std::unique_ptr<amrex::MultiFab>>& charge_cp,
     const int icomp, const int dcomp)
 {
+    if (charge_fp[0] == nullptr) return;
```
EZoni, line 357: Added this to avoid complex logic where this function is called, such as what we had previously:
```cpp
    // Do rho FFTs only if needed
    if (WarpX::update_with_rho || WarpX::current_correction || WarpX::do_dive_cleaning)
    {
        PSATDForwardTransformRho(rho_fp, rho_cp, 0, 0); // rho old
        PSATDForwardTransformRho(rho_fp, rho_cp, 1, 1); // rho new
    }
```

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -398,19 +398,43 @@ WarpX::OneStep_nosub (Real cur_time)
...
+    // filter (if used), exchange guard cells, interpolate across MR levels
+    if (maxwell_solver_id == MaxwellSolverAlgo::PSATD)
+    {
+        if (fft_periodic_single_box)
+        {
+            // With periodic single box, synchronize J and rho here,
+            // even with current correction or Vay deposition
+            if (current_deposition_algo == CurrentDepositionAlgo::Vay)
+            {
+                SyncCurrent(current_fp_vay, current_cp);
```
RemiLehe: Should `current_cp` be replaced by `current_cp_vay` in theory?

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -398,19 +398,43 @@ WarpX::OneStep_nosub (Real cur_time)
...
-    // Synchronize J and rho: filter, exchange boundary, interpolate across levels.
-    // With Vay current deposition, the current deposited at this point is not yet
-    // the actual current J. This is computed later in WarpX::PushPSATD, by calling
-    // WarpX::PSATDVayDeposition. The function SyncCurrent is called after that,
-    // instead of here, so that we synchronize the correct current.
-    // With current centering, the nodal current is deposited in 'current_fp_nodal':
-    // SyncCurrent stores the result of its centering into 'current_fp' and then
-    // performs both filtering, if used, and exchange of guard cells.
-    if (WarpX::current_deposition_algo != CurrentDepositionAlgo::Vay)
+    // Synchronize J and rho:
```
RemiLehe, line 401: Maybe this all block of code could be extracted and put in a separate function, in order to improve the readability of the code.

Source/Evolve/WarpXEvolve.cpp
EZoni: In theory yes, but I think we do not have the `current_cp_vay` MultiFab defined anywhere yet, because Vay deposition is not currently fully implemented with mesh refinement. Maybe I could add a note as inline comment ("TODO") to remember that this will need to be changed once Vay deposition is fully implemented with mesh refinement.

Source/FieldSolver/WarpXPushFieldsEM.cpp
```diff
@@ -606,46 +608,118 @@ WarpX::PushPSATD ()
...
+            PSATDCurrentCorrection();
 
-    PSATDForwardTransformJ(J_fp, current_cp);
+            // Inverse FFT of J
+            PSATDBackwardTransformJ(current_fp, current_cp);
+        }
+        else if (current_deposition_algo == CurrentDepositionAlgo::Vay)
+        {
+            // FFT of D and rho (if used)
+            PSATDForwardTransformJ(current_fp_vay, current_cp);
```
RemiLehe, line 664: Same comment on `current_cp` here.

Source/WarpX.cpp
```diff
@@ -1327,10 +1334,15 @@ WarpX::ReadParameters ()
...
-        if (WarpX::current_deposition_algo == CurrentDepositionAlgo::Vay)
+        // Without periodic single box, fill guard cells with backward FFTs,
+        // with current correction or Vay deposition
+        if (fft_periodic_single_box == false)
         {
-            WarpX::fill_guards = amrex::IntVect(1);
+            if (current_correction ||
+                current_deposition_algo == CurrentDepositionAlgo::Vay)
+            {
+                fill_guards = amrex::IntVect(1);
```
RemiLehe: Does this flag also affect the inverse FFT of E and B?

Source/Evolve/WarpXEvolve.cpp
EZoni:
```suggestion
                // TODO Replace current_cp with current_cp_vay once Vay deposition is implemented with MR
                SyncCurrent(current_fp_vay, current_cp);
```

EZoni: Inline comment added in https://github.com/ECP-WarpX/WarpX/pull/3290/commits/7fcd631358aa22a57f8e3f63e93b8450a9ee99d0.

RemiLehe (commented):
Thanks a lot for this PR. 
I have a few comments and questions (see inline).

Source/FieldSolver/WarpXPushFieldsEM.cpp
EZoni:
```suggestion
            // FFT of D and rho (if used)
            // TODO Replace current_cp with current_cp_vay once Vay deposition is implemented with MR
            PSATDForwardTransformJ(current_fp_vay, current_cp);
```

Source/WarpX.cpp
EZoni: Yes, I believe it also affects the inverse FFT of E and B, because the flag `fill_guards` is passed to the constructor of the spectral solver, and from there to the relevant methods of the class `SpectralFieldData`. We can discuss offline if we want to fill the guard cells only for J by adding more control on this.

EZoni: Thanks to #3302, the new behavior with current correction affects only J.

EZoni:
Even though these bug fixes solve every issue reported in #3232, the code seems to be buggy when tested on the case reported in #2462 (Galilean PSATD in RZ geometry), which was actually the only case to be seemingly working fine before the issues reported in #3232 came out. I will try to investigate more and report if needed.

Note that we now perform FFTs of J and rho two times, before and after the current correction. This could be problematic when filtering is done in k-space from within the FFT functions, as it is the case for PSATD in RZ geometry. So I believe we currently apply filtering twice, which is not correct.

EZoni:
The bug reported above in https://github.com/ECP-WarpX/WarpX/pull/3290#issuecomment-1217029995 was fixed in the last commit, https://github.com/ECP-WarpX/WarpX/pull/3290/commits/67de99d4a6aee4a09bdefe5f08504fc0e4b855b3.

@RemiLehe @dpgrote
I had to introduce an additional boolean flag to control when to apply filtering in k-space for RZ PSATD, which is relevant when we compute FFTs of J and rho twice (e.g., before and after the current correction, in the case without periodic single box). Let me know if you have suggestions for a different implementation of the bug fix.

RemiLehe (approved):

RemiLehe:
Thanks a lot for this PRs and for fixing the associated bugs!

PR #3291, opened 2022-08-07
Doc: PICMI Parameters

ax3l:
This adds all currently defined PICMI parameters to the user-facing docs. See #3292

## Follow-up PRs: Help wanted :smiley: 

- [ ] As a follow-up, we should add a section on additional APIs that we already expose in WarpX for field data, particle data, boundary scraping, hooks, etc. (This has been on my todo list @dpgrote)
- [ ] We furthermore need to document the additional PICMI `warpx_` kwargs that we allow and not yet standardize.


dpgrote (approved):
Thanks for working on this, looks good to me

PR #2336, opened 2021-09-24
[WIP] Decrease number of guard cells allocated as a function of interpolation order

NeilZaim:
I'm wondering if we're not overly cautious in the number of guard cells we allocate as a function of interpolation order. For instance, for a regular simulation with order 3 shape factors, we will end up with 4 guard cells (because we round it up later on to the closest even integer), while I thought that 2 would be necessary.

So I'm trying to reduce this number in this PR and see if this affects our automated tests.

(This should not affect PSATD simulations, where the number of allocated guard cells is dominated by the needs of the field solver and not by the interpolation order)

**Edit**: Updates after the CI tests. Some tests have been failing and I've had to make a few modifications:
- Some tests were PEC boundary conditions were failing. This is because the boxes used inside PEC routines were directly grown by the shape factor, which is no longer compatible with this PR (the total number of guard cells can be lower than the shape factor, e.g. in the case of order 3). This resulted in out-of-bound array accesses.
- The benchmarks for the test `averaged_galilean_2d_psatd_hybrid` have changed. After investigating, this is because this PR changes the number of allocated guard cells from `(9,16)` to `(8,16)`. Indeed, the number of guard cells in the x direction was interestingly not determined by the field solver order, but by the high cfl used, which means that particles can move by 4 cells in the x direction within a timestep, and resulted in a computed value of `ng_alloc_Rho` of `(9,6)` (eventually the number of allocated guard cells is the same for all arrays, so all arrays ended up with `(9,16)` guard cells). However, the value of `9` was obtained by assuming that 3 guard cells are necessary for a shape factor of 3, which is not the case (only 2 are needed). Hence, I think it makes sense that this test only needs 8 guard cells in the x direction and that we can reset the benchmarks (which I've done).
- I've also had to reset the benchmark for the momentum of the ions in the x direction in the test `LaserAccelerationBoost` (which was off by ~1e-7). I've investigated this and it seems to be machine precision errors in the current deposition (which end up making 1e-7 errors in the ion momentum, which is practically 0 in the x direction). With the changes of this PR, the number of guard cells for the J field decrease from 4 to 3 in this test. This can affect the results (machine precision errors) via the calculation of `xyzmin` here:
https://github.com/ECP-WarpX/WarpX/blob/051234024144992c37dac526d92d10f2b8553bc6/Source/Particles/WarpXParticleContainer.cpp#L420
[...]

PR #2282, opened 2021-09-03
Fix uninitialized particle data

NeilZaim:
Another issue that I have observed using valgrind. When we use `AddPlasma` to add new particles, we usually create too many particles and discard the invalid ones by setting their id to -1. The parameters (positions, momenta, etc.) of the discarded particles are never initialized. However, nothing guarantees that these uninitialized values will not be used before the next call the `Redistribute`.

This can happen when we use continuous injection. In this case, we call the routine, `mypc->ApplyBoundaryConditions()` after the continuous injection and before any `Redistribute`. And since this routine uses the particle positions, valgrind complains about uninitialized values being used. (I think it's quite harmless however, because it only affects particles that should be deleted anyways)

I thought it'd be safest to always initialize the values of discarded particles to 0, so I wrote a function that does that and sets the id of the particle to -1. However, I find the result a bit cumbersome, so maybe there's a better/simpler way to remove the valgrind warning (e.g. just initializing all values to 0 at the beginning of the loop on the particles in `AddPlasma`, or just calling `Redistribute` immediatly after continuous injection).

PR #3294, opened 2022-08-08
Fix out of bounds array access with Particle Scraping + Continuous Injection

NeilZaim:
When we use `AddPlasma` to add new particles, we usually create too many particles and discard the invalid ones by setting their id to `-1`. Since #2282, we also set all particle quantities to `0` in order to make sure that no uninitialized quantities are used between the `AddPlasma` routine and the next call to `Redistribute`. However, errors can still occur with invalid particles if we call a routine that does some form of field gathering before calling `Redistribute` and if the position `(0,0,0)` is not in the current box. I've found that such a segfault can occur if we use continuous injection combined with EB particle scraping (I've observed the issue locally and it's very likely that this was causing the segfault in the macOS CI test of #2336).

This issue can easily be fixed by adding a line like `if (particles[ip].id() < 0) return;` inside the `scrapeParticles` function. A more robust solution could be to call `Redistribute` after every call to the continuous injection routines but that would probably make the code slower.

### DIFF
--- a/Source/EmbeddedBoundary/ParticleScraper.H
+++ b/Source/EmbeddedBoundary/ParticleScraper.H
@@ -156,7 +156,7 @@ scrapeParticles (PC& pc, const amrex::Vector<const amrex::MultiFab*>& distance_t
             auto& tile = pti.GetParticleTile();
             auto ptd = tile.getParticleTileData();
             const auto np = tile.numParticles();
-            auto& particles = tile.GetArrayOfStructs();
+            amrex::ArrayOfStructs<0, 0, amrex::DefaultAllocator>& particles = tile.GetArrayOfStructs();
             auto phi = (*distance_to_eb[lev])[pti].array();  // signed distance function
             amrex::ParallelForRNG( np,
             [=] AMREX_GPU_DEVICE (const int ip, amrex::RandomEngine const& engine) noexcept
--- a/Source/EmbeddedBoundary/ParticleScraper.H
+++ b/Source/EmbeddedBoundary/ParticleScraper.H
@@ -156,10 +156,14 @@ scrapeParticles (PC& pc, const amrex::Vector<const amrex::MultiFab*>& distance_t
             auto& tile = pti.GetParticleTile();
             auto ptd = tile.getParticleTileData();
             const auto np = tile.numParticles();
+            auto& particles = tile.GetArrayOfStructs();
             auto phi = (*distance_to_eb[lev])[pti].array();  // signed distance function
             amrex::ParallelForRNG( np,
             [=] AMREX_GPU_DEVICE (const int ip, amrex::RandomEngine const& engine) noexcept
             {
+                // skip particles that are already flagged for removal
+                if (particles[ip].id() < 0) return;
+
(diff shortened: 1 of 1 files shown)


Source/EmbeddedBoundary/ParticleScraper.H
```diff
@@ -156,10 +156,14 @@ scrapeParticles (PC& pc, const amrex::Vector<const amrex::MultiFab*>& distance_t
             auto& tile = pti.GetParticleTile();
             auto ptd = tile.getParticleTileData();
             const auto np = tile.numParticles();
+            auto& particles = tile.GetArrayOfStructs();
             auto phi = (*distance_to_eb[lev])[pti].array();  // signed distance function
             amrex::ParallelForRNG( np,
             [=] AMREX_GPU_DEVICE (const int ip, amrex::RandomEngine const& engine) noexcept
             {
+                // skip particles that are already flagged for removal
+                if (particles[ip].id() < 0) return;
```
dpgrote, line 165: Others can chime in, but I think using `return` in the lambda expression is not a good idea. On a CPU, the `ParallelFor` gets expressed as a loop, so the `return` returns out of the whole routine.

dpgrote:
Could `AddPlasma` use the same technique that is used in `AddPlasmaFlux`? It creates a temporary particle container for the new particles and calls `Redistribute` on that, and then copies the new particles into the appropriate container.

Source/EmbeddedBoundary/ParticleScraper.H
NeilZaim: I've seen `return` used in other `ParallelFor` in the code (see below for example), so I guess it should be ok (also, the CPU tests have all passed).

https://github.com/ECP-WarpX/WarpX/blob/5a5f8f89ecd4294d271bb66aa0e236887243ba67/Source/Particles/WarpXParticleContainer.cpp#L1052-L1059

NeilZaim:
Regarding whether we could call `Redistribute` from inside `AddPlasma`, I'll let @atmyers reply. But if it is a negligible performance loss, then I'd say that this is a good idea.

Source/EmbeddedBoundary/ParticleScraper.H
atmyers: `return` will return from the lambda function call for that loop element, rather than the enclosing routine. Using `continue` in this context, however, which is easy to do when porting a regular C++ `for` loop to an `amrex::ParallelFor`, is usually a bug because it doesn't actually break out of the lambda function.

atmyers (approved):

atmyers:
`AddPlasma` could be implemented that way, but I think we should do some performance tests at scale if we go that route.

Source/EmbeddedBoundary/ParticleScraper.H
dpgrote: Ok, thanks. I remember reading about the problem with returns in the past. Did this change at some point?

atmyers: Not for `ParallelFor`, but `return` was and still is problematic for macros like `AMREX_FOR_1D`.

PR #3295, opened 2022-08-08
[WIP] Laser initialization from spectral information

RTSandberg:
Allow a user to launch a laser using spectral intensity and phase.  The spectral information is converted to a temproal profile and stored in openPMD format. The openPMD file is loaded in WarpX and used to initialize a laser pulse


RTSandberg:
This is precluded by LASY

PR #3298, opened 2022-08-10
Enable field ionization from PICMI

NeilZaim:
The goal of this PR is to take into account in WarpX the changes made in this PICMI Pull Request: https://github.com/picmi-standard/picmi/pull/75.

I haven't tested it yet because I first wanted to make sure that this is generally how it should be done. cc @dpgrote

Also, how do we synchronously merge this PR and picmi's PR? Is the general workflow to
1. Merge the picmi PR
2. Update the picmi version
3. Use the new version inside this PR and then we can merge it? (and we possibly have to do this for other codes that use picmi like warp or fbpic)


Python/pywarpx/picmi.py
```diff
@@ -163,14 +163,13 @@ def initialize_inputs(self, layout,
...
 
-        for interaction in self.interactions:
-            assert interaction[0] == 'ionization'
-            assert interaction[1] == 'ADK', 'WarpX only has ADK ionization model implemented'
-            self.species.do_field_ionization=1
-            self.species.physical_element=self.particle_type
-            self.species.ionization_product_species = interaction[2].name
-            self.species.ionization_initial_level = self.charge_state
-            self.species.charge = 'q_e'
+    def activate_ionization(self, product_species):
```
dpgrote: Thanks for the PR. This looks ok, but is different than what I would have done. I would have put the code here in the  `initialize_inputs` method in the `FieldIonization` class. This more closely follows the pattern of how things are done in PICMI. That class has the references to the species and so can set these parameters that are related to the field ionization setup. Also, there is no need to use `add_new_attr` - these parameters can be set directly, `self.species.do_field_ionization = 1`. `add_new_attr` is there for when the attributes have special characters or is constructed.

Python/pywarpx/picmi.py
```diff
@@ -1126,7 +1119,11 @@ def initialize_inputs(self):
 class FieldIonization(picmistandard.PICMI_FieldIonization):
     def initialize_inputs(self):
         assert self.model == 'ADK', 'WarpX only has ADK ionization model implemented'
```
dpgrote, line 1216: Can you add the WarpX specific doc string for this class that states that only ADK is supported here?

NeilZaim: I've just added a docstring which I hope is what is needed here.

dpgrote:
I'll update the picmi version now that PR #75 was merged. picmistandard is now up to version 0.0.21.

NeilZaim:
@dpgrote Sorry I had abandonned this PR for a while, but it should be ready now.

dpgrote (approved):
Looks good! Thanks!

PR #3302, opened 2022-08-10
Fill Guard Cells of Fields and Currents Independently

EZoni:
There are situations where it is necessary to fill the guard cells when inverse FFTs of fields and/or currents are computed. For example, we want to fill the guard cells of fields when damped BCs are applied, or we want to fill the guard cells of currents when current correction or Vay deposition are used.

However, so far we only had one single parameter which would apply to both fields and currents. This is not necessary and we prefer to control what we do with the guard cells of fields and currents independently.

The RZ PSATD solver always fills the guard cells along the z direction, so no additional control is needed there. This is the reason why we have different interfaces for the relevant functions (leading to `#ifdef` branching in the implementation). Let me know if we want to change that and have default values (aka, always false, always ignored) in the RZ case.

Other comments to be discussed were added as single comments below.

### DIFF
--- a/Source/FieldSolver/SpectralSolver/SpectralAlgorithms/SpectralBaseAlgorithm.cpp
+++ b/Source/FieldSolver/SpectralSolver/SpectralAlgorithms/SpectralBaseAlgorithm.cpp
@@ -31,9 +31,7 @@ SpectralBaseAlgorithm::SpectralBaseAlgorithm(const SpectralKSpace& spectral_kspa
     const amrex::DistributionMapping& dm,
     const SpectralFieldIndex& spectral_index,
     const int norder_x, const int norder_y,
-    const int norder_z, const bool nodal,
-    const amrex::IntVect& fill_guards):
-        m_fill_guards(fill_guards),
+    const int norder_z, const bool nodal):
         m_spectral_index(spectral_index),
     // Compute and assign the modified k vectors
         modified_kx_vec(spectral_kspace.getModifiedKComponent(dm,0,norder_x,nodal)),
@@ -65,8 +63,6 @@ SpectralBaseAlgorithm::ComputeSpectralDivE (
     field_data.ForwardTransform(lev, *Efield[0], Idx.Ex, 0 );
     field_data.ForwardTransform(lev, *Efield[1], Idx.Ey, 0 );
     field_data.ForwardTransform(lev, *Efield[2], Idx.Ez, 0 );
-
-    const amrex::IntVect& fill_guards = m_fill_guards;
 
     // Loop over boxes
     for (MFIter mfi(field_data.fields); mfi.isValid(); ++mfi){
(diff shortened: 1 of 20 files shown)


Source/FieldSolver/SpectralSolver/SpectralAlgorithms/SpectralBaseAlgorithm.cpp
```diff
@@ -107,5 +103,6 @@ SpectralBaseAlgorithm::ComputeSpectralDivE (
     }
 
     // Backward Fourier transform
-    field_data.BackwardTransform(lev, divE, Idx.divE, 0, fill_guards);
+    const amrex::IntVect& fill_guards = amrex::IntVect(0);
+    field_data.BackwardTransform(lev, divE, Idx.divE, fill_guards, 0);
```
EZoni, line 107: Here I'm setting `fill_guards` so that the guard cells are **not** filled for the spectral div(E) diagnostics, because I feel like it isn't really necessary to do so. Please let me know if you think they should be filled instead.

Source/FieldSolver/SpectralSolver/SpectralFieldData.H
```diff
@@ -127,7 +127,7 @@ class SpectralFieldData
                                const int i_comp);
 
         void BackwardTransform (const int lev, amrex::MultiFab& mf, const int field_index,
-                                const int i_comp, const amrex::IntVect& fill_guards);
+                                const amrex::IntVect& fill_guards, const int i_comp);
```
EZoni, line 130: Here I swapped the arguments `i_comp` and `fill_guards` to match the same interface of the method `BackwardTransform` in the class `SpectralSolver`.

Source/FieldSolver/WarpXPushFieldsEM.cpp
```diff
@@ -152,15 +159,22 @@ void WarpX::PSATDBackwardTransformEBavg (
 {
     const SpectralFieldIndex& Idx = spectral_solver_fp[0]->m_spectral_index;
 
+    // No need to fill guard cells for averaged fields
+    const amrex::IntVect& fill_guards = amrex::IntVect(0);
```
EZoni: Here I'm setting `fill_guards` so that the guard cells are **not** filled for the **averaged** E and B fields, since those are not damped when we apply "damped" boundary conditions (which is the only case where we currently want to fill the guard cells for inverse FFTs of the E and B fields). I could change this if we think it would more robust.

Source/FieldSolver/SpectralSolver/SpectralAlgorithms/SpectralBaseAlgorithm.cpp
RemiLehe: I agree.

Source/FieldSolver/WarpXPushFieldsEM.cpp
EZoni: Changed this in 844df7f: we now treat the averaged E and B fields in the same way as the regular E and B fields.

RemiLehe (approved):

PR #3307, opened 2022-08-11
`isort`: AMReX, WarpX, etc. as First Party

ax3l:
Mark the `amrex` and `pywarpx`, `picmistandard`, ... imports as first party, so that they do not change if run locally or remotely pre/post install.

- https://pycqa.github.io/isort/docs/configuration/custom_sections_and_ordering.html
- https://github.com/PyCQA/isort/issues/448#issuecomment-456595042

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -35,3 +35,4 @@ indent_size = unset
 [*.py]
 # isort config
 force_sort_within_sections = true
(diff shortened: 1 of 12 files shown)


roelof-groenewald (approved):

dpgrote (approved):
Yes, this looks good. This is the PEP8 standard import order, built-ins first, then third party, then your own packages.

PR #3308, opened 2022-08-11
[WIP] BTD openPMD: Finalize w/o Particles

ax3l:
Handle the corner-case that we finalize a labframe simulation without having particles in that last dump.

Otherwise `SetConstParticleRecordsEDPIC` will cause a
```
A Record can not be written without any contained RecordComponents: id
```

Once we configure the record component attributes.

## Notes to Reviewer

- [ ] We investigate a crash from an inputs file in RZ for particles, so this might be not what we want to do. In our inputs file, we have a weird situation that there are no writes ever to the lab station for particles, but suddenly it shows up as if it had written 50k particles. For the same case in 3D, we see that there actually is a pc passed with 50k particles at some point, which is missing in RZ.

Thus, this patch here **might better not be merged** if we find the root cause of this.

I suspect this is caused by the bookkeeping logic in `BTDiagnostics::Flush`` with regards to passed particle containers for actual writes and book keeping in `UpdateTotalParticlesFlushed` for `m_totalParticles_flushed_already`.


ax3l:
This was a work-around and is addressed in the book keeping that goes into the function.

Issue #3292, opened 2022-08-07
Documenting the Python APIs


ax3l:
PICMI APIs and usage is, besides links to outside docs and examples, not really documented on the same level as the C++ app w/ inputs file usage. Let's change that.

- [x] document the standard PICMI APIs in the WarpX docs: #3291 @ax3l
  - [ ] We furthermore need to document the additional PICMI `warpx_` kwargs that we allow and not yet standardize. @dpgrote 
- [x] As a follow-up, we should add a section on additional APIs that we already expose in WarpX for field data, particle data, boundary scraping, hooks, etc.: #3310 @roelof-groenewald 

@dpgrote @RemiLehe @peterscherpelz @roelof-groenewald et al. help wanted :)

ax3l:
Unassigned, since I am looking for help with the latter two points :-)

PR #3310, opened 2022-08-12
Docs: Add description of Python APIs in `libwarpx`

roelof-groenewald:
This PR is in response to @ax3l's request for increased Python docs (Issue #3292).

### DIFF
--- a/Docs/source/usage/python.rst
+++ b/Docs/source/usage/python.rst
@@ -145,6 +145,10 @@ The input file should have the line ``sim.step()`` which runs the simulation.
 where ``<n_ranks>`` is the number of MPI ranks used, and ``<python_script>``
 is the name of the script.
 
+
+Extending a Simulation from Python
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
 When running WarpX directly from Python it is possible to interact with the simulation
 by installing ``CallbackFunctions``, which will execute a given Python function at a
 specific location in the WarpX simulation loop.
--- a/Docs/source/usage/python.rst
+++ b/Docs/source/usage/python.rst
@@ -145,6 +145,56 @@ The input file should have the line ``sim.step()`` which runs the simulation.
 where ``<n_ranks>`` is the number of MPI ranks used, and ``<python_script>``
 is the name of the script.
 
+When running WarpX directly from Python it is possible to interact with the simulation
+by installing ``CallbackFunctions``, which will execute a given Python function at a
+specific location in the WarpX simulation loop.
+
+.. autoclass:: pywarpx.callbacks.CallbackFunctions
+
+Places in the WarpX loop where callbacks are available include:
+``afterinit``, ``beforecollisions``, ``aftercollisions``, ``beforeEsolve``, ``afterEsolve``,
+``beforedeposition``, ``afterdeposition``, ``beforestep``, ``afterstep``, ``afterdiagnostics``,
+``afterrestart`` and ``oncheckpointsignal``.
+
+There are several "hooks" available via the ``libwarpx`` shared library to access and manipulate
+simulation objects (particles, fields and memory buffers) as well as general properties
+(such as processor number). These "hooks" are accessible through the `Simulation.extension` object.
+
+.. autofunction:: pywarpx.libwarpx.getNProcs
+
+.. autofunction:: pywarpx.picmi.Simulation.extension.getNProcs
+
+.. autofunction:: pywarpx.picmi.Simulation.extension.getMyProc
+
+.. autofunction:: pywarpx.picmi.Simulation.extension.get_nattr
+
+.. autofunction:: pywarpx.picmi.Simulation.extension.get_nattr_species
+
+.. autofunction:: pywarpx.picmi.Simulation.extension.getistep
+
+.. autofunction:: pywarpx.picmi.Simulation.extension.gett_new
+
+.. autofunction:: pywarpx.picmi.Simulation.extension.evolve
+
+.. autofunction:: pywarpx.picmi.Simulation.extension.finalize
+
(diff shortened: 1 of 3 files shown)


roelof-groenewald:
I wanted to open an early (unfinished) PR for this to make sure I am on the right path here in terms of how you guys would like to structure the Python documentation.
@ax3l you suggested that we add user-facing function docs as sub-sections to Parameters, but I think for the APIs that allow interaction with the simulation it makes sense to add it to the "Running WarpX directly from Python" section. Please let me know if you disagree.

Docs/source/usage/python.rst
```diff
@@ -145,6 +145,56 @@ The input file should have the line ``sim.step()`` which runs the simulation.
...
+``beforedeposition``, ``afterdeposition``, ``beforestep``, ``afterstep``, ``afterdiagnostics``,
+``afterrestart`` and ``oncheckpointsignal``.
+
+There are several "hooks" available via the ``libwarpx`` shared library to access and manipulate
+simulation objects (particles, fields and memory buffers) as well as general properties
+(such as processor number). These "hooks" are accessible through the `Simulation.extension` object.
+
+.. autofunction:: pywarpx.libwarpx.getNProcs
+
+.. autofunction:: pywarpx.picmi.Simulation.extension.getNProcs
```
roelof-groenewald: @ax3l and @dpgrote please let me know which of the above forms you guys prefer. The intent is that users always interact with `libwarpx` through the `simulation` object, right? For that reason I think referencing the functions as part of `Simulation.extension` makes sense.

dpgrote: I agree that the latter is better since the is the recommended way of accessing the functions.

Python/pywarpx/_libwarpx.py
```diff
@@ -456,8 +462,11 @@ def getistep(self, level=0):
         Parameter
         ---------
 
-        level : the refinement level to reference
+        level : int
```
dpgrote, line 465: My own preference is to keep the parameter documentation on the one line (with extra lines indented). For me, this is especially important when there are multiple parameters, like with `add_particles`. It is a lot easier for me to parse the text in the original format.

dpgrote (commented):
This is a good start, though see my comments.

Python/pywarpx/_libwarpx.py
roelof-groenewald: I definitely agree with you that in the `.py` file the inline formatting is much easier to read. The reason I changed it is because the auto generated documentation is not so readable with that formatting. The way it was renders the documentation as:
old

while the reformated doc-string produces:
new

If there is a way to get the documentation formatted appropriately while keeping the inline doc-strings in the `.py` file I'd be all for that.

dpgrote: In the picmi scripts, for example fields.py, uses this format

`- grid: grid object to be used by the solver (grid object)`

giving a bulleted list on the web page. I see now that your updated format is what is used in simulations.py, so it is consistent with that.

roelof-groenewald: The bulleted list is good for readability as well but I noticed there is a downside with it as well when dealing with multi-line descriptions as with `unique_particles` below
```
        - w                : Particle weights. Default = 0. (array or scalars)
        - unique_particles : Whether the particles are unique or duplicated on several
                             processes. Default = ``True``. (bool)
        - kwargs           : Containing an entry for all the extra particle attribute arrays.
                             If an attribute is not given it will be set to 0. (dict)
```
which renders as 
old

To render properly the docstring must be:
```
        - w                : Particle weights. Default = 0. (array or scalars)
        - unique_particles : Whether the particles are unique or duplicated on several
          processes. Default = ``True``. (bool)
        - kwargs           : Containing an entry for all the extra particle attribute arrays.
          If an attribute is not given it will be set to 0. (dict)
```

I think sticking with the formatting used in `simulations.py` will be best, especially with blank lines added between arguments, which helps a lot with readability in the `.py` file.

roelof-groenewald: Just for posterity I found out this specific way of formatting docstrings is called "Numpydoc".
https://stackoverflow.com/questions/3898572/what-are-the-most-common-python-docstring-formats

dpgrote: Ok, sounds good

roelof-groenewald:
There are still a bunch of functions in `_libwarpx.py` related to getting mesh data that is not covered in the docs but maybe we can tackle those in a next PR.
Apparently the package [pyment](https://pypi.org/project/pyment/) can be used to change all the docstrings in a file to a different style. I haven't tested it yet but might be a nice tool to homogenize the formatting style for all docstrings in `pywarpx`.

Docs/source/usage/python.rst
```diff
@@ -145,6 +145,98 @@ The input file should have the line ``sim.step()`` which runs the simulation.
 where ``<n_ranks>`` is the number of MPI ranks used, and ``<python_script>``
 is the name of the script.
 
+When running WarpX directly from Python it is possible to interact with the simulation
```
ax3l, line 152: Probably easier to read if we add a subsection here:
```suggestion

Extending a Simulation from Python
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

When running WarpX directly from Python it is possible to interact with the simulation
```

ax3l (commented):
This looks great to me, feel free to finalize and then it can be merged :)

ax3l (approved):
Thank you, that's splendid :rocket: :sparkles:

roelof-groenewald:
@ax3l was there something more you wanted me to change on this PR or can we merge this one?

ax3l:
Forgot to hit the button after doing the final update. Thanks for the ping :)

Issue #3305, opened 2022-08-11
Boosted-frame diagnostic can crash when `prob_hi` is >0 in `z`


RemiLehe:
I found an instance in which the boosted-frame diagnostic crashes with SegFault.
This is specifically in the case where `prob_hi` is strictly positive in `z`, which in general we don't recommend for boosted-frame simulation. Therefore, this issue is maybe not critical.

Here is a reproducing script:
[inputs_3d.txt](https://github.com/ECP-WarpX/WarpX/files/9306103/inputs_3d.txt)

In the case where `prob_hi` is 0, the code runs without any issue:
[inputs_3d_zmax0.txt](https://github.com/ECP-WarpX/WarpX/files/9306104/inputs_3d_zmax0.txt)

RemiLehe:
We discussed this issue offline with @ax3l @RevathiJambunathan @atmyers 

Here are a few steps that we discussed could be taken in order to understand the issue better:
- With the current input script (using openPMD output): we could compile in debug mode, in order to understand **where** the issue is happening.
- With a modified input script using plotfiles instead of openPMD output: the code does not SegFault, but instead raises a `locateParticle` error. It would be good to check whether this happens in `Redistribute` or when writing plotfiles.
- With could try to increase the tolerance of the `Redistribute` from one guard cells to two guard cells.

We also discussed #3311 in more detail, and it seems that the half cell shift is not needed because of the `floor` function.

RevathiJambunathan:
sorry for the incorrect information during our offline discussion. 

I have some notes and updates below. 

The test segfaults even with plotfile, not with `locateParticle`
Let me look into this more. but here is a preliminary update :
It segfaults at line https://github.com/ECP-WarpX/WarpX/blob/ab94900feb60cfdc50430cb7f5059587d6dba2e5/Source/Diagnostics/ComputeDiagFunctors/BackTransformFunctor.cpp#L120
Because k_lab = 1279 , but the bounds of the box for the buffer is ((0,0,1280) (15,15,1535) (0,0,0)) )
This is what prompted me to check k-index in PR #3311 

 Also , here is how k-index evolves for every timestep for this case : 
 
![Screenshot from 2022-08-19 14-47-15](https://user-images.githubusercontent.com/41089244/185711735-82186adf-2e0e-4221-89c0-e5da5dc83133.png)


Note that : STEP 256, ideally klab should have been 1280, again prompting me to do check k-index computation in PR #3311

[...]

PR #3311, opened 2022-08-12
[WIP ]BTD : add half cell shift when computing kindex for the slices in lab-=frame

RevathiJambunathan:
Two changes in the PR  : 

When getting k-index for the zslice : 
we do `(zlab - problo)/dz `
But the data is cell-centered
so it should be `(zlab - (problo + 0.5*dz)  ) / dz` 

When running the input in Issue #3305 with this change , I ran into an assert that ensures that the z-slice is wtthin
the bounds of the lab-frame buffer (buffer_lo, buffer_hi) 
But, since one of the slices was co-incident with the bound, I added a little epsilon to the assert, such that we check if zlice is within (lo-0.1*dz, hi+0.1*dz)

and now the input file in Issue #3305 runs without segfault.

### DIFF
--- a/Source/Diagnostics/BTDiagnostics.cpp
+++ b/Source/Diagnostics/BTDiagnostics.cpp
@@ -555,9 +555,11 @@ BTDiagnostics::PrepareFieldDataForOutput ()
                         }
                         DefineFieldBufferMultiFab(i_buffer, lev);
                     }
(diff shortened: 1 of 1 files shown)


RevathiJambunathan:
but CI test fails

RevathiJambunathan:
PR #3330 is the right solution to fix Issue #3305 .Hence closing this PR.

Issue #3306, opened 2022-08-11
Particles can cross the plasma lens without seeing it, for large timesteps


RemiLehe:
As mentioned in the title, macroparticles can sometimes cross the plasma lens without being affected. See the following plot obtained with this simulation script:
[inputs_3d_zmax0.txt](https://github.com/ECP-WarpX/WarpX/files/9310034/inputs_3d_zmax0.txt)
![Screen Shot 2022-08-11 at 9 00 46 AM](https://user-images.githubusercontent.com/6685781/184177811-e303c2ec-363a-4820-adcd-603aecb698fc.png)

(This uses the following plotting code:
```python
from openpmd_viewer import OpenPMDTimeSeries, ParticleTracker
ts = OpenPMDTimeSeries('./diags/back_3d/')
pt = ParticleTracker( ts, iteration=0, preserve_particle_index=True )
z_trajectories, x_trajectories, y_trajectories = ts.iterate( ts.get_particle, ['z', 'x', 'y'], select=pt)
plt.plot(z_trajectories[:,::200], x_trajectories[:,::200], 'b', alpha=0.1)
plt.axvline( x=0.0145, color='r' )
plt.axvline( x=0.0155, color='r' )
```

When checking which particles are focused, it seems that this depends essentially on their longitudinal position within the bunch:
![Screen Shot 2022-08-11 at 9 00 06 AM](https://user-images.githubusercontent.com/6685781/184177839-2f1bba26-21e5-4c4d-a821-f12fc10dfcaf.png)

It seems that, in this simulation, the distance traveled in z by a macroparticle during one timestep (1.11e-05 m) is comparable to the length of the lens in the boosted-frame (1.66e-05 m). Thus, considering that the lens and particles are counter-propagating in the boosted-frame, it could be that some macroparticles never fall within the range of the lens.

dpgrote:
Currently, the plasma lens requires particles to land at least one time step in the lens for the field to be applied. This can possibly be fixed by adjusting the residence correction that is done for the lens to account for the case when the particle skips over the lens. A PR will be posted soon.

dpgrote:
This should be fixed by PR #3318

dpgrote:
PR #3318 is merged.

PR #3318, opened 2022-08-15
Fixed plasma lens residence correction to allow a short lens

dpgrote:
This updates the residence correction to include the case when the particles do not land inside of the lens on a time step (i.e. the time step is too big or the lens too short). When this happens, the correction will be the length of the element divided by the length of the time step (`vz*dt`).


RemiLehe (changes requested):

Would you have time to add the case featured in [this issue](https://github.com/ECP-WarpX/WarpX/issues/3306) as an automated test? If not, I can also try to add it myself to this PR.

dpgrote:
@RemiLehe Instead of adding the example you showed in the issue, I added a new test that copies a previous plasma lens test but uses a much shorter lens. I did it this way so that the test is consistent with the existing tests and so it can use the existing analysis file.

RemiLehe (approved):

PR #3320, opened 2022-08-16
Wheel: Separate Directory

ax3l:
The `pip_install` target did not yet work on Windows, because it has no universal wildcard `*` support for our pip install logic.

Fix this by creating the wheel in a sub-directory and installing it by "finding all wheels for `pywarpx` in a given prefix".

- Avoid wildcards by using a separate directory.
- Avoid downloading deps during `wheel` build again.
- Fix Manifest: follow-up to #3265


EZoni (approved):
As far as I can understand this, it looks good to me. I'll let @dpgrote merge this.

dpgrote (approved):
This looks good to me!

PR #3322, opened 2022-08-17
CI: Clean up Analysis of Galilean Tests

EZoni:
While adding some new Galilean tests in #3290, I noticed that the analysis scripts for the existing tests could be cleaned up and improved.

To-do:
- [x] Merge 3D analysis script with 2D/RZ one (to avoid code duplication)
- [x] Parse input parameters from output file `warpx_used_inputs` instead of parsing from test name
- [x] Double check reference energy values (`energy_ref`), and reset if needed (including tolerances)

New tests will then be added in #3290, to cover the cases without periodic single box (which will require `tol_charge` to be set on a case-by-case basis, too).


Examples/Tests/galilean/analysis.py
```diff
@@ -0,0 +1,99 @@
...
+if re.search('geometry.dims\s*=\s*2', warpx_used_inputs):
+    dims = '2D'
+elif re.search('geometry.dims\s*=\s*RZ', warpx_used_inputs):
+    dims = 'RZ'
+elif re.search('geometry.dims\s*=\s*3', warpx_used_inputs):
+    dims = '3D'
+if re.search('psatd.current_correction\s*=\s*1', warpx_used_inputs):
+    current_correction = True
+if re.search('psatd.do_time_averaging\s*=\s*1', warpx_used_inputs):
+    time_averaging = True
```
EZoni, line 41: @dpgrote @ax3l Please feel free to suggest any better way to parse these input parameters from our output file `warpx_used_inputs` (note that the flags set here are used below to set reference energy values and tolerances independently).

Examples/Tests/galilean/analysis.py
```diff
@@ -0,0 +1,99 @@
...
+# RuntimeError: yt attempted to read outside the boundaries
+# of a non-periodic domain along dimension 0.
+if 'force_periodicity' in dir(ds): ds.force_periodicity()
+
+all_data = ds.covering_grid(level=0, left_edge=ds.domain_left_edge, dims=ds.domain_dimensions)
+Ex = all_data['boxlib', 'Ex'].squeeze().v
+Ey = all_data['boxlib', 'Ey'].squeeze().v
+Ez = all_data['boxlib', 'Ez'].squeeze().v
+
+# Set tolerance for numerical stability and charge conservation
```
EZoni:
```suggestion
# Set reference energy values, and tolerances for numerical stability and charge conservation
```

RemiLehe (approved):

PR #3326, opened 2022-08-18
Docs: Spack Desktop Environment

ax3l:
Add Spack environment files for developer desktop/laptop computers.

- [x] link in the docs
- [ ] test each env
  - [x] Linux: OpenMP w/ Ascent (install + run)
  - [ ] Linux: CUDA w/ Ascent
  - [ ] Linux: ROCm w/ Ascent
  - [x] macOS M1: OpenMP w/ Ascent (install + run)
  - [ ] macOS x86_64: OpenMP w/ Ascent

## Dependencies in Spack
- [x] https://github.com/spack/spack/pull/32285
- [x] https://github.com/spack/spack/pull/32284

## Spack Version

Tested with Spack `v0.13.2-14196-g7fc78b8b0f` (tag search off, essentially post 0.18.1 `develop`)

### DIFF
--- a/Tools/machines/desktop/spack-macos-openmp.yaml
+++ b/Tools/machines/desktop/spack-macos-openmp.yaml
@@ -1,0 +1,120 @@
+# This is a Spack environment file.
+#
+# This environment can be used to install all dependencies to build the manual
+# locally.
+#
+# Activating and installing this environment will provide all dependencies
+# that are needed for full-feature development.
+#   https://spack.readthedocs.io/en/latest/environments.html
+#
+# Inside the directory of this file
+#   spack env create warpx-openmp-dev spack-macos-openmp.yaml
+#   spack env activate warpx-openmp-dev
+#   spack install  # only needed the first time
+#
+spack:
+  specs:
+  - adios2 ~fortran
+  - ascent +adios2 +python ~fortran
+  - blaspp
+  - boost
+  - ccache
+  - cmake
+  - conduit ~fortran
+  - fftw
+  - hdf5 ~fortran
+  - lapackpp
+  - mpi
+  - llvm-openmp
+  - pkgconfig
+  - python
+  - py-cython
+  - py-h5py
+  - py-libensemble +nlopt
+  - py-mpi4py
+  - py-numpy
+  - py-pip
+  - py-setuptools
+  - py-wheel
+  - sensei +ascent ~catalyst +python
+# not yet ready for macOS prime time
+#   https://github.com/spack/spack/issues/32283
+#   https://github.com/spack/spack/pull/32285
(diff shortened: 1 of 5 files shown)


Tools/machines/desktop/spack-macos-openmp.yaml
```diff
@@ -0,0 +1,120 @@
...
+  compilers:
+  # macOS
+  # preparation: you first need to install xcode (app store) and gcc (homebrew) and gpg2 (homebrew)
+  - compiler:
+      spec: apple-clang@13.1.6
+      paths:
+        cc: /usr/bin/clang
+        cxx: /usr/bin/clang++
+        f77: /opt/homebrew/bin/gfortran
+        fc: /opt/homebrew/bin/gfortran
```
ax3l, line 85: We might even be able to put `null` here and compile without any Fortran compiler and thus no need to get `brew` first... test ongoing
```suggestion
        f77: null
        fc: null
```

Tools/machines/desktop/spack-macos-openmp.yaml
```diff
@@ -0,0 +1,120 @@
...
+      modules: []
+      environment: {}
+      extra_rpaths: []
+  - compiler:
+      spec: apple-clang@13.1.6
+      paths:
+        cc: /usr/bin/clang
+        cxx: /usr/bin/clang++
+        f77: /opt/homebrew/bin/gfortran
+        fc: /opt/homebrew/bin/gfortran
```
ax3l, line 98: See above:
```suggestion
        f77: null
        fc: null
```

Tools/machines/desktop/spack-macos-openmp.yaml
ax3l: OpenBLAS:
```
raise InstallError("OpenBLAS requires both C and Fortran compilers!")
```

RemiLehe (approved):

n01r:
Hey @ax3l, on my x86_64 MacBook the [spack-macos-openmp.yaml](https://github.com/ECP-WarpX/WarpX/blob/development/Tools/machines/desktop/spack-macos-openmp.yaml) does not work like that, yet. I am missing `/opt/homebrew/bin/gfortran` in my setup and `brew cask install gfortran` as well as `brew install gfortran --cask` do not work, anymore, since the `gfortran` cask is deprecated since the end of 2020.

There is a separate repository that exists but this would install it under a different path.
https://github.com/fxcoudert/gfortran-for-macOS/releases

It is also part of `brew install gcc` now. So one could also do that and then change the path accordingly.

ax3l:
Related follow-up: https://github.com/spack/spack/pull/32398

PR #3329, opened 2022-08-19
Major update of the Python/picmi documentation

dpgrote:
This PR is a major update and clean up of the documentation. This relies on PR#79 in picmistandard, which sets up the classes so that the doc strings of the picmistandard and WarpX implementations are combined.

- Use Napoleon Sphinx extension (to process numpydocs)
- Clean up existing docs in picmi.py
- Add the WarpX specific documentation for picmi classes
- In RST files, reference the pywarpx classes instead of the picmistandard classes
- Also, in python.rst, added section head for each class (for easy reference)
- Reformat the docs for the Checksum classes
- Fixes in the RST documentation files

This removes all warnings from Sphinx related to the python and RST files. Note that there a many warnings remaining related to problems in the documentation for C++ routines.


ax3l:
> This does require one big change - that during the generation of the html of the documentation, the pywarpx package needs to be installed. Is this something that is easily doable? This PR does not handle that.

That's do-able, if we install it in a way that it does not require to build WarpX or AMReX as well.

dpgrote:
@ax3l I note that PR #33310 is making the same assumption since the documentation is referencing 'pywarp.picmi`. If that one works, then this one will work also.

ax3l:
Awesome.

Added doc infrastructure, added Napoleon extension and fixes some warnings in doc strings :)

ax3l (approved):

PR #3330, opened 2022-08-20
shift BTD lab frame bounds by half a cell to prevent round off errors

RevathiJambunathan:
The BTD domain [prob_hi, prob_lo] in z are transformed to get the [lo,hi] in lab-frame.
We then take the back-transformed 'hi' in the z-direction) and set that it as zmax. 
Thus, we always has a z-slice in the lab-frame close to the edge of the cell, and susceptible to round off errors. 

Thanks to the discussion with @RemiLehe @atmyers @ax3l and a suggestion from @RemiLehe 
to shift the zmax by half a cell, as done in this PR, such that, the z-slices in the lab-frame are somewhere close to the cell-center, instead of on the edge of the cell.

This fixes the segfault in Issue #3305

### DIFF
--- a/Source/Diagnostics/BTDiagnostics.cpp
+++ b/Source/Diagnostics/BTDiagnostics.cpp
@@ -367,7 +367,7 @@ BTDiagnostics::InitializeBufferData ( int i_buffer , int lev)
     m_snapshot_domain_lab[i_buffer].setHi(m_moving_window_dir,
                                   zmax_buffer_lab + warpx.moving_window_v * m_t_lab[i_buffer]);
     // To prevent round off errors, moving the snapshot domain by half a cell so that all the slices
-    // lie close to the cell-centers in the lab-frame grid instead of on the edge of cell.    
+    // lie close to the cell-centers in the lab-frame grid instead of on the edge of cell.
     amrex::Real new_hi = m_snapshot_domain_lab[i_buffer].hi(m_moving_window_dir)
                        + 0.5_rt * dz_lab(warpx.getdt(lev), ref_ratio[m_moving_window_dir]);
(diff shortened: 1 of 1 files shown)


RemiLehe (approved):

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -367,6 +366,11 @@ BTDiagnostics::InitializeBufferData ( int i_buffer , int lev)
                                   zmin_buffer_lab + warpx.moving_window_v * m_t_lab[i_buffer]);
     m_snapshot_domain_lab[i_buffer].setHi(m_moving_window_dir,
                                   zmax_buffer_lab + warpx.moving_window_v * m_t_lab[i_buffer]);
+    // To prevent round off errors, moving the snapshot domain by half a cell so that all the slices
+    // lie close to the cell-centers in the lab-frame grid instead of on the edge of cell.
+    amrex::Real new_hi = m_snapshot_domain_lab[i_buffer].hi(m_moving_window_dir)
+                       + 0.5_rt * dz_lab(warpx.getdt(lev), ref_ratio[m_moving_window_dir]);
+    m_snapshot_domain_lab[i_buffer].setHi(m_moving_window_dir,new_hi);
     amrex::Real new_lo = m_snapshot_domain_lab[i_buffer].hi(m_moving_window_dir) -
                          num_z_cells_in_snapshot *
```
ax3l, line 375: wouldn't it be correct to move the low down by 0.5 cells to prevent similar errors (see PR description in #3440)? Wouldn't this then be:
```suggestion
                         (num_z_cells_in_snapshot +1) *
```
note the `+1`

ax3l: clarified: not needed

PR #3337, opened 2022-08-23
Spack Desktop: Simpler CUDA Visualization

ax3l:
Reduce the dependencies for in situ vis, avoiding to build DevilRay in Ascent, which takes very long right now.
We mostly use Ascent with VTK-m for now.

Spack prep.:
```console
# create new development environment
spack env create warpx-cuda-dev Tools/machines/desktop/spack-ubuntu-cuda.yaml
spack env activate warpx-cuda-dev

# installation
spack install
python3 -m pip install jupyter matplotlib numpy openpmd-api openpmd-viewer pandas scipy virtualenv yt
```

Compiled full-feature WarpX on Ubuntu 20.04 with:
```console
# work-around for Ascent/VTK-m/Vis: only needed on Pascal GPUs?
#export CUDAFLAGS="-Xptxas --disable-optimizer-constants"
# confuses GCC 9.4.0 now during link logic

cmake -S . -B build -DWarpX_COMPUTE=CUDA -DWarpX_ASCENT=ON -DWarpX_EB=ON -DWarpX_PSATD=ON -DWarpX_QED_TABLE_GEN=ON -DWarpX_LIB=ON
cmake --build build -j 4
cmake --build build --target pip_install -j 4
```

Run:
```console
cd build/bin
cp ../../Examples/Physics_applications/laser_acceleration/ascent_actions.yaml .
cp ../../Examples/Physics_applications/laser_acceleration/inputs_3d .
cp ../../Examples/Physics_applications/laser_acceleration/PICMI_inputs_3d.py .

# app w/ ascent
./warpx inputs_3d amr.n_cell=64 64 256 diag1.intervals=10 diag1.format=ascent

# Python PICMI w/o ascent
./PICMI_inputs_3d.py
```
- [ ] output

### DIFF
--- a/Tools/machines/desktop/spack-ubuntu-cuda.yaml
+++ b/Tools/machines/desktop/spack-ubuntu-cuda.yaml
@@ -14,7 +14,8 @@
 #
 spack:
   specs:
-  - adios2
+  # https://github.com/ornladios/ADIOS2/issues/3332
+  - adios2 ~cuda
   - ascent +adios2 +python ~fortran ~shared
   - blaspp
   - boost
--- a/Tools/machines/desktop/spack-ubuntu-cuda.yaml
+++ b/Tools/machines/desktop/spack-ubuntu-cuda.yaml
@@ -15,7 +15,7 @@
 spack:
   specs:
   - adios2
-  - ascent +adios2 +python ~fortran
+  - ascent +adios2 +python ~fortran ~shared
   - blaspp
   - boost
   - ccache
@@ -37,6 +37,7 @@ spack:
   - py-setuptools
   - py-wheel
   - sensei +ascent ~catalyst +python
+  - vtk-m ~shared
 # This always enables DevilRay, which builds too long on CUDA and we mainly use VTK-m
 #  - ecp-data-vis-sdk +adios2 +ascent +hdf5 +sensei
 # skipped to save time: 3D post-processing
--- a/Tools/machines/desktop/spack-ubuntu-cuda.yaml
+++ b/Tools/machines/desktop/spack-ubuntu-cuda.yaml
@@ -15,11 +15,12 @@
 spack:
   specs:
   - adios2
+  - ascent +adios2 +python ~fortran
   - blaspp
   - boost
   - ccache
   - cmake
-  - ecp-data-vis-sdk +adios2 +ascent +hdf5 +sensei
+  - conduit ~fortran
   - cuda
   - fftw
   - hdf5
@@ -35,6 +36,9 @@ spack:
   - py-pip
   - py-setuptools
   - py-wheel
+  - sensei +ascent ~catalyst +python
+# This always enables DevilRay, which builds too long on CUDA and we mainly use VTK-m
+#  - ecp-data-vis-sdk +adios2 +ascent +hdf5 +sensei
 # skipped to save time: 3D post-processing
 #  - paraview +adios2 +python3 +qt
 # skipped to save time, because they are faster installed via pip afterwards


Tools/machines/desktop/spack-ubuntu-cuda.yaml
```diff
@@ -15,11 +15,12 @@
 spack:
   specs:
   - adios2
+  - ascent +adios2 +python ~fortran
```
ax3l: @cyrush recommends to add static builds for DevilRay and VTK-m.


The moment I have Ascent on, I get linker errors of the kind:
```
nvlink error   : Entry function '_ZN5amrex13launch_globalILi256EZNS_14ParallelForRNGIZN25PhysicalParticleContainer9AddPlasmaEiNS_7RealBoxEEUliiiRKNS_12RandomEngineEE_EENSt9enable_ifIXsr5amrex19MaybeDeviceRunnableIT_vEE5valueEvE4typeERKNS_3BoxEOS9_EUlvE_EEvT0_' uses too much data for compiler-generated constants; please recompile with -Xptxas --disable-optimizer-constants
```
?

- SM: 6.1
- CUDA: 11.7.64
- GCC: 9.4.0

ax3l: @cyrush @kwryankrattiger with the static lib requirement, I wonder if one of the packages in the vis stack (Ascent, VTK-m, ...) forgets to forward CUDA RDC flags via CMake or so? update: see below, probably hardware limitation trigged by VTK-m...?

cyrush: Ascent handles all of this (it was a major focus a few years ago). We also had all of the spack pacages in balance, and this is how we test compiling and linking CUDA in CI. 

Doesn't mean things don't break occasionally :-)

With VTK-m 1.8 shared is now possible, but that won't be supported until Ascent 0.9.0.

kwryankrattiger: The SDK prefers shared libs because it is less weird to link.

ax3l: test: linker problem is _not_ resolved with `~static` as currently pushed - @cyrush any more ideas? :)

cyrush: Iterations at my P100 build are going painfully slow.  

I am trying an experiment with sm=60 (tesla) instead of sm=61 (consumer) I hope to see it finish today.

I think what is happening here is that the composition of ascent stuff and warpx cude stuff when linked together are spilling over a device limit for pascal consumer. 

We are setting this flag in Ascent for our use of VTK-m:

(ascent develop)
https://github.com/Alpine-DAV/ascent/blob/deef65e39f3b2792a40281439c4f614488349c0b/src/cmake/thirdparty/SetupVTKm.cmake#L45

(vtk-h alone, which supports ascent 0.8.0)
https://github.com/Alpine-DAV/vtk-h/blob/47852e2bcb23b0f87db4f6ad61e5867f682f232d/src/cmake/thirdparty/SetupVTKm.cmake#L87


Since we are compiling the vtk-m filters, that should take care of in Ascent. But again, maybe we are just hitting some limit when everything is pulled together.

ax3l: If I compile everything for SM 7.0, then things link. Maybe VTK-m really is too heavy to run on pre-Volta... What is the officially supported SM range for Ascent, VTK-m et al.? @cyrush 

I did a test with SM 6.1 before with `-Xptxas --disable-optimizer-constants` but forgot the outcome. Will repeat.
Ref.: https://docs.nvidia.com/cuda/pdf/PTX_Compiler_API_User_Guide.pdf
```
Disable use of optimizer constant bank.
```

kwryankrattiger: officially

```
VTKm CUDA Architecture Specifies what GPU architecture(s) to build CUDA for. The options include
native, fermi, kepler, maxwell, pascal, volta, and turing
```

ax3l: Thank you :) Ui, maybe a bit unlikely Fermi & Kepler are still working, not even the CUDA Toolkit supports those anymore :D

cyrush: Ahh - that's correct VTK-m will compile for all pascal variants. 

But when WarpX goes to link - I expect it will be providing a single sm case as part of its linking args.  Ascent may as well.  Many moving parts here - I am not sure what survives and what is gone as we travel down the rapids. 

My hypothesis is that the composed software might be too beefy for consumer pascal w/o `-Xptxas --disable-optimizer-constants`. We hit a hardware resource wall.  

I wish I had a good answer to verify specifically (cuobject dump like inspection, etc).  But even if we do have the exact answer of why, the prescription will still be to use `-Xptxas --disable-optimizer-constants`.

My pascal (sm 60 model) builds are painfully slow, which is adding to delay sorry :-(

cyrush: I was able to link when (and only when) I used `-Xptxas --disable-optimizer-constants.`

However I can't run:

```
STEP 1 starts ...
amrex::Abort::0::GPU last error detected in file /usr/WS1/cyrush/2022.08.warpx/WarpX/build/_deps/fetchedamrex-src/Src/Base/AMReX_GpuLaunchFunctsG.H line 1053: invalid device function !!!
SIGABRT
```
```
cat Backtrace.0 
=== If no file names and line numbers are shown below, one can run
            addr2line -Cpfie my_exefile my_line_address
    to convert `my_line_address` (e.g., 0x4a6b) into file name and line number.
    Or one can use amrex/Tools/Backtrace/parse_bt.py.

=== Please note that the line number reported by addr2line may not be accurate.
    One can use
            readelf -wl my_exefile | grep my_line_address'
    to find out the offset for that line.

 0: ./warpx() [0x8d80de]
    amrex::BLBackTrace::print_backtrace_info(_IO_FILE*) at ??:?

 1: ./warpx() [0x8d96cb]
    amrex::BLBackTrace::handler(int) at ??:?

 2: ./warpx() [0x4b5ce9]
    amrex::Gpu::ErrorCheck(char const*, int) at ??:?

 3: ./warpx() [0x6c881c]
    void FiniteDifferenceSolver::EvolveBCartesian<CartesianYeeAlgorithm>(std::array<std::unique_ptr<amrex::MultiFab, std::default_delete<amrex::MultiFab> >, 3ul>&, std::array<std::unique_ptr<amrex::MultiFab, std::default_delete<amrex::MultiFab> >, 3ul> const&, std::unique_ptr<amrex::MultiFab, std::default_delete<amrex::MultiFab> > const&, int, double) at ??:?

 4: ./warpx() [0x6c73da]
    FiniteDifferenceSolver::EvolveB(std::array<std::unique_ptr<amrex::MultiFab, std::default_delete<amrex::MultiFab> >, 3ul>&, std::array<std::unique_ptr<amrex::MultiFab, std::default_delete<amrex::MultiFab> >, 3ul> const&, std::unique_ptr<amrex::MultiFab, std::default_delete<amrex::MultiFab> > const&, std::array<std::unique_ptr<amrex::MultiFab, std::default_delete<amrex::MultiFab> >, 3ul> const&, std::array<std::unique_ptr<amrex::MultiFab, std::default_delete<amrex::MultiFab> >, 3ul> const&, std::array<std::unique_ptr<amrex::MultiFab, std::default_delete<amrex::MultiFab> >, 3ul>&, std::array<std::unique_ptr<amrex::MultiFab, std::default_delete<amrex::MultiFab> >, 3ul>&, std::array<std::unique_ptr<amrex::iMultiFab, std::default_delete<amrex::iMultiFab> >, 3ul>&, std::array<std::unique_ptr<amrex::LayoutData<FaceInfoBox>, std::default_delete<amrex::LayoutData<FaceInfoBox> > >, 3ul>&, int, double) at ??:?

 5: ./warpx() [0x6b71cf]
    WarpX::EvolveB(int, PatchType, double, DtType) at ??:?

 6: ./warpx() [0x6b7372]
    WarpX::EvolveB(int, double, DtType) at ??:?

 7: ./warpx() [0x6b7406]
    WarpX::EvolveB(double, DtType) at ??:?

 8: ./warpx() [0x69fece]
    WarpX::OneStep_nosub(double) at ??:?

 9: ./warpx() [0x6a26b8]
    WarpX::Evolve(int) at ??:?

10: ./warpx() [0x48921d]
    main at ??:?

11: /lib64/libc.so.6(__libc_start_main+0xf5) [0x2aaabf588555]
    __libc_start_main
../csu/libc-start.c:300

12: ./warpx() [0x49979e]
    _start at ??:?
```

I think this is before we get to Ascent.

cyrush: But it could still be sad about something in Ascent.

ax3l: Yep, maybe we cannot handle that flag well... `invalid device function`, curious. could be the last kernel called before `FiniteDifferenceSolver::EvolveBCartesian<CartesianYeeAlgorithm>` is called and checks the error. You could set the env var `CUDA_LAUNCH_BLOCKING=1` to see the actually failing call.

RemiLehe (approved):

PR #3338, opened 2022-08-23
order-independent rho_<species> in RZ + openPMD

RTSandberg:
This fixes some order dependency in the rho per species diagnostic when using RZ + openPMD.  

In [this input deck](https://github.com/ECP-WarpX/WarpX/files/9411099/inputs.txt), three species (e1, e2, e3) are spatially ordered and ordered in the declaration:

``particles.species_names = e1 e2 e3``
``e1.zmin =  10.e-6``
``e1.zmax =  20.e-6``
``e2.zmin =  40.e-6``
``e2.zmax =  50.e-6``
``e3.zmin =  60.e-6``
``e3.zmax =  70.e-6``
but the diagnostic order is mixed up:
``diag1.fields_to_plot = rho_e2 rho_e3 rho_e1``

Previously, their density plot had the species mixed up
Screen Shot 2022-08-23 at 4 29 53 PM
but with this fix the densities are read correctly, regardless of the differing order between species and diagnostic declarations:
Screen Shot 2022-08-23 at 4 37 34 PM

### DIFF
--- a/Examples/Tests/openpmd_rz/analysis_openpmd_rz.py
+++ b/Examples/Tests/openpmd_rz/analysis_openpmd_rz.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import openpmd_api as io
+import numpy as np
 
 series = io.Series("LaserAccelerationRZ_opmd_plt/openpmd_%T.h5", io.Access.read_only)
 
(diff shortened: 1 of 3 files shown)


EZoni (commented):
@RTSandberg Thank you for the fix! Just reporting here what we discussed offline: it would be great if we could test this automatically by adding some simple check in one of the existing CI tests (for example, by checking the sign of the charge density, if there is a test with one positive species and one negative species, and where the order of the species is not consistent among the various input parameters).

EZoni (approved):

PR #3341, opened 2022-08-26
CI: GH Action Key Update

ax3l:
It looks like GH Actions slightly changed the syntax for cache keys.
This fixes the resulting breakage.

- [x] spot if there is an issue (e.g. https://github.com/kngott/WarpX/actions/runs/2929984658/workflow#L100)
Screenshot from 2022-08-26 11-07-43

- [ ] fix the issue

### DIFF
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -1,7 +1,6 @@
 name: 🐧 CUDA
 
 on: [push, pull_request]
-# test
 
 concurrency:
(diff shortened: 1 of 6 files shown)


kngott:
Did not fix my error. 🤷

ax3l:
Probably some local setting with your repo then?

ax3l:
Does your branch have a `cuda.yaml` file?

PR #3342, opened 2022-08-26
Python: 3.7+

ax3l:
Python 3.6 is now end-of-life. Bump support to 3.7+.

Also updated docs & requirements to use yt 4.0.1+ only (as we don't test older releases anymore).


dpgrote:
I also see the file Tools/Ascent/ascent_replay_warpx.ipynb which references version 3.6.3.

ax3l:
I think the problem is just the old, Ubuntu 20.04-default Python 3.6 on the CUDA CI runner.

dpgrote (approved):
This looks good, thanks!

Note, maybe a comment to the PR that this also upgrades the version of yt.

PR #3346, opened 2022-08-29
Fix few "magic constants" & undefined variables found with clang-tidy

lucafedeli88:
This PR fixes few "magic constants" & undefined variables found using the `clang-tidy` tool

### DIFF
--- a/Source/Diagnostics/BTDiagnostics.cpp
+++ b/Source/Diagnostics/BTDiagnostics.cpp
@@ -43,6 +43,11 @@
 
 using namespace amrex::literals;
 
+namespace
+{
+    const int permission_flag_rwxrxrx = 0755;
+}
+
 BTDiagnostics::BTDiagnostics (int i, std::string name)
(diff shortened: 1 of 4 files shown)


ax3l (approved):
Thanks a lot! :)

cc @RevathiJambunathan fyi, I will go ahead and merge this :)

ax3l:
@lucafedeli88 which clang-tidy rules did you use to find those?

ax3l:
cc @RevathiJambunathan fyi, I will go ahead and merge this :)

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -43,6 +43,11 @@
 
 using namespace amrex::literals;
 
+namespace
+{
+    const int permission_flag_rwxrxrx = 0755;
```
ax3l, line 48: @PhilMiller mentioned that those can be `constexpr`

PR #3349, opened 2022-08-29
Perlmutter: Work-Around CUDA-Aware MPI & Slurm

ax3l:
There are known HPE bugs with cgroups on Perlmutter that can blow up simulations (segfault) with CUDA-aware MPI.

We avoid the respective Slurm options now and just manually control the exposed GPUs per MPI rank.

### DIFF
--- a/Tools/machines/perlmutter-nersc/perlmutter.sbatch
+++ b/Tools/machines/perlmutter-nersc/perlmutter.sbatch
@@ -16,6 +16,7 @@
 #SBATCH -C gpu
 #SBATCH -c 32
 #SBATCH --ntasks-per-node=4
+#SBATCH --gpus-per-node=4
 #SBATCH -o WarpX.o%j
 #SBATCH -e WarpX.e%j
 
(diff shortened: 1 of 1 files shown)


Tools/machines/perlmutter-nersc/perlmutter.sbatch
```diff
@@ -16,8 +16,6 @@
 #SBATCH -C gpu
 #SBATCH -c 32
 #SBATCH --ntasks-per-node=4
-#SBATCH --gpus-per-task=1
-#SBATCH --gpu-bind=single:1
 #SBATCH -o WarpX.o%j
```
kngott, line 20: Need `-#SBATCH --gpus-per-node=4`, to make sure you're getting all 4 nodes to begin with.

ax3l: ok, though that might be default
```suggestion
#SBATCH --gpus-per-node=4
#SBATCH -o WarpX.o%j
```

kngott (approved):

ax3l:
Regression reported in https://github.com/ECP-WarpX/WarpX/issues/3356#issuecomment-1238724908

PR #3355, opened 2022-08-31
Remove some magic numbers

lucafedeli88:
This PR removes almost all "magic numbers" from the following files (issues found with the `readability-magic-numbers` option of the `clang-tidy` tool):
```
BackgroundStopping.cpp
GetExternalFields.cpp
QuantumSyncEngineWrapper.cpp
BreitWheelerEngineWrapper.cpp
ParticleBoundaryBuffer.cpp
LaserProfileGaussian.cpp
LaserProfileHarris.cpp
ParticleExtrema.cpp
```

### DIFF
--- a/Source/Diagnostics/ReducedDiags/ParticleExtrema.cpp
+++ b/Source/Diagnostics/ReducedDiags/ParticleExtrema.cpp
@@ -82,16 +82,41 @@ ParticleExtrema::ParticleExtrema (std::string rd_name)
         // get WarpXParticleContainer class object
         auto & myspc = mypc.GetParticleContainer(i_s);
 
-        if (myspc.DoQED())
-        {
-            // resize data array for QED species
-            const int num_quantities = 18;
-            m_data.resize(num_quantities,0.0);
-        } else
-        {
-            // resize data array for regular species
-            const int num_quantities = 16;
-            m_data.resize(num_quantities,0.0);
+        auto all_diag_names = std::vector<std::string> {};
+        auto add_diag = [&,c=0] (
+            const std::string& name, const std::string& header) mutable {
+            m_headers_indices[name] = aux_header_index{header, c++};
+            all_diag_names.push_back(name);
+        };
+
+        add_diag("xmin", "xmin(m)");
+        add_diag("xmax", "xmax(m)");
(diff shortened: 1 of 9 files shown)


Source/Diagnostics/ReducedDiags/ParticleExtrema.cpp
```diff
@@ -487,27 +487,35 @@ void ParticleExtrema::ComputeDiags (int step)
...
+        constexpr auto
+            idx_xmin   = 0  , idx_xmax   = 1   , idx_ymin   = 2 , idx_ymax   = 3,
+            idx_zmin   = 4  , idx_zmax   = 5   , idx_uxminm = 6 , idx_uxmaxm = 7,
+            idx_uyminm = 8  , idx_uymaxm = 9   , idx_uzminm = 10, idx_uzmaxm = 11,
+            idx_gmin   = 12 , idx_gmax   = 13  , idx_wmin   = 14, idx_wmax   = 15;
+        [[maybe_unused]] constexpr auto
+            idx_chimin_f = 16, idx_chimax_f = 17;
+
+        m_data[idx_xmin]  = xmin;
+        m_data[idx_xmax]  = xmax;
```
ax3l: I think this index change might take it too far for readability comparing to the left... what do you think?

It would be worth it if we use the index again at another location, but since we only use it here it looks a bit too confusing to me.

Source/Particles/ParticleBoundaryBuffer.cpp
```diff
@@ -73,24 +73,41 @@ ParticleBoundaryBuffer::ParticleBoundaryBuffer ()
         m_do_boundary_buffer[i].resize(numSpecies(), 0);
     }
 
+#if defined(WARPX_DIM_1D_Z)
+    constexpr auto idx_zlo = 0;
```
ax3l, line 77: ok here, since we use each index twice.

Source/Diagnostics/ReducedDiags/ParticleExtrema.cpp
lucafedeli88: You are right... is it better now ? otherwise I will revert to the initial state...

ax3l (approved):
Looks good, thank you! :)

Issue #3335, opened 2022-08-23
Initialize particles only where density is non-zero


n01r:
Currently, particles are initialized everywhere in the box inside the bounds of `zmin, zmax, xmin, ...` if they are set.
Otherwise, these bounds are expanded to the full box size, which can lead to running out of memory quickly when very dense, spatially limited targets (e.g. foils, spheres, etc.) are initialized.

Following discussions from 2022/08/23 this is identified as a potential topic for the September hackathon.

Approaches for solving this: 
- evaluate the density function twice and create an `iMultiFab` mask for non-zero density (preferred)
- Find a global bounding box for particles
- Find a local (per box) bounding box for particles

ax3l:
tentatively assigning @atmyers for this, potentially hackathon; generally high priority for all non-trivial (non-uniform) simulations :)

PR #1892, opened 2021-04-13
Added injection of a thermal flux from a plane

dpgrote:
This implements the injection of a flux of particles from a planar surface. The surface can only be tangent to one of the grid axes. Here is an example input setting up the species `electrons` that will be injected from the plane at xmin=0.

electrons.species_type = electron
electrons.injection_style = NFluxPerCell
electrons.surface_flux_pos = 0.
electrons.flux_normal_axis = x
electrons.flux_direction = +1
electrons.num_particles_per_cell = 1.5
electrons.profile = constant
electrons.density = 1.e8 #  #/m^2/s
electrons.momentum_distribution_type = gaussianflux
electrons.ux_th = 0.01
electrons.uy_th = 0.01
electrons.uz_th = 0.01

The profile can be any of the standard profiles with the density given as #/m^2/s. The momentum distribution type `gaussianflux` is Gaussian in the plane and v*Gaussian normal to it.

PR #3359, opened 2022-09-01
Initialize user defined runtime attributes in AddPlasmaFlux

NeilZaim:
With this PR, all runtime attributes should be properly initialized in the 3 main routines that inject particles (`AddPlasma`, `AddPlasmaFlux` and `AddNParticles`).

The code here is a copy-paste of the code in `AddPlasma`, following the logic of how `AddPlasmaFlux` is implemented (see discussion in #1892). The code duplication between the 2 functions is not trivially removed but also makes the code harder to maintain (as we can see here), so we could consider adding common routines that are shared by the 2 functions. On the other hand we were also considering to refactor the `AddPlasma` function (#3335) so maybe we should wait until this is done to remove the code duplication.

### DIFF
--- a/Source/Particles/PhysicalParticleContainer.cpp
+++ b/Source/Particles/PhysicalParticleContainer.cpp
@@ -1393,8 +1393,10 @@ PhysicalParticleContainer::AddPlasmaFlux (amrex::Real dt)
     InjectorPosition* inj_pos = plasma_injector->getInjectorPosition();
     InjectorDensity*  inj_rho = plasma_injector->getInjectorDensity();
     InjectorMomentum* inj_mom = plasma_injector->getInjectorMomentum();
-    Real density_min = plasma_injector->density_min;
-    Real density_max = plasma_injector->density_max;
+    const amrex::Real density_min = plasma_injector->density_min;
+    const amrex::Real density_max = plasma_injector->density_max;
+    constexpr int level_zero = 0;
+    const amrex::Real t = WarpX::GetInstance().gett_new(level_zero);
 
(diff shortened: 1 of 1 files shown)


Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1560,6 +1562,26 @@ PhysicalParticleContainer::AddPlasmaFlux (amrex::Real dt)
             pa[ia] = soa.GetRealData(ia).data() + old_size;
         }
 
+        // user-defined integer and real attributes
+        const int n_user_int_attribs = m_user_int_attribs.size();
+        const int n_user_real_attribs = m_user_real_attribs.size();
+        amrex::Gpu::DeviceVector<int*> pa_user_int(n_user_int_attribs);
+        amrex::Gpu::DeviceVector<ParticleReal*> pa_user_real(n_user_real_attribs);
+        amrex::Gpu::DeviceVector< amrex::ParserExecutor<7> > user_int_attrib_parserexec(n_user_int_attribs);
```
dpgrote, line 1570: Could the ParserExecutor object be precalculated, perhaps right after user input of the attribute parser expressions? That would save a little bit of code duplication since it could be used in the three particle creation routines.

dpgrote (approved):
This looks good, thanks! I have the one comment, though perhaps for a later PR.

RevathiJambunathan (approved):
Thank you for this PR, @NeilZaim !
Looks good to me.

ax3l (approved):

PR #3363, opened 2022-09-01
CI: Add 2D/3D Langmuir Tests w/ Multi-J PSATD

EZoni:
Add initial 2D and 3D Langmuir tests with **multi-J PSATD**, and remove old 2D tests. These were LWFA tests (not even that sensitive to NCI, as far as I remember) performing only regression analysis with the checksum benchmarks, while the Langmuir tests also compare the simulation results with theoretical predictions.


oshapoval:
Thank you very much for this PR, @EZoni! I agree that having Langmuir test for multiJ PSATD is a must have. Changes look good to me!
Small observation: in 3D nodal multi-J PSATD case relative error is very close to the tolerance of 0.05. 

Screen Shot 2022-09-06 at 10 34 54 AM
Screen Shot 2022-09-06 at 10 34 37 AM

RemiLehe (approved):

PR #3364, opened 2022-09-02
Docs: abort_on_unused_inputs

n01r:
Add documentation of the AMReX option `abort_on_unused_inputs` to WarpX docs.


Docs/source/usage/parameters.rst
```diff
@@ -139,6 +139,10 @@ Overall simulation parameters
     Note that even with this set to ``1`` WarpX will not catch all out-of-memory events yet when operating close to maximum device memory.
     `Please also see the documentation in AMReX <https://amrex-codes.github.io/amrex/docs_html/GPU.html#inputs-parameters>`_.
 
+* ``amrex.abort_on_unused_inputs`` (``0`` or ``1``; default is ``1`` for true)
```
ax3l: The default is false. We only set it to true in CI, as we want to catch all warnings:
```suggestion
* ``amrex.abort_on_unused_inputs`` (``0`` or ``1``; default is ``0`` for false)
```

ax3l (approved):

PR #3366, opened 2022-09-02
Frontier/Crusher: rocFFT Cache Control

ax3l:
rocFFT in ROCm 5.1+ tries to [write to a cache](https://rocfft.readthedocs.io/en/latest/library.html#runtime-compilation) in the home area by default. This does not scale and caused timeouts for PSATD runs.

### DIFF
--- a/Docs/source/install/hpc/crusher.rst
+++ b/Docs/source/install/hpc/crusher.rst
@@ -106,3 +106,13 @@ Known System Issues
    .. code-block:: bash
 
       export FI_MR_CACHE_MAX_COUNT=0  # libfabric disable caching
+
+.. warning::
+
+   Sep 2nd, 2022 (OLCFDEV-1079):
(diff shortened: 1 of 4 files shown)


atmyers (approved):

ax3l:
cc @WeiqunZhang question for AMD-Matt, because the HPE recommended `/dev/null` might actually be writable

EZoni:
Comment from AMD-Matt reported on Slack by @WeiqunZhang, regarding setting `ROCFFT_RTC_CACHE_PATH=/dev/null`:
> I confirmed that setting the environment variable to `/dev/null` is OK; and that when you do so, an in-memory cache will be created. This means that a particular length will only be runtime-compiled once per application run. Instead of `/dev/null`, you can also use `:memory:`.  This might make the behaviour more obvious to others.

EZoni (approved):

PR #3367, opened 2022-09-03
BTD diagnostics specified by intervals

RTSandberg:
Specify diagnostics via intervals of the specified step size ``dt_snapshots_lab`` or ``dz_snapshots_lab`` rather than ``num_snapshots_lab``

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -2167,7 +2167,7 @@ BackTransformed Diagnostics (with support for Plotfile/openPMD output)
     Using the `Intervals parser`_ syntax, this string defines the lab frame times at which data is dumped,
     given as multiples of the step size ``dt_snapshots_lab`` or ``dz_snapshots_lab`` described below.
     Example: ``btdiag1.intervals = 10:11,20:24:2`` and ``btdiag1.dt_snapshots_lab = 1.e-12``
-    indicate to dump at lab times ``1e-11``, ``1.1e-11``, ``2e-11``, ``2.2e-11``, and ``2.2e-11`` seconds.
+    indicate to dump at lab times ``1e-11``, ``1.1e-11``, ``2e-11``, ``2.2e-11``, and ``2.4e-11`` seconds.
     Note that the stop interval, the second number in the slice, must always be specified.
     Only this option or ``num_snapshots_lab`` should be specified;
     a run-time error occurs if the user attempts to set both ``num_snapshots_lab`` and ``intervals``.
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -2156,9 +2156,13 @@ BackTransformed Diagnostics (with support for Plotfile/openPMD output)
 
 ``BackTransformed`` diag type are used when running a simulation in a boosted frame, to reconstruct output data to the lab frame. This option can be set using ``<diag_name>.diag_type = BackTransformed``. Note that this diagnostic is not currently supported for RZ.  Additional options for this diagnostic include:
 
-* ``<diag_name>.num_snapshots_lab`` (`integer`)
-    Only used when ``<diag_name>.diag_type`` is ``BackTransformed``.
-    The number of lab-frame snapshots that will be written.
+* ``<diag_name>.intervals`` (`string`)
+    Using the `Intervals parser`_ syntax, this string defines the lab frame times at which data is dumped, 
+    given as multiples of the step size ``dt_snapshots_lab`` or ``dz_snapshots_lab`` described below.
+    .. Use a negative number or 0 to disable data dumping.
+    example: ``btdiag1.intervals = 10:11,20:24:2`` and ``btdiag1.dt_snapshots_lab = 1.e-12``
+    indicate to dump at lab times ``1e-11``, ``1.1e-11``, ``2e-11``, ``2.2e-11``, and ``2.2e-11`` seconds.
+    Note that the stop interval, the second number in the slice, must always be specified.
 
 * ``<diag_name>.dt_snapshots_lab`` (`float`, in seconds)
     Only used when ``<diag_name>.diag_type`` is ``BackTransformed``.
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -2156,13 +2156,21 @@ BackTransformed Diagnostics (with support for Plotfile/openPMD output)
 
 ``BackTransformed`` diag type are used when running a simulation in a boosted frame, to reconstruct output data to the lab frame. This option can be set using ``<diag_name>.diag_type = BackTransformed``. Note that this diagnostic is not currently supported for RZ.  Additional options for this diagnostic include:
 
+* ``<diag_name>.num_snapshots_lab`` (`integer`)
+    Only used when ``<diag_name>.diag_type`` is ``BackTransformed``.
+    The time interval in between the lab-frame snapshots (where this time interval is expressed in the laboratory frame).
+    Only this option or ``intervals`` should be specified; 
+    a run-time error occurs if the user attempts to set both ``num_snapshots_lab`` and ``intervals``.
+
 * ``<diag_name>.intervals`` (`string`)
+    Only used when ``<diag_name>.diag_type`` is ``BackTransformed``.
     Using the `Intervals parser`_ syntax, this string defines the lab frame times at which data is dumped,
     given as multiples of the step size ``dt_snapshots_lab`` or ``dz_snapshots_lab`` described below.
-    .. Use a negative number or 0 to disable data dumping.
-    example: ``btdiag1.intervals = 10:11,20:24:2`` and ``btdiag1.dt_snapshots_lab = 1.e-12``
+    Example: ``btdiag1.intervals = 10:11,20:24:2`` and ``btdiag1.dt_snapshots_lab = 1.e-12``
     indicate to dump at lab times ``1e-11``, ``1.1e-11``, ``2e-11``, ``2.2e-11``, and ``2.2e-11`` seconds.
     Note that the stop interval, the second number in the slice, must always be specified.
+    Only this option or ``num_snapshots_lab`` should be specified; 
+    a run-time error occurs if the user attempts to set both ``num_snapshots_lab`` and ``intervals``.
 
 * ``<diag_name>.dt_snapshots_lab`` (`float`, in seconds)
     Only used when ``<diag_name>.diag_type`` is ``BackTransformed``.
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -2159,7 +2159,7 @@ BackTransformed Diagnostics (with support for Plotfile/openPMD output)
 * ``<diag_name>.num_snapshots_lab`` (`integer`)
     Only used when ``<diag_name>.diag_type`` is ``BackTransformed``.
     The time interval in between the lab-frame snapshots (where this time interval is expressed in the laboratory frame).
-    Only this option or ``intervals`` should be specified; 
+    Only this option or ``intervals`` should be specified;
     a run-time error occurs if the user attempts to set both ``num_snapshots_lab`` and ``intervals``.
 
 * ``<diag_name>.intervals`` (`string`)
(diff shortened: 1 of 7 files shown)


Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -159,8 +159,18 @@ BTDiagnostics::ReadParameters ()
     WARPX_ALWAYS_ASSERT_WITH_MESSAGE(m_do_back_transformed_fields, " fields must be turned on for the new back-transformed diagnostics");
     if (m_do_back_transformed_fields == false) m_varnames.clear();
 
-    getWithParser(pp_diag_name, "num_snapshots_lab", m_num_snapshots_lab);
-    m_num_buffers = m_num_snapshots_lab;
+
+    std::vector<std::string> intervals_string_vec = {"0"};
```
ax3l, line 165: Please include `<string>` in this file

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -159,8 +159,18 @@ BTDiagnostics::ReadParameters ()
...
+    bool num_snapshots_specified = queryWithParser(pp_diag_name, "num_snapshots_lab", m_num_snapshots_lab);
+    bool intervals_specified = pp_diag_name.queryarr("intervals", intervals_string_vec);
```
ax3l:
```suggestion
    bool const num_snapshots_specified = queryWithParser(pp_diag_name, "num_snapshots_lab", m_num_snapshots_lab);
    bool const intervals_specified = pp_diag_name.queryarr("intervals", intervals_string_vec);
```

Source/Utils/IntervalsParser.H
```diff
@@ -148,4 +155,53 @@ private:
...
+    * has a strictly positive period.
+    */
+    bool isActivated () const;
+
+private:
+    std::vector<int> m_btd_iterations;
+    std::vector<SliceParser> m_slices;
+    std::vector<int> m_slice_starting_i_buffer;
+    int m_n_snapshots;
+    std::string m_separator = ",";
```
ax3l:
```suggestion
    std::string const m_separator = ",";
```

Source/Utils/IntervalsParser.cpp
```diff
@@ -116,3 +122,74 @@ int IntervalsParser::localPeriod (const int n) const
 }
 
 bool IntervalsParser::isActivated () const {return m_activated;}
+
+BTDIntervalsParser::BTDIntervalsParser (const std::vector<std::string>& instr_vec)
+{
+    std::string inconcatenated;
+    for (const auto& instr_element : instr_vec) inconcatenated +=instr_element;
+
+    auto insplit = WarpXUtilStr::split<std::vector<std::string>>(inconcatenated, m_separator);
```
ax3l:
```suggestion
    auto const insplit = WarpXUtilStr::split<std::vector<std::string>>(inconcatenated, m_separator);
```

ax3l (commented):
This looks great to me. Tiny comments inline, then ready to go.

Source/Diagnostics/BTDiagnostics.H
```diff
@@ -9,6 +9,7 @@
 
 #include "Diagnostics.H"
 #include "Diagnostics/ComputeDiagFunctors/ComputeDiagFunctor.H"
+#include "Utils/IntervalsParser.H"
 #include "Utils/WarpXConst.H"
```
RTSandberg, line 12:
```suggestion
#include <string>
```

Source/Diagnostics/BTDiagnostics.cpp
RTSandberg: ok, will do

Source/Diagnostics/BTDiagnostics.H
```diff
@@ -9,7 +9,8 @@
 
 #include "Diagnostics.H"
 #include "Diagnostics/ComputeDiagFunctors/ComputeDiagFunctor.H"
-#include "Utils/WarpXConst.H"
+#include "Utils/IntervalsParser.H"
+#include <string>
```
ax3l: Is `#include "Utils/WarpXConst.H"` not used in this file anymore?

Please include `<string>` last: https://warpx.readthedocs.io/en/latest/developers/repo_organization.html#c-includes

RTSandberg: the deletion is an accident... still learning how to effectively do in-line commits from github

Docs/source/usage/parameters.rst
```diff
@@ -2147,7 +2147,19 @@ BackTransformed Diagnostics (with support for Plotfile/openPMD output)
 
 * ``<diag_name>.num_snapshots_lab`` (`integer`)
     Only used when ``<diag_name>.diag_type`` is ``BackTransformed``.
-    The number of lab-frame snapshots that will be written.
+    The time interval in between the lab-frame snapshots (where this time interval is expressed in the laboratory frame).
```
EZoni: I wonder if the documentation of this input parameter is clear. It is called `num_snapshots_lab`, but the docs now describe it as a "time interval" (in the laboratory frame). It is still a number, though, right?

Docs/source/usage/parameters.rst
```diff
@@ -2147,7 +2147,19 @@ BackTransformed Diagnostics (with support for Plotfile/openPMD output)
...
+    Only this option or ``intervals`` should be specified;
+    a run-time error occurs if the user attempts to set both ``num_snapshots_lab`` and ``intervals``.
+
+* ``<diag_name>.intervals`` (`string`)
+    Only used when ``<diag_name>.diag_type`` is ``BackTransformed``.
+    Using the `Intervals parser`_ syntax, this string defines the lab frame times at which data is dumped,
+    given as multiples of the step size ``dt_snapshots_lab`` or ``dz_snapshots_lab`` described below.
+    Example: ``btdiag1.intervals = 10:11,20:24:2`` and ``btdiag1.dt_snapshots_lab = 1.e-12``
+    indicate to dump at lab times ``1e-11``, ``1.1e-11``, ``2e-11``, ``2.2e-11``, and ``2.2e-11`` seconds.
```
```suggestion
    indicate to dump at lab times ``1e-11``, ``1.1e-11``, ``2e-11``, ``2.2e-11``, and ``2.4e-11`` seconds.
```

Source/Diagnostics/BTDiagnostics.H
```diff
@@ -37,6 +38,8 @@ private:
     bool m_plot_raw_fields_guards = false;
     /** Read relevant parameters for BTD */
     void ReadParameters ();
+    /** Determines timesteps at which full diagnostics are written to file */
```
EZoni: Is it "BTD" instead of "full" here?
```suggestion
    /** Determines timesteps at which BTD diagnostics are written to file */
```

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -159,8 +159,18 @@ BTDiagnostics::ReadParameters ()
...
-    getWithParser(pp_diag_name, "num_snapshots_lab", m_num_snapshots_lab);
-    m_num_buffers = m_num_snapshots_lab;
+
+    std::vector<std::string> intervals_string_vec = {"0"};
+    bool const num_snapshots_specified = queryWithParser(pp_diag_name, "num_snapshots_lab", m_num_snapshots_lab);
+    bool const intervals_specified = pp_diag_name.queryarr("intervals", intervals_string_vec);
+    if (num_snapshots_specified)
+    {
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(!(num_snapshots_specified && intervals_specified),
+            "For back-transformed diagnostics, user should specify either num_snapshots_lab or intervals, not both");
```
EZoni: Isn't it enough to check that `intervals_specified == false` here, since we are already inside an `if (num_snapshots_specified)` block?
```suggestion
        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(!intervals_specified,
            "For back-transformed diagnostics, user should specify either num_snapshots_lab or intervals, not both");
```
I might be wrong, it's always difficult to get this logic right.

Source/Utils/IntervalsParser.cpp
```diff
@@ -116,3 +122,74 @@ int IntervalsParser::localPeriod (const int n) const
 }
 
 bool IntervalsParser::isActivated () const {return m_activated;}
+
+BTDIntervalsParser::BTDIntervalsParser (const std::vector<std::string>& instr_vec)
+{
+    std::string inconcatenated;
```
EZoni, line 128: More of a C++ question, I guess: is `inconcatenated` always going to be initialized as an empty string `''` by default (which has to be the case, due to the upcoming `+=` assignements)?

Source/Utils/IntervalsParser.cpp
```diff
@@ -116,3 +122,74 @@ int IntervalsParser::localPeriod (const int n) const
 }
 
 bool IntervalsParser::isActivated () const {return m_activated;}
+
+BTDIntervalsParser::BTDIntervalsParser (const std::vector<std::string>& instr_vec)
```
EZoni, line 126: Could you maybe add a few brief inline comments in the body of this method, to describe what each part of the implementation does? Just to make it a bit more readable to someone who hasn't written the code.

EZoni (changes requested):
Thank you for this PR, Ryan! This is a much needed feature. I left a few comments and questions on my side.

Docs/source/usage/parameters.rst
RTSandberg: I agree; I'm reverting to the original description: "The number of lab-frame snapshots that will be written."

Source/Utils/IntervalsParser.cpp
RTSandberg: I assume so because that was copied from the ``IntervalsParser`` constructor but is not code I thought about

Source/Diagnostics/BTDiagnostics.cpp
RTSandberg: You are right.  I thought it might be easier to see the logic requires both statements, but your comment points out that being in nested if statements with the two statements so close in the code makes this redunancy unnecessary

Source/Utils/IntervalsParser.cpp
RTSandberg: yes, I remember wondering at the time if I would need this, let alone someone else.  Now it is somewhat annotated

EZoni (approved):
Thank you for responding to my comments, Ryan. The PR looks good to me.

@RemiLehe @RevathiJambunathan 
Since this is a useful feature, I will merge the PR by the end of today, unless you have urgent requests to be addressed. Further minor improvements can be done in a follow-up PR.

EZoni:
@RTSandberg 
The CI test `BTD_ReducedSliceDiag`, recently extended in #3371, currently fails with this error message:
```
  File "./analysis_3Dbacktransformed_diag.py", line 70, in <module>
    assert(np.allclose(Ez_plotfile, Ez_openpmd, rtol=rtol, atol=atol))
AssertionError
```
It might be trying to compare iteration 4 of the plotfile data with iteration 3 of the openPMD data.

Update: fixed by 069d741.

EZoni (approved):

Source/Utils/IntervalsParser.cpp
```diff
@@ -116,3 +122,94 @@ int IntervalsParser::localPeriod (const int n) const
...
+            {
+                // find where iteration ii should go in m_btd_iterations
+                while (ii > m_btd_iterations[btd_iter_ind] && btd_iter_ind < static_cast<int>(m_btd_iterations.size()))
+                {
+                    btd_iter_ind++;
+                }
+                if (ii != m_btd_iterations[btd_iter_ind])
+                {
+                    m_btd_iterations.insert(m_btd_iterations.begin() + btd_iter_ind, ii);
+                }
```
ax3l, line 199: Looking at this right now with @WeiqunZhang @RevathiJambunathan :
- `m_btd_iterations[...]` should be `m_btd_iterations.at(...)` because it can catch at runtime out-of-bounds access, and the loop here is cheap
- there are out-of-bounds accesses to `m_btd_iterations` in both the `while` condition and the `if` later on

Let's find a safer formulation.

WeiqunZhang: There are two issues. `btd_iter_ind < static_cast<int>(m_btd_iterations.size())` should be checked before `m_btd_iterations[btd_iter_ind]` is used. The second issue is `btd_iter_ind` could be as big as the size of the vector because of `++` in the loop body. So the next if test `if (ii != m_btd_iterations[btd_iter_ind])` is not safe.

ax3l: Fix coming in #4216

PR #3368, opened 2022-09-06
Shared memory charge and current deposition

atmyers:
This adds support for shared memory current deposition to WarpX. This work was mostly done by @kaplannp, building on work by @PhilMiller and myself. Currently it only does the "direct" algorithm, not Esirkepov.

The performance depends on the runtime parameter for the shared memory tile size. Using 4 GPUs on Crusher,  the time spent in current deposition is: 

```
2x2x2     4.84s
2x2x4     3.79s
2x2x8     3.525s 
4x4x4     2.912s
4x4x8     3.005s
8x8x8     3.636s
dev       9.595s
```

In the best case, this is over a factor of 3 improvement. The inputs file was our standard warm plasma benchmark:

```
max_step = 100
amr.n_cell =  256 256 256
amr.max_grid_size = 128
amr.blocking_factor = 32
amr.max_level = 0
geometry.dims = 3
boundary.field_lo = Periodic Periodic Periodic
boundary.field_hi = Periodic Periodic Periodic
geometry.prob_lo     = -20.e-6   -20.e-6   -20.e-6    # physical domain
geometry.prob_hi     =  20.e-6    20.e-6    20.e-6

warpx.verbose = 1
warpx.cfl = 0.999
algo.maxwell_solver = yee
algo.particle_shape = 3

electrons.charge = -q_e
electrons.mass = m_e
electrons.injection_style = "NUniformPerCell"
electrons.num_particles_per_cell_each_dim = 1 2 2
electrons.profile = constant
electrons.density = 1.e25  # number of electrons per m^3
electrons.ux_th  = 1. # uth the std of the (unitless) momentum
electrons.uy_th  = 1. # uth the std of the (unitless) momentum
electrons.uz_th  = 1. # uth the std of the (unitless) momentum

positrons.charge = q_e
positrons.mass = m_e
positrons.injection_style = "NUniformPerCell"
positrons.num_particles_per_cell_each_dim = 1 2 2
positrons.profile = constant
positrons.density = 1.e25  # number of positrons per m^3
positrons.momentum_distribution_type = "gaussian"
positrons.ux_th  = 1. # uth the std of the (unitless) momentum
positrons.uy_th  = 1. # uth the std of the (unitless) momentum
positrons.uz_th  = 1. # uth the std of the (unitless) momentum
```

Answers are unchanged w.r.t. dev:

```
[atmyers2@login2 SharedTestCrusher_9_6_2022]$ ./fcompare.gnu.x86-trento.ex diags/diag1000050 shared_diags/diag1000050/

            variable name            absolute error            relative error
                                        (||A - B||)         (||A - B||/||A||)
 ----------------------------------------------------------------------------
 level = 0
 Bx                                 7.815970093e-14           1.520984891e-15
 By                                 7.405187574e-14            1.69736307e-15
 Bz                                 7.371880884e-14           1.647847181e-15
 Ex                                  2.74181366e-05           1.782039511e-15
 Ey                                  2.53804028e-05           1.662466781e-15
 Ez                                 2.574920654e-05            1.79266237e-15
 jx                                         0.34375           2.302983221e-15
 jy                                     0.306640625           2.457765782e-15
 jz                                      0.33984375            2.80149781e-15
 rho                                 1.92085281e-09           2.335405994e-15
```

On A100, the speedup is less dramatic, but the shared algorithm still beats the dev one by ~30% if the tile size is chosen well.


Source/Particles/Deposition/SharedDepositionUtils.H
```diff
@@ -0,0 +1,281 @@
...
+    }
+    if (j_type[zdir] == CELL) {
+        l_cell = compute_shape_factor(sz_cell, zmid - 0.5);
+    }
+    amrex::Real sz_j[depos_order + 1] = {0._rt};
+    for (int iz=0; iz<=depos_order; iz++)
+    {
+        sz_j[iz] = ((j_type[zdir] == NODE) ? amrex::Real(sz_node[iz]) : amrex::Real(sz_cell[iz]));
+    }
+    int const l_j = ((j_type[zdir] == NODE) ? l_node : l_cell);
```
AlexanderSinn, line 240:
```suggestion
    const double zmid = ((zp - zmin) + relative_time*vz)*dzi;
    double sz_j[depos_order + 1] = {0.};
    int const l_j = compute_shape_factor(sz_j, (j_type[zdir] == NODE) ? zmid : zmid - 0.5);
```
Same for x and y direction

Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -335,6 +336,247 @@ void doDepositionShapeN(const GetParticlePosition& GetPosition,
...
+            // Looks like it sets buffer box to contain the tile
+            // iv is a point. first box is the entire domain, last box gets written with size bin_size
+            getTileIndex(iv, box, true, bin_size, buffer_box_x);
+            getTileIndex(iv, box, true, bin_size, buffer_box_y);
+            getTileIndex(iv, box, true, bin_size, buffer_box_z);
+        }
+
+        Box tbox_x = convert(buffer_box_x, jx_type);
+        Box tbox_y = convert(buffer_box_y, jy_type);
+        Box tbox_z = convert(buffer_box_z, jz_type);
```
AlexanderSinn: Surely it must be possible to calculate the buffer_box from bin_id directly without having to sample the position of a particle. I would also suggest tiling in a way where the user specifies the maximum tile size instead of the minimum size. This could avoid using too much shared memory, while also simplifying the index calculation compared to what getTileIndex does. Additionally, an inverse function is available. Here is an example for 1D:
cell to tile ID:
```
ntiles = (ncells + maxtilesize - 1) / maxtilesize;
tile_id = ((i_cell - domain_lo) * ntiles + ntiles - 1) / ncells
```
tile ID to box:
```
box_lo = (tile_id * ncells) / ntiles + domain_lo
box_hi = ((tile_id + 1) * ncells) / ntiles - 1 + domain_lo
```
In 3d the 3 different tile_ids would be flattened to one dimension similar to Array4 (inverse: like ParallelFor over a box).

Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -335,6 +336,247 @@ void doDepositionShapeN(const GetParticlePosition& GetPosition,
...
+
+        tbox_x.grow(depos_order);
+        tbox_y.grow(depos_order);
+        tbox_z.grow(depos_order);
```
AlexanderSinn:
```suggestion
            getTileIndex(iv, box, true, bin_size, buffer_box);
        }
        
        buffer_box.grow((depos_order + 1) / 2);

        Box tbox_x = convert(buffer_box, jx_type);
        Box tbox_y = convert(buffer_box, jy_type);
        Box tbox_z = convert(buffer_box, jz_type);
```

`(depos_order + 1) / 2` guard cells should be sufficient right?

Source/Particles/Deposition/SharedDepositionUtils.H
kaplannp: Looks good! Tested in 3D, 2D, and RZ, then committed this.

Source/Particles/Deposition/CurrentDeposition.H
kaplannp: Looks good! Tested in 3D, 2D, and RZ, then committed this.

Source/Particles/Deposition/CurrentDeposition.H
kaplannp: Will consider this, but recommend we proceed with merge for now as the current changes already demonstrate a good speedup. There's also something to be said for maintainability, readability, and simplicity of reusing AMReX code to calculate tilebox size

kaplannp:
Including some figures here with heatmaps showing speed performance varying tilesize and ppc on Perlmutter (first two plots) and Crusher (last two plots) in 2 and 3 d. Timings are given in terms of ratios shared/dev (smaller is better). We use tsize=144 as default in 2d and 288 in 3d. These defaults have potential for 3X speed increase, and the plots show problem specific tuning can improve performance further.
pm2DExpanded
pm3DNewSort
cr3D
cr2DIncreasedProbSize

### LinePlots  
Here are two more figures demonstrating raw speedup looking at the times in seconds varying PPC based on default tilesize for 3d on Perlmutter and Crusher.  The times for Perlmutter use the latest code that incorporates PR #3762 for improved sorting efficiency. SortingForDeposition is currently turned off for crusher, so it's difficult to compare these two plots, but in each case, the Working or shared times are plotted in relation to development for the default tilesize. For every PPC, shared outperforms Dev, and for large PPC, it outperforms Dev by a wide margin.

lineTSize288
cr3DTSize288

Source/Particles/Deposition/SharedDepositionUtils.H
kaplannp: Actually, after further testing, this breaks things for single precision. 
amrex::Gpu::Atomic::AddNoRet requires that sz_j be an amrex::Real  while compute_shape_factor requires sz_j be a double.
The previous version of the code called compute_shape_factor on the double sz_cell, and AddNoRet on the amrex::Real sz_j. so there was no issue. I saw two obvious ways to fix this. The first is to cast sz_j when we call AddNoRet, and the second is to add a for loop after compute_shape_factor to initialize a new sz_j of type amrex::Real. I tested both methods and found them slower (results below). I'm therefore reverting to previous version of code. If anyone finds a better way to keep the simplicity of the proposed changes without sacrificing runtime, will review.
outCast:Total Time                     : 2.67980957
outFor:Total Time                     : 2.624755859
outDev:Total Time                     : 2.297363281

Source/Particles/Deposition/CurrentDeposition.H
kaplannp: Did a little testing, and looks like we got a small performance boost from this. These are for 1 2 2 and 2 4 4 ppc:
```
outWithDepChange:Total Time                     : 3.574548124
outPreDep:Total Time                     : 3.656580654
outWithDepChangeL:Total Time                     : 19.51583009
outPreDepL:Total Time                     : 20.03403573
```

Docs/source/usage/parameters.rst
```diff
@@ -2093,6 +2093,15 @@ Additional parameters
      If ``sort_intervals`` is activated and ``sort_particles_for_deposition`` is ``false``, particles are sorted in bins of ``sort_bin_size`` cells.
      In 2D, only the first two elements are read.
 
+* ``warpx.do_shared_mem_charge_deposition`` (`bool`) optional (default `false`)
```
ax3l, line 2134: Why default to false?

Source/Particles/Deposition/ChargeDeposition.H
```diff
@@ -201,6 +202,302 @@ void doChargeDepositionShapeN (const GetParticlePosition& GetPosition,
 #endif
         }
         );
+#if defined(WARPX_USE_GPUCLOCK)
+    if( load_balance_costs_update_algo == LoadBalanceCostsUpdateAlgo::GpuClock) {
```
ax3l: Source of many bugs in the past:
```suggestion
    if (cost && load_balance_costs_update_algo == LoadBalanceCostsUpdateAlgo::GpuClock) {
```

This function is sometimes called early, before cost arrays are allocated.

Please also add in line 93 above to avoid a memory leak of `cost_real`.

Source/Particles/Deposition/ChargeDeposition.H
```diff
@@ -201,6 +202,302 @@ void doChargeDepositionShapeN (const GetParticlePosition& GetPosition,
...
+    const auto offsets_ptr = a_bins.offsetsPtr();
+    const int threads_per_block = 256;
+
+    std::size_t shared_mem_bytes = npts*sizeof(amrex::Real);
+    amrex::IntVect bin_size = WarpX::shared_tilesize;
+
+    const std::size_t max_shared_mem_bytes = amrex::Gpu::Device::sharedMemPerBlock();
+
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(shared_mem_bytes <= max_shared_mem_bytes,
+                                      "Tile size too big for GPU shared memory charge deposition");
```
ax3l:
```suggestion
    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(shared_mem_bytes <= max_shared_mem_bytes,
                                     "Tile size too big for GPU shared memory charge deposition");
```

Source/Particles/Deposition/ChargeDeposition.H
```diff
@@ -201,6 +202,302 @@ void doChargeDepositionShapeN (const GetParticlePosition& GetPosition,
...
+        Box tbx = convert( buffer_box, ix_type);
+        tbx.grow(depos_order);
+
+        Gpu::SharedMemory<amrex::Real> gsm;
+        amrex::Real* const shared = gsm.dataPtr();
+
+        amrex::Array4<amrex::Real> buf(shared, amrex::begin(tbx), amrex::end(tbx), 1);
+
+        // Zero-initialize the temporary array in shared memory
+        volatile amrex::Real* vs = shared;
```
ax3l, line 375: Why do we need `volatile` here?
(I remember an ancient CUDA quirk with multiple shared memory locations that needed this.)

Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -334,6 +335,214 @@ void doDepositionShapeN(const GetParticlePosition& GetPosition,
...
+    amrex::IntVect const jz_type = jz_fab.box().type();
+
+    constexpr int zdir = WARPX_ZINDEX;
+    constexpr int NODE = amrex::IndexType::NODE;
+    constexpr int CELL = amrex::IndexType::CELL;
+
+    // Loop over particles and deposit into jx_fab, jy_fab and jz_fab
+#if defined(WARPX_USE_GPUCLOCK)
+    amrex::Real* cost_real = nullptr;
+    if( load_balance_costs_update_algo == LoadBalanceCostsUpdateAlgo::GpuClock) {
```
ax3l:
```suggestion
    if (cost && load_balance_costs_update_algo == LoadBalanceCostsUpdateAlgo::GpuClock) {
```

Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -334,6 +335,214 @@ void doDepositionShapeN(const GetParticlePosition& GetPosition,
...
+            depositComponent<depos_order>(GetPosition, wp, uxp, uyp, uzp, ion_lev, jz_buff, jz_type,
+                                          relative_time, dx, xyzmin, lo, q, n_rz_azimuthal_modes,
+                                          ip, zdir, NODE, CELL, 2);
+        }
+
+        __syncthreads();
+        addLocalToGlobal(tbox_z, jz_arr, jz_buff);
+    });
+#if defined(WARPX_USE_GPUCLOCK)
+    if( load_balance_costs_update_algo == LoadBalanceCostsUpdateAlgo::GpuClock) {
```
ax3l:
```suggestion
    if (cost && load_balance_costs_update_algo == LoadBalanceCostsUpdateAlgo::GpuClock) {
```

Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -334,6 +335,214 @@ void doDepositionShapeN(const GetParticlePosition& GetPosition,
...
+        cost_real = (amrex::Real *) amrex::The_Managed_Arena()->alloc(sizeof(amrex::Real));
+        *cost_real = 0._rt;
+    }
+#endif
+    const auto dxiarr = geom.InvCellSizeArray();
+    const auto plo = geom.ProbLoArray();
+    const auto domain = geom.Domain();
+
+    amrex::Box sample_tbox(IntVect(AMREX_D_DECL(0,0,0)), a_tbox_max_size - 1);
+    sample_tbox.grow((depos_order+1) / 2);
```
ax3l: Is this the same grow for all staggering and even/odd shapes?

Source/Particles/Deposition/SharedDepositionUtils.H
ax3l: The general solution here is to avoid `double` that we use right now and instead calculate the position deltas from positions that were reduced by the offset of the current box (or even cell).

The only reason we introduced double here as a work-around is that our positions are global and frequently run into precision issues z = z + dz for particles in moving window simulations.

The additionally needed type conversions at runtime (float<->double) can actually make things slower, as you saw.

Source/WarpX.cpp
```diff
@@ -132,6 +132,22 @@ bool WarpX::do_divb_cleaning = false;
...
+bool WarpX::do_shared_mem_current_deposition = false;
+#endif
+#if defined(WARPX_DIM_3D)
+amrex::IntVect WarpX::shared_tilesize(AMREX_D_DECL(6,6,8));
+#elif defined(WARPX_DIM_2D)
+amrex::IntVect WarpX::shared_tilesize(AMREX_D_DECL(14,14));
+#else
+amrex::IntVect WarpX::shared_tilesize(AMREX_D_DECL(0,0,0)); //Not used
+#endif
+int WarpX::shared_mem_current_tpb = 128;
```
ax3l: Let's document the defaults in the user-facing `parameters.rst`?

Docs/source/usage/parameters.rst
kaplannp: We haven't done as rigorous testing with charge dep as we have on current dep. We've also made some changes to Current dep that we haven't carried over to charge dep yet. Plan to mimic the work on Current Dep in Charge Dep in another PR, but since Current Dep works well, wanted to get that merged first

Source/WarpX.cpp
kaplannp: Yes, good call. Docs are updated now.

Source/Particles/Deposition/ChargeDeposition.H
kaplannp: I think you're right that we technically don't need it. From what I understand, the `volatile` stops compiler optimizations from storing shared values in registers, which can generally cause problems when threads are accessing the same value in shared memory, as we do. I think because of the places we `__syncthreads`, and because most of our accesses to shared are atomics, we should be ok, but it's the kind of thing we'd need to make absolute sure of, and even a slight change later on, like moving the `__syncthreads`, could lead to some inexplicable bugs.  I ran some tests, and didn't see any performance benefit from removing it, so I'd recommend keeping it.

Source/Particles/Deposition/CurrentDeposition.H
kaplannp: Whoops! That didn't work for shape 2, but I put things back to `grow(depos_order)`, and now passes tests

ax3l:
@atmyers @kaplannp can you please update to solve the small conflict?

Source/Particles/Deposition/CurrentDeposition.H
ax3l: Don't we overgrow now? I think for odd shapes you need to grow less than for even shapes.

Examples/Tests/particle_fields_diags/analysis_particle_diags_impl.py
```diff
@@ -218,5 +218,6 @@ def do_analysis(single_precision = False):
         print(k, 'relative error openPMD = ', error_opmd[k])
 
 
+
```
ax3l, line 221:
```suggestion
```

ax3l (approved):
Thank you, LGTM! Please feel free to present at the developer meeting on the AMReX-WarpX meeting on the details for everyone :)

PR #3327, opened 2022-08-19
CI: Test New v. Legacy BTD in `RigidInjection_BTD`

EZoni:
First step to add more CI coverage of the new BTD, see #3300. Follow up on the initial tests added in #1973.

Main changes:
- We now compare particle positions and momenta between legacy BTD and new BTD (both plotfiles and openPMD).
- The checksums are evaluated on the BTD plotfile data (so lab-frame data, instead of "regular" boosted-frame data).

To-do:
- [x] Add regular checksum analysis (on boosted-frame data)
- [x] If possible, move checksum analysis to back-transformed data
- [x] Compare more particle attributes between legacy and new BTD

Issue #3300, opened 2022-08-10
BTD Testing: Analysis Scripts


ax3l:
- [ ] for plotfile and openPMD diagnostics with BTD, we need to add CI analysis scripts that actually read the data to see its valid and in the range we expect (fields, particles)
- [ ] more runtime tests need to use openPMD to increase functional coverage

cc @RevathiJambunathan @RemiLehe @ax3l @dpgrote et al.

ax3l:
cc @EZoni and I will start this.

PR #3371, opened 2022-09-07
CI: Test New v. Legacy BTD in `BTD_ReducedSliceDiag`

EZoni:
Keep adding more CI coverage of the new BTD, see #3300. Similar to #3327.

Main changes:
- [x] Add checksums for the CI test `BTD_ReducedSliceDiag`, evaluated on the BTD plotfile data

To-do:
- [x] Understand why `ions` species is not written to output (new BTD): clarified, see comments in #3420.

Note:
We cannot compare legacy BTD data with new BTD data (either plotfile or openPMD), because the array shapes are different (see comment below). As far as I understood from @RTSandberg, this is a known discrepancy and it is not expected to be taken care of any time soon.


EZoni:
@RevathiJambunathan

When I compare the legacy BTD output with the new BTD output in the Python analysis (see files changed in this PR), I get Ez arrays of different shapes:
- legacy BTD: `(32, 32, 49)`
- new BTD (plotfile): `(32, 32, 64)`
- new BTD (openPMD): `(32, 32, 64)`

Note that the new BTD array with plotfile and openPMD formats agree with each other up to machine precision, so that's consistent. I cannot yet compare those with the legacy array because of the mismatch in the shapes.

Do you have any hints on where the `49` cells in the third dimension could be coming from for the legacy output in this case? Note that the legacy array with shape `(32, 32, 49)` is loaded in the Python analysis as follows:
```py
snapshot = './lab_frame_data/snapshots/snapshot00003'
header   = './lab_frame_data/snapshots/Header'
allrd, info = read_raw_data.read_lab_snapshot(snapshot, header)
Ez_legacy = allrd['Ez']
```

RTSandberg (approved):
Thanks for working on these tests!  They look good to me

RemiLehe (approved):

PR #3374, opened 2022-09-08
Fix: ABLASTR nvcc Phi Fine/Coarse

ax3l:
Poisson solver: work around an NVCC compile error downstream in ImpactX. This is likely due to nesting of lambdas.

### DIFF
--- a/Source/ablastr/fields/PoissonSolver.H
+++ b/Source/ablastr/fields/PoissonSolver.H
@@ -59,6 +59,41 @@
 
 namespace ablastr::fields {
 
+namespace details
+{
+    /** Local interpolation from phi_cp to phi[lev+1]
+     *
+     * This is needed to work-around an NVCC limitation in downstream code (ImpactX),
+     * when nesting lambdas. Otherwise this could be written directly into the
+     * ParallelFor.
+     *
+     * @param[out] phi_fp_arr phi on the fine level
+     * @param[in] phi_cp_arr phi on the coarse level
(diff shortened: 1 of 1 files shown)


Source/ablastr/fields/PoissonSolver.H
```diff
@@ -264,17 +299,14 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
...
-                amrex::ParallelFor( b,
-                                    [=] AMREX_GPU_DEVICE (int i, int j, int k) noexcept
-                {
-                    amrex::mf_nodebilin_interp(i, j, k, 0, phi_fp_arr, 0, phi_cp_arr, 0, refratio);
-                });
+            for (amrex::MFIter mfi(*phi[lev + 1], amrex::TilingIfNotGPU()); mfi.isValid(); ++mfi) {
+                amrex::Array4<amrex::Real> const phi_fp_arr = phi[lev + 1]->array(mfi);
+                amrex::Array4<amrex::Real const> const phi_cp_arr = phi_cp.array(mfi);
+
+                details::PoissonInterpCPtoFP const interp(phi_fp_arr, phi_cp_arr, refratio);
```
ax3l, line 306: I intentionally copy the Array4's and `IntVect` here, which should not be an issue...
(we used `=` before as well in the lambda and all things are const-ifyied where possible)

RemiLehe (approved):

PR #3381, opened 2022-09-12
Add species name to 2 error messages

lucafedeli88:
This PR makes two error messages more readable by adding the name of the species that has some issues.


ax3l (approved):
:tada:

PR #3388, opened 2022-09-12
Vay Deposition: Filter D, Exchange Guard Cells of J

EZoni:
Replace #3208 using the new functions introduced in #3222.

With Vay deposition (in the case without periodic single box), we want to filter D at the beginning of the Vay scheme and exchange the guard cells of J at the end (instead of performing all operations on J).

This removes some spurious signal that was observed in the current density (more evident in log scale, as shown below), with the following two-particle test ([inputs_2d.txt](https://github.com/ECP-WarpX/WarpX/files/9551776/inputs_2d.txt)):
<p align="center">
    
    
</p> 

Note that this solution, based on #3222, will not work with mesh refinement (where the splitting between filtering and exchange of guard cells over different MR patches is more complicated and still needs to be understood).

Note also that this affects only the standard PSATD loop, and not yet the multi-J one (where Vay deposition is not yet implemented).


oshapoval:
@EZoni Thanks for this PR! Looks good for me. Just reminder for everyone that current implementation of Vay deposition will not work with `psatd.periodic_single_box_fft=1` and we have assert warning message for that:

oshapoval (approved):
@EZoni Thanks for this PR! Looks good for me. Just reminder for everyone that current implementation of Vay deposition will not work with `psatd.periodic_single_box_fft=1`, and we have an abort warning message for that:

```Option algo.current_deposition=vay must be used with psatd.periodic_single_box_fft=0.```

File: Source/FieldSolver/WarpXPushFieldsEM.cpp [hunk 1]
```diff
@@ -725,7 +725,9 @@ WarpX::PushPSATD ()
             PSATDSubtractCurrentPartialSumsAvg();
 
             // Synchronize J and rho (if used)
-            SyncCurrent(current_fp, current_cp);
+            // TODO This works only without mesh refinement
```
RemiLehe, line 731: Could you add a comment here saying that we use `SumBoundaryJ` instead of `SyncCurrent` because the current was already filtered in the `Evolve` function?

RemiLehe (changes requested):
Thanks for this PR.
I just have one minor comment (see inline). Btw, is there an `Abort` in place, in order to check the user does not try to run WarpX with both Vay deposition and mesh refinement?

EZoni:
Thank you, Remi. I will address the comment asap. As far as aborting the code is concerned, here's the code for it:
https://github.com/ECP-WarpX/WarpX/blob/5761b4bf998eab84ea0f7e4b132026593f3ddf9f/Source/WarpX.cpp#L980-L983

RemiLehe (approved):

PR #3323, opened 2022-08-18
Add warning to set zmin and zmax for ion-acc example

n01r:
This PR adds a warning to the documentation of the laser-ion acceleration example to set `<species>.zmin / zmax / xmin / ... ` for simulations of highly dense, spatially confined/limited targets to avoid particle creation throughout the whole box which could lead to memory overflows on the device.

It also changes the ion acceleration example use these parameters.

PR #3392, opened 2022-09-13
Improve memory footprint for ParticleInitialize

RevathiJambunathan:
This PR solves the issue #3335  where we run out of memory when initializing particles within a region whose bounding box area is larger than the profile itself.

- [ ] docs: remove outdated recommendation paragraph from https://github.com/ECP-WarpX/WarpX/pull/3323#discussion_r1005011585

To recreate the issue, particles are initialized in a spherical ring as shown below 
Screenshot from 2022-09-13 09-27-52

On crusher I tried the following : 
1. Run development with the input (attached below) with random initialilization with 512 particles per cell (to exaggerate the error)
This simulation ran out of memory on crusher with the following message
```
amrex::Abort::0::Out of gpu memory. Free: 529225728 Asked: 574015488 !!!
SIGABRT
See Backtrace.0 file for details
MPICH ERROR [Rank 0] [job id 180341.5] [Tue Sep 13 11:58:51 2022] [crusher131] - Abort(6) (rank 0 in comm 496): application called MPI_Abort(comm=0x84000000, 6) - process 0

srun: error: crusher131: task 0: Exited with exit code 6
srun: launch/slurm: _step_signal: Terminating StepId=180341.5
```

2. With this PR, the same input file generates the spherical ring. 

To validate the implementation and ensure same number of particles are injected, I also did uniform particle injection and the total particles injected is the same (2 particles per cell per direction in this case). 
I also did a random number injection case with fewer particles and the difference in total particles were approximately equal with some difference that we can attribute to the randomness.

```
# Description:
#
# This initializes the electrons and positrons within a sphere
#

max_step = 1
amr.n_cell = 128 128 128
amr.max_grid_size = 32
amr.blocking_factor = 32
amr.max_level = 0
geometry.coord_sys   = 0                  # 0: Cartesian
geometry.dims = 3
geometry.prob_lo     = 0.0 0.0 0.0
geometry.prob_hi     = 8 8 8
boundary.field_lo = pml pml pml
boundary.field_hi = pml pml pml


algo.maxwell_solver = yee
warpx.verbose = 1
warpx.do_dive_cleaning = 0
warpx.use_filter = 1
warpx.cfl = .95
warpx.do_pml_in_domain = 1
warpx.pml_has_particles = 1
warpx.do_pml_j_damping = 1

my_constants.xc  = 4.
my_constants.yc  = 4.
my_constants.zc  = 4.
my_constants.dens = 1.e16
algo.particle_shape = 3
my_constants.r0 = 3
my_constants.dR = 0.25
particles.species_names = plasma_e plasma_p

plasma_e.charge = -q_e
plasma_e.mass = m_e
plasma_e.injection_style = "NRandomPerCell"
plasma_e.num_particles_per_cell = 512
#plasma_e.injection_style = "NUniformPerCell"
#plasma_e.num_particles_per_cell_each_dim = 2 2 2
plasma_e.profile = parse_density_function
plasma_e.density_function(x,y,z) = "( ((( (z-zc)*(z-zc) + (y-yc)*(y-yc) + (x-xc)*(x-xc) )^(0.5))<=(r0 + dR)) * ((( (z-zc)*(z-zc) + (y-yc)*(y-yc) + (x-xc)*(x-xc) )^(0.5))>=(r0-dR)) )*dens"
plasma_e.momentum_distribution_type = parse_momentum_function
## Inject stationary pairs
plasma_e.momentum_function_ux(x,y,z) = "0.0"
plasma_e.momentum_function_uy(x,y,z) = "0.0"
## uz is always 0 for the injection methods above
plasma_e.momentum_function_uz(x,y,z) = "0.0"
plasma_e.do_continuous_injection = 0
plasma_e.density_min = 0.1

plasma_p.charge = q_e
plasma_p.mass = m_e
plasma_p.injection_style = "NRandomPerCell"
plasma_p.num_particles_per_cell = 512 
plasma_p.profile = parse_density_function
plasma_p.density_function(x,y,z) = "( ((( (z-zc)*(z-zc) + (y-yc)*(y-yc) + (x-xc)*(x-xc) )^(0.5))<=(r0 + dR)) * ((( (z-zc)*(z-zc) + (y-yc)*(y-yc) + (x-xc)*(x-xc) )^(0.5))>=(r0-dR)) )*dens"
#plasma_p.injection_style = "NUniformPerCell"
#plasma_p.num_particles_per_cell_each_dim = 2 2 2
plasma_p.momentum_distribution_type = parse_momentum_function
## Inject stationary pairs
plasma_p.momentum_function_ux(x,y,z) = "0.0"
plasma_p.momentum_function_uy(x,y,z) = "0.0"
## uz ip always 0 for the injection methods above
plasma_p.momentum_function_uz(x,y,z) = "0.0"
plasma_p.do_continuous_injection = 0
plasma_p.density_min = 0.1

diagnostics.diags_names = plt
plt.diag_type = Full
plt.intervals = 1
plt.fields_to_plot = Ex Ey Ez Bx By Bz jx jy jz rho

warpx.reduced_diags_names = PartNum
PartNum.intervals = 1
PartNum.type = ParticleNumber
```

### DIFF
--- a/Source/Particles/PhysicalParticleContainer.cpp
+++ b/Source/Particles/PhysicalParticleContainer.cpp
@@ -990,41 +990,22 @@ PhysicalParticleContainer::AddPlasma (int lev, RealBox part_realbox)
                     r = 1;
                 }
                 pcounts[index] = num_ppc*r;
-                // update pcount by checking which particles have non-zero density
-                int flag_pcount = 0;
-                for (int ip = 0; ip < num_ppc*r; ++ip) {
-                    amrex::Real dens1 = 0;
-                    amrex::Real dens2 = 0;
-                    amrex::Real dens3 = 0;
-                    amrex::Real dens4 = 0;
-                    amrex::Real dens5 = 0;
-                    amrex::Real dens6 = 0;
-                    amrex::Real dens7 = 0;
-                    amrex::Real dens8 = 0;
-                    // To ensure density parser does not encounter erroneous
-                    // arithmetic operation, such as log(0) or sqrt(-1)
-                    // check if cell-corner is within injection bounds
-                    if (inj_pos->insideBounds(lo.x, lo.y, lo.z))
-                        dens1 = inj_rho->getDensity(lo.x, lo.y, lo.z);
-                    if (inj_pos->insideBounds(lo.x, lo.y, hi.z))
-                        dens2 = inj_rho->getDensity(lo.x, lo.y, hi.z);
-                    if (inj_pos->insideBounds(lo.x, hi.y, lo.z))
-                        dens3 = inj_rho->getDensity(lo.x, hi.y, lo.z);
-                    if (inj_pos->insideBounds(hi.x, lo.y, lo.z))
-                        dens4 = inj_rho->getDensity(hi.x, lo.y, lo.z);
-                    if (inj_pos->insideBounds(lo.x, hi.y, hi.z))
-                        dens5 = inj_rho->getDensity(lo.x, hi.y, hi.z);
-                    if (inj_pos->insideBounds(hi.x, lo.y, hi.z))
-                        dens6 = inj_rho->getDensity(hi.x, lo.y, hi.z);
-                    if (inj_pos->insideBounds(hi.x, hi.y, lo.z))
-                        dens7 = inj_rho->getDensity(hi.x, hi.y, lo.z);
-                    if (inj_pos->insideBounds(hi.x, hi.y, hi.z))
-                        dens8 = inj_rho->getDensity(hi.x, hi.y, hi.z);
-                    if (  dens1 > 0 || dens2 > 0 || dens3 > 0 || dens4 > 0
-                       || dens5 > 0 || dens6 > 0 || dens7 > 0 || dens8 > 0) {
-                        flag_pcount = 1;
-                    }
-                }
+                // update pcount by checking if cell-corners or cell-center
+                // has non-zero density
+                const auto xlim = GpuArray<Real, 3>{lo.x,(lo.x+hi.x)/2._rt,hi.x};
+                const auto ylim = GpuArray<Real, 3>{lo.y,(lo.y+hi.y)/2._rt,hi.y};
+                const auto zlim = GpuArray<Real, 3>{lo.z,(lo.z+hi.z)/2._rt,hi.z};
+
+                const auto checker = [&](){
+                    for (const auto& x : xlim)
+                        for (const auto& y : ylim)
+                            for (const auto& z : zlim)
+                                if (inj_pos->insideBounds(x,y,z)) {
+                                    if (inj_rho->getDensity(x,y,z) > 0) return 1;
+                                }
+                    return 0;
+                };
+                const int flag_pcount = checker();
                 if (flag_pcount == 1) {
                     pcounts[index] = num_ppc*r;
                 } else {
@@ -1038,6 +1019,8 @@ PhysicalParticleContainer::AddPlasma (int lev, RealBox part_realbox)
             amrex::ignore_unused(j,k);
 #endif
         });
+        // Need synchronize since we have tmp xlim, ylim, zlim arrays
+        amrex::Gpu::synchronize();
 
         // Max number of new particles. All of them are created,
         // and invalid ones are then discarded
--- a/Source/Particles/PhysicalParticleContainer.cpp
+++ b/Source/Particles/PhysicalParticleContainer.cpp
@@ -993,14 +993,33 @@ PhysicalParticleContainer::AddPlasma (int lev, RealBox part_realbox)
                 // update pcount by checking which particles have non-zero density
                 int flag_pcount = 0;
                 for (int ip = 0; ip < num_ppc*r; ++ip) {
-                    amrex::Real dens1 = inj_rho->getDensity(lo.x, lo.y, lo.z);
-                    amrex::Real dens2 = inj_rho->getDensity(lo.x, lo.y, hi.z);
-                    amrex::Real dens3 = inj_rho->getDensity(lo.x, hi.y, lo.z);
-                    amrex::Real dens4 = inj_rho->getDensity(hi.x, lo.y, lo.z);
-                    amrex::Real dens5 = inj_rho->getDensity(lo.x, hi.y, hi.z);
-                    amrex::Real dens6 = inj_rho->getDensity(hi.x, lo.y, hi.z);
-                    amrex::Real dens7 = inj_rho->getDensity(hi.x, hi.y, lo.z);
-                    amrex::Real dens8 = inj_rho->getDensity(hi.x, hi.y, hi.z);
+                    amrex::Real dens1 = 0;
+                    amrex::Real dens2 = 0;
+                    amrex::Real dens3 = 0;
+                    amrex::Real dens4 = 0;
+                    amrex::Real dens5 = 0;
+                    amrex::Real dens6 = 0;
+                    amrex::Real dens7 = 0;
+                    amrex::Real dens8 = 0;
+                    // To ensure density parser does not encounter erroneous
+                    // arithmetic operation, such as log(0) or sqrt(-1)
+                    // check if cell-corner is within injection bounds
+                    if (inj_pos->insideBounds(lo.x, lo.y, lo.z))
+                        dens1 = inj_rho->getDensity(lo.x, lo.y, lo.z);
+                    if (inj_pos->insideBounds(lo.x, lo.y, hi.z))
+                        dens2 = inj_rho->getDensity(lo.x, lo.y, hi.z);
+                    if (inj_pos->insideBounds(lo.x, hi.y, lo.z))
+                        dens3 = inj_rho->getDensity(lo.x, hi.y, lo.z);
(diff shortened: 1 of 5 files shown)


Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -986,6 +986,29 @@ PhysicalParticleContainer::AddPlasma (int lev, RealBox part_realbox)
                     r = 1;
                 }
                 pcounts[index] = num_ppc*r;
+                // update pcount by checking which particles have non-zero density
+                int flag_pcount = 0;
+                for (int ip = 0; ip < num_ppc*r; ++ip) {
+                    const XDim3 r_loc =
+                        inj_pos->getPositionUnitBox(ip, lrrfac, amrex::RandomEngine{});
+                    amrex::Real dens1 = inj_rho->getDensity(lo.x, lo.y, lo.z);
```
atmyers: Are you sure this will do the correct thing in 1D / 2D / 2D RZ geometry?

RevathiJambunathan: oops. forgot to remove this when cleaning it up

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -986,6 +986,29 @@ PhysicalParticleContainer::AddPlasma (int lev, RealBox part_realbox)
                     r = 1;
                 }
                 pcounts[index] = num_ppc*r;
+                // update pcount by checking which particles have non-zero density
+                int flag_pcount = 0;
+                for (int ip = 0; ip < num_ppc*r; ++ip) {
+                    const XDim3 r_loc =
+                        inj_pos->getPositionUnitBox(ip, lrrfac, amrex::RandomEngine{});
```
RevathiJambunathan:
```suggestion
```

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -986,6 +986,27 @@ PhysicalParticleContainer::AddPlasma (int lev, RealBox part_realbox)
                     r = 1;
                 }
                 pcounts[index] = num_ppc*r;
+                // update pcount by checking which particles have non-zero density
```
RevathiJambunathan:
```suggestion
                // Flag cells for pcount injection if at-least one cell-corner
                // gives a non-zero density

```

Source/Particles/PhysicalParticleContainer.cpp
RevathiJambunathan: and will test for 1D/2D/RZ

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -986,6 +986,46 @@ PhysicalParticleContainer::AddPlasma (int lev, RealBox part_realbox)
...
+                    if (inj_pos->insideBounds(hi.x, lo.y, lo.z))
+                        dens4 = inj_rho->getDensity(hi.x, lo.y, lo.z);
+                    if (inj_pos->insideBounds(lo.x, hi.y, hi.z))
+                        dens5 = inj_rho->getDensity(lo.x, hi.y, hi.z);
+                    if (inj_pos->insideBounds(hi.x, lo.y, hi.z))
+                        dens6 = inj_rho->getDensity(hi.x, lo.y, hi.z);
+                    if (inj_pos->insideBounds(hi.x, hi.y, lo.z))
+                        dens7 = inj_rho->getDensity(hi.x, hi.y, lo.z);
+                    if (inj_pos->insideBounds(hi.x, hi.y, hi.z))
+                        dens8 = inj_rho->getDensity(hi.x, hi.y, hi.z);
```
RevathiJambunathan: @atmyers it was necessary to actually check insideBounds for all the points where we compute density
Otherwise I get an erroneous arithmetic operation especially for functions like `log(x)` where x is 0
Overlap checks if cell(lo,hi) are within particles bounds. 
So if hi is within particle min,max, then density would be computed for the lo as well .

Because of this -- I need to add a little buffer for test-cases such that cell-corners satisfy the insideBound condition.

lucafedeli88: Maybe this could be made more compact by doing something like (unless you need the densities):
```
const auto xlim = GpuArray<Real, 2>{lo.x,hi.x};
const auto ylim = GpuArray<Real, 2>{lo.y,hi.y};
const auto zlim = GpuArray<Real, 2>{lo.z,hi.z};

const auto checker = [&](){
    for (const auto& x : xlim)
        for (const auto& y : ylim)
            for (const auto& z : zlim)
                if (inj_pos->insideBounds(x,y,z) &&  inj_rho->getDensity(x,y,z) > 0) return 1;
    return 0;
};

const auto flag_pcount = checker();
```

RevathiJambunathan: oh thank you @lucafedeli88 !! 
I ll test this and then add it!

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -994,6 +1015,8 @@ PhysicalParticleContainer::AddPlasma (int lev, RealBox part_realbox)
             amrex::ignore_unused(j,k);
 #endif
         });
+        // Need synchronize since we have tmp xlim, ylim, zlim arrays
+        amrex::Gpu::synchronize();
```
RevathiJambunathan: is this needed?

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -986,6 +986,27 @@ PhysicalParticleContainer::AddPlasma (int lev, RealBox part_realbox)
...
+                const auto xlim = GpuArray<Real, 3>{lo.x,(lo.x+hi.x)/2._rt,hi.x};
+                const auto ylim = GpuArray<Real, 3>{lo.y,(lo.y+hi.y)/2._rt,hi.y};
+                const auto zlim = GpuArray<Real, 3>{lo.z,(lo.z+hi.z)/2._rt,hi.z};
+
+                const auto checker = [&](){
+                    for (const auto& x : xlim)
+                        for (const auto& y : ylim)
+                            for (const auto& z : zlim)
+                                if (inj_pos->insideBounds(x,y,z)) {
+                                    if (inj_rho->getDensity(x,y,z) > 0) return 1;
```
RevathiJambunathan: need to compute `getDensity` only if position is within particle bounds so that we dont run into density profiles with terms including log(0) or sqrt(-ve)

RevathiJambunathan:
@atmyers for the two tests cases that fail, 
the differences in checksums are very small. Do you agree? 
I checked `level_thinning` case and found that the particles injected are the same. 
I changed particle min,max in the input file.

ax3l:
@RevathiJambunathan can you potentially plot the final density and E fields of the two failing tests `LaserIonAcc2d` and `leveling_thinning`? Just to be sure.

RevathiJambunathan:
> @RevathiJambunathan can you potentially plot the final density and E fields of the two failing tests `LaserIonAcc2d` and `leveling_thinning`? Just to be sure.

oh sure! I am just seeing your comment here!

ax3l:
We added a note about the old cutoff recommendation in https://github.com/ECP-WarpX/WarpX/pull/3323#discussion_r1005011585

We can remove that when this PR is getting merged :)
(please rebase / merge `development`)

Source/Particles/PhysicalParticleContainer.cpp
atmyers: I don't think so - a streamSynchronize is implied by the below can to `ExclusiveScan`.

atmyers:
I agree that the test can simply be reset - could you rebase and do that, then I think this can be merged?

RevathiJambunathan:
@atmyers Thank you. 
Ys. I will rebase and let the tests run again and reset them!

Source/Particles/PhysicalParticleContainer.cpp
RevathiJambunathan: thank you. I will remove it then

Source/Particles/PhysicalParticleContainer.cpp
lucafedeli88: if you want to avoid the nested ifs, this should be fine:
`if(inj_pos->insideBounds(x,y,z) &&  inj_rho->getDensity(x,y,z) > 0)` 
I think that C++ standard guarantees that `inj_rho->getDensity(x,y,z) > 0` will never be evaluated if `inj_pos->insideBounds(x,y,z)` is false

RevathiJambunathan: ah okay :) Thank you!

RevathiJambunathan:
Charge density and Efield from LaserIonAcc2D test (updated checksum benchmark in [f177e98](https://github.com/ECP-WarpX/WarpX/pull/3392/commits/f177e98225f93b8ee260163df800be5dcae81ba5))
Screenshot from 2022-12-09 13-14-58

Distribution of resampled particles for leveling_thinning test compared to development.
(updated checksum in [f177e98](https://github.com/ECP-WarpX/WarpX/pull/3392/commits/f177e98225f93b8ee260163df800be5dcae81ba5))

Screenshot from 2022-12-09 13-15-11

atmyers (approved):

PR #3395, opened 2022-09-13

WeiqunZhang:
The GatherAndPush kernel in the PushPX function has a very low occupancy due to register pressure.  There are a number of reasons.  By default, we compile with QED module on, even if we do not use it at run time.  Another culprit is the GetExternalEB functor that contains 7 Parsers.  Again, we have to pay a high runtime cost, even if we do not use it.  In this PR, we move some runtime logic out of the GPU kernel, and use the following pattern to eleminate the unnecessary cost if QED and GetExternalEB are not used at run time.

    struct NoExternalEB{};
    struct HasExternalEB{};
    auto make_lambda = [=] (auto exteb_flag) {
        return [=] AMREX_GPU_DEVICE () {
            if constexpr (std::is_same<decltype(exteb_flag),HasQED>::value) {
                getExternalEB(...); // slow code
            }
        };
    };

    if (need_to_use_external_eb) {
        auto lambda = make_lambda(HasExternalEB{});
        amrex::ParallelFor(..., lambda);
    } else {
        auto lambda = make_lambda(NoExternalEB{});
        amrex::ParallelFor(..., lambda);
    }

    NumSgprs: 108
    NumVgprs: 256
    NumAgprs: 40
    TotalNumVgprs: 296
    ScratchSize: 264
    Occupancy: 1

    NumSgprs: 104
    NumVgprs: 249
    NumAgprs: 0
    TotalNumVgprs: 249
    ScratchSize: 144
    Occupancy: 2

### DIFF
--- a/Source/Particles/Gather/GetExternalFields.H
+++ b/Source/Particles/Gather/GetExternalFields.H
@@ -13,13 +13,12 @@
 #include <AMReX_Parser.H>
 #include <AMReX_REAL.H>
 
-enum ExternalFieldInitType { None, Constant, Parser, RepeatedPlasmaLens, Unknown };
-
 /** \brief Functor class that assigns external
  *         field values (E and B) to particles.
 */
 struct GetExternalEBField
 {
+    enum ExternalFieldInitType { None, Constant, Parser, RepeatedPlasmaLens, Unknown };
 
     GetExternalEBField () = default;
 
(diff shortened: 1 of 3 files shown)


WeiqunZhang:
Unfortunately, nvcc does not support extended device lambda inside a generic lambda.

ax3l:
Oh no, so my prototype was too simple today:
https://cuda.godbolt.org/z/nxqe6Errf

AlexanderSinn:
I tucked everything away in a special parallel_for_compiletime_options function using variadic templates
https://cuda.godbolt.org/z/z3WYjoxPr
This compiles even for CUDA, if it actually works in practice is another question.

ax3l:
That is very solid Alex, awesome. Weiqun is testing this on HIP rn and we try to make this a general helper in AMReX.

ax3l:
The current version makes a lot of sense and can provide what we do.

We might want to generalize it to allow that the options are not mutually exclusive. Both variants are useful.

WeiqunZhang:
See #3399

PR #3399, opened 2022-09-14

WeiqunZhang:
The GatherAndPush kernel in the PushPX function has a very low occupancy due to register pressure.  There are a number of reasons.  By default, we compile with QED module on, even if we do not use it at run time.  Another culprit is the GetExternalEB functor that contains 7 Parsers.  Again, we have to pay a high runtime cost, even if we do not use it.  In this PR, we move some runtime logic out of the GPU kernel to eleminate the unnecessary cost if QED and GetExternalEB are not used at run time.  This uses a fold expression trick that @AlexanderSinn taught us.

    NumSgprs: 108
    NumVgprs: 256
    NumAgprs: 40
    TotalNumVgprs: 296
    ScratchSize: 264
    Occupancy: 1

    NumSgprs: 104
    NumVgprs: 249
    NumAgprs: 0
    TotalNumVgprs: 249
    ScratchSize: 144
    Occupancy: 2

### DIFF
--- a/Source/Particles/PhysicalParticleContainer.cpp
+++ b/Source/Particles/PhysicalParticleContainer.cpp
@@ -134,7 +134,7 @@ void ParallelFor(Long N, F&& f, CompiletimeOptions<Op...>, int runtime_option)
               )
         ),...
     );
-    AMREX_ASSERT(option_miss < sizeof...(option_pack));
+    AMREX_ASSERT(option_miss + 1 == sizeof...(Op));
 }
 
 namespace
--- a/Source/Particles/PhysicalParticleContainer.cpp
+++ b/Source/Particles/PhysicalParticleContainer.cpp
@@ -2894,18 +2894,6 @@ PhysicalParticleContainer::PushPX (WarpXParIter& pti,
         scaleFields(xp, yp, zp, Exp, Eyp, Ezp, Bxp, Byp, Bzp);
 
 #ifdef WARPX_QED
-        if constexpr (control.value ==  NoExtEB_HasQED ||
-                      control.value == HasExtEB_HasQED) {
-            if (do_sync) {
-                doParticlePush<1>(getPosition, setPosition, copyAttribs, ip,
-                                  ux[ip], uy[ip], uz[ip],
-                                  Exp, Eyp, Ezp, Bxp, Byp, Bzp,
-                                  ion_lev ? ion_lev[ip] : 0,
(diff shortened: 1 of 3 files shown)


WeiqunZhang:
It works with HIP.  Let's see if it compiles with CUDA and SYCL.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -109,6 +109,34 @@
...
+                  amrex::ParallelFor(N, [f] AMREX_GPU_DEVICE (Long i) {
+                      f(i, CTOption<Op>{});
+                  })
+              )
+            : (
+                ++option_miss, void()
+              )
+        ),...
+    );
+    AMREX_ASSERT(option_miss < sizeof...(option_pack));
```
AlexanderSinn:
```suggestion
    AMREX_ASSERT(option_miss + 1 == sizeof...(Op));
```
Should be a bit safer this way

WeiqunZhang:
Unfortunately, it does not work with nvcc >= 15.
```
An extended __device__ lambda cannot first-capture variable in constexpr-if context.
```

AlexanderSinn:
I had that problem a few times before, the default lambda capture `[=]` needs to be replaced with one explicitly mentioning all of the needed variables, including the one only used inside the constexpr-if.

https://github.com/Hi-PACE/hipace/blob/98c62be645c8a42694b9c75c880da1ba1fe5bdb1/src/particles/deposition/PlasmaDepositCurrentInner.H#L188-L201

WeiqunZhang:
The problem is it might be very expensive to capture something that is not needed in the case of `GetExternalEB`.

AlexanderSinn:
Lambda capture is a fundamental C++ inefficiency and I think there is no way to reduce that without adding a wrapper around PushPX, which calls one multiple PushPX templates.
I had a setup to somewhat fix this here: https://github.com/Hi-PACE/hipace/pull/775/commits/470f6ed1be9a90a00241d9c0ec775838d7d71dc8 but removed it because of the negligible performance impact in that case.
The hope is that the GPU code doesn’t access all of the captured data, maybe capturing pointers to data in GPU memory instead would help.

WeiqunZhang:
After working about the lambda capture issue, we now have a redefinition problem with the new nvcc compiler.  :disappointed:   I will try to do this in yet another way.

WeiqunZhang:
Yet another attempt #3402

PR #3400, opened 2022-09-14
Update profiling instructions for Perlmutter

RemiLehe:


RemiLehe:
This fix does not actually work. I get:
```
Failed to connect to Agent: The timeout expired.
srun: error: nid002433: task 0: Exited with exit code 1
srun: launch/slurm: _step_signal: Terminating StepId=3154495.0
```

PR #3402, opened 2022-09-14

WeiqunZhang:
The GatherAndPush kernel in the PushPX function has a very low occupancy due to register pressure.  There are a number of reasons.  By default, we compile with QED module on, even if we do not use it at run time.  Another culprit is the GetExternalEB functor that contains 7 Parsers.  Again, we have to pay a high runtime cost, even if we do not use it.  In this PR, we move some runtime logic out of the GPU kernel to eleminate the unnecessary cost if QED and GetExternalEB are not used at run time.

    NumSgprs: 108
    NumVgprs: 256
    NumAgprs: 40
    TotalNumVgprs: 296
    ScratchSize: 264
    Occupancy: 1

    NumSgprs: 104
    NumVgprs: 249
    NumAgprs: 0
    TotalNumVgprs: 249
    ScratchSize: 144
    Occupancy: 2

### DIFF
--- a/Source/Particles/PhysicalParticleContainer.cpp
+++ b/Source/Particles/PhysicalParticleContainer.cpp
@@ -2624,7 +2624,24 @@ PhysicalParticleContainer::PushPX (WarpXParIter& pti,
 
     const auto t_do_not_gather = do_not_gather;
 
-    amrex::ParallelFor( np_to_push, [=] AMREX_GPU_DEVICE (long ip)
+    enum exteb_flags : int { no_exteb, has_exteb };
+    enum qed_flags : int { no_qed, has_qed };
+
+    int exteb_runtime_flag = getExternalEB.isNoOp() ? no_exteb : has_exteb;
+#ifdef WARPX_QED
+    int qed_runtime_flag = (local_has_quantum_sync || do_sync) ? has_qed : no_qed;
+#else
+    int qed_runtime_flag = no_qed;
+#endif
+
+    // Using this version of ParallelFor with compile time options could
+    // improve performance when qed or external EB is not used by reducing
+    // register pressure.
+    amrex::ParallelFor(TypeList<CompileTimeOptions<no_exteb,has_exteb>,
+                                CompileTimeOptions<no_qed  ,has_qed>>{},
+                       {exteb_runtime_flag, qed_runtime_flag},
+                       np_to_push, [=] AMREX_GPU_DEVICE (long ip, auto exteb_control,
+                                                         [[maybe_unused]] auto qed_control)
     {
         amrex::ParticleReal xp, yp, zp;
         getPosition(ip, xp, yp, zp);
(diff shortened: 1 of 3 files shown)


WeiqunZhang:
This version works for HIP and nvcc 17.  Hopefully it will pass the CIs.

Source/Particles/Pusher/PushSelector.H
```diff
@@ -128,18 +121,16 @@ void doParticlePush(const GetParticlePosition& GetPosition,
...
                                    Ex, Ey, Ez, Bx,
                                    By, Bz, qp, m, dt);
         amrex::ParticleReal x, y, z;
         GetPosition(i, x, y, z);
         UpdatePosition(x, y, z, ux, uy, uz, dt );
         SetPosition(i, x, y, z);
-    } else {
-        amrex::Abort("Unknown particle pusher");
-    }
+    } //else {
```
ax3l, line 131: Commented to safe registers or to avoid compile issues?
if the former, then we could leave the `else` in during debug mode.

ax3l (approved):
That is excellent. Awesome work-around for nvcc :tada:

Source/Particles/Pusher/PushSelector.H
```diff
@@ -128,18 +121,16 @@ void doParticlePush(const GetParticlePosition& GetPosition,
...
+//        amrex::Abort("Unknown particle pusher");
+//    }
```
WeiqunZhang, line 133:
```suggestion
    }
#ifdef AMREX_DEBUG
    else {
        amrex::Abort("Unknown particle pusher");
    }
#endif
```

WeiqunZhang:
@maikel showed us a very cool trick. https://cuda.godbolt.org/z/edxEMY7YG

ax3l:
Oh awesome, that's the Cartesian product we need :rocket: :sparkles: 
Shall we wait for this to be finished? We can use this in many places :)

WeiqunZhang:
Yes, we could wait till the functionality is in amrex.

dpgrote:
If I can comment here, can this be done with templating instead of using the more obscure `std::is_same<decltype...`, similar to the templating for `doParticlePush`? A few comments in the code would be helpful, saying that this is being done to reduce register pressure, avoiding calls to the external fields and QED stuff when not used. And also, the lambda is probably big enough to be a separate routine. Otherwise, this is great with a very nice speed up!

WeiqunZhang:
Yes, the lambda is big enough.  So we do not want to write more than once.  Also it captures so many variables, it will also be error prone if we use a non-lambda function because we might mess up the order of the variables in a function's parameters.

AlexanderSinn:
> @maikel showed us a very cool trick. https://cuda.godbolt.org/z/edxEMY7YG

Here is a N-dimensional version of that, it even compiles with gcc7.5. I am still unsure about that NVCC/NVHPC redefinition problem, however.

https://cuda.godbolt.org/z/xP9nKMYdM

WeiqunZhang:
What's the redefinition problem with nvcc?  The compiler explorer link compiles with nvcc.

AlexanderSinn:
https://github.com/ECP-WarpX/WarpX/pull/3399#issuecomment-1247354064

WeiqunZhang:
Oh that.  I have no ideas.

WeiqunZhang:
https://github.com/AMReX-Codes/amrex/pull/2954

ax3l:
@WeiqunZhang we merged in the update of https://github.com/AMReX-Codes/amrex/pull/2954 to WarpX now. Feel free to ping me when this PR is rebased and ready to go :rocket:

WeiqunZhang:
Yes.  Thanks for reminding me!

WeiqunZhang:
@ax3l It's ready for review.

Source/Particles/Pusher/PushSelector.H
RemiLehe: Did you mean to apply this change @WeiqunZhang ?

RemiLehe (changes requested):
Thanks for this PR!
This looks almost ready to merge. But it looks like there are some remaining commented lines that still need to be converted to a debug-only code path; is that correct?

Source/Particles/Pusher/PushSelector.H
WeiqunZhang: Yes, I did mean to comment out or even remove the else part, because it could affect performance and it's not really necessary.  The runtime parameter is converted from a string to int.  If there is a typo in inputs, it would have be caught in WarpX::ReadParameters.  Admittedly, it's still possible that something could go wrong.  But it's very unlikely.  If we really want to make sure there is no unknown type, we could turn the `int` into a strong type like `enum struct`.

RemiLehe (approved):

RemiLehe:
@WeiqunZhang There seems to be a remaining compilation error with `clang` in the CI ; is that correct?

WeiqunZhang:
```
Failed to build pywarpx
```
I will try to rerun the job.  If it still fails, I will merge development into this to see if that will fix it.

WeiqunZhang:
Oh, I think I know why.  It needs a more recent version of amrex due to clang is not happy with an amrex function.  So I just merged development into this branch.

WeiqunZhang:
All checks have passed.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -2624,7 +2624,24 @@ PhysicalParticleContainer::PushPX (WarpXParIter& pti,
...
+
+    int exteb_runtime_flag = getExternalEB.isNoOp() ? no_exteb : has_exteb;
+#ifdef WARPX_QED
+    int qed_runtime_flag = (local_has_quantum_sync || do_sync) ? has_qed : no_qed;
+#else
+    int qed_runtime_flag = no_qed;
+#endif
+
+    // Using this version of ParallelFor with compile time options could
+    // improve performance when qed or external EB is not used by reducing
```
ax3l: Since this is tested and found to be true:
```suggestion
    // Using this version of ParallelFor with compile time options
    // improves performance when qed or external EB are not used by reducing
```

ax3l:
ping @lucafedeli88 FYI, as discussed :)

Source/Particles/Pusher/PushSelector.H
```diff
@@ -40,6 +40,8 @@
  * \param t_chi_max                 Cutoff chi for QSR
  * \param dt                        Time step size
  */
+
+template <int do_sync>
```
ax3l, line 44: Please add Doxygen for `\tparam do_sync`, since it should be clear from the API if we mean Synchrotron QED (or some sync in compute).

ax3l: done

ax3l (approved):
Awesome, great hackathon success :tada:

PR #3406, opened 2022-09-16
ABLASTR: Fix Stray Include in ChargeDeposition

ax3l:
The include in #3393 was not stray, the file uses `GetPosition`.

We might want to add a development branch (#3407) that also changes the WarpX particle type for hacking in https://github.com/ECP-WarpX/impactx/pull/245

Otherwise we need to template ABLASTR container stuff, which is more work and will not be needed long term with the new layout.


lucafedeli88 (approved):
Thanks for this PR!

PR #3413, opened 2022-09-19
Fixes to Physics_applications/capacitive_discharge/PICMI*

dpgrote:
A change was made in PICMI so that for the grid objects, the separate names for the number of cells in each dimension (i.e. nx, ny, nz) are no longer saved as attributes, and the implementations need to use the vector `number_of_cells` instead. This PR fixes several PICMI input files that were using the separate names.

This PR is needed in preparation for PR #3329 which increments the PICMI version which includes the change above.


ax3l (approved):
Thank you! :)

PR #3418, opened 2022-09-21
Fix compilation of RZ version on GPU

RemiLehe:
The recent PR https://github.com/ECP-WarpX/WarpX/pull/3255 broke compilation for the RZ code on GPU. (It seems that this was not tested in CI.)

This is due to a check which we already do anyways here:
https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/Collision/CollisionHandler.cpp#L35

Therefore, I removed the duplicated check here, in order to fix the compilation on GPU


EZoni (approved):
Thank you for the fix, Remi!

PR #3421, opened 2022-09-22
Add 1d support to `_libwarpx.py` functions `get_particle_X`

roelof-groenewald:
The handy libwarpx functions to get particle coordinates do not yet support 1d simulations. This PR adds that support.
One thing that might need some discussion is how to treat dimensions not included in the coordinate system chosen. For example, if `get_particle_y()` is called in 2d the previous implementation was to return the same coordinate as when calling `get_particle_z()`. But if `get_particle_r()` was called in a 2d Cartesian simulation an exception was raised saying r is not defined in 2d Cartesian.

### DIFF
--- a/Examples/Physics_applications/capacitive_discharge/PICMI_inputs_1d.py
+++ b/Examples/Physics_applications/capacitive_discharge/PICMI_inputs_1d.py
@@ -77,18 +77,6 @@ class PoissonSolver1D(picmi.ElectrostaticSolver):
         system."""
         self.nsolve = self.nz + 1
 
-        # Set up the tridiagonal computation matrix in order to solve A*phi =
-        # rho for phi.
-        self.A_ldiag = np.ones(self.nsolve-1) / self.dz**2
-        self.A_mdiag = -2.*np.ones(self.nsolve) / self.dz**2
-        self.A_udiag = np.ones(self.nsolve-1) / self.dz**2
-
-        self.A_mdiag[0] = 1.
-        self.A_udiag[0] = 0.0
-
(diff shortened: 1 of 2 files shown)


dpgrote (approved):
This looks good to me!

roelof-groenewald:
Just re-triggering the CI since there was a random failure.

ax3l:
Awesome! Want to use this in a 1D test to keep it covered in CI?

roelof-groenewald:
> Awesome! Want to use this in a 1D test to keep it covered in CI?

Sure thing! I added a call to the `get_particle_z` function in the 1d capacitive discharge test to cover that functionality.

ax3l (approved):
Thanks a lot :)

ax3l:
@rscohn2 We noticed that since roughly last week, we have an elevated rate of timeouts in Intel oneAPI builds on GitHub Action runners. This affects both `icx` host builds and `dpcpp` device build runners. Classic `icpc` runners seem to be unaffected.

It looks to me like these runners might get stuck in the linker steps. Sometimes, restarting seems to solve the issue.

Is there a new oneAPI release that came out recently?

rscohn2:
> Is there a new oneAPI release that came out recently?

Yes. I am not sure what day it was published, but sometime in the last 2 weeks.

ax3l:
I wonder if it's maybe also possible that we just run out of storage on runners again if the release maybe sized up a little - and the linker hang is just a symptom of that.

ax3l:
Will continue to track this compiler regression in #3442

PR #3426, opened 2022-09-23
[WIP] Profiler tags for visualizing BTD in NsightSystem 

RevathiJambunathan:


AlexanderSinn:
Looking at a Nsight profile I made a while ago for HiPACE++ I noticed two things:
-	Properly synchronized function/profiler names are available under 
Processes/[…] hipace/CUDA HW …/99.7% Context 1/82.4% Stream 17/NVTX
So basically, not in Threads and not in [All Streams] but in the category of the most used Stream(s).

-	The automatic synchronizing after a MFIter loop (and likely also ParIter) can take 20-30µs with one Rank, one GPU and one MF Box. If BTD uses a lot of these with little compute load, this might be the reason for the slow performance.

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -517,6 +518,9 @@ BTDiagnostics::UpdateBufferData ()
 void
 BTDiagnostics::PrepareFieldDataForOutput ()
 {
+    WARPX_PROFILE("BTDiagnostics::PrepFieldData()");
```
ax3l, line 750: Is the first line not sufficient?

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -728,6 +733,10 @@ BTDiagnostics::GetZSliceInDomainFlag (const int i_buffer, const int lev)
 void
 BTDiagnostics::Flush (int i_buffer)
 {
+    WARPX_PROFILE("BTDiagnostics::Flush()");
```
ax3l, line 736: same q as above

Source/Diagnostics/ComputeDiagFunctors/BackTransformFunctor.cpp
```diff
@@ -43,6 +44,9 @@ BackTransformFunctor::BackTransformFunctor (amrex::MultiFab const * mf_src, int
 void
 BackTransformFunctor::operator ()(amrex::MultiFab& mf_dst, int /*dcomp*/, const int i_buffer) const
 {
+    WARPX_PROFILE("BTDiagnostics::FieldOperator()");
```
ax3l, line 49: same q as above

Source/Diagnostics/ComputeDiagFunctors/BackTransformParticleFunctor.cpp
```diff
@@ -80,6 +81,9 @@ BackTransformParticleFunctor::BackTransformParticleFunctor (
 void
 BackTransformParticleFunctor::operator () (PinnedMemoryParticleContainer& pc_dst, int &totalParticleCounter, int i_buffer) const
 {
+    WARPX_PROFILE("BTDiagnostics::ParticleOperator()");
```
ax3l, line 84: same q as above

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -1130,6 +1142,9 @@ BTDiagnostics::InitializeParticleBuffer ()
 void
 BTDiagnostics::PrepareParticleDataForOutput()
 {
+    WARPX_PROFILE("BTDiagnostics::PrepParticleData()");
```
ax3l, line 1393: same q as above

PR #945, opened 2020-04-24
Generalize coarsening for MR

EZoni:
**Goal**

Resolve third point of issue #797 ("MR: Interpolation of charge and currents from fine to coarse"): extend functions used for interpolation from fine to coarse patch for mesh refinement (MR) applications.

**Description**

The PR introduces the **new namespace** `CoarsenMR` in the files `Source/Utils/CoarsenMR.H(/.cpp)`.

Because of the similarity between the new namespace `CoarsenMR` and the old namespace `Average` (PRs #795, #832, #853), the **old namespace** `Average` has been **renamed** as `CoarsenIO` **and cleaned-up**, in order to resemble the implementation of the new namespace.

The highest-level function of the new namespaces `CoarsenIO` and `CoarsenMR` is now named `Coarsen`. It relies on the two additional functions `Loop` (intermediate layer) and `Interp` (lowest-level kernel).

The functionality provided by the new namespace `CoarsenMR` replaces the one provided by the old classes `InterpolateDensityFineToCoarse` and `InterpolateCurrentFineToCoarse`, now removed.

The formulas below describe the interpolation implemented, in the 1D case (extensions to 2D and 3D are trivial). The subscripts 'f' and 'c' stand for fine and coarse, the parameter c<sub>r</sub> denotes the coarsening ratio and g denotes the quantity being coarsened (charge density or each component of the current density in our use cases). The trivial case c<sub>r</sub> = 1 is implemented (even though, clearly, not expected to be used for MR purposes) but not shown in the formulas below. The weights shown below can be crosschecked with the hard-coded weights of the old implementation for c<sub>r</sub> = 2 in 2D and 3D. 

Coarsening_with_staggering_for_IO_and_MR-pages-6-1

Three important notes:
1. The interpolation implemented is expected to be valid for **any coarsening ratio c<sub>r</sub> ≥ 1** (integer powers of 2 only, namely c<sub>r</sub> = 1, 2, 4, 8, ... ), independently along each direction.
2. The interpolation implemented is expected to be valid only when the source and destination MultiFabs have the **same staggering**.
3. The interpolation points and weights in 1D can be crosschecked with the help of the **new Python script** `Source/Utils/check_interp_points_and_weights.py`.

**Testing**

[...]

PR #853, opened 2020-03-26
Interpolation with coarsening for general staggering

EZoni:
**Note** This PR is related to PR #856 and issue #797. CPU/GPU regression tests ongoing.

**Goal**

Implement interpolation (namely, averaging over either 1 point or 2 equally distant points) with coarsening for general staggering of input and output data.

**Description**

New functions defined in the namespace `Average`: `CoarsenAndInterpolate`, `CoarsenAndInterpolateLoop` and `CoarsenAndInterpolateKernel`. The following cases are covered:
- fine nodal to coarse nodal
- fine centered to coarse nodal
- fine nodal to coarse centered
- fine centered to coarse centered

The labels above are meaningful in 1D: extending the same idea to 2D and 3D, face-centered data and edge-centered data are covered as well. Previous calls to `amrex::average_down` (should not give the exact same result) and `Average::ToCellCenter` (should give the same result) have been replaced with calls to `Average::CoarsenAndInterpolate`.

**Testing the PR**

The core loop implemented in `Average::CoarsenAndInterpolateKernel` has been tested by checking the indices of the input data used for interpolation with the following Python script:

```py
import numpy as np

# ic: index of the destination coarse array
# si: staggering of the source fine MultiFab
# so: staggering of the destination coarse MultiFab
# cr: coarsening ratio
def indices_onestag( ic, si, so, cr ):
    if ( cr==0 or ( cr!=1 and np.ceil(np.log2(cr))!=np.floor(np.log2(cr)) ) ):
        print( '\n ABORT: coarsening ratio must be an integer power of 2' )
        return
    elif (cr==1):
        n    = 1+abs(si-so) #2-(2*si*so-si-so+1)
        imin = ic-so*(1-si)
    else:
        n    = 2-si
        imin = ic*cr+cr//2*(1-so)-(1-si)
    if ( si==1 and so==1 ):
        print( '\n fine nodal to coarse nodal (coarsening ratio={})'.format( cr ) )
    elif ( si==0 and so==1 ):
        print( '\n fine centered to coarse nodal (coarsening ratio={})'.format( cr ) )
    elif ( si==1 and so==0 ):
        print( '\n fine nodal to coarse centered (coarsening ratio={})'.format( cr ) )
    elif ( si==0 and so==0 ):
        print( '\n fine centered to coarse centered (coarsening ratio={})'.format( cr ) )
    print( ' interpolate over n={} points'.format( n ), end='' )
    print( ' to get value at i_c={}:'.format( ic ), end='' )
[...]

PR #3433, opened 2022-09-28
ABLASTR: Coarsen Functions

ax3l:
Move coarsen functions to ABLASTR.

Needed for https://github.com/ECP-WarpX/impactx/pull/261

Follow-up to:
- #853
- #945

Edit by @EZoni:
Here's a few slides describing the interpolation operations refactored in this PR (interpolation between grids with arbitrary index type and mesh refinement ratio, for I/O and MR): [Slides-Interpolation-IO-MR.pdf](https://github.com/ECP-WarpX/WarpX/files/9904632/Slides-Interpolation-IO-MR.pdf). The slides date back to 2020, when the PRs listed above were opened and merged.


Source/ablastr/mesh_refinement/coarsen.H
```diff
@@ -1,25 +1,35 @@
...
-    using namespace amrex;
 
+namespace ablastr::mesh_refinement
+{
     /**
      * \brief Interpolates the floating point data contained in the source Array4
      *        \c arr_src, extracted from a fine MultiFab, with weights defined in
      *        such a way that the total charge is preserved.
      *
+     * The input (sf) and output (sc) staggering need to be the same.
```
ax3l: @EZoni let us clarify this and the difference to the other function.

Source/ablastr/mesh_refinement/coarsen.H
```diff
@@ -34,60 +44,72 @@ namespace CoarsenMR{
...
+        int const k,
+        int const comp
+    )
     {
+        using namespace amrex::literals;
+
+        AMREX_ASSERT(sf[0] == sc[0]);
+        AMREX_ASSERT(sf[1] == sc[1]);
+        AMREX_ASSERT(sf[2] == sc[2]);
+
```
ax3l: Not sure if true, but had we had the impression that is a requirements given line 74.

Source/ablastr/mesh_refinement/coarsenToCellCentered.H
```diff
@@ -1,26 +1,36 @@
...
-    using namespace amrex;
 
+namespace ablastr::mesh_refinement::to_cell_centered
+{
     /**
      * \brief Interpolates the floating point data contained in the source Array4
      *        \c arr_src, extracted from a fine MultiFab, by averaging over either
      *        1 point or 2 equally distant points.
      *
+     * The output (sc) staggering is always cell-centered.
```
ax3l: @EZoni I think I might be wrong with this detail. Let us clarify so its clear what the difference is to the other function.

Source/ablastr/mesh_refinement/coarsenToCellCentered.H
```diff
@@ -70,19 +88,18 @@ namespace CoarsenIO{
         const int jmin = idx_min[1];
         const int kmin = idx_min[2];
         int  ii, jj, kk;
-        Real wx, wy, wz;
+        amrex::Real const wx = 1.0_rt / static_cast<amrex::Real>(numx);
```
ax3l: I moved this out as a tiny optimization. The compiler probably did this already.

Source/ablastr/mesh_refinement/coarsen.H
EZoni: I think I agree with this. Also looking at my notes, I think the interpolation operation (fine to coarse, or coarsening indeed) used for mesh refinement purposes (not I/O) was derived and implemented only for the case where the fine and the coarse grids have the same index type. Which, I think, represents a very common and reasonable scenario.

Source/ablastr/mesh_refinement/coarsen.H
EZoni: I think this is correct, see my reply to the comment above. However, I wonder if we could move this check one layer up, where `Interp` is called, or something along these lines?

Source/ablastr/mesh_refinement/coarsenToCellCentered.H
EZoni: In this case I don't think the output index type has to be cell-centered. We do use only cell-centered data for I/O, but I think this `Interp` function was written in a general way, such that it works for an arbitrary output index type (nodal or cell-centered). 

Thus, I would suggest two changes (can be discussed further, if necessary):
- remove the `AMREX_ASSERT`s added below, as I think the function is expected to return a correct result even if the output index type is nodal;
- rename the file differently, since this is not specialized for cell-centered output only.

Alternatively, if we want this to be specialized for cell-centered output only, we could also clean up the formulas by setting `sc` to zero everywhere. But I'd be more in favor of the solution above, since the functions were implemented for arbitrary index type.

EZoni:
@ax3l I added some comments above, which hopefully clarify the scope of the various interpolation functions that you are refactoring here, to the best of my knowledge. As a general remark, note that the two sets of interpolation routines had been named originally with the tags `IO` and `MR`, because the type of interpolation we chose to do for I/O and MR is different (simple nearest-neighbors interpolation for I/O, and more complicated weights to guarantee certain conservation properties for MR - more details in the slides linked in the PR description), regardless of whether there could be constraints on the index type of the input/output data in certain cases.

Source/ablastr/mesh_refinement/coarsen.H
ax3l: Thanks, perfect.

Source/ablastr/mesh_refinement/coarsen.H
ax3l: Yes, this is pretty deep here - but also only used in Debug mode.

```suggestion
        AMREX_ASSERT_WITH_MESSAGE(sf[0] == sc[0], "Interp: Staggering for component 0 does not match!");
        AMREX_ASSERT_WITH_MESSAGE(sf[1] == sc[1], "Interp: Staggering for component 1 does not match!");
        AMREX_ASSERT_WITH_MESSAGE(sf[2] == sc[2], "Interp: Staggering for component 2 does not match!");

```

Source/ablastr/mesh_refinement/coarsenToCellCentered.H
ax3l: Thanks for checking and the details, @EZoni!

I think all that is left is finding a good name for this - I think coarsen "MR"/"IO" is not very descriptive right now...

ax3l:
Thanks @EZoni - I think all that is left then is finding a good name, especially for the I/O stuff, which I find not very descriptive. Do you think we could describe the properties of the two classes (e.g., conservation properties and time/space average) in the name(s)?

Source/ablastr/mesh_refinement/coarsenToCellCentered.H
```diff
@@ -35,15 +45,23 @@ namespace CoarsenIO{
...
+        const int j,
+        const int k,
+        const int comp
+    )
     {
+        using namespace amrex::literals;
+
+        AMREX_ASSERT(sc[0] == 0);
+        AMREX_ASSERT(sc[1] == 0);
+        AMREX_ASSERT(sc[2] == 0);
```
ax3l: Remove invalid assert:
```suggestion
```

EZoni:
@ax3l We discussed with @jlvay and @WeiqunZhang and had the following ideas:
- The old `CoarsenIO` operation could be renamed `sample`;
- The old `CoarsenMR` operation could be renamed `average` (as in the AMReX `AverageDown`) or `interpolate`.

Let's discuss here more if you have any comments or concerns on these suggestions.

P.S. Again, if we want the I/O one to be specialized for cell-centered output, there is the option of simplifying the formulas, by setting the corresponding output index type to zero in the code, and then calling it `sample_to_cell_centers` or something along these lines.

ax3l:
Thanks a lot - I updated the PR accordingly and it's now ready for review with the new names :)

EZoni:
Copying here what I commented on Slack.

Would it be possible and would it make sense to consider removing one intermediate layer? Like doing something like this:
- rename `namespace ablastr::coarsen::sample` as `namespace ablastr:sample_with_coarsening`
- rename `namespace ablastr::coarsen::average` as `namespace ablastr::average_with_coarsening`

(or similar names), which keeps the information about "coarsening" in the name but removes one layer in terms of structure?

Then one would have things like
`ablastr::sample_with_coarsening::Coarsen` or  `ablastr::sample_with_coarsening::Interp`
instead of things like (current code)
`ablastr::coarsen::sample::Coarsen` or `ablastr::coarsen::sample::Interp`.

What do you think?

ax3l:
Thanks a ton! Discussed between Edoardo, Remi & Axel: we'll keep the current naming for now since it mirrors the directory structure.
I'll rebase. - Done.

Source/Diagnostics/WarpXIO.cpp
```diff
@@ -13,11 +13,11 @@
 #endif
 #include "FieldIO.H"
 #include "Particles/MultiParticleContainer.H"
-#include "Utils/CoarsenIO.H"
 #include "Utils/TextMsg.H"
 #include "Utils/WarpXProfilerWrapper.H"
 #include "WarpX.H"
 
+#include <ablastr/coarsen/sample.H>
```
EZoni: Do you think we can remove this, since it doesn't seem to be used in the file?
```suggestion
```

Source/Parallelization/WarpXComm.cpp
```diff
@@ -13,13 +13,14 @@
 #   include "BoundaryConditions/PML_RZ.H"
 #endif
 #include "Filter/BilinearFilter.H"
-#include "Utils/CoarsenMR.H"
+#include "Utils/Parser/IntervalsParser.H"
```
EZoni: Was this added on purpose?
```suggestion
```

EZoni (commented):
Thank you, @ax3l! This looks great. I just left two minor comments, see below.

Source/Diagnostics/WarpXIO.cpp
ax3l: Thanks, true!

Source/Parallelization/WarpXComm.cpp
ax3l: I think it was missing, but could have been prior to rebase. Will remove, too - thx!

EZoni (approved):
Great, ready to merge! I think I'll open an issue to keep track of the fact that it would be good to add some documentation of these interpolation methods (sampling, averaging), including a link to the slides attached in the PR description, as you suggested some time ago.

ax3l:
Good idea! You could copy the slide content in a new theory sub-section of the manual.

PR #3436, opened 2022-09-29
Adding an if statement for the last step diagnostic

### LINKED ISSUE #3432: Possible PICMI bug: `sim.step(max_steps)` and `sim.step(1)` in a `for loop` produces different number of diagnostics
When testing a PICMI script, I found the two different ways of running the simulation **lead to different number of output diagnostics**.
This script executes the simulation by `sim.step(max_steps)`, which produce outputs in the `diags` every 10 time steps as expected.
[picmi_maxstep.txt](https://github.com/ECP-WarpX/WarpX/files/9669076/picmi_maxstep.txt)
However, the script below, after changing to `sim.step(1)` in a for loop, produces outputs at every time steps, even if the output period is set to `10`.
[picmi_step.txt](https://github.com/ECP-WarpX/WarpX/files/9669085/picmi_step.txt)

In addition, **`sim.step(max_steps)` produces a warning `rho is 0` that is expected at step 1, but `sim.step(1)` produces this warning at every time steps**.

```
STEP 1 starts ...


**** WARNINGS ******************************************************************
* GLOBAL warning list  after  [ FIRST STEP ]
*
* --> [!  ] [ElectrostaticSolver] [raised once]
*     Max norm of rho is 0
*     @ Raised by: ALL
*
********************************************************************************

STEP 1 ends. TIME = 1.5e-11 DT = 1.5e-11
Evolve time = 0.0899578 s; This step = 0.0899578 s; Avg. per step = 0.0899578 s

STEP 2 starts ...
```

However, `sim.step(1)` produces this warning at every step.
```
STEP 1 starts ...


**** WARNINGS ******************************************************************
* GLOBAL warning list  after  [ FIRST STEP ]
*
* --> [!  ] [ElectrostaticSolver] [raised once]
*     Max norm of rho is 0
*     @ Raised by: ALL
*
********************************************************************************

STEP 1 ends. TIME = 1.5e-11 DT = 1.5e-11
Evolve time = 0.0804025 s; This step = 0.0804025 s; Avg. per step = 0.0804025 s
--- INFO    : Writing plotfile diags/diag_part000001
--- INFO    : Writing plotfile diags/diag_field000001

STEP 2 starts ...


**** WARNINGS ******************************************************************
* GLOBAL warning list  after  [ FIRST STEP ]
*
* --> [!  ] [ElectrostaticSolver] [raised once]
*     Max norm of rho is 0
*     @ Raised by: ALL
*
********************************************************************************

STEP 2 ends. TIME = 3e-11 DT = 1.5e-11
Evolve time = 0.0981157 s; This step = 0.0177132 s; Avg. per step = 0.0981157 s
--- INFO    : Writing plotfile diags/diag_part000002
--- INFO    : Writing plotfile diags/diag_field000002

STEP 3 starts ...


**** WARNINGS ******************************************************************
* GLOBAL warning list  after  [ FIRST STEP ]
*
* --> [!  ] [ElectrostaticSolver] [raised once]
*     Max norm of rho is 0
*     @ Raised by: ALL
*
********************************************************************************

STEP 3 ends. TIME = 4.5e-11 DT = 1.5e-11
```
The good news is they seems to produce the same output results.
Potential obtained by `sim.step(max_steps)`:
phi_all10
Potential obtained by `sim.step(1)`:
phi_step10

@ax3l @dpgrote Do you have any ideas how to fix it?

Yin-YinjianZhao:
This PR is a quick fix that closes #3432 .


Source/Evolve/WarpXEvolve.cpp
```diff
@@ -370,7 +370,9 @@ WarpX::Evolve (int numsteps)
 
         // End loop on time steps
     }
-    multi_diags->FilterComputePackFlushLastTimestep( istep[0] );
+    if (istep[0] >= max_step || cur_time >= stop_time) {
+        multi_diags->FilterComputePackFlushLastTimestep( istep[0] );
```
dpgrote, line 366: Thanks for the PR! Perhaps add a comment here that this check is needed for the Python version, which allows the Evolve routine to be called multiple times.

dpgrote:
I notice that one of the CI tests is failing. Looking into it, the problem is that the `max_steps` parameter is not set, it is not passed in as an input parameter. What this means is that when running from Python, with the check added, the final diagnostic will never be written. In the case that is failing, the period of the diagnostic is set to 0, so the diagnostic is never written out, causing the failure.

I think the way to fix this correctly, is to have this PR also make the change so that the max_steps parameter is passed in as input. I can clone your repo and push the needed change. I think it will be just a few lines of Python.

Source/Evolve/WarpXEvolve.cpp
Yin-YinjianZhao: Thanks, just added a comment, maybe the PR is ready.
@dpgrote @ax3l

dpgrote (dismissed):
This looks good to me - I can approve it. However, it would be good if someone else double checks my change to the picmi interface.

Yin-YinjianZhao:
> This looks good to me - I can approve it. However, it would be good if someone else double checks my change to the picmi interface.

I ran this version using my old picmi inputs and didn't see any issues.

Python/pywarpx/picmi.py
```diff
@@ -1145,16 +1145,11 @@ def initialize_warpx(self, mpi_comm=None):
             return
 
         self.warpx_initialized = True
-        pywarpx.warpx.init(mpi_comm)
+        pywarpx.warpx.init(mpi_comm, max_step=self.max_steps, stop_time=self.max_time)
```
ax3l, line 1587: Let's not add this in the `init` method but as regular inputs parameters via ParamParse?

There are so many ways to control how to stop a sim, this seems overly specific.

dpgrote: These are the two parameters that WarpX uses to stop the simulation. They are passed in here and then added to to the list of arguments that are passed to the call to `amrex_init` which adds them to the ParmParse database of parameters. If a Python user has some other means for stopping the simulation, they would leave the max_steps and max_time unset, in which case these parameters would not be added to the argument list.

ax3l (approved):

ax3l:
It looks like this breaks existing Python-API tests, i.e., `embedded_boundary_python_API`

dpgrote:
Ah yes. That input file tries to take two time steps even though it is setting max_step = 1. This should presumably be allowed. I would suggest that if the user specifies a number of time steps when calling `step` then max_step (and stop_time) be ignored. Note that this changes the `if` test of whether to call the last diagnostic flush to be `if (istep[0] == max_step || (stop_time -  - 1.e-3*dt[0] <= cur_time && cur_time < stop_time + dt[0])` so that it is only done once.

ax3l:
> I would suggest that if the user specifies a number of time steps when calling step then max_step (and stop_time) be ignored.

That sounds reasonable because explicit. But on the other hand, our "end simulation" logic so far follows the principle "when at least one of the abort criteria is met, then abort". So maybe it's easier to keep it that way? I am not sure.

ax3l:
Ah, detail: `when calling step()` ...

ax3l:
@Yin-YinjianZhao can you take a look at the failing test and update it accordingly please ? :)

Yin-YinjianZhao:
@ax3l I thought you guys had a solution and agreement on how to fix it, but I didn't get it?
How do you want to fix it?

dpgrote:
My suggested change above should fix it.

Yin-YinjianZhao:
@dpgrote You mean `if (istep[0] == max_step || (stop_time - 1.e-3*dt[0] <= cur_time && cur_time < stop_time + dt[0]))`?

dpgrote:
Correct. The `- -` was a typo.

Yin-YinjianZhao:
@dpgrote But it seems like the test still fails.

ax3l:
Is it possible that you actually need to update `Examples/Modules/embedded_boundary_python_API/analysis.py` to read ` == 1` now? After `step(1)`, have a dump for 0 and 1, don't we?

dpgrote:
I had forgotten that line 73 in Source/Evolve/WarpXEvolve.cpp needs to be changed. Now it is
```
        numsteps_max = std::min(istep[0]+numsteps, max_step);
```
The `min` needs to be remove so it becomes
```
        numsteps_max = istep[0] + numsteps;
```
What is happening in the test case is that when the second step command is called, that `min` constrains `numsteps_max` and so the second step doesn't happen (and hence there is not the second diagnostic file).

EZoni:
CI should be fixed now, following @ax3l's suggestion in https://github.com/ECP-WarpX/WarpX/pull/3436#issuecomment-1295266185. 
Will just need to be rebased one more time as soon as #3520 and #3521 are merged.

dpgrote:
I looked at this and wondered about the change made in the CI test case. There is still a problem - in that test case, with the second call to `sim.step`, a time step is not being done. I traced it to the `stop_time` variable. The input file is not setting it, but it is being set in the routine `ComputeMaxStep`, setting to essentially `stop_time = max_steps*dt`. Is this needed, having `max_step` and `stop_time` be consistent? It doesn't seem like it since the evolve loop is checking both.

dpgrote (approved):
It all looks good now. Thanks!

PR #3441, opened 2022-09-30
BTD : remove redundant filterParticle call

RevathiJambunathan:


Source/Diagnostics/ComputeDiagFunctors/BackTransformParticleFunctor.cpp
```diff
@@ -126,7 +126,6 @@ BackTransformParticleFunctor::operator () (PinnedMemoryParticleContainer& pc_dst
                 const auto& ptile_src = particles.at(index);
                 auto src_data = ptile_src.getConstParticleTileData();
                 // Flag particles that need to be copied if they cross the z-slice
-                // setting this to 1 for testing (temporarily)
                 amrex::ParallelFor(np,
```
RevathiJambunathan, line 129: The call to amrex::filterParticles would do the same operation as the parallel for and it would also copy the selected particles to the dst pc. We transform the particles and then copy them in the next parallelfor

RemiLehe:
It seems that the tests are failing.
It now occurs to me that `filterParticles` is maybe also copying quantities that are not Lorentz-transformed (e.g. the weight of particles) and which are therefore not copied as part of `GetParticleLorentzTransform`.

Source/Diagnostics/ComputeDiagFunctors/BackTransformParticleFunctor.cpp
atmyers: I agree you don't need this here because you're doing the filter and transform operations manually. If you want, you could save some lines of code by using the `filterAndTransformParticles` function in amrex: https://github.com/AMReX-Codes/amrex/blob/development/Src/Particle/AMReX_ParticleTransformation.H#L478

RevathiJambunathan:
> It seems that the tests are failing. It now occurs to me that `filterParticles` is maybe also copying quantities that are not Lorentz-transformed (e.g. the weight of particles) and which are therefore not copied as part of `GetParticleLorentzTransform`.

@RemiLehe Thank you for noticing this. I will add a copy for the weight as well in the operator for Transform. 
and check if this passes. Then will try the more elegant `filterAndTransformParticles` call in amrex.

RevathiJambunathan:
@RemiLehe we actually do copy the weight of the particles in the LorentzTransform operator. 

https://github.com/ECP-WarpX/WarpX/blob/7d3dab494a605caa8edf775bf71984d8e062e4c5/Source/Diagnostics/ComputeDiagFunctors/BackTransformParticleFunctor.H#L160

RevathiJambunathan:
@RemiLehe you are correct in the the lorentz transform operator does not copy integers, run time attributes, etc. 
I am going to close this PR and the amrex::filter call is necessary and may be follow up with a better way to implement `filterParticleAndTransform` method in amrex.

PR #3445, opened 2022-10-03
[WIP] Avoid side-effect: Undo the scaling of rho in the `computePhi`

RemiLehe:
When calling `computePhi`, we modify the array of `rho` in-place.
This PR avoids this side-effect.

### DIFF
--- a/Source/ablastr/fields/PoissonSolver.H
+++ b/Source/ablastr/fields/PoissonSolver.H
@@ -161,7 +161,7 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
     // scale rho appropriately; also determine if rho is zero everywhere
     amrex::Real max_norm_b = 0.0;
     for (int lev=0; lev<=finest_level; lev++) {
-        rho[lev]->mult(-1._rt/PhysConst::ep0); // TODO: when do we "un-multiply" this? We need to document this side-effect!
+        rho[lev]->mult(-1._rt/PhysConst::ep0);
         max_norm_b = amrex::max(max_norm_b, rho[lev]->norm0());
     }
     amrex::ParallelDescriptor::ReduceRealMax(max_norm_b);
@@ -317,6 +317,11 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
                 post_phi_calculation.value()(mlmg, lev);
 
     } // loop over lev(els)
+
+    // Undo the scaling of rho
+    for (int lev=0; lev<=finest_level; lev++) {
+        rho[lev]->mult(-PhysConst::ep0);
+    }
 }
 
 } // namespace ablastr::fields
--- a/Source/ablastr/fields/PoissonSolver.H
+++ b/Source/ablastr/fields/PoissonSolver.H
@@ -161,7 +161,6 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
     // scale rho appropriately; also determine if rho is zero everywhere
     amrex::Real max_norm_b = 0.0;
     for (int lev=0; lev<=finest_level; lev++) {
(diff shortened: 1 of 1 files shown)


Source/ablastr/fields/PoissonSolver.H
```diff
@@ -317,6 +317,11 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
                 post_phi_calculation.value()(mlmg, lev);
 
     } // loop over lev(els)
+
+    // Undo the scaling of rho
+    for (int lev=0; lev<=finest_level; lev++) {
```
dpgrote: This is not always necessary since the rho passed in is often a temporary. Perhaps add a flag to computePhi whether to reset rho and have it default to true. Then the calling routine can turn it off when not needed and save a little bit of time.

Along the same line, the initial scaling done above is not always needed either since sometimes rho is zero (for example in AddBoundaryField) and it could be skipped as well to save even more time.

Source/ablastr/fields/PoissonSolver.H
```diff
@@ -161,7 +161,7 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
     // scale rho appropriately; also determine if rho is zero everywhere
     amrex::Real max_norm_b = 0.0;
     for (int lev=0; lev<=finest_level; lev++) {
-        rho[lev]->mult(-1._rt/PhysConst::ep0); // TODO: when do we "un-multiply" this? We need to document this side-effect!
+        rho[lev]->mult(-1._rt/PhysConst::ep0);
```
ax3l: Proposal: let's remove this here, make `rho` a `amrex::Vector<amrex::MultiFab const*> const & rho` named `source` and explicitly scale up before we call `computePhi` (and undo it afterwards in case we did not pass a temporary)?

dpgrote: The advantage of doing the scaling here is that it reduces code duplication for most use cases.

RemiLehe:
Thanks for the feedback! 
I updated this PR to include the `-epsilon_0` directly into the definition of the operator that we apply on `phi`.

RemiLehe:
OK, that was a bad idea: it seems that the MLMG solver diverges in this case:
```
MLMG: Initial rhs               = 6.56305248e-05
MLMG: Initial residual (resid0) = 6.56305248e-05
MLMG: Iteration   1 Fine resid/bnorm = 5.799033105e+54
```

Source/ablastr/fields/PoissonSolver.H
```diff
@@ -226,12 +225,14 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
         // one of the axes of the grid, i.e. that only *one* of the
         // components of `beta` is non-negligible. // we use this
 #if defined(WARPX_DIM_RZ)
-        linop.setSigma({0._rt, 1._rt-beta_solver[1]*beta_solver[1]});
+        linop.setSigma({
+            -PhysConst::ep0,
```
PhilMiller, line 229: If you're intending to multiply through by `ep0`, this looks like a mistake - the first entry would still be 0, wouldn't it?

WeiqunZhang:
https://github.com/AMReX-Codes/amrex/pull/2986

WeiqunZhang:
> OK, that was a bad idea: it seems that the MLMG solver diverges in this case:
> 
> ```
> MLMG: Initial rhs               = 6.56305248e-05
> MLMG: Initial residual (resid0) = 6.56305248e-05
> MLMG: Iteration   1 Fine resid/bnorm = 5.799033105e+54
> ```

@RemiLehe What test is this?

WeiqunZhang:
Also, we might need to adjust the absolute tolerance, because the solver compares that with the residual.  Now the residual has been scaled by a factor.

PR #3447, opened 2022-10-03
Docs: Fix PICMI Builds

ax3l:
Unbreak the RTD build. We need to do this all relative to the root directory.

### DIFF
--- a/Docs/README.md
+++ b/Docs/README.md
@@ -14,7 +14,7 @@ python3 -m pip install -r Docs/requirements.txt
 
 ### Compiling the documentation
 
-`cd` into this directory and type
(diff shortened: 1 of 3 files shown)


dpgrote:
What is the reason for this change? It looks like the only difference is to make sure that the `Docs/requirements.txt` doesn't get confused with the `requirements.txt` in the top directory.

ax3l:
The reason is that ReadTheDocs can only call the requirements file from the root directory. So we need to be relative to the source root.
Screenshot from 2022-10-03 11-38-52

https://readthedocs.org/projects/warpx/builds/18230914/

dpgrote (approved):
Ok, looks good.

Issue #3377, opened 2022-09-09
Invalid memory access when moving window and timers-based load-balancing is used


lucafedeli88:
I am opening this issue because I have observed an invalid memory access when moving window and load-balancing based on timers are used in combination.

Here I provide a small reproducer:
```
#################################
####### GENERAL PARAMETERS ######
#################################
max_step = 10
amr.n_cell =  64 64 64
amr.max_grid_size = 32
amr.blocking_factor = 32
amr.max_level = 0
geometry.dims = 3
geometry.prob_lo = -10.e-6   -10.e-6   -10.e-6    # physical domain
geometry.prob_hi =  10.e-6    10.e-6    10.e-6

algo.load_balance_intervals = 3::100
algo.load_balance_with_sfc = 0
algo.load_balance_costs_update = timers

warpx.do_moving_window = 1
warpx.moving_window_dir = z
warpx.moving_window_v = 1.0
warpx.start_moving_window_step = 2


#################################
####### Boundary condition ######
#################################
boundary.field_lo = pml pml pml
boundary.field_hi = pml pml pml

#################################
############ NUMERICS ###########
#################################
warpx.verbose = 1
warpx.cfl = 0.99

# Order of particle shape factors
algo.particle_shape = 3

#################################
############ PLASMA #############
#################################
particles.species_names = electrons

electrons.species_type = electron
electrons.injection_style = "NUniformPerCell"
electrons.num_particles_per_cell_each_dim = 1 1 2
electrons.profile = constant
electrons.density = 1.e25  # number of electrons per m^3
electrons.momentum_distribution_type = "gaussian"
electrons.ux_th  = 0.01 # uth the std of the (unitless) momentum
electrons.uy_th  = 0.01 # uth the std of the (unitless) momentum
electrons.uz_th  = 0.01 # uth the std of the (unitless) momentum
```

When WarpX runs this inputfile (even without GPUs or OMP support), `valgrind` detects the following issue:

```
STEP 3 starts ...
==41155== Invalid read of size 4
==41155==    at 0x55CFBD: Add<float> (AMReX_GpuAtomic.H:584)
==41155==    by 0x55CFBD: WarpX::shiftMF(amrex::MultiFab&, amrex::Geometry const&, int, int, int, float, bool, amrex::ParserExecutor<3> const&) (WarpXMovingWindow.cpp:435)
[...]

PR #3450, opened 2022-10-05
Fix Invalid Mem w/ Moving Window, LB & PMLs

lucafedeli88:
This PR is a workaround to fix an invalid memory access when moving window is used together with timers-based load balancing.
See issue https://github.com/ECP-WarpX/WarpX/issues/3377


Source/Utils/WarpXMovingWindow.cpp
```diff
@@ -428,7 +428,9 @@ WarpX::shiftMF (amrex::MultiFab& mf, const amrex::Geometry& geom,
             dstfab(i,j,k,n) = srcfab(i+shift.x,j+shift.y,k+shift.z,n);
         })
 
-        if (cost && WarpX::load_balance_costs_update_algo == LoadBalanceCostsUpdateAlgo::Timers)
+        if (cost &&
+            mfi.index() < cost->size() && //Workaround to fix invalid memory accesses when mf is a PML
```
ax3l: attn @atmyers @WeiqunZhang : PMLs have a different domain decomposition than regular MFs. How can we handle costs for them (see: `&(*cost)[mfi.index()]` below)?

atmyers: One thought is - do we actually want to track costs for the PML grids? Usually, the local cost of the PML update is small compared to other costs in the simulation, and in particular to the cost of the parallel copy from the PML grids to the parent grids (maybe this is different for PSATD or maybe there are future planned PML implementations that will be expensive, though?). This function could simply check whether it's working on PML grids and skip updating the costs if so. 

Another is that currently, we never actually apply load balancing separately to the PML grids. The default behavior is to assign the PML box to whatever rank the closest parent grid box is assigned to, in order to minimize communication costs. So if you do want include the pml cost somewhere, if `WarpX::do_similar_dm_pml = 1`, perhaps it makes the most sense to give it to whatever box the pml is "following". 

If `WarpX::do_similar_dm_pml = 0`, the PML will be load balanced using the default SFC anyway, so don't use the costs anyway. If, in the future, we do want to load balance the PML grids separately, then we'd need to track a separate set of pml_costs, defined on the pml grids, and use them in the load balancing step, but that would be a task for a separate, non-bugfix PR.

In summary, I think it's probably fine just to skip updating the costs for PML grids. If you want to do a little better than that, you could skip them if `WarpX::do_similar_dm_pml = 0` and include them in the "parent" box if "WarpX::do_similar_dm_pml = 1", but I expect that the improvement in the cost accuracy won't be significant, at least for FDTD.

ax3l: > This function could simply check whether it's working on PML grids and skip updating the costs if so.
> ... In summary, I think it's probably fine just to skip updating the costs for PML grids.

I think that would be a good solution for now. @lucafedeli88 can you please update accordingly once you are back from Japan? :)

ax3l: ping @lucafedeli88 :)

lucafedeli88: Sure! Sorry... I'll try to do that now...

lucafedeli88: @ax3l , would this be OK as a temporary workaround ?

ax3l: Thank you, that LGTM - @atmyers do you finalize the review and merge? :)

atmyers (approved):

PR #3451, opened 2022-10-05
openPMD: Shape of Constant Particle Records

ax3l:
We forgot to set the shape (particle number) of constant particle records, which then defaulted to `1`. This violates the openPMD standard and complicates post-processing.

First seen in openPMD-api's `.to_df()` pandas function.

This affects particle:
- mass
- charge

Issue introduced in #2980


RemiLehe (approved):

n01r (approved):
Great find!

PR #3453, opened 2022-10-06
[WIP] Handled expanded particle ID range in Python 

atmyers:
Convert to long and int, respectively, when getting the particle id and cpu numbers through the Python interface. This is handled by a couple of numpy ufuncs that can convert from the bit-packed representation to the id and cpu.

### DIFF
--- a/Python/pywarpx/_libwarpx.py
+++ b/Python/pywarpx/_libwarpx.py
@@ -143,13 +143,13 @@ class LibWarpX():
         # our particle data type, depends on _ParticleReal_size
         _p_struct = (
             [(d, self._numpy_particlereal_dtype) for d in 'xyz'[:self.dim]]
-            + [('id', 'i4'), ('cpu', 'i4')]
+            + [('idcpu', 'u8')]
         )
         self._p_dtype = np.dtype(_p_struct, align=True)
 
         _numpy_to_ctypes = {}
         _numpy_to_ctypes[self._numpy_particlereal_dtype] = c_particlereal
-        _numpy_to_ctypes['i4'] = ctypes.c_int
+        _numpy_to_ctypes['u8'] = ctypes.c_uint64
 
         class Particle(ctypes.Structure):
             _fields_ = [(v[0], _numpy_to_ctypes[v[1]]) for v in _p_struct]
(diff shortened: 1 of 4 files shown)


Python/pywarpx/amrex_particle_id_utils.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include "numpy/ufuncobject.h"
+#include "numpy/npy_3kcompat.h"
+#include <math.h>
+
+static PyMethodDef GetIDMethods[] = {
+    {NULL, NULL, 0, NULL}
+};
+
+static long uint64_t_to_long_id (uint64_t idcpu) {
+    long r = 0;
```
atmyers, line 13: TODO - call amrex functions for this so we don't duplicate this logic

Python/setup.py
```diff
@@ -61,5 +63,12 @@
       package_data = package_data,
       install_requires = ['numpy', 'picmistandard==0.0.20', 'periodictable'],
       python_requires = '>=3.7',
+      ext_modules=[
+          Extension(
```
dpgrote, line 67: This extension should only be defined when the installation is including the compiled library, i.e. when `args.with_libwarpx` or `args.with_lib_dir or PYWARPX_LIB_DIR` is true. When those are not true, the setup.py installs a pure Python package. You can do (after line 54)

```
ext_modules = []
if package_data:
    ext_modules.append(Extension(...))
```

and then, `ext_modules = ext_modules` in the call to setup.

Python/pywarpx/_libwarpx.py
```diff
@@ -882,22 +882,24 @@ def get_particle_z(self, species_name, level=0):
...
-        Return a list of numpy arrays containing the particle 'id'
-        positions on each tile.
+        Return a list of numpy arrays containing the particle id
+        numbers on each tile.
 
         '''
+        from .amrex_particle_id_utils import get_id
         structs = self.get_particle_structs(species_name, level)
-        return [struct['id'] for struct in structs]
+        return [get_id(struct['idcpu']) for struct in structs]
```
ax3l, line 891: Wouldn't it be  simpler if you skipped the Python extension stuff, implemented a clean C++ version of the two int->int functions in WarpX as a helper function and just exposed that to Python via `libwarpx.so`?

ax3l: Side note: in pyAMReX and thus the future we kicked off today, we could declutter id and cpu by reinterpreting the idcpu as two 32bit numpy ints.
https://github.com/AMReX-Codes/pyamrex/blob/13f46bb79b3e07647b9c369f59525ef00054a6f6/src/Particle/ArrayOfStructs.cpp#L45-L75

Python/setup.py
ax3l: I would not go via extensions, this complicates the deployment and deps logic too much.
https://github.com/ECP-WarpX/WarpX/pull/3453#discussion_r989608087

PR #3456, opened 2022-10-07
Add SymPy Notebook to Derive PSATD Equations

EZoni:
This adds the SymPy jupyter notebook `psatd.ipynb` in the new directory `Tools/Algorithms`. 

The notebook derives the PSATD equations by integrating analytically the first-order Maxwell equations (with the additional fields F and/or G, if the div(E) and/or div(B) cleaning options are activated). This derivation is different than the one currently used in the literature and in the WarpX implementation, which is based on the second-order Maxwell equations.

The notebook also handles the time dependency of J and rho (constant, linear, or quadratic in time). The execution time varies depending on the complexity of the equations to be integrated.

**Note**
The code is currently implemented as a jupyter notebook, not sure whether this is handled well by git (?). The "Files changed" tab does not display the notebook well. If you select "View file" separately, it is instead displayed as a notebook.

To-do:
- [x] Add automated verification of solution of linear system of ODEs


ax3l:
@EZoni that is awesome, thanks a lot.

Can you please clean the cell output before checking in notebooks? We clone WarpX on the fly to build ImpactX and other codes, so keeping the source code repo small is quite important for developer productivity. The current file adds 733 KB to every clone.

ax3l:
Oh, as discussed: feel free to add a separate `requirements.txt` in the same dir as the notebook.

EZoni:
@ax3l Thank you for your comments. Cell output cleared. Regarding `requirements.txt`, I didn't add anything because I'm not importing anything special for this particular notebook. So far I have been running the notebook with SymPy 1.9, but I'm not sure if that should be listed as a "basic dependency" in some `requirements.txt` file. In other words, I'm using very basic SymPy methods, so I guess other older versions of SymPy would be compatible, too. Let me know if this makes sense to you.

ax3l:
Ypo, `sympy` is a bit heavy in a more central `requirements.txt` file of WarpX, but for `Tools/Algorithms` dirs it would be legit. Given the clarity of the notebook import, it's not super urgent to add here

ax3l (approved):

Tools/Algorithms/psatd.ipynb
```diff
@@ -0,0 +1,631 @@
...
+    "    dim += 1\n",
+    "if divB_cleaning:\n",
+    "    dim += 1\n",
+    "\n",
+    "# Define symbols for physical constants\n",
+    "c = sp.symbols(r'c', real=True, positive=True)\n",
+    "mu0 = sp.symbols(r'\\mu_0', real=True, positive=True)\n",
+    "\n",
+    "# Define symbols for time variables\n",
+    "s = sp.symbols(r's', real=True, positive=True)\n",
```
RemiLehe, line 106: Could you add a comment explaining what `s` is here?

Tools/Algorithms/psatd.ipynb
```diff
@@ -0,0 +1,631 @@
...
+    "# and related additional terms in the first-order time derivative of the magnetic field\n",
+    "if divB_cleaning:\n",
+    "    dBx_dt += I*c**2*G*kx\n",
+    "    dBy_dt += I*c**2*G*ky\n",
+    "    dBz_dt += I*c**2*G*kz\n",
+    "    dG_dt = I*(kx*Bx+ky*By+kz*Bz)\n",
+    "\n",
+    "# Define array of first-order time derivatives of the electric and magnetic fields\n",
+    "dE_dt = Matrix([[dEx_dt], [dEy_dt], [dEz_dt]])\n",
+    "dB_dt = Matrix([[dBx_dt], [dBy_dt], [dBz_dt]])"
```
RemiLehe, line 179: It seems that these variables are not used and could be removed.

Tools/Algorithms/psatd.ipynb
```diff
@@ -0,0 +1,631 @@
...
+    "\n",
+    "def simple_mat(W):\n",
+    "    \"\"\"\n",
+    "    Simplify matrix W.\n",
+    "    \"\"\"\n",
+    "    for i in range(W.shape[0]):\n",
+    "        for j in range(W.shape[1]):\n",
+    "            W[i,j] = W[i,j].expand().simplify()\n",
+    "            \n",
+    "def limit(f, x, l):\n",
```
RemiLehe: I think that this function could be removed altogether, as it is not used in the rest of the notebook, it seems.

Tools/Algorithms/psatd.ipynb
```diff
@@ -0,0 +1,631 @@
...
+    "# Define array of first-order time derivatives of all fields\n",
+    "fields_list = [dEx_dt, dEy_dt, dEz_dt, dBx_dt, dBy_dt, dBz_dt]\n",
+    "if divE_cleaning:\n",
+    "    fields_list.append(dF_dt)\n",
+    "if divB_cleaning:\n",
+    "    fields_list.append(dG_dt)\n",
+    "dEBFG_dt = zeros(dim, 1)\n",
+    "for i in range(dEBFG_dt.shape[0]):\n",
+    "    dEBFG_dt[i] = fields_list[i]\n",
+    "dEBFG_dt = dEBFG_dt.expand()\n",
```
RemiLehe, line 239: Just checking: is this line needed here?

Tools/Algorithms/psatd.ipynb
```diff
@@ -0,0 +1,631 @@
...
+    "        M[i,j] = dEBFG_dt[i].coeff(EBFG[j], 1)\n",
+    "print(r'M = ')\n",
+    "display(M)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Diagonalization of $W_1 = M (t-t_n) = P_1 D_1 P_1^{-1}$:"
```
RemiLehe, line 324: Do we really need to calculate the diagonalization of $W_1$, $W_2$ and $W_3$?
It seems that diagonalizing $M$ only would be enough (and then you would multiply the resulting diagonal matrix by $(t-t_n)$, $t$ or $s$ to get the diagonalization of $W_1$, $W_2$ and $W_3$.

Tools/Algorithms/psatd.ipynb
```diff
@@ -0,0 +1,631 @@
...
+    "display(P3)\n",
+    "print('D3 = ')\n",
+    "display(D3)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### General solution (homogeneous system):\n",
```
RemiLehe, line 390: Could you move this `markdown` cell up, so that it appears before the diagonalization code?
I think it would help the reader understand why we are doing the diagonalization.

Tools/Algorithms/psatd.ipynb
EZoni: This one yes: for some reason some of the elements of the matrix M, computed by calling the SymPy method `coeff` from each element of `dEBFG_dt`, are wrong (zero, instead of non-zero) if I do not expand `dEBFG_dt` before calling `coeff`.

Tools/Algorithms/psatd.ipynb
EZoni: That's right, thank you. Removed in 71e5bae.

Tools/Algorithms/psatd.ipynb
EZoni: It is true that these variables are not used in the notebook. However, I think we can expect this notebook to be used in an interactive way (namely, by running individual cells and checking results) and I think it is useful/handy to have access to the first time derivatives of E and B independently through `dE_dt` and `dB_dt`, rather than having access to them only through the more general array `dEBFG_dt`. This would be my rationale for keeping these variables available to the user.

Tools/Algorithms/psatd.ipynb
EZoni: Thank you. Comment added in ff98459, let me know if it isn't clear enough.

Tools/Algorithms/psatd.ipynb
EZoni: Yes, thank you. Done in b26e018.

Tools/Algorithms/psatd.ipynb
EZoni: Thank you for the suggestion. I just added in e6e1df8 and 66fa38e an automated verification of the solution of the linear system of ODEs. Now I'll be able to make changes like the ones you suggested while making sure that I do not introduce any bugs. Will try your suggestions and report on the results soon.

EZoni: Done in 86b93d4: we now diagonalize only $M=P D P^{-1}$ and then use $P$ and $D$ for the diagonalization of $W_1$, $W_2$ and $W_3$.

EZoni:
@RemiLehe Thank you very much for your review! I think I have addressed all of your comments.

RemiLehe:
Thanks for updating the PR!

PR #3458, opened 2022-10-07
Updated 1D CI tests with particles because of update of yt

dpgrote:
This resets the benchmark data for several 1D tests. Previously, yt could not read in the particles from 1D plot data files so the benchmarks for the particle data was all zeros. A fix was made to yt (PR #4112) that allowed it to read in the particle data. With that fix, the checksum evaluation now reads in the particles and generates nonzero values, causing the benchmarks to fail. This PR updates the benchmark values to use the correct nonzero values.

This PR also updates the required version of yt to 4.1.0 (which includes the above mentioned PR).


RemiLehe (approved):
Thanks for fixing this!

PR #2872, opened 2022-02-17
openPMD: Add ADIOS2 Engine Parameter Control

kshitij-v-mehta:
Add support for setting ADIOS engine types and their parameters for openPMD. Similar to how ADIOS operators are setup.
Addresses #2866

PR #3460, opened 2022-10-10
openPMD Engines: Allow Params w/o Type

ax3l:
The current implementation only accepted parameters for ADIOS2 engines if the engine type was also specified. That is not necessary, the default (bp4) will usually be taken and thus the documentation did not work.

This fixes it, keeping the engine type optional in the inputs.

```
<diag_name>.adios2_engine.type = bp4  # now optional, as documented

<diag_name>.adios2_engine.parameters.NumAggregators = 2048
<diag_name>.adios2_engine.parameters.BurstBufferPath="/mnt/bb/username"
```

Follow-up to @kshitij-v-mehta's #2872 :)

cc @guj I think you saw that in recent tests with burst buffers as well (engine params were ignored)?
cc @franzpoeschel just fyi

### DIFF
--- a/Source/Diagnostics/WarpXOpenPMD.cpp
+++ b/Source/Diagnostics/WarpXOpenPMD.cpp
@@ -181,34 +181,44 @@ namespace detail
                 op_block += R"END(,
           "parameters": {
 )END";
-            op_block += op_parameters + "}";
+            op_block += op_parameters +
+                        "\n          }";
         }
             op_block += R"END(
         }
       ]
     })END";
-        if (!engine_type.empty())
-            op_block += ",";
-
+            if (!engine_type.empty() || !en_parameters.empty())
+                op_block += ",";
         }  // end operator string block
 
         // add the engine string block
-        if (!engine_type.empty()) {
+        if (!engine_type.empty() || !en_parameters.empty())
+        {
             en_block = R"END(
-    "engine": {
-      "type": ")END";
-            en_block += engine_type + "\"";
+    "engine": {)END";
 
+            // non-default engine type
+            if (!engine_type.empty()) {
+                en_block += R"END(
+      "type": ")END";
+                en_block += engine_type + "\"";
+
+                if(!en_parameters.empty())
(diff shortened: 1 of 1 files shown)


kshitij-v-mehta:
Overall, it is fine to have a default engine. However, if a specified parameter is not applicable for the default engine (e.g. a parameter for the SST engine doesn't apply to the BP engine), the parameter is silently ignored. Would it instead make sense to enforce an engine type if parameters are specified?

guj:
I think enforcing engine type and parameter better come from ADIOS.
  
> Overall, it is fine to have a default engine. However, if a specified parameter is not applicable for the default engine (e.g. a parameter for the SST engine doesn't apply to the BP engine), the parameter is silently ignored. Would it instead make sense to enforce an engine type if parameters are specified?

guj (commented):
Looks like with this change,  if there is no engine parameters, the operators will still be ignored. 
(e.g. if a user wants to use default engine and just want to use a compression operator.) 

Is this intended?

ax3l:
> Looks like with this change, if there is no engine parameters, the operators will still be ignored.

I cannot follow. With this change, you do not need to specify `adios2_engine.type` to have `adios2_engine.parameters` honored. before that, they were ignored unless `adios2_engine.type` was specified.

Update: clarified on Slack.
This PR is about engine params, not operator params (which include compression). The latter work as expected.

ax3l:
> Overall, it is fine to have a default engine. However, if a specified parameter is not applicable for the default engine (e.g. a parameter for the SST engine doesn't apply to the BP engine), the parameter is silently ignored. Would it instead make sense to enforce an engine type if parameters are specified?

I did a couple of tests with this patch yesterday. By default, it looks like BP4 is picked as default engine.

Even if we enforce an engine type, we have no mechanism to validate the parameters for it. These warnings (or errors) need to be thrown on the ADIOS2 side if we pass invalid parameters for the current engine.

RemiLehe (approved):

PR #3463, opened 2022-10-12
Correct particle positions outside refined injection

RemiLehe:
The current code for the refined injection was using the refinement ratio (`lrref`) in order to determine the position of macroparticles within one cell **even outside of the refined injection region**.

This lead to slightly incorrect/asymmetrical macroparticle positions, as shown in this plot, which corresponds to the same case as our CI test `RefinedInjection`:
index

which in turn leads to incorrect charge density `rho`:
index

This PR fixes the issue, and updates the CI test. In particular, I added an extra test that checks the uniformity of `rho`. This test fails with the current dev branch, but passes with this PR.


prkkumar:
I tested this with a mesh refinement simulation with `refine_plasma` turned ON. It fixes the artifact that I was seeing before. Here is the ion density plot before this PR:
Screen Shot 2022-10-12 at 9 07 54 AM
And here it is after this PR:
Screen Shot 2022-10-12 at 9 07 16 AM

Examples/Physics_applications/laser_acceleration/analysis_refined_injection.py
```diff
@@ -32,7 +32,7 @@
 
 # the number of coarse particle streams
 # (odd because one that exactly hits the fine patch is not refined)
```
RemiLehe:
```suggestion
```

Examples/Physics_applications/laser_acceleration/analysis_refined_injection.py
```diff
@@ -50,5 +49,15 @@
 
 assert( np == np_expected )
 
+# Test uniformity of rho, by taking a slice of rho that
+# crosses the edge of the refined injection region
+# (but is ahead of the mesh refinement patch)
+ds.force_periodicity()
```
RemiLehe, line 55: Strangely, I get an error in this code if I don't add `ds.force_periodicity()`:
```
RuntimeError: yt attempted to read outside the boundaries of a non-periodic domain along dimension 1.
```

atmyers:
Nice catch!

atmyers (approved):

PR #3467, opened 2022-10-14
ABLASTR: Always Sync Option

ax3l:
Add an option to ABLASTR/WarpX to always sync nodal cells that are owned by multiple blocks. This can be used for debugging.


Docs/source/usage/parameters.rst
```diff
@@ -2840,3 +2814,41 @@ This is essentially the python slicing syntax except that the stop is inclusive
...
+
+* ``warpx.serialize_initial_conditions`` (`0` or `1`) optional (default `0`)
+    Serialize the initial conditions for reproducible testing, e.g, in our continuous integration tests.
+    Mainly whether or not to use OpenMP threading for particle initialization.
+
+* ``warpx.safe_guard_cells`` (`0` or `1`) optional (default `0`)
+    Run in safe mode, exchanging more guard cells, and more often in the PIC loop (for debugging).
+
+* ``ablastr.fillboundary_always_sync`` (`0` or `1`) optional (default `0`)
+    Run all ``FillBoundary`` operations on ``MultiFab`` to force-synchronize shared nodal points points.
```
EZoni: Just a typo:
```suggestion
    Run all ``FillBoundary`` operations on ``MultiFab`` to force-synchronize shared nodal points.
```

EZoni (commented):
This PR will be very useful, thank you @ax3l! Looks good to me, I just left one minor comment on a typo in the docs.

I'd wait a day or two before merging, just to give other developers the chance to chime in, if they have suggestions. 

Also, just one question: did you try running some input file with the new option activated? Shall we turn it ON in, say, one CI test (even if not crucial for the particular test), just to make sure that we don't break this feature in the future?

ax3l:
> did you try running some input file with the new option activated?

Not yet tested in detail :)

> Shall we turn it ON in, say, one CI test (even if not crucial for the particular test), just to make sure that we don't break this feature in the future?

Not generally, because we want to catch errors in CI that come from missing syncs. But we could turn it on for one tests just to make sure it does something.

Update by @EZoni: new option added to CI test `pml_psatd_dive_divb_cleaning` in 6bf4fa4.

Source/ablastr/utils/Communication.cpp
```diff
@@ -60,27 +62,18 @@ void ParallelAdd(amrex::MultiFab &dst, const amrex::MultiFab &src, int src_comp,
...
-                  amrex::IntVect ng,
-                  bool do_single_precision_comms,
-                  const amrex::Periodicity &period,
-                  const bool nodal_sync)
-{
-    BL_PROFILE("ablastr::utils::communication::FillBoundary");
+    // allow developers to always enforce nodal sync
+    bool do_nodal_sync = nodal_sync.value_or(false);
+    amrex::ParmParse pp_ablastr("ablastr");
+    pp_ablastr.query("fillboundary_always_sync", do_nodal_sync);
```
ax3l: Update me: this needs to be fixes so that `fillboundary_always_sync = 0` does not become an "always no-sync"

ax3l: Thx @RemiLehe

EZoni (approved):

PR #3472, opened 2022-10-20
[mini] Fix unused import warnings in python analysis scripts

NeilZaim:
These warnings were issued by the lgtm automated check in #3298


EZoni (approved):
Thanks for cleaning this up, @NeilZaim!

PR #3474, opened 2022-10-21
Transition to pyAMReX

ax3l:
Transitions our Python bindings to [pyAMReX](https://github.com/AMReX-Codes/pyamrex).

## Tasks

- [x] build system
  - [x] ensure multi-dim builds continue to work #3933
- [x] add callback hooks; we can use [this](https://pybind11.readthedocs.io/en/stable/advanced/cast/functional.html) to avoid fiddling with low-level C function pointers: https://github.com/ax3l/WarpX/pull/4
- [x] add functions exposed previously in `WarpXWrappers`
- [x] update (and pot. rename) `_libwarpx.py`
- [x] update logic that called into `_libwarpx.py`

Follow-ups:
- `setup.py` pip logic:
  - [ ] rename `PYWARPX_LIB_DIR` to `WARPX_PYMODULE_DIR`?
  - [ ] rename pip package: `pywarpx` -> `warpx`
- [x] fix `WarpX (Unknown)` in init output msg (already an issue in `development`) #4218
- [ ] add MPI comms to pyAMReX, then re-enable `Python_pass_mpi_comm`
- [x] remove `Source/Python/WarpXWrappers.H/cpp` #4233
- [ ] remove `_libwarpx.py` indirection level
- [ ] SoA particle transition for particle position & id/cpu access on GPU - will enable zero-copy cupy support for particle positions & ids #3850


.github/workflows/hip.yml
```diff
@@ -107,7 +107,7 @@ jobs:
           -DWarpX_DIMS=2              \
           -DWarpX_COMPUTE=HIP         \
           -DWarpX_EB=ON               \
-          -DWarpX_LIB=ON              \
+          -DWarpX_PYTHON=ON           \
           -DWarpX_MPI=ON              \
           -DWarpX_OPENPMD=ON          \
           -DWarpX_PRECISION=DOUBLE    \
```
ax3l, line 114: Here and in docs, `PYWARPX_LIB_DIR` probably needs to generally be updated to
```
PYWARPX_LIB_DIR=$PWD/build/lib/site-packages/pywarpx/
```

RemiLehe:
Discussed with @ax3l and @dpgrote :

- The `WarpX` Python object currently has a `multifab` method, that returns a `pyamrex.MultiFab` object
- In `pyamrex`, we will add a method `.to_global_view` to the `pyamrex.MultiFab` object, which returns an object of type `MFGlobalView`, which is very similar to the current `_MultiFabWrapper`. In particular, it would have the global `__setitem__` and `__getitem__`.

A user that would like to globally set values on a multifab would do:
```
Ex = w.multifab('Ex')
Ex.to_global_view()[1:10, 2:30, 1:10] = 0
```

Eventually, this `MFGlobalView` could be moved to the `pyamrex` source code.

RemiLehe:
Similarly, for the particles, we would use `pyamrex.ParticleContainer` for low-level, performant access, and have a `.to_global_view()` method that would return an object that is similar to the current `PGroup` object. 

The methods `get_particle_arrays`, `get_particle_structs`, etc... would probably be removed, as the user can instead use the methods of the `pyamrex.ParticleContainer`.

Python/pywarpx/_libwarpx.py
```diff
@@ -358,44 +141,21 @@ def _get_boundary_number(self, boundary):
...
-        return np.frombuffer(buf, dtype=dtype, count=size)
-
     def getNProcs(self):
         '''
 
         Get the number of processors
 
         '''
-        return self.libwarpx_so.warpx_getNProcs()
+        return self.libwarpx_so.getNProcs()
```
RemiLehe, line 118: @ax3l I wrapped the function `amrex::ParallelDescriptor::NProcs` in the `pybind11` module for WarpX (defined in `pyWarpX.cpp`). However, could we instead call this function directly from `pyamrex`?

Examples/Tests/particle_data_python/PICMI_inputs_2d.py
```diff
@@ -130,7 +130,7 @@ def add_particles():
         w=w, newPid=newPid, unique_particles=args.unique
     )
 
-callbacks.installbeforestep(add_particles)
+sim.extension.libwarpx_so.add_python_callback( "installbeforestep", add_particles )
```
RemiLehe: @ax3l @roelof-groenewald @dpgrote With the new Python wrappings, this seems to be a valid way to setup the callback functions, but is maybe a bit cumbersome (due to `sim.extension.libwarpx_so.`). Is there maybe a simpler way that I am overlooking?

roelof-groenewald: I just pushed a change to restore the old API.

roelof-groenewald: My thinking in refactoring things were that maybe we change the API to `callbacks.install('beforestep', add_particles)`, but we can decide on that later.

Python/pywarpx/callbacks.py
```diff
@@ -316,299 +322,129 @@
...
-    _beforeInitEsolve.uninstallfuncinlist(f)
-def isinstalledbeforeInitEsolve(f):
-    "Checks if the function is called before the initial E solve"
-    return _beforeInitEsolve.isinstalledfuncinlist(f)
+    installcallback('beforeInitEsolve', f)
 
 # ----------------------------------------------------------------------------
 def callfromafterInitEsolve(f):
-    installafterInitEsolve(f)
+    installafterInitEsolve('afterInitEsolve', f)
```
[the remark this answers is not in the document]
roelof-groenewald:
```suggestion
    installcallback('afterInitEsolve', f)
```

Python/pywarpx/fields.py
```diff
@@ -279,7 +279,7 @@ def __getitem__(self, index):
             raise Exception('Too many indices given')
 
         # Expand the indices to length 3
-        ii = self._get_indices(index, None)
+        ii = self._get_indices(index, slice(None))
```
roelof-groenewald: @dpgrote I've been trying to get the test in `Examples/Physics_applications/capacitive_discharge/PICMI_inputs_1d.py` to run, but I encountered an error in the call to `fields.py` to grab `rho` due to this line. The error was:
```
  File "/home/roelof/scratch/build/warpx/warpx_pybind/lib/site-packages/pywarpx/fields.py", line 179, in _find_start_stop
    iistop = ii + 1
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'
```
which is circumvented by this change but you might prefer a different fix (perhaps dealing with the None in `_find_start_stop()`.

With this fix the test unfortunately still doesn't run due to errors in the part that writes to `phi`.

Examples/Tests/particle_data_python/PICMI_inputs_2d.py
RemiLehe: Sounds good. Thanks!

Python/pywarpx/fields.py
dpgrote: This fix is good, thanks! I also just did a push to make the same fix for the `__setitem__`. Can you check if this fixes the problem when writing `phi`?

Python/pywarpx/_libwarpx.py
dpgrote: To avoid code duplication, I think it would be better to rely on the function from pyamrex. And the same for `MyProc` - this should be added to pyamrex.

dpgrote: I went ahead and made a PR in pyamrex to add `MyProc` and some other routines. https://github.com/AMReX-Codes/pyamrex/pull/146

Python/pywarpx/fields.py
roelof-groenewald: Sorry, I should've been more clear. With that same fix in the `__setitem__` function there is still an error in the part that writes to `phi`:
```
  File "PICMI_inputs_1d.py", line 105, in _run_solve
    fields.PhiFPWrapper(0, True)[...] = self.phi
  File "/home/roelof/scratch/build/warpx/warpx_pybind/lib/site-packages/pywarpx/fields.py", line 390, in __setitem__
    mf_arr[block_slices] = value3d[global_slices]
IndexError: too many indices for array: array is 1-dimensional, but 3 were indexed
```

roelof-groenewald: In this specific case `mf_arr` has shape (131, 1, 1, 1) while in the current execution `value3d` (i.e. `phi`) has shape (131). The test runs if I change in the input script
```
fields.PhiFPWrapper(0, True)[...] = self.phi[:]
```
to
```
fields.PhiFPWrapper(0, True)[...] = self.phi[:,None,None,None]
```
But is this what we would want?

roelof-groenewald: Also, just to let you know, I'm seeing that `__getitem__` gives an array of the correct length but always with all the values in the array the same (at least in the 1d case I'm working on).

dpgrote: I just pushed some changes that should fix this issues with the get and set item methods. Also, for convenience, I added the `shape` attribute, similar to the  numpy shape, but here it returns the shape of the global array.

Python/pywarpx/_libwarpx.py
```diff
@@ -478,7 +233,8 @@ def getistep(self, level=0):
             The refinement level to reference
         '''
 
-        return self.libwarpx_so.warpx_getistep(level)
+        warpx = self.libwarpx_so.get_instance()
+        return warpx.getistep(level)
```
roelof-groenewald: Code of this sort should be,
```suggestion
        return self.warpx.getistep(level)
```
right?

Python/pywarpx/fields.py
roelof-groenewald: Thanks! Looks like the 1d test case is happy now :slightly_smiling_face:

Source/Python/pyWarpX.cpp
```diff
@@ -55,6 +56,7 @@ PYBIND11_MODULE(PYWARPX_MODULE_NAME, m) {
     )pbdoc";
 
     // note: order from parent to child classes
+    init_WarpXParIter(m);
```
ax3l: Let's move this after we know the PC, one line down, because we have it as an argument to the pti constructor.

Source/Python/WarpXParticleContainer.cpp
```diff
@@ -29,5 +43,43 @@ void init_WarpXParticleContainer (py::module& m)
...
+        //         const int, const long, const long,
+        //         const int, const int, const int
+        //     )>(&WarpXParticleContainer::DepositCharge),
+        //     py::arg("pti"), py::arg("wp"), py::arg("ion_lev"),
+        //     py::arg("rho"), py::arg("icomp"),
+        //     py::arg("offset"), py::arg("np_to_depose"),
+        //     py::arg("thread_num"), py::arg("lev"), py::arg("depos_lev")
+        // )
+        .def("deposit_charge",
+            [](WarpXParticleContainer& pc, WarpXParIter& pti,
```
ax3l: Let's try to replace the `&` with `*` for these types to avoid copies.

roelof-groenewald: This has been resolved by changing the approach to do the loop in C++ code.

roelof-groenewald:
The following tests are still failing.
1d:
- [x] `Python_background_mcc_1d` and
- [x] `Python_background_mcc_1d_tridiag` due to rho collected from fields.py not matching the benchmark
2d and 3d:
- [ ] Tests timing out due to longer compilation time (and 2h limit on pipeline). For example in `development` the `Langmuir_multi_2d_MR` test takes 46.465 s to compile but in this PR it takes 772.696 s. 
- [ ] `Python_id_cpu_read`
RZ:
- [x] `Python_Langmuir_rz_multimode` - E-field error too large
- [x] `Python_magnetostatic_eb_rz` - E-field error too large
EB:
- [ ] `Python_magnetostatic_eb_3d` - E-field error too large
- [x] `Python_particle_scrape` - boundary buffer runtime int component not showing up in the bound container
- [ ] `embedded_boundary_python_API` - need to transition `get_mesh_edge_lengths` to `fields.py`

(cc @ax3l @dpgrote )

ax3l:
Latest test list:
- [x] Python_wrappers:
  - `return warpx.multifab(f'{self.mf_name}[level={self.level}]')` -> `RuntimeError: The MultiFab 'f_fp[level=0]' is unknown or is not allocated!`
  - Just renamed?
- [ ] Python_id_cpu_read:
  - `get_particle_id`  -> `NotImplementedError()`
  - to implement together with `warpx_convert_id_to_long` and `warpx_convert_cpu_to_int`
- [x] Python_pass_mpi_comm:
  - `_libwarpx.py", line 141, in amrex_init` -> `Exception: mpi_comm argument not yet supported`
  - not yet added in pyAMReX
  - not used in practice yet, we will re-activate later

ax3l:
Ah, we now need to include these changes by @lucafedeli88 from #4104.
I'll push an update following https://pybind11.readthedocs.io/en/stable/advanced/classes.html#custom-constructors

ax3l:
Ok, only a Windows DLL loading issue left and this test:
- `Python_wrappers` in `Examples/Tests/python_wrappers`
```
  File "PICMI_inputs_2d.py", line 259, in <module>
    check_values(1013263608.6369569, Ex[:,:], rtol, atol)
  File "PICMI_inputs_2d.py", line 253, in check_values
```

ax3l:
I think `__setitem__` in `_MultiFABWrapper` manipulates only a copy due to the `.T`
Update: no, `.T` returns a view :)

EZoni:
@ax3l @dpgrote 

Regarding the test `Python_wrappers`, there seems to be an issue related to domain decomposition. The checksum values (coded by hand in Python) seem to match between this PR and `development` when the test is set up on one single box (by modifying `max_grid_size` and `blocking_factor` in the PICMI script). I checked this on the first failing value, the one computed for the `Ex` field.

Another observation is that, in the multi-box case (the one where this PR does not match `development`), the shape of `Ex[:,:]` seems to be the same for both branches, which I guess is a good thing. However:
- in `development`, I cannot do, say, `Ex[-9,:]` because "Dimension 0 lower index is out of bounds" (which makes sense, we have only `8` ghost cells and the lower index bound is `-8`);
- in this branch, I can do `Ex[-9,:]` and I can even use indices lower than `-9`, not sure why and/or if that is intended.

Will try to collect more info and investigate further.

ax3l:
@roelof-groenewald @dpgrote @RemiLehe ready for merge :)

Python/pywarpx/_libwarpx.py
```diff
@@ -82,7 +61,7 @@ def load_library(self):
 
         if 'libwarpx_so' in self.__dict__:
             raise RuntimeError(
-                "Invalid attempt to load libwarpx_so library multiple times."
+                "Invalid attempt to load libwarpx_so... library multiple times."
```
lucafedeli88: I am not sure that the meaning of these three dots is clear here.

Python/pywarpx/_libwarpx.py
```diff
@@ -464,9 +181,15 @@ def finalize(self, finalize_mpi=1):
         Call finalize for WarpX and AMReX. Registered to run at program exit.
 
         '''
+        # TODO: simplify, part of pyAMReX already
         if self.initialized:
-            self.libwarpx_so.warpx_finalize()
-            self.libwarpx_so.amrex_finalize(finalize_mpi)
+            del self.warpx
+            # The call to warpx_finalize causes a crash - don't know why
+            #self.libwarpx_so.warpx_finalize()
```
lucafedeli88, line 190: What is the error message?

Python/pywarpx/fields.py
```diff
@@ -101,833 +144,604 @@ def mesh(self, direction):
...
-        elif self.dim == 3:
-            return index[0], index[1], index[2]
+    def _get_field(self, mfi):
+        """Return the field at the given mfi.
+        If include ghosts is true, return the whole array, otherwise
+        return the interior slice that does not include the ghosts.
+        """
+        # Note that the array will always have 4 dimensions.
+        # even when self.dim < 3.
+        # The transpose is taken since the thing returned by self.mf.array(mfi) is in C ordering.
```
lucafedeli88:
```suggestion
        # The transpose is taken since the array returned by self.mf.array(mfi) is in C ordering.
```
?

lucafedeli88 (dismissed):
Thanks a lot for this monumental work! I've left only three small comments.

Python/pywarpx/_libwarpx.py
ax3l: This segfaults. Let me attempt to remove some logic from our constructor - I bet it operates on an already removed instance.

Generally, in Python, we want to have no heavy logic in destructors and thus I move it to `Finalize()`.

Python/pywarpx/_libwarpx.py
ax3l: Thx, will rephrase :)

Source/Python/WarpX.cpp
```diff
@@ -146,6 +147,29 @@ void init_WarpX (py::module& m)
             },
             py::arg("potential")
         )
+
+        .def("set_plasma_lens_strength",
+            [](WarpX& wx, int i_lens, amrex::Real strength_E, amrex::Real strength_B) {
+                auto & mypc = wx.GetPartContainer();
```
dpgrote: Since this is affecting only `mypc`, should this go into `Source/Python/Particles/MultiParticleContainer.cpp` instead?

Python/pywarpx/_libwarpx.py
ax3l: Ok, I tried reworking this and this will need more work later on. Too many cross-dependencies between living Python objects and the instance destruction order for now.

Source/Python/WarpX.cpp
ax3l: Good point!  cc @RemiLehe fyi, I will move it

ax3l: Done! :)

roelof-groenewald (approved):
Whoohoo!!

ax3l:
Let's ship it! 😁

PR #3475, opened 2022-10-22
Add reduced diagnostics to `picmi.py`

roelof-groenewald:
Allow reduced diagnostics to be set up through Python.

Also created a new parent class `WarpXDiagnosticBase` that handles functionality that is common to all diagnostics in order to reduce code duplication.


Examples/Tests/reduced_diags/PICMI_inputs_loadbalancecosts.py
```diff
@@ -0,0 +1,90 @@
...
+        density=1e14, rms_velocity=[0]*3,
+        upper_bound=[xmax, ymax, 1.0]
+    )
+)
+layout = picmi.GriddedLayout(
+    n_macroparticle_per_cell=[1, 1, 1], grid=grid
+)
+
+# Reduced diagnostic
+reduced_diag = picmi.ReducedDiagnostic(
```
dpgrote, line 53: This is a picky comment you can ignore if you wish, but there is inconsistent formatting in the use of spaces around equals signs, and with one argument per line.

Python/pywarpx/Diagnostics.py
```diff
@@ -7,6 +7,7 @@
 from .Bucket import Bucket
 
 diagnostics = Bucket('diagnostics', _diagnostics_dict={})
+reduced_diagnostics = Bucket('warpx', _diagnostics_dict={})
```
dpgrote, line 10: I notice the inconsistency in the instance name argument, but this comes from the naming used in the C++ parm parsing, using `diagnostics.diag_names` and `warpx.reduced_diag_names`. I don't know which prefix is better, but it would be good to use the same prefix (though probably should be a different PR).

Python/pywarpx/picmi.py
```diff
@@ -2033,3 +2027,114 @@ def initialize_inputs(self):
...
+
+class ReducedDiagnostic(picmistandard.base._ClassWithInit, WarpXDiagnosticBase):
+    """
+    Sets up a reduced diagnostic in the simulation.
+
+    See `Input Parameters <https://warpx.readthedocs.io/en/latest/usage/parameters.html#reduced-diagnostics>`_
+    for more information.
+
+    Parameters
+    ----------
```
dpgrote, line 2040: There is the reference above, but I think it would still be helpful to list the parameter documentation here as well, with very brief descriptions and typing.

dpgrote (approved):
This looks good overall, though I have a couple of minor comments that are not essential.

Python/pywarpx/Diagnostics.py
roelof-groenewald: I agree using the diagnostic prefix for both would be nicer, but, as you said, probably a follow-up PR.

ax3l (approved):

ax3l:
x-ref @RTSandberg to touch base on #3179

PR #3476, opened 2022-10-22
Move shared functionality between `picmi.UniformDistribution` and `picmi.AnalyticDistribution` into a parent class

roelof-groenewald:
Just a quick PR to allow a density function to be used in Python initialized simulations.

### DIFF
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -384,7 +384,7 @@ class DensityDistributionBase(object):
             species.__setattr__(f'momentum_function_u{sdir}(x,y,z)', f'({expression})/{constants.c}')
 
 
-class UniformDistribution(DensityDistributionBase, picmistandard.PICMI_UniformDistribution):
+class UniformDistribution(picmistandard.PICMI_UniformDistribution, DensityDistributionBase):
     def initialize_inputs(self, species_number, layout, species, density_scale):
 
         self.set_mangle_dict()
@@ -397,7 +397,7 @@ class UniformDistribution(DensityDistributionBase, picmistandard.PICMI_UniformDi
             species.density *= density_scale
 
 
-class AnalyticDistribution(DensityDistributionBase, picmistandard.PICMI_AnalyticDistribution):
+class AnalyticDistribution(picmistandard.PICMI_AnalyticDistribution, DensityDistributionBase):
     def initialize_inputs(self, species_number, layout, species, density_scale):
 
         self.set_mangle_dict()
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -342,11 +342,19 @@ class UniformDistribution(picmistandard.PICMI_UniformDistribution):
         species.zmin = self.lower_bound[2]
         species.zmax = self.upper_bound[2]
 
-        # --- Only constant density is supported at this time
-        species.profile = "constant"
-        species.density = self.density
-        if density_scale is not None:
-            species.density *= density_scale
+        # --- Check if the density is a simple number
+        if isinstance(self.density, float):
+            species.profile = "constant"
+            species.density = self.density
+            if density_scale is not None:
+                species.density *= density_scale
+        else:
+            species.profile = "parse_density_function"
+            if density_scale is not None:
+                raise AttributeError(
+                    "Cannot use `density_scale` with a parsable density function."
+                )
+            species.__setattr__('density_function(x,y,z)', self.density)
 
         # --- Note that WarpX takes gamma*beta as input
         if np.any(np.not_equal(self.rms_velocity, 0.)):
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -322,12 +322,13 @@ class GaussianBunchDistribution(picmistandard.PICMI_GaussianBunchDistribution):
 
 
 class DensityDistributionBase(object):
-    """This is a base class for several predefined density distributions, meant
-    to capture universal initialization logic."""
-    def init(self, kw):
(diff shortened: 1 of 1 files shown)


dpgrote:
The `parse_density_function` can already be specified using the `AnalyticDistribution`.

roelof-groenewald:
> The `parse_density_function` can already be specified using the `AnalyticDistribution`.

Oh wow, I missed that entirely. Thanks for pointing it out to me!
But I'm wondering now if there isn't room for improvement here to avoid others from having the same confusion. I think the issue comes in because the C++ input 'injection style' is `NUniformPerCell` referring to a uniform number of macro-particles per cell, whereas the picmi `UniformDistribution` refers to a uniform density. I guess that is actually pretty clear in itself so probably fine to keep things as is.

roelof-groenewald:
I realized (after @dpgrote pointed its existence out to me, thanks!) that there is a large amount of shared logic between `UniformDistribution` and `AnalyicDistribution`. So now this PR has changed to a simple code clean-up where that shared logic is moved into a shared parent class.

Python/pywarpx/picmi.py
```diff
@@ -402,34 +406,6 @@ def initialize_inputs(self, species_number, layout, species, density_scale):
         else:
             species.__setattr__('density_function(x,y,z)', "{}*({})".format(density_scale, expression))
```
roelof-groenewald, line 412: This could lead to issues in cases where people are using the parsers ability to calculate sequential variable values, for example:
`density_function(x,y,z) = "r=sqrt(x**2+y**2);1e15*(r < 50e-6)"`
Maybe we should remove the `density_scale` parameter when a density function is used?

Python/pywarpx/picmi.py
```diff
@@ -341,14 +342,16 @@ def initialize_inputs(self, species_number, layout, species, density_scale):
...
-        species.profile = "constant"
-        species.density = self.density
-        if density_scale is not None:
-            species.density *= density_scale
+        if self.fill_in:
+            species.do_continuous_injection = 1
 
         # --- Note that WarpX takes gamma*beta as input
-        if np.any(np.not_equal(self.rms_velocity, 0.)):
+        if (hasattr(self, "momentum_expressions")
```
dpgrote, line 360: This all looks ok. However, I might imagine that a user would look at this code and decide to hack something like this
```
dist = picmi.UniformDistribution(...)
dist.momentum_expressions = ['a', 'b', 'c']
```

Python/pywarpx/picmi.py
dpgrote: Good point. This was written before that type of expression was allowed. However, the resulting expression like this is valid
`3*(r=sqrt(x**2+y**2);1e15*(r < 50e-6))`
I just confirmed a test of it.

Python/pywarpx/picmi.py
roelof-groenewald: We could add the `momentum_expressions` argument to `PICMI_UniformDistribution` to properly allow it. Would you like me to do that?

Python/pywarpx/picmi.py
roelof-groenewald: Oh cool! That is good to know. Thanks for testing.

RemiLehe:
@roelof-groenewald It seems that the test `Python_Langmuir_rz_multimode` does not pass ; would you be able to take a closer look?

roelof-groenewald:
@RemiLehe thanks for pointing that out. Yeap, I'll take a look later today and fix the failing test.

roelof-groenewald:
@RemiLehe I fixed the failing test. The current failure (CUDA / nVCC 11.0.2) I see is also failing on other PRs.

RemiLehe:
@roelof-groenewald Thanks! Would you have time to rebase this on `development`? This should fix the failing test and allow merging.

Python/pywarpx/picmi.py
```diff
@@ -432,6 +383,34 @@ def setup_parse_momentum_functions(self, species):
                 expression = f'{self.directed_velocity[idir]}'
             species.__setattr__(f'momentum_function_u{sdir}(x,y,z)', f'({expression})/{constants.c}')
 
+
+class UniformDistribution(DensityDistributionBase, picmistandard.PICMI_UniformDistribution):
```
dpgrote: For the documentation to work properly (i.e. for it to include the doc from the picmistandard class), the picmistandard class should be first in the list of base classes. You can check this by locally generating the documentation.

roelof-groenewald: That is very interesting, thanks for letting me know! I'll swap the order around.

Python/pywarpx/picmi.py
```diff
@@ -432,6 +383,34 @@ def setup_parse_momentum_functions(self, species):
...
+        self.set_species_attributes(species, layout)
+
+        # --- Only constant density is supported by this class
+        species.profile = "constant"
+        species.density = self.density
+        if density_scale is not None:
+            species.density *= density_scale
+
+
+class AnalyticDistribution(DensityDistributionBase, picmistandard.PICMI_AnalyticDistribution):
```
```suggestion
class AnalyticDistribution(picmistandard.PICMI_AnalyticDistribution, DensityDistributionBase):
```

Python/pywarpx/picmi.py
```suggestion
class UniformDistribution(picmistandard.PICMI_UniformDistribution, DensityDistributionBase):
```

RemiLehe:
@roelof-groenewald Thanks again for this PR. Would you be able to fix the failing CI tests?
(Regarding the failure for MacOS compilation, you can simply merge the `development` branch into this PR.)

dpgrote (approved):
This looks good. Thanks!

PR #3477, opened 2022-10-22
Allow arbitrary laser antenna normal in picmi

roelof-groenewald:
The `LaserAntenna` class in `picmi.py` is hardcoded for an antenna normal vector (0, 0, 1). This PR extends the class to allow other antenna normal directions.

### DIFF
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -1057,10 +1057,26 @@ class AnalyticLaser(picmistandard.PICMI_AnalyticLaser):
 class LaserAntenna(picmistandard.PICMI_LaserAntenna):
     def initialize_inputs(self, laser):
         laser.laser.position = self.position  # This point is on the laser plane
+        if not np.allclose(laser.laser.direction, [0, 0, 1]):
+            if not np.allclose(laser.laser.direction, self.normal_vector):
+                raise AttributeError(
+                    'The specified laser direction does not match the '
+                    'antenna normal'
+                )
         laser.laser.direction = self.normal_vector  # The plane normal direction
         if isinstance(laser, GaussianLaser):
-            laser.laser.profile_focal_distance = laser.focal_position[2] - self.position[2]  # Focal distance from the antenna (in meters)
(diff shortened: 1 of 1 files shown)


Python/pywarpx/picmi.py
```diff
@@ -1056,10 +1056,26 @@ def initialize_inputs(self):
 class LaserAntenna(picmistandard.PICMI_LaserAntenna):
     def initialize_inputs(self, laser):
         laser.laser.position = self.position  # This point is on the laser plane
+        if not np.allclose(laser.laser.direction, [0, 0, 1]):
```
dpgrote: Is this first check necessary? Presumably, the laser direction and antenna normal should be the same even if the laser direction is [0,0,1], correct?

roelof-groenewald: I wrote it in this way since a user might only specify the normal direction of the antenna in which case the laser direction would be along z (the default), so in that case the laser direction is just overwritten. I'm not sure what the best approach would be since requiring the direction to be specified twice doesn't seem ideal. What are your thoughts?

dpgrote: I understand the concern about duplicating parameters. Though, I prefer things to be explicitly written out, so forcing both the laser direction and antenna normal to be supplied would be ok. This seems to be an issue with the picmi standard which has the duplicated input parameters. Do any codes support having the laser direction not be normal to the antenna plane? The better fix would be to change the picmi standard and remove the default values for the laser parameters, and make the antenna normal an optional argument, defaulting to the laser direction.

roelof-groenewald: That makes sense to me. So I'll change this to always check that the laser direction and antenna normal is the same (if antenna normal is given) and then open a PR in `picmistandard` to remove the default laser direction.

ax3l: > Do any codes support having the laser direction not be normal to the antenna plane?

This could be useful for a few use cases (cc with @RemiLehe):
- oblique incidence angles for solid targets with small sim boxes
- travel-wave thomson pulses or flying focus init on the boundaries of a sim

roelof-groenewald:
@dpgrote as per the discussion at the developers meeting on Tuesday, are you happy with this PR as is?

dpgrote (approved):
This looks good to me!

Issue #2009, opened 2021-06-09
amrex.the_arena_is_managed=0


ax3l:
Since WarpX runs w/o managed memory, we might want to change the default of `amrex.the_arena_is_managed` to `0`.

At least on Windows, managed memory also creates performance regressions.

ax3l:
At the moment, running WarpX with `amrex.the_arena_is_managed=0` segfaults on Summit:

```
 3: ./warpx.3d.MPI.CUDA.DP.OPMD() [0x101bb24c]
    amrex::PODVector<double, amrex::ArenaAllocator<double> >::push_back(double const&) at ??:?

 4: ./warpx.3d.MPI.CUDA.DP.OPMD() [0x1068dfa0]
    LaserParticleContainer::InitData(int) at ??:?

 5: ./warpx.3d.MPI.CUDA.DP.OPMD() [0x1068cd00]
    LaserParticleContainer::InitData() at ??:?

 6: ./warpx.3d.MPI.CUDA.DP.OPMD() [0x105e1978]
    MultiParticleContainer::InitData() at ??:?

 7: ./warpx.3d.MPI.CUDA.DP.OPMD() [0x1057af90]
    WarpX::InitFromScratch() at ??:?

 8: ./warpx.3d.MPI.CUDA.DP.OPMD() [0x1057aa34]
    WarpX::InitData() at ??:?

 9: ./warpx.3d.MPI.CUDA.DP.OPMD() [0x10704254]
    main at ??:?
```

ax3l:
With #2018 applied, the next segfault is triggered also in init from:
```
 3: ./warpx.3d.MPI.CUDA.DP.OPMD() [0x101928f4]
    void amrex::unpackRemotes<amrex::ParticleContainer<0, 0, 4, 0, amrex::ArenaAllocator>, amrex::PODVector<char, amrex::ArenaAllocator<char> >, amrex::RedistributeUnpackPolicy, 0>(amrex::ParticleContainer<0, 0, 4, 0, amrex::ArenaAllocator>&, amrex::ParticleCopyPlan const&, amrex::PODVector<char, amrex::ArenaAllocator<char> >&, amrex::RedistributeUnpackPolicy&&) at ??:?

 4: ./warpx.3d.MPI.CUDA.DP.OPMD() [0x1019022c]
    amrex::ParticleContainer<0, 0, 4, 0, amrex::ArenaAllocator>::RedistributeGPU(int, int, int, int) at ??:?

[...]

PR #3480, opened 2022-10-25
[WIP] Turn off managed memory by default

RemiLehe:
The default in `amrex` is to have managed memory ON by default.
However, this can affect performance if in the code is inadvertently accessing GPU memory from the CPU. @WeiqunZhang also reported that this can in some cases lower the performance of GPU-aware MPI.

It is probably better to turn managed memory OFF by default.

### DIFF
--- a/Source/Initialization/WarpXAMReXInit.cpp
+++ b/Source/Initialization/WarpXAMReXInit.cpp
@@ -23,6 +23,8 @@ namespace {
         // https://amrex-codes.github.io/amrex/docs_html/GPU.html#inputs-parameters
         bool abort_on_out_of_gpu_memory = true; // AMReX' default: false
         pp_amrex.queryAdd("abort_on_out_of_gpu_memory", abort_on_out_of_gpu_memory);
+        bool the_arena_is_managed = false; // AMReX' default: true
+        pp_amrex.queryAdd("the_arena_is_managed", the_arena_is_managed);
 
         // Work-around:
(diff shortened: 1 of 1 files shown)


WeiqunZhang (approved):

RemiLehe:
I will run our automated test on GPU for this branch, to make sure that this does not break anything.

RemiLehe:
Hm, it's not going well...
The first three tests (`BTD_ReducedSliceDiag`, `Deuterium_Tritium_Fusion_3D` and `ElectrostaticSphere`) all crash with 
```
amrex::Abort::0::CUDA error 2 in file /tmp/ci-v5Kx7DGbiA/warpx/builddir/_deps/fetchedamrex-src/Src/Base/AMReX_Arena.cpp line 187: out of memory !!!
```

I think that the "out-of-memory" message is confusing and probably points instead to the CPU trying to access GPU memory, because I don't get those error messages when I enable managed memory (i.e. when I use the `development` branch).

RemiLehe:
This will require more work to track down the error.
(See also #2009 )

WeiqunZhang:
I think here is what happened.  The tests are MPI jobs with multiple processes using the same GPU.  AMReX's The_Arena on each process is by default trying to allocate 3/4 of the GPU memory.  Without managed memory to back it up, cudaMalloc will throw an OOM.  In principle, we cold do something to prevent this in amrex.  We will need to gather the device ids and figure out how many processes are using the same GPU.

WeiqunZhang:
We can reopen this after we make changes in amrex.

PR #2942, opened 2022-03-08
Fix stair-case solver in 2D Cartesian

RemiLehe:
The stair-case solver does not work properly for the `Ey` component in 2D Cartesian. This PR fixes this, both for the vacuum Maxwell solver and the macroscopic Maxwell solver.

PR #3481, opened 2022-10-25
[WIP] Fix behavior of PML with embedded boundaries in 2D

RemiLehe:
This is a follow-up on #2942 ; the update of the PML cells needs to be modified in a similar way as the valid cells were modified in #2942.


RemiLehe:
Superseded by https://github.com/BLAST-WarpX/warpx/pull/5574

PR #3484, opened 2022-10-26
Centralize the multi fab allocation

dpgrote:
This PR is based off of PR #3238, starting from its creation of the routine `AllocInitMultiFab` to handle common operations when allocating MultiFabs. This PR retains the call to `setval` though makes it optional (to keep the existing behavior). It also adds the `multifab_map` which maintains a map of pointers to all MultiFabs based on the field names. The ultimate goal of the map is to allow easy access to the MultiFabs from the Python interface, avoiding the need for a separate list of all MultiFabs. This avoids code duplication and will make any new MultiFabs automatically accessible from Python.

- [x] All MultiFabs and iMultiFabs are now updated.
- [x] RemakeMultiFab now calls the new function to generate the MultiFabs and iMultiFabs.


EZoni:
Just a quick question on something I noticed, even if the PR is still work in progress.

If I'm not mistaken, it seems that the following MultiFabs are allocated but not initialized with a value: `Bfield_<fp,cp>`, `Efield_<fp,cp>`, `Bfield_avg_<fp,cp>`, `Efield_avg_<fp,cp>`, `m_distance_to_eb`, `m_edge_lengths`, `m_face_areas`, `m_flag_info_face`, `m_flag_ext_face`, `m_area_mod`, `Venl`, `current_store`, `Bfield_aux` (when not aliased), `Efield_aux` (when not aliased), `Bfield_cax`, `Efield_cax`, `gather_buffer_masks`, `current_buf`, and `current_buffer_masks`.

Do you plan to initialize these too (with `0.0`), or is there a specific reason why you want them to be uninitialized?

dpgrote:
@EZoni In this PR, I want to keep the behavior of the code unchanged (as much as possible). Those MFs that are uninitialized here were originally uninitialized in the development branch (as far as I can tell). I don't feel strongly about this, so if there is consensus that all MFs should be initialized, then I can add that.

EZoni:
@dpgrote In general, I'd have a preference for initializing to zero all MultiFabs. If we want to see the effects of it on a case-by-case basis, which might be informative indeed, then I guess we could add the "new" ones in a separate PR. I'll let @ax3l and @WeiqunZhang chime in as well, when they'll review the PR.

ax3l:
I think it depends a little.
We often want to defer _allocation_, to only be done if a certain algorithm is truly used. For instance, EB iMultiFabs should not use memory if not used, but we can still already create the empty, unallocated MultiFab.
With respect to _initialization_ at allocation ([RAII](https://en.cppreference.com/w/cpp/language/raii)), there is an argument to make that it is not useful if it will be overwritten when used - but for safety a simple set at allocation is not too costly and fulfills then RAII safety.

Side note on RAII: does _not_ mean we must init on MultiFab creation - just when we allocate aka `define` it. Similar to `std::vector` creation with size 0 vs. resize.

Either way, I would do this separately from this PR if we want to unify this.

EZoni:
@dpgrote I see some of the EB tests are crashing, because of erroneous arithmetic operations according to the CI output:
- embedded_boundary_cube
- embedded_boundary_cube_2d
- embedded_boundary_cube_macroscopic
- embedded_boundary_rotated_cube
- embedded_boundary_rotated_cube_2d

Source/WarpX.cpp
```diff
@@ -2205,37 +2199,37 @@ WarpX::AllocLevelMFs (int lev, const BoxArray& ba, const DistributionMapping& dm
...
         if (n_current_deposition_buffer > 0) {
-            current_buf[lev][0] = std::make_unique<MultiFab>(amrex::convert(cba,jx_nodal_flag),dm,ncomps,ngJ,tag("current_buf[x]"));
-            current_buf[lev][1] = std::make_unique<MultiFab>(amrex::convert(cba,jy_nodal_flag),dm,ncomps,ngJ,tag("current_buf[y]"));
-            current_buf[lev][2] = std::make_unique<MultiFab>(amrex::convert(cba,jz_nodal_flag),dm,ncomps,ngJ,tag("current_buf[z]"));
+            AllocInitMultiFab(current_buf[lev][0], amrex::convert(cba,jx_nodal_flag),dm,ncomps,ngJ,tag("current_buf[x]"));
+            AllocInitMultiFab(current_buf[lev][1], amrex::convert(cba,jy_nodal_flag),dm,ncomps,ngJ,tag("current_buf[y]"));
+            AllocInitMultiFab(current_buf[lev][2], amrex::convert(cba,jz_nodal_flag),dm,ncomps,ngJ,tag("current_buf[z]"));
             if (rho_cp[lev]) {
-                charge_buf[lev] = std::make_unique<MultiFab>(amrex::convert(cba,rho_nodal_flag),dm,2*ncomps,ngRho,tag("charge_buf"));
+                AllocInitMultiFab(charge_buf[lev], amrex::convert(cba,rho_nodal_flag),dm,2*ncomps,ngRho,tag("charge_buf"), 0.0_rt);
```
EZoni: If we want to keep the behavior of the code unchanged, it looks like `charge_buf` is now initialized but was maybe uninitialized in the development branch? Saying this as I couldn't find other instances of `charge_buf` among the changes (namely the old initialization that would now appear as deleted), but maybe I overlooked it.

Source/WarpX.H
```diff
@@ -331,6 +332,43 @@ public:
...
+    // Map of all of the MultiFabs used (this can include MFs from other classes)
+    static std::map<std::string, amrex::MultiFab *> multifab_map;
+    static std::map<std::string, amrex::iMultiFab *> imultifab_map;
+
+    static void AddToMutiFabMap(const std::string name, const std::unique_ptr<amrex::MultiFab>& mf) {
+        multifab_map[name] = mf.get();
+    }
+
+    static void AddToMutiFabMap(const std::string name, const std::unique_ptr<amrex::iMultiFab>& mf) {
+        imultifab_map[name] = mf.get();
```
EZoni: I also see that these two functions might have a typo in the name, "Muti" instead of "Multi".

Source/WarpX.cpp
dpgrote: Initialization removed.

Source/WarpX.cpp
```diff
@@ -1851,100 +1854,94 @@ WarpX::AllocLevelMFs (int lev, const BoxArray& ba, const DistributionMapping& dm
...
+    AllocInitMultiFab(Bfield_fp[lev][0], amrex::convert(ba, Bx_nodal_flag), dm, ncomps, ngEB, tag("Bfield_fp[x]"));
+    AllocInitMultiFab(Bfield_fp[lev][1], amrex::convert(ba, By_nodal_flag), dm, ncomps, ngEB, tag("Bfield_fp[y]"));
+    AllocInitMultiFab(Bfield_fp[lev][2], amrex::convert(ba, Bz_nodal_flag), dm, ncomps, ngEB, tag("Bfield_fp[z]"));
 
-    Efield_fp[lev][0] = std::make_unique<MultiFab>(amrex::convert(ba,Ex_nodal_flag),dm,ncomps,ngEB,tag("Efield_fp[x]"));
-    Efield_fp[lev][1] = std::make_unique<MultiFab>(amrex::convert(ba,Ey_nodal_flag),dm,ncomps,ngEB,tag("Efield_fp[y]"));
-    Efield_fp[lev][2] = std::make_unique<MultiFab>(amrex::convert(ba,Ez_nodal_flag),dm,ncomps,ngEB,tag("Efield_fp[z]"));
+    AllocInitMultiFab(Efield_fp[lev][0], amrex::convert(ba, Ex_nodal_flag), dm, ncomps, ngEB, tag("Efield_fp[x]"));
+    AllocInitMultiFab(Efield_fp[lev][1], amrex::convert(ba, Ey_nodal_flag), dm, ncomps, ngEB, tag("Efield_fp[y]"));
+    AllocInitMultiFab(Efield_fp[lev][2], amrex::convert(ba, Ez_nodal_flag), dm, ncomps, ngEB, tag("Efield_fp[z]"));
```
EZoni, line 1871: @dpgrote I just tested this locally and noticed that the EB tests that currently fail (listed in https://github.com/ECP-WarpX/WarpX/pull/3484#issuecomment-1302786830) are all fixed if we initialize the MultiFabs `Bfield_fp` and `Efield_fp` to zero here.

I don't understand if this is a symptom of an EB bug that somehow is uncovered by the changes in this PR but does not appear in the main branch. The MultiFabs `Bfield_fp` and `Efield_fp` do not seem to be initialized to zero in the main branch, so not sure why this seems to be causing a problem only in this PR.

dpgrote: This is now fixed. In the development branch, `Bfield_aux` is always being zeroed, but here `Bfield_aux` was only zeroed when a new MF is created and not when it is aliased (from `Bfield_fp`). The fix is to add a call to `setVal` when doing the alias.

Source/BoundaryConditions/PML_RZ.cpp
```diff
@@ -43,17 +43,17 @@ PML_RZ::PML_RZ (const int lev, const amrex::BoxArray& grid_ba, const amrex::Dist
...
     const amrex::MultiFab & Br_fp = WarpX::GetInstance().getBfield_fp(lev,0);
     const amrex::MultiFab & Bt_fp = WarpX::GetInstance().getBfield_fp(lev,1);
-    WarpX::AllocInitMultiFab(pml_B_fp[0], amrex::convert( grid_ba, Br_fp.ixType().toIntVect() ), grid_dm,
-                             Br_fp.nComp(), Br_fp.nGrowVect(), "pml_B_fp[0]", 0.0_rt);
-    WarpX::AllocInitMultiFab(pml_B_fp[1], amrex::convert( grid_ba, Bt_fp.ixType().toIntVect() ), grid_dm,
-                             Bt_fp.nComp(), Bt_fp.nGrowVect(), "pml_B_fp[1]", 0.0_rt);
+    const amrex::BoxArray ba_Br = amrex::convert(grid_ba, Br_fp.ixType().toIntVect())
+    const amrex::BoxArray ba_Bt = amrex::convert(grid_ba, Bt_fp.ixType().toIntVect())
+    WarpX::AllocInitMultiFab(pml_B_fp[0], ba_Br, grid_dm, Br_fp.nComp(), Br_fp.nGrowVect(), "pml_B_fp[0]", 0.0_rt);
+    WarpX::AllocInitMultiFab(pml_B_fp[1], ba_Bz, grid_dm, Bt_fp.nComp(), Bt_fp.nGrowVect(), "pml_B_fp[1]", 0.0_rt);
```
EZoni:
```suggestion
    WarpX::AllocInitMultiFab(pml_B_fp[1], ba_Bt, grid_dm, Bt_fp.nComp(), Bt_fp.nGrowVect(), "pml_B_fp[1]", 0.0_rt);
```

Source/BoundaryConditions/PML_RZ.cpp
```diff
@@ -43,22 +43,17 @@ PML_RZ::PML_RZ (const int lev, const amrex::BoxArray& grid_ba, const amrex::Dist
...
-                                                     Br_fp.nComp(), Br_fp.nGrow() );
-    pml_B_fp[1] = std::make_unique<amrex::MultiFab>( amrex::convert( grid_ba, Bt_fp.ixType().toIntVect() ), grid_dm,
-                                                     Bt_fp.nComp(), Bt_fp.nGrow() );
-
-    pml_E_fp[0]->setVal(0.0);
-    pml_E_fp[1]->setVal(0.0);
-    pml_B_fp[0]->setVal(0.0);
-    pml_B_fp[1]->setVal(0.0);
+    const amrex::BoxArray ba_Br = amrex::convert(grid_ba, Br_fp.ixType().toIntVect())
+    const amrex::BoxArray ba_Bt = amrex::convert(grid_ba, Bt_fp.ixType().toIntVect())
```
EZoni:
```suggestion
    const amrex::BoxArray ba_Br = amrex::convert(grid_ba, Br_fp.ixType().toIntVect());
    const amrex::BoxArray ba_Bt = amrex::convert(grid_ba, Bt_fp.ixType().toIntVect());
```

Source/BoundaryConditions/PML_RZ.cpp
```diff
@@ -43,22 +43,17 @@ PML_RZ::PML_RZ (const int lev, const amrex::BoxArray& grid_ba, const amrex::Dist
 
     const amrex::MultiFab & Er_fp = WarpX::GetInstance().getEfield_fp(lev,0);
     const amrex::MultiFab & Et_fp = WarpX::GetInstance().getEfield_fp(lev,1);
-    pml_E_fp[0] = std::make_unique<amrex::MultiFab>( amrex::convert( grid_ba, Er_fp.ixType().toIntVect() ), grid_dm,
-                                                     Er_fp.nComp(), Er_fp.nGrow() );
-    pml_E_fp[1] = std::make_unique<amrex::MultiFab>( amrex::convert( grid_ba, Et_fp.ixType().toIntVect() ), grid_dm,
-                                                     Et_fp.nComp(), Et_fp.nGrow() );
+    const amrex::BoxArray ba_Er = amrex::convert(grid_ba, Er_fp.ixType().toIntVect())
+    const amrex::BoxArray ba_Et = amrex::convert(grid_ba, Et_fp.ixType().toIntVect())
```
EZoni:
```suggestion
    const amrex::BoxArray ba_Er = amrex::convert(grid_ba, Er_fp.ixType().toIntVect());
    const amrex::BoxArray ba_Et = amrex::convert(grid_ba, Et_fp.ixType().toIntVect());
```

EZoni (approved):

EZoni:
@ax3l @WeiqunZhang Ready for review, good to go on my side.

Source/WarpX.cpp
```diff
@@ -181,6 +181,9 @@ bool WarpX::do_multi_J = false;
 int WarpX::do_multi_J_n_depositions;
 bool WarpX::safe_guard_cells = 0;
 
+std::map<std::string, amrex::MultiFab *> WarpX::multifab_map;
+std::map<std::string, amrex::iMultiFab *> WarpX::imultifab_map;
```
ax3l, line 185: Storing as raw pointers is not ideal for lifetime control.

We could either continue to store as unique pointers or store the object directly as `amrex::MultiFab`/`amrex::iMultiFab`, as we do in ImpactX.

cc @WeiqunZhang for guidance since I am at SC22.

dpgrote: I agree that saving pointers can be problematic, but here, the lifetime of the MFs is controlled by the unique pointers in the WarpX class, e.g. `Efield_fp`. The map is only for convenience for the Python interface and isn't used anywhere in the C++.

WeiqunZhang: Yes, this should be a raw pointer because it does not own the memory.

ax3l: ah I overlooked that. makes sense (we could check if we like to migrate storage location later on)

ax3l (approved):

PR #3486, opened 2022-10-27
Add Python interface for flux injection

RemiLehe:
This PR is a follow-up of https://github.com/picmi-standard/picmi/pull/83.
It implements the new `picmi.UniformFluxInjection` in WarpX.


roelof-groenewald:
Maybe this can wait for #3476 to be merged since the new `UniformFluxDistribution` would also be able to inherit heavily from the parent class `DensityDistributionBase` created in that PR.

dpgrote (approved):
Looks good!

PR #3487, opened 2022-10-27
Add neumann BC in WarpX PICMI interface

RemiLehe:
This is a follow-up on #3483 ; this allows the neumann boundary conditions to be used with PICMI.


dpgrote (approved):
Thanks for this change (that I overlooked)!

PR #3494, opened 2022-11-01
[WIP] Adding Higginson's correction and a test for the Coulomb collision module

Yin-YinjianZhao:
We try to add a test that the simulation result should match better with an analytical solution for the Coulomb collision module.
The test is described by Smilei (3. Temperature isotropization, Fig 18.):
https://smileipic.github.io/Smilei/Understand/collisions.html#test-cases-for-collisions
Their scripts can be found here:
https://smileipic.github.io/tutorials/advanced_collisions.html
https://github.com/SmileiPIC/Smilei/tree/master/benchmarks/collisions

When implementing the test, we found our Coulomb module has a bug that occurs among collisions within a single species, which should be related to Higginson's correction:
https://www.sciencedirect.com/science/article/pii/S0021999120302242?via%3Dihub

Thus, we will add Higginson's correction and that test in this PR.


Yin-YinjianZhao:
So far, this version of the PR still has some issues, the results do not match that well.
Smilei's plot:
Screen Shot 2022-10-24 at 10 11 38 AM
Ours:
Screen Shot 2022-11-01 at 10 19 24 AM

Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
```diff
@@ -248,7 +251,7 @@ void UpdateMomentumPerezElastic (
 
     // Rejection method
     r = amrex::Random(engine);
-    if ( w2 > r*amrex::max(w1, w2) )
+    if ( w2/amrex::max(w1,w2)>r )
```
ax3l, line 254: 2x: If you keep the term on the RHS, then things should be identical mathematically, but we save division for a faster multiplication.

Also, since `amrex::max(w1,w2)` is used multiple times, consider storing it in a `const` variable?

Yin-YinjianZhao: Yes, will modify these later, currently there is still a major issue that has not been solved.

EZoni:
@Yin-YinjianZhao @RemiLehe Would you be able to point to the lines in the Smilei code where the Higginson correction is implemented? Is it supposed to be found somewhere in their file [Collisions/BinaryProcesses.cpp](https://github.com/SmileiPIC/Smilei/blob/master/src/Collisions/BinaryProcesses.cpp)?

RemiLehe:
Yes, the intra-species correction can be found here:
https://github.com/SmileiPIC/Smilei/blob/master/src/Collisions/BinaryProcesses.cpp#L304
More broadly, the Higginson scheme itself is implemented throughout this file so it is difficult to point a specific line.

RemiLehe:
Superseded by https://github.com/ECP-WarpX/WarpX/pull/5091

PR #3493, opened 2022-10-31
Perlmutter (NERSC): Update Modules & Jobscript

ax3l:
Update after major system upgrade.

Tested:
```
Initializing CUDA...
CUDA initialized with 1 GPU per MPI rank; 8 GPU(s) used in total
MPI initialized with 8 MPI processes
MPI initialized with thread support level 3
AMReX (22.10-25-g735c3513153f) initialized
PICSAR (4252e567089f)
WarpX (22.10-38-g886e495dd471)
# ...
# all good
```

cc @SeverinDiederichs 

- [ ] Affinity: test that pinning works and 4 GPUs per node are faster than 1 or 2

PR #3495, opened 2022-11-01
Perlmutter (NERSC): Finish Affinity Control

ax3l:
Slurm on Perlmutter is not yet configured to do proper affinity control (aka CPU-GPU-NIC pinning) itself. Thus, we do it ourselves.

- [x] testing: test that pinning works

Thanks to @kngott for investigating this!

cc @SeverinDiederichs for HiPACE++ usage

Follow-up to #3493


ax3l:
ping @EZoni @RemiLehe ready for merge :)

EZoni (approved):

PR #3501, opened 2022-11-03
Docs: Improve MPI Threading User FAQ

ax3l:
Clarify that this FAQ entry explains the initial output of WarpX in stdout.

### DIFF
--- a/Docs/source/usage/faq.rst
+++ b/Docs/source/usage/faq.rst
@@ -6,16 +6,27 @@ FAQ
 This section lists frequently asked usage questions.
 
 
-What is MPI thread support level?
----------------------------------
+What is "MPI initialized with thread support level ..."?
+--------------------------------------------------------
 
-We report this in output on startup together with other information.
+When we start up WarpX, we report a couple of information on used MPI processes across parallel compute processes, CPU threads or GPUs and further capabilities.
+For instance, a parallel, multi-process, multi-threaded CPU run could output::
 
-That is the `MPI support for threaded execution <https://www.mpich.org/static/docs/v3.1/www3/MPI_Init_thread.html>`__, e.g., with OpenMP or system threads.
+   MPI initialized with 4 MPI processes
+   MPI initialized with thread support level 3
+   OMP initialized with 8 OMP threads
+   AMReX (22.10-20-g3082028e4287) initialized
+   ...
 
+The 1st line is the number of parallel MPI processes (also called *MPI ranks*).
+
+The 2nd line reports on the `support level of MPI functions to be called from threads <https://www.mpich.org/static/docs/v3.1/www3/MPI_Init_thread.html>`__.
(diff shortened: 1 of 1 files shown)


Docs/source/usage/faq.rst
```diff
@@ -6,17 +6,28 @@ FAQ
...
+   OMP initialized with 8 OMP threads
+   AMReX (22.10-20-g3082028e4287) initialized
+   ...
 
+The 1st line is the number of parallel MPI processes (also called *MPI ranks*).
+
+The 2nd line reports on the `support level of MPI functions to be called from threads <https://www.mpich.org/static/docs/v3.1/www3/MPI_Init_thread.html>`__.
 We currently only use this for optional, :ref:`async IO with AMReX plotfiles <running-cpp-parameters-diagnostics>`.
 In the past, requesting MPI threading support had performance penalties, but we have not seen such anymore on recent systems.
 Thus, we request it by default but you can overwrite it with a compile time option if it ever becomes needed.
```
kngott, line 27: "Edit" is greyed out for me, so I'll do it here. 😄 

> The 2nd line reports on the `support level of the MPI library <https://www.mpich.org/static/docs/v3.1/www3/MPI_Init_thread.html>`__.
> A support level of 4 means WarpX has been ran with MPI_THREAD_MULTIPLE support.
>  We currently only use this for optional, :ref:`async IO with AMReX plotfiles <running-cpp-parameters-diagnostics>`.
>  In the past, requesting MPI_THREAD_MULTIPLE support had performance penalties, but we have not seen such anymore on recent systems.
>  Thus, we now request it by default but you can overwrite it with a compile time option for testing or if you see any effect on your application's performance.

(Should we put the option here? Or refer to the option list in the docs?)

ax3l: Good point,you mean `amrex.async_out_nfiles`? Already linked, but now also mentioned.

kngott: n_files just says the number of files to be used, right? It doesn't turn off Async I/O and THREAD_MULTIPLE initialization. 

(Hence why just pointing to the appropriate list of all the options might be better... then we also aren't updating in two places at once. 😄)

ax3l: Sure. Which options am I missing to link? Build or run-time?

kngott: To turn THREAD_MULTIPLE off (which I think is the discussion being had here),

WarpX_MPI_THREAD_MULTIPLE=OFF.

I don't know how async io works/responds if you have it on but THREAD_MULTIPLE off. Been a while. Might need to include that being off too (which I think is runtime).

ax3l: Ah ok, this flag is mentioned and further linked i the runtime inputs that are linked. Let me finalize and link directly.

RemiLehe (approved):

PR #3504, opened 2022-11-04
Allow `None` for Maxwell solver

roelof-groenewald:
This PR is in preparation for adding more solver options to WarpX (specifically a hybrid solver). It generalizes a non-Maxwellian option for field solvers i.e. options other than explicit electromagnetic field evolutions. As a side benefit is also immediately allows simulations to be run with no field solver (untested).

### DIFF
--- a/Source/Evolve/WarpXComputeDt.cpp
+++ b/Source/Evolve/WarpXComputeDt.cpp
@@ -36,8 +36,7 @@ WarpX::ComputeDt ()
     const amrex::Real* dx = geom[max_level].CellSize();
     amrex::Real deltat = 0.;
 
-    if (maxwell_solver_id == MaxwellSolverAlgo::None){ }
-    else if (maxwell_solver_id == MaxwellSolverAlgo::PSATD) {
+    if (maxwell_solver_id == MaxwellSolverAlgo::PSATD) {
         // Computation of dt for spectral algorithm
         // (determined by the minimum cell size in all directions)
 #if defined(WARPX_DIM_1D_Z)
@@ -64,8 +63,10 @@ WarpX::ComputeDt ()
             deltat = cfl * CartesianCKCAlgorithm::ComputeMaxDt(dx);
 #endif
         } else {
-            amrex::Abort(Utils::TextMsg::Err(
-                "ComputeDt: Unknown algorithm"));
+            if (maxwell_solver_id != MaxwellSolverAlgo::None){
+                amrex::Abort(Utils::TextMsg::Err(
+                    "ComputeDt: Unknown algorithm"));
+            }
         }
     }
 
--- a/Source/Evolve/WarpXComputeDt.cpp
+++ b/Source/Evolve/WarpXComputeDt.cpp
@@ -32,7 +32,15 @@
 void
 WarpX::ComputeDt ()
 {
-    // Determine
+    // Handle cases where the timestep is not limited by the speed of light
+    if (electromagnetic_solver_id == ElectromagneticSolverAlgo::None) {
+        for (int lev=0; lev<=max_level; lev++) {
+            dt[lev] = const_dt;
+        }
+        return;
+    }
+
+    // Determine the appropriate timestep as limited by the speed of light
     const amrex::Real* dx = geom[max_level].CellSize();
     amrex::Real deltat = 0.;
 
@@ -63,10 +71,7 @@ WarpX::ComputeDt ()
             deltat = cfl * CartesianCKCAlgorithm::ComputeMaxDt(dx);
 #endif
         } else {
-            if (electromagnetic_solver_id != ElectromagneticSolverAlgo::None){
-                amrex::Abort(Utils::TextMsg::Err(
-                    "ComputeDt: Unknown algorithm"));
-            }
+            amrex::Abort(Utils::TextMsg::Err("ComputeDt: Unknown algorithm"));
         }
     }
 
@@ -76,12 +81,6 @@ WarpX::ComputeDt ()
     if (do_subcycling) {
         for (int lev = max_level-1; lev >= 0; --lev) {
             dt[lev] = dt[lev+1] * refRatio(lev)[0];
-        }
-    }
-
-    if (electromagnetic_solver_id == ElectromagneticSolverAlgo::None) {
-        for (int lev=0; lev<=max_level; lev++) {
-            dt[lev] = const_dt;
         }
     }
 }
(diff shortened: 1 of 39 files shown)


EZoni:
@roelof-groenewald @RemiLehe  Thank you for this PR! Since we are cleaning things up, I was wondering if it would make things more consistent to rename `MaxwellSolverAlgo` as `ElectromagneticSolverAlgo`, as in `ElectrostaticSolverAlgo`. I feel like the `Electromagnetic` keyword could be slightly more self-explanatory than the `Maxwell` keyword, but this is just a suggestion. I'm fine with keeping the name as it is now. What do you think?

roelof-groenewald:
@EZoni Renaming to `ElectromagneticSolverAlgo` sounds like a good idea to me. I will make that change. There is also somewhat of an inconsistency in the variable names that hold the configuration in that `maxwell_solver_id` holds the specifics of the EM solver but `do_electrostatic` holds the equivalent for the ES case. So maybe `do_electrostatic` should also become `electrostatic_solver_id`?

roelof-groenewald:
I see there are a number of CI failures due to the above changes but I'm not entirely clear on why. I believe the failures are a consequence of an early return in `FiniteDifferenceSolver` that happens now. Does anybody know of the top of their head which parts of that function is needed for the failing electrostatic cases? Or maybe more to the point, where is `m_h_stencil_coefs_(x,y,z)` used in the ES case?

Actually I think the failing tests are due to `WarpX::InitializeEBGridData` not executing. I will fix.

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1240,7 +1240,8 @@ void WarpX::InitializeEBGridData (int lev)
               "particles are close to embedded boundaries");
         }
 
-        if (WarpX::maxwell_solver_id == MaxwellSolverAlgo::Yee ||
+        if (WarpX::maxwell_solver_id == MaxwellSolverAlgo::None ||
+            WarpX::maxwell_solver_id == MaxwellSolverAlgo::Yee ||
             WarpX::maxwell_solver_id == MaxwellSolverAlgo::CKC ||
             WarpX::maxwell_solver_id == MaxwellSolverAlgo::ECT) {
```
roelof-groenewald: There are several code blocks like this. Would it make sense to replace these with 
`if (WarpX::maxwell_solver_id != MaxwellSolverAlgo::PSATD)`
or do we prefer keeping it explicitly specifying the cases for which that block should be executed?

Source/Parallelization/GuardCellManager.cpp
```diff
@@ -251,7 +251,7 @@ guardCellManager::Init (
         ng_FieldSolverG = ng_alloc_EB;
     }
 #ifdef WARPX_DIM_RZ
-    else if (maxwell_solver_id == MaxwellSolverAlgo::Yee) {
+    else if (maxwell_solver_id == MaxwellSolverAlgo::None || maxwell_solver_id == MaxwellSolverAlgo::Yee) {
```
roelof-groenewald: I'm not sure if this is necessary, just tying to get the simulation to run the same way as before first.

Source/Parallelization/GuardCellManager.cpp
```diff
@@ -262,8 +262,9 @@ guardCellManager::Init (
             ng_FieldSolver  = CartesianNodalAlgorithm::GetMaxGuardCell();
             ng_FieldSolverF = CartesianNodalAlgorithm::GetMaxGuardCell();
             ng_FieldSolverG = CartesianNodalAlgorithm::GetMaxGuardCell();
-        } else if (maxwell_solver_id == MaxwellSolverAlgo::Yee
-                   || maxwell_solver_id == MaxwellSolverAlgo::ECT) {
+        } else if (maxwell_solver_id == MaxwellSolverAlgo::None ||
```
roelof-groenewald: Same comment as above.

Source/Parallelization/GuardCellManager.cpp
roelof-groenewald: Indeed there are failures if this is not included since the EB logic uses `ng_FieldSolver`.

RemiLehe:
Thanks a lot for this PR @roelof-groenewald !
Could you actually activate the `none` electromagnetic solver in the following CI test:
https://github.com/ECP-WarpX/WarpX/tree/development/Examples/Tests/collision
https://github.com/ECP-WarpX/WarpX/tree/development/Examples/Tests/scraping
(These tests are currently using the electrostatic solver, but with `<species>.do_not_deposit = 1`, so that the fields are zero throughout the simulation. This was a "hack" that we used at the time, because we did not have a `none` maxwell solver. You can simply replace `warpx.do_electrostatic = ...` by `algo.maxwell_solver = none`, and I think that it should still do the same thing.)

Source/Initialization/WarpXInitData.cpp
RemiLehe: Yes, I think it makes sense to have `!= MaxwellSolverAlgo::PSATD` here.

Source/Evolve/WarpXComputeDt.cpp
```diff
@@ -77,7 +79,7 @@ WarpX::ComputeDt ()
         }
     }
 
-    if (do_electrostatic != ElectrostaticSolverAlgo::None) {
+    if (electromagnetic_solver_id == ElectromagneticSolverAlgo::None) {
```
RemiLehe: Maybe this whole block could be moved to the beginning of the function (along with an early `return` statement, or an `else` condition that only executes the rest of the function if `electromagnetic_solver_id` is not `None`.)
This might be more readable.

Source/Initialization/WarpXInitData.cpp
roelof-groenewald: Sounds good, thanks!

roelof-groenewald:
@RemiLehe I updated those tests to use `algo.maxwell_solver = none` and it seems to work fine :)

EZoni:
> @RemiLehe I updated those tests to use algo.maxwell_solver = none and it seems to work fine :)

That's great, thanks @roelof-groenewald! Do we need to keep `<species>.do_not_deposit = 1` in the input files of those tests then? I think we should be able to remove that hack if we use `algo.maxwell_solver = none`.

Source/Parallelization/GuardCellManager.H
```diff
@@ -36,12 +36,11 @@ public:
      * \param noy_fft order of PSATD in y direction
      * \param noz_fft order of PSATD in z direction
      * \param nci_corr_stencil stencil of NCI corrector
-     * \param maxwell_solver_id if of Maxwell solver
+     * \param electromagnetic_solver_id if of Maxwell solver
```
EZoni:
```suggestion
     * \param electromagnetic_solver_id Integer corresponding to the type of Maxwell solver
```

roelof-groenewald: Nice catch!

EZoni (approved):
Great, looks good to me! @RemiLehe As soon as you approve it, I can merge the PR.

PR #3505, opened 2022-11-04
Clean Species Physical Properties

NeilZaim:
This fixes a couple of minor omissions regarding the species physical properties:
- `alpha` was listed twice in the `string_to_species` map.
- The properties of aluminium were not given. This currently leads to a crash if one tries to use aluminium. I've added the data using the NIST database, following what was done in #3090.
- beryllium9, carbon14 and fluorine19 were not included in the documentation.


RemiLehe (approved):

RemiLehe:
Thanks for cleaning up!

PR #3508, opened 2022-11-05
Load balancing bug fix: remake MultiFabs for Vay deposition, current centering, time averaging

RevathiJambunathan:
Additional MultiFabs for
- Vay deposition
- current centering
- time averaging

need to be allocated after load-balancing, otherwise the simulation crashes. 

Thanks to @hklion for reporting this and testing it as well.

### DIFF
--- a/Source/Parallelization/WarpXRegrid.cpp
+++ b/Source/Parallelization/WarpXRegrid.cpp
@@ -173,6 +173,13 @@ WarpX::RemakeLevel (int lev, Real /*time*/, const BoxArray& ba, const Distributi
             if (current_deposition_algo == CurrentDepositionAlgo::Vay) {
                 RemakeMultiFab(current_fp_vay[lev][idim], dm, false);
             }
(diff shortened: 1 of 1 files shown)


Source/Parallelization/WarpXRegrid.cpp
```diff
@@ -1,6 +1,6 @@
 /* Copyright 2019 Andrew Myers, Ann Almgren, Axel Huebl
  * David Grote, Maxence Thevenet, Michael Rowan
- * Remi Lehe, Weiqun Zhang, levinem
+ * Remi Lehe, Weiqun Zhang, levinem, Revathi Jambunathan
```
RevathiJambunathan, line 3:
```suggestion
 * Remi Lehe, Weiqun Zhang, levinem
```

EZoni (approved):
Thank you, @RevathiJambunathan and @hklion, for this bug fix! Looks good to me.

PR #3510, opened 2022-11-07
2D/RZ Embedded Boundaries Bug Fix

EZoni:
While investigating some of the CI failures appearing in #3484, I noticed that the CI test `embedded_boundary_cube_2d` fails also on the `development` branch, when running in debug mode.

There seems to be an out-of-bound access here:
https://github.com/ECP-WarpX/WarpX/blob/ada6fc49cf16f40d1764d3a95c8338126360f329/Source/Initialization/WarpXInitData.cpp#L1027

The bug is somehow triggered by the changes in #3484, but since it happens on the `development` branch too, I believe it is independent of those changes.

To-do:
- [x] Add debug flag to CI test (trigger crash)
- [x] Fix bug
- [x] Remove debug flag from CI test

### DIFF
--- a/Source/Initialization/WarpXInitData.cpp
+++ b/Source/Initialization/WarpXInitData.cpp
@@ -976,10 +976,10 @@ WarpX::InitializeExternalFieldsOnGridUsingParser (
        amrex::Array4<amrex::Real> const& Sz = face_areas[2]->array(mfi);
 
 #if defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
-       const amrex::Dim3 lx_lo = lx.begin;
-       const amrex::Dim3 lx_hi = lx.end;
-       const amrex::Dim3 lz_lo = lz.begin;
-       const amrex::Dim3 lz_hi = lz.end;
+       const amrex::Dim3 lx_lo = amrex::lbound(lx);
+       const amrex::Dim3 lx_hi = amrex::ubound(lx);
+       const amrex::Dim3 lz_lo = amrex::lbound(lz);
+       const amrex::Dim3 lz_hi = amrex::ubound(lz);
 #endif
 
 #if defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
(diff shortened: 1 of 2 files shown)


EZoni:
It is also possible that the issue seen here is separate from the issue seen in #3484. Namely, when running in debug mode, a different bug gets triggered, which hides the original bug that causes the CI failure in #3484. Not 100% sure at this point.

I think the bug here is simply caused by the fact that, when we try to read `lx(i-1, j, k)` (see code linked in the PR description above), the index `i-1` goes out of bound at the edge (e.g. minimum `i` allowed is `-1`, but `i-1=-2` when `i=-1`).

@lgiacome @RemiLehe @WeiqunZhang 
Do you have an opinion on how this should be fixed?

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1024,7 +1032,10 @@ WarpX::InitializeExternalFieldsOnGridUsingParser (
                 if((field=='E' and ly(i, j, k)<=0) or (field=='B' and Sy(i, j, k)<=0))  return;
 #elif defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
                 //In XZ and RZ Ey is associated with a mesh node, so we need to check if  the mesh node is covered
-                if((field=='E' and (lx(i, j, k)<=0 || lx(i-1, j, k)<=0 || lz(i, j, k)<=0 || lz(i, j-1, k)<=0)) or
+                if((field=='E' and (lx(std::min(i  , lx_hi.x-1), std::min(j  , lx_hi.y-1), k)<=0
+                                 || lx(std::max(i-1, lx_lo.x  ), std::min(j  , lx_hi.y-1), k)<=0
+                                 || lz(std::min(i  , lz_hi.x-1), std::min(j  , lz_hi.y-1), k)<=0
+                                 || lz(std::min(i  , lz_hi.x-1), std::max(j-1, lz_lo.y  ), k)<=0)) or
```
EZoni: @WeiqunZhang @RemiLehe 
Not sure I understand why I needed to subtract `1` from `a_hi.<x,z>` (computed above as `a.end`) in order for this to work. 

The out-of-bound error looked like
```
(33,-1,0,0) is out of bound (-1:32,-1:33,0:0,0:0)
```
but `a.end` gives `(33,34,1)` which seems inconsistent with the error message (hence the subtraction of `1`).

EZoni: Answered by @WeiqunZhang on Slack: the attributes `begin` and `end` follow the C++ convention. For indexing that follows the Fortran convention (starting from `0`), one can use `amrex::lbound(a)` or `amrex::ubound(a)` instead.

RemiLehe (approved):

PR #3514, opened 2022-11-09
Take into account mean velocity in gaussian flux injection

RemiLehe:
## Overview

In the current version of the code, when using the `gaussian_flux` injection, we ignore the mean velocity in the direction of the flux (`u_m`), and we therefore generate the momentum distribution:
Screen Shot 2022-11-08 at 8 00 00 PM
instead of the correct momentum distribution:
Screen Shot 2022-11-08 at 8 00 24 PM

This PR fixes this (i.e. generates the correct distribution) by using the rejection method. More specifically, it uses two different generation methods, depending on whether the mean velocity is low or high compared to the thermal velocity.

## Low mean velocity (compared to the thermal velocity)

In this case, we write the momentum distribution as:
Screen Shot 2022-11-08 at 8 16 09 PM
where:
- The red term is a probability distribution that we already know how to generate, using the method of the inverse of the cumulative distribution (provided that `u_m < u_th`).
- The blue term is a number between 0 and 1, and can thus be used as an acceptance probability, in the rejection method.
- The remaining black term is a constant (it does not depend on `u`), and thus does not matter in the rejection method.

This plots the approximate distribution (red term) and the acceptance probability (blue term) for different values of u_m and u_th:
image
As can be seen, for `u_m << u_th`, the acceptance probability is very close to 1 over the approximate distribution. 

## High mean velocity (compared to the thermal velocity)

In this case, we write the momentum distribution as:
Screen Shot 2022-11-08 at 9 19 03 PM

- The red term is a probability distribution that we already know how to generate.
- The blue term is a number between 0 and 1, and can thus be used as an acceptance probability, in the rejection method.
- The remaining black term is a constant (it does not depend on `u`), and thus does not matter in the rejection method.

This plots the approximate distribution (red term) and the acceptance probability (blue term) for different values of `u_m` and `u_th`:
image
As can be seen, for `u_m >> u_th`, the acceptance probability is very close to 1 over the approximate distribution. 

## Automated test

I also added an automated test that checks both of the above methods, by injecting the particles in the simulation with the `gaussian_flux` distribution and then checking (in post-processing)  that the particles have the right velocity distribution.


RemiLehe:
It seems that the test `FluxInjection` was not doing quite what we wanted in the `development` version of WarpX.

From the description:
```
This script tests the continuous injection of particles in
the azimuthal direction, in RZ geometry.

In this tests, relativistic electrons are continuously injected
in a uniform Bz field, at a radius that corresponds to their
Larmor radius. Therefore, throughout the simulations, these
electrons always stay at the same radius, while gyrating
around the z axis.
```
and the script had the following lines:
```
electron.momentum_distribution_type = gaussianflux
electron.uy_th = 0.
electron.uy_m = 10.
```

However, `uy_m` was silently ignored in this script. As a consequence, the electrons were injected with zero velocity, and their gyroradius is 0 in this case.
Screenshot from 2022-11-11 08-57-11
(note that different macroparticles start with slightly different initial `r` and `theta`, but they all have the same gyroradius)

The current PR fixes this issue, and the electrons are now indeed gyrating with the correct gyroradius, as illustrated by a few sample trajectories from this test: 


Note that I had to increase the tolerance of the test, from:
```
assert np.all( (r >= 1.5) & (r <=1.9) )
```
to 
```
assert np.all( (r >= 1.48) & (r <=1.92) )
```

Examples/Tests/FluxInjection/analysis_flux_injection_rz.py
```diff
@@ -55,7 +55,7 @@
 # Check that the number of particles matches the expected one
 assert np.allclose( w.sum(), n_tot, rtol=0.05 )
 # Check that the particles are at the right radius
-assert np.all( (r >= 1.5) & (r <=1.9) )
+assert np.all( (r >= 1.49) & (r <=1.91) )
```
RemiLehe: See the PR description for an explanation of why this test had to be updated.

RemiLehe, line 58:
```suggestion
assert np.all( (r >= 1.49) & (r <=1.91) )
```

Source/Initialization/InjectorMomentum.H
```diff
@@ -86,6 +87,66 @@ private:
     amrex::Real m_ux_th, m_uy_th, m_uz_th;
 };
 
+namespace {
```
RemiLehe, line 90: I had to use an anonymous namespace here. Otherwise, when compiling, the linker says that the function `generateGaussianFluxDist` is defined multiple times (since this `.H` file is included is several of the `.cpp` files). I am not sure if there is a better/cleaner way to do this.

Examples/Tests/FluxInjection/analysis_flux_injection_3d.py
```diff
@@ -0,0 +1,118 @@
...
+
+# Open plotfile specified in command line
+fn = sys.argv[1]
+ds = yt.load( fn )
+ad = ds.all_data()
+t_max = ds.current_time.item()  # time of simulation
+
+# Total number of electrons expected:
+# Simulation parameters determine the total number of particles emitted (Ntot)
+flux = 1. # in m^-2.s^-1, from the input script
```
dpgrote, line 46: It's not necessary, but this could be generalized by having these parameters obtained from the `ds.parameters` dictionary rather than hardcoded.

Examples/Tests/FluxInjection/analysis_flux_injection_3d.py
```diff
@@ -0,0 +1,118 @@
...
+emission_surface = 8*8 # in m^2
+Ntot = flux * emission_surface * t_max
+
+# Parameters of the histogram
+hist_bins = 50
+hist_range = [-0.5, 0.5]
+
+# Define function that histogram and check the data
+
+def gaussian_dist(u, u_th, u_m=0):
```
dpgrote: This may be a little confusing having the `u_m` argument since it isn't used. Were you intending to test a case with it nonzero?

Source/Initialization/InjectorMomentum.H
```diff
@@ -86,6 +87,66 @@ private:
...
+        // Momentum to be returned at the end of this function
+        amrex::Real u = 0._rt;
+
+        if (u_th == 0._rt) {
+            u = u_m; // Trivial case ; avoids division by 0 in the rest of the code below
+        } else if (u_m < 0.6*u_th) {
+            // Mean velocity is lower than thermal velocity
+            // Use the distribution u*exp(-u**2*(1-u_m/u_th)/(2*u_th**2)) as an approximation
+            // and then use the rejection method to correct it
+            // ( stop rejecting with probability exp(-u_m/(2*u_th**3)*(u-u_th)**2) )
```
dpgrote, line 113: Maybe the math is clear enough, but perhaps a comment could be added that when `u_m = 0` the first guess is always used (so that this doesn't add any inefficiency for that common case).

Examples/Tests/FluxInjection/inputs_3d
```diff
@@ -0,0 +1,68 @@
...
+electron.surface_flux_pos = -1.
+electron.flux_normal_axis = y
+electron.flux_direction = +1
+electron.profile = constant
+electron.density = 1.
+electron.momentum_distribution_type = gaussianflux
+electron.ux_th = 0.1
+electron.uy_th = 0.1
+electron.uy_m = 0.07
+electron.uz_th = 0.1
```
dpgrote, line 47: A detail I noticed. This injects particles over the whole plane of the domain. Since the particles have a transverse velocity spread, it is possible that some particles are lost out the sides. If so, this would have a small effect on the comparison made in the analysis since the number of particles would be slightly too small.

dpgrote (approved):
This looks good and I can approve it, though I have a couple of comments regarding possible changes.

Examples/Tests/FluxInjection/inputs_3d
RemiLehe: Agreed. However, in this particular case, the boundary condition is set to `periodic` ; therefore particles are in fact not lost.

Examples/Tests/FluxInjection/analysis_flux_injection_3d.py
RemiLehe: Agreed. I will update this.

Source/Initialization/InjectorMomentum.H
RemiLehe: OK, I will add this.

PR #3515, opened 2022-11-09
Abort when using Vay Deposition with yee or ckc

RevathiJambunathan:
Vay deposition is currently only implemented for PSATD and does not work with yee/ckc

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -1553,8 +1553,7 @@ Numerics and algorithms
     Available options are: ``direct``, ``esirkepov``, and ``vay``. The default choice
     is ``esirkepov`` for FDTD maxwell solvers and ``direct`` for standard or
     Galilean PSATD solver (that is, with ``algo.maxwell_solver = psatd``).
-    Note that ``vay`` is only available for ``algo.maxwell_solver = psatd`` and not available for
-    ``algo.maxwell_solver = yee`` or ``algo.maxwell_solver = ckc``.
+    Note that ``vay`` is only available for ``algo.maxwell_solver = psatd``.
 
     1. ``direct``
 
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -1553,6 +1553,8 @@ Numerics and algorithms
     Available options are: ``direct``, ``esirkepov``, and ``vay``. The default choice
     is ``esirkepov`` for FDTD maxwell solvers and ``direct`` for standard or
     Galilean PSATD solver (that is, with ``algo.maxwell_solver = psatd``).
+    Note that ``vay`` is only available for ``algo.maxwell_solver = psatd`` and not available for
+    ``algo.maxwell_solver = yee`` or ``algo.maxwell_solver = ckc``.
 
     1. ``direct``
 
--- a/Source/Evolve/WarpXEvolve.cpp
+++ b/Source/Evolve/WarpXEvolve.cpp
@@ -917,12 +917,15 @@ WarpX::PushParticlesandDepose (int lev, amrex::Real cur_time, DtType a_dt_type, 
         current_y = current_fp_nodal[lev][1].get();
         current_z = current_fp_nodal[lev][2].get();
     }
-    else if (WarpX::current_deposition_algo == CurrentDepositionAlgo::Vay)
+#ifdef PSATD
+    else if (WarpX::current_deposition_algo == CurrentDepositionAlgo::Vay and
+             electromagnetic_solver_id == ElectromagneticSolverAlgo::PSATD)
     {
         current_x = current_fp_vay[lev][0].get();
         current_y = current_fp_vay[lev][1].get();
         current_z = current_fp_vay[lev][2].get();
     }
+#endif
     else
     {
         current_x = current_fp[lev][0].get();
--- a/Source/Evolve/WarpXEvolve.cpp
+++ b/Source/Evolve/WarpXEvolve.cpp
@@ -917,15 +917,13 @@ WarpX::PushParticlesandDepose (int lev, amrex::Real cur_time, DtType a_dt_type, 
         current_y = current_fp_nodal[lev][1].get();
         current_z = current_fp_nodal[lev][2].get();
     }
-#ifdef PSATD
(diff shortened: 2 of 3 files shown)


RevathiJambunathan:
@EZoni since Vay does not work without psatd, should there be an additional condition here for  `electromagnetic_solver_id == ElectromagneticSolverAlgo::PSATD`

The reason I ask is because when debugging a crash for Vay deposition with FDTD, this line suggests that Vay is implemented for FDTD. When discussing with @RemiLehe , he mentioned its best to ask you whether to modify this or not, depending on if Vay will be implemented in the future with FDTD. There are some other places where `current_fp_vay` is allocated and initialized with Vay and we could add the same conditional there. 


  https://github.com/ECP-WarpX/WarpX/blob/3e98c31a491fb438cb98692c7a62dbcdd717c39b/Source/Evolve/WarpXEvolve.cpp#L913

Source/WarpX.cpp
```diff
@@ -1010,6 +1010,12 @@ WarpX::ReadParameters ()
             maxLevel() <= 0,
             "Vay deposition not implemented with mesh refinement");
 
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+            !(WarpX::current_deposition_algo == CurrentDepositionAlgo::Vay and
+             (electromagnetic_solver_id == ElectromagneticSolverAlgo::Yee ||
+              electromagnetic_solver_id == ElectromagneticSolverAlgo::CKC)),
+            "Vay deposition not implemented with ckc or yee");
+
```
EZoni: Responding to your question, yes, Vay deposition is only implemented for the PSATD solver. I do not recall discussing its implementation for the FDTD case in the near future. The algorithm itself, as it is described in the literature ([Vay et al, JCP, 2013](https://doi.org/10.1016/j.jcp.2013.03.010), equations (20)-(24)), is designed for a spectral method, as the true current J is computed from the deposited current D through some operations in k-space.

Hence, I agree that we could add a condition `electromagnetic_solver_id == ElectromagneticSolverAlgo::PSATD` here. Personally, I find the conditions like the one implemented now, `!(a && (b || c))`, relatively intricated and not easy to decipher. I would be in favor of a more verbose but easier implementation, such as
```
if (WarpX::current_deposition_algo == CurrentDepositionAlgo::Vay)
{
        electromagnetic_solver_id == ElectromagneticSolverAlgo::PSATD,
        "Vay deposition implemented only for algo.maxwell_solver=psatd");
}
```

Docs/source/usage/parameters.rst
```diff
@@ -1558,6 +1558,8 @@ Numerics and algorithms
     Available options are: ``direct``, ``esirkepov``, and ``vay``. The default choice
     is ``esirkepov`` for FDTD maxwell solvers and ``direct`` for standard or
     Galilean PSATD solver (that is, with ``algo.maxwell_solver = psatd``).
+    Note that ``vay`` is only available for ``algo.maxwell_solver = psatd`` and not available for
+    ``algo.maxwell_solver = yee`` or ``algo.maxwell_solver = ckc``.
```
EZoni: Here I would maybe only state "Note that `vay` is only available for `algo.maxwell_solver = psatd`", without stating the options for which it is not available, otherwise we risk missing some of those as new ones are added.

Source/WarpX.cpp
RevathiJambunathan: sounds good. I will add this and will also add that condition whereever needed then

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -924,12 +924,15 @@ WarpX::PushParticlesandDepose (int lev, amrex::Real cur_time, DtType a_dt_type,
...
-    else if (WarpX::current_deposition_algo == CurrentDepositionAlgo::Vay)
+#ifdef PSATD
+    else if (WarpX::current_deposition_algo == CurrentDepositionAlgo::Vay and
+             electromagnetic_solver_id == ElectromagneticSolverAlgo::PSATD)
     {
         current_x = current_fp_vay[lev][0].get();
         current_y = current_fp_vay[lev][1].get();
         current_z = current_fp_vay[lev][2].get();
     }
+#endif
```
EZoni: Thanks, @RevathiJambunathan! 

If you want to leave the `#ifdef` directive, I think it will need to be `#ifdef WARPX_USE_PSATD` instead of `#ifdef PSATD` (that's why CI is currently failing, just tested locally).

However, I would have two requests/comments:

1) I would be more in favor of removing the `#ifdef` directive. Personally, I don't think it's needed here, as those few lines of code can be compiled even without PSATD.

2) I would argue that even the extra condition in the `if` statement, namely `electromagnetic_solver_id == ElectromagneticSolverAlgo::PSATD`, is not necessary here. Thanks to the assert message that you added in `WarpX::ReadParameters`, a situation where `WarpX::current_deposition_algo == CurrentDepositionAlgo::Vay ` is true and `electromagnetic_solver_id == ElectromagneticSolverAlgo::PSATD` is false cannot occurr anymore. If your goal here was simply to make sure that the code clearly shows that Vay deposition and the "Vay current" exist only with the PSATD solver, I think we could add an inline comment stating `// Note: Vay deposition is supported only for PSATD, code currently aborts otherwise` or something like that. Otherwise, I think there are many other places in the code where a check on Vay deposition would need to be coupled to a check on the electromagnetic solver. But again, thanks to the assert message that you added in `WarpX::ReadParameters`, I think that's not necessary anymore: that single check takes care of all problems. What do you think?

RevathiJambunathan: Thanks Edoardo! 
I agree that there would be too many places to add an additional if condition.
However, the way it read now, when debugging, it seemed as if its implemented for FDTD

With this PR, we have added comments in the documentation, an abort, and adding a comment makes it clear. So I agree with removing this ifdef and if condition altogether, but adding a meaningful comment.

EZoni (approved):

EZoni:
@RevathiJambunathan Thank you, the PR looks good to me and I approved the changes. Some macOS dependencies for CI are still broken, hopefully they'll be fixed by #3521. We will need to rebase this after #3521, in order to fix CI here and merge the PR.

PR #3516, opened 2022-11-09
[WIP] Remove particle counter in BTD

RemiLehe:
As discussed in past meetings with @RevathiJambunathan , the variable `m_totalParticles_in_buffer` is updated at each iteration (using expensive MPI calls) but is only used in one place in the code (which may not be critical).

This PR attempts to remove this variable.

TODO:
- [ ] Test performance on milestone case.

### DIFF
--- a/Source/Diagnostics/BTDiagnostics.cpp
+++ b/Source/Diagnostics/BTDiagnostics.cpp
@@ -124,7 +124,6 @@ void BTDiagnostics::DerivedInitData ()
     }
     m_particles_buffer.resize(m_num_buffers);
     m_totalParticles_flushed_already.resize(m_num_buffers);
-    m_totalParticles_in_buffer.resize(m_num_buffers);
 
     // check that simulation can fill all BTD snapshots
     const int lev = 0;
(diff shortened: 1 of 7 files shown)


Source/Diagnostics/ComputeDiagFunctors/BackTransformParticleFunctor.cpp
```diff
@@ -149,7 +149,6 @@ BackTransformParticleFunctor::operator () (PinnedMemoryParticleContainer& pc_dst
             }
         }
     }
-    totalParticleCounter = pc_dst.TotalNumberOfParticles();
```
RemiLehe, line 152: As discussed with @RevathiJambunathan, this is called for every BTD snapshot and at every PIC iteration, and generates MPI calls. It is therefore an expensive operation (AllReduce: https://github.com/AMReX-Codes/amrex/blob/development/Src/Particle/AMReX_ParticleContainerI.H#L375).

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -1382,21 +1378,19 @@ BTDiagnostics::PrepareParticleDataForOutput()
                 // Check if the zslice is in domain
                 bool ZSliceInDomain = GetZSliceInDomainFlag (i_buffer, lev);
                 if (ZSliceInDomain) {
-                    if ( m_totalParticles_in_buffer[i_buffer][i] == 0) {
```
RemiLehe, line 1385: It is unclear to me how critical this `if` condition is. (My understanding is that it detects whether a buffer has just been flushed at the previous iteration.)
If it is indeed critical, we could maybe think of a way to replace it with some other boolean which is less expensive to compute/update. In fact, having a flag that specifically checks whether a buffer has been flushed in a previous iteration is probably more accurate than the above check - for instance in the case where no particles of a given species are captured by the BTD diagnostics)

PR #3518, opened 2022-11-11
Fix warnings with ceil in BTD code

dpgrote:
This is a small fix that removes warnings about implicit conversions from real to int.


RTSandberg (approved):

RevathiJambunathan (approved):

RevathiJambunathan:
Thank you @dpgrote !

Issue #3446, opened 2022-10-03
Possible diagnostic bug: `NFluxPerCell` with different `surface_flux_pos` generates different electron density.


jcyu96:
Hello everyone,

While using the field `NFluxPerCell` for electron, I got the different electron density as different parameter `surface_flux_pos` was used, that is not supposed to happen. Below are the settings in my input file and electron density figures for the two species: 
```
electron.charge = -q_e
electron.mass = m_e
electron.injection_style = NFluxPerCell
electron.num_particles_per_cell = 8
electron.surface_flux_pos = -2.e-6
electron.flux_normal_axis = z
electron.flux_direction = +1
electron.xmin =  15.e-6
electron.xmax =  25.e-6
electron.zmin = -5.e-6
electron.zmax =  5.e-6
electron.profile = constant
electron.density = 1.e21
electron.momentum_distribution_type = "maxwell_juttner"
electron.theta = 1.2
```
![2022-10-03 10 44](https://user-images.githubusercontent.com/57207865/193494287-3e422963-7200-4b98-a792-5db3694faf30.jpg)

```
electron1.charge = -q_e
electron1.mass = m_e
electron1.injection_style = NFluxPerCell
electron1.num_particles_per_cell = 8
electron1.surface_flux_pos = 0.e-6
electron1.flux_normal_axis = z
electron1.flux_direction = +1
electron1.xmin =  -25.e-6
electron1.xmax =  -15.e-6
electron1.zmin = -5.e-6
electron1.zmax =  5.e-6
electron1.profile = constant
electron1.density = 1.e21
electron1.momentum_distribution_type = "maxwell_juttner"
electron1.theta = 1.2
```
![image](https://user-images.githubusercontent.com/57207865/193494416-c37b674a-bda4-4b90-8ac1-61f6a2a3d46b.png)

Can anyone explain why this conflict is there? Thanks to all！

dpgrote:
I notice that you are using the Maxwell-Juttner distribution. This distribution is not really supported in this context. Only the `gaussianflux` distribution is designed as a flux velocity distribution. Do you see the same difference when using that?

jcyu96:
Hi @dpgrote 
I tried `gaussianflux` and got other unexplainable phenomenon. When I inject multiple electron species, as long as two species have equal absolute values of `<species_name>.surface_flux_pos`, one of them will not be injected. Like the following: 
```
electron.charge = -q_e
electron.mass = m_e
[...]

PR #3519, opened 2022-11-11
Flux injection: move particle only after performing checks

RemiLehe:
In the flux injection, we push the position of the injected particles forward by a random time, which is evenly sampled between 0 and `dt` - right after creating the particle. (This is in order to produce a continuous-looking flow of particles, as opposed to a set of discrete packets of particles, separated by `v*dt`.)

This PR moves this push (i.e. the call to `UpdatePosition`) **after** a certain number of operations:
- Checking that the particle is within the local tile. (This is check is maybe not strictly needed, since we have a call to a global `Redistribute` at the end of the `AddPlasmaFlux`.) I think that we should move the particle only **after** performing this check. Otherwise, particles may get systematically removed if the injection plane is very close to the boundary of a local subdomain. (Even though the injection is inside the subdomain, particles will exit the subdomain after being pushed, and will therefore be removed.) This probably what is causing issue #3446.
- Calculating the flux of particle. (`dens = inj_rho->getDensity(ppos.x, ppos.y, ppos.z);`) In the case where the user-specified flux is position-dependent (and given by a 3D function of space - e.g. a parsed expression), I think it would be more intuitive to the user if this expression is evaluated at the position of the injection plane, rather than at the position of the injected particle **after** it has been pushed.
- For RZ simulations: Converting the position of the particles from cylindrical to Cartesian. My understanding is that the function `UpdatePosition` expects its argument to be in Cartesian coordinates.

### DIFF
--- a/Source/Particles/PhysicalParticleContainer.cpp
+++ b/Source/Particles/PhysicalParticleContainer.cpp
@@ -1661,9 +1661,6 @@ PhysicalParticleContainer::AddPlasmaFlux (amrex::Real dt)
                 pu.y *= PhysConst::c;
                 pu.z *= PhysConst::c;
 
-                const amrex::Real t_fract = amrex::Random(engine)*dt;
-                UpdatePosition(ppos.x, ppos.y, ppos.z, pu.x, pu.y, pu.z, t_fract);
-
 #if defined(WARPX_DIM_3D)
                 if (!tile_realbox.contains(XDim3{ppos.x,ppos.y,ppos.z})) {
                     p.id() = -1;
@@ -1682,7 +1679,6 @@ PhysicalParticleContainer::AddPlasmaFlux (amrex::Real dt)
                     continue;
                 }
 #endif
-
                 // Save the x and y values to use in the insideBounds checks.
                 // This is needed with WARPX_DIM_RZ since x and y are modified.
                 Real xb = ppos.x;
@@ -1726,8 +1722,10 @@ PhysicalParticleContainer::AddPlasmaFlux (amrex::Real dt)
                     p.id() = -1;
                     continue;
                 }
-
                 Real dens = inj_rho->getDensity(ppos.x, ppos.y, ppos.z);
+
+                const amrex::Real t_fract = amrex::Random(engine)*dt;
+                UpdatePosition(ppos.x, ppos.y, ppos.z, pu.x, pu.y, pu.z, t_fract);
 
                 // Remove particle if density below threshold
                 if ( dens < density_min ){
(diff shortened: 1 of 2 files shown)


RemiLehe:
Closes #3446

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1726,9 +1722,11 @@ PhysicalParticleContainer::AddPlasmaFlux (amrex::Real dt)
                     p.id() = -1;
                     continue;
                 }
-
                 Real dens = inj_rho->getDensity(ppos.x, ppos.y, ppos.z);
 
+                const amrex::Real t_fract = amrex::Random(engine)*dt;
```
dpgrote: Thanks for this fix. This is the correct thing to do. It seems obvious now, I don't know why I didn't notice this when first implementing this code.

Note that the position update could be moved further down a few lines to after the lines setting up the density. This will skip the update for particles removed because the density is too low.

Also, I notice that `theta` and `xb` will need to be updated to the advanced position since they are used below. Note that for RZ, `xb` will be the radius. It might be easier to completely separate the RZ and XY code below (in their own `ifdef` blocks).

RemiLehe: Thanks for the review.
I updated the PR and moved `UpdatePosition` further down.
I also removed the variables `xb` and `yb` (which, in practice, are only used for the RZ case) and replaced it by the more explicit variable `radial_position` (which correspond to the `r` position **before** updating the particle position.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1682,13 +1679,17 @@ PhysicalParticleContainer::AddPlasmaFlux (amrex::Real dt)
...
-
-                // Save the x and y values to use in the insideBounds checks.
-                // This is needed with WARPX_DIM_RZ since x and y are modified.
-                Real xb = ppos.x;
-                Real yb = ppos.y;
+                // Lab-frame simulation
+                // If the particle is not within the species's
+                // xmin, xmax, ymin, ymax, zmin, zmax, go to
+                // the next generated particle.
+                if (!inj_pos->insideBounds(ppos.x, ppos.y, ppos.z)) {
```
RemiLehe, line 1686: This check is now done **before** calling `theta = 2._prt*MathConst::pi*amrex::Random(engine);` a few lines below. This will therefore change the random number generation, and probably explains why I had to reset the benchmarks.

dpgrote (approved):
This looks good now - thanks for the updates.

PR #3526, opened 2022-11-15
Fix bug for intra-species Coulomb collisions

Yin-YinjianZhao:
For inner species collisions, there is a bug of calculating the number density.
This PR aims to fix this bug.


Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
```diff
@@ -82,18 +83,23 @@ void ElasticCollisionPerez (
...
-      {
-        n12 += amrex::min( w1[ I1[i1] ], w2[ I2[i2] ] );
-        ++i1; if ( i1 == static_cast<int>(I1e) ) { i1 = I1s; }
-        ++i2; if ( i2 == static_cast<int>(I2e) ) { i2 = I2s; }
-      }
-      n12 = n12 / dV;
+    if ( isSameSpecies ) {
+        for (int i1=I1s; i1<static_cast<int>(I1e); ++i1) { n1 += w1[ I1[i1] ]; }
+        for (int i2=I2s; i2<static_cast<int>(I2e); ++i2) { n1 += w2[ I2[i2] ]; }
+        n1 = n1 / dV; n2 = n1; n12 = n1;
```
RemiLehe: I agree with the calculation of `n1` and `n2` here, but I think that the calculation of `n12` is too simple (i.e. only valid in the case where all weights are equal). I think that we should instead use the same code that calculates `n12` below.

Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
```diff
@@ -82,18 +83,23 @@ void ElasticCollisionPerez (
...
+    } else {
+        for (int i1=I1s; i1<static_cast<int>(I1e); ++i1) { n1 += w1[ I1[i1] ]; }
```
RemiLehe: It seems that the calculation of `n2` is missing here...

Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
```diff
@@ -82,18 +83,23 @@ void ElasticCollisionPerez (
...
+        {
+          int i1 = I1s; int i2 = I2s;
+          for (int k = 0; k < amrex::max(NI1,NI2); ++k)
+          {
+            n12 += amrex::min( w1[ I1[i1] ], w2[ I2[i2] ] );
+            ++i1; if ( i1 == static_cast<int>(I1e) ) { i1 = I1s; }
+            ++i2; if ( i2 == static_cast<int>(I2e) ) { i2 = I2s; }
+          }
+          n12 = n12 / dV;
+        }
```
RemiLehe: I think that this should be outside of the `else` statement, as it also applies for intraspecies.
The only thing is that we would need to add these additional lines at the end, in order to match formula 38 from the Higginson paper:
```
if (isSameSpecies) {
    n12 *= 2;
}
```

Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
```diff
@@ -82,18 +83,23 @@ void ElasticCollisionPerez (
...
+    if ( isSameSpecies ) {
+        for (int i1=I1s; i1<static_cast<int>(I1e); ++i1) { n1 += w1[ I1[i1] ]; }
+        for (int i2=I2s; i2<static_cast<int>(I2e); ++i2) { n1 += w2[ I2[i2] ]; }
+        n1 = n1 / dV; n2 = n1;
+        n12 *= 2;
+    } else {
+        for (int i1=I1s; i1<static_cast<int>(I1e); ++i1) { n1 += w1[ I1[i1] ]; }
+        for (int i2=I2s; i2<static_cast<int>(I2e); ++i2) { n2 += w2[ I2[i2] ]; }
+        n1 = n1 / dV; n2 = n2 / dV;
     }
```
EZoni: Seems like the temporary declarations of `i1` and `i2` inside the `for` loops shadow previous local declarations of the same variables (line 86 above). The CI builds treat all warnings as errors and thus fail.

Regression/WarpX-tests.ini
```diff
@@ -2204,6 +2204,24 @@ compareParticles = 0
...
+addToCompileString =
+cmakeSetupOpts = -DWarpX_DIMS=3
+restartTest = 0
+useMPI = 1
+numprocs = 1
+useOMP = 1
+numthreads = 1
+compileTest = 0
+doVis = 0
+compareParticles = 0
```
RemiLehe:
```suggestion
compareParticles = 1
```

Examples/Tests/collision/analysis_collision_3d_isotropization.py
```diff
@@ -0,0 +1,68 @@
...
+
+tolerance = 0.05
+error = np.maximum(abs(Tx-Tx0/e)/Tx,abs(Ty-Ty0/e)/Ty)
+
+print('error = ', error)
+print('tolerance = ', tolerance)
+assert(error < tolerance)
+
+test_name = os.path.split(os.getcwd())[1]
+checksumAPI.evaluate_checksum(test_name, fn, do_particles=False)
```
RemiLehe:
```suggestion
checksumAPI.evaluate_checksum(test_name, fn)
```

Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
```diff
@@ -82,18 +83,25 @@ void ElasticCollisionPerez (
...
+    n12 = n12 / dV;
+    if ( isSameSpecies ) {
+        for (int i1=I1s; i1<static_cast<int>(I1e); ++i1 ) { n1 += w1[ I1[i1] ]; }
+        for (int i2=I2s; i2<static_cast<int>(I2e); ++i2 ) { n1 += w2[ I2[i2] ]; }
+        n1 = n1 / dV; n2 = n1;
+        n12 *= 2;
+    } else {
+        for (int i1=I1s; i1<static_cast<int>(I1e); ++i1 ) { n1 += w1[ I1[i1] ]; }
+        for (int i2=I2s; i2<static_cast<int>(I2e); ++i2 ) { n2 += w2[ I2[i2] ]; }
+        n1 = n1 / dV; n2 = n2 / dV;
```
RemiLehe:
```suggestion
    for (int i1=I1s; i1<static_cast<int>(I1e); ++i1 ) { n1 += w1[ I1[i1] ]; }
    for (int i2=I2s; i2<static_cast<int>(I2e); ++i2 ) { n1 += w2[ I2[i2] ]; }
    n1 = n1 / dV; n2 = n1;
    if ( isSameSpecies ) {
        n12 *= 2;
```

EZoni:
Note that the test `collisionXZ` is currently failing with an error of `0.0010291030859550358` vs a tolerance of `0.001`. I guess we can probably increase the tolerance slightly to, say, `0.002`, but maybe, @RemiLehe or @Yin-YinjianZhao, would you like to confirm that this is expected?

RemiLehe:
Yes, I think it is expected that the changes in this PR are affecting the test `collisionXZ`, because this test does involve intra-species collisions.

It is unclear to me, however, whether increasing the tolerance is the best course of action. We could instead update the "theoretical" predictions that we are comparing the simulation to. If I remember correctly, the "theoretical" prediction was not based on analytical theory, but was instead just a fit of the simulation results for a given version of WarpX. @Yin-YinjianZhao Could you confirm? 
If so, I think that we should update this fit - or even better: find the relevant analytical predictions.

Yin-YinjianZhao:
> Yes, I think it is expected that the changes in this PR are affecting the test `collisionXZ`, because this test does involve intra-species collisions.
> 
> It is unclear to me, however, whether increasing the tolerance is the best course of action. We could instead update the "theoretical" predictions that we are comparing the simulation to. If I remember correctly, the "theoretical" prediction was not based on analytical theory, but was instead just a fit of the simulation results for a given version of WarpX. @Yin-YinjianZhao Could you confirm? If so, I think that we should update this fit - or even better: find the relevant analytical predictions.

@RemiLehe Yes, it was based on a fit. Maybe we could follow Smeli's other tests and use the formulas they mentioned.

RemiLehe:
@Yin-YinjianZhao OK, thanks, could you send a link to that specific Smilei test?

RemiLehe:
After discussing with @Yin-YinjianZhao , the current `collisionXZ` test maybe not one of the Smilei test, but in the future, we could replace it by the first test here: https://smileipic.github.io/Smilei/Understand/collisions.html#test-cases-for-collisions since it has an analytical formula.

Examples/Tests/collision/analysis_collision_3d_isotropization.py
```diff
@@ -0,0 +1,68 @@
...
+ad  = ds.all_data()
+vx  = ad['electron', 'particle_momentum_x'].to_ndarray()/m
+vy  = ad['electron', 'particle_momentum_y'].to_ndarray()/m
+Tx = np.mean(vx**2)*m/e
+Ty = np.mean(vy**2)*m/e
+
+nt = 100
+Tx0 = T_par
+Ty0 = T_per
+for i in range(nt-1):
```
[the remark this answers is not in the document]
EZoni:
```suggestion
for _ in range(nt-1):
```

Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
```diff
@@ -95,6 +103,8 @@ void ElasticCollisionPerez (
       }
       n12 = n12 / dV;
     }
+    // Intra-species: Apply correction in collision rate
+    if ( isSameSpecies ) n12 *= T_PR(2.0);
```
EZoni:
```suggestion
    if (isSameSpecies) n12 *= T_PR(2.0);
```

Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
```diff
@@ -84,6 +86,12 @@ void ElasticCollisionPerez (
     T_PR n12 = T_PR(0.0);
     for (int i1=I1s; i1<static_cast<int>(I1e); ++i1) { n1 += w1[ I1[i1] ]; }
     for (int i2=I2s; i2<static_cast<int>(I2e); ++i2) { n2 += w2[ I2[i2] ]; }
+    // Intra-species: the density is in fact the sum of the density of
+    // each sub-group of particles
+    if ( isSameSpecies ) {
```
EZoni:
```suggestion
    if (isSameSpecies) {
```

EZoni (changes requested):

RemiLehe (approved):

EZoni (approved):

PR #3529, opened 2022-11-17
Speed up Vay current deposition

AlexanderSinn:
The Vay deposition deposits eight different values into each of the three current arrays jx, jy and jz. However, these eight values are the same between the three currents, just with different prefactors.
This PR proposes to use four temporary arrays to store the values, and to obtain jx, jy and jz after that through linear combinations of the temporary arrays. This approach reduces the number of AtomicAdds by a factor of three (3d) or 1.5 (2d).

Test on one A100:
dev:
```
TinyProfiler total time across processes [min...avg...max]: 21.53 ... 21.53 ... 21.53

----------------------------------------------------------------------------------------------------------
Name                                                       NCalls  Excl. Min  Excl. Avg  Excl. Max   Max %
----------------------------------------------------------------------------------------------------------
WarpXParticleContainer::DepositCurrent::CurrentDeposition      40      16.07      16.07      16.07  74.63%
WarpX::OneStep_nosub()                                         20      2.111      2.111      2.111   9.80%
PhysicalParticleContainer::Evolve::GatherAndPush               40      0.865      0.865      0.865   4.02%
SpectralSolver::ForwardTransform                              240     0.7541     0.7541     0.7541   3.50%
SpectralSolver::BackwardTransform                             180     0.5473     0.5473     0.5473   2.54%

...

END REGION WarpX::Evolve()
Total GPU global memory (MB) spread across MPI: [40388 ... 40388]
Free  GPU global memory (MB) spread across MPI: [29786 ... 29786]
[The         Arena] space (MB) allocated spread across MPI: [30291 ... 30291]
[The         Arena] space (MB) used      spread across MPI: [0 ... 0]
[The  Device Arena] space (MB) allocated spread across MPI: [8 ... 8]
[The  Device Arena] space (MB) used      spread across MPI: [0 ... 0]
[The  Pinned Arena] space (MB) allocated spread across MPI: [8 ... 8]
[The  Pinned Arena] space (MB) used      spread across MPI: [0 ... 0]
AMReX (22.11-17-gd8bc97f92a1a) finalized
```

PR:
```
TinyProfiler total time across processes [min...avg...max]: 10.51 ... 10.51 ... 10.51

----------------------------------------------------------------------------------------------------------
Name                                                       NCalls  Excl. Min  Excl. Avg  Excl. Max   Max %
----------------------------------------------------------------------------------------------------------
WarpXParticleContainer::DepositCurrent::CurrentDeposition      40      5.046      5.046      5.046  48.00%
WarpX::OneStep_nosub()                                         20      2.111      2.111      2.111  20.08%
PhysicalParticleContainer::Evolve::GatherAndPush               40     0.8649     0.8649     0.8649   8.23%
SpectralSolver::ForwardTransform                              240     0.7541     0.7541     0.7541   7.17%
SpectralSolver::BackwardTransform                             180     0.5472     0.5472     0.5472   5.20%

...

END REGION WarpX::Evolve()
Total GPU global memory (MB) spread across MPI: [40388 ... 40388]
Free  GPU global memory (MB) spread across MPI: [28890 ... 28890]
[The         Arena] space (MB) allocated spread across MPI: [30291 ... 30291]
[The         Arena] space (MB) used      spread across MPI: [0 ... 0]
[The  Device Arena] space (MB) allocated spread across MPI: [8 ... 8]
[The  Device Arena] space (MB) used      spread across MPI: [0 ... 0]
[The  Pinned Arena] space (MB) allocated spread across MPI: [8 ... 8]
[The  Pinned Arena] space (MB) used      spread across MPI: [0 ... 0]
AMReX (22.11-17-gd8bc97f92a1a) finalized
```

input:
```
# algo
algo.charge_deposition = standard
algo.current_deposition = vay
algo.maxwell_solver = psatd
algo.particle_pusher = vay
algo.particle_shape = 3

# amr
amr.max_level = 0
amr.n_cell = 256 256 256

# boundary
boundary.field_hi = periodic periodic periodic
boundary.field_lo = periodic periodic periodic

# diag

# diagnostics
diagnostics.enable = 0

# geometry
geometry.coord_sys = 0
geometry.dims = 3
geometry.prob_hi =  45.  45.  45.
geometry.prob_lo = -45. -45. -45.

amr.max_grid_size = 512

# max_step
max_step = 20

# particles

# psatd
psatd.nox = 4
psatd.noy = 4
psatd.noz = 4
psatd.nx_guard = 20
psatd.ny_guard = 20
psatd.nz_guard = 20
psatd.periodic_single_box_fft = 0
psatd.update_with_rho = 0

# warpx
warpx.cfl = 0.9999
warpx.do_nodal = 1
warpx.use_filter = 1
warpx.verbose = 1

electrons.charge = -q_e
electrons.mass = m_e
electrons.injection_style = "NUniformPerCell"
electrons.num_particles_per_cell_each_dim = 1 1 1
electrons.profile = constant
electrons.density = 1.e25
electrons.ux_th  = 1. # uth the std of the (unitless) momentum
electrons.uy_th  = 1. # uth the std of the (unitless) momentum
electrons.uz_th  = 1. # uth the std of the (unitless) momentum

positrons.charge = q_e
positrons.mass = m_e
positrons.injection_style = "NUniformPerCell"
positrons.num_particles_per_cell_each_dim = 1 1 1
positrons.profile = constant
positrons.density = 1.e25
positrons.momentum_distribution_type = "gaussian"
positrons.ux_th  = 1. # uth the std of the (unitless) momentum
positrons.uy_th  = 1. # uth the std of the (unitless) momentum
positrons.uz_th  = 1. # uth the std of the (unitless) momentum
```


EZoni (commented):
Thank you, @AlexanderSinn! This looks great, thanks for speeding up this kernel.

A couple of minor questions, that you could maybe comment on:
1. My understanding is that the code blocks within the `if (i_new==i_old && j_new==j_old && k_new==k_old)` kind of conditions are a simpler and more optimized case of the more general code implemented in the corresponding `else` blocks. Is that right? If so, did you have a chance to measure how much removing these `if` blocks impact the overall performance of the kernel for the test case that you considered? 

2. I was kind of expecting some issues with using the indices of the `ParallelFor` loops on the Jx box to access the data of Jy and Jz, due to the different index types (staggering) among the three quantities in the non-nodal case (a typical issue would be out-of-bound access due to missing last nodal point in directions where the data being accessed are cell-centered instead of nodal). However, CI didn't flag any issue, not even in DEBUG mode, which I tried briefly locally. So maybe I'm missing something here. What's your opinion on this?

AlexanderSinn:
1.	Correct. Here is the Excl. runtime of CurrentDeposition for different u_th:
```
no if
0.01:  5.130s
0.1:   5.143s
1:     5.454s
10:    5.500s
100:   5.498s

with if
0.01:  4.699s
0.1:   4.701s
1:     5.047s
10:    5.084s
100:   5.087s
```
2.	I did add an assert that the jy box and jz box should be equal to the jx box. This also tests for nodal vs. cell centered. I am not to familiar with the staggering of the Vay deposition but I assume it is supposed to be always the same for jx, jy and jz as there is only one shape factor calculated per direction and new/old position unlike the normal current deposition that has two.

EZoni:
Ah, of course, yes, I forgot about that while writing my question earlier.

The Vay current (which is not yet the true physical current "J" at this point in the PIC loop) has the same staggering as rho, for all components: https://github.com/ECP-WarpX/WarpX/blob/2775ac17fc78b3433d313da46a1c81f932e3912e/Source/WarpX.cpp#L1891-L1896

I got confused because everything here was already called `jx`, `jy`, `jz`, etc., but all of these should be actually called `Dx`, `Dy`, `Dz`, etc., because that is the notation used in the literature (see, for example, ([Vay et al, JCP, 2013](https://doi.org/10.1016/j.jcp.2013.03.010), equations (20)-(24)).

Maybe this PR is also a good chance to rename those quantities from `j`s to `D`s, and make the code consistent with the notation used in the literature, similarly to what I did here when I implemented the cumulative sums: https://github.com/ECP-WarpX/WarpX/blob/2775ac17fc78b3433d313da46a1c81f932e3912e/Source/FieldSolver/WarpXPushFieldsEM.cpp#L435-L437

Do you think you could do that within this PR? If you want to keep the PR smaller instead, we/I can do it in a separate PR. Just let me know what is your preference. Thanks again for working on this!

AlexanderSinn:
You can do it in a separate PR. The naming will be difficult to fix in the `WarpXParticleContainer::DepositCurrent` function.

EZoni (approved):

RemiLehe:
I agree with having a follow-up PR that changes the naming.
In addition, a follow-up PR could also add a few more comments to explain the overall scheme here.

PR #3532, opened 2022-11-21
Assert for Diag Type

RevathiJambunathan:
Many times, we can have a typo in setting diag_type
For example : BackTranformed  instead of BackTransformed

and the results will be incorrect. 
This PR asserts that the user-input matches the three available diagnostic types


EZoni (approved):
Thank you for adding this check, @RevathiJambunathan!

PR #3536, opened 2022-11-24
Allow initial analytic fields to be set through the `picmi` interface

roelof-groenewald:
The python interface allows external fields applied directly to particles to be set but not initial fields which will be evolved with the field solver. This PR adds a `AnalyticInitialField` class to `picmi.py` that sets the initial fields in a simulation.


dpgrote (approved):
This look good. Thanks for adding it!

RemiLehe (approved):

Issue #3513, opened 2022-11-09
Possible bug: Numerical heating instability in 1D


snicks11:
I am attempting a simple benchmark of a uniform plasma of ions and electrons in 1D. I am using the Yee electromagnetic solver (energy-conserving mode) and default Boris particle push. The grid spacing is set to the Debye length, and the boundaries are periodic for particles and fields. I observe that the total particle energy grows exponentially over the course of ~ 100 plasma oscillation periods without saturation. (If this were simply a case of the grid spacing being too large, the heating would saturate when the new Debye length matches the grid spacing.) Unphysical waves are also present, while the expected plasma oscillation frequency is apparently absent. The problem is even worse when an electrostatic solver is used. 

When I run the same setup in 2D, however (essentially turning the 1D domain into a square), energy conservation works correctly. Any help would be appreciated. I have attached pictures of my PICMI input file below. 

![image](https://user-images.githubusercontent.com/102836570/200706345-36760928-20f6-4439-84a2-3862b4c42382.png)
![image](https://user-images.githubusercontent.com/102836570/200706276-bd67385c-1bc3-4d3d-b657-1eae53432588.png)
![image](https://user-images.githubusercontent.com/102836570/200706128-edf9e5e4-044c-40b0-97fc-9e1d20fdcd36.png)
![image](https://user-images.githubusercontent.com/102836570/200706513-cad5bf5e-5590-4a03-9abb-7f727a0bd982.png)
![image](https://user-images.githubusercontent.com/102836570/200706682-aab11877-5dc8-4c4a-9b4d-b04e7972fe15.png)

[Edit] Here is a plot of the total particle energy in a typical run that I am seeing. The instability is quite severe.
![image](https://user-images.githubusercontent.com/102836570/200904226-dd6d7c9a-3706-4e4e-84ce-aa14378a423a.png)

roelof-groenewald:
@prkkumar do you have any thoughts on what might be causing this numerical heating in 1D? When you implemented the 1D support did you by chance run a similar benchmark? I'm wondering if this has always been present or if some recent changes broke something.

prkkumar:
Hi, thanks for reporting this. The two 1D benchmarks with particles and Yee solver that we originally ran are these:

1. https://github.com/ECP-WarpX/WarpX/blob/development/Examples/Physics_applications/laser_acceleration/inputs_1d
[...]

Issue #2819, opened 2022-02-02
Zero Output for Particle Diags in 1D


EZoni:
Looking at the checksum benchmarks for the tests
- `Langmuir_multi_1d`
- `LaserAcceleration_1d`
- `Python_LaserAcceleration_1d`
- `Python_PlasmaAcceleration1d`

it seems like there might be some issue with the particle diagnostics in 1D, as all particle attributes written to output seem to be zero:
https://github.com/ECP-WarpX/WarpX/blob/295d78082e1b9cdb9cbd9ad5c128fcc44610d1d0/Regression/Checksum/benchmarks_json/Langmuir_multi_1d.json#L2-L10
https://github.com/ECP-WarpX/WarpX/blob/295d78082e1b9cdb9cbd9ad5c128fcc44610d1d0/Regression/Checksum/benchmarks_json/Langmuir_multi_1d.json#L22-L30
https://github.com/ECP-WarpX/WarpX/blob/295d78082e1b9cdb9cbd9ad5c128fcc44610d1d0/Regression/Checksum/benchmarks_json/LaserAcceleration_1d.json#L2-L10
https://github.com/ECP-WarpX/WarpX/blob/295d78082e1b9cdb9cbd9ad5c128fcc44610d1d0/Regression/Checksum/benchmarks_json/Python_LaserAcceleration_1d.json#L2-L10
https://github.com/ECP-WarpX/WarpX/blob/295d78082e1b9cdb9cbd9ad5c128fcc44610d1d0/Regression/Checksum/benchmarks_json/Python_PlasmaAcceleration1d.json#L11-L19

ax3l:
I think the reason could be that we by default apply a geometry filter for the ProbDomain.

In 1D, if we initialize the particles not with x and y exactly on axis, we might filter them out...?

ax3l:
cc @prkkumar do you like to take a look at this 1D issue?

prkkumar:
Yes, thank you! I will look into it.

RevathiJambunathan:
Update : 
this seems to be a post-processing error.
in yt, when processing plotfile, 
we get electrons and positrons listed, but possibly a yt-thing that it's not being read correctly

EZoni:
@atmyers @dpgrote 
Please feel free to confirm and close this issue,
it seems to me like https://github.com/yt-project/yt/pull/4112 fixed it.

dpgrote:
[...]

PR #3537, opened 2022-11-27
Fix Esirkepov deposition in 1D

RemiLehe:
There were a few typos in the 1D code for Esirkepov.

This correction naturally leads to small changes in the benchmark files.


Regression/Checksum/benchmarks_json/LaserAcceleration_1d.json
```diff
@@ -1,23 +1,23 @@
...
-    "particle_orig_z": 0.022432812500000038,
-    "particle_position_x": 0.02243266637270741,
-    "particle_regionofinterest": 40.0,
-    "particle_weight": 5.20625e+18
+    "particle_momentum_y": 0.0,
+    "particle_momentum_z": 0.0,
+    "particle_orig_z": 0.0,
+    "particle_position_x": 0.0,
+    "particle_regionofinterest": 0.0,
+    "particle_weight": 0.0
```
EZoni: Note that these zeros might be related to #2819.

EZoni (approved):
Thank you for this fix, @RemiLehe! Looks good to me.

EZoni:
Note that the checksums of the CI test `LaserAcceleration_1d` seem to need another reset.

It might be that our automatic tool to reset checksums does not work for particle data in 1D
(maybe because of some missing fix on the `yt` side).

Update: correct, the problem was only local, though.

roelof-groenewald (approved):
Thanks for looking into and fixing this issue. 
Fixes #3513.

PR #3538, opened 2022-11-28
Enable the use of `FieldProbe` default parameter values

roelof-groenewald:
The reduced diagnostic `FieldProbe` currently assigns default values to various parameters but the diagnostic initialization uses `utils::parser::getWithParser()` to extract user provided values for these parameters - which errors if the keyword is not in the input file. This PR switches those instances to using `utils::parser::queryWithParser()` which allows the default values to be used if the user didn't specify an alternative value.


RemiLehe (approved):

PR #3542, opened 2022-12-01
Clean up documentation

dpgrote:
I was testing the build of the documentation and noticed that there were many warnings being reported, both in the rst files and in the documenting comments in the source code. I had a few spare minutes so I went through and fixed them. Most of the warnings were from arguments to function methods being missed. There were also a number of syntax problems in the comments. This PR touches many places in the C++, but nearly all are in the comments.

The CI tests already build the docs checking for errors. Could this be changed so that it fails on warnings as well? This can be done by added the `-W` option when running Sphinx. This will help reduce comment creep, where the comments become outdated over time and they are not updated along with the associated code.

Another comment is that there are several different styles of comments. For example, some use "\param" and some "@param" for arguments. Also, some use "[in]" and "[out]" while many others do not. Is it worth making it consistent?

### DIFF
--- a/Docs/source/index.rst
+++ b/Docs/source/index.rst
@@ -96,7 +96,6 @@ Data Analysis
    dataanalysis/visit
    dataanalysis/visualpic
    dataanalysis/picviewer
-   dataanalysis/backtransformed_diags
    dataanalysis/reduced_diags
 
 Theory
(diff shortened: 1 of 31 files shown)


Docs/source/index.rst
```diff
@@ -96,7 +96,6 @@ Data Analysis
    dataanalysis/visit
    dataanalysis/visualpic
    dataanalysis/picviewer
-   dataanalysis/backtransformed_diags
```
ax3l, line 99: Did we remove the file already?
Shall we (in a follow-up) add a new one? cc @RevathiJambunathan ?

ax3l (approved):
Thank you for the fixes!!

PR #3543, opened 2022-12-01
[WIP] Clean up examples folders

EZoni:
With @RemiLehe we discussed today that it could be a good idea to move all examples folders under `Examples/` instead of keeping the three separate folders `Examples/Modules/`, `Examples/Physics_applications/` and `Examples/Tests/`, which do not make much more sense at this stage of the WarpX development.

Besides moving all examples folders under `Examples/`, I also renamed all folders following the `snake_case` naming convention, ensuring consistency. It would be nice to keep such consistency in the future, I personally think it makes the repository look less messy to new users/developers.

This could be a good opportunity to see whether we want to improve any of the test folder names under `Examples/`, possibly making them more descriptive and/or self-explanatory.

Also to be discussed whether we want to enfore the same naming convention for the test names themselves, too.

To-do:
- [ ] Reinstate `Examples/Physics_applications` and just merge `Examples/Modules/` with `Examples/Tests/`
- [x] Move folders
- [x] Use `snake_case` naming convention for folders
- [x] Update CI tools and docs with new paths 
- [x] Rename `nci_corrector` as `nci_fdtd`
- [ ] Make dir `nci_psatd` and merge NCI PSATD inputs
- [ ] Use snake case naming convention for test names?


EZoni:
Discussed again with @RemiLehe:
We actually want to keep `Examples/Physics_applications/` and just merge `Examples/Modules/` with `Examples/Tests`. It'll be faster to close this PR and open a new one, rather than fixing this one.

PR #3545, opened 2022-12-01
Clean up examples folders

EZoni:
Replace #3543.

With @RemiLehe we discussed that it could be a good idea to merge all examples folders under `Examples/Modules` with those under `Examples/Tests/`.

We could also rename all folders following the `snake_case` naming convention, ensuring consistency. It would be nice to keep such consistency in the future, I personally think it makes the repository look less messy to new users/developers.

This could be a good opportunity to see whether we want to improve any of the test folder names, possibly making them more descriptive and/or self-explanatory.

Also to be discussed whether we want to enfore the same naming convention for the test names themselves, too.

To-do:
- [x] Merge `Examples/Modules` with `Examples/Tests`
- [x] Use `snake_case` naming convention for folders
- [x] Update CI tools and docs with new paths 
- [x] Rename `nci_corrector` as `nci_fdtd`


RemiLehe (approved):
Great, thanks a lot for cleaning up!

PR #3546, opened 2022-12-02
Update docutils version

dpgrote:
This updates the Docutils and Sphinx to recent versions. The `Docs/requirements.txt` had commented that there were problems with the new Docutils, but this has since been fixed. Also, the `Docs/Doxyfile` was updated doing `doxygen -u`.

My local tests show that the html is properly generated with the new versions.


ax3l:
Thanks a lot. I took a look at the PR RTD preview and this looks great!

Looks like docutils and sphinx finally synced up their HTML5 style issues :)

ax3l (approved):

PR #3548, opened 2022-12-06
[WIP] CI: improve analysis for test using div(E) cleaning

EZoni:
For CI tests using the div(E) cleaning method, we can simply check charge conservation in the same way as we do for current correction or Vay deposition. Errors on div(E)=rho are supposed to be propagated away through the field F. 

Thank you to @oshapoval for pointing this out!


EZoni:
I opened this PR some time ago, but I think we can leave this specific test as is for now. A more thorough improvement of all custom analysis scripts in our regression tests suite should better fit into a large-scale strategy.

PR #3549, opened 2022-12-07
Fix the CUDA Aware part of the Perlmutter script.

kngott:
As a Gordon Bell winner, I take no responsibility for this bug.

But, I'll fix it. And it gives solid performance enhancement, so I will take responsibility for that. 😄


RemiLehe (approved):

RemiLehe:
Thanks for fixing this!

PR #3551, opened 2022-12-07
Feature magnetostatic solver

clarkse:
A magnetostatic field solver has been implemented  to allow for calculation of beam self fields in the presence of embedded conducting boundaries using the same linear operator and the Amrex MLMG Poisson solver as the laboratory electrostatic solver.  This solver is nodal, so requires currents deposited at nodal locations (i.e. do_current_centering = true).  This is required for using the unedited linear operator that is solving the electrostatic solver.  The A field solution is then nodal in each direction, and when gradients are taken, the derivatives are located on the cell edges, which need to be interpolated to the appropriate face when computing curl operation for different components of B.

A test problem was devised where a uniform cylindrical beam is transiting a beam pipe and the E and B fields are computed, then compared to the simple analytical form.  The magnetostatic solution has roughly a factor of 2 larger error than the electrostatic solver, and this probably originates from combining multiple solutions into the curl and the additional interpolation steps.

bth_rz

The is the output from the RZ test script comparing Btheta from the solved solution to the analytical case.  The error is computed inside of the beampipe and matches on a similar order to the electrostatic solution of Er.  The larger error in the solution compared to the residual appears to be due to the solution not being centered appropriately, and would probably match the analytical solution better if corrected based on the locations of the field components with respect to the mesh/node location.

The boundary conditions on the domain edges are correct for PEC and Neumann boundary conditions, however the embedded boundary currently only supports homogeneous Dirichlet boundaries. 

Next steps for the magnetostatic solver to make it relevant for injector simulation and design include potentially adding a new linear operator to apply the correct boundary conditions at the embedded boundary.  The PEC boundary will couple the components of A such that they must be solved simultaneously.

https://github.com/AMReX-Codes/amrex/pull/3035 

The above AMREX PR should allow for 3 Multifabs to be solved simultaneously, and development of an extended operator to implement the boundary conditions will need to be done in WarpX/AMREX.  This will alleviate the need for current centering, and the gradients will naturally reside in the appropriate face centered locations for computing the curl.


Source/Utils/WarpXAlgorithmSelection.cpp
```diff
@@ -33,7 +33,8 @@ const std::map<std::string, int> electromagnetic_solver_algo_to_int = {
 
 const std::map<std::string, int> electrostatic_solver_algo_to_int = {
     {"none", ElectrostaticSolverAlgo::None },
-    {"relativistic", ElectrostaticSolverAlgo::Relativistic},
+    {"relativistic", ElectrostaticSolverAlgo::RelativisticApproximate},
+    {"relativistic-magnetostatic", ElectrostaticSolverAlgo::RelativisticMagnetostatic},
```
RemiLehe: Could you update the WarpX documentation here: https://github.com/ECP-WarpX/WarpX/blob/development/Docs/source/usage/parameters.rst for the parameter `do_electrostatic`, and include the option `relativistic-magnetostatic`, with a brief description of what distinguishes it from the `labframe` and `relativistic` solver?

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -0,0 +1,543 @@
...
+                     self_fields_verbosity );
+}
+
+/* Compute the vector potential `A` by solving the Poisson equation with `J` as
+   a source.
+   This uses the amrex solver.
+
+   More specifically, this solves the equation
+   \f[
+       \vec{\nabla}^2 r \phi - (\vec{\beta}\cdot\vec{\nabla})^2 r \phi = -\frac{r \rho}{\epsilon_0}
```
RemiLehe: Could you update the documentation to include the equation for A?

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -0,0 +1,611 @@
...
+
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(do_current_centering == true,
+                                     "Error: Magnetostatic solver requires warpx.do_current_centering=true");
+
+    AddCurrentDensityFieldLabFrame();
+
+    // Transfer fields from 'fp' array to 'aux' array.
+    // This is needed when using momentum conservation
+    // since they are different arrays in that case.
+    UpdateAuxilaryData();
```
RemiLehe: I am not sure I understand this part. Maybe we could discuss this off line.

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -0,0 +1,611 @@
...
+    for (int lev = 0; lev <= max_level; lev++) {
+        ApplyInverseVolumeScalingToCurrentDensity(current_fp[lev][0].get(),
+                                                  current_fp[lev][1].get(),
+                                                  current_fp[lev][2].get(), lev);
+    }
+#endif
+
+    // NOTE:  (SEC 12/2/22)
+    // Interpolate fields back to nodal current
+    // This could be done faster, but want to check if answer is right
```
RemiLehe: Yes, I think that the code could be potentially simplified by adding the following block here:
https://github.com/ECP-WarpX/WarpX/blob/development/Source/WarpX.cpp#L1838
```
if (WarpX::electrostatic_solver == RelativisticMagnetostatic) {
        jx_nodal_flag  = IntVect::TheNodeVector();
        jy_nodal_flag  = IntVect::TheNodeVector();
        jz_nodal_flag  = IntVect::TheNodeVector();
}
```
In this case, all the operations in `AddCurrentDensityFieldLabFrame` could be done directly on `current_fp` (without ever having to touch `current_nodal_fp` and/or interpolate back and forth between `currenf_fp` and `current_nodal_fp`). We would also remove the need to set `do_current_centering`.

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -0,0 +1,611 @@
...
+                                                        true);
+
+        }
+    }
+
+    // set the boundary and current density potentials
+    setVectorPotentialBC(vector_potential_fp_nodal);
+    // setVectorPotentialBC(current_fp_nodal);
+
+    // Compute the potential phi, by solving the Poisson equation
```
RemiLehe:
```suggestion
    // Compute the vector potential A, by solving the Poisson equation
```

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -0,0 +1,611 @@
...
+   This uses the amrex solver.
+
+   More specifically, this solves the equation
+   \f[
+       \vec{\nabla}^2 r \phi - (\vec{\beta}\cdot\vec{\nabla})^2 r \phi = -\frac{r \rho}{\epsilon_0}
+   \f]
+
+   \param[in] rho The charge density a given species
+   \param[out] phi The potential to be computed by this function
+   \param[in] beta Represents the velocity of the source of `phi`
```
RemiLehe:
```suggestion
   \param[in] curr The current density
   \param[out] A The vector potential to be computed by this function
```

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver_K.H
```diff
@@ -0,0 +1,251 @@
+/* Copyright 2022 S. Eric Clark, LLNL
```
RemiLehe: Could this file be moved into the `ablastr/fields` folder, in order to be consistent with this similar file:
https://github.com/ECP-WarpX/WarpX/tree/development/Source/ablastr/fields
In this case, we could rename this file `VectorPoissonSolver`.

RemiLehe (changes requested):
Thanks a lot for this PR, and sharing this code!
This could be very useful to other WarpX users!

I had a first rudimentary pass through the code, and this looks great. I added a few comments and suggestions ; would you have time to address them? In addition, I could you add a CI test that uses the regular WarpX interface (without Python) - so as to provide an example for users that do not use Python? This could be done e.g. by converting one of the existing Python tests to the regular WarpX interface, or by adding a new test.

I'll try to have a second pass through the code at the end of the week.

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -0,0 +1,611 @@
...
+                                                        WarpX::do_single_precision_comms,
+                                                        curr_period,
+                                                        true);
+
+        }
+    }
+
+    // set the boundary and current density potentials
+    setVectorPotentialBC(vector_potential_fp_nodal);
+    // setVectorPotentialBC(current_fp_nodal);
```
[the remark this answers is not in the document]
clarkse: Resolved this issue.

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
clarkse: So this scheme works well in 3D and eliminates interpolating back and forth between nodal and e-field staggered mesh locations..  The ApplyInverseVolumeScalingToCurrentDensity routine throws an Assertion error since it only works for cell edges.  It will need to be extended to work for Nodal currents for it to work correctly.

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
clarkse: I am unsure of what uses the aux arrays in other parts of the code, so to be consistent with the electrostatic solver I update these fields after modifying the regular E/B fields.

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
clarkse: So the weighting of Jr has been adjusted to allow for the nodal inverse current density scaling.  Please take a look.  Otherwise I think this should be implemented now.  Please take a look at the latest commit.

Examples/Tests/magnetostatic_eb/PICMI_inputs_3d.py
```diff
@@ -0,0 +1,280 @@
...
+
+matplotlib.use('agg')
+
+import matplotlib.pyplot as plt
+import numpy as np
+import scipy.constants as con
+
+from pywarpx import fields, picmi
+
+# from pywarpx.Algo import algo
```
RemiLehe: Could you clean the script by removing all the commented lines?
Also: could you add a summary of what this test does, at the top of the file (see for instance: https://github.com/ECP-WarpX/WarpX/blob/development/Examples/Tests/electrostatic_sphere/analysis_electrostatic_sphere.py#L10)

Examples/Tests/magnetostatic_eb/PICMI_inputs_rz.py
```diff
@@ -0,0 +1,232 @@
+#!/usr/bin/env python3
```
RemiLehe, line 1: Same comment as for the corresponding 3d script: could you clean up the script and add a high-level summary?

Examples/Tests/magnetostatic_eb/analysis_rz.py
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env python3
```
RemiLehe, line 1: This file should maybe be name `analysis_3d.py`

Python/pywarpx/_libwarpx.py
```diff
@@ -1462,6 +1472,32 @@ def get_mesh_magnetic_field_fp(self, level, direction, include_ghosts=True):
 
         return self._get_mesh_field_list(self.libwarpx_so.warpx_getBfieldFP, level, direction, include_ghosts)
 
+    def get_mesh_vector_potential_fp(self, level, direction, include_ghosts=True):
+        '''
+
+        This returns a list of numpy arrays containing the mesh magnetic field
```
RemiLehe:
```suggestion
        This returns a list of numpy arrays containing the mesh vector potential
```

Python/pywarpx/_libwarpx.py
```diff
@@ -2700,6 +2804,42 @@ def get_Jz_nodal_flag(self):
...
+
+    def get_JzNodal_nodal_flag(self):
+        '''
+        This returns a 1d array of the nodal flags for Jz along each direction. A 1 means node centered, and 0 cell centered.
+        '''
+        return self._get_nodal_flag(self.libwarpx_so.warpx_getJzNodal_nodal_flag)
+
+    def get_Ax_nodal_flag(self):
+        '''
+        This returns a 1d array of the nodal flags for Bx along each direction. A 1 means node centered, and 0 cell centered.
```
RemiLehe:
```suggestion
        This returns a 1d array of the nodal flags for Ax along each direction. A 1 means node centered, and 0 cell centered.
```

Python/pywarpx/_libwarpx.py
```diff
@@ -2700,6 +2804,42 @@ def get_Jz_nodal_flag(self):
...
+        '''
+        return self._get_nodal_flag(self.libwarpx_so.warpx_getAx_nodal_flag)
+
+    def get_Ay_nodal_flag(self):
+        '''
+        This returns a 1d array of the nodal flags for By along each direction. A 1 means node centered, and 0 cell centered.
```
RemiLehe:
```suggestion
        This returns a 1d array of the nodal flags for Ay along each direction. A 1 means node centered, and 0 cell centered.
```

Python/pywarpx/_libwarpx.py
```diff
@@ -2700,6 +2804,42 @@ def get_Jz_nodal_flag(self):
...
+        '''
+        return self._get_nodal_flag(self.libwarpx_so.warpx_getAy_nodal_flag)
+
+    def get_Az_nodal_flag(self):
+        '''
+        This returns a 1d array of the nodal flags for Bz along each direction. A 1 means node centered, and 0 cell centered.
```
RemiLehe:
```suggestion
        This returns a 1d array of the nodal flags for Az along each direction. A 1 means node centered, and 0 cell centered.
```

Python/pywarpx/_libwarpx.py
```diff
@@ -196,6 +196,8 @@ class Particle(ctypes.Structure):
         self.libwarpx_so.warpx_getCurrentDensityCPLoVects.restype = _LP_c_int
         self.libwarpx_so.warpx_getCurrentDensityFP.restype = _LP_LP_c_real
         self.libwarpx_so.warpx_getCurrentDensityFPLoVects.restype = _LP_c_int
+        self.libwarpx_so.warpx_getCurrentDensityFPNodal.restype = _LP_LP_c_real
```
RemiLehe: I am a bit worried that users of this interface might not understand the distinction between `CurrentDensityFP` and `CurrentDensityFPNodal` and might get confused. (Since the usage of the nodal current density is somewhat non-standard, and only used in some specific simulations of plasma accelerators.) 

Since the nodal current density is not needed anymore for this PR, would it be okay to remove these Python functions?

Python/pywarpx/_libwarpx.py
```diff
@@ -242,6 +246,12 @@ class Particle(ctypes.Structure):
         self.libwarpx_so.warpx_getJx_nodal_flag.restype = _LP_c_int
         self.libwarpx_so.warpx_getJy_nodal_flag.restype = _LP_c_int
         self.libwarpx_so.warpx_getJz_nodal_flag.restype = _LP_c_int
+        self.libwarpx_so.warpx_getJxNodal_nodal_flag.restype = _LP_c_int
```
RemiLehe: Same comment as above regarding the nodal J: this interface could be removed, as it is not strictly needed, and might be confusing.

Python/pywarpx/_libwarpx.py
```diff
@@ -1597,6 +1633,32 @@ def get_mesh_current_density_fp(self, level, direction, include_ghosts=True):
 
         return self._get_mesh_field_list(self.libwarpx_so.warpx_getCurrentDensityFP, level, direction, include_ghosts)
 
+    def get_mesh_current_density_fp_nodal(self, level, direction, include_ghosts=True):
```
RemiLehe: Same comment as above regarding the nodal J: this interface could be removed, as it is not strictly needed, and might be confusing.

Python/pywarpx/_libwarpx.py
```diff
@@ -2310,6 +2393,27 @@ def get_mesh_current_density_fp_lovects(self, level, direction, include_ghosts=T
         '''
         return self._get_mesh_array_lovects(level, direction, include_ghosts, self.libwarpx_so.warpx_getCurrentDensityFPLoVects)
 
+    def get_mesh_current_density_fp_nodal_lovects(self, level, direction, include_ghosts=True):
```
RemiLehe: Same comment as above regarding the nodal J: this interface could be removed, as it is not strictly needed, and might be confusing.

Python/pywarpx/_libwarpx.py
```diff
@@ -2700,6 +2804,42 @@ def get_Jz_nodal_flag(self):
         '''
         return self._get_nodal_flag(self.libwarpx_so.warpx_getJz_nodal_flag)
 
+    def get_JxNodal_nodal_flag(self):
```
RemiLehe: Same comment as above regarding the nodal J: this interface could be removed, as it is not strictly needed, and might be confusing.

Python/pywarpx/picmi.py
```diff
@@ -73,6 +73,10 @@ class Species(picmistandard.PICMI_Species):
         Whether or not to deposit the charge and current density for
         for this species
 
+    warpx_random_theta: bool, default=True
+        Whether or not to add random angle to the particles in theta
+        when in 2D-RZ mode.
```
RemiLehe: We usually simply call this mode RZ in WarpX.
```suggestion
        when in RZ mode.
```

Source/Diagnostics/Diagnostics.cpp
```diff
@@ -75,10 +75,18 @@ Diagnostics::BaseReadParameters ()
     // Sanity check if user requests to plot phi
     if (utils::algorithms::is_in(m_varnames_fields, "phi")){
         WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
-            warpx.electrostatic_solver_id==ElectrostaticSolverAlgo::LabFrame,
+            warpx.electrostatic_solver_id==ElectrostaticSolverAlgo::LabFrame ||
+            warpx.electrostatic_solver_id==ElectrostaticSolverAlgo::RelativisticMagnetostatic,
             "plot phi only works if do_electrostatic = labframe");
```
RemiLehe:
```suggestion
            "plot phi only works if do_electrostatic = labframe or do_electrostatic = relativistic-magnetostatic");
```

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -309,6 +309,12 @@ WarpX::Evolve (int numsteps)
             // and so that the fields are at the correct time in the output.
             bool const reset_fields = true;
             ComputeSpaceChargeField( reset_fields );
+            if (electrostatic_solver_id == ElectrostaticSolverAlgo::RelativisticMagnetostatic) {
+                // Call Magnetostatic SOlver to solver for the vector potential A and compute the
```
RemiLehe:
```suggestion
                // Call Magnetostatic Solver to solve for the vector potential A and compute the
```

Python/pywarpx/fields.py
```diff
@@ -669,6 +669,48 @@ def PhiFPWrapper(level=0, include_ghosts=False):
                             get_nodal_flag=libwarpx.get_Phi_nodal_flag,
                             level=level, include_ghosts=include_ghosts)
 
+def JxNodalWrapper(level=0, include_ghosts=False):
```
RemiLehe: Same comment as above regarding the nodal J: this interface could be removed, as it is not strictly needed, and might be confusing.

Python/pywarpx/WarpX.py
```diff
@@ -40,6 +41,7 @@ def create_argv_list(self, **kw):
         argv += warpx.attrlist()
         argv += my_constants.attrlist()
         argv += amr.attrlist()
+        argv += amrex.attrlist()
```
RemiLehe: It's not clear to me why this change is needed as part of this PR ; could you explain this a bit more?

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.H
```diff
@@ -0,0 +1,57 @@
...
+                                           const amrex::Vector<amrex::Array<std::unique_ptr<amrex::MultiFab>, 3>>& grad_buf_e_stag,
+                                           const amrex::Vector<amrex::Array<std::unique_ptr<amrex::MultiFab>, 3>>& grad_buf_b_stag)
+                : m_b_field(b_field),
+                  m_grad_buf_e_stag(grad_buf_e_stag),
+                  m_grad_buf_b_stag(grad_buf_b_stag)
+        {}
+
+        void operator()(amrex::Array<std::unique_ptr<amrex::MLMG>,3> & mlmg, int const lev);
+
+        // Function to perform interpolation from E-staggered fields to B-staggered fields
```
RemiLehe:
```suggestion
        // Function to perform interpolation from cell edges to cell faces
```

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -309,6 +309,12 @@ WarpX::Evolve (int numsteps)
...
+                // B field.  Time varying A contribution to E field is neglected.
+                // This is currently a lab frame calculation.
+                ComputeCurrentDensityField();
```
RemiLehe: Could we call this function `ComputeMagnetostaticField`?

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -0,0 +1,462 @@
...
+#include <string>
+
+using namespace amrex;
+
+void
+WarpX::ComputeCurrentDensityField()
+{
+    WARPX_PROFILE("WarpX::ComputeCurrentDensityField");
+    // Fields have been reset in Electrostatic solver for this time step, these fields
+    // are added into the E & B fields after electrostatic solve
```
RemiLehe:
```suggestion
    // are added into the B fields after electrostatic solve
```

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -0,0 +1,462 @@
...
+
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(this->max_level == 0, "Magnetostatic solver not implemented with mesh refinement.");
+
+    AddCurrentDensityFieldLabFrame();
```
RemiLehe:
```suggestion
    AddMagnetostaticFieldLabFrame();
```

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -0,0 +1,462 @@
...
+#endif
+
+#include <array>
+#include <memory>
+#include <string>
+
+using namespace amrex;
+
+void
+WarpX::ComputeCurrentDensityField()
```
RemiLehe: I would suggest:
```suggestion
WarpX::ComputeMagnetostaticField()
```

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -0,0 +1,462 @@
...
+{
+    WARPX_PROFILE("WarpX::ComputeCurrentDensityField");
```
RemiLehe:
```suggestion
    WARPX_PROFILE("WarpX::ComputeMagnetostaticField");
```

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -0,0 +1,462 @@
...
+    // Fields have been reset in Electrostatic solver for this time step, these fields
+    // are added into the E & B fields after electrostatic solve
+
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(this->max_level == 0, "Magnetostatic solver not implemented with mesh refinement.");
+
+    AddCurrentDensityFieldLabFrame();
+}
+
+void
+WarpX::AddCurrentDensityFieldLabFrame()
```
RemiLehe:
```suggestion
WarpX::AddMagnetostaticLabFrame()
```

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -0,0 +1,462 @@
...
+{
+    WARPX_PROFILE("WarpX::AddCurrentDensityFieldLabFrame");
```
RemiLehe:
```suggestion
    WARPX_PROFILE("WarpX::AddMagnetostaticFieldLabFrame");
```

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -0,0 +1,462 @@
...
+    // This will interpolate to staggered grid, filter, and synchronize
+    SyncCurrent(current_fp, current_cp); // Apply filter, perform MPI exchange, interpolate across levels
+
+#ifdef WARPX_DIM_RZ
+    for (int lev = 0; lev <= max_level; lev++) {
+        ApplyInverseVolumeScalingToCurrentDensity(current_fp[lev][0].get(),
+                                                  current_fp[lev][1].get(),
+                                                  current_fp[lev][2].get(), lev);
+    }
+#endif
```
RemiLehe: We could exchange the order of these 2 operations, in order to be more consistent with `AddSpaceChargeLabFrame`:
```suggestion
#ifdef WARPX_DIM_RZ
    for (int lev = 0; lev <= max_level; lev++) {
        ApplyInverseVolumeScalingToCurrentDensity(current_fp[lev][0].get(),
                                                  current_fp[lev][1].get(),
                                                  current_fp[lev][2].get(), lev);
    }
#endif
SyncCurrent(current_fp, current_cp); // Apply filter, perform MPI exchange, interpolate across levels
```

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -0,0 +1,462 @@
...
+
+    // set the boundary and current density potentials
+    setVectorPotentialBC(vector_potential_fp_nodal);
+
+    // Compute the vector potential A, by solving the Poisson equation
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE( !IsPythonCallBackInstalled("poissonsolver"),
+        "Python Level Poisson Solve not supported for Magnetostatic implementation.");
+
+    computeVectorPotential( current_fp, vector_potential_fp_nodal, self_fields_required_precision,
```
RemiLehe, line 120: There is potentially a bit of an issue here, since `self_fields_absolute_tolerance` is compared to both `phi` (in the electrostatic solver` and `A` (in the magnetostatic solver), but these two quantities do not have the same units.

One solution could be to use `self_fields_absolute_tolerance` for `phi`, and then use `self_fields_absolute_tolerance*c` for `A` (since `A` is of roughly of order `phi*c`, unless the particles are non-relativistic). Another solution could be to have a different parameter for `phi` and `A`, which the user sets independently.

Note that this problem does not occur for `self_fields_required_precision`, since this is a relative tolerance (not absolute).

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -0,0 +1,462 @@
...
+                     self_fields_verbosity );
+}
+
+/* Compute the vector potential `A` by solving the Poisson equation with `J` as
+   a source.
+   This uses the amrex solver.
+
+    More specifically, this solves the equation
+    \f[
+        \vec{\nabla}^2 r \vec{A} - (\vec{\beta}\cdot\vec{\nabla})^2 r \vec{A} = - r \mu_0 \vec{J}
```
RemiLehe: `beta` is not used here.
```suggestion
        \vec{\nabla}^2 r \vec{A} = - r \mu_0 \vec{J}
```

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -0,0 +1,462 @@
...
+#endif
+    };
+
+    // This will grab the gradient values for Ax
+    mlmg[0]->getGradSolution({buf_ptr});
+
+    // Interpolate dAx/dz to By grid buffer, then add to By
+    this->doInterp(m_grad_buf_e_stag[lev][2],
+                   m_grad_buf_b_stag[lev][1]);
+    MultiFab::Add(*(m_b_field[lev][1]), *(m_grad_buf_b_stag[lev][1]), 0, 0, 1, m_b_field[lev][1]->nGrowVect() );
```
RemiLehe: I would remove the guard cells here:
```suggestion
    MultiFab::Add(*(m_b_field[lev][1]), *(m_grad_buf_b_stag[lev][1]), 0, 0, 1, 0 );
```
This would be consistent with e.g. `ComputeE` and `ComputeB` from the relativistic electrostatic solver, where we only compute E and B in the valid cells. Then the guard cells are refreshed at the beginning of the next PIC iteration.

Source/FieldSolver/WarpXPushFieldsEM.cpp
```diff
@@ -1227,9 +1227,6 @@ WarpX::ApplyInverseVolumeScalingToCurrentDensity (MultiFab* Jx, MultiFab* Jy, Mu
     const Real dr = dx[0];
```
RemiLehe: Could you open a separate pull request for the changes contained in this file - since this can be reviewed independently from the rest of the PR?

Feel free to simultaneously keep these changes in the current PR, though, since my understanding is that this is necessary for the CI tests to pass.

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -0,0 +1,462 @@
...
+        {
+            warpx_interp(j, k, l, dst_arr, src_arr, dst_stag, src_stag, fg_nox, fg_noy, fg_noz,
+                stencil_coeffs_x, stencil_coeffs_y, stencil_coeffs_z);
+        });
+    }
+
+    // Synchronize the ghost cells, do halo exchange
+    ablastr::utils::communication::FillBoundary(*dst,
+                                                dst->nGrowVect(),
+                                                WarpX::do_single_precision_comms);
```
RemiLehe: I think that guard cell exchange can be skipped here.
```suggestion
    // Synchronize the ghost cells, do halo exchange
    ablastr::utils::communication::FillBoundary(*dst,
                                                dst->nGrowVect(),
                                                WarpX::do_single_precision_comms);
```

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -0,0 +1,462 @@
...
+    MultiFab::Add(*(m_b_field[lev][1]), *(m_grad_buf_b_stag[lev][1]), 0, 0, 1, m_b_field[lev][1]->nGrowVect() );
+
+    // Synchronize B fields
+    for (int idim = 0; idim < 3; idim++)
+    {
+        // Synchronize the ghost cells, do halo exchange
+        ablastr::utils::communication::FillBoundary(*(m_b_field[lev][idim]),
+                                                    m_b_field[lev][idim]->nGrowVect(),
+                                                    WarpX::do_single_precision_comms);
+    }
```
RemiLehe: This could be removed: the guard cells are refreshed at the beginning of the next PIC iteration anyway ; so I think that this will add unnecessary communications (which comes with a computational cost).
```suggestion
```

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -0,0 +1,462 @@
...
+#endif
+    for (MFIter mfi(*dst, TilingIfNotGPU()); mfi.isValid(); ++mfi)
+    {
+        IntVect const src_stag = src->ixType().toIntVect();
+        IntVect const dst_stag = dst->ixType().toIntVect();
+
+        Array4<amrex::Real const> const& src_arr = src->const_array(mfi);
+        Array4<amrex::Real> const& dst_arr = dst->array(mfi);
+
+        Box bx = mfi.growntilebox(dst->nGrowVect());
```
RemiLehe: I would calculate this only over the valid cells:
```suggestion
        Box bx = mfi.tilebox();
```

Source/WarpX.H
```diff
@@ -887,6 +890,19 @@ public:
     void computePhiTriDiagonal (const amrex::Vector<std::unique_ptr<amrex::MultiFab> >& rho,
                                       amrex::Vector<std::unique_ptr<amrex::MultiFab> >& phi) const;
 
+    // Magnetostatic Solver Interface
+    MagnetostaticSolver::VectorPoissonBoundaryHandler m_vector_poisson_boundary_handler;
+    void ComputeCurrentDensityField ();
```
RemiLehe:
```suggestion
    void ComputeMagnetostaticField ();
```

Source/WarpX.H
```diff
@@ -887,6 +890,19 @@ public:
...
+    void AddCurrentDensityFieldLabFrame ();
```
RemiLehe:
```suggestion
    void AddMagnetostaticFieldLabFrame ();
```

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -0,0 +1,462 @@
...
+        m_grad_buf_e_stag[lev][1].get(),
+        m_grad_buf_e_stag[lev][2].get()
+#elif defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
+        m_grad_buf_e_stag[lev][0].get(),
+        m_grad_buf_e_stag[lev][2].get()
+#endif
+    };
+
+    // This will grab the gradient values for Ax
+    mlmg[0]->getGradSolution({buf_ptr});
```
RemiLehe, line 409: This function call needs to fill the guard cells (at least one cell deep) in order for the following `doInterp` function to work properly ; do you know whether that is the case?

Source/Python/WarpXWrappers.cpp
```diff
@@ -300,6 +303,8 @@ namespace
     WARPX_GET_LOVECTS(warpx_getCurrentDensityLoVects, WarpX::GetInstance().get_pointer_current_fp)
     WARPX_GET_LOVECTS(warpx_getCurrentDensityCPLoVects, WarpX::GetInstance().get_pointer_current_cp)
     WARPX_GET_LOVECTS(warpx_getCurrentDensityFPLoVects, WarpX::GetInstance().get_pointer_current_fp)
+    WARPX_GET_LOVECTS(warpx_getCurrentDensityFPNodalLoVects, WarpX::GetInstance().get_pointer_current_fp_nodal)
```
RemiLehe: As discussed above, this interface could be removed.

Source/Python/WarpXWrappers.cpp
```diff
@@ -313,6 +318,12 @@ namespace
     int* warpx_getJx_nodal_flag() {return getFieldNodalFlagData( WarpX::GetInstance().get_pointer_current_fp(0,0) );}
     int* warpx_getJy_nodal_flag() {return getFieldNodalFlagData( WarpX::GetInstance().get_pointer_current_fp(0,1) );}
     int* warpx_getJz_nodal_flag() {return getFieldNodalFlagData( WarpX::GetInstance().get_pointer_current_fp(0,2) );}
+    int* warpx_getJxNodal_nodal_flag() {return getFieldNodalFlagData( WarpX::GetInstance().get_pointer_current_fp_nodal(0,0) );}
```
RemiLehe: As discussed above, this interface could be removed.

Source/Python/WarpXWrappers.H
```diff
@@ -171,6 +171,9 @@ extern "C" {
   amrex::Real** warpx_getCurrentDensityCP (int lev, int direction, int *return_size, int *ncomps, int **ngrowvect, int **shapes);
   amrex::Real** warpx_getCurrentDensityFP (int lev, int direction, int *return_size, int *ncomps, int **ngrowvect, int **shapes);
 
+  amrex::Real** warpx_getCurrentDensityFPNodal (int lev, int direction, int *return_size, int *ncomps, int **ngrowvect, int **shapes);
```
RemiLehe: As discussed above, this interface could be removed.

Source/Python/WarpXWrappers.H
```diff
@@ -182,6 +185,9 @@ extern "C" {
   int* warpx_getCurrentDensityLoVects (int lev, int direction, int *return_size, int **ngrowvect);
   int* warpx_getCurrentDensityCPLoVects (int lev, int direction, int *return_size, int **ngrowvect);
   int* warpx_getCurrentDensityFPLoVects (int lev, int direction, int *return_size, int **ngrowvect);
+  int* warpx_getCurrentDensityFPNodalLoVects (int lev, int direction, int *return_size, int **ngrowvect);
```
RemiLehe: As discussed above, this interface could be removed.

Source/Python/WarpXWrappers.H
```diff
@@ -195,9 +201,15 @@ extern "C" {
   int* warpx_getBx_nodal_flag ();
   int* warpx_getBy_nodal_flag ();
   int* warpx_getBz_nodal_flag ();
+  int* warpx_getAx_nodal_flag ();
+  int* warpx_getAy_nodal_flag ();
+  int* warpx_getAz_nodal_flag ();
   int* warpx_getJx_nodal_flag ();
   int* warpx_getJy_nodal_flag ();
   int* warpx_getJz_nodal_flag ();
+  int* warpx_getJxNodal_nodal_flag ();
```
RemiLehe: As discussed above, this interface could be removed.

RemiLehe (changes requested):
Thanks again for this great PR.
I made a more thorough review. Would you have time to address the inline comments?

Thanks!

Examples/Tests/magnetostatic_eb/analysis_rz.py
clarkse: This is the benchmark script for the Python RZ CI test, which doesn't check particles because of the random_theta issue I am  having and the non-reproducible random numbers from my platform to the Ubuntu container on Azure.

Python/pywarpx/WarpX.py
clarkse: I needed the amrex bucket to set flags for managing the arena in the Python input files when using CUDA in WSL.  These lines were commented out in the PICMI input files when preparing for CI tests in OMP.  This could be removed and pushed through as a different PR.

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
clarkse: I believe I was able to fix this by doing a halo exchange in the doInterp function.

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
clarkse: I went with the tolerance*c

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
RemiLehe: Ah, yes, sorry I overlooked this.

RemiLehe (approved):

PR #3552, opened 2022-12-08
Highlights & References: Gordon Bell

ax3l:
Add our award-winning Gordon Bell paper in documentation :)
Update landing pages.

I think our paper DOI might be `10.5555/3571885.3571889` according to the ACM page https://dl.acm.org/doi/proceedings/10.5555/3571885 - but it does not yet resolve to the paper. Will update this in a follow-up PR once it resolves.


lucafedeli88 (approved):
Thanks a lot for having added this! Very good idea!

PR #3554, opened 2022-12-08
Allow surface flux injection from the domain boundary

dpgrote:
This PR fixes an oversight of PR #3519. After that PR, the injection plane could not be on the domain boundary. If it is on the boundary, no particles would be injected. This happens because the check is made using `tile_realbox.contains` which checks if the point is within the tile box but returns false if the point is on the boundary. To fix this, a custom version of the routine was written, `ParticleUtils::containsInclusive`, which checks if the point is within the domain or on the boundary. Perhaps this routine could be moved to AMReX?

For a similar reason, the new routine `InjectorPosition::insideBoundsInclusive` was added which returns true if the particle is within the injection region or on the lower or upper boundary. The routine `InjectorPosition::insideBounds` returns false if the particle is on the upper boundary.

The CI test FluxInjection3D was updated, putting the injection planes on the domain boundary. Also, the direction of the proton flux was made negative, testing injection from the upper boundary. And further, the `max_grid_size` is set to 8 so that there are multiple blocks, ensuring the that checks are correct. The benchmark is updated since the particle positions changed. The momentum benchmarks did not change (except for a roundoff difference).


RemiLehe (approved):
Thanks for this PR! This looks good.

PR #3557, opened 2022-12-08
Renamed currents in Vay deposition from `j` to `D`

### LINKED ISSUE #3550: Rename currents used in Vay deposition (`j` to `D`)
All the variables related to currents in the function `doVayDepositionShapeN` in Source/Particles/Deposition/CurrentDeposition.H should be renamed following the convention used in the literature and in other parts of WarpX.

This was suggested in https://github.com/ECP-WarpX/WarpX/pull/3529#issuecomment-1320734540 as a follow-up on #3529, and it is summarized again here.

All of the `jx`, `jy`, `jz` name tags in the function `doVayDepositionShapeN` should be renamed as `Dx`, `Dy`, `Dz`, because that is the notation used in the literature (see, for example, ([Vay et al, JCP, 2013](https://doi.org/10.1016/j.jcp.2013.03.010), equations (20)-(24)), similarly to what was done when cumulative sums were implemented: https://github.com/ECP-WarpX/WarpX/blob/2775ac17fc78b3433d313da46a1c81f932e3912e/Source/FieldSolver/WarpXPushFieldsEM.cpp#L435-L437

aeriforme:
Renamed Vay currents fabs and arrays from `jx`, `jy`, `jz` to `Dx`, `Dy`, `Dz` when doing Vay current deposition,
according to the notation in the literature [[Vay et al, JCP, 2013](https://doi.org/10.1016/j.jcp.2013.03.010)]. 

Fix #3550.

### DIFF
--- a/Source/Particles/Deposition/CurrentDeposition.H
+++ b/Source/Particles/Deposition/CurrentDeposition.H
@@ -713,7 +713,7 @@ void doEsirkepovDepositionShapeN (const GetParticlePosition& GetPosition,
  * \brief Vay current deposition
  * (<a href="https://doi.org/10.1016/j.jcp.2013.03.010"> Vay et al, 2013</a>)
  * for thread \c thread_num: deposit \c D in real space and store the result in
- * \c jx_fab, \c jy_fab, \c jz_fab
+ * \c Dx_fab, \c Dy_fab, \c Dz_fab
  *
  * \tparam depos_order  deposition order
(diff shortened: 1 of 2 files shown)


EZoni (approved):
@aeriforme Thank you for cleaning this up and congrats on your first PR in WarpX!

ax3l:
Welcome again and thank you for your first PR to WarpX! :tada:

PR #3558, opened 2022-12-09
CI: Source Checks on ubuntu 22.04

ax3l:
Updated Doxygen from 1.8.17 to 1.9.1

Follow-up to #3546

Note: RTD still builds with a 1.8* Doxygen


dpgrote (approved):
This looks good. Presumably, updating the version of Ubuntu does not affect any of the other tests.

PR #3561, opened 2022-12-09
Allow const_dt with EM and check it for ES

dpgrote:
This small PR allows more control of the time step for EM simulations and ensures that it is set for ES.

### DIFF
--- a/Source/WarpX.H
+++ b/Source/WarpX.H
@@ -1339,7 +1339,8 @@ private:
     int num_injected_species = -1;
     amrex::Vector<int> injected_plasma_species;
 
-    amrex::Real const_dt = amrex::Real(0.5e-11);
+    amrex::Real const_dt;
+    bool const_dt_specified;
 
     // Macroscopic properties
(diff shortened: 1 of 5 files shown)


ax3l:
@dpgrote it looks like `embedded_boundary_cube` and `embedded_boundary_cube` crash with the current PR diff.

Source/WarpX.H
```diff
@@ -1339,7 +1339,8 @@ private:
     int num_injected_species = -1;
     amrex::Vector<int> injected_plasma_species;
 
-    amrex::Real const_dt = amrex::Real(0.5e-11);
+    amrex::Real const_dt;
+    bool const_dt_specified;
```
ax3l: Let's use a single `std::optional<amrex::Real>` as type and call the member variable `m_const_dt` according to our conventions?

ax3l: ping @dpgrote when you have the time :)

RemiLehe:
I just updated the PR according to @ax3l's comments.

RemiLehe (approved):

PR #3570, opened 2022-12-14
Include fusion reactions in documentation

RemiLehe:
This PR documents the fusion module.

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -1397,7 +1397,7 @@ WarpX provides several particle collision models, using varying degrees of appro
     in this documentation we use ``<collision_name>`` as a placeholder.
 
 * ``<collision_name>.type`` (`string`) optional
-    The type of collsion. The types implemented are:
+    The type of collision. The types implemented are:
 
     - ``pairwisecoulomb`` for pairwise Coulomb collisions, the default if unspecified.
       This provides a pair-wise relativistic elastic Monte Carlo binary Coulomb collision model,
@@ -1405,7 +1405,7 @@ WarpX provides several particle collision models, using varying degrees of appro
       When the RZ mode is used, `warpx.n_rz_azimuthal_modes` must be set to 1 at the moment,
       since the current implementation of the collision module assumes axisymmetry.
     - ``nuclearfusion`` for fusion reactions.
-      This implements the pair-wise fusion model by `Higginson et al. (JCP 388, 439-453, 2019) <https://www.sciencedirect.com/science/article/pii/S0021999119302037>`_.
+      This implements the pair-wise fusion model by `Higginson et al. (JCP 388, 439-453, 2019) <https://doi.org/10.1016/j.jcp.2019.03.020>`_.
       Currently, WarpX supports deuterium-deuterium, deuterium-tritium, deuterium-helium and proton-boron fusion.
       When initializing the reactant and product species, you need to use ``species_type`` (see the documentation
       for this parameter), so that WarpX can identify the type of reaction to use.
@@ -1455,7 +1455,7 @@ WarpX provides several particle collision models, using varying degrees of appro
     More specifically, in a fusion reaction between two macroparticles with weight ``w_1`` and ``w_2``,
     the weight of the product macroparticles will be ``min(w_1,w_2)/fusion_multipler``.
     (And the weights of the reactant macroparticles are reduced correspondingly after the reaction.)
-    See `Higginson et al. (JCP 388, 439-453, 2019) <https://www.sciencedirect.com/science/article/pii/S0021999119302037>`_
+    See `Higginson et al. (JCP 388, 439-453, 2019) <https://doi.org/10.1016/j.jcp.2019.03.020>`_
     for more details. The default value of ``fusion_multiplier`` is 1.
 
 * ``<collision_name>.fusion_probability_threshold``(`float`) optional.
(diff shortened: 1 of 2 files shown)


aeriforme:
Thank you for adding this! 
I have some comments that came up in my mind while reading: 
* There's a typo in line 1400: collsion.
* Can a species take part in ``pairwisecoulomb`` and ``nuclearfusion`` collisions 
at the same time (whether reactant or product) without any caveats? 
* From the ``fusion_probability_threshold`` bullet point:  
"_If the fusion multiplier is too high and results in a fusion probability
that approaches 1 (for a given collision between two macroparticles), then
there is a risk of underestimating the total fusion yield._": why underestimating? 
* Would it be better to cite [Higginson et al. (JCP 388, 439-453, 2019)] using the DOI https://doi.org/10.1016/j.jcp.2019.03.020?

RemiLehe:
Thanks a lot for reviewing this! 
I updated the documentation according to your comments.
Regarding the above questions:
- Yes, one species can indeed take part in several collision modules at the same time. The collision modules are called one after the other, at each PIC iteration (or each `ndt` PIC iterations).
- Regarding underestimating: this is not very easy to explain online, but I'll give it a try. This is because we can emit only 1 product macroparticle per timestep. Let us imagine that we have one macroparticle representing, say, 1000 physical particle, and that the physical parameters are such that the fusion reaction should occur within ~10 timesteps. Let us also imagine that the fusion_multiplier is 200, i.e. the weight of the product macroparticles is 5. In this case, because the algorithm can only emit 1 product macroparticle per timestep, it will take 1000/5 = 200 timesteps for the original macroparticle to be completely transformed into product macroparticles (instead of the physical answer i.e. ~10 timesteps).

Does the above answer your question? We could also discuss it offline.

aeriforme (approved):
I see. So the higher the `fusion_multiplier`, the longer is the time that takes for the reaction to complete. Therefore the risk is of underestimating the yield per unit time (where "yield" is the number of product particles per reaction event) and, as a consequence, possibly underestimating the total yield (where "total" I interpret as integrated in time). 

Does this make sense?

We could talk more in person.

RemiLehe:
Yes, that's correct. 👍

Docs/source/usage/parameters.rst
```diff
@@ -1386,24 +1386,30 @@ External fields
...
-    The type of collsion. The types implemented are:
+    The type of collision. The types implemented are:
 
     - ``pairwisecoulomb`` for pairwise Coulomb collisions, the default if unspecified.
       This provides a pair-wise relativistic elastic Monte Carlo binary Coulomb collision model,
       following the algorithm given by `Perez et al. (Phys. Plasmas 19, 083104, 2012) <https://doi.org/10.1063/1.4742167>`_.
       When the RZ mode is used, `warpx.n_rz_azimuthal_modes` must be set to 1 at the moment,
       since the current implementation of the collision module assumes axisymmetry.
+    - ``nuclearfusion`` for fusion reactions.
+      This implements the pair-wise fusion model by `Higginson et al. (JCP 388, 439-453, 2019) <https://doi.org/10.1016/j.jcp.2019.03.020>`_.
```
ax3l: Avoid double label ref warnings:
```suggestion
      This implements the pair-wise fusion model by `Higginson et al. (JCP 388, 439-453, 2019) <https://doi.org/10.1016/j.jcp.2019.03.020>`__.
```

Docs/source/usage/parameters.rst
```diff
@@ -1434,8 +1446,32 @@ WarpX provides several particle collions models, using varying degrees of approx
...
+* ``<collision_name>.fusion_multiplier`` (`float`) optional.
+    Only for ``nuclearfusion``.
+    Increasing ``fusion_multiplier`` creates more macroparticles of fusion
+    products, but with lower weight (in such a way that the corresponding
+    total number of physical particle remains the same). This can improve
+    the statistics of the simulation, in the case where fusion reactions are very rare.
+    More specifically, in a fusion reaction between two macroparticles with weight ``w_1`` and ``w_2``,
+    the weight of the product macroparticles will be ``min(w_1,w_2)/fusion_multipler``.
+    (And the weights of the reactant macroparticles are reduced correspondingly after the reaction.)
+    See `Higginson et al. (JCP 388, 439-453, 2019) <https://doi.org/10.1016/j.jcp.2019.03.020>`_
```
```suggestion
    See `Higginson et al. (JCP 388, 439-453, 2019) <https://doi.org/10.1016/j.jcp.2019.03.020>`__
```

Docs/source/usage/parameters.rst
```diff
@@ -1434,8 +1446,32 @@ WarpX provides several particle collions models, using varying degrees of approx
     a Coulomb logarithm will be computed automatically according to the algorithm in
     `Perez et al. (Phys. Plasmas 19, 083104, 2012) <https://doi.org/10.1063/1.4742167>`_.
```
```suggestion
    `Perez et al. (Phys. Plasmas 19, 083104, 2012) <https://doi.org/10.1063/1.4742167>`__.
```

Docs/source/usage/parameters.rst
```diff
@@ -1386,24 +1386,30 @@ External fields
...
     This is then used in the rest of the input deck;
     in this documentation we use ``<collision_name>`` as a placeholder.
 
 * ``<collision_name>.type`` (`string`) optional
-    The type of collsion. The types implemented are:
+    The type of collision. The types implemented are:
 
     - ``pairwisecoulomb`` for pairwise Coulomb collisions, the default if unspecified.
       This provides a pair-wise relativistic elastic Monte Carlo binary Coulomb collision model,
       following the algorithm given by `Perez et al. (Phys. Plasmas 19, 083104, 2012) <https://doi.org/10.1063/1.4742167>`_.
```
```suggestion
      following the algorithm given by `Perez et al. (Phys. Plasmas 19, 083104, 2012) <https://doi.org/10.1063/1.4742167>`__.
```

ax3l (approved):

PR #3573, opened 2022-12-15
Refactor update of auxiliary data for electrostatic solver

RemiLehe:
In the current version of WarpX, the update of the auxiliary data (i.e. interpolating across levels ; interpolating from a staggering grid to a nodal grid when using the momentum-conserving algorithm) is done in a different parts of the code, for the electrostatic and electromagnetic solver.

This PR simplifies this by always doing the update of the auxiliary data in the same part of the code, for both the electrostatic and electromagnetic solver.


ax3l (approved):
Thx!

PR #3576, opened 2022-12-17
Avoid managed memory : user-defined particle attribs in AddPlasma

RevathiJambunathan:
Thanks to @WeiqunZhang for the offline discussion. 
This PR is to fix the managed memory error reported in Issue #2009 

I tested this with particle attributes [this input](https://github.com/ECP-WarpX/WarpX/blob/development/Examples/Physics_applications/laser_acceleration/inputs_3d) and set `amrex.the_arena_is_managed=0` and it works on the GPU.
Note that I had to turn off the FieldProbe reduced diagnostic, since it seems that there is some managed memory used there.

### DIFF
--- a/Source/Particles/PhysicalParticleContainer.cpp
+++ b/Source/Particles/PhysicalParticleContainer.cpp
@@ -1059,23 +1059,43 @@ PhysicalParticleContainer::AddPlasma (int lev, RealBox part_realbox)
         // user-defined integer and real attributes
         const int n_user_int_attribs = m_user_int_attribs.size();
         const int n_user_real_attribs = m_user_real_attribs.size();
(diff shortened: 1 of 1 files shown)


WeiqunZhang (approved):

WeiqunZhang:
LGTM.  As for the issue in FieldProbe, I can see there is an issue there and it can be fixed.  @RevathiJambunathan since you have the test set up already, would you be able to test it?

RevathiJambunathan:
@WeiqunZhang Thank you for taking a look at the FieldProbe issue. Yes. I will be happy to test it!

WeiqunZhang:
#3579 @RevathiJambunathan Note that the draft PR has your commit included.

PR #3577, opened 2022-12-19
Fix: NullifyMF Z-Box Check

ax3l:
The box check did not take the right Z component in non-3D geometries.

This results in our mirrors not being applied in some cases, depending on the concrete simulation dimensions used for the other components.


RemiLehe:
Thanks for this PR!
However, I think `LowerCorner` always returns `z` as the number 2 vector component:
https://github.com/ECP-WarpX/WarpX/blob/development/Source/WarpX.cpp#L2377

ax3l:
Thx! ok, but is `amrex::Real dz = WarpX::CellSize(lev)[WARPX_ZINDEX];` not wrong to use `[2]` in non-3D?

update: dang, no it's not:
https://github.com/ECP-WarpX/WarpX/blob/c3eb6ea1efeff3910e8520d94e260d94367d68f7/Source/WarpX.cpp#L2354-L2366

RemiLehe:
It's the same thing with `CellSize`:
https://github.com/ECP-WarpX/WarpX/blob/development/Source/WarpX.cpp#L2355

This is certainly very confusing, and we could consider whether to change the behavior of `LowerCorner` and `CellSize`. I don't remember if we had a good reason to define them this way.

PR #3579, opened 2022-12-20
Fieldprobe managed memory

WeiqunZhang:
The implementation here can be optimized by using shared memory.  However,
instead of doing it here, I am going to add a helper function to AMReX.
    
Also removed an unnecessary synchronization, because there is a
synchronization in the destructor of the particle iterator.


RevathiJambunathan:
@WeiqunZhang Thank you for this PR!
I tested it on summit with `amrex.the_arena_is_managed=0`and it works now.

RevathiJambunathan (approved):

WeiqunZhang:
The optimization mentioned is in https://github.com/AMReX-Codes/amrex/pull/3076.

ax3l (approved):
Thanks a lot 😌

PR #3578, opened 2022-12-19
AMReX/PICSAR: Weekly Update

ax3l:
Weekly update to latest AMReX.
Weekly update to latest PICSAR (no changes).

```console
./Tools/Release/updateAMReX.py
./Tools/Release/updatePICSAR.py
```

## Fix Intel CI: dpcpp -> icpx

Intel has deprecated `dpcpp`.  We are supposed to use `icpx -fsycl` now for SYCL code.

PR #3581, opened 2022-12-20
Fix Intel CI: dpcpp -> icpx

WeiqunZhang:
Intel has deprecated `dpcpp`.  We are supposed to use `icpx -fsycl` now for SYCL code.

### DIFF
--- a/.github/workflows/intel.yml
+++ b/.github/workflows/intel.yml
@@ -157,8 +157,8 @@ jobs:
         set +e
         source /opt/intel/oneapi/setvars.sh
         set -e
-        export CXX=$(which dpcpp)
-        export CC=$(which clang)
+        export CXX=$(which icpx)
+        export CC=$(which icx)
 
         cmake -S . -B build_sp         \
           -DCMAKE_CXX_FLAGS_RELEASE="-O1 -DNDEBUG" \
--- a/.github/workflows/intel.yml
+++ b/.github/workflows/intel.yml
@@ -159,6 +159,7 @@ jobs:
         set -e
         export CXX=$(which icpx)
         export CC=$(which icx)
+        export CXXFLAGS="-fsycl ${CXXFLAGS}"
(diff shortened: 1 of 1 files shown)


.github/workflows/intel.yml
```diff
@@ -157,8 +157,8 @@ jobs:
         set +e
         source /opt/intel/oneapi/setvars.sh
         set -e
-        export CXX=$(which dpcpp)
-        export CC=$(which clang)
+        export CXX=$(which icpx)
+        export CC=$(which icx)
```
ax3l, line 161: But should actually be added in CMake, for now here:
```suggestion
        export CC=$(which icx)
        export CXXFLAGS="-fsycl ${CXXFLAGS}"
```

ax3l (approved):

.github/workflows/intel.yml
WeiqunZhang: Since we use `-DWarpX_COMPUTE=SYCL`, `-fsycl` will be added in the cmake setup in amrex.  But I am not sure if this is explicitly needed when compiling OpenPMD.

ax3l:
@WeiqunZhang 
```
Src/Base/AMReX_GpuLaunchFunctsG.H:33:105: error: 'local' is deprecated: use `local_accessor` instead [-Werror,-Wdeprecated-declarations]
            sycl::accessor<unsigned long long, 1, sycl::access::mode::read_write, sycl::access::target::local>
                                                                                                        ^
```

Can you fix this upstream? :)
(Already on my phone, otherwise I can push it tomorrow)

WeiqunZhang:
It's fixed in amrex already.  That's why I suggest that you merge this PR with the weekly update PR.

ax3l:
Oh, got it. Lemme rebase

ax3l:
Merged into  #3578

PR #3582, opened 2022-12-20
Fix two "Inconsistent nullness checks" of costs pointer (for load balancing)  found with CodeQL 

lucafedeli88:
@ax3l , nerd sniping works


ax3l (approved):
:rocket: :sparkling_heart:

PR #3584, opened 2022-12-20
Import external E/B fields from openPMD files

### LINKED ISSUE #3048: Define external fields from an openPMD file
Discussed with @ax3l @Yin-YinjianZhao @dpgrote : in order to easily supported domain decomposition, we will store these external fields as `MultiFabs` that have the same resolution and distribution mapping as the regular fields of the simulation.

The implementation need the following steps:

- [ ] Conditionally, allocate `MultiFabs` for the external `E` and `B` fields (same resolution and distribution mapping)
  - only if the user specified to load data from file, otherwise ask for extra GPU memory for these fields
- [ ] Read the openPMD file:
     - Find out which chunks to read in the original file (we do not want to load the whole file)
     - Read this data into a temporary array (this array is a the resolution of the original file)
       - https://openpmd-api.readthedocs.io/en/0.14.4/usage/firstread.html
       - https://openpmd-api.readthedocs.io/en/0.14.4/usage/parallel.html#c
     - Call kernel to interpolate from that temporary array to the above-mentioned `MultiFab`
     
@ax3l Do you want to draft a signature for a function that does some of that.
     
- [ ] Modify the function `GetExternalFields`, to pass the above-mentioned `MultiFab` (see `Particles/Gather`). We should be able to reuse the usual function for `Gather` from the regular grid (since this uses the same resolution)

### LINKED ISSUE #3224: Fixed arbitrary external magnetic field from file
Greetings,

I'm wandering if I can specify a fixed external Bx, By, Bz values defined from a text file.

I looked into the documentation and found the warpx.Bx_external_grid_function(x,y,z) parameter, but it's no quite what a need, I'm trying a simulate a LINAC lattice.

Do I need a python installation of WaprX and specify a PICMI input script.

Thank you in advance.

PhilMiller:
Sets the initial fields from openPMD files for the EM solver.
Sets them every time step for the ES solver.

Close #3048
Close #3224


PhilMiller:
Big things to address:
* [x] Adding external field contribution to particle-influence field solutions
* [x] Use `Array4` syntax in all paths
* [x] Use existing linear interpolation functions https://github.com/ECP-WarpX/WarpX/blob/development/Source/Utils/Algorithms/LinearInterpolation.H
* [x] Behaving sensibly in simulations with mesh refinement
* [x] Not crashing with load balancing (add appropriately to `WarpX::RemakeLevel` in `Parallelization/WarpXRegrid.cpp`)
* [x] Handling input reading to non-managed host memory and copying to device memory appropriately - allocating and copying from host at load time
* [x] Ensure guard cells are filled appropriately
* [ ] Absence of performance problems
* [x] Check indexing for XZ case
* [x] Fix indexing for 3d case
* [x] Assert that data order and axis labels match expectations

PhilMiller:
Big squash and rebase to address merge conflicts with current code base

Source/WarpX.cpp
```diff
@@ -646,12 +652,36 @@ WarpX::ReadParameters ()
...
+                                             && (electrostatic_solver_id == ElectrostaticSolverAlgo::LabFrame
+                                                 || electrostatic_solver_id == ElectrostaticSolverAlgo::None),
+                                             "External B field is mutually exclusive with B field solution");
+        }
+        if (add_external_E_field) {
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(electromagnetic_solver_id == ElectromagneticSolverAlgo::None
+                                             && electrostatic_solver_id == ElectrostaticSolverAlgo::None,
+                                             "External E field is mutually exclusive with E field solution");
+        }
+
```
roelof-groenewald: Is this limitation necessary for a reason I'm not seeing? The external field would just set the initial values of the E and B fields, right? It should then be fine to let those initial fields evolve consistently with the particle dynamics, right?

PhilMiller:
Huh, the test failures this was hitting in CI and my local machine at startup, of not being able to read the file, don't reproduce on Perlmutter. It runs fine there.

PhilMiller:
Ok, so on Perlmutter, the analysis script hits the same failure as CI:

```
  File "/global/homes/p/pmiller/avalanche/repos/WarpX/Examples/Tests/LoadExternalField/analysis_3d.py", line 39, in <module>
    x  = ad['proton','particle_position_x'].to_ndarray()
  File "/tmp/ci-E3NhykLmqc/py-venv/lib/python3.9/site-packages/yt/data_objects/data_containers.py", line 258, in __getitem__
    f = self._determine_fields([key])[0]
  File "/tmp/ci-E3NhykLmqc/py-venv/lib/python3.9/site-packages/yt/data_objects/data_containers.py", line 1473, in _determine_fields
    finfo = self.ds._get_field_info(ftype, fname)
  File "/tmp/ci-E3NhykLmqc/py-venv/lib/python3.9/site-packages/yt/data_objects/static_output.py", line 858, in _get_field_info
    field_info, candidates = self._get_field_info_helper(ftype, fname)
  File "/tmp/ci-E3NhykLmqc/py-venv/lib/python3.9/site-packages/yt/data_objects/static_output.py", line 978, in _get_field_info_helper
    raise YTFieldNotFound(field=INPUT, ds=self)
yt.utilities.exceptions.YTFieldNotFound: Could not find field ('proton', 'particle_position_x') in LoadExternalField3D_plt000300.
```

PhilMiller:
When I shorten the diagnostics interval to every 30 steps, the analysis script finds the particle at step 60, obviously in the wrong place (failing the later assertion), and then it's gone by step 90.

This maybe exposes a deficiency of WarpX's diagnostics, that it doesn't emit any sort of placeholder for particle species the input specified, but of which there are none in the domain at a particular time point.

PhilMiller:
plot
Welp, particle is doing that. I think the field isn't influencing it very well...

Source/WarpX.cpp
```diff
@@ -2168,6 +2186,33 @@ WarpX::AllocLevelMFs (int lev, const BoxArray& ba, const DistributionMapping& dm
...
+        if (add_external_E_field) {
+            AllocInitMultiFab(Efield_fp_external[lev][0], nba, dm, ncomps, ngEB, tag("Efield_fp_external[x]"), 0.0_rt);
+            AllocInitMultiFab(Efield_fp_external[lev][1], nba, dm, ncomps, ngEB, tag("Efield_fp_external[y]"), 0.0_rt);
+            AllocInitMultiFab(Efield_fp_external[lev][2], nba, dm, ncomps, ngEB, tag("Efield_fp_external[z]"), 0.0_rt);
+        }
+    } else {
+        if (add_external_B_field) {
+            AllocInitMultiFab(Bfield_fp_external[lev][0], amrex::convert(ba, Bx_nodal_flag), dm, ncomps, ngEB, tag("Bfield_fp_external[x]"));
+            AllocInitMultiFab(Bfield_fp_external[lev][1], amrex::convert(ba, By_nodal_flag), dm, ncomps, ngEB, tag("Bfield_fp_external[y]"));
+            AllocInitMultiFab(Bfield_fp_external[lev][2], amrex::convert(ba, Bz_nodal_flag), dm, ncomps, ngEB, tag("Bfield_fp_external[z]"));
```
PhilMiller: The flags probably need adjustment for the RZ case?

Source/Initialization/WarpXInitData.cpp
```diff
@@ -859,6 +885,34 @@ WarpX::InitLevelData (int lev, Real /*time*/)
...
     }
 
+    // Reading external fields from data file
+    if (add_external_B_field && lev == 0) {
+        std::string read_fields_from_path="./";
+        pp_warpx.query("read_fields_from_path", read_fields_from_path);
+#if defined(WARPX_DIM_RZ)
+        ReadExternalFieldsFromFile(read_fields_from_path, Bfield_fp_external[lev][0].get(), "B", "r");
+        ReadExternalFieldsFromFile(read_fields_from_path, Bfield_fp_external[lev][1].get(), "B", "t");
+        ReadExternalFieldsFromFile(read_fields_from_path, Bfield_fp_external[lev][2].get(), "B", "z");
```
PhilMiller: Feels like it ought to be this:

```diff
> git diff
diff --git a/Source/Initialization/WarpXInitData.cpp b/Source/Initialization/WarpXInitData.cpp
index af8f7c0e..36a4ec7e 100644
--- a/Source/Initialization/WarpXInitData.cpp
+++ b/Source/Initialization/WarpXInitData.cpp
@@ -891,8 +891,8 @@ WarpX::InitLevelData (int lev, Real /*time*/)
         pp_warpx.query("read_fields_from_path", read_fields_from_path);
 #if defined(WARPX_DIM_RZ)
         ReadExternalFieldsFromFile(read_fields_from_path, Bfield_fp_external[lev][0].get(), "B", "r");
-        ReadExternalFieldsFromFile(read_fields_from_path, Bfield_fp_external[lev][1].get(), "B", "t");
-        ReadExternalFieldsFromFile(read_fields_from_path, Bfield_fp_external[lev][2].get(), "B", "z");
+        ReadExternalFieldsFromFile(read_fields_from_path, Bfield_fp_external[lev][1].get(), "B", "z");
+        //ReadExternalFieldsFromFile(read_fields_from_path, Bfield_fp_external[lev][???].get(), "B", "t");
 #else
         ReadExternalFieldsFromFile(read_fields_from_path, Bfield_fp_external[lev][0].get(), "B", "x");
         ReadExternalFieldsFromFile(read_fields_from_path, Bfield_fp_external[lev][1].get(), "B", "y");
@@ -904,8 +904,8 @@ WarpX::InitLevelData (int lev, Real /*time*/)
         pp_warpx.query("read_fields_from_path", read_fields_from_path);
 #if defined(WARPX_DIM_RZ)
         ReadExternalFieldsFromFile(read_fields_from_path, Efield_fp_external[lev][0].get(), "E", "r");
-        ReadExternalFieldsFromFile(read_fields_from_path, Efield_fp_external[lev][1].get(), "E", "t");
-        ReadExternalFieldsFromFile(read_fields_from_path, Efield_fp_external[lev][2].get(), "E", "z");
+        ReadExternalFieldsFromFile(read_fields_from_path, Efield_fp_external[lev][1].get(), "E", "z");
+        //ReadExternalFieldsFromFile(read_fields_from_path, Efield_fp_external[lev][???].get(), "E", "t");
 #else
         ReadExternalFieldsFromFile(read_fields_from_path, Efield_fp_external[lev][0].get(), "E", "x");
         ReadExternalFieldsFromFile(read_fields_from_path, Efield_fp_external[lev][1].get(), "E", "y");
```

But that doesn't seem to produce sensible resulting fields

PhilMiller: Basically, whatever we do here needs to end up consistent with `AddExternalFields` just above

PhilMiller:
I'm seeing two separate problems as I try to debug this
*  ~~Diagnostics are showing that the effective field `Bfield_aux` doesn't contain any contribution from the external field, even though the addition is called right before the diagnostic output~~
* ~~When I hack diagnostics to output the external field `Bfield_fp_external` instead of `Bfield_aux`, it's mangled by some sort of indexing swap~~

PhilMiller:
With the last commit, the hacked diagnostics show that the external fields are actually being read in the right order:
image
image

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1215,3 +1269,152 @@ void WarpX::CheckKnownIssues()
...
+        // Now, the full range of data is loaded.
+        // Loading chunk data can speed up the process.
+        // Thus, `chunk_offset` and `chunk_extent` should be modified accordingly in another PR.
+        openPMD::Offset chunk_offset = {0,0,0};
+        openPMD::Extent chunk_extent = {extent[0],extent[1],extent[2]};
+
+        auto FC_chunk_data = FC.loadChunk<double>(chunk_offset,chunk_extent);
+        series.flush();
+        auto FC_data_host = FC_chunk_data.get();
+
```
PhilMiller, line 1351: Add assertions about axis labels and data order matching expectation below, or generalize to respect them

PhilMiller:
Need to regenerate the checksum files, but the particle position assertions as Yin-Yinjian Zhao wrote them now pass!

PhilMiller:
I had to correct for #3683

PhilMiller:
With the adjustment to the input files to replace `do_nodal = 1` with `grid_type = collocated`, the benchmark resets only meaningfully change the recorded magnetic field values. The particle positions are numerically perturbed, but basically identical to what Yin-Yinjian Zhao had in the earlier PR.

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1228,3 +1282,152 @@ void WarpX::CheckKnownIssues()
                 ablastr::warn_manager::WarnPriority::low);
         }
 }
+
+#if defined(WARPX_USE_OPENPMD) && !defined(WARPX_DIM_1D_Z)
+void
+WarpX::ReadExternalFieldsFromFile (std::string read_fields_from_path, MultiFab* mf,
```
PhilMiller: This should really be `ReadExternalField[FromFile]` or `ReadExternalFieldComponent[FromFile]`, since it only operates on one field and component at a time.

PhilMiller:
Regarding mesh refinement - I think more work would be needed, since the external field is only added onto the coarsest level, while particles may gather their effective field from a finer patch. I'll add an assertion about `max_level`

EZoni:
@PhilMiller When you use the functions `utils::algorithms::bilinear_interp` or `utils::algorithms::trilinear_interp`, there seems to be an inconsistency between `double` and `amrex::Real` due to the fact that all arguments have to be of the same type specified by the template, while that isn't the case in the current implementation. At least, that's my understanding based on CI's build logs.

Source/Initialization/WarpXInitData.cpp
```diff
@@ -431,6 +436,11 @@ WarpX::InitData ()
         if (electrostatic_solver_id == ElectrostaticSolverAlgo::LabFrameElectroMagnetostatic)
             ComputeMagnetostaticField();
 
+        // Set up an invariant that any code besides the field solver
+        // that looks at field values will see the composite of the
+        // field solution and any external field
```
EZoni: I'm not sure I fully understand the meaning of the word "invariant" in this docstring. Do you think it could be explained differently?

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1228,3 +1289,152 @@ void WarpX::CheckKnownIssues()
                 ablastr::warn_manager::WarnPriority::low);
         }
 }
+
+#if defined(WARPX_USE_OPENPMD) && !defined(WARPX_DIM_1D_Z)
+void
+WarpX::ReadExternalFieldFromFile (
+       std::string read_fields_from_path, MultiFab* mf,
```
EZoni:
```suggestion
       std::string read_fields_from_path, amrex::MultiFab* mf,
```

Source/Initialization/WarpXInitData.cpp
```diff
@@ -447,6 +457,23 @@ WarpX::InitData ()
     CheckKnownIssues();
 }
 
+void
+WarpX::AddExternalFields () {
+    for (int lev = 0; lev <= finest_level; ++lev) {
+        // FIXME: RZ multimode has more than one component for all these
```
EZoni, line 464: Do you plan to implement RZ support in this PR or in a separate PR? If it'll be a separate PR, it would be good to abort in case the user requests this feature with an RZ simulation.

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1228,3 +1289,152 @@ void WarpX::CheckKnownIssues()
...
+    // Read external field openPMD data
+    auto series = openPMD::Series(read_fields_from_path, openPMD::Access::READ_ONLY);
+    auto iseries = series.iterations.begin()->second;
+    auto F = iseries.meshes[F_name];
+    auto offset = F.gridGlobalOffset();
+    auto offset0 = offset[0];
+    auto offset1 = offset[1];
+#if defined(WARPX_DIM_3D)
+    auto offset2 = offset[2];
+#endif
```
EZoni: Would it be possible to use the explicit types instead of `auto`, at least for the easy cases? So that code reades can understand immediately if they are dealing with, say, an integer or floating point offset, for example.

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1228,3 +1289,152 @@ void WarpX::CheckKnownIssues()
...
+    auto d = F.gridSpacing<long double>();
+    Real d0 = d[0];
+    Real d1 = d[1];
+#if defined(WARPX_DIM_3D)
+    Real d2 = d[2];
+#endif
```
EZoni:
```suggestion
    amrex::Real d0 = d[0];
    amrex::Real d1 = d[1];
#if defined(WARPX_DIM_3D)
    amrex::Real d2 = d[2];
#endif
```

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1228,3 +1289,152 @@ void WarpX::CheckKnownIssues()
...
+    amrex::Gpu::DeviceVector<double> FC_data_gpu(total_extent);
+    auto FC_data = FC_data_gpu.data();
+    amrex::Gpu::copy(amrex::Gpu::hostToDevice, FC_data_host, FC_data_host + total_extent, FC_data);
+
+    // Loop over boxes
+    for (MFIter mfi(*mf, TilingIfNotGPU()); mfi.isValid(); ++mfi)
+    {
+        auto box = mfi.growntilebox();
+        const amrex::Box& tb = mfi.tilebox(nodal_flag, mf->nGrowVect());
+        auto const& mffab = mf->array(mfi);
```
EZoni: Same comment on the `auto` types, I guess we could easily replace them with the explicit types, like `amrex::Box` and `amrex::Array4<amrex::Real>` or similar. Whenever possible, I personally find it less criptic and easier for developers to use for debugging, extensions, etc.

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1228,3 +1289,152 @@ void WarpX::CheckKnownIssues()
...
+
+        // Start ParallelFor
+        amrex::ParallelFor (tb,
+            [=] AMREX_GPU_DEVICE (int i, int j, int k) {
+                // i,j,k denote x,y,z indicies in 3D xyz.
+                // i,j,k denote r,z,mode indicies in 2D rz.
```
EZoni:
```suggestion
                // i,j,k denote x,y,z indices in 3D xyz.
                // i,j,k denote r,z,mode indices in 2D rz.
```

EZoni (commented):
Thank you for this PR, @PhilMiller! I left a few minor comments, in addition to the separate one I made earlier on the CI build failures.

EZoni:
Another question on this comment:

> With the adjustment to the input files to replace do_nodal = 1 with grid_type = collocated, the benchmark resets only meaningfully change the recorded magnetic field values. The particle positions are numerically perturbed, but basically identical to what Yin-Yinjian Zhao had in the earlier PR.

Maybe I didn't understand the comment properly, but does this mean that using the new grid type interface caused changes in the benchmarks somehow? I think it should not be the case, since it was simply a change in the user interface.

PhilMiller:
> Another question on this comment:
> 
> > With the adjustment to the input files to replace do_nodal = 1 with grid_type = collocated, the benchmark resets only meaningfully change the recorded magnetic field values. The particle positions are numerically perturbed, but basically identical to what Yin-Yinjian Zhao had in the earlier PR.
> 
> Maybe I didn't understand the comment properly, but does this mean that using the new grid type interface caused changes in the benchmarks somehow? I think it should not be the case, since it was simply a change in the user interface.

Yes, sort of, but in the good direction. When I initially updated the branch to incorporate that change, `do_nodal = 1` started being ignored, and so the benchmarks were not matching. After I converted the test input files to read `grid_type = collocated`, the benchmark values got back into alignment with what they were before.

Source/Initialization/WarpXInitData.cpp
PhilMiller: RZ works just fine. It's RZ with more than one mode that doesn't. I can add an assertion against that.

Source/Initialization/WarpXInitData.cpp
PhilMiller: I'm using it in the sense described here: https://softwareengineering.stackexchange.com/questions/32727/what-are-invariants-how-can-they-be-used-and-have-you-ever-used-it-in-your-pro

PhilMiller:
I think guard cells are fine, since we fill the entirety of each `Box` present in the field solution.

PhilMiller:
The basics of performance should be fine even on GPUs, since the ultimate data is only ever touched on GPU. Some benchmarking may be necessary to confirm though.

Source/WarpX.cpp
```diff
@@ -2030,6 +2048,18 @@ WarpX::AllocLevelMFs (int lev, const BoxArray& ba, const DistributionMapping& dm
...
+        AllocInitMultiFab(Bfield_fp_external[lev][0], amrex::convert(ba, Bx_nodal_flag), dm, ncomps, ngEB, tag("Bfield_fp_external[x]"));
+        AllocInitMultiFab(Bfield_fp_external[lev][1], amrex::convert(ba, By_nodal_flag), dm, ncomps, ngEB, tag("Bfield_fp_external[y]"));
+        AllocInitMultiFab(Bfield_fp_external[lev][2], amrex::convert(ba, Bz_nodal_flag), dm, ncomps, ngEB, tag("Bfield_fp_external[z]"));
+    }
+    if (add_external_E_field) {
+        AllocInitMultiFab(Efield_fp_external[lev][0], amrex::convert(ba, Ex_nodal_flag), dm, ncomps, ngEB, tag("Efield_fp_external[x]"));
+        AllocInitMultiFab(Efield_fp_external[lev][1], amrex::convert(ba, Ey_nodal_flag), dm, ncomps, ngEB, tag("Efield_fp_external[y]"));
+        AllocInitMultiFab(Efield_fp_external[lev][2], amrex::convert(ba, Ez_nodal_flag), dm, ncomps, ngEB, tag("Efield_fp_external[z]"));
+    }
+
```
EZoni, line 2066: I wonder if this causes a segfault in case one tries to read external fields with more than one MR level.

I saw you assert that in `WarpX::InitLevelData` in Source/Initialization/WarpXInitData.cpp.

However, the order of function calls seems to me to be the following:
- `AmrCore::InitFromScratch`
  - `WarpX::MakeNewLevelFromScratch(lev, ...)`
    - `AllocLevelData(lev, ...)`
      - `AllocLevelMFs(lev, ...)`
      (which attempts to allocate the external field MultiFabs for a given level)
    - `InitLevelData(lev, ...)`
    (which asserts that external fields cannot be read with more than one level)

It seems to me that, with more than one MR level, one could encounter a segfault before the assert message is triggered. Is it possible that this is the case or am I missing something? May be worth double checking.

PhilMiller: You're almost certainly right. I'll check and adjust those guards

PhilMiller: Ok, looking more closely at this, I'm not seeing definite segfault potential, other than that these calls add the allocation footprint of one more MultiFab over the same boxes as the rest of the simulation already allocates. 

That said, I'll move the assertions here or to parameter reading, so we don't even risk that.

run_test.sh
```diff
@@ -74,6 +74,8 @@ git clone https://github.com/AMReX-Codes/amrex.git
 cd amrex && git checkout --detach 1ffc6e427e3aea9302602190c8fc1cfbea0c9725 && cd -
 # warpx-data contains various required data sets
 git clone --depth 1 https://github.com/ECP-WarpX/warpx-data.git
+# openPMD-example-datasets contains various required data sets
+git clone --depth 1 --branch draft https://github.com/openPMD/openPMD-example-datasets
```
RemiLehe: Could you instead use `wget` or `curl` to download only the two files that you need from this repo?
Otherwise, we might spend a lot of time downloading large datasets (from `openPMD-example-datasets`) that are not actually used in the test.

RemiLehe (changes requested):

run_test.sh
PhilMiller: An earlier version of the PR as written by Yin-Yinjian Zhao did that, and I could never get it to actually run successfully in the automated CI environment. If there's a working example, I'd be happy to try following it

Source/Initialization/WarpXInitData.cpp
```diff
@@ -431,6 +436,11 @@ WarpX::InitData ()
         if (electrostatic_solver_id == ElectrostaticSolverAlgo::LabFrameElectroMagnetostatic)
             ComputeMagnetostaticField();
 
+        // Set up an invariant that any code besides the field solver
```
EZoni: I still have a bit of trouble understanding this comment.
Was it maybe meant to be like this ("so" or "such" missing)?
```suggestion
        // Set up an invariant so that any code besides the field solver
```

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1228,3 +1287,174 @@ void WarpX::CheckKnownIssues()
...
+                amrex::Real x2;
+                if ( box.type(2)==amrex::IndexType::CellIndex::NODE )
+                     { x2 = real_box.lo(2) + k*dx[2]; }
+                else { x2 = real_box.lo(2) + k*dx[2] + 0.5*dx[2]; }
+                int const ix2 = floor( (x2-offset2)/d2 );
+                amrex::Real const xx2 = offset2 + ix2*d2;
+#endif
+
+#if defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
+                amrex::Array4<double> fc_array(FC_data, {0,0,0}, {extent0, extent2, extent1}, 1);
```
EZoni, line 1435: Is there a specific reasons why here, and in the call below for the 3D case, we cannot use directly `extent[0]`, `extent[1]` and `extent[2]` (which I think would allow us to remove the additional variables `extent0`, `extent1` and `extent2`)?

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1228,3 +1287,174 @@ void WarpX::CheckKnownIssues()
...
+                double
+                    f00 = fc_array(0, ix1  , ix0  ),
+                    f01 = fc_array(0, ix1  , ix0+1),
+                    f10 = fc_array(0, ix1+1, ix0  ),
+                    f11 = fc_array(0, ix1+1, ix0+1);
```
EZoni: Here I just want to double check.

I see that the order of the indices to access `fc_array` in the 2D/RZ cases is:
- 1st index always 0
- 2nd index related to `ix1`
- 3rd index related to `ix0`

and the extents are passed in the order `{extent0, extent2, extent1}`.

On the other hand, the order of the indices to access `fc_array` in the 3D cases is:
- 1st index related to `ix2`
- 2nd index related to `ix1`
- 3rd index related to `ix0`

and the extents are passed in the order `{extent2, extent1, extent0}`.

Somehow the order of the indices vs the order of the extents seem to me to match in 3D but not in 2D/RZ.
Could you help me clarify why this would be the case?

EZoni (commented):
Thank you for the latest code changes, @PhilMiller! I left a couple of more questions.

Source/Initialization/WarpXInitData.cpp
PhilMiller: The array contains elements of `size_t`, while the `Array4` constructor expects arguments of type `int`. The implicit narrowing conversion if they're passed through directly produces compiler warnings.

Source/Initialization/WarpXInitData.cpp
PhilMiller: In the RZ case, 1st index is theta, which is currently asserted as unused because it's not implemented here.

PhilMiller: For complete transparency, this code is utterly untested in the XZ case, because that seemingly garnered no interest from anyone who's been paid to look into this functionality.

PhilMiller: I could add an assertion against that, too, I guess, and leave it for someone who wants to use it to do the testing.

Source/Initialization/WarpXInitData.cpp
PhilMiller: It's written as I meant it, but the wording may be too heavily 'academic CS'. The 'invariant' is "any code besides the solver will see the composite values".

I'm happy to reword it as requested if you find that more comprehensible.

run_test.sh
PhilMiller: ping @RemiLehe

Source/Initialization/WarpXInitData.cpp
EZoni: Thanks Phil. If that's okay with you, yes, I would be in favor of rewording it a little bit.

Source/Initialization/WarpXInitData.cpp
EZoni: Thanks for the comments. I discussed this briefly with @RemiLehe offline, we think it could be nice to choose variable names that create a bit less confusion, with respect to these index "tags" such as `0`, `1` and `2` added to variable names. Maybe we could add `r`, `z` and `mode` suffixes instead and use them only for RZ with `#ifdef` blocks. Or other solutions would be welcome, too. This might require a bit of live brainstorming with you and @RemiLehe.

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1228,3 +1287,174 @@ void WarpX::CheckKnownIssues()
...
+    {
+        amrex::Box box = mfi.growntilebox();
+        amrex::Box tb = mfi.tilebox(nodal_flag, mf->nGrowVect());
+        auto const& mffab = mf->array(mfi);
+
+        // Start ParallelFor
+        amrex::ParallelFor (tb,
+            [=] AMREX_GPU_DEVICE (int i, int j, int k) {
+                // i,j,k denote x,y,z indices in 3D xyz.
+                // i,j,k denote r,z,mode indices in 2D rz.
```
EZoni: Also discussed with @RemiLehe offline, we are not sure that `k` here represents the mode index. Isn't this maybe just an empty index, which can get only one single value (`0`) in RZ (aka, in the 2D version of AMReX's `ParallelFor`), rather than the mode index spanning from `0` to the number of modes?

ax3l:
Hi @PhilMiller, CI complains about a few unused variables.
Can you please push an `amrex::ignore_unused` or remove them where appropriate? :)

Docs/source/usage/parameters.rst
```diff
@@ -1359,6 +1359,14 @@ Grid initialization
     Note that the current implementation of the parser for external B-field
     does not work with RZ and the code will abort with an error message.
 
+    If `B_ext_grid_init_style` is set to be `read_from_file`, an additional parameter,
+    indicating the path of an openPMD data file,
+    `warpx.read_fields_from_path` must be specified,
+    from which external B field data can be loaded into WarpX.
+    One can refer to input files in `Examples/Tests/LoadExternalField` for more information.
+    Regarding how to prepare the openPMD data file, one can refer to
+    the [openPMD-example-datasets](https://github.com/openPMD/openPMD-example-datasets).
```
ax3l: rst not md :)
```suggestion
    the `openPMD-example-datasets <https://github.com/openPMD/openPMD-example-datasets>`__.
```

Docs/source/usage/parameters.rst
```diff
@@ -1359,6 +1359,14 @@ Grid initialization
     Note that the current implementation of the parser for external B-field
     does not work with RZ and the code will abort with an error message.
 
+    If `B_ext_grid_init_style` is set to be `read_from_file`, an additional parameter,
+    indicating the path of an openPMD data file,
+    `warpx.read_fields_from_path` must be specified,
+    from which external B field data can be loaded into WarpX.
+    One can refer to input files in `Examples/Tests/LoadExternalField` for more information.
```
```suggestion
    One can refer to input files in ``Examples/Tests/LoadExternalField`` for more information.
```

Docs/source/usage/parameters.rst
```diff
@@ -1359,6 +1359,14 @@ Grid initialization
     Note that the current implementation of the parser for external B-field
     does not work with RZ and the code will abort with an error message.
 
+    If `B_ext_grid_init_style` is set to be `read_from_file`, an additional parameter,
+    indicating the path of an openPMD data file,
+    `warpx.read_fields_from_path` must be specified,
```
ax3l: rst being rst
```suggestion
    If ``B_ext_grid_init_style`` is set to be ``read_from_file``, an additional parameter,
    indicating the path of an openPMD data file,
    ``warpx.read_fields_from_path`` must be specified,
```

Docs/source/usage/parameters.rst
```diff
@@ -1384,6 +1392,17 @@ Grid initialization
...
 
+    If `E_ext_grid_init_style` is set to be `read_from_file`, an additional parameter,
+    indicating the path of an openPMD data file,
+    `warpx.read_fields_from_path` must be specified,
+    from which external E field data can be loaded into WarpX.
+    One can refer to input files in `Examples/Tests/LoadExternalField` for more information.
+    Regarding how to prepare the openPMD data file, one can refer to
+    the [openPMD-example-datasets](https://github.com/openPMD/openPMD-example-datasets).
+    Note that if both `B_ext_grid_init_style` and `E_ext_grid_init_style` are set to
+    `read_from_file`, the openPMD file specified by `warpx.read_fields_from_path`
```
```suggestion
    If ``E_ext_grid_init_style`` is set to be ``read_from_file``, an additional parameter,
    indicating the path of an openPMD data file,
    ``warpx.read_fields_from_path`` must be specified,
    from which external E field data can be loaded into WarpX.
    One can refer to input files in ``Examples/Tests/LoadExternalField`` for more information.
    Regarding how to prepare the openPMD data file, one can refer to
    the `openPMD-example-datasets <https://github.com/openPMD/openPMD-example-datasets>`__.
    Note that if both ``B_ext_grid_init_style`` and ``E_ext_grid_init_style`` are set to
    ``read_from_file``, the openPMD file specified by ``warpx.read_fields_from_path``
```

Regression/WarpX-tests.ini
```diff
@@ -3905,6 +3905,42 @@ compareParticles = 1
...
+numthreads = 1
+compileTest = 0
+doVis = 0
+compareParticles = 1
+particleTypes = proton
+analysisRoutine = Examples/Tests/LoadExternalField/analysis_rz.py
+
+[LoadExternalField3D]
+buildDir = .
+inputFile = Examples/Tests/LoadExternalField/inputs_3d
```
ax3l, line 3928: Do we also need a 1D and 2D test?

Can be a follow-up PR, too.

Source/Initialization/WarpXInitData.cpp
```diff
@@ -440,6 +449,23 @@
     CheckKnownIssues();
 }
 
+void
+WarpX::AddExternalFields () {
+    for (int lev = 0; lev <= finest_level; ++lev) {
+        // FIXME: RZ multimode has more than one component for all these
```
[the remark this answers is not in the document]
ax3l: Is this fixed with the RZ example or still to do?

if to do, then let us:
- open a new issue documenting RZ, 2D and 1D needs for follow-up work,
- add runtime asserts here for the things not yet supported (and link the PR in the message) - I see 1D and RZ N>1 already has an assert :+1: but I think Cartesian 2D might be missing.

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1228,3 +1288,189 @@ void WarpX::CheckKnownIssues()
...
+    const amrex::RealBox& real_box = geom0.ProbDomain();
+    const auto dx = geom0.CellSizeArray();
+    amrex::IntVect nodal_flag = mf->ixType().toIntVect();
+
+    // Read external field openPMD data
+    auto series = openPMD::Series(read_fields_from_path, openPMD::Access::READ_ONLY);
+    auto iseries = series.iterations.begin()->second;
+    auto F = iseries.meshes[F_name];
+
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(F.getAttribute("dataOrder").get<std::string>() == "C",
```
ax3l, line 1310: Excellent assert work here, thanks!

run_test.sh
ax3l: Phil, here is an example:
https://github.com/openPMD/openPMD-api/blob/dev/share/openPMD/download_samples.sh

ax3l (commented):
This looks all excellent - I added comments to finalize and then we can track follow-ups in a separate issue.

Source/Initialization/WarpXInitData.cpp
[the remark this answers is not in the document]
PhilMiller: the XZ case has an `amrex::Abort` guarding it in `ReadExternalFieldFromFile`.

I'll open issues for the follow-up work.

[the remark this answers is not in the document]
PhilMiller: #3828, #3829, and #3830 opened and linked in the source.

Source/Initialization/WarpXInitData.cpp
PhilMiller: Fixed.

PhilMiller:
Just need to shift downloading of the example data used in the tests away from cloning the repo, and I think this is ready to go.

run_test.sh
ax3l: ping @PhilMiller to finalize :)

ax3l:
```suggestion
curl --create-dirs --output-dir openPMD-example-datasets -sOL https://github.com/openPMD/openPMD-example-datasets/raw/4ba1d257c5b4897c0a3cd57742bb0987343a902e/example-femm-thetaMode.h5
curl --create-dirs --output-dir openPMD-example-datasets -sOL https://github.com/openPMD/openPMD-example-datasets/raw/4ba1d257c5b4897c0a3cd57742bb0987343a902e/example-femm-3d.h5
```

Examples/Tests/LoadExternalField/inputs_rz
```diff
@@ -0,0 +1,68 @@
+warpx.serialize_initial_conditions = 0
+warpx.do_dynamic_scheduling = 0
+particles.do_tiling = 0
+
+warpx.B_ext_grid_init_style = "read_from_file"
+warpx.read_fields_from_path = "../../../../openPMD-example-datasets/example-femm-thetaMode.h5"
```
ax3l, line 6:
```suggestion
warpx.read_fields_from_path = "../../../../example-femm-thetaMode.h5"
```

run_test.sh
```diff
@@ -74,6 +74,9 @@ git clone https://github.com/AMReX-Codes/amrex.git
 cd amrex && git checkout --detach 1ffc6e427e3aea9302602190c8fc1cfbea0c9725 && cd -
 # warpx-data contains various required data sets
 git clone --depth 1 https://github.com/ECP-WarpX/warpx-data.git
+# openPMD-example-datasets contains various required data sets
+curl --create-dirs --output-dir openPMD-example-datasets -sOL https://github.com/openPMD/openPMD-example-datasets/raw/4ba1d257c5b4897c0a3cd57742bb0987343a902e/example-femm-thetaMode.h5
+curl --create-dirs --output-dir openPMD-example-datasets -sOL https://github.com/openPMD/openPMD-example-datasets/raw/4ba1d257c5b4897c0a3cd57742bb0987343a902e/example-femm-3d.h5
```
ax3l:
```suggestion
mkdir -p openPMD-example-datasets
cd openPMD-example-datasets
curl -sOL https://github.com/openPMD/openPMD-example-datasets/raw/4ba1d257c5b4897c0a3cd57742bb0987343a902e/example-femm-thetaMode.h5
curl -sOL https://github.com/openPMD/openPMD-example-datasets/raw/4ba1d257c5b4897c0a3cd57742bb0987343a902e/example-femm-3d.h5
cd -
```

ax3l (approved):

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1228,3 +1288,195 @@ void WarpX::CheckKnownIssues()
...
+                                     "Reading from files with non-C dataOrder is not implemented");
+
+    auto axisLabels = F.getAttribute("axisLabels").get<std::vector<std::string>>();
+    auto fileGeom = F.getAttribute("geometry").get<std::string>();
+
+#if defined(WARPX_DIM_3D)
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(fileGeom == "cartesian", "3D can only read from files with cartesian geometry");
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(axisLabels[0] == "x" && axisLabels[1] == "y" && axisLabels[2] == "z",
+                                     "3D expects axisLabels {x, y, z}");
```
ax3l, line 1319: Things we need to improve:
- openPMD 1.* defines a `dataOrder` attribute, which tells us if we need to invert the arguments in `axisLabel`, spacing, etc. during read
- AMReX uses Fortran data order to label `AMReXF[x,y,z]` which is for pure C/C++ order `F[z,y,x]` with `x` the fastest running index in both cases.

PR #3588, opened 2022-12-21
Docs: Fix HTML5 for Install Logos

ax3l:
With the update tot HTML5 in docutils, we need to modernize our CSS code that does fancy logos in the user install section.

Follow-up to #3546


RemiLehe (approved):

dpgrote (approved):
One of the smallest PRs ever. Thanks!

PR #3589, opened 2022-12-21
Adding radial weighting at r=0 for Jr if nodal

clarkse:
In case of nodal currents, as in the magnetostatic solver, during radial current weighting, the case of handling the radial current at r=0 needs to be handled.  This adds in this logic and pins the radial current to zero at r=0 similarly to how Jt is pinned to zero at r=0.


dpgrote (approved):

PR #3590, opened 2022-12-21
Adding Amrex Bucket into WarpX Python interface.

clarkse:
This PR adds the Amrex bucket into the Python WarpX interface to allow setting of Amrex parameters from the PICMI input scripts.

- [x] PICMI-like options used
- [x] user-facing documentation

### DIFF
--- a/Python/pywarpx/Amrex.py
+++ b/Python/pywarpx/Amrex.py
@@ -1,0 +1,9 @@
+# Copyright 2022 S. Eric CLark, LLNL
+#
+# This file is part of WarpX.
+#
+# License: BSD-3-Clause-LBNL
+
+from .Bucket import Bucket
+
+amrex = Bucket('amrex')
--- a/Python/pywarpx/WarpX.py
+++ b/Python/pywarpx/WarpX.py
@@ -10,6 +10,7 @@ import re
 from . import Particles
 from .Algo import algo
 from .Amr import amr
+from .Amrex import amrex
 from .Boundary import boundary
 from .Bucket import Bucket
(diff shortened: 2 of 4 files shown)


dpgrote:
Which amrex parameters do you need to set?

clarkse:
> 

In WSL, I need to set amrex.the_arena_is_managed=0, and amrex.the_arena_init_size, as when the arena is not managed, the default arena memory allocation oversubscribes the GPU if using more than one MPI process/GPU.

dpgrote:
Ok. It would be more PICMI like if these were added as arguments to the `Simulation` class in `picmi.py`. See for example how the argument `warpx_amr_check_input` is setup. You would add arguments `warpx_amrex_the_arena_is_managed` and `warpx_amrex_the_arena_init_size`.

clarkse:
@dpgrote I have added the PICMI like interface.

dpgrote (approved):

ax3l (dismissed):
Awesome!

As a last step, can you please add the new options in the user-facing docs?
`Docs/source/usage/python.rst` builds the page https://warpx.readthedocs.io/en/latest/usage/python.html

ax3l:
(oops, did not mean to approve yet)

clarkse:
> (oops, did not mean to approve yet)

@ax3l It appears the docs pull from the inline comments in picmi.py.  I do see anything that I should be changing in the python.rst file.  I built the docs locally and they appear to have updated appropriately.  

image

ax3l:
Oh, cool! That makes sense :)

ax3l (approved):

Issue #3541, opened 2022-11-29
Compilation issue with ADIOS+conda on MacOS


RemiLehe:
When installing dependencies with conda (https://warpx.readthedocs.io/en/latest/install/dependencies.html#conda-linux-macos-windows) on MacOS, I get the following error when compiling WarpX:
```
[  3%] Linking CXX shared library ../../lib/libopenPMD.ADIOS1.Serial.dylib
ld: warning: -pie being ignored. It is only used when linking a main executable
Undefined symbols for architecture x86_64:
  "openPMD::internal::operationAsString(openPMD::Operation)", referenced from:
      openPMD::ADIOS1IOHandlerImpl::flush() in ADIOS1IOHandler.cpp.o
ld: symbol(s) not found for architecture x86_64
clang-13: error: linker command failed with exit code 1 (use -v to see invocation)
gmake[2]: *** [_deps/fetchedopenpmd-build/CMakeFiles/openPMD.ADIOS1.Serial.dir/build.make:133: lib/libopenPMD.ADIOS1.Serial.dylib] Error 1
```

RemiLehe:
Interestingly, compiling on Ubuntu (using the same instructions, i.e. using `conda` to install dependencies) does not cause any issue.

ax3l:
oh, I fixed this already and have to do a release of openPMD-api.
https://github.com/openPMD/openPMD-api/issues/1287

So long, please use `-DopenPMD_USE_ADIOS1=OFF` or rely on conda's pre-build openPMD-api by using `-DWarpX_openpmd_internal=OFF`.

ax3l:
#3591 will unbreak this even before the next release - we can disable ADIOS1 by default (as we will also in openPMD-api in the upcoming release, which deprecates ADIOS1 in favor of ADIOS2 that we use anyway)

PR #3591, opened 2022-12-22
CMake: ADIOS1 off by default

ax3l:
We will deprecate ADIOS1 and will not build it by default anymore in the upcoming release. Since we only use ADIOS2 in WarpX and builds are sometimes fragile with ADIOS1, we just disable it now in superbuilds already.

Related to #3541 (fixes the defaults on macOS with the 0.14.5 release :) )


RemiLehe (approved):

RemiLehe:
Fixes #3541

PR #3597, opened 2022-12-28
Fix a Bug in 1D ES

### LINKED ISSUE #3559: Boundary issue in 1D ES solve
The current ES solver generates spurious fields close to the left boundary.  Please see plot of E_z below ran with periodic BC.  Roelof already found a solution using a SuperLU solve as you can observe on the bottom . 

image

image

WeiqunZhang:
After the tri-diagonal solve, the data in the special single-box MultiFab need to be copied back to the normal MultiFab.  However, when there was a bug for periodic boundary.  The issue is the special MultiFab is really special.  It has a nodal Box that grows the original nodal Box by 1 node in the lower boundary and 2 (why 2?) nodes in the upper boundary and these grown nodes are considered valid nodes, even though they do not have valid data.  Therefore amrex::ParallelCopy with periodicity does not work in this case.  For example, for a domain with 96 cells, the data at nodes 1 and 2 of the normal MultiFab could be written with the invalid data at nodes 97 and 98 of the special MultiFab.  This bug is fixed by replacing it with a ParallelCopy without periodicity followed by a FillBoundary.

### DIFF
--- a/Source/FieldSolver/ElectrostaticSolver.cpp
+++ b/Source/FieldSolver/ElectrostaticSolver.cpp
@@ -859,10 +859,9 @@ WarpX::computePhiTriDiagonal (const amrex::Vector<std::unique_ptr<amrex::MultiFa
 
     }
 
-    // Copy phi1d to phi, including the x guard cell
-    const IntVect xghost(AMREX_D_DECL(1,0,0));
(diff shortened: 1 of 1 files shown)


roelof-groenewald (approved):
Thanks for looking into and fixing this issue!

ax3l (commented):
Thanks a lot! :tada:

PR #3600, opened 2023-01-03
1D ES: Remove an unnecessary FillBoundary

WeiqunZhang:
There is no need to fill the ghost nodes of phi, because the stencil used to compute cell-centered E field from nodal values of phi does not use ghost nodes.


WeiqunZhang:
Superseded by #3604.

PR #3604, opened 2023-01-05
For the tridiag, remove the guard cells since they are not needed

dpgrote:
This is a clean up of the tridiag Poisson solver used with ES 1D. The guard cells of the potential array are never used so this PR cleans up the code to not set the guard cells.

### DIFF
--- a/Source/FieldSolver/ElectrostaticSolver.cpp
+++ b/Source/FieldSolver/ElectrostaticSolver.cpp
@@ -718,24 +718,20 @@ WarpX::computePhiTriDiagonal (const amrex::Vector<std::unique_ptr<amrex::MultiFa
         nx_solve_max = nx_full_domain;
     }
 
-    // Create a 1-D MultiFab that covers all of x, including guard cells on each end.
+    // Create a 1-D MultiFab that covers all of x.
     // The tridiag solve will be done in this MultiFab and then copied out afterwards.
-    const amrex::IntVect lo_full_domain(AMREX_D_DECL(-1,0,0));
-    const amrex::IntVect hi_full_domain(AMREX_D_DECL(nx_full_domain+1,0,0));
-    const amrex::Box box_full_domain(lo_full_domain, hi_full_domain);
-    const BoxArray ba_full_domain(box_full_domain);
-    amrex::DistributionMapping dm_full_domain;
+    const amrex::IntVect lo_full_domain(AMREX_D_DECL(0,0,0));
+    const amrex::IntVect hi_full_domain(AMREX_D_DECL(nx_full_domain,0,0));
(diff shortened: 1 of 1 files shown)


Source/FieldSolver/ElectrostaticSolver.cpp
```diff
@@ -718,24 +718,20 @@ WarpX::computePhiTriDiagonal (const amrex::Vector<std::unique_ptr<amrex::MultiFa
...
     // the data readily accessible from the GPU.
-    auto phi1d_mf = MultiFab(ba_full_domain_node, dm_full_domain, ncomps1d, nguard1d, MFInfo().SetArena(The_Pinned_Arena()));
-    auto zwork1d_mf = MultiFab(ba_full_domain_node, dm_full_domain, ncomps1d, nguard1d, MFInfo().SetArena(The_Pinned_Arena()));
-    auto rho1d_mf = MultiFab(ba_full_domain_node, dm_full_domain, ncomps1d, nguard1d, MFInfo().SetArena(The_Pinned_Arena()));
+    auto phi1d_mf = MultiFab(ba_full_domain_node, dm_full_domain, 1, 0, MFInfo().SetArena(The_Pinned_Arena()));
+    auto zwork1d_mf = MultiFab(ba_full_domain_node, dm_full_domain, 1, 0, MFInfo().SetArena(The_Pinned_Arena()));
+    auto rho1d_mf = MultiFab(ba_full_domain_node, dm_full_domain, 1, 0, MFInfo().SetArena(The_Pinned_Arena()));
 
     // Copy previous phi to get the boundary values
     phi1d_mf.ParallelCopy(*phi[lev], 0, 0, 1, Geom(lev).periodicity());
```
WeiqunZhang: Do we need to do periodic copy?

dpgrote: No, the periodic copy is not needed. This line is only needed when there are Dirichlet boundaries. I'll update this accordingly.

For rho, the periodicity may not be needed either since the periodicity can be handled in the solver.

dpgrote: Though, what does the periodicity argument do here? When true, does it mean that the value on the upper boundary will be obtained from the lower boundary? Does `SynchRho` set the rho on the upper boundary, in which case the periodicity is not needed?

WeiqunZhang: It means the value at the lower and upper boundaries can be set to either the lower or the upper boundary value in the source MultiFab.  If the two values are not consistent, the behavior is undefined in some sense.,

WeiqunZhang (approved):

PR #3615, opened 2023-01-11
RZ: added abort message if warpx.numprocs[0] > 1 is set.

oshapoval:
RZ: If domain decomposition is set using number of `numprocs`, we need to add an abort error message if `warpx.numprocs[0] > 1`, because domain decomposition in RZ only works along `z` direction.


Source/WarpX.cpp
```diff
@@ -504,6 +504,10 @@ WarpX::ReadParameters ()
         std::vector<int> numprocs_in;
         utils::parser::queryArrWithParser(
             pp_warpx, "numprocs", numprocs_in, 0, AMREX_SPACEDIM);
+        #ifdef WARPX_DIM_RZ
+                WARPX_ALWAYS_ASSERT_WITH_MESSAGE( numprocs_in[0] == 1,
+                        "Domain decomposition in RZ works only along z direction");
+        #endif
 
         if (not numprocs_in.empty()) {
```
EZoni: Thank you, @oshapoval! I think the assert needs to go inside the `if` block below, because it needs to be checked only if `numprocs` is provided in the inputs (and since it is an optional parameter, that's not always the case). This should fix CI.
```suggestion

        if (not numprocs_in.empty()) {
#ifdef WARPX_DIM_RZ
            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(numprocs_in[0] == 1,
                "Domain decomposition in RZ works only along z direction");
#endif
```

Source/WarpX.cpp
```diff
@@ -506,6 +506,10 @@ WarpX::ReadParameters ()
             pp_warpx, "numprocs", numprocs_in, 0, AMREX_SPACEDIM);
 
         if (not numprocs_in.empty()) {
+#ifdef WARPX_DIM_RZ
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE( numprocs_in[0] == 1,
+            "Domain decomposition in RZ works only along z direction");
```
```suggestion
            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(numprocs_in[0] == 1,
                "Domain decomposition in RZ works only along z direction");
```

EZoni (dismissed):

EZoni:
@dpgrote
I can't remember right now, is radial domain decomposition forbidden in RZ for PSATD only or for both PSATD and FDTD? Seems to me like it's only a constraint for PSATD from this function: https://github.com/ECP-WarpX/WarpX/blob/b7945ae972f4a09b650c8ac7b18083847e0a1f9b/Source/Utils/WarpXUtil.cpp#L298

dpgrote:
@EZoni It is only a limitation for the spectral solver. The FDTD can be decomposed radially.

EZoni:
Thanks, @dpgrote.

Then, @oshapoval, could you put the assert within an additional `if` block like the following one,
```
if (electromagnetic_solver_id == ElectromagneticSolverAlgo::PSATD)
```
so that we check, and possibly abort, only when we use the PSATD Maxwell solver?

oshapoval:
> @EZoni It is only a limitation for the spectral solver. The FDTD can be decomposed radially.

Thanks for checking this, @dpgrote .

oshapoval:
> Then, @oshapoval, could you put the assert within an additional `if` block like the following one,
> 
> ```
> if (electromagnetic_solver_id == ElectromagneticSolverAlgo::PSATD)
> ```
> 
> so that we check, and possibly abort, only when we use the PSATD Maxwell solver?
@EZoni Fixed in d8f865def5ae436de306d283a0c07a92612c8a0c.

EZoni (approved):
Thank you for this PR, @oshapoval!

PR #3560, opened 2022-12-09
Doxygen CI: Warnings as Error

ax3l:
Make sure that no Doxygen warnings get in, as we do in other CI runs already.

Depends on https://github.com/AMReX-Codes/amrex/pull/3061

PR #3616, opened 2023-01-12
Fix Doxygen Warnings

ax3l:
Fix a few Doxygen warnings.
Pulled out of the infrastructure update in #3560 that currently seems to hang.


dpgrote (approved):
Thanks for separating these out from the other PR.

PR #3620, opened 2023-01-13
CI: Cache & Checkout Action v3

ax3l:
Update the Cache, Checkout and Python Setup actions to the latest versions.


EZoni (commented):
Thank you, @ax3l! I see that you removed `usr/local/` from some paths (see files cuda.yml, nvcc11-0.sh, nvcc11-8.sh). Are these changes left over from #3621 or are they still needed in this PR?

ax3l:
Thanks! Oh that is not intentional, let me clean up.

EZoni (approved):

ax3l:
While we are on it, let me also update the checkout action :)

PR #3621, opened 2023-01-13
CI: /usr/local lost PATH

ax3l:
It looks like on Linux runners, `/usr/local/` is not in the default `${PATH}` anymore.


ax3l:
This looks more like a general connectivity issue of Ubuntu runners on Azure (incl. GH Actions).

PR #3624, opened 2023-01-13
EB potential python interface

dpgrote:
This adds the Python level routine `set_potential_EB` that allows changing the expression for the embedded boundary potential.


RemiLehe (approved):
Great! Thanks a lot for adding this!

PR #3627, opened 2023-01-17
Fix few includes in WarpXAMReXInit.cpp/H and TextMsg.cpp using IWYU tool

lucafedeli88:
This fixes a couple of issues found with the Include What You Use tool


Source/Initialization/WarpXAMReXInit.H
```diff
@@ -9,6 +9,10 @@
 
 #include <AMReX.H>
 
+#ifdef AMREX_USE_MPI
+#   include <mpi.h>
+#endif
```
ax3l: Since `MPI_Comm` us used unconditionally below, this include would be better:
```suggestion
#include <AMReX_ccse-mpi.H>
```

It is also pulled by `<AMReX.H>`.

ax3l (commented):
Thanks!
Only one that I would either change or not include at all.

Source/Initialization/WarpXAMReXInit.H
lucafedeli88: ah ! Maybe by doing this we could avoid including AMReX.H in that file

ax3l (approved):

PR #3629, opened 2023-01-18
Docs: Summit Jupyter HDF5 Back

ax3l:
The issues on OLCF Jupyter to read HDF5 data from GPFS production directories have been resolved. We can use the previous work-around again and reads will not get stuck.

### DIFF
--- a/Docs/source/install/hpc/summit.rst
+++ b/Docs/source/install/hpc/summit.rst
@@ -16,12 +16,12 @@ If you are new to this system, **please see the following resources**:
 * `Jupyter service <https://jupyter.olcf.ornl.gov>`__
 * `Production directories <https://docs.olcf.ornl.gov/data/index.html#data-storage-and-transfers>`_:
 
-  * ``$PROJWORK/$proj/``: shared with all members of a project, purged every 90 days (recommended)
-  * ``$MEMBERWORK/$proj/``: single user, purged every 90 days (usually smaller quota)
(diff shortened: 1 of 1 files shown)


Docs/source/install/hpc/summit.rst
```diff
@@ -264,6 +265,9 @@ Known System Issues
    The above **HDF5 Jupyter read** work-around for OLCFHELP-3685 does not work anymore, due to the way that GPFS is mounted via NSF on Jupyter nodes.
    As a work-around until this is fixed, please copy your HDF5 data to ``/ccs``, ``$HOME`` or use ADIOS2 BP instead of HDF5 files.
 
+   Jan 18th, 2023:
+   The work-around above works again with GPFS data.
+
```
RemiLehe: Should we remove this whole warning then? 
We can still see the history of this from the `git` history, if we need it.

RemiLehe (changes requested):
It's great that this is working again!
Should we remove the previous warning?

Docs/source/install/hpc/summit.rst
ax3l: True, simpler. And we have also versions in our RTD pages.

RTSandberg (approved):
Glad to have a more straightforward access to HDF5 from GPFS again

RemiLehe (approved):

PR #3632, opened 2023-01-23
Clean loadbalance params

hklion:
- Added default values of `algo.costs_heuristic_particles_wt` and `algo.costs_heuristic_cells_wt` to documentation.
- Only read in knapsack factor and heuristic weight params if they will actually be used.


ax3l (approved):
Perfect, thanks a lot for this, Hannah!

PR #3633, opened 2023-01-23
added space to warning message when too many boxes per GPU

hklion:
There was a space missing in the warning message that gets output if there are too many boxes per GPU. I've added it.


ax3l (approved):
Thanks for this! :)

Issue #3631, opened 2023-01-21
Installation problem


shefalys:
Hello everyone, I am a new member of warpx community. I am struggling with installation of warpx on my desktop (OS: ubuntu 22.04)

I followed the documentation for installation of the code and was trying a given 2d example script. I stuck with an error. I followed the following procedures to install Warpx on my desktop computer:

(1) conda create -n warpx -c conda-forge warpx
(2) conda activate warpx

Now the command terminal looks like:
(warpx) shafaley@shafaley: ~/Documents/WarpX

Then, I used the following command to test the 2d script on my desktop
mpirun -np 16 ./warpx inputs_2d

Following error message appear:
[proxy:0:0@shafaley] HYDU_create_process (utils/launch/launch.c:73): execvp error on file ./warpx (No such file or directory)

How to fix this error ?

#### Additional information ####

If I do warpx.(tab)
I get:
Warpx.1d.NOMPI.OMP.DP.OPMD.PSATD.QED
Warpx.2d.NOMPI.OMP.DP.OPMD.PSATD.QED
Warpx.3d.NOMPI.OMP.DP.OPMD.PSATD.QED
Warpx.RZ.NOMPI.OMP.DP.OPMD.PSATD.QED

that means it install .... am I right ?

========================================
#### Try with PICMI ####

Not giving up and I tried python command:  **mpirun -np 16 python3 PICMI_inputs_2d.py** 
and it looks like I can run the python example script: **PICMI_inputs_2d.py**
but what is it giving has no idea !!!  generating folders with name, for example, Python_LaserAccelerationMR_plt000000.old.9127500 and so on ...

at the moment difficult to see where the data is  ....... 

I need some help up to the point where I can at least run the example scripts properly and can post-process some data files with openpmd-viewer …. 

**once again, please note that I only used the two command to install warpx**
(1) conda create -n warpx -c conda-forge warpx
(2) conda activate warpx

Thanks
Shefalys

ax3l:
@shefalys thanks for reaching out, you are almost there.

## WarpX as an Application

If you install with conda, you will not get a `warpx` symbolic link but as you correctly noticed a series of pre-installed WarpX binaries.

So, in order to run WarpX as an application replace this:
```console
./warpx inputs_2d
```
with
```console
warpx.2d.NOMPI.OMP.DP.OPMD.PSATD.QED inputs_2d
```

### What happened here?

[...]

PR #3636, opened 2023-01-24
Docs: Conda MPI Variants

ax3l:
Document how to install the MPI variants of WarpX with conda/mamba.

Seen in https://github.com/ECP-WarpX/WarpX/issues/3631#issuecomment-1401357414


ax3l:
Ah no, we do not yet have MPI variants in conda for WarpX.
(OpenMP is best for desktop anyway.)

PR #3637, opened 2023-01-24
Fix bug with PML in 2D when compiling with EB

RemiLehe:
In the current version of WarpX, the PML field update is not done properly **in 2D, if the code is compiled with EB support**.

There are 2 bugs that are causing this, which are being fixed in this PR:

- When calling the `ScaleEdges` function, we were passing an array of cellsizes that, in 2D, looked like `{dx, dz, 0}` while the function expected something of the form `{dx, 0, dz}`. This is now fixed by calling `WarpX::CellSize(lev)`.

- The `if` condition that ensures that the PML field is not updated in the EB was not implemented correctly. We now use the same `if` condition for the PML field as for the regular field:
https://github.com/ECP-WarpX/WarpX/blob/development/Source/FieldSolver/FiniteDifferenceSolver/EvolveE.cpp#L174

Many thanks to @anecas1979 for finding this bug and helping with the debugging process!


EZoni (approved):
Thank you for this fix, @RemiLehe!

PR #3643, opened 2023-01-27
Docs: Fix `n_rz_azimuthal_modes`

ax3l:
Renamed a while ago and docs were outdated.

Thanks to @titoiride for reporting!


dpgrote (approved):
Thanks for the fix!

Issue #3641, opened 2023-01-27
Add Titanium for ionization


kevinambro:
We would like to use ionization for a simulation of a laser interacting with a titanium target, would it be possible to add support for Z=22 in IonizationEnergiesTables.H?

n01r:
Hi @kevinambro, thanks for using WarpX!
I prepared a quick pull request. Feel free to test the branch.
This should hopefully be all that is needed!

ax3l:
Implemented via #3645, further updates in #3646.

@kevinambro let us know if this works or if you have further questions :) Closing for now as solved.

PR #3645, opened 2023-01-28
Add ionization support for Titanium

n01r:
This PR adds ionization support for Titanium as asked for in WarpX Issue #3641.


dpgrote:
While you're at it, is there a reason that we don't just include all of the elements? Yes, the files are larger, but it is still a tiny fraction of the overall size of the WarpX source.

lucafedeli88:
> While you're at it, is there a reason that we don't just include all of the elements? Yes, the files are larger, but it is still a tiny fraction of the overall size of the WarpX source.

I think that this is a good idea. We should consider adding that in a future PR

lucafedeli88 (approved):
Looks good to me! Thanks for this PR!

PR #3656, opened 2023-01-31
Clean compilation warning on unused function

RemiLehe:
This removes the compilation warning:
```
warning #177-D: function "<unnamed>::PDim3::operator=(const amrex::XDim3 &)" was declared but never referenced
```

### DIFF
--- a/Source/Particles/PhysicalParticleContainer.cpp
+++ b/Source/Particles/PhysicalParticleContainer.cpp
@@ -150,8 +150,6 @@ namespace
 
         AMREX_GPU_HOST_DEVICE
         PDim3& operator=(const PDim3&) = default;
-        AMREX_GPU_HOST_DEVICE
-        PDim3& operator=(const amrex::XDim3 &a) { x = a.x; y = a.y; z = a.z; return *this; }
     };
 
     AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE


ax3l:
@KZhu-ME @roelof-groenewald do you remember this one? Was this fixed in new NVCC versions already and is transitory?

roelof-groenewald:
I actually don't remember where that came from but if it is unused then I don't see any issue with removing it :)

roelof-groenewald (approved):

ax3l:
Logic was from here: #3065

Cannot find the discussion anymore on the warning...

I thought @KZhu-ME had a point where it is used in downstream and that this diagnostics warning is actually a bug in NVCC, that was already fixed...

I would not remove this if fixed in newer NVCC and also now without also removing the constructor for the same type (rule of 5): https://en.cppreference.com/w/cpp/language/rule_of_three

KZhu-ME (approved):

KZhu-ME:
I think @PhilMiller can provide more insight on this issue, since I think he was the one who first pointed it out to me.

PhilMiller:
I think we were expecting to use it in downstream code at the time, but maybe never did. I see no reason not to remove it, given that it's so easy to reintroduce if needed

ax3l:
Ok, thanks. Unused class methods should not throw in diagnostics, this is a  compiler bug in NVCC.

I think we should either also remove the constructor `PDim3(const amrex::XDim3& a)` or leave it in and add a compiler warning suppression for NVCC. But ok to remove it for sake of "not used yet".

ax3l (approved):

Issue #3658, opened 2023-01-31
Particle initialization


tmiethlinger:
Hello,

I have two questions / possible issues regarding particle initialization, for `WarpX_DIMS=1`.

### 1.) NUniformPerCell

My input file:
[input_NUniformPerCell.txt](https://github.com/ECP-WarpX/WarpX/files/10551749/input_NUniformPerCell.txt)

> NUniformPerCell: injection with a fixed number of evenly-spaced particles per cell

Using NUniformPerCell, I expected evenly-spaced particles per cell, which is however not the case.

If I plot the particle positions, then they are all located at their respective cell center: 
![image](https://user-images.githubusercontent.com/91594368/215903596-aed74c47-5643-4d13-8213-26eed35fd29b.png)
The "step size" in the staircase shape comes from ppc=500

Plot script:
```
sim = 1
simpath = campaign_dir + "/runs/" + str(sim)
filepath = simpath + "/diags/openPMD/openpmd_" + s + ".h5"
f = h5py.File(filepath, 'r')
Rz = f['data/' + str(i) + '/particles/electron/position/z'][()]
```

### 2.) NRandomPerCell
Basically the same input script as above, but with
```
NUniformPerCell -> NRandomPerCell
num_particles_per_cell_each_dim -> num_particles_per_cell
```
[input_NRandomPerCell.txt](https://github.com/ECP-WarpX/WarpX/files/10551807/input_NRandomPerCell.txt)

I don't know if bug or intended, but if I do the same simulation but different `warpx.random_seed` values, I end up with the same initial particle positions.
![image](https://user-images.githubusercontent.com/91594368/215905221-c928a516-a9ab-4b73-a8be-e37dff8bbb0a.png)

My script to read the data, and compare Rz1 with Rz2. Since they have different seeds, I'd have expected different values.
```
i = 0 # Inspect initial particle positions
s = str(i).zfill(6)

sim = 1
simpath = campaign_dir + "/runs/" + str(sim)
filepath = simpath + "/diags/openPMD/openpmd_" + s + ".h5"
f = h5py.File(filepath, 'r')
Rz1 = f['data/' + str(i) + '/particles/electron/position/z'][()]

sim = 2
simpath = campaign_dir + "/runs/" + str(sim)
filepath = simpath + "/diags/openPMD/openpmd_" + s + ".h5"
f = h5py.File(filepath, 'r')
Rz2 = f['data/' + str(i) + '/particles/electron/position/z'][()]
```

[...]

PR #3659, opened 2023-01-31
Fix NUniformPerCell for 1D

dpgrote:
It was noticed by several people that when running 1D with NUniformPerCell, all of the particles in each cell were being placed at the same location, at the center of the cell. See issue #3658. This was caused by the wrong value being used by the code that was generating the positions. This PR fixes it.

This will break several CI tests so that the benchmarks will need to be updated before merging.


RemiLehe (approved):

RemiLehe:
Thanks for fixing this!

RemiLehe (changes requested):
As can be expected, this broke the checksums for a number of 1D tests ; could you regenerate those checksums and update them as part of this PR?

dpgrote:
Yep, I will update the benchmarks. I wanted to see which ones failed.

RemiLehe (approved):

PR #3661, opened 2023-02-01
updates classical radiation reaction entry in docs

aeriforme:
This PR adds missing information to the documentation regarding the usage of
classical radiation reaction together with the quantum synchrotron module.


lucafedeli88 (approved):
Thanks @aeriforme for this PR!

PR #3663, opened 2023-02-01
change plotfile RZ allowed diag to r,t from x,y

RTSandberg:
Previously there were multiple diagnostic naming conventions in RZ: openPMD RZ transverse field were `r`,`t` and plotfile RZ transverse fields were `x`,`y`.  Furthermore, if `<diag>.dump_rz_modes=1` and all modes are written to plotfile then the modes are labeled as `<field>r_<mode>_<real/imag>`.  Now all RZ transverse fields are labeled `r` and `t`.


Python/pywarpx/picmi.py
```diff
@@ -1744,27 +1744,39 @@ def initialize_inputs(self):
...
-                    fields_to_plot.add('By')
-                    fields_to_plot.add('Bz')
+                    for field_name in B_fields_list:
+                        fields_to_plot.add(field_name)
                 elif dataname == 'J':
-                    fields_to_plot.add('jx')
-                    fields_to_plot.add('jy')
-                    fields_to_plot.add('jz')
+                    for field_name in J_fields_list:
+                        fields_to_plot.add(field_name.lower())
```
EZoni, line 1767: What does `lower()` do here and why is it done only for `J`?

Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -635,7 +619,49 @@ FullDiagnostics::InitializeFieldFunctors (int lev)
...
+            } else if ( m_varnames[comp] == "Bt" ){
+                m_all_field_functors[lev][comp] = std::make_unique<CellCenterFunctor>(warpx.get_pointer_Bfield_aux(lev, 1), lev, m_crse_ratio);
+            } else if ( m_varnames[comp] == "jr" ){
+                m_all_field_functors[lev][comp] = std::make_unique<CellCenterFunctor>(warpx.get_pointer_current_fp(lev, 0), lev, m_crse_ratio);
+            } else if ( m_varnames[comp] == "jt" ){
+                m_all_field_functors[lev][comp] = std::make_unique<CellCenterFunctor>(warpx.get_pointer_current_fp(lev, 1), lev, m_crse_ratio);
+            } else if ( m_varnames[comp] == "Ar" ){
+                m_all_field_functors[lev][comp] = std::make_unique<CellCenterFunctor>(warpx.get_pointer_vector_potential_fp(lev, 0), lev, m_crse_ratio);
+            } else if ( m_varnames[comp] == "At" ){
+                m_all_field_functors[lev][comp] = std::make_unique<CellCenterFunctor>(warpx.get_pointer_vector_potential_fp(lev, 1), lev, m_crse_ratio);
```
EZoni, line 639: I'm noticing that this seems to be done for the `fp` data only. Is the `cp` completely missing from the file?

EZoni (commented):
Thank you for this PR, @RTSandberg! It looks good to me, I just left a couple of comments.

EZoni:
@dpgrote Dave, I feel it could be useful if you chime in here and check that the changes look good to you.

Python/pywarpx/picmi.py
RTSandberg: ``lower()`` changes the case of the string to all lower case.  This was done because that is how ``J`` is treated elsewhere in the input parsing.  Also, ``j`` is given in lower case in the documentation.

RTSandberg:
I just realized that I only made this change to the Full Diagnostic, not the Back-Transformed diagnostic.

Source/Diagnostics/FullDiagnostics.cpp
EZoni: @RevathiJambunathan  
It seems like we don't do the same in this file for the `_cp` data/pointers (for example for the current J) inside `InitializeFieldFunctors`: does it mean that the coarse-patch field functors are initialized in some other file? I'm asking to know if Ryan should update the RZ part of those, too, somewhere else.

Docs/source/usage/parameters.rst
```diff
@@ -2148,12 +2148,11 @@ In-situ capabilities can be used by turning on Sensei or Ascent (provided they a
     Possible scalar fields: ``part_per_cell`` ``rho`` ``phi`` ``F`` ``part_per_grid`` ``divE`` ``divB`` and ``rho_<species_name>``, where ``<species_name>`` must match the name of one of the available particle species. Note that ``phi`` will only be written out when do_electrostatic==labframe. Also, note that for ``<diag_name>.diag_type = BackTransformed``, the only scalar field currently supported is ``rho``.
     Possible vector field components in Cartesian geometry: ``Ex`` ``Ey`` ``Ez`` ``Bx`` ``By`` ``Bz`` ``jx`` ``jy`` ``jz``.
     Possible vector field components in RZ geometry: ``Er`` ``Et`` ``Ez`` ``Br`` ``Bt`` ``Bz`` ``jr`` ``jt`` ``jz``.
-    Default is ``<diag_name>.fields_to_plot = Ex Ey Ez Bx By Bz jx jy jz``,
-    unless in RZ geometry with ``<diag_name>.format == openpmd``,
-    then default is ``<diag_name>.fields_to_plot = Er Et Ez Br Bt Bz jr jt jz``.
+    The default is ``<diag_name>.fields_to_plot = Ex Ey Ez Bx By Bz jx jy jz``,
+    unless in RZ geometry,
+    then the default is ``<diag_name>.fields_to_plot = Er Et Ez Br Bt Bz jr jt jz``.
```
EZoni: This could be maybe rephrased (more clearly?) as
```suggestion
    The default is ``<diag_name>.fields_to_plot = Ex Ey Ez Bx By Bz jx jy jz`` in Cartesian geometry and ``<diag_name>.fields_to_plot = Er Et Ez Br Bt Bz jr jt jz`` in RZ geometry.
```

Examples/Tests/silver_mueller/analysis_silver_mueller_rz.py
```diff
@@ -0,0 +1,38 @@
...
+
+# Copyright 2021
+#
+# This file is part of WarpX.
+#
+# License: BSD-3-Clause-LBNL
+"""
+This tests verifies the Silver-Mueller boundary condition.
+A laser pulse is emitted and propagates towards the boundaries ; the
+test check that the reflected field at the boundary is negligible.
```
EZoni:
```suggestion
This test verifies the Silver-Mueller boundary condition.
A laser pulse is emitted and propagates towards the boundaries;
the test checks that the reflected field at the boundary is negligible.
```

Examples/Tests/silver_mueller/analysis_silver_mueller_rz.py
```diff
@@ -0,0 +1,38 @@
...
+all_data_level_0 = ds.covering_grid(level=0,left_edge=ds.domain_left_edge, dims=ds.domain_dimensions)
+Er = all_data_level_0['boxlib', 'Er'].v.squeeze()
+Et = all_data_level_0['boxlib', 'Et'].v.squeeze()
+Ez = all_data_level_0['boxlib', 'Ez'].v.squeeze()
+# The peak of the initial laser pulse is on the order of 6 V/m
+# Check that the amplitude after reflection is less than 0.01 V/m
+max_reflection_amplitude = 0.01
+assert np.all( abs(Er) < max_reflection_amplitude )
+assert np.all( abs(Et) < max_reflection_amplitude )
+assert np.all( abs(Ez) < max_reflection_amplitude )
```
EZoni: Is this analysis here significantly different from what was done in `analysis_silver_mueller.py` (the original Cartesian version)? If not, we could simply have an `if` condition in the original analysis script, to distinguish the Cartesian and RZ cases. And we could parse the output file `warpx_used_inputs` to learn which test is being run (this is already done in other analysis scripts). If the differences are minimal, I would be in favor of not duplicating the analysis script. Especially because we don't do it consistently throughout the WarpX test suite (we don't enforce the policy "one test - one input - one analysis script", it was discussed in the past, but never done).

EZoni (commented):
Thank you, @RTSandberg! I left some more comments in the review.

Also, since you now made the change for BTD, too, on the PICMI side, shouldn't we make changes similar to those in FullDiagnostics.cpp also in BTDiagnostics.cpp (for the function `BTDiagnostics::InitializeFieldFunctors`), to complete the BTD part?

RTSandberg (commented):
Clarifying suggestions were accepted

Examples/Tests/silver_mueller/analysis_silver_mueller_rz.py
RTSandberg: The analysis is not significantly different, this is a good idea.  I did not think to use ``warpx_used_inputs`` to do logic on the geometry. Thanks, @EZoni !

EZoni:
> Also, since you now made the change for BTD, too, on the PICMI side, shouldn't we make changes similar to those in FullDiagnostics.cpp also in BTDiagnostics.cpp (for the function BTDiagnostics::InitializeFieldFunctors), to complete the BTD part?

Clarified offline, this is done separately in the function `InitializeFieldFunctorsRZopenPMD` for RZ geometry.

Examples/Tests/silver_mueller/analysis_silver_mueller.py
```diff
@@ -24,15 +25,27 @@
 
 ds = yt.load( filename )
 all_data_level_0 = ds.covering_grid(level=0,left_edge=ds.domain_left_edge, dims=ds.domain_dimensions)
-Ex = all_data_level_0['boxlib', 'Ex'].v.squeeze()
-Ey = all_data_level_0['boxlib', 'Ey'].v.squeeze()
-Ez = all_data_level_0['boxlib', 'Ez'].v.squeeze()
+warpx_used_inputs = open('./warpx_used_inputs', 'r').read()
+if re.search('geometry.dims\s*=\s*RZ', warpx_used_inputs):
```
EZoni: Just a suggestion, maybe we could have an auxiliary boolean and use `re.search` only once:
```suggestion
geom_RZ = re.search('geometry.dims = RZ', warpx_used_inputs)
if geom_RZ:
```
Note that I also removed the `\s*` expression, the output in that file should always be written with one single white space, so this should work.

Examples/Tests/silver_mueller/analysis_silver_mueller.py
```diff
@@ -24,15 +25,27 @@
...
+    Ey = all_data_level_0['boxlib', 'Ey'].v.squeeze()
+    Ez = all_data_level_0['boxlib', 'Ez'].v.squeeze()
 # The peak of the initial laser pulse is on the order of 6 V/m
 # Check that the amplitude after reflection is less than 0.01 V/m
 max_reflection_amplitude = 0.01
-assert np.all( abs(Ex) < max_reflection_amplitude )
-assert np.all( abs(Ey) < max_reflection_amplitude )
-assert np.all( abs(Ez) < max_reflection_amplitude )
+
+if re.search('geometry.dims\s*=\s*RZ', warpx_used_inputs):
```
EZoni: And we could reuse the boolean here:
```suggestion
if geom_RZ:
```

EZoni (commented):
Thank you for the latest changes, @RTSandberg! I added only one small suggestion.

Examples/Tests/silver_mueller/analysis_silver_mueller.py
RTSandberg: ok that sounds good to me.  Making the change now

EZoni (approved):

PR #3665, opened 2023-02-02
Ohm's law solver (hybrid kinetic-fluid extension)

roelof-groenewald:
In this PR a different solver for the E-field is added that uses Ohm's law (with assumption of zero electron mass). The result is an algebraic expression for the E-field. See https://arxiv.org/pdf/2204.01676.pdf for details of the model.

Still to do before this PR is ready to be merged:
- [x] Add description of model to documentation
- [x] Add support for PEC boundary conditions (PML can come later)
- [x] Merge PR #3711 
- [x] More validation of implementation
- [x] Add CI tests

Some early validation studies have been performed looking at EM modes as described in [Munoz et al. (2018)](https://www.sciencedirect.com/science/article/abs/pii/S0010465517303521). The input and analysis for these tests are included in `Examples/Tests/ohm_solver/`.

spectrum_par_1d_400_substeps_1e-07_eta

spectrum_perp_1d_400_substeps_1e-07_eta


roelof-groenewald:
I performed the ion-beam instability test from Munoz et al.'s paper as well, with the WarpX results generally agreeing with theirs.

ion_beam_R_instability_resonant_eta_1e-07_substeps_400

ion_beam_R_instability_non resonant_eta_1e-07_substeps_400

roelof-groenewald:
Just posting the 2d simulation results for the perpendicularly propagating modes as well so that the plot can be included in the documentation.
spectrum_perp_2d_750_substeps_1e-05_eta

ax3l (commented):
Looked quickly at the required data containers & allocations: all runtime controlled, so perfect from that side for integration.

Source/FieldSolver/WarpXPushFieldsHybrid.cpp
```diff
@@ -0,0 +1,258 @@
...
+    // recalculated to include the resistivity before evolving B.
+
+    // Firstly J_i^{n} is calculated as the average of J_i^{n-1/2} and J_i^{n+1/2}
+    for (int lev = 0; lev <= finest_level; ++lev)
+    {
+        // Loop through the grids, and over the tiles within each grid
+#ifdef AMREX_USE_OMP
+#pragma omp parallel if (amrex::Gpu::notInLaunchRegion())
+#endif
+        for ( MFIter mfi(*current_fp[lev][0], TilingIfNotGPU()); mfi.isValid(); ++mfi ) {
```
roelof-groenewald: I'm thinking this should be doable with simpler multifab operations. All that is happening here is `current_fp_temp = 0.5 * (current_fp_temp + current_fp)`.

roelof-groenewald:
Results of Landau damping simulations. The straight lines show the theoretical damping rates.
ion_Landau_damping

roelof-groenewald:
With the fixes from #3711 and some other Ohm's law specific PEC boundary handling, I can now stably run a magnetic reconnection problem from a force-free sheet as described in [Le et al. (2016)](https://aip.scitation.org/doi/10.1063/1.4943893). In this problem the z-domain boundaries are PEC while the x-boundaries are periodic.
mag_reconnection

roelof-groenewald:
@RemiLehe @EZoni I know you guys are super busy with various proposal / milestone deadlines, so I wanted to make sure there is a long time window for reviewing this PR. I'm planning on presenting the addition of the hybrid-PIC algorithm in WarpX at the [Sherwood Fusion Theory Conference](http://www.sherwoodtheory.org/sw2023/index.php) in early May. It would be awesome if at that point the algorithm could be in the main WarpX release (i.e. the May release). Please let me know how I can help to make the review easier.

roelof-groenewald:
I realize this PR is pretty large, which makes it difficult to review. So I was thinking a good way to tackle it might be to have different people review different (mostly independent) parts of it. I compiled a checklist below, linking people to parts I think they might be interested in reviewing. If you are willing to review those parts, please do so and mark the item as "done" if you are satisfied (of course everyone is welcome to review the whole PR as well 😬 ).
- [ ] `Docs`, `Examples/Tests/ohm_solver` and `Regression` changes (@EZoni, @ax3l) 
- [ ] `Python/pywarpx` and `Source/Python` changes (@dpgrote)
- [ ] `Source/BoundaryConditions` changes (@RevathiJambunathan)
- [ ] Other changes in `Source` (@RemiLehe - since you mentioned before that you are interested in reviewing this PR)

Thank you all in advance for your time!

P.S. There are several validation tests buried in the long list of commits / comments above, meant to provide confidence in the accuracy of the implementation.

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,141 @@
...
+computational cost of resolving the electron dynamics can be avoided by modeling
+the electrons as a fluid rather than kinetic particles. The simulation resolution
+can then be set by the ion time and length scales (commonly the ion cyclotron
+period :math:`1/\Omega_i` and ion skin depth :math:`l_i`, respectively), which
+can reduce the total simulation time drastically compared to a simulation that
+has to resolve the electron Debye length and CFL-condition based on the speed
+of light.
+
+Many authors have described variations of the kinetic ion & fluid electron model,
+generally refered to as particle-fluid hybrid or just hybrid-PIC models. The implementation
```
EZoni:
```suggestion
generally referred to as particle-fluid hybrid or just hybrid-PIC models. The implementation
```

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,141 @@
...
+neutrality (:math:`n_e=n_i`) and the Maxwell-Ampere equation can be simplified by
+neglecting the displacement current term (Darwin approximation, see `Nielson and Lewis (1976) <https://www.sciencedirect.com/science/article/abs/pii/B9780124608160500154>`_),
+giving,
+
+    .. math::
+
+        \mu_0\vec{J} = \vec{\nabla}\times\vec{B}.
+
+where :math:`\vec{J} = \vec{J}_i - \vec{J}_e` is the total electrical current i.e.
+the sum of electron and ion currents. Seeing as ions are treated in the regular
```
EZoni:
```suggestion
the sum of electron and ion currents. Since ions are treated in the regular
```

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,141 @@
...
+The basic justification for the hybrid model is that the system to which it is
+applied is dominated by ion kinetics, with ions moving much slower than electrons
+and photons. In this scenario two critical approximations can be made, namely,
+neutrality (:math:`n_e=n_i`) and the Maxwell-Ampere equation can be simplified by
+neglecting the displacement current term (Darwin approximation, see `Nielson and Lewis (1976) <https://www.sciencedirect.com/science/article/abs/pii/B9780124608160500154>`_),
+giving,
+
+    .. math::
+
+        \mu_0\vec{J} = \vec{\nabla}\times\vec{B}.
```
EZoni:
```suggestion
        \mu_0\vec{J} = \vec{\nabla}\times\vec{B},
```

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,141 @@
...
+
+where :math:`\vec{J} = \vec{J}_i - \vec{J}_e` is the total electrical current i.e.
```
EZoni:
```suggestion
where :math:`\vec{J} = \vec{J}_i - \vec{J}_e` is the total electrical current, i.e.
```

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,141 @@
...
+the sum of electron and ion currents. Seeing as ions are treated in the regular
+PIC manner, the ion current, :math:`\vec{J}_i`, is known during a simulation. Therefore,
+given the magnetic field, the electron current can be calculated.
+
+If we now furher assume electrons are inertialess, the electron momentum
```
EZoni:
```suggestion
If we now further assume electrons are inertialess, the electron momentum
```

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,141 @@
...
+
+.. note::
+
+    Various verification tests of the hybrid model implementation can be found in
+    the :ref:`examples section <examples-hybrid-model>`.
+
+The kinetic-fluid hybrid extension mostly uses the same routines as the standard
+PIC algorithm with the only exception that the E-field is calculated from the
+above equation rather than it being integrated from a differential equation. The
+function `WarpX::HybridPICEvolveFields()` handles the logic to update the E&M fields
```
EZoni: Maybe, here and below, we can render these code function names with "``" instead of italic?
```suggestion
function ``WarpX::HybridPICEvolveFields()`` handles the logic to update the E&M fields
```

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,141 @@
...
+when the "hybridPIC" model is used. This function is executed after particle pushing
+and deposition (charge and current density) has been completed. Therefore, based
+on the usual time-staggering in the PIC algorithm, when `HybridPICEvolveFields()` is called
```
EZoni:
```suggestion
on the usual time-staggering in the PIC algorithm, when ``HybridPICEvolveFields()`` is called
```

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,141 @@
...
+at timestep :math:`t=n`, the quantities :math:`\rho^n`, :math:`\rho^{n+1}`, :math:`\vec{J}_i^{n-1/2}`
```
EZoni:
```suggestion
at timestep :math:`t=t_n`, the quantities :math:`\rho^n`, :math:`\rho^{n+1}`, :math:`\vec{J}_i^{n-1/2}`
```

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,141 @@
...
+
+Field update
+^^^^^^^^^^^^
+
+The field update is done in three steps as described below.
+
+First half step
+"""""""""""""""
+
+Firstly the E-field at :math:`t=n` is calculated for which the current density needs to
```
EZoni:
```suggestion
Firstly the E-field at :math:`t=t_n` is calculated for which the current density needs to
```

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,141 @@
...
+be interpolated to the correct time, using :math:`\vec{J}_i^n = 1/2(\vec{J}_i^{n-1/2}+ \vec{J}_i^{n+1/2})`.
+The electron pressure is simply calculated using :math:`\rho^n` and the B-field is also already
+known at the correct time since it was calculated for :math:`t=n` at the end of the last step.
```
EZoni:
```suggestion
known at the correct time since it was calculated for :math:`t=t_n` at the end of the last step.
```

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,141 @@
...
+Once :math:`\vec{E}^n` is calculated, it is used to push :math:`\vec{B}^n` forward in time
+(using Faraday's law i.e. the same as in the regular PIC routine with `WarpX::EvolveB()`)
```
EZoni:
```suggestion
(using Faraday's law, i.e. the same as in the regular PIC routine with ``WarpX::EvolveB()``)
```

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,141 @@
...
+using the known fields :math:`\vec{B}^{n+1/2}`, :math:`\vec{J}_i^{n+1/2}` and
+interpolated charge density :math:`\rho^{n+1/2}=1/2(\rho^n+\rho^{n+1})` (which is
+also used to calculate the electron pressure). Similarly as before, the B-field
+is then pushed forward to get :math:`\vec{B}^{n+1}` using the newly calculated
+:math:`\vec{E}^{n+1/2}` field.
+
+Extrapolation step
+""""""""""""""""""
+
+Obtaining the E-field at timestep :math:`t=n+1` is a well documented issue for
```
EZoni:
```suggestion
Obtaining the E-field at timestep :math:`t=t_{n+1}` is a well documented issue for
```

Docs/source/usage/examples.rst
```diff
@@ -121,3 +121,116 @@ An example of using Python to access the simulation charge density, solve the Po
 An example of initializing the fields by accessing their data through Python, advancing the simulation for a chosen number of time steps, and plotting the fields again through Python. The simulation runs with 128 regular cells, 8 guard cells, and 10 PML cells, in each direction. Moreover, it uses div(E) and div(B) cleaning both in the regular grid and in the PML and initializes all available electromagnetic fields (E,B,F,G) identically.
 
 * :download:`Unit pulse with PML <../../../Examples/Tests/python_wrappers/PICMI_inputs_2d.py>`
+
+.. _examples-hybrid-model:
+
+Kinetic-fluid hybrid model
+--------------------------
+
+Several examples / benchmarks of the kinetic-fluid hybrid model are shown below. The first few examples are replications
```
EZoni:
```suggestion
Several examples and benchmarks of the kinetic-fluid hybrid model are shown below. The first few examples are replications
```

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,141 @@
...
+the electrons as a fluid rather than kinetic particles. The simulation resolution
+can then be set by the ion time and length scales (commonly the ion cyclotron
+period :math:`1/\Omega_i` and ion skin depth :math:`l_i`, respectively), which
+can reduce the total simulation time drastically compared to a simulation that
+has to resolve the electron Debye length and CFL-condition based on the speed
+of light.
+
+Many authors have described variations of the kinetic ion & fluid electron model,
+generally refered to as particle-fluid hybrid or just hybrid-PIC models. The implementation
+in WarpX follows the outline from `Winske et al. (2022) <https://arxiv.org/abs/2204.01676>`_.
```
EZoni: Ideally, it would be great if you could use BibTeX, as started in #3787, when adding references in the theory section (in this file and all other relevant files). It should be possible to add the reference entries to the existing `refs.bib` file and just cite them wherever needed.

Docs/source/usage/examples.rst
```diff
@@ -121,3 +121,116 @@ An example of using Python to access the simulation charge density, solve the Po
...
+
+Ion Landau damping
+^^^^^^^^^^^^^^^^^^
+
+Landau damping is a well known process in which electrostatic (acoustic) waves
+are damped by transferring energy to particles satisfying a resonance condition.
+The process can be simulated by seeding a plasma with a specific acoustic mode
+(density perturbation) and tracking the strength of the mode as a function of
+time. The figure below shows a set of such simulations with parameters matching
+those described in section 4.5 of Munoz et al. (2018). The straight lines show
```
EZoni: Link to reference is missing here.

Docs/source/usage/parameters.rst
```diff
@@ -2110,6 +2110,29 @@ Additional parameters
...
+
+* ``hybrid_pic_model.plasma_resistivity(rho)`` (`float` or `str`) optional (default ``0``)
+     If ``algo.maxwell_solver`` is set to ``hybrid``, this sets the plasma resistivity in :math:`\Omega m`.
+
+* ``hybrid_pic_model.n_floor`` (`float`) optional (default ``1``)
+     If ``algo.maxwell_solver`` is set to ``hybrid``, this sets the plasma density floor, in :math:`m^{-3}`, which is useful since the generalized Ohm's law used to calculate the E-field includes a :math:`1/n` term.
+
+* ``hybrid_pic_model.substeps`` (`int`) optional (default ``100``)
+     If ``algo.maxwell_solver`` is set to ``hybrid``, this sets the number of sub-steps to take during the B-field update.
+
```
EZoni: Would it make sense to group these parameters under a specific subsection, rather than the current one, which is called "Additional parameters"?
```
Additional parameters
^^^^^^^^^^^^^^^^^^^^^
```
Maybe something like this, or similar?
```
Maxwell solver: kinetic-fluid hybrid
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

EZoni (commented):
Thank you for the PR, @roelof-groenewald! And thank you for suggesting a way to split the review among different reviewers. I start by leaving a few minor comments on the docs part, and I will then proceed to the Python examples/tests part.

Docs/source/theory/kinetic_fluid_hybrid_model.rst
roelof-groenewald: This is awesome! Thanks for pointing it out. Quick question though: you mention switching to bibtex for references in the `theory` section. Should I not use bibtex for the references in the `Examples` section?

Docs/source/usage/parameters.rst
roelof-groenewald: Sounds good! I'll move it to be below the `Maxwell solver: macroscopic media` section.

Docs/source/theory/kinetic_fluid_hybrid_model.rst
roelof-groenewald: I went ahead and changed the references in the `Examples` section to also use bibtex. We can undo if you don't like it.

EZoni: Thanks, I think it's good if we use it everywhere. I just mentioned the theory section because that's where references are typically used the most. Thank you for propagating the change to other sections of the docs, too.

Source/WarpX.H
```diff
@@ -691,6 +728,16 @@ public:
     void ApplyEfieldBoundary (const int lev, PatchType patch_type);
     void ApplyBfieldBoundary (const int lev, PatchType patch_type, DtType dt_type);
 
+    /**
+     * \brief When the Ohm's law solver is used, the elctron pressure values
```
```suggestion
     * \brief When the Ohm's law solver is used, the electron pressure values
```

Source/WarpX.H
```diff
@@ -604,6 +608,39 @@ public:
     void MacroscopicEvolveE (int lev, amrex::Real dt);
     void MacroscopicEvolveE (int lev, PatchType patch_type, amrex::Real dt);
 
+    /**
+     * \brief
+     * Function to update the E-field using Ohm's law (hybrid-PIC model).
+     */
+    void HybridPICSolveE (                               DtType dt_type);
+    void HybridPICSolveE (int lev,                       DtType dt_type);
+    void HybridPICSolveE (int lev, PatchType patch_type, DtType dt_type);
```
```suggestion
    void HybridPICSolveE (const int lev,                       DtType dt_type);
    void HybridPICSolveE (const int lev, PatchType patch_type, DtType dt_type);
```

Source/WarpX.H
```diff
@@ -604,6 +608,39 @@ public:
...
+
+    /**
+     * \brief
+     * Function to calculate the total current based on Ampere's law while
+     * neglecting displacement current (J = curl x B). Used in the Ohm's law
+     * solver (kinetic-fluid hybrid model).
+     */
+    void CalculateCurrentAmpere ();
+    void CalculateCurrentAmpere (int lev);
```
```suggestion
    void CalculateCurrentAmpere (const int lev);
```

Source/Initialization/WarpXInitData.cpp
```diff
@@ -729,6 +737,11 @@ WarpX::InitLevelData (int lev, Real /*time*/)
 
     for (int i = 0; i < 3; ++i) {
 
+        if (WarpX::electromagnetic_solver_id == ElectromagneticSolverAlgo::HybridPIC)
+        {
+            current_fp_temp[lev][i]->setVal(0.0);
+        }
+
```
EZoni: I think you should be able to skip this here, since in WarpX.cpp you allocated `current_fp_temp` using the function `AllocInitMultiFab` and passing `0.0_rt` as the last argument, which should be precisely the initialization value across all levels and dimensions.
```suggestion
```

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -0,0 +1,254 @@
...
+    HybridPICSolveE(lev, PatchType::fine, a_dt_type);
+    if (lev > 0)
+    {
+        amrex::Abort(Utils::TextMsg::Err(
+        "HybridPICSolveE: Only one level implemented for hybrid-PIC solver."));
+        HybridPICSolveE(lev, PatchType::coarse, a_dt_type);
+    }
+}
+
+void WarpX::HybridPICSolveE (int lev, PatchType patch_type, DtType a_dt_type)
```
EZoni:
```suggestion
void WarpX::HybridPICSolveE (const int lev, PatchType patch_type, DtType a_dt_type)
```

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -0,0 +1,254 @@
...
+void WarpX::HybridPICSolveE (DtType a_dt_type)
+{
+    for (int lev = 0; lev <= finest_level; ++lev)
+    {
+        HybridPICSolveE(lev, a_dt_type);
+    }
+    FillBoundaryE(guard_cells.ng_FieldSolver, WarpX::sync_nodal_points);
+}
+
+void WarpX::HybridPICSolveE (int lev, DtType a_dt_type)
```
EZoni:
```suggestion
void WarpX::HybridPICSolveE (const int lev, DtType a_dt_type)
```

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -0,0 +1,254 @@
...
+{
+    WARPX_PROFILE("WarpX::HybridPICSolveE()");
+
+    HybridPICSolveE(lev, PatchType::fine, a_dt_type);
+    if (lev > 0)
+    {
+        amrex::Abort(Utils::TextMsg::Err(
+        "HybridPICSolveE: Only one level implemented for hybrid-PIC solver."));
+        HybridPICSolveE(lev, PatchType::coarse, a_dt_type);
```
EZoni: Can we remove this call, since the code aborts anyways?
```suggestion
```
Or was it meant to be a hint on how to implement this in the future?

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+void WarpX::CalculateCurrentAmpere ()
+{
+    for (int lev = 0; lev <= finest_level; ++lev)
+    {
+        CalculateCurrentAmpere(lev);
+    }
+}
+
+void WarpX::CalculateCurrentAmpere (int lev)
```
EZoni:
```suggestion
void WarpX::CalculateCurrentAmpere (const int lev)
```

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -0,0 +1,254 @@
...
+    // Copy the rho^{n+1} values to rho_fp_temp and the J_i^{n+1/2} values to
+    // current_fp_temp since at the next step those values will be needed as
+    // rho^{n} and J_i^{n-1/2}.
+    for (int lev = 0; lev <= finest_level; ++lev)
+    {
+        MultiFab::Copy(*rho_fp_temp[lev], *rho_fp[lev],
+                        0, 0, 1, rho_fp_temp[lev]->nGrowVect());
+        for (int idim = 0; idim < 3; ++idim) {
+            MultiFab::Copy(*current_fp_temp[lev][idim], *current_fp[lev][idim],
+                           0, 0, 1, current_fp_temp[lev][idim]->nGrowVect());
```
EZoni, line 166: I always find it useful to add a brief inline comment explaining what the various "0" and "1" represent in these AMReX function calls (it would be nice if C++ had a Python-like syntax such as `f(arg1=val1, arg2=val2, ...)`). I know we don't do it consistently throughout the code, but if you too think that it's a good idea, this could be a good place to start. If you do so, it could be good to do it consistently at least in this file.

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+void WarpX::HybridPICEvolveFields ()
+{
+    // The below deposition is hard coded for a single level simulation
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+        finest_level == 0,
+        "Ohm's law E-solve only works with a single level.");
+
+    // Perform charge deposition in component 0 of rho_fp
+    mypc->DepositCharge(rho_fp, 0._rt);
```
EZoni: I wonder if in this comment we should also make it clear that, since we pass `0._rt` as the second argument, we are depositing the charge density at the time corresponding to the current particle positions, which I guess is $n+1$ (given that the shift `-0.5_rt*dt[0]` below corresponds to $n+1/2$)?

EZoni (commented):
Hi @roelof-groenewald, I'm posting a few simple comments. I'm slowly going through various parts of the PR, but I think it's better if I post my comments in relatively small chunks, otherwise it might be more difficult to keep track of them on GitHub.

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
roelof-groenewald: I did leave it with the intent for it to be a hint as to what the function call would look like if mesh refinement is implemented for the hybrid solver at some point. But I agree we can actually remove it since it isn't a very profound hint.

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
roelof-groenewald: Thanks for continuing the review!! I agree that such comments would be helpful for the next person trying to read the code. I'll add the comments (also for the next point) and re-ping you for a review when it is done.

Python/pywarpx/picmi.py
```diff
@@ -1006,6 +1007,59 @@ def initialize_inputs(self):
         pywarpx.warpx.pml_has_particles = self.pml_has_particles
         pywarpx.warpx.do_pml_j_damping = self.do_pml_j_damping
 
+
+class HybridPICSolver():
```
dpgrote: This should inherit from `picmistandard.base._ClassWithInit`.

Source/Evolve/WarpXComputeDt.cpp
```diff
@@ -33,8 +33,12 @@ void
 WarpX::ComputeDt ()
 {
     // Handle cases where the timestep is not limited by the speed of light
-    if (electromagnetic_solver_id == ElectromagneticSolverAlgo::None) {
-        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(m_const_dt.has_value(), "warpx.const_dt must be specified with the electrostatic solver.");
+    if (electromagnetic_solver_id == ElectromagneticSolverAlgo::None ||
```
dpgrote, line 36: Can this be separated out into the different cases so that the assert error message is specific to the setup?

roelof-groenewald: Yeap, can do!

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -136,7 +136,32 @@ WarpX::Evolve (int numsteps)
...
+
+            // The hybrid-PIC algorithm uses the charge and current density
+            // from both the current and previous step when updating the
+            // fields, so we deposit the ion charge and current now in the
+            // temp multifab locations.
+            if (electromagnetic_solver_id == ElectromagneticSolverAlgo::HybridPIC)
+            {
+                mypc->DepositCharge(rho_fp_temp, 0._rt);
+                mypc->DepositCurrent(current_fp_temp, dt[0], 0._rt);
+                SyncCurrent(current_fp_temp, current_cp);
```
dpgrote: Is there a reason to not use `SyncRho` here instead of calling `AddRhoFromFineLevelandSumBoundary` directly?

Also, perhaps add new routines like `void WarpX::ApplyRhofieldBoundary (PatchType patch_type)` that have the loop over levels, as is done elsewhere.

dpgrote:
This looks great overall! I have one general comment.

The code in the Evolve allows particles to move more than one cell per time step, correct? Since there is a current deposition, perhaps add a check that `Esirkepov` is not turned on since that assumes that particles move no more than one cell per step.

Source/Evolve/WarpXEvolve.cpp
roelof-groenewald: If there was a reason for not calling `SyncRho` directly, I don't remember what it was - maybe I only saw the `SyncRho()` function which always operates on `rho_fp` and not the version below it that takes multifabs as input - but I will change that to be more consistent with the current density treatment. Thanks for pointing it out!

I think implementing `ApplyRhofieldBoundary (PatchType patch_type)` functions that will include the loop over levels can maybe be done in a separate PR since for consistency it should be done for E, B, J and rho.

roelof-groenewald:
Thanks a lot for the review @dpgrote! 

> The code in the Evolve allows particles to move more than one cell per time step, correct? Since there is a current deposition, perhaps add a check that `Esirkepov` is not turned on since that assumes that particles move no more than one cell per step.

I struggled quite a bit with the decision on whether to allow particles to move more than one cell per time step and actually the initial implementation did not allow it. Since the model is electromagnetic particles really shouldn't move multiple cells per time step, but the way in which this model is often run (trying to stretch the timescales which can be reached in a feasible time) creates instances where a very small fraction of particles sometimes achieve speeds that make them jump over a couple cells. That usually doesn't impact the overall simulation accuracy much since the offending particles are very few (otherwise the simulation is clearly under resolved). But the optimized deposition logic in `WarpX::PushParticlesandDepose` cannot handle even a single particle having moved too far - it segfaults in that case. So really the compromise was that using the deposition logic that allows particles to move multiple cells per step ensures that the simulation will not crash due to a single particle going too fast. 
We normally want to use the Esirkepov current deposition logic so we do constrain the time step so that the overwhelming majority of particles move less than cell width per step.

Source/Evolve/WarpXEvolve.cpp
dpgrote: I add these routines for E and B in PR #3842. This PR can add them for J and rho. But a separate PR is fine also.

dpgrote:
With Esirkepov, I think you will get a segfault if a particle moves more than one cell. The shape factor arrays only allow for movement of one cell. Beyond that, the deposition loops will go out of bounds on those arrays.

Source/Evolve/WarpXEvolve.cpp
roelof-groenewald: Sounds good. I'll keep an eye on your PR and submit a follow-up for J and rho once it is merged.

roelof-groenewald:
Ah, I see. In the testing of the magnetic reconnection problem at lower resolution I commonly ran into problems at https://github.com/ECP-WarpX/WarpX/blob/453c6326904a7ba27eb49e1dcf35e7eeafa72b60/Source/ablastr/particles/DepositCharge.H#L133, which were cleared up by doing the scraping of particles outside the simulation domain (and global redistribute) before depositing the charge. I guess since I don't run into segfaults from the Esirkepov deposition it means that my problem was with some particles flying out of the domain too far, instead of traveling over multiple cells inside the domain.
I will add a warning at initialization if Esirkepov is used telling the user that they should be careful with the choice of dt since a segfault could occur, that way we don't block Esirkepov from being used with the hybrid-PIC solver entirely.

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,144 @@
+.. _theory-kinetic-fluid-hybrid-model:
+
+Kinetic-fluid Hybrid Model
+==========================
+
+Many problems in plasma physics fall in a class where electron kinetics do not
```
RemiLehe:
```suggestion
Many problems in plasma physics fall in a class where both electron kinetics and electromagnetic waves do not
```

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,144 @@
+.. _theory-kinetic-fluid-hybrid-model:
+
+Kinetic-fluid Hybrid Model
+==========================
+
```
RemiLehe, line 5: The sentence "For these kinds of problems the computational cost of resolving the electron dynamics can be avoided by modeling the electrons as a fluid rather than kinetic particles." only highlights the fluid approximation for electrons, but in my understanding the hybrid PIC models makes more assumptions:
- The electrons are assumed to be a neutralizing fluid (:math:`n_e=n_i`), which effectively suppresses Langmuir waves.
- Instead of updating the electric field with the Maxwell-Faraday law, the electric field is instead computed from a generalized Ohm's. This effectively suppresses light waves.

This points are made further down (using equations), but I think that it is worth making them early on.

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,144 @@
...
+This description follows mostly from that reference.
+
+Model
+-----
+
+The basic justification for the hybrid model is that the system to which it is
+applied is dominated by ion kinetics, with ions moving much slower than electrons
+and photons. In this scenario two critical approximations can be made, namely,
+neutrality (:math:`n_e=n_i`) and the Maxwell-Ampere equation can be simplified by
+neglecting the displacement current term (Darwin approximation :cite:p:`c-NIELSON1976`),
```
RemiLehe:
```suggestion
neglecting the displacement current term,
```
In my understanding, the Darwin approximation does not neglect the displacement current term entirely, but instead it only neglects its transverse part. (Which effectively discards light waves, but retains the physics of the space charge field of a relativistic particle.)

It might be important to make this distinction since we might implement a Darwin solver later on in the code.

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,144 @@
...
+
+If we now further assume electrons are inertialess, the electron momentum
+equation yields,
+
+    .. math::
+
+        \frac{d(n_em_e\vec{V}_e)}{dt} = 0 = -en_e\vec{E}-\vec{J}_e\times\vec{B}-\nabla\cdot\vec{P}_e+en_e\vec{\eta}\cdot\vec{J},
+
+where :math:`\vec{V_e}=\vec{J}_e/(en_e)`, :math:`\vec{P}_e` is the electron pressure
+tensor and :math:`\vec{\eta}` is the resistivity. An expression for the electric field
```
RemiLehe:
```suggestion
tensor and :math:`\vec{\eta}` is the resistivity tensor. An expression for the electric field
```

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,144 @@
...
+
+Implementation details
+----------------------
+
+.. note::
+
+    Various verification tests of the hybrid model implementation can be found in
+    the :ref:`examples section <examples-hybrid-model>`.
+
+The kinetic-fluid hybrid extension mostly uses the same routines as the standard
```
RemiLehe:
```suggestion
The kinetic-fluid hybrid extension mostly uses the same routines as the standard electromagnetic
```

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,144 @@
...
+PIC algorithm with the only exception that the E-field is calculated from the
+above equation rather than it being integrated from a differential equation. The
```
RemiLehe:
```suggestion
above equation rather than it being updated from the full Maxwell-Ampere equation. The
```

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,144 @@
...
+
+First half step
+"""""""""""""""
+
+Firstly the E-field at :math:`t=t_n` is calculated for which the current density needs to
+be interpolated to the correct time, using :math:`\vec{J}_i^n = 1/2(\vec{J}_i^{n-1/2}+ \vec{J}_i^{n+1/2})`.
+The electron pressure is simply calculated using :math:`\rho^n` and the B-field is also already
+known at the correct time since it was calculated for :math:`t=t_n` at the end of the last step.
+Once :math:`\vec{E}^n` is calculated, it is used to push :math:`\vec{B}^n` forward in time
+(using Faraday's law, i.e. the same as in the regular PIC routine with ``WarpX::EvolveB()``)
```
RemiLehe: I am being a bit pedantic here (sorry!), but technically the Faraday's law is integrated form of the Maxwell-Faraday equation.
```suggestion
(using the Maxwell-Faraday equation, i.e. the same as in the regular PIC routine with ``WarpX::EvolveB()``)
```

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,144 @@
...
+
+Electron pressure
+^^^^^^^^^^^^^^^^^
+
+The electron pressure is assumed to be a scalar quantity and calculated using the given
+input parameters, :math:`T_e`, :math:`n_0` and :math:`\gamma` using
+
+    .. math::
+
+        P_e = n_0T_e\left( \frac{n_e}{n_0} \right)^\gamma.
```
RemiLehe: Would it make sense to denote the temperature as `T_e0` to emphasize that this is the initial temperature (just like `n_0` is the initial density.
```suggestion
input parameters, :math:`T_{e0}`, :math:`n_0` and :math:`\gamma` using

    .. math::

        P_e = n_0T_{e0}\left( \frac{n_e}{n_0} \right)^\gamma.
```

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -0,0 +1,144 @@
...
+
+        \vec{J}_i^{n+1} = \frac{3}{2}\vec{J}_i^{n+1/2} - \frac{1}{2}\vec{J}_i^{n-1/2}.
+
+With this extrapolation all fields required to calculate :math:`\vec{E}^{n+1}`
+are known and the simulation can proceed.
+
+Sub-stepping
+^^^^^^^^^^^^
+
+It is also well known that hybrid PIC routines require the B-field to be
```
RemiLehe, line 123: Out of curiosity: what happens if we do not sub-step B? Does the simulation become numerically unstable?

Docs/source/usage/examples.rst
```diff
@@ -121,3 +121,119 @@ An example of using Python to access the simulation charge density, solve the Po
...
 * :download:`Unit pulse with PML <../../../Examples/Tests/python_wrappers/PICMI_inputs_2d.py>`
+
+.. _examples-hybrid-model:
+
+Kinetic-fluid hybrid model
+--------------------------
+
+Several examples and benchmarks of the kinetic-fluid hybrid model are shown below. The first few examples are replications
+of the verification tests described in :cite:t:`ex-MUNOZ2018`.
+The hybrid-PIC model was added to WarpX in April 2023 (`PR #3665 <https://github.com/ECP-WarpX/WarpX/pull/3665>`_) - the below figures
```
RemiLehe:
```suggestion
The hybrid-PIC model was added to WarpX in `PR #3665 <https://github.com/ECP-WarpX/WarpX/pull/3665>`_ - the figures below
```

Examples/Tests/ohm_solver/EM_modes/PICMI_inputs.py
```diff
@@ -0,0 +1,365 @@
+#!/usr/bin/env python3
```
RemiLehe: I think that `Tests` folder has a flat directory structure. 
For the sake of consistency, would it be okay to move the subfolders of `ohm_solver` to folders directly in `Tests` (for instance: `ohm_solver_em_modes`)

Examples/Tests/ohm_solver/EM_modes/analysis.py
```diff
@@ -0,0 +1,188 @@
...
+from pywarpx import picmi
+
+constants = picmi.constants
+
+matplotlib.rcParams.update({'font.size': 20})
+
+B_dir = 'z'
+
+# load simulation parameters
+with open(f'sim_parameters_{B_dir}.dpkl', 'rb') as f:
```
RemiLehe: The fact that this part of the script is hard-coded for `B_dir = 'z'` might be a bit confusing.
Could we simply do
```suggestion
# load simulation parameters
with open(f'sim_parameters.dpkl', 'rb') as f:
```
and modify the `PICMI_inputs.py` so that the name of the saved file is always `sim_parameters.dpkl`?

Examples/Tests/ohm_solver/EM_modes/analysis.py
```diff
@@ -0,0 +1,188 @@
...
+    sim = dill.load(f)
+
+if sim.test:
```
RemiLehe: In our analysis scripts, we usually put this part at the very end. 
And we also usually include a physics test, in the form of an `assert` statement that automatically compares some simulation result with the theory (up to some precision). In the case of the test here, it is not immediately clear to me how we could check that the code predicts the correct modes from a simple `assert` statement, but maybe we could discuss this offline.

Examples/Tests/ohm_solver/EM_modes/PICMI_inputs.py
```diff
@@ -0,0 +1,365 @@
...
+#
+# --- Test script for the kinetic-fluid hybrid model in WarpX wherein ions are
+# --- treated as kinetic particles and electrons as an isothermal, inertialess
+# --- background fluid. The script is set up to produce either parallel or
+# --- perpendicular (Bernstein) EM modes and can be run in 1d, 2d or 3d
+# --- Cartesian geometries. As a CI test only a small number of steps are taken
+# --- using the 1d version.
```
RemiLehe: For consistency with the other tests, could you also add the Munoz reference here?

Examples/Tests/ohm_solver/ion_Landau_damping/PICMI_inputs.py
```diff
@@ -0,0 +1,344 @@
...
+    # Numerical parameters
+    NPPC        = [8192, 4096, 1024] # Seed number of particles per cell
+    DZ          = 1.0 / 6.0 # Cell size (ion skin depths)
+    DT          = 1e-3 # Time step (ion cyclotron periods)
+
+    # density perturbation strength
+    epsilon = 0.03
+
+    # Plasma resistivity - used to dampen the mode excitation
+    eta = 1e-7
```
RemiLehe: Out of curiosity: is this technically needed here? What happens if we set this to 0.
(Also: do I understand correctly that the theoretical formula that we compare the simulation with assumes 0 resistivity?)

Examples/Tests/ohm_solver/ion_Landau_damping/analysis.py
```diff
@@ -0,0 +1,107 @@
...
+
+constants = picmi.constants
+
+matplotlib.rcParams.update({'font.size': 20})
+
+# load simulation parameters
+with open(f'sim_parameters.dpkl', 'rb') as f:
+    sim = dill.load(f)
+
+if sim.test:
```
RemiLehe: Same comment as for the other script: maybe we could put this at the end of the script and at the same time add a physics test that checks the damping rate with an `assert` statement.

RemiLehe (changes requested):
Thanks a lot for this very thorough PR!
I started reviewing it and had a few comments on the documentation and tests. Instead of waiting until I review the whole thing, I thought I would already post the comments from the first part of the review.

Docs/source/theory/kinetic_fluid_hybrid_model.rst
roelof-groenewald: Sounds good. I changed the intro paragraph a bit to include those points. Let me know if you think we should make further refinements.

Docs/source/theory/kinetic_fluid_hybrid_model.rst
roelof-groenewald: I agree we should be careful with using "Darwin approximation", so I removed that part. I'd like to keep the reference though as it deals with particle codes in the non-radiative limit which is appropriate for this.

Docs/source/theory/kinetic_fluid_hybrid_model.rst
roelof-groenewald: Yes, exactly. In my experience the field values grow until a particle is kicked out of bounds.

Examples/Tests/ohm_solver/ion_Landau_damping/PICMI_inputs.py
roelof-groenewald: It is needed just for numerical stability. Inclusion of the resistivity doesn't shift the dispersion though, it only decreases the intensity.

roelof-groenewald:
@RemiLehe Following up on our discussion from the other day. The ion-beam R instability example runs to ~25 ion cyclotron periods in 2.5 minutes on my machine using 2 processors (using 1d). This gives the following mode growth curve:
ion_beam_R_instability_resonant_eta_1e-07_substeps_100_low_modes
(where the slopes of the straight lines are set by the theoretical growth rates).

Do you think 2.5 minutes is an okay duration for a CI test? I think at least the m=4 and m=5 modes can already be used in this case for a physics-based test.

roelof-groenewald:
The physics test described above, for a reason a don't know, gives different results on the Azure system than on my Mac. Luckily, running the test on a Linux system gave the same result as the Azure test. The RMS error to the theoretical growth rates on these systems are higher than on the Mac (plot below). 
ion_beam_R_instability_resonant_eta_1e-07_substeps_100_low_modes

But the full test still produces the expected result:
ion_beam_R_instability_resonant_eta_1e-07_substeps_400_low_modes

Source/FieldSolver/FiniteDifferenceSolver/FiniteDifferenceSolver.H
```diff
@@ -129,6 +131,34 @@ class FiniteDifferenceSolver
...
+        /**
+          * \brief E-update in the hybrid PIC algorithm as described in
+          * Winske et al. (2003) Eq. 10.
+          * https://link.springer.com/chapter/10.1007/3-540-36530-3_8
+          *
+          * \param[out] Efield  vector of electric field MultiFabs updated at a given level
+          * \param[in] Jfield   vector of total current MultiFabs at a given level
+          * \param[in] Jifield  vector of ion current density MultiFabs at a given level
+          * \param[in] Bfield   vector of magnetic field MultiFabs at a given level
+          * \param[in] rhofield scalar ion chrge density Multifab at a given level
```
RemiLehe:
```suggestion
          * \param[in] rhofield scalar ion charge density Multifab at a given level
```

Source/FieldSolver/FiniteDifferenceSolver/FiniteDifferenceSolver.H
```diff
@@ -129,6 +131,34 @@ class FiniteDifferenceSolver
...
+                      std::array< std::unique_ptr<amrex::MultiFab>, 3>& Jfield,
+                      std::array< std::unique_ptr<amrex::MultiFab>, 3 > const& Jifield,
+                      std::array< std::unique_ptr<amrex::MultiFab>, 3> const& Bfield,
+                      std::unique_ptr<amrex::MultiFab> const& rhofield,
+                      std::unique_ptr<amrex::MultiFab> const& Pefield,
+                      std::array< std::unique_ptr<amrex::MultiFab>, 3 > const& edge_lengths,
+                      int lev, std::unique_ptr<HybridPICModel> const& hybrid_pic_model,
+                      DtType a_dt_type );
+
+        void CalculateCurrentAmpere (
```
RemiLehe, line 168: Could you add also a docstring here?

Source/FieldSolver/FiniteDifferenceSolver/FiniteDifferenceSolver.H
```diff
@@ -184,6 +214,27 @@ class FiniteDifferenceSolver
...
+            std::array< std::unique_ptr<amrex::MultiFab>, 3> const& Jifield,
+            std::array< std::unique_ptr<amrex::MultiFab>, 3> const& Bfield,
+            std::unique_ptr<amrex::MultiFab> const& rhofield,
+            std::unique_ptr<amrex::MultiFab> const& Pefield,
+            std::array< std::unique_ptr<amrex::MultiFab>, 3 > const& edge_lengths,
+            int lev, std::unique_ptr<HybridPICModel> const& hybrid_pic_model,
+            DtType a_dt_type );
+
+        template<typename T_Algo>
+        void CalculateTotalCurrentCylindrical (
```
RemiLehe: For consistency?
```suggestion
        void CalculateCurrentAmpereCylindrical (
```

Source/FieldSolver/FiniteDifferenceSolver/FiniteDifferenceSolver.H
```diff
@@ -267,6 +318,26 @@ class FiniteDifferenceSolver
...
+            std::array< std::unique_ptr<amrex::MultiFab>, 3> const& Jifield,
+            std::array< std::unique_ptr<amrex::MultiFab>, 3> const& Bfield,
+            std::unique_ptr<amrex::MultiFab> const& rhofield,
+            std::unique_ptr<amrex::MultiFab> const& Pefield,
+            std::array< std::unique_ptr<amrex::MultiFab>, 3 > const& edge_lengths,
+            int lev, std::unique_ptr<HybridPICModel> const& hybrid_pic_model,
+            DtType a_dt_type );
+
+        template<typename T_Algo>
+        void CalculateTotalCurrentCartesian (
```
RemiLehe:
```suggestion
        void CalculateCurrentAmpereCartesian (
```

Source/WarpX.H
```diff
@@ -471,11 +473,13 @@ public:
     amrex::MultiFab * get_pointer_Bfield_fp  (int lev, int direction) const { return Bfield_fp[lev][direction].get(); }
     amrex::MultiFab * get_pointer_current_fp  (int lev, int direction) const { return current_fp[lev][direction].get(); }
     amrex::MultiFab * get_pointer_current_fp_nodal  (int lev, int direction) const { return current_fp_nodal[lev][direction].get(); }
+    amrex::MultiFab * get_pointer_current_fp_ampere  (int lev, int direction) const { return current_fp_ampere[lev][direction].get(); }
     amrex::MultiFab * get_pointer_rho_fp  (int lev) const { return rho_fp[lev].get(); }
     amrex::MultiFab * get_pointer_F_fp  (int lev) const { return F_fp[lev].get(); }
     amrex::MultiFab * get_pointer_G_fp  (int lev) const { return G_fp[lev].get(); }
     amrex::MultiFab * get_pointer_phi_fp  (int lev) const { return phi_fp[lev].get(); }
     amrex::MultiFab * get_pointer_vector_potential_fp  (int lev, int direction) const { return vector_potential_fp_nodal[lev][direction].get(); }
+    amrex::MultiFab * get_pointer_electron_pressure_fp  (int lev) const { return electron_pressure_fp[lev].get(); }
```
RemiLehe: It seems that this function is not currently being used ; should we remove it?

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -0,0 +1,252 @@
...
+    mypc->DepositCurrent(current_fp, dt[0], -0.5_rt * dt[0]);
+
+    // Synchronize J and rho:
+    // filter (if used), exchange guard cells, interpolate across MR levels
+    // and apply boundary conditions
+    SyncCurrentAndRho();
+
+    // SyncCurrent does not include a call to FillBoundary, but it is needed
+    // for the hybrid-PIC solver since current values are interpolated to
+    // a nodal grid
```
RemiLehe, line 40: `SyncCurrent` does call `SumBoundaryJ`. Even though `SumBoundaryJ` currently does not fill guard cells for the FDTD solvers, we could potentially change this: https://github.com/ECP-WarpX/WarpX/blob/development/Source/Parallelization/WarpXSumGuardCells.H#L37

RemiLehe: Although this would then also sum guard cells for `rho`, which is maybe not desired...

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -0,0 +1,511 @@
...
+void FiniteDifferenceSolver::CalculateCurrentAmpere (
+    std::array< std::unique_ptr<amrex::MultiFab>, 3>& Jfield,
+    std::array< std::unique_ptr<amrex::MultiFab>, 3> const& Bfield,
+    std::array< std::unique_ptr<amrex::MultiFab>, 3 > const& edge_lengths,
+    int lev )
+{
+   // Select algorithm (The choice of algorithm is a runtime option,
+   // but we compile code for each algorithm, using templates)
+    if (m_fdtd_algo == ElectromagneticSolverAlgo::Yee ||
+        m_fdtd_algo == ElectromagneticSolverAlgo::HybridPIC) {
```
RemiLehe:
```suggestion
    if (m_fdtd_algo == ElectromagneticSolverAlgo::HybridPIC) {
```
Should we only allow this for `Hybrid`?

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -0,0 +1,511 @@
...
+            // Jx calculation
+            [=] AMREX_GPU_DEVICE (int i, int j, int k){
+#ifdef AMREX_USE_EB
+                // Skip if this cell is fully covered by embedded boundaries
+                if (lx(i, j, k) <= 0) return;
+#endif
+                Jx(i, j, k) = (
+                    - T_Algo::DownwardDz(By, coefs_z, n_coefs_z, i, j, k)
+                    + T_Algo::DownwardDy(Bz, coefs_y, n_coefs_y, i, j, k)
+                ) / PhysConst::mu0;
```
RemiLehe: Should we define:
```
Real const one_over_mu0 = 1. / PhysConst::mu0;
```
outside of the loop, as a small optimization? (maybe the compiler already automatically does this.)

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -0,0 +1,511 @@
...
+    //    energy conservation.
+    // 2) The nodal E-field values are averaged onto the Yee grid and the
+    //    electron pressure & resistivity terms are added (these terms are
+    //    naturally located on the Yee grid).
+
+    // Create a temporary multifab to hold the nodal E-field values
+    // Note the multifab has 3 values for Ex, Ey and Ez which we can do here
+    // since all three components will be calculated on the same grid.
+    auto const& ba = convert(rhofield->boxArray(), IntVect::TheNodeVector());
+    MultiFab enE_nodal_mf(ba, rhofield->DistributionMap(), 3, rhofield->nGrow());
```
RemiLehe: Naively, it seemed to me that `enE_nodal_mf` actually does not need to have guard cells.
```suggestion
    MultiFab enE_nodal_mf(ba, rhofield->DistributionMap(), 3, IntVect::TheZeroVector);
```
since we then only interpolate it to the staggered positions of E than are *inside* the area delimited by the outermost nodes of the valid grid. Is that correct?
In this case, we could get rid of the ` enE_nodal_mf.FillBoundary` further below.

Source/WarpX.H
```diff
@@ -1322,8 +1370,12 @@ private:
     amrex::Vector<            std::unique_ptr<amrex::MultiFab>      > F_fp;
     amrex::Vector<            std::unique_ptr<amrex::MultiFab>      > G_fp;
     amrex::Vector<            std::unique_ptr<amrex::MultiFab>      > rho_fp;
+    amrex::Vector<            std::unique_ptr<amrex::MultiFab>      > rho_fp_temp;
```
RemiLehe: Could you make these variables members of the `HybridPICModel` instead? 
This would provide more encapsulation and make it clearer which data members go along with which solver/scheme.

Source/WarpX.H
```diff
@@ -616,6 +620,39 @@ public:
     void MacroscopicEvolveE (int lev, amrex::Real dt);
     void MacroscopicEvolveE (int lev, PatchType patch_type, amrex::Real dt);
 
+    /**
+     * \brief
+     * Function to update the E-field using Ohm's law (hybrid-PIC model).
+     */
+    void HybridPICSolveE (                               DtType dt_type);
+    void HybridPICSolveE (const int lev,                       DtType dt_type);
+    void HybridPICSolveE (const int lev, PatchType patch_type, DtType dt_type);
```
RemiLehe: Could these functions become members of the `HybridModel` class?
This could provide more encapsulation and make it clearer which members go along with with solver/scheme, as well as de-clutter the `WarpX.H` file.
Note that, as a consequence, you may need to pass the `Efield_fp`, `Bfield_fp` as argument to these function.

Source/WarpX.cpp
```diff
@@ -2176,6 +2198,38 @@ WarpX::AllocLevelMFs (int lev, const BoxArray& ba, const DistributionMapping& dm
             dm, ncomps, ngEB, tag("vector_potential_grad_buf_b_stag[z]"), 0.0_rt);
     }
 
+    // Allocate extra multifabs needed by the kinetic-fluid hybrid algorithm.
```
RemiLehe, line 2190: If these are members of the `HybridPICModel` class, then the code below will need to be moved to somewhere in the `HybridPICModel` class.

Source/WarpX.H
RemiLehe: Also: these new data structures may cause problems when doing load-balancing, unless they are redistributed along with the other arrays.
It may be good to raise an error when hybrid is used along with load-balancing.

RemiLehe (changes requested):
This is still only a partial review, but I wanted to post some of the comments that we discussed offline with @roelof-groenewald

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -136,7 +136,39 @@ WarpX::Evolve (int numsteps)
...
+                for (int lev=0; lev <= finest_level; ++lev) {
+                    // SyncCurrent does not include a call to FillBoundary, but it is needed
+                    // for the hybrid-PIC solver since current values are interpolated to
+                    // a nodal grid
+                    current_fp_temp[lev][0]->FillBoundary(Geom(lev).periodicity());
+                    current_fp_temp[lev][1]->FillBoundary(Geom(lev).periodicity());
+                    current_fp_temp[lev][2]->FillBoundary(Geom(lev).periodicity());
+
+                    ApplyRhofieldBoundary(lev, rho_fp_temp[lev].get(), PatchType::fine);
```
RemiLehe: Is this actually needed?

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -136,7 +136,39 @@ WarpX::Evolve (int numsteps)
                             *Efield_aux[lev][0],*Efield_aux[lev][1],*Efield_aux[lev][2],
                             *Bfield_aux[lev][0],*Bfield_aux[lev][1],*Bfield_aux[lev][2]);
             }
+
+            // The hybrid-PIC algorithm uses the charge and current density
+            // from both the current and previous step when updating the
+            // fields, so we deposit the ion charge and current now in the
+            // temp multifab locations.
+            if (electromagnetic_solver_id == ElectromagneticSolverAlgo::HybridPIC)
```
RemiLehe: As discussed offline, this relies on `is_synchronized`, but `is_synchronized` may not always correspond to the very first timestep. It may be better to instead check `step == step_begin`.

Source/WarpX.H
roelof-groenewald: It is used in the `WarpX::ApplyElectronPressureBoundary()` function.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
roelof-groenewald: That reasoning makes sense to me, thanks for pointing it out! I've made this change, presumably the CI tests will all still pass.

Source/WarpX.H
roelof-groenewald: Moved to `HybridPICModel()` now.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
roelof-groenewald: This improvement cut the field solve time in half for large GPU runs :tada:

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -160,6 +162,38 @@ WarpX::Evolve (int numsteps)
...
+
+                ApplyRhofieldBoundary(lev, rho_fp_temp[lev].get(), PatchType::fine);
+                // Set current density at PEC boundaries, if needed.
+                ApplyJfieldBoundary(
+                    lev, current_fp_temp[lev][0].get(),
+                    current_fp_temp[lev][1].get(),
+                    current_fp_temp[lev][2].get(),
+                    PatchType::fine
+                );
+            }
```
RemiLehe, line 194: Could you wrap this in a small helper function (defined e.g. in the file `WarpXPushFieldsHybridPIC.cpp`), so that it takes only one line, in the `WarpX::Evolve` function.
```suggestion
HybridPICDepositOldRhoAndJ();
```

Source/WarpX.cpp
```diff
@@ -2233,8 +2253,10 @@ WarpX::AllocLevelMFs (int lev, const BoxArray& ba, const DistributionMapping& dm
     }
 #endif
 
-    bool deposit_charge = do_dive_cleaning || (electrostatic_solver_id == ElectrostaticSolverAlgo::LabFrame  ||
-                                               electrostatic_solver_id == ElectrostaticSolverAlgo::LabFrameElectroMagnetostatic);
+    bool deposit_charge = do_dive_cleaning ||
+                          (electrostatic_solver_id == ElectrostaticSolverAlgo::LabFrame) ||
+                          (electrostatic_solver_id == ElectrostaticSolverAlgo::LabFrameElectroMagnetostatic) ||
+                          (electromagnetic_solver_id == ElectromagneticSolverAlgo::HybridPIC);
```
RemiLehe, line 2259: We could do a follow-up PR that only allocates `rho_ncomps = ncomps` instead of `rho_ncomps = 2*ncomps` in these cases. (See a few lines below.)

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1294,13 +1302,32 @@ void WarpX::CheckKnownIssues()
...
         {
             ablastr::warn_manager::WMRecordWarning(
-                "PML",
-                "Using PSATD together with PML may lead to instabilities if the plasma touches the PML region. "
-                "It is recommended to leave enough empty space between the plasma boundary and the PML region.",
+                "Hybrid-PIC",
+                "When using Esirkepov current deposition together with the hybrid-PIC "
+                "algorithm, a segfault will occur if a particle moves over multiple cells "
+                "in a single step, so be careful with your choice of time step.",
                 ablastr::warn_manager::WarnPriority::low);
```
RemiLehe, line 1322: As discussed previously, could we set the default current deposition to `direct` with the hybrid solver, instead of Esirkepov (and then modify the documentation accordingly)

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -0,0 +1,511 @@
...
+    std::unique_ptr<amrex::MultiFab> const& rhofield,
+    std::unique_ptr<amrex::MultiFab> const& Pefield,
+    std::array< std::unique_ptr<amrex::MultiFab>, 3 > const& edge_lengths,
+    int lev, HybridPICModel const* hybrid_model,
+    DtType a_dt_type )
+{
+
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+        WarpX::grid_type == GridType::Staggered,
+        "Ohm's law E-solve only works with a staggered (Yee) grid.");
```
RemiLehe, line 204: I would recommend directly checking the `IndexType` of `Efield`, `Jfield`, `Bfield`, etc. to make sure that they match the expected one.
It is more verbose, but probably more robust (since the `grid_type` argument may be subject to changes in the future).

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -0,0 +1,338 @@
...
+    }
+}
+
+void HybridPICModel::HybridPICSolveE (
+    std::array< std::unique_ptr<amrex::MultiFab>, 3> & Efield,
+    std::array< std::unique_ptr<amrex::MultiFab>, 3> const& Jfield,
+    std::array< std::unique_ptr<amrex::MultiFab>, 3> const& Bfield,
+    std::unique_ptr<amrex::MultiFab> const& rhofield,
+    std::array< std::unique_ptr<amrex::MultiFab>, 3> const& edge_lengths,
+    const int lev, PatchType patch_type, DtType a_dt_type)
```
RemiLehe, line 243: Instead of passing `a_dt_type`, I think it would be clearer and more transparent to directly call this function by passing `current_fp_ampere[lev]` or `Jfield` (and `rho_fp_temp[lev]` or `rhofield`) as arguments, directly from `HybridPICEvolveFields`. We could also add a boolean flag `include_resistivity_term` (instead of relying of `DtType::Full`).

RemiLehe (approved):
This looks good! Thanks again for this PR! 
I added a few additional comments, but I think that they could be addressed as follow-up PRs. @roelof-groenewald Would you have time to implement this as a follow-up?

roelof-groenewald:
Thanks for the final review @RemiLehe! I will submit a series follow-up PRs for the comments you left (probably some time next week).

Source/Evolve/WarpXEvolve.cpp
roelof-groenewald: Addressed in #4032.

Source/Initialization/WarpXInitData.cpp
roelof-groenewald: Addressed in #4033.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
roelof-groenewald: Addressed on #4034. Note that I moved the check to `HybridPICModel::InitData()` since there is no need to check this on every step.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
roelof-groenewald: Addressed in #4044.

PR #3652, opened 2023-01-31
Add Sr, Y, Zr, Nb, Mo, Tc, Ru, Rh, Pd, Ag, Cd, In, Sn, Sb, Te, I, Xe to the ionization table

lucafedeli88:
As a follow-up to https://github.com/ECP-WarpX/WarpX/pull/3646 this PR adds a new row of the periodic table to the ionization table.

PR #3666, opened 2023-02-02
Add all the elements from Z=55 (Cs) to Z=85 (At) to the ionization table

lucafedeli88:
As a follow-up to https://github.com/ECP-WarpX/WarpX/pull/3652, this PR adds new elements to the ionization table


dpgrote (approved):
Thanks!

ax3l (approved):

PR #3669, opened 2023-02-02
Fix a bug in 1D PEC

### LINKED ISSUE #3664: Possible bug: PEC boundary in 1D
I am using a 1D uniform magnetized plasma with reflecting boundaries for particles and Dirichlet (PEC) boundaries for fields with an EM solver. I am not explicitly setting the boundary potentials; I understand that they are set to zero by default, which is what I want. As I understand, the PEC boundaries should be clamping the tangential (x and y) electric fields to zero. In fact, what I see is that only Ey is clamped to zero. (The PEC boundary also sets the normal component of magnetic fields (Bz) to zero, but in 1D this is automatically satisfied.) Possibly as a consequence, I am seeing an unphysical instability occur, with both total particle and field energy increasing when there are no physical sources of free energy in the plasma. Evidential plots are below, as well as the input file. The input file is also included in a zip folder. 

A plot of Ey vs z at the final timestep. The field is correctly clamped to zero at the boundaries.
image

A plot of Ex vs z at the final timestep. The field is not clamped to zero at the boundaries. 
image

A colormap of Bx in space and time, showing wave activity. (Time runs downward on the vertical axis, and space runs to the right on the horizontal axis.)
image

A plot of total particle energy vs time.
image

A plot of total field energy vs time. This plot and the plot above show the presence of an instability. 
image

PICMI-translated input deck
image
[inputs_1d_picmi.zip](https://github.com/ECP-WarpX/WarpX/files/10561982/inputs_1d_picmi.zip)

prkkumar:
This PR fixes a bug in PEC boundary condition in 1D. In particular, the way we calculate which component is tangential and which is normal was not properly done for the 1D case. This PR fixes that. 

This should fix #3664.


ax3l (approved):
Thanks a lot for the fix! :)

ax3l:
CI has some hickups (network issues during install phase), I will restart the ones that fail.

prkkumar:
LaserInjection_1d benchmark needs to be updated, I think. It uses pec

ax3l:
Makes sense, please feel free to push an updated checksum

Regression/Checksum/benchmarks_json/LaserInjection_1d.json
```diff
@@ -1,13 +1,23 @@
 {
+  "electrons": {
+    "particle_momentum_x": 0.0,
+    "particle_momentum_y": 0.0,
+    "particle_momentum_z": 0.0,
+    "particle_orig_z": 0.0,
+    "particle_position_x": 0.0,
+    "particle_regionofinterest": 0.0,
+    "particle_weight": 0.0
+  },
```
ax3l: Is this part off?
Did you change the inputs file locally, maybe?
```suggestion
```

ax3l: oops: mismatch LaserInjection_1d vs LaserAcceleration_1d

ax3l (approved):
Thank you for the quick debugging and fix, this is great! :rocket: :sparkles:

PR #3670, opened 2023-02-03
Add all the elements from Z=87 (Fr) to Z=100 (Fm) to the ionization table

lucafedeli88:
This PR is a follow up of https://github.com/ECP-WarpX/WarpX/pull/3666 . Now all the elements up to Z=100 are in the ionization table. I've also updated the documentation accordingly.
Supporting higher Z elements is problematic because some entries have missing rows in the NIST database. However, I doubt that a WarpX user will ever need to simulate the ionization of a target with Z>100 !


dpgrote (approved):
Thanks! I agree that 100 is a good place to stop.

n01r:
If we ever wanted to extend this we could fill the missing entries with model-calculated values and then document the used model.

n01r (approved):

ax3l (approved):

Issue #1916, opened 2021-04-27
Naming Convention for Particle Momenta and Proper Velocities


EZoni:
As discussed offline, we would like to have more consistency in the naming conventions for the variables that represent the particle momenta and proper velocities in the code.

For instance, we have:

1. `<species_name>.ux` for particles with constant momentum profile, where `ux` represents the normalized (non-dimensional) momentum `ux = px/(mc)`:
https://github.com/ECP-WarpX/WarpX/blob/5035351505823a0f2ba4070434453746064589cc/Source/Initialization/PlasmaInjector.cpp#L421-L430

2. the particle attribute accessed via `PIdx::ux`, where `ux` represents the proper velocity `ux = px/m`:
https://github.com/ECP-WarpX/WarpX/blob/5035351505823a0f2ba4070434453746064589cc/Source/Particles/WarpXParticleContainer.H#L33-L43

</br>

Observations, comments and suggestions so far:
1. historically, the name `ux` used to represent proper velocities in Warp (and that's still the case);

2. shall we change the name of the normalized momentum accessed by users in the input file?
Possible choices to replace `<species_name>.ux`:
   - `<species_name>.gamma_beta_x`
   - `<species_name>.px_mc`

3. shall we change the name of the proper velocity attribute accessed internally via `PIdx::ux`?
Possible choices to replace `PIdx::ux`:
   - `PIdx::gammavx`
   - `PIdx::gamma_vx`
   - `PIdx::px_m`

4. shall we use particle momenta directly in the equations themselves, without switching to proper velocities?

In general, about points 2. and 3. I personally like `px_mc` to denote the normalized momentum `px/(mc)` and `px_m` to denote the proper velocity `px/m`. They seem both short and self-explanatory. However, I understand the concern of impacting all users by changing the name in point 2.

Please feel free to add further suggestions and/or give your opinions on points 2., 3., 4. above.

PR #3672, opened 2023-02-03
Renamed "particle_vel" to "particle_mom"

titoiride:
In the input file, when initializing particles either using the `SingleParticle` or the `MultipleParticles` injection methods,
the parameter used to define the initial particle momentum is called `single_particle_vel` and `multiple_particle_vel_<direction>` respectively.
In this PR we change the naming from `vel` to `mom` since the parameters represent normalized momentum, in order to avoid confusion and specify the particle relativistic beta instead


dpgrote:
This is a good change to make, though "mom" is still not quite right since it could cause confusion (about whether or not it includes the mass). The most precise would be "single_particle_gammabeta". Alternatively, this could be "single_particle_u" or "single_particle_uxyz", since WarpX does use the notation that "u" represents "gamma*beta". This latter would be consist with other input parameter names. It would also be "multiple_particle_ux" etc.

titoiride:
Thanks for the comment, I totally agree. I think using the `u` would be a perfect choice because as you mentioned it's consistent with the other parameters.
PR related question: how should I fix the PR? Are you fine with multiple commits or should I rebase it and clean it?

dpgrote:
How you update the PR won't matter much since all of the commits will be squashed into a single commit anyway when it is merged.

ax3l:
Thanks, great change!

Explicit naming is a great suggestion, I think _betagamma_ is a bit more common than _gammabeta_ but leave that up to you.

Please do not forget to update the documentation
https://warpx.readthedocs.io/en/latest/usage/parameters.html
which is in the file
`Docs/source/usage/parameters.rst`
with this PR.

Please add guidance for users in `WarpX::BackwardCompatibility` on the changed name :)
https://github.com/ECP-WarpX/WarpX/blob/a6a466b0362a60a74f095ce50a1e2d5ed22cdd3a/Source/WarpX.cpp#L1446-L1448

titoiride:
In the end I chose `single_particle_u` and `multiple_particles_u<x,y,z>` for the initialization. Betagamma is common, but it was not consistent with the other initialization parameters that all used u.
A question regarding the `WarpX::BackwardCompatibility` guidance: how can I parse from the input file the species name at that level? I saw that in `PlasmaInjector.cpp` the species name is passed to the constructor, but I'm not sure where I can find the knowledge about the species_names within `WarpX.cpp`.
Should I issue a re-parse of the line `particles.species_names` of the input file within the `BackwardCompatibility()` function?

EZoni:
Just linking an old issue which is partially related: https://github.com/ECP-WarpX/WarpX/issues/1916

ax3l:
> A question regarding the WarpX::BackwardCompatibility guidance: how can I parse from the input file the species name at that level? I saw that in PlasmaInjector.cpp the species name is passed to the constructor, but I'm not sure where I can find the knowledge about the species_names within WarpX.cpp.
Should I issue a re-parse of the line particles.species_names of the input file within the BackwardCompatibility() function?

Yes, you just iterate the species and warn if any of them uses the old parameter - please include the new name in the warning message.

Iterating the species names is cheap: this is all cached after first read.

.gitignore
```diff
@@ -56,6 +56,11 @@ cmake-build-*/
 .kdev?/
 *.kdev?
 
+##########
+# VsCode #
+##########
+.vscode
+
```
dpgrote: This change is ok here, but would be better in a separate PR since it isn't related to the main changes.

Python/pywarpx/picmi.py
```diff
@@ -450,9 +450,9 @@ def initialize_inputs(self, species_number, layout, species, density_scale):
         species.multiple_particles_pos_x = self.x
         species.multiple_particles_pos_y = self.y
         species.multiple_particles_pos_z = self.z
-        species.multiple_particles_vel_x = np.array(self.ux)/constants.c
-        species.multiple_particles_vel_y = np.array(self.uy)/constants.c
-        species.multiple_particles_vel_z = np.array(self.uz)/constants.c
+        species.multiplt_particle_ux = np.array(self.ux)/constants.c
```
dpgrote: I notice a typo here - this should be "multiple_particle_ux", not "multiplt_particle_ux". I see the same typo every place this substitution is made.

.gitignore
titoiride: I can revert it and propose a new small PR for it, no problem for me.

Python/pywarpx/picmi.py
titoiride: Thanks for checking 😓

Source/WarpX.cpp
```diff
@@ -1627,6 +1627,25 @@ WarpX::BackwardCompatibility ()
...
+    pp_particles.queryarr("species_names", backward_sp_names);
+    for(std::string speciesiter : backward_sp_names){
+        ParmParse pp_species(speciesiter);
+        std::vector<amrex::Real> backward_vel;
+
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+            !pp_species.queryarr("multiple_particles_vel_x", backward_vel) &&
+            !pp_species.queryarr("multiple_particles_vel_y", backward_vel) &&
+            !pp_species.queryarr("multiple_particles_vel_z", backward_vel),
+            "multiple_particles_vel_<x,y,z> are not supported anymore. "
```
ax3l: These messages are great!

Please also include the species name `speciesiter` in the text? That way users can find it quicker in their input.

Something like:
```
electrons.multiple_particles_vel_<x,y,z> are not supported anymore. ...
```

titoiride: Thanks, good suggestion, I implemented it.

ax3l (approved):
Thanks a lot! :)

PR #3680, opened 2023-02-07
CI & Bug Fixes: 1D & 2D Compile

ax3l:
Add a modern CI entry that compiles 1D and 2D with `-Werror` and many warnings enabled.

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -49,7 +49,7 @@ jobs:
         cmake --build build_RZ -j 2
         ./build_RZ/bin/warpx Examples/Physics_applications/laser_acceleration/inputs_rz
 
-  build_2D_rz:
+  build_1D_2D:
     name: GCC 1D & 2D w/ MPI
     runs-on: ubuntu-22.04
     if: github.event.pull_request.draft == false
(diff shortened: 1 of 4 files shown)


prkkumar (approved):

ax3l:
Something still defaults in 2D, let's test locally

.github/workflows/ubuntu.yml
```diff
@@ -49,6 +49,53 @@ jobs:
         cmake --build build_RZ -j 2
         ./build_RZ/bin/warpx Examples/Physics_applications/laser_acceleration/inputs_rz
 
+  build_2D_rz:
```
dpgrote: This all looks good. Though, this name should presumably be "build_1D_2D". For consistency, perhaps the name above should be "build_cxxminimal_3D_RZ"?

dpgrote:
Maybe I missed the discussion, but presumably gcc12 is being used because it is better at doing the checks. Is there a reason that gcc12 is also not used for build_cxxminimal?

.github/workflows/ubuntu.yml
ax3l: Good point :)

ax3l:
The reason I use a newer one here is to have a broad coverage in compilers we test. We should update some of the other tests to use it too, though. Will review that separately.

ax3l:
> Something still defaults in 2D, let's test locally

Hm, I think EB in 1D and 2D (and likely RZ, too) is still segfaulting - even it its just enabled and not used.

To be followed up on Tue. #3686

.github/workflows/ubuntu.yml
ax3l: (not minimal because minimal is with respect to third-party dependencies)

PR #3688, opened 2023-02-10
Update Crusher Modules to cce/15.0.0 and others accordingly

n01r:
Module changes were suggested by HPE on 2023/02/08 due to problems with building ImpactX on cce/14.0.2.
Hence we update WarpX dependencies accordingly, too.

We build `c-blosc` and `adios2` manually but this might change again with ticket `OLCFHELP-10863`.


ax3l (approved):
Great job, thanks a lot!

PR #3689, opened 2023-02-13
use correct positions for moving window for BTD at restart

### LINKED ISSUE #2468: BTD bug-fix for restart
PR #2356 fixed the starting position for BTD in the old implementation. 
This needs to be added in the new BTD implementation.

RevathiJambunathan:
This was done in the old BTD in [this PR ](https://github.com/ECP-WarpX/WarpX/pull/2356/files), but not in the new BTD
With this PR, we should be able to run boosted frame simulation at restart, without BTD crashing.

Close #2468

### DIFF
--- a/Source/Diagnostics/BTDiagnostics.cpp
+++ b/Source/Diagnostics/BTDiagnostics.cpp
@@ -367,12 +367,16 @@ BTDiagnostics::InitializeBufferData ( int i_buffer , int lev)
             (diag_ba.getCellCenteredBox( diag_ba.size()-1 ).bigEnd(idim) + 1) * warpx.Geom(lev).CellSize(idim));
     }
 
+    // When restarting boosted simulations, use correct positions for the moving window
+    amrex::Real boosted_moving_window_v = (warpx.moving_window_v - m_beta_boost*PhysConst::c)
+                                        / (1._rt - m_beta_boost * warpx.moving_window_v/PhysConst::c)
+
     // Define buffer_domain in lab-frame for the i^th snapshot.
     // Replace z-dimension with lab-frame co-ordinates.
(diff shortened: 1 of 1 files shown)


ax3l (approved):
Thanks a lot - LGTM.

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -367,12 +367,16 @@ BTDiagnostics::InitializeBufferData ( int i_buffer , int lev)
...
+    amrex::Real boosted_moving_window_v = (warpx.moving_window_v - m_beta_boost*PhysConst::c)
+                                        / (1._rt - m_beta_boost * warpx.moving_window_v/PhysConst::c);
+
     // Define buffer_domain in lab-frame for the i^th snapshot.
     // Replace z-dimension with lab-frame co-ordinates.
-    amrex::Real zmin_buffer_lab = diag_dom.lo(m_moving_window_dir)
-                                 / ( (1.0_rt + m_beta_boost) * m_gamma_boost);
-    amrex::Real zmax_buffer_lab = diag_dom.hi(m_moving_window_dir)
-                                 / ( (1.0_rt + m_beta_boost) * m_gamma_boost);
+    amrex::Real zmin_buffer_lab = ( diag_dom.lo(m_moving_window_dir) - boosted_moving_window_v * warpx.gett_new(0) )
```
n01r, line 376: Just a note: this looks like we are currently limited to boosts in z whereas the moving window could in principle move in any direction, correct?

RevathiJambunathan: yes, BTD assumes that moving window is in z 
(this can be generalized)

RevathiJambunathan:
PR #3734 is all we would need. 
We store the meta-data instead of re-computing them

PR #3692, opened 2023-02-15
RZ PSATD: tell users that number of cells in r should be a power of 2

RemiLehe:
In RZ PSATD, it seems that the function `CheckGriddingForRZSpectral` tries to set the `blocking_factor` as close as possible to `n_r` (the number of cells in `r`), in order to prevent domain decomposition along `r`.

However, if `n_r` is not exactly equal to the `blocking_factor`, the user gets an error message saying `domain size not divisible by blocking_factor`. This can confuse users, as they may not know what to change in the input script. (Changing the `blocking_factor` in the input script will not help, since it is overwritten by `CheckGriddingForRZSpectral`).

Overall, it seems that the user needs to set `n_r` to a power of 2, so that it can be exactly equal to the `blocking_factor`. This PR adds a corresponding warning message.

### DIFF
--- a/Source/Utils/WarpXUtil.cpp
+++ b/Source/Utils/WarpXUtil.cpp
@@ -328,6 +328,9 @@ void CheckGriddingForRZSpectral ()
     }
     blocking_factor_x[0] = k;
     max_grid_size_x[0] = k;
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(n_cell[0] == k,
+        "With RZ spectral, the number of cells in r must be a power of 2.");
+
 
     for (int lev=1 ; lev <= max_level ; lev++) {
         // For this to be correct, this needs to read in any user specified refinement ratios.


dpgrote:
If I remember, there is nothing in the RZ spectral solver that requires n_r to be a power of two.

This error message can be removed by setting `max_grid_size_x[0] = n_cell[0];` at line 330 in `Source/Utils/WarpXUtil.cpp`. What is happening is that the `max_grid_size_x` was being set equal to `blocking_factor_x`, and when this is true is when n_r must be a multiple of the blocking factor. However, if `max_grid_size_x` is set to `n_r` (which would be less than or equal to `blocking_factor_x`), then the check for a multiple is not done. I tried this in a test case and it seemed to run fine.

EZoni:
@dpgrote Was this PR superseded by #4073, given what you explained in your comment above?

dpgrote:
That is correct - this PR should be closed, superseded by #4073. I had forgotten about this PR. My comment on Feb 15 is exactly what #4073 is doing.

PR #3693, opened 2023-02-15
add option warpx.compute_max_step_from_btd to increase max_step to fill BTD

RTSandberg:
This PR adds an extra boolean flag ``warpx.compute_max_step_from_btd`` that allows the WarpX to automatically set or increase ``max_step`` as necessary to fill all BTD snapshots.  Previously users had to guess a sufficient ``max_step`` or ``zmax_plasma_to_compute_max_step`` to fill all BTD snapshots.  If incorrect, a warning was printed when ``max_step`` was insufficient to fill all BTD snapshots and gave the value that would probably fill all snapshots, but this two-run setup complicates simulation workflows.  This PR simplifies the work needed to fill all BTD snapshots.

TODO
- [x] add CI test / example input

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -60,9 +60,11 @@ Overall simulation parameters
 * ``warpx.do_compute_max_step_from_btd`` (`integer`; 0 by default) optional
     Can be useful when computing back-transformed diagnostics.  If specified,
     automatically calculates the number of iterations required in the boosted
-    frame for all back-transformed diagnostics to be completed. The value of
-    ``max_step`` is overwritten, and printed to standard output. Currently
-    only works if the Lorentz boost and the moving window are along the z direction.
+    frame for all back-transformed diagnostics to be completed. If the current
+    value of ``max_step`` is too low to fill all BTD, the value of
+    ``max_step`` is overwritten with the new value and printed to standard output. 
+    Currently only works if the Lorentz boost and the moving window are along 
+    the z direction.
 
 * ``warpx.random_seed`` (`string` or `int` > 0) optional
     If provided ``warpx.random_seed = random``, the random seed will be determined
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -61,8 +61,8 @@ Overall simulation parameters
     Can be useful when computing back-transformed diagnostics.  If specified,
     automatically calculates the number of iterations required in the boosted
     frame for all back-transformed diagnostics to be completed. If ``max_step``
-    or ``zmax_plasma_to_compute_max_step`` are not specified or the current 
-    value of ``max_step`` is too low to fill all BTD, the value of ``max_step`` 
+    or ``zmax_plasma_to_compute_max_step`` are not specified or the current
+    value of ``max_step`` is too low to fill all BTD, the value of ``max_step``
     is overwritten with the new value and printed to standard output.
 
 * ``warpx.random_seed`` (`string` or `int` > 0) optional
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -60,11 +60,10 @@ Overall simulation parameters
 * ``warpx.do_compute_max_step_from_btd`` (`integer`; 0 by default) optional
     Can be useful when computing back-transformed diagnostics.  If specified,
     automatically calculates the number of iterations required in the boosted
-    frame for all back-transformed diagnostics to be completed. If the current
-    value of ``max_step`` is too low to fill all BTD, the value of
-    ``max_step`` is overwritten with the new value and printed to standard output.
-    Currently only works if the Lorentz boost and the moving window are along
-    the z direction.
+    frame for all back-transformed diagnostics to be completed. If ``max_step``
+    or ``zmax_plasma_to_compute_max_step`` are not specified or the current 
+    value of ``max_step`` is too low to fill all BTD, the value of ``max_step`` 
+    is overwritten with the new value and printed to standard output.
 
 * ``warpx.random_seed`` (`string` or `int` > 0) optional
     If provided ``warpx.random_seed = random``, the random seed will be determined
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -63,7 +63,7 @@ Overall simulation parameters
     frame for all back-transformed diagnostics to be completed. If ``max_step``,
     ``stop_time``, or ``warpx.zmax_plasma_to_compute_max_step`` are not specified,
     or the current values of ``max_step`` and/or ``stop_time`` are too low to fill
-    all BTD snapshots, the values of ``max_step`` and/or ``stop_time`` are 
+    all BTD snapshots, the values of ``max_step`` and/or ``stop_time`` are
     overwritten with the new values and printed to standard output.
 
 * ``warpx.random_seed`` (`string` or `int` > 0) optional
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -60,8 +60,8 @@ Overall simulation parameters
 * ``warpx.do_compute_max_step_from_btd`` (`integer`; 0 by default) optional
     Can be useful when computing back-transformed diagnostics.  If specified,
     automatically calculates the number of iterations required in the boosted
-    frame for all back-transformed diagnostics to be completed. The value of 
-    ``max_step`` is overwritten, and printed to standard output. Currently 
+    frame for all back-transformed diagnostics to be completed. The value of
+    ``max_step`` is overwritten, and printed to standard output. Currently
     only works if the Lorentz boost and the moving window are along the z direction.
 
 * ``warpx.random_seed`` (`string` or `int` > 0) optional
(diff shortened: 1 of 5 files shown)


RTSandberg:
It should be noted that the computed ``max_step`` to fill all snapshots may be off by a few iterations.  Perhaps adding a cushion of a few snapshots is a good idea?

Docs/source/usage/parameters.rst
```diff
@@ -57,6 +57,14 @@ Overall simulation parameters
     printed to standard output. Currently only works if the Lorentz boost and
     the moving window are along the z direction.
 
+* ``warpx.do_compute_max_step_from_btd`` (`integer`; 0 by default) optional
```
RTSandberg:
```suggestion
* ``warpx.compute_max_step_from_btd`` (`integer`; 0 by default) optional
```

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -158,14 +158,25 @@ void BTDiagnostics::DerivedInitData ()
...
+
     if (final_snapshot_fill_iteration > warpx.maxStep()) {
-        std::string warn_string =
-            "\nSimulation might not run long enough to fill all BTD snapshots.\n"
-            "Final step: " + std::to_string(warpx.maxStep()) + "\n"
-            "Last BTD snapshot fills around step: " + std::to_string(final_snapshot_fill_iteration);
-        ablastr::warn_manager::WMRecordWarning(
-            "BTD", warn_string,
-            ablastr::warn_manager::WarnPriority::low);
+        if (warpx.do_compute_max_step_from_btd) {
```
RTSandberg:
```suggestion
        if (warpx.compute_max_step_from_btd) {
```

Source/WarpX.H
```diff
@@ -295,6 +295,9 @@ public:
     //! simulation domain along z reaches #zmax_plasma_to_compute_max_step in the boosted frame
     static bool do_compute_max_step_from_zmax;
 
+    //! If true, the code will compute max_step from the back transformed diagnostics
+    static bool do_compute_max_step_from_btd;
```
RTSandberg:
```suggestion
    static bool compute_max_step_from_btd;
```

Source/WarpX.cpp
```diff
@@ -114,6 +114,7 @@ Real WarpX::gamma_boost = 1._rt;
 Real WarpX::beta_boost = 0._rt;
 Vector<int> WarpX::boost_direction = {0,0,0};
 bool WarpX::do_compute_max_step_from_zmax = false;
+bool WarpX::do_compute_max_step_from_btd = false;
```
RTSandberg:
```suggestion
bool WarpX::compute_max_step_from_btd = false;
```

Source/WarpX.cpp
```diff
@@ -622,6 +623,9 @@ WarpX::ReadParameters ()
             pp_warpx, "zmax_plasma_to_compute_max_step",
             zmax_plasma_to_compute_max_step);
 
+        pp_warpx.query("do_compute_max_step_from_btd",
```
RTSandberg:
```suggestion
        pp_warpx.query("compute_max_step_from_btd",
```

Source/WarpX.cpp
```diff
@@ -622,6 +623,9 @@ WarpX::ReadParameters ()
...
+            do_compute_max_step_from_btd);
```
RTSandberg:
```suggestion
            compute_max_step_from_btd);
```

RTSandberg (commented):
shorten the flag name a bit, to ``compute_max_step_from_btd``

Source/WarpX.H
```diff
@@ -779,6 +782,7 @@ public:
     bool getis_synchronized() const {return is_synchronized;}
 
     int maxStep () const {return max_step;}
+    void updateMaxStep (int const new_max_step) {max_step = new_max_step;}
```
EZoni: Are there other parts of the code where `max_step` is overwritten and which could reuse this new function? If so, I would be in favor of reusing this function there, too, since it is now available. For example, we might be doing something similar when we compare `max_step` and `zmax_plasma_to_compute_max_step`, but I'm not 100% sure.

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -158,14 +158,25 @@ void BTDiagnostics::DerivedInitData ()
...
+                <<final_snapshot_fill_iteration<<std::endl;
+        } else {
+            std::string warn_string =
+                "\nSimulation might not run long enough to fill all BTD snapshots.\n"
+                "Final step: " + std::to_string(warpx.maxStep()) + "\n"
+                "Last BTD snapshot fills around step: " + std::to_string(final_snapshot_fill_iteration);
+            ablastr::warn_manager::WMRecordWarning(
+                "BTD", warn_string,
+                ablastr::warn_manager::WarnPriority::low);
+        }
```
EZoni: Do I understand the code behavior correctly?
- If both `max_step` and `compute_max_step_from_btd` are set in the input, and if `max_step` is not enough to fill all BTD, then we overwrite `max_step` and print a message;
- If `max_step` is set in the input but `compute_max_step_from_btd`  isn't, and if `max_step` is not enough to fill all BTD, then we print a warning, but do not overwrite `max_step` (so we end up not filling all BTD);
- If `max_step` is not set in the input but `compute_max_step_from_btd` is, then `max_step` is set automatically.

Question: shouldn't we want to kind of merge points 1 and 2 above, namely overwrite `max_step` regardless of whether `compute_max_step_from_btd` is set in the input? In this case, the new parameter `compute_max_step_from_btd` would be still useful for users who do not want to specify `max_step`.

Happy to discuss more offline, if my comment isn't clear.

EZoni (commented):
Thank you for this PR, @RTSandberg! It will be very useful. I left only a couple of comments.

Source/Diagnostics/BTDiagnostics.cpp
RTSandberg: Yes, the PR was specifically designed to achieve these 3 points.  I'm glad they came out so clearly.  I thought about combining points 1 and 2, but I didn't want to overwrite ``max_step`` unless a user specifically requests for this to be done.  The decision to combine points 1 and 2 seems beyond my paygrade to make, I yield to higher authorities

Source/WarpX.H
RTSandberg: The parameter``zmax_plasma_to_compute_max_step``  is used to compute ``max_step`` in a ``WarpX`` member function, so ``max_step`` is directly available.  The use of ``updateMaxStep`` in this function is deep in a Diagnostics initialization function and so perhaps dangerous.  I couldn't think of a safer way to change ``max_step`` or I would avoid the use of this function.

ax3l (commented):
This looks very useful, thank you! :)

One comment: since we have multiple end-of-sim conditions, the first one that matches will be taken.  You might want to update the set physical time as well (as a float)?

Source/WarpX.H
```diff
@@ -779,7 +782,9 @@ public:
...
     amrex::Real stopTime () const {return stop_time;}
+    void updateStopTime (int amrex::Real new_stop_time) {stop_time = new_stop_time;}
```
```suggestion
    void updateStopTime (amrex::Real const new_stop_time) {stop_time = new_stop_time;}
```

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -158,7 +158,29 @@ void BTDiagnostics::DerivedInitData ()
...
+
+        } else if (warpx.maxStep() == std::numeric_limits<int>::max() && warpx.stopTime() == std::numeric_limits<amrex::Real>::max()) {
+            amrex::Print()<<"max_step unspecified.  Setting to "
+                <<final_snapshot_fill_iteration<< " to fill all BTD snapshots." << std::endl;
+            warpx.updateMaxStep(final_snapshot_fill_iteration);
+            amrex::Print()<<"stop_time unspecified.  Setting to "
+                <<final_snapshot_fill_time<< " to fill all BTD snapshots." << std::endl;
+            warpx.updateMaxStep(final_snapshot_fill_time);
+
+        }
```
EZoni: Is this the case where `zmax_plasma_to_compute_time_step` is specified? Or is it possible that neither of the three are specified by the user? If this corresponds to the case where `zmax_plasma_to_compute_time_step` is specified, when does the computation of `max_step` (or `stop_time`) based on `zmax_plasma_to_compute_time_step` occur, before or after this function is called? If before, I would assume that we actually do not end up in this `else if` block, but then I'm lost about which case this covers. Might be useful to clarify this offline.

RTSandberg: This (or rather a similar block in a newer version) is the case when none of ``max_step``, ``stop_time``, or ``zmax_plasma_to_compute_time_step`` are specified but ``compute_max_step_from_btd=1``.

RTSandberg:
Responding to @ax3l : I forgot to account for ``stop_time`` being set.  This PR now works regardless of the termination criterion specified.  The warning may look a bit odd -- I don't know how to get the warning string 
        ``std::string warn_string =``
            ``"\nSimulation might not run long enough to fill all BTD snapshots.\n"``
            ``"Final step: " + std::to_string(warpx.maxStep()) + "\n"``
            ``"Stop time: " + std::to_string(warpx.stopTime()) + "\n" ...``
to format in scientific notation.

Update: solved using `std::stringstream`.

EZoni (approved):

PR #3698, opened 2023-02-17
Add waterbag generation in injector

titoiride:
With this PR a new option for particle generation is added.
It's now possible to request a **Uniform** particle momentum distribution, that is a uniform distribution between two values of the momentum, `u<x,y,z>_min < u < u<x,y,z>_max`.

To-Do:
- [x] Update documentation
  - https://warpx.readthedocs.io/en/latest/usage/parameters.html#particle-initialization
  - source file: `Docs/source/usage/parameters.rst`
- [x] Add/modify a test: `Examples/Tests/initial_distribution`
- [ ] PICMPI Python API?


Source/Initialization/InjectorMomentum.H
```diff
@@ -220,6 +220,49 @@ private:
...
+    InjectorMomentumWaterbag (amrex::Real a_ux_min, amrex::Real a_uy_min,
+                              amrex::Real a_uz_min, amrex::Real a_ux_max,
+                              amrex::Real a_uy_max, amrex::Real a_uz_max) noexcept
+        : m_ux_min(a_ux_min), m_uy_min(a_uy_min), m_uz_min(a_uz_min),
+          m_ux_max(a_ux_max), m_uy_max(a_uy_max), m_uz_max(a_uz_max)
+        {
+            m_Dux = m_ux_max - m_ux_min;
+            m_Duy = m_uy_max - m_uy_min;
+            m_Duz = m_uz_max - m_uz_min;
+            m_ux_h = 0.5 * (m_ux_max - m_ux_min);
```
dpgrote: This should presumably be "+" to take the average.

Source/Initialization/InjectorMomentum.H
```diff
@@ -220,6 +220,49 @@ private:
...
+            m_uy_h = 0.5 * (m_uy_max - m_uy_min);
+            m_uz_h = 0.5 * (m_uz_max - m_uz_min);
+        }
+
+    AMREX_GPU_HOST_DEVICE
+    amrex::XDim3
+    getMomentum (amrex::Real /*x*/, amrex::Real /*y*/, amrex::Real /*z*/,
+                 amrex::RandomEngine const& engine) const noexcept
+    {
+        return amrex::XDim3{m_ux_min + amrex::Random(engine) * m_Dux,
```
dpgrote, line 247: This generates a cube in velocity space and so will have a correlation along the coordinate axes. Would a sphere in velocity space be better since it would not have that correlation?

n01r (commented):
Thanks for this implementation, @titoiride!

Source/Initialization/InjectorMomentum.H
titoiride: I think you raise a really interesting point and I'm not sure I get the full picture on when this would be an actual issue or not. The classical waterbag is spherical in the momenta, so maybe the naming I'm using is inappropriate, but spherical momentum distributions would constrain all the momentum direction to the same maximum value (i.e. random direction for a given kinetic energy), which sometimes might be unnecessary if I already know I'm interested in a subset of those particles. Elliptical, perhaps? I wonder if all of those possibilities serve for different purposes (and at that point it might be worth having all of them available). In fact, it probably depends on the application and often one would have different constraints, for example when generating a beam you select the energy (with energy spread), main direction and RMS divergence.

Source/Initialization/InjectorMomentum.H
titoiride: Thanks, when I copy-pasted the lines I missed the typo!

Source/Initialization/InjectorMomentum.H
cemitch99: A general approach to the classical waterbag allows for a distribution on a phase space of any dimension (2D, 4D, or 6D), with an arbitrary covariance matrix.  The idea is to uniformly populate a phase space ball of the appropriate dimension, and then to apply a linear coordinate transformation to produce the correct covariance matrix.  The 6D version is used often in Impact-T.  In 2D, this would produce a uniformly-filled ellipse.

cemitch99: The details of how we do this in the 6D case in ImpactX can be found here:
https://github.com/ECP-WarpX/impactx/blob/development/src/particles/distribution/Waterbag.H

ax3l: @titoiride do you potentially have time to update this PR, doing the same as in ImpactX? :)

titoiride: Thanks for the feedback, I'll resume working on this PR ASAP.

titoiride: > A general approach to the classical waterbag allows for a distribution on a phase space of any dimension (2D, 4D, or 6D), with an arbitrary covariance matrix. The idea is to uniformly populate a phase space ball of the appropriate dimension, and then to apply a linear coordinate transformation to produce the correct covariance matrix. The 6D version is used often in Impact-T. In 2D, this would produce a uniformly-filled ellipse.

The "design" problem I see here is that the struct we're introducing here is an `InjectorMomentum`. So, it only outputs a momentum vector as in WarpX the coordinates of the particle distribution are specified using other parameters, so it's impractical to introduce correlations. Generating momenta in a uniform sphere on the other hand is easy.

cemitch99: Understood, thanks.  In that case, I don't see a direct analogy with anything in Impact-T, but I agree with dpgrote that having a distribution that is isotropic in momentum space would be more relevant physically.  One option could be to start with a uniformly-populated sphere in momentum space, and then apply a linear transformation to allow user-specified correlations between px,py,pz if you want them (ie, an ellipsoid).

titoiride:
Reporting here for the records a conversation happened with @n01r and @cemitch99:
>We identified that we actually want two new new momentum injectors where one is a more unphysical testing distribution like it is implemented in this PR right now. This one is for Lagrangian sampling of a beam phase space which can often help us with simulations where we would, e.g., like to test various beam starting positions and momenta in just one simulation with non-interacting particles.
Such a distribution is uniformly sampled in limits for each momentum component, and lacks correlation between the components or with position.
On the other hand, a real waterbag implementation can be implemented in a second pull request later and could be used to model space-charge dominated beams of low emittance. Here, the momentum components are correlated and we would sample uniformly in a 3D ball in momentum space and then later possibly also correlate this with positions if needed.
To avoid confusion, we decided to change the naming of the new class (and of this PR) to `InjectorMomentumUniform`

Source/Initialization/PlasmaInjector.cpp
```diff
@@ -639,6 +639,22 @@ void PlasmaInjector::parseMomentum (amrex::ParmParse& pp)
...
+        amrex::Real uy_max = 0._rt;
+        amrex::Real uz_max = 0._rt;
+        utils::parser::queryWithParser(pp, "ux_min", ux_min);
+        utils::parser::queryWithParser(pp, "uy_min", uy_min);
+        utils::parser::queryWithParser(pp, "uz_min", uz_min);
+        utils::parser::queryWithParser(pp, "ux_max", ux_max);
+        utils::parser::queryWithParser(pp, "uy_max", uy_max);
+        utils::parser::queryWithParser(pp, "uz_max", uz_max);
+        // Construct InjectorMomentum with InjectorMomentumUniform.
+        h_inj_mom.reset(new InjectorMomentum((InjectorMomentumUniform*)nullptr,
```
n01r, line 656: @ax3l Should we check before this line that for all directions `min` is never larger than `max`?

n01r (approved):
Thanks a lot @titoiride ✨

Source/Initialization/PlasmaInjector.cpp
titoiride: The distribution is symmetrical and works either way. If `min` is larger than `max`, the momentum is still going to be generated between `min` and `max` (or `max` and `min`). It should also work if they're equal, in which case it collapses onto a delta.

n01r: Ah, missed that 👍 thanks :)

titoiride:
@ax3l @n01r The test is implemented and hopefully should work. The CI results in the error `ERROR: Benchmark and plotfile checksum have different outer keys:`.
The way I structured the test in the file is that it reads the distribution generated and perform a statistical analysis comparing to the analytical value, *i.e.*, it checks that it is within 3\sigma (~99.9% certainty) from the expected value. In that sense, it does not require any external value or file to counter-check, it is in fact self-consistent. I'm not sure if the problem is that the CI expects some kind of confirmation value. In any case, let me know if you can take a look in Examples/Tests/initial_distribution/analysis_distribution.py

ax3l (approved):
Thank you, Davide! :tada:

ax3l:
Follow-ups we discussed:
- a separate distribution for waterbag as in ImpactX (no volunteer yet)
- @dpgrote mentioned we need to update the PICMI interfaces to expose this to the Python layer

PR #3699, opened 2023-02-17
Add `write_species` flag to picmi `FieldDiagnostic`

roelof-groenewald:
The `warpx_write_species` flag sets whether particle data should be saved along with field data.

### DIFF
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -1699,6 +1699,9 @@ class FieldDiagnostic(picmistandard.PICMI_FieldDiagnostic, WarpXDiagnosticBase):
     warpx_plot_raw_fields_guards: bool, optional
         Flag whether the raw fields should include the guard cells
 
+    warpx_write_species: bool, optional
+        Flag whether to output particle data with the diagnostic
+
     warpx_format: {plotfile, checkpoint, openpmd, ascent, sensei}, optional
         Diagnostic file format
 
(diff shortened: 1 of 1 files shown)


dpgrote:
This is Ok as is and I can approve it. However, there is already the separate `ParticleDiagnostic`. To be rigorous, perhaps `FieldDiagnostic` should set `write_species` to false and `ParticleDiagnostic` set it to true?

roelof-groenewald:
@dpgrote I agree it would be better in the way you suggested, but seeing as the default up to now has been that `FieldDiagnostic` outputs the particle data, it would be a breaking change for users that were using it to get both field and particle data. It would presumably also cause several CI tests to break that had a `FieldDiagnostic` installed but also compared the particle checksum values to known values. The motivation for this PR is to have the option to save time and disk-space in simulations where we don't care about the particles. Maybe we should merge this now, and make an issue to remind us later to switch the default to `write_species=False`?

dpgrote:
@roelof-groenewald Yes, that sounds good. I can approve and merge this PR.

dpgrote (approved):
Thanks!

roelof-groenewald:
> @dpgrote I agree it would be better in the way you suggested, but seeing as the default up to now has been that `FieldDiagnostic` outputs the particle data, it would be a breaking change for users that were using it to get both field and particle data. It would presumably also cause several CI tests to break that had a `FieldDiagnostic` installed but also compared the particle checksum values to known values. The motivation for this PR is to have the option to save time and disk-space in simulations where we don't care about the particles. Maybe we should merge this now, and make an issue to remind us later to switch the default to `write_species=False`?

Just wanted to close the above discussion by linking #3745, in which the defaults are fixed.

PR #3705, opened 2023-02-21
Fix getBulkMomentum for Maxwellian distributions

dpgrote:
This fixes a bug that I noticed in the `getBulkMomentum` routine for the Maxwellian distributions. The gamma was being calculated incorrectly. Also, it adds the "_rt" to several constants.


ax3l (approved):

ax3l:
Ooops, good catch!
Interesting that no test uses this function and fails now.

ax3l:
@RemiLehe checked: only used in "applyBallisticCorrection" - which is unused in MW-Jüttner in the test that is checked in

PR #3708, opened 2023-02-22
Small fixes for Coulomb collisions

dpgrote:
There are two equations that use the expression (gamma - 1), which is problematic when gamma is close to one (for nonrelativistic particles). Both can be rewritten to avoid this subtraction. This is a small issue and is unlikely to have a significant effect on most simulations, but there is no reason to do it correctly. In some tests with a nonrelativistic plasma, without this fix there was a very small change in the total energy, showing heating an order of magnitude or so above machine precision after several hundred time steps. With this fix, the heating goes away.

### DIFF
--- a/Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
+++ b/Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
@@ -77,8 +77,10 @@ void UpdateMomentumPerezElastic (
     T_PR p1sz;
     if ( vcms > std::numeric_limits<T_PR>::min() )
     {
+        /* lorentz_tansform_factor = ( (gc-1.0)/vcms*vcDv1 - gc )*m1*g1;
+         * Rewrite to avoid roundoff problems when gc is close to 1 */
         T_PR const lorentz_tansform_factor =
-            ( (gc-T_PR(1.0))/vcms*vcDv1 - gc )*m1*g1;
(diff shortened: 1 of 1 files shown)


EZoni (approved):
Thank you, @dpgrote! This fix looks good to me.

Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
```diff
@@ -77,8 +77,10 @@ void UpdateMomentumPerezElastic (
     T_PR p1sz;
     if ( vcms > std::numeric_limits<T_PR>::min() )
     {
+        /* lorentz_tansform_factor = ( (gc-1.0)/vcms*vcDv1 - gc )*m1*g1;
```
PhilMiller, line 80: Typo 'transform'

Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
```diff
@@ -77,8 +77,10 @@ void UpdateMomentumPerezElastic (
...
+         * Rewrite to avoid roundoff problems when gc is close to 1 */
```
PhilMiller, line 81: To be really precise, it's an issue of cancellation, not rounding

PhilMiller:
I'll submit a follow-on PR for those notes

PhilMiller:
#3770

PR #3709, opened 2023-02-22
PICMI Boundary: Silver-Mueller

### LINKED ISSUE #3702: Absorbing Silver-Mueller not yet in PICMI
@anecas1979 found that silver-mueller bondaries are not yet available in our Python PICMI interface.

ax3l:
Add support for silver-mueller boundary conditions in PICMI.

Documentation in the standard: https://github.com/picmi-standard/picmi/pull/90

Close #3702

### DIFF
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -23,6 +23,7 @@ picmistandard.register_codename(codename)
 # dictionary to map field boundary conditions from picmistandard to WarpX
 BC_map = {
     'open':'pml', 'dirichlet':'pec', 'periodic':'periodic', 'damped':'damped',
+    'absorbing_silver_mueller':'absorbing_silver_mueller',
     'neumann':'neumann', 'none':'none', None:'none'
 }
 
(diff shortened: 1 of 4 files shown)


Python/pywarpx/picmi.py
```diff
@@ -23,6 +23,7 @@
 # dictionary to map field boundary conditions from picmistandard to WarpX
 BC_map = {
     'open':'pml', 'dirichlet':'pec', 'periodic':'periodic', 'damped':'damped',
+    'silver_mueller':'absorbing_silver_mueller',
```
ax3l: Let me know if we shall keep the long name:
```suggestion
    'absorbing_silver_mueller':'absorbing_silver_mueller',
```

dpgrote: I would use the long name to be consistent with the name used in input files, since (at least for now) this is only used in WarpX (and not yet supported in standard PICMI).

ax3l: But I want to add it to PICMI :)
See https://github.com/picmi-standard/picmi/pull/90

Otherwise it would not be well documented.

ax3l: Updated for longer name.

dpgrote (approved):
This looks good to me.

Issue #1943, opened 2021-05-05
`warpx.use_filter` produces strange results in RZ FDTD mode


RemiLehe:
It seems that using `warpx.use_filter=1` produces strange results, for the RZ **FDTD** solver. See for instance the snapshot for a laser-wakefield simulation below:
![Screen Shot 2021-05-05 at 1 21 59 PM](https://user-images.githubusercontent.com/6685781/117204242-e5686d00-ada4-11eb-9746-f7b186c2d54f.png)
(note: labels are swapped; this shows the charge density `rho`)

Maybe we should **add an assert** to prevent users from selecting this option.

Note by @EZoni:
The CI test `LaserAccelerationRZ` has `warpx.use_filter = 1` in this configuration (RZ FDTD).
This would need to be modified, with subsequent benchmark reset, if we add an assert as above.

EZoni:
Is there an input file that we can use to reproduce this issue?

dpgrote:
In PR #2031, filtering is turned off by default with RZ FDTD. If filtering is turned on with RZ FDTD, a warning will be printed since it has not been implemented. Can this issue be closed or left open as a reminder that filtering in not implemented in RZ FDTD?

ax3l:
Until we can look into this properly, #2775 will add an abort for now.

roelof-groenewald:
When adding PEC boundary handling to Cartesian geometry cases (see above link), strange behavior was observed when applying the boundary correction before filtering and therefore the boundary handling had to be done after filtering. The RZ volume and boundary correction is applied before filtering. Should that also be moved to happen afterwards instead?

PR #3711, opened 2023-02-23
Reflect density from non-periodic boundaries during deposition

roelof-groenewald:
As @clarkse pointed out, during charge and current deposition density is lost from particles that are close to non-periodic boundaries. This can be seen in the plots below for example, where a uniform density plasma was created (in 1D) with a set drift velocity. Dirichlet boundaries were used in this case.

charge
current

With the present changes uniform charge and current density is obtained in the full simulation domain.
after_current

Still to do:
- [x] Fix failing CI tests


dpgrote:
Could this be done at the grid level? See for example `ApplyInverseVolumeScalingToCurrentDensity` which does this at the r=0 axis for RZ geometry.

roelof-groenewald:
@dpgrote thanks for looking at the changes. My initial thought was also to do it on the field data rather than with individual particles by using a sort of `ApplyRhoBoundary()` function, so I'm certainly open to switching to that approach. The reason I decided to go this route is primarily due to the fact that the node on the domain edge basically has its value doubled. Nodes in the ghost cells could be robustly handled by removing the deposited charge on those nodes and adding it to the appropriate nodes in the domain, but that isn't the case for the boundary nodes. Consequently calling such an `ApplyRhoBoundary()` function more than once would result in incorrect results and therefore it adds complexity in ensuring that all depositions are accompanied by exactly 1 call to `ApplyRhoBoundary()`. The current approach is more robust in that the deposition is fully completed for every particle (except in the RZ case).
That being said, the RZ approach, as you mentioned, uses a post-deposition correction so doing the same for Cartesian geometries is probably okay too. Is there a reason why correcting after deposition would be a preferred approach here?

ax3l:
Thank you for catching and fixing this @roelof-groenewald.

As @dpgrote suggests, moving the correction to be a post-deposition grid function might be better. As a rationale, the charge deposition is our top cost function in the PIC loop, any runtime complexity that we can spare from it is good.

I don't see a bit issue that the correction can only be called once (per step), that is the case for many other algorithmic sub-routines as well :)

Source/BoundaryConditions/WarpXFieldBoundaries.cpp
```diff
@@ -76,3 +76,8 @@ void WarpX::ApplyBfieldBoundary (const int lev, PatchType patch_type, DtType a_d
         }
     }
 }
+
+void WarpX::ApplyRhofieldBoundary (const int lev, MultiFab* rho)
+{
+    if (PEC::isAnyBoundaryPEC()) PEC::ApplyPECtoRhofield(rho, lev);
```
roelof-groenewald: We should make sure the fine vs coarse patches are properly treated.

roelof-groenewald:
After a discussion with Jean-Luc and @dpgrote, the guard cell values are now also corrected based on the motion of image charges. Furthermore, the handling of particle density that crosses a PEC boundary now depends on the particle boundary type used. If an absorbing boundary is used, the particle density in cells next to conducting walls are reduced by the image charge density in the conductor. If reflecting boundaries are used, the particle density is increased by the portion of density that extends over the boundary (the density is reflected). The result of the deposition logic is shown the plot below where the left boundary was set to absorb and the right boundary to reflect.
after_Jfield

Source/BoundaryConditions/WarpX_PEC.H
```diff
@@ -295,6 +309,343 @@ using namespace amrex;
...
+                if (!( (iside == 0 )
+                    ? is_boundary_PEC(fbndry_lo, idim)
+                    : is_boundary_PEC(fbndry_hi, idim) )) continue;
+
+                // mirror guard cell outside the domain by "ig" number cells, in direction, idim
+                const int ig = ( (iside == 0)
+                                ? (ijk_vec[idim] - dom_lo[idim])
+                                : (dom_hi[idim] + is_nodal[idim] - ijk_vec[idim]) );
+
+                if (ig > 0 && ig <= nguard[idim]) {
```
dpgrote: What happens if the grid is not nodal when ig == 0? I think the update should be done since the data point is at the cell center (and not on the boundary). Perhaps this could be `if (ig >= is_nodal[idim] && ig <= nguard[idim])`?

Source/BoundaryConditions/WarpX_PEC.H
```diff
@@ -295,6 +309,343 @@ using namespace amrex;
...
+
+                    // Mirror location outside the domain by "ig" number of cells
+                    // across PEC boundary in direction, idim, and side, iside
+                    ijk_mirror[idim] = ( (iside == 0)
+                                     ? (dom_lo[idim] - ig)
```
dpgrote: I think this needs to be `(dom_lo[idim] - (1 - is_nodal[idim]) - ig)` since when the grid is not nodal, the mirror to `ig=0` is -1 (with dom_lo == 0).

Source/BoundaryConditions/WarpX_PEC.H
roelof-groenewald: Thanks! I didn't pay very much attention to the actual case for non-nodal fields here since `rho` is always nodal.

Source/BoundaryConditions/WarpX_PEC.H
roelof-groenewald: Thanks for catching this! Basically the same problem is present in the current E and B field updates, so I fixed those as well.

ax3l (commented):
Implementation-architecture-wise, looks good to me :+1: 

@roelof-groenewald please go ahead resetting the benchmarks.

roelof-groenewald:
@MaxThevenet the blinear filter test (which I believe you added) is currently failing for this PR. Would you mind taking a look and seeing if that failure is something we should worry about or if it is something we can reset/adjust?

ax3l:
Ref.:
- https://github.com/ECP-WarpX/WarpX/blob/4d790b0802b3e70e1aeaf7518381795856e66ac0/Regression/WarpX-tests.ini#L353-L367
- https://github.com/ECP-WarpX/WarpX/blob/development/Examples/Tests/single_particle/inputs_2d
- https://github.com/ECP-WarpX/WarpX/blob/4d790b0802b3e70e1aeaf7518381795856e66ac0/Examples/Tests/single_particle/analysis_bilinear_filter.py
  - maybe reset lines 29 and 30?

roelof-groenewald:
@ax3l it seems the `bilinear_filter` test failure is actually pointing out a real problem with the changes in this PR when filtering is used. Running the test locally, I see something like a race-condition at play. When using 2 procs (like the CI test) the result of the test changes from run to run. Sometimes the `Jx` data has overflow values while other times the `Jz` data has overflows. This does not happen when running on 1 proc or with filtering turned off. Any ideas what could be going on?

roelof-groenewald:
@RemiLehe your suggestion of applying the PEC boundary to rho and J after `syncRho` and `syncCurrent` did fix the failure in the `bilinear_filter` test. 

I'm not sure if all the simulation configurations are currently covered, since the way things are laid out the application of the boundary is done in different places for the standard E&M case versus the electrostatic case versus diagnostic cases, so please keep an eye out for that during review.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -971,8 +984,18 @@ WarpX::PushParticlesandDepose (int lev, amrex::Real cur_time, DtType a_dt_type,
...
+// #else
+//         if (rho_fp[lev].get()) {
+//             // Reflect density over PEC boundaries, if needed.
+//             WarpX::GetInstance().ApplyRhofieldBoundary(lev, rho_fp[lev].get());
+//         }
+//         // Set current density at PEC boundaries, if needed.
+//         WarpX::GetInstance().ApplyJfieldBoundary(
+//             lev, current_fp[lev][0].get(), current_fp[lev][1].get(),
+//             current_fp[lev][2].get()
+//         );
```
roelof-groenewald: I left this commented out code here as a reminder that currently the boundary handling for cartesian grids are not matching the RZ handling (done in the `ApplyInverseScalingToChargeDensity` function). The Cartesian grid code had to be moved from here to after the application of the filter to avoid incorrect results. Might this be related to why the filter does not work in RZ (see #1943)?

Source/FieldSolver/ElectrostaticSolver.cpp
```diff
@@ -228,6 +229,12 @@ WarpX::AddSpaceChargeFieldLabFrame ()
     }
 #endif
     SyncRho(); // Apply filter, perform MPI exchange, interpolate across levels
+#ifndef WARPX_DIM_RZ
+    for (int lev = 0; lev <= max_level; lev++) {
+        // Reflect density over PEC boundaries, if needed.
+        ApplyRhofieldBoundary(lev, rho_fp[lev].get());
+    }
+#endif
```
roelof-groenewald: Here is another instance of the RZ and Cartesian boundary handling being treated differently - in RZ the boundary is handled before `SyncRho()` but in Cartesian it is done afterwards since filtering is done in `SyncRho()`.

Source/Evolve/WarpXEvolve.cpp
EZoni: @roelof-groenewald 
Thank you, I think this is a good point. Would you like to add a similar comment directly to #1943, just to keep track of your idea about why the RZ issue might be happening?

Source/WarpX.H
```diff
@@ -656,6 +656,15 @@ public:
...
 
+    /**
+     * \brief Depending on the boundary conditions used, the charge and current
+     * density deposited in the guard cells may have to be reflected back into
+     * the simulation domain. These function perform that logic.
+     */
+    void ApplyRhofieldBoundary (const int lev, amrex::MultiFab* Rho);
+    void ApplyJfieldBoundary (const int lev, amrex::MultiFab* Jx,
+                              amrex::MultiFab* Jy, amrex::MultiFab* Jz);
+
```
EZoni: Thank you for adding this docstring. If you look at the Doxygen output, the docstring shows up only for the first function, though. Could you add a second docstring for the second function, `ApplyJfieldBoundary`, so that both functions are equally documented in our Doxygen reference?

Source/FieldSolver/ElectrostaticSolver.cpp
```diff
@@ -228,6 +229,12 @@ WarpX::AddSpaceChargeFieldLabFrame ()
     }
 #endif
     SyncRho(); // Apply filter, perform MPI exchange, interpolate across levels
+#ifndef WARPX_DIM_RZ
+    for (int lev = 0; lev <= max_level; lev++) {
```
EZoni: Should we use the current finest level available here (so something like `warpx.finestLevel()` if `warpx` is an instance of the WarpX object), instead of `max_level`, in order to be more robust with respect to cases where the two don't match (and thus avoid potential segfaults)?

Source/Evolve/WarpXEvolve.cpp
EZoni: Also, on top of these code lines that are commented out, could you maybe add a brief inline comment before them, explaining why these lines are commented out here, basically repeating exactly what you described in your GitHub comment above?

Source/BoundaryConditions/WarpX_PEC.H
```diff
@@ -295,6 +309,331 @@ using namespace amrex;
...
+                if (!( (iside == 0 )
+                    ? is_boundary_PEC(fbndry_lo, idim)
+                    : is_boundary_PEC(fbndry_hi, idim) )) continue;
+
+                // mirror guard cell outside the domain by "ig" number cells, in direction, idim
+                const int ig = ( (iside == 0)
+                                ? (ijk_vec[idim] - dom_lo[idim])
+                                : (dom_hi[idim] + is_nodal[idim] - ijk_vec[idim]) );
+
+                if (ig >= is_nodal[idim]  && ig <= nguard[idim]) {
```
EZoni: Could you clarify to me why we need to check whether `ig` is greater than or equal to 0 or 1, depending on whether the field is cell-centered or nodal along `idim`? I might be missing what `ig` exactly represents here.

Source/BoundaryConditions/WarpX_PEC.H
```diff
@@ -295,6 +309,331 @@ using namespace amrex;
...
+            ijk_mirror = ijk_vec;
+            // Loop over sides, iside = 0 (lo), iside = 1 (hi)
+            for (int iside = 0; iside < 2; ++iside) {
+                // check if the boundary in this direction is a PEC, otherwise
+                // continue to the next direction
+                if (!( (iside == 0 )
+                    ? is_boundary_PEC(fbndry_lo, idim)
+                    : is_boundary_PEC(fbndry_hi, idim) )) continue;
+
+                // mirror guard cell outside the domain by "ig" number cells, in direction, idim
```
EZoni:
```suggestion
                // mirror guard cell outside the domain by "ig" number of cells, in direction, idim
```

Source/BoundaryConditions/WarpX_PEC.H
```diff
@@ -295,6 +309,331 @@ using namespace amrex;
...
+                               amrex::GpuArray<int, 3> const& fbndry_lo,
+                               amrex::GpuArray<int, 3> const& fbndry_hi)
+    {
+        amrex::IntVect ijk_mirror = ijk_vec;
+        bool OnPECBoundary = false;
+        bool GuardCell = false;
+        // Loop over all the dimensions
+        for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
+            // Loop over sides, iside = 0 (lo), iside = 1 (hi)
+            // Loop over sides, iside = 0 (lo), iside = 1 (hi)
```
EZoni:
```suggestion
```

Source/BoundaryConditions/WarpX_PEC.H
```diff
@@ -295,6 +309,331 @@ using namespace amrex;
...
+            for (int iside = 0; iside < 2; ++iside) {
+                // check if the boundary in this direction is a PEC, otherwise
+                // continue to the next direction
+                if (!( (iside == 0 )
+                    ? is_boundary_PEC(fbndry_lo, idim)
+                    : is_boundary_PEC(fbndry_hi, idim) )) continue;
+
+                // guard cell outside the domain by "ig" number cells, in direction, idim
```
EZoni:
```suggestion
                // guard cell outside the domain by "ig" number of cells, in direction, idim
```

Source/Evolve/WarpXEvolve.cpp
roelof-groenewald: Sure thing!

Source/BoundaryConditions/WarpX_PEC.cpp
```diff
@@ -207,7 +206,254 @@ PEC::ApplyPECtoBfield (std::array<amrex::MultiFab*, 3> Bfield, const int lev,
...
+#pragma omp parallel if (amrex::Gpu::notInLaunchRegion())
+#endif
+    for (amrex::MFIter mfi(*rho, amrex::TilingIfNotGPU()); mfi.isValid(); ++mfi) {
+
+        // Extract field data
+        amrex::Array4<amrex::Real> const& rho_array = rho->array(mfi);
+
+        // Construct a tilebox to loop over the grid
+        tilebox = mfi.tilebox();
+        amrex::Box tb = convert( tilebox, rho_nodal );
```
EZoni: Not sure, but you might be able to do this in one line without the temporary `tilebox`:
```cpp
amrex::Box tb = convert(mfi.tilebox(), rho_nodal);
```

EZoni (commented):
Thank you for the PR, @roelof-groenewald! I left a few minor comments.

Source/FieldSolver/ElectrostaticSolver.cpp
roelof-groenewald: I followed the logic on line 208 above, where `max_level` is also used. I'll replace that with `finestLevel()` as well.

EZoni: Cc: @RemiLehe, @ax3l, @dpgrote, @WeiqunZhang 
My understanding is that `finest_level()` should be more robust and/or correct, but I might be wrong.
Others please feel free to chime in, too, on how these MR loops should be handled in the most general case.

Source/BoundaryConditions/WarpX_PEC.H
roelof-groenewald: `ig` gives the mirrored guard cell number. For nodal fields, the 0'th guard cell is on the boundary, so sort of both inside and outside the domain but we don't need to do anything to it since in `SetRhoFieldOnPEC` it will be set to 0. For cell-centered fields the 0'th ghost cell is the cell center next to the PEC (outside the domain) and so it mirrors the first valid entry in the domain, and we do need to reflect it's density back into the domain. Does that clear things up?

EZoni: Thank you, it does. Maybe it could be worth adding a similar inline comment (about the differences between the nodal and cell-centered cases) in the various functions, wherever `ig` is defined or wherever you think it's most appropriate, just to make it easier to read through the code. What do you think?

roelof-groenewald: Yeah, that sounds good to me. I'll do so!

Source/BoundaryConditions/WarpX_PEC.H
```diff
@@ -295,6 +309,346 @@ using namespace amrex;
...
+            ijk_mirror = ijk_vec;
+            // Loop over sides, iside = 0 (lo), iside = 1 (hi)
+            for (int iside = 0; iside < 2; ++iside) {
+                // check if the boundary in this direction is a PEC, otherwise
+                // continue to the next direction
+                if (!( (iside == 0 )
+                    ? is_boundary_PEC(fbndry_lo, idim)
+                    : is_boundary_PEC(fbndry_hi, idim) )) continue;
+
+                // Mirror guard cell outside the domain by "ig" number cells,
```
roelof-groenewald:
```suggestion
                // Mirror guard cell outside the domain by "ig" number of cells,
```

Source/BoundaryConditions/WarpX_PEC.H
```diff
@@ -295,6 +309,346 @@ using namespace amrex;
...
+            // Loop over sides, iside = 0 (lo), iside = 1 (hi)
+            // Loop over sides, iside = 0 (lo), iside = 1 (hi)
+            for (int iside = 0; iside < 2; ++iside) {
+                // check if the boundary in this direction is a PEC, otherwise
+                // continue to the next direction
+                if (!( (iside == 0 )
+                    ? is_boundary_PEC(fbndry_lo, idim)
+                    : is_boundary_PEC(fbndry_hi, idim) )) continue;
+
+                // Mirror guard cell outside the domain by "ig" number cells,
```
roelof-groenewald:
```suggestion
                // Mirror guard cell outside the domain by "ig" number of cells,
```

Source/BoundaryConditions/WarpX_PEC.H
```diff
@@ -295,6 +309,346 @@ using namespace amrex;
...
+                // For 2D : for icomp==1, (Ey in XZ, Etheta in RZ),
+                //          icomp=1 is tangential to both x and z boundaries
+                //          The logic below ensures that the flags are set right for 2D
+                const bool is_tangent_to_PEC = ( (icomp == AMREX_SPACEDIM*idim)
+                                             ? false : true );
+#else
+                const bool is_tangent_to_PEC = ( ( icomp == idim) ? false : true );
+#endif
+
+                // Mirror guard cell outside the domain by "ig" number cells, in
```
roelof-groenewald:
```suggestion
                // Mirror guard cell outside the domain by "ig" number of cells, in
```

Source/BoundaryConditions/WarpX_PEC.H
```diff
@@ -295,6 +309,346 @@ using namespace amrex;
...
+                // For 2D : for icomp==1, (Ey in XZ, Etheta in RZ),
+                //          icomp=1 is tangential to both x and z boundaries
+                //          The logic below ensures that the flags are set right for 2D
+                const bool is_tangent_to_PEC = ( (icomp == AMREX_SPACEDIM*idim)
+                                             ? false : true );
+#else
+                const bool is_tangent_to_PEC = ( ( icomp == idim) ? false : true );
+#endif
+
+                // Mirror guard cell outside the domain by "ig" number cells,
```
roelof-groenewald:
```suggestion
                // Mirror guard cell outside the domain by "ig" number of cells,
```

Examples/Physics_applications/capacitive_discharge/analysis_2d.py
```diff
@@ -15,5 +15,5 @@
 
 my_check = checksumAPI.evaluate_checksum(
     'background_mcc', 'Python_background_mcc_plt000050',
-    do_particles=True, rtol=3.5e-3
+    do_particles=True, rtol=4e-3
```
roelof-groenewald: I'm not sure why this test is now failing but it is probably okay to loosen the tolerance a little bit more.
```suggestion
    do_particles=True, rtol=5e-3
```

Source/BoundaryConditions/WarpX_PEC.cpp
```diff
@@ -207,7 +206,253 @@ PEC::ApplyPECtoBfield (std::array<amrex::MultiFab*, 3> Bfield, const int lev,
...
+        fbndry_hi[idim] = WarpX::field_boundary_hi[idim];
+        pbndry_lo[idim] = WarpX::particle_boundary_lo[idim];
+        pbndry_hi[idim] = WarpX::particle_boundary_hi[idim];
     }
+    amrex::IntVect rho_nodal = rho->ixType().toIntVect();
+    amrex::IntVect ng_fieldgather = rho->nGrowVect();
+
+    const int nComp = rho->nComp();
+
+    amrex::Box tilebox;
```
roelof-groenewald:
```suggestion
```

Source/BoundaryConditions/WarpXFieldBoundaries.cpp
EZoni: The level info is currently used only to extract the correct domain box for a given level. Do we need to do more in order to treat refined patches correctly? If that's the case, it might be worth adding a warning or something like that.

Source/BoundaryConditions/WarpX_PEC.H
```diff
@@ -275,8 +289,8 @@ using namespace amrex;
                         // Mirror location inside the domain by "ig" number of cells
                         // across PEC boundary in direction, idim, and side, iside
                         ijk_mirror[idim] = ( (iside == 0)
-                                       ? (dom_lo[idim] + ig)
-                                       : (dom_hi[idim] + is_nodal[idim] - ig));
+                                        ? (dom_lo[idim] + ig - (1 - is_nodal[idim]))
+                                        : (dom_hi[idim] + 1 - ig));
```
EZoni, line 321: Why does the second line (the high side) not take into account the staggering along `idim` (here and in the similar code lines above too)?

I would have expected something like the following, if we are mirroring _inside_ the domain (which seems to be the case, since we are adding, not subtracting, `ig` in the low side case, but I may be missing something):
```suggestion
                                        : (dom_hi[idim] + (1 - is_nodal[idim]) - ig));
```
I'm thinking of situations like the following cases (and thinking of a case where `ig=1`):
```
low side nodal:
--o-----[-----o--
        0

low side staggered:
--|--x--[--x--|--
           0

high side nodal:
--o-----]-----o--
        n

high side staggered:
--|--x--]--x--|--
    n-1
```

roelof-groenewald: Thanks for checking and thinking about these indices. I think you are right that there is still a bug here. The asymmetry between the lo and hi sides incorporating the nodal flag in the highlighted code comes from the fact that the nodal flag is used when calculating `ig` for the hi case but not for the low case. If you plug the calculations for `ig` into the calculation of the mirror cells you get:

lo case: `ijk_mirror = dom_lo - (1 - is_nodal) - (dom_lo - ijk_vec)`
hi case: 'ijk_mirror = dom_hi + (1 + is_nodal) - (ijk_vec - dom_hi)`

which shows the current treatment of the two cases more clearly. I think the fix is to change the calculation of `ig` in the hi case to just be `ig = ijk_vec[idim] - dom_hi[idim]` and add your suggested change. I'll plot the field values to confirm the fix.

roelof-groenewald: Looking at it again, actually for nodal fields the high side boundary is at cell `dom_hi + 1` (so n+1 in your diagram), that is why we need to compare to   `dom_hi + is_nodal`.

roelof-groenewald: So basically as is, the high side mirror calculation can be written as:
`ijk_mirror = (dom_hi + is_nodal) + (1 - is_nodal) - [ijk_vec - (dom_hi + is_nodal)]`
which I believe is correct.

Source/BoundaryConditions/WarpXFieldBoundaries.cpp
roelof-groenewald: Thank you for reminding me about this. I think with the latest commit we now correctly handle the coarse versus fine patches.

ax3l (dismissed):
Thanks a lot, this looks great :+1: 

Update: saw an open comment left...

Source/BoundaryConditions/WarpX_PEC.H
ax3l: @roelof-groenewald, @EZoni is this fixed? :)

Source/FieldSolver/ElectrostaticSolver.cpp
ax3l: You are correct - always only loop over the active levels, which is `finest_level()`.

`max_level` is the maximum allowed level in a simulation, but less levels might be allocated and used at any time.

ax3l:
> @ax3l it seems the bilinear_filter test failure is actually pointing out a real problem with the changes in this PR when filtering is used. Running the test locally, I see something like a race-condition at play. When using 2 procs (like the CI test) the result of the test changes from run to run. Sometimes the Jx data has overflow values while other times the Jz data has overflows. This does not happen when running on 1 proc or with filtering turned off. Any ideas what could be going on?

No immediate idea, let's track in an issue?

roelof-groenewald:
> > @ax3l it seems the bilinear_filter test failure is actually pointing out a real problem with the changes in this PR when filtering is used. 
> 
> No immediate idea, let's track in an issue?

This issue was actually fixed by Remi's suggestion to apply the BC after the filter. So no worries on this.

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -785,6 +783,9 @@ WarpXParticleContainer::GetChargeDensity (int lev, bool local)
 
 #ifdef WARPX_DIM_RZ
     WarpX::GetInstance().ApplyInverseVolumeScalingToChargeDensity(rho.get(), lev);
+#else
+    // Reflect density over PEC boundaries, if needed.
+    WarpX::GetInstance().ApplyRhofieldBoundary(lev, rho.get(), PatchType::fine);
 #endif
 
     if (local == false) { ablastr::utils::communication::SumBoundary(*rho, WarpX::do_single_precision_comms, gm.periodicity()); }
```
dpgrote, line 793: Something I've seen is that the AMReX routines handling the periodicity will only copy data from valid cells to the periodic guard cells. So, if there is a corner where one side is periodic and one is PEC, if the periodicity is handled after the PEC boundary, the data in the corner guard cells won't be set. So perhaps move this line above the lines setting the PEC boundaries.

roelof-groenewald: Thanks Dave! I've updated the order for both this function and the `DepositCharge` function.

Source/BoundaryConditions/WarpX_PEC.H
RevathiJambunathan: Update : no I need to think more about this

RevathiJambunathan: I agree with @roelof-groenewald 
changing the hi condition to `ijk_mirror = (dom_hi + is_nodal) + (1 - is_nodal) - ig`
is general and correct for nodal and cell-centered staggering. 
Thank you @roelof-groenewald !

Source/BoundaryConditions/WarpX_PEC.H
```diff
@@ -295,6 +309,345 @@ using namespace amrex;
...
+                                ? (ijk_vec[idim] - dom_lo[idim])
+                                : (dom_hi[idim] + is_nodal[idim] - ijk_vec[idim]) );
+
+                if (ig >= is_nodal[idim] && ig <= nguard[idim]) {
+
+                    // Mirror location outside the domain by "ig" number of cells
+                    // across PEC boundary in direction, idim, and side, iside
+                    ijk_mirror[idim] = ( (iside == 0)
+                                     ? (dom_lo[idim] - (1 - is_nodal[idim]) - ig)
+                                     : (dom_hi[idim] + is_nodal[idim] + ig));
```
RevathiJambunathan: would the same suggestions/fixes for ijk_mirror from above apply here as well?

RevathiJambunathan: or may be not, becuase ijk_mirror here is outside the domain, 
unlike in PEC for E-field, where ijk_mirror was inside the domain
Is that right?

roelof-groenewald: No, you are right, this should be `dom_hi[idim] + 1 + ig`. It doesn't show up as a problem since rho is always nodal. But for the sake of accuracy I'll update this to be consistent with the other cases.

Source/BoundaryConditions/WarpX_PEC.H
```diff
@@ -295,6 +309,345 @@ using namespace amrex;
...
+                                ? (ijk_vec[idim] - dom_lo[idim] + (1 - is_nodal[idim]))
+                                : (dom_hi[idim] + 1 - ijk_vec[idim]) );
+
+                if (ig > 0 && ig <= nguard[idim]) {
+
+                    // Mirror location outside the domain by "ig" number of cells
+                    // across PEC boundary in direction, idim, and side, iside
+                    ijk_mirror[idim] = ( (iside == 0)
+                                     ? (dom_lo[idim] - ig)
+                                     : (dom_hi[idim] + ig + is_nodal[idim]));
```
RevathiJambunathan: @roelof-groenewald here as well right? 
`dom_hi[idim] + 1 + ig`

roelof-groenewald: I just pushed a change where the `ig` and `ijk_mirror` calculation is made uniform for all the functions in `WarpX_PEC.H`, so hopefully that will take care of all those instances.

RevathiJambunathan (approved):
Thank you for this PR @roelof-groenewald !

RevathiJambunathan:
https://github.com/ECP-WarpX/WarpX/issues/3833 
From @dpgrote's message offline to ensure this also works with RZ.

PR #3720, opened 2023-03-02
PSATD: call FFT of rho with explicit spectral index

EZoni:
Make the spectral index of rho (currently `rho_old` or `rho_new`) more evident when calling `PSATDForwardTransformRho`, replacing the old hidden logic
```cpp
    int dst_comp;
    if (rho_in_time == RhoInTime::Constant)
    {
        dst_comp = Idx.rho_mid;
    }
    else // rho_in_time == RhoInTime::Linear
    {
        dst_comp = (dcomp == 0) ? Idx.rho_old : Idx.rho_new;
    }
```

Note that rho constant in time, which corresponds to the spectral index `rho_mid`, is not yet implemented, so we now abort in case a user sets `psatd.rho_in_time=constant` (which should not happen anyways, because `constant` is not among the possible values listed for `psatd.rho_in_time` in the docs).

### DIFF
--- a/Source/Evolve/WarpXEvolve.cpp
+++ b/Source/Evolve/WarpXEvolve.cpp
@@ -546,6 +546,9 @@ WarpX::OneStep_multiJ (const amrex::Real cur_time)
         "multi-J algorithm not implemented for FDTD"
     );
 
+    const int rho_mid = spectral_solver_fp[0]->m_spectral_index.rho_mid;
+    const int rho_new = spectral_solver_fp[0]->m_spectral_index.rho_new;
+
     // Push particle from x^{n} to x^{n+1}
     //               from p^{n-1/2} to p^{n+1/2}
     const bool skip_deposition = true;
@@ -571,7 +574,7 @@ WarpX::OneStep_multiJ (const amrex::Real cur_time)
         // Filter, exchange boundary, and interpolate across levels
         SyncRho();
         // Forward FFT of rho
-        PSATDForwardTransformRho(rho_fp, rho_cp, 0, 1);
+        PSATDForwardTransformRho(rho_fp, rho_cp, 0, rho_new);
     }
 
     // 4) Deposit J at relative time -dt with time step dt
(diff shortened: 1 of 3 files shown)


Source/Evolve/WarpXEvolve.cpp
```diff
@@ -571,7 +574,7 @@ WarpX::OneStep_multiJ (const amrex::Real cur_time)
         // Filter, exchange boundary, and interpolate across levels
         SyncRho();
         // Forward FFT of rho
-        PSATDForwardTransformRho(rho_fp, rho_cp, 0, 1);
+        PSATDForwardTransformRho(rho_fp, rho_cp, 0, rho_new);
```
EZoni, line 577: This is inside the `if` condition `if (rho_fp[0] && rho_in_time == RhoInTime::Linear)`, thus it is executed only when rho is linear in time, which means that the spectral index `rho_new` is well defined.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -635,7 +638,8 @@ WarpX::OneStep_multiJ (const amrex::Real cur_time)
             // Filter, exchange boundary, and interpolate across levels
             SyncRho();
             // Forward FFT of rho
-            PSATDForwardTransformRho(rho_fp, rho_cp, 0, 1);
+            const int rho_idx = (rho_in_time == RhoInTime::Linear) ? rho_new : rho_mid;
+            PSATDForwardTransformRho(rho_fp, rho_cp, 0, rho_idx);
```
EZoni, line 642: Unlike the previous call, this is executed regardless of whether rho is constant or linear in time, which means that we have to choose the rho spectral index carefully.

Source/WarpX.cpp
```diff
@@ -1209,6 +1209,14 @@ WarpX::ReadParameters ()
...
         rho_in_time = GetAlgorithmInteger(pp_psatd, "rho_in_time");
 
+        if (psatd_solution_type != PSATDSolutionType::FirstOrder || do_multi_J == false)
+        {
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                rho_in_time == RhoInTime::Linear,
+                "psatd.rho_in_time=constant not yet implemented, "
+                "except for psatd.solution_type=first-order and warpx.do_multi_J=1");
+        }
+
```
EZoni, line 1219: The function `PushPSATD`, modified above in Source/FieldSolver/WarpXPushFieldsEM.cpp, is used only without the multi-J algorithm, which means that we have only the spectral indices `rho_old` and `rho_new` since rho constant in time is not implemented without the multi-J algorithm, as this statement asserts.

RemiLehe (approved):

PR #3725, opened 2023-03-02
Moving window: shift G MultiFab w/ div(B) cleaning

EZoni:
Fix bug occuring with moving window and div(B) cleaning: the scalar field G MultiFab must be shifted as all other fields.


RemiLehe (approved):

oshapoval (commented):
Thanks for spotting this!

PR #3726, opened 2023-03-03
RZ FDTD: fix bug w/ div(E) cleaning (extra c2)

EZoni:
There seems to be an extra factor $c^2$ in the imaginary part of RZ modes higher than 0 (imaginary part means fourth component `2*m` for $m \geq 1$) for the F update equation with the FDTD solver.


RemiLehe (approved):
Thanks for catching this!

PR #3732, opened 2023-03-07
fix zmax_plasma_... calculation upon restart

RTSandberg:
This PR fixes the ``max_step`` determined by ``WarpX::computeMaxStepBoostAccelerator`` when performing a restart from a checkpoint file.

Previously, the code on the attached example would output 
``max_step computed in computeMaxStepBoostAccelerator: 513
``
initially, and then upon restart,
``max_step computed in computeMaxStepBoostAccelerator: 488
``
With this fix, the max_step computed upon restart remains the same value of 513

[inputs_2D.txt](https://github.com/ECP-WarpX/WarpX/files/10904149/inputs_2D.txt)

### DIFF
--- a/Source/Initialization/WarpXInitData.cpp
+++ b/Source/Initialization/WarpXInitData.cpp
@@ -379,7 +379,7 @@ WarpX::InitData ()
     // needs to start from the initial zmin_domain_boost,
     // even if restarting from a checkpoint file
     if (do_compute_max_step_from_zmax) {
-      zmin_domain_boost_step_0 = geom[0].ProbLo(WARPX_ZINDEX);
+        zmin_domain_boost_step_0 = geom[0].ProbLo(WARPX_ZINDEX);
     }
     if (restart_chkfile.empty())
     {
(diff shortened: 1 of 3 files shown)


Source/Initialization/WarpXInitData.cpp
```diff
@@ -375,6 +375,12 @@ WarpX::InitData ()
 
     Print() << utils::logo::get_logo();
 
+    // WarpX::computeMaxStepBoostAccelerator
+    // needs to start from the initial zmin_domain_boost,
+    // even if restarting from a checkpoint file
+    if (do_compute_max_step_from_zmax) {
+      zmin_domain_boost_step_0 = geom[0].ProbLo(WARPX_ZINDEX);
+    }
```
EZoni:
```suggestion
    if (do_compute_max_step_from_zmax) {
        zmin_domain_boost_step_0 = geom[0].ProbLo(WARPX_ZINDEX);
    }
```

EZoni (approved):
Thank you for this fix, @RTSandberg! Looks good to me.

PR #3735, opened 2023-03-08
Fix handling of name passed into SetTag

dpgrote:
This PR does some clean up of the handling of the `name` variable that is passed into `SetTag` during the creation of MultiFabs. This fixes a compilation error that arises due to recent changes in AMReX. As Weiqun says, "Top level const as function parameter is bad" so the `name` was changed to `const std::string&`.

Note that PR https://github.com/AMReX-Codes/amrex/pull/3188 in AMReX would also fix the compilation error, but it is good to also clean up the code here.

### DIFF
--- a/Source/WarpX.H
+++ b/Source/WarpX.H
@@ -349,7 +349,7 @@ public:
         const amrex::DistributionMapping& dm,
         const int ncomp,
         const amrex::IntVect& ngrow,
-        const std::string name,
+        const std::string& name,
         std::optional<const amrex::Real> initial_value = {});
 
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
We could change the two `static void AddToMultiFabMap` as well to avoid creating a copy.

ax3l (approved):

ax3l:
Alternatively, we could have used non-const and `std::move()` if we call another destination with it.

WeiqunZhang (approved):

PR #3738, opened 2023-03-08
Deposition: Fix LB Cost Segfault

ax3l:
We sometimes call depositions before the load balance cost arrays are allocated, e.g., during initialization. In this case, we need to avoid writing into invalid memory.

Fix #3721

Introduced in #1406

PR #3739, opened 2023-03-08
Load Balance & Regrid: Cost Guards

ax3l:
This is adding guards if the `cost` array was already defined. Not sure if sensible, seen via similar error-prune patterns in the code.

Related to #3738


atmyers (approved):

atmyers:
I think this is principle the correct thing to do, since `ComputeCostsHeuristic` does assume that `costs` has already been defined. However, I think we want to check `costs[0]`, and then assert that if `costs` are defined on level 0 they are defined on every level.

ax3l:
Looked at the accesses above, I think these locations already assume it's either allocated or checking for lvl 0

PR #3740, opened 2023-03-08
Implement PICMI flag for setting max_step from btd

RemiLehe:
This adds the missing PICMI interface for the parameter `compute_max_step_from_btd`.


Python/pywarpx/picmi.py
```diff
@@ -1489,6 +1489,11 @@ class Simulation(picmistandard.PICMI_Simulation):
     warpx_zmax_plasma_to_compute_max_step: float, optional
         Sets the simulation run time based on the maximum z value
 
+    warpx_compute_max_step_from_btd: integer, optional
```
EZoni: Is this technically a `bool` rather than an `integer`?
```suggestion
    warpx_compute_max_step_from_btd: bool, default=0
```

EZoni (approved):

PR #3741, opened 2023-03-09
Fix figure reference in docs for input and output

n01r:
Fixes the label and linking of Fig. 16 in the docs


ax3l (approved):
Thanks for fixing this! :)

PR #3683, opened 2023-02-08
New user input for grid type (collocated, staggered, hybrid)

EZoni:
- Old input config for hybrid mode (full hybrid, w/ fields and currents centering):
```sh
warpx.do_nodal = 0 #optional
algo.field_gathering = momentum-conserving #necessary, otherwise set to energy-conserving by default
interpolation.field_centering_nox = 8 #necessary, otherwise set to 2 by default
interpolation.field_centering_noy = 8 #necessary, otherwise set to 2 by default
interpolation.field_centering_noz = 8 #necessary, otherwise set to 2 by default
warpx.do_current_centering = 1 #necessary, otherwise set to 0 (false) by default
interpolation.current_centering_nox = 8 #necessary, otherwise set to 2 by default
interpolation.current_centering_noy = 8 #necessary, otherwise set to 2 by default
interpolation.current_centering_noz = 8 #necessary, otherwise set to 2 by default
```
- New input config for hybrid mode (full hybrid, w/ fields and currents centering):
```sh
warpx.grid_type = hybrid #necessary, otherwise set to staggered by default
```
- The new input parameter `warpx.grid_type` can be set to `collocated` (which corresponds to nodal currently, but could be changed to fully cell-centered in the future, e.g. to avoid symmetry breaking), `staggered` (which corresponds to purely staggered), and `hybrid`.
- The old input parameter `warpx.do_nodal` (0/1) was removed.
- If the user does want to tweak details of the hybrid algo, for instance switch off currents centering or change the order of the finite centering of fields and/or currents, all relevant parameters are now grouped under the same namespace `warpx`, for convenience: `warpx.field_centering_no<x,y,z>`, `warpx.current_centering_no<x,y,z>`, `warpx.do_current_centering`.
- The field gathering algo must be momentum-conserving, so fields centering is mandatory for the hybrid mode. On the other hand, currents centering is still optional.

To-do:
- [x] Fix bug: `grid_type` probably still used as `bool` somewhere
  (hybrid tests pass if `GridType::Hybrid` equals `0` instead of `2`)
- [x] Set default algo parameters for hybrid case
- [x] More changes to interface (e.g., all hybrid input parameters under `warpx` namespace)
- [x] Update examples with new user input syntax
- [x] Update documentation
- [x] Improve PR description, add details

PR #3743, opened 2023-03-09
PICMI: add missing flags for latest algorithms

EZoni:
Add missing flags for latest algorithms:
- [x] `warpx.do_multi_J`
- [x] `warpx.do_multi_J_n_depositions`
- [x] `psatd.J_in_time`
- [x] `psatd.rho_in_time`
- [x] `warpx.grid_type`
- [x] `warpx.field_centering_no<x,y,z>`
- [x] `warpx.do_current_centering`
- [x] `warpx.current_centering_no<x,y,z>`

See #3683 for reference about `warpx.grid_type` and other parameters related to the hybrid scheme.

Tested locally.

### DIFF
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -901,12 +901,12 @@ class ElectromagneticSolver(picmistandard.PICMI_ElectromagneticSolver):
     warpx_psatd_do_time_averaging: bool, optional
         Whether to do the time averaging for the spectral solver
 
-    warpx_psatd_J_in_time: string, {'constant', 'linear'}, default='constant'
+    warpx_psatd_J_in_time: {'constant', 'linear'}, default='constant'
         This determines whether the current density is assumed to be constant
         or linear in time, within the time step over which the electromagnetic
         fields are evolved.
 
-    warpx_psatd_rho_in_time: string, {'linear'}, default='linear'
+    warpx_psatd_rho_in_time: {'linear'}, default='linear'
         This determines whether the charge density is assumed to be linear
         in time, within the time step over which the electromagnetic fields are evolved.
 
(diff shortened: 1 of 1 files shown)


Python/pywarpx/picmi.py
```diff
@@ -901,6 +901,15 @@ class ElectromagneticSolver(picmistandard.PICMI_ElectromagneticSolver):
     warpx_psatd_do_time_averaging: bool, optional
         Whether to do the time averaging for the spectral solver
 
+    warpx_psatd_J_in_time: string, {'constant', 'linear'}, default='constant'
```
dpgrote: In other places where there are multiple options like this, the `string` type is not specified. I don't have a strong opinion about this, just noting the inconsistency.

Python/pywarpx/picmi.py
```diff
@@ -1448,6 +1460,46 @@ class Simulation(picmistandard.PICMI_Simulation):
...
+        (fields and currents are interpolated back and forth between a staggered grid
+        and a nodal grid, must be used with momentum-conserving field gathering algorithm.
+
+    warpx_do_current_centering: bool, optional
+        If true, the current is deposited on a nodal grid and then centered
+        to a staggered grid (Yee grid), using finite-order interpolation.
+        Default: warpx.do_current_centering=0 with collocated or staggered grids,
+        warpx.do_current_centering=1 with hybrid grids.
+
+    warpx_field_centering_nox, warpx_field_centering_noy, warpx_field_centering_noz:
```
dpgrote: This might need to be all on one line for the doc generation to be correct. Check the html. Also, the parenthesis are not needed.

Python/pywarpx/picmi.py
EZoni: Sounds good, I removed `string`, even though I personally think it's better to state the type explicitly.

Python/pywarpx/picmi.py
EZoni: Thank you, yes, they needed to be on a single line. Note that I had to use the `/` separator, rather than `,` (comma), in order to have a single entry instead of three repeated entries. I could not find another way around it.

RemiLehe (approved):

Issue #3733, opened 2023-03-07
cudart error when running the code


hnakahara79:
Hi @ax3l ,

I build the code as:  `cmake -S . -B build -DWarpX_DIMS=3`

The executable generated in the `build/bin/` folder was `warpx.3d.MPI.OMP.DP.PDP.OPMD.QED`

I copied the executable in the PWD and run example file of the code to test
```
mpirun -np 80 ./warpx.3d.MPI.OMP.DP.PDP.OPMD.QED inputs_3d
```
I received the following error:
```
########
**No component were able to opened in the pml framework.**
**PML ucx cannot be selected**
**PMIX ERROR: UNREACHABLE in file server/pmix_server.c**
########
```
I used `openmpi/4.1.4`

########

During building of executable, the messages that appear, can be seen in the attachment file.

Thank you and Best regards.

hnakahara79:
Hi @ax3l , 

do you have any suggestions how to fix the above problem.

ax3l:
Hi @hnakahara79 , since your machine is GPU-enabled, use this to build WarPX for CUDA GPUs via `-DWarpX_COMPUTE=CUDA`:

```console
rm -rf build

cmake -S . -B build -DWarpX_DIMS=3 -DWarpX_COMPUTE=CUDA
cmake --build build -j 8
```

The default build is for OpenMP (CPU only).

Changing the compute backend to GPU should solve the runtime error of WarpX as well as the linker warning you see for ADIOS2.

Ref.:
https://warpx.readthedocs.io/en/latest/install/cmake.html#build-options

ax3l:
With regards to the HDF5 warning in your PDF: Did you load HDF5 via a module or did you build it yourself?

By default, HDF5 is non-parallel (has no MPI support). If you like to use HDF5 as an additional openPMD backend (you already have ADIOS2, which is totally sufficient as well), when you just need a HDF5 dependency that enabled MPI "parallel" support.
Maybe your system has another, MPI-parallel HDF5 module that you can load (or ask support to build one for your compilers & MPI)?

**Otherwise** this is how you build HDF5 from source: (load your compilers and MPI modules first!)
```console
# install once
wget https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.12/hdf5-1.12.2/src/hdf5-1.12.2.tar.gz
tar -xvf hdf5-1.12.2.tar.gz

[...]

PR #3744, opened 2023-03-09
Spack: OpenMP w/o GPU

ax3l:
Spack desktop environments:
Make sure that we are explicit about not matching GPU packages if we want to build a pure OpenMP version on a GPU-enabled machine.

X-ref: #3733


ax3l:
the issue was in the end that `LD_LIBRARY_PATH` took precedence over the `RUNPATH` of dependencies - and there was another, CUDA-enabled `libblaspp.so` in `/usr/local/lib` installed. I saw it via `libtree warpx` and removed the stray system-wide install.

RemiLehe (approved):

PR #3745, opened 2023-03-09
[WIP] Change default behavior of `picmi.FieldDiagnostic` and `picmi.ParticleDiagnostic`

RemiLehe:
Right now, with `picmi.FieldDiagnostic` and `picmi.LabFrameFieldDiagnostic`, the default is to write **all particle species** (along with the fields). This is probably not what the user expects (note that there is a separate class -- `picmi.ParticleDiagnostic` -- to be used when writing particles). In addition, this default behavior could be computationally costly, if there is a large number of particles in the simulation.

Similar, with `picmi.ParticleDiagnostic`, the default is to write **all fields**, which is again probably not what the user expects. 

This PR changes the default behavior, so that species are not written by default for `picmi.FieldDiagnostic`, and fields are not written by default for `picmi.ParticleDiagnostic`. It also removes the parameter `warpx_write_species`, since the user can use `picmi.ParticleDiagnostic` instead.


Python/pywarpx/picmi.py
```diff
@@ -1808,8 +1804,7 @@ def initialize_inputs(self):
         self.diagnostic.plot_raw_fields_guards = self.plot_raw_fields_guards
         self.diagnostic.plot_finepatch = self.plot_finepatch
         self.diagnostic.plot_crsepatch = self.plot_crsepatch
-        self.diagnostic.write_species = self.write_species
-
+        self.diagnostic.write_species = False
```
dpgrote: Thanks for this PR. Note that care is needed here since `FieldDiagnostic` and `ParticleDiagnostic` can be used to setup the same diagnostic, e.g. `diag1`. The `ParticleDiagnostic` should set `write_species` to True, and this should only set it to false if it hadn't already been set. Something like this
```
if 'write_species' not in self.diagnostic.argvattrs:
    self.diagnostic.write_species = False
```

Python/pywarpx/picmi.py
```diff
@@ -1683,6 +1683,10 @@ def add_diagnostic(self):
             self.diagnostic = pywarpx.Diagnostics.Diagnostic(
                 self.name, _species_dict={}
             )
+            # By default, disable particle output and field output
+            # (This will be reset in FieldDiagnostic and ParticleDiagnostic)
+            self.diagnostic.fields_to_plot = 'none'
```
dpgrote: This is a good idea, but unfortunately this class is also used by `ReducedDiagnostic`. These two parameters are not used by that class.

RemiLehe: OK, yes, that's true. Would it be fine with you if I add:
```
if not isinstance(self, ReducedDiagnostic):
```
before setting these two attributes?

dpgrote: That's getting a bit kludgy. I think it would be better to set these in the two classes, with the appropriate checks of whether they had already been set. Note that also `LabFrameFieldDiagnostic` doesn't use `write_species`.

Python/pywarpx/picmi.py
```diff
@@ -2067,8 +2065,7 @@ def initialize_inputs(self):
             # --- is the same on all processors.
             fields_to_plot = list(fields_to_plot)
             fields_to_plot.sort()
-            self.diagnostic.fields_to_plot = fields_to_plot
-
+            self.diagnostic.set_or_replace_attr('fields_to_plot', fields_to_plot)
```
dpgrote, line 2330: I don't think that this is needed here now since there is only one type of lab frame diagnostic. Though, perhaps the `LabFrameParticleDiagnostic` should be implemented (in a separate PR). Does the BTD in WarpX allow turning on and off the particles and fields?

Python/pywarpx/picmi.py
```diff
@@ -1809,14 +1805,13 @@ def initialize_inputs(self):
...
         self.diagnostic.plot_raw_fields = self.plot_raw_fields
         self.diagnostic.plot_raw_fields_guards = self.plot_raw_fields_guards
         self.diagnostic.plot_finepatch = self.plot_finepatch
         self.diagnostic.plot_crsepatch = self.plot_crsepatch
-        self.diagnostic.write_species = self.write_species
-
+        if 'write_species' not in self.diagnostic.argvattrs:
+            self.diagnostic.write_species = False
```
dpgrote, line 2063: This looks good. Perhaps add a comment noting that the check is needed since ParticleDiagnostic will set write_species.

Python/pywarpx/picmi.py
```diff
@@ -1921,7 +1916,9 @@ def initialize_inputs(self):
         self.diagnostic.openpmd_backend = self.openpmd_backend
         self.diagnostic.file_min_digits = self.file_min_digits
         self.diagnostic.intervals = self.period
-
+        self.diagnostic.set_or_replace_attr('write_species', True)
+        if 'fields_to_plot' not in self.diagnostic.argvattrs:
+            self.diagnostic.fields_to_plot = 'none'
```
dpgrote, line 2171: Same here. Perhaps add a comment that the check is needed since FieldDiagnostic sets fields_to_plot.

PR #3746, opened 2023-03-09
Sort `FieldProbe` output using probe particle ids

roelof-groenewald:


Source/Diagnostics/ReducedDiags/FieldProbe.cpp
```diff
@@ -427,13 +427,17 @@ void FieldProbe::ComputeDiags (int step)
             // reset m_data vector to clear pushed values. Reserves data
             m_data.clear();
             m_data.shrink_to_fit();
-            m_data.reserve(numparticles * noutputs);
+            // the +1 below is to temporarily store the particle ID in order
+            // to sort the output
+            m_data.reserve(numparticles * (noutputs + 1));
```
ax3l: why all the +1 instead of adding it in `noutputs` and removing the id in the final print to file? :)

roelof-groenewald: No strong reason. I just kept `noutputs` as its original meaning of how many quantities of interest are saved. I'll change it to have the +1 included in `noutputs`.

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
```diff
@@ -573,13 +575,15 @@ void FieldProbe::ComputeDiags (int step)
                 if (m_intervals.contains(step+1) && np > 0)
                 {
                     // This could be optimized by using shared memory.
+                    // +1 is for storing the particle ID for sorting
```
ax3l:
```suggestion
```

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
```diff
@@ -573,13 +575,15 @@ void FieldProbe::ComputeDiags (int step)
...
                     amrex::Gpu::DeviceVector<amrex::Real> dv(np*noutputs);
                     amrex::Real* dvp = dv.data();
                     amrex::ParallelFor(np, [=] AMREX_GPU_DEVICE (long ip)
                     {
                         amrex::ParticleReal xp, yp, zp;
                         getPosition(ip, xp, yp, zp);
                         long idx = ip*noutputs;
+                        dvp[idx++] = m_structs[ip].id();
```
ax3l, line 585: Let's potentially add an explicit cast here?

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
```diff
@@ -652,30 +656,52 @@ void FieldProbe::ComputeDiags (int step)
 
 void FieldProbe::WriteToFile (int step) const
 {
-    if (ProbeInDomain() && amrex::ParallelDescriptor::IOProcessor())
+    if (!(ProbeInDomain() && amrex::ParallelDescriptor::IOProcessor())) return;
+
+    // loop over num valid particles to find the proper starting ID
```
ax3l: You mean the "lowest" id of a probe particle?

roelof-groenewald: That is correct yes.

ax3l: I would make the comment more explicit then. How about:
```suggestion
    // loop over num valid particles to find the lowest particle ID for later sorting
```

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
```diff
@@ -652,30 +655,53 @@ void FieldProbe::ComputeDiags (int step)
...
+    for (int i = 0; i < m_valid_particles; i++)
     {
-        // open file
-        std::ofstream ofs{m_path + m_rd_name + "." + m_extension,
-                          std::ofstream::out | std::ofstream::app};
+        if (m_data_out[i*noutputs] < first_id)
+            first_id = m_data_out[i*noutputs];
+    }
+
+    // Create a new array to store ordered data that will be printed to file.
```
ax3l:
```suggestion
    // Create a new array to store probe data ordered by id, which will be printed to file.
```

ax3l (approved):
Thanks a lot! :)

PR #3747, opened 2023-03-09
Fix Bucket __getattr__

dpgrote:
This is a small fix in the Python interface code that had incorrect Python code. Note that this bit of code has never been used since this error had not been noticed.

The `object` class has no attribute `__getattr__`. The appropriate attribute to use is `__getattribute__`.


RemiLehe:
Thanks. Could you explain what was wrong with calling `__getattr__` in this case?

RemiLehe (approved):

PR #3751, opened 2023-03-14
remove spaces from headers in `FieldMaximum`, `RhoMaximum` and `ParticleNumber` red diags

aeriforme:
This PR removes the white spaces in the headers of the `FieldMaximum`, `RhoMaximum` and `ParticleNumber` reduced diagnostics. 
The white spaces separated the name of the quantity saved in that column and its units.  
The new behavior is consistent with the other reduced diagnostics. 
Example: from `[2]max_Ex_lev0 (V/m)` to `[2]max_Ex_lev0(V/m)`.


ax3l (approved):
Thanks! That should be easier to parse in pandas et al.

PR #3752, opened 2023-03-14
adds tools for the analysis of reduced diagnostics data 

aeriforme:
This PR adds some python modules to simplify the reading and plotting of reduced diagnostics data. 

For now it works 
* in 3D Cartesian geometry upon merging of PR #3751 
* for `ParticleEnergy`, `ParticleMomentum`, `ParticleNumber`, `FieldEnergy`, `FieldMaximum` reduced diags. 

The directory `WarpX/Examples/Tests/reduced_diags_tools` contains examples of usage of these tools.

A list of open issues to address follows (and more to be found):
- [ ] geometry:
  - [x] 1D Cartesian 
  - [x] 2D Cartesian
  - [ ] RZ
- [ ] `RhoMaximum` diag does not include photon species and one cannot deduce if a species (given its name) is a photon species only from reduced diagnostics data
- [ ] `ParticleExtrema` diag does not include information on the species: maybe standardize name of diagnostic as `ParticleExtrema_<species_names>`
- [ ] integrate the tools in the python workflow 
- [ ] documentation (separate page in postprocessing section)


Python/pywarpx/diagnostics/data/BaseReader.py
```diff
@@ -0,0 +1,173 @@
...
+            raise IOError('The file {} does not exist.\n'
+                          'Did the simulation already run?'
+                          .format(data_file_path))
+
+        return data_file_path
+
+    def get_ncols(self):
+        ''' Returns the number of columns of the reduced diagnostic.'''
+        data_file_path = self.get_data_path()
+        data = np.loadtxt(data_file_path)
```
dpgrote: This may not matter much in general, but if the data files become large, it will be more efficient to read in the data once rather than the way here where the data is read each time a query is made.
Also, are there any cases where ndim == 1, since the files always include the step and time?

Python/pywarpx/diagnostics/data/BaseReader.py
```diff
@@ -0,0 +1,173 @@
...
+                        for i in range(nspecies):
+                            result=re.search('](.+?)'+string,data[start+i*step])
+                            species_names.append(result.group(1))
+            return species_names
+        else:
+            raise ValueError('The current diagnostics does not involve any species!')
+
+    def get_steps(self):
+        data_file_path = self.get_data_path()
+        df = pd.read_csv(data_file_path, sep=" ", header=0)
```
dpgrote: Same here, it will be more efficient to read the data in once and save it as an attribute.

Python/pywarpx/diagnostics/data/BaseReader.py
```diff
@@ -0,0 +1,173 @@
...
+            return nspecies
+        else:
+            return 0
+
+    def get_species_names(self, string=None, start=None, step=None):
+        '''Returns the names of the species involved in the diagnostic.'''
+        # deduces names of species from the header by considering only the relevant entries
+        # finds the string by locating the units
+        if self.has_species == True:
+            species_names = []
```
dpgrote, line 82: Doesn't pandas read in the header? Can you get the species names from there?

Python/pywarpx/diagnostics/data/BaseReader.py
```diff
@@ -0,0 +1,173 @@
...
+        return steps
+
+    def get_times(self):
+        data_file_path = self.get_data_path()
+        df = pd.read_csv(data_file_path, sep=" ", header=0)
+        times = np.unique(df['[1]time(s)'].values)
+        return times
+
+
+    def restrict_data(self, data, **kwargs): # check if works for all diags (e.g. probes, field reduction)
```
dpgrote: This would be cleaner python if the keyword arguments were specified
`def restrict_data(self, data, steps=None, times=None):`
This would make the logic simpler, something like below:
```
if steps is not None and times is not None:
    raise Exception()
elif steps is not None:
    if not isinstance(steps, collections.abc.Iterable):
         steps = np.array([steps])
elif times is not None:
    # set steps from times
if steps is None:
    steps = self.get_steps()
```

Python/pywarpx/diagnostics/data/FieldEnergy.py
```diff
@@ -0,0 +1,91 @@
...
+        """Returns the number of mesh refinement levels in the simulation."""
+
+        return int((self.get_ncols() - 2.)/3.)
+
+    def get_valid_args(self):
+        nlev = self.get_nlev()
+        valid_args = [s for n in range(nlev) for s in ['total_lev'+str(n), 'E_lev'+str(n), 'B_lev'+str(n)]]
+        return valid_args
+
+    def get_data(self, *args, **kwargs):
```
dpgrote: Same here for the keyword args.

Since you have the list of valid_args, it looks like this routine could be generalized for all of the cases and put in the parent class.

dpgrote:
Thanks for this PR! This will be very useful to have. I have some comments aimed toward improving the code.

Python/pywarpx/diagnostics/data/BaseReader.py
aeriforme: Makes sense. What about if it only reads the first column? 
`ndim = 1` was meant for the case where only 1 timestep is dumped. 
Anyway the following should solve all these issues:
```py
header=pd.read_csv(data_file_path, header=0, sep=" ", nrows=0)
ncols = len(header.columns)
```
Do you agree?

Python/pywarpx/diagnostics/data/BaseReader.py
aeriforme: Yes, I started using pandas after doing this. I will fix it.

Python/pywarpx/diagnostics/data/BaseReader.py
aeriforme: Agreed.

Python/pywarpx/diagnostics/data/BaseReader.py
aeriforme: Here it would be nice to have something like `wc -l <filename>` in python.

aeriforme: What do you think about this:
`num_lines = sum(1 for line in open('myfile.txt'))`
?

aeriforme: Ok, no, it does not work.
I forgot about those diagnostics that might have multiple lines for a given timestep.

Python/pywarpx/diagnostics/data/BaseReader.py
dpgrote: Yes, that would be good, using pandas here.
Note that another advantage of reading in the data once is that the loadtxt or read_csv command appear in only one place, reducing code duplication.

Python/pywarpx/diagnostics/data/BaseReader.py
aeriforme: What if we use again pandas capabilities and load only one column, as in:
```py
data_file_path = self.get_data_path()
df = pd.read_csv(data_file_path, sep=" ", usecols=['#[0]step()'])
steps = np.unique(df.values)
```
?

Python/pywarpx/diagnostics/data/BaseReader.py
aeriforme: I understand your point about loading the data just once, but I'm not sure I understand how to implement it given the current design.  

Right now the idea is that the user can directly extract the desired data without having to load the whole file.
Anytime the user calls a `get_data` kind of function, the whole file will be loaded and the desired data will be selected. 
The user can also ask for the available steps and times. 
If we keep this logic, I'm not sure how to have just one occurrence of loadtxt or read_csv command. 

Another approach would be to make the user load the whole file on their own at first, and after that use these tools to extract the desired data. 
To give an idea, something like this:
```
all_data = diag.load(filename) # here the whole file is loaded
steps  = all_data.steps()
desired_data = all_data.get_data()
```
Do you think the latter is a better rationale?

dpgrote: I think it would be fine to have methods that query the file without reading in all of the data, for example reading only the headers or reading in one line of data to get the number of columns. For anything else though that would process the whole file, I think it would be simpler and faster to just read in the data once and query the result as needed. However, it is unlikely that the files will be truly large so just reading in the data right away would likely not be an issue. (My timings show that loadtxt and pandas can read in 1,000,000 lines of data in under a second.)

Python/pywarpx/diagnostics/data/BaseReader.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def __init__(self, run_directory, file_prefix=None):
+        """
+        Parameters
+        ----------
+        run_directory: string
+            Path to the run directory of WarpX.
+        """
+        if run_directory is None:
+            raise ValueError('The run_directory parameter can not be None!')
```
roelof-groenewald: Does it make sense to make the current directory (`"."`) the default? I believe that would work to use the directory from which python is executed, which would allow a workflow where the simulation input file and analysis file could be in the same directory, and the user wouldn't need to specify the run directory.

Python/pywarpx/diagnostics/data/BaseReader.py
```diff
@@ -0,0 +1,189 @@
...
+        Parameters
+        ----------
+        subdir: string
+            Name of reducedfiles directory, default is 'reducedfiles'.
+        Returns
+        -------
+        A string with a path to the current diagnostic (assumed unique).
+        """
+
+        sim_output_dir = os.path.join(self.run_directory, 'diags/'+subdir+'/')
```
dpgrote: The directory will not always be 'diags/' - this should be an input argument that can default to 'diags/'.

Python/pywarpx/diagnostics/data/BaseReader.py
```diff
@@ -0,0 +1,189 @@
...
+
+        data_file_path = os.path.join(
+            sim_output_dir,
+            self.data_file_prefix +
+            self.data_file_suffix
+        )
+        if not os.path.isfile(data_file_path):
+            raise IOError('The file {} does not exist.\n'
+                          'Did the simulation already run?'
+                          .format(data_file_path))
```
roelof-groenewald: This is purely a preference thing, but I like to use f-strings nowadays, since they are more readable in my opinion:
```suggestion
            raise IOError(f'The file {data_file_path} does not exist.\n'
                          'Did the simulation already run?'
                          )
```

Python/pywarpx/diagnostics/data/BaseReader.py
```diff
@@ -0,0 +1,189 @@
...
+        return data_file_path
+
+    def get_ncols(self):
+        ''' Returns the number of columns of the reduced diagnostic.'''
+        data_file_path = self.get_data_path()
+        header=pd.read_csv(data_file_path, header=0, sep=" ", nrows=0)
+        ncols = len(header.columns)
+        return ncols
+
+    def get_nspecies(self, subtract=None, divide=None):
```
dpgrote: I might have used less abstract names. Perhaps 'initial_columns' and 'columns_per_species' as examples.

Python/pywarpx/diagnostics/data/BaseReader.py
```diff
@@ -0,0 +1,189 @@
...
+        # deduces number of species from number of columns
+        if self.has_species == True:
+            ncols = self.get_ncols()
+            # remove <substract> columns and divide to avoid counting multiple entries
+            nspecies = int((ncols - subtract)/divide)
+            return nspecies
+        else:
+            return 0
+
+    def get_species_names(self, string=None, start=None, step=None):
```
dpgrote: The example names above would also work here.

Examples/Tests/reduced_diags_tools/analysis.py
```diff
@@ -0,0 +1,170 @@
...
+
+import ParticleEnergy as PE
+
+diag = PE.ParticleEnergyData(fname='PartEne.txt')
+
+# extract some general info
+times = diag.get_times()
+print('available times\n {}'.format(times))
+
+steps = diag.get_steps()
```
dpgrote, line 24: Could some `asserts` be added to check that the values obtained from the data files are correct?

Examples/Tests/reduced_diags_tools/inputs_3d
```diff
@@ -0,0 +1,100 @@
+# Maximum number of time steps
```
dpgrote, line 1: Rather than having separate input file, could these tests be moved into the `Examples/Tests/reduced_diags` directory and use the existing input file there?

Looking at this a bit more, perhaps the file `Examples/Tests/reduced_diags/analysis_reduced_diags_impl.py` could be modified to use the new tools rather than `np.genfromtxt`?

Python/pywarpx/diagnostics/data/BaseReader.py
```diff
@@ -0,0 +1,185 @@
+import collections
```
dpgrote, line 1: As WarpX is setup now, perhaps a better place for these Python scripts would be in `Tools/PostProcessing`? Also, the more descriptive `reduced_data` would be a better name for the subdirectory instead of `data`.

This would be a separate PR, but I think it would be good if all of the post-processing scripts, including these, would be installed as part of the Python interface.

Python/pywarpx/diagnostics/data/ParticleExtrema.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def get_valid_args(self):
+        """
+        Returns the valid strings to extract the data.
+        """
+
+        if self.get_ncols() == 18:
+            valid_args = ('xmin', 'xmax', 'ymin', 'ymax', 'zmin', 'zmax', 'pxmin', 'pxmax', 'pymin', 'pymax', 'pzmin', 'pzmax', 'gmin', 'gmax', 'wmin', 'wmax')
+        elif self.get_ncols() == 20:
+            valid_args = ('xmin', 'xmax', 'ymin', 'ymax', 'zmin', 'zmax', 'pxmin', 'pxmax', 'pymin', 'pymax', 'pzmin', 'pzmax', 'gmin', 'gmax', 'wmin', 'wmax' 'chimin', 'chimax')
```
dpgrote, line 38: You can add an `else` block that raises an error if the number of columns isn't what was expected.

Python/pywarpx/diagnostics/data/BaseReader.py
ax3l: Yes, the strategy is as follows:
These tools are part of the default Python install, and thus well placed here. But we will also allow to install them separately, for the app purists. I'll add some install logic for this.

Python/pywarpx/diagnostics/data/BaseReader.py
```diff
@@ -0,0 +1,185 @@
...
+import os
+import re
+
+import numpy as np
+import pandas as pd
+
+
+class DataReader():
```
ax3l, line 9: Maybe more descriptive:
```suggestion
class ReducedDiagnosticsData():
```

Python/pywarpx/diagnostics/data/BaseReader.py
```diff
@@ -0,0 +1,185 @@
...
+            raise ValueError('The current diagnostics does not involve any species!')
+
+    def load_data(self):
+        data_file_path = self.get_data_path()
+        df = pd.read_csv(data_file_path, sep=self.separator, header=0)
+        return df
+
+    def get_steps(self):
+        data_file_path = self.get_data_path()
+        df = pd.read_csv(data_file_path, sep=self.separator, usecols=['#[0]step()'])
```
ax3l, line 102: We don't need to do this too early, but technically we can and should cache these reads on first access to any of these methods to avoid repeated filesystem access.

Examples/Tests/reduced_diags_tools/analysis.py
```diff
@@ -0,0 +1,170 @@
+#!/usr/bin/env python3
+
+#####################################################
+# examples of usage of the reduced diagnostic tools #
+#####################################################
+
+import sys
+
+sys.path.append('../../../Python/pywarpx/diagnostics/data/')
+mydir = '.'
```
ax3l, line 10: TODO, will take care of that.

PR #3756, opened 2023-03-15
Allow using different types for coordinates and values in {bi,tri}-linear interpolation functions

lucafedeli88:
This PR changes the {bi,tri}-linear inteporlation functions adding the possibility to use arguments of different types for coordinates and values.


PhilMiller (approved):
Looks like what I expected

RemiLehe (approved):
Thanks! This looks good!

PR #3762, opened 2023-03-16
Enable calling SortParticlesForDeposition

atmyers:
This enables calling the new `SortParticlesForDeposition` that @AlexanderSinn developed and added to amrex. 

Some timing data I collected on Perlmutter (using the "cell" idx_type for the sort):

> 
> Two-node KPP test on Perlmutter (sorting both every 4 steps): 
> 
> 	New:
> 	
> 		TinyProfiler total time across processes [min...avg...max]: 688.3 ... 688.3 ... 688.3
> 	
> 	Old:
> 	
> 		TinyProfiler total time across processes [min...avg...max]: 685.6 ... 685.6 ... 685.6
> 	
> 4 ppc per species uniform plasma (sorting both every 4 steps):
> 
> 	New:
> 	
> 		TinyProfiler total time across processes [min...avg...max]: 14.88 ... 14.88 ... 14.88
> 		
> 	Old:
> 	
> 		TinyProfiler total time across processes [min...avg...max]: 14.35 ... 14.35 ... 14.35
> 	
> 32 ppc per specific uniform plasma (sorting both runs every step):
> 
> 	New:
> 	
> 		TinyProfiler total time across processes [min...avg...max]: 97.25 ... 97.25 ... 97.25
> 		
> 	Old:
> 	
> 		TinyProfiler total time across processes [min...avg...max]: 172 ... 172 ... 172

So this is a massive improvement at high ppc and essentially no difference at low ppc. Thus, there's no reason not to make it the new default.


.github/workflows/cuda.yml
```diff
@@ -110,7 +110,7 @@ jobs:
         which nvcc || echo "nvcc not in PATH!"
 
         git clone https://github.com/AMReX-Codes/amrex.git ../amrex
-        cd ../amrex && git checkout --detach 6d30e83c944e4e6167178f2d145df2a3e67d2b24 && cd -
+        cd amrex && git checkout --detach 254d3faba017c9b13800d68499abcfe7d0405128 && cd -
```
ax3l: Depends on #3767

ax3l: merged - please undo change here :) @atmyers

Source/WarpX.H
```diff
@@ -307,6 +307,9 @@ public:
     static utils::parser::IntervalsParser sort_intervals;
     static amrex::IntVect sort_bin_size;
 
+    static bool sort_particles_for_deposition;
+    static amrex::IntVect sort_idx_type;
```
ax3l: Please add Doxygen strings via `//! ...`

AlexanderSinn (commented):
Looks Good! Maybe after lots of benchmarking, an optimal default value for `sort_idx_type` can be set depending on particle shape factor, deposition type and grid staggering. But for now `0 0 0` is fine. I am not sure why the CI tests fail.

.github/workflows/cuda.yml
ax3l: @atmyers I think this line is still changed

Docs/source/usage/parameters.rst
```diff
@@ -2045,12 +2045,25 @@ Additional parameters
...
+     If ``true``, particles will be sorted by cell to optimize deposition with many particles per cell, in the order x -> y -> z -> ppc.
+     If ``false``, particles will be sorted by bin, using the ``sort_bin_size`` parameter below, in the order ppc -> x -> y -> z.
+     ``true`` is recommend for best performance on NVIDIA GPUs, especially if there are many particles per cell.
+
+* ``warpx.sort_idx_type`` (list of `int`) optional (default: ``0 0 0``)
+    This controls the type of grid used to sort the particles when ``sort_particles_for_deposition`` is ``true``. Possible values are:
+    ``idx_type = {0, 0, 0}``: Sort particles to a cell centered grid
+    ``idx_type = {1, 1, 1}``: Sort particles to a node centered grid
+    ``idx_type = {2, 2, 2}``: Compromise between a cell and node centered grid.
+     In 2D, only the first two elements are read.
```
ax3l: Just for completion:

2D (and RZ)?

In 1D, only the first element is read?

ax3l:
Notes:
- we see no improvement (but even a perf regression) on ROCm/HIP 5.2 yet
- we have not tested SYCL yet

Thus, this default is changed only for CUDA.

.github/workflows/cuda.yml
```diff
@@ -110,7 +110,7 @@ jobs:
         which nvcc || echo "nvcc not in PATH!"
 
         git clone https://github.com/AMReX-Codes/amrex.git ../amrex
-        cd ../amrex && git checkout --detach 6d30e83c944e4e6167178f2d145df2a3e67d2b24 && cd -
+        cd ../amrex && git checkout --detach 5f8030af0ddfe99139968bcf896b5662cccf94a4 && cd -
```
ax3l:
```suggestion
        cd ../amrex && git checkout --detach 6d30e83c944e4e6167178f2d145df2a3e67d2b24 && cd -
```

ax3l (approved):
Thanks a lot!

.github/workflows/cuda.yml
atmyers: This need to be here - this PR uses functions from AMReX that don't exist in version `6d30e83c944e4e6167178f2d145df2a3e67d2b24`

ax3l: Oh, yes - sorry for this. I forgot that #3772 was not yet in.

 #3788 will fix this in the next weekly update.

PR #3764, opened 2023-03-16
Fix bugs related to azimuthal modes for multi-J algo in RZ geometry

oshapoval:
For reference, this is how all fields, with modes, are stored in spectral space:
https://github.com/ECP-WarpX/WarpX/blob/f85998603fecd102e6f5ed06ecc08cf85b9a8bfb/Source/FieldSolver/SpectralSolver/SpectralFieldDataRZ.cpp#L37-L43

Added capability to loop over all azimuthal modes in `CopySpectralDataComp`, `ZeroOutDataComp` and `ScaleDataComp` by taking into account the shift `mode*n_fields`.

This PR fixes a few bugs in RZ with:
- multi-J PSATD solver w/o averaging;
- multi-J PSATD solver w/ averaging.

### DIFF
--- a/Source/FieldSolver/SpectralSolver/SpectralAlgorithms/PsatdAlgorithmRZ.cpp
+++ b/Source/FieldSolver/SpectralSolver/SpectralAlgorithms/PsatdAlgorithmRZ.cpp
@@ -408,14 +408,14 @@ void PsatdAlgorithmRZ::InitializeSpectralCoefficients (SpectralFieldDataRZ const
 
                 if (om != 0.0_rt)
                 {
-                    X5(i,j,k) = c2 / ep0 * (S_ck(i,j,k) / om2 - (1._rt - C(i,j,k)) / (om4 * dt)
+                    X5(i,j,k,mode) = c2 / ep0 * (S_ck(i,j,k,mode) / om2 - (1._rt - C(i,j,k,mode)) / (om4 * dt)
                                             - 0.5_rt * dt / om2);
-                    X6(i,j,k) = c2 / ep0 * ((1._rt - C(i,j,k)) / (om4 * dt) - 0.5_rt * dt / om2);
+                    X6(i,j,k,mode) = c2 / ep0 * ((1._rt - C(i,j,k,mode)) / (om4 * dt) - 0.5_rt * dt / om2);
                 }
                 else
                 {
-                    X5(i,j,k) = - c2 * dt3 / (8._rt * ep0);
-                    X6(i,j,k) = - c2 * dt3 / (24._rt * ep0);
+                    X5(i,j,k,mode) = - c2 * dt3 / (8._rt * ep0);
+                    X6(i,j,k,mode) = - c2 * dt3 / (24._rt * ep0);
                 }
             }
         });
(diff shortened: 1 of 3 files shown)


EZoni:
Please rebase on `development` to get the latest CI fix (#3765).

Source/FieldSolver/SpectralSolver/SpectralFieldDataRZ.H
```diff
@@ -73,9 +73,14 @@ class SpectralFieldDataRZ
          */
         void CopySpectralDataComp (const int src_comp, const int dest_comp)
         {
-            // The last two arguments represent the number of components and
-            // the number of ghost cells to perform this operation
-            Copy(fields, fields, src_comp, dest_comp, n_rz_azimuthal_modes, 0);
+            const int nmodes = n_rz_azimuthal_modes;
+            for (int imode=0 ; imode < nmodes ; imode++)
```
EZoni: I would add a comment that explains why the shift is needed, and refers to the allocation of `fields` in the constructor of the class. Also, I think we could use `n_rz_azimuthal_modes` directly in the `for` loop, instead of allocating an extra variable:
```suggestion
            // The fields of each mode are grouped together, so that the index of a 
            // field for a specific mode is given by field_index + mode*n_fields
            // (see allocation of fields in SpectralFieldDataRZ.cpp)
            for (int imode=0; imode < n_rz_azimuthal_modes; imode++)
```

I would do this here as well as in the other two functions below.

EZoni (commented):
Thank you for this bug fix, @oshapoval! I left one suggestion, otherwise it looks good to me.

Source/FieldSolver/SpectralSolver/SpectralFieldDataRZ.H
oshapoval: Thank you for the review, @EZoni. This was fixed in 5d17ad222c3149eb5463fa0570ce46f205259e4f and 81530393add446fddd62909b3daf280bf3ed043c.

Source/FieldSolver/SpectralSolver/SpectralAlgorithms/PsatdAlgorithmRZ.cpp
```diff
@@ -382,7 +382,7 @@ void PsatdAlgorithmRZ::InitializeSpectralCoefficients (SpectralFieldDataRZ const
             amrex::Real const k_norm = std::sqrt(kr*kr + kz*kz);
 
             // Calculate coefficients
-            constexpr amrex::Real c = PhysConst::c;
+            constexpr amrex::Real c = PhysConst::c * PhysConst::c;
```
EZoni: This is called `c`, and not `c2`.
I think we might need to leave this line as it was and fix the $c^2$ factor in line 403, right?

oshapoval:
This is called bug on top of the previous bug 😂
Yes, this fix needs to go to line 403. Thanks for spotting this @ezoni !

EZoni:
@oshapoval Thank you for the latest fixes. Could you reset the benchmark of the test `multi_J_rz_psatd`? They changed because of the bug fix related to the coefficients used with time averaging (which is used in that test). Once that's done, I think the PR is ready to merge.

EZoni (approved):

PR #3765, opened 2023-03-16
CI: fix bug in NVCC GNUMake yml file

EZoni:
We found out that the CI build "NVCC 11.8.0 GNUMake" was not using the version of AMReX that we thought it would use. The `cd amrex` command was wrong, hence the specific AMReX commit hash was not checked out and the latest version of AMReX was used, instead of the desired one.

Thank you to @WeiqunZhang and @atmyers for help with debugging.

### DIFF
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -110,7 +110,7 @@ jobs:
         which nvcc || echo "nvcc not in PATH!"
 
         git clone https://github.com/AMReX-Codes/amrex.git ../amrex
-        cd amrex && git checkout --detach 6d30e83c944e4e6167178f2d145df2a3e67d2b24 && cd -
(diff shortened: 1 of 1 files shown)


WeiqunZhang (approved):

.github/workflows/cuda.yml
```diff
@@ -110,7 +110,7 @@ jobs:
         which nvcc || echo "nvcc not in PATH!"
 
         git clone https://github.com/AMReX-Codes/amrex.git ../amrex
-        cd amrex && git checkout --detach 6d30e83c944e4e6167178f2d145df2a3e67d2b24 && cd -
```
ax3l, line 113: Please dbl check the script in `/home/axel/src/warpx/Tools/Release/updateAMReX.py` updates this location correctly :)

PR #3768, opened 2023-03-20
Allow 8-fold symmetry option

RTSandberg:
This PR extends the beam symmetrization functionality.  It was observed that initial beam emittances in `x` and `y` were different due to statistical noise.  To eliminate the possible effect of the initial emittance value on emittance growth, 8-fold beam symmetry ensures that initial emittances in the `x`- and `y`- directions are the same.

This PR adds an additional optional parameter ``warpx.symmetrization_order``, which can currently be 4 or 8 and defaults to 4.  If ``warpx.do_symmetrize=1``, then the value of ``warpx.symmetrization_order`` determines whether the beam is symmetrized in x and y only (``warpx.symmetrization_order=4``, ``(x,y) (-x,y) (x,-y) (-x,-y)``) or is symmetrized in `x`,`y`, and across diagonals (``warpx.symmetrization_order=8``, also ``(y,x), (-y,x), (y,-x), (-y,-x)``).

With order 4, an example input script
[inputs_rz.txt](https://github.com/ECP-WarpX/WarpX/files/11021726/inputs_rz.txt)
leads to the following initial emittance values:
`emitt_x = 9.46075997990898e-07, emitt_y = 1.1325899636804437e-06`
WIth order 8, the initial emittance values are:
`emitt_x = 1.0348160840903872e-06, emitt_y = 1.0348160840903872e-06`

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -674,7 +674,7 @@ Particle initialization
       ``<species_name>.x/y/z_rms`` (standard deviation in `x/y/z`),
       ``<species_name>.x/y/z_cut`` (optional, particles with ``abs(x-x_m) > x_cut*x_rms`` are not injected, same for y and z. ``<species_name>.q_tot`` is the charge of the un-cut beam, so that cutting the distribution is likely to result in a lower total charge),
       and optional arguments ``<species_name>.do_symmetrize`` (whether to
-      symmetrize the beam) and ``<species_name>.symmetrization_order`` (Extent of symmetrization, default is 4, can be 4 or 8.
+      symmetrize the beam) and ``<species_name>.symmetrization_order`` (order of symmetrization, default is 4, can be 4 or 8).
       If ``<species_name>.do_symmetrize`` is 0, no symmetrization occurs.  If ``<species_name>.do_symmetrize`` is 1,
       then the beam is symmetrized according to the value of ``<species_name>.symmetrization_order``.
       If set to 4, symmetrization is in the x and y direction, (x,y) (-x,y) (x,-y) (-x,-y).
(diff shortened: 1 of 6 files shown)


Docs/source/usage/parameters.rst
```diff
@@ -673,8 +673,12 @@ Particle initialization
       ``<species_name>.x/y/z_m`` (average position in `x/y/z`),
       ``<species_name>.x/y/z_rms`` (standard deviation in `x/y/z`),
       ``<species_name>.x/y/z_cut`` (optional, particles with ``abs(x-x_m) > x_cut*x_rms`` are not injected, same for y and z. ``<species_name>.q_tot`` is the charge of the un-cut beam, so that cutting the distribution is likely to result in a lower total charge),
-      and optional argument ``<species_name>.do_symmetrize`` (whether to
-      symmetrize the beam in the x and y directions).
+      and optional arguments ``<species_name>.do_symmetrize`` (whether to
+      symmetrize the beam) and ``<species_name>.symmetrization_order`` (Extent of symmetrization, default is 4, can be 4 or 8.
```
EZoni: Maybe "order" is clearer than "extent" in this case (see also how you called the parameter itself)?
```suggestion
      symmetrize the beam) and ``<species_name>.symmetrization_order`` (order of symmetrization, default is 4, can be 4 or 8).
```

Examples/Tests/multi_j/inputs_rz
```diff
@@ -83,7 +84,8 @@ driver_back.npart = 1000000
 driver_back.q_tot = 1e-9
 driver_back.momentum_distribution_type = "at_rest"
 driver_back.initialize_self_fields = 0
-driver_back.do_symmetrize = 1
+driver.do_symmetrize = 1
```
EZoni: You noticed this earlier yourself, but the fix still needs to be pushed:
```suggestion
driver_back.do_symmetrize = 1
```

EZoni (commented):
Thank you for this PR, @RTSandberg! It looks good to me, I just left a couple of minor comments. Also, let me know if the CI checksum issue is progressing well or if any help is needed there.

Examples/Tests/multi_j/inputs_rz
RTSandberg: pushed!

EZoni (approved):

PR #3771, opened 2023-03-21
openPMD-api 0.15.1

ax3l:
Update WarpX to use openPMD-api `0.15.1`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -270,7 +270,6 @@ if(WarpX_PSATD)
 endif()
 
 if(WarpX_OPENPMD)
-    target_compile_definitions(ablastr PUBLIC WARPX_USE_OPENPMD)
     target_link_libraries(ablastr PUBLIC openPMD::openPMD)
 endif()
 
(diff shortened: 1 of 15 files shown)


CMakeLists.txt
```diff
@@ -116,7 +116,7 @@ option(WarpX_IPO                                "Compile WarpX with interprocedu
 #   this is temporary and we don't want to set it long term
 #   mainly to work-around a global setting in openPMD-api <= 0.14.5 that changes the
 #   default in superbuilds by accident
-if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
+if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)  # update me
```
ax3l: update me

Tools/machines/summit-olcf/summit_warpx.profile.example
```diff
@@ -28,7 +28,7 @@ module load hdf5/1.10.7
 # optional: for openPMD support (GNUmake only)
 #module load ums
 #module load ums-aph114
-#module load openpmd-api/0.14.2
+#module load openpmd-api/0.15.0
```
ax3l: Deploy me (optional, GNUmake only)

CMakeLists.txt
ax3l: To do: also need to update the default in `openPMD.cmake` from shared to static.

ax3l: done

RemiLehe (approved):
Thanks! ✨

PR #3772, opened 2023-03-21
Update `updateAMReX.py` to match new CUDA yml file

EZoni:
Fix in `updateAMReX.py` needed after #3765, as pointed out by @ax3l in [this comment](https://github.com/ECP-WarpX/WarpX/pull/3765/files#r1143641098).
I tested it locally and it seemed to work.

### DIFF
--- a/Tools/Release/updateAMReX.py
+++ b/Tools/Release/updateAMReX.py
@@ -123,9 +123,9 @@ ci_gnumake_path = str(REPO_DIR.joinpath(".github/workflows/cuda.yml"))
 with open(ci_gnumake_path, encoding='utf-8') as f:
     ci_gnumake_content = f.read()
     #   branch/commit/tag (git fetcher) version
-    #     cd amrex && git checkout COMMIT_TAG_OR_BRANCH && cd -
+    #     cd ../amrex && git checkout COMMIT_TAG_OR_BRANCH && cd -
     ci_gnumake_content = re.sub(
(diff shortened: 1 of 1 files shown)


Tools/Release/updateAMReX.py
```diff
@@ -123,9 +123,9 @@
 with open(ci_gnumake_path, encoding='utf-8') as f:
     ci_gnumake_content = f.read()
     #   branch/commit/tag (git fetcher) version
-    #     cd amrex && git checkout COMMIT_TAG_OR_BRANCH && cd -
+    #     cd ../amrex && git checkout COMMIT_TAG_OR_BRANCH && cd -
     ci_gnumake_content = re.sub(
-        r'(.*cd\s+amrex.+git checkout\s+--detach\s+)(.+)(\s+&&\s.*)',
+        r'(.*cd\s+../amrex.+git checkout\s+--detach\s+)(.+)(\s+&&\s.*)',
```
ax3l: Thank you very much for this! :pray: 

Technically, the `.` will match any character since this is a regex.
Explicit would be
```suggestion
        r'(.*cd\s+\.\./amrex.+git checkout\s+--detach\s+)(.+)(\s+&&\s.*)',
```
to escape a literal `.`

Python regex special characters are: `.`, `+`, `*`, `?`, `^`, `$`, `(`, `)`, `[`, `]`, `{`, `}`, `|`, `\`

ax3l (approved):
Thank you very much :)

PR #3773, opened 2023-03-21
Add more PICMI options

RemiLehe:
This exposes two new parameters via the PICMI interface:
- `flux_tmin` and `flux_tmax`, which allows to turn the particle flux injection on and off.
- `cover_multiple_cuts`, which allows to run cases where the EB cuts a single cell in multiple places. (When this option is turned on, `AMReX` considers that the cell is entirely covered by the EB, instead of returning the usual error.)


Python/pywarpx/picmi.py
```diff
@@ -416,6 +416,10 @@ def initialize_inputs(self, species_number, layout, species, density_scale):
         species.flux_normal_axis = self.flux_normal_axis
         species.surface_flux_pos = self.surface_flux_position
         species.flux_direction = self.flux_direction
+        if self.flux_tmin is not None:
```
dpgrote: The check `is not None` is not needed. If parameters are set to None, they are ignored.

Python/pywarpx/picmi.py
```diff
@@ -1341,12 +1345,15 @@ class EmbeddedBoundary(picmistandard.base._ClassWithInit):
...
+    cover_multiple_cuts: bool, default=False
+        Whether to cover cells with multiple cuts.
+        (If False, this will raise an error if some cells have multiple cuts)
 
     Parameters used in the analytic expressions should be given as additional keyword arguments.
 
     """
     def __init__(self, implicit_function=None, stl_file=None, stl_scale=None, stl_center=None, stl_reverse_normal=False,
-                 potential=None, **kw):
+                 potential=None, cover_multiple_cuts=False, **kw):
```
dpgrote: The default value of `cover_multiple_cuts` should be None. This avoids having the default value set in two different places, here and in the C++.

Python/pywarpx/picmi.py
```diff
@@ -1393,6 +1402,9 @@ def initialize_inputs(self, solver):
             pywarpx.eb2.stl_center = self.stl_center
             pywarpx.eb2.stl_reverse_normal = self.stl_reverse_normal
 
+        if self.cover_multiple_cuts:
```
dpgrote: This can be a direct assignment
```
pywarpx.eb2.cover_multiple_cuts = self.cover_multiple_cuts
```
The interface already handles the setting of booleans.

Python/pywarpx/picmi.py
```diff
@@ -416,6 +416,8 @@ def initialize_inputs(self, species_number, layout, species, density_scale):
         species.flux_normal_axis = self.flux_normal_axis
         species.surface_flux_pos = self.surface_flux_position
         species.flux_direction = self.flux_direction
+        species.flux_tmin = self.flux_tmin
```
dpgrote, line 419: This is relying on `flux_tmin` and `max` being added as attributes to `picmistandard.PICMI_UniformFluxDistribution`. Is there going to be a PR in picmi to add these?

RemiLehe: Did you mean this PR?
https://github.com/picmi-standard/picmi/pull/98/files
Or do we need something in addition to this?

dpgrote: Sorry, yes, never mind. My local repo wasn't updating properly and I didn't see the options (now fixed).

dpgrote (approved):
This looks good! Thanks!

PR #3776, opened 2023-03-22
Fix add_particles to account for theta attribute with RZ

dpgrote:
In the Python routine `add_particles`, when handling the real attributes, it was taking into account that the velocities are included in the attributes. However, when using RZ, `theta` is also an attribute and this was not being accounted for. This was leading to a seg fault when using `add_particles` with RZ. This PR fixes the issue.

This PR also adds assertions to avoid seg faults in similar circumstances.


ax3l (approved):
LGTM, thank you!! :)

PR #3777, opened 2023-03-22
Fix SyncCurrent and SyncRho for multi-levels

WeiqunZhang:
Previously these two function did not work when max_level > 1.  For example, when max_level = 2, the coarse patch data from level 2 were coarsened and added to the fine patch on level 1. However, the coarse patch on level 1 did not have any contribution from level 2 as it should have, because the coarsening was performed before the fine patch on level 1 received the level 2 contribution.  The order has been fixed in this commit.


Source/Parallelization/WarpXComm.cpp
```diff
@@ -863,6 +863,8 @@ WarpX::SyncCurrent (
     // If warpx.do_current_centering = 1, center currents from nodal grid to staggered grid
     if (WarpX::do_current_centering)
     {
+        AMREX_ALWAYS_ASSERT_WITH_MESSAGE(finest_level <= 1,
+                                         "warpx.do_current_centering=1 not supported with more than one fine levels");
```
WeiqunZhang, line 867: To be safe, I added an assertion here, because I am not sure if we need to do anything special for do_current_centering.

ax3l:
Looks like only one test failed in checksums
```
Langmuir_multi_2d_MR_psatd FAILED
```

Source/Parallelization/WarpXComm.cpp
EZoni: The assertion is a good idea for now, I think current centering is not fully supported with mesh refinement yet.

Source/Parallelization/WarpXComm.cpp
```diff
@@ -871,33 +873,78 @@ WarpX::SyncCurrent (
         }
     }
 
-    // Restrict fine patch current onto the coarse patch, before
-    // summing the guard cells of the fine patch
-    for (int lev = 1; lev <= finest_level; ++lev)
+    std::unique_ptr<MultiFab> mf_comm; // for communication between levels
```
EZoni, line 942: Do we need to initialize this MultiFab to something (zero)? I'm just wondering, what happens the first time that `fine_lev_cp.ParallelAdd(*mf_comm, ...)` is called, if `mf_comm` is just allocated but not initialized?

Source/Parallelization/WarpXComm.cpp
```diff
@@ -871,33 +873,78 @@ WarpX::SyncCurrent (
...
+        {
+            const int ncomp = J_fp[lev][idim]->nComp();
+            auto const& period = Geom(lev).periodicity();
 
-        const IntVect& refinement_ratio = refRatio(lev-1);
+            if (lev < finest_level)
+            {
+                // On a coarse level, the data in mf_comm comes from the
+                // fine level. They are unfiltered and uncommuncatied. We
+                // need to add it to the current level.
```
EZoni: I wonder if talking about "coarse level" and "fine level" in this comment could be somewhat misleading, in the sense that any given level beyond 0 has both fine patch data (typically denoted with fp) and coarse patch data (typically denoted with cp), right? Is there a way to make this comment clearer with respect to this nuance, possibly clearly stating which type of patch (fine, coarse) of which level (finer, coarser) is supposed to get data from which type of patch (fine, coarse) of which level (finer, coarser)?

Source/Parallelization/WarpXComm.cpp
```diff
@@ -871,33 +873,78 @@ WarpX::SyncCurrent (
...
+                        (*J_cp[lev][idim], amrex::make_alias, 0, ncomp);
+                }
+            }
 
-    // For each level
-    // - apply filter to the coarse patch/buffer of `lev+1` and fine patch of `lev` (same resolution)
-    // - add the coarse patch/buffer of `lev+1` into the fine patch of `lev`
-    // - sum guard cells of the coarse patch of `lev+1` and fine patch of `lev`
-    for (int lev=0; lev <= finest_level; ++lev) {
-        AddCurrentFromFineLevelandSumBoundary(J_fp, J_cp, lev);
```
EZoni, line 900: It seems that the `AddCurrentFromFineLevelandSumBoundary` function call has been removed, and the same for `AddRhoFromFineLevelandSumBoundary` below. Does this mean that the two functions themselves can be removed as well?

EZoni (commented):
Thank you for this fix, @WeiqunZhang. I left a few comments to start with, but I'll need one or two more rounds of review, since it's a relatively complicated part of the code.

Source/Parallelization/WarpXComm.cpp
WeiqunZhang: I think we are safe, because `mf_comm` is an alias MultiFab whose internal pointers pointing to the data in `current_buf` or `J_cp`.

Source/Parallelization/WarpXComm.cpp
WeiqunZhang:
```suggestion
                // On a coarse level, the data in mf_comm comes from the
                // coarse patch of the fine level. They are unfiltered and uncommunicated.
                // We need to add it to the fine patch of the current level.
```
Is this better?

Source/Parallelization/WarpXComm.cpp
WeiqunZhang: They are still being used in WarpXEvolve.cpp.

Source/Parallelization/WarpXComm.cpp
```diff
@@ -909,21 +956,65 @@ WarpX::SyncRho ()
...
-        const IntVect& refinement_ratio = refRatio(lev-1);
-        ablastr::coarsen::average::Coarsen(*rho_cp[lev], *rho_fp[lev], refinement_ratio );
-    }
+        if (lev < finest_level)
+        {
+            auto const& period = Geom(lev).periodicity();
+
+            // On a coarse level, the data in mf_comm comes from the
+            // fine level. They are unfiltered and uncommuncatied. We
+            // need to add it to the current level.
```
EZoni: I think we should improve this inline comment too, as the one above:
```suggestion
            // On a coarse level, the data in mf_comm comes from the
            // coarse patch of the fine level. They are unfiltered and uncommunicated.
            // We need to add it to the fine patch of the current level.
```

Source/Parallelization/WarpXComm.cpp
EZoni: Got it, because the loop starts from the finest level and goes back down to the coarsest. Thanks for clarifying.

EZoni:
One more question:
Do you have a guess about why only one CI test with mesh refinement (`Langmuir_multi_2d_MR_psatd`, as Axel pointed out above - and I just confirmed locally) needs a benchmark update? I would have expected these code changes to modify the checksums of most CI tests with mesh refinement, to an extent higher than the default relative tolerance that we use.

WeiqunZhang:
I don't know.  Maybe that test is more sensitive to roundoff errors.

ax3l:
@EZoni @WeiqunZhang let's add this to the release today. Can you fix the conflict and merge?

WeiqunZhang:
> @EZoni @WeiqunZhang let's add this to the release today. Can you fix the conflict and merge?

Will do it this morning.

ax3l:
Thank you!

Discussed offline: @RemiLehe  like's to take a look at inline details and we need a bit more to make it user-facing, thus will not include it in the release just yet to not rush it.

Source/Parallelization/WarpXComm.cpp
```diff
@@ -871,33 +873,78 @@ WarpX::SyncCurrent (
...
+                // On a coarse level, the data in mf_comm comes from the
+                // coarse patch of the fine level. They are unfiltered and uncommunicated.
+                // We need to add it to the fine patch of the current level.
+                MultiFab fine_lev_cp(J_fp[lev][idim]->boxArray(),
+                                     J_fp[lev][idim]->DistributionMap(),
+                                     ncomp, 0);
+                fine_lev_cp.setVal(0.0);
+                fine_lev_cp.ParallelAdd(*mf_comm, 0, 0, ncomp, mf_comm->nGrowVect(),
+                                        IntVect(0), period);
+                // We now need to create a mask to fix the double counting.
```
ax3l, line 961: Can you elaborate in this inline comment what `the double counting` risk that is avoided is exactly? Which parts/points would be double-countered without a mask?

Source/Parallelization/WarpXComm.cpp
```diff
@@ -871,33 +873,78 @@ WarpX::SyncCurrent (
...
+                auto const& sma = fine_lev_cp.const_arrays();
+                auto const& dma = J_fp[lev][idim]->arrays();
+                amrex::ParallelFor(fine_lev_cp, IntVect(0), ncomp,
+                [=] AMREX_GPU_DEVICE (int bno, int i, int j, int k, int n)
+                {
+                    if (mma[bno](i,j,k) && sma[bno](i,j,k,n) != 0.0_rt) {
+                        dma[bno](i,j,k,n) += sma[bno](i,j,k,n);
+                    }
+                });
+                // Now it's safe to apply filter and sumboundary on J_cp
```
ax3l, line 973: Can you elaborate in the inline comment on what is the safe aspect?

Is it this?
```suggestion
                // With the fine patch J_fp on the current level updated,
                // we can now apply the filter and SumBoundary to the
                // coarse patch J_cp on the higher level.
```

Source/Parallelization/WarpXComm.cpp
```diff
@@ -871,33 +873,78 @@ WarpX::SyncCurrent (
...
-        ablastr::coarsen::average::Coarsen(*crse[1], *fine[1], refinement_ratio );
-        ablastr::coarsen::average::Coarsen(*crse[2], *fine[2], refinement_ratio );
-    }
+            if (lev > 0)
+            {
+                // On a fine level, we need to coarsen the current onto the
+                // coarse level. This needs to be done before filtering because
+                // filtering depends on the level. This is also done before any
+                // same-level communication because it's easier this way to
+                // avoid double counting.
```
ax3l, line 987: As above: double counting of which points is referred to? Between levels or on the same level between neighboring boxes that share ownership of the same index?
```suggestion
                // avoid double counting of ...
```

Source/Parallelization/WarpXComm.cpp
```diff
@@ -871,33 +873,78 @@ WarpX::SyncCurrent (
...
+                J_cp[lev][idim]->setVal(0.0);
+                ablastr::coarsen::average::Coarsen(*J_cp[lev][idim],
+                                                   *J_fp[lev][idim],
+                                                   refRatio(lev-1));
+                if (current_buf[lev][idim])
```
ax3l, line 992: Can you add a comment why we can take an alias or need to copy based on which condition?

Source/Parallelization/WarpXComm.cpp
```diff
@@ -916,21 +963,65 @@ WarpX::SyncRho (
...
+            // On a coarse level, the data in mf_comm comes from the
+            // coarse patch of the fine level. They are unfiltered and uncommunicated.
+            // We need to add it to the fine patch of the current level.
+            MultiFab fine_lev_cp(charge_fp[lev]->boxArray(),
+                                 charge_fp[lev]->DistributionMap(),
+                                 ncomp, 0);
+            fine_lev_cp.setVal(0.0);
+            fine_lev_cp.ParallelAdd(*mf_comm, 0, 0, ncomp, mf_comm->nGrowVect(),
+                                    IntVect(0), period);
+            // We now need to create a mask to fix the double counting.
```
ax3l, line 1050: As above :)

Source/Parallelization/WarpXComm.cpp
```diff
@@ -916,21 +963,65 @@ WarpX::SyncRho (
...
+            auto const& sma = fine_lev_cp.const_arrays();
+            auto const& dma = charge_fp[lev]->arrays();
+            amrex::ParallelFor(fine_lev_cp, IntVect(0), ncomp,
+            [=] AMREX_GPU_DEVICE (int bno, int i, int j, int k, int n)
+            {
+                if (mma[bno](i,j,k) && sma[bno](i,j,k,n) != 0.0_rt) {
+                    dma[bno](i,j,k,n) += sma[bno](i,j,k,n);
+                }
+            });
+            // Now it's safe to apply filter and sumboundary on charge_cp
```
ax3l, line 1062: As above :)

Source/Parallelization/WarpXComm.cpp
```diff
@@ -916,21 +963,65 @@ WarpX::SyncRho (
...
+            ApplyFilterandSumBoundaryRho(lev+1, lev, *charge_cp[lev+1], 0, ncomp);
+        }
+
+        if (lev > 0)
+        {
+            // On a fine level, we need to coarsen the data onto the coarse
+            // level. This needs to be done before filtering because
+            // filtering depends on the level. This is also done before any
+            // same-level communication because it's easier this way to
+            // avoid double counting.
```
ax3l, line 1072: As above :)

Source/Parallelization/WarpXComm.cpp
```diff
@@ -916,21 +963,65 @@ WarpX::SyncRho (
...
+            charge_cp[lev]->setVal(0.0);
+            ablastr::coarsen::average::Coarsen(*charge_cp[lev],
+                                               *charge_fp[lev],
+                                               refRatio(lev-1));
+            if (charge_buf[lev])
```
ax3l, line 1077: As above :)

WeiqunZhang:
More comments added.

Source/Parallelization/WarpXComm.cpp
```diff
@@ -871,33 +873,143 @@ WarpX::SyncCurrent (
...
+    // to coarse levels. In the implementation below, we loop over levels
+    // from the finest to the coarsest. On each level, filtering and
+    // SumBoundary are done as the last two things. So the communication
+    // data on the sender side are always unfiltered and unsummed. The
+    // receivers are responsible for filtering that is dependent on the grid
+    // resolution. On the finest level, we coarsen the unsummed fp data onto
+    // cp grids, which are the coarsened version of the fp grids with the
+    // same DistributionMapping. Then on the level below, We use ParallelAdd
+    // to add the finer level's cp data onto the current level's fp
+    // data. After that, we apply filter and SumBounadry to the finer
```
ax3l:
```suggestion
    // data. After that, we apply filter and SumBoundary to the finer
```

Source/Parallelization/WarpXComm.cpp
```diff
@@ -871,33 +873,143 @@ WarpX::SyncCurrent (
...
+    // there are levels below this, we need to process this level's cp data
+    // just like we have done for the finer level. The iteration continues
+    // until we reache level 0. There are however two additional
+    // complications.
+    //
+    // The first complication is that simply calling ParallelAdd to add the
+    // finer level's cp data to the current level's fp data does not
+    // work. Suppose there are multiple boxes on the current level (or just
+    // a single box with periodic bounaries). A given (i,j,k) can be present
+    // in more than one box. At the time of calling ParallelAdd, the current
```
ax3l: Only for nodal staggering of rho/j, right?
```suggestion
    // a single box with periodic bounaries). A given (i,j,k) can be present
    // in more than one box for nodal data in AMReX.
    // At the time of calling ParallelAdd, the current
```

Source/Parallelization/WarpXComm.cpp
```diff
@@ -871,33 +873,143 @@ WarpX::SyncCurrent (
...
+    // level's fp data have not been summed. Because of how ParallelAdd
+    // works, all boxes with that (i,j,k) will receive the data. So there is
+    // a double counting issue of that data existing on multiple boxes. Note
```
ax3l:
```suggestion
    // a double counting issue of those data points existing on multiple boxes. Note
```

Source/Parallelization/WarpXComm.cpp
WeiqunZhang: Right, only for non-cell-centered rho/j.

RemiLehe (approved):
Thanks a lot for this PR!

PR #3778, opened 2023-03-22
Slurm & LSF: How to Checkpoint & Terminate

ax3l:
Add a first concrete example how to gracefully checkpoint and terminate a WarpX simulation that runs into an HPC system's walltime.

If this works well, we will add this to all HPC templates.

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -195,16 +195,15 @@ We follow the same naming, but remove the ``SIG`` prefix, e.g., the WarpX signal
 
 .. tip::
 
-   For example, the following logic can be added to `Slurm batch scripts <https://docs.gwdg.de/doku.php?id=en:services:application_services:high_performance_computing:running_jobs_slurm:signals>`__ (`signal name to number mapping here <https://en.wikipedia.org/wiki/Signal_(IPC)#Default_action>`__) to write a checkpoint 6 min prior to walltime and gracefully shut down 1 min prior to walltime:
+   For example, the following logic can be added to `Slurm batch scripts <https://docs.gwdg.de/doku.php?id=en:services:application_services:high_performance_computing:running_jobs_slurm:signals>`__ (`signal name to number mapping here <https://en.wikipedia.org/wiki/Signal_(IPC)#Default_action>`__) to gracefully shut down 6 min prior to walltime.
+   If you have a checkpoint diagnostics in your inputs file, this automatically will write a checkpoint due to the default ``<diag_name>.dump_last_timestep = 1`` option in WarpX.
 
    .. code-block:: bash
 
       #SBATCH --signal=B:1@360
-      #SBATCH --signal=B:2@60
 
-      srun ...                        \
-        warpx.checkpoint_signals=HUP  \
-        warpx.break_signals=INT       \
+      srun ...                   \
+        warpx.break_signals=HUP  \
         > output.txt
 
    For `LSF batch systems <https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=options-wa>`__, the equivalent job script lines are:
@@ -212,11 +211,9 @@ We follow the same naming, but remove the ``SIG`` prefix, e.g., the WarpX signal
    .. code-block:: bash
 
       #BSUB -wa 'HUP' -wt '6'
-      #BSUB -wa 'INT' -wt '1'
 
-      jsrun ...                       \
-        warpx.checkpoint_signals=HUP  \
-        warpx.break_signals=INT       \
+      jsrun ...                  \
+        warpx.break_signals=HUP  \
         > output.txt
 
 .. _running-cpp-parameters-box:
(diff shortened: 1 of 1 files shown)


PhilMiller:
Sounds reasonable to me

roelof-groenewald:
In this example a checkpoint will be written 6 minutes before reaching the walltime limit, but then again when the simulation gracefully exits since if checkpoints are enabled, WarpX always writes a checkpoint on exit, right? Should that be noted in the example so users are not surprised by the double checkpoint?

n01r:
@roelof-groenewald, I tried this on Summit. We cannot configure two warning actions, it seems. That's because we're setting `LSB_SUB_JOB_WARNING_ACTION` with `bsub -wa`.
I also found that the termination from receiving the signal at `warpx.break_signals` does not create a checkpoint for me.

PhilMiller:
One minute may not be enough time for graceful shutdown to actually finish the preceding timestep (in which the signal is received), run through the next (after which the signal will be handled), write the checkpoint, and exit. So, the earlier 'warning' may often end up generating the last fully-written checkpoint.

On that note, @ax3l, are there mechanisms in place to validate a checkpoint, or to not finalize it in some fashion, until the data and files have been synced to storage?

ax3l:
> WarpX always writes a checkpoint on exit, right? Should that be noted in the example so users are not surprised by the double checkpoint?

Oh really? I totally forgot... In that case, we would just need to send a single graceful exit signal like 6min before walltime (and need to have checkpoints enabled).

Update: oh, that's just because we trigger every full diagnostics again on the last time step by default, and checkpoints or part of them. There is now also an option to disable that behavior: `<diag_name>.dump_last_timestep = 1/0`

> On that note, @ax3l, are there mechanisms in place to validate a checkpoint, or to not finalize it in some fashion, until the data and files have been synced to storage?

I am not aware that AMReX plotfiles, which we use for checkpointing, have some kind of journaling. We could investigate if one can keep key attributes/status sub-files delayed in AMReX until its done. Will forward to a separate Slack discussion.

ax3l:
I pushed a simplification.

roelof-groenewald (approved):
Thanks! Looks good to me!

roelof-groenewald:
> I also found that the termination from receiving the signal at `warpx.break_signals` does not create a checkpoint for me.

@n01r did you have the `dump_last_timestep` flag disabled for the run where the checkpoint was not created on the break signal?

n01r:
@roelof-groenewald nope, I never touched this parameter. It's not present in my input and all the other simulations write out full diags at the last step.

roelof-groenewald:
Thanks for bringing that up @n01r. Looking more into it, the behavior has been changed by this addition in `WarpXEvolve.cpp` (lines 412-417):

```
    // This if statement is needed for PICMI, which allows the Evolve routine to be
    // called multiple times, otherwise diagnostics will be done at every call,
    // regardless of the diagnostic period parameter provided in the inputs.
    if (istep[0] == max_step || (stop_time - 1.e-3*dt[0] <= cur_time && cur_time < stop_time + dt[0])) {
        multi_diags->FilterComputePackFlushLastTimestep( istep[0] );
    }
```

Now the break signal is no longer treated as a final step the way it used to be.
@ax3l do you think we should revert the default behavior? Maybe we can just change those lines to:
```
    if (m_dump_last_timestep) {
        multi_diags->FilterComputePackFlushLastTimestep( istep[0] );
    }
```

ax3l:
@roelof-groenewald I think that is a good idea, but for a separate PR.
Can you discuss it briefly in the developer meeting tomorrow to check with everyone?

I cannot attend in the morning but will follow-up on the PR or Slack afterwards.

For the PR here, the docs should be working now - do you want to go ahead and merge it?

Docs/source/usage/parameters.rst
```diff
@@ -193,6 +193,28 @@ We follow the same naming, but remove the ``SIG`` prefix, e.g., the WarpX signal
 
    The ``FPE`` signal should not be overwritten in WarpX, as it is `controlled by AMReX <https://amrex-codes.github.io/amrex/docs_html/Debugging.html#breaking-into-debuggers>`__ for :ref:`debug workflows that catch invalid floating-point operations <debugging_warpx>`.
 
+.. tip::
+
+   For example, the following logic can be added to `Slurm batch scripts <https://docs.gwdg.de/doku.php?id=en:services:application_services:high_performance_computing:running_jobs_slurm:signals>`__ (`signal name to number mapping here <https://en.wikipedia.org/wiki/Signal_(IPC)#Default_action>`__) to gracefully shut down 6 min prior to walltime.
+   If you have a checkpoint diagnostics in your inputs file, this automatically will write a checkpoint due to the default ``<diag_name>.dump_last_timestep = 1`` option in WarpX.
```
ax3l, line 199: Ok, this is no true until we fix it...

roelof-groenewald: I'll bring up the topic at the devs meeting tomorrow, and if it is agreed that we should recover this behavior I'll open a PR to be merged just before this one.

PR #3782, opened 2023-03-23
Workaround for nvcc

WeiqunZhang:
Fixes an issue in #3696. CUDA device lambda appears to have a limitation that explicit by-copy captures in addition to by-copy capture default would result in invalid device function at runtime.


aeriforme (approved):
It works on Summit with the [input](https://github.com/ECP-WarpX/WarpX/files/11066378/input.txt) file that initially raised the bug. 
Thanks a lot!

ax3l (approved):
Thank you for the work-around.

Documenting version range of the issue below.
CI does not seem to catch it:
- NVCC 11.0.3 runs w/ QED
- NVCC 11.8.0 GNUmake runs w/o QED
- NVHPC@21.11 NVCC/NVC++ runs w/ QED

We saw the issue on Summit with NVCC 11.3 & 11.5.2.

PR #3784, opened 2023-03-24
Refactoring of main.cpp to improve readability

lucafedeli88:
I would like to propose to improve the readability of `main.cpp` by moving some code to helper functions  and helper classes.  Specifically, in this PR, I did the following:

- I moved rocfft initialization to a new helper function, defined in the `utils` namespace
- I created a simple timer in `ablastr` , which is used to compute the wall time in WarpX
- I moved the call to MPI Finalize into a helper function (MPI initialization was already in a helper function)
- I removed unnecessary `#include`  (unnecessary after the refactoring)

what do you think?


ax3l (approved):
Love it, thank you very much!

PR #3786, opened 2023-03-24
include std::set, int template parameter

RTSandberg:
This PR fixes an issue introduced in #3768, where `std::set` is used without explicitly including the `set` library.  It also adds an explicit type value for the set keys, `std::set<int>`.  The previous PR could be build on some machines but not others (e.g. Summit).


ax3l (approved):
Thx!

ax3l (approved):

PR #3787, opened 2023-03-24
Docs: start using BibTeX in theory section

EZoni:
I think it could be a good idea to start using BibTeX for our bibliography in the Theory section of the documentation (and elsewhere, wherever needed).

This is just a small PR to show how I think we could do this. It modifies as an example the subsection "Introduction".

Besides the improved visual result, I think it would make it easier to keep the references in the docs up-to-date, by having a centralized `refs.bib` file where one can add new references.

Example of old docs:
<p align="center">
    
    
</p>

Example of new docs:
<p align="center">
    
    
</p>


ax3l (approved):
Very cool, I did not know that is possible!

Do you want to merge this first PR and add more later?

EZoni:
@ax3l Thank you, yes, I would suggest that we merge this PR and get the work started.

PR #3789, opened 2023-03-27
DD fusion anneutronic branch

jcarlsson-aed:
For deuterium-deuterium (DD) fusion there are two branches:

                  | He-3 + n (neutronic branch, 50% probability)
 d + d -> <
                  | t + p    (anneutronic branch, 50% probability)

Currently in Warpx anneutronic DD fusion only works when the species of the fusion products are tritium (or hydrogen3) and proton. If protium (or hydrogen1) are specified instead of proton, the simulation aborts with the following error message:

amrex::Abort::0::ERROR: Product species of proton-boron fusion must be of type helium3 and neutron, or tritium and proton !!!

The proposed diff allows anneutronic DD fusion to produce either protium / hydrogen1 or proton. It also corrects the error message.

It's a very small diff to the file Source/Particles/Collision/BinaryCollision/BinaryCollisionUtils.cpp: four added lines of code, and one modified line (the error message). I also added an input deck that demonstrates the issue.


jcarlsson-aed:
I've discussed this with @RemiLehe so it might be slightly less painful for him to review?

RemiLehe:
Thanks for this PR!
I removed the test file because WarpX's CI impose that any test file should also have an associated Python analysis script that checks that the results are correct.

RemiLehe (approved):

PR #3790, opened 2023-03-28
Output enabled diagnostics (and checkpoint) on interrupt by default

roelof-groenewald:
The current change will restore the earlier behavior where checkpoints are output when the simulation is interrupted by OS signals (as can be configured for jobs reaching their walltime limit, see #3778).


roelof-groenewald:
As discussed during the devs meeting, I will change this to add another flag for whether the for-loop was terminated due to an interrupt signal or not.

dpgrote (approved):
This looks good. Thanks for updating.

PhilMiller (approved):

PR #3798, opened 2023-03-31
Diagnostics: Loop up to Finest Level

### LINKED ISSUE #3716: Mesh refinement in RZ geometry
Hello there,

I tried to run simulation in RZ geometry with mesh refinement, but the code is not able to build my computational domain (I am getting a "Segfault"), although the simulation runs fine without the MR. Can you please take a look at my input file and suggest a solution of the problem? Thank you very much.

[inputRZ_MR.txt](https://github.com/ECP-WarpX/WarpX/files/10861162/inputRZ_MR.txt)

EZoni:
There are situations where the mesh refinement patch is outside the computational domain at the start of the simulation. In these situations, we need to use the current finest level available, rather than the maximum level set in the input, when we loop over mesh refinement levels.

This is an attempt to start implementing #3716.

### DIFF
--- a/Source/Diagnostics/Diagnostics.cpp
+++ b/Source/Diagnostics/Diagnostics.cpp
@@ -376,6 +376,11 @@ Diagnostics::InitDataAfterRestart ()
 void
 Diagnostics::InitData ()
 {
+    auto& warpx = WarpX::GetInstance();
+
+    // Get current finest level available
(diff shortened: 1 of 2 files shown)


ax3l (approved):
This looks like a bug in diagnostics to me and should be generally correct. We fixed something similar in ImpactX a while ago. Thx!

PR #3802, opened 2023-04-03
Make verboncoeur correction optional

dpgrote:
This PR makes the Verboncoeur correction of the charge and current density on the axis when using RZ optional. It adds the new input parameter `boundary.verboncoeur_axis_correction`, which defaults to true.

In the development of the implicit energy conserving mover, applying the correction spoils the energy conservation.


ax3l (approved):
This LGTM, thanks a lot!

PR #3809, opened 2023-04-05
Charge on EB: Use `fp` field instead of `aux`

RemiLehe:
When computing the `ChargeOnEB`, we use the electric field. However, the current PR is using the `aux` version of `E`. This is not ideal, since the typical PIC loop is:
```
loop over iterations:
   - Compute aux from fp
   - Update fp (using the Maxwell equations or the Poisson equation)
   - Run the reduced diagnostics
```
Therefore, when computing the reduced diagnostics, the `aux` field is outdated by one step. In addition, when using momentum-conserving gather, the `aux` field is a smoothed version of `fp`, and it is more accurate to compute the above diagnostic from the unsmoothed version.

Therefore, the current PR uses the `fp` fields instead of the `aux` fields.

This PR should not affect the automated test, which uses `energy-conserving` gather - since `fp` and `aux` are actually the same array (on level 0) when using `energy-conserving` gather.

### DIFF
--- a/Source/Diagnostics/ReducedDiags/ChargeOnEB.cpp
+++ b/Source/Diagnostics/ReducedDiags/ChargeOnEB.cpp
@@ -93,9 +93,9 @@ void ChargeOnEB::ComputeDiags (const int step)
     int const lev = 0;
 
     // get MultiFab data at lev
-    const amrex::MultiFab & Ex = warpx.getEfield(lev,0);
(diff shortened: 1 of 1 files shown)


EZoni (approved):
Thanks for this fix, @RemiLehe! The PR looks good to me. I see that many CI checks were skipped, could you trigger them again (e.g., `git commit --amend` to change the hash tag and then force-push)?

PR #3810, opened 2023-04-06
Fix loading particles from openPMD

RemiLehe:
This PR fixes a few bugs related to particle loading in openPMD:

- When reading particles using openPMD, we forgot to read the quantity `positionOffset`. In practice, this can result in having the loaded particles outside of the simulation domain, whereas the particles are physically supposed to be inside the domain.

- We were incorrectly assuming that the mass of the species is the electron mass.


ax3l (approved):
Excellent, thanks!

PR #3807, opened 2023-04-05
Code cleanup: use `DepositCurrent()` in `AddSpaceChargeFieldLabFrame()`

roelof-groenewald:
Simple code clean-up: Functionality in `AddSpaceChargeFieldLabFrame` is currently exactly duplicated in `MultiParticleContainer::DepositCharge`.

PR #2446, opened 2021-10-18
[WIP] Use the function SyncRho and the array rho_fp in the relativistic Electrostatic solver

RemiLehe:
This is a follow-up on #1811.
It uses the array `rho_fp` and the function `SyncRho` in the relativistic Electrostatic solver.
As a consequence, filtering is automatically taken into account.

PR #3814, opened 2023-04-07
`SyncRho`: Pass References To Charge MultiFabs

RemiLehe:
This PR is similar to #3277. It allows to call `SyncRho` on arbitrary multifabs. 

This would be useful in order to cleanup the relativistic Poisson solver. (Related PRs: #3807,  #2446)

It would also be useful for the charge density field diagnostic (and species-wise charge density diagnostic), which currently deposits `rho` in a different multifab than `rho_fp` and `rho_cp`.

### DIFF
--- a/Source/WarpX.H
+++ b/Source/WarpX.H
@@ -776,7 +776,9 @@ public:
         const amrex::Vector<std::array<std::unique_ptr<amrex::MultiFab>,3>>& J_fp,
         const amrex::Vector<std::array<std::unique_ptr<amrex::MultiFab>,3>>& J_cp);
 
-    void SyncRho ();
+    void SyncRho (
+        const amrex::Vector<std::unique_ptr<amrex::MultiFab>>& charge_fp,
+        const amrex::Vector<std::unique_ptr<amrex::MultiFab>>& charge_cp);
 
     amrex::Vector<int> getnsubsteps () const {return nsubsteps;}
     int getnsubsteps (int lev) const {return nsubsteps[lev];}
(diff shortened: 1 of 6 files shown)


Source/WarpX.H
```diff
@@ -778,6 +778,10 @@ public:
 
     void SyncRho ();
```
EZoni, line 780: I would expect that, similarly to what was done in #3277, it should be possible to remove this function and update the Python interface accordingly. I guess this would require to modify both Source/Python/WarpXWrappers.H and Source/Python/WarpXWrappers.cpp. Maybe you initially modified only the cpp file and that's why it seemed like that change needed to be reverted in order to fix the Python interface?

RemiLehe: Yes, I decided not to change the Python interface after realizing that it is being called here:
https://github.com/ECP-WarpX/WarpX/blob/development/Python/pywarpx/_libwarpx.py#L1219
Changing this would require adding a Python interface to extract the list of `rho_fp` and `rho_cp` at each level, so that the Python code would instead read:
```
self.libwarpx_so.warpx_SyncRho( self.libwarpx_so.warpx_GetRhoFP(), self.libwarpx_so.warpx_GetRhoCP() )
```
While there are existing Python interfaces that return the `rho_fp` and `rho_cp` for **one given level**, we don't have one yet for the full list of levels. I think that this could be a follow-up PR, potentially after the transition to `pyamrex`.
Any opinion on your side?

RemiLehe: Incidentally, note that we have a call to `warpx_SyncCurrent()` here:
https://github.com/ECP-WarpX/WarpX/blob/development/Python/pywarpx/timestepper.py#L60
which in principle should fail because it does not conform to the syntax defined in #3277 - but it does not seem that this file is being tested. Maybe we could discuss this offline.

dpgrote: I would suggest keeping `SyncRho()` for now and deal with this after transitioning to pyamrex.
Also, for now I wouldn't worry about timestepper.py. It has probably never been really tested and it likely to have other problems. This is also something that can be fixed after pyamrex.

EZoni (approved):
Thanks, discussed offline, looks good to me!

PR #3063, opened 2022-04-26
Add accelerator lattice, starting with quadrupoles

dpgrote:
This PR is the initial implementation of an accelerator lattice, starting with hard edge quadrupoles. The implementation follows the scheme in Warp, where the lattice is laid out along the `z` axis. Each lattice element is specified by an extent in `z` and information about its field structure.

To locate which element particles are in, a grid is laid out in `z` and at each grid cell, the index of the nearest element is saved. Then for each particle, a look up is done from the table. This grid is updated as needed in a moving frame and with load balancing.

For the GPU, host level and device level classes are defined. The data is read in and setup in the host level class and the needed data copied to the GPU. The device level class only holds pointers to the needed data. This hasn't been tested but does compile.

To do:
 - [x] Add a CI test in the moving window
 - [x] Test on GPU
 - [x] Add handling of the boosted frame
 - [x] Add plasma lens using the new lattice infrastructure
 - [x] Add documentation

This PR assumes that elements within a type do not overlap each other. A future PR can be added to allow the overlapping.

PR #3815, opened 2023-04-07
Lattice: Line Docs & Rename

ax3l:
Rename the `lattice` element type to `line` and add documentation for it.

Follow-up to #3063

In sync with ImpactX:
- https://github.com/ECP-WarpX/impactx/pull/335

### DIFF
--- a/Source/AcceleratorLattice/AcceleratorLattice.cpp
+++ b/Source/AcceleratorLattice/AcceleratorLattice.cpp
@@ -44,8 +44,9 @@ AcceleratorLattice::ReadLattice (std::string const & root_name, amrex::ParticleR
     bool reverse = false;
     pp_lattice.queryAdd("reverse", reverse);
 
-    if (reverse)
+    if (reverse) {
         std::reverse(lattice_elements.begin(), lattice_elements.end());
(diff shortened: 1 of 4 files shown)


Source/AcceleratorLattice/AcceleratorLattice.cpp
```diff
@@ -35,6 +41,12 @@ AcceleratorLattice::ReadLattice (std::string const & root_name, amrex::ParticleR
         m_lattice_defined = true;
     }
 
+    bool reverse = false;
+    pp_lattice.queryAdd("reverse", reverse);
+
+    if (reverse)
+        std::reverse(lattice_elements.begin(), lattice_elements.end());
```
dpgrote, line 48: Here's my pickiness - can you wrap this in braces `if () {}`?

dpgrote (approved):
Thanks! This looks good (except for my picky comment).

Source/AcceleratorLattice/AcceleratorLattice.cpp
ax3l: no prob, done!

PR #3820, opened 2023-04-07
Fix FieldEnergy diagnostic for RZ

dpgrote:
Several fixes are needed. The different cell volumes on axis and at the outer boundary are calculated. Also, when the boundary is not periodic, for nodal quantities, the value on the last grid cell is included. Note that this in general won't matter since the nodal quantities on the boundary will be zero for PEC.

### DIFF
--- a/Source/Diagnostics/ReducedDiags/FieldEnergy.cpp
+++ b/Source/Diagnostics/ReducedDiags/FieldEnergy.cpp
@@ -158,8 +158,11 @@ void FieldEnergy::ComputeDiags (int step)
 amrex::Real
 FieldEnergy::ComputeNorm2RZ(const amrex::MultiFab& field, const int lev)
 {
-    const std::array<amrex::Real,3>& dx = WarpX::CellSize(lev);
-    const amrex::Real dr = dx[0];
+    // get a reference to WarpX instance
+    auto & warpx = WarpX::GetInstance();
+
+    Geometry const & geom = warpx.Geom(lev);
+    const amrex::Real dr = geom.CellSize(0);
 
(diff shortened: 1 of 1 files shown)


Source/Diagnostics/ReducedDiags/FieldEnergy.cpp
```diff
@@ -177,20 +180,35 @@ FieldEnergy::ComputeNorm2RZ(const amrex::MultiFab& field, const int lev)
...
+                // on the upper edge of the domain
+                tb.enclosedCells(idir);
+            }
+        }
+
+        reduce_ops.eval(tb, ncomp, reduce_data,
             [=] AMREX_GPU_DEVICE (int i, int j, int k, int n) -> ReduceTuple
             {
                 const amrex::Real r = rmin + (i - irmin)*dr;
+                amrex::Real volume_factor = r;
```
ax3l, line 204: Hi @dpgrote , let's calculate this factor before the `eval` and capture the variable `volume_factor` automatically in the lambda.

Source/Diagnostics/ReducedDiags/FieldEnergy.cpp
```diff
@@ -177,20 +180,35 @@ FieldEnergy::ComputeNorm2RZ(const amrex::MultiFab& field, const int lev)
...
+                if (r == 0._rt) {
+                    volume_factor = dr/8._rt;
```
ax3l, line 206: We could move the constant dr/8._rt outside of the kernel and pass it by value, but let's keep it here for clarity.

ax3l (approved):
Thank you! :)

Issue #3797, opened 2023-03-31
Conflicting error message - PB fusion


PhysicsDan:
Hi - great work implementing nuclear fusion!

Just to bring to your attention - this error message has a small typo in it.

When I set `particle.species_type` to `alpha` (or `helium4`) I get the error
```
0::Assertion `product_species.AmIA<PhysicalSpecies::helium>()' failed, file "src/warpx/Source/Particles/Collision/BinaryCollision/BinaryCollisionUtils.cpp", line 93, Msg: 
### ERROR   : ERROR: Product species of proton-boron fusion must be of type
#            alpha
```
While the code will start running when I set the `species_type` to `helium` (this matches the test inputs in this repo too).

Thanks,
Dan

RemiLehe:
Thanks for raising this issue! 
This is now corrected in https://github.com/ECP-WarpX/WarpX/pull/4480

PR #3823, opened 2023-04-09
Correction to proton-boron fusion error check addressing issue #3797

eebasso:
Changed PhysicalSpecies "helium" to "helium4" and made the error message string mention "helium4" instead of "alpha." It appears that PhysicalSpecies::alpha is currently unused. This is intended to address issue #3797.


RemiLehe (approved):

RemiLehe:
Thanks for this PR!

I actually also allowed `alpha` in addition to `helium4`. (See [here](https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/SpeciesPhysicalProperties.cpp#L176) for the difference between `alpha` and `helium4` in WarpX: `helium4` is the atom and includes the mass of the electrons, whereas `alpha` is the nucleus only.)

I also updated the proton-boron fusion test, since it was (incorrectly) using `helium` instead of `helium4`/`alpha`.

eebasso:
> Thanks for this PR!
> 
> I actually also allowed `alpha` in addition to `helium4`. (See [here](https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/SpeciesPhysicalProperties.cpp#L176) for the difference between `alpha` and `helium4` in WarpX: `helium4` is the atom and includes the mass of the electrons, whereas `alpha` is the nucleus only.)
> 
> I also updated the proton-boron fusion test, since it was (incorrectly) using `helium` instead of `helium4`/`alpha`.

Thank you for the response. I was thinking of doing that change at first, but it seemed to me that the "helium4 OR alpha" check isn't currently done elsewhere in BinaryCollisionUtils, and so I was trying to be consistent. For example, deuterium-helium3 fusion allows helium4 as a product species, but not alpha. I assume that had to do with conserving the number of electrons.

Perhaps every instance of helium4 check needs to have an alpha check alternative? Technically, neither  3 x helium4 nor 3 x alpha conserves the number of electrons from proton + boron11 fusion, but 3*helium4 comes closer. The process of hydrogen1 + boron11 -> 3 x helium4 would conserve charge.

Edit: another alternative I can think of is to make one of `alpha` or `helium4` as an alternative name to other in the map `string_to_species` defined in SpeciesPhysicalProperties.cpp.

Looking at the map `species_to_properties,` it seems that `helium4` is still considered a +2 charged particle, but it has a larger mass than `alpha` due to two electrons.
```
        {PhysicalSpecies::helium4, Properties{
             amrex::Real(4.00260325413) * PhysConst::m_u,
             amrex::Real(2) * PhysConst::q_e}},
        {PhysicalSpecies::alpha, Properties{
             amrex::Real(4.00260325413) * PhysConst::m_u - amrex::Real(2) * PhysConst::m_e,
             amrex::Real(2) * PhysConst::q_e}},
```

RemiLehe:
That's a very good point. 
There are indeed some inconsistencies between the mass and the charge of the species. I will discuss this with the other WarpX developers.

eebasso:
Excellent. I am happy to assist you in any way I can.

RemiLehe:
Superseded by https://github.com/ECP-WarpX/WarpX/pull/4480
(as discussed offline in the WarpX developer meeting)

PR #2687, opened 2021-12-17
Fixes removing implicit double to float conversions

dpgrote:
I compiled with single precision on my Mac and found a large number of warnings about implicit conversion from double to float. This PR fixes most of them. Most are missing `_rt`. There are a few places where `double` was explicitly used - were these intentional?

The routine `amrex::second()` is explicitly double so causes warnings when used with `amrex::Real` variables when that is float. This will be fixed in a separate PR since the correct fix is to use doubles for the timing variables.

I did not fix CurrentDeposition.H, which is going to be messy. The `s[xyz]_new` and `_old` are explicitly double but used in assignments to floats. Adding the static casts will make ugly code since many will be needed.

PR #3824, opened 2023-04-11
Add replacement of std::pow for integer exponents known at compile time

lucafedeli88:
This PR adds a replacement of std::pow(x,p) to be used when p is integer and known at compile time.
On my laptop, for p > 2 and as high as 20 I have observed that `ipow<p>(x)` is roughly one order of magnitude faster with respect to `std::pow(x,p)` if applied to a vector (`std::pow` is usually a function call for p != 1 or 2).

(note: if `-Ofast` is used, there is no speed-up with clang or g++)


AlexanderSinn:
It might be worth considering adding the following divide and conquer approach for large exponents:
```
if constexpr (IPower % 2 == 0 && IPower >= 10)
    return ipow<2>(ipow<IPower/2>(x));
```

lucafedeli88:
> It might be worth considering adding the following divide and conquer approach for large exponents:
> 
> ```
> if constexpr (IPower % 2 == 0 && IPower >= 10)
>     return ipow<2>(ipow<IPower/2>(x));
> ```

this is a very good idea! On my laptop it improves performances starting from `IPower ~ 7`. 
I would even omit the `IPower >= 10`, since performance loss for smaller exponents seems negligible.

ax3l:
> (note: if -Ofast is used, there is no speed-up with clang or g++)

When we recently visited this topic, we arrived at the conclusion that modern compilers in `-O3` (our default) and fastmath do identify compile-time int arguments to pow and choose the fastest implementation instead of calling into the `cmath` stdlib unconditionally with two float arguments of pow. This could explain why `-Ofast` shows no difference for you in performance.

X-ref was here https://github.com/ECP-WarpX/WarpX/pull/2687#issuecomment-1007359428 but we unfortunately seem to have finished the discussion on Slack. There are SO posts documenting the behavior of modern compilers though.

We turn fastmath flags on on GPU by default and should do so on CPU, too: https://github.com/AMReX-Codes/amrex/issues/1769

ax3l:
If we want to add this, then I think it should go upstream in AMReX Math namespace maybe :)
cc @WeiqunZhang

WeiqunZhang:
@lucafedeli88 When you test it, are you testing `std::pow(x, 5)` or `std::pow(x,5.0)`? The warpx code that has changes was using `5_prt` (i.e., a floating point number). The standard library has overloaded pow for `std::pow(double,int)`.

WeiqunZhang:
https://github.com/gcc-mirror/gcc/blob/0c5e64c4249322a178e1a0e843874e4d6b43b992/libstdc%2B%2B-v3/include/c_std/cmath#L360 I can believe the template version is faster because the integer is known at compile time. But it's hard to believe it could be 20x faster.

lucafedeli88:
> @lucafedeli88 When you test it, are you testing `std::pow(x, 5)` or `std::pow(x,5.0)`? The warpx code that has changes was using `5_prt` (i.e., a floating point number). The standard library has overloaded pow for `std::pow(double,int)`.

@WeiqunZhang ,  it seems that unless `-Ofast` is used, `std::pow(x, 5)` is still a function call in the assembly code.
For instance, with `-O3` and g++ 12.2 :  https://godbolt.org/z/bfqqhne7c
If I use `-Ofast` instead, I get exactly the same assembly:  https://godbolt.org/z/odzKWeebh
(but we don't use `-Ofast` by default on CPUs).

Specifically, the code that I use for the tests is:
```
#include <algorithm>
#include <cmath>
#include <chrono>
#include <iostream>
#include <random>
#include <vector>

template <unsigned int Power, typename RealType>
inline constexpr RealType ipow (RealType x) noexcept
{
    static_assert(Power >= 1, "ERROR!");
    if constexpr (Power == 1)
        return x;
    else if constexpr (Power == 2)
        return x*x;
    else if constexpr (Power%2 == 0)
        return ipow<2>(ipow<Power/2>(x));
    else
        return x * ipow<Power-1>(x);
}


template <unsigned int Power, unsigned int HowMany>
double do_test (std::mt19937& rng)
{	
    auto data = std::vector<double>(HowMany);
    auto res1 = std::vector<double>(HowMany);
    auto res2 = std::vector<double>(HowMany);
    auto res3 = std::vector<double>(HowMany);
    auto fdist = std::uniform_real_distribution<double>(-1000.0, 1000.0);
    std::generate(data.begin(), data.end(), [&](){return fdist(rng);});
    
    const auto s1 = std::chrono::system_clock::now();
    std::transform(data.begin(), data.end(), res1.begin(), 
    	[&](const auto x){return std::pow(x, Power);});
    const auto e1 = std::chrono::system_clock::now();
    const std::chrono::duration<double> d1 = e1 - s1;

    const auto s2 = std::chrono::system_clock::now();
    std::transform(data.begin(), data.end(), res2.begin(), 
    	[&](const auto x){return ipow<Power>(x);});
    const auto e2 = std::chrono::system_clock::now();
    const std::chrono::duration<double> d2 = e2 - s2;
    
    std::cout << "POWER: " << Power 
        << "    --> std::pow:  " << d1.count() 
        << "  ipow: " << d2.count()
        << "  speedup:  " << d1.count()/d2.count()
        << "\n";
    
    auto udist = std::uniform_int_distribution<int>{0, HowMany};

    // this is done to avoid that the compiler optimizes away the code
    return res1[udist(rng)] + res2[udist(rng)];
}

int main()
{
    auto rng = std::mt19937{};
    double checksum = 0.0;
    
    constexpr auto how_many = 10'000'000;
    
    checksum += do_test<1,how_many>(rng);
    checksum += do_test<2,how_many>(rng);
    checksum += do_test<3,how_many>(rng);
    checksum += do_test<4,how_many>(rng);
    checksum += do_test<5,how_many>(rng);
    checksum += do_test<6,how_many>(rng);
    checksum += do_test<7,how_many>(rng);
    checksum += do_test<8,how_many>(rng);
    checksum += do_test<9,how_many>(rng);
    checksum += do_test<10,how_many>(rng);
    checksum += do_test<11,how_many>(rng);
    checksum += do_test<12,how_many>(rng);
    checksum += do_test<13,how_many>(rng);
    checksum += do_test<14,how_many>(rng);
    checksum += do_test<15,how_many>(rng);
    checksum += do_test<16,how_many>(rng);
    checksum += do_test<17,how_many>(rng);
    checksum += do_test<18,how_many>(rng);
    checksum += do_test<19,how_many>(rng);
    checksum += do_test<20,how_many>(rng);
 
    std::cout << checksum;
    return 0;
}
```

On my laptop (I have gcc 12.2), if I compile with `-O3` I get:

```
POWER: 1    --> std::pow:  0.00493031  ipow: 0.00476413  speedup:  1.03488
POWER: 2    --> std::pow:  0.00819809  ipow: 0.00811547  speedup:  1.01018
POWER: 3    --> std::pow:  0.152703  ipow: 0.00820557  speedup:  18.6097
POWER: 4    --> std::pow:  0.153436  ipow: 0.0081764  speedup:  18.7658
POWER: 5    --> std::pow:  0.153256  ipow: 0.00826171  speedup:  18.5501
POWER: 6    --> std::pow:  0.153436  ipow: 0.0088534  speedup:  17.3308
POWER: 7    --> std::pow:  0.154391  ipow: 0.00912723  speedup:  16.9155
POWER: 8    --> std::pow:  0.157955  ipow: 0.00860278  speedup:  18.361
POWER: 9    --> std::pow:  0.153521  ipow: 0.00833349  speedup:  18.4221
POWER: 10    --> std::pow:  0.153022  ipow: 0.00835848  speedup:  18.3074
POWER: 11    --> std::pow:  0.152927  ipow: 0.00834943  speedup:  18.3159
POWER: 12    --> std::pow:  0.153212  ipow: 0.00831651  speedup:  18.4226
POWER: 13    --> std::pow:  0.153075  ipow: 0.00834966  speedup:  18.3331
POWER: 14    --> std::pow:  0.153219  ipow: 0.00840341  speedup:  18.2329
POWER: 15    --> std::pow:  0.153442  ipow: 0.00847801  speedup:  18.0988
POWER: 16    --> std::pow:  0.153615  ipow: 0.00832498  speedup:  18.4522
POWER: 17    --> std::pow:  0.152917  ipow: 0.00834704  speedup:  18.3199
POWER: 18    --> std::pow:  0.153045  ipow: 0.00840551  speedup:  18.2077
POWER: 19    --> std::pow:  0.153002  ipow: 0.00847431  speedup:  18.0548
POWER: 20    --> std::pow:  0.152929  ipow: 0.00839568  speedup:  18.2152
```

The speedup disappears if I compile with `-Ofast`:

```
POWER: 1    --> std::pow:  0.00487771  ipow: 0.0047411  speedup:  1.02881
POWER: 2    --> std::pow:  0.00825916  ipow: 0.00809785  speedup:  1.01992
POWER: 3    --> std::pow:  0.0083529  ipow: 0.00818178  speedup:  1.02091
POWER: 4    --> std::pow:  0.00834421  ipow: 0.00823907  speedup:  1.01276
POWER: 5    --> std::pow:  0.00838926  ipow: 0.00821934  speedup:  1.02067
POWER: 6    --> std::pow:  0.00847286  ipow: 0.00834778  speedup:  1.01498
POWER: 7    --> std::pow:  0.00851694  ipow: 0.00835189  speedup:  1.01976
POWER: 8    --> std::pow:  0.00897998  ipow: 0.00883556  speedup:  1.01635
POWER: 9    --> std::pow:  0.0127758  ipow: 0.00952083  speedup:  1.34188
POWER: 10    --> std::pow:  0.00862572  ipow: 0.00855036  speedup:  1.00881
POWER: 11    --> std::pow:  0.00874994  ipow: 0.00864405  speedup:  1.01225
POWER: 12    --> std::pow:  0.00856567  ipow: 0.00840156  speedup:  1.01953
POWER: 13    --> std::pow:  0.00857595  ipow: 0.00845045  speedup:  1.01485
POWER: 14    --> std::pow:  0.00856031  ipow: 0.00841503  speedup:  1.01726
POWER: 15    --> std::pow:  0.00855174  ipow: 0.00855129  speedup:  1.00005
POWER: 16    --> std::pow:  0.00850642  ipow: 0.00836502  speedup:  1.0169
POWER: 17    --> std::pow:  0.00871724  ipow: 0.00845794  speedup:  1.03066
POWER: 18    --> std::pow:  0.00860011  ipow: 0.00842802  speedup:  1.02042
POWER: 19    --> std::pow:  0.00991384  ipow: 0.0102766  speedup:  0.964703
POWER: 20    --> std::pow:  0.0103174  ipow: 0.00963396  speedup:  1.07094
```
I have the same results with clang++ 14.

The documentation of the GNU compiler states that:
```
Built-in Function: double __builtin_powi (double, int)
Built-in Function: float __builtin_powif (float, int)
Built-in Function: long double __builtin_powil (long double, int)

    Returns the first argument raised to the power of the second. Unlike the pow function no guarantees about precision and rounding are made. 
```
therefore I guess that the compiler is allowed to use the optimized version only if fast math is enabled.

WeiqunZhang:
Interesting. I always thought the compiler will use the fast way if the exponent is an integer type. https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html
```
Built-in Function: double __builtin_powi (double, int)
Built-in Function: float __builtin_powif (float, int)
Built-in Function: long double __builtin_powil (long double, int)

    Returns the first argument raised to the power of the second. Unlike the pow function no guarantees about precision and rounding are made. 
```
The last sentence is probably why the `__builtin_powi` is not used unless it's fast math. If `__builtin_powi` is used, even `-O1` is sufficient. https://godbolt.org/z/P3844cvvq

Yes, It's a good idea to add this to `amrex::Math`. It does not even need to use `AMREX_GPU_DEVICE` because it's `constexpr`.

lucafedeli88:
> Interesting. I always thought the compiler will use the fast way if the exponent is an integer type. https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html
> 
> ```
> Built-in Function: double __builtin_powi (double, int)
> Built-in Function: float __builtin_powif (float, int)
> Built-in Function: long double __builtin_powil (long double, int)
> 
>     Returns the first argument raised to the power of the second. Unlike the pow function no guarantees about precision and rounding are made. 
> ```
> 
> The last sentence is probably why the `__builtin_powi` is not used unless it's fast math. If `__builtin_powi` is used, even `-O1` is sufficient. https://godbolt.org/z/P3844cvvq
> 
> Yes, It's a good idea to add this to `amrex::Math`. It does not even need to use `AMREX_GPU_DEVICE` because it's `constexpr`.

I will prepare a PR to add this function to `amrex::Math`!

lucafedeli88:
I am closing this PR because, as agreed, I've opened a PR to include this function in AMReX: https://github.com/AMReX-Codes/amrex/pull/3251

PR #3825, opened 2023-04-11
amrex.the_arena_is_managed=0

### LINKED ISSUE #2009: amrex.the_arena_is_managed=0

ax3l:
Disable managed memory by default for performance and non-bugness.

Please disable explicitly if needed for some of the current scripting tasks, e.g., in the current Python bindings.
```py
# Set up simulation
sim = picmi.Simulation(
    solver = solver,
    max_steps = max_steps,
    # ...
    warpx_amrex_the_arena_is_managed=1  # overwrite the new default if needed
)
```

Close #2009

Tests:
- [x] 3D LWFA test w/o and w/ openPMD ran (CUDA 11.7)

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -161,7 +161,13 @@ Overall simulation parameters
     When running on GPUs, memory that does not fit on the device will be automatically swapped to host memory when this option is set to ``0``.
     This will cause severe performance drops.
     Note that even with this set to ``1`` WarpX will not catch all out-of-memory events yet when operating close to maximum device memory.
-    `Please also see the documentation in AMReX <https://amrex-codes.github.io/amrex/docs_html/GPU.html#inputs-parameters>`_.
+    `Please also see the documentation in AMReX <https://amrex-codes.github.io/amrex/docs_html/GPU.html#inputs-parameters>`__.
+
+* ``amrex.the_arena_is_managed``  (``0`` or ``1``; default is ``0`` for false)
+    When running on GPUs, device memory that is accessed from the host will automatically be transferred with managed memory.
+    This is useful for convenience during development, but has sometimes severe performance and memory footprint implications if relied on (and sometimes vendor bugs).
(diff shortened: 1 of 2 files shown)


dpgrote:
Presumably, this is something a user would never set, but would only be used temporarily by someone developing new code before the memory management was added.
When is the managed memory needed when the Python interface is being used?

ax3l:
I think it might be used if we pass some raw pointers out currently, but I am not sure what people do in their private scripts...

Yes, managed memory should just be used for hacking/prototyping, due to the many side effects and performance implications.

dpgrote:
The wrapper routine `getMultiFabPointers` returns a list of pointers to MFs, using `mf[mfi].dataPtr()`, and returns the list to Python. When on GPU, is that `dataPtr` a pointer to the GPU memory and not valid on the CPU? With managed memory, does a call to `dataPtr` on the CPU force a copy of the data from the GPU to the CPU?
Maybe there's no need to worry about this since the pyamrex wrapper will handle such details more precisely.
I'll approve this PR.

dpgrote (approved):
This looks good. Thanks!

ax3l:
Hi Dave, for you question above:

Exactly. For the case of a device memory fab, the CPU code accessing the device pointer triggered a memory migration due to managed memory.
With this PR, will need to pass `amrex.the_arena_is_managed=1` now in Python scripts that directly access data via its GPU pointer from the CPU.

So follow-ing #3590 by @clarkse, I think this is the PICMI syntax when needed:
```py
# Set up simulation
sim = picmi.Simulation(
    solver = solver,
    max_steps = max_steps,
    # ...
    warpx_amrex_the_arena_is_managed=1)
```

Managed memory does the migration currently instead of causing an error (segfault). In pyAMReX, we can implement an explicit copy helper like this (or also rely on managed memory where we want to):
https://github.com/ECP-WarpX/impactx/blob/30f93ec73ef30343d436ac51380bd38f928f8004/tests/python/test_charge_deposition.py#L59-L70

PR #3832, opened 2023-04-12
Doc: Frontier Module Updates

ax3l:
Also fixes the HDF5 workaround for the older Frontier Cray Modules.

cc @SeverinDiederichs @lucafedeli88 both of you are currently on LUMI, let's check if similar modules are also available there and update the docs, too :)


ax3l:
Oh, problem solved with `cray-hdf5-parallel/1.12.2.3` & `cmake/3.23.2` it seems. Will update PR with new module update.

ax3l:
I tested this, so I'll merge it for now :)

PR #3835, opened 2023-04-12
Doc: Perlmutter CFS for SW & Optimas

ax3l:
Updates the installation location of software to use CFS. HOME has only 40GB of quota, which can fill quickly when pulling in large ML packages.

Also adds optimas for GPU including pytorch dependency.

cc @AngelFP


RemiLehe:
Should we wait for `optimas` to be `open-source` before merging this?

ax3l:
I think this is fine to merge - we can update the last line to simplify after it's released :)

PR #3837, opened 2023-04-13
Fix particle reflecting boundaries in RZ

### LINKED ISSUE #3821: Found a bug when using reflecting BC in 2D RZ
I found some particles that are supposed to be reflected with a specular reflection are not, their reflected speeds are somehow reduced such that a number of particles are accumulated and trapped on a very thin layer at the bigger r boundary.

As can be seen in this gif after about 6000 time steps.

zoom1

As well as in this zoomed gif.

zoom

I was using this picmy.py script:
[picmi.txt](https://github.com/ECP-WarpX/WarpX/files/11182317/picmi.txt)
in which electron's do_not_gather=1, such that they are moving without feeling the fields.

The running output reads:
[output.txt](https://github.com/ECP-WarpX/WarpX/files/11182320/output.txt)

I used the latest WarpX cloned yesterday.

Could someone help me take a look, it should be a small fix I believe.

@RemiLehe @dpgrote

dpgrote:
This PR has two separate changes.

The first fixes the reflection at the boundary of the velocity with RZ. Previously, the code was only changing the sign of `ux`. But this is not correct. Now the code changes the sign of `ur` and recalculates `ux` and `uy`. This fixes the bug reported in issue #3821. Also, a comment was added as a reminder that the reflection of the position, `r = 2*rmax - r` is only an approximation in RZ.

The second is allowing setting of the input parameter `boundary.reflect_all_velocities` from the PICMI interface.

Fix #3821

### DIFF
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -537,6 +537,10 @@ class CylindricalGrid(picmistandard.PICMI_CylindricalGrid):
 
     warpx_potential_hi_z: float, default=0.
        Electrostatic potential on the upper longitudinal boundary
+
+    warpx_reflect_all_velocities: bool default=False
+        Whether the sign of all of the particle velocities are changed upon
+        reflection on a boundary, or only the velocity normal to the surface
     """
     def init(self, kw):
         self.max_grid_size = kw.pop('warpx_max_grid_size', 32)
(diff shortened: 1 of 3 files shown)


Yin-YinjianZhao:
I took a try.
This one is by using this PR, and it looks good, the bug should have been scussefuly fixed.
zoom
Compared with the one with the bug.
zoom

Yin-YinjianZhao (approved):

Yin-YinjianZhao:
@dpgrote BTW, I don't quite understand this:
```
    warpx_reflect_all_velocities: bool default=False
        Whether the sign of all of the particle velocities are changed upon
        reflection on a boundary, or only the velocity normal to the surface
```
For my case, should I turn it on or not? Will it affect much?

dpgrote:
You would have to experiment with `warpx_reflect_all_velocities`. This was put in for a different project that is studying sheath behavior. For the boundary opposite to the sheath, which is effectively in the bulk of the plasma, it had been found that the plasma is better behaved and more quiescent near that boundary if all of the velocities are reversed upon reflection.

RemiLehe (approved):

RemiLehe:
Thanks for fixing this!

PR #3840, opened 2023-04-14
Doc: HPSS Permissions

ax3l:
Document to change the default permissions of HPSS archives to be group (as in *project*) readable.


EZoni (approved):
Thank you, @ax3l!

n01r (approved):
Thanks, LGTM :)

PR #3842, opened 2023-04-15
Set guard cells in corners with mixed boundary conditions

### LINKED ISSUE #3900: PEC boundaries: Copy values in guard cells
With PEC boundaries, charge and current density deposited in guard cells outside the domain (inside PEC boundaries) are either reflected back into the simulation domain or subtracted from internal cells (to represent image charge contributions). Currently, the values in guard cells at the corner of simulation boxes are not gathered during the `SumBoundary` call, which means that small amounts of density is lost in those overlapping regions.

[cc. @WeiqunZhang ]

dpgrote:
In the evolve loop, whenever fields are updated, the WarpX specific boundary conditions are applied first (e.g. PEC) and then the periodic (in a call to the appropriate `FillBoundary` routine). However, the `FillBoundary` routines will only fill guard cells using valid cells. For mixed BCs then, the guard cells in the corners don't get set to the periodic repeat of the side guard cells along the periodic direction. So, since WarpX applies its BCs first, the changes that are made in the side guard cells are not copied to the corners along the periodic direction, leaving them either zero or with incorrect values.

The fix is to call the PEC boundary routines after the `FillBoundary` routines, moving them out of the `EvolveE` and `EvolveB` routines and making direct calls to them.

Fix #3900

Tasks:
- [x] Move `ApplyEfieldBoundary` for FDTD
- [x] Move `ApplyBfieldBoundary` for FDTD
- [x] Do the same changes of PSATD
- [x] Fix CI benchmarks as needed. With careful checking, the tests were failing because the development version was not correct, with the code using corner cells that had not been updated.


WeiqunZhang:
There are amrex functions for filling ghost cells at the corner of periodic/non-periodic boundaries. https://github.com/AMReX-Codes/amrex/blob/68aa69a11750b3f6ac36ec13b26249f0be80d527/Src/Base/AMReX_FabArray.H#LL1098C35-L1098C43 Not sure if that could be an easier fix than the changes in this PR.

There is also a known issue with PEC boundaried. https://github.com/ECP-WarpX/WarpX/issues/3900. Fixing that by passing the correct number of ghost cells to SumGuargCells should also fix the corners.

(Admittedly, I have not read the changes in this PR in detail.)

dpgrote:
@WeiqunZhang `FillBoundary` is already being called with the number of ghost cells specified. This seems to be the same as calling `EnforcePeriodicity` with nghost since both are calling `FBEP_nowait`. One of the reasons I like my method is that it puts the handling of the boundary in one place, rather than scattered in different routines. In fact, I think I'll update this pull request to put the `Apply*FieldBoundary` in the 'FillBoundary` calls to reduce code duplication.

The fix for #3900 may be the same as is done here.

dpgrote:
There are now too many conflicts and it is not clear if the issue has been fixed by other PRs.

PR #3846, opened 2023-04-17
Use amrex::Math::powi<N>(x) in WarpX

lucafedeli88:
This PR replaces some exponentiations  where  the exponent is an integer known at compile time with the newly added `amrex::Math::powi<N>(x)` AMReX function


lucafedeli88:
Sorry, WarpX is not using the version of AMReX that contains `powi` yet. We need to wait 1 week to merge this PR

ax3l (approved):
Thank you! :+1:

PR #3849, opened 2023-04-18
Add python callback call when a break signal is recieved

aveksler1:
There are a couple actions that we want to perform only once, on a break signal, and we want to ensure they are the last things to be executed. An `onbreaksignal` callback would help us do this clearly and easily.


Source/Evolve/WarpXEvolve.cpp
```diff
@@ -370,6 +370,7 @@ WarpX::Evolve (int numsteps)
 
         exit_loop_due_to_interrupt_signal = SignalHandling::TestAndResetActionRequestFlag(SignalHandling::SIGNAL_REQUESTS_BREAK);
         if (cur_time >= stop_time - 1.e-3*dt[0] || exit_loop_due_to_interrupt_signal) {
+            ExecutePythonCallback("onbreaksignal");
```
dpgrote: Having the `onbreaksignal` call back is a good idea. Note that this will call `onbreaksignal` if the stop time is reached. Is this intended? Should this be
```
if (exit_loop_due_to_interrupt_signal) ExecutePythonCallback("onbreaksignal");
```

aveksler1: Yes, it should be. Thanks for catching that!

roelof-groenewald: In my opinion it would be better to move this callback execution down a couple lines (to 385) so that the callbacks execute after the diagnostic output. That would also make it more consistent with the proposed documentation saying that the `onbreaksignal` callback executes lasts before the code exits.

Python/pywarpx/callbacks.py
```diff
@@ -554,6 +560,20 @@ def isinstalledoncheckpointsignal(f):
     "Checks if the function is called on checkpoint signal"
     return _oncheckpointsignal.isinstalledfuncinlist(f)
 
+# ----------------------------------------------------------------------------
+def onbreaksignal(f):
+    installonbreaksignal(f)
+    return f
+def installonbreaksignal(f):
+    "Adds a function to the list of functions called on checkpoint signal"
```
roelof-groenewald:
```suggestion
    "Adds a function to the list of functions called on a break signal"
```

Python/pywarpx/callbacks.py
```diff
@@ -554,6 +560,20 @@ def isinstalledoncheckpointsignal(f):
...
+    _onbreaksignal.installfuncinlist(f)
+def uninstallonbreaksignal(f):
+    "Removes the function from the list of functions called on checkpoint signal"
```
roelof-groenewald:
```suggestion
    "Removes the function from the list of functions called on a break signal"
```

Python/pywarpx/callbacks.py
```diff
@@ -554,6 +560,20 @@ def isinstalledoncheckpointsignal(f):
...
+    _onbreaksignal.uninstallfuncinlist(f)
+def isinstalledonbreaksignal(f):
+    "Checks if the function is called on checkpoint signal"
```
roelof-groenewald:
```suggestion
    "Checks if the function is called on a break signal"
```

roelof-groenewald (changes requested):
Thanks @aveksler1! Just a few more changes to the doc-strings for the new callback.

Python/pywarpx/callbacks.py
aveksler1: Done, thank you!

roelof-groenewald (approved):
Looks good! Thanks!

Python/pywarpx/callbacks.py
```diff
@@ -289,6 +293,8 @@ def printcallbacktimers(tmin=1.,lminmax=False,ff=None):
               _beforestep,_afterstep,
               _afterdiagnostics,
               _afterrestart,
+              _oncheckpointsignal,
```
PhilMiller, line 296: Was this one just missing previously?

PhilMiller (approved):

Python/pywarpx/callbacks.py
aveksler1: Yes.

PR #2315, opened 2021-09-20
BinaryCollision: implement particle creation functor

NeilZaim:
This PR is based on #2301, which is based on #2290, ~~which is based on #2217~~, so it should probably be reviewed once these ~~3~~ 2 PRs have been merged.

It implements the binary collision particle creation functor that effectively creates new particles from binary collisions and sets their initial quantities. It also removes the weight of the product particles from the parent colliding particles, and if needed deletes the parent particles whose weight reaches 0.
Specific functions (depending on the binary collision type) will need to be implemented in order to initialize the momentum of the product particles. Currently, for proton-boron fusion, we just zero-initialize these momenta.

PR #3850, opened 2023-04-19
Update Particle Container to Pure SoA

ax3l:
Transition particle containers to pure SoA layouts.

- [x] AMReX: update SoA PC `Increment`, `IncrementWithTotal`, `Checkpoint`, `WritePlotFile`, `Restart`; SoA Particle `atomicSetID`
- [x] Update in situ vis
  - [x] Ascent https://github.com/AMReX-Codes/amrex/pull/3350
  - [x] SENSEI https://github.com/AMReX-Codes/amrex/pull/3349
- [x] double check openPMD output.
- [x] double check plotfile output.
- [x] fix newly unearthed clang-tidy issues
- [x] fix QED tests
- [x] fix EB/scraping tests
- [x] fix `Python_restart_runtime_components` (in AMReX restarts)
- [x] update & pass Python scripts that read structs (positions, ids)
- [x] run important tests manually on GPU before merge

### Fun Mini-Benchmarks on CPU, DP & SP

Same before/after PR on my laptop
```bash
./warpx.3d ../../Examples/Tests/performance_tests/automated_test_1_uniform_rest_32ppc amr.max_grid_size=64 amr.n_cell=64 64 64 max_step=5 &
taskset -cp 6 $!
```
We expect to need to do more work for vectorizing in the coming months.

### Fun Mini-Benchmarks on GPU, DP

```bash
./warpx.3d ../../Examples/Tests/performance_tests/automated_test_1_uniform_rest_32ppc amr.max_grid_size=256 amr.n_cell=256 256 256 max_step=5
```
~1-2%ish faster total RT on Perlmutter A100.

Current deposition: same
GatherAndPush: 1-2%
Redistribute_partition: 12%
AddPlasma: same
PushP: same
SortParticlesForDeposition: 230%


Source/Particles/NamedComponentParticleContainer.H
```diff
@@ -16,12 +16,19 @@
...
 
-/** Particle Attributes stored in amrex::ParticleContainer's struct of array
+/** Real Particle Attributes stored in amrex::ParticleContainer's struct of array
  */
 struct PIdx
 {
     enum {
-        w = 0,      ///< weight
+#if !defined (WARPX_DIM_1D)
+        x,
```
ax3l, line 27: We could even call this `r` for RZ sims :)

Source/Particles/NamedComponentParticleContainer.H
```diff
@@ -16,12 +16,19 @@
...
+#endif
+#if defined (WARPX_DIM_3D)
+        y,
+#endif
+        z,
+        w,      ///< weight
         ux, uy, uz,
 #ifdef WARPX_DIM_RZ
         theta,      ///< RZ needs all three position components
```
ax3l, line 36: We could move this forward in the enum to `r,z,theta`.

Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H
```diff
@@ -204,14 +204,21 @@ public:
...
                 // setting its id to -1
-                constexpr amrex::Long minus_one_long = -1;
                 if (w1[p_pair_indices_1[i]] <= amrex::ParticleReal(0.))
                 {
-                    particle_ptr_1[p_pair_indices_1[i]].atomicSetID(minus_one_long);
+                    //particle_ptr_1[p_pair_indices_1[i]].atomicSetID(minus_one_long); // TODO
+                    // or?
+                    //soa_1.m_idata[PIdxInt::id] = -1;
+                    // or?
+                    particle_ptr_1->id() = -1;
```
ax3l: Does this need to be atomic @NeilZaim (re: #2315)?
- Does this need to be atomic to prevent data races between different particles?

Does it matter if the ID is negative or exactly -1?
- @atmyers and I seem to remember any negative number will do (to mark invalid)

NeilZaim: Hi @ax3l 

1.  Yes this is to prevent a race condition. This is within a loop on pairs of colliding particles. A single particle can be in multiple pairs so potentially multiple GPU threads could try to set the ID of a single particle to -1 at the same time.
2. Yes, I also think that any negative ID will do here (@atmyers can probably confirm). The id is set to -1 solely for the particle to be removed during redistribute.

atmyers: It does not need to be `-1`, any negative value will cause the redistribute in AMReX to remove that particle. 

Question - what happens if we just let multiple threads set the id to negative 1 simultaneously? It might show up in some profiling software that helps detect race conditions, but is it actually dangerous?

NeilZaim: I thought these race conditions were always undefined behaviour, even if multiple threads are writing the same value to the variable. If it's not UB, then it's ok to remove the atomic.

PhilMiller: Multiple accesses to a location, at least one of which is a write, are a data race, resulting in UB. You can specify relaxed atomic stores, but they still must be atomic.

Source/Particles/ParticleCreation/SmartCreate.H
```diff
@@ -47,32 +47,32 @@ struct SmartCreate
         const int id = 0) const noexcept
     {
 #if defined(WARPX_DIM_3D)
-        prt.m_aos[i_prt].pos(0) = x;
-        prt.m_aos[i_prt].pos(1) = y;
-        prt.m_aos[i_prt].pos(2) = z;
+        prt.m_rdata[0][i_prt] = x;
```
RevathiJambunathan:
```suggestion
        prt.m_rdata[PIdx::x][i_prt] = x;
```

This was used in BTDParticleFunctor, so it can be used here to

Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H
ax3l: Excellent, that clarifies it. We will do an atomic change to the `id` and keep `cpu` untouched.

Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H
```diff
@@ -204,14 +204,12 @@ public:
 
                 // If the colliding particle weight decreases to zero, remove particle by
                 // setting its id to -1
-                constexpr amrex::Long minus_one_long = -1;
                 if (w1[p_pair_indices_1[i]] <= amrex::ParticleReal(0.))
                 {
-                    particle_ptr_1[p_pair_indices_1[i]].atomicSetID(minus_one_long);
+                    amrex::Gpu::Atomic::Exch(&particle_ptr_1->id(), -1);
```
PhilMiller: If there's an atomic store available rather than exchange, that would be preferable, since it would avoid loading the ID as part of the atomic transaction, just to throw it away.

ax3l: Yep, we thought so too but @WeiqunZhang and I are not aware of one right away.

(The old impl. also used atomic exchange to implement `atomicSetID`.)

WeiqunZhang: For SYCL, we could create a sycl::atomic_ref and then call load. For CUDA, it's not clear what's best way to do it. We could do something like https://github.com/NVIDIA/cub/blob/9ebcf693eedd90919cded7161a1560f8ad8b352d/cub/detail/strong_store.cuh#L47, but it would be hard to maintain. Maybe we could use https://nvidia.github.io/libcudacxx/extended_api.html in the future.

Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H
```diff
@@ -206,14 +203,14 @@ public:
...
+                constexpr uint64_t minus_1 = -1;
                 if (w1[p_pair_indices_1[i]] <= amrex::ParticleReal(0.))
                 {
-                    particle_ptr_1[p_pair_indices_1[i]].atomicSetID(minus_one_long);
+                    amrex::Gpu::Atomic::Exch(&idcpu1[i], minus_1);
                 }
                 if (w2[p_pair_indices_2[i]] <= amrex::ParticleReal(0.))
                 {
-                    particle_ptr_2[p_pair_indices_2[i]].atomicSetID(minus_one_long);
+                    amrex::Gpu::Atomic::Exch(&idcpu2[i], minus_1);
```
ax3l: Ok, CUDA compiles do not like this.

ax3l: Changed to a standalone implementation of `atomicSetID` for now.

ax3l:
Windows check:
```
[noarch_python] Noarch package contains binary script: vpic.exe
##[error]Cmd.exe exited with code '1'.
```

Source/Diagnostics/FlushFormats/FlushFormatCheckpoint.cpp
```diff
@@ -178,8 +178,8 @@ FlushFormatCheckpoint::CheckpointParticles (
         Vector<std::string> real_names;
         Vector<std::string> int_names;
 
+        // note: positions skipped here, since we reconstruct a plotfile SoA from them
```
ax3l, line 181: @atmyers Here, I tried both adding positions (see openPMD output) and not adding (current status). In both cases, it gets a bit confused and either misses attribute names or names them twice. Thus, I think there is just an offset issue how we iterate this name list somewhere.

In the version here, I see that the plotfile writes the positions but seems to forget later attribute names. I have not checked yet if the written "positions" are correctly labeled (are actually the values from position attributes).

Source/Diagnostics/FlushFormats/FlushFormatPlotfile.cpp
```diff
@@ -355,8 +355,8 @@ FlushFormatPlotfile::WriteParticles(const std::string& dir,
         Vector<int> int_flags;
         Vector<int> real_flags;
 
+        // note: positions skipped here, since we reconstruct a plotfile SoA from them
```
ax3l, line 358: @atmyers See above comment for checkpoints, which applies here as well.

Source/Diagnostics/FlushFormats/FlushFormatPlotfile.cpp
```diff
@@ -366,13 +366,14 @@ FlushFormatPlotfile::WriteParticles(const std::string& dir,
 #endif
 
         // get the names of the real comps
-        real_names.resize(tmp.NumRealComps());
+        real_names.resize(tmp.NumRealComps() - AMREX_SPACEDIM);  // skip mandatory positions
         auto runtime_rnames = tmp.getParticleRuntimeComps();
         for (auto const& x : runtime_rnames) { real_names[x.second+PIdx::nattribs] = x.first; }
```
ax3l: This is out of bounds now for `ImplicitPicard_1d`.
Should probably read:
```suggestion
for (auto const& x : runtime_rnames) { real_names[x.second+PIdx::nattribs-AMREX_SPACEDIM] = x.first; }
```

Source/Particles/NamedComponentParticleContainer.H
ax3l: And even better, we should make `theta` a runtime parameter (see reason in https://github.com/AMReX-Codes/pyamrex/pull/243)

Source/Particles/ElementaryProcess/QEDPairGeneration.H
```diff
@@ -167,7 +167,7 @@ public:
             p_ux, p_uy, p_uz,
             engine);
 
-        src.m_aos[i_src].id() = -1; //destroy photon after pair generation
+        src.m_idata[0][i_src] = -1;  // destroy photon after pair generation
```
ax3l: That looks wrong.
```suggestion
        amrex::ParticleIDWrapper{src.m_idcpu[i_src]} = -1;  // destroy photon after pair generation
```

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -195,55 +195,57 @@ WarpXParticleContainer::AddNParticles (int /*lev*/, long n,
...
-
-    if (np > 0)
-    {
         pinned_tile.push_back_real(PIdx::w , attr_real[0].data() + ibegin, attr_real[0].data() + iend);
-        pinned_tile.push_back_real(PIdx::ux, ux.data() + ibegin, ux.data() + iend);
-        pinned_tile.push_back_real(PIdx::uy, uy.data() + ibegin, uy.data() + iend);
-        pinned_tile.push_back_real(PIdx::uz, uz.data() + ibegin, uz.data() + iend);
+        pinned_tile.push_back_real(PIdx::ux,     ux.data() + ibegin,     ux.data() + iend);
+        pinned_tile.push_back_real(PIdx::uy,     uy.data() + ibegin,     uy.data() + iend);
+        pinned_tile.push_back_real(PIdx::uz,     uz.data() + ibegin,     uz.data() + iend);
```
ax3l: Unnecessary white spaces, not sure how they got it :)

ax3l:
Merging this after discussion with @atmyers on Friday.

ax3l:
Congrats to all who helped @Thierry992, @atmyers, @AlexanderSinn et al. :tada: :clap:

ax3l:
@atmyers @roelof-groenewald feel free to give this a final review and then we will merge :)

atmyers (approved):

roelof-groenewald:
Congrats! This is awesome! Besides the (upcoming) performance improvements, it will make writing particle routines much easier going forward.

ax3l:
Yes, and as nice Python GPU goodies we can now:
- read & write directly (zero-copy) from/to particle positions and ids via `cupy`
- can implement `cudf` for particle tiles with [zero-copy read/write support](https://github.com/AMReX-Codes/pyamrex/issues/244)
- ...

ax3l:
Merge coming for the 24.03 release series in #4653

PR #3851, opened 2023-04-19
Correct sign of the `ChargeOnEB` diagnostic

RemiLehe:
The `ChargeOnEB` diagnostic had the wrong sign: it computed $$-\epsilon_0 \oint E\cdot dS$$ instead of $$\epsilon_0 \oint E\cdot dS$$.

Fundamentally, this is because I had misunderstood the `eb_area_fraction` data, and thought that it represented the area covered by the EB, while it actually represents the opposite (uncovered area):
https://amrex-codes.github.io/amrex/docs_html/EB.html#embedded-boundary-data

I also updated the sign in the automated test. The charge on a sphere that is set to a positive potential is now also positive, which makes more sense. 😅 

I also properly indented the documentation for `ChargeOnEB`.

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -2793,25 +2793,25 @@ Reduced Diagnostics
         so the time of the diagnostic may be long
         depending on the simulation size.
 
-* ``ChargeOnEB``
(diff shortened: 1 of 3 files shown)


dpgrote (approved):
Looks good!

PR #3857, opened 2023-04-20
Docs: Boosted Frame Inputs

ax3l:
Outdated docs: No, all inputs are lab-frame, besides `stop_time`.

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -28,6 +28,8 @@ Overall simulation parameters
     ``max_step`` and ``stop_time`` are provided, both criteria are used and the simulation stops
     when the first criterion is hit.
 
+    Note: in boosted-frame simulations, ``stop_time`` refers to the time in the boosted frame.
+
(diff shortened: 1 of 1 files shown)


Docs/source/usage/parameters.rst
```diff
@@ -28,6 +28,8 @@ Overall simulation parameters
     ``max_step`` and ``stop_time`` are provided, both criteria are used and the simulation stops
     when the first criterion is hit.
 
+    Note: in boosted-frame simulations, ``stop_time`` refers to the time in the boosted frame.
```
dpgrote, line 31: Would it make sense to convert `stop_time` so that the user inputs the quantity in the lab frame? The boosted value would have to be calculated for some position in the boosted frame domain. Though, maybe this is tricky enough for it not to be worth it for a rarely used feature.

ax3l: Good point, added it for discussion to the Tue agenda.

If we go for it, then we can update this in a follow-up PR. For now we just document behavior.

ax3l:
Ping @dpgrote @RemiLehe

dpgrote (approved):
Looks good!

PR #3863, opened 2023-04-21
Move KernelTimer.H to ablastr/parallelization

lucafedeli88:
This PR moves `KernelTimer.H` from `Parallelization` to `ablastr\parallelization`. The file is adapted so that there is no dependency on `WarpX`. This requires to move the check 
```
#   ifndef WARPX_USE_GPUCLOCK
        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(WarpX::load_balance_costs_update_algo!=LoadBalanceCostsUpdateAlgo::GpuClock,
            "`algo.load_balance_costs_update = gpuclock` requires to compile with `-DWarpX_GPUCLOCK=ON`.");
#   endif // !WARPX_USE_GPUCLOCK
```
 to `WarpX.H`, where the parameter `load_balance_costs_update_algo` is read 
 
 
 On a second though, should we maybe place this function into `ablastr/utils/timer` instead ?


ax3l (approved):
LGTM, thank you :)

PR #3404, opened 2022-09-16
Partial refactoring of the utils directory

lucafedeli88:
This PR is an attempt to a partial refactor of th Utils directory, which has become a bit "messy" ;-) .

In particular:

- I have created a subfolder "Algorithms", where I put the algorithms is_in (which I have improved), linear_interpolation (bilinear and trilinear as well), and
upper_bound. Those algorithms are inside a namespace called utils::algorithms

- I have created a subfolder "Physics", where I put stuff related to ionization tables. Those functions are inside the namespace utils::physics

- I have created the subfolder "Strings", where I put the split function and the automatic_text_wrap function

- I have created the subfolder "Parser", where I put the interval parser and the utility functions related to the amrex::Parser. These functions are inside the utils::parser namespace.

All the other changes in the code are a consequence of these changes.

PR #3864, opened 2023-04-21
Move StringUtils.H to ablastr

lucafedeli88:
This PR moves `StringUtils.H` to ablastr, as suggested some time ago by @ax3l  (https://github.com/ECP-WarpX/WarpX/pull/3404 ).
It also removes some code duplication.


ax3l (approved):
Thank you, this is great!

PR #3865, opened 2023-04-21
Docs: use bibtex citations in collisions theory section

roelof-groenewald:
Switch references in the Collisions theory section of the docs to use bibtex instead of hyperlinks.


ax3l (approved):
Thanks!!

EZoni (commented):
Thank you, @roelof-groenewald! I tried compiling the docs in this branch and I see that the references show up in the text, say as "bla bla [36]", but I think the actual bibliography item shows up in the Introduction page rather than the Collisions page, maybe because of the fact that the BibTeX entries are added to one single file.

I think you might need to add the line `.. bibliography::` at the end of the file `collisions.rst`. After this, we might need to check whether the new bibliography entries show up only in the Collisions page or if they show up both here and in the Introduction page, in which case we might want to find a way to hide them from the page where they are not used (either using separate `bib` files or in some other way that I'm not aware of at the moment).

roelof-groenewald:
> I think you might need to add the line `.. bibliography::` at the end of the file `collisions.rst`. After this, we might need to check whether the new bibliography entries show up only in the Collisions page or if they show up both here and in the Introduction page...

@EZoni thanks for the review! I noticed also with the citations for the hybrid-PIC algo that they all showed up on the `Introduction` page. The last commit solves that problem by only citing references with the `b-` key prefix. This I think would be a fine solution - every page will just have it's own prefix. The question is whether we want to add the `::labelprefix` part. This causes the citations to have an extra prefix i.e. they show up as " bla bla [B1]", and same in the bibliography:

image

Without the `::labelprefix` sphinx gives warnings that multiple references are cited as "[1]" and "[2]", etc. But the links work fine, so if we could just suppress the warning, neglecting the `::labelprefix` would be my preference.

EZoni:
Thanks @roelof-groenewald! It would be also my preference to avoid `::labelprefix`. I think we can remove it for now, and we can search for a way to suppress the multiple references warning later on, if that's OK with you.

Docs/source/theory/collisions.rst
```diff
@@ -89,3 +92,7 @@ The impact of incorporating relativistic effects in the MCC routine can be seen
 .. figure:: https://user-images.githubusercontent.com/40245517/170900079-74e505a5-2790-44f5-ac84-5847eda954e6.png
    :alt: Classical v relativistic MCC
    :width: 96%
+
+.. bibliography::
+    :labelprefix: B
```
roelof-groenewald:
```suggestion
```

EZoni (approved):

ax3l:
We currently only build a HTML page.
If we wanted to later on build a LaTeX document again, then `::labelprefix` references, e.g., per chapter could be more scalable. Realistically, it might be that far down the line / something we will not bother doing.

PR #3866, opened 2023-04-23
Documentation For Using NCU

kaplannp:
Adds documentation for Nsight Compute tool on Perlmutter in developers/workflows/profiling\ the\ code


Docs/source/developers/profiling.rst
```diff
@@ -208,3 +212,80 @@ In these examples, the individual lines for recording a trace profile are:
...
+`Vendor homepage <https://developer.nvidia.com/nsight-compute>`__ and `product manual <https://docs.nvidia.com/nsight-compute/>`__.
+
+Nsight-Compute captures fine grained information at the kernel level
+concerning resource utilization. It collects a lot of data, and runs slowly (can
+be a few minutes per step), but provides detailed information about occupancy,
+and memory bandwidth for a kernel.
+
+
+Example
+""""""""""""""""""
```
ax3l: rst is a bit picky about those alignments:
```suggestion
Example
"""""""
```

Docs/source/developers/profiling.rst
```diff
@@ -208,3 +212,80 @@ In these examples, the individual lines for recording a trace profile are:
...
 This can be done on another system than the one that recorded the traces.
 For example, if you record on a cluster and open the analysis GUI on your laptop, it is recommended to make sure that versions of Nsight-Systems match on the remote and local system.
+
+Nvidia Nsight-Compute
+---------------------
+
+`Vendor homepage <https://developer.nvidia.com/nsight-compute>`__ and `product manual <https://docs.nvidia.com/nsight-compute/>`__.
+
+Nsight-Compute captures fine grained information at the kernel level
+concerning resource utilization. It collects a lot of data, and runs slowly (can
```
ax3l: Just FYI: Slowlyness depends on the stats and number of stats to collect.
What makes it slow is that for complicated stats, it has to run multiple times the same kernel. To achieve that, it does swap the whole GPU memory in and out. Wild, isn't it.

Docs/source/developers/profiling.rst
```diff
@@ -208,3 +212,80 @@ In these examples, the individual lines for recording a trace profile are:
...
+    single process can take a few minutes for every step, but since we're
+    profiling at the kernel level, a single step of a single process is often
+    enough to extract relevant information.
+
+Details
+"""""""
+In the example above, the individual lines for recording a trace profile are:
+
+* ``dcgmi profile --pause`` other profiling tools can't be collecting data,
+  `see this Q&A <https://forums.developer.nvidia.com/t/profiling-failed-because-a-driver-resource-was-unavailable/205435>`_.
```
ax3l: Anonymous link might be better here
```suggestion
  `see this Q&A <https://forums.developer.nvidia.com/t/profiling-failed-because-a-driver-resource-was-unavailable/205435>`__.
```

ax3l (commented):
Thanks a lot for this!

I thought we added this last summer but it slipped.

Docs/source/developers/profiling.rst
kaplannp: Very nifty, and I guess the cache state doesn't matter cause they'll have to flush it anyways.
I added this to the note in the docs. Maybe it'll help people who are trying to run more steps.

ax3l (approved):
:+1:

Issue #3870, opened 2023-04-24
Managed Memory Leftovers: MR Init, QED, Ionization


RemiLehe:
The field ionization module seems to be failing on GPU when managed memory is turned off. (which is now the default in WarpX)

For instance, running our ionization test on GPU now fails on Perlmutter with `SegFault` in `doFieldIonization`:
https://github.com/ECP-WarpX/WarpX/blob/development/Examples/Tests/ionization/inputs_2d_rt
but runs correctly if one adds `amrex.the_arena_is_managed=1` in the inputs.

ax3l:
After looking at the logic, I think this affects QED, too.
Tested with `../../Examples/Tests/qed/quantum_synchrotron/inputs_2d`
CC @lucafedeli88 

Both use the same logic for creating new particles cc @atmyers 
https://github.com/ECP-WarpX/WarpX/blob/03bfb024201da4e01a3bf3a2bba3a80dba0cb554/Source/Particles/MultiParticleContainer.cpp#L889-L924

I see some page faults that are mitigated by managed memory with nsys, first DtoH and then also HtoD. Possible that the particle tiles of the source are device tiles, accessed without a copy from host, and that the dst tile is not actively copied back to device.

ax3l:
#3825 is not yet released, removing respective tag.

Ionization fixed #3874 
QED fix incoming via #3873

ax3l:
`Examples/Physics_applications/laser_acceleration/inputs_2d` fails due to managed memory usage in `WarpX::BuildBufferMasksInBox` for all MR setups.
https://github.com/ECP-WarpX/WarpX/blob/03bfb024201da4e01a3bf3a2bba3a80dba0cb554/Source/WarpX.cpp#L2863-L2918

It's consequently also very slow. To fix, we need an update using `ParallelFor` on the right box extents.

cc @WeiqunZhang @EZoni

ax3l:
With #3880 merged, we have all known issues fixed again :)

Thanks everyone that reported, tested & helped! 🤝

PR #3873, opened 2023-04-26
Fix missing GPU synchronization when QED lookup tables are loaded or generated

lucafedeli88:
~This PR adds a thin wrapper around `amrex::Gpu::DeviceVector`, in order to allow PICSAR to call  `amrex::Gpu::synchronize()` when required during the loading or the generation of the QED lookup tables.~
This is part of what is required to make the QED module work also when managed memory is not used.
Essentially picsar-qed needs something that behaves similarly to the Device Vector with managed memory. In the past I tested picsar-qed with similar thin wrappers around Kokkos::vector<Real> (https://github.com/ECP-WarpX/picsar/blob/development/multi_physics/QED/kokkos_example/kokkos_example_commons.hpp) or thrust::device_vector<RealType> (https://github.com/ECP-WarpX/picsar/blob/development/multi_physics/QED/test_gpu/test_breit_wheeler.cu).
picsar-qed calls the `pxr_sync()` function only if the vector type implements it and at the right moment (when a data transfer to the GPU is required). `data()` is used to generate a non-owning vector pointing to the data on the device and suitable to be used in GPU kernels.

Related to #3870

Follow-up to #3825


Source/Particles/ElementaryProcess/QEDInternals/QedWrapperCommons.H
```diff
@@ -36,4 +40,38 @@
...
+    {}
+
+    /**
+    * This function may be called in some places inside picsar_table classes.
+    * It forces a copy of the CPU data to the GPU
+    *
+    */
+    void pxr_sync()
+    {
+        amrex::Gpu::synchronize();
```
ax3l: Is stream sync sufficient?
```suggestion
        amrex::Gpu::Device::streamSynchronize();
```

Source/Particles/ElementaryProcess/QEDInternals/QedWrapperCommons.H
```diff
@@ -36,4 +40,38 @@
...
+    *
+    * @tparam Args the constructor arguments
+    */
+    template<typename... Args>
+    AmrexDeviceVectorWrapper(Args&&... args) : AGDV(std::forward<Args>(args)...)
+    {}
+
+    /**
+    * This function may be called in some places inside picsar_table classes.
+    * It forces a copy of the CPU data to the GPU
```
ax3l, line 76: The comment is a bit misleading: this just forces that async copy operations, which have to be done separately from this call according to the impl., are done.

Source/Particles/ElementaryProcess/QEDInternals/QedWrapperCommons.H
```diff
@@ -6,16 +6,19 @@
  */
 #ifndef WARPX_amrex_qed_wrapper_commons_h_
 #define WARPX_amrex_qed_wrapper_commons_h_
-
 /*
  * This header contains some common #define directives
  * used by the QED library wrappers and related
- * components.
+ * components. It also defines a thin wrapper around
```
ax3l: Not 100% sure I understand why you define a wrapper for DeviceVector - you could also define a free standing function for `amrex::Gpu::Device::streamSynchronize();` and call that?

lucafedeli88:
@ax3l , I am sorry, this PR was just a draft and it wasn't complete when you reviewed it. I should add maybe more comments, but the idea is that for picsar-qed I need something that behaves similarly to the Device Vector with managed memory. In the past I tested picsar-qed with similar thin wrappers around `Kokkos::vector<Real>` (https://github.com/ECP-WarpX/picsar/blob/development/multi_physics/QED/kokkos_example/kokkos_example_commons.hpp)  or `thrust::device_vector<RealType>` (https://github.com/ECP-WarpX/picsar/blob/development/multi_physics/QED/test_gpu/test_breit_wheeler.cu)

ax3l:
No worries :)

Don't forget to run `./Tools/Release/updatePICSAR.py` once you are done updating PICSAR, to bump the version in WarpX for this PR :)

Source/Particles/ElementaryProcess/QEDInternals/QedWrapperCommons.H
lucafedeli88: I agree. I am sorry, but the PR was not complete when you reviewed it. Now the function actually performs the copy

Source/Particles/ElementaryProcess/QEDInternals/QedWrapperCommons.H
lucafedeli88: I will change this comment. You are right.

lucafedeli88:
> ./Tools/Release/updatePICSAR.py

Actually there is no need to perform changes in PICSAR-QED (PICSAR-QED has no knowledge of AMReX). It should be enough to define a new Vector Type (based on `DeviceVector`) to be used as a building block of the QED lookup tables.

Source/Particles/ElementaryProcess/QEDInternals/QuantumSyncEngineWrapper.cpp
```diff
@@ -207,1034 +205,1031 @@ void QuantumSynchrotronEngine::init_builtin_phot_em_table()
...
--3.86979e+00_prt, -3.72360e+00_prt, -3.57740e+00_prt, -3.43121e+00_prt,
--3.28502e+00_prt, -3.13883e+00_prt, -2.99264e+00_prt, -2.84646e+00_prt,
--2.70029e+00_prt, -2.55412e+00_prt, -2.40797e+00_prt, -2.26184e+00_prt,
--2.11575e+00_prt, -1.96971e+00_prt, -1.82374e+00_prt, -1.67790e+00_prt,
--1.53224e+00_prt, -1.38687e+00_prt, -1.24192e+00_prt, -1.09763e+00_prt,
--9.54333e-01_prt, -8.12547e-01_prt, -6.73013e-01_prt, -5.36770e-01_prt,
--4.05101e-01_prt, -2.79009e-01_prt, -1.56354e-01_prt, 0.00000e+00_prt};
-
-    amrex::Gpu::synchronize();
+    const auto vals = std::vector<amrex::ParticleReal>{
```
ax3l, line 208: That diff, ahhh xD why 😅

ax3l (approved):

PR #3878, opened 2023-04-27
CI: macOS Cache Less

ax3l:
Cache only selected brew paths in macOS, not all of `/usr/local`, which contains more runner specific software.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -27,7 +27,9 @@ jobs:
       # - for a refresh the key has to change, e.g., hash of a tracked file in the key
       with:
         path: |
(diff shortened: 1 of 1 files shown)


AlexanderSinn:
I’m not sure that this will work, I solved it a bit differently https://github.com/Hi-PACE/hipace/pull/926/files#diff-7b74d73de878e0e53f6768bbf3a3d4b50c3c6f1f16f9b64518690028b8206b1c .
Definitely test that the CI works and is fast when rerunning without a change.

ax3l:
Oops, should not merge on my phone. Let's apply your patch in a separate PR, looks good 👍

PR #3880, opened 2023-04-27
Build buffer masks on GPU

WeiqunZhang:


Source/WarpX.cpp
```diff
@@ -2893,57 +2893,23 @@ void
...
+    auto const& gmsk = guard_mask.const_array();
+    amrex::Dim3 ng3 = amrex::IntVect(ng).dim3();
+    amrex::ParallelFor(tbx, [=] AMREX_GPU_DEVICE (int i, int j, int k)
+    {
+        for         (int kk = k-ng3.z; kk <= k+ng3.z; ++kk) {
+            for     (int jj = j-ng3.y; jj <= j+ng3.y; ++jj) {
+                for (int ii = i-ng3.x; ii <= i+ng3.x; ++ii) {
+                    if ( gmsk(ii,jj,kk) == 0 ) {
+                        msk(i,j,k) = 0;
+                        return;
```
EZoni, line 2906: Is it correct to have the `return` statement in the innermost `for` loop over `ii` here? I'm wondering what happens if `gmsk(ii,jj,kk)==0` is true for more than one combination of indices `(ii,jj,kk)`. Once the first combination is encountered, the corresponding `msk(i,j,k)` is set to zero and then the function "returns". Does it then reach the second combination of indices `(ii,jj,kk)` where `gmsk(ii,jj,kk)==0` is true? Maybe I don't understand exactly where `return` brings us in this case.

ax3l: Looks like a bug to me, too.
```suggestion
```

Surprised that no test fails.

WeiqunZhang: Once it reaches `return`, the iteration will stop and the rest of gmsk values will not have any impact. It's equivalent to
```
if (gmsk(i-1,j-1,k-1) == 0) {
    msk(i,j,k) = 0;
} else if (gmsk(i,j-1,k-1) == 0) {
    msk(i,j,k) = 0;
} else if (...) {
    msk(i,j,k) = 0;
}
....
else if (gmsk(i+1,j+1,k+1) == 0) {
    msk(i,j,k) = 0;
else {
    msk(i,j,k) = 1;
}
```
I think this is equivalent to the original logic.

WeiqunZhang: Note that the `return` returns from the lambda function, not `ParallelFor`.

ax3l: Thanks for clarifying, I overlooked the lambda wrapping:)

ax3l (approved):

EZoni (approved):

PR #3881, opened 2023-04-27
Clean up to handling of the stopping of the simulation

dpgrote:
This PR does some cleanup of the handling of how the stopping point of the simulation is handled.

The changes include:

- Adding a section in the documentation on simulation stopping
- Adding a print out about why the simulation stopped
- Changed `stop_time` to use its exact value for consistency (removing the small slop factor)
- Minor code and formatting changes to improve clarity

A suggestion was made to add a utility routine to check if the simulation should stop. I do not think that this would help since each of the checks is somewhat difference (for instance checking the step number against `numsteps_max` versus `max_step`).


Docs/source/usage/parameters.rst
```diff
@@ -21,12 +21,12 @@ Overall simulation parameters
     When provided, this information is added as metadata to (openPMD) output files.
 
 * ``max_step`` (`integer`)
-    The number of PIC cycles to perform.
+    The number of PIC cycles to perform. See :ref:`stopping the simulation <stopping-the-simulation>`.
 
 * ``stop_time`` (`float`; in seconds)
     The maximum physical time of the simulation. Can be provided instead of ``max_step``. If both
```
RTSandberg:
```suggestion
    The maximum physical (in the boosted frame if doing a boosted-frame calculation) time of the simulation. Can be provided instead of ``max_step``. If both
```

RTSandberg: This is redundant but I think is worth pointing out.

Docs/source/usage/parameters.rst
```diff
@@ -224,6 +228,34 @@ We follow the same naming, but remove the ``SIG`` prefix, e.g., the WarpX signal
...
+
+There are several ways to control when a simulation will stop.
+The basic ways are to set either the maximum number of time steps (by setting
+``max_step``) or to set the maximum simulation time (by setting
+``stop_time``, noting the value will be in the boosted frame).
+When using the boosted frame and/or the back-transformed diagnostics, other
+criteria can be used.
+The quantity ``warpx.zmax_plasma_to_compute_max_step`` can be specified, setting the
+stopping based on the location of the boosted frame.
+Also, ``warpx.compute_max_step_from_btd`` can set, in which case the the
```
RTSandberg:
```suggestion
Also, ``warpx.compute_max_step_from_btd`` can be set, in which case the
```

Docs/source/usage/parameters.rst
```diff
@@ -224,6 +228,34 @@ We follow the same naming, but remove the ``SIG`` prefix, e.g., the WarpX signal
...
+^^^^^^^^^^^^^^^^^^^^^^^
+
+There are several ways to control when a simulation will stop.
+The basic ways are to set either the maximum number of time steps (by setting
+``max_step``) or to set the maximum simulation time (by setting
+``stop_time``, noting the value will be in the boosted frame).
+When using the boosted frame and/or the back-transformed diagnostics, other
+criteria can be used.
+The quantity ``warpx.zmax_plasma_to_compute_max_step`` can be specified, setting the
+stopping based on the location of the boosted frame.
```
RTSandberg:
```suggestion
stopping criteria based on the location of the boosted frame.
```

Docs/source/usage/parameters.rst
```diff
@@ -224,6 +228,34 @@ We follow the same naming, but remove the ``SIG`` prefix, e.g., the WarpX signal
...
+
+In the Python interface, the ``step`` command is called directly, with each
+call specifying the number of time steps.
+In this case, the ``max_step`` input parameter is ignored and the simulation
+will continue past it to complete the number of steps specified in the call to ``step``.
+Note that the simulation will stop if the time exceeds ``stop_time``.
+Any further calls to ``step`` will do nothing.
+
+The simulation can also be stopped by sending the appropriate signal (see :ref:`signal handling <signal-handling>`.).
+When a breaking signal is received, WarpX will finish it's current time step, write out a check point file, and then exit.
```
RTSandberg:
```suggestion
When a breaking signal is received, WarpX will finish its current time step, write out a check point file, and then exit.
```

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -189,9 +189,10 @@ void BTDiagnostics::DerivedInitData ()
         std::stringstream warn_string;
             warn_string << "\nSimulation might not run long enough to fill all BTD snapshots.\n"
             << "Final step: " << warpx.maxStep() << "\n"
-            <<"Stop time: " << warpx.stopTime() << "\n"
-            <<"Last BTD snapshot fills around step: " << final_snapshot_fill_iteration << "\n"
-            <<" or time: " << final_snapshot_fill_time << "\n";
+            << "Stop time: " << warpx.stopTime() << "\n"
+            << "Last BTD snapshot fills around step: " << final_snapshot_fill_iteration << "\n"
+            << " or time: " << final_snapshot_fill_time << "\n"
+            << "Set `warpx.compute_max_step_from_btd = 1` to have simulation stopping calculated automatically\n";;
```
RTSandberg, line 193: This is a nice addition, thanks!

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -370,20 +370,35 @@ WarpX::Evolve (int numsteps)
         }
 
         exit_loop_due_to_interrupt_signal = SignalHandling::TestAndResetActionRequestFlag(SignalHandling::SIGNAL_REQUESTS_BREAK);
-        if (cur_time >= stop_time - 1.e-3*dt[0] || exit_loop_due_to_interrupt_signal) {
+        if (exit_loop_due_to_interrupt_signal) {
```
RTSandberg, line 302: This change makes sense to me.  The fudge factor is 7 years old or older, do you know why it was there?

Source/Initialization/WarpXInitData.cpp
```diff
@@ -638,6 +638,8 @@ WarpX::computeMaxStepBoostAccelerator() {
     max_step = computed_max_step;
     Print()<<"max_step computed in computeMaxStepBoostAccelerator: "
            <<max_step<<std::endl;
+    // Reset stop_time to override any user input
+    stop_time = std::numeric_limits<amrex::Real>::max();
```
RTSandberg, line 687: Would it be useful to print a warning that `stop_time` was overridden here?

RTSandberg (commented):
This PR is helpful, thanks @dpgrote !

Docs/source/usage/parameters.rst
dpgrote: I'll add this, but with a change in the wording.

Source/Evolve/WarpXEvolve.cpp
dpgrote: I think there was concern about what happens if `cur_time` lands right near `stop_time`, for instance if a user had set `stop_time = 100*dt`. I don't think this would be an issue, and if so, the user should set `max_step = 100`.
Note that the check of `cur_time` here is unnecessary since it is checked at the start of the loop anyway.

RTSandberg: I did see that.  It makes sense to me to take out this seemingly unnecessary magic number and logic

PR #3884, opened 2023-04-28
Moving Window: Fix the logic to shift F and G fields when doing divE and divB cleaning in pml

prkkumar:
This PR fixes the logic to shift `F` and `G` fields when doing divE and divB cleaning in `pml` in simulations with moving window. Prior to this fix, shifts to `F` and `G` were applied by checking if `do_dive_cleaning` and `do_divb_cleaning` is `true` everywhere including in the pml region. However, in some cases, particularly with the PSATD solver, we need to do this only in the pml region. This PR fixes the logic to enable that.

Here is a result from test done with the psatd solver and mesh refinement in 2D with and without the fixes in the PR: 
fixed_logic_mr

### DIFF
--- a/Source/Utils/WarpXMovingWindow.cpp
+++ b/Source/Utils/WarpXMovingWindow.cpp
@@ -233,12 +233,23 @@ WarpX::MoveWindow (const int step, bool move_j)
             }
         }
 
-        // Shift scalar field F with div(E) cleaning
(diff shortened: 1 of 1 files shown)


EZoni (approved):
Thank you for this bug fix, @prkkumar! Looks good to me.

PR #3890, opened 2023-05-04
Bug fix - PEC - Remove duplicate boundary cell updates for rho and J

roelof-groenewald:
Thanks to @aeriforme, we found that in some cases the logic to apply a PEC boundary condition to the charge density field (PR #3711) causes some cells to be updated multiple times.
The issue is that the PEC boundary is applied in 2 steps, each requiring a loop over the grid cells. The first one updates cells inside the simulation domain based on values in their mirror cells across the PEC boundary. The second updates guard cells based on their mirror cells inside the domain. It is very important that these steps happen in sequence. However, it seems that with the loop over tiles (boxes) it is possible for some external grid points to be updated before the loop over internal cells is complete. In this PR the issue is fixed by having separate loops over the tiles, not only separate loops over the grid cells in a given `box`.

Is there a more elegant / better performant way to approach this double for-loop requirement? [@ax3l @WeiqunZhang @atmyers ]

Charge density in uniform seed density test problem (3d) before fix:
current_state

Charge density in same test with this fix:
after_fix

I assume the same issue will also come up in the function that applies the PEC boundary condition to the current density. ~~If there isn't a different preferred approach I will also update that function.~~


WeiqunZhang:
https://github.com/WeiqunZhang/WarpX/blob/1f5b4298d8ce0bc225a8bb94d86b5e1b08def6c5/Source/BoundaryConditions/WarpX_PEC.cpp#L261
I think it can be done like this. The code may not even compile. It is just for demonstration.

roelof-groenewald:
Thanks for the idea and explanation of how to robustly handle the boundary cells, @WeiqunZhang. I have applied your suggestion to both the charge and current density PEC boundary application.

dpgrote:
Thanks for the fix! One question - the kernels for J and rho are almost identical. Can there be one kernel that is used in both cases?

ax3l:
I agree this is a good idea, but would say we perform code deduplication work separately after this fix.

Source/BoundaryConditions/WarpX_PEC.cpp
```diff
@@ -220,82 +230,105 @@ PEC::ApplyPECtoRhofield (amrex::MultiFab* rho, const int lev, PatchType patch_ty
...
+        is_pec[idim][1] = WarpX::field_boundary_hi[idim] == FieldBoundaryType::PEC;
+        if (!is_pec[idim][0]) grown_domain_box.growLo(idim, ng_fieldgather[idim]);
+        if (!is_pec[idim][1]) grown_domain_box.growHi(idim, ng_fieldgather[idim]);
+
+        psign[idim][0] = (WarpX::particle_boundary_lo[idim] == ParticleBoundaryType::Reflecting)
+                         ? 1._rt : -1._rt;
+        psign[idim][1] = (WarpX::particle_boundary_hi[idim] == ParticleBoundaryType::Reflecting)
+                         ? 1._rt : -1._rt;
+        mirrorfac[idim][0] = 2*domain_lo[idim] - (1 - rho_nodal[idim]);
+        mirrorfac[idim][1] = 2*domain_hi[idim] - (1 - rho_nodal[idim]);
```
WeiqunZhang: Should it be `2*domain_hi[idim] + (1 - rho_nodal[idim]);` (unless domain_box is converted to nodal like in ApplyPECtoJfield)?

roelof-groenewald: Yes, that is true. Thanks for catching it! I thought about the J case where I force the domain box to be nodal in order to get a consistent `mirrorfac` calculation for both the nodal and cell centered directions for the J-field and then just applied the same calculation here (where I just consider rho to be nodal since it usually is). 
```suggestion
        mirrorfac[idim][1] = 2*domain_hi[idim] + (1 - rho_nodal[idim]);
```

roelof-groenewald:
Thanks for the review @dpgrote. I will look to re-use a single kernel for both `rho` and `J` when I get back from Sherwood conference (either as part of this PR if it is still open or in a follow-up PR).

WeiqunZhang (approved):

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -541,6 +541,7 @@ void WarpX::SyncCurrentAndRho ()
         SyncRho(rho_fp, rho_cp);
     }
 
+#ifndef WARPX_DIM_RZ
```
dpgrote: What is the reason that this is turned off for RZ? The z-boundaries could be PEC.

roelof-groenewald: I believe the outer r boundary can also be set to PEC, right? Currently the PEC boundary application does not have any checks that it is only applied in the z-direction when RZ is used. I think it would be good at some point to make the handling of RZ boundaries more consistent with these changes (moving parts of the `ApplyInverseVolumeScalingToChargeDensity` logic to be handled by `ApplyRhofieldBoundary`).

dpgrote: Yes, the outer boundary can also be PEC. Note that there will be some trickiness with the outer boundaries because of the r-dependence. I agree about being consistent about handling the axis - it is doing the same thing, copying the charge/current from below the axis to above.

roelof-groenewald: So do you think it is okay to merge this as is and then follow-up with the RZ handling later?

dpgrote: I'm not sure. If someone is using RZ and PEC on the z boundaries, this will break the code since it turns this off for those boundaries.

roelof-groenewald: Maybe the best is then that, for now, I will remove the limitation to non-RZ but add a warning that there will be incorrect handling of the boundary if someone uses a PEC at the upper R boundary.

roelof-groenewald: Actually that solution might not work since the test `LaserInjectionFromLASYFile_RZ` uses a PEC boundary for the upper R boundary.

dpgrote (approved):
Looks good!

PR #3893, opened 2023-05-04
Replace 3 occurrencess of NULL with std::nullptr

lucafedeli88:
This PR replaces all the occurrences of  `NULL` in  WarpX with `std::nullptr`


AlexanderSinn:
It’s just `nullptr`, no `std::`.

ax3l (approved):

PR #3894, opened 2023-05-04
Fix: Laser-Ion Example Gather

ax3l:
Guarantee improved plasma stability for 2D with very low initial target temperature when using Esirkepov current deposition with energy-conserving field gather (default).

Thank you to @joshualudwig8 for raising this!


n01r (approved):
Looks good, thanks for the fix!

PR #3585, opened 2022-12-21
Adding additional check in addPlasma and addPlasmaFlux to respect the…

clarkse:
… random_theta flag, otherwise acts deterministically.

When generating benchmarks on Ubuntu 22.04 LTS in WSL, the random number generator does not generate the same numbers as in the Azure pipeline.  If using deterministic particle locations (e.g. a gridded layout) then in RZ it would be expected to have deterministic theta locations that do not rely on the random number generator package.  These cahnges in addPlasma and addPlasmaFlux would respect the <species>.random_theta flag and lay particles down deterministically.

PR #3895, opened 2023-05-05
Prevent implicit capture on GPU for RZ sims with flux injection

RemiLehe:
After #3585, WarpX crashes on GPU for RZ simulations with flux plasma injection, with:
```
an illegal memory access was encountered !!!
```
(e.g. for this example: https://github.com/ECP-WarpX/WarpX/blob/development/Examples/Tests/flux_injection/inputs_rz) 

This was due to an implicit capture from the variable `m_rz_random_theta`. This PR fixes it.


dpgrote (approved):
Looks good! Thanks for fixing!

PR #3897, opened 2023-05-05
Moving window: check pointers to F,G while applying shift

bensoub:
Small cleanup after #3884.

We want to favor this implementation
```cpp
// In file1.cpp:
multifab* F;
if (a and b or c) allocate(F);

// In file2.cpp:
if (F) do_something(F);
```
instead of this implementation
```cpp
// In file1.cpp:
multifab* F;
if (a and b or c) allocate(F);

// In file2.cpp:
if (a and b or c) do_something(F);
```
so that the code in the second file, `file2.cpp`, is more robust to changes in the logic that determines whether `F` is allocated or not (e.g., condition `if (a and b or c)` replaced by another condition), hence less error-prone.


EZoni:
@bensoub We need to rebase this PR since #3896 introduced a fix for the Read the Docs build that is currently failing here. We can discuss offline how to do this.

Source/Utils/WarpXMovingWindow.cpp
```diff
@@ -250,6 +250,7 @@ WarpX::MoveWindow (const int step, bool move_j)
         if (do_pml_dive_cleaning)
         {
             // Fine grid
+
```
EZoni:
```suggestion
```

Source/Utils/WarpXMovingWindow.cpp
```diff
@@ -263,12 +264,12 @@ WarpX::MoveWindow (const int step, bool move_j)
                     amrex::MultiFab* pml_F = pml[lev]->GetF_cp();
                     shiftMF(*pml_F, geom[lev-1], num_shift_crse, dir, lev, dont_update_cost);
                 }
-            }
+           }
```
EZoni:
```suggestion
            }
```

Source/Utils/WarpXMovingWindow.cpp
```diff
@@ -290,9 +291,9 @@ WarpX::MoveWindow (const int step, bool move_j)
             }
             if (lev > 0)
             {
-                // Coarse grid
-                if (do_pml && pml[lev]->ok())
+            if (do_pml && pml[lev]->ok())
```
EZoni:
```suggestion
                // Coarse grid
                if (do_pml && pml[lev]->ok())
```

Source/Utils/WarpXMovingWindow.cpp
```diff
@@ -290,9 +291,9 @@ WarpX::MoveWindow (const int step, bool move_j)
...
                 {
+                // Coarse grid
```
EZoni:
```suggestion
```

EZoni (changes requested):

prkkumar (approved):

EZoni (approved):
Thank you for your first WarpX PR, @bensoub! Looks good to me.

PR #3898, opened 2023-05-05
Compute dt for ckc solver in 1D

IlianCS:
This PR fixes a bug in which `dt` was set by default to 0 when using ckc solver in 1D.


RemiLehe (approved):
Thanks for fixing this!

lucafedeli88:
Good catch, @IlianCS !

PR #3915, opened 2023-05-09
Do not duplicate protected member of base class

EZoni:
In the base classes `SpectralBaseAlgorithm` and `SpectralBaseAlgorithmRZ` the member variable `m_spectral_index`, of type `SpectralFieldIndex`, is declared as `protected`. Therefore, it is meant to be used in classes derived from the base ones and should not need to be duplicated in each derived class.


ax3l (approved):
Good catch!

PR #3917, opened 2023-05-10
Use clang-tidy in CI

lucafedeli88:
This PR adds the workflow to use `clang-tidy` in CI tests.
For the moment, just two checks are included:
- [cppcoreguidelines-avoid-goto](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/cppcoreguidelines-avoid-goto.html)
- [modernize-use-nullptr](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize-use-nullptr.html)

Other tests could be added gradually by adding new lines to the `.clang-tidy` file.

⚠️ Before merging this PR, it could be useful to:
- [x]  exclude AMReX, PICSAR, and OpenPMD from the check with `clang-tidy`

The workflow generates a wrapper around `clang-tidy` that only calls `clang-tidy` on source files under `WarpX\Source` :
```
#!/bin/bash
REGEX="[a-z_A-Z0-9\/]*WarpX\/Source[a-z_A-Z0-9\/]+.cpp"
if [[ $4 =~ $REGEX ]];then
  clang-tidy $@
fi
```

The line `HeaderFilterRegex: 'Source[a-z_A-Z0-9\\/]+.H$'` in `.clang-tidy` ensures that also WarpX header files are checked.


WeiqunZhang:
I think the issue is WarpX does a superbuild that compiles amrex, picsar and openpmd cpp files. However, you can choose which files are sent to clang-tidy. You can take a look at how this is done in amrex, where only ccache-missing files are run with clang-tidy. See https://github.com/AMReX-Codes/amrex/blob/development/.github/workflows/clang.yml for examples. This has some explanation on how it's done. https://github.com/AMReX-Codes/amrex/pull/3219/ The purpose of doing this in amrex is to speed up clang-tidy. But the idea can be used in this case too.

WeiqunZhang:
Another approach could be to give your own script to cmake as clang-tidy. In your own script, you can decide whether to invoke clang-tidy or not.

lucafedeli88:
> Another approach could be to give your own script to cmake as clang-tidy. In your own script, you can decide whether to invoke clang-tidy or not.

Thanks for the suggestions! I have tried to implement your second suggestion: a wrapper around clang-tidy that only process the file if it is under `*/WarpX/Source`

.github/workflows/clang_tidy.yml
```diff
@@ -0,0 +1,54 @@
+name: 🧹 clang-tidy
+
+on: [push, pull_request]
+
+concurrency:
+  group: ${{ github.ref }}-${{ github.head_ref }}-ubuntu
```
ax3l: The CI got cancelled because this collides with he ubuntu CI workflows:
```suggestion
  group: ${{ github.ref }}-${{ github.head_ref }}-clangtidy
```

ax3l:
Let's try to systematically reach these levels of coverage after the PR is merged:
- https://github.com/AMReX-Codes/amrex/blob/development/.clang-tidy
- https://github.com/openPMD/openPMD-api/blob/dev/.clang-tidy

ax3l:
@WeiqunZhang does this cache ccache correctly or do we need to apply the same logic you added to AMReX CI for ccache? :)

WeiqunZhang:
Currently the cache data of ccache are not saved.

ax3l:
@lucafedeli88 can you see the error in this job?

I increased to build all dims for coverage, but cannot see an error besides the non-zero exit status.

lucafedeli88:
> @lucafedeli88 can you see the error in this job?
> 
> I increased to build all dims for coverage, but cannot see an error besides the non-zero exit status.

~~I can't see any error... it's a bit weird... I've noticed that the job duration is ~ 1h. Is it possible that we have a runtime limit set at 1h ?~~   It's not that... this time the job took longer

.clang-tidy
```diff
@@ -0,0 +1,6 @@
+Checks: '-*,
+    cppcoreguidelines-avoid-goto,
+    modernize-use-nullptr
+    '
+
+HeaderFilterRegex: 'Source[a-z_A-Z0-9\\/]+.H$'
```
WeiqunZhang: Is there a typo? `\\/` or just `\/`.

`.` -> `\.`?

lucafedeli88: For some reasons I thought that I had to double-escape "/", but apparently it is not necessary. 
Escaping "." does not seem to be necessary either
Thanks for spotting this!

WeiqunZhang: Just `.` will match `Source/fooH`, not just `Source/foo.H`. So `\.` is more precise.

lucafedeli88: You are right! I've replaced `.` with `\.`

WeiqunZhang (approved):

PR #3852, opened 2023-04-19
Add documentation to compile and use WarpX on Adastra (CINES) supercomputer

lucafedeli88:
This PR adds the documentation to compile and use WarpX on Adastra (CINES) supercomputer

PR #3918, opened 2023-05-10
Fix typo in Adastra cluster documentation

NeilZaim:
I was a bit slow to test and review #3852, sorry about that.

But I've tested the compilation and usage of the code on Adastra and it worked as expected.

I just found this typo, which prevented the `adastra_warpx.profile` file from being displayed in the documentation.


lucafedeli88 (approved):
Thanks for having fixed this!

ax3l (approved):
Thx! :)

PR #3921, opened 2023-05-12
Enforce const correctness using clang-tidy CI test

lucafedeli88:
This PR enforces const correctness in WarpX. 
It also adds `misc-const-correctness` to the `clang-tidy`-based CI tests.


ax3l (approved):
This is great!

The only thing I wish is that it were east-consts as in most the examples here :D https://clang.llvm.org/extra/clang-tidy/checks/misc/const-correctness.html

But I guess that is a job for `clang-format` (e.g., [this example](https://github.com/openPMD/openPMD-api/blob/dev/.clang-format)) at some point :)

PR #3924, opened 2023-05-13
Add a link to CMake documentation in the WarpX build section

IlianCS:
This PR aims to inform people that there is another build type available for CMake which is not in the WarpX documentation.


ax3l (approved):
Thank you, great idea! :)

Issue #429, opened 2019-10-02
Implement binary particle collisions


RemiLehe:
For dense plasmas, it would be useful to have binary particle collisions implemented. 

Here are a few references that could be useful:
- https://journals.aps.org/pre/abstract/10.1103/PhysRevE.55.4642
- https://www.sciencedirect.com/science/article/pii/S0021999198960491?via%3Dihub
- https://aip.scitation.org/doi/10.1063/1.4742167
- https://www.sciencedirect.com/science/article/pii/S0021999108001988?via%3Dihub
- http://www.maisondelasimulation.fr/smilei/collisions.html

Yin-YinjianZhao:
@RemiLehe @MaxThevenet @jlvay 

Some updates of my progress:

I have implemented the model proposed in Perez's paper (https://aip.scitation.org/doi/10.1063/1.4742167) by writing a small c++ code.
The code is attached in the code.zip file.
[code.zip](https://github.com/ECP-WarpX/WarpX/files/3765276/code.zip)

Currently I can reproduce the results of non-relativistic tests in Nanbu's paper (https://www.sciencedirect.com/science/article/pii/S0021999198960491?via%3Dihub).

An attached ongoing report presents my implementation in details and the test results.
[main.pdf](https://github.com/ECP-WarpX/WarpX/files/3765277/main.pdf)

Next, I will work on some tests that are relativistic.

Any comments are very welcome :)

MaxThevenet:
Thanks! Note that the smile documentation mentions a typo in Perez's paper, but it is yet another one, see http://www.maisondelasimulation.fr/smilei/collisions.html

Yin-YinjianZhao:
@MaxThevenet @RemiLehe @jlvay 

Some updates of my progress:

Finally added a relativistic test into the report: relaxation to Maxwell-Jüttner distribution. Also added some equations to present the full version of Perez's elastic model and fixed a stupid mistake under Maxence's help.
[main.pdf](https://github.com/ECP-WarpX/WarpX/files/3790358/main.pdf)

In addition, I have been working on some other tests, e.g. electrical conductivity (Perez), stoping power (Perez), energy transfer rate (Sentoku), beam relaxation (Sentoku), etc. But so far the results are not good. The reasons could be that I don't fully understand the corresponding physics behind these tests, and some details of these tests are not described clearly enough at least to me, so I applied something wrong, or there are still some issues of my implement of the algorithm. I will keep working on it.

[...]

PR #3928, opened 2023-05-16
Fix calculation of b0 in binary collisions

SeverinDiederichs:
I believe there is a typo in the calculation of b0 in the binary collisions.
Looking at Eq. 22 from the Pérez paper (https://pubs.aip.org/aip/pop/article/19/8/083104/108747/Improved-modeling-of-relativistic-collisions-and)
Screenshot 2023-05-16 at 17 09 22

the square of the last term seems to be missing. 

This PR fixes the equation.


RemiLehe (approved):
Good catch! Thanks a lot for fixing this!

EZoni (approved):
Thanks for the fix!

Comment by @RemiLehe on Slack:
The reason why the collision tests are not affected is that we explicitly give the value of the Coulomb log in the input file, while the code changed here applies to the case where the user does not give the value of the Coulomb log and WarpX tries to compute it on the fly.

ax3l:
X-ref: @Yin-YinjianZhao spotted a few mistakes in the org papers, documented in #429 and other locations.

Please see e.g., https://github.com/ECP-WarpX/WarpX/files/3799803/main.pdf in his comment about issues he found.

This PR is reverted in #3961

PR #3929, opened 2023-05-16
Remove Harris laser

### LINKED ISSUE #3927: Remove Harris laser?
Currently in WarpX, we have some specific code that defines the Harris laser profile. However, it is very rarely used in practice, and it could instead easily entered be by the user with `parse_field_function` option.

I would suggest to remove the Harris profile, as this would slightly de-clutter the code and the documentation.

RemiLehe:
Remove Harris laser, fix #3927.


EZoni (approved):
Thanks for the PR, looks good to me.

PR #2300, opened 2021-09-13
C++17, CMake 3.17+

ax3l:
Update C++ requirements to compile with C++17 or newer.

Proper C++17 compilers:
- [x] Clang 6+
- [x] GCC 7+
- [x] NVCC 11.0+
- [x] NVC++ 19.1+ 
- [x] ROCm `hipcc`/`clang++ -x hip` (already used as C++17)
- [x] oneAPI `dpcpp`/`icpx` (already used as C++17)
- [x] Intel `icpc` 19.0.1+ (version 2021+ recommended)
- [x] MSVC 19.15+ (already used as C++17)
- [x] etc. https://en.cppreference.com/w/cpp/compiler_support#C.2B.2B17_features

Checklist:
- [x] depends on #2551
- [x] depends on #2596
- [x] depends on #2595
- [x] test and update documented HPC machines
  - [x] Cori (NERSC)
  - [ ] Perlmutter (NERSC): major update downtime and will need a general module update afterwards anyway (separate PR)
  - [x] Summit (OLCF): note - slight `cuda/` module update needed (see below)
  - [x] Juwels (JSC)
  - [x] Lassen (LLNL)
  - [x] Quartz (LLNL): will need openPMD-api 0.14.4: https://github.com/openPMD/openPMD-api/pull/1157 or a bump `intel/19.1.2` to `intel/2021.4`
  - [x] Ookami (Stony Brook)
    - [x] GCC (recommended)
    - [ ] FujitsuClang (wrote to support for newer CMake, not tested yet)

Potentially force to build AMReX, PICSAR and openPMD-api in super-builds to C++17, too:
  - [x] ensured to be fine if build in C++14 and used in C++17 with openPMD-api [0.14.3+](https://github.com/openPMD/openPMD-api/releases/tag/0.14.3) #2551 
  - [x] AMReX and PICSAR: seem to cause no problems yet
  - [x] BLAS++/LAPACK++: seem to cause no problems yet

PR #3932, opened 2023-05-17
CUDA: Recommend 11.3+

ax3l:
We already saw C++17 issues with the earlier CUDA releases in the 11.0 line and found on Summit (and with ImpactX) that CUDA 11.3+ is more reliable - and now widely available.

X-Ref:
- #2300
- https://github.com/ECP-WarpX/impactx/pull/361


lucafedeli88 (approved):
Thanks for this PR!

PR #3937, opened 2023-05-23
CI: No Draft CodeQL

ax3l:
Do not run CodeQL scans on *Draft* PRs, since its code annotations can be noisy for drafts.


lucafedeli88 (approved):
Thanks for this PR!

PR #3940, opened 2023-05-24
Correctly handle `do_pml_dive_cleaning` for RZ PSATD

RemiLehe:
In the current version of WarpX, the code crashes when running RZ PSATD simulations with PML, because of this line:
https://github.com/ECP-WarpX/WarpX/blob/development/Source/Utils/WarpXMovingWindow.cpp#L253
(`pml[lev]` is not allocated ; instead with RZ PSATD, we use another object: `pml_rz[lev]`).

However, more generally, I think that `do_pml_dive_cleaning` should always be `false` in RZ PSATD, since the actual PML RZ equations implemented in WarpX do not cover this case.

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -400,7 +400,7 @@ Additional PML parameters
     Whether to use divergence cleaning for E in the PML region.
     The value must match ``warpx.do_pml_divb_cleaning`` (either both false or both true).
     This option seems to be necessary in order to avoid strong Nyquist instabilities in 3D simulations with the PSATD solver, open boundary conditions and PML in all directions. 2D simulations and 3D simulations with open boundary conditions and PML only in one direction might run well even without divergence cleaning.
-    This option is implemented only for the PSATD solver ; it is turned on by default for PSATD in Cartesian geometry.
+    This option is implemented only for the Cartesian PSATD solver ; it is turned on by default in this case.
 
 * ``warpx.do_pml_divb_cleaning`` (`bool`)
     Whether to use divergence cleaning for B in the PML region.
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -396,13 +396,13 @@ Additional PML parameters
 * ``warpx.v_particle_pml`` (`float`; default: 1)
     When ``warpx.do_pml_j_damping = 1``, the assumed velocity of the particles to be absorbed in the PML, in units of the speed of light `c`.
 
-* ``warpx.do_pml_dive_cleaning`` (`bool`; default: 1)
+* ``warpx.do_pml_dive_cleaning`` (`bool`)
     Whether to use divergence cleaning for E in the PML region.
     The value must match ``warpx.do_pml_divb_cleaning`` (either both false or both true).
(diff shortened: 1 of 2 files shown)


Docs/source/usage/parameters.rst
```diff
@@ -396,13 +396,13 @@ Additional PML parameters
 * ``warpx.v_particle_pml`` (`float`; default: 1)
     When ``warpx.do_pml_j_damping = 1``, the assumed velocity of the particles to be absorbed in the PML, in units of the speed of light `c`.
 
-* ``warpx.do_pml_dive_cleaning`` (`bool`; default: 1)
+* ``warpx.do_pml_dive_cleaning`` (`bool`)
     Whether to use divergence cleaning for E in the PML region.
     The value must match ``warpx.do_pml_divb_cleaning`` (either both false or both true).
     This option seems to be necessary in order to avoid strong Nyquist instabilities in 3D simulations with the PSATD solver, open boundary conditions and PML in all directions. 2D simulations and 3D simulations with open boundary conditions and PML only in one direction might run well even without divergence cleaning.
-    This option is implemented only for the PSATD solver.
+    This option is implemented only for the PSATD solver ; it is turned on by default for PSATD in Cartesian geometry.
```
```suggestion
    This option is implemented only for the Cartesian PSATD solver ; it is turned on by default in this case.
```

Docs/source/usage/parameters.rst
```diff
@@ -396,13 +396,13 @@ Additional PML parameters
 * ``warpx.v_particle_pml`` (`float`; default: 1)
     When ``warpx.do_pml_j_damping = 1``, the assumed velocity of the particles to be absorbed in the PML, in units of the speed of light `c`.
 
-* ``warpx.do_pml_dive_cleaning`` (`bool`; default: 1)
+* ``warpx.do_pml_dive_cleaning`` (`bool`)
     Whether to use divergence cleaning for E in the PML region.
     The value must match ``warpx.do_pml_divb_cleaning`` (either both false or both true).
     This option seems to be necessary in order to avoid strong Nyquist instabilities in 3D simulations with the PSATD solver, open boundary conditions and PML in all directions. 2D simulations and 3D simulations with open boundary conditions and PML only in one direction might run well even without divergence cleaning.
-    This option is implemented only for the PSATD solver.
+    This option is implemented only for the Cartesian PSATD solver ; it is turned on by default in this case.
```
EZoni: Could we add a similar statement also to the docs of `warpx.do_pml_divb_cleaning` below?

Source/WarpX.cpp
```diff
@@ -880,17 +880,16 @@ WarpX::ReadParameters ()
...
+        // true for Cartesian PSATD solver, false otherwise
+        do_pml_dive_cleaning = false;
+        do_pml_divb_cleaning = false;
+#ifndef WARPX_DIM_RZ
+        if (electromagnetic_solver_id == ElectromagneticSolverAlgo::PSATD)
         {
             do_pml_dive_cleaning = true;
             do_pml_divb_cleaning = true;
         }
+#endif
```
EZoni: Just below this code block there's another couple of lines that read
```cpp
 if (do_dive_cleaning) do_pml_dive_cleaning = true;
```
and similar for `do_pml_divb_cleaning`.

I think these should also be executed only in Cartesian geometry.
I believe this should be enough to fix the CI tests that currently fail.

EZoni (commented):
Thank you for catching this, @RemiLehe! I left a couple of comments.

EZoni (approved):
Thank you for this fix!

PR #3944, opened 2023-05-26
Explicitly delete NamedComponentParticleContainer's copy constructor

RemiLehe:
Fulfill the [rule-of-five](https://en.cppreference.com/w/cpp/language/rule_of_three).
The original AMReX PCs that this class wraps are also only move-able, but not copy-able.


ax3l (approved):
Thank you, LGTM! :)

PR #3946, opened 2023-05-26
[WIP] Remove radial expansion profile

RemiLehe:
The "radial expansion" profile is not commonly used, and could easily be reproduced with the parser.

Thus, in order to de-clutter the code and the documentation, I would suggest to remove it.


EZoni:
@RemiLehe 
I was assigned to this one, so I went ahead and rebased on `development`. I also removed the member function `clear` from the class `InjectorMomentum` since it seemed to be empty at this point (after removing the "custom" and "radial" options): let's see if this causes any issues that I overlooked.

EZoni:
The CI test `Python_gaussian_beam` fails because the corresponding PICMI script attempts to initialize an electron beam with a Gaussian distribution that includes radial expansion (non-zero `velocity_divergence`):
https://github.com/ECP-WarpX/WarpX/blob/eaf0e0d1965e7c97560ea4bf71c9a5623a577732/Examples/Tests/gaussian_beam/PICMI_inputs_gaussian_beam.py#L54-L56
Would you know how to reproduce this density with a parser expression that would match what WarpX would do given this input? If this is not "easy", then maybe we should consider keeping this profile instead?

RemiLehe:
Superseded by #6677

PR #3949, opened 2023-05-30
Use GPU compatible kernel in `sumParticleCharge`

roelof-groenewald:
Currently the `WarpXParticleContainer::sumParticleCharge` function does not support GPU execution. This PR fixes that.

### DIFF
--- a/Source/Particles/WarpXParticleContainer.cpp
+++ b/Source/Particles/WarpXParticleContainer.cpp
@@ -1126,6 +1126,9 @@ amrex::ParticleReal WarpXParticleContainer::sumParticleCharge(bool local) {
 
     const int nLevels = finestLevel();
 
+#ifdef AMREX_USE_OMP
+#pragma omp parallel if (amrex::Gpu::notInLaunchRegion())
+#endif
     for (int lev = 0; lev <= nLevels; ++lev) {
         for (WarpXParIter pti(*this, lev); pti.isValid(); ++pti)
         {
--- a/Source/Particles/WarpXParticleContainer.cpp
+++ b/Source/Particles/WarpXParticleContainer.cpp
@@ -1121,22 +1121,23 @@ WarpXParticleContainer::GetChargeDensity (int lev, bool local)
 amrex::ParticleReal WarpXParticleContainer::sumParticleCharge(bool local) {
 
     amrex::ParticleReal total_charge = 0.0;
+    ReduceOps<ReduceOpSum> reduce_op;
+    ReduceData<ParticleReal> reduce_data(reduce_op);
 
     const int nLevels = finestLevel();
-    for (int lev = 0; lev <= nLevels; ++lev)
-    {
 
-#ifdef AMREX_USE_OMP
-#pragma omp parallel reduction(+:total_charge)
-#endif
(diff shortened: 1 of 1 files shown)


EZoni (commented):
Thank you for this PR! It looks good to me.

One question for you, @roelof-groenewald, and/or @WeiqunZhang:
I see that similar reduce operations elsewhere in the file (e.g., the function `meanParticleVelocity`) are coded within nested conditional structures such as:
```cpp
#ifdef AMREX_USE_GPU
    if (Gpu::inLaunchRegion())
    {
        ...
    }
    else
#endif
    {
        ...
#ifdef AMREX_USE_OMP
#pragma omp parallel reduction(...)
#endif
        ...
     }
```
Do we need to do something similar here? If not, is that needed for the other reduce operations in the file or could that be simplified as the one here? Maybe I'm missing why the two cases (the function modified here and the one mentioned above) should require different implementations.

WeiqunZhang:
I don't think we need the Gpu::inLaunchRegion logic. There are a lot of those in amrex to allow users to gradually convert their codes for GPU. But that's not needed in WarpX.

WeiqunZhang:
You can keep the omp pragma but remove the reduction part. The reduce functions are omp thread safe.
```
#ifdef AMREX_USE_OMP
#pragma omp parallel if (Gpu::notInLaunchRegion())
#endif
```

WeiqunZhang:
But add `if (Gpu::notInLaunchRegion())` in case both omp and gpu are on.

EZoni:
@roelof-groenewald, I think this PR is good to go, provided that we address @WeiqunZhang's comments based on my earlier [question](https://github.com/ECP-WarpX/WarpX/pull/3949#pullrequestreview-1466376670). Or were you planning to do this separately?

roelof-groenewald:
@EZoni Thanks for reminding me about this. Your question and @WeiqunZhang's comments are for updating the `meanParticleVelocity()` function, right? So I think we can complete this PR and update that function in a new PR (I have a use case for the `sumParticleCharge()` function that currently forces me to turn of managed memory again when running on GPU - this PR will alleviate that issue in the specific case).

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -1121,23 +1121,24 @@ WarpXParticleContainer::GetChargeDensity (int lev, bool local)
...
+    ReduceData<ParticleReal> reduce_data(reduce_op);
 
     const int nLevels = finestLevel();
-    for (int lev = 0; lev <= nLevels; ++lev)
-    {
 
-#ifdef AMREX_USE_OMP
-#pragma omp parallel reduction(+:total_charge)
-#endif
+    for (int lev = 0; lev <= nLevels; ++lev) {
```
WeiqunZhang, line 1132:
```suggestion
#ifdef AMREX_USE_OMP
#pragma omp parallel if (amrex::Gpu::notInLaunchRegion())
#endif
    for (int lev = 0; lev <= nLevels; ++lev) {
```
is what I meant.

roelof-groenewald: Oh I see. Thanks for clarifying!

WeiqunZhang (approved):

PR #3950, opened 2023-05-30
Field Probe reduced diag: harmonize header format

NeilZaim:
I propose here to use the same format for the field probe headers as for the other reduced diagnostics headers. In particular, adding the `#` symbol fixes my workflow based on numpy's `genfromtext()`.

### DIFF
--- a/Examples/Tests/field_probe/analysis_field_probe.py
+++ b/Examples/Tests/field_probe/analysis_field_probe.py
@@ -24,7 +24,7 @@ filename = "diags/reducedfiles/FP_line.txt"
 
 # Open data file
 df = pd.read_csv(filename, sep=' ')
-df.columns = df.columns.str.replace('#','')
+df.columns = df.columns.str.replace('#', '', n=1)
 df = df.sort_values(by=['[2]part_x_lev0(m)'])
 
 # Select position and Intensity of timestep 500
--- a/Examples/Tests/field_probe/analysis_field_probe.py
+++ b/Examples/Tests/field_probe/analysis_field_probe.py
@@ -24,11 +24,11 @@ filename = "diags/reducedfiles/FP_line.txt"
 
 # Open data file
 df = pd.read_csv(filename, sep=' ')
-df = df.sort_values(by=['[2]part_x_lev0-(m)'])
+df = df.sort_values(by=['[2]part_x_lev0(m)'])
 
 # Select position and Intensity of timestep 500
-x = df.query('`[0]step()` == 500')['[2]part_x_lev0-(m)']
-S = df.query('`[0]step()` == 500')['[11]part_S_lev0-(W*s/m^2)']
+x = df.query('`[0]step()` == 500')['[2]part_x_lev0(m)']
+S = df.query('`[0]step()` == 500')['[11]part_S_lev0(W*s/m^2)']
 xvals = x.to_numpy()
 svals = S.to_numpy()
 
--- a/Examples/Tests/field_probe/analysis_field_probe.py
+++ b/Examples/Tests/field_probe/analysis_field_probe.py
@@ -24,6 +24,7 @@ filename = "diags/reducedfiles/FP_line.txt"
 
 # Open data file
 df = pd.read_csv(filename, sep=' ')
+df.columns = df.columns.str.replace('#','')
 df = df.sort_values(by=['[2]part_x_lev0(m)'])
 
 # Select position and Intensity of timestep 500
--- a/Examples/Tests/field_probe/analysis_field_probe.py
+++ b/Examples/Tests/field_probe/analysis_field_probe.py
@@ -24,7 +24,6 @@ filename = "diags/reducedfiles/FP_line.txt"
 
 # Open data file
 df = pd.read_csv(filename, sep=' ')
-df.columns = df.columns.str.replace('#', '', n=1)
 df = df.sort_values(by=['[2]part_x_lev0(m)'])
 
 # Select position and Intensity of timestep 500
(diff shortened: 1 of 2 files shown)


NeilZaim:
Ok, so looks like adding the `#` symbol breaks the `FieldProbe` CI test which uses `pandas` to read the file and actually wants to read the header line rather than treat it as a comment.

I still think we should use a single header format for all reduced diagnostics, but I don't want to break anything if someone relies heavily on `pandas` for their FieldProbe workflow. Options are:
1) **Start with `#`**. Currently done for all reduced diags except FieldProbe. Works with `np.genfromtext` which is what we use in nearly all CI tests. To make it work in the `pandas` FieldProbe CI test, we just need to add a line like `df.columns=df.columns.str.replace('#','')`.
2) **Start with no `#`**.  Currently only done for FieldProbe. Works fine with `pandas`, for `numpy` we need to replace `np.genfromtext(filename)` with `np.genfromtext(filename, skip_header=1)`.
3) Maybe another more standard/robust format?

Personally I'd go for option 1), but I'm fine with anything. @ax3l Any preference or it doesn't really matter?

NeilZaim:
I've implemented option 1 to make sure it actually passes the CI test, but of course this can still be discussed.

NeilZaim:
I've actually just noticed #3752 so the present PR might be redundant.

ax3l:
Thank you for the update, @NeilZaim !

Yes, I actually try to motivate people to go away from pure numpy to dataframes (pandas, dask) for particle analysis, since this has a ton of more functions (and can always be extracted to numpy). Your solution looks legit to get both working, let me add a small comment based on its docs.

Examples/Tests/field_probe/analysis_field_probe.py
```diff
@@ -24,11 +24,12 @@
 
 # Open data file
 df = pd.read_csv(filename, sep=' ')
-df = df.sort_values(by=['[2]part_x_lev0-(m)'])
+df.columns = df.columns.str.replace('#','')
```
ax3l: To make this efficient, maybe stop after first match? https://pandas.pydata.org/docs/reference/api/pandas.Series.str.replace.html
```suggestion
df.columns = df.columns.str.replace('#', '', n=1)
```

ax3l: You can also use https://pandas.pydata.org/docs/reference/api/pandas.Series.str.removeprefix.html
```suggestion
df.columns = df.columns.str.removeprefix('#')
```

NeilZaim:
Thanks, I've commited your first suggestion (it looks like the second one will check every line for a prefix so also probably less efficient).

ax3l:
@NeilZaim thank you for the PR.

Why I am a bit hesitant to merge this is, that I actually want to motivate people to avoid directly numpy usage (list of arrays) where a dataframe (table) with pandas is the better option.

This PR kinda goes in the wrong direction - making pandas use harder and numpy use easier, for something that is better processed as a dataframe. Maybe instead we might want to add a small section how to process the data in the probe to the docs instead, which can show pandas and what needs to be done for numpy. Would that be something useful?

Examples/Tests/field_probe/analysis_field_probe.py
```diff
@@ -24,11 +24,12 @@
 
 # Open data file
 df = pd.read_csv(filename, sep=' ')
-df = df.sort_values(by=['[2]part_x_lev0-(m)'])
+df.columns = df.columns.str.replace('#', '', n=1)
```
ax3l: Only this is the part I am concerned about in the comment above.

The other part is all good :+1:

NeilZaim:
@ax3l Thanks for the message.

Just to make sure, when you say that pandas is a better option, you mean specifically for the FieldProbe reduced diag, which can have multiple particles for a given timestep so it makes less sense to read the whole diag as a 2D numpy array?
If so, I have to admit that I did not consider this because I was using a point probe. So yes, it could make sense to treat the FieldProbe diag differently from the other diags.

ax3l:
> Just to make sure, when you say that pandas is a better option, you mean specifically for the FieldProbe reduced diag, which can have multiple particles for a given timestep so it makes less sense to read the whole diag as a 2D numpy array?

Yes, exactly.

Examples/Tests/field_probe/analysis_field_probe.py
ax3l: If you are ok with this, I would not add the `#`. All other breaking changes below (column names) are good and will enable other readers to pick it up easier, even though they are not the preferred interface to read this tabular data (dataframes are).

Examples/Tests/field_probe/analysis_field_probe.py
```diff
@@ -24,11 +24,12 @@
...
+df = df.sort_values(by=['[2]part_x_lev0(m)'])
```
NeilZaim:
```suggestion
df = df.sort_values(by=['[2]part_x_lev0(m)'])
```

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
```diff
@@ -164,6 +164,7 @@ FieldProbe::FieldProbe (std::string rd_name)
 
             // write header row
             int c = 0;
+            ofs << "#";
```
NeilZaim:
```suggestion
            // No "#" symbol in FieldProbe header, because the recommended way to read FieldProbe
            // output is to use pandas, which requires reading the header rather than treating it as a
            // commented line.
            // ofs << "#";
```

NeilZaim:
Sounds good, I've done as you suggested and now this PR only removes the `-` before the units in the FieldProbe diag.

PR #3953, opened 2023-05-30
Update picmi interface for MCCCollisions

dpgrote:
For MCCColiisions, allow string expressions for the background density and temperature.


roelof-groenewald (approved):
Thanks!

PR #3956, opened 2023-05-31
Fix: PSATD+RZ in SP

ax3l:
Fix the FFTW logic for RZ simulations using the PSATD field solver in single precision. Previously, the API was using the double precision interfaces for FFTW unconditionally.

Thx to @lucafedeli88 for reporting :)

### DIFF
--- a/Source/FieldSolver/SpectralSolver/SpectralFieldDataRZ.H
+++ b/Source/FieldSolver/SpectralSolver/SpectralFieldDataRZ.H
@@ -30,7 +30,11 @@ class SpectralFieldDataRZ
 #elif defined(AMREX_USE_HIP)
         using FFTplans = amrex::LayoutData<rocfft_plan>;
 #else
+#  ifdef AMREX_USE_FLOAT
+        using FFTplans = amrex::LayoutData<fftwf_plan>;
+#else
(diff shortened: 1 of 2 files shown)


Source/FieldSolver/SpectralSolver/SpectralFieldDataRZ.H
```diff
@@ -29,8 +29,12 @@ class SpectralFieldDataRZ
         using FFTplans = amrex::LayoutData<cufftHandle>;
```
dpgrote: This could be cleaned up by using `AnyFFT`,
```
using FFTplans = amrex::LayoutData<AnyFFT::VendorFFTPlan>
```

dpgrote:
This looks good. I have the one comment.

Source/FieldSolver/SpectralSolver/SpectralFieldDataRZ.H
ax3l: Good idea, done :)

dpgrote (approved):
This looks good! Thanks!

lucafedeli88 (approved):
I've tested it on my computer and it fixes the issue. Thanks a lot for having fixed this so quickly, @ax3l !

PR #3910, opened 2023-05-09
Read RZ laser files produced by lasy

IlianCS:
This PR is focused on the next implementations for lasy files reading (https://github.com/ECP-WarpX/WarpX/pull/3871).

TODO list:

- [x] Implement reading of RZ files
- [x] Implement reading of azimuthal modes from RZ files (https://fbpic.github.io/overview/pic_algorithm.html#cylindrical-grid-with-azimuthal-decomposition)
- [x] Update the docs

For a following PR:

- [ ] Read wavelength, polarization, and `e_max` from file ; make this optional for the user to pass in the input script.
- [ ] Read the flag `isLaserEnvelope` ; read the full field if it is not present

PR #3957, opened 2023-05-31
Read wavelength and flag 'isLaserEnvelope' from lasy file

IlianCS:
Next implementations for lasy files reading, following https://github.com/ECP-WarpX/WarpX/pull/3910

Files that need to be changed:

```
LaserProfiles.H
LaserProfileFromFile.cpp
LaserProfileGaussian.cpp
LaserProfileFieldFunction.cpp
LaserParticleContainer.cpp
```
Originally we wanted to read the wavelength, polarization and e_max from the lasy file. We also wanted to read the full field if isLaserEnvelope is set to False in the file.

**Update:** We are not looking for reading the polarization and e_max in the file anymore.
_e_max_: we are waiting for e_max to be stored inside the lasy file because we don't want to loop over the whole file in WarpX in order to compute it.
_polarization_: complex meta data set by the user when creating a lasy file to define, for instance, circular and elliptical lasers. As WarpX is not able to read complex polarization, we cannot read this attribute so we let it as a choice for the user to provide a linear polarization in the inputs file.
_isLaserEnvelope_: we want to read the boolean inside the lasy file but lasy does not have the capability yet to store the amplitude instead of the envelope.

TODO list (updated):

- [ ] Read wavelength from file.
- [ ] Read the flag isLaserEnvelope.


ax3l:
cc @MaxThevenet @RemiLehe I think this might need an addition of a few attributes in the openPMD laser envelope extension? Did we already start discussing this, sounds like a good idea :)

https://github.com/openPMD/openPMD-standard/blob/upcoming-2.0.0/EXT_LaserEnvelope.md

PR #3958, opened 2023-06-01
Continuous injection of moving plasma

EZoni:
To-do:
- [x] Fix GPU segfaults related to managed arena
- [x] Fix `FIXME` code
- [x] Check case of arbitrary moving window direction
- [x] Test with counter-streaming plasmas test case


Source/WarpX.H
```diff
@@ -830,7 +830,7 @@ public:
     amrex::Real getdt (int lev) const {return dt[lev];}
     int getdo_moving_window() const {return do_moving_window;}
     amrex::Real getmoving_window_x() const {return moving_window_x;}
-    amrex::Real getcurrent_injection_position () const {return current_injection_position;}
+    amrex::Vector<amrex::Real> getcurrent_injection_position () const {return current_injection_position;}
```
WeiqunZhang:
```suggestion
    amrex::Vector<amrex::Real> const& getcurrent_injection_position () const {return current_injection_position;}
```

EZoni: Thank you! Done in [c21e0f9](https://github.com/ECP-WarpX/WarpX/pull/3958/commits/c21e0f967e8da90be86d85112d8193de0ef609f1).

Source/Diagnostics/FlushFormats/FlushFormatPlotfile.cpp
```diff
@@ -302,7 +302,11 @@ FlushFormatPlotfile::WriteWarpXHeader(
 
         warpx.GetPartContainer().WriteHeader(HeaderFile);
 
-        HeaderFile << warpx.getcurrent_injection_position() << "\n";
+        for (auto& inj_pos : warpx.getcurrent_injection_position())
+        {
+             amrex::Print(0) << "FFP: inj_pos = " << inj_pos << std::endl;
```
EZoni: Leftover from debugging:
```suggestion
```

Source/Diagnostics/WarpXIO.cpp
```diff
@@ -211,8 +211,12 @@ WarpX::InitFromCheckpoint ()
         }
 
         mypc->ReadHeader(is);
-        is >> current_injection_position;
-        GotoNextLine(is);
+        for (auto& inj_pos : current_injection_position)
+        {
+             amrex::Print(0) << "WIO: inj_pos = " << inj_pos << std::endl;
```
EZoni: Leftover from debugging:
```suggestion
```

Source/Utils/WarpXMovingWindow.cpp
```diff
@@ -56,25 +56,82 @@ void
...
+                u_bulk = plasma_injector->getInjectorMomentum()->getBulkMomentum(current_injection_position[i], 0._rt, 0._rt);
+                v_bulk = PhysConst::c * u_bulk.x / std::sqrt(1._rt + u_bulk.x*u_bulk.x);
+#endif
+            }
+            else if (dir == 1)
+            {
+                // dir=1 is nothing in 1D, z in 2D, y in 3D
+                // (we do not expect to enter this code block in 1D)
+#if defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
+                u_bulk = plasma_injector->getInjectorMomentum()->getBulkMomentum(0._rt, 0._rt, current_injection_position[i]);
```
EZoni: @WeiqunZhang
I tested the code on a local GPU and I confirm that the code crashes with segfault at this line (running in DEBUG mode), unless I set `amrex.the_arena_is_managed = 1` in the input file. Do you have ideas of why calling `plasma_injector->getInjectorMomentum()->getBulkMomentum` here would cause an issue like that?

RemiLehe: It looks like `getInjectorMomentum` returns the device version of the injector (`d_inj_mom`):
https://github.com/ECP-WarpX/WarpX/blob/development/Source/Initialization/PlasmaInjector.cpp#L743

It seems that we should create another function that return `h_inj_mom` instead. (Or make that attribute `public`.)

EZoni: Thanks for catching this. Do you have a preference between the two solutions?

WeiqunZhang: It's probably better to have a new function returning the host version of the injector.

EZoni: Thanks both! Should be fixed by [ba39a94](https://github.com/ECP-WarpX/WarpX/pull/3958/commits/ba39a94e15b85df7b81ee5b30243dcd7ad3f4088).

Source/Diagnostics/FlushFormats/FlushFormatPlotfile.cpp
```diff
@@ -302,7 +302,10 @@ FlushFormatPlotfile::WriteWarpXHeader(
 
         warpx.GetPartContainer().WriteHeader(HeaderFile);
 
-        HeaderFile << warpx.getcurrent_injection_position() << "\n";
+        for (auto& inj_pos : warpx.getcurrent_injection_position())
```
WeiqunZhang:
```suggestion
        for (auto const& inj_pos : warpx.getcurrent_injection_position())
```
Tools like clang-tidy might not like it, even though it's not a big deal in this particular case.

WeiqunZhang: Actually you may have to use `auto const&` or `auto` but not `auto &`, because the function returns `Real` by value.

Source/Diagnostics/WarpXIO.cpp
```diff
@@ -211,8 +211,11 @@ WarpX::InitFromCheckpoint ()
         }
 
         mypc->ReadHeader(is);
-        is >> current_injection_position;
-        GotoNextLine(is);
+        for (auto& inj_pos : current_injection_position)
```
WeiqunZhang:
```suggestion
        for (auto const& inj_pos : current_injection_position)
```

Source/Diagnostics/FlushFormats/FlushFormatPlotfile.cpp
WeiqunZhang: Oh, I was looking at the old WarpX code. The new code return `Vector<Real> const&`. Anyway, I think it makes sense to use `auto const&`.

Source/Initialization/PlasmaInjector.H
```diff
@@ -131,7 +131,9 @@ public:
 
     InjectorPosition* getInjectorPosition ();
     InjectorDensity*  getInjectorDensity ();
-    InjectorMomentum* getInjectorMomentum ();
+
+    InjectorMomentum* getInjectorMomentumDevice ();
+    InjectorMomentum& getInjectorMomentumHost ();
```
WeiqunZhang: Should we make the return types consistent?

Source/Diagnostics/WarpXIO.cpp
EZoni: This `const` does not compile, I think because we modify `inj_pos` with `is >> inj_pos` in the `for` loop.

Source/Initialization/PlasmaInjector.H
EZoni: Do you mean that you would like both functions to return a reference `InjectorMomentum&`?

Source/Diagnostics/WarpXIO.cpp
WeiqunZhang: Yes, not this one.

Source/Initialization/PlasmaInjector.H
WeiqunZhang: Both `InjectorMomentum*`.

EZoni: Got it, done in [72b2ee2](https://github.com/ECP-WarpX/WarpX/pull/3958/commits/72b2ee27f99ff6489044c8fe84476693da2aff81).

Source/Diagnostics/FlushFormats/FlushFormatPlotfile.cpp
EZoni: Got it, done in [72b2ee2](https://github.com/ECP-WarpX/WarpX/pull/3958/commits/72b2ee27f99ff6489044c8fe84476693da2aff81).

RevathiJambunathan (approved):
Looks good to me @EZoni 
Would the documentation for continuous injection change based on the new implementation?

Source/Utils/WarpXMovingWindow.cpp
```diff
@@ -56,25 +56,82 @@ void
...
+    const int n_species = mypc->nSpecies();
+    for (int i=0; i<n_species; i++)
+    {
+        const WarpXParticleContainer& pc = mypc->GetParticleContainer(i);
+        const std::vector<std::string>& species_names = mypc->GetSpeciesNames();
+
+        // Continuously inject plasma in new cells (by default only on level 0)
+        if (pc.doContinuousInjection())
+        {
+            const std::unique_ptr<PlasmaInjector> plasma_injector
```
RemiLehe: Could we get the member variable `plasma_injector` of the instance of `PhysicalParticleContainer` instead of creating a new one?

Source/WarpX.H
```diff
@@ -1450,7 +1450,7 @@ private:
     amrex::Real v_particle_pml;
 
     amrex::Real moving_window_x = std::numeric_limits<amrex::Real>::max();
-    amrex::Real current_injection_position = 0;
+    amrex::Vector<amrex::Real> current_injection_position;
```
RemiLehe: Could we store this as a member variable of `PhysicalParticleContainer`, instead of storing it as an array in the WarpX class?

RevathiJambunathan:
Discussed offline : we need to ensure that there is no ambiguity for the case where moving_window_v==0 
and when there is no moving window (i.e. update docs to ensure that continuous injection is not turned on without moving window)

Source/WarpX.H
EZoni: I think we would have to store it as a member variable of `WarpXParticleContainer` because we would access it from objects like `pc` above, defined as `WarpXParticleContainer& pc = mypc->GetParticleContainer(i);`. Would you want to have it in `WarpXParticleContainer`?

Source/Utils/WarpXMovingWindow.cpp
EZoni: Similarly to the comment below, `plasma_injector` is a member variable of the class `PhysicalParticleContainer`, but the object from which we would try to get it is `pc` defined as `WarpXParticleContainer& pc = mypc->GetParticleContainer(i);`, unless there is a way to get only the physical particle containers from `mypc`.

EZoni:
@RemiLehe 
Do you agree with the changes in commit [9084909](https://github.com/ECP-WarpX/WarpX/pull/3958/commits/908490941827b9d90156777f5fa6589726c05079) related to the fact that we don't want to include the case `moving_window_v = 0` in the operations related to continuous injection?

Source/Utils/WarpXMovingWindow.cpp
EZoni: Discussed offline: let's define a virtual method in `WarpXParticleContainer` to get the plasma injector whenever available in a derived class.

Source/WarpX.H
EZoni: Discussed offline: let's add `current_injection_position` as a member variable of the class `WarpXParticleContainer`.

Source/Utils/WarpXMovingWindow.cpp
EZoni: Done in [74e333f](https://github.com/ECP-WarpX/WarpX/pull/3958/commits/74e333fb43efa73036798ff9c9aa5dc971a2fa1a).

Source/WarpX.H
EZoni: Done in [15bad72](https://github.com/ECP-WarpX/WarpX/pull/3958/commits/15bad72491d41c15c6ed251a5278727b9c7007a5).

Source/Utils/WarpXMovingWindow.cpp
```diff
@@ -56,25 +56,81 @@ void
...
+    const int n_species = mypc->nSpecies();
+    for (int i=0; i<n_species; i++)
+    {
+        WarpXParticleContainer& pc = mypc->GetParticleContainer(i);
+
+        // Continuously inject plasma in new cells (by default only on level 0)
+        if (pc.doContinuousInjection())
+        {
+            PlasmaInjector* plasma_injector = pc.GetPlasmaInjector();
+            if (plasma_injector == nullptr) return;
```
RemiLehe: Should we use `continue` here instead of `return`, so that we go to the next species in the `for` loop?

RemiLehe (changes requested):
Thanks for updating the PR.
I have one more minor comment.

Source/Utils/WarpXMovingWindow.cpp
EZoni: Thanks, I agree, I think this would be a bug. Fixed in [d2af9e9](https://github.com/ECP-WarpX/WarpX/pull/3958/commits/d2af9e9cadb7b8deed5a16c28835dcf30b24dfa0).

RemiLehe (approved):

EZoni:
Just a note: `clang-tidy` was complaining about something here, we might see it on `development` as well now. We will have to fix it if it shows up there.

PR #3959, opened 2023-06-01
Update required PICMI version

RemiLehe:
Because of [recent changes in WarpX](https://github.com/ECP-WarpX/WarpX/pull/3773), the required version is now 0.24.0.


lucafedeli88 (approved):
Thanks for this update, @RemiLehe

PR #3960, opened 2023-06-02
Add max_background_density to MCCollision

dpgrote:
This is required when the `background_density` is an expression.


ax3l (approved):
THanks, LGTM

roelof-groenewald (approved):

PR #3972, opened 2023-06-06
Docs: Make sbatch command consistent with the example file (Perlmutter)

IlianCS:
Just a tiny modification in the sbatch command line to make it consistent with the name of the GPU example file above.


ax3l (approved):
Thank you, good catch! :)

PR #3974, opened 2023-06-08
`geometry.is_periodic`: Only Warn

ax3l:
We want to be able to directly run a simulation from the created `warpx_used_inputs` file #3132 of a previous simulation. Since we use `geometry.is_periodic` internally to communicate boundary conditions to AMReX solvers and communication routines, we cannot fail if it is present but instead should only warn.


lucafedeli88 (approved):
Thanks for this PR, @ax3l !

PR #3978, opened 2023-06-09
Add documentation describing handling of PICMI input

dpgrote:
This adds documentation for developers describing the process used to convert PICMI input into WarpX parameters.


Docs/source/developers/python.rst
```diff
@@ -1,7 +1,87 @@
 .. _development-python:
 
-Python interface
-================
+Handling PICMI input files
+==========================
+
+There are two levels in the handling of the input files. The first is the Python level API that mirrors the input parameters and the
+second is the translation from the PICMI input to the input parameters. The two levels are described below.
```
ax3l: The title is very confusing to me, how about this? :)

```suggestion
Processing PICMI Input Options
==============================

The input parameters in a WarpX PICMI file are processed in two steps.
The first step is the Python level API, which mirrors the :ref:`C++ application input structure <running-cpp-parameters>`; the second is the translation from the PICMI input to the equivalent :ref:`app (AMReX) input file parameters <running-cpp-parameters>`.

The two steps are described below.
```

Docs/source/developers/python.rst
```diff
@@ -1,7 +1,87 @@
...
+Many of the prefix instances are predefined, for instance ``geometry`` is created in the file ``Python/pywarpx/Geometry.py``.
+In that case, ``geometry`` is an instance of the class ``Bucket`` (specified in ``Python/pywarpx/Bucket.py``),
+the general class for prefixes.
+It is called ``Bucket`` since its main purpose is a place to hold attributes.
+Most of the instances are instances of the ``Bucket`` class.
+There are exceptions, such as ``constants`` and ``diagnostics`` where extra processing is needed.
+
+There can also be instances created as needed.
+For example, for the particle species, an instance is created for each species listed in ``particles.species_names``.
+This gives a place to hold the parameters for the species, for instance ``electrons.mass``.
```
ax3l: To avoid confusion with all the *instance(s)*:
```suggestion
This gives a place to hold the parameters for the species, e.g., ``electrons.mass``.
```

ax3l (commented):
Thank you so much, this is great!!

Small suggestions added inline.

Docs/source/developers/python.rst
dpgrote: Not sure that "step" is the correct word here. How about "layers", the PICMI input layer and the C++ input layer?

ax3l: Layer is good :)
```suggestion
Processing PICMI Input Options
==============================

The input parameters in a WarpX PICMI file are processed in two layers.
The first layer is the Python level API, which mirrors the :ref:`C++ application input structure <running-cpp-parameters>`; the second is the translation from the PICMI input to the equivalent :ref:`app (AMReX) input file parameters <running-cpp-parameters>`.

The two layers are described below.
```

ax3l (approved):
:rocket: :sparkles:

PR #3979, opened 2023-06-09
Add test for MR with 2 levels

RemiLehe:
Now that #3777 is merged, it should be possible to do simulations with 2 levels of refinement.

- [x] update notes in https://warpx.readthedocs.io/en/latest/usage/parameters.html on max MR level limit (now relaxed)


RemiLehe:
The results of the test look correct, but the error (~5%) is slightly too high at this point.
Screenshot 2023-06-09 at 12 51 56 PM

ax3l:
attn @WeiqunZhang @atmyers

ax3l:
@RemiLehe oops, looks like this does not pass the test it adds. Can you please rebase and check? :)

ax3l (approved):
Ok, the changes were small:
A tiny relaxation of the threshold for the analytical solution.

I added a checksum and the values on level 0 and 1 are also reasonably close to the 1 level MR case.

PR #3980, opened 2023-06-09
Adding current density to Field Reduction reduced diagnostic

lzcorrales:
In this PR we add current density to the Field Reduction reduced diagnostic.
This will create a change in the input file where the parser function will now also be a function of jx, jy, jz
for example` <diag_name>.reduced_function(x,y,z,Ex,Ey,Ez,Bx,By,Bz,jx,jy,jz) = "parser_function"`


Source/Diagnostics/ReducedDiags/FieldReduction.H
```diff
@@ -176,8 +190,18 @@ public:
...
+                const amrex::Real jy_interp = ablastr::coarsen::sample::Interp(arrjy, jytype, cellCenteredtype,
+                                                                               reduction_coarsening_ratio, i, j, k, reduction_comp);
+                const amrex::Real jz_interp = ablastr::coarsen::sample::Interp(arrjz, jztype, cellCenteredtype,
+                                                                               reduction_coarsening_ratio, i, j, k, reduction_comp);
+
+        return reduction_function_parser(x, y, z, Ex_interp, Ey_interp, Ez_interp,
+
+                Bx_interp, By_interp, Bz_interp,
+
+                jx_interp, jy_interp, jz_interp);
```
lzcorrales:
```suggestion
        return reduction_function_parser(x, y, z, Ex_interp, Ey_interp, Ez_interp,
                Bx_interp, By_interp, Bz_interp,
                jx_interp, jy_interp, jz_interp);
```

Source/Diagnostics/ReducedDiags/FieldReduction.H
```diff
@@ -176,8 +190,16 @@ public:
...
+                const amrex::Real jx_interp = ablastr::coarsen::sample::Interp(arrjx, jxtype, cellCenteredtype,
+                                                                               reduction_coarsening_ratio, i, j, k, reduction_comp);
+                const amrex::Real jy_interp = ablastr::coarsen::sample::Interp(arrjy, jytype, cellCenteredtype,
+                                                                               reduction_coarsening_ratio, i, j, k, reduction_comp);
+                const amrex::Real jz_interp = ablastr::coarsen::sample::Interp(arrjz, jztype, cellCenteredtype,
+                                                                               reduction_coarsening_ratio, i, j, k, reduction_comp);
+
+        return reduction_function_parser(x, y, z, Ex_interp, Ey_interp, Ez_interp,
+                Bx_interp, By_interp, Bz_interp,
+                jx_interp, jy_interp, jz_interp);
```
NeilZaim:
```suggestion
                return reduction_function_parser(x, y, z, Ex_interp, Ey_interp, Ez_interp,
                        Bx_interp, By_interp, Bz_interp,
                        jx_interp, jy_interp, jz_interp);
```

Source/Diagnostics/ReducedDiags/FieldReduction.H
```diff
@@ -176,8 +190,16 @@ public:
...
                 const amrex::Real Bz_interp = ablastr::coarsen::sample::Interp(arrBz, Bztype, cellCenteredtype,
                                                                                reduction_coarsening_ratio, i, j, k, reduction_comp);
-                return reduction_function_parser(x, y, z, Ex_interp, Ey_interp, Ez_interp,
-                                                          Bx_interp, By_interp, Bz_interp);
+                const amrex::Real jx_interp = ablastr::coarsen::sample::Interp(arrjx, jxtype, cellCenteredtype,
+                                                                               reduction_coarsening_ratio, i, j, k, reduction_comp);
+                const amrex::Real jy_interp = ablastr::coarsen::sample::Interp(arrjy, jytype, cellCenteredtype,
+                                                                               reduction_coarsening_ratio, i, j, k, reduction_comp);
+                const amrex::Real jz_interp = ablastr::coarsen::sample::Interp(arrjz, jztype, cellCenteredtype,
+                                                                               reduction_coarsening_ratio, i, j, k, reduction_comp);
```
NeilZaim, line 204: We could consider doing these in loops, but I don't know if it's worth the effort.

NeilZaim (approved):
Thanks for the PR, it looks good to me (I just added a suggestion to fix an indent).

Did you test that the reduction works with the current density?

NeilZaim:
Ah, and also it could be worth adding a check that will provide an explicit error message if someone tries to use the old format (`<diag_name>.reduced_function(x,y,z,Ex,Ey,Ez,Bx,By,Bz)`).

See for instance what is done here:

https://github.com/ECP-WarpX/WarpX/blob/f6cda9ab07b683b7a2d61065439fe564b112876b/Source/Diagnostics/ReducedDiags/ReducedDiags.cpp#L84-L93

You could for example add a similar `BackwardCompatibility` function in the `FieldReduction` class, that you would call inside its constructor.

Source/Diagnostics/ReducedDiags/FieldReduction.H
lzcorrales: Thank you for the suggestion.

lzcorrales:
Thank you for taking the time to review, suggest, and comment. All comments where addressed except rewriting into a loop. 
I added a test to the reduction to ensure current density functionality. I also added a the BackwardCompatibility functionality. No loops where created to remain consistent with the rest of the code.

Source/Diagnostics/ReducedDiags/FieldReduction.H
RevathiJambunathan: HI @NeilZaim 
Thank you for this suggestion. For now, we wanted to be consistent with whatever is done for the E and B fields. 
If required, this can be done for all the fields in a separate PR

NeilZaim:
Thanks for updating this PR!
Since you've added an automated test that this feature runs correctly in the file `Examples/Tests/reduced_diags/inputs`, I think that you can also add an automated check that it produces the correct result. This can be done by updating the file `Examples/Tests/reduced_diags/analysis_reduced_diags_impl.py` to take into account the new Edotj reduced diagnostic that you've added in the input file.

RevathiJambunathan:
@NeilZaim thank you for the suggestion. 
It would be easy to do a check if we can initialize the current density from the input file, like we do for the E and B fields. But we dont have that feature at the moment.
Do you have a suggestion for a test to check the analytical jdotE result? 
For now, we just checked that it writes a result, depending on the initialization for E and B 
Similar to the FieldReduction test case already present in the file that does B^2

NeilZaim:
@RevathiJambunathan I didn't necessarily mean to compare with an analytical result (which would indeed be more difficult) but to add a check that doing the reduction manually from the plotfile produces the same result as the new reduced diagnostic (same thing as what is done for B^2 actually).

RevathiJambunathan:
@NeilZaim Ah I see that. Thanks.

RevathiJambunathan (approved):
Thanks @lzcorrales for this PR! 
Congrats on your first PR!!!

RevathiJambunathan:
@NeilZaim thanks for the suggestions! they have all been addressed now

PR #3981, opened 2023-06-12
precommit: make sure isort output is compatible with black

RTSandberg:
It was observed recently that isort and black pre-commit filters were incompatible.  This PR adds some configuration options to make the isort output compatible with what black expects


ax3l (approved):
Thank you! :)

PR #3982, opened 2023-06-12
adds nodal electrostatic solver

aeriforme:
This PR adds a nodal version of the electrostatic solver
that is used if the fields are nodal, 
i.e. if the user sets `warpx.grid_type = collocated`.


Source/FieldSolver/ElectrostaticSolver.cpp
```diff
@@ -493,58 +488,118 @@ WarpX::computeE (amrex::Vector<std::array<std::unique_ptr<amrex::MultiFab>, 3> >
...
+                            +beta_y*beta_z*0.5_rt*inv_dz*( phi_arr(i+1,j,k)-phi_arr(i-1,j,k));
+                    },
+                    [=] AMREX_GPU_DEVICE (int i, int j, int k) {
+                        Ez_arr(i,j,k) +=
+                            +(beta_z*beta_z-1)*0.5_rt*inv_dz*( phi_arr(i+1,j,k)-phi_arr(i-1,j,k) );
+                    }
+                );
+#endif
+            }
+            else{
```
RemiLehe: Could you actually do an `else if` that checks that the staggering of the `E` field is the one we expect for a Yee grid (and executes the code below if that is the case) and then an `else` clause that abort by saying that the staggering of `E` is unexpected.

Source/FieldSolver/ElectrostaticSolver.cpp
```diff
@@ -608,61 +668,114 @@ WarpX::computeB (amrex::Vector<std::array<std::unique_ptr<amrex::MultiFab>, 3> >
 
             // Calculate the magnetic field
             // Use discretized derivative that matches the staggering of the grid.
+            // Nodal solver
+            if ((bx_type == IntVect::TheNodeVector()) && (by_type == IntVect::TheNodeVector()) && (bz_type == IntVect::TheNodeVector())){
```
RemiLehe: Same comment as above.

Examples/Tests/nodal_electrostatic/analysis_3d.py
```diff
@@ -0,0 +1,27 @@
...
+sys.path.insert(1, '../../../../warpx/Regression/Checksum/')
+import checksumAPI
+
+# this will be the name of the plot file
+fn = sys.argv[1]
+
+# check that the maximum chi value is small
+fname = 'diags/reducedfiles/ParticleExtrema_beam_p.txt'
+chi_max = np.loadtxt(fname)[:,19]
+assert(np.sum(chi_max<2.e-8) == len(chi_max))
```
EZoni: I think this is equivalent, right? May be easier to read, up to your preference:
```suggestion
assert np.all(chi_max < 2e-8)
```

EZoni:
The changes in `ElectrostaticSphereLabFrame.json`, needed in order to fix CI, are not related to the changes in this PR. Once #4198 is merged, we can rebase here and merge this one.

EZoni (approved):
I think this is good to go, thanks @aeriforme!

Issue #3952, opened 2023-05-30
hdf5 files


rezaplasma:
hi every one
I am new in Warp, i have a question about output formats
i have read that Warp can write diagnostics data as` openPMD format` (e.g. hdf5), but when i run (script).py with python the results are not  as ().h5, i would just to know how i can create HDF5 files
does i need add feature in my script or not? 
thank you

ax3l:
Hi @rezaplasma,

Thanks for the question!

We implement multiple diagnostics in WarpX, ultimately moving fully to openPMD. Our openPMD output is either in `.h5` (HDF5) or `.bp` (ADIOS2): https://warpx.readthedocs.io/en/latest/dataanalysis/formats.html

> when i run (script).py with python the results are not as ().h5, i would just to know how i can create HDF5 files does i need add feature in my script or not?

Yes, there is an option to add until we change the default. In [PICMI Python input, the options you need to add are:
](https://warpx.readthedocs.io/en/latest/usage/python.html#diagnostics)
```py
ParticleDiagnostic( # ...
    warpx_format='openpmd',
    warpx_openpmd_backend='h5'
)
```
and similar for `FieldDiagnostic`.

Does this work for you? Let us know if it does or if you have further questions :)

rezaplasma:
hi @ax3l 
thanks for your answering
I will try it and let you know
does these 
` warpx_format='openpmd',` and
   ` warpx_openpmd_backend='h5'`
are arguments for
 `ParticleDiagnostic` and `FieldDiagnostic` objects ?

rezaplasma:
hi @ax3l 
i have set these arguments but i have again faced this warning
```
--- INFO    : Writing openPMD file diags/Python_LaserAcceleration_1d_plt000000
terminate called after throwing an instance of 'openPMD::error::WrongAPIUsage'
  what():  Wrong API usage: openPMD-api built without support for backend 'HDF5'.
```
does i must add an option `warpx_file_prefix`
what's the issue?
thank you

ax3l:
Hi @rezaplasma,

Thank you for the update!

How did you install WarpX? Do you mind sharing the exact details (setup of dependencies, CMake commands and output)? It looks like HDF5 was not provided during compilation, and thus cannot be used.

Best,
Axel

rezaplasma:
Hi @ax3l 

Thanks for replying
This is the error that i got

```
[...]

PR #3983, opened 2023-06-13
Docs: Debian/Ubuntu no-MPI

ax3l:
Add details how to install dependencies with `apt` w/o MPI. Add details how to compile w/ and w/o MPI.

First seen as a question in #3952.


ax3l:
Need this for further updates and for ImpactX, so going ahead and merge in.

PR #3984, opened 2023-06-13
Add Frontier Checkpointing and IO Warning

n01r:
We observed that I/O and checkpointing can be critically slow at scale with the default settings. We recommend testing and making changes to striping or AMReX checkpointing.


lucafedeli88 (approved):
Thanks for this PR!

PR #3985, opened 2023-06-14
Doc: Perlmutter (NERSC) Install Helpers

ax3l:
Based on user feedback, this streamlines the HPC dependency installation docs and makes sure one does not forget to configure the `$proj` environment variable during setup.

The new structure is would then be used blueprint for other systems.

Contains a few follow-up updates to the equivalent Frontier profile #3986.

- [x] compilation workflows tested after downtime
- [x] runtime tested on the head node


RTSandberg (approved):
Looks good to me, thanks!

PR #3986, opened 2023-06-14
Doc: Frontier (OLCF) hipcc

ax3l:
On Frontier, we switch back to `hipcc` as compiler since we see:
- HPE/Cray compilers have the performance regression we saw for 5.3-5.4 even when we load the ROCm 5.2 modules in the latest CCE programming environment (PE): `cce/15.0.0`/`cce/15.0.1`
- ROCm 5.5 still has the performance regressions

This also adds RZ+PSATD dependencies of BLAS++ & LAPACK++ and Python dependencies, and modernizes/streamlines the dependency install as in #3985.

- [x] runtime tested

### DIFF
--- a/Tools/machines/frontier-olcf/install_dependencies.sh
+++ b/Tools/machines/frontier-olcf/install_dependencies.sh
@@ -90,7 +90,7 @@ python3 -m pip install --upgrade openpmd-api
 python3 -m pip install --upgrade matplotlib
 python3 -m pip install --upgrade yt
 # install or update WarpX dependencies such as picmistandard
-python3 -m pip install --upgrade -r requirements.txt
+python3 -m pip install --upgrade -r $HOME/src/warpx/requirements.txt
 # optional: for libEnsemble
 python3 -m pip install -r $HOME/src/warpx/Tools/LibEnsemble/requirements.txt
 # optional: for optimas (based on libEnsemble & ax->botorch->gpytorch->pytorch)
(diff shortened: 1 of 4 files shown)


Tools/machines/frontier-olcf/install_dependencies.sh
```diff
@@ -0,0 +1,98 @@
...
+python3 -m pip install --upgrade cython
+python3 -m pip install --upgrade numpy
+python3 -m pip install --upgrade pandas
+python3 -m pip install --upgrade scipy
+MPICC="cc -shared" python3 -m pip install --upgrade mpi4py --no-cache-dir --no-build-isolation --no-binary mpi4py
+python3 -m pip install --upgrade openpmd-api
+python3 -m pip install --upgrade matplotlib
+python3 -m pip install --upgrade yt
+# install or update WarpX dependencies such as picmistandard
+python3 -m pip install --upgrade -r requirements.txt
```
lucafedeli88:
```suggestion
python3 -m pip install --upgrade -r $HOME/src/warpx/requirements.txt
```
Shouldn't be like this?

lucafedeli88 (dismissed):
Thanks a lot for this PR, @ax3l ! The script to install all the dependencies is especially useful!
I've left only a minor comment.

I'll take care of updating LUMI and Adastra documentation in a similar way.

Tools/machines/frontier-olcf/install_dependencies.sh
ax3l: That is definitely more robust and does not assume we are in $HOME/src/warpx, yes! :+1:

atmyers (approved):

PR #3965, opened 2023-06-03
Replace amrex::Abort with WARPX_ABORT_WITH_MESSAGE

lucafedeli88:
This PR replaces all the occurrences of `amrex::Abort` not in a GPU kernel with the `WARPX_ABORT_WITH_MESSAGE` macro, which standardizes the format and also adds some information on where the issue that triggered the abort message occurred.

PR #3987, opened 2023-06-14
Fix: 1D Abort in Kernels

ax3l:
Regression to #3965


lucafedeli88 (changes requested):
You are right! Thanks for catching this! 
Should we add a new CI test to catch similar bugs in the future?

ax3l:
I was wondering if we shall add it to the CUDA build test, now that we have multi-dim builds.
They take already quite some time, but maybe worth it.

https://github.com/ECP-WarpX/WarpX/blob/565f734a4100e9f224da8ffa7afed978df59d40e/.github/workflows/cuda.yml#L64-L75

Would do in a separate PR.

ax3l:
Merging to unbreak 1D GPU builds :)

PR #3988, opened 2023-06-14
Fix: Unused Cost GPUClock

ax3l:
By default, seems unused on ROCm.

### DIFF
--- a/Source/Particles/Deposition/CurrentDeposition.H
+++ b/Source/Particles/Deposition/CurrentDeposition.H
@@ -985,7 +985,7 @@ void doVayDepositionShapeN (const GetParticlePosition& GetPosition,
     WARPX_ABORT_WITH_MESSAGE("Vay deposition not implemented in cartesian 1D geometry");
 #endif
 
-#if !defined(AMREX_USE_GPU)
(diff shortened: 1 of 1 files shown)


Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -985,7 +985,7 @@ void doVayDepositionShapeN (const GetParticlePosition& GetPosition,
     WARPX_ABORT_WITH_MESSAGE("Vay deposition not implemented in cartesian 1D geometry");
 #endif
 
-#if !defined(AMREX_USE_GPU)
+#if !defined(WARPX_USE_GPUCLOCK)
```
ax3l, line 988: There seems more to it in 1D, maybe a simple `[[maybe_unused]]` or a more precise check here helps.

lucafedeli88 (approved):
Thanks for this PR. Since you mention that there are more in 1D, are you planning to further looking into this?

ax3l:
I tried a bit around but could not nail it down immediately (had to do the other large PRs currently open).
When you have a chance today, feel free to check if you see the same warning on LUMI? Otherwise I'll pick it up when I have time again.

PR #3995, opened 2023-06-15
Generalize buffers for `SyncRho`, `SyncCurrent`, and related functions

### LINKED ISSUE #3826: Generalize buffers in SyncRho and SyncCurrent
PRs #3277 #3814 have generalized `SyncRho` and `SyncCurrent`, so that we can now pass arbitrary `rho_fp` and `rho_cp`. However, this still uses `charge_buf` (`j_buf` for current) for particles close to the MR patch edge, which is an attribute of the WarpX class, instead of passing the buffers as an additional argument.

EZoni:
Generalize buffers for `SyncRho`, `SyncCurrent`, `AddRhoFromFineLevelandSumBoundary`, and `AddCurrentFromFineLevelandSumBoundary`.

Fixes #3826.

### DIFF
--- a/Source/Python/WarpXWrappers.cpp
+++ b/Source/Python/WarpXWrappers.cpp
@@ -710,9 +710,10 @@ namespace
     }
     void warpx_SyncCurrent (
         const amrex::Vector<std::array<std::unique_ptr<amrex::MultiFab>,3>>& J_fp,
-        const amrex::Vector<std::array<std::unique_ptr<amrex::MultiFab>,3>>& J_cp) {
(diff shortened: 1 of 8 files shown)


Source/Python/WarpXWrappers.cpp
```diff
@@ -710,9 +710,10 @@ namespace
     }
     void warpx_SyncCurrent (
```
ax3l, line 711: There is a symbol error left `undefined symbol: warpx_SyncCurrent` can you update `Source/Python/WarpXWrappers.H` as well? :)

ax3l (approved):
Looks great, thank you! :)

PR #3996, opened 2023-06-15
Impose t_lab in Lorentz Transformation for externally loaded particles

IlianCS:
Currently, starting a simulation using gamma boost will set t_lab=0 for all species. We would want to perform the Lorentz Transformation by imposing t_lab from the openPMD file when the species is loaded externally and set t_lab=0 by default for handwritten species.
t_lab is the time of the snapshot where the species is extracted (openPMD file's metadata).

### DIFF
--- a/Source/Particles/PhysicalParticleContainer.cpp
+++ b/Source/Particles/PhysicalParticleContainer.cpp
@@ -434,7 +434,6 @@ void PhysicalParticleContainer::MapParticletoBoostedFrame (
     }
     constexpr int lev = 0;
     const amrex::Real t0 = WarpX::GetInstance().gett_new(lev);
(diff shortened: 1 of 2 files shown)


Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -434,8 +432,9 @@ void PhysicalParticleContainer::MapParticletoBoostedFrame (
         x = xpr - tpr*vxpr;
```
dpgrote: Should these also be `tpr - t0`? And the comment above should be `Move the particles to where they will be at t = t0, the current simulation time in the boosted frame`?

RemiLehe (approved):

PR #3997, opened 2023-06-16
Doc: Summit (OLCF) Quick Install + Torch

ax3l:
This streamlines the Summit (OLCF) docs to use install helper scripts (see: #3985 #3986). It also adds a step to install pyTorch on ppc64le with CUDA acceleration, consistent with our main environment for our codes.

- [x] tested


RTSandberg (approved):
Looks good, thanks!

PR #4002, opened 2023-06-16
Docs: Install Dep Scripts Build Dirs

ax3l:
Clean the temporary build directories of dependencies immediately after successful install, to save on FS quotas.

Follows #3985 #3986 etc.


lucafedeli88 (approved):
Thanks for this PR, @ax3l !

PR #4005, opened 2023-06-18
Move AnyFFT.H and FFT wrappers into ablastr

lucafedeli88:
This PR moves `AnyFFT.H` and related FFT Wrappers into ablastr. It also adds a new wrapper for when FFT is not used (this is done to allow for always calling fft setup and cleanup in `main.cpp`.

We may want to discuss some issues, like: 
- do we want to place this under `ablastr/fft`  or `ablastr/utils/fft` or `ablastr/math/fft` ?
- should we add a macro like `ABLASTR_USE_FFT` in order to avoid using `WARPX_USE_PSATD` in `ablastr` ?


ax3l:
> do we want to place this under ablastr/fft or ablastr/utils/fft or ablastr/math/fft ?

I prefer `ablastr/math/fft`.

ax3l:
> should we add a macro like ABLASTR_USE_FFT in order to avoid using WARPX_USE_PSATD in ablastr ?

Sounds good :+1:

ax3l:
@lucafedeli88 can you please rebase so CI is happy? :)

ax3l (approved):
Thank you! :tada:

PR #4013, opened 2023-06-20
Doc: HPC QED & Multi-Dim

ax3l:
Update HPC machine instructions to enable multi-dims, PSATD and QED table non-defaults for all machines where these dependencies are met. This ensures users will get an as-feature-complete-as-possible install.


EZoni (approved):
Thanks for the PR, looks good to me.

PR #4014, opened 2023-06-20
Clean up of MultiFab tagging

dpgrote:
The MultiFab tags all have the suffix "[l=0]" (with the refinement level number). This puts the code adding the level suffix into the routines that setup and allocate the MultiFabs, and changes the suffix to "[level=0]". This removes the `tag` lambda which was defined in two different places. This also fixes the tags for the MultiFabs defined in the PML class which did not have the level suffix added.


EZoni (approved):
Thanks, Dave. This looks good to me. I wonder if we want to have a more verbose tag, such as `[level=0]` rather than `[l=0]`, but I leave it up to you.

RemiLehe:
I think I agree with @EZoni.
@dpgrote Do you want to update the PR accordingly?

EZoni (approved):

PR #4015, opened 2023-06-20
Fixes to remove compilation warnings

dpgrote:
This make small changes to fix warnings. It adds `static_cast` in several places and makes consistent the use of long int in particle loops in the field probes. It also adds a `ignore_unused` that arises when compiling with OpenPMD turned off.

### DIFF
--- a/Source/Diagnostics/ReducedDiags/FieldProbe.cpp
+++ b/Source/Diagnostics/ReducedDiags/FieldProbe.cpp
@@ -631,9 +631,6 @@ void FieldProbe::ComputeDiags (int step)
             amrex::Vector<int> displs_vector;
             if (amrex::ParallelDescriptor::IOProcessor()) {
                 displs_vector.resize(mpisize, 0);
-                if (mpisize > 0) {
-                    displs_vector[0] = 0;
-                }
                 total_data_size += length_vector[0];
                 for (int i=1; i<mpisize; i++) {
                     displs_vector[i] = (displs_vector[i-1] + length_vector[i-1]);
--- a/Source/Diagnostics/ReducedDiags/FieldProbe.cpp
+++ b/Source/Diagnostics/ReducedDiags/FieldProbe.cpp
@@ -631,7 +631,9 @@ void FieldProbe::ComputeDiags (int step)
             amrex::Vector<int> displs_vector;
             if (amrex::ParallelDescriptor::IOProcessor()) {
                 displs_vector.resize(mpisize, 0);
-                displs_vector[0] = 0;
+                if (mpisize > 0) {
+                    displs_vector[0] = 0;
+                }
                 total_data_size += length_vector[0];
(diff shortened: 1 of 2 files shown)


Source/Diagnostics/ReducedDiags/FieldProbe.cpp
```diff
@@ -631,7 +631,9 @@ void FieldProbe::ComputeDiags (int step)
             amrex::Vector<int> displs_vector;
             if (amrex::ParallelDescriptor::IOProcessor()) {
                 displs_vector.resize(mpisize, 0);
-                displs_vector[0] = 0;
+                if (mpisize > 0) {
```
ax3l: Shouldn't `mpisize` be at least `1` even in serial runs?

dpgrote: Yes, we know that, but the compiler doesn't. It was giving a warning that `dispels_vector` is uninitialized. That is one way to remove the warning. Is there a different way?

juliettepech (approved):

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
ax3l: Got it. We could call [vector::assign](https://en.cppreference.com/w/cpp/container/vector/assign) instead of `resize`, even if we technically only need to ensure the value for the first element is zero.

dpgrote: Looking at it again, that line `displs_vector[0] = 0;` shouldn't be needed at all since the vector is already initialized to `0` in the `resize`.

ax3l: No, resize only sets _new_ elements to the value provided.

https://en.cppreference.com/w/cpp/container/vector/resize

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
```diff
@@ -631,7 +631,6 @@ void FieldProbe::ComputeDiags (int step)
             amrex::Vector<int> displs_vector;
             if (amrex::ParallelDescriptor::IOProcessor()) {
                 displs_vector.resize(mpisize, 0);
```
ax3l: This is a bug, resize only sets _new_ elements to the value provided.

https://en.cppreference.com/w/cpp/container/vector/resize
```suggestion
                displs_vector.assign(mpisize, 0);
```
https://en.cppreference.com/w/cpp/container/vector/assign

ax3l (approved):

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
dpgrote: Yes, but here `displs_vector` is declared as an empty vector at line 631 so all elements are new.

PR #4016, opened 2023-06-20
Improve script to compute PSATD stencil

EZoni:
Improve the Python script `Tools/Algorithms/stencil.py` used to compute the PSATD stencil and estimate the number of guard cells needed for a given simulation setup. The script now parses the input file and computes boosted-frame quantities (such as cell size and time step) automatically.


Tools/Algorithms/stencil.py
```diff
@@ -429,29 +433,77 @@ def run_main(dx, dy, dz, dt, nox, noy, noz, gamma=1., galilean=False, path='.',
...
-    # Lorentz boost
-    gamma = 30.
+    # Parse path to input file from command line
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--path')
+    args = parser.parse_args()
+    path = args.path
+
+    # Open input file
+    input_file = open(path, 'r').read()
```
dpgrote: A suggestion here is to have a separate python script that reads in the input file into a dictionary (like the `parameters` dictionary in `yt`). Then all of the `re` processing wouldn't be needed. To get the number of cells would then be perhaps `nx = input_dict['amr.n_cell']` or `nx = [int(n) for n in  input_dict['amr.n_cell']]` depending on how much the script processes the input. That script would do something like this
```
for line in open(inputfilename):
    sline = line.split('=')
    parameters[sline[0].strip()] = sline[1].split()
```
(with extra checks to skip comments and blank lines and handle multiline parameters).

EZoni: Thank you for the suggestion, Dave! I have implemented it in [690d8aa](https://github.com/ECP-WarpX/WarpX/pull/4016/commits/690d8aa895db2f87f209305e5b231aa65e99a08e).
Just to confirm, do you want the parsing to be implemented in a separate Python file?

dpgrote: I was thinking that it would be a generally useful thing to have a python script that could read in an input file. That's why I suggested putting it in a separate file.

Tools/Algorithms/stencil.py
```diff
@@ -142,87 +144,115 @@ def compute_stencils(coeff_nodal, coeff_stagg, axis):
...
 
     Parameters
     ----------
-    dx : float
-        Cell size along x.
-    dy : float
-        Cell size along y.
-    dz : float
-        Cell size along z.
+    dims : int
```
dpgrote: Can you get `dims = len(dx_boosted)` instead of requiring it as an input parameter?

Tools/Algorithms/stencil.py
```diff
@@ -142,87 +144,115 @@ def compute_stencils(coeff_nodal, coeff_stagg, axis):
...
     """
+    # Default value for nx
+    if nx is None:
+        nx = np.full(shape=dims, fill_value=256)
+
     # k vectors
-    kx_arr = 2*np.pi*np.fft.fftfreq(nx, dx)
-    ky_arr = 2*np.pi*np.fft.fftfreq(ny, dy)
-    kz_arr = 2*np.pi*np.fft.fftfreq(nz, dz)
+    kx_arr = 2*np.pi*np.fft.fftfreq(nx[0], dx_boosted[0])
```
dpgrote: Could these be put in a list, making simpler code like this? Same for below.
```
k_arr = []
for i in dims:
    k_arr.append(2*np.pi*np.fft.fftfreq(nx[i], dx_boosted[I])
```

Tools/Algorithms/stencil.py
```diff
@@ -417,41 +446,87 @@ def run_main(dx, dy, dz, dt, nox, noy, noz, gamma=1., galilean=False, path='.',
...
+        sline = line.split('=')
+        # skip lines that are commented out or blank
+        skip_line = sline[0].startswith('#') or sline[0].startswith('\n')
+        if not skip_line:
+            key = sline[0].strip()
+            val = sline[1].split()
+            input_dict[key] = val
+            # The value corresponding to a given key of input_dict is a list
+            # of strings, from which we remove any leftover comments
+            for w in input_dict[key]:
```
dpgrote: I would remove the comments from `val` first, then add it to the dictionary. I think that would be cleaner.

Also, maybe do `if len(val) == 1: val = val[0]`

EZoni: I'm thinking that maybe keeping a list of length 1 for one-dimensional input values makes it easier to make the script handle 1D without further changes. For example, a line like this 
```py
    nx = np.array([int(w) for w in input_dict['amr.n_cell']])
```
probably would not work if `input_dict['amr.n_cell']` were a single integer instead of a list of integers of length 1, which would be the case for 1D input files.

Tools/Algorithms/stencil.py
EZoni: Thank you for the suggestion. I've changed the code to use lists wherever possible, avoiding conditional statements on dimensionality.

Tools/Algorithms/stencil.py
EZoni: The function to parse the WarpX input is now called `parse_input` and defined in the separate file `parser.py`. This could be a good starting point for a more comprehensive WarpX input parser. For now, I would leave it in the same folder as the script `stencil.py`, since it is the only script using it.

Tools/Algorithms/parser.py
```diff
@@ -0,0 +1,31 @@
+def parse_input(input_file):
```
dpgrote: I would make the file name more descriptive, something like `input_file_parser.py`. And the function name `parse_input_file`.

Tools/Algorithms/parser.py
```diff
@@ -0,0 +1,31 @@
...
+        Path to input file.
+
+    Returns
+    -------
+    input_dict : dictionary
+        Dictionary storing WarpX input parameters
+        (parameter's name stored as key, parameter's value stored as value).
+    """
+    input_dict = dict()
+    for line in open(input_file):
```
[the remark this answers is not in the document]
dpgrote: Using `with` would better (and would remove this warning).
```
with open(input_file) as ff:
    for line in ff:
```

Tools/Algorithms/parser.py
```diff
@@ -0,0 +1,31 @@
...
+        sline = line.split('=')
+        # skip lines that are commented out or blank
+        skip_line = sline[0].startswith('#') or sline[0].startswith('\n')
```
dpgrote: The parameter values can extend over multiple lines (for instance function expressions), so there may be cases where `len(sline) == 1` when it is not a comment nor blank line, which would cause an error below when using `sline[1]`.

Tools/Algorithms/parser.py
```diff
@@ -0,0 +1,31 @@
...
+        if not skip_line:
+            key = sline[0].strip()
+            val = sline[1].split()
+            # The value corresponding to a given key of input_dict is a list
+            # of strings, from which we remove any leftover comments
+            for w in val:
```
dpgrote: Modifying the iterator in a loop is not good practice. In this case, if `val` has two elements that start with '#', the loop crashes with a ValueError. This can be avoided by adding a `break`. Here's an alternative, somewhat simpler version
```
for i in range(len(val)):
    if val[i].startswith('#'):
        val = val[:i]
        break
```

Tools/Algorithms/stencil.py
```diff
@@ -130,112 +133,127 @@ def compute_stencils(coeff_nodal, coeff_stagg, axis):
...
+              stencil_avg_stagg  = np.sum(stencil_stagg, axis=-1)
+              stencil_avg_stagg /= stencil_stagg.shape[-1]
+        elif axis == 1 or axis == -1:
+              # Averaged over kx
+              stencil_avg_nodal  = np.sum(stencil_nodal, axis=0)
+              stencil_avg_nodal /= stencil_nodal.shape[0]
+              stencil_avg_stagg  = np.sum(stencil_stagg, axis=0)
+              stencil_avg_stagg /= stencil_stagg.shape[0]
+    elif dims == 3:
+        if axis == 0:
```
dpgrote: Here's a trick I sometimes use that can remove these `if` blocks (plus some other simplification)
```
# Average over the other two directions
i1 = (axis + 1) % 3
i2 = (axis + 2) % 3
stencil_avg_nodal = (stencil_nodal.sum(axis=(i1,i2))/
                                  (stencil_nodal.shape[i1]*stencil_nodal.shape[i2]))
stencil_avg_stagg = (stencil_stagg.sum(axis=(i1,i2))/
                                  (stencil_stagg.shape[i1]*stencil_stagg.shape[i2]))
```

Tools/Algorithms/stencil.py
```diff
@@ -130,112 +133,127 @@ def compute_stencils(coeff_nodal, coeff_stagg, axis):
...
+            k_arr_s.append(k_arr[i])
 
     # Mesh in k space
-    kx_c, ky_c, kz_c = np.meshgrid(kx_arr_c, ky_arr_c, kz_arr_c)
-    kx_s, ky_s, kz_s = np.meshgrid(kx_arr_s, ky_arr_s, kz_arr_s)
+    k_c = np.meshgrid(*k_arr_c)
+    k_s = np.meshgrid(*k_arr_s)
+    kk_c = 0.
+    kk_s = 0.
+    for k in k_c:
```
dpgrote: How about
`kk_c = np.sqrt((k_c**2).sum())`

Tools/Algorithms/stencil.py
```diff
@@ -342,68 +354,62 @@ def run_main(dx, dy, dz, dt, nox, noy, noz, gamma=1., galilean=False, path='.',
...
     # Maximum number of cells
-    ncx = 65
-    ncy = 65
-    ncz = 65
-
-    # Array of cell numbers
-    cx = np.arange(ncx)
-    cy = np.arange(ncy)
-    cz = np.arange(ncz)
+    nc = []
```
dpgrote: This will also work
`nc = dims*[65]`

Tools/Algorithms/stencil.py
```diff
@@ -342,68 +354,62 @@
...
+        label = ['x', 'z']
+    elif dims == 3:
+        label = ['x', 'y', 'z']
 
     # Plot stencils
-    plot_stencil(cx, stencils['x']['nodal'], stencils['x']['stagg'], 'x', path, name)
-    plot_stencil(cy, stencils['y']['nodal'], stencils['y']['stagg'], 'y', path, name)
-    plot_stencil(cz, stencils['z']['nodal'], stencils['z']['stagg'], 'z', path, name)
+    for i, s in enumerate(stencils):
+        plot_stencil(np.arange(nc[i]), s['nodal'], s['stagg'], label[i], path, name)
```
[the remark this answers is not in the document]
dpgrote: This failure is because the `if` block setting label could leave label undefined if dims is not 1, 2, or 3. Would this be too hard to understand?
`label = ['x', 'y', 'z'][::4-dims]`
Or
```
label = ['x']
if dims == 3: label.append('y')
if dims > 1: label.append('z')
```

Tools/Algorithms/stencil.py
```diff
@@ -416,42 +422,68 @@ def run_main(dx, dy, dz, dt, nox, noy, noz, gamma=1., galilean=False, path='.',
...
-    dt = 1e-14
-    # Spectral order
-    nox = 8
-    noy = 8
-    noz = 16
-    # Lorentz boost
-    gamma = 30.
+    # Parse path to input file from command line
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--path')
```
dpgrote: A more descriptive input argument would be better, perhaps '--input_file'

Tools/Algorithms/parser.py
EZoni: About this, it's a good point. For now I added an extra condition that checks whether the line has only one entry, in which case it's skipped, but I still have to figure out how to possibly add that entry to a previous key of the dictionary, assuming it's always a continuation of a previous input parameter. It is not relevant for the purpose of this PR, but in general it is something that the parser should be able to do correctly.

Tools/Algorithms/stencil.py
EZoni: I went for this solution, it is still better than my previous implementation:
```py
kk_c = np.sqrt(sum(k**2 for k in k_c))
```
The other solution didn't seem to work, I think because the operator `**` was mixing a list (`k_c`) and an int (the exponent).

Tools/Algorithms/stencil.py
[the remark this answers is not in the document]
EZoni: I went for your second solution, I think it's more readable than the first one, even though slightly more verbose.

Tools/Algorithms/stencil.py
```diff
@@ -416,42 +413,68 @@ def run_main(dx, dy, dz, dt, nox, noy, noz, gamma=1., galilean=False, path='.',
...
+        psatd_order[-1] = int(input_dict['psatd.noz'][0])
+
     # Galilean flag
-    galilean = True
-    # Output path
-    path = '.'
-    # Output name tag (can be empty: '')
-    name = 'test'
-    # --
+    # TODO Extend for alternative syntax psatd.v_galilean
```
EZoni: Remove this comment:
```suggestion
```

Tools/Algorithms/input_file_parser.py
```diff
@@ -0,0 +1,32 @@
...
+    input_dict : dictionary
+        Dictionary storing WarpX input parameters
+        (parameter's name stored as key, parameter's value stored as value).
+    """
+    input_dict = dict()
+    with open(input_file) as ff:
+        for line in ff:
+            sline = line.split('=')
+            # skip lines that are commented out, blank, or continuation of previous parameters
+            skip_line = sline[0].startswith('#') or sline[0].startswith('\n') or len(sline) == 1
```
dpgrote, line 21: This is good enough for now. The format of the input file is much more flexible though, for example allowing multiple parameters in a line, though those things probably are not used often.

Tools/Algorithms/stencil.py
```diff
@@ -130,112 +133,121 @@ def compute_stencils(coeff_nodal, coeff_stagg, axis):
     coeff_stagg : numpy.ndarray
         Leading spectral staggered coefficient of the general PSATD equations.
     axis : int
-        Axis or direction (must be 0, 1, or 2).
+        Axis or direction (must be 0, 1, 2 or -1).
```
dpgrote: Can you add the meaning of `-1`, the `z` axis for both 2D and 3D?

Tools/Algorithms/stencil.py
```diff
@@ -5,13 +5,16 @@
 measure is not larger than the error threshold).
 Reference: https://doi.org/10.1016/j.cpc.2022.108457
 
-Run the script simply with "python stencil.py" (or with "run stencil.py"
-using IPython). The user can modify the input parameters set in the main
-function at the end of the file.
+How to run the script:
+$ python stencil.py --path path_to_input_file
```
dpgrote: This needs to be updated to use the correct option `--input_file`.

EZoni: Thanks, done in [2c21e08](https://github.com/ECP-WarpX/WarpX/pull/4016/commits/2c21e0848753de989ed1edbf5d502a62f7a8098f).

Tools/Algorithms/stencil.py
EZoni: Thanks, done in [2c21e08](https://github.com/ECP-WarpX/WarpX/pull/4016/commits/2c21e0848753de989ed1edbf5d502a62f7a8098f).

Tools/Algorithms/stencil.py
EZoni: Here I'm thinking that maybe the `if` blocks make the code a bit easier to grasp, even though more verbose, so I would be inclinded to keep it as it is, if that's ok with you?

EZoni: Oh no, wait, I thought you were talking about the dimensional `if` blocks. I think I can implement the change you suggest here actually.

EZoni: Done in [2c21e08](https://github.com/ECP-WarpX/WarpX/pull/4016/commits/2c21e0848753de989ed1edbf5d502a62f7a8098f).

Tools/Algorithms/stencil.py
```diff
@@ -130,112 +133,110 @@ def compute_stencils(coeff_nodal, coeff_stagg, axis):
...
-          # Averaged over kx and ky
-          stencil_avg_nodal  = np.sum(np.sum(stencil_nodal, axis=1), axis=0)
-          stencil_avg_nodal /= (stencil_nodal.shape[1]*stencil_nodal.shape[0])
-          stencil_avg_stagg  = np.sum(np.sum(stencil_stagg, axis=1), axis=0)
-          stencil_avg_stagg /= (stencil_stagg.shape[1]*stencil_stagg.shape[0])
+    if dims == 1:
+        stencil_avg_nodal = stencil_nodal
+        stencil_avg_stagg = stencil_stagg
+    elif dims == 2:
+        if axis == 0:
```
dpgrote: The same simplification that was done for 3D can be done here, with
```
i1 = (axis + 1) % 2
```

Tools/Algorithms/stencil.py
```diff
@@ -130,112 +133,110 @@ def compute_stencils(coeff_nodal, coeff_stagg, axis):
...
+              stencil_avg_nodal  = np.sum(stencil_nodal, axis=-1)
+              stencil_avg_nodal /= stencil_nodal.shape[-1]
+              stencil_avg_stagg  = np.sum(stencil_stagg, axis=-1)
+              stencil_avg_stagg /= stencil_stagg.shape[-1]
+        elif axis == 1 or axis == -1:
+              # Average over kx
+              stencil_avg_nodal  = np.sum(stencil_nodal, axis=0)
+              stencil_avg_nodal /= stencil_nodal.shape[0]
+              stencil_avg_stagg  = np.sum(stencil_stagg, axis=0)
+              stencil_avg_stagg /= stencil_stagg.shape[0]
```
EZoni:
```suggestion
        # Average over the other direction
        i1 = (axis + 1) % 2
        stencil_avg_nodal = stencil_nodal.sum(axis=i1) / stencil_nodal.shape[i1]
        stencil_avg_stagg = stencil_stagg.sum(axis=i1) / stencil_stagg.shape[i1]
```

dpgrote (approved):
This looks good, thanks!

PR #4017, opened 2023-06-20
Doc: IPAC23 Paper

ax3l:
Document the IPAC23 paper, which uses WarpX to train an ML model for ImpactX.


ax3l:
ping @RTSandberg :)

ax3l:
have to merge because I will add another one in the same section in #4027

Issue #3938, opened 2023-05-23
WarpX on new HPC cluster at UCI


floresv299:
Hello, I'm trying to install and run Warpx on my institution's cluster. I have some experience installing softwares but there are some dependancies that I'm not sure what they do or how to get them. Thank you for the help !

ax3l:
Hi @floresv299,

Thank you for reaching out! We are glad to help you to get started.

The best way to run on institutional clusters is to use their modules for base dependencies, e.g., compilers, MPI and any dependency we can find, and only add what we need manually.

In general, all (potential and optional) dependencies of WarpX are listed here:
https://warpx.readthedocs.io/en/latest/install/dependencies.html#install-dependencies

I would recommend to start with the following, which you should get from `module load` on your HPC system:
- compiler
- CMake
- MPI

Let's then get into a few details of your cluster:
- Is there a public documentation to the machine you are referring to? Can you provide us with that link to guide you?
  - Is this a CPU or GPU machine?
- Do you already run other codes on the machine?
  - Are there job scripts that you already use?
- For WarpX, which features do you like to use?
  - GPU acceleration?
  - 1D/2D/3D and/or RZ geometry?
  - Advanced field solvers (PSATD)?

Based on this, we can guide you best how to do the next steps.

Just to share upfront, we document a lot of clusters around the world in our manual's HPC section. We can derive from one that is close to yours and even add your workflow there as well if you like.
https://warpx.readthedocs.io/en/latest/install/hpc.html

ax3l:
Hi @floresv299, how is it going? :)

floresv299:
Dear Axel,

Thank you very much for reaching out, apologies for the late reply. Here
are the answers to some of your questions:

1) Here is the link to the documentation for our HPC
https://rcic.uci.edu/hpc3/index.html
This machine is able to run on a CPU and GPU.

2) I have used both CPU and GPU computing for EPOCH and most recently
for FBPIC. Below, I will attach some of the job scripts.

3) As for features it would be nice to have GPU acceleration and 1/2/3D
geometry. In terms of advanced field solvers, what would you recommend? The
work that our lab does is LWFA and solid-target interaction,
utilizing boosted frames to look at betatron emission and high harmonic
generation.

Thank you again for the help, it means a lot !

ax3l:
Hi @floresv299 ,

[...]

PR #4021, opened 2023-06-21
Doc: HPC3 (UCI) HDF5 & GPU Visibility

ax3l:
Switch the HDF5 module to one that is picked up well by our build logic.

Slurm is not well configured to properly address GPU visibility per MPI rank. Thus, we define the node layouts during compile-time.

Related to tests by @floresv299 in https://github.com/ECP-WarpX/WarpX/issues/3938#issuecomment-1601556070

### DIFF
--- a/Docs/source/install/hpc/hpc3.rst
+++ b/Docs/source/install/hpc/hpc3.rst
@@ -94,7 +94,7 @@ Use the following :ref:`cmake commands <building-cmake>` to compile:
    cd $HOME/src/warpx
    rm -rf build
 
-   cmake -S . -B build -DWarpX_COMPUTE=CUDA -DGPUS_PER_SOCKET=${GPUS_PER_SOCKET} -DGPUS_PER_NODE=${GPUS_PER_NODE} -DWarpX_PSATD=ON -DWarpX_QED_TABLE_GEN=ON -DWarpX_LIB=ON -DWarpX_DIMS="1;2;RZ;3"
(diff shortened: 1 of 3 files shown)


ax3l:
@WeiqunZhang can you please double check if I controlled GPU visibility here? :)

Should we add env-var control to define the defaults in the build system, so users do not need to run an extended configure line for building?

RE: https://github.com/ECP-WarpX/WarpX/issues/3938#issuecomment-1601751687

ax3l:
@floresv299 @WeiqunZhang I updated the PR once more to try to properly match MPI rank to GPU device.

I have not received an answer from HPC3 support at UCI yet about optimal pinning, so let us hope the GPUs are ordered reasonably.

This is based on the work-around for under-configured Slurm on Perlmutter (NERSC).

floresv299:
Hi Axel, apologies for the delay, there wasn't enough resources to run a simulation on my hpc all of yesterday. However I recompiled the code using the -DGPUS_PER_SOCKET=${GPUS_PER_SOCKET} -DGPUS_PER_NODE=${GPUS_PER_NODE} along with having export GPUS_PER_SOCKET=2
export GPUS_PER_NODE=4 in my profile. That seemed to do the job, no more errors encountered. Attached is the output file.
[Warpx_output_4_GPUS.txt](https://github.com/ECP-WarpX/WarpX/files/11851099/Warpx_output_4_GPUS.txt)

PR #4022, opened 2023-06-21
For ReducedDiags, write header on restart if file does not exist

dpgrote:
This is a little nicety needed since sometimes restarts will be done in a new place so new reduced diags files are written.


lucafedeli88 (approved):
Thanks for this @dpgrote !

PR #4024, opened 2023-06-22
add collider-relevant reduced diags 

aeriforme:
This PR adds collider-relevant reduced diagnostics given two arbitrary species. 
The two species should be the ones associated to two colliding bunches propagating along `z`. 
The diagnostic does not work in RZ geometry yet. 

For each species, define the transverse coordinates `x` in 2D and `x`, `y` in 3D as well as the angles with respect to the propagation direction `theta_x` in 2D and `theta_x`, `theta_y` in 3D (e.g. `theta_x = arctan2(u_x, u_z)` where `u_x` is the normalized momentum along `x`).

Then, the quantities calculated by the collider-relevant reduced diagnostics are:
* time derivative of the luminosity
* for each species
  * average transverse positions: `x_ave` in 2D; `x_ave`, `y_ave` in 3D for each species
  * standard deviation of the transverse positions: `x_std` in 2D; `x_std`, `y_std` in 3D
  * minimum angles: `theta_x_min` in 2D; `theta_x_min`, `theta_y_min` in 3D
  * average angles: `theta_x_ave` in 2D; `theta_x_ave`, `theta_y_ave` in 3D
  * maximum angles: `theta_x_max` in 2D; `theta_x_max`, `theta_y_max` in 3D
  * standard deviation of the angles: `theta_x_std` in 2D; `theta_x_std`, `theta_y_std` in 3D

In the analysis script of the CI test, we also reuse the Python input file parser originally in Tools/Algorithms, now moved to the new directory Tools/Parser.


Examples/Tests/collider_relevant_diags/analysis_multiple_particles.py
```diff
@@ -0,0 +1,162 @@
...
+    ux1, ux2, ux3 = [float(w) for w in input_dict[f'{species}.multiple_particles_ux']]
+    uy1, uy2, uy3 = [float(w) for w in input_dict[f'{species}.multiple_particles_uy']]
+    uz1, uz2, uz3 = [float(w) for w in input_dict[f'{species}.multiple_particles_uz']]
+
+    x = np.array([float(w) for w in input_dict[f'{species}.multiple_particles_pos_x']])
+    y = np.array([float(w) for w in input_dict[f'{species}.multiple_particles_pos_y']])
+
+    w = np.array([float(w) for w in input_dict[f'{species}.multiple_particles_weight']])
+
+    px1, px2, px3 = ux1*m_e*c, ux2*m_e*c, ux3*m_e*c
```
EZoni:
```suggestion
```

Examples/Tests/collider_relevant_diags/analysis_multiple_particles.py
```diff
@@ -0,0 +1,162 @@
...
+    py1, py2, py3 = uy1*m_e*c, uy2*m_e*c, uy3*m_e*c
```
EZoni:
```suggestion
```

Examples/Tests/collider_relevant_diags/analysis_multiple_particles.py
```diff
@@ -0,0 +1,162 @@
...
+    pz1, pz2, pz3 = uz1*m_e*c, uz2*m_e*c, uz3*m_e*c
+
```
EZoni:
```suggestion
```

Examples/Tests/collider_relevant_diags/analysis_multiple_particles.py
```diff
@@ -0,0 +1,162 @@
...
+    g1, g2, g3 = np.sqrt(1.+ux1**2+uy1**2+uz1**2), np.sqrt(1.+ux2**2+uy2**2+uz2**2), np.sqrt(1.+ux3**2+uy3**2+uz3**2)
+
+    E1, E2, E3 = m_e*c**2*(g1-1.), m_e*c**2*(g2-1.), m_e*c**2*(g3-1.)
```
EZoni:
```suggestion
```

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,563 @@
+/* Copyright 2019-2020 Yinjian Zhao
+ *
+ * This file is part of WarpX.
+ *
+ * License: BSD-3-Clause-LBNL
+ */
```
EZoni:
```suggestion
/* Copyright 2023 The WarpX Community
 *
 * This file is part of WarpX.
 *
 * Authors: Yinjian Zhao, Arianna Formenti
 * License: BSD-3-Clause-LBNL
 */
```

Source/Diagnostics/ReducedDiags/ColliderRelevant.H
```diff
@@ -0,0 +1,53 @@
+/* Copyright 2019-2020 Yinjian Zhao
+ *
+ * This file is part of WarpX.
+ *
+ * License: BSD-3-Clause-LBNL
+ */
```
EZoni:
```suggestion
/* Copyright 2023 The WarpX Community
 *
 * This file is part of WarpX.
 *
 * Authors: Yinjian Zhao, Arianna Formenti
 * License: BSD-3-Clause-LBNL
 */
```

Source/Diagnostics/ReducedDiags/ColliderRelevant.H
```diff
@@ -0,0 +1,54 @@
...
+    ColliderRelevant(std::string rd_name);
+
+    /// name of species
+    std::vector<std::string> m_beam_name;
+
+    /**
+     * This funciton computes the particle extrema
+     *
+     * @param[in] step current time step
+     */
```
EZoni: Do you want to update this docstring with the list of all new reduced diags that this function actually computes? I would also add a `\brief` before the text so that Doxygen can process it correctly.

Docs/source/usage/parameters.rst
```diff
@@ -2985,6 +2985,56 @@ Reduced Diagnostics
         1 or 0, it is possible to compute the charge on only some part of the
         embedded boundary.
 
+    * ``Collider-relevant``
```
ax3l: The existing options seem to use CamelCase, so maybe for consistency:
```suggestion
    * ``ColliderRelevant``
```
(+ impl below :) )

Source/Diagnostics/ReducedDiags/ColliderRelevant.H
```diff
@@ -0,0 +1,61 @@
+/* Copyright 2023 The WarpX Community
+ *
+ * This file is part of WarpX.
+ *
+ * Authors: Yinjian Zhao, Arianna Formenti
```
ax3l: Unique enough on this file :)
```suggestion
 * Authors: Arianna Formenti
```

Source/Diagnostics/ReducedDiags/ColliderRelevant.H
```diff
@@ -0,0 +1,61 @@
...
+    {
+        std::string header;
+        int idx;
+    };
+
+    /// map to store header texts and indices of the reduced diagnostics
+    std::map<std::string, aux_header_index> m_headers_indices;
+};
+
+#endif
```
ax3l: optional, just nice to have
```suggestion
#endif  // WARPX_DIAGNOSTICS_REDUCEDDIAGS_COLLIDERRELEVANT_H_
```

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,559 @@
+/* Copyright 2023 The WarpX Community
+ *
+ * This file is part of WarpX.
+ *
+ * Authors: Yinjian Zhao, Arianna Formenti
```
ax3l: Mostly you, you can credit Yin for his earlier reduced diags here if sufficiently similar
```suggestion
 * Authors: Arianna Formenti, Yinjian Zhao
```

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,559 @@
...
+
+        const bool is_photon = myspc.AmIA<PhysicalSpecies::photon>();
+
+        // photon number density is not available yet
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+            !is_photon,
+            "Collider-relevant diagnostic does not work for colliding photons yet");
+    }
+
+    auto all_diag_names = std::vector<std::string>{};
```
ax3l: This block deserves an inline comment :sweat_smile: 
_What does it do?_ (Not _how_.)

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,559 @@
...
+
+#include <ablastr/coarsen/sample.H>
+#include <ablastr/warn_manager/WarnManager.H>
+
+#include <algorithm>
+#include <array>
+#include <cmath>
+#include <fstream>
+#include <limits>
+#include <map>
```
ax3l, line 56: For the unique_ptr et al.
```suggestion
#include <map>
#include <memory>
```

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,559 @@
...
+        thetax_ave = amrex::get<1>(r_ave);
+        amrex::ParallelDescriptor::ReduceRealSum({x_ave, thetax_ave}, amrex::ParallelDescriptor::IOProcessorNumber());
+        x_ave = x_ave / wtot;
+        thetax_ave = thetax_ave / wtot;
+
+        // x_std, y_std, thetax_std, thetay_std
+        amrex::Real x_std = 0.0_rt;
+        amrex::Real thetax_std = 0.0_rt;
+
+        amrex::ReduceOps<ReduceOpSum, ReduceOpSum> reduce_ops_std;
```
ax3l: We might be able to avoid a few of the repeated reduces over all particles if we reformulate some of the statistical quantities needed. I sent you a PDF with the concept.

ax3l (commented):
Nice, thank you and very clean! I added a few small suggestions.

Docs/source/usage/parameters.rst
```diff
@@ -2985,6 +2985,56 @@ Reduced Diagnostics
...
+        .. math::
+
+            \frac{dL}{dt} = 2 c \iiint  n_1(x,y,z) n_2(x,y,z) dx dy dz
+
+        where :math:`n_1`, :math:`n_2` are the number densities of the two colliding species.
+
+        [3], [4], [5]: The minimum, average and maximum values of the quantum parameter :math:`\chi` of species 1:
+        :math:`\chi_{min}`,
+        :math:`\langle \chi \rangle`,
+        :math:`\chi_{max}`.
```
lucafedeli88: What happens with this diagnostic if QED is not enabled? We may want to write this in the documentation.

Examples/Tests/collider_relevant_diags/inputs_3d_multiple_particles
```diff
@@ -0,0 +1,137 @@
...
+qed_qs.photon_creation_energy_threshold = 0.
+qed_qs.lookup_table_mode = builtin
+qed_qs.chi_min = 1.e-3
+warpx.do_qed_schwinger = 0
+
+#################################
+######### DIAGNOSTICS ###########
+#################################
+# FULL
+diagnostics.diags_names = diag1 diag2
```
lucafedeli88, line 98: Maybe for clarity we could call these something like `diag_yt` and `diag_opmd`

Regression/WarpX-tests.ini
```diff
@@ -209,6 +209,22 @@ compileTest = 0
 doVis = 0
 analysisRoutine = Examples/Tests/btd_rz/analysis_BTD_laser_antenna.py
 
+[collider_diagnostics]
+buildDir = .
+inputFile = Examples/Tests/collider_relevant_diags/inputs_3d_multiple_particles
+runtime_params = warpx.abort_on_warning_threshold=high
```
lucafedeli88, line 215: Why do we set this at `high` ? Does the simulation generate some `low` or `medium` priority warnings?

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,559 @@
...
+{
+    // read colliding species names - must be 2
+    amrex::ParmParse pp_rd_name(rd_name);
+    pp_rd_name.getarr("species", m_beam_name);
+
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+        m_beam_name.size() == 2u,
+        "Collider-relevant diagnostics must involve exactly two species");
+
+    // RZ coordinate is not working
```
lucafedeli88:
```suggestion
    // RZ coordinate is not supported
```
I would suggest writing this.

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,559 @@
...
+#if (defined WARPX_DIM_RZ)
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(false,
+        "Collider-relevant diagnostics do not work in RZ geometry.");
```
lucafedeli88:
```suggestion
    WARPX_ABORT_WITH_MESSAGE(
        "Collider-relevant diagnostics do not work in RZ geometry.");
```
You can use this here.

lucafedeli88 (changes requested):
Thanks a lot for this PR, @aeriforme ! I've left few minor comments

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,559 @@
...
+#include <cmath>
+#include <fstream>
+#include <limits>
+#include <map>
+#include <vector>
+
+using namespace amrex;
+
+ColliderRelevant::ColliderRelevant (std::string rd_name)
+: ReducedDiags{rd_name}
```
WeiqunZhang:
```suggestion
: ReducedDiags{std::move(rd_name)}
```
Oh, we also need to change from `amrex::ParmParse pp_rd_name(rd_name);` to `amrex::ParmParse pp_rd_name(m_rd_name);`.

Maybe you can also add a change to the constructor of ReduceDiags from
```
ReducedDiags::ReducedDiags (std::string rd_name)
{
    m_rd_name = rd_name;
```
to
```
ReducedDiags::ReducedDiags (std::string rd_name)
    : m_rd_name(std::move(rd_name))
{
```

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
WeiqunZhang: Is that just `std::vector<std::string> all_diag_names;`?

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,559 @@
...
+                const amrex::Real uz = p.rdata(PIdx::uz);
+                const amrex::Real thetax = std::atan2(ux, uz);
+                return {w*x, w*thetax};
+            },
+            reduce_ops_ave);
+        x_ave = amrex::get<0>(r_ave);
+        thetax_ave = amrex::get<1>(r_ave);
+        amrex::ParallelDescriptor::ReduceRealSum({x_ave, thetax_ave}, amrex::ParallelDescriptor::IOProcessorNumber());
+        x_ave = x_ave / wtot;
+        thetax_ave = thetax_ave / wtot;
```
WeiqunZhang: What if wtot is zero? Since the mpi_reduce above only reduces to the I/O process, if one of the non-I/O processes has no particles, wtot will be zero on that process. So maybe do this only `if (amrex::ParallelDescriptror::IOProcessor())`.

WeiqunZhang (changes requested):

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,559 @@
...
+        thetax_max = amrex::get<1>(r_minmax);
+        amrex::ParallelDescriptor::ReduceRealMin(thetax_min, amrex::ParallelDescriptor::IOProcessorNumber());
+        amrex::ParallelDescriptor::ReduceRealMax(thetax_max, amrex::ParallelDescriptor::IOProcessorNumber());
+
+        // x_ave, thetax_ave
+        amrex::Real x_ave = 0.0_rt;
+        amrex::Real thetax_ave = 0.0_rt;
+
+        amrex::ReduceOps<ReduceOpSum, ReduceOpSum> reduce_ops_ave;
+        auto r_ave = amrex::ParticleReduce<amrex::ReduceData<Real, Real>>(
```
dpgrote: Could these ReduceOps be combined to reduce the number of kernels and avoid the duplication of the calculation of theta?

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
dpgrote: I'll second this comment.

Regression/WarpX-tests.ini
aeriforme: Yes, I have added two low priority warnings for the `ColliderRelevant` diagnostic to let the user know that the colliding species should propagate along `z` and that the various `chi` values are computed on the coarsest mesh refinement level.

Examples/Tests/collider_relevant_diags/inputs_3d_multiple_particles
aeriforme: I think that the CI machine uses `diag1` as a default name to compare the plotfiles with the checksums, right @EZoni?

Docs/source/usage/parameters.rst
aeriforme: Oh right! I'll fix this. 
Anyway, what happens is that the various `chi` values are not dumped in the file.

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
aeriforme: If I understand correctly, the idea is to compute as many quantities as possible with a single pass over all the particles. 

In this case, the quantities that the diagnostic calculates are simple averages and standard deviations over all the particles of the relevant species (e.g. there are not cross-correlations). 

To avoid doing multiple passes, I understand that the idea would be to use the formula for the variance:
`Var(X) = <X^2> - <X>^2` 
so to get in a single pass `<X^2>` and `<X>^2` (`<*>` is the weighted average) and then the variance without having to loop again through the particles. 

Is this what you are suggesting?

Isn't this calculation prone to numerical cancellation? 

On a side note, should I also compute the minimum and maximum angles together with the averages?

dpgrote: Yes, it is generally better to have few kernel launches. Note that the same is true for the parallel operations. Those should also be combined as much as possible. See for example PR #4050.

That's a good question regarding numerical cancellation. I just did some quick tests comparing the two ways of calculating an rms. Using  `<X^2> - <X>^2` does break down when the average becomes much larger than the rms, around a ratio of 1.e6. So it is probably a good idea then to use the two loops, calculating the averages first, then the variances.

Examples/Tests/collider_relevant_diags/inputs_3d_multiple_particles
EZoni: Yes, I think the `diag1` name is hard-coded here for all tests:
https://github.com/ECP-WarpX/WarpX/blob/cf3b8a9258fa723a109b46c8154b836f10287d01/Regression/WarpX-tests.ini#L51-L53

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,560 @@
...
+                const amrex::Real uz = p.rdata(PIdx::uz);
+                const amrex::Real thetax = std::atan2(ux, uz);
+                const amrex::Real thetay = std::atan2(uy, uz);
+                return {w, w*x, w*x*x, w*y, w*y*y,
+                        thetax, w*thetax, w*thetax*thetax, thetax,
+                        thetay, w*thetay, w*thetay*thetay, thetay};
+            },
+            reduce_ops);
+
+        w_tot  = amrex::get<0>(r);
```
dpgrote: The code could be simplified by doing the declaration here and removing the ones above
```amrex::Real w_tot = amrex::get<0>(r);```

File: Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp [hunk 24]
```diff
@@ -0,0 +1,560 @@
...
+            thetay2_ave = 0.0_rt;
+        }
+
+        // x_std, y_std, thetax_std, thetay_std
+        amrex::Real x_std = 0.0_rt;
+        amrex::Real y_std = 0.0_rt;
+        amrex::Real thetax_std = 0.0_rt;
+        amrex::Real thetay_std = 0.0_rt;
+
+        x_std = std::sqrt(x2_ave - x_ave*x_ave);
```
dpgrote: I've changed my mind on this. The tests I did showed that doing it this way does break down if the average is much larger than the std. I now think doing two loops is better and more robust.

aeriforme: Oh ok, I haven't tested it yet in fact. 
@ax3l, what do you think?

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,572 @@
...
+            auto val = reduce_data.value();
+            chimin_f = get<0>(val);
+            chimax_f = get<1>(val);
+            chiave_f = get<2>(val);
+            amrex::ParallelDescriptor::ReduceRealMin(chimin_f, amrex::ParallelDescriptor::IOProcessorNumber());
+            amrex::ParallelDescriptor::ReduceRealMax(chimax_f, amrex::ParallelDescriptor::IOProcessorNumber());
+            amrex::ParallelDescriptor::ReduceRealSum(chiave_f, amrex::ParallelDescriptor::IOProcessorNumber());
+
+            m_data[get_idx("chimin_"+species_names[i_s])] = chimin_f;
+            m_data[get_idx("chiave_"+species_names[i_s])] = chiave_f/w_tot;
```
EZoni: This line produces the error `‘w_tot’ was not declared in this scope` when compiled in 1D. I wonder what we should do in 1D anyways. Do these diagnostics make sense? If not, we could simply skip them, as we currently do for RZ. Otherwise, we need to make sure that `w_tot` is defined also in 1D.

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,572 @@
...
+    // get species names (std::vector<std::string>)
+    const std::vector<std::string> species_names = mypc.GetSpeciesNames();
+
+    // get a reference to WarpX instance
+    auto& warpx = WarpX::GetInstance();
+
+    // get cell size
+    amrex::Geometry const & geom = warpx.Geom(0);
+#if defined(WARPX_DIM_1D_Z)
+        amrex::Real dV = geom.CellSize(0);
```
dpgrote: This can be simplified as
```
amrex::Real dV = AMREX_D_TERM(geom.CellSize(0), *geom.CellSize(1), *geom.CellSize(2));
```

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,572 @@
...
+        amrex::Real thetax_min  = amrex::get<2>(r);
+        amrex::Real thetax_ave  = amrex::get<3>(r);
+        amrex::Real thetax_max  = amrex::get<4>(r);
+
+        amrex::ParallelDescriptor::ReduceRealSum({w_tot, x_ave, thetax_ave},
+                                                  amrex::ParallelDescriptor::IOProcessorNumber());
+        amrex::ParallelDescriptor::ReduceRealMin(thetax_min, amrex::ParallelDescriptor::IOProcessorNumber());
+        amrex::ParallelDescriptor::ReduceRealMax(thetax_max, amrex::ParallelDescriptor::IOProcessorNumber());
+
+        if (w_tot > std::numeric_limits<Real>::min())
```
dpgrote: Is there a reason that this isn't the more simple `w_tot > 0.`?

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,572 @@
...
+    {
+        // get WarpXParticleContainer class object
+        WarpXParticleContainer& myspc = mypc.GetParticleContainerFromName(m_beam_name[i_s]);
+        // get charge
+        amrex::ParticleReal const q = myspc.getCharge();
+
+        using PType = typename WarpXParticleContainer::SuperParticleType;
+
+        num_dens[i_s] = myspc.GetChargeDensity(0);
+        num_dens[i_s]->mult(1./q);
```
dpgrote, line 225: This requires that `q` is not zero. Maybe add a check above similar to the photon check to make sure that the species are not neutral.

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,572 @@
...
+
+#if defined(WARPX_DIM_XZ)
```
dpgrote: I'm not sure if it would be better, but could these separate blocks of code be rewritten intertwined since there is a fair amount of code duplication with completely separate blocks for 2D and 3D?

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,572 @@
...
+        amrex::Real thetax_min  = amrex::get<3>(r);
+        amrex::Real thetax_ave  = amrex::get<4>(r);
+        amrex::Real thetax_max  = amrex::get<5>(r);
+        amrex::Real thetay_min  = amrex::get<6>(r);
+        amrex::Real thetay_ave  = amrex::get<7>(r);
+        amrex::Real thetay_max  = amrex::get<8>(r);
+
+        amrex::ParallelDescriptor::ReduceRealSum({w_tot, x_ave, y_ave,
+                                                  thetax_ave, thetay_ave},
+                                                  amrex::ParallelDescriptor::IOProcessorNumber());
```
dpgrote: I've seen this in some places where a temporary will be defined to shorten the code
```
const int IOProc = amrex::ParallelDescriptor::IOProcessorNumber();
```

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,572 @@
...
+        if (w_tot > std::numeric_limits<Real>::min())
+        {
+            x_ave = x_ave / w_tot;
+            y_ave = y_ave / w_tot;
+            thetax_ave = thetax_ave / w_tot;
+            thetay_ave = thetay_ave / w_tot;
+        }
+        else
+        {
+            x_ave = 0.0_rt;
```
dpgrote: Is this block of code needed? If `w_tot` is zero, it means that there are no particles so these values will have been set to zero already.

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,572 @@
...
+
+        amrex::ParallelDescriptor::ReduceRealSum({w_tot, x_ave, y_ave,
+                                                  thetax_ave, thetay_ave},
+                                                  amrex::ParallelDescriptor::IOProcessorNumber());
+        amrex::ParallelDescriptor::ReduceRealMin({thetax_min, thetay_min}, amrex::ParallelDescriptor::IOProcessorNumber());
+        amrex::ParallelDescriptor::ReduceRealMax({thetax_max, thetay_max}, amrex::ParallelDescriptor::IOProcessorNumber());
+
+        if (w_tot > std::numeric_limits<Real>::min())
+        {
+            x_ave = x_ave / w_tot;
```
dpgrote: Perhaps move this block of code into the `if` block below so there is only one check of `w_tot`. This also removes the need for the `else` block.

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,572 @@
...
+        m_data[get_idx("thetay_min_"+species_names[i_s])] = thetay_min;
+        m_data[get_idx("thetay_ave_"+species_names[i_s])] = thetay_ave;
+        m_data[get_idx("thetay_max_"+species_names[i_s])] = thetay_max;
+        m_data[get_idx("thetay_std_"+species_names[i_s])] = thetay_std;
+#endif
+
+#if (defined WARPX_QED)
+        // get mass
+        amrex::ParticleReal m = myspc.getMass();
+        const bool is_photon = myspc.AmIA<PhysicalSpecies::photon>();
```
dpgrote, line 420: `is_photon` will always be false here since the checks above disallow photon species.

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
dpgrote: Also, this should ensure the `w_tot` is not zero.

aeriforme: Some of the quantities do make sense in 1D: the derivative of luminosity in time and the `chi` values.
I guess we could go back to the previous version where `w_tot` is computed in a separate reduce operation before the other ones.

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
aeriforme: Not that I know of. I will simplify it.

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
aeriforme: @EZoni has been looking into this, and it seems that the code would become quite fragmented with a lot of `#if defined` statements.

dpgrote: Use of `AMREX_D_TERM` and `AMREX_D_DECL` might help.

EZoni: Hi Dave, I had tried considering those to reduce duplication, but somehow it still did not seem that feasible/readable overall.

The reason is that there are parts of the code where we have, for instance,
```cpp
auto r_std = amrex::ParticleReduce<amrex::ReduceData<Real, Real>>
```
in 2D, and
```cpp
auto r_std = amrex::ParticleReduce<amrex::ReduceData<Real, Real, Real, Real>>(
```
in 3D.

You see that here the reduce syntax needs 2 args in 2D and 4 args in 3D: I don't know how we could handle this in a dimension-agnostic way using `AMREX_D_TERM` or `AMREX_D_DECL`, which map only 1D to 1 arg, 2D to 2 args, 3D to 3 args. Do you see an easy solution for such cases?

EZoni: Hence I thought that we could investigate such a refactoring, which didn't seem trivial to me, possibly later in a separate PR, to avoid stalling this one for too long.

dpgrote: Doing the refactoring in a subsequent PR is fine.

I think you can do things like `AMREX_D_DECL(,Real,Real)`, so you can do this
```amrex::ParticleReduce<amrex::ReduceData<AMREX_D_DECL(,Real,Real),AMREX_D_DECL(,Real,Real)>```
I do agree that it starts to get messy.

File: Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp [hunk 34]
```diff
@@ -0,0 +1,563 @@
...
+        num_dens[i_s]->mult(1./q);
+
+#if defined(WARPX_DIM_1D_Z)
+        // w_tot
+        amrex::Real w_tot = ReduceSum( myspc,
+            [=] AMREX_GPU_HOST_DEVICE (const PType& p)
+            {
+                return p.rdata(PIdx::w);
+            });
+        amrex::ParallelDescriptor::ReduceRealSum(wtot, IOProc);
```
EZoni:
```suggestion
        amrex::ParallelDescriptor::ReduceRealSum(w_tot, IOProc);
```

Docs/source/usage/parameters.rst
```diff
@@ -2985,6 +2985,57 @@ Reduced Diagnostics
...
+        :math:`{\theta_y}_{min}`,
+        :math:`\langle \theta_y \rangle`,
+        :math:`{\theta_y}_{max}`,
+        :math:`\sqrt{\langle \theta_y- \langle \theta_y \rangle \rangle^2}`.
+
+        [18], ..., [32]: Analogous quantities for species 2.
+
+        For 2D-XZ, :math:`y`-related quantities are not outputted.
+        For 1D-Z, :math:`x`-related and :math:`y`-related quantities are not outputted.
+        RZ geomtry is not supported yet.
```
```suggestion
        RZ geometry is not supported yet.
```
small typo

Examples/Tests/collider_relevant_diags/analysis_multiple_particles.py
```diff
@@ -0,0 +1,144 @@
...
+    iterations = np.asarray(series.iterations)
+    lumi = []
+    for n,ts in enumerate(iterations):
+        it = series.iterations[ts]
+        rho1 = it.meshes["rho_beam_e"]
+        dV = np.prod(rho1.grid_spacing)
+        rho1 = it.meshes["rho_beam_e"][io.Mesh_Record_Component.SCALAR].load_chunk()
+        rho2 = it.meshes["rho_beam_p"][io.Mesh_Record_Component.SCALAR].load_chunk()
+        q1 = np.unique(it.particles["beam_e"]["charge"][io.Mesh_Record_Component.SCALAR].load_chunk())
+        q2 = np.unique(it.particles["beam_p"]["charge"][io.Mesh_Record_Component.SCALAR].load_chunk())
```
lucafedeli88: Are you using `np.unique` here because all the electrons (positrons) have the same charge? If it is the case I would suggest  checking this explicitly.

Regression/WarpX-tests.ini
lucafedeli88, line 215:
```suggestion
runtime_params = warpx.abort_on_warning_threshold=medium
```
If the test is expected to generate only `low` priority warnings I would suggest setting this threshold to `medium`

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,555 @@
...
+        "Collider-relevant diagnostics must involve exactly two species");
+
+    // RZ coordinate is not supported
+#if (defined WARPX_DIM_RZ)
+    WARPX_ABORT_WITH_MESSAGE(
+        "Collider-relevant diagnostics do not work in RZ geometry.");
+#endif
+
+    ablastr::warn_manager::WMRecordWarning(
+        "DIAGNOSTICS",
```
lucafedeli88:
```suggestion
        "Diagnostics",
```
I would suggest to change this to lowercase for consistency with other warnings in the code.

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,555 @@
...
+        "The collider-relevant reduced diagnostic is meant for \
+        colliding species propagating along the z direction.",
+        ablastr::warn_manager::WarnPriority::low);
+
+    ablastr::warn_manager::WMRecordWarning(
+        "DIAGNOSTICS",
```
lucafedeli88:
```suggestion
        "Diagnostics",
```
Same as above

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,555 @@
...
+        add_diag("thetay_max_"+species_names[i_s], "theta_y_max_"+species_names[i_s]+"(rad)");
+        add_diag("thetay_std_"+species_names[i_s], "theta_y_std_"+species_names[i_s]+"(rad)");
+
+#elif (defined WARPX_DIM_XZ)
+        add_diag("x_ave_"+species_names[i_s], "x_ave_"+species_names[i_s]+"(m)");
+        add_diag("x_std_"+species_names[i_s], "x_std_"+species_names[i_s]+"(m)");
+        add_diag("thetax_min_"+species_names[i_s], "theta_x_min_"+species_names[i_s]+"(rad)");
+        add_diag("thetax_ave_"+species_names[i_s], "theta_x_ave_"+species_names[i_s]+"(rad)");
+        add_diag("thetax_max_"+species_names[i_s], "theta_x_max_"+species_names[i_s]+"(rad)");
+        add_diag("thetax_std_"+species_names[i_s], "theta_x_std_"+species_names[i_s]+"(rad)");
```
lucafedeli88: Should we really use `species_names` in this section? A simulation may contain several species and the first two species may not be the two beams we are interested in.

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,555 @@
...
+            x_std = std::sqrt(x_std / w_tot);
+            thetax_std = std::sqrt(thetax_std / w_tot);
+        }
+
+        m_data[get_idx("x_ave_"+species_names[i_s])] = x_ave;
+        m_data[get_idx("x_std_"+species_names[i_s])] = x_std;
+        m_data[get_idx("thetax_min_"+species_names[i_s])] = thetax_min;
+        m_data[get_idx("thetax_ave_"+species_names[i_s])] = thetax_ave;
+        m_data[get_idx("thetax_max_"+species_names[i_s])] = thetax_max;
+        m_data[get_idx("thetax_std_"+species_names[i_s])] = thetax_std;
```
lucafedeli88: same as above for the usage of `species_names`

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,555 @@
...
+        m_data[get_idx("y_ave_"+species_names[i_s])] = y_ave;
+        m_data[get_idx("y_std_"+species_names[i_s])] = y_std;
+        m_data[get_idx("thetax_min_"+species_names[i_s])] = thetax_min;
+        m_data[get_idx("thetax_ave_"+species_names[i_s])] = thetax_ave;
+        m_data[get_idx("thetax_max_"+species_names[i_s])] = thetax_max;
+        m_data[get_idx("thetax_std_"+species_names[i_s])] = thetax_std;
+        m_data[get_idx("thetay_min_"+species_names[i_s])] = thetay_min;
+        m_data[get_idx("thetay_ave_"+species_names[i_s])] = thetay_ave;
+        m_data[get_idx("thetay_max_"+species_names[i_s])] = thetay_max;
+        m_data[get_idx("thetay_std_"+species_names[i_s])] = thetay_std;
```
lucafedeli88: same as above for the usage of `species_names`

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
```diff
@@ -0,0 +1,555 @@
...
+        if (myspc.DoQED())
+        {
+            // define variables in preparation for field gatheeduce_data.value()ring
+            const int n_rz_azimuthal_modes = WarpX::n_rz_azimuthal_modes;
+            const int nox = WarpX::nox;
+            const bool galerkin_interpolation = WarpX::galerkin_interpolation;
+            const amrex::IntVect ngEB = warpx.getngEB();
+
+            // TO DO: loop over refinement levels: for (int lev = 0; lev <= level_number; ++lev)
+            int lev = 0;
```
lucafedeli88: Shouldn't this be a `const int` for the moment?

lucafedeli88 (changes requested):
Thanks again for this very useful PR, @aeriforme ! I've left few new comments, but I think that it will be possible to merge it very soon.

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
EZoni: Thanks, done in [f4800d8](https://github.com/ECP-WarpX/WarpX/pull/4024/commits/f4800d862293e09777d30ebc1f675be886223ee5).

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
EZoni: Thanks, done in [f4800d8](https://github.com/ECP-WarpX/WarpX/pull/4024/commits/f4800d862293e09777d30ebc1f675be886223ee5).

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
EZoni: Thanks, done in [f4800d8](https://github.com/ECP-WarpX/WarpX/pull/4024/commits/f4800d862293e09777d30ebc1f675be886223ee5).

Regression/WarpX-tests.ini
EZoni: Here we have to keep `high` because this other warning, unrelated to this PR, gets triggered:
```
[medium][QED] The built-in Quantum Synchrotron table will be used.
              This low resolution table is intended for testing purposes only.
```

Docs/source/usage/parameters.rst
EZoni: Thanks, done in [f4800d8](https://github.com/ECP-WarpX/WarpX/pull/4024/commits/f4800d862293e09777d30ebc1f675be886223ee5).

EZoni:
@lucafedeli88 Note that in [f4800d8](https://github.com/ECP-WarpX/WarpX/pull/4024/commits/f4800d862293e09777d30ebc1f675be886223ee5) I had to remove one instance of `#include <memory>`, it seems like it was added twice.

Regression/WarpX-tests.ini
```diff
@@ -209,6 +209,22 @@ compileTest = 0
...
+dim = 3
+addToCompileString =
+cmakeSetupOpts = -DWarpX_DIMS=3 -DCMAKE_BUILD_TYPE=Debug
```
EZoni: `Debug` compilation mode was being used for local debugging purposes only, should not slip through:
```suggestion
cmakeSetupOpts = -DWarpX_DIMS=3
```

Examples/Tests/collider_relevant_diags/analysis_multiple_particles.py
EZoni: Thanks @lucafedeli88, is commit [f207759](https://github.com/ECP-WarpX/WarpX/pull/4024/commits/f207759e082902dac199a331a0a9323d9fb844b2) accomplishing what you had in mind?

Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp
EZoni: Thanks @lucafedeli88, I think you're right. Now we're using `m_beam_name` instead, which was set by
```cpp
    // read colliding species names - must be 2
    amrex::ParmParse pp_rd_name(m_rd_name);
    pp_rd_name.getarr("species", m_beam_name);
```
and contains correctly only the collider-relevant beam species names.

See [e5d815d](https://github.com/ECP-WarpX/WarpX/pull/4024/commits/e5d815d184aa3238fcddb0cb3d94bbf3066ce64b) for reference. I also switched the order of the particle species in the input file (we start with photons now), to show that this indeed works.

lucafedeli88 (approved):
All my comments have been addressed. I think that we can merge this PR now!

EZoni (approved):

WeiqunZhang (approved):

PR #4026, opened 2023-06-22
Move MPIInitHelpers into ablastr

lucafedeli88:
This PR moves moves `MPIInitHelpers.cpp` and `MPIInitHelpers.H` into `ablastr`, since they are sufficiently general to be shared with other codes


ax3l (approved):
Thank you! :tada:

PR #4028, opened 2023-06-22
Clean up GpuArray init in ABLASTR interp modules

EZoni:
As suggested by @RemiLehe, there should be a simpler way to initalize the `GpuArray` data that we use to store index types and coarsening ratios in the interpolation modules available in ABLASTR (sampling and averaging), similarly to what is done in Source/Diagnostics/ReducedDiags/FieldMaximum.cpp.


RemiLehe (approved):
Great! Thanks a lot for simplifying this code.

PR #4037, opened 2023-06-26
Remove Custom momentum and density injector

RemiLehe:
The "Custom" momentum and density injectors where introduced at the beginning of the development of WarpX, and required the user to modify the C++ files `CustomDensityProb.H`/`CustomMomentumProb.H` and recompile the code.

This was a bit of a klunky workflow, and it essentially stopped being used after we introduced injectors based on the `amrex` parser. For this reason, it is not currently documented, nor tested. 

I would recommend removing these files to cleanup the code. If we encounter a case that could not be handled with the parser, my impression is that we would now recommend to use Python callback functions.


lucafedeli88 (approved):
I agree that this is a good idea. No need to support this feature in C++ since it can be done with python (and the AMReX parser is already very general and adequate for most custom targets).

PR #4040, opened 2023-06-26
Define new `InjectorFlux` object, and use it in `AddPlasmaFlux`

RemiLehe:
In the current version of WarpX, the `AddPlasmaFlux` function uses the `InjectorDensity` object (which was originally meant for `AddPlasma`) to determine the flux to be emitted at each timestep. This has a few drawbacks:

- The user has to specify `<species>.density = ...` to set the value of the **flux**, in the input script, which is confusing. (e.g. a density is typically in `m^-3`, while the flux that the user actually needs to enter here is in `m^-2.s^-1`
- It was not possible to define a flux that depends on time e.g. with a parser expression, since the `DensityInjector` reads `<species>.density_function(x,y,z)` which cannot depend on time.

Instead, in this PR, I added a new `InjectorFlux` object - which is very similar to the `InjectorDensity` and `InjectorMomentum` objects. With this new object, the syntax in the input script is:
- `<species>.flux`  (instead of `<species>.density`) for a constant flux
- `<species>.flux_function(x,y,z,t)` (instead of `<species>.density_function(x,y,z)`) for a parsed expression


Examples/Tests/flux_injection/inputs_3d
```diff
@@ -38,8 +38,8 @@ electron.num_particles_per_cell = 100
 electron.surface_flux_pos = -4.
 electron.flux_normal_axis = y
 electron.flux_direction = +1
-electron.profile = constant
-electron.density = 1.
+electron.flux_profile = parse_flux_function
+electron.flux_function(x,y,z,t) = "1."
```
RemiLehe, line 42: Even though the flux is constant in this example, I did use the `parse_flux_function` here just to test the corresponding code path.

Source/Initialization/InjectorFlux.H
```diff
@@ -0,0 +1,136 @@
+/* Copyright 2023 Remi Lehe
+ *
+ * This file is part of WarpX.
+ *
+ * License: BSD-3-Clause-LBNL
+ */
+#ifndef INJECTOR_FLUX_H_
+#define INJECTOR_FLUX_H_
```
RemiLehe, line 8: This whole file closely mimics `InjectorDensity.H`

Source/Initialization/PlasmaInjector.cpp
```diff
@@ -342,7 +342,7 @@ PlasmaInjector::PlasmaInjector (int ispecies, const std::string& name)
             (InjectorPositionRandomPlane*)nullptr,
             xmin, xmax, ymin, ymax, zmin, zmax,
             flux_normal_axis);
-        parseDensity(pp_species_name);
+        parseFlux(pp_species_name);
```
RemiLehe, line 345: This part of the code is inside `if (injection_style == "nfluxpercell") {`.

Source/Initialization/PlasmaInjector.cpp
```diff
@@ -572,6 +582,33 @@ void PlasmaInjector::parseDensity (const amrex::ParmParse& pp)
     }
 }
 
+// Depending on injection type at runtime, initialize inj_flux
+// so that inj_flux->getFlux calls
+// InjectorFlux[Constant or Parser or etc.].getFlux.
+void PlasmaInjector::parseFlux (const amrex::ParmParse& pp)
```
RemiLehe, line 588: Again, this closely mimics `parseDensity`.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1821,14 +1819,12 @@ PhysicalParticleContainer::AddPlasmaFlux (amrex::Real dt)
                     pu.y = sin_theta*ur + cos_theta*ut;
                 }
 #endif
-                Real dens = inj_rho->getDensity(ppos.x, ppos.y, ppos.z);
-                // Remove particle if density below threshold
-                if ( dens < density_min ){
+                Real flux = inj_flux->getFlux(ppos.x, ppos.y, ppos.z, t);
```
RemiLehe, line 1822: The time `t` is now taken into account.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1475,10 +1475,8 @@ PhysicalParticleContainer::AddPlasmaFlux (amrex::Real dt)
     }
 
     InjectorPosition* inj_pos = plasma_injector->getInjectorPosition();
-    InjectorDensity*  inj_rho = plasma_injector->getInjectorDensity();
+    InjectorFlux*  inj_flux = plasma_injector->getInjectorFlux();
     InjectorMomentum* inj_mom = plasma_injector->getInjectorMomentum();
-    const amrex::Real density_min = plasma_injector->density_min;
-    const amrex::Real density_max = plasma_injector->density_max;
```
RemiLehe, line 1481: I removed `density_min` and `density_max` here (i.e. in `AddPlasmaFlux`), since I think that the name of this variable is confusing for the case of flux injection, and since I don't think it is used in practice flux-injected particles.

`density_min` and `density_max` are still used in `AddPlasma`.

dpgrote (approved):
These changes look good to me.

PR #4041, opened 2023-06-26
Weibel instability test case

juliettepech:
## Original description 

A new test case highlighting the Weibel instability with two counter-current electron beams. The test implemented in the code is for a temperature of 1.10⁻² eV. Here is the animation obtained for a temperature of 0eV : 

Two_streams_0eV


EZoni:
@aeriforme @JustinRayAngus 
I found this while looking at the list of PRs that are not labeled as "WIP" and it looks to me like this could be good material for the examples and/or tutorials in our documentation.

EZoni:
I had promised myself to review and work on this PR, since it could be good material for examples and/or tutorials, so I pushed a few commits with updates to the test case (different resolution and time scale) and its analysis.

Here are examples of the figures and animation that the new analysis produces:

<p align="center">
weibel_magnetic_energy_growth
</p>

<p align="center">
weibel_rho_electrons_1
</p>

<p align="center">
weibel_Bz
</p>

Regression/requirements.txt
```diff
@@ -8,5 +8,6 @@ openpmd-api
 openpmd-viewer
 pandas
 periodictable
+scikit-learn
```
EZoni: Remove this obsolete dependency:

```suggestion
```

EZoni (changes requested):

ax3l (dismissed):
This is great, thank you for finishing it @EZoni!

Let's make it an example in a follow-up?

aeriforme:
We could also update the related tutorial!

EZoni:
Just merged `development` here to have the latest CI workflows run properly.

PR #4042, opened 2023-06-26
Add simple string input parameter parsing

dpgrote:
This adds the capability of simple parsing of string input parameters, using the f-string like syntax from Python. In string parameters, when the names from `my_constants` parameters are used in curly braces, it is replaced by the value. For example with 
```
my_constants.run_name = case22
diag1.file_prefix = diags_{run_name}/diag
```
The value of `file_prefix` will be read in as "diags_case22/diag".

For all string parameter input, replaced `query` with `queryWithParser`, `get` with `getWithParser`, and the same for `queryarr` and `getarr`.

The changes here required other changes:
- In `Source/Utils/Parser/ParserUtils.H`, updated the parameter names from `char *` to `std::string_view`, matching what is now done in `AMReX_ParmParse.H`.
- Removed the now unnecessary use of `c_str` in calls to the `queryWithParser` routines.
- For consistency, removed a few other unnecessary uses of `c_str`.

Side notes:
- There are a number of places where expressions are read in using `query`. These were not updated. Should these be using `Store_parserString` which allows reading strings with spaces?
- In general, in some places `queryWithParser` is used, and others `query`. The latter presumably allows expression because AMReX now handles it. But the use is inconsistent - should WarpX use a single standard?
- On that note, perhaps the string parameter parsing implemented here should go in AMReX?


AlexanderSinn:
Based on this PR, I made a PR for the same feature in HiPACE++ (https://github.com/Hi-PACE/hipace/pull/990), however instead of copying the string from my_constants directly, the expression inside `{}` is evaluated as a double using the parser (that has access to my_constants). Additionally I made `{{` and `}}` escape sequences.

dpgrote:
@AlexanderSinn I implemented it intentionally as a direct string replacement (rather than a parsing and evaluation). I was thinking of cases where perhaps a scan is being done over an input parameter and having the diagnostic file names include that parameter. Based on your example, I would rather have the output names be like "test_sin(2)" than "test_0.9092974", the former being more human readable. I think a string replacement would offer ore flexibility, since the `my_constant` parameter value could be a string it self. I could add the "{{" escape, but using curly braces in file names is very rare.

AlexanderSinn:
Good point! Now I made it so that only if it isn’t found in `my_constants` it will be parsed to double otherwise it will be read as a string.

AlexanderSinn:
@dpgrote @ax3l 
After using this in hipace for quite a while, I now miss this feature in WarpX. I think this would be quite usefull to add, even just for simple string concatinations to to construct file paths.
Specifically I want to do
```powershell
# Input script
atoms_H.read_density_from_path = "{base_path}/density_H.h5"
atoms_Ar.read_density_from_path = "{base_path}/density_Ar.h5"
elec.read_density_from_path = "{base_path}/density_e.h5"

# Add as a command line argument
my_constants.base_path = "/some/path"
```

ax3l:
Huh, very cool.

Can we make this work with our `toml` syntax? Probably works there , too if we keep the rhs a string including the f-string placeholders.

PR #4023, opened 2023-06-21
Doc: Summit Scipy Install

RTSandberg:
Some aspects of the automated WarpX build on Summit fail non-deterministically.  This PR adds some fixes to assist the build.  The scipy version was too modern for Summit's power pc and was downgraded.  Also, some small typos in directory names were fixed

PR #4045, opened 2023-06-27
Doc: Summit SciPy -j10

ax3l:
We can use the latest SciPy if we avoid over-using the provided resources on the head-node. Compile with `-j 10` instead of `-j` now.

Follow-up to #4023 

X-ref: https://github.com/scipy/scipy/issues/18766


RTSandberg (approved):
I like this more than being pinned to a specific version of scipy.  I checked and it works reliably for me!

PR #4048, opened 2023-06-27
Use default `nsys` on Perlmutter

RemiLehe:
The default on `nsys` on Perlmutter is now 
```
/global/common/software/nersc/pm-2021q4/easybuild/software/Nsight-Systems/2022.2.1/bin/nsys
```

As far as I can tell, it works properly and has the right annotations. I think that we can therefore remove the warnings.


ax3l (approved):
That's great, thanks!

PR #4047, opened 2023-06-27
Clang-tidy CI test: add few more bugprone checks in clang-tidy CI 

lucafedeli88:
This PR adds the following checks to clang-tidy CI test:
- [bugprone-copy-constructor-init](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/bugprone-copy-constructor-init.html)
- [bugprone-dangling-handle](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/bugprone-dangling-handle.html)
- [bugprone-dynamic-static-initializers](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/bugprone-dynamic-static-initializers.html)

PR #4049, opened 2023-06-27
[WIP] Clang tidy CI test: add bugprone_easily_swappable_parameters check

lucafedeli88:
This is an attempt to add this check to clang-tidy CI test:
- [bugprone-easily-swappable-parameters](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/easily-swappable-parameters.html)

My guess is that there will be too many issues of this kind to fix them, but we'll see

**Warning** : depends on https://github.com/ECP-WarpX/WarpX/pull/4047


lucafedeli88:
too many issues... we have to discuss if we want to seriously implement this check

PR #4051, opened 2023-06-27
Merge MPI Reduction Calls

WeiqunZhang:


Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -1223,9 +1223,7 @@ std::array<ParticleReal, 3> WarpXParticleContainer::meanParticleVelocity(bool lo
     }
 
     if (local == false) {
-        ParallelDescriptor::ReduceRealSum(vx_total);
-        ParallelDescriptor::ReduceRealSum(vy_total);
-        ParallelDescriptor::ReduceRealSum(vz_total);
+        ParallelDescriptor::ReduceRealSum({vx_total,vy_total,vz_total});
```
EZoni: There seems to be a build error (e.g., CI test `background_mcc_dp_psp`) when we build for geometries different than 3D. Should this be replaced  by something along the following lines, instead?
```cpp
ParallelDescriptor::ReduceRealSum({AMREX_D_DECL(vx_total,vy_total,vz_total)});
```
In this case the labels `x`, `y` and `z` might be inconsistent, though.

Also, I don't understand why the build fails only when `WarpX_PARTICLE_PRECISION=SINGLE`.

WeiqunZhang: I don't think we should use `AMREX_D_DECL`. There are always 3 velocities. The build issue can be fixed by explicitly informing ReduceRealSum the type because it has trouble deducing it in this case.

lucafedeli88 (approved):
Thanks for this PR! It looks good to me!

PR #4057, opened 2023-06-28
Allow `FieldProbe` in RZ geometry

### LINKED ISSUE #4031: Field Probes in RZ
It is often unwieldly to output all of the field data from a 2D simulation. It would therefore be very useful to have field probe functionality for the RZ geometry. 

(Linking @roelof-groenewald because we discussed this previously)

roelof-groenewald:
Currently the `FieldProbe` reduced diagnostic is not supported in RZ. This PR adds support for it. The one tricky thing is that `FieldProbePIdx != PIdx` which causes a problem with getting the probe positions in `GetParticlePosition()` which uses `m_theta = soa.GetRealData(PIdx::theta).dataPtr() + a_offset;`. So we need to ensure `PIdx::theta == FieldProbePIdx::theta`.

Fixes #4031.

### DIFF
--- a/Source/Diagnostics/ReducedDiags/FieldProbeParticleContainer.H
+++ b/Source/Diagnostics/ReducedDiags/FieldProbeParticleContainer.H
@@ -19,13 +19,20 @@
  * This enumerated struct is used to index the field probe particle
  * values that are being stored as SoA data. Nattribs
  * is enumerated to give the number of attributes stored.
+ * The strange insertion of `theta` below is due to the use of
+ * GetParticlePosition for the field probe locations, which reads the probe
+ * theta value from PIdx::theta = 4.
  */
 struct FieldProbePIdx
 {
     enum
     {
         Ex = 0, Ey, Ez,
-        Bx, By, Bz,
+        Bx,
+#ifdef WARPX_DIM_RZ
+        theta,      ///< RZ needs all three position components
+#endif
+        By, Bz,
         S, //!< the Poynting vector
         nattribs
     };
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -2659,7 +2659,8 @@ Reduced Diagnostics
         to file. The electromagnetic field components are interpolated to the measurement point
         by default, but can they be saved as non-averaged by setting
         ``<reduced_diags_name>.raw_fields = true``, in which case the raw fields for the cell
-        containing the measurement point are saved.
+        containing the measurement point are saved. In RZ geometry, this only saves the
+        0'th azimuthal mode component of the fields.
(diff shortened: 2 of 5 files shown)


Source/Diagnostics/ReducedDiags/FieldProbeParticleContainer.H
```diff
@@ -19,13 +19,20 @@
  * This enumerated struct is used to index the field probe particle
  * values that are being stored as SoA data. Nattribs
  * is enumerated to give the number of attributes stored.
+ * The strange insertion of `theta` below is due to the use of
```
dpgrote, line 22: This is fine for now, though a bit kludgy. Perhaps a fix is to add an optional argument to `GetParticlePosition` that specifies the index for theta (defaulting to `PIdx::theta`). That would be messy though since it would have to be inside of `ifdef WARPX_DIM_RZ` macros.

dpgrote (approved):
This looks good, thanks!

As an aside, though, what the code is doing for `raw_fields` doesn't make much sense for the line and plane probes. Unless I missed something, when `raw_fields` is true, it is only fetching the field at the single coordinate `x, y, z_probe` and writing that out repeatedly for each probe location. Perhaps an easy fix is to instead call `doGatherShapeN` with interp_order = 0, to do nearest grid point interpolation. In fact, the `raw_fields` option is not really needed since the interpolation order can be specified. Stuff for a later PR.

roelof-groenewald:
Thanks for the review @dpgrote!

I agree with your comment about the `PIdx::theta` handling being kludgy. A robust solution might be to have the particle iterator have access to an attribute of the parent particle container that will tell `GetParticlePosition` what index holds the `theta` value.

Source/Diagnostics/ReducedDiags/FieldProbeParticleContainer.H
```diff
@@ -19,13 +19,20 @@
...
 {
     enum
     {
         Ex = 0, Ey, Ez,
-        Bx, By, Bz,
+        Bx,
+#ifdef WARPX_DIM_RZ
+        theta,      ///< RZ needs all three position components
+#endif
+        By, Bz,
```
ax3l, line 35: @roelof-groenewald this order or arguments in the enum is so... confusing :sweat_smile:

ax3l: Generalization posted in #4386

ax3l: @dpgrote @roelof-groenewald the implementation to assume an enum order across different types, and ensuring it via comment instead of static assert, reminds me of this Simpsons sketch:
https://www.youtube.com/watch?v=QRVExJZKIT8

PR #4059, opened 2023-06-29
RigidInjection: properly transform zinject to boosted frame

RemiLehe:
When the simulation starts with `t_boost` different than 0, the injection position needs to be transformed differently.


lucafedeli88 (approved):
Thanks for this PR, @RemiLehe

PR #4061, opened 2023-06-29
Doc: Update openPMD tutorials link

IlianCS:
Changed the link for the openPMD tutorials because the previous one is broken.


EZoni (approved):
Thanks for the fix, @IlianCS!

PR #4071, opened 2023-07-05
Implicit electromagnetic solver using Picard iterations

dpgrote:
This PR introduces the implicit EM PIC solver that uses an energy conserving scheme. It uses a simple Picard type iteration to solve the system. This method is for demonstration only to show that the scheme works. It is not very efficient and does not work well with $\omega_{pe} \Delta t$ close to or greater than 1.

As a straight-forward extension, the semi-implicit Picard scheme is also added. This is an approximately energy conserving scheme but still has the explicit $\omega_{pe} \Delta t < 1$  constraint.

Note that this PR includes some fixes for the RZ code, correcting the handling of the fields near the axis. These fixes were needed to obtained exact energy convergence in RZ. Because of this fix, the results of some of the RZ CI tests needed to be updated. The changes were all small, in the 3rd to 5th decimal place.

## Tests

- [x] energy conservation test


RemiLehe:
Thanks for this PR @dpgrote. I will start taking a look now.
Btw, could you explain a bit more what you meant by the fact that the current implementation of the fully implicit scheme "does not work well with w_pe*dt close to or greater than 1." Did you mean that the performance is not very good (but results are correct), or that one gets incorrect results (e.g. energy conservation is not exact), or even that you get code crashes (e.g. segault) in this case?

dpgrote:
> Thanks for this PR @dpgrote. I will start taking a look now. Btw, could you explain a bit more what you meant by the fact that the current implementation of the fully implicit scheme "does not work well with w_pe*dt close to or greater than 1." Did you mean that the performance is not very good (but results are correct), or that one gets incorrect results (e.g. energy conservation is not exact), or even that you get code crashes (e.g. segault) in this case?

This is reporting Justin's experience, that with larger dt more iterations are required to reach convergence as the system becomes more stiff. Running tests on a sample 1D uniform plasma test case, if w_pe*dt = 0.1 requires ~9 iterations to converge to 1.e-7, but w_pe*dt = 0.24 requires ~168 iterations, and at 0.25 it fails to converge at all.

Source/WarpX.H
```diff
@@ -1435,6 +1476,12 @@ private:
     amrex::Vector<std::array< std::unique_ptr<amrex::MultiFab>, 3 > > Efield_avg_fp;
     amrex::Vector<std::array< std::unique_ptr<amrex::MultiFab>, 3 > > Bfield_avg_fp;
 
+    // Implicit, fields at start of step and from the previous iteration
```
RemiLehe, line 1474: Could you add more comments to indicate which field is which?
(e.g. is `Efield_n` the one at the start of the step, or the one from the previous iteration)

Source/WarpX.cpp
```diff
@@ -2008,7 +2014,9 @@ WarpX::ClearLevel (int lev)
 void
 WarpX::AllocLevelData (int lev, const BoxArray& ba, const DistributionMapping& dm)
 {
-    const bool aux_is_nodal = (field_gathering_algo == GatheringAlgo::MomentumConserving);
+    const bool aux_is_nodal = ((field_gathering_algo == GatheringAlgo::MomentumConserving) &&
```
RemiLehe: I am not sure I understand why this is needed ; could you explain?
In general, wouldn't we want to use the energy-conserving gather instead of the momentum-conserving gather, when using the implicit or semi-implicit method?

Docs/source/usage/parameters.rst
```diff
@@ -77,6 +77,20 @@ Overall simulation parameters
...
 
+* ``algo.evolve_scheme`` (`string`) optional (default `explicit`)
+    Specifies the evolve scheme used by WarpX.
+
+    * ``explicit``: Use an explicit solver, such as the standard FDTD or PSATD
+
+    * ``implicit_picard``: Use an energy conserving implicit solver that uses a Picard iteration to solve the system.
+      Note that this method is for demonstration only. It is inefficient and does not work well when
+      :math:`\omega_{pe} \delta t` is close to or greater than one.
+      The method is described in `Angus et al., On numerical energy conservation for an implicit particle-in-cell method coupled with a binary Monte-Carlo algorithm for Coulomb collisions <https://doi.org/10.1016/j.jcp.2022.111030>`__.
```
RemiLehe, line 92: My understanding is that the WarpX implementation is actually a relativistic generalization (while the above paper is written in a non-relativistic form). Could you make this clear in this sentence?

Source/WarpX.H
```diff
@@ -436,6 +451,25 @@ public:
...
+     * and optionally initialize it. This also adds the MultiFab
+     * to the map of MultiFabs (used to ease the access to MultiFabs from the Python
+     * interface
+     *
+     * \param mf[out] The MultiFab unique pointer to be allocated
+     * \param mf_model[in] The MultiFab to model
+     * \param name[in] The name of the MultiFab to use in the map
+     * \param initial_value[in] The optional initial value
+     */
+    static void AllocInitMultiFabFromModel (
```
RemiLehe, line 450: Would it be worth to introduce this in a separate PR?
(In order to make the current PR more focused.)

Docs/source/usage/parameters.rst
```diff
@@ -77,6 +77,20 @@ Overall simulation parameters
...
+
+    * ``semi_implicit_picard``: Use an energy conserving semi-implicit solver that uses a Picard iteration to solve the system.
+      Note that this method has the CFL limitation :math:`\omega_{pe} \delta t < 1`.
+      The method is described in `Chen et al., A semi-implicit, energy- and charge-conserving particle-in-cell algorithm for the relativistic Vlasov-Maxwell equations <https://doi.org/10.1016/j.jcp.2020.109228>`__.
+
```
RemiLehe, line 120: Could you also add documentation for:
- `max_picard_iterations`
- `picard_iteration_tolerance`

Source/WarpX.cpp
```diff
@@ -1138,6 +1141,9 @@ WarpX::ReadParameters ()
         current_deposition_algo = static_cast<short>(GetAlgorithmInteger(pp_algo, "current_deposition"));
         charge_deposition_algo = static_cast<short>(GetAlgorithmInteger(pp_algo, "charge_deposition"));
         particle_pusher_algo = static_cast<short>(GetAlgorithmInteger(pp_algo, "particle_pusher"));
+        evolve_scheme = static_cast<short>(GetAlgorithmInteger(pp_algo, "evolve_scheme"));
+        utils::parser::queryWithParser(pp_algo, "max_picard_iterations", max_picard_iterations);
+        utils::parser::queryWithParser(pp_algo, "picard_iteration_tolerance", picard_iteration_tolerance);
```
RemiLehe: Should we raise an error if the `evolve_scheme` is not explicit, but the user chooses anything else than the electromagnetic Yee solver (e.g. electrostatic solver, or PSATD solver), since it seems that the code would in this case ignore the type of solver that was chosen.

Source/Evolve/WarpXEvolveImplicitPicard.cpp
```diff
@@ -0,0 +1,580 @@
...
+    // Only one refinement level is supported
+    const int nlevs_max = maxLevel() + 1;
+    Efield_n.resize(nlevs_max);
+    Efield_save.resize(nlevs_max);
+    if (evolve_scheme == EvolveScheme::ImplicitPicard) {
+        Bfield_n.resize(nlevs_max);
+        Bfield_save.resize(nlevs_max);
+    }
+
+    // Strange, the WarpX::DistributionMap(0) is not consistent with Ex_fp.DistributionMap()???
```
RemiLehe: Is this comment still up to date?

Source/WarpX.cpp
dpgrote: In the explicit code, with the momentum conserving scheme, it does a double interpolation, from the Yee mesh to a node centered mesh and then to the particles. But for implicit, the deposition and gather shape functions need to be the same in order to conserve energy - this double interpolation breaks the symmetry. So for momentum conserving, that double interpolation is turned off. The real difference here though is charge conservation. With the momentum conserving scheme, it is using direct deposition and so charge is not conserved. This is really only to test the mover and the momentum conserving scheme would probably never be used because of the lack of charge conservation.

Docs/source/usage/parameters.rst
dpgrote: A comment was added to this effect.

Source/WarpX.H
dpgrote: Comments have been added.

Docs/source/usage/parameters.rst
dpgrote: Done

Source/WarpX.cpp
dpgrote: I've added a number of checks.

Source/Evolve/WarpXEvolveImplicitPicard.cpp
dpgrote: The comment would still be true but is not needed here. I deleted it.

Source/WarpX.H
dpgrote: This routine is only used in the implicit solver so would be an unused routine in a separate PR.

This routine may be useful elsewhere if other MultiFabs are created in a similar way, but I didn't look through the source for any such places.

Docs/source/usage/parameters.rst
```diff
@@ -81,6 +81,38 @@ Overall simulation parameters
...
     even if a fixed ``warpx.random_seed`` is provided.
 
+* ``algo.evolve_scheme`` (`string`, default: `explicit`)
+    Specifies the evolve scheme used by WarpX.
+
+    * ``explicit``: Use an explicit solver, such as the standard FDTD or PSATD
+
+    * ``implicit_picard``: Use an energy conserving implicit solver that uses a Picard iteration to solve the system.
+      Note that this method is for demonstration only. It is inefficient and does not work well when
+      :math:`\omega_{pe} \delta t` is close to or greater than one.
```
```suggestion
      :math:`\omega_{pe} \Delta t` is close to or greater than one.
```
I think that we use the notation `\Delta t` rather than `\delta t` in the rest of the documentation.

Docs/source/usage/parameters.rst
```diff
@@ -81,6 +81,38 @@ Overall simulation parameters
...
+      The method is described in `Angus et al., On numerical energy conservation for an implicit particle-in-cell method coupled with a binary Monte-Carlo algorithm for Coulomb collisions <https://doi.org/10.1016/j.jcp.2022.111030>`__.
+      The version implemented is an updated version that is relativistically correct, including the relativistic gamma factor for the particles.
+
+    * ``semi_implicit_picard``: Use an energy conserving semi-implicit solver that uses a Picard iteration to solve the system.
+      Note that this method has the CFL limitation :math:`\omega_{pe} \delta t < 1`.
```
RemiLehe: My understanding is that this solver has the standard CFL limit for the Yee solver, i.e. 
`\Delta t < c/\sqrt( \sum_i 1/\Delta x_i^2 )` to ensure numerical stability. 

Other hand, I think that the solver would still be numerically stable for `\omega_{pe} \Delta t < 1` (provided that the Picard iteration does converge).

If you agree with the above, could you replace 
```suggestion
      Note that this method has the CFL limitation :math:`\Delta t < c/\sqrt( \sum_i 1/\Delta x_i^2 )`. It is inefficient and does not work well when :math:`\omega_{pe} \Delta t` is close to or greater than one.
```

Docs/source/usage/parameters.rst
```diff
@@ -81,6 +81,38 @@ Overall simulation parameters
...
+      The method is described in `Chen et al., A semi-implicit, energy- and charge-conserving particle-in-cell algorithm for the relativistic Vlasov-Maxwell equations <https://doi.org/10.1016/j.jcp.2020.109228>`__.
+      The version implemented is an updated version that is relativistically correct, including the relativistic gamma factor for the particles.
```
```suggestion
```
I think that the Chen-Chacon paper is already relativistic.

Docs/source/usage/parameters.rst
```diff
@@ -81,6 +81,38 @@ Overall simulation parameters
...
+
+* ``algo.max_picard_iterations`` (`integer`, default: 10)
+    When `algo.evolve_scheme` is either `implicit_picard` or `semi_implicit_picard`, this sets the maximum number of Picard
+    itearations that are done each time step.
+
+* ``algo.picard_iteration_tolerance`` (`float`, default: 1.e-7)
+    When `algo.evolve_scheme` is either `implicit_picard` or `semi_implicit_picard`, this sets the convergence tolerance of
+    the iterations, the maximum of the fractional change of the field from one iteration to the next.
```
```suggestion
    the iterations, the maximum of the relative change of the L2 norm of the field from one iteration to the next.
```

Docs/source/usage/parameters.rst
```diff
@@ -81,6 +81,38 @@ Overall simulation parameters
...
+    If this is set to zero, the maximum number of iterations will always be done with the change only calculated on the last
+    iteration (for a slight optimization).
+
+* ``algo.require_picard_convergence`` (`bool`, default: 1)
+    When `algo.evolve_scheme` is either `implicit_picard` or `semi_implicit_picard`, this sets whether the iteration each step
+    is required to converge.
+    If it is required, an abort is raised if it does not converge and the code then exits.
+    If not, then a warning is issued and the calculation continues.
```
RemiLehe, line 119: Great! Thanks for adding this option!

Regression/WarpX-tests.ini
```diff
@@ -4443,3 +4443,20 @@ compileTest = 0
 doVis = 0
 compareParticles = 0
 analysisRoutine = Examples/Tests/nodal_electrostatic/analysis_3d.py
+
```
RemiLehe, line 4446: Could you also run the same test for the semi-implicit Picard method?

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -221,31 +228,33 @@ WarpX::Evolve (int numsteps)
...
-            FillBoundaryB(guard_cells.ng_FieldGather);
-            if (fft_do_time_averaging)
-            {
-                FillBoundaryE_avg(guard_cells.ng_FieldGather);
-                FillBoundaryB_avg(guard_cells.ng_FieldGather);
+        if (evolve_scheme == EvolveScheme::Explicit) {
+            if (num_mirrors>0){
+                applyMirrors(cur_time);
+                // E : guard cells are NOT up-to-date
+                // B : guard cells are NOT up-to-date
```
RemiLehe, line 235: I think that `applyMirrors` could be moved outside of the `if` condition for 
```
if (evolve_scheme == EvolveScheme::Explicit) {
```

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -286,9 +295,10 @@ WarpX::Evolve (int numsteps)
 
         m_particle_boundary_buffer->gatherParticles(*mypc, amrex::GetVecOfConstPtrs(m_distance_to_eb));
 
-        // Non-Maxwell solver: particles can move by an arbitrary number of cells
+        // Non-Maxwell solver or implicit: particles can move by an arbitrary number of cells
         if( electromagnetic_solver_id == ElectromagneticSolverAlgo::None ||
-            electromagnetic_solver_id == ElectromagneticSolverAlgo::HybridPIC )
+            electromagnetic_solver_id == ElectromagneticSolverAlgo::HybridPIC ||
+            evolve_scheme == EvolveScheme::ImplicitPicard)
```
RemiLehe: Hm, even though the timestep may be very large, the motion of the particles during one timestep still needs to be contained within the guard cells, for the current deposition to work correctly.

Therefore, I think that we can still use `RedistributeLocal`. In fact, this is also what we use with PSATD, for which we sometimes also take large timesteps (but have similar restrictions on the fact that particles need to be contained within the guard cells).

Source/Evolve/WarpXPushType.H
```diff
@@ -0,0 +1,16 @@
...
+ * License: BSD-3-Clause-LBNL
+ */
+#ifndef WARPX_PUSHTYPE_H_
+#define WARPX_PUSHTYPE_H_
+
+enum struct PushType : int
+{
+    Explicit = 0,
+    Implicit
+};
```
RemiLehe: Could you add a few comments explaining what this controls:
```suggestion
enum struct PushType : int
// Controls the exact definition of the velocity v used in both the current deposition 
// and the update of the particle position
{
    Explicit = 0, // v^{n+1/2} = u^{n+1/2}/gamma^{n+1/2}
    Implicit // v^{n+1/2} = (u^{n+1} - u^{n})/(gamma^{n+1} - gamma^n)
    // The implicit definition is required in the current deposition in order to have 
    // exact energy conservation ; see Eq. 17 in Chen, JCP  407 (2020) 109228
};
```

Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -126,208 +362,179 @@ void doDepositionShapeN (const GetParticlePosition<PIdx>& GetPosition,
...
-#if (AMREX_SPACEDIM >= 2)
-            // x direction
-            // Get particle position after 1/2 push back in position
-#if defined(WARPX_DIM_RZ)
-            // Keep these double to avoid bug in single precision
-            const double xmid = (rpmid - xmin)*dxi;
-#else
-            const double xmid = ((xp - xmin) + relative_time*vx)*dxi;
+/**
+ * \brief Direct current deposition for thread thread_num for the implicit scheme
```
RemiLehe, line 408: Could you add a comment saying:
```
The only difference with `doDepositionShapeN` is in the exact discrete definition of the particle velocity
See WarpXPushType.H
```

RemiLehe: Also, maybe in a follow-up PR, could we implement a single function for `doDepositionShapeN` and `doDepositionImplicitShapeN` with a template parameter (`push_type`), in order to reduce code duplication and make the differences more obvious. The function would always take both `uxp_n,uyp_n,uzp_n` and `uxp,uyp,uzp` as argument, but `uxp_n,uyp_n,uzp_n` would be ignored if `push_type` is `0` (explicit pusher)

Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -927,6 +1134,401 @@ void doEsirkepovDepositionShapeN (const GetParticlePosition<PIdx>& GetPosition,
...
+ * \param dx           3D cell size
+ * \param xyzmin       Physical lower bounds of domain.
+ * \param lo           Index lower bounds of domain.
+ * \param q            species charge.
+ * \param n_rz_azimuthal_modes Number of azimuthal modes when using RZ geometry.
+ * \param cost Pointer to (load balancing) cost corresponding to box where present particles deposit current.
+ * \param load_balance_costs_update_algo Selected method for updating load balance costs.
+ */
+template <int depos_order>
+void doChargeConservingDepositionShapeNImplicit (const amrex::ParticleReal * const xp_n,
```
RemiLehe, line 1164: Would there be a way to combine this with `doEsirkepovDepositionShapeN` (maybe as a follow-up PR), to avoid code duplication?

Source/Particles/PhysicalParticleContainer.H
```diff
@@ -134,7 +135,8 @@ public:
                          amrex::Real t,
                          amrex::Real dt,
                          DtType a_dt_type=DtType::Full,
```
RemiLehe, line 138: Could you add `push_type` to the docstring?
```
\param push_type Controls the exact definition of the velocity v used in both the current deposition and the update of the particle position
```

Source/Particles/Gather/FieldGather.H
```diff
@@ -451,6 +451,434 @@ void doGatherShapeN (const amrex::ParticleReal xp,
...
+    int dil_By = 1, diu_By = 1;
+    if (i_By_old < i_By_new) { dil_By = 0; }
+    if (i_By_old > i_By_new) { diu_By = 0; }
+    int dkl_By = 1, dku_By = 1;
+    if (k_By_old < k_By_new) { dkl_By = 0; }
+    if (k_By_old > k_By_new) { dku_By = 0; }
+#endif
+
+#if defined(WARPX_DIM_3D)
+
```
RemiLehe, line 679: Do I understand correctly that this is the gather that has the shape-factor that corresponds to the Esirkepov?

If yes, I think it would be good to make this more explicit by:
- Reflecting this in the function name, e.g. `doGatherShapeNImplicitEsirkepovLike`
- Mentioning the differences with the regular gather function in the docstring, namely that this uses the Esirkepov shape

Source/Particles/Gather/FieldGather.H
```diff
@@ -606,4 +1034,92 @@ void doGatherShapeN (const amrex::ParticleReal xp,
...
+ * \param bx_type,by_type,bz_type IndexType of the magnetic field
+ * \param dx_arr                  3D cell spacing
+ * \param xyzmin_arr              Physical lower bounds of domain in x, y, z.
+ * \param lo                      Index lower bounds of domain.
+ * \param n_rz_azimuthal_modes    Number of azimuthal modes when using RZ geometry
+ * \param nox                     order of the particle shape function
+ * \param galerkin_interpolation  whether to use lower order in v
+ */
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+void doGatherShapeNImplicit (
```
RemiLehe, line 1058: Is this function used ?

Source/Particles/Gather/FieldGather.H
```diff
@@ -606,4 +1034,92 @@ void doGatherShapeN (const amrex::ParticleReal xp,
     }
 }
 
+
+/**
+ * \brief Field gather for a single particle
+ *
+ * \param xp,yp,zp                Particle position coordinates
```
RemiLehe: Could you update the docstring to include `xp_n,yp_n,zp_n`?

Source/Particles/Gather/FieldGather.H
RemiLehe: Also, should it be used in `doGatherShapeN`?

Source/Evolve/WarpXPushType.H
RemiLehe: Hm, maybe we should also mention that it gathers the field at the position `(x^{n+1}+x^n)/2`.

Source/Particles/Deposition/CurrentDeposition.H
RemiLehe: Discussed offline: this will be done in a follow-up PR.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -2924,26 +2940,277 @@ PhysicalParticleContainer::PushPX (WarpXParIter& pti,
...
+                UpdatePosition(xp, yp, zp, ux[ip], uy[ip], uz[ip], dt);
+                setPosition(ip, xp, yp, zp);
+            }
+        }
+#endif
+
+#ifdef WARPX_QED
+        [[maybe_unused]] auto foo_local_has_quantum_sync = local_has_quantum_sync;
+        [[maybe_unused]] auto *foo_podq = p_optical_depth_QSR;
+        [[maybe_unused]] const auto& foo_evolve_opt = evolve_opt; // have to do all these for nvcc
```
RemiLehe: Out of curiosity, what is the error with `nvcc` when not initializing this variables?

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -2924,26 +2940,277 @@ PhysicalParticleContainer::PushPX (WarpXParIter& pti,
...
+            }
+        }
+#else
+            amrex::ignore_unused(qed_control);
+#endif
+    });
+}
+
+/* \brief Perform the implicit particle push operation in one fused kernel
+ *        The main difference is the order.
```
RemiLehe:
```suggestion
 *        The main difference with `PushPX` is the order.
```

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -2924,26 +2940,277 @@ PhysicalParticleContainer::PushPX (WarpXParIter& pti,
...
+ *         - push position 1/2 dt
```
RemiLehe:
```suggestion
 *         - push position by 1/2 dt
```

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -2924,26 +2940,277 @@ PhysicalParticleContainer::PushPX (WarpXParIter& pti,
...
+ *         - gather fields
+ *         - push velocity dt
```
RemiLehe:
```suggestion
 *         - push velocity by dt
```

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -2924,26 +2940,277 @@ PhysicalParticleContainer::PushPX (WarpXParIter& pti,
...
+
+        UpdatePositionImplicit(xp, yp, zp, ux_n[ip], uy_n[ip], uz_n[ip], ux[ip], uy[ip], uz[ip], 0.5_rt*dt);
+        setPosition(ip, xp, yp, zp);
+
+        amrex::ParticleReal Exp = 0._rt, Eyp = 0._rt, Ezp = 0._rt;
+        amrex::ParticleReal Bxp = 0._rt, Byp = 0._rt, Bzp = 0._rt;
+
+        if(!t_do_not_gather){
+            // first gather E and B to the particle positions
+            doGatherShapeNImplicit(xp_n, yp_n, zp_n, xp, yp, zp, Exp, Eyp, Ezp, Bxp, Byp, Bzp,
```
RemiLehe, line 3179: My understanding is that this function will always gather with Esirkepov-like shape factor, and that this will therefore be energy conserving when using the Esirkepov current deposition. However, when using the direct current deposition (which also has a new implicit version in `CurrentDeposition.H`), I don't think we will get energy conservation.

Did I misunderstand? Or should we had an `if` statement here to use the correct shape function in the case of the direct current deposition for direct deposition? Or do we simply do not want to use the direct deposition with the implicit scheme? (in which case we could maybe revert the corresponding changes in `CurrentDeposition.H`.)

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -2924,26 +2940,277 @@ PhysicalParticleContainer::PushPX (WarpXParIter& pti,
...
+
+/* \brief Perform the implicit particle push operation in one fused kernel
+ *        The main difference is the order.
+ *         - push position 1/2 dt
+ *         - gather fields
+ *         - push velocity dt
+ *         - average old and new velocity to get time centered value
+ */
+void
+PhysicalParticleContainer::ImplicitPushXP (WarpXParIter& pti,
```
RemiLehe, line 3007: Again here: would there be a way to avoid code duplication with `PushPX`, by having a single function with a template parameter?

Source/Particles/Pusher/UpdatePosition.H
```diff
@@ -42,4 +42,39 @@ void UpdatePosition(amrex::ParticleReal& x, amrex::ParticleReal& y, amrex::Parti
     z += uz * inv_gamma * dt;
 }
 
+/** \brief Push the particle's positions over one timestep,
+ *    given the value of its momenta `ux`, `uy`, `uz` */
```
RemiLehe: Maybe we could be more explicit in the docstring here:
```
More specifically, instead of using the standard leapfrog algorithm
(`x^{n+1} - x^n = dt*u^{n+1/2}/gamma^{n+1/2}`), here we use the implicit version
`x^{n+1} - x^n = dt*(u^{n+1} - u^{n})/(gamma^{n+1} - gamma^n)
see Eq. 17 in Chen, JCP  407 (2020) 109228
```

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -538,27 +541,75 @@ WarpXParticleContainer::DepositCurrent (WarpXParIter& pti,
...
+                        GetPosition, wp.dataPtr() + offset,
+                        uxp_n.dataPtr() + offset, uyp_n.dataPtr() + offset, uzp_n.dataPtr() + offset,
+                        uxp.dataPtr() + offset, uyp.dataPtr() + offset, uzp.dataPtr() + offset, ion_lev,
+                        jx_arr, jy_arr, jz_arr, np_to_deposit, dt, dx, xyzmin, lo, q,
+                        WarpX::n_rz_azimuthal_modes, cost,
+                        WarpX::load_balance_costs_update_algo);
+                }
             }
         } else if (WarpX::current_deposition_algo == CurrentDepositionAlgo::Vay) {
             if        (WarpX::nox == 1){
```
RemiLehe, line 618: Could raise an error here, that prevents the user from trying to use both the Vay deposition and the implicit algorithm at the same time.

Source/WarpX.H
```diff
@@ -120,6 +121,12 @@ public:
 
     void Evolve (int numsteps = -1);
 
+    void EvolveImplicitPicardInit (int lev);
+    void SaveParticlesAtStepStart (WarpXParticleContainer& pc, int lev);
```
RemiLehe: Could we change the function name to make it clear that this is used only in the implicit algorithm:
```suggestion
    void SaveParticlesAtImplicitStepStart (WarpXParticleContainer& pc, int lev);
```

Source/WarpX.cpp
```diff
@@ -1222,6 +1227,38 @@ WarpX::ReadParameters ()
...
+                electromagnetic_solver_id == ElectromagneticSolverAlgo::CKC,
+                "Only the Yee EM solver is supported with the implicit and semi-implicit schemes");
+
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                particle_pusher_algo == ParticlePusherAlgo::Boris,
+                "Only the Boris particle pusher is supported with the implicit and semi-implicit schemes");
+
+            if (current_deposition_algo == CurrentDepositionAlgo::Direct) {
+                WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                    field_gathering_algo == GatheringAlgo::MomentumConserving,
```
RemiLehe: I am not sure I understand this. Maybe we could discuss this offline.

Source/WarpX.cpp
```diff
@@ -1222,6 +1227,38 @@ WarpX::ReadParameters ()
...
+                current_deposition_algo == CurrentDepositionAlgo::Direct,
+                "Only Esirkepov or Direct current deposition supported with the implicit and semi-implicit schemes");
+
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                electromagnetic_solver_id == ElectromagneticSolverAlgo::Yee ||
+                electromagnetic_solver_id == ElectromagneticSolverAlgo::CKC,
+                "Only the Yee EM solver is supported with the implicit and semi-implicit schemes");
+
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                particle_pusher_algo == ParticlePusherAlgo::Boris,
```
RemiLehe: My impression is that the exact energy conservation would be preserved with any pusher for which the `vxB` force does not modify the norm of `u`, when updating `u` with `u^{n+1} - u^n = q*dt/m * (E + vxB)`. Using a dot product with `(u^{n+1} + u^n)`, we can see that this happens if `(u^{n+1} + u^n).(vxB) = 0`. In particular, this works if `v` is colinear with `u^{n+1} + u^n`, which is the case for the Boris and [Higuera-Carey]
(https://pubs.aip.org/aip/pop/article/24/5/052104/989750/Structure-preserving-second-order-integration-of) pusher (but not for the Vay pusher)
```suggestion
                particle_pusher_algo == ParticlePusherAlgo::Boris ||
                particle_pusher_algo == ParticlePusherAlgo::HigueraCarey
```

RemiLehe (changes requested):

Source/WarpX.cpp
RemiLehe: Sounds good. 
In the case of direct deposition: should ask the user to set `interpolation.galerkin_scheme = 0` (this should be enough to ensure that the shape factors are symmetrical) instead of asking to set the field gathering to `momentum-conserving`, so as to preserve a consistent behavior of `momentum-conserving` (in which case we would remove the above `if` statement)

Source/Particles/PhysicalParticleContainer.cpp
RemiLehe: This could maybe be done in a follow-up PR.

Source/Particles/PhysicalParticleContainer.cpp
RemiLehe: Ah, actually now I see that, when using `galerkin_interpolation = 0`, this will use the regular shape factors.

Docs/source/usage/parameters.rst
dpgrote: Yes, you are correct. I don't remember why I thought that it wasn't.

Source/Evolve/WarpXEvolve.cpp
dpgrote: Since the mirrors affect the fields, it is included in the iterations in `OneStep_ImplicitPicard` so doesn't need to also be called here.

Source/Evolve/WarpXEvolve.cpp
dpgrote: Ok. I was just thinking that the issue is whether particles can jump across processors. But yes, since the iterations assume locality the particles cannot go beyond the guard cells. I'll undo this change.

Source/Evolve/WarpXPushType.H
dpgrote: Your comments are good, though the change in how gamma is calculated is a side effect of the overall particle advance, with the explicit being a leap frog advance and the implicit being Crank-Nicholson. Perhaps this should be the comment.

RemiLehe: OK, sounds good. Feel free to adapt my suggestions accordingly.

Source/Particles/Pusher/UpdatePosition.H
dpgrote: I added comments. For completeness, I added a comment to the explicit version as well.

Source/Particles/WarpXParticleContainer.cpp
dpgrote: Once this PR gets merged, I would like to submit another PR that puts each of the `if` blocks into separate routines. This routine is becoming unwieldy making it hard to follow the logic of the outermost `if` blocks.

Source/WarpX.cpp
RemiLehe: I think I now get it. Could you replace this by check for `interpolation.galerkin_shape = 0`?

Source/WarpX.cpp
dpgrote: Yes, the Higuera-Cary algorithm does work - energy is conserved.

Source/WarpX.cpp
dpgrote: This is to keep the deposition and gather consistent, using the same stencil. Without this consistency, the energy would not be exactly converged.

I'll update this based on the comments below.

Source/WarpX.cpp
dpgrote: Ok, yes, this works. I hadn't thought about it this way.  I'll update the code accordingly.

Source/Particles/PhysicalParticleContainer.cpp
dpgrote: This I don't know. This code was copied `PushPX`

Source/Particles/PhysicalParticleContainer.cpp
dpgrote: Maybe, but there are enough differences that the code could be messy with lots of template blocks.

BTW, comparing the two, I noticed that I had not made the change in the implicit version related to the initialization of the external fields. This is now fixed.

Source/Particles/Deposition/CurrentDeposition.H
dpgrote: Yes, templating can be done. Though with `push_type = 0` the `uxp_n` etc would be null pointers since those particle attributes would not have been defined.

Source/Particles/Gather/FieldGather.H
dpgrote: Yes, correct. I'll add comments and change the name. Maybe I should call it `doGatherShapeNImplicitGrote` and name it after the inventor :)

Source/Particles/Gather/FieldGather.H
dpgrote: This is called by `ImplicitPushXP`.

RemiLehe (approved):

RemiLehe:
As a summary of potential follow-up PRs that were mentioned in the review:

- [ ] Adding a CI test for the semi-implicit solver
- [ ] Trying to reduce code duplication:
    - [ ] for the explicit and implicit version of the direct deposition kernel (possibly with a template parameter)
    - [ ] for the explicit and implicit version of the Esirkepov deposition (if I understood correctly, this will be combined with adding subcycling for the Esirkepov kernel)
    - [ ] possibly for `PushXP` / `PushPXImplicit` if this does not become too messy

Issue #4038, opened 2023-06-26
PEC boundary conditions for RZ PSATD


rezaplasma:
hi

I'm running with `rz `geometry. when I used `PSATD `option for method in `ElectromagneticSolver`, I got below error:
`PEC boundary not implemented for PSATD, yet
`
is it possible to resolve that error or it doesn't work for this geometry?
although , the other options also don't work and i found only `Yee `method can be used for `rz `geometry

EZoni:
Hi @rezaplasma, thank you for reporting this.

Would you like to attach here the input file that was used and also the cmake compilation flags that were used to build the executable, so that we can try to reproduce the issue?

In general, not all boundary conditions work with all algorithmic options in WarpX, but PSATD is definitely supported also for RZ geometry, so we might want to look more in detail to your input file.

rezaplasma:
Hi @EZoni 

Thanks for your reply
I'm running with python (PICMI scripts)
The script is attached

When i set method `PSATD` i face that error

rezaplasma:
[rz1.txt](https://github.com/ECP-WarpX/WarpX/files/11872419/rz1.txt)

rezaplasma:
@EZoni 

I have installed warpx with conda package
Do i need to install build option `Warpx_PSATD`?

RemiLehe:
@rezaplasma Did you try to use `'open'` instead of `'dirichlet'` in the PICMI script?
This should work, in principle, with the PSATD solver.

rezaplasma:
@RemiLehe 

Hi
I used `open` for boundaries
```
lower_boundary_conditions = ['none', 'open'],
upper_boundary_conditions = ['open', 'open'],

```
The below error was received :

```
0::Assertion `field_boundary_lo[1] != FieldBoundaryType::PML && field_boundary_hi[1] != FieldBoundaryType::PML' failed, file "/home/conda/feedstock_root/build_artifacts/warpx_1683172402631/work/Source/WarpX.cpp", line 934, Msg:
### ERROR   : PML are not implemented in RZ geometry along z; please set a
#            different boundary condition using boundary.field_lo and
#            boundary.field_hi.
```
I think the `open` condition is not available for PSATD solver!!

RemiLehe:
Yes, that's correct.
In summary:
- with RZ PSATD: you can use `open`
- with RZ FDTD: you can use `dirichlet`
Does that solve your issue?

rezaplasma:
[...]

PR #4072, opened 2023-07-06
Fix clang-tidy `bugprone-branch-clone` warning

EZoni:
PR #3958 introduced code that triggers the clang-tidy `bugprone-branch-clone` warning, treated as error in CI. This PR fixes it. Please feel free to post suggestions for better solutions.

PR #4073, opened 2023-07-07
Fix max_grid_size for RZ PSATD

dpgrote:
This PR sets max_grid_size_x = n_cell[0]. When using RZ PSATD, all of the radial dimension must be in the same block, so the values of blocking factor and max grid size are modified so this happens. The code was setting the max grid size to be equal to the blocking factor but this fails in cases when the number of grid cells is not a power of 2. The blocking factor must be a power of 2 and the max grid size must be a integer multiple of the blocking factor unless it is less than the blocking factor. By setting the max grid size equal to the blocking factor, amrex checks whether the number of grid cells is a multiple of the blocking factor and this fails when the number of cells is not a power of 2. By setting the max grid size equal to the number of grid cells the check is skipped since the max grid size will be less than the blocking factor.

This PR solves an issue brought up in issue #4038.

### DIFF
--- a/Source/Utils/WarpXUtil.cpp
+++ b/Source/Utils/WarpXUtil.cpp
@@ -322,14 +322,15 @@ void CheckGriddingForRZSpectral ()
     Vector<int> max_grid_size_x(max_level+1);
 
     // Set the radial block size to be the power of 2 greater than or equal to
-    // the number of grid cells. The blocking_factor must be a power of 2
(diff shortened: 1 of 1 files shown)


EZoni (approved):
Thank you for the fix, @dpgrote! Looks good to me. The PR needs to be rebased once #4072 is merged, so that the `clang-tidy` check is fixed.

Issue #4065, opened 2023-06-30
Possibly remove `raw_fields` parameter for `FieldProbe` reduced diagnostic.


roelof-groenewald:
As an aside, though, what the code is doing for `raw_fields` doesn't make much sense for the line and plane probes. Unless I missed something, when `raw_fields` is true, it is only fetching the field at the single coordinate `x, y, z_probe` and writing that out repeatedly for each probe location. Perhaps an easy fix is to instead call `doGatherShapeN` with interp_order = 0, to do nearest grid point interpolation. In fact, the `raw_fields` option is not really needed since the interpolation order can be specified. Stuff for a later PR.

_Originally posted by @dpgrote in https://github.com/ECP-WarpX/WarpX/pull/4057#pullrequestreview-1507588245_

roelof-groenewald:
Close due to above PR.

PR #2430, opened 2021-10-14
FieldProbe using Particle

ElisaRheaume:
Adjusting `FieldProbe.cpp` to use test particles. Once this is done, can be expanded with options to include many more particles and varied outputs.

- [x] add particle container (with one particle for now)
- [x] add Poynting vector
- [x] add time-integration functionality
- [x] generalize to RZ support
- [x] add new test for integrated value in inputs file
- [x] fix: in MPI-parallel simulations, only plot the probe output line once per step
- [x] redistribute particle on load balance
- [x] check values still match `development`
- [x] runtime test on GPU

PR #4074, opened 2023-07-11
Remove field probe raw_fields option

dpgrote:
This PR addresses the issue raised in #4065 and removes the `raw_fields` option.

X-ref:
- #4057
- #3746
- ...
- #2430


ax3l (approved):
This is great, thank you!

PR #4075, opened 2023-07-11
 Clang-tidy CI test: make bugprone checks list more compact

lucafedeli88:
This PR makes the list of the "bugprone" clang-tidy checks more compact: all checks are enabled except very few explicitly disabled ones.


EZoni (approved):
Thanks @lucafedeli88!

PR #4079, opened 2023-07-12
Add reduced diag precision input parameter

dpgrote:
This adds a parameter that allows the user to control the precision of the data written to the reduced diagnostics output files.

### DIFF
--- a/Source/Diagnostics/ReducedDiags/ReducedDiags.H
+++ b/Source/Diagnostics/ReducedDiags/ReducedDiags.H
@@ -42,6 +42,9 @@ public:
     /// separator in the output file
     std::string m_sep = " ";
 
+    /// precision for data in the output file
+    int m_precision = 14;
+
     /// output data
     std::vector<amrex::Real> m_data;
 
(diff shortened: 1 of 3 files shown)


Source/Diagnostics/ReducedDiags/ReducedDiags.H
```diff
@@ -42,6 +42,9 @@ public:
     /// separator in the output file
     std::string m_sep = " ";
 
+    /// precision for data in the output file
+    int m_precision = 14;
```
lucafedeli88, line 46:
```suggestion
    constexpr int default_precision = 14;
    int m_precision = default_precision;
```
I suggest this small change to improve readability.

lucafedeli88 (commented):
Thanks for this PR! I left a tiny comment

Source/Diagnostics/ReducedDiags/ReducedDiags.H
dpgrote: I can make this change, but note that none of the other variables have a separate constexpr to hold the default value.

lucafedeli88: You are right. We should maybe address these issues all at once if at some point we decide to enforce the `readability-magic-numbers` check in `clang-tidy` !

lucafedeli88 (approved):

PR #4080, opened 2023-07-13
ablastr: fix some missing or superfluous #include directives 

lucafedeli88:
This PR fixes some missing or superfluous `#include`  directives in ablastr. 
Issues were found using the `iwyu` (Include What You Use) tool.

### DIFF
--- a/Source/ablastr/utils/msg_logger/MsgLogger.cpp
+++ b/Source/ablastr/utils/msg_logger/MsgLogger.cpp
@@ -13,10 +13,10 @@
 #ifdef AMREX_USE_MPI
 #   include <AMReX_ParallelDescriptor.H>
 #endif
-#include <AMReX_Print.H>
 
-#include <iostream>
-#include <sstream>
+#include <algorithm>
+#include <array>
+#include <memory>
 #include <numeric>
 
 namespace abl_msg_logger = ablastr::utils::msg_logger;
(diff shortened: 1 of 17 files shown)


Source/ablastr/utils/msg_logger/MsgLogger.cpp
```diff
@@ -211,11 +211,19 @@ MsgWithCounterAndRanks::deserialize (std::vector<char>::const_iterator&& it)
...
+        m_num_procs{amrex::ParallelDescriptor::NProcs()},
+        m_io_rank{amrex::ParallelDescriptor::IOProcessorNumber()}
+    {}
+#else
+    Logger::Logger() :
+        m_rank{0},
+        m_num_procs{0},
+        m_io_rank{0}
+    {}
+#endif
```
EZoni: Was this code change part of the PR or was it accidentally included?

lucafedeli88: It is actually part of the PR! `AMReX_ParallelDescriptor.H` was probably included transitively even if `AMREX_USE_MPI` was not defined.

EZoni: I see, thanks! So I guess `m_num_procs` is not used at all in the rest of the code, when `AMREX_USE_MPI` is not defined? Is that why we can set `m_num_procs{0}` in this case? I'm asking because intuitively I was expecting `m_num_procs{1}` as minimum value set in the trivial case of one single process.

lucafedeli88: Good observation indeed... on a second thought I think that we can always include `AMReX_ParallelDescriptor.H` in this file, since it should already be able to deal with the case in which `AMREX_USE_MPI` is undefined.

EZoni (approved):

PR #4083, opened 2023-07-14
Clean main.cpp by moving some functions into WarpX constructor

lucafedeli88:
This PR moves some functions that manipulate WarpX inputfile into WarpX constructor. It also removes a redundant check of the dimensionality of the simulation.

### DIFF
--- a/Source/Utils/WarpXUtil.cpp
+++ b/Source/Utils/WarpXUtil.cpp
@@ -300,8 +300,6 @@ void CheckDims ()
 void CheckGriddingForRZSpectral ()
 {
 #ifdef WARPX_DIM_RZ
-    // Ensure that geometry.dims is set properly.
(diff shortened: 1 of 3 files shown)


ax3l:
@lucafedeli88 thank you for the PR! It looks like some of the tests failed with precision issues, maybe because adjustments to boundaries or blocks so are made too late?

lucafedeli88:
> @lucafedeli88 thank you for the PR! It looks like some of the tests failed with precision issues, maybe because adjustments to boundaries or blocks so are made too late?

@ax3l , I don't think that moving these functions into WarpX constructor can work... I suspect that they need to be called before `AmrCore` constructor is called, and `AmrCore` is WarpX's parent class. For this reason, I think that we'll have to close this PR...

PR #4086, opened 2023-07-14
[WIP] Fix rho diagnostic, when mesh refinement is on

RemiLehe:
While the diagnostics for the fields E, B, J directly access the `MultiFab` used in the PIC loop, the diagnostic for `rho` instead redeposits the charge density (this is because `rho` is not always needed in the PIC loop, and is therefore not always allocated). 

However, inside the diagnostic functor that deposits `rho`, the communication across MR levels was not done properly. As a result, on the `dev` branch, the diagnostic of `rho` typically shows an empty patch (see example below, which uses [this script](https://github.com/ECP-WarpX/WarpX/blob/development/Examples/Physics_applications/laser_acceleration/inputs_2d) with `max_step=1000`). This PR fixes the issue by properly depositing the charge on each level, and by calling the function `SyncRho`.

Screenshot 2023-07-14 at 1 46 52 PM

TODO:
- [ ] Add automated test
- [ ] Add documentation/explanation for the parameter `bottom_level`

### DIFF
--- a/Source/Diagnostics/ComputeDiagFunctors/RhoFunctor.cpp
+++ b/Source/Diagnostics/ComputeDiagFunctors/RhoFunctor.cpp
@@ -43,7 +43,7 @@ RhoFunctor::operator() ( amrex::MultiFab& mf_dst, const int dcomp, const int /*i
     rho_fp.resize(nlevs_max);
     rho_buf.resize(nlevs_max);
     const int ng_rho = warpx.get_ng_depos_rho().max();
-    for (int lev=0; lev<=warpx.maxLevel(); lev++) {
+    for (int lev=m_lev; lev<=warpx.maxLevel(); lev++) {
         // Allocate rho_fp
         const amrex::DistributionMapping& dm = warpx.DistributionMap(lev);
         amrex::BoxArray nba = warpx.boxArray(lev);
@@ -51,7 +51,7 @@ RhoFunctor::operator() ( amrex::MultiFab& mf_dst, const int dcomp, const int /*i
         rho_fp[lev] = std::make_unique<amrex::MultiFab>(nba,dm,WarpX::ncomps,ng_rho);
         rho_fp[lev]->setVal(0.0);
         // Allocate rho_cp
-        if (lev>0) {
+        if (lev>m_lev) {
             nba.coarsen(warpx.refRatio(lev-1));
             rho_cp[lev] = std::make_unique<amrex::MultiFab>(nba,dm,WarpX::ncomps,ng_rho);
         }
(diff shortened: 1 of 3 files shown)


Source/Diagnostics/ComputeDiagFunctors/RhoFunctor.cpp
```diff
@@ -43,17 +64,24 @@ RhoFunctor::operator() ( amrex::MultiFab& mf_dst, const int dcomp, const int /*i
...
+            rho_fp[lev] = mypc.GetChargeDensity(lev, true);
+        }
     }
 
-    // Handle the parallel transfers of guard cells and
-    // apply the filtering if requested.
-    warpx.ApplyFilterandSumBoundaryRho(m_lev, m_lev, *rho, 0, rho->nComp());
+    // Handle parallel transfers of guard cells,
+    // summation between levels, and filtering
+    warpx.SyncRho(rho_fp, rho_cp, rho_buf, m_lev);
```
RemiLehe, line 81: We call `SyncRho` with `bottom_level=m_lev`. This is a small optimization: if we want `rho` at `m_lev`, we only need to sum `rho` from the highest level down to `lev=m_lev` - no need to go all the way down to `lev=0`.

prkkumar:
I tested this with the attached input script:  
[input.txt](https://github.com/ECP-WarpX/WarpX/files/12885665/input.txt)

I still see the unexpected high values of rho on level 1
Screenshot 2023-10-12 at 10 50 59 AM

Note that this result is obtained from a simulation run in boosted frame (see the attached input file). The attached image is obtained from the BTD data at the final snapshot.

When I plot the boosted frame data directly (i.e. without using the BTD data), I see vertical lines at subdomain boundaries)
Screenshot 2023-10-12 at 11 02 21 AM

RemiLehe:
In order to go further:
- Keep changes in this PR (since it enables SyncRho across levels) and fix the above mentioned bug showing vertical lines in subdomain boundaries.
- Add handling of deposition buffers (this could be done by abstracting the code that we have in the regular deposition of `rho`, which sorts particles according to whether they are in a deposition buffer or not, see https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/PhysicalParticleContainer.cpp#L1966), by abstracting the code in a separate function.
- (optional) Grab `rho` if it already exists, just like we do for `J` https://github.com/ECP-WarpX/WarpX/blob/development/Source/Diagnostics/ComputeDiagFunctors/JFunctor.cpp#L33C1-L33C27
- (later/reevaluate) Refactor the code for gather buffers and deposition buffers to not rely on sorting and instead use masks in the kernels.
- (later) Fix slight time difference between `J` (half-timestep) and `rho` (integer-timestep) when we combine them in the BTD. One solution would be to write a single BTD functor for deposit `J/rho` that does it at the right time (only sorts once) and combines them immediately. This would be more computationally expensive.

RemiLehe:
As suggested by @RevathiJambunathan : we should set the deposition buffer size to 0 in order to test that it solves the weird-looking border in the plot of rho.

PR #4087, opened 2023-07-17
 Update LUMI documentation 

lucafedeli88:
This PR updates documentation for LUMI supercomputer using the same format of the documentation for Frontier.


Docs/source/install/hpc/lumi.rst
```diff
@@ -22,64 +22,112 @@ If you are new to this system, **please see the following resources**:
...
+Use the following :ref:`cmake commands <building-cmake>` to compile:
 
 .. code-block:: bash
 
    cd $HOME/src/warpx
-   rm -rf build
-
-   cmake -S . -B build -DWarpX_DIMS="1;2;3" -DWarpX_COMPUTE=HIP -DWarpX_PSATD=ON -DWarpX_QED_TABLE_GEN=ON
-   cmake --build build -j 16
+   rm -rf build_lumi
```
ax3l, line 89: I usually take a `build_<computer>` name if the home directory is shared between different computers. This is for instance the case at NERSC and OLCF.

Maybe CSC the same, but otherwise you can just call it `build` :)

ax3l (approved):
Thank you very much! :)

PR #4090, opened 2023-07-18
Fix: Ablastr PoissonSolver Constant

ax3l:
Seen in ImpactX:
```
build/_deps/fetchedablastr-src/Source/ablastr/fields/PoissonSolver.H:163:31: error: use of undeclared identifier 'PhysConst'
        rho[lev]->mult(-1._rt/PhysConst::ep0); // TODO: when do we "un-multiply" this? We need to document this side-effect!
                                   ^
```

### DIFF
--- a/Source/ablastr/fields/PoissonSolver.H
+++ b/Source/ablastr/fields/PoissonSolver.H
@@ -160,7 +160,8 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
     // scale rho appropriately; also determine if rho is zero everywhere
     amrex::Real max_norm_b = 0.0;
     for (int lev=0; lev<=finest_level; lev++) {
-        rho[lev]->mult(-1._rt/PhysConst::ep0); // TODO: when do we "un-multiply" this? We need to document this side-effect!
+        using namespace ablastr::constant::SI;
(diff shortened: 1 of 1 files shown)


Source/ablastr/fields/PoissonSolver.H
```diff
@@ -160,7 +160,8 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
     // scale rho appropriately; also determine if rho is zero everywhere
     amrex::Real max_norm_b = 0.0;
     for (int lev=0; lev<=finest_level; lev++) {
-        rho[lev]->mult(-1._rt/PhysConst::ep0); // TODO: when do we "un-multiply" this? We need to document this side-effect!
+        using namespace ablastr::constant::SI;
```
EZoni, line 163: Does the using-declaration need to be inside the `for` loop over levels?

lucafedeli88: I would suggest placing the `using` declaration at the beginning of the function rather then in the for loop.

ax3l: The using declaration is performance neutral and the only type we need is inside this loop. Thus, I would not pollute the whole scope of the body of the function just yet and keep it localized here.

lucafedeli88: Ok! you have convinced me

lucafedeli88 (approved):

PR #4091, opened 2023-07-18
Add write_diagonstics_on_restart input option

dpgrote:
This PR adds the `warpx.write_diagonstics_on_restart` input parameter, which does as it says when true. Sometimes it is convenient to have the code rewrite the diagnostics at the time of the restart, for instance if the restart is being done in a different directory than the original simulation. This mirrors the AMReX parameter `amr.plotfile_on_restart` but is more general since is works for diagnostics of any format and the reduced diagnostics.

### DIFF
--- a/Source/WarpX.H
+++ b/Source/WarpX.H
@@ -1535,6 +1535,7 @@ private:
 
     std::string restart_chkfile;
 
+    /** When `true`, write the diagnostics after restart at the time of the restart. */
     bool write_diagonstics_on_restart = false;
 
     amrex::VisMF::Header::Version plotfile_headerversion  = amrex::VisMF::Header::Version_v1;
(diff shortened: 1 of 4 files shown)


ax3l (approved):
Ok. Thanks! :)

ax3l:
@dpgrote Do you know if this works with plotfiles and openPMD?

Source/WarpX.H
```diff
@@ -1535,6 +1535,8 @@ private:
 
     std::string restart_chkfile;
 
+    bool write_diagonstics_on_restart = false;
```
ax3l, line 1539: @dpgrote please add a doxygen string for this member variable.
```suggestion
    /** When `true`, write the diagnostics after restart at the time of the restart. */
    bool write_diagonstics_on_restart = false;
```

dpgrote:
> @dpgrote Do you know if this works with plotfiles and openPMD?

Yes, it should. This controls to the call to `FilterComputePackFlush` which makes the diagnostics for any of the formats.

PR #4097, opened 2023-07-20
WarpX class: move a couple of function definitions from WarpX.H to WarpX.cpp

lucafedeli88:
This PR moves the definition of a couple of functions from the header to the cpp file of the WarpX class

### DIFF
--- a/Source/WarpX.cpp
+++ b/Source/WarpX.cpp
@@ -2883,6 +2883,30 @@ WarpX::getCosts (int lev)
 }
 
 void
+WarpX::setLoadBalanceEfficiency (const int lev, const amrex::Real efficiency)
+    {
+        if (m_instance)
+        {
+            m_instance->load_balance_efficiency[lev] = efficiency;
+        } else
+        {
+            return;
+        }
+    }
+
+amrex::Real
+WarpX::getLoadBalanceEfficiency (const int lev)
+    {
+        if (m_instance)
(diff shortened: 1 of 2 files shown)


Source/WarpX.cpp
```diff
@@ -2882,6 +2882,30 @@ WarpX::getCosts (int lev)
...
+WarpX::setLoadBalanceEfficiency (const int lev, const amrex::Real efficiency)
+    {
+        if (m_instance)
+        {
+            m_instance->load_balance_efficiency[lev] = efficiency;
+        } else
+        {
+            return;
+        }
+    }
```
```suggestion
{
    if (m_instance)
    {
        m_instance->load_balance_efficiency[lev] = efficiency;
    } else
    {
        return;
    }
}
```

Source/WarpX.cpp
```diff
@@ -2882,6 +2882,30 @@ WarpX::getCosts (int lev)
...
+WarpX::getLoadBalanceEfficiency (const int lev)
+    {
+        if (m_instance)
+        {
+            return m_instance->load_balance_efficiency[lev];
+        } else
+        {
+            return -1;
+        }
+    }
```
```suggestion
{
    if (m_instance)
    {
        return m_instance->load_balance_efficiency[lev];
    } else
    {
        return -1;
    }
}
```

ax3l (approved):
Thx!

PR #4099, opened 2023-07-20
Disable Multi-Line Support in amrex::ParmParse

WeiqunZhang:
By disabling the multi-line support, we can avoid mistakes like

    # galilean
    psatd.use_default_v_galilean  # Forgot = 1

Here, `psatd.use_default_v_galilean` is silently ignored because with the multi-line support it becomes part of the values for the previous definition (i.e., `algo.current_deposition = direct psatd.use_default_v_galilean`). This case will now abort. Note that using multiple lines in a pair of double quotes is still supported. For example,

    electrons.density_function(x,y,z) = "nc*n0*(
            if(abs(z)<=r0, 1.0, if(abs(z)<r0+Lcut, exp((-abs(z)+r0)/L), 0.0)) )"
    hydrogen.density_function(x,y,z) = "nc*n0*(
            if(abs(z)<=r0, 1.0, if(abs(z)<r0+Lcut, exp((-abs(z)+r0)/L), 0.0)) )"

### DIFF
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -110,7 +110,7 @@ jobs:
         which nvcc || echo "nvcc not in PATH!"
 
         git clone https://github.com/AMReX-Codes/amrex.git ../amrex
-        cd ../amrex && git checkout --detach 10b6cb26d0ac359922276ce08eeb728a14624b70 && cd -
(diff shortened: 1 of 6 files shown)


WeiqunZhang:
Due to latest changed in amrex, this PR will be closed and I will submit a new PR.

PR #4101, opened 2023-07-21
fix `chi_min` and `chi_max` in `ParticleExtrema` red diags

aeriforme:
This PR fixes the computation of `chi_min` and `chi_max` in the `ParticleExtrema` reduced diagnostic. 
Based on a conversation with @WeiqunZhang and @EZoni on Slack, the modifications consist in the following:
* the values resulting from the reduce operations are extracted outside of the loop over the boxes
* final min and max values are computed outside of the loop over mesh refinement levels 
* added `ParallelDescriptor::IOProcessorNumber()` to perform MPI_Reduce instead of MPI_Allreduce,  so that only the I/O process has the reduction result.

### DIFF
--- a/Source/Diagnostics/ReducedDiags/ParticleExtrema.cpp
+++ b/Source/Diagnostics/ReducedDiags/ParticleExtrema.cpp
@@ -403,6 +403,11 @@ void ParticleExtrema::ComputeDiags (int step)
                 const MultiFab & By = warpx.getBfield(lev,1);
                 const MultiFab & Bz = warpx.getBfield(lev,2);
 
+                // declare reduce_op
+                ReduceOps<ReduceOpMin, ReduceOpMax> reduce_op;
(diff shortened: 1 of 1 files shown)


WeiqunZhang (approved):

NeilZaim:
Many thanks for finding this bug and for the fix!

I'm interested because I have used this diag in the past and it seemed to be producing correct results.
From what I understand, the old code would produce wrong results when there are multiple boxes/tiles per MPI rank in the `WarpXParIter` loop because in this case the values inside the `chimin` and `chimax` vectors are rewritten at each iteration of the loop, meaning that we only keep the min and max values in the last box/tile of the iteration.
This would mean that the old code still gave the right result when using GPU (no tiling) + `warpx.numprocs` for domain decomposition (one box per MPI). Is that correct?

WeiqunZhang:
Yes, I believe so.

EZoni (approved):
Thanks for the fix!

PR #4102, opened 2023-07-21
Clang tidy CI test: add cppcoreguidelines-init-variables check to clang tidy CI test

lucafedeli88:
This PR adds the following check to clang-tidy CI test:

- [cppcoreguidelines-init-variables](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/cppcoreguidelines-init-variables.html)


ax3l:
cc @WeiqunZhang 

I have the following question: yes, init on definition is the right thing to do.
But in HPC there are a few cases, where we immediately overwrite values.

Do compilers have an optimization pass to omit a prior write to a variable, if they find the value is immediately overwritten?

WeiqunZhang:
The compiler probably will optimize it away. But they are also equally capable of warning us about a variable may be used uninitialized. My opinion is it's very misleading and messy if we have to initialize everything. Since we have `-Wall -Wextra ...` set in the build system and we use `-Wall` in CIs, this clang-tidy check is arguably unnecessary.

lucafedeli88:
We will not enforce this check.

PR #4103, opened 2023-07-21
Update inputs files in WarpX

WeiqunZhang:
Due to the recent changes in amrex::ParmParse, we must use the line continuation character '\' for values spanning multiple lines. For Parser expression, we could either double quotes to enclose the whole string or use `\` as line continuation.

The old style multi-line values were removed in AMReX to avoid mistakes like
```

    # galilean
    psatd.use_default_v_galilean  # Forgot = 1

    # multi-line
    #   preferred
    some.function(x,y,z) = "x*x + sin(y)
        + z*z"
    #   old-school
    some.other(x,y,z) = x*x + sin(y)  \
        + z*z
```
Here, `psatd.use_default_v_galilean` is silently ignored because with the old style multi-line support it becomes part of the values for the previous definition (i.e., `algo.current_deposition = direct psatd.use_default_v_galilean`). This case will now abort.

### DIFF
--- a/Examples/Tests/electrostatic_sphere_eb/inputs_3d
+++ b/Examples/Tests/electrostatic_sphere_eb/inputs_3d
@@ -33,5 +33,5 @@ eb_charge.type = ChargeOnEB
 eb_charge.intervals = 1
 eb_charge_one_eighth.type = ChargeOnEB
 eb_charge_one_eighth.intervals = 1
-// Select only one eighth of the sphere
+# Select only one eighth of the sphere
 eb_charge_one_eighth.weighting_function(x,y,z) = (x>0)*(y<0)*(z>0)
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -110,7 +110,7 @@ jobs:
         which nvcc || echo "nvcc not in PATH!"
 
         git clone https://github.com/AMReX-Codes/amrex.git ../amrex
-        cd ../amrex && git checkout --detach 10b6cb26d0ac359922276ce08eeb728a14624b70 && cd -
+        cd ../amrex && git checkout --detach 587542a395e4bd1991c3e731b289ca89fe43746e && cd -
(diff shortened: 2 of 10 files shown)


Examples/Tests/electrostatic_sphere_eb/inputs_3d
```diff
@@ -33,5 +33,5 @@ eb_charge.type = ChargeOnEB
 eb_charge.intervals = 1
 eb_charge_one_eighth.type = ChargeOnEB
 eb_charge_one_eighth.intervals = 1
-// Select only one eighth of the sphere
```
atmyers, line 36: What will happen if someone uses this comment syntax now?

WeiqunZhang: It's a runtime error now. It used to be that `eb_charge_one_eighth.intervals` has multiple values, `1`, `//`, `Select`, etc.

atmyers (approved):

ax3l:
Discussed today in dev meeting, @jlvay & @ax3l might desire:
- trimming whitespaces after `\` if possible (users forgot)
- allowing comments to the right?
```
some.func(x,y,z) = "
   x * x      # something
   + sqrt(y)  # important comment
   - z        # here is part three
"
```
if it that is easily possible?

WeiqunZhang:
Trimming white spaces after `\` is fine. Allowing `#` after line continuation is also fine. But allowing comments inside a pair of double quotes is problematic, because unlike lines continuation `\`, everything inside double quotes are considered a *single* string that contains `\n`s. We would need to escape `#` like `\#` when we want to have `#` in a string, unless we say `#` is not allowed as part of string at all.

dpgrote:
My vote would be to allow white space and comments after `\`, but not allow comments in double quoted strings (since there doesn't need to be two ways to do the same thing). Also, this avoid the confusion of what happens if there is a double quote in a comment in a double quoted string.

PR #4104, opened 2023-07-21
Refactoring: instantiate WarpX class via MakeWarpX function

lucafedeli88:
Before calling the WarpX constructor, we need to call several other functions. In order to clean `main.cpp` and following @WeiqunZhang 's suggestion, these functions, as well as the actual call to the WarpX constructor, are grouped in a new method of the WarpX class:
```c++
void WarpX::MakeWarpX ()
{
    ParseGeometryInput();

    ConvertLabParamsToBoost();
    ReadBCParams();

#ifdef WARPX_DIM_RZ
    CheckGriddingForRZSpectral();
#endif

    m_instance = new WarpX();
}
```

`m_instance` is, as it was before, a pointer to the unique WarpX instance, and the `GetInstance` method returns a reference to this instance:
```c++
WarpX&
WarpX::GetInstance ()
{
    if (!m_instance) {
        MakeWarpX();
    }
    return *m_instance;
}
```
The WarpX constructor is now private, as well as the `MakeWarpX ` function. Therefore, calling `GetInstance()` is now the only way to instantiate a WarpX object. 

I had to move the following two lines from the WarpX constructor to the `InitData` member function, since internally they use `WarpX::GetInstance()`, but the instance is not ready until the constructor has been executed and the unique pointer has been created.

```c++
    // Diagnostics
    multi_diags = std::make_unique<MultiDiagnostics>();

    /** create object for reduced diagnostics */
    reduced_diags = std::make_unique<MultiReducedDiags>();
```

For the same reason, I had to slightly modify the `PlasmaInjector` class, in order to avoid calls to `WarpX::GetInstance()`.

I think that, as of today, we abuse a little bit of the `WarpX::GetInstance()`. We often use it to manipulate **a lot** of global variables, while passing just the right variables as function parameters would sometimes be preferable. However, this issue is too complex to address in a single PR.

Finally, I added a new method to call at the end of the simulation to delete the WarpX instance:
```c++
void WarpX::Finalize()
{
    WarpX::ResetInstance();
}
```
where
```c++
void
WarpX::ResetInstance ()
{
    if (m_instance){
        delete m_instance;
        m_instance = nullptr;
    }
}
```


ax3l:
@lucafedeli88 sorry, needs little rebase :)

ax3l (approved):
I like this, that makes sense and cleans the `m̀ain.cpp` to a level where I like to read it again :)

WeiqunZhang (approved):

ax3l:
@lucafedeli88 I wonder if `WarpX::~WarpX` should also call `ResetInstance` to clean out the static instance member?

PR #4105, opened 2023-07-21
Docs: Fix timestep duration plot script download link

n01r:
Previously, this just was a relative link to the file which could not be resolved. Now it downloads the plot script.


ax3l (approved):
Cool! Thx!

Issue #4035, opened 2023-06-23
Only allocates `rho_ncomps = ncomps` instead of `rho_ncomps = 2*ncomps` for ES, MS and hybrid-PIC solvers


roelof-groenewald:
We could do a follow-up PR that only allocates `rho_ncomps = ncomps` instead of `rho_ncomps = 2*ncomps` in these cases. (See a few lines below.)

_Originally posted by @RemiLehe in https://github.com/ECP-WarpX/WarpX/pull/3665#discussion_r1227202244_

PR #4114, opened 2023-07-24
Fix the writing of raw rho_fp

dpgrote:
PR #4035 changed the number of components in the rho_fp MultiFab, but the writing of rho_fp raw fields was assuming that the number of components was 2*ncomps. This PR fixes this and writes out the correct number of components.


lucafedeli88 (approved):
Thanks for this PR!

PR #4116, opened 2023-07-25
Include `J` in diagnostic output when an electromagnetic solver is not used

roelof-groenewald:
Currently if no field solver or the electrostatic solver used, the current density is not deposited so the field diagnostic does not show the actual current density values (just zeros). In this PR the current density is deposited when field diagnostics are output in such cases.


Source/Diagnostics/ComputeDiagFunctors/JFunctor.cpp
```diff
@@ -0,0 +1,89 @@
...
+#include "Utils/WarpXAlgorithmSelection.H"
+#include "WarpX.H"
+
+#include <ablastr/coarsen/sample.H>
+
+#include <AMReX.H>
+#include <AMReX_IntVect.H>
+#include <AMReX_MultiFab.H>
+
+JFunctor::JFunctor(const int dir, int lev,
```
EZoni:
```suggestion
JFunctor::JFunctor (const int dir, int lev,
```

Source/Diagnostics/ComputeDiagFunctors/JFunctor.cpp
```diff
@@ -0,0 +1,89 @@
...
+JFunctor::operator()(amrex::MultiFab& mf_dst, int dcomp, const int /*i_buffer*/) const
+{
+    auto& warpx = WarpX::GetInstance();
+    /** pointer to source multifab (can be multi-component) */
+    amrex::MultiFab* m_mf_src = warpx.get_pointer_current_fp(m_lev, m_dir);
+
+    // Deposit current if no solver or the electrostatic solver is being used
+    if ( WarpX::electromagnetic_solver_id == ElectromagneticSolverAlgo::None &&
+        WarpX::electrostatic_solver_id != ElectrostaticSolverAlgo::LabFrameElectroMagnetostatic
+    ) {
```
EZoni: Just alignment adjustments:
```suggestion
    if (WarpX::electromagnetic_solver_id == ElectromagneticSolverAlgo::None &&
        WarpX::electrostatic_solver_id != ElectrostaticSolverAlgo::LabFrameElectroMagnetostatic)
    {
```

Source/Diagnostics/ComputeDiagFunctors/JFunctor.cpp
```diff
@@ -0,0 +1,89 @@
...
+        auto& current_fp_y = warpx.getcurrent_fp(m_lev,1);
+        current_fp_temp[0][1] = std::make_unique<amrex::MultiFab>(
+            current_fp_y.boxArray(), current_fp_y.DistributionMap(),
+            current_fp_y.nComp(), current_fp_y.nGrow()
+        );
+        auto& current_fp_z = warpx.getcurrent_fp(m_lev,2);
+        current_fp_temp[0][2] = std::make_unique<amrex::MultiFab>(
+            current_fp_z.boxArray(), current_fp_z.DistributionMap(),
+            current_fp_z.nComp(), current_fp_z.nGrow()
+        );
```
EZoni: Shall we use nGrowVect to take into account different number of ghost cells in different directions?
```suggestion
        auto& current_fp_x = warpx.getcurrent_fp(m_lev,0);
        current_fp_temp[0][0] = std::make_unique<amrex::MultiFab>(
            current_fp_x.boxArray(), current_fp_x.DistributionMap(),
            current_fp_x.nComp(), current_fp_x.nGrowVect()
        );
        auto& current_fp_y = warpx.getcurrent_fp(m_lev,1);
        current_fp_temp[0][1] = std::make_unique<amrex::MultiFab>(
            current_fp_y.boxArray(), current_fp_y.DistributionMap(),
            current_fp_y.nComp(), current_fp_y.nGrowVect()
        );
        auto& current_fp_z = warpx.getcurrent_fp(m_lev,2);
        current_fp_temp[0][2] = std::make_unique<amrex::MultiFab>(
            current_fp_z.boxArray(), current_fp_z.DistributionMap(),
            current_fp_z.nComp(), current_fp_z.nGrowVect()
        );
```

Source/Diagnostics/ComputeDiagFunctors/JFunctor.cpp
```diff
@@ -0,0 +1,89 @@
...
+        ablastr::coarsen::sample::Coarsen( mf_dst, mf_dst_stag, dcomp, 0, nComp(), 0,  m_crse_ratio);
+    } else {
+        ablastr::coarsen::sample::Coarsen( mf_dst, *m_mf_src, dcomp, 0, nComp(), 0, m_crse_ratio);
+    }
+#else
+    // In cartesian geometry, coarsen and interpolate from simulation MultiFab, m_mf_src,
+    // to output diagnostic MultiFab, mf_dst.
+    ablastr::coarsen::sample::Coarsen(mf_dst, *m_mf_src, dcomp, 0, nComp(), mf_dst.nGrowVect(), m_crse_ratio);
+    amrex::ignore_unused(m_lev, m_convertRZmodes2cartesian);
+#endif
```
EZoni: This code is in common with `RhoFunctor::operator()`, provided that one replaces `rho` with `m_mf_src`.
I wonder if it could be possible to put this into a common inline function and avoid code duplication?

Source/Diagnostics/ComputeDiagFunctors/JFunctor.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+JFunctor::JFunctor(const int dir, int lev,
+                   amrex::IntVect crse_ratio,
+                   bool convertRZmodes2cartesian, int ncomp)
+    : ComputeDiagFunctor(ncomp, crse_ratio), m_dir(dir), m_lev(lev),
+      m_convertRZmodes2cartesian(convertRZmodes2cartesian)
+{ }
+
+void
+JFunctor::operator()(amrex::MultiFab& mf_dst, int dcomp, const int /*i_buffer*/) const
```
EZoni:
```suggestion
JFunctor::operator() (amrex::MultiFab& mf_dst, int dcomp, const int /*i_buffer*/) const
```

Source/Diagnostics/ComputeDiagFunctors/JFunctor.H
```diff
@@ -0,0 +1,47 @@
...
+     *
+     * \param[in] mf_src source multifab.
+     * \param[in] lev level of multifab. Used for averaging in rz.
+     * \param[in] crse_ratio for interpolating field values from the simulation MultiFab, src_mf,
+                  to the output diagnostic MultiFab, mf_dst.
+     * \param[in] convertRZmodes2cartesian (in cylindrical) whether to
+     *            sum all modes in mf_src before cell-centering into dst multifab.
+     * \param[in] ncomp Number of component of mf_src to cell-center in dst multifab.
+     */
+    JFunctor(const int dir, const int lev,
```
EZoni: We tend to write `Function (...)` rather than `Function(...)` (extra white space between function's name and parenthesis) in function definitions, in order to make "git grepping" easier:
```suggestion
    JFunctor (const int dir, const int lev,
```

Source/Diagnostics/ComputeDiagFunctors/JFunctor.H
```diff
@@ -0,0 +1,47 @@
...
+public:
+    /** Constructor.
+     *
+     * \param[in] mf_src source multifab.
+     * \param[in] lev level of multifab. Used for averaging in rz.
+     * \param[in] crse_ratio for interpolating field values from the simulation MultiFab, src_mf,
+                  to the output diagnostic MultiFab, mf_dst.
+     * \param[in] convertRZmodes2cartesian (in cylindrical) whether to
+     *            sum all modes in mf_src before cell-centering into dst multifab.
+     * \param[in] ncomp Number of component of mf_src to cell-center in dst multifab.
```
EZoni: The Doxygen might need to be updated.

Source/Diagnostics/ComputeDiagFunctors/JFunctor.H
```diff
@@ -0,0 +1,47 @@
...
+     *
+     * \param[out] mf_dst output MultiFab where the result is written
+     * \param[in] dcomp first component of mf_dst in which cell-centered
+     *            data is stored
+     */
+    virtual void operator()(amrex::MultiFab& mf_dst, int dcomp, const int /*i_buffer=0*/) const override;
+private:
+    /** direction of the current density to save */
+    const int m_dir;
+    /** level on which mf_src is defined (used in cylindrical) */
```
EZoni:
```suggestion
    /** level on which mf_src is defined */
```

EZoni (commented):
Thank you for the PR, @roelof-groenewald! I left a few minor comments/questions.

Source/Diagnostics/ComputeDiagFunctors/JFunctor.cpp
roelof-groenewald: I'm a big fan of reducing code duplication so thanks for pointing this out. I'll take a closer look and make the needed modifications.

roelof-groenewald: I was able to move the code from these parts into the parent class, let me know if you spot more opportunity for removing code duplication.

roelof-groenewald:
As a follow-up to this PR it would be great to implement a `mypc.GetCurrentDensity()` function similar to the one for charge density and then allow users to specify per species current density to be saved the same way that per species `rho` can currently be saved.

Regression/Checksum/benchmarks_json/Python_prev_positions.json
```diff
@@ -1,23 +1,23 @@
...
+    "Ez": 646727.8074440088,
+    "jx": 4745.078379617619,
+    "jy": 368.4331779923921,
+    "jz": 632.1508106460103
+  },
   "electrons": {
     "particle_momentum_x": 8.78764082600202e-23,
     "particle_momentum_y": 3.78128494181519e-24,
-    "particle_momentum_z": 5.396907550474659e-24,
+    "particle_momentum_z": 5.3969075504746586e-24,
```
dpgrote, line 16: Do you understand why the position was affected?

Regression/Checksum/benchmarks_json/hard_edged_quadrupoles_moving.json
```diff
@@ -1,22 +1,22 @@
...
-    "jz": 0.0
+    "Ex": 6.0282565190090465e-05,
+    "Ey": 6.38479659567398e-05,
+    "Ez": 7.880459213065183e-05,
+    "jx": 1.1659465170956616e-09,
+    "jy": 2.6115239381823616e-17,
+    "jz": 1.5293971084510288e-07
+  },
+  "electron": {
+    "particle_momentum_x": 2.0819392998019267e-25,
```
dpgrote, line 14: Same question here - do you know why the particle quantities changed? (though only in the last two digits)

Source/Diagnostics/ComputeDiagFunctors/CellCenterFunctor.cpp
```diff
@@ -21,29 +16,7 @@ CellCenterFunctor::CellCenterFunctor(amrex::MultiFab const * mf_src, int lev,
...
-        ablastr::coarsen::sample::Coarsen( mf_dst, *m_mf_src, dcomp, 0, nComp(), 0, m_crse_ratio);
-    }
-#else
-    // In cartesian geometry, coarsen and interpolate from simulation MultiFab, m_mf_src,
-    // to output diagnostic MultiFab, mf_dst.
-    ablastr::coarsen::sample::Coarsen(mf_dst, *m_mf_src, dcomp, 0, nComp(), mf_dst.nGrowVect(), m_crse_ratio);
-    amrex::ignore_unused(m_lev, m_convertRZmodes2cartesian);
-#endif
+    auto& warpx = WarpX::GetInstance();
+    InterpolateToDst(mf_dst, *m_mf_src, dcomp, warpx.DistributionMap(m_lev),
```
dpgrote: Can a more descriptive name be used for the new function? `InterpolateMFForDiag`?

Source/Diagnostics/ComputeDiagFunctors/JFunctor.cpp
```diff
@@ -0,0 +1,64 @@
...
+            current_fp_y.nComp(), current_fp_y.nGrowVect()
+        );
+        auto& current_fp_z = warpx.getcurrent_fp(m_lev,2);
+        current_fp_temp[0][2] = std::make_unique<amrex::MultiFab>(
+            current_fp_z.boxArray(), current_fp_z.DistributionMap(),
+            current_fp_z.nComp(), current_fp_z.nGrowVect()
+        );
+
+        auto& mypc = warpx.GetPartContainer();
+        mypc.DepositCurrent(current_fp_temp, warpx.getdt(m_lev), 0.0);
```
dpgrote, line 54: This will do the full deposition in all directions even though only one will be used, which means that the deposition will be done three times to write out the `jx`, `jy`, and `jz`. Is there a way to do the deposition only once each time the diagnostic is done?

Regression/Checksum/benchmarks_json/Python_prev_positions.json
roelof-groenewald: This I would think comes from a difference in the number of digits displayed in the new'ish printing of the updated benchmark values in the Azure output (from which I copied the values to the json files) compared to the digits included in the json writing when the benchmark reset functionality is used.

Source/Diagnostics/ComputeDiagFunctors/JFunctor.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+        // copy the deposited current into the src mf
+        amrex::MultiFab::Copy(*m_mf_src, *current_fp_temp[0][m_dir], 0, 0, 1, m_mf_src->nGrowVect());
```
dpgrote: Since this overwrites `m_mf_src` anyway, can the deposition be done directly to it rather than creating the temporary and copying?

Source/Diagnostics/ComputeDiagFunctors/CellCenterFunctor.cpp
roelof-groenewald: Sure, I'll change that.

dpgrote:
I left some comments. Also, note that the current density per species can be calculated by doing this

```
diag.particle_fields_to_plot = jz
diag.particle_fields.jz(x,y,z,ux,uy,uz) = uz*q_e
diag.particle_fields.jz.do_average = 0
```

Though this would only give mode 0 when in RZ.

Source/Diagnostics/ComputeDiagFunctors/JFunctor.cpp
roelof-groenewald: I'll think about this a bit to see if I can come up with a way to achieve that which doesn't deviate too much from how the other diagnostic quantities are handled.

Source/Diagnostics/ComputeDiagFunctors/JFunctor.cpp
roelof-groenewald: I'm actually not sure how to do that since `warpx.current_fp` is private, so I can't access it directly from `JFunctor`.

roelof-groenewald:
@dpgrote Thanks for the review and the tip on how to get the per species current within the current functionality.

Source/Diagnostics/ComputeDiagFunctors/JFunctor.cpp
roelof-groenewald: Thanks for the suggestion. I've updated the implementation to only deposit once per diagnostic output.

dpgrote:
What you have for controlling the current deposition looks like it'll work. Here's another option - there is already the routine `PrepareFieldDataForOutput`. You could make the flag `deposit_current` a member of the `Diagnostic` class and have it set only once based on the field solve. Then the routine `PrepareFieldDataForOutput` could call deposit current if needed, putting the results in `current_fp`. Actually, if you do it this way, the `JFunctor` is not needed at all. Note that the same could be done with rho as well.

roelof-groenewald:
That wouldn't work for `rho` since it needs to be separately deposited for the different species (if `rho_species` is specified as a diagnostic, right?). I'd also still like to implement the same for `J` at some point since the `particle_field` approach does not account for the particle shape, so the given fields wouldn't be the same as the ones used in the field solves.

Source/Diagnostics/ComputeDiagFunctors/JFunctor.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+    // Deposit current if no solver or the electrostatic solver is being used
+    if (m_deposit_current)
+    {
+        // allocate temporary multifab to deposit current density into
+        amrex::Vector<std::array< std::unique_ptr<amrex::MultiFab>, 3 > > current_fp_temp;
+        current_fp_temp.resize(1);
+
+        auto& current_fp_x = warpx.getcurrent_fp(m_lev,0);
+        current_fp_temp[0][0] = std::make_unique<amrex::MultiFab>(
```
dpgrote, line 40: This should work instead of creating temporary MultiFabs and then copying.
```
current_fp_temp[0][0] = std::make_unique<amrex::MultiFab>(current_fp_x, amrex::make_alias, 0, 1);
```
I tested this and it at least compiles.

roelof-groenewald: Thanks! I didn't know about the `make_alias` functionality. That is great to learn about and I'm sure it'll come in handy again in the future.

Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -192,6 +193,13 @@ FullDiagnostics::InitializeFieldFunctorsRZopenPMD (int lev)
     m_all_field_functors[lev].clear();
     m_all_field_functors[lev].resize(m_varnames_fields.size());
 
+    // Boolean flag for whether the current density should be deposited before
+    // diagnostic output
+    bool deposit_current = (
```
dpgrote: I can approve this as it is, but it would be nice if this bit of duplicated code could be avoided. Maybe create a class parameter like `solver_deposits_current` that is set in the constructor. Then this variable is set to `bool deposit_current = !solver_deposits_current`.

dpgrote (approved):
This looks good! Thanks for making the changes (and sorry I have one more suggestion :) ).

Source/Diagnostics/FullDiagnostics.cpp
roelof-groenewald: That's a good point since if we add more solvers that do not deposit current we will now only need to update this flag in one place. Thanks for the suggestion!

PR #4117, opened 2023-07-25
Initialize variables at declaration if it improves readability

lucafedeli88:
With this PR variables are initialized when they are declared when this is beneficial for readability.


Source/Diagnostics/ComputeDiagFunctors/ParticleReductionFunctor.cpp
```diff
@@ -93,9 +93,8 @@ ParticleReductionFunctor::operator() (amrex::MultiFab& mf_dst, const int dcomp,
                 const amrex::ParticleReal ux = p.rdata(PIdx::ux) / PhysConst::c;
                 const amrex::ParticleReal uy = p.rdata(PIdx::uy) / PhysConst::c;
                 const amrex::ParticleReal uz = p.rdata(PIdx::uz) / PhysConst::c;
-                amrex::Real value;
-                if ((do_filter) && (!filter_fn(xw, yw, zw, ux, uy, uz))) value = 0._rt;
-                else value = map_fn(xw, yw, zw, ux, uy, uz);
+                const bool filtered_out_flag = ((do_filter) && (filter_fn(xw, yw, zw, ux, uy, uz) <= 0.0));
```
WeiqunZhang:
```suggestion
                const bool filtered_out_flag = ((do_filter) && (filter_fn(xw, yw, zw, ux, uy, uz) == 0.0_prt));
```

Source/Initialization/WarpXInitData.cpp
```diff
@@ -636,13 +636,9 @@ WarpX::computeMaxStepBoostAccelerator() {
...
         (moving_window_v-v_plasma_boost);
     // Divide by dt, and update value of max_step.
-    int computed_max_step;
-    if (do_subcycling){
-        computed_max_step = static_cast<int>(interaction_time_boost/dt[0]);
-    } else {
-        computed_max_step =
-            static_cast<int>(interaction_time_boost/dt[maxLevel()]);
-    }
+    const int computed_max_step = (do_subcycling)?
```
WeiqunZhang:
```suggestion
    const auto computed_max_step = (do_subcycling)?
```
I don't remember which clang-tidy check. But one of them will want this to be `auto`.

Source/Utils/Algorithms/UpperBound.H
```diff
@@ -25,12 +25,10 @@ namespace utils::algorithms
     template<typename T> AMREX_GPU_DEVICE AMREX_FORCE_INLINE
     const T* upper_bound(const T* first, const T* last, const T& val)
     {
-        const T* it;
-        size_t count, step;
-        count = last-first;
+        size_t count = last-first;
```
WeiqunZhang: We could use `amrex::upper_bound` you put into amrex.

Source/Initialization/WarpXInitData.cpp
lucafedeli88: Possibly the `modernize-use-auto` . Thanks for the suggestion!

Source/Diagnostics/ComputeDiagFunctors/ParticleReductionFunctor.cpp
lucafedeli88: I am not totally sure about this. `filter_fn` is  an `amrex::ParserExecutor`. Doesn't the `()` operator return a `double` ? I think that, if this is the case, it would be more appropriate to do `== 0.0`.

I wrote `<= 0.0` because I think that this is what we should do with the filter function, but `== 0.0` actually does what we write in the documentation.

Source/Utils/Algorithms/UpperBound.H
lucafedeli88: Actually I've checked and the algorithm is no longer used! Moreover, since it is already implemented in AMReX, I have decided to remove it: https://github.com/ECP-WarpX/WarpX/pull/4123

Source/Diagnostics/ComputeDiagFunctors/ParticleReductionFunctor.cpp
WeiqunZhang: The return type is the same as the arguments (ParticleReal in this case).

lucafedeli88: Ah, I see.. I stopped at the first definition. I didn't see the one with `std::enable_if_t<sizeof...(Ts) == N &&  amrex::Same<float,Ts...>::value, float>` . You are totally right

lucafedeli88:
```suggestion
                const bool filtered_out_flag = ((do_filter) && (filter_fn(xw, yw, zw, ux, uy, uz) == 0.0_prt));
```

WeiqunZhang (approved):

PR #4120, opened 2023-07-25
Tools: Update Copyright String

ax3l:
Use a modern, short version referring our license model and keep per-file authorship notes.

This tool uses the git history, so might include more people than still present.
The alternative, git blame, might exclude still present authors due to reformatting changes.


lucafedeli88 (approved):
Thanks @ax3l !

PR #4125, opened 2023-07-26
Add DSMC module

roelof-groenewald:
A direct simulation Monte Carlo (DSMC) module is added to handle non-Coulomb binary collisions between particles i.e. charge exchange, elastic scattering, etc.

Currently only non-particle producing collisions are supported (i.e. not things like dissociation or ionization). I'm thinking those will be added in a future PR.

Still to do for this PR:
- [x] Add benchmark / CI test

Other changes:

- Renamed `MCCProcess` to `ScatteringProcess` and moved the functionality to the collisions base folder to allow the functionality to read cross-sections from file to be re-used in other collision models.
- Added `get_collision_parameters()` to binary collision utils since that functionality is also needed for different collision models.


roelof-groenewald:
As a test of the DSMC collision logic I performed the same benchmark test as used for the MCC module (see [here](https://warpx.readthedocs.io/en/latest/usage/examples.html#capacitive-discharge)) but using DSMC collisions for the ions. In the DSMC case I reset the kinetic neutral particles' temperature periodically since the MCC test does not have an evolving neutral species.
This test shows the collision probabilities and scattering physics is likely correct.
benchmark

As a check of how sensitive the above test is to the collision frequency, I also ran a case where I artificially lowered the collision frequency by 20%. This showed a significant change in the final ion density, indicating that the test is sensitive to the exact collision parameters.
image

Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.cpp
```diff
@@ -25,16 +25,16 @@ ParticleCreationFunc::ParticleCreationFunc (const std::string collision_name,
...
-                m_num_products_host.push_back(3);
+        {
+            // Proton-Boron fusion only produces alpha particles
+            m_num_product_species = 1;
+            // Proton-Boron fusion produces 3 alpha particles per fusion reaction
+            m_num_products_host.push_back(3);
 #ifndef AMREX_USE_GPU
-                // On CPU, the device vector can be filled immediately
-                m_num_products_device.push_back(3);
+            // On CPU, the device vector can be filled immediatly
```
roelof-groenewald:
```suggestion
            // On CPU, the device vector can be filled immediately
```

Docs/source/theory/collisions.rst
```diff
@@ -3,18 +3,36 @@
 Collisions
 ==========
 
-Monte Carlo Collisions
-----------------------
+WarpX includes several different models to capture collisional processes
+including collisions between kinetic particles (Coulomb collisions, DSMC,
+nuclear fusion) as well as collisions between kinetic particles and a fluid
```
RemiLehe:
```suggestion
nuclear fusion) as well as collisions between kinetic particles and a fixed (i.e. non-evolving)
```
The word `fluid` here might be confusing ; some might understand this as interacting with the hybrid fluid-kinetic solver, or the fluid species in WarpX.

Docs/source/theory/collisions.rst
```diff
@@ -3,18 +3,36 @@
...
-Several types of collisions between simulation particles and a neutral background gas are supported including elastic scattering, back scattering, charge exchange, excitation collisions and impact ionization.
-An instance of the class :cpp:class:`MCCProcess` is created for each type of collision included in a simulation. This class saves information about the type of collision and the collision cross-section as a function of energy.
+.. _theory-collisions-mcc:
 
-The so-called null collision strategy is used in order to minimize the computational burden of the MCC module.
-This strategy is standard in PIC-MCC and a detailed description can be found elsewhere, for example in :cite:t:`b-Birdsall1991`.
-In short the maximum collision probability is found over a sensible range of energies and is used to pre-select the appropriate number of macroparticles for collision consideration. Only these pre-selected particles are then individually considered for a collision based on their energy and the cross-sections of all the different collisional processes included.
+Monte Carlo Collisions
+----------------------
```
RemiLehe:
```suggestion
Monte Carlo Collisions (MCC)
------------------------------
```

Docs/source/theory/collisions.rst
```diff
@@ -23,19 +41,44 @@ The MCC implementation assumes that the background neutral particles are **therm
...
+expression reduces to the classical equation
+:math:`E_{coll} = \frac{1}{2}\frac{Mm}{M+m} u^2`. The collision cross-sections
+for all scattering processes are evaluated at the energy as calculated above.
 
 Once a particle is selected for a specific collision process, that process determines how the particle is scattered as outlined below.
 
+.. _theory-collisions-dsmc:
+
+Direct Simulation Monte Carlo
+-----------------------------
```
RemiLehe:
```suggestion
Direct Simulation Monte Carlo (DSMC)
--------------------------------------
```

Docs/source/theory/collisions.rst
```diff
@@ -23,19 +41,44 @@ The MCC implementation assumes that the background neutral particles are **therm
...
+
+(Under construction)
```
RemiLehe: Could you add some text here (even if it is short) to explain the approach?

Docs/source/theory/collisions.rst
```diff
@@ -23,19 +41,44 @@ The MCC implementation assumes that the background neutral particles are **therm
...
+with the neutral velocity and vice-versa.
 
 Elastic scattering
 ^^^^^^^^^^^^^^^^^^
 
-This scattering process as well as the ones below that relate to it, are all performed in the center-of-momentum (COM) frame. Designating the COM velocity of the particle as :math:`\vec{u}_c` and its labframe velocity as :math:`\vec{u}_l`, the transformation from lab frame to COM frame is done with a general Lorentz boost (see function ``ParticleUtils::doLorentzTransform()``):
+The ``elastic`` option uses hard-sphere scattering, with a differential
+cross section that is independent of angle.
+This scattering process as well as the ones below that relate to it, are all
+performed in the center-of-momentum (COM) frame. Designating the COM velocity of
```
RemiLehe, line 89: When we use the MCC framework to perform elastic scattering, this text seems in contradiction with the paragraph saying "The particle velocity is then boosted to the stationary frame of the neutral through a Galilean transformation."

Docs/source/theory/collisions.rst
```diff
@@ -74,6 +117,12 @@ Excitation
 
 The process is also the same as for elastic scattering except the excitation energy cost is subtracted from the particle energy. This is done by reducing the velocity before a scattering angle is chosen.
 
+Benchmarks
+----------
+
+See the :ref:`MCC example <examples-mcc-turner>` for a bencmark of the MCC
```
RemiLehe:
```suggestion
See the :ref:`MCC example <examples-mcc-turner>` for a benchmark of the MCC
```

Docs/source/theory/collisions.rst
RemiLehe: Actually, should this be "Background Monte Carlo Collisions" (background MCC) ?

Docs/source/theory/collisions.rst
```diff
@@ -23,19 +41,44 @@ The MCC implementation assumes that the background neutral particles are **therm
...
-This is the simplest scattering process. Under charge exchange the simulation particle's velocity is simply replaced with the sampled velocity of the neutral particle. Charge exchange usually has a cooling effect on the ions in the simulation by exchanging energetic ions for lower energy neutrals.
+This process can occur when an ion and neutral (of the same species) collide
+and results in the exchange of an electron. The ion velocity is simply replaced
+with the neutral velocity and vice-versa.
 
 Elastic scattering
 ^^^^^^^^^^^^^^^^^^
 
-This scattering process as well as the ones below that relate to it, are all performed in the center-of-momentum (COM) frame. Designating the COM velocity of the particle as :math:`\vec{u}_c` and its labframe velocity as :math:`\vec{u}_l`, the transformation from lab frame to COM frame is done with a general Lorentz boost (see function ``ParticleUtils::doLorentzTransform()``):
+The ``elastic`` option uses hard-sphere scattering, with a differential
```
RemiLehe: I am not sure I understand the term `hard-sphere` in this context. 
Should we simply say that the scattering is isotropic, i.e. has a differential cross-section that is independent of the angle.

Examples/Physics_applications/capacitive_discharge/PICMI_inputs_1d.py
```diff
@@ -269,24 +285,26 @@ def setup_run(self):
...
-                # }
-            }
-        )
+        ion_scattering_processes={
+            'elastic': {'cross_section': cross_sec_direc+'ion_scattering.dat'},
+            'back': {'cross_section': cross_sec_direc+'ion_back_scatter.dat'},
+            # 'charge_exchange': {'cross_section': cross_sec_direc+'charge_exchange.dat'}
+        }
+        if self.dsmc:
+            ion_colls = picmi.DSMCCollisions(
```
RemiLehe, line 294: Could you add this in the Python section of the WarpX documentation?

Source/Particles/Collision/BinaryCollision/BinaryCollisionUtils.H
```diff
@@ -36,6 +38,89 @@ namespace BinaryCollisionUtils{
                                       MultiParticleContainer const * mypc);
 
     CollisionType nuclear_fusion_type_to_collision_type (NuclearFusionType fusion_type);
+
+    /**
+     * \brief Return (relativistic) collision energy, collision speed and
+     * Lorentz factor for transforming between the lab and center-of-momentum
+     * frames. Note the use of `double` for energy since this calculation is
```
RemiLehe: This says that the energy is `double`, but I see `amrex::ParticleReal& E_coll`. Did I misunderstood?

Also: for "collision energy", could the documentation be more precise and specify: "Kinetic energy of the incident colliding particles in the center of mass frame". 
Same thing for "collision speed": "relative speed in the center-of-mass frame"

Source/Particles/Collision/BinaryCollision/BinaryCollisionUtils.H
```diff
@@ -36,6 +38,89 @@ namespace BinaryCollisionUtils{
...
+     * prone to error with single precision.
+     */
+    AMREX_GPU_HOST_DEVICE AMREX_INLINE
+    void get_collision_parameters (
+        const amrex::ParticleReal& u1x, const amrex::ParticleReal& u1y,
+        const amrex::ParticleReal& u1z, const amrex::ParticleReal& u2x,
+        const amrex::ParticleReal& u2y, const amrex::ParticleReal& u2z,
+        const amrex::ParticleReal& m1, const amrex::ParticleReal& m2,
+        amrex::ParticleReal& E_coll, amrex::ParticleReal& v_coll,
+        amrex::ParticleReal& lab_to_COM_factor )
```
RemiLehe: Consistently with the above remarks, could the variables have more explicit names:
```suggestion
        amrex::ParticleReal& E_kin_COM, amrex::ParticleReal& v_rel_COM,
        amrex::ParticleReal& lab_to_COM_lorentz_factor )
```

Source/Particles/Collision/BinaryCollision/BinaryCollisionUtils.H
```diff
@@ -36,6 +38,89 @@ namespace BinaryCollisionUtils{
...
+        const auto g2_star = std::sqrt(one_pr + p_star_sq / static_cast<amrex::ParticleReal>(m2_sq*c_sq));
+
+        // relative velocity in the center of mass frame
+        v_coll = std::sqrt(p_star_sq) * (one_pr/(m1*g1_star) + one_pr/(m2*g2_star));
+
+        // Cross sections and relative velocity are computed in the center of mass frame.
+        // On the other hand, the particle densities (weight over volume) in the lab frame are used. To
+        // take into account this discrepancy, we need to multiply the fusion probability by the ratio
+        // between the Lorentz factors in the COM frame and the Lorentz factors in the lab frame
+        // (see Perez et al., Phys.Plasmas.19.083104 (2012))
```
RemiLehe: This comment should be moved to `SingleNuclearFusionEvent.H`

RemiLehe (changes requested):

Source/Particles/Collision/BinaryCollision/BinaryCollisionUtils.H
RemiLehe: Maybe the comment on `double` can be moved to the body of the function.

Source/Particles/Collision/BinaryCollision/DSMC/DSMC.H
```diff
@@ -0,0 +1,85 @@
...
+#include "Utils/ParticleUtils.H"
+#include "Utils/WarpXProfilerWrapper.H"
+
+#include <AMReX_DenseBins.H>
+#include <AMReX_ParmParse.H>
+#include <AMReX_Random.H>
+
+
+/**
+ * \brief This class perfoms DSMC (direct simulation Monte Carlo) collisions
```
RemiLehe:
```suggestion
 * \brief This class performs DSMC (direct simulation Monte Carlo) collisions
```

Source/Particles/Collision/BinaryCollision/DSMC/DSMC.cpp
```diff
@@ -0,0 +1,300 @@
...
+            {
+                amrex::Gpu::synchronize();
+                wt = static_cast<amrex::Real>(amrex::second()) - wt;
+                amrex::HostDevice::Atomic::Add( &(*cost)[mfi.index()], wt);
+            }
+        }
+
+        // Call redistribute to remove particles with negative ids
+        species1.Redistribute(lev, lev, 0, true, true);
+        species2.Redistribute(lev, lev, 0, true, true);
```
RemiLehe, line 128: Now I wonder if we should call `Redistribute` in the corresponding line of
https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/Collision/BinaryCollision/BinaryCollision.H#L199
(Since the above link points to code that may be used e.g. for fusion reactions, where particles [may get negative IDs](https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H#L207), as part of the call to the `copy_transform` functor, within `doCollisionsWithinTile`)

Docs/source/theory/collisions.rst
roelof-groenewald: There is actually two separate boosts in the collision process. First, we assume a non-relativistic neutral velocity which allows us to use a Galilean transformation to get to the rest frame of the neutral. In this frame the relativistic collision energy is easy to calculate with the function `ParticleUtils::getCollisionEnergy()`. This calculation happens very often and therefore is good to have a minimum number of FLOPS. Particles that are then chosen for a collision (presumably a relatively small number), is then Lorentz transformed to the COM frame (this therefore allows relativistic simulation particles) where the scattering happens. Does that clear things up?

Source/Particles/Collision/BinaryCollision/BinaryCollisionUtils.H
roelof-groenewald: I updated the comment to not mention fusion specifically, but the principle is generally true for all collision processes.

Source/Particles/Collision/BinaryCollision/DSMC/CollisionFilterFunc.H
```diff
@@ -0,0 +1,228 @@
...
+ *
+ * @param[in] u1x,u1y,u1z momenta of the first colliding particle
+ * @param[in] u2x,u2y,u2z momenta of the second colliding particle
+ * @param[in] m1,m2 masses
+ * @param[in] w1,w2 effective weight of the colliding particles
+ * @param[in] dt is the time step length between two collision calls.
+ * @param[in] dV is the volume of the corresponding cell.
+ * @param[in] pair_index is the index of the colliding pair
+ * @param[out] p_mask is a mask that will be set to true if fusion occurs for that pair
+ * @param[out] p_pair_reaction_weight stores the weight of the product particles
```
RemiLehe, line 30: Could you add documentation for `process_count` and `scattering_processes` ?

Docs/source/usage/parameters.rst
```diff
@@ -1799,22 +1800,17 @@ WarpX provides several particle collision models, using varying degrees of appro
...
     identifier for each excitation process. For each scattering process specified
-    a path to a cross-section data file must  also be given. We use
+    a path to a cross-section data file must also be given. We use
     ``<scattering_process>`` as a placeholder going forward.
 
 * ``<collision_name>.<scattering_process>_cross_section`` (`string`)
-    Only for ``background_mcc``. Path to the file containing cross-section data
+    Only for ``dsmc`` and ``background_mcc``. Path to the file containing cross-section data
     for the given scattering processes. The cross-section file must have exactly
     2 columns of data, the first containing equally spaced energies in eV and the
```
RemiLehe, line 1815: Here, could also define more specifically the definition of energies in the first column: is it the kinetic energy of the colliding particles *in the center-of-mass* frame?

Source/Particles/Collision/BinaryCollision/DSMC/CollisionFilterFunc.H
```diff
@@ -0,0 +1,228 @@
...
+
+    amrex::ParticleReal sigma_tot = 0._prt;
+    for (int ii = 0; ii < process_count; ii++) {
+        auto const& scattering_process = *(scattering_processes + ii);
+        sigma_tot += scattering_process.getCrossSection(E_coll);
+    }
+
+    // calculate total collision probability
+    amrex::ParticleReal exponent = (
+        lab_to_COM_factor * multiplier_ratio * w_max
```
RemiLehe, line 72: I think that the factor `lab_to_COM_factor` is not applied, in the corresponding code for the MCC module:
https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/Collision/BackgroundMCC/BackgroundMCCCollision.cpp#L398
Should we apply it?

Source/Particles/Collision/BinaryCollision/DSMC/CollisionFilterFunc.H
```diff
@@ -0,0 +1,228 @@
...
+ * @param[in] cell_start_pair is the start index of the pairs in that cell.
+ * @param[out] p_mask is a mask that will be set to true if a fusion event occurs for a given
+ * pair. It is only needed here to store information that will be used later on when actually
+ * creating the product particles.
+ * @param[out] p_pair_indices_1,p_pair_indices_2 arrays that store the indices of the
+ * particles of a given pair. They are only needed here to store information that will be used
+ * later on when actually creating the product particles.
+ * @param[out] p_pair_reaction_weight stores the weight of the product particles. It is only
+ * needed here to store information that will be used later on when actually creating the
+ * product particles.
```
RemiLehe, line 134: Could you add documentation for process_count and scattering_processes ?

Source/Particles/Collision/BinaryCollision/DSMC/DSMC.cpp
```diff
@@ -0,0 +1,300 @@
...
+    const int num_added = splitScatteringParticles(
+        n_total_pairs,
+        ptile_1, ptile_2,
+        p_mask,
+        copy_species1, copy_species2,
+        m1, m2,
+        p_pair_indices_1, p_pair_indices_2,
+        p_pair_reaction_weight);
+
+    if (num_added > 0) {
```
RemiLehe, line 296: Does the code below imply that we add `num_added` particles in both `ptile_1` and `ptile_2`?
I am not sure this corresponds with what I thought the algorithm was doing.

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -0,0 +1,168 @@
...
+#ifndef SPLIT_AND_SCATTER_FUNC_H_
+#define SPLIT_AND_SCATTER_FUNC_H_
+
+#include "Particles/Collision/ScatteringProcess.H"
+
+/**
+ * \brief Function that performs the particle scattering and injection due
+ * to binary collisions.
+ *
+ * \return num_added the number of particles that were written to dst.
```
RemiLehe: Here, there is no object called `dst`. Should this comment be updated?

Source/Particles/Collision/BinaryCollision/DSMC/CollisionFilterFunc.H
```diff
@@ -0,0 +1,228 @@
...
+ * pair of particles.
+ *
+ * @param[in] u1x,u1y,u1z momenta of the first colliding particle
+ * @param[in] u2x,u2y,u2z momenta of the second colliding particle
+ * @param[in] m1,m2 masses
+ * @param[in] w1,w2 effective weight of the colliding particles
+ * @param[in] dt is the time step length between two collision calls.
+ * @param[in] dV is the volume of the corresponding cell.
+ * @param[in] pair_index is the index of the colliding pair
+ * @param[out] p_mask is a mask that will be set to true if fusion occurs for that pair
```
RemiLehe: The documentation for `p_mask` should be updated, I think.

Source/Particles/Collision/BinaryCollision/DSMC/CollisionFilterFunc.H
```diff
@@ -0,0 +1,228 @@
...
+ *
+ * @param[in] I1s,I2s is the start index for I1,I2 (inclusive).
+ * @param[in] I1e,I2e is the stop index for I1,I2 (exclusive).
+ * @param[in] I1,I2 index arrays. They determine all elements that will be used.
+ * @param[in] ptd_1,ptd_2 contain the particle data of the two species
+ * @param[in] m1,m2 are masses.
+ * @param[in] dt is the time step length between two collision calls.
+ * @param[in] dV is the volume of the corresponding cell.
+ * @param[in] cell_start_pair is the start index of the pairs in that cell.
+ * @param[out] p_mask is a mask that will be set to true if a fusion event occurs for a given
```
RemiLehe: The documentation should be updated here, I think.

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -0,0 +1,168 @@
...
+    if (n_total_pairs == 0) {
+        return 0;
+    }
+
+    Gpu::DeviceVector<Index> offsets(n_total_pairs);
+    Index* AMREX_RESTRICT p_offsets = offsets.dataPtr();
+    Index offset = 0;
+    // the following is used to calculate the appropriate offset instead
+    // of a cummulative sum since the mask is also used to specify the type
+    // of collision and can therefore have values >1
```
RemiLehe: In order to be able to use `ExclusiveSum` (and thus run on GPU), could we create a temporary mask array here, which has the same number of elements as `mask` but has 0 wherever `mask` is 0 and 1 wherever `mask` is non-zero?

Or alternatively, we could ask `amrex` for a version of `ExclusiveSum` that takes a lambda function, in addition to the mask array.

Source/Particles/Collision/BinaryCollision/DSMC/CollisionFilterFunc.H
```diff
@@ -0,0 +1,228 @@
...
+    if (amrex::Random(engine) < probability)
+    {
+        const amrex::ParticleReal random_number = amrex::Random(engine);
+        amrex::ParticleReal sigma = 0._prt;
+        for (int ii = 0; ii < process_count; ii++) {
+            auto const& scattering_process = *(scattering_processes + ii);
+            sigma += scattering_process.getCrossSection(E_coll);
+            if (random_number <= sigma / sigma_tot)
+            {
+                p_mask[pair_index] = int(scattering_process.m_type);
```
RemiLehe, line 103: Is there a risk that `m_type` might give `0` when converted to `int` here?

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -0,0 +1,168 @@
...
+                uz2 = -uz1 * m1 / m2;
+            } else if (mask[i] == int(ScatteringProcessType::BACK)) {
+                // reverse the velocity vectors of both particles
+                ux1 *= -1.0_prt;
+                uy1 *= -1.0_prt;
+                uz1 *= -1.0_prt;
+                ux2 *= -1.0_prt;
+                uy2 *= -1.0_prt;
+                uz2 *= -1.0_prt;
+            } else if (mask[i] == int(ScatteringProcessType::CHARGE_EXCHANGE)) {
```
RemiLehe, line 133: In the future, for charge exchange through binary collisions, maybe we could replace this implementation by a separate one that works directly with the ionization level array (which is used already for field ionization).
This would allow field ionization, charge exchange and (later) impact ionization to all work self-consistently within the same simulation.

Source/Particles/Collision/BinaryCollision/DSMC/CollisionFilterFunc.H
RemiLehe: Ah, I guess no: only `INVALID` will map to 0.

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -0,0 +1,168 @@
...
+                    uz1 = uz2;
+                    ux2 = temp_ux;
+                    uy2 = temp_uy;
+                    uz2 = temp_uz;
+                }
+                else {
+                    Abort("Uneven mass charge-exchange not implemented yet.");
+                }
+            }
+            else {
```
RemiLehe, line 149: It looks like `EXCITATION` was not implemented here. Is it intended?

RemiLehe (commented):
Thanks for updating the PR! I added a few more comments.

Source/Particles/Collision/BinaryCollision/DSMC/DSMC.cpp
roelof-groenewald: I think so. The "bug" might be hidden by the fact that `Redistribute` will be called at a different point in the PIC loop, so the negative ID particles probably won't be around for too long after the collisions.

Source/Particles/Collision/BinaryCollision/DSMC/CollisionFilterFunc.H
roelof-groenewald: It's a good point, but I'm not sure it can be applied in quite the same way since in the MCC routine we pre-calculate a maximum collision frequency (which does not take into account any specific colliding pair for which we would know `lab_to_COM_factor` - see https://github.com/ECP-WarpX/WarpX/blob/e460e608a303b642023b2c8418bf5ae2ec64a5f5/Source/Particles/Collision/BackgroundMCC/BackgroundMCCCollision.cpp#L232). The individual collisions are then evaluated based on the ratio of their calculated collision frequency to that maximum collision frequency. In this approach would both the individual and maximum collision frequencies be scaled by the `lab_to_COM_factor`, which would effectively cancel it out?

Source/Particles/Collision/BinaryCollision/DSMC/DSMC.cpp
roelof-groenewald: That is correct yes since we don't know which species had the lower weight during the collision. So we add new particles for both species but for one there will also be a particle removed (weight -> 0 and the ID set to negative), resulting in no net increase in the total number of particles for that species.

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
roelof-groenewald: I'm open to either solution, maybe we can raise it at the next development meeting for broader discussion?

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
roelof-groenewald: That would be awesome!

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
roelof-groenewald: Yes, because in the DSMC scheme excitation would be a particle producing collision since the excited particle would need to be tracked as a separate kinetic species. With MCC since we don't track neutrals we can also ignore tracking excited neutrals and just assume they will relax before undergoing more collisions (since we are always "creating" a neutral to collide we just don't choose excited neutrals in that case).

RemiLehe:
We discussed a number of potential follow-up PRs offline. Here is a quick summary:

- [ ] Reduce code duplication between [BinaryCollision.H](https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/Collision/BinaryCollision/BinaryCollision.H) and `DSMC.cpp` by removing the `DSMC` class, and instead achieve the same thing by instantiating a `BinaryCollision` object with the right collision functor and copy-transform functor (namely `CollisionFilter` and `splitScatteringParticles` respectively).
This would require to modify the `BinaryCollision` class to account for the fact that the mask can now contain various integers (not just `0` or `1`), and to allow `scattering_processes` and `process_count` to be passed to the first functor. In addition, we might need to reconcile the syntax that was used to access the particles, since `BinaryCollision` uses [SoA syntax](https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/Collision/BinaryCollision/BinaryCollision.H#L371), while the `DSMC` class uses the [particle tile syntax](https://github.com/ECP-WarpX/WarpX/pull/4125/files#diff-9a87070f8e0486ad5bc78a927249eea39f88d271363c27947e0aa202d454dbb3R274).

- [ ] Make the charge exchange module (and future impact ionization module) modify the `ionizationLevel` variable, so that it can be run self-consistently together with field ionization.

- [ ] For collision processes that do not create new physical particles: consider using an algorithm that does not create new macroparticles, but instead scatters the heavier (i.e. higher weight) macroparticle only with a finite probability (equal to the ratio of weights). This would not conserve energy and momentum exactly, but would avoid creating large numbers of macroparticles.

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
WeiqunZhang:
```
auto const num_add = amrex::Scan::PrefixSum<Index>
    (n_total_pairs,
     [=] AMREX_GPU_DEVICE (Index i) -> Index { return mask[i] ? 1 : 0; },
     [=] AMREX_GPU_DEVICE (Index i, Index s) { p_offset[i] = s; },
     amrex::Scan::Type::exclusive, amrex::Scan::retSum);
```
something like this

roelof-groenewald: Thanks @WeiqunZhang!

RemiLehe (commented):
Thanks a lot for this PR!

RemiLehe (approved):

PR #4127, opened 2023-07-26
Add option to store quantum parameter during QED particle creation

NeilZaim:
This pull request adds an option to store the quantum parameter during QED particle creation (either quantum synchrotron or Breit-Wheeler).

In concrete terms, a user can write in the input file `photon.store_chi_at_creation = 1`, in which case the quantum parameter chi of the particle that emitted the photon is stored as an extra runtime attribute of the photon species.

This feature is useful for obtaining distributions of chi representative of QED processes rather than just relying on the `ParticleExtrema` diagnostic which only gives the maximum chi value at a given timestep.


Source/Particles/ElementaryProcess/QEDPairGeneration.H
```diff
@@ -119,7 +124,7 @@ public:
     * @param[in] engine random number generator engine
     */
     template <typename DstData, typename SrcData>
-    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    AMREX_GPU_DEVICE AMREX_FORCE_INLINE
```
NeilZaim, line 127: I made this change because the cuda tests failed due to calling the device only function `chi_photon` from within this functor.

I think this change is ok because these QED transform functors are only called on the device.

Docs/source/usage/parameters.rst
```diff
@@ -1105,6 +1105,11 @@ Particle initialization
     (the name of an existing positron species must be provided).
     **This feature requires to compile with QED=TRUE**
 
+* ``<species>.store_chi_at_creation`` (`0` or `1`) optional (default `0`)
+    For species which are product of Quantum synchrotron or Breit-Wheeler processes, enable this parameter to store the quantum chi parameter of the parent particle at the time of product particle creation.
```
```suggestion
    For species that are product of Quantum synchrotron or Breit-Wheeler processes, enable this parameter to store the quantum chi parameter of the parent particle at the time of product particle creation.
```
just a small suggestion

Source/Particles/ElementaryProcess/QEDPairGeneration.H
lucafedeli88: I agree with you that this change is fine.

Source/Particles/MultiParticleContainer.cpp
```diff
@@ -1607,6 +1630,13 @@ void MultiParticleContainer::doQedQuantumSync (int lev,
         pc_source ->defineAllParticleTiles();
         pc_product_phot->defineAllParticleTiles();
 
+        const bool store_chi_at_creation = pc_product_phot->store_chi_at_creation();
+        int chi_at_creation_runtime_comp = 0;
+        if (store_chi_at_creation){
+            chi_at_creation_runtime_comp =
+                                    pc_product_phot->particle_runtime_comps["qs_chi_at_creation"];
+        }
```
lucafedeli88: If you want to make this more compact and if you want to make `chi_at_creation_runtime_comp` a const quantity you can do:
```suggestion
        const int chi_at_creation_runtime_comp =  (store_chi_at_creation)?
            pc_product_phot->particle_runtime_comps["qs_chi_at_creation"] : 0;
```

lucafedeli88 (changes requested):
Thanks a lot for this PR, @NeilZaim ! This feature can be really useful.

I have a question: if the product species is for example an electron species that has particles at t=0, what happens to the `bw_chi_at_creation` quantity for these particles? Is it 0 or is it a random number due to uninitialized memory?
Maybe you could add a line in the documentation to warn about this potential issue, or you could initialize the quantity at -1 for particles that are not created via a QED process (possibly tricky for continuously injected species).

Examples/Tests/qed/breit_wheeler/analysis_core.py
```diff
@@ -183,6 +184,11 @@ def check_opt_depths(phot_data, ele_data, pos_data):
         assert( np.abs(scale - 1) < tol_red )
     print("  [OK] optical depth distributions are still exponential")
 
+def check_chi_at_creation(ele_data, pos_data, chi_theory):
+    assert(small_diff(ele_data["chi_at_creation"], chi_theory))
```
ax3l, line 188: Little wish, can you document this convention for the member here?
https://warpx.readthedocs.io/en/latest/developers/particles.html#particle-attributes

ax3l: Also, pls use camelCaseNames because we use currently the underscore sign in I/O to separate scalar from vector/tensor properties that have multiple components.

NeilZaim:
@lucafedeli88 

> I have a question: if the product species is for example an electron species that has particles at t=0, what happens to the bw_chi_at_creation quantity for these particles? Is it 0 or is it a random number due to uninitialized memory?
Maybe you could add a line in the documentation to warn about this potential issue, or you could initialize the quantity at -1 for particles that are not created via a QED process (possibly tricky for continuously injected species).

I think you are right, the value would be uninitialized, thanks for catching this! I have now set the value of the attribute to -1 in this case (a bit cumbersome to do this because there are multiple ways to add particles, but this is another topic) and stated this in the documentation.

Source/Particles/MultiParticleContainer.cpp
NeilZaim: Thanks for the suggestion, I've used this construct wherever possible.

Examples/Tests/qed/breit_wheeler/analysis_core.py
NeilZaim: These are good points, thanks. I've switched to camel case and added documentation in the page above (the formatting of the table is slightly off when I compile the doc but I see no simple way of improving this with my archaic rst knowledge :smiley: ).

NeilZaim:
@lucafedeli88 @ax3l Thanks for the reviews, I should have taken all your comments into account.

lucafedeli88:
I will close this PR, since it has received no updates for more than 1 year. If needed, we will re-open in or open a new one.

PR #4131, opened 2023-07-28
Remove superfluous #include from LatticeElements

lucafedeli88:
This PR removes some superfluous `#include` directives from LatticeElements source files.


dpgrote (approved):
This looks good, thanks!

PR #4132, opened 2023-07-28
Remove WarpX::GetInstance call from ParserFilter functor

lucafedeli88:
This PR removes the call to `WarpX::GetInstance` from ParserFilter functor. As a consequence, `time` has to be passed as an argument.

### DIFF
--- a/Source/Diagnostics/FlushFormats/FlushFormatPlotfile.H
+++ b/Source/Diagnostics/FlushFormats/FlushFormatPlotfile.H
@@ -48,7 +48,7 @@ public:
     /** \brief Write particles data to file.
      * \param[in] filename name of output directory
      * \param[in] particle_diags Each element of this vector handles output of 1 species.
-     * \param[in] time the simulation time
+     * \param[in] time the simulation time on the coarsest level
      * \param[in] isBTD whether this is a back-transformed diagnostic
      */
     void WriteParticles(const std::string& filename,
--- a/Source/Diagnostics/FlushFormats/FlushFormatPlotfile.H
+++ b/Source/Diagnostics/FlushFormats/FlushFormatPlotfile.H
@@ -48,10 +48,12 @@ public:
     /** \brief Write particles data to file.
      * \param[in] filename name of output directory
      * \param[in] particle_diags Each element of this vector handles output of 1 species.
+     * \param[in] time the simulation time
      * \param[in] isBTD whether this is a back-transformed diagnostic
      */
     void WriteParticles(const std::string& filename,
                         const amrex::Vector<ParticleDiag>& particle_diags,
+                        const amrex::Real time,
                         bool isBTD = false) const;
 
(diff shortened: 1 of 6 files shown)


Source/Diagnostics/FlushFormats/FlushFormatPlotfile.cpp
```diff
@@ -339,7 +339,7 @@ FlushFormatPlotfile::WriteWarpXHeader(
 void
 FlushFormatPlotfile::WriteParticles(const std::string& dir,
                                     const amrex::Vector<ParticleDiag>& particle_diags,
-                                    bool isBTD) const
+                                    const amrex::Real time, bool isBTD) const
```
NeilZaim, line 342: Shouldn't line 102 of this file be modified as well? Maybe there are other instances of this function being called?
Edit: Had a quick look and did not see any other instance

NeilZaim: Good news is that the automated tests appear to have caught this :)

lucafedeli88: you are totally right! I should have corrected the bug!

Source/Diagnostics/FlushFormats/FlushFormatPlotfile.H
```diff
@@ -48,10 +48,12 @@ public:
     /** \brief Write particles data to file.
      * \param[in] filename name of output directory
      * \param[in] particle_diags Each element of this vector handles output of 1 species.
+     * \param[in] time the simulation time
```
ax3l:
```suggestion
     * \param[in] time the simulation time on the coarsest level
```

Source/Particles/Filter/FilterFunctors.H
```diff
@@ -91,14 +90,15 @@ struct ParserFilter
      * \param a_is_active whether the test is active
      * \param a_filter_parser parser taking t, x, y, z, ux, uy, and uz, and returning a boolean for selected particle
      * \param a_mass mass of the particle species
+     * \param time simulation time
```
ax3l:
```suggestion
     * \param time simulation time on the coarsest level
```

Source/Particles/Filter/FilterFunctors.H
```diff
@@ -136,10 +136,10 @@ private:
 public:
     /** Parser function with 7 input variables, t,x,y,z,ux,uy,uz */
     amrex::ParserExecutor<7> const m_function_partparser;
-    /** Store physical time. */
-    amrex::ParticleReal m_t;
     /** Mass of particle species */
     amrex::ParticleReal m_mass;
+    /** Store physical time. */
```
ax3l:
```suggestion
    /** Store physical time on the coarsest level. */
```

ax3l (approved):
Thank you! :)

PR #4139, opened 2023-07-31
Move GotoNextLine helper function into ablastr 

lucafedeli88:
This PR moves the static function `WarpX::GotoNextLine (std::istream& is)` into the `ablastr` library (as a free function). It also changes the first argument of `is.ignore` from `constexpr std::streamsize bl_ignore_max { 100000 };`  to  `std::numeric_limits<std::streamsize>::max()`

### DIFF
--- a/Source/ablastr/utils/text/StreamUtils.H
+++ b/Source/ablastr/utils/text/StreamUtils.H
@@ -1,0 +1,24 @@
+/* Copyright 2023
+ *
+ * This file is part of WarpX.
+ *
+ * License: BSD-3-Clause-LBNL
+ */
+
+#ifndef ABLASTR_UTILS_TEXT_STREAMUTILS_H_
+#define ABLASTR_UTILS_TEXT_STREAMUTILS_H_
+
+#include <iostream>
+
+namespace ablastr::utils::text
+{
+    /**
+    * \brief This function drops the rest of the current line of
+    * the input stream "is"
+    *
+    * @param[in,out] is the input stream
+    */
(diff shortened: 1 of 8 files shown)


Source/ablastr/utils/text/StreamUtils.H
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * This file is part of WarpX.
+ *
+ * License: BSD-3-Clause-LBNL
+ */
+
+#ifndef ABLASTR_UTILS_TEXT_STREAMUTILS_H_
+#define ABLASTR_UTILS_TEXT_STREAMUTILS_H_
+
+#include <iostream>
```
ax3l: That is a lighter include to pull in:
```suggestion
#include <istream>
```

Source/ablastr/utils/text/StreamUtils.cpp
```diff
@@ -0,0 +1,16 @@
+/* Copyright 2023
+ *
+ * This file is part of WarpX.
+ *
+ * License: BSD-3-Clause-LBNL
+ */
+
+#include "StreamUtils.H"
+
+#include <limits>
```
ax3l, line 11: For `std::streamsize`:
```suggestion
#include <ios>
#include <limits>
```

lucafedeli88: ah, yes, you are right

ax3l (approved):
Wohoo, thx! :+1:

PR #4141, opened 2023-07-31
Move CheckGuardCells(amrex::MultiFab const& mf) function into anonymous namespace in WarpXInitData

lucafedeli88:
This PR is part of an effort to make `WarpX.H` a bit lighter


ax3l (approved):
LGTM :+1:

PR #4150, opened 2023-08-02
Clang tidy CI test: add some misc-* checks 

lucafedeli88:
This PR adds the following checks to clang-tidy CI test:

- [misc-misleading-identifier](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/misc-misleading-identifier.html)
- [misc-misplaced-const](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/misc-misplaced-const.html)

It explicitly excludes this check since it would be difficult to enforce it in WarpX (we may reconsider this in the future):
- [misc-definitions-in-headers](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/misc-definitions-in-headers.html)


Source/Initialization/WarpXAMReXInit.H
```diff
@@ -31,7 +31,7 @@ namespace warpx::initialization
         int& argc,
         char**& argv,
         bool const build_parm_parse = true,
```
ax3l, line 33: Interesting - this one should not be const either.
I thought we have a separate clang tidy check for that which catches this... Maybe only in AMReX so far.

ax3l (approved):

PR #4151, opened 2023-08-02
Remove unused Timer function from WarpXOpenPMD.H

lucafedeli88:
This PR removes an unused function from WarpXOpenPMD.H


ax3l (approved):
Thx! :)

PR #4154, opened 2023-08-03
Fix printing of warning messages at the end of the simulation

lucafedeli88:
Warning messages were not printed at the end of the simulation. This PR fixes this bug.


ax3l (approved):
Oh, oops! Good catch!

PR #4155, opened 2023-08-03
WarpX: Rule-of-Five

ax3l:
Explicitly declares the copy and move constructors for the `WarpX` class.
https://en.cppreference.com/w/cpp/language/rule_of_three


lucafedeli88 (approved):
Thanks @ax3l! There is actually a `clang-tidy` check that we could add to CI tests to enforce this rule in WarpX

PR #4157, opened 2023-08-04
Docs fix: extra whitespace in perlmutter gpu batch script 

aeriforme:
Removes an extra whitespace from perlmutter batch script 
in `Tools/machines/perlmutter-nersc/perlmutter_gpu.sbatch`.

### DIFF
--- a/Tools/machines/perlmutter-nersc/perlmutter_gpu.sbatch
+++ b/Tools/machines/perlmutter-nersc/perlmutter_gpu.sbatch
@@ -15,7 +15,7 @@
 # A100 40GB (most nodes)
 #SBATCH -C gpu
 # A100 80GB (256 nodes)
-#S BATCH -C gpu&hbm80g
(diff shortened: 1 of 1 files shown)


Tools/machines/perlmutter-nersc/perlmutter_gpu.sbatch
```diff
@@ -15,7 +15,7 @@
 # A100 40GB (most nodes)
 #SBATCH -C gpu
 # A100 80GB (256 nodes)
-#S BATCH -C gpu&hbm80g
+#SBATCH -C gpu&hbm80g
```
lucafedeli88, line 18: Sometimes these spaces are intentional (e.g., they are typically used to mark features that are not yet available). 
By making this change, do you mean that 80GB GPUs are generally available on Perlmutter and that a user should prefer them by default over the 40 GB GPUs?

PR #4158, opened 2023-08-04
Clang tidy CI test: add two performance-* checks to clang-tidy CI test 

lucafedeli88:
This PR adds the following checks to clang-tidy CI test:

- [performance-no-automatic-move](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-no-automatic-move.html)
- [performance-no-int-to-ptr](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-no-int-to-ptr.html)


Source/Particles/MultiParticleContainer.cpp
```diff
@@ -672,7 +672,7 @@ MultiParticleContainer::GetZeroParticlesInGrid (const int lev) const
 {
     const WarpX& warpx = WarpX::GetInstance();
     const int num_boxes = warpx.boxArray(lev).size();
-    const Vector<Long> r(num_boxes, 0);
+    Vector<Long> r(num_boxes, 0);
```
AlexanderSinn, line 675: This is exactly what I was referring to multiple weeks ago about const preventing moves and hurting performance.

lucafedeli88: Yes! That's why I asked for your review! This check https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-no-automatic-move.html should prevent this issue even if we enforce const correctness in the code.

ax3l (approved):
Thank you!! :)
Tricky!

PR #4161, opened 2023-08-04
Add RZ support for Ohm's law solver

roelof-groenewald:
This PR adds support for RZ geometry in the hybrid-PIC algorithm.

An initial test of the implementation successfully produced l = 0 (azimuthal mode number) normal modes in a metallic cylinder as shown in the attached plot. 
normal_modes_disp

To do:

- [x] Add mode analysis to the CI test's analysis file


roelof-groenewald:
Two tests broke with are unrelated to the hybrid-PIC solver (`LaserInjectionFromRZLASYFile` and `LaserInjectionFromLASYFile_RZ`). These broke because the charge and current density deposited outside the domain at r_max is "appropriately" handled with PEC boundaries. The checksum values for these tests all changed by less than 1%, so I will just reset the benchmark values.

Examples/Tests/ohm_solver_EM_modes/PICMI_inputs_rz.py
```diff
@@ -0,0 +1,258 @@
...
+
+from pywarpx import picmi
+
+constants = picmi.constants
+
+comm = mpi.COMM_WORLD
+
+simulation = picmi.Simulation(verbose=0)
+# make a shorthand for simulation.extension since we use it a lot
+sim_ext = simulation.extension
```
RemiLehe: It looks like the `sim_ext` variable is actually not used in the script ; do you want to remove it?

Examples/Tests/ohm_solver_EM_modes/analysis_rz.py
```diff
@@ -0,0 +1,169 @@
+#!/usr/bin/env python3
+#
+# --- Analysis script for the hybrid-PIC example producing EM modes.
```
RemiLehe, line 3: Is there a reference that gives the dispersion relation of the cylindrical modes that are being tested here?

Examples/Tests/ohm_solver_EM_modes/PICMI_inputs_rz.py
```diff
@@ -0,0 +1,258 @@
...
+                warpx_file_prefix='Python_ohms_law_solver_EM_modes_rz_plt'
+            )
+            simulation.add_diagnostic(part_diag)
+
+        #######################################################################
+        # Initialize simulation                                               #
+        #######################################################################
+
+        simulation.initialize_inputs()
+        simulation.initialize_warpx()
```
RemiLehe: Is this needed? My impression is that this would be called directly by `simulation.step`

Examples/Tests/ohm_solver_EM_modes/analysis_rz.py
```diff
@@ -0,0 +1,169 @@
...
+
+import dill
+import matplotlib
```
RemiLehe:
```suggestion
```

Examples/Tests/ohm_solver_EM_modes/analysis_rz.py
```diff
@@ -0,0 +1,169 @@
...
+    4.0 * np.pi * r_max**2 / j_1M**2
+    * j1(np.outer(jn_zeros(1, max(r_modes)+1)[r_modes], jn_zeros(1, nr)) / j_1M)
+    / jn(2 ,jn_zeros(1, nr))**2
+)
+
+# No transformation for z
+B = np.identity(nz)
+
+# combine projection arrays
+proj = np.einsum('ab,cd->acbd', A, B)
```
RemiLehe, line 76: I am not quite sure I understand why this operation is needed. My impression from e.g. [this paper](https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=46a789bdeaf36df1f2de813e9ddd5bd2aac23919) is that we can obtain the Hankel transform by simply doing a matrix multiplication with `A`. Could you explain more?

Examples/Tests/ohm_solver_EM_modes/analysis_rz.py
```diff
@@ -0,0 +1,169 @@
...
+    pm1 = ax.pcolormesh(
+        kz*sim.l_i, omega/sim.w_ci,
+        abs(F_kw[:, m, :])/np.max(abs(F_kw[:, m, :])),
+        norm=colors.LogNorm(vmin=vmin[ii], vmax=vmax),
+        cmap='inferno'
+    )
+    cb = fig.colorbar(pm1, ax=ax)
+    cb.set_label(r'Normalized $E_\theta(k_z, m, \omega)$')
+
+    # Get dispersion relation
```
RemiLehe: Could you add the reference that gives the dispersion relation below?
Also: for the parameters in this test, does the cylindrical dispersion relation (e.g. using the zeros of the Bessel function) differ significantly from the corresponding Cartesian dispersion relation? In other words, is this test able to confirm that we correctly capture the geometry-specific terms that arise in RZ?

Source/Initialization/WarpXInitData.cpp
```diff
@@ -803,28 +803,35 @@ WarpX::InitLevelData (int lev, Real /*time*/)
...
 
+#ifdef WARPX_DIM_RZ
+        std::stringstream warnMsg;
+        warnMsg << "Parser for external B (r and theta) fields does not work with RZ\n"
+            << "The initial Br and Bt fields are currently hardcoded to 0.\n"
+            << "The initial Bz field should only be a function of z.\n";
+        ablastr::warn_manager::WMRecordWarning(
+          "Inputs", warnMsg.str(), ablastr::warn_manager::WarnPriority::high);
+        str_Bx_ext_grid_function = "0";
+        str_By_ext_grid_function = "0";
```
RemiLehe, line 819: As a side-note: if this is needed, I could try to finalize this PR:
https://github.com/ECP-WarpX/WarpX/pull/3913

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -238,12 +416,256 @@ void FiniteDifferenceSolver::HybridPICSolveECylindrical (
...
+
+        // Extract stencil coefficients
+        Real const * const AMREX_RESTRICT coefs_r = m_stencil_coefs_r.dataPtr();
+        int const n_coefs_r = m_stencil_coefs_r.size();
+        Real const * const AMREX_RESTRICT coefs_z = m_stencil_coefs_z.dataPtr();
+        int const n_coefs_z = m_stencil_coefs_z.size();
+
+        // Extract cylindrical specific parameters
+        Real const dr = m_dr;
+        // int const nmodes = m_nmodes;
```
RemiLehe:
```suggestion
```

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -64,11 +64,189 @@ void FiniteDifferenceSolver::CalculateCurrentAmpereCylindrical (
...
+                            m * Br(i, j, 0, 2*m-1) / r
+                            + T_Algo::DownwardDrr_over_r(Bt, r, dr, coefs_r, n_coefs_r, i, j, 0, 2*m  )
+                        ); // Imaginary part
+                    }
+                // r==0: on-axis corrections
+                } else {
+                    // For m==0, Bt is linear in r, for small r
+                    // Therefore, the formula below regularizes the singularity
+                    Jz(i, j, 0, 0) = one_over_mu0 * 4 * Bt(i, j, 0, 0) / dr;
+                    // Ensure that Ez remains 0 for higher-order modes
```
RemiLehe:
```suggestion
                    // Ensure that Jz remains 0 for higher-order modes
```

RemiLehe (changes requested):
Thanks a lot for this PR!!
I added a few comments ; would you have time to address them?
Thanks,

roelof-groenewald:
Thanks @RemiLehe! I'll make the requested changes a bit later today and request a new review once that is done.

Examples/Tests/ohm_solver_EM_modes/PICMI_inputs_rz.py
roelof-groenewald: That is true. I'll remove it here. I usually have some print statements after initialization, so this was left over from the script I used to create this example.

Examples/Tests/ohm_solver_EM_modes/analysis_rz.py
roelof-groenewald: It is just a trick for doing the Hankel transform for all z-slices through a single matrix multiplication, instead of multiplying by `A` inside a for-loop over `nz`. 
Basically, `A` has shape `m` x `nr` (where `m` is the number of radial modes we want to keep), while the field data has shape `nr x nz`. So I create an identity matrix of shape `n x nz` (where `n` is the number of axial modes - in this case `n = nz`), and then reorder the axis so that `proj` has shape `m x n x nr x nz` so that I can do a one shot transform from the field data array.
This technique is nice if you only care about a small number of axial modes as well since you can do the Hankel and Fourier transform in a single multiplication (setting up `B` appropriately).

Examples/Tests/ohm_solver_EM_modes/analysis_rz.py
roelof-groenewald: Sure thing, I'll add this: T. Stix, Waves in Plasmas (American Inst. of Physics, 1992), Chap 6, Sec 2.

That's a good question regarding the test confirming geometric-specific aspects. The radial wave numbers that are supported come from the zeros of the first order Bessel function such that $J_1(\nu_m R) = 0$, which certainly do not generally align well with wave numbers where $\sin(k_r R)=0$. In the specific test shown here we have $R\approx0.4369$ m so the first few Bessel versus $sin$ zeros are:
|      | Bessel | sin |
| -------- | ------- | ---- |
| m = 1 | 8.77 m$^{-1}$ | 14.38 m$^{-1}$ |
| m = 2 | 16.06 m$^{-1}$ | 28.76 m$^{-1}$ |
| m = 3 | 23.29 m$^{-1}$ | 43.14 m$^{-1}$ |

It looks like the Cartesian dispersion would be noticeably removed from the dispersion as plotted. So I'd say the test is robust to confirm an RZ specific outcome.

Source/Initialization/WarpXInitData.cpp
roelof-groenewald: It would be helpful but not urgently. Specifically, I could see a situation in which we would like to apply an analytic perturbation to the axial field.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
roelof-groenewald: I'm hoping we can extend the implementation soon to support m > 0 modes.

Examples/Tests/ohm_solver_EM_modes/analysis_rz.py
roelof-groenewald: Added lower down.

RemiLehe (approved):

Examples/Tests/ohm_solver_EM_modes/PICMI_inputs_rz.py
```diff
@@ -0,0 +1,249 @@
...
+        field_diag = picmi.FieldDiagnostic(
+            name='field_diag',
+            grid=self.grid,
+            period=self.diag_steps,
+            data_list=['B', 'E'],
+            write_dir='diags',
+            warpx_file_prefix='field_diags',
+            warpx_format='openpmd',
+            warpx_openpmd_backend='h5',
+            warpx_write_species=False
```
roelof-groenewald: Remove outdated `write_species` flag.
```suggestion
```

PR #4166, opened 2023-08-07
Show compiler warnings with appleclang

johnson452:
Show compiler warnings with appleclang.


ax3l (approved):
Thank you, Grant! :tada:

PR #4175, opened 2023-08-10
Use more consistent particle type in pusher

RemiLehe:
The particle pusher uses `amrex::Real` in a few places where `amrex::ParticleReal` should be used instead.


lucafedeli88 (approved):
Thanks for this PR!

PR #4180, opened 2023-08-11
In picmi.py, add G to fields to plot check

dpgrote:
This adds `G` to the field names that can be included in the list of fields to be saved in `FieldDiagnostic` for the PICMI interface.


RemiLehe (approved):
Thanks!

Issue #3385, opened 2022-09-12
EBs: Should not Fail 2D or RZ if not used


ax3l:
When compiling with `-DWarpX_EB=ON -DWarpX_PSATD=ON`, the RZ geometry cannot be used anymore.

This can be seen by running the `Examples/Physics_applications/laser_acceleration/inputs_rz` test, which **does not use EB** (or PSATD).

```
./warpx.RZ.MPI.HIP.DP.PDP.OPMD.PSATD.EB.QED ../../Examples/Physics_applications/laser_acceleration/inputs_rz 
Initializing HIP...
HIP initialized.
MPI initialized with 1 MPI processes
MPI initialized with thread support level 3
AMReX (22.09) initialized
0::Assertion `do_electrostatic!=ElectrostaticSolverAlgo::None' failed, file "/ccs/home/huebl/src/warpx/Source/WarpX.cpp", line 680, Msg: 
### ERROR   : Currently, the embedded boundary in RZ only works for
#            electrostatic solvers.
 !!!
SIGABRT
See Backtrace.0 file for details
MPICH ERROR [Rank 0] [job id ] [Mon Sep 12 14:53:37 2022] [crusher068] - Abort(6) (rank 0 in comm 496): application called MPI_Abort(comm=0x84000000, 6) - process 0
```

The current unconditional assert should check if EBs are used at runtime.

Related to #3280

RemiLehe:
It would be nice to have a runtime helper function do determine whether the EB are being used (impl. function or STL or constructive geometry) - instead of using compilation flags.

ax3l:
Regular 2D simulations also crash if EBs are on, even if not used: #3686.

> It would be nice to have a runtime helper function do determine whether the EB are being used (impl. function or STL or constructive geometry) - instead of using compilation flags.

@WeiqunZhang are you able help with that? :) I think you are code author of these input routines.

ax3l:
Discussed today:
We can also add a plain runtime option
```
warpx.embedded_boundaries = true
```

ax3l:
Also gives us the opportunity to raise an error if EBs are to be used in inputs but were not compiled.

PR #4187, opened 2023-08-15
Assert moving window is off when EB is enabled.

atmyers:
This is currently not implemented and requires some thought - add an assertion to fail early for now.


atmyers:
It looks like a couple of test do compile WarpX with `AMREX_USE_EB=TRUE`, then run with the moving window; however, they don't set any geometry. Perhaps we should only fail the assert if `warpx.eb_implicit_function` is set? Alternatively, we could change the tests.

ax3l:
Thanks Andrew - I will take this on and add this:
https://github.com/ECP-WarpX/WarpX/issues/3385#issuecomment-1579481282

PR #4182, opened 2023-08-14
Docs: HPC3 (UCI) with Only `-j 12`

ax3l:
We use too many cores when we compile and thus get our compile processes killed by the system. Let's use a bit less to stay in the limits.

Fix #4181

PR #4189, opened 2023-08-15
Update Python installation instructions on HPC

RemiLehe:
We should use the new flag `WarpX_PYTHON`

Follow-up to #3474


ax3l:
Rebasing to resolve conflict with #4182

Docs/source/install/cmake.rst
```diff
@@ -315,21 +315,16 @@ So another sophisticated example might be: use Clang as the compiler, build with
 Here we wrote this all in one line, but one can also set all environment variables in a development environment and keep the pip call nice and short as in the beginning.
```
ax3l, line 315: @roelof-groenewald I might want to simplify this further in a follow-up, as in https://impactx.readthedocs.io/en/latest/install/cmake.html

PR #4196, opened 2023-08-16
Allow multiple injection sources per species

dpgrote:
This PR allows multiple injection sources to be specified for each species. The existing input parameter format for specifying a single injection source is unchanged (since that is the most common case and since changing it would break all existing input files). This PR adds a new input parameter `<species_name>.injection_sources` which is a list of names of new sources. The input parameters for each source are specified as usual but with he source name added. For instance, with `<species_name>.injection_sources = source1 source2` there will be two input parameters `<species_name>.source1.injection_style` and `<species_name>.source2.injection_style`. The code is written so that all of the injection sources can be specified this way, ignoring the existing method for a single source.

CI tests were added, the nuclear fusion test and the plasma lens test (for the PICMI version).


Docs/source/usage/parameters.rst
```diff
@@ -669,6 +669,18 @@ Particle initialization
...
     If periodic boundary conditions are used in direction ``i``, then the default (i.e. if the range is not specified) range will be the simulation box, ``[geometry.prob_hi[i], geometry.prob_lo[i]]``.
 
+* ``<species_name>.injection_sources`` (``list of strings``) optional
+    Names of additional injection sources. By default, WarpX assumes one injection source per species, hence all of the input
+    parameters below describing the injection are parameters directly of the species. However, this option allows
+    additional sources, the names of which are specified here. For each source, the name of the source is added to the
+    input parameters below. For instance, with ``<species_name>.injection_sources = source1 source2`` there can be the two input
+    parameters ``<species_name>.source1.injection_style`` and ``<species_name>.source2.injection_style``.
+    For the parameters of each source, the parameter with the name of the source will be used.
+    If it is not given, the value of the parameter without the source name will be used. This allows parameters used for all
```
RemiLehe, line 679:
```suggestion
    If ``<species_name>.injection_sources`` is not given, the value of the parameter without the source name will be used. This allows parameters used for all
```

Examples/Tests/langmuir/inputs_2d
```diff
@@ -47,16 +47,25 @@ my_constants.k = 2.*pi/20.e-6  # perturbation wavenumber
 # Particles
 particles.species_names = electrons positrons
 
+# Split the loading of the plasma to test multiple injection sources
+# The two sources s1 and s2 are identical except their zmin and zmax
```
RemiLehe: Could we actually do this in some other test?
This is because the langmuir input file is one of the first test that people run when developing new algorithms, and it would be nice to keep the input script relatively simple.

Examples/Tests/nuclear_fusion/inputs_deuterium_tritium_3d
```diff
@@ -74,18 +74,22 @@ neutron_1.do_not_deposit = 1
...
-deuterium_2.num_particles_per_cell = 1000
-deuterium_2.profile = "parse_density_function"
 ## A tenth of the macroparticles in each cell is made of immobile high-density background deuteriums.
 ## The other nine tenths are made of fast low-density beam deuteriums.
-deuterium_2.density_function(x,y,z) = if(y - floor(y) < 0.1, 10.*background_dens, 10./9.*beam_dens)
-deuterium_2.momentum_distribution_type = "parse_momentum_function"
-deuterium_2.momentum_function_ux(x,y,z) = 0.
-deuterium_2.momentum_function_uy(x,y,z) = 0.
-deuterium_2.momentum_function_uz(x,y,z) = "if(y - floor(y) < 0.1,
-                                          0., sqrt(2*m_deuterium*Energy_step*(floor(z)**2))/(m_deuterium*clight))"
```
RemiLehe, line 88: Nice! This makes things a lot cleaner.

Source/Utils/ParmParseWithOptionalGroup.cpp
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+}
+
+void ParmParseWithOptionalGroup::getWithParser (const char* name, int& ref) const
+{
+    // First, query name with the group name
+    const amrex::ParmParse pp_with_group(prefix_dot_group);
+    bool is_specified = utils::parser::queryWithParser(pp_with_group, name, ref);
+    if (!is_specified) {
```
RemiLehe: Should this be
```suggestion
    if (!is_specified && !group.empty()) {
```

Source/Utils/WarpXMovingWindow.cpp
```diff
@@ -81,7 +81,8 @@ WarpX::UpdateInjectionPosition (const amrex::Real a_dt)
             current_injection_position[dir] = pc.m_current_injection_position;
 #endif
 
-            PlasmaInjector* plasma_injector = pc.GetPlasmaInjector();
+            // This only uses the base plasma injector
+            PlasmaInjector* plasma_injector = pc.GetPlasmaInjector(0);
```
RemiLehe, line 85: Could you explain this a bit more? Does it mean that the injection with the moving window could not work with multiple plasma injection sources?

Examples/Tests/langmuir/inputs_2d
dpgrote: I can remove this test since there is also the nuclear fusion test.

Examples/Tests/nuclear_fusion/inputs_deuterium_tritium_3d
dpgrote: Note that I tried doing the same thing in the RZ version of the test, but it was failing one of the tolerance tests. I couldn't figure out why so I removed it. I can give you the modified input file if you would want to take a look at it.

Source/Particles/PhysicalParticleContainer.H
```diff
@@ -72,9 +72,9 @@ public:
     void InitIonizationModule () override;
 
     /*
-     * \brief Returns a pointer to the plasma injector.
+     * \brief Returns a reference to the i'th plasma injector.
```
RemiLehe: It seems that it is still a pointer.
```suggestion
     * \brief Returns a pointer to the i'th plasma injector.
```

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -2951,9 +3022,13 @@ PhysicalParticleContainer::getIonizationFunc (const WarpXParIter& pti,
                                 ion_atomic_number};
 }
 
-PlasmaInjector* PhysicalParticleContainer::GetPlasmaInjector ()
+PlasmaInjector* PhysicalParticleContainer::GetPlasmaInjector (int i)
 {
-    return plasma_injector.get();
+    if (i < 0 || i >= static_cast<int>(plasma_injectors.size())) {
+        return nullptr;
```
RemiLehe, line 3052: Should we call `amrex::Abort` or `WARPX_ALWAYS_ASSERT_WITH_MESSAGE` here?

RemiLehe (changes requested):
Thanks a lot for this PR! I added a few comments.

Source/Utils/ParmParseWithOptionalGroup.cpp
dpgrote: Since this is a `get` the parameter must be supplied. So if `is_speficied` is false and the group is empty, this is an error since the parameter wasn't given. As I have, it then calls `get` which would raise an exception. Though, maybe this could be rearranged so that the exception would reference the parameter name with the group name included - this might be more user friendly.

Source/Utils/WarpXMovingWindow.cpp
dpgrote: Basically yes, the moving window would not work in general with multiple sources. Though, it is only using the bulk velocity, so if that was the same for all of the sources, then this would be correct. Here, I punted and just used the first source, which keeps the code consistent with what was already there.
I could add a comment about this. Perhaps a check could be added here to ensure this?

Source/Particles/PhysicalParticleContainer.cpp
dpgrote: This routine is now only used by the moving window code, and it checks if the result is a null pointer. There can be species for which the injection was not defined in which case `plasma_injectors.size()` will be zero, so the call from the moving window passing in zero would then get the null pointer.

Source/Utils/Parser/ParserUtils.H
```diff
@@ -63,9 +63,28 @@ namespace utils::parser
...
+     *
+     * \param[in] pp used to read the query_string `pp.<function>=string`
+     * \param[in] group name of the optional group
+     * \param[in] query_string ParmParse.query will look for this string
+     * \param[out] stored_string variable in which the string to parse is stored
+     */
     void Store_parserString(
         const amrex::ParmParse &pp,
-        std::string query_string,
+        std::string const& group,
```
ax3l, line 86: instead of an empty string, this could be a `std::optional<std::string>`, to make it more explicit. This way, one can also default to `std::nullopt` instead of `""`.

dpgrote: This is used by `PlasmaInjector` where the `source_name` is passed in for the group. The `source_name` will be the empty string when the original single injection source method is used for input.

RemiLehe (approved):
Thanks for updating the PR for the PICMI interface!

RemiLehe:
@dpgrote It looks like the PR has a few merge conflicts. Would you have time to fix them?

Issue #4185, opened 2023-08-15
WarpX hanging on Lassen


bzdjordje:
There were frequent hangs with WarpX with the last version and with the recent release it seems to have become persistent. I checked with a computer scientist at Livermore and when analyzing an ongoing run with a stack trace analysis tool we have it was concluded that most tasks were getting tied up in mpi communications (PAMI?). Attached are example stat traces as well as the files for the run in question. 

[output.txt](https://github.com/ECP-WarpX/WarpX/files/12340344/output.txt)
[warpx_rz_stat_00.pdf](https://github.com/ECP-WarpX/WarpX/files/12340345/warpx_rz_stat_00.pdf)
[warpx_rz_stat_01.pdf](https://github.com/ECP-WarpX/WarpX/files/12340346/warpx_rz_stat_01.pdf)
[warpx_rz_stat_02.pdf](https://github.com/ECP-WarpX/WarpX/files/12340347/warpx_rz_stat_02.pdf)
[warpx_rz_stat_03.pdf](https://github.com/ECP-WarpX/WarpX/files/12340348/warpx_rz_stat_03.pdf)
[warpx_rz_stat_04.pdf](https://github.com/ECP-WarpX/WarpX/files/12340349/warpx_rz_stat_04.pdf)
[warpx_rz_stat_05.pdf](https://github.com/ECP-WarpX/WarpX/files/12340350/warpx_rz_stat_05.pdf)
[warpx_rz_stat_check.pdf](https://github.com/ECP-WarpX/WarpX/files/12340351/warpx_rz_stat_check.pdf)
[WarpXe.5125616.txt](https://github.com/ECP-WarpX/WarpX/files/12340352/WarpXe.5125616.txt)
[WarpXo.5125616.txt](https://github.com/ECP-WarpX/WarpX/files/12340353/WarpXo.5125616.txt)

dpgrote:
I have access to Lassen and can take a look. What modules are you using when you build WarpX?

joshualudwig8:
I have also been experiencing frequent hangs with WarpX with the latest version on Lassen.
I use the instructions from [here](https://warpx.readthedocs.io/en/latest/install/hpc/lassen.html) to compile it.
The hangs I experience are always happening during outputs, but appear to be random otherwise. (I ran the same simulation twice and both times it hung on the same output, I reran the simulation for a third time with `warpx.random_seed = random` and it ran fine.)

ax3l:
[...]

PR #2874, opened 2022-02-18
Lassen: Work-Around MPI Allgatherv

ax3l:
The implementation of AllGatherv in IBM's MPI optimization library "libcollectives" is broken and leads to HDF5 crashes for multi-node runs.

IBM Spectrum MPI, Version 10 Release 1, User's Guide:
  https://www.ibm.com/docs/en/SSZTET_EOS/eos/guide_101.pdf

See:
https://github.com/ECP-WarpX/WarpX/pull/2863#issuecomment-1043623849

PR #4199, opened 2023-08-17
Docs: Lassen (LNL) Skip IBM MPI_AllGather

ax3l:
IBM libcollective keeps on giving.

See #4185 #2874 

Also documents further bugs in the system that are already in our runscript template.


ax3l:
does not apply

PR #4200, opened 2023-08-17
Fix: Remove Overly Exported Classes

ax3l:
Luckily, many of them should not contain global state.

Avoids that we need to include the WarpX "lib" export header downstream in ImpactX, which only needs ABLASTR. ABLASTR currently still side-includes some WarpX headers.

Follow-up to #3474

### DIFF
--- a/Source/Particles/Deposition/ChargeDeposition.H
+++ b/Source/Particles/Deposition/ChargeDeposition.H
@@ -13,7 +13,6 @@
 #include "Particles/Pusher/GetAndSetPosition.H"
 #include "Particles/ShapeFactors.H"
 #include "Utils/WarpXAlgorithmSelection.H"
-#include "Utils/WarpXProfilerWrapper.H"
 #ifdef WARPX_DIM_RZ
 #   include "Utils/WarpX_Complex.H"
 #endif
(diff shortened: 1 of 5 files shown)


Source/Particles/Deposition/ChargeDeposition.H
```diff
@@ -13,10 +13,10 @@
 #include "Particles/Pusher/GetAndSetPosition.H"
 #include "Particles/ShapeFactors.H"
 #include "Utils/WarpXAlgorithmSelection.H"
-#include "Utils/WarpXProfilerWrapper.H"
 #ifdef WARPX_DIM_RZ
 #   include "Utils/WarpX_Complex.H"
 #endif
+#include "WarpX.H"
```
ax3l, line 19: The file uses `WarpX::shared_tilesize`.

roelof-groenewald: That's curious that the above change caused this to break - was it effectively including `WarpX.H` through the profile wrapper?

ax3l:
Exactly. The profiler include masked the missing direct include.

Source/Particles/Deposition/ChargeDeposition.H
ax3l: Exactly. The profiler include masked the missing direct include.

roelof-groenewald (approved):
Looks good! The CUDA builds seem to be failing for all PRs (some issue with pulling needed packages).

PR #4203, opened 2023-08-18
Use cupy when available with the particle container wrapper

dpgrote:
This avoid relying on managed memory by explicitly using `cupy` when it is available.

- [x] we might be able to replace this mostly with https://github.com/AMReX-Codes/pyamrex/pull/88

### DIFF
--- a/Python/pywarpx/particle_containers.py
+++ b/Python/pywarpx/particle_containers.py
@@ -8,6 +8,13 @@
 
 import numpy as np
 
+try:
+    import cupy as cp
+    cnp = cp
+except ImportError:
+    cp = None
+    cnp = np
+
 from ._libwarpx import libwarpx
 
 
@@ -185,12 +192,12 @@ class ParticleContainerWrapper(object):
 
     def get_particle_structs(self, level):
         '''
-        This returns a list of numpy arrays containing the particle struct data
+        This returns a list of numpy or cupy arrays containing the particle struct data
         on each tile for this process. The particle data is represented as a structured
-        numpy array and contains the particle 'x', 'y', 'z', and 'idcpu'.
+        array and contains the particle 'x', 'y', 'z', and 'idcpu'.
 
-        The data for the numpy arrays are not copied, but share the underlying
-        memory buffer with WarpX. The numpy arrays are fully writeable.
+        The data for the arrays are not copied, but share the underlying
+        memory buffer with WarpX. The arrays are fully writeable.
 
         Parameters
         ----------
@@ -201,22 +208,22 @@ class ParticleContainerWrapper(object):
         Returns
         -------
 
-        List of numpy arrays
+        List of arrays
             The requested particle struct data
         '''
         particle_data = []
         for pti in libwarpx.libwarpx_so.WarpXParIter(self.particle_container, level):
-            aos_arr = np.array(pti.aos(), copy=False)
+            aos_arr = cnp.array(pti.aos(), copy=False)
             particle_data.append(aos_arr)
         return particle_data
 
     def get_particle_arrays(self, comp_name, level):
         '''
-        This returns a list of numpy arrays containing the particle array data
+        This returns a list of numpy or cupy arrays containing the particle array data
         on each tile for this process.
 
-        The data for the numpy arrays are not copied, but share the underlying
-        memory buffer with WarpX. The numpy arrays are fully writeable.
+        The data for the arrays are not copied, but share the underlying
+        memory buffer with WarpX. The arrays are fully writeable.
 
         Parameters
         ----------
@@ -230,7 +237,7 @@ class ParticleContainerWrapper(object):
         Returns
         -------
 
-        List of numpy arrays
+        List of arrays
             The requested particle array data
         '''
         comp_idx = self.particle_container.get_comp_index(comp_name)
@@ -238,13 +245,15 @@ class ParticleContainerWrapper(object):
         data_array = []
         for pti in libwarpx.libwarpx_so.WarpXParIter(self.particle_container, level):
             soa = pti.soa()
-            data_array.append(np.array(soa.GetRealData(comp_idx), copy=False))
+            idx = soa.GetRealData(comp_idx)
+            idx_arr = cnp.array(idx, copy=False)
+            data_array.append(idx_arr)
         return data_array
 
     def get_particle_id(self, level=0):
         '''
 
-        Return a list of numpy arrays containing the particle 'id'
+        Return a list of numpy or cupy arrays containing the particle 'id'
         numbers on each tile.
 
         '''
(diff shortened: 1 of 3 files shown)


Python/pywarpx/particle_containers.py
```diff
@@ -341,48 +384,48 @@ def get_particle_z(self, level=0):
...
         z momentum on each tile.
 
         '''
 
-        return self.get_particle_arrays('uz', level)
+        return self.get_particle_arrays('uz', level, copy_to_host=copy_to_host)
     uzp = property(get_particle_uz)
 
-    def get_species_charge_sum(self, local=False):
+    def get_species_charge_sum(self, local=False, copy_to_host=copy_to_host):
```
ax3l:
```suggestion
    def get_species_charge_sum(self, local=False):
```

Python/pywarpx/particle_containers.py
```diff
@@ -264,71 +307,71 @@ def get_particle_cpu(self, level=0):
     def get_particle_x(self, level=0):
         '''
 
-        Return a list of numpy arrays containing the particle 'x'
+        Return a list of numpy or cupy arrays containing the particle 'x'
         positions on each tile.
 
         '''
         structs = self.get_particle_structs(level)
```
ax3l:
```suggestion
        xp, cupy_status = load_cupy()

        structs = self.get_particle_structs(level)
```

Python/pywarpx/particle_containers.py
```diff
@@ -264,71 +307,71 @@ def get_particle_cpu(self, level=0):
...
         if libwarpx.geometry_dim == '3d' or libwarpx.geometry_dim == '2d':
             return [struct['x'] for struct in structs]
         elif libwarpx.geometry_dim == 'rz':
-            return [struct['x']*np.cos(theta) for struct, theta in zip(structs, self.get_particle_theta(species_name))]
+            return [struct['x']*cnp.cos(theta) for struct, theta in zip(structs, self.get_particle_theta(species_name))]
```
ax3l:
```suggestion
            return [struct['x']*xp.cos(theta) for struct, theta in zip(structs, self.get_particle_theta(species_name))]
```

Python/pywarpx/particle_containers.py
```diff
@@ -264,71 +307,71 @@ def get_particle_cpu(self, level=0):
...
 
     def get_particle_y(self, level=0):
         '''
 
-        Return a list of numpy arrays containing the particle 'y'
+        Return a list of numpy or cupy arrays containing the particle 'y'
         positions on each tile.
 
         '''
         structs = self.get_particle_structs(level)
```
ax3l:
```suggestion
        xp, cupy_status = load_cupy()

        structs = self.get_particle_structs(level)
```

Python/pywarpx/particle_containers.py
```diff
@@ -264,71 +307,71 @@ def get_particle_cpu(self, level=0):
...
         if libwarpx.geometry_dim == '3d':
             return [struct['y'] for struct in structs]
         elif libwarpx.geometry_dim == 'rz':
-            return [struct['x']*np.sin(theta) for struct, theta in zip(structs, self.get_particle_theta(species_name))]
+            return [struct['x']*cnp.sin(theta) for struct, theta in zip(structs, self.get_particle_theta(species_name))]
```
ax3l:
```suggestion
            return [struct['x']*xp.sin(theta) for struct, theta in zip(structs, self.get_particle_theta(species_name))]
```

Python/pywarpx/particle_containers.py
```diff
@@ -264,71 +307,71 @@ def get_particle_cpu(self, level=0):
...
 
     def get_particle_r(self, level=0):
         '''
 
-        Return a list of numpy arrays containing the particle 'r'
+        Return a list of numpy or cupy arrays containing the particle 'r'
         positions on each tile.
 
         '''
         structs = self.get_particle_structs(level)
```
ax3l:
```suggestion
        xp, cupy_status = load_cupy()

        structs = self.get_particle_structs(level)
```

Python/pywarpx/particle_containers.py
```diff
@@ -264,71 +307,71 @@ def get_particle_cpu(self, level=0):
...
         if libwarpx.geometry_dim == 'rz':
             return [struct['x'] for struct in structs]
         elif libwarpx.geometry_dim == '3d':
-            return [np.sqrt(struct['x']**2 + struct['y']**2) for struct in structs]
+            return [cnp.sqrt(struct['x']**2 + struct['y']**2) for struct in structs]
```
ax3l:
```suggestion
            return [xp.sqrt(struct['x']**2 + struct['y']**2) for struct in structs]
```

Python/pywarpx/particle_containers.py
```diff
@@ -264,71 +307,71 @@ def get_particle_cpu(self, level=0):
...
 
     def get_particle_theta(self, species_name, level=0):
         '''
 
-        Return a list of numpy arrays containing the particle
+        Return a list of numpy or cupy arrays containing the particle
         theta on each tile.
 
         '''
         if libwarpx.geometry_dim == 'rz':
```
ax3l, line 458:
```suggestion
        xp, cupy_status = load_cupy()

        if libwarpx.geometry_dim == 'rz':
```

Python/pywarpx/particle_containers.py
```diff
@@ -264,71 +307,71 @@ def get_particle_cpu(self, level=0):
...
             return self.get_particle_arrays('theta', level)
         elif libwarpx.geometry_dim == '3d':
             structs = self.get_particle_structs(level)
-            return [np.arctan2(struct['y'], struct['x']) for struct in structs]
+            return [cnp.arctan2(struct['y'], struct['x']) for struct in structs]
```
ax3l:
```suggestion
            return [xp.arctan2(struct['y'], struct['x']) for struct in structs]
```

RTSandberg:
I tested this on Perlmutter and successfully ran with cupy in Python call-backs and no managed memory

ax3l:
I pushed one more update based on the new functions in https://github.com/AMReX-Codes/pyamrex/pull/88.

Now, we avoid cupy for the workflow side that only creates a copy to host.
Instead, we do an explicit copy in AMReX. This way, we can support
- host access of data copies w/o even having cupy installed
- **support on HIP** (we did not yet implement DLPack APIs and cuda array APIs _for ROCm_ are since recently unsupported in cupy)
- AoS data copies (which are currently unsupported types in cupy) - but I already had a work around for this

ax3l (approved):
Good to go now from my side.

Python/pywarpx/particle_containers.py
```diff
@@ -183,156 +184,308 @@ def add_real_comp(self, pid_name, comm=True):
...
+        array and contains the particle 'x', 'y', 'z', and 'idcpu'.
 
-        The data for the numpy arrays are not copied, but share the underlying
-        memory buffer with WarpX. The numpy arrays are fully writeable.
+        Unless copy_to_host is specified, the data for the structs are
+        not copied, but share the underlying memory buffer with WarpX. The
+        arrays are fully writeable.
+
+        Note that cupy does not support structs:
+        ...
```
ax3l:
```suggestion
        https://github.com/cupy/cupy/issues/2031
```

PR #4205, opened 2023-08-19
Add Low-noise Mode GPU-to-CPU mapping for Frontier

n01r:
Frontier uses a Low-noise Mode Layout for its nodes which pins system processes via the default core specification `-S 8` to the first core every L3 region.


ax3l (approved):
Great catch, thank you! :)

PR #4208, opened 2023-08-21
 BoxArray in BTD has single box

RevathiJambunathan:
The BoxArray max size in BTD was set equal to buffer-size -- i.e., number of cells in the z-direction. 
However, if number of cells in x and y > z, then the box array will have more than one box. 
BTD is designed to have only one box -- since we write one buffer at a time 

This is fixed in this PR

### DIFF
--- a/Source/Diagnostics/BTDiagnostics.cpp
+++ b/Source/Diagnostics/BTDiagnostics.cpp
@@ -259,9 +259,7 @@ BTDiagnostics::ReadParameters ()
     WARPX_ALWAYS_ASSERT_WITH_MESSAGE(snapshot_interval_is_specified,
         "For back-transformed diagnostics, user should specify either dz_snapshots_lab or dt_snapshots_lab");
 
-    if (utils::parser::queryWithParser(pp_diag_name, "buffer_size", m_buffer_size)) {
-        if(m_max_box_size < m_buffer_size) m_max_box_size = m_buffer_size;
-    }
+    utils::parser::queryWithParser(pp_diag_name, "buffer_size", m_buffer_size);
 #ifdef WARPX_DIM_RZ
     const amrex::Vector< std::string > BTD_varnames_supported = {"Er", "Et", "Ez",
                                                            "Br", "Bt", "Bz",
@@ -847,7 +845,7 @@ BTDiagnostics::PrepareFieldDataForOutput ()
                                              i_buffer, ZSliceInDomain,
                                              m_current_z_boost[i_buffer],
                                              m_buffer_box[i_buffer],
-                                             k_index_zlab(i_buffer, lev), m_max_box_size,
+                                             k_index_zlab(i_buffer, lev),
                                              m_snapshot_full[i_buffer] );
 
             }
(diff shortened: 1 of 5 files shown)


Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -897,7 +897,6 @@ BTDiagnostics::DefineFieldBufferMultiFab (const int i_buffer, const int lev)
     m_buffer_box[i_buffer].setSmall( m_moving_window_dir, hi_k_lab - m_buffer_size + 1);
     m_buffer_box[i_buffer].setBig( m_moving_window_dir, hi_k_lab );
     amrex::BoxArray buffer_ba( m_buffer_box[i_buffer] );
-    buffer_ba.maxSize(m_max_box_size);
```
RemiLehe, line 900: Should we then remove `m_max_box_size` from the member variables of this class?

ax3l:
@RevathiJambunathan  @WeiqunZhang can we add in `ParticleContainer::SetParticleBoxArray` an assert that checks the size of the passed BoxArray?

Or in WarpX `Source/Particles/MultiParticleContainer.cpp`'s `MultiParticleContainer::SetParticleBoxArray` if we need additional, WarpX-specific info?

Source/Diagnostics/BTDiagnostics.cpp
ax3l: Seems to be used elsewhere as well:
```
$ ag m_max_box_size
Source/Diagnostics/BTDiagnostics.H
178:    int m_max_box_size = 256;

Source/Diagnostics/ComputeDiagFunctors/BackTransformFunctor.cpp
166:    m_max_box_size = max_box_size;

Source/Diagnostics/ComputeDiagFunctors/BackTransformFunctor.H
127:    int m_max_box_size = 256;

Source/Diagnostics/BTDiagnostics.cpp
263:        if(m_max_box_size < m_buffer_size) m_max_box_size = m_buffer_size;
850:                                             k_index_zlab(i_buffer, lev), m_max_box_size,
```

ax3l (approved):
Fixes the issue for me in first tests :crossed_fingers:

Source/Diagnostics/BTDiagnostics.cpp
RevathiJambunathan: yes we can remove all references to `m_max_box_size`

RevathiJambunathan: @RemiLehe this is now fixed

RevathiJambunathan:
> @RevathiJambunathan @WeiqunZhang can we add in `ParticleContainer::SetParticleBoxArray` an assert that checks the size of the passed BoxArray?
> 
> Or in WarpX `Source/Particles/MultiParticleContainer.cpp`'s `MultiParticleContainer::SetParticleBoxArray` if we need additional, WarpX-specific info?

@ax3l as discussed offline, this should be checked in BTD -- since the requirement for one box in the box-array is specific to BTD

ax3l:
Perfect, agreed :+1:

ax3l (approved):
LGTM, thank you!! :tada:

PR #4210, opened 2023-08-21
[WIP] Remove `none` boundary condition, use `r_origin` for r=0

### LINKED ISSUE #3694: Rename/Improve `none` boundary conditions with PSATD solver
As suggested by @jlvay in recent meetings, the boundary conditions labeled as `none` (guard cells set to zero, I believe) should be renamed (and possibly also improved), because it is not clear what they correspond to physically.

RemiLehe:
## Overview

Changes:
- Remove `none` boundary condition for all solvers and all geometries (poorly defined)
- Rename old `none` to `r_origin`, to be used only in RZ geometry at the boundary r=0

## To do
- [ ] Discuss and decide whether we really want to remove `none` for all cases (e.g., cases with no Maxwell solver)
- [ ] Update PICMI interface

## Notes

Close #3694.


EZoni:
Thanks, Remi! 

Should we add a check in the function `BackwardCompatibility`, 
with a warning that says that `none` is not supported anymore?
https://github.com/ECP-WarpX/WarpX/blob/3d29f8714fc299e429376339c2913c1dbbc94c53/Source/WarpX.cpp#L1641

Docs/source/usage/parameters.rst
```diff
@@ -343,7 +343,7 @@ Domain Boundary Conditions
     * ``pec``: This option can be used to set a Perfect Electric Conductor at the simulation boundary. Please see the :ref:`PEC theory section <theory-bc-pec>` for more details. Note that PEC boundary is invalid at `r=0` for the RZ solver. Please use ``none`` option. This boundary condition does not work with the spectral solver.
       If an electrostatic field solve is used the boundary potentials can also be set through ``boundary.potential_lo_x/y/z`` and ``boundary.potential_hi_x/y/z`` (default `0`).
 
-    * ``none``: No boundary condition is applied to the fields with the electromagnetic solver. This option must be used for the RZ-solver at `r=0`.
+    * ``none``: This option must be used for the RZ-solver at `r=0`.
```
dpgrote: Was the 'none' going to be changed to 'r_origin'? (Though I like 'axis')

ax3l: yes, but should be a separate PR if that works :)

ax3l: Yes, might be worth in one go.

I think during the meeting most liked `r_origin`, although `r_axis` et al are also reasonable choices.

EZoni:
Putting "[WIP]" back in the PR title, waiting for CI to run and catch all instances of `none` to be replaced.

EZoni:
Just dropping an update here to mention that we decided for `axis` instead of `r_origin` in the WarpX hackathon last week:

Screenshot from 2025-05-19 13-58-22

My advice would be to replace this PR with a new one from scratch, given the large number of conflicts that need to be resolved.

PR #4211, opened 2023-08-21
CMake: Fix logic that triggers shared AMReX

ax3l:
Accidentally always built AMReX as a shared library, which is only needed if we build with Python (aka: only needed if we share the AMReX library and its global variables over multiple libs).

Follow-up to #3474

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,7 +135,7 @@ endif()
 # shared WarpX library or a third party, like ImpactX, uses ablastr in a
 # shared library (e.g., for Python bindings), then we need relocatable code.
 option(ABLASTR_POSITION_INDEPENDENT_CODE
-    "Build ABLASTR with position independent code" ${WarpX_LIB})
+       "Build ABLASTR with position independent code" ON)
 mark_as_advanced(ABLASTR_POSITION_INDEPENDENT_CODE)
 
 # this defined the variable BUILD_TESTING which is ON by default
@@ -233,7 +233,7 @@ foreach(D IN LISTS WarpX_DIMS)
         list(APPEND _ALL_TARGETS app_${SD})
     endif()
 
-    if(WarpX_PYTHON OR WarpX_LIB)
+    if(WarpX_PYTHON OR (WarpX_LIB AND BUILD_SHARED_LIBS))
         set(ABLASTR_POSITION_INDEPENDENT_CODE ON CACHE BOOL
             "Build ABLASTR with position independent code" FORCE)
     endif()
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -135,7 +135,7 @@ endif()
 # shared WarpX library or a third party, like ImpactX, uses ablastr in a
 # shared library (e.g., for Python bindings), then we need relocatable code.
 option(ABLASTR_POSITION_INDEPENDENT_CODE
-    "Build ABLASTR with position independent code" ${WarpX_LIB})
+       "Build ABLASTR with position independent code" ON)
```
ax3l, line 138: We can default to ON here, which is the reasonable thing to do to use ABLASTR in both shared and static libs downstream.

There is a small performance penalty on CPU sections for relocatable code, which we do not worry about.

ax3l:
Separate PR: If someone builds Python & App at the same time, we have to document a workflow to copy also the `libamrex_Nd.so` files to the run dir if one decides to copy the application around...

cmake/dependencies/AMReX.cmake
```diff
@@ -87,9 +87,13 @@ macro(find_amrex)
         endif()
 
         # shared libs, i.e. for Python bindings, need relocatable code
-        #   openPMD: currently triggers shared libs (TODO)
-        if(WarpX_PYTHON OR (WarpX_LIB AND BUILD_SHARED_LIBS) OR ABLASTR_POSITION_INDEPENDENT_CODE OR WarpX_OPENPMD)
-            set(AMReX_PIC ON CACHE INTERNAL "")
+        if(WarpX_PYTHON OR
```
dpgrote, line 90: Since `ABLASTR_POSITION_INDEPENDENT_CODE` defaults to `ON`, doesn't this still always use `-fPIC` by default?

ax3l: Yes it will set this for AMReX as well (this causes at best negligible performance penalties on CPU for the sake of significantly easier usability of static libs. We did/could disable `-fPIC` for pure app builds or we have `WarpX_IPO` if we want to get some of these optimizations on CPU back. But our central kernels are not affected really, because we in-place lambda so much in `ParallelFor`s.)

The issue I want to avoid is actually more about the shared libs (below). Let me update the PR description.

dpgrote: Right, sorry, I misread this and was thinking that this PR was turning off the PIC. It is instead turning off the shared lib. I can approve.

dpgrote (approved):
Looks good!

PR #4214, opened 2023-08-22
Fix mass and charge reading from an external openPMD file

IlianCS:
This PR adds flushes after loading chunks which fix the reading of mass and charge from the openPMD file.


ax3l (approved):
Excellent, good catch!! :)

(This now works with non-constant record components for mass and charge.)

PR #4216, opened 2023-08-22
fix undefined behavior in btd intervals

RTSandberg:
This PR fixes a bug added in PR #3367, which provides interval parsing for the back-transformed diagnostic.  This bug caused unsafe memory access and had indeterminate behavior that likely resulted in a non-deterministic failure to add certain BTD snapshots, or worse to perhaps not register all snapshots on all ranks.

### DIFF
--- a/Source/Utils/Parser/IntervalsParser.cpp
+++ b/Source/Utils/Parser/IntervalsParser.cpp
@@ -193,12 +193,12 @@ utils::parser::BTDIntervalsParser::BTDIntervalsParser (
         const int start = temp_slice.getStart();
         const int period = temp_slice.getPeriod();
         int btd_iter_ind;
-        // for Slice temp_slice in m_slices,
-        // determine the index in m_btd_iterations where temp_slice's starting value goes
-        //
         // Implementation note:
-        // assuming the user mostly lists slices in ascending order,
         // start at the end of m_btd_iterations and search backward
+        // with the thinking that the user mostly lists slices in ascending order
+        //
+        // for Slice temp_slice in m_slices,
+        // determine the largest index in m_btd_iterations smaller than start
         if (m_btd_iterations.size() == 0)
         {
             btd_iter_ind = 0;
@@ -206,7 +206,7 @@ utils::parser::BTDIntervalsParser::BTDIntervalsParser (
         else
         {
             btd_iter_ind = m_btd_iterations.size() - 1;
-            while (start < m_btd_iterations[btd_iter_ind] and btd_iter_ind>0)
+            while (start < m_btd_iterations.at(btd_iter_ind) and btd_iter_ind>0)
             {
                 btd_iter_ind--;
             }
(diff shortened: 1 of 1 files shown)


WeiqunZhang (approved):

EZoni:
@RTSandberg @ax3l @WeiqunZhang @RevathiJambunathan 
If the only CI failure here is the one related to the docs (already fixed in #4215), we could also go ahead and merge this bug fix without rebasing on `development` and having to wait for all CI checks to complete again.

Source/Utils/Parser/IntervalsParser.cpp
```diff
@@ -193,43 +193,51 @@ utils::parser::BTDIntervalsParser::BTDIntervalsParser (
...
-        for (int ii = start; ii <= temp_slice.getStop(); ii += period)
+        for (int slice_iter = start; slice_iter <= temp_slice.getStop(); slice_iter += period)
         {
             if (m_btd_iterations.size() > 0)
             {
-                // find where iteration ii should go in m_btd_iterations
-                while (ii > m_btd_iterations[btd_iter_ind] && btd_iter_ind < static_cast<int>(m_btd_iterations.size()))
+                // increment btd_iter_ind until slice_iter <= m_btd_iterations.at(btd_iter_ind)
+                while (btd_iter_ind < static_cast<int>(m_btd_iterations.size())
+                        && slice_iter > m_btd_iterations.at(btd_iter_ind))
```
ax3l: - Using a named variable for the upper bound could increase clarity?
- the comment before the while loop is the same as the code and adds no further info right now (please check the comment and adjust as needed - please describe the `why` of the next lines)
```suggestion
                auto const num_btd_iterations = static_cast<int>(m_btd_iterations.size());

                // increment btd_iter_ind for each existing iteration,
                // if slice_iter is larger than an existing one
                while (btd_iter_ind < num_btd_iterations)
                        && slice_iter > m_btd_iterations.at(btd_iter_ind))
```

Source/Utils/Parser/IntervalsParser.cpp
```diff
@@ -193,43 +193,51 @@ utils::parser::BTDIntervalsParser::BTDIntervalsParser (
...
                 {
                     btd_iter_ind++;
                 }
-                if (ii != m_btd_iterations[btd_iter_ind])
+                // this is the place to insert slice_iter if it is not in m_btd_iterations
+                // if slice_iter > all entries in m_btd_iterations, append slice_iter
+                if (btd_iter_ind == static_cast<int>(m_btd_iterations.size()))
```
ax3l:
```suggestion
                if (btd_iter_ind == num_btd_iterations)
```

WeiqunZhang: One could argue that the old style is better because one does not need to go back to find out what `num_btw_iterations` is to know what this test is about.

ax3l: I don't think so with a proper name: is only few lines up and increases readability by naming things properly and shortening the line.
It also avoids confusion by avoiding to copy-pasting the cast + query of size.


(Also `.size()` on vectors is more costly than querying a member, but that is less of an issue.)

ax3l (approved):

PR #4218, opened 2023-08-22
Fix: WarpX Version

ax3l:
Fix the WarpX version info.

- [x] test on app
- [x] test in Python

Introduced in #3933 for app and #3474 for Python bindings.

### DIFF
--- a/Source/Python/pyWarpX.cpp
+++ b/Source/Python/pyWarpX.cpp
@@ -1,4 +1,4 @@
-/* Copyright 2021-2022 The ImpactX Community
+/* Copyright 2021-2023 The WarpX Community
  *
  * Authors: Axel Huebl
  * License: BSD-3-Clause-LBNL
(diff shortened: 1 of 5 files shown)


Source/Python/pyWarpX.cpp
```diff
@@ -71,8 +71,8 @@ PYBIND11_MODULE(PYWARPX_MODULE_NAME, m) {
 
     // API runtime version
     //   note PEP-440 syntax: x.y.zaN but x.y.z.devN
-#ifdef PYIMPACTX_VERSION_INFO
-    m.attr("__version__") = MACRO_STRINGIFY(PYIMPACTX_VERSION_INFO);
+#ifdef pyWarpX_VERSION_INFO
```
dpgrote: Should this be all caps?

ax3l: Yes, but gotta change it in the CMake file as well.

ax3l: Good catch!

Let me also update `setup.py`.

Source/Python/pyWarpX.cpp
```diff
@@ -71,8 +71,8 @@ PYBIND11_MODULE(PYWARPX_MODULE_NAME, m) {
...
+    m.attr("__version__") = MACRO_STRINGIFY(pyWarpX_VERSION_INFO);
```
ax3l:
```suggestion
#ifdef PYWARPX_VERSION_INFO
    m.attr("__version__") = MACRO_STRINGIFY(PYWARPX_VERSION_INFO);
```

dpgrote (approved):
This looks good!

PR #4219, opened 2023-08-23
Add custom input parameter t_lab for externally loaded species in the boosted frame

IlianCS:
This PR adds the possibility to parse a custom real parameter `t_lab` instead of imposing it from file (`t_lab_from_file`) for externally loaded species in the boosted frame.

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -718,7 +718,7 @@ Particle initialization
       ``<species_name>.mass`` (`double`) optional (default is read from openPMD file) when set this will be the charge of the physical particle represented by the injected macroparticles.
       ``<species_name>.z_shift`` (`double`) optional (default is no shift) when set this value will be added to the longitudinal, ``z``, position of the particles.
       ``<species_name>.t_lab_from_file`` (`bool`) optional (default is false) only read if warpx.gamma_boost > 1., it allows to set t_lab for the Lorentz Transform as being the time stored in the openPMD file.
-      ``<species_name>.t_lab`` (`double`) optional (default is 0) which allows to set a custom t_lab for the Lorentz Transform. This option overrides ``t_lab_from_file`` if the latter is defined.
+      ``<species_name>.t_lab`` (`double`) optional, which allows to set a custom t_lab for the Lorentz Transform. This option overrides ``t_lab_from_file`` if the latter is defined.
       Warning: ``q_tot!=0`` is not supported with the ``external_file`` injection style. If a value is provided, it is ignored and no re-scaling is done.
       The external file must include the species ``openPMD::Record`` labeled ``position`` and ``momentum`` (`double` arrays), with dimensionality and units set via ``openPMD::setUnitDimension`` and ``setUnitSI``.
       If the external file also contains ``openPMD::Records`` for ``mass`` and ``charge`` (constant `double` scalars) then the species will use these, unless overwritten in the input file (see ``<species_name>.mass``, ``<species_name>.charge`` or ``<species_name>.species_type``).
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -718,7 +718,7 @@ Particle initialization
       ``<species_name>.mass`` (`double`) optional (default is read from openPMD file) when set this will be the charge of the physical particle represented by the injected macroparticles.
       ``<species_name>.z_shift`` (`double`) optional (default is no shift) when set this value will be added to the longitudinal, ``z``, position of the particles.
       ``<species_name>.t_lab_from_file`` (`bool`) optional (default is false) only read if warpx.gamma_boost > 1., it allows to set t_lab for the Lorentz Transform as being the time stored in the openPMD file.
-      One can use instead ``<species_name>.t_lab`` (`double`) optional (default is 0) which allows to set a custom t_lab for the Lorentz Transform.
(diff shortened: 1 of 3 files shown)


Docs/source/usage/parameters.rst
```diff
@@ -717,7 +717,8 @@ Particle initialization
       ``<species_name>.charge`` (`double`) optional (default is read from openPMD file) when set this will be the charge of the physical particle represented by the injected macroparticles.
       ``<species_name>.mass`` (`double`) optional (default is read from openPMD file) when set this will be the charge of the physical particle represented by the injected macroparticles.
       ``<species_name>.z_shift`` (`double`) optional (default is no shift) when set this value will be added to the longitudinal, ``z``, position of the particles.
-      ``<species_name>.impose_t_lab_from_file`` (`bool`) optional (default is false) only read if warpx.gamma_boost > 1., it allows to set t_lab for the Lorentz Transform as being the time stored in the openPMD file.
+      ``<species_name>.t_lab_from_file`` (`bool`) optional (default is false) only read if warpx.gamma_boost > 1., it allows to set t_lab for the Lorentz Transform as being the time stored in the openPMD file.
+      One can use instead ``<species_name>.t_lab`` (`double`) optional (default is 0) which allows to set a custom t_lab for the Lorentz Transform.
```
```suggestion
      ``<species_name>.t_lab`` (`double`) optional (default is 0) which allows to set a custom t_lab for the Lorentz Transform. This option overrides ``t_lab_from_file`` if the latter is defined.
```
I would suggest to start the line with the option name. Moreover, I think that we should write in the documentation what happens if both `t_lab` and `t_lab_from_file` are defined.

lucafedeli88 (changes requested):
Thanks for this PR, @IlianCS ! I left just a small comment

Docs/source/usage/parameters.rst
```diff
@@ -717,7 +717,8 @@ Particle initialization
       ``<species_name>.charge`` (`double`) optional (default is read from openPMD file) when set this will be the charge of the physical particle represented by the injected macroparticles.
       ``<species_name>.mass`` (`double`) optional (default is read from openPMD file) when set this will be the charge of the physical particle represented by the injected macroparticles.
       ``<species_name>.z_shift`` (`double`) optional (default is no shift) when set this value will be added to the longitudinal, ``z``, position of the particles.
-      ``<species_name>.impose_t_lab_from_file`` (`bool`) optional (default is false) only read if warpx.gamma_boost > 1., it allows to set t_lab for the Lorentz Transform as being the time stored in the openPMD file.
+      ``<species_name>.t_lab_from_file`` (`bool`) optional (default is false) only read if warpx.gamma_boost > 1., it allows to set t_lab for the Lorentz Transform as being the time stored in the openPMD file.
+      ``<species_name>.t_lab`` (`double`) optional (default is 0) which allows to set a custom t_lab for the Lorentz Transform. This option overrides ``t_lab_from_file`` if the latter is defined.
```
dpgrote: Just a comment. Would it work to have this setup where if `t_lab` is not supplied, it means that it will be obtained from the file (instead of having the extra input parameter flag to specify that)?

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -593,12 +593,22 @@ PhysicalParticleContainer::AddPlasmaFromFile(ParticleReal q_tot,
         // assumption asserts: see PlasmaInjector
         openPMD::Iteration it = series->iterations.begin()->second;
         const ParmParse pp_species_name(species_name);
-        pp_species_name.query("impose_t_lab_from_file", impose_t_lab_from_file);
         double t_lab = 0._prt;
-        if (impose_t_lab_from_file) {
+        pp_species_name.query("t_lab_from_file", t_lab_from_file);
+        if (t_lab_from_file) {
```
ax3l, line 598:
```suggestion
        bool t_lab_from_file_set = pp_species_name.query("t_lab_from_file", t_lab_from_file);
        if (t_lab_from_file_set) {
```

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -593,12 +593,22 @@ PhysicalParticleContainer::AddPlasmaFromFile(ParticleReal q_tot,
...
             // Impose t_lab as being the time stored in the openPMD file
             t_lab = it.time<double>() * it.timeUnitSI();
         }
+        const bool t_lab_from_input =
+            utils::parser::queryWithParser(pp_species_name, "t_lab", t_lab);
+        if ( t_lab_from_input && t_lab_from_file ){
```
ax3l, line 604:
```suggestion
        if ( t_lab_from_input && t_lab_from_file_set ){
```

ax3l (changes requested):
What I would do is this logic:
- is `t_lab_from_file` set?
- yes: read it
- no: require that `t_lab` is set in the input (by using get instead of query) - remove `(default is 0)` for it

IlianCS:
Thank you for the review @dpgrote @ax3l.
Actually I think that the current implementation is correct. I want the default behavior to be that the injection is done with `t_lab=0`, whether or not there is a time stored in the openPMD file (e.g when the file comes from a first WarpX simulation).

Then, the user would need to explicitly use the input parameter `t_lab_from_file` if they want to use the time in the file or `t_lab` if they want to use a custom time.

Still, we should indeed remove "default is 0" because when `t_lab_from_file` is provided and not `t_lab` then `t_lab=0` does not overwrite the time from the openPMD file (despite what the description can imply).

PR #4220, opened 2023-08-23
Fix Repeated Plasma Lens: Start

ax3l:
The current implementation has problems finding the right index of the current plasma lens if:
- the lens started before z=0
- the lens did not start between z=0...first-period

This should fix it.

- [x] testing

### DIFF
--- a/Source/Particles/Gather/GetExternalFields.H
+++ b/Source/Particles/Gather/GetExternalFields.H
@@ -145,6 +145,7 @@ struct GetExternalEBField
             const amrex::ParticleReal gamma = std::sqrt(1._prt + (uxp*uxp + uyp*uyp + uzp*uzp)*inv_c2);
             const amrex::ParticleReal vzp = uzp/gamma;
 
+            // the current slice in z between now and the next time step
             amrex::ParticleReal zl = z;
             amrex::ParticleReal zr = z + vzp*m_dt;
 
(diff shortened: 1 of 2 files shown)


Source/Particles/Gather/GetExternalFields.H
```diff
@@ -153,27 +153,31 @@ struct GetExternalEBField
...
-            amrex::ParticleReal const zr_bounded = std::min(std::max(zr, lens_start), lens_end);
-            amrex::ParticleReal const frac = ((zr - zl) == 0._rt ? 1._rt : (zr_bounded - zl_bounded)/(zr - zl));
-
-            // Note that "+=" is used since the fields may have been set above
-            // if a different E or Btype was specified.
-            Ex += x*frac*m_repeated_plasma_lens_strengths_E[i_lens];
-            Ey += y*frac*m_repeated_plasma_lens_strengths_E[i_lens];
-            Bx += +y*frac*m_repeated_plasma_lens_strengths_B[i_lens];
-            By += -x*frac*m_repeated_plasma_lens_strengths_B[i_lens];
+            if (zl > m_repeated_plasma_lens_starts[0]) {
```
ax3l: Here, we should also exclude if we are outside of the last plasma lens, otherwise we can have an out-of-bound access as well if we have further logic after the last lens.

Or do we simply repeat the lenses after `m_n_lenses` have passed? The `i_lens = i_lens % m_n_lenses;` seems to suggest so?

dpgrote:
This looks good, though it is changing the code somewhat, since the old code would allow periodic repeats of the whole set of lenses. But that probably doesn't matter much.

Since you are making changes, do you want to add a new input parameter that specifies the start of the lenses, something like `repeated_plasma_lens_zinit`, instead of hardwiring it to zero?

dpgrote:
BTW,  a comment here is that this repeated plasma lens should probably be removed anyway and use the accelerator lattice, since also have the plasma lens element.

ax3l:
In that case, I would not add more params now and just fix it somewhat :)

dpgrote (approved):
Ok, looks good.

PR #4223, opened 2023-08-23
Docs: Perlmutter (NERSC) Update

ax3l:
Update slurm changes and pinning.
Changing the GPU order in Slurm currently does not seem to show performance differences with these options set.

### DIFF
--- a/Tools/machines/perlmutter-nersc/perlmutter_gpu.sbatch
+++ b/Tools/machines/perlmutter-nersc/perlmutter_gpu.sbatch
@@ -17,24 +17,26 @@
 # A100 80GB (256 nodes)
 #S BATCH -C gpu&hbm80g
 #SBATCH --exclusive
+# ideally single:1, but NERSC cgroups issue
 #SBATCH --gpu-bind=none
+#SBATCH --ntasks-per-node=4
 #SBATCH --gpus-per-node=4
 #SBATCH -o WarpX.o%j
 #SBATCH -e WarpX.e%j
 
 # executable & inputs file or python interpreter & PICMI script here
 EXE=./warpx
-INPUTS=inputs_small
+INPUTS=inputs
 
 # pin to closest NIC to GPU
 export MPICH_OFI_NIC_POLICY=GPU
 
 # threads for OpenMP and threaded compressors per MPI rank
+#   note: 16 avoids hyperthreading (32 virtual cores, 16 physical)
 export SRUN_CPUS_PER_TASK=16
 
-# depends on https://github.com/ECP-WarpX/WarpX/issues/2009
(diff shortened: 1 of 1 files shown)


kngott:
Were those tests performed with GPU_AWARE_MPI? That's when the GPU ordering should have an effect. No effect is extremely odd, as some configuration should definitely perform better than the others (distance from GPU to NIC is lowest somewhere).

kngott:
This just worked for me on Perlmutter with the KPP:

```#!/bin/bash -l
#SBATCH -C gpu
#SBATCH -t 00:30:00 
#SBATCH -J WarpX_KPP
#SBATCH -o WarpX_KPP.o%A_%a
#SBATCH -A nstaff_g
#SBATCH -N 2
#SBATCH -q regular 
#SBATCH --exclusive
#SBATCH --gpu-bind=none
#SBATCH --ntasks-per-node=4
#SBATCH --gpus-per-node=4
#SBATCH --array=0-2

EXE=../warpx.3d.MPI.CUDA.DP.PDP.OPMD.QED
INPUTS=./inputs

export MPICH_OFI_NIC_POLICY=GPU
GPU_AWARE_MPI="amrex.use_gpu_aware_mpi=1"
SRUN_FLAGS="--cpus-per-task=16 --cpu-bind=cores"

srun --cpu-bind=cores bash -c "
    export CUDA_VISIBLE_DEVICES=\$((3-SLURM_LOCALID));
    ${EXE} ${INPUTS} ${GPU_AWARE_MPI}" \
```

RTSandberg:
Both approaches work for me; Kevin's ran faster for me by a factor of about 50% but that was only one run each so my statistics are bad

ax3l:
@RTSandberg updated and ready to test :)

kngott:
Looks good to me. I won't approve so it isn't accidentally merged before testing.

There might be some argument that gpu-bind=closest is the preferred, (not single:1), but 🤷.

RTSandberg:
I tested this with and without the line ``export CUDA_VISIBLE_DEVICES=\$((3-SLURM_LOCALID));``
a) it runs!
b) runtime is the same with and without (1145 vs 1143 s) on my test script

ax3l:
Will merge for now to unbreak docs - missing time difference might also be from load imbalance.

RTSandberg:
I realized the ``amrex.the_arena_is_managed`` parameter was set inconsistently in my previous test, so the comparison I ran between two different submission scripts is invalid and they should probably be tested again

RTSandberg:
[output_14301424.txt](https://github.com/ECP-WarpX/WarpX/files/12458857/output_14301424.txt)
[perlmutter_batch.txt](https://github.com/ECP-WarpX/WarpX/files/12458867/perlmutter_batch.txt)
[inputs.txt](https://github.com/ECP-WarpX/WarpX/files/12458868/inputs.txt)

PR #4224, opened 2023-08-24
CI: Shrink NVHPC Size

ax3l:
Try to shrink the size occupied in the NVHPC CI runs. Currently running out-of-disk...

Install:
```
After this operation, 9028 MB of additional disk space will be used.
```
o.0 our limit is 10G or so on CI runners.

Also filed as suggestion to Nvidia as Bug no. 4257086.


lucafedeli88 (approved):
Thanks for this PR, @ax3l !

PR #4225, opened 2023-08-24
Doc: Summit CUDA 11.7

ax3l:
Working around compiler issues seen with 11.3.
This update requires a rebuild of all dependencies.

- [ ] tested


ax3l:
The CUDA 11.7 update shows issues with cuBLAS for our dgemm.
Will stay with 11.3 and we simplify the code in AMReX instead to help nvcc help itself: https://github.com/AMReX-Codes/amrex/pull/3510

PR #4226, opened 2023-08-24
Doc: HPC Python Separate (Static/Shared AMReX)

ax3l:
Currently, due to global variables used in AMReX to store implicit state instead of using handlers, we cannot build shared libraries for Python (pyAMReX + pyWarpX + pyImpactX) with a *static* `libamrex_Nd.a`. Thus, we automatically build a shared AMReX library in our superbuild the moment that we need Python bindings.

That can confuse people that just use the C++ application with AMReX inputs files, which then also depends on this shared AMReX library (which also needs to be copied around, for instance).

This update the HPC docs to just build App and bindings twice (twice the time spent), to work around this current imitation of AMReX by linking the app statically and the Python bindings shared.

Also, for dependency install fetch via `git fetch --prune`.
Fixes an issue first seen in blaspp/lapackpp where fetch failed because an unused remote branch was gone.

### DIFF
--- a/Docs/source/install/hpc/adastra.rst
+++ b/Docs/source/install/hpc/adastra.rst
@@ -85,16 +85,24 @@ Finally, since Adastra does not yet provide software modules for some of our dep
 Compilation
 -----------
 
-Use the following :ref:`cmake commands <building-cmake>` to compile:
+Use the following :ref:`cmake commands <building-cmake>` to compile the application:
 
 .. code-block:: bash
 
    cd $HOME/src/warpx
    rm -rf build_adastra
 
-   cmake -S . -B build_adastra -DWarpX_COMPUTE=HIP -DWarpX_PSATD=ON -DWarpX_QED_TABLE_GEN=ON -DWarpX_PYTHON=ON -DWarpX_DIMS="1;2;RZ;3"
+   cmake -S . -B build_adastra -DWarpX_COMPUTE=HIP -DWarpX_PSATD=ON -DWarpX_QED_TABLE_GEN=ON -DWarpX_DIMS="1;2;RZ;3"
    cmake --build build_adastra -j 16
-   cmake --build build_adastra -j 16 --target pip_install
+
+and the following for the Python module:
+
+.. code-block:: bash
+
+   rm -rf build_adastra_py
+
+   cmake -S . -B build_adastra_py -DWarpX_COMPUTE=HIP -DWarpX_PSATD=ON -DWarpX_QED_TABLE_GEN=ON -DWarpX_APP=OFF -DWarpX_PYTHON=ON -DWarpX_DIMS="1;2;RZ;3"
+   cmake --build build_adastra_py -j 16 --target pip_install
 
 **That's it!**
 The WarpX application executables are now in ``$HOME/src/warpx/build_adastra/bin/`` and we installed the ``pywarpx`` Python module.
--- a/Docs/source/install/hpc/adastra.rst
+++ b/Docs/source/install/hpc/adastra.rst
@@ -95,6 +95,7 @@ Use the following :ref:`cmake commands <building-cmake>` to compile the applicat
    cmake -S . -B build_adastra -DWarpX_COMPUTE=HIP -DWarpX_PSATD=ON -DWarpX_QED_TABLE_GEN=ON -DWarpX_DIMS="1;2;RZ;3"
    cmake --build build_adastra -j 16
 
+The WarpX application executables are now in ``$HOME/src/warpx/build_adastra/bin/``.
 Additionally, the following commands will install WarpX as a Python module:
 
 .. code-block:: bash
@@ -104,8 +105,6 @@ Additionally, the following commands will install WarpX as a Python module:
    cmake -S . -B build_adastra_py -DWarpX_COMPUTE=HIP -DWarpX_PSATD=ON -DWarpX_QED_TABLE_GEN=ON -DWarpX_APP=OFF -DWarpX_PYTHON=ON -DWarpX_DIMS="1;2;RZ;3"
    cmake --build build_adastra_py -j 16 --target pip_install
 
-**That's it!**
-The WarpX application executables are now in ``$HOME/src/warpx/build_adastra/bin/`` and we installed the ``pywarpx`` Python module.
 
 Now, you can :ref:`submit Adstra compute jobs <running-cpp-adastra>` for WarpX :ref:`Python (PICMI) scripts <usage-picmi>` (:ref:`example scripts <usage-examples>`).
 Or, you can use the WarpX executables to submit Adastra jobs (:ref:`example inputs <usage-examples>`).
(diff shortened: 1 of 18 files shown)


ax3l:
fyi cc @berceanu for Karolina, @erny123 @floresv299 for HPC3

Docs/source/install/hpc/adastra.rst
```diff
@@ -85,16 +85,24 @@ Finally, since Adastra does not yet provide software modules for some of our dep
 Compilation
 -----------
 
-Use the following :ref:`cmake commands <building-cmake>` to compile:
+Use the following :ref:`cmake commands <building-cmake>` to compile the application:
```
RemiLehe: I think we refer to it as "the application **executable**" in the rest of the text below.
```suggestion
Use the following :ref:`cmake commands <building-cmake>` to compile the application executable:
```

Docs/source/install/hpc/adastra.rst
```diff
@@ -85,16 +85,24 @@ Finally, since Adastra does not yet provide software modules for some of our dep
...
 
    cd $HOME/src/warpx
    rm -rf build_adastra
 
-   cmake -S . -B build_adastra -DWarpX_COMPUTE=HIP -DWarpX_PSATD=ON -DWarpX_QED_TABLE_GEN=ON -DWarpX_PYTHON=ON -DWarpX_DIMS="1;2;RZ;3"
+   cmake -S . -B build_adastra -DWarpX_COMPUTE=HIP -DWarpX_PSATD=ON -DWarpX_QED_TABLE_GEN=ON -DWarpX_DIMS="1;2;RZ;3"
    cmake --build build_adastra -j 16
-   cmake --build build_adastra -j 16 --target pip_install
+
+and the following for the Python module:
```
RemiLehe:
```suggestion
Or, alternatively, use the following command if you want to use WarpX as a Python module, instead of an executable:
```

Docs/source/install/hpc/adastra.rst
```diff
@@ -85,16 +85,24 @@ Finally, since Adastra does not yet provide software modules for some of our dep
...
 
 .. code-block:: bash
 
    cd $HOME/src/warpx
    rm -rf build_adastra
 
-   cmake -S . -B build_adastra -DWarpX_COMPUTE=HIP -DWarpX_PSATD=ON -DWarpX_QED_TABLE_GEN=ON -DWarpX_PYTHON=ON -DWarpX_DIMS="1;2;RZ;3"
+   cmake -S . -B build_adastra -DWarpX_COMPUTE=HIP -DWarpX_PSATD=ON -DWarpX_QED_TABLE_GEN=ON -DWarpX_DIMS="1;2;RZ;3"
    cmake --build build_adastra -j 16
-   cmake --build build_adastra -j 16 --target pip_install
```
RemiLehe, line 97: **That's it!** The WarpX application executables are now in ``$HOME/src/warpx/build_adastra/bin/``.

Docs/source/install/hpc/adastra.rst
```diff
@@ -85,16 +85,24 @@ Finally, since Adastra does not yet provide software modules for some of our dep
...
+
+.. code-block:: bash
+
+   rm -rf build_adastra_py
+
+   cmake -S . -B build_adastra_py -DWarpX_COMPUTE=HIP -DWarpX_PSATD=ON -DWarpX_QED_TABLE_GEN=ON -DWarpX_APP=OFF -DWarpX_PYTHON=ON -DWarpX_DIMS="1;2;RZ;3"
+   cmake --build build_adastra_py -j 16 --target pip_install
 
 **That's it!**
 The WarpX application executables are now in ``$HOME/src/warpx/build_adastra/bin/`` and we installed the ``pywarpx`` Python module.
```
RemiLehe:
```suggestion
```

dpgrote (approved):
This looks good. It has been my habit to build the Python version separately anyway.

Docs/source/install/hpc/adastra.rst
```diff
@@ -85,16 +85,24 @@ Finally, since Adastra does not yet provide software modules for some of our dep
...
-   cmake --build build_adastra -j 16 --target pip_install
+
+and the following for the Python module:
+
+.. code-block:: bash
+
+   rm -rf build_adastra_py
+
+   cmake -S . -B build_adastra_py -DWarpX_COMPUTE=HIP -DWarpX_PSATD=ON -DWarpX_QED_TABLE_GEN=ON -DWarpX_APP=OFF -DWarpX_PYTHON=ON -DWarpX_DIMS="1;2;RZ;3"
+   cmake --build build_adastra_py -j 16 --target pip_install
```
RemiLehe, line 106: The above command automatically installs the ``pywarpx`` Python module in the current Python environment.

RemiLehe:
I added a few suggestions in order to clarify the instructions, in the case of Ad Astra.

@ax3l If you agree with these suggestions, they would need to be propagated to the other files.

Docs/source/install/hpc/adastra.rst
ax3l: You can install both at the same time, thus I would not write *alternatively*.

ax3l:
@RemiLehe thanks for the suggestions! My intention is to not separate the App Exe and Py Module install too far from each other, both visually and in text.
I still want that people by default just install both of them, so they have all features at hand. Thus I kept the filler text between the two paragraphs very brief and short, so it is clear both sections shall be executed for full features by default.

Docs/source/install/hpc/adastra.rst
ax3l: See comment below. I like to summarize the result of both steps afterwards, so people do both to have a full install by default.

Docs/source/install/hpc/adastra.rst
ax3l: Maybe?
```suggestion
Additionally, the following commands will install WarpX as a Python module:
```

Docs/source/install/hpc/adastra.rst
ax3l: Yes, since the workflow above is to activate the warpx.profile script and run the install deps - there is always a python venv active (that users do not even need to take care of or know about).

Docs/source/install/hpc/adastra.rst
```diff
@@ -85,16 +85,24 @@ Finally, since Adastra does not yet provide software modules for some of our dep
...
 
    cd $HOME/src/warpx
    rm -rf build_adastra
 
-   cmake -S . -B build_adastra -DWarpX_COMPUTE=HIP -DWarpX_PSATD=ON -DWarpX_QED_TABLE_GEN=ON -DWarpX_PYTHON=ON -DWarpX_DIMS="1;2;RZ;3"
+   cmake -S . -B build_adastra -DWarpX_COMPUTE=HIP -DWarpX_PSATD=ON -DWarpX_QED_TABLE_GEN=ON -DWarpX_DIMS="1;2;RZ;3"
    cmake --build build_adastra -j 16
-   cmake --build build_adastra -j 16 --target pip_install
+
+Additionally, the following commands will install WarpX as a Python module:
```
ax3l, line 99:
```suggestion
The WarpX application executables are now in ``$HOME/src/warpx/build_adastra/bin/``.
Additionally, the following commands will install WarpX as a Python module:
```

RemiLehe (approved):

PR #4227, opened 2023-08-24
Fix: openPMD 1D/2D3V Constant Positions

ax3l:
Add constant record components for position and positionOffset in 1D3V and 2D3V simulations.
Follows our conventions in
  https://warpx.readthedocs.io/en/latest/developers/dimensionality.html#conventions
and simplifies post-processing, e.g., emittance calculations.

### DIFF
--- a/Source/Diagnostics/WarpXOpenPMD.cpp
+++ b/Source/Diagnostics/WarpXOpenPMD.cpp
@@ -214,26 +214,31 @@ namespace detail
     /** Return the component labels for particle positions
      */
     inline std::vector< std::string >
-    getParticlePositionComponentLabels ()
+    getParticlePositionComponentLabels (bool ignore_dims=false)
     {
         using vs = std::vector< std::string >;
+        vs positionComponents;
+        if (ignore_dims) {
+            positionComponents = vs{"x", "y", "z"};
+        } else {
 #if defined(WARPX_DIM_1D_Z)
-        vs positionComponents{"z"};
+            positionComponents = vs{"z"};
 #elif defined(WARPX_DIM_XZ)
-        vs positionComponents{"x", "z"};
+            positionComponents = vs{"x", "z"};
 #elif defined(WARPX_DIM_RZ)
-        // note: although we internally store particle positions
-        //       for AMReX in r,z and a theta attribute, we
-        //       actually need them for algorithms (e.g. push)
-        //       and I/O in Cartesian.
-        //       Other attributes like momentum are consequently
-        //       stored in x,y,z internally.
-        vs positionComponents{"x", "y", "z"};
+            // note: although we internally store particle positions
+            //       for AMReX in r,z and a theta attribute, we
+            //       actually need them for algorithms (e.g. push)
+            //       and I/O in Cartesian.
+            //       Other attributes like momentum are consequently
+            //       stored in x,y,z internally.
+            positionComponents = vs{"x", "y", "z"};
 #elif defined(WARPX_DIM_3D)
-        vs positionComponents{"x", "y", "z"};
+            positionComponents = vs{"x", "y", "z"};
 #else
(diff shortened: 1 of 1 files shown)


RemiLehe (approved):

IlianCS:
Works on my side, thank you!
In addition, I think we should use the `<species_names>.y_m` for the gaussian_beam injection style and impose the y coordinate accordingly.

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -1039,6 +1051,14 @@ WarpXOpenPMDPlot::SetConstParticleRecordsEDPIC (
     }
     currSpecies["charge"][scalar].makeConstant( charge );
     currSpecies["mass"][scalar].makeConstant( mass );
+    //   convention: in 1D3V and 2D3V, omitted positions are set to zero
+#if defined(WARPX_DIM_1D_Z)
+    currSpecies["position"]["x"].makeConstant( 0._prt );
+    currSpecies["position"]["y"].makeConstant( 0._prt );
+#endif
+#if defined(WARPX_DIM_XZ)
+    currSpecies["position"]["y"].makeConstant( 0._prt );
```
IlianCS, line 1060: maybe use the variable from the input parameter `<species_name>.y_m` for gaussian beam? (and `<species_name>.x_m` in 1D)

IlianCS (changes requested):

ax3l:
>  In addition, I think we should use the <species_names>.y_m for the gaussian_beam injection style and impose the y coordinate accordingly.

There is a general convention for that geometry already tough, when would you need a non-zero y in 2D3V? @IlianCS 
https://warpx.readthedocs.io/en/latest/developers/dimensionality.html#conventions

Source/Diagnostics/WarpXOpenPMD.cpp
ax3l: I think that would not be ideal, as it breaks the current geometry assumption:
https://github.com/ECP-WarpX/WarpX/pull/4227#issuecomment-1693626344

ax3l: We can add a warning to  the input of `y_m` in 2D3V (and lower d): warn the user that the value is overwritten with 0 if they set it to something else.

IlianCS:
> > In addition, I think we should use the <species_names>.y_m for the gaussian_beam injection style and impose the y coordinate accordingly.
> 
> There is a general convention for that geometry already tough, when would you need a non-zero y in 2D3V? @IlianCS https://warpx.readthedocs.io/en/latest/developers/dimensionality.html#conventions

It is about being consistent with the inputs script. Please check the other comments above. I think that is a good idea to put a warning at least to say that we overwrite the values because we require the user to parse all `x_m`, `y_m`, `z_m` and `x_rms`, `y_rms`, `z_rms` even if it is a 1D or 2D simulation.

Source/Diagnostics/WarpXOpenPMD.cpp
IlianCS: If we have to follow the convention, could it be possible to prevent reading `y_m, y_rms` in 2D (and `x_m, x_rms` in 1D) from the inputs file and directly impose the values? I think that would be less confusing.

IlianCS: Or make it optionnal and if one were to parse these parameters so we should implement this warning message.

ax3l: yes, that's what I mean as well. Would be a good follow-up to warn the user if they set it to non-zero and run a 1D/2D sim. Do you want to take that on? :)

ax3l:
Fully agree, great idea! Do you like to add such warnings in the input parsing sections? :pray:

Issue #4193, opened 2023-08-16
Missing particles in BTD output


RevathiJambunathan:
This is an issue reported by @joshualudwig8 

A simulation performed by them have strips of particles missing from the BTD output

![Screenshot 2023-08-16 at 11 34 04 AM](https://github.com/ECP-WarpX/WarpX/assets/41089244/4069a911-1c3d-480a-9ee6-58d399df7c0b)
![Screenshot 2023-08-16 at 11 34 38 AM](https://github.com/ECP-WarpX/WarpX/assets/41089244/518389d5-0940-4748-b978-35421e3c76f5)
![Screenshot 2023-08-16 at 11 35 15 AM](https://github.com/ECP-WarpX/WarpX/assets/41089244/433bd813-8851-4446-aa1e-a23c88a2e02c)

@jlvay was able to reproduce a similar error with a reduced number of snapshots (41 snapshots) on summit
![Screenshot 2023-08-16 at 11 36 29 AM](https://github.com/ECP-WarpX/WarpX/assets/41089244/6d487b1c-a9d9-4189-b63f-f291eb6d24d7)

RevathiJambunathan:
Update so far : 
1. I first changed the output format to plotfile instead of openpmd and the code crashed with an ASSERT 
![Screenshot 2023-08-11 at 5 01 06 PM](https://github.com/ECP-WarpX/WarpX/assets/41089244/a6b4f608-4601-422e-9c0d-3f872a2c23d2)
Looking into this further -- it became clear that the input has ionization turned on and `ionization_level` is a runtime particle attribute added as a result 
However, for BTD we do not copy this attribute (we only copy/store position, momenta and particle weight) 
So I fixed this such that we do not write the ionization level and its not included in the output consistent with the implementation


2. The number of box-arrays was assumed to be 1, but in some cases where the non-moving window direction had more cells than buffer-size, we would have to boxes in a box-array. 
This is a minor detail and is now fixed. 

More to investigate as the code crashes when writing particles with plotfile format

RevathiJambunathan:
https://github.com/ECP-WarpX/WarpX/pull/4228 
With this PR, the code does not crash with plotfile output on summit. I used @jlvay's modified input file with 41 snapshots 

here is the electrons in red : 
![Screenshot 2023-08-24 at 1 51 07 PM](https://github.com/ECP-WarpX/WarpX/assets/41089244/1423aa98-0e39-478e-8b6e-2b977f5b8d43)

[...]

PR #4228, opened 2023-08-24
Fix ParticleContainer for BTD plotfile

RevathiJambunathan:
Screenshot 2023-08-24 at 1 49 01 PM
In trying to debug Issue #4193, I would get the above error with plotfile

This PR fixes it by ensuring that the real_names and int_names are obtained for the correct particle container


RemiLehe (approved):

ax3l (approved):
Great! The pinned container looks identical to the openPMD version :+1: 

Let's double check the runtime attributes...
https://github.com/ECP-WarpX/WarpX/blob/b88db1668e1882f6542c5f2bd04ca3ef4cf2034f/Source/Diagnostics/WarpXOpenPMD.cpp#L551-L569
seems similar?

RevathiJambunathan:
@ax3l Both openpmd and plotfile will pick the runtime attributes in the particle container (pinned_pc for BTD) 
However, this particle container only copies position, momenta, and particle weight

So in essence, the run time attributes are not supported for BTD at the moment.

PR #4229, opened 2023-08-24
Fix Python wrapper to extract `theta`

RemiLehe:
It seems that there is typo that broke the `get_particle_theta` code, but was not detected by CI.

Follow-up to #3474


ax3l (approved):
Thank you, yes that must have slipped through I think :)

I see no CI coverage of it, thus it did not fail anything.

roelof-groenewald (approved):
Thanks for fixing!

PR #4231, opened 2023-08-25
Do not call pre-processing routine in Python initialization

RemiLehe:
We used to call the following pre-processing routines in `main.cpp` (which is used when building the C++ executable):
- `ParseGeometryInput`
- `ConvertLabParamsToBoost`
- `ReadBCParams`

As a consequence, these routines are also directly called in `_libwarpx.py` (which is used when WarpX is run as a Python module).

However, in https://github.com/ECP-WarpX/WarpX/pull/4104, these calls were moved from `main.cpp` to inside the function `MakeWarpX` (which is called by `GetInstance`). Since `_libwarpx.py` itself does call `GetInstance`, does not need to call the pre-processing routines anymore (otherwise these routines are called twice, which is erroneous - esp. in the case of boosted-frame simulations)

### DIFF
--- a/Python/pywarpx/_libwarpx.py
+++ b/Python/pywarpx/_libwarpx.py
@@ -169,10 +169,6 @@ class LibWarpX():
         if argv is None:
             argv = sys.argv
         self.amrex_init(argv, mpi_comm)
-        self.libwarpx_so.convert_lab_params_to_boost()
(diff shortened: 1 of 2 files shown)


dpgrote (approved):
This looks good.

ax3l (approved):
Good catch, thank you!!

Issue #4234, opened 2023-08-25
Crash with large 3D simulations on Frontier


tmsclark2:
Hi, 
I got crashs with large 3D simulations on Frontier. The crash is concerning a `MPI_Waitall` routine :
``` 
MPICH ERROR [Rank 8281] [job id 1410467.0] [Wed Aug 23 06:05:55 2023] [frontier01982] - Abort(136982671) (rank 8281 in comm 0): Fatal error in PMPI_Waitall: Other MPI error, error stack:
PMPI_Waitall(339)..............: MPI_Waitall(count=26, req_array=0x1006e9b0, status_array=0xfefafb0) failed
MPIR_Waitall(167)..............: 
MPIR_Waitall_impl(51)..........: 
MPID_Progress_wait(193)........: 
MPIDI_Progress_test(89)........: 
MPIDI_OFI_handle_cq_error(1062): OFI poll failed (ofi_events.c:1064:MPIDI_OFI_handle_cq_error:Input/output error - UNDELIVERABLE)
``` 

 Those crashs are not at the same time and don't produce backtraces. 

Here is the error output of the simulations and the submit and input file of one of those cases to reproduce it : [input.txt](https://github.com/ECP-WarpX/WarpX/files/12442633/input.txt) [out.txt](https://github.com/ECP-WarpX/WarpX/files/12442634/out.txt) [batch.txt](https://github.com/ECP-WarpX/WarpX/files/12442713/batch.txt)

Here are the modules used for the compilation :
[warpx_profile.txt](https://github.com/ECP-WarpX/WarpX/files/12443912/warpx_profile.txt)

ax3l:
I think this might be a known issue:
https://docs.olcf.ornl.gov/systems/frontier_user_guide.html#olcfdev-1597-ofi-poll-failed-undeliverable-errors

Let's try with this set:
```bash
export MPICH_SMP_SINGLE_COPY_MODE=NONE
```

**Update:** and this set at the same time, too:
```bash
export FI_CXI_RX_MATCH_MODE=software
```

ax3l:
According to our earlier ticket in Summer, OLCFHELP-12850 cc @n01r, we were recommended at the time to also set

```bash
export FI_CXI_RX_MATCH_MODE=software
```

We saw no improvement setting the three env vars (this PR + the one above) at the time. `FI_CXI_RX_MATCH_MODE` is not mentioned in the OLCF docs at this time.
[...]

PR #4235, opened 2023-08-25
Frontier (OLCF): MPI_Waitall issues

ax3l:
Document to use `export MPICH_SMP_SINGLE_COPY_MODE=NONE` and `export FI_CXI_RX_MATCH_MODE=software` in job scripts.

According to earlier attempts by Marco, this might not yet solve the issue, but it is the [documented workaround from the OLCF side so far](https://docs.olcf.ornl.gov/systems/frontier_user_guide.html#olcfdev-1597-ofi-poll-failed-undeliverable-errors), so we should add it for now.

X-ref #4234

### DIFF
--- a/Tools/machines/frontier-olcf/submit.sh
+++ b/Tools/machines/frontier-olcf/submit.sh
@@ -31,6 +31,12 @@
 # or, less invasive:
 export FI_MR_CACHE_MONITOR=memhooks  # alternative cache monitor
 
+# Seen since August 2023
+# OLCFDEV-1597: OFI Poll Failed UNDELIVERABLE Errors
+# https://docs.olcf.ornl.gov/systems/frontier_user_guide.html#olcfdev-1597-ofi-poll-failed-undeliverable-errors
+export MPICH_SMP_SINGLE_COPY_MODE=NONE
+export FI_CXI_RX_MATCH_MODE=software
+
 # note (9-2-22, OLCFDEV-1079)
 # this environment setting is needed to avoid that rocFFT writes a cache in
 # the home directory, which does not scale.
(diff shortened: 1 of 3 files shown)


Tools/machines/frontier-olcf/submit.sh
```diff
@@ -31,6 +31,11 @@
 # or, less invasive:
 export FI_MR_CACHE_MONITOR=memhooks  # alternative cache monitor
 
+# Seen since August 2023
+# OLCFDEV-1597: OFI Poll Failed UNDELIVERABLE Errors
+# https://docs.olcf.ornl.gov/systems/frontier_user_guide.html#olcfdev-1597-ofi-poll-failed-undeliverable-errors
+export MPICH_SMP_SINGLE_COPY_MODE=NONE
```
ax3l, line 37: According to our earlier ticket in Summer, OLCFHELP-12850, we were recommended at the time to also set

```bash
export FI_CXI_RX_MATCH_MODE=software
```

We saw no improvement setting the three env vars (this PR + the one above) at the time. `FI_CXI_RX_MATCH_MODE` is not mentioned in the OLCF docs at this time.

n01r: My last test for this was on June 7th which had all 3 options and it ran for 12 hours without failing.
That was on 384 nodes and `export FI_CXI_RX_MATCH_MODE=software` was the one that was newly added. However, I did not have the time to test if this was reproducible. Apologies! :pray:

n01r: Otherwise, the issue appeared randomly throughout the simulations and it would run fine for hours sometimes.

ax3l: Will add!

ax3l:
@tmsclark @lucafedeli88 I added comments for this in LUMI's template, but did not comment them in yet since you did not see the same issues there (yet?).

Please feel free to update LUMI anytime afterwards if you should see the  same issue there.

Issue #4236, opened 2023-08-25
Crash with large 3D simulations on LUMI


tmsclark2:
Hi,
I got crashs with large 3D simulations on LUMI. The crash is concerning a MPI_Allgather routine :

```
MPICH ERROR [Rank 0] [job id 4292261.0] [Thu Aug  3 00:01:58 2023] [nid006593] - Abort(1616271) (rank 0 in comm 0): Fatal error in PMPI_Init_thread: Other MPI error, error stack:
MPIR_Init_thread(170).......:
MPID_Init(501)..............:
MPIDI_OFI_mpi_init_hook(805):
MPIDU_bc_table_create(204)..:  PMI_Allgather failed: -1
``` 

This crash happens before warpx starts and does not produce traces.

Here is the error output of the simulations and the submit file : [warpx-4292261.txt](https://github.com/ECP-WarpX/WarpX/files/12444020/warpx-4292261.txt) [batch.txt](https://github.com/ECP-WarpX/WarpX/files/12444022/batch.txt)

Here are the modules used for the compilation : [Recipe_warpx.txt](https://github.com/ECP-WarpX/WarpX/files/12444025/Recipe_warpx.txt)

ax3l:
This looks exactly like this bug:
https://docs.olcf.ornl.gov/systems/crusher_quick_start_guide.html#olcfdev-1655-occasional-seg-fault-during-mpi-init

ax3l:
@tmsclark2 can you please open a support ticket with LUMI about this, pointing this to the OLCF issue and asking what they recommend to do at LUMI?
Please CC the people on this issue in your ticket :pray:

ax3l:
@tmsclark2 can you try this work-around on LUMI?
https://github.com/ECP-WarpX/WarpX/pull/4237

PR #4237, opened 2023-08-25
Work-Around: Segfault in MPI_Init with HIP

ax3l:
See:
  https://docs.olcf.ornl.gov/systems/crusher_quick_start_guide.html#olcfdev-1655-occasional-seg-fault-during-mpi-init

Proposed work-around for @tmsclark in #4236.

I think this might be a general defect of GPU-aware MPI implementations from HPE/Cray at this point, explicit device context init before MPI init can help establishing GPU-aware MPI init assumptions at this point.
Clarifying with AMD if we can have a check for an already existing HIP/ROCm initialized runtime, in case we want to move this safety net to AMReX at some point, too. (Also, not clear if `hipInit` is idempotent.)

- [x] make part of `ablastr::parallelization::mpi_init`?
- [x] test on AMD GPUs, i.e. Frontier, LUMI

### DIFF
--- a/Source/ablastr/parallelization/MPIInitHelpers.cpp
+++ b/Source/ablastr/parallelization/MPIInitHelpers.cpp
@@ -23,6 +23,7 @@
 #include <iostream>
 #include <string>
 #include <utility>
+#include <stdexcept>
 #include <sstream>
 
 
@@ -53,7 +54,7 @@ namespace ablastr::parallelization
         hipError_t hip_ok = hipInit(0);
         if (hip_ok != hipSuccess) {
             std::cerr << "hipInit failed with error code " << hip_ok << "! Aborting now.\n";
-            return 1;
+            throw std::runtime_error("hipInit failed. Did not proceeding with MPI_Init_thread.");
         }
 #endif
 
--- a/Source/ablastr/parallelization/MPIInitHelpers.cpp
+++ b/Source/ablastr/parallelization/MPIInitHelpers.cpp
@@ -1,7 +1,6 @@
-/* Copyright 2020 Axel Huebl
+/* This file is part of ABLASTR.
  *
- * This file is part of ABLASTR.
- *
+ * Authors: Axel Huebl
  * License: BSD-3-Clause-LBNL
  */
 #include "MPIInitHelpers.H"
@@ -15,9 +14,17 @@
 #   include <mpi.h>
 #endif
 
+// OLCFDEV-1655: Segfault during MPI_Init & in PMI_Allgather
+// https://docs.olcf.ornl.gov/systems/crusher_quick_start_guide.html#olcfdev-1655-occasional-seg-fault-during-mpi-init
+#if defined(AMREX_USE_HIP)
+#include <hip/hip_runtime.h>
+#endif
+
+#include <iostream>
 #include <string>
 #include <utility>
 #include <sstream>
+
 
 namespace ablastr::parallelization
 {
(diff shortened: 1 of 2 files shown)


Source/ablastr/parallelization/MPIInitHelpers.cpp
```diff
@@ -15,10 +14,18 @@
...
+// OLCFDEV-1655: Segfault during MPI_Init & in PMI_Allgather
+// https://docs.olcf.ornl.gov/systems/crusher_quick_start_guide.html#olcfdev-1655-occasional-seg-fault-during-mpi-init
+#if defined(AMREX_USE_HIP)
+#include <hip/hip_runtime.h>
+#endif
+
+#include <iostream>
 #include <string>
 #include <utility>
 #include <sstream>
```
ax3l, line 27:
```suggestion
#include <stdexcept>
#include <sstream>
```

Source/ablastr/parallelization/MPIInitHelpers.cpp
```diff
@@ -40,6 +47,16 @@ namespace ablastr::parallelization
     std::pair< int, int >
     mpi_init (int argc, char* argv[])
     {
+        // OLCFDEV-1655: Segfault during MPI_Init & in PMI_Allgather
+        // https://docs.olcf.ornl.gov/systems/crusher_quick_start_guide.html#olcfdev-1655-occasional-seg-fault-during-mpi-init
+#if defined(AMREX_USE_HIP) && defined(AMREX_USE_MPI)
+        hipError_t hip_ok = hipInit(0);
+        if (hip_ok != hipSuccess) {
+            std::cerr << "hipInit failed with error code " << hip_ok << "! Aborting now.\n";
+            return 1;
```
ax3l:
```suggestion
            throw std::runtime_error("hipInit failed. Did not proceeding with MPI_Init_thread.");
```

Source/ablastr/parallelization/MPIInitHelpers.cpp
```diff
@@ -40,6 +48,16 @@ namespace ablastr::parallelization
     std::pair< int, int >
     mpi_init (int argc, char* argv[])
     {
+        // OLCFDEV-1655: Segfault during MPI_Init & in PMI_Allgather
+        // https://docs.olcf.ornl.gov/systems/crusher_quick_start_guide.html#olcfdev-1655-occasional-seg-fault-during-mpi-init
+#if defined(AMREX_USE_HIP) && defined(AMREX_USE_MPI)
+        hipError_t hip_ok = hipInit(0);
+        if (hip_ok != hipSuccess) {
+            std::cerr << "hipInit failed with error code " << hip_ok << "! Aborting now.\n";
```
lucafedeli88, line 56: Is there a reason for not using `ABLASTR_ALWAYS_ASSERT_WITH_MESSAGE` here?

lucafedeli88 (commented):
Thanks really a lot for this PR! I've left a small comment. I'll test these changes as soon as possible.

Source/ablastr/parallelization/MPIInitHelpers.cpp
ax3l: Yes: anything calling into AMReX functions cannot be used (safely) before AMReX is initialized. `amrex::Assert` implements multiple things, not all of them work with an uninitialized AMReX context.

For instance, not even `amrex::Print()` works before init, I had to develop a work-around falling back to all-print in pyAMReX: https://github.com/AMReX-Codes/pyamrex/pull/174

Raising a standard exception is a clean thing to do here - we init MPI and technically there is no AMReX involved at this point in time yet.

ax3l:
I tested this on Frontier and it does not cause issues.

ax3l:
@tmsclark @lucafedeli88 let me know if it fixes your LUMI issue :)

PR #4238, opened 2023-08-26
Cleaning: Explicit Capture in Copy Lambda

ax3l:
In the spirit of a recent nvcc compiler bug, simplify the capture logic from implicit to explicit for the particle filter lambda.

X-ref: in the spirit of https://github.com/AMReX-Codes/amrex/pull/3510


RevathiJambunathan (approved):
Thanks for this PR!

PR #4239, opened 2023-08-26
Remove some unnecessary static variables from WarpX.H

lucafedeli88:
The static variables 
```
std::string WarpX::str_Ex_ext_grid_function;
std::string WarpX::str_Ey_ext_grid_function;
std::string WarpX::str_Ez_ext_grid_function;
std::string WarpX::str_Bx_ext_grid_function;
std::string WarpX::str_By_ext_grid_function;
std::string WarpX::str_Bz_ext_grid_function;
```
are actually used only once to parse the field functions from the inputfile. For this reason, I don't think that they need to be static variables of the WarpX class.


ax3l (approved):
Thank you, very good. All globals must goooo :)

PR #4240, opened 2023-08-27
Replace static function WarpX::CheckSignals() with direct call to SignalHandling::CheckSignals()

lucafedeli88:
This PR is part of an ongoing effort to simplify the WarpX class.


ax3l (approved):
Great work, thank you! :)

PR #4242, opened 2023-08-27
Remove AddToMultiFabMap functions from WarpX class

lucafedeli88:
As part of the ongoing effort to clean the WarpX class, this PR removes the static `AddToMultiFabMap` functions from it.
These functions actually contained just the instruction `multifab_map[name] = mf.get();` (or `imultifab_map[name] = mf.get();`). These instructions are now used directly in the code.


ax3l (approved):
LGTM, thanks!!

Adding @dpgrote for review as well to make sure I do not miss anything.

dpgrote (approved):
This looks good! This for cleaning up (my mess :) )

PR #4244, opened 2023-08-28
Add Python bindings to extract the timestep dt

RemiLehe:
In many applications, it is useful to extract the timestep of the simulation in Python callbacks.

### DIFF
--- a/Source/Python/WarpX.cpp
+++ b/Source/Python/WarpX.cpp
@@ -143,8 +143,7 @@ void init_WarpX (py::module& m)
             [](WarpX const & wx, int lev){ return wx.gett_new(lev); },
             py::arg("lev")
         )
-        .def("getdt",
-            [](WarpX const & wx, int lev){ return wx.getdt(lev); },
+        .def("getdt", &WarpX::getdt,
             py::arg("lev")
         )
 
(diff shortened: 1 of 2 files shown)


Source/Python/WarpX.cpp
```diff
@@ -143,6 +143,10 @@ void init_WarpX (py::module& m)
             [](WarpX const & wx, int lev){ return wx.gett_new(lev); },
             py::arg("lev")
         )
+        .def("getdt",
+            [](WarpX const & wx, int lev){ return wx.getdt(lev); },
```
dpgrote, line 147: This is fine for now, but there is a potential for out of bounds access if a bad lev is passed in. `getdt` is doing a simple array access, i.e. `dt[lev]`. You can add bounds checking in this interface, or `getdt` could be changed to use `dt.at(lev)` and then pybind11 would catch the `std::out_of_range` exception.

Note that this comment applies to other routines there as well.

RemiLehe:
Thanks for the suggestion. I added the `.at` operator for this particular function.

dpgrote (approved):
Thanks for the update! This looks good to me.

Source/Python/WarpX.cpp
ax3l, line 147: Does this work as well? :)
```suggestion
        .def("getdt", &WarpX::getdt,
```

PR #4253, opened 2023-08-31
Clang tidy CI test: add misc-definitions-in-headers check

lucafedeli88:
This PR adds the following check to `clang-tidy` CI test:

- [misc-definitions-in-headers](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/misc-definitions-in-headers.html)

The changes introduced with this PR consist essentially in moving some functions from headers to cpp files or adding the `AMREX_FORCE_INLINE` attribute


Source/FieldSolver/SpectralSolver/SpectralHankelTransform/BesselRoots.cpp
```diff
@@ -0,0 +1,153 @@
...
+!                                          (www.jpmoreau.fr)
+! ------------------------------------------------------------------------ */
+
+#include "BesselRoots.H"
+
+#include "Utils/WarpXConst.H"
+
+#include <cmath>
+
+namespace{
```
EZoni, line 41: Thanks @lucafedeli88, the PR looks good to me. I have only one question: why is this unnamed `namespace` needed here, or in other words, why can't we treat `SecantRootFinder` in the same way as `GetBesselRoots`?

lucafedeli88: Hi @EZoni ! I moved `SecantRootFinder` into the cpp file to simplify the header file, since this function appears to be an helper function used only in the `GetBesselRoots` function. The use of an unnamed namespace avoids potential collisions with functions having the same name and defined elsewhere (https://google.github.io/styleguide/cppguide.html#Internal_Linkage) .

EZoni (approved):

PR #4257, opened 2023-09-01
Add particle resampling inputs to `picmi.py`

roelof-groenewald:
As described in the title this PR just adds some missing inputs to `picmi.py`.

For context, I need this in testing of #4125.


ax3l (approved):
Thank you! :+1:

PR #4258, opened 2023-09-01
Use input parameter parser for lattice elements

dpgrote:
As the title says, for the accelerator lattice elements, this changes the reading of the input parameters to use the parser instead of direct query or get.


lucafedeli88 (approved):
Good idea! Thanks for this PR, @dpgrote !

PR #4259, opened 2023-09-01
[WIP] PSATD: less radial ghost cells by default in RZ

EZoni:
In RZ geometry, there is no domain decomposition along r, hence no need to account for the extent of the spectral stencil along that direction. Note that some variables were also renamed from `ngFFt_...` to `ngFFT_...`, the lower-case `t` looked like a typo.

@RemiLehe @dpgrote 
Is this what you had in mind?

### DIFF
--- a/Regression/Checksum/benchmarks_json/Langmuir_multi_rz_psatd.json
+++ b/Regression/Checksum/benchmarks_json/Langmuir_multi_rz_psatd.json
@@ -1,27 +1,27 @@
 {
   "electrons": {
-    "particle_momentum_x": 3.676049145834066e-36,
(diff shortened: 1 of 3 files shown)


dpgrote (dismissed):
This looks ok for me, but also let Remi review.

Though it is failing two tests. The differences look like round-off errors, but it's not clear why the round-off would be different.

EZoni:
I agree with the fact that it's not fully understood why this would change the checksums of some PSATD tests.

RemiLehe:
In offline discussion: is this OK with we use `damped` in the `r` direction?

PR #4262, opened 2023-09-01
Particle fields diagnostic support for PICMI

aveksler1:
This PR adds support for particle field functions in the `FieldDiagnostic` class in PICMI.

The user specifies the field names and their corresponding field functions and parameters in a dictionary of dictionaries, as described by the doc string. Happy to make improvements or provide an example of the format in the doc string if necessary!

### DIFF
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -1912,11 +1912,26 @@ from dataclasses import dataclass
 
 @dataclass(frozen=True)
 class ParticleFieldDiagnostic(object):
-    def __init__(self, name, func, do_average=1, filter=None):
-        self.name = name
-        self.func = func
-        self.do_average = do_average
-        self.filter = filter
+    """
+    Parameters
+    ----------
+    name: str
+        Name of particle field diagnostic. If a component of a vector field, the coordinate
+        (i.e x, y, z) should be the last character.
+    func: parser str
+        Parser function to be calculated for each particle per cell. Should be of the form
+        f(x,y,z,ux,uy,uz)
+    do_average: (0 or 1) optional, default 1
+        Whether the diagnostic is averaged by the sum of particle weights in each cell
+    filter: parser str, optional
+        Parser function returning a boolean for whether to include a particle in the diagnostic.
+        If not specified, all particles will be included. The function arguments are the same
+        as the `func` above.
+    """
+    name: str
+    func: str
+    do_average: int = 1
+    filter: str = None
 
 
 class FieldDiagnostic(picmistandard.PICMI_FieldDiagnostic, WarpXDiagnosticBase):
@@ -1953,12 +1968,6 @@ class FieldDiagnostic(picmistandard.PICMI_FieldDiagnostic, WarpXDiagnosticBase):
         List of ParticleFieldDiagnostic classes to install in the simulation. Error
         checking is handled in the class itself.
 
-    # warpx_particle_fields_to_plot: dict, optional
-    #     Keys: particle field diagnostic name (string)
-    #     Values: Dictionaries with key/value pairs of corresponding parser function
-    #     and parameters. The keys must be "func" (parser string, mandatory),
-    #     "do_average" (0 or 1, optional, default 1), and "filter" (parser string, optional).
-
     warpx_particle_fields_species: list of strings, optional
         Species for which to calculate particle_fields_to_plot functions. Fields will
         be calculated separately for each specified species. If not passed, default is
@@ -1977,7 +1986,6 @@ class FieldDiagnostic(picmistandard.PICMI_FieldDiagnostic, WarpXDiagnosticBase):
         self.file_min_digits = kw.pop('warpx_file_min_digits', None)
         self.dump_rz_modes = kw.pop('warpx_dump_rz_modes', None)
         self.particle_fields_to_plot = kw.pop('warpx_particle_fields_to_plot', None)
-        # self.particle_fields_to_plot = kw.pop('warpx_particle_fields_to_plot', {})
         self.particle_fields_species = kw.pop('warpx_particle_fields_species', None)
 
     def initialize_inputs(self):
@@ -2057,7 +2065,6 @@ class FieldDiagnostic(picmistandard.PICMI_FieldDiagnostic, WarpXDiagnosticBase):
             self.diagnostic.fields_to_plot = fields_to_plot
 
         particle_fields_to_plot = list()
-        ####### list of ParticleFieldDiagnostics implementation
         if self.particle_fields_to_plot is not None:
             for pfd in self.particle_fields_to_plot:
                 particle_fields_to_plot.append(pfd.name)
@@ -2070,26 +2077,6 @@ class FieldDiagnostic(picmistandard.PICMI_FieldDiagnostic, WarpXDiagnosticBase):
                 self.diagnostic.__setattr__(
                     f'particle_fields.{dataname}.filter(x,y,z,ux,uy,uz)', pfd.filter
                 )
-
-        # ####### dictionary implementation
-        # for dataname, params in self.particle_fields_to_plot.items():
-        #     particle_fields_to_plot.append(dataname)
-        #     try:
-        #         self.diagnostic.__setattr__(
-        #             f'particle_fields.{dataname}(x,y,z,ux,uy,uz)', params.pop('func')
-        #         )
-        #     except KeyError:
-        #         raise Exception("A function must be provided for a particle field!")
-        #     self.diagnostic.__setattr__(
-        #         f'particle_fields.{dataname}.do_average', params.pop('do_average', None)
-        #     )
-        #     self.diagnostic.__setattr__(
-        #         f'particle_fields.{dataname}.filter(x,y,z,ux,uy,uz)', params.pop('filter', None)
-        #     )
-
-        #     # catches unexpected keywords
-        #     if params:
-        #         raise TypeError(f"Unexpected keyword argument for particle field {dataname} - {list(params)}")
 
         # --- Convert to a sorted list so that the order
         # --- is the same on all processors.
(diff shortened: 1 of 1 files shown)


Python/pywarpx/picmi.py
```diff
@@ -1938,6 +1938,17 @@ class FieldDiagnostic(picmistandard.PICMI_FieldDiagnostic, WarpXDiagnosticBase):
...
         Flag whether to dump the data for all RZ modes
+
+    warpx_particle_fields_to_plot: dict, optional
+        Keys: particle field diagnostic name (string)
+        Values: Dictionaries with key/value pairs of corresponding parser function
+        and parameters. The keys must be "func" (parser string, mandatory),
+        "do_average" (0 or 1, optional, default 1), and "filter" (parser string, optional).
+
+    warpx_particle_fields_species: list of strings, optional
+        Species for which to calcualte particle_fields_to_plot functions. Fields will
```
dpgrote: Typo `calculate`

Python/pywarpx/picmi.py
```diff
@@ -2028,6 +2041,38 @@ def initialize_inputs(self):
             fields_to_plot.sort()
             self.diagnostic.fields_to_plot = fields_to_plot
 
+        particle_fields_to_plot = set()
+        if self.particle_fields_to_plot is not None:
+            for dataname, params in self.particle_fields_to_plot.items():
+                particle_fields_to_plot.add(dataname)
+                if "func" not in params:
```
dpgrote: I might have done something like this
```
try:
    self.diagnostic.__setattr__(
                        f'particle_fields.{dataname}(x,y,z,ux,uy,uz)', params.pop('func')
                    )
except KeyError:
    raise Exception('A function must be provided for a particle field!')
```

Python/pywarpx/picmi.py
```diff
@@ -2028,6 +2041,38 @@ def initialize_inputs(self):
...
+                    raise ValueError("A function must be provided for a particle field!")
+                else:
+                    self.diagnostic.__setattr__(
+                        f'particle_fields.{dataname}(x,y,z,ux,uy,uz)', params['func']
+                    )
+                    params.pop('func', None)
+                if "do_average" in params:
```
dpgrote: Here you can do
```
self.diagnostic.__setattr__(
                        f'particle_fields.{dataname}.do_average', params.pop('do_average', None)
                    )
```
Parameters set to None are ignored.

Python/pywarpx/picmi.py
```diff
@@ -1951,6 +1962,8 @@ def init(self, kw):
         self.file_prefix = kw.pop('warpx_file_prefix', None)
         self.file_min_digits = kw.pop('warpx_file_min_digits', None)
         self.dump_rz_modes = kw.pop('warpx_dump_rz_modes', None)
+        self.particle_fields_to_plot = kw.pop('warpx_particle_fields_to_plot', None)
```
dpgrote: This can default to an empty dictionary, `{}`, avoiding the check of `is not None` below.

Python/pywarpx/picmi.py
```diff
@@ -2028,6 +2041,38 @@ def initialize_inputs(self):
             fields_to_plot.sort()
             self.diagnostic.fields_to_plot = fields_to_plot
 
+        particle_fields_to_plot = set()
+        if self.particle_fields_to_plot is not None:
+            for dataname, params in self.particle_fields_to_plot.items():
+                particle_fields_to_plot.add(dataname)
```
dpgrote: Perhaps check to make sure that a data name is not repeated, since that is presumably an error.
```
if dataname in particle_fields_to_plot:
    raise Exception('A particle fields data name can not be repeated')
```
Note that particle_fields_to_plot can be a list (instead of a set).

ax3l:
Hi @aveksler1 , just checking in if you saw @dpgrote 's review comments :)

aveksler1:
@ax3l Yes I did, thanks so much @dpgrote for the suggestions!!! I hope to make the corresponding changes soon, sorry about that 😄

RemiLehe:
Thanks for this PR!

However, I am not sure that this is the approach that we want to take... 
This seems somewhat inconsistent with the PICMI standard, which has a dedicated `ParticleDiagnostic`. I was starting to use `ParticleDiagnostic` in this PR: https://github.com/ECP-WarpX/WarpX/pull/3745

roelof-groenewald:
@RemiLehe I think the `particle_fields` should be associated with a `FieldDiagnostic` since these quantities are output on the grid - they are just custom fields. I think of it as the `rho` and `J` fields just have hard coded `particle_field` functions.

Python/pywarpx/picmi.py
aveksler1: This is avoided by the fact that the `dataname` is a key in a dictionary, so duplicate values are not an issue

Python/pywarpx/picmi.py
```diff
@@ -1938,6 +1938,17 @@ class FieldDiagnostic(picmistandard.PICMI_FieldDiagnostic, WarpXDiagnosticBase):
 
     warpx_dump_rz_modes: bool, optional
         Flag whether to dump the data for all RZ modes
+
+    warpx_particle_fields_to_plot: dict, optional
```
dpgrote: I don't know if this would be better, but perhaps instead of a dictionary, accept a list of instances of a class with the predefined attributes. Something like this
```
from dataclasses import dataclass

@dataclass(frozen=True)
class ParticleFieldDiagnostic:
    name: str
    func: str
    do_average: int = 1
    filter: str = None
```
This will let Python do some of the error checking, making sure that the input is correct. The "frozen=True" prevents extra attributes from being added.

aveksler1: Hi David, 

Sorry for the late response again (I'm going to dedicate today and tomorrow to close this out). I think thats a great idea!

The majority of the classes in `picmi.py` are derived from the PICMI standard or from the `WarpXDiagnosticBase` class, and I don't think that this fits into either of these categories. Would it be fine for this to be a standalone class with just an `__init__ `function?

dpgrote: Using a stand alone class for this would be good. Note that a doc string should be added (I left it out of my example for simplification).

A future PR might add this to the picmi standard, but get it working in WarpX first.

aveksler1: Thank you for the suggestion, its working with the class implementation.

Python/pywarpx/picmi.py
```diff
@@ -1908,6 +1908,32 @@ def set_write_dir(self):
             file_prefix = (self.file_prefix or self.name)
             self.diagnostic.file_prefix = os.path.join(write_dir, file_prefix)
 
+from dataclasses import dataclass
```
dpgrote: This import line should go at the top of the file after the line "import re".

Python/pywarpx/picmi.py
```diff
@@ -1908,6 +1908,32 @@ def set_write_dir(self):
...
+
+
+@dataclass(frozen=True)
+class ParticleFieldDiagnostic(object):
```
dpgrote: Thanks for adding this - it looks good! Though I have comments.
Here, the `(object)` is not necessary.
```class ParticleFieldDiagnostic:```

File: Python/pywarpx/picmi.py [hunk 9]
```diff
@@ -1908,6 +1908,32 @@ def set_write_dir(self):
...
+    """
+    Parameters
+    ----------
+    name: str
+        Name of particle field diagnostic. If a component of a vector field, the coordinate
+        (i.e x, y, z) should be the last character.
+    func: parser str
```
dpgrote, line 1924: For the doc to be parsed properly, there needs to be an empty line between each parameter definition.

Python/pywarpx/picmi.py
```diff
@@ -1908,6 +1908,32 @@ def set_write_dir(self):
...
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class ParticleFieldDiagnostic(object):
+    """
+    Parameters
+    ----------
+    name: str
+        Name of particle field diagnostic. If a component of a vector field, the coordinate
```
dpgrote: For this comment, perhaps be more specific.
`If a component of a vector field, for the openPMD viewer to to treat it as a vector, the coordinate (i.e. x, y, z) should be the last character.`

Python/pywarpx/picmi.py
dpgrote: With this version, this can be an empty list, `[]`, for the same reason.

Python/pywarpx/picmi.py
```diff
@@ -2028,6 +2065,28 @@ def initialize_inputs(self):
             fields_to_plot.sort()
             self.diagnostic.fields_to_plot = fields_to_plot
 
+        particle_fields_to_plot = list()
```
dpgrote: To avoid duplicate variable names and make its use more clear, this variable should be something like `particle_fields_to_plot_names`.

Python/pywarpx/picmi.py
```diff
@@ -2028,6 +2065,28 @@ def initialize_inputs(self):
...
+                    raise Exception('A particle fields data name can not be repeated.')
+                particle_fields_to_plot.append(pfd.name)
+                self.diagnostic.__setattr__(
+                    f'particle_fields.{pfd.name}(x,y,z,ux,uy,uz)', pfd.func
+                )
+                self.diagnostic.__setattr__(
+                    f'particle_fields.{pfd.name}.do_average', pfd.do_average
+                )
+                self.diagnostic.__setattr__(
+                    f'particle_fields.{dataname}.filter(x,y,z,ux,uy,uz)', pfd.filter
```
dpgrote: This should presumably `pfd.name` instead of `dataname`.

aveksler1: Thanks for catching this!

dpgrote (approved):
Looks good to me! Thanks!

roelof-groenewald (approved):

PR #4264, opened 2023-09-02
Print resampling message only if `verbose` is on

roelof-groenewald:
This PR includes 2 changes to the message print to stdout during particle resampling:
1. Only output the message about resampling if `WarpX::verbose = true`
2. Include the time step at which resampling happens

In tests of #4125 I use resampling and in the current setup the frequent output of resampling messages obscures other run diagnostics I'm periodically printing.

### DIFF
--- a/Source/Particles/MultiParticleContainer.H
+++ b/Source/Particles/MultiParticleContainer.H
@@ -179,7 +179,7 @@ public:
     *
     * @param[in] timestep the current timestep.
     */
-    void doResampling (int timestep);
+    void doResampling (int timestep, const bool verbose);
 
 #ifdef WARPX_QED
     /** If Schwinger process is activated, this function is called at every
--- a/Source/Particles/MultiParticleContainer.H
+++ b/Source/Particles/MultiParticleContainer.H
@@ -179,7 +179,7 @@ public:
     *
     * @param[in] timestep the current timestep.
     */
-    void doResampling (int timestep, const bool verbose);
+    void doResampling (int timestep, bool verbose);
 
 #ifdef WARPX_QED
     /** If Schwinger process is activated, this function is called at every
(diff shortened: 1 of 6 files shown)


Source/Particles/MultiParticleContainer.H
```diff
@@ -179,7 +179,7 @@ public:
     *
     * @param[in] timestep the current timestep.
     */
-    void doResampling (int timestep);
+    void doResampling (int timestep, const bool verbose);
```
roelof-groenewald:
```suggestion
    void doResampling (int timestep, bool verbose);
```

Source/Particles/PhysicalParticleContainer.H
```diff
@@ -307,7 +307,7 @@ public:
     *
     * @param[in] timestep the current timestep.
     */
-    void resample (int timestep) override final;
+    void resample (int timestep, const bool verbose=true) override final;
```
roelof-groenewald:
```suggestion
    void resample (int timestep, bool verbose=true) override final;
```

Source/Particles/WarpXParticleContainer.H
```diff
@@ -381,7 +381,7 @@ public:
      * override the method for every derived class. Note that in practice this function is never
      * called because resample() is only called for PhysicalParticleContainers.
      */
-    virtual void resample (const int /*timestep*/) {}
+    virtual void resample (const int /*timestep*/, const bool /*verbose*/) {}
```
roelof-groenewald:
```suggestion
    virtual void resample (const int /*timestep*/, bool /*verbose*/) {}
```

lucafedeli88 (approved):
Thanks for this PR, @roelof-groenewald !

PR #4266, opened 2023-09-06
Add lower and upper bound params to regular diags in PICMI

n01r:
The parameters `warpx_lower_bound` and `warpx_upper_bound` can now be used in the definition of a `FieldDiagnostic` in PICMI.

### DIFF
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -1951,6 +1951,8 @@ class FieldDiagnostic(picmistandard.PICMI_FieldDiagnostic, WarpXDiagnosticBase):
         self.file_prefix = kw.pop('warpx_file_prefix', None)
         self.file_min_digits = kw.pop('warpx_file_min_digits', None)
         self.dump_rz_modes = kw.pop('warpx_dump_rz_modes', None)
+        self.lower_bound = kw.pop('warpx_lower_bound', None)
+        self.upper_bound = kw.pop('warpx_upper_bound', None)
 
     def initialize_inputs(self):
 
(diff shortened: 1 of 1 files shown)


RTSandberg (approved):
This looks good and works for me!

Python/pywarpx/picmi.py
```diff
@@ -1951,6 +1951,8 @@ def init(self, kw):
         self.file_prefix = kw.pop('warpx_file_prefix', None)
         self.file_min_digits = kw.pop('warpx_file_min_digits', None)
         self.dump_rz_modes = kw.pop('warpx_dump_rz_modes', None)
+        self.lower_bound = kw.pop('warpx_lower_bound', None)
```
ax3l, line 1962: @n01r thank you! please add to the doc string for documentation (line 1940 :) )

RemiLehe (approved):

dpgrote:
I missed this PR. It's not clear why the new input parameters `warpx_lower_bound` and upper are needed since the PICMI standard already includes the `lower_bound` and `upper_bound` input parameters and they were already included in the WarpX implementation.

n01r:
@dpgrote, there still was no option to limit the output to, e.g. just x-z slices, on a per-diagnostic level. Or did I overlook the option to do that?

dpgrote:
I'm not sure what the difference is. The changes in this PR only add the new arguments and sets `self.lower_bound` and upper. That setting of `self.lower_bound` overwrites the input argument from the PICMI standard, but ends up at the same place, being assigned to `self.diagnostic.diag_lo` (since that line was already there).

n01r:
Ah, got it @dpgrote. Right, the lower and upper bounds are already here in this class https://github.com/picmi-standard/picmi/blob/a73a0f14f18cec3233b47fa8983de85596318ac3/PICMI_Python/diagnostics.py#L42 from which we inherit. Indeed, we do not need this. Sorry, I missed that. Will create a PR to revert.

PR #4267, opened 2023-09-06
Clang tidy CI test: add selected google-* checks

lucafedeli88:
This PR adds the following checks to `clang-tidy` CI test:

- [google-build-explicit-make-pair](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/google-build-explicit-make-pair.html)
- [google-build-namespaces](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/google-build-namespaces.html)
- [google-global-names-in-headers](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/google-global-names-in-headers.html)

These are the `google-*` checks currently enabled in AMReX (see [AMReX .clang-tidy file](https://github.com/AMReX-Codes/amrex/blob/development/.clang-tidy) `google-build-namespaces` is actually an alias of a check that we already enable (`cert-dcl59-cpp`)


ax3l:
Ok, just FYI
https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/google-build-explicit-make-pair.html

We do not support GCC 4.6 anymore :)

ax3l:
I do not understand the issue with
https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/google-build-namespaces.html

Can you elaborate? :)

ax3l:
This is great :+1: 
https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/google-global-names-in-headers.html

There is a small exception to this for our Python include helper though, that shall always define some aliases.

lucafedeli88:
> Ok, just FYI https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/google-build-explicit-make-pair.html
> 
> We do not support GCC 4.6 anymore :)

I see... do you want to remove this check?

lucafedeli88:
> I do not understand the issue with https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/google-build-namespaces.html
> 
> Can you elaborate? :)

Sure! It's not really an issue. I've enabled this check because it is included in the AMReX `.clang-tidy` file. However, the `cert-dcl59-cpp` check that  we've already enabled does exactly the same thing. This is not problematic, it's just that we are enabling the same check twice. It also means that we can safely remove this check from the list if we want.

lucafedeli88:
> This is great 👍 https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/google-global-names-in-headers.html
> 
> There is a small exception to this for our Python include helper though, that shall always define some aliases.

Which file do you mean? At the moment `clang-tidy` is not complaining

ax3l:
I meant the pyWarpX.H and pyImpactX.H files, but seems it is happy for now.

> > Just FYI [...]  We do not support GCC 4.6 anymore :)
> I see... do you want to remove this check?

Ok to leave, since it's  clean to use that way.

ax3l:
I meant the pyWarpX.H and pyImpactX.H files, but seems it is happy for now.

ax3l (approved):

PR #4270, opened 2023-09-06
Fix macos CI

AlexanderSinn:
Fix from https://github.com/Hi-PACE/hipace/pull/937


lucafedeli88 (approved):
Thanks for fixing this, @AlexanderSinn !

PR #4274, opened 2023-09-07
Docs: mpi4py in MPI-Conda Deps

ax3l:
Fix missing mpi4py dependency that we can get from conda.

### DIFF
--- a/Docs/source/install/dependencies.rst
+++ b/Docs/source/install/dependencies.rst
@@ -214,7 +214,7 @@ The `Advanced Package Tool (APT) <https://en.wikipedia.org/wiki/APT_(software)>`
       .. code-block:: bash
 
          sudo apt update
-         sudo apt install build-essential ccache cmake g++ git libfftw3-mpi-dev libfftw3-dev libhdf5-openmpi-dev libopenmpi-dev pkg-config python3 python3-matplotlib python3-numpy python3-pandas python3-pip python3-scipy python3-venv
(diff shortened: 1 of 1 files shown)


Docs/source/install/dependencies.rst
```diff
@@ -75,7 +75,7 @@ Conda (Linux/macOS/Windows)
 
       .. code-block:: bash
 
-         conda create -n warpx-cpu-mpich-dev -c conda-forge blaspp boost ccache cmake compilers git lapackpp "openpmd-api=*=mpi_mpich*" python numpy pandas scipy yt "fftw=*=mpi_mpich*" pkg-config matplotlib mamba ninja mpich pip virtualenv
+         conda create -n warpx-cpu-mpich-dev -c conda-forge blaspp boost ccache cmake compilers git lapackpp "openpmd-api=*=mpi_mpich*" python numpy pandas scipy yt "fftw=*=mpi_mpich*" pkg-config matplotlib mamba mpich mpi4py ninja pip virtualenv
```
RemiLehe, line 78: Thanks! 
Should we also add `mpi4py` in the `spack`, `brew` and `apt-get` instructions?

ax3l: I think I installed those there via pip, lemme check

ax3l: Spack: `py-mpi4py` is in the specs already :)

ax3l: Adding to Ubuntu/Debian now.

Brew has no Python at all yet... maybe another time. Tricky to get its shipped Python right sometimes.

RemiLehe (approved):

PR #4275, opened 2023-09-08
[WIP] Reorder WarpX.H and move function definitions to WarpX.cpp

lucafedeli88:
This PR reorders `WarpX.H` in such a way that `public`, `protected` and `private` appear exactly once.
Moreover, it moves all the function definitions into `WarpX.cpp`

### DIFF
--- a/Source/WarpX.H
+++ b/Source/WarpX.H
@@ -440,10 +440,10 @@ public:
     static std::map<std::string, amrex::iMultiFab *> imultifab_map;
 
     std::array<const amrex::MultiFab* const, 3>
-    get_array_Bfield_aux  (const int lev) const;
+    get_array_Bfield_aux  (int lev) const;
 
     std::array<const amrex::MultiFab* const, 3>
-    get_array_Efield_aux  (const int lev) const;
+    get_array_Efield_aux  (int lev) const;
 
     amrex::MultiFab * get_pointer_Efield_aux  (int lev, int direction) const;
     amrex::MultiFab * get_pointer_Bfield_aux  (int lev, int direction) const;
--- a/Source/WarpX.H
+++ b/Source/WarpX.H
@@ -110,18 +110,18 @@ public:
     static std::string Version (); //!< Version of WarpX executable
     static std::string PicsarVersion (); //!< Version of PICSAR dependency
 
-    int Verbose () const { return verbose; }
+    int Verbose () const;
 
     void InitData ();
 
     void Evolve (int numsteps = -1);
 
-    MultiParticleContainer& GetPartContainer () { return *mypc; }
-    MacroscopicProperties& GetMacroscopicProperties () { return *m_macroscopic_properties; }
-    HybridPICModel& GetHybridPICModel () { return *m_hybrid_pic_model; }
-    MultiDiagnostics& GetMultiDiags () {return *multi_diags;}
+    MultiParticleContainer& GetPartContainer ();
+    MacroscopicProperties& GetMacroscopicProperties ();
+    HybridPICModel& GetHybridPICModel ();
+    MultiDiagnostics& GetMultiDiags ();
 
-    ParticleBoundaryBuffer& GetParticleBoundaryBuffer () { return *m_particle_boundary_buffer; }
+    ParticleBoundaryBuffer& GetParticleBoundaryBuffer ();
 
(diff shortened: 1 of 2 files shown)


Source/WarpX.H
```diff
@@ -110,18 +110,18 @@ public:
...
-    MultiParticleContainer& GetPartContainer () { return *mypc; }
-    MacroscopicProperties& GetMacroscopicProperties () { return *m_macroscopic_properties; }
-    HybridPICModel& GetHybridPICModel () { return *m_hybrid_pic_model; }
-    MultiDiagnostics& GetMultiDiags () {return *multi_diags;}
+    MultiParticleContainer& GetPartContainer ();
+    MacroscopicProperties& GetMacroscopicProperties ();
+    HybridPICModel& GetHybridPICModel ();
+    MultiDiagnostics& GetMultiDiags ();
 
-    ParticleBoundaryBuffer& GetParticleBoundaryBuffer () { return *m_particle_boundary_buffer; }
```
ax3l, line 124: For these one-line definitions, it is indeed ok to have them in the header and also supports the compiler to inline them.

We can move them to cpp for stylistic reasons if we want, probably slightly increases link time.

ax3l (approved):
I am ok with this, mostly because it looks a bit cleaner for the long lines, even given the potential downsides.

ax3l:
@WeiqunZhang can you take a 2nd look? :)

WeiqunZhang:
Fine with me.

WeiqunZhang (approved):

Source/WarpX.H
dpgrote: If I could give my opinion, I generally prefer one line routines like these to be in the header file. This reduces a lot of duplication of code that has little benefit. It also makes reading the code and tracing through the call chain easier since I quickly see the variable that is being returned. With the little routines in the .cpp file, there is now one more step in the call chain (and one more file) to look through just to find a variable name. This is something that I do often enough, that having this one more level of indirection will be annoying.

ax3l:
@lucafedeli88 please see @dpgrote 's comment above

lucafedeli88:
I think that a possible solution to simplify all these getters could be something like the code below (the actual details must be discussed carefully). The idea is that all the fields that are currently individual member variables of the WarpX class could be grouped together in a map and accessed with the same function. What do you think?

```
#include <map>
#include <memory>
#include <utility>

enum FieldType
{
    E,
    B,
    J,
    RHO,
};

enum Component
{
    S, //scalar
    X,
    Y,
    Z
};

class FieldData
{};

auto all_fields = std::map<std::pair<FieldType,Component>, std::unique_ptr<FieldData>>{};

void init_all_fields()
{
    all_fields[std::make_pair(FieldType::E, Component::X)] = std::make_unique<FieldData>();
    all_fields[std::make_pair(FieldType::E, Component::Y)] = std::make_unique<FieldData>();
    all_fields[std::make_pair(FieldType::E, Component::Z)] = std::make_unique<FieldData>();
    all_fields[std::make_pair(FieldType::B, Component::X)] = std::make_unique<FieldData>();
    all_fields[std::make_pair(FieldType::B, Component::Y)] = std::make_unique<FieldData>();
    all_fields[std::make_pair(FieldType::B, Component::Z)] = std::make_unique<FieldData>();
    all_fields[std::make_pair(FieldType::J, Component::X)] = std::make_unique<FieldData>();
    all_fields[std::make_pair(FieldType::J, Component::Y)] = std::make_unique<FieldData>();
    all_fields[std::make_pair(FieldType::J, Component::Z)] = std::make_unique<FieldData>();
    all_fields[std::make_pair(FieldType::RHO, Component::S)] = std::make_unique<FieldData>();
}

FieldData& get_field(FieldType type, Component comp = Component::S)
{
    return *all_fields[std::make_pair(type, comp)];
}    

```

dpgrote:
Note that the MultiFabs are already saved in a map, `multifab_map`. The keys are the tags, the full field names, for example `"Bfield_fp[x][level=0]"`. You could write a function that takes the field name, component, and qualifier as strings and the level, and puts together the key and returns the appropriate MultiFab. The qualifiers are things like `fp`, `cp`, `aux`, and `pml`.

lucafedeli88:
I would suggest closing this PR. The reordering part has been already accepted (https://github.com/ECP-WarpX/WarpX/pull/4340) and the other part requires more work.

PR #4281, opened 2023-09-08
Quartz (LLNL): New Modules, Clang

ax3l:
Module update on Quartz requires changes.
Switching to Clang 14 as the compiler.


ax3l:
tested with @bzdjordje 
cc @joshualudwig8 fyi

PR #4288, opened 2023-09-11
PICMI: Add warpx_intervals option to BTD

dpgrote:
This allows the user to set the `intervals` option for the back-transformed diagnostics from the PICMI interface.


RTSandberg:
Thanks @dpgrote !  Could you add ``warpx_intervals`` to ``LabFrameParticleDiagnostic``, too?

RTSandberg (approved):
Thanks, @dpgrote !  This looks good and will be very convenient to the few of us using BTD intervals and wanting to move to PyWarpX

EZoni (approved):

PR #4291, opened 2023-09-12
RZ m=0 mode support for particle field diagnostics 

aveksler1:
Hi all,

This PR adds support for particle field diagnostics for m=0 RZ simulations. 

I intended for the names of the diagnostics to match with what openPMD-viewer expects so that different direction components (r, theta, and z) custom field diagnostics can be specified and OpenPMD-viewer would automatically treat it them as a components of a single vector field.

### DIFF
--- a/Source/Diagnostics/Diagnostics.cpp
+++ b/Source/Diagnostics/Diagnostics.cpp
@@ -122,7 +122,7 @@ Diagnostics::BaseReadParameters ()
             "Diagnostics",
             "Particle field diagnostics will output the 0th mode only.",
             ablastr::warn_manager::WarnPriority::low
-        )
+        );
     }
 #endif
 
--- a/Source/Diagnostics/Diagnostics.cpp
+++ b/Source/Diagnostics/Diagnostics.cpp
@@ -117,8 +117,8 @@ Diagnostics::BaseReadParameters ()
     }
 #ifdef WARPX_DIM_RZ
     WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
-        m_pfield_varnames.empty(),
-        "Input error: cannot use particle_fields_to_plot with RZ"
+        (WarpX::ncomps == 1),
+        "particle_fields_to_plot only work for m=0 RZ modes."
     );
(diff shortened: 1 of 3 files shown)


Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -322,13 +328,26 @@ FullDiagnostics::InitializeFieldFunctorsRZopenPMD (int lev)
                 "Error: " + m_varnames_fields[comp] + " is not a known field output type in RZ geometry");
         }
     }
+
+    for (int pcomp=0; pcomp<int(m_pfield_varnames.size()); pcomp++) {
```
ax3l, line 334: Please add an inline comment before this block, to make this easier readable :)

ax3l (approved):
Thank you, LGTM.

I'll let @RemiLehe and @dpgrote take a 2nd look :)

ax3l:
Can this be used in an example/test to make sure we execute the new code in CI? :)

Source/Diagnostics/Diagnostics.cpp
```diff
@@ -117,8 +117,8 @@ Diagnostics::BaseReadParameters ()
     }
 #ifdef WARPX_DIM_RZ
     WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
-        m_pfield_varnames.empty(),
-        "Input error: cannot use particle_fields_to_plot with RZ"
+        (WarpX::ncomps == 1),
```
dpgrote: Would this diagnostic work if there are more than one mode, but have the diagnostic only write out the mode 0 values? If so, this check would not be needed.

aveksler1: The way this is currently implemented this would be true. 

I'm not sure how mode decomposed output data looks like actually, so it may not be obvious that it is the 0th mode only.

aveksler1:
> Can this be used in an example/test to make sure we execute the new code in CI? :)

A test will be added to #4161

RemiLehe (approved):

PR #4294, opened 2023-09-12
Revert PR4266 - lower and upper bounds

n01r:
Reverting #4266 .
The lower and upper bounds are already present within PICMI.

We can just specify, e.g.,
```Python
field_diag = picmi.FieldDiagnostic(
    name='field_diags',
    ...
    lower_bound=[xmin, 0., zmin],
    upper_bound=[xmax, 0., zmax],
    ...
)
```


dpgrote (approved):
Thanks! Looks good

lucafedeli88 (approved):
Looks good to me

Issue #4269, opened 2023-09-06
Bug in PML-SilverMuller assert


lucafedeli88:
Hi, 
thanks to `clang-tidy` (https://github.com/ECP-WarpX/WarpX/pull/4268) I've found out this issue:

```
            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
                !(
                    ( WarpX::field_boundary_lo[idim] == FieldBoundaryType::PML &&
                    WarpX::field_boundary_lo[idim] == FieldBoundaryType::Absorbing_SilverMueller ) ||
                    ( WarpX::field_boundary_hi[idim] == FieldBoundaryType::PML &&
                     WarpX::field_boundary_hi[idim] == FieldBoundaryType::Absorbing_SilverMueller )
                ),
                "PML and Silver-Mueller boundary conditions cannot be activated at the same time.");
```

The assert is always true because a boundary condition cannot be `PML` and `SilverMuller` at the same time. What was the intent of the check?

dpgrote:
Perhaps this should be
```
WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
                !(
                    ( WarpX::field_boundary_lo[idim] == FieldBoundaryType::PML &&
                    WarpX::field_boundary_hi[idim] == FieldBoundaryType::Absorbing_SilverMueller ) ||
                    ( WarpX::field_boundary_hi[idim] == FieldBoundaryType::PML &&
                     WarpX::field_boundary_lo[idim] == FieldBoundaryType::Absorbing_SilverMueller )
                ),
                "PML and Silver-Mueller boundary conditions cannot be activated at the same time.");
```
Making sure that one end isn't one type and the other end the other.

RemiLehe:
When tracking this, it seems that this was added here:
https://github.com/ECP-WarpX/WarpX/pull/2051/files#diff-1450fa5afa66d2e4dde0bd48fe21eb084d339554ad937d3604376713e053e3c4

RemiLehe:
It seems that the original intent was: if **any** boundary is PML **and** **any** other boundary is Silver-Mueller, then abort.

lucafedeli88:
This PR should fix the issue: https://github.com/ECP-WarpX/WarpX/pull/4297

EZoni:
Fixed by #4297.

PR #2051, opened 2021-07-02
Interface silver mueller

RevathiJambunathan:
Adding the interface for Silver-Mueller boundary condition.
This boundary conditions is applied only for the B-field and only in the first half of the EvolveB routine.

To ensure that we can call all boundary conditions from the same function call -- `ApplyBfieldBoundary()`, the DtType::FirstHalf and DtType::SecondHalf is added in the argument so that the silverMueller boundary condition can only be applied to the FirstHalf.

Note that in the current implementation, for Cartesian simulations, SilverMueller is applied to all boundaries.
and for RZ, silver mueller is applied to r_max, z_min and z_max boundaries.
Thus, I include an abort if the user has specified mixed boundaries with silver-mueller at one boundary and a different boundary condition at another domain boundary.
The capability to have mixed boundaries with silver-mueller will be added in a separate PR

PR #4297, opened 2023-09-13
Fix check of PML + Silver Mueller compatibility

lucafedeli88:
This PR should address issue https://github.com/ECP-WarpX/WarpX/issues/4269 by replacing an always true check with what was originally intended.

### DIFF
--- a/Source/WarpX.cpp
+++ b/Source/WarpX.cpp
@@ -874,10 +874,11 @@ WarpX::ReadParameters ()
         for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
             WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
                 !(
-                    ( WarpX::field_boundary_lo[idim] == FieldBoundaryType::PML &&
-                    WarpX::field_boundary_lo[idim] == FieldBoundaryType::Absorbing_SilverMueller ) ||
-                    ( WarpX::field_boundary_hi[idim] == FieldBoundaryType::PML &&
-                     WarpX::field_boundary_hi[idim] == FieldBoundaryType::Absorbing_SilverMueller )
+                    std::any_of(WarpX::field_boundary_lo.begin(), WarpX::field_boundary_lo.end(),
+                        [](const auto& cc){return cc == FieldBoundaryType::PML;})
+                    &&
+                    std::any_of(WarpX::field_boundary_lo.begin(), WarpX::field_boundary_lo.end(),
+                        [](const auto& cc){return cc == FieldBoundaryType::Absorbing_SilverMueller;})
                 ),
                 "PML and Silver-Mueller boundary conditions cannot be activated at the same time.");
 
--- a/Source/WarpX.cpp
+++ b/Source/WarpX.cpp
@@ -871,35 +871,30 @@ WarpX::ReadParameters ()
             quantum_xi_c2 = static_cast<amrex::Real>(quantum_xi * PhysConst::c * PhysConst::c);
         }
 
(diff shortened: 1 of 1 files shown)


Source/WarpX.cpp
```diff
@@ -874,10 +874,11 @@ WarpX::ReadParameters ()
...
                 !(
-                    ( WarpX::field_boundary_lo[idim] == FieldBoundaryType::PML &&
-                    WarpX::field_boundary_lo[idim] == FieldBoundaryType::Absorbing_SilverMueller ) ||
-                    ( WarpX::field_boundary_hi[idim] == FieldBoundaryType::PML &&
-                     WarpX::field_boundary_hi[idim] == FieldBoundaryType::Absorbing_SilverMueller )
+                    std::any_of(WarpX::field_boundary_lo.begin(), WarpX::field_boundary_lo.end(),
+                        [](const auto& cc){return cc == FieldBoundaryType::PML;})
+                    &&
+                    std::any_of(WarpX::field_boundary_lo.begin(), WarpX::field_boundary_lo.end(),
+                        [](const auto& cc){return cc == FieldBoundaryType::Absorbing_SilverMueller;})
```
EZoni: Thanks, @lucafedeli88! 

I think I still need to have a clearer picture of the original intention here.

Quoting from #2051:
> Thus, I include an abort if the user has specified mixed boundaries with silver-mueller at one boundary and a different boundary condition at another domain boundary. The capability to have mixed boundaries with silver-mueller will be added in a separate PR.

@RevathiJambunathan 
When it says "mixed boundaries", what does it mean exactly? Does it mean different boundary conditions at the lo and hi boundary, or different boundary conditions along different directions, or a combination of both?

I'm asking because I see that `WarpX::field_boundary_hi` doesn't appear in the changes here and I'm not sure if that matches the original intention of #2051.

RevathiJambunathan: By "mixed-boundaries", we meant different boundary conditions at lo and hi boundary. 
But with Silver-mueller, there was a condition, that it should be used in all boundaries 
So with silver-mueller, its also true that it could not support mixed boundary at other valid boundaries in different directions.

EZoni (dismissed):
Thanks for the PR, @lucafedeli88! Looks good to me.

Source/WarpX.cpp
```diff
@@ -874,10 +874,19 @@ WarpX::ReadParameters ()
...
+                    std::any_of(WarpX::field_boundary_hi.begin(), WarpX::field_boundary_hi.end(),
+                        [](const auto& cc){return cc == FieldBoundaryType::PML;})
+                    )
+                    &&
+                    (std::any_of(WarpX::field_boundary_lo.begin(), WarpX::field_boundary_lo.end(),
+                        [](const auto& cc){return cc == FieldBoundaryType::Absorbing_SilverMueller;})
+                    ||
+                    std::any_of(WarpX::field_boundary_hi.begin(), WarpX::field_boundary_hi.end(),
+                        [](const auto& cc){return cc == FieldBoundaryType::Absorbing_SilverMueller;})
+                    )
```
EZoni: Actually, I think this assert block now can be placed outside the loop over `idim`, right? Logically it does not depend on `idim` anymore, thanks to the usage of `std::any_of`, if I understand correctly.

lucafedeli88: Ah! Right! You are absolutely right! I've refactored the code to entirely avoid the loop.

EZoni (approved):
Thank you so much for fixing the bug and cleaning this up!

PR #4298, opened 2023-09-13
Add flag to set GPU-aware MPI from PICMI

RemiLehe:


ax3l (approved):
Excellent, thank you!

PR #4299, opened 2023-09-13
Refinement patch parser

RevathiJambunathan:
This PR adds a parser to define refinement patches. 
Thanks to @WeiqunZhang for the offline discussions

Input using `fine_tag_lo` and `fine_tag_hi` : 
[inputs_3d_mr.txt](https://github.com/ECP-WarpX/WarpX/files/12600471/inputs_3d_mr.txt)

results in : 
using_finetaglo_tag_hi

The same result can be obtained when using 
`warpx.ref_patch_function(x,y,z) = (x >-9.e-6) * (x < 9.e-6) * (y > -9.e-6) * (y < 9.e-6) * (z > -9.e-6) * (z < 9.e-6)`

using_function

Additionally, the parser can be used to define distinct patches : 
input : 
[inputs_3d_mr.txt](https://github.com/ECP-WarpX/WarpX/files/12600506/inputs_3d_mr.txt)

using_function_distinct_patches

### DIFF
--- a/Source/Utils/WarpXTagging.cpp
+++ b/Source/Utils/WarpXTagging.cpp
@@ -47,10 +47,10 @@ WarpX::ErrorEst (int lev, TagBoxArray& tags, Real /*time*/, int /*ngrow*/)
                                        (k+0.5_rt)*dx[2]+problo[2])};
 #if defined (WARPX_DIM_3D)
             amrex::Real tag_val = ref_parser(pos[0], pos[1], pos[2]);
-#elif defined (WARPX_DIM_XZ)
+#elif defined (WARPX_DIM_XZ) || defined (WARPX_DIM_RZ)
             amrex::Real unused = 0.0;
             amrex::Real tag_val = ref_parser(pos[0], unused, pos[1]);
-#elif defined (WARPX_DIM_1D)
+#elif defined (WARPX_DIM_1D_Z)
             amrex::Real unused = 0.0;
             amrex::Real tag_val = ref_parser(unused, unused, pos[0]);
 #endif
--- a/Source/Utils/WarpXTagging.cpp
+++ b/Source/Utils/WarpXTagging.cpp
@@ -15,6 +15,7 @@
 #include <AMReX_GpuControl.H>
 #include <AMReX_IntVect.H>
 #include <AMReX_MFIter.H>
+#include <AMReX_Parser.H>
 #include <AMReX_REAL.H>
 #include <AMReX_RealVect.H>
 #include <AMReX_SPACE.H>
@@ -30,8 +31,7 @@ WarpX::ErrorEst (int lev, TagBoxArray& tags, Real /*time*/, int /*ngrow*/)
     const auto problo = Geom(lev).ProbLoArray();
     const auto dx = Geom(lev).CellSizeArray();
 
-    const auto ftlo = fine_tag_lo;
-    const auto fthi = fine_tag_hi;
+    amrex::ParserExecutor<3> const& ref_parser = ref_patch_parser->compile<3>();
 
 #ifdef AMREX_USE_OMP
 #pragma omp parallel if (amrex::Gpu::notInLaunchRegion())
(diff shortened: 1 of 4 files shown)


WeiqunZhang:
if both `ref_patch_function` and `fine_tag_*` are specified in inputs, do we want to tag the union of both? If not, could we add a warning or error message?

Source/WarpX.cpp
```diff
@@ -993,10 +993,47 @@ WarpX::ReadParameters ()
...
+            bool fine_tag_lo_specified = utils::parser::queryArrWithParser(pp_warpx, "fine_tag_lo", lo);
+            bool fine_tag_hi_specified = utils::parser::queryArrWithParser(pp_warpx, "fine_tag_hi", hi);
+            std::string ref_patch_function;
+            bool parser_specified = pp_warpx.query("ref_patch_function(x,y,z)",ref_patch_function);
+            const std::string error_string = std::string( "For max_level > 0, you need to either set\
+                                             warpx.fine_tag_lo and warpx.fine_tag_hi\
+                                             or warpx.ref_patch_function(x,y,z)");
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE( ((fine_tag_lo_specified && fine_tag_hi_specified) ||
+                                                parser_specified ),
+                                                error_string);
```
dpgrote: Elsewhere, the standard has been to put the error message directly in the call to `WARPX_ALWAYS_ASSERT_WITH_MESSAGE`, instead of defining a temporary string.

Source/Utils/WarpXTagging.cpp
```diff
@@ -45,7 +45,16 @@ WarpX::ErrorEst (int lev, TagBoxArray& tags, Real /*time*/, int /*ngrow*/)
             const RealVect pos {AMREX_D_DECL((i+0.5_rt)*dx[0]+problo[0],
                                        (j+0.5_rt)*dx[1]+problo[1],
                                        (k+0.5_rt)*dx[2]+problo[2])};
```
dpgrote, line 49: I might have done this differently, to do the check between the fine tag and parser here, something like this (though this isn't complete).
```
bool tag;
if (ref_patch_parser) {
    tag = (ref_parser(pos[0], pos[1], pos[2]) == 1);
} else {
    tag = (pos > ftlo && pos < fthi);
}
```
This would avoid the somewhat messy definitions of ref_patch_function below.

RevathiJambunathan: Yes I will fix this

Source/WarpX.cpp
RevathiJambunathan: I see! 
I saw this code for reference https://github.com/ECP-WarpX/WarpX/blob/ce82a694bcf0bdcb425ad0ad73c9658430ea9663/Source/Diagnostics/BoundaryScrapingDiagnostics.cpp#L54 

Would be nice to document the standard practice for writing error messages?

Source/Utils/WarpXTagging.cpp
RevathiJambunathan: @dpgrote fixed this

dpgrote: Much nicer!

Source/WarpX.cpp
dpgrote: I used "standard" loosely. I was just noticing what was done elsewhere, but I don't think there is any official standard. I agree that it would be a good idea to have some list of standard practices for details such as this.

RevathiJambunathan:
> if both `ref_patch_function` and `fine_tag_*` are specified in inputs, do we want to tag the union of both? If not, could we add a warning or error message?

@WeiqunZhang I added a warning. If both are specified, currently, only fine_tag_* will be used, and the parser is ignored.

Source/Utils/WarpXTagging.cpp
RevathiJambunathan: Thanks for the suggestions!

Source/WarpX.cpp
RevathiJambunathan: sounds good.  Either way, I have removed the temp string :)

Source/Utils/WarpXTagging.cpp
```diff
@@ -45,7 +47,21 @@ WarpX::ErrorEst (int lev, TagBoxArray& tags, Real /*time*/, int /*ngrow*/)
...
-            if (pos > ftlo && pos < fthi) {
+            bool tag_val = 0;
```
WeiqunZhang, line 50:
```suggestion
            bool tag_val = false;
```

Source/Utils/WarpXTagging.cpp
```diff
@@ -45,7 +47,21 @@ WarpX::ErrorEst (int lev, TagBoxArray& tags, Real /*time*/, int /*ngrow*/)
...
+                amrex::Real unused = 0.0;
+                tag_val = (ref_parser(pos[0], unused, pos[1]) == 1);
+#elif defined (WARPX_DIM_1D_Z)
+                amrex::Real unused = 0.0;
+                tag_val = (ref_parser(unused, unused, pos[0]) == 1);
+#endif
+            } else {
+                tag_val = (pos > ftlo && pos < fthi);
+            }
+            if ( tag_val == 1) {
```
WeiqunZhang, line 64:
```suggestion
            if (tag_val) {
```

WeiqunZhang (changes requested):
Looks good. Have some minor comments.

RevathiJambunathan:
@WeiqunZhang done :)

WeiqunZhang (approved):

dpgrote (approved):
Looks good to me!

ax3l (approved):
Very cool, thank you!

PR #4300, opened 2023-09-13
Add `get_charge_density` to python `WarpXParticleContainer` 

roelof-groenewald:
Besides what is mentioned in the title I also changed the `_MultiFABWrapper` class in `fields.py` to directly take in a multifab instead of the WarpX tag for a predefined multifab. I then created a child class, named `_WarpXMultiFABWrapper`, to access WarpX specific (predefined) multifabs as is currently done.

The new functionality in this PR is used, for example, as follows:
```py
  pc = particle_containers.ParticleContainerWrapper(species.name).particle_container
  rho_mf = fields._MultiFABWrapper(
      pc.get_charge_density(lev=0, local=False),
      level=0
  )
  density = rho_mf[...]
```

I specifically need this for a simulation without a field solver where I need to access the species density, but since there is no field solver `rho_fp` is not initialized.

### DIFF
--- a/Python/pywarpx/fields.py
+++ b/Python/pywarpx/fields.py
@@ -77,7 +77,8 @@ class _MultiFABWrapper(object):
          ghost cells.
     """
     def __init__(self, mf, level, include_ghosts=False):
-        self.mf = mf
+        if mf is not None:
+            self.mf = mf
         self.level = level
         self.include_ghosts = include_ghosts
 
@@ -558,15 +559,7 @@ class _WarpXMultiFABWrapper(_MultiFABWrapper):
     """
     def __init__(self, mf_name, level, include_ghosts=False):
         self.mf_name = mf_name
-        self.level = level
-        self.include_ghosts = include_ghosts
-
-        self.dim = libwarpx.dim
-
-        # The overlaps list is one along the axes where the grid boundaries overlap the neighboring grid,
-        # which is the case with node centering.
-        ix_type = self.mf.box_array().ix_type()
-        self.overlaps = self._get_indices([int(ix_type.node_centered(i)) for i in range(self.dim)], 0)
+        _MultiFABWrapper.__init__(self, None, level, include_ghosts)
 
     @property
     def mf(self):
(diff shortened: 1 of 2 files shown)


RemiLehe:
Thanks for this PR @roelof-groenewald !
Could you update address the issue raised by CodeQL?

Python/pywarpx/fields.py
```diff
@@ -77,8 +76,8 @@ class _MultiFABWrapper(object):
          Note that when True, the first n-ghost negative indices will refer to the lower
          ghost cells.
     """
-    def __init__(self, mf_name, level, include_ghosts=False):
-        self.mf_name = mf_name
+    def __init__(self, mf, level, include_ghosts=False):
```
dpgrote: This is fine as you have it, but I will often do this differently, where the init has both arguments, `mf` and `mf_name` and does the appropriate thing depending on which is set (instead of defining a separate class).

Python/pywarpx/fields.py
```diff
@@ -77,8 +76,8 @@ class _MultiFABWrapper(object):
...
+        self.mf = mf
         self.level = level
         self.include_ghosts = include_ghosts
```
dpgrote: Maybe put the following lines in a method that can be called from here and from the inheritor (avoiding the code duplication).

Python/pywarpx/fields.py
```diff
@@ -544,206 +536,246 @@ def max_index(self, *args):
...
+    def mf(self):
+        # Always fetch this anew in case the C++ MultiFab is recreated
+        warpx = libwarpx.libwarpx_so.get_instance()
+        # All MultiFab names have the level suffix
+        return warpx.multifab(f'{self.mf_name}[level={self.level}]')
+
+
 def ExWrapper(level=0, include_ghosts=False):
-    return _MultiFABWrapper(mf_name=f'Efield_aux[x]', level=level, include_ghosts=include_ghosts)
+    return _WarpXMultiFABWrapper(mf_name=f'Efield_aux[x]', level=level, include_ghosts=include_ghosts)
```
dpgrote: While you are making changes here, I notice that the `f` before the string is not necessary. It is leftover from a previous version that did need it and can now be removed.

Python/pywarpx/fields.py
roelof-groenewald: I thought about that but I wasn't sure how to nicely do it while supporting the current setup where the `multifab` is newly grabbed from WarpX:
```
    @property
    def mf(self):
        # Always fetch this anew in case the C++ MultiFab is recreated
        ....
```
In this way we wouldn't be able to set `self.mf` in the `__init__` function of `_MultiFABWrapper`, but would instead need to add a new flag - something like `self.grab_mf_from_warpx = (mf is None)` and then in the `mf` property definition check if that flag is true for whether to return a fetched multifab or the one given at initialization. I thought the current approach is somewhat cleaner than that. What do you think? Do you have a nicer way in mind?

Python/pywarpx/fields.py
roelof-groenewald: I changed things to actually call the `_MultiFABWrapper` constructor to also address the codeQL issue.

Python/pywarpx/fields.py
dpgrote: I was thinking of something similar, but using a hidden variable, `self._mf`.
In the __init__, `self._mf = mf`
Then the property would be
```
@property
def mf(self):
    if self._mf is not None:
        return self._mf
    else:
        warpx = libwarpx.libwarpx_so.get_instance()
        return warpx.multifab(f'{self.mf_name}[level={self.level}]')
```

dpgrote (approved):
Looks good!

PR #4304, opened 2023-09-15
Prevent NaNs in Coulomb collision module

### LINKED ISSUE #2886: Wakefield simulation crashes with collisions
This [input file](https://github.com/ECP-WarpX/WarpX/files/8126520/inputs.txt), a standard beam-driven wakefield acceleration case, runs well on GPU (A100, on the JUWELS Booster) when collisions are turned off but crashes in the current deposition after 69 iterations when electron-electron binary Coulomb collisions are on, with error
```
amrex::Abort::0::CUDA error 700 in file /p/home/jusers/thevenet1/juwels/src/warpx/build/_deps/fetchedamrex-src/Src/Base/AMReX_GpuDevice.cpp line 626: an illegal memory access was encountered !!!
SIGABRT
See Backtrace.0 file for details
```
The same crash happens on single-GPU or multi-GPU runs. Also, I tried to reduce the problem size but the problem disappeared for a smaller problem. I have not tried on CPU.

Here are the [standard output](https://github.com/ECP-WarpX/WarpX/files/8126525/output.txt) and relevant [Backtrace](https://github.com/ECP-WarpX/WarpX/files/8126528/Backtrace.0.txt) files.

RemiLehe:
In the Coulomb collision module, we sample pairs of macroparticles and compute how the scatter against each other. 

When doing this computation, the code divides certain quantities by the relative velocity of the two particles (or rather, by their net momentum in the center-of-mass frame). If the relative motion of the two particles is very very small (so that the norm of the net momentum is 0 - or underflows for the chosen precision), this will result in NaNs. But physically, the particles should not scatter, in this case.

I added a few `if` condition to avoid this:
- When computing the parameter `s` (which determines how much scattering there is), I check that the norm of the net momentum is non-zero - otherwise `s` is set to 0
- The rest of the code is not executed when `s` is found to be equal to 0.

The Github `diff` is hard to parse, but it is really only adding these two `if` condition, and reindenting the code.

Fixes #2886

### DIFF
--- a/Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
+++ b/Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
@@ -125,161 +125,168 @@ void UpdateMomentumPerezElastic (
     }
 
     // Compute s
-    const auto tts = m1*g1s*m2*g2s/(inv_c2*p1sm*p1sm) + T_PR(1.0);
-    const auto tts2 = tts*tts;
-    T_PR s = n1*n2/n12 * dt*lnLmd*q1*q1*q2*q2 /
-          ( T_PR(4.0) * MathConst::pi * PhysConst::ep0 * PhysConst::ep0 *
-            m1*g1*m2*g2/(inv_c2*inv_c2) ) * gc*p1sm/mass_g * tts2;
+    T_PR s = 0;
+    if (p1sm > std::numeric_limits<T_PR>::min()) {
+        // s is non-zero (i.e. particles scatter) only if the relative
+        // motion between particles is not negligible (p1sm non-zero)
 
-    // Compute s'
-    const auto cbrt_n1 = std::cbrt(n1);
-    const auto cbrt_n2 = std::cbrt(n2);
-    const auto coeff = static_cast<T_PR>(
-        std::pow(4.0*MathConst::pi/3.0,1.0/3.0));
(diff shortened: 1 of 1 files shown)


RemiLehe:
Fixes #2886

Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
```diff
@@ -125,161 +125,168 @@ void UpdateMomentumPerezElastic (
     }
 
     // Compute s
-    const auto tts = m1*g1s*m2*g2s/(inv_c2*p1sm*p1sm) + T_PR(1.0);
-    const auto tts2 = tts*tts;
-    T_PR s = n1*n2/n12 * dt*lnLmd*q1*q1*q2*q2 /
-          ( T_PR(4.0) * MathConst::pi * PhysConst::ep0 * PhysConst::ep0 *
-            m1*g1*m2*g2/(inv_c2*inv_c2) ) * gc*p1sm/mass_g * tts2;
+    T_PR s = 0;
+    if (p1sm > std::numeric_limits<T_PR>::min()) {
```
ax3l, line 107: Is there a potential issue for the `b0` parameter above, too? There is a division by p1sm^2 above.

RemiLehe: That's a good point. I will update this accordingly.

ax3l (approved):
Thank you! 🚀

PR #4305, opened 2023-09-15
Feature lenses x y

RTSandberg:
This PR modifies the periodic plasma lens to allow for separate lens strengths in x and y.

TODO

- [x] Lenses in x and y
- [ ] Document
- [ ] Test in CI?


Python/pywarpx/picmi.py
```diff
@@ -1551,14 +1551,22 @@ class PlasmaLens(picmistandard.base._ClassWithInit):
...
         self.strengths_E = strengths_E
         self.strengths_B = strengths_B
-
-        assert (self.strengths_E is not None) or (self.strengths_B is not None),\
+        self.warpx_strengths_Ex = warpx_strengths_Ex
+        self.warpx_strengths_Ey = warpx_strengths_Ey
+        self.warpx_strengths_Bx = warpx_strengths_Bx
+        self.warpx_strengths_By = warpx_strengths_By
+
+        assert (self.strengths_E is not None) or (self.strengths_B is not None)\
```
dpgrote, line 1674: This check is not needed here since the same check is done in C++.

Python/pywarpx/picmi.py
```diff
@@ -1570,8 +1578,18 @@ def initialize_inputs(self):
         pywarpx.particles.repeated_plasma_lens_period = self.period
         pywarpx.particles.repeated_plasma_lens_starts = self.starts
         pywarpx.particles.repeated_plasma_lens_lengths = self.lengths
-        pywarpx.particles.repeated_plasma_lens_strengths_E = self.strengths_E
-        pywarpx.particles.repeated_plasma_lens_strengths_B = self.strengths_B
+        if self.warpx_strengths_Ex is not None or self.warpx_strengths_Ey is not None:
+            print('Using Ex and Ey, ignoring \'strengths_E\' if provided',flush=True)
```
dpgrote, line 1689: Is this print statement needed, or is this left over from debugging?
BTW, a handy trick in Python is to take advantage of single and double quotes.
```print('Using Ex and Ey, ignoring "strengths_E" if provided',flush=True)```

Also, the `if` statement is not needed. The three variables can be directly assigned. Anything set to `None` is ignored when the parameters are processed.
```
        pywarpx.particles.repeated_plasma_lens_strengths_Ex = self.warpx_strengths_Ex
        pywarpx.particles.repeated_plasma_lens_strengths_Ey = self.warpx_strengths_Ey
        pywarpx.particles.repeated_plasma_lens_strengths_E = self.strengths_E
```

And the same for `B`

File: Source/Python/Particles/MultiParticleContainer.cpp [hunk 3]
```diff
@@ -22,25 +22,39 @@ void init_MultiParticleContainer (py::module& m)
         )
 
         .def("set_plasma_lens_strength",
```
dpgrote, line 24: There should be two versions of set_plasma_lens_strength, the original one with two arguments and the new one with four.

RTSandberg:
Thanks for your comments @dpgrote !  I haven't finished checking or documenting this and then it fell by the wayside (as happens so often...).  Hopefully I get to finishing the test in the near future

PR #1600, opened 2020-12-24
Divergence cleaning for PSATD in PML

EZoni:
**Goal**
Implement divergence cleaning methods for PSATD solver in PMLs, to mitigate instabilities arising from edges of valid domain in 3D simulations.

**Description**
The main changes are:
- New PSATD equations including divergence cleaning for E and B implemented in Source/FieldSolver/SpectralSolver/SpectralAlgorithms/PMLPsatdAlgorithm.cpp;
- Damping of new diagonal components as well as F and G MultiFabs added to Source/BoundaryConditions/WarpX_PML_kernels.H;
- New input parameters `warpx.do_pml_dive_cleaning` and `warpx.do_pml_divb_cleaning` (`true` by default with PSATD);
- New CI 3D test `pml_psatd_dive_divb_cleaning`, with regression analysis only. The current runtime of the test is about 2 minutes, approximately. In order to keep the runtime as short as possible, the test runs only up to 200 iterations (where the instability would start to show up, without the divergence cleaning method) and with spectral order 8 (instead of the default 16).

**Tests**
The Nyquist noise arising in the PML region for the setup given in [inputs_3d.txt](https://github.com/ECP-WarpX/WarpX/files/5772012/inputs_3d.txt) seems to be largely mitigated by using the new divergence cleaning PSATD equations in the PML region. I tested the 3D input both on CPU and GPU (both single GPU and many GPUs on Summit).

<p align="center">
    
    
</p>

<p align="center">
    
    
</p>

PR #1513, opened 2020-11-13
Fix PSATD equations used with PML

EZoni:
**Goal**
Fix the PSATD equations used with PML and close #1420.

**Description**
The new PSATD equations require 22 distinct coefficients. The implementation did not change the design of the relevant classes.

**Tests**
The PR was tested separately, using the setup of the CI test `pml_x_psatd`:

<p align="center">
    
</p>

The benchmark for the CI test `pml_x_psatd` was updated. Also the new analysis simply checks whether the numerical reflectivity is below a certain threshold (`1e-6`). I also removed a check on charge conservation with very large tolerance (`2e-2`), since this test does not run with current correction.

PR #733, opened 2020-02-20
Implement Galilean PML

RemiLehe:
Implement Galilean PSATD equations in the PML, following FBPIC's model.

Note:
We don't merge the new class with `PsatdAlgorithmPml`, because the latter implements the correct equations obtained by analytical integration of Maxwell's equations in the PML (split equations), rather than a naive copy of the regular PSATD equations, hence the implementation between the two classes diverges quite a bit.

To-do:
- [x] Implement correct analytical equations, with F and G, as in `PsatdAlgorithmPml`
- [x] Remove extra class `PsatdAlgorithmGalileanPml` if not needed anymore

PR #4316, opened 2023-09-20
Add SymPy notebook to derive PSATD equations in PML

EZoni:
This adds the SymPy jupyter notebook `psatd_pml.ipynb` in the directory Tools/Algorithms, similar to the one implemented in #3456 but for the PSATD equations in the PML formulation.

The notebook derives the PSATD equations by integrating analytically the second-order Maxwell equations (with the additional fields F and G) in the PML formulation.

The coefficients obtained by solving the PSATD equations with this notebook are the ones currently implemented in WarpX (see #1513, #1600).

Note for the reviewers:
The "Files changed" tab does not display the notebook well.
If you select "View file" separately, it is instead displayed as a notebook.

Possible follow-up work:
- Other methods of solution could be explored (such as solving directly the first-order equations, as done in #3456 for the standard PSATD equations);
- The scalar fields F and G could be optional;
- The solution of the same equation in a Galilean frame of reference could be explored (while #733 implements a "naive" extension of the standard PSATD equations for the PML, instead of solving analytically the actual equations).


RemiLehe (approved):
Thanks!

PR #4317, opened 2023-09-22
Change default behavior of picmi.FieldDiagnostic and picmi.ParticleDiagnostic & fix CI tests

oshapoval:
This PR fixes couples of Python CI tests in PR #3745. Close #3745

Description of the initial PR #3745 by @RemiLehe: 

Right now, with picmi.FieldDiagnostic and picmi.LabFrameFieldDiagnostic, the default is to write all particle species (along with the fields). This is probably not what the user expects (note that there is a separate class -- picmi.ParticleDiagnostic -- to be used when writing particles). In addition, this default behavior could be computationally costly, if there is a large number of particles in the simulation.

Similar, with picmi.ParticleDiagnostic, the default is to write all fields, which is again probably not what the user expects.

This PR changes the default behavior, so that species are not written by default for picmi.FieldDiagnostic, and fields are not written by default for picmi.ParticleDiagnostic. It also removes the parameter warpx_write_species, since the user can use picmi.ParticleDiagnostic instead.

- [ ] update affected tests

### DIFF
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -2014,9 +2014,6 @@ class FieldDiagnostic(picmistandard.PICMI_FieldDiagnostic, WarpXDiagnosticBase):
     warpx_plot_raw_fields_guards: bool, optional
         Flag whether the raw fields should include the guard cells
 
-    warpx_write_species: bool, optional
-        Flag whether to output particle data with the diagnostic
-
     warpx_format: {plotfile, checkpoint, openpmd, ascent, sensei}, optional
         Diagnostic file format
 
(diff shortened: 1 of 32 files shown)


Python/pywarpx/picmi.py
```diff
@@ -2058,14 +2054,14 @@ def initialize_inputs(self):
...
+            self.diagnostic.set_or_replace_attr('fields_to_plot', fields_to_plot)
         self.diagnostic.plot_raw_fields = self.plot_raw_fields
         self.diagnostic.plot_raw_fields_guards = self.plot_raw_fields_guards
         self.diagnostic.plot_finepatch = self.plot_finepatch
         self.diagnostic.plot_crsepatch = self.plot_crsepatch
-        self.diagnostic.write_species = self.write_species
-
+        if 'write_species' not in self.diagnostic.argvattrs:
+            self.diagnostic.write_species = False
+            self.diagnostic.do_back_transformed_particles = False
```
dpgrote: This line is not needed (since this is not class is not the BTD).
```suggestion
```

ax3l:
cc just FYI @roelof-groenewald @aveksler1 (breaking) change coming to better comply with PICMI.

ax3l:
Looks like the runtime attribute (prev. position) of `Python_prev_positions`  was not written yet and is the last test that fails.

RemiLehe (approved):

PR #4318, opened 2023-09-22
Fix distribution map initialization for particle buffers in BTD

RevathiJambunathan:
Particle buffers are now initialized when the buffer is empty and when the k-index of the buffer is within the snapshot box, similar to the times when the field buffer is initialized.

In the current dev branch, the particle buffers are initialized everytime the number of particles in the buffer is 0. And this could lead to performance decrease as observed in the BTD simulations, where, the number of calls to the distribution mapping constructor was very high.


atmyers (approved):

ax3l (approved):
Thank you! :rocket: :sparkles:

PR #4321, opened 2023-09-25
Fix bug in clang-tidy configuration file 

lucafedeli88:
This PR fixes a bug in the clang-tidy configuration file.
As a result of this bug, some checks were not performed. This PR also addresses the issues found after the bugfix.

I've added the option
```
- key:          bugprone-narrowing-conversions.WarnOnIntegerToFloatingPointNarrowingConversion
  value:        "false"
```
because otherwise we would be forced to change code like:

```
Real xi = x + i*dx;
```
into 
```
Real xi = x + static_cast<Real>(i)*dx;
```
which I find way too pedantic.


lucafedeli88:
I've apparently introduced a bug, since a couple of tests are failing, but I can't find the bug...

lucafedeli88:
> I've apparently introduced a bug, since a couple of tests are failing, but I can't find the bug...

I've found the bug 🐛

ax3l (approved):
Thank you! :sparkles:

PR #4326, opened 2023-09-27
Maxlevel user option for external field initialization

RevathiJambunathan:
For some mesh-refinement simulations, it is useful to only initiaize the external electric and magnetic fields only on the parent grid. This PR adds an option for maximum levels upto which the fields are initialized with the external field options.

A suggested change was to remove default option for external field initialization, and instead use `AllocInitMultiFab` to initialize fields to 0.0 
This simplified a lot of the code in the InitLevelData function

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -1398,10 +1398,10 @@ External fields
 Grid initialization
 ^^^^^^^^^^^^^^^^^^^
 
-* ``warpx.B_ext_grid_init_style`` (string) optional (default is "default")
+* ``warpx.B_ext_grid_init_style`` (string) optional
     This parameter determines the type of initialization for the external
-    magnetic field. The "default" style initializes the
-    external magnetic field (Bx,By,Bz) to (0.0, 0.0, 0.0).
+    magnetic field. By default, the
+    external magnetic field (Bx,By,Bz) is initialized to (0.0, 0.0, 0.0).
     The string can be set to "constant" if a constant magnetic field is
     required to be set at initialization. If set to "constant", then an
     additional parameter, namely, ``warpx.B_external_grid`` must be specified.
@@ -1429,9 +1429,9 @@ Grid initialization
     Regarding how to prepare the openPMD data file, one can refer to
     the `openPMD-example-datasets <https://github.com/openPMD/openPMD-example-datasets>`__.
 
-* ``warpx.E_ext_grid_init_style`` (string) optional (default is "default")
+* ``warpx.E_ext_grid_init_style`` (string) optional
     This parameter determines the type of initialization for the external
-    electric field. The "default" style initializes the
+    electric field. By default, the
     external electric field (Ex,Ey,Ez) to (0.0, 0.0, 0.0).
     The string can be set to "constant" if a constant electric field is
     required to be set at initialization. If set to "constant", then an
@@ -1474,12 +1474,9 @@ Grid initialization
     than periodic.
 
 * ``warpx.maxlevel_extEMfield_init`` (default is maximum number of levels in the simulation)
-    With this parameter, the externally applied electric and magnetic fields with parameters
-    ``warpx.E_ext_grid_init_style="constant"``,
-    ``warpx.B_ext_grid_init_style="constant"``,
-    ``warpx.E_ext_grid_init_style="parse_E_ext_grid_function"`` and
-    ``warpx.B_ext_grid_init_style="parse_B_ext_grid_function"`` will not be initialized with the external field
-    for levels strictly greater than ``warpx.maxlevel_extEMfield_init``. For some mesh-refinement simulations,
+    With this parameter, the externally applied electric and magnetic fields
+    will not be applied for levels greater than ``warpx.maxlevel_extEMfield_init``.
+    For some mesh-refinement simulations,
     the external fields are only applied to the parent grid and not the refined patches. In such cases,
     ``warpx.maxlevel_extEMfield_init`` can be set to 0. Note that the other levels will be initialized to
     a default value of 0.
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -1478,8 +1478,8 @@ Grid initialization
     will not be applied for levels greater than ``warpx.maxlevel_extEMfield_init``.
     For some mesh-refinement simulations,
     the external fields are only applied to the parent grid and not the refined patches. In such cases,
-    ``warpx.maxlevel_extEMfield_init`` can be set to 0. Note that the other levels will be initialized to
-    a default value of 0.
+    ``warpx.maxlevel_extEMfield_init`` can be set to 0.
+    In that case, the other levels have external field values of 0.
 
 Applied to Particles
 ^^^^^^^^^^^^^^^^^^^^
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -1474,7 +1474,7 @@ Grid initialization
     than periodic.
 
 * ``warpx.maxlevel_extEMfield_init`` (default is maximum number of levels in the simulation)
-    With this parameter, the externally applied electric and magnetic fields with paramters
+    With this parameter, the externally applied electric and magnetic fields with parameters
     ``warpx.E_ext_grid_init_style="constant"``,
     ``warpx.B_ext_grid_init_style="constant"``,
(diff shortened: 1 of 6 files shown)


Docs/source/usage/parameters.rst
```diff
@@ -1440,6 +1440,17 @@ Grid initialization
     the field solver. In particular, do not use any other boundary condition
     than periodic.
 
+* ``warpx.maxlevel_extEMfield_init`` (default is maximum number of levels in the simulation)
+    With this parameter, the externally applied electric and magnetic fields with paramters
```
```suggestion
    With this parameter, the externally applied electric and magnetic fields with parameters
```

Source/WarpX.H
```diff
@@ -159,6 +159,12 @@ public:
     //! User-defined parser to initialize z-component of the electric field on the grid
     std::unique_ptr<amrex::Parser> Ezfield_parser;
 
+    /** Maximum level upto which the externally defined electric and magnetic fields are initialized.
```
```suggestion
    /** Maximum level up to which the externally defined electric and magnetic fields are initialized.
```

Source/Initialization/WarpXInitData.cpp
```diff
@@ -747,6 +747,27 @@ WarpX::InitLevelData (int lev, Real /*time*/)
...
+        }
+    }
+    if ( (E_ext_grid_s == "constant") || (E_ext_grid_s == "parse_e_ext_grid_function")) {
+        if (lev > maxlevel_extEMfield_init) {
+            Efield_fp[lev][0]->setVal(0.);
+            Efield_fp[lev][1]->setVal(0.);
+            Efield_fp[lev][2]->setVal(0.);
+            return;
+        }
+    }
```
EZoni: Just a question to double check if I understand what is meant to be done here: is it correct that the code executes `return;` within these `if` blocks, hence skipping the rest of the function, including what possibly needs to be done for the other field?

EZoni (commented):
Thank you, @RevathiJambunathan! I left only one question and a couple of typo fixes.

Source/Initialization/WarpXInitData.cpp
RevathiJambunathan: Hey @EZoni !! 
Thank you so much for catching that. I have fixed it now in this commit [d4a7eb6](https://github.com/ECP-WarpX/WarpX/pull/4326/commits/d4a7eb657d5a5a0ada20d4aeef8374716b88e5a5)

Docs/source/usage/parameters.rst
```diff
@@ -1452,6 +1452,17 @@ Grid initialization
     the field solver. In particular, do not use any other boundary condition
     than periodic.
 
+* ``warpx.maxlevel_extEMfield_init`` (default is maximum number of levels in the simulation)
```
ax3l, line 1476: Can you pls add this to PICMI while we add it, too? :)

RevathiJambunathan: Sure. I ll discuss with @dpgrote on how to do this!

RevathiJambunathan: done! 
Thanks to @dpgrote for the offiline discussion

Docs/source/usage/parameters.rst
```diff
@@ -1452,6 +1452,17 @@ Grid initialization
...
+    With this parameter, the externally applied electric and magnetic fields with parameters
+    ``warpx.E_ext_grid_init_style="constant"``,
```
dpgrote: This might be too much detail in this documentation. I think it might be cleaner to say something like
"With this parameter, any externally applied electric and magnetic fields will not be used to initialize the fields for levels greater than ``warpx.maxlevel_extEMfield_init``. ..."
My concern is that if in the future a new method of initializing the field is implemented, this comment will become out of date and incomplete.

Source/Initialization/WarpXInitData.cpp
```diff
@@ -747,6 +747,25 @@ WarpX::InitLevelData (int lev, Real /*time*/)
                    E_ext_grid_s.begin(),
                    ::tolower);
 
+    // Externally imposed fields are only initialized until the user-defined maxlevel_extEMfield_init.
+    // The default maxlevel_extEMfield_init value is the total number of levels in the simulation
+    // Note that if the external fields are not provided, the fields, E and B, on all levels
+    // will still be initialized with the default value of 0.
+    if ( (B_ext_grid_s == "constant") || (B_ext_grid_s == "parse_b_ext_grid_function")) {
```
dpgrote: I would suggest that this should be
```if (B_ext_grid_s != "default" && lev > maxlevel_extEMfield_init) {```
For the same reason above, in case a new method is added, and to simplify the nested `if` statements.

A general comment, this bit of code could be removed and the logic below simplified if the E and B fields were initialized when allocated, by passing in zero as the initial value when `AllocInitMultiFab` is called for these MultiFabs. This would be more robust anyway to mistakes that could leave the E and B uninitialized.

Comment (RevathiJambunathan) [reply Source/Initialization/WarpXInitData.cpp]: This is now fixed.

**Review by dpgrote — APPROVED** [review]
Looks good! Thanks!

File: Docs/source/usage/parameters.rst [hunk 7]
```diff
@@ -1452,6 +1452,14 @@ Grid initialization
     the field solver. In particular, do not use any other boundary condition
     than periodic.
 
+* ``warpx.maxlevel_extEMfield_init`` (default is maximum number of levels in the simulation)
+    With this parameter, the externally applied electric and magnetic fields
+    will not be applied for levels greater than ``warpx.maxlevel_extEMfield_init``.
+    For some mesh-refinement simulations,
+    the external fields are only applied to the parent grid and not the refined patches. In such cases,
+    ``warpx.maxlevel_extEMfield_init`` can be set to 0. Note that the other levels will be initialized to
+    a default value of 0.
```
ax3l: "Note that the other levels will be initialized to a default value of 0."

I am not sure that is true of we do an initial poisson solve.
Better would be:
```suggestion
    ``warpx.maxlevel_extEMfield_init`` can be set to 0.
    In that case, the other levels will not have external fields applied.
```
or
```suggestion
    ``warpx.maxlevel_extEMfield_init`` can be set to 0.
    In that case, the other levels have external field values of 0.
```

Source/WarpX.cpp
```diff
@@ -2191,28 +2194,28 @@ WarpX::AllocLevelMFs (int lev, const BoxArray& ba, const DistributionMapping& dm
     //
     const std::array<Real,3> dx = CellSize(lev);
 
-    AllocInitMultiFab(Bfield_fp[lev][0], amrex::convert(ba, Bx_nodal_flag), dm, ncomps, ngEB, lev, "Bfield_fp[x]");
-    AllocInitMultiFab(Bfield_fp[lev][1], amrex::convert(ba, By_nodal_flag), dm, ncomps, ngEB, lev, "Bfield_fp[y]");
-    AllocInitMultiFab(Bfield_fp[lev][2], amrex::convert(ba, Bz_nodal_flag), dm, ncomps, ngEB, lev, "Bfield_fp[z]");
+    AllocInitMultiFab(Bfield_fp[lev][0], amrex::convert(ba, Bx_nodal_flag), dm, ncomps, ngEB, lev, "Bfield_fp[x]", 0.0_rt);
```
ax3l, line 2213: Change LGTM, we did not init the value yet (oddly).

Source/Initialization/WarpXInitData.cpp
```diff
@@ -764,7 +764,10 @@ WarpX::InitLevelData (int lev, Real /*time*/)
 
     for (int i = 0; i < 3; ++i) {
 
-        if (B_ext_grid_s == "constant" || B_ext_grid_s == "default") {
+        // Externally imposed fields are only initialized until the user-defined maxlevel_extEMfield_init.
+        // The default maxlevel_extEMfield_init value is the total number of levels in the simulation
+        if ( ( B_ext_grid_s == "constant") && (lev <= maxlevel_extEMfield_init) )
```
ax3l, line 769: I do not understand the `<=` here:

As far as I am aware, `lev` goes `0...maxLevel()-1`, no?

ax3l: aho, got it

ax3l (approved):
LGTM, thank you :)

Source/WarpX.cpp
RevathiJambunathan: it was relying on the InitData function

ax3l (approved):

PR #4336, opened 2023-10-03
Electron current diagnostic for Hybrid solver

aveksler1:
This PR adds a diagnostic for electron current for the Hybrid model to C++ exposes it to Python.

There's currently some code duplication with `JFunctor.cpp` regarding allowing current to be output with no field solve.


Python/pywarpx/picmi.py
```diff
@@ -2084,6 +2087,9 @@ def initialize_inputs(self):
             B_fields_list = ['Bx', 'By', 'Bz']
             J_fields_list = ['Jx', 'Jy', 'Jz']
             A_fields_list = ['Ax', 'Ay', 'Az']
+        # Je fields list if Hybrid solver is used.
+        if type(solver) == HybridPICSolver:
+            Je_fields_list = ['Jer', 'Jet', 'Jez']
```
dpgrote: The `Je_fields_list` can always be set and doesn't need to depend on the solver type. This would avoid having to pass in `solver` since this is the only place it is used.

ax3l:
```
  File "/home/runner/.local/lib/python3.8/site-packages/pywarpx/_libwarpx.py", line 109, in load_library
    from . import warpx_pybind_3d as cxx_3d
ImportError: /home/runner/.local/lib/python3.8/site-packages/pywarpx/warpx_pybind_3d.cpython-38-x86_64-linux-gnu.so: undefined symbol: _ZN9JeFunctorC1EiiN5amrex7IntVectEbi
```
means
```
JeFunctor::JeFunctor(int, int, amrex::IntVect, bool, int)
```

Source/Diagnostics/ComputeDiagFunctors/JeFunctor.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include "FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.H"
+#include "Particles/MultiParticleContainer.H"
+
+#include <AMReX.H>
+#include <AMReX_IntVect.H>
+#include <AMReX_MultiFab.H>
+
+JeFunctor::JeFunctor (const int dir, int lev,
+                     amrex::IntVect crse_ratio,
+                     bool deposit_current, int ncomp)
```
ax3l:
```suggestion
JeFunctor::JeFunctor (int dir, int lev,
                     amrex::IntVect crse_ratio,
                     bool deposit_current, int ncomp)
```

Source/Diagnostics/ComputeDiagFunctors/JeFunctor.cpp
```diff
@@ -0,0 +1,71 @@
...
+*/
+#include "JeFunctor.H"
+
+#include "WarpX.H"
+#include "FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.H"
+#include "Particles/MultiParticleContainer.H"
+
+#include <AMReX.H>
+#include <AMReX_IntVect.H>
+#include <AMReX_MultiFab.H>
```
ax3l:
```suggestion
```

Source/Diagnostics/ComputeDiagFunctors/JeFunctor.H
```diff
@@ -0,0 +1,44 @@
+#ifndef WARPX_JEFUNCTOR_H_
+#define WARPX_JEFUNCTOR_H_
+
+#include "ComputeDiagFunctor.H"
+
+#include <AMReX_BaseFwd.H>
```
ax3l:
```suggestion
#include <AMReX_IntVect.H>
#include <AMReX_MultiFab.H>
```

ax3l:
@aveksler1 you forgot to add the new .cpp file to the CMakeLists.txt (and legacy Makefile) files.

Thus, it is not build and also not linked.

Source/Diagnostics/ComputeDiagFunctors/JeFunctor.H
```diff
@@ -0,0 +1,44 @@
...
+     * \param[out] mf_dst output MultiFab where the result is written
+     * \param[in] dcomp first component of mf_dst in which cell-centered
+     *            data is stored
+    */
+    virtual void operator()(amrex::MultiFab& mf_dst, int dcomp, int /*i_buffer=0*/) const override;
+private:
+    /** direction of the electron current density to save */
+    const int m_dir;
+    /** level on which mf_src is defined */
+    int m_lev;
```
ax3l: I think `m_deposit_current` is missing here.

aveksler1:
image
image
I wanted to move these here and follow up on this. 
I opted to change how I named the electron current field diagnostic from `"je{component_name}"` to `"j{component_name}_e"`, rather than rewriting `WarpXOpenPMDPlot::GetMeshCompNames` because I couldn't come up with a string parsing solution to respect all the current ways field diagnostics are named. For example:
1. The most common ones are 2 character, `Ex`, `jr`, etc.
2. There are also particle field diagnostics, where the species name is appended to the end of the diagnostics name. For example, if you want current from a particular species in a simulation, entering `jx` turns into `jx_{species_name}`, which is respected by the parsing implementation by moving the component to the end.
3. There are also scalars, like `parts_per_cell`, which have underscores but aren't vectors.

Any string parsing here would have a lot of edge cases, and room for misidentification of vectors and scalars. The standard naming convention for vectors right now is `"j/E/B{component_name}{extra_stuff}"` and changing this could be a big breaking change. Maybe a second identifier of a field diagnostic type not tied to the structure of the name.

Regardless of the above, this PR is ready for review again! @ax3l @dpgrote

Python/pywarpx/picmi.py
```diff
@@ -2241,11 +2241,13 @@ def diagnostic_initialize_inputs(self):
             E_fields_list = ['Er', 'Et', 'Ez']
             B_fields_list = ['Br', 'Bt', 'Bz']
             J_fields_list = ['Jr', 'Jt', 'Jz']
+            Jdisp_fields_list = ['Jr_disp', 'Jt_disp', 'Jz_disp']
```
roelof-groenewald: I think it would be better to call this `J_displacement_field_list`. It is longer but much clearer (when I see disp I think dispersion).

Source/Diagnostics/ComputeDiagFunctors/JdispFunctor.H
```diff
@@ -0,0 +1,42 @@
+#ifndef WARPX_JDISPFUNCTOR_H_
+#define WARPX_JDISPFUNCTOR_H_
+
+#include "ComputeDiagFunctor.H"
+
+#include <AMReX_BaseFwd.H>
+
+/**
+ * \brief Functor to cell-center MF for electron current density from Kinetic-Fluid
```
roelof-groenewald: This comment should be updated since we are now calling this the displacement current.

Source/Diagnostics/ComputeDiagFunctors/JdispFunctor.H
```diff
@@ -0,0 +1,42 @@
...
+     * \param[in] lev level of multifab. Used for averaging in rz.
+     * \param[in] crse_ratio for interpoalting field values from the simulation Multifab, src_mf,
+     *            to the output diagnostic MultiFab, mf_dst.
+     * \param[in] ncomp Number of component of mf_src to cell-center in dst multifab.
+    */
+    JdispFunctor (int dir, int lev, amrex::IntVect crse_ratio, int ncomp=1);
+    /** \brief Cell-center m_mf_src and write the result in mf_dst.
+     *
+     * Currently, the Kinetic-Fluid Hybrid model only supports the m=0 mode, so only
+     * that component is written.
```
roelof-groenewald: I don't see where this is enforced, could you elaborate?

Source/Diagnostics/ComputeDiagFunctors/JdispFunctor.cpp
```diff
@@ -0,0 +1,63 @@
...
+    : ComputeDiagFunctor(ncomp, crse_ratio), m_dir(dir), m_lev(lev)
+{ }
+
+void
+JdispFunctor::operator() (amrex::MultiFab& mf_dst, int dcomp, const int /*i_buffer*/) const
+{
+    auto& warpx = WarpX::GetInstance();
+    auto* hybrid_pic_model = warpx.get_pointer_HybridPICModel();
+
+    /** pointer to source1 (Ji) multifab */
```
roelof-groenewald: Comment should be changed to say total simulation current or something like that (not `Ji`).

roelof-groenewald (changes requested):
Looks good so far. Could you please add a paragraph to the Ohm's law solver documentation saying that the displacement current diagnostic is really the electron current when the hybrid-PIC scheme is used?

Source/Diagnostics/ComputeDiagFunctors/JdispFunctor.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+void
+JdispFunctor::operator() (amrex::MultiFab& mf_dst, int dcomp, const int /*i_buffer*/) const
+{
+    auto& warpx = WarpX::GetInstance();
+    auto* hybrid_pic_model = warpx.get_pointer_HybridPICModel();
+
+    /** pointer to source1 (total simulation current J) multifab */
+    amrex::MultiFab* m_mf_j = warpx.get_pointer_current_fp(m_lev, m_dir);
+    amrex::MultiFab* m_mf_j_ampere;
```
roelof-groenewald: I think we can be a bit more explicit here, rather than using the specific language of the hybrid-PIC model.
```suggestion
    amrex::MultiFab* m_mf_curlB;
```

Source/Diagnostics/ComputeDiagFunctors/JdispFunctor.cpp
```diff
@@ -0,0 +1,71 @@
...
+    amrex::MultiFab* m_mf_j_external;
+    if (hybrid_pic_model) {
+        /** pointer to source2 (current calculated from Ampere's Law, Jamp) multifab */
+        m_mf_j_ampere = hybrid_pic_model->get_pointer_current_fp_ampere(m_lev, m_dir);
```
roelof-groenewald:
```suggestion
        m_mf_curlB = hybrid_pic_model->get_pointer_current_fp_ampere(m_lev, m_dir);
```

Source/Diagnostics/ComputeDiagFunctors/JdispFunctor.cpp
```diff
@@ -0,0 +1,71 @@
...
+        // Multiply J multifab by mu0
+        m_mf_j->mult(PhysConsts::mu0)
+
+        // Displacement current = curlB - mu0J
+        amrex::MultiFab::LinComb(
+            Jdisp, 1, *m_mf_curl_b, 0,
+            -1, *m_mf_j, 0, 0, 1, Jdisp.nGrowVect()
+        )
+
+    } */
```
roelof-groenewald:
```suggestion
    // J_displacement = curl x B / mu0 - J
    amrex::MultiFab::LinComb(
        Jdisp, 1, *m_mf_curlB, 0,
        -1, *m_mf_j, 0, 0, 1, Jdisp.nGrowVect()
    );
    
    // subtract external current in the hybrid-PIC case
    if (hybrid_pic_model) {
        amrex::MultiFab::Subtract(Jdisp, *m_mf_j_external, 0, 0, 1, Jdisp.nGrowVect());
    } 
```

Source/Diagnostics/ComputeDiagFunctors/JdispFunctor.cpp
```diff
@@ -0,0 +1,71 @@
...
+    if (hybrid_pic_model) {
+        /** pointer to source2 (current calculated from Ampere's Law, Jamp) multifab */
+        m_mf_j_ampere = hybrid_pic_model->get_pointer_current_fp_ampere(m_lev, m_dir);
+        /** pointer to source3 (external currents, Jext) multifab */
+        m_mf_j_external = hybrid_pic_model->get_pointer_current_fp_external(m_lev, m_dir);
+    } else {
+        // To finish this implementation, we need to implement a method to
+        // calculate (∇ x B).
+        WARPX_ABORT_WITH_MESSAGE(
+            "Displacement current diagnostic is only implemented for the HybridPICModel.");
```
roelof-groenewald, line 44:
```suggestion
        WARPX_ABORT_WITH_MESSAGE(
            "Displacement current diagnostic is only implemented for the HybridPICModel.");
            
        // Skeleton for future implementation for solvers other than HybridPIC.
        // Get curlB multifab

        // Multiply J multifab by mu0
        m_mf_curlB->mult(1.0/PhysConsts::mu0) */
```

Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -700,11 +727,17 @@ FullDiagnostics::InitializeFieldFunctors (int lev)
             } else if ( m_varnames[comp] == "jy" ){
                 m_all_field_functors[lev][comp] = std::make_unique<JFunctor>(1, lev, m_crse_ratio, true, deposit_current);
                 deposit_current = false;
-            } else if ( m_varnames[comp] == "Ax" ){
+            } else if ( m_varnames[comp] == "jx_displacement" ){
+                m_all_field_functors[lev][comp] = std::make_unique<JdispFunctor>(0, lev, m_crse_ratio);
+            } else if ( m_varnames[comp] == "jy_displacement" ){
+                m_all_field_functors[lev][comp] = std::make_unique<JdispFunctor>(1, lev, m_crse_ratio);
+            }
+            else if ( m_varnames[comp] == "Ax" ){
```
roelof-groenewald: Just to stick with the convention used in the rest of this switch statement.
```suggestion
            } else if ( m_varnames[comp] == "Ax" ){
```

roelof-groenewald (changes requested):
Just a few suggested changes to reduce code duplication.

Source/Diagnostics/ComputeDiagFunctors/JdispFunctor.cpp
```diff
@@ -0,0 +1,71 @@
...
+    }
+    // A Jdisp multifab is generated to hold displacement current.
+    amrex::MultiFab Jdisp( m_mf_j->boxArray(), m_mf_j->DistributionMap(), 1, m_mf_j->nGrowVect() );
+
+    if (hybrid_pic_model) {
+        // Jdisp = Jamp - Ji
```
aveksler1, line 47:
```suggestion
        // Je = (1/mu0)*curl(B) - Ji - Jext. curl(B) from hybrid model already has (1/mu0) prefactor.
```

Source/Diagnostics/ComputeDiagFunctors/JdispFunctor.cpp
```diff
@@ -0,0 +1,65 @@
...
+        // To finish this implementation, we need to implement a method to
+        // calculate (∇ x B).
+        WARPX_ABORT_WITH_MESSAGE(
+            "Displacement current diagnostic is only implemented for the HybridPICModel.");
+
+        // Skeleton for future implementation for solvers other than HybridPIC.
+        // Get curlB multifab
+
+        // Multiply J multifab by mu0
+        m_mf_curlB->mult(1.0/PhysConsts::mu0)
```
aveksler1:
```suggestion
        // Divide curlB multifab by mu0 to get units of current
        // m_mf_curlB->mult(1.0/PhysConsts::mu0)
```

roelof-groenewald (approved):
Looks good! Thanks for the effort on this.

PR #4337, opened 2023-10-03
Ensure that mass is specified for particle species.

lucafedeli88:
If `mass` and `species_type` are not specified for a given species, mass is set to zero, which leads to a crash for non-photon particles. This check ensures that either `mass` or `species_type` are specified, or that particles are read from an external file.


RemiLehe (approved):

Source/Utils/SpeciesUtils.cpp
```diff
@@ -58,6 +50,22 @@ namespace SpeciesUtils {
             species_name + "'."
         );
 
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+            mass_is_specified ||
+            species_is_specified ||
+            (injection_style == "external_file"),
```
ax3l, line 56: In the external file logic, is the particle mass truly required or optional?

If it is the latter case, you need to check there as well.

PR #4133, opened 2023-07-28
Clang tidy CI test: add some modernize checks to clang-tidy CI test

lucafedeli88:
Specifically, this PR adds the following checks:

- [modernize-deprecated-headers](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize-deprecated-headers.html)
- [modernize-deprecated-ios-base-aliases](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize-deprecated-ios-base-aliases.html)
- [modernize-loop-convert](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize-loop-convert.html)
- [modernize-make-shared](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize-make-shared.html)
- [modernize-make-unique](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize-make-unique.html)
- [modernize-pass-by-value](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize-pass-by-value.html)
- [modernize-raw-string-literal](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize-raw-string-literal.html)
- [modernize-redundant-void-arg](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize-redundant-void-arg.html)
- [modernize-replace-auto-ptr](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize-replace-auto-ptr.html)
- [modernize-replace-disallow-copy-and-assign-macro](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize-replace-disallow-copy-and-assign-macro.html)
- [modernize-replace-random-shuffle](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize-replace-random-shuffle.html)
- [modernize-return-braced-init-list](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize-return-braced-init-list.html)
[...]

PR #4338, opened 2023-10-03
Close #4331 segfault when activating nuclearfusion collisions

### LINKED ISSUE #4331: segfault when activating nuclearfusion collisions
Hi all, thanks for the amazing software package.
I was able to compiled and run the software, but once I include fusion reactions the simulation segfaults during the first timestep. I have attached a minimal example. (I have redacted a few paths)
The simulaton runs fine when I comment out the collision. The simulation also runs fine when I instead include a pairwisecoulomb collision. However the simulation also segfaults if I include a pairwisecoulomb collision but do not comment out the product_species line.

i.e.
collisions.collision_names = dt
dt.type = nuclearfusion
dt.species = t d
dt.product_species = n a
=> SEGFAULT

collisions.collision_names = dt
dt.type = pairwisecoulomb
dt.species = t d
dt.product_species = n a
=> SEGFAULT

collisions.collision_names = dt
dt.type = pairwisecoulomb
dt.species = t d
=> no SEGFAULT

I am not sure if it is a problem with the config file, the build, or the software in itself.
[files.zip](https://github.com/ECP-WarpX/WarpX/files/12777665/files.zip)

RemiLehe:
This PR fixes two issues that were raised in #4331:

- The code segfaults if the user specifies product species when using the Coulomb collisions. (Coulomb collisions do not produce new particles, and thus no product species should be specified.)

- The code segfaults if the user selects the nuclear fusion module, but the reacting species contain no macroparticles. This issue is due to a bug in WarpX, where we wanted the `ParticleCreationFunctor` to return an `amrex::Vector<int>` filled with 0, but used the wrong syntax.

Regression from #4133


Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H
```diff
@@ -115,7 +115,7 @@ public:
     {
         using namespace amrex::literals;
 
-        if (n_total_pairs == 0) return {m_num_product_species, 0};
```
RemiLehe, line 118: Our intent was to return an `amrex::Vector<int>` of size `m_num_product_species` filled with `0`, but instead this actually returns an `amrex::Vector<int>` of size `2`, filled with `[ m_num_product_species, 0 ]`.

ax3l: Interesting, this complains about it:
```
Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H:118:40: warning: avoid repeating the return type from the declaration; use a braced initializer list instead [modernize-return-braced-init-list]
        if (n_total_pairs == 0) return amrex::Vector<int>(m_num_product_species, 0);
                                       ^
```

@lucafedeli88 I think #4133 changed this by accident.

RemiLehe: Oh, ok. But then I am not sure how get the correct behavior, while still making `clang-tidy` happy. 
Any suggestion @ax3l @lucafedeli88 ? As a reminder, using the following braced initializer list does not do what we want: 
```
return {m_num_product_species, 0}
```

lucafedeli88: ah! I see... that's true... I am very sorry. 
I blindly followed this suggestion, which is misleading in this case
```
2023-10-03T19:15:46.7620300Z /home/runner/work/WarpX/WarpX/Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H:118:40: warning: avoid repeating the return type from the declaration; use a braced initializer list instead [modernize-return-braced-init-list]
2023-10-03T19:15:46.7621127Z         if (n_total_pairs == 0) return amrex::Vector<int>(m_num_product_species, 0);
```

ax3l: @lucafedeli88 I think this test has a bug and does not exclude transformations of non-init-list constructors.
[Looking at the tests here](https://reviews.llvm.org/D28768), we might want to:
- exclude the test for now
- report a bug with LLVM

ax3l (approved):
Thank you! Sorry I overlooked this!

There is indeed a bug in clang-tidy for this constructor -.-

Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H
lucafedeli88: I agree

lucafedeli88 (approved):
I agree with @ax3l : we should exclude the buggy check in clang-tidy for now.

Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H
ax3l: @lucafedeli88 I'll file the LLVM bug.

ax3l: @lucafedeli88 filed in https://github.com/llvm/llvm-project/issues/68159

ax3l:
A fix for this bug in clang-tidy just landed in LLVM `main`: https://github.com/llvm/llvm-project/pull/68491

PR #4339, opened 2023-10-03
Set the particle fields directly to the constant external fields

dpgrote:
This PR is a quick fix to the problem of register pressure that arises because the large block of code for adding external fields is included in the case when only a simple constant field is needed. This PR moves the constant field out of the `GetExternalEBField` class and instead applies the constant field directly to the particles. The fields for the particles are initialized to the constant field instead of to zero (with the constant defaulting to zero).

This implies a small change to the user interface. Now, when setting the constant fields, the user can set particles.E_external_particle and/or particles.B_external_particle without the need to set `particles.E_ext_particle_init_style = constant` (and same for B). This is perhaps useful since the constant field can be set in addition to another external field. For now, if the init style is set to constant, it is ignored.


ax3l:
Hi @dpgrote, this is great!

> For now, if the init style is set to constant, it is ignored.

Should we make this throw a warning? (Or does it anyway throw an unused parameter warning now?)

Can you update the user-facing docs for this?

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -90,6 +91,17 @@ WarpXParticleContainer::WarpXParticleContainer (AmrCore* amr_core, int ispecies)
     SetParticleSize();
     ReadParameters();
 
+    // This needs to be here since ReadParameters is static but the m_E_external_particle and B are not
+    // Should they be declared static?
```
ax3l: This is ok, not ideal that we use so many statics that `ReadParameters` is static, too.
```suggestion
```

dpgrote:
> Hi @dpgrote, this is great!
> 
> > For now, if the init style is set to constant, it is ignored.
> 
> Should we make this throw a warning? (Or does it anyway throw an unused parameter warning now?)
> 
> Can you update the user-facing docs for this?

Either way is fine. There are three options. Which is your preference?

1. Keep it as is now, where constant is ignored (with no warnings)
2. Add a warning if constant is used
3. Revert to original interface where constant is needed to specify the fields

ax3l (approved):
This looks great, but the documentation in `parameters.rst` is not changed to reflect the change in logic. Please post as a follow-up PR?

PR #4340, opened 2023-10-04
Reorder WarpX.H so that public, protected, and private only appear once

lucafedeli88:
This PR is a minor refactoring of WarpX.H: now `public:` , `protected:`, and `private:` only appear once.  
This PR implements some parts of https://github.com/ECP-WarpX/WarpX/pull/4275 , which is currently on hold due to ongoing discussions on how to simplify the interface of the WarpX class (not related to the changes introduced with this PR)


Source/WarpX.H
```diff
@@ -1042,6 +1042,108 @@ public:
...
+#endif
+
+    void InitEB ();
+    /**
+    * \brief Compute the length of the mesh edges. Here the length is a value in [0, 1].
+    *        An edge of length 0 is fully covered.
+    */
+
+#ifdef AMREX_USE_EB
+    static void ComputeEdgeLengths (std::array< std::unique_ptr<amrex::MultiFab>, 3 >& edge_lengths,
```
ax3l:
```suggestion

#ifdef AMREX_USE_EB
    /**
    * \brief Compute the length of the mesh edges. Here the length is a value in [0, 1].
    *        An edge of length 0 is fully covered.
    */
    static void ComputeEdgeLengths (std::array< std::unique_ptr<amrex::MultiFab>, 3 >& edge_lengths,
```

Source/WarpX.H
```diff
@@ -1042,6 +1042,108 @@ public:
...
+    *          - 0 for unstable cells
+    *          - 1 for stable cells which have not been intruded
+    *          - 2 for stable cells which have been intruded
+    *        Here we cannot know if a cell is intruded or not so we initialize all stable cells with 1
+    */
+    void MarkCells();
+    /**
+    * \brief Compute the level set function used for particle-boundary interaction.
+    */
+#endif
```
ax3l:
```suggestion
#endif

    /**
    * \brief Compute the level set function used for particle-boundary interaction.
    */
```

ax3l (approved):
Thanks, LGTM.

PR #4351, opened 2023-10-09
Add new PRL paper using WarpX

RemiLehe:


lucafedeli88 (approved):
Thanks a lot for updating the list of papers using WarpX.

PR #4357, opened 2023-10-10
[WIP] Correct behavior of Python field wrapper when extracting guard cells

RemiLehe:


RemiLehe:
The entire file has since been rewritten with a completely new `MultiFabWrapper` class and imaginary-number indexing interface for ghost cells — neither `_MultiFABWrapper` nor `_get_intersect_slice` exist in the current code.

PR #4360, opened 2023-10-10
[WIP]Call WriteJobInfo for openpmd diagnostics

dpgrote:
This PR has been rewritten to write out the job info as attributes in the openpmd file. This makes the attributes directly accessible from the Python interface.

Todo - is there a nice way to write out the input parameters to the openpmd file? They could be written out as attributes also, but that is a large number of attributes and would make it hard to find others. Is there a way to write out a dictionary?


ax3l:
Sounds good, @dpgrote we could also add more of this info as attributes to openPMD diags.
We can add arbitrary attributes we like to our files (does not have to be standardized).

ax3l (dismissed):

Source/Diagnostics/FlushFormats/FlushFormat.H
```diff
@@ -27,7 +27,10 @@ public:
         bool isLastBTDFlush = false,
         const amrex::Vector<int>& totalParticlesFlushedAlready = amrex::Vector<int>() ) const = 0;
 
-     virtual ~FlushFormat() {}
+    virtual ~FlushFormat() {}
+
+    /** Write general info of the run into the plotfile */
+    void WriteJobInfo(const std::string& dir) const;
```
ax3l: @dpgrote I wonder if this should be written by full diagnostics at all - or instead rather like the `warpx_used_inputs` file just in general, when the simulation starts?

dpgrote: Good question. I like this kind of thing to be written out in the same place as the diagnostics so that it is associated with the output of the simulation. If this is written at the startup, the code would have to figure out where the diagnostics will be written.

dpgrote:
Writing out this info as attributes in the openPMD data file is probably a better idea, so that it could be accessed in that standard way. This could include all of the input parameters as well which would be useful. I submitted this PR as a quick way to have this data accessible to see if anyone was interested. I needed it in one case so thought I would submit it. I do notice that the way this is setup for openPMD, this file will be written out and overwritten each time the diagnostics are written out.

PR #4361, opened 2023-10-11
CMake: `pip_install_nodeps` Target

ax3l:
Add a target that does not search of check any dependencies with pip. Useful in package managers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -686,6 +686,17 @@ if(WarpX_PYTHON)
             ${WarpX_CUSTOM_TARGET_PREFIX}pip_install_requirements
             ${_EXTRA_INSTALL_DEPENDS}
     )
+
+    # this is for package managers only
+    add_custom_target(${WarpX_CUSTOM_TARGET_PREFIX}pip_install_nodeps
+        ${CMAKE_COMMAND} -E env WARPX_MPI=${WarpX_MPI}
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -686,6 +686,17 @@ if(WarpX_PYTHON)
             ${WarpX_CUSTOM_TARGET_PREFIX}pip_install_requirements
             ${_EXTRA_INSTALL_DEPENDS}
     )
+
+    # this is for package managers only
+    add_custom_target(${WarpX_CUSTOM_TARGET_PREFIX}pip_install_nodeps
+        ${CMAKE_COMMAND} -E env WARPX_MPI=${WarpX_MPI}
+            python3 -m pip install --force-reinstall --no-index --no-deps ${PYINSTALLOPTIONS} --find-links=warpx-whl pywarpx
```
ax3l:
```suggestion
            ${Python_EXECUTABLE} -m pip install --force-reinstall --no-index --no-deps ${PYINSTALLOPTIONS} --find-links=warpx-whl pywarpx
```

roelof-groenewald (approved):
In the case of someone installing pywarpx from a package manager the version would presumably change between different releases, so in this case the `--force-reinstall` is not really needed, right? It probably doesn't hurt to keep it, just asking out of curiosity. I guess the standard is to not have it by default.

ax3l:
It does not hurt if no package is found, but it is needed if you want to iterate fast in development and have already all deps (skip the check)

PR #4362, opened 2023-10-11
Electrostatic Diagnostics for `j`: Default to Direct Deposition

ax3l:
Fix #4358

- Change Default: Direct Current Deposition for ES (Used as a diagnostics for `j`.)
- ~Esirkepov: Ensure `dt` fits into cells with an assert~ #4364

Issue #4358, opened 2023-10-10
When specifying domain sizes using variables, writing output files smashes the stack


electroflow:
Hi all, thanks for patching the last issue so fast!

When I compile the latest version under Ubuntu 22.04.3 running in WSL i get "*** stack smashing detected ***: terminated" when I specify the domain in terms of a variable and write an output file. The problem does not occour when either not writing out files or specifying the domain size using constants. Example:

```
my_constants.rho = 1e20
my_constants.eV = 1.6e-19
my_constants.c = 299792458

my_constants.vm = 0.004

my_constants.energy = 10 * eV
my_constants.ve = (2 * energy / m_e) ** 0.5

my_constants.th_e = (2 * energy / m_e) ** 0.5 / c
my_constants.th_p = (2 * energy / m_p) ** 0.5 / c

my_constants.debye = (energy * epsilon0 / rho / q_e ** 2) ** 0.5

warpx.const_dt = 0.1 * debye / ve
warpx.do_electrostatic = labframe

amr.n_cell =  32 32 32

amr.blocking_factor = 32

geometry.dims = 3
geometry.prob_lo = 0.0 0.0 0.0  
geometry.prob_hi = debye debye debye

amr.max_level = 0

stop_time = 2e-10

boundary.field_lo = pec pec periodic 
boundary.field_hi = pec pec periodic
boundary.particle_lo = reflecting reflecting periodic
boundary.particle_hi = reflecting reflecting periodic

particles.species_names = e p

algo.particle_shape = 1

e.species_type = electron
e.injection_style = "NRandomPerCell"
e.num_particles_per_cell = 1
e.momentum_distribution_type = "gaussian"

e.uz_th = th_e
e.uy_th = th_e
e.ux_th = th_e

e.profile = constant
e.density = rho

p.species_type = proton
p.injection_style = "NRandomPerCell"
p.num_particles_per_cell = 1
p.momentum_distribution_type = "gaussian"

p.uz_th = th_p
p.uy_th = th_p
p.ux_th = th_p

p.profile = constant
p.density = rho

diagnostics.diags_names = particlesr

particlesr.intervals = 1000
particlesr.diag_type = Full
particlesr.format = openpmd
particlesr.write_species = 1
particlesr.species = e p
```

Gives the output

```
AMReX (23.10) initialized
WarpX (23.10)

    __        __             __  __
    \ \      / /_ _ _ __ _ __\ \/ /
[...]

PR #4364, opened 2023-10-11
[WIP] Esirkepov: Ensure Particle Does not Run out Dt

ax3l:
Esirkepov: Ensure `dt` fits into cells with an assert

Re #4358
Follow-up to #4362

- [ ] Resolve comment by @roelof-groenewald in https://github.com/ECP-WarpX/WarpX/pull/4362#discussion_r1355752120

> Can this be an AMREX_ASSERT_WITH_MESSAGE instead? The actual dt*PhysConst::c < min_dx is a stronger constraint than really necessary - would only be needed for particles traveling at the speed of light. I frequently use Esirkepov deposition with the hybrid solver (since it is faster than direct current deposition - not sure why - and less noisy).
The message might also be less rigid - something like "Time step could be too large for Esirkepov..."

### DIFF
--- a/Source/Particles/WarpXParticleContainer.cpp
+++ b/Source/Particles/WarpXParticleContainer.cpp
@@ -354,12 +354,33 @@ WarpXParticleContainer::DepositCurrent (WarpXParIter& pti,
     // Jx, Jy and Jz have the same number of guard cells, hence it is sufficient to check for Jx
     const amrex::IntVect range = jx->nGrowVect() - shape_extent;
 #endif
+
+    const std::array<Real,3>& dx = WarpX::CellSize(std::max(depos_lev,0));
+
+    // what is the length of the smallest cell edge?
+#if defined(WARPX_DIM_3D)
+    const Real min_dx = dx[0] < dx[1] ? dx[0] : (dx[1] < dx[2] ? dx[1] : dx[2]);
+#elif defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
+    const Real min_dx = dx[0] < dx[2] ? dx[0] : dx[2];
+#else
+    const Real min_dx = dx[2];
+#endif
+
+    // dt shift would be out-of-range?
+    // We also deposit the current for diagnostics reasons, e.g., in ES simulations that have no CFL
+    // Esirkepov's current deposition algorithm assumes that dt*c does not cross a cell boundary,
+    // let's ensure that.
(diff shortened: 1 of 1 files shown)


ax3l:
@roelof-groenewald in https://github.com/ECP-WarpX/WarpX/pull/4362#discussion_r1355802414
> Thanks for the reply. A runtime option to skip the assert would be fine. There is also currently a warning raised if a user chooses to use Esirkepov with the hybrid solver saying that if a particle gains too much speed a segfault will occur in the current deposition. It is not great for ensuring smooth code execution but it at least gives the user a starting place to look if they find that their simulation crashes.
>
> An alternative option I have seen used is to simply limit particle speeds - using the same expression you have to calculate the maximum allowable speed and if a particle is pushed so that it would exceed that speed its speed is set equal to the maximal allowed speed. This obviously violates energy conservation but it prevents the code from crashing. I guess a different strategy could be to set the speed of light to the maximum supported speed in which case the speed limit will be naturally applied through the relativistic corrections... All of that is just for interest sake, I'm not suggesting we do that.
>
> Regarding Esirkepov being faster than direct deposition. I see that with few GPU runs but with many particles, using the hybrid solver. For example, a simulation using 2 Perlmutter GPUs with 100 x 100 x 184 grid size and 400M particles achieves ~1.4 steps per second with Esirkepov but only ~1.15 steps per second with direct deposition. I'll add this to the dev meeting agenda for next week.

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -354,12 +354,33 @@ WarpXParticleContainer::DepositCurrent (WarpXParIter& pti,
...
+    const Real min_dx = std::min(dx[0],dx[2]);
+#else
+    const Real min_dx = dx[2];
+#endif
+
+    // dt shift would be out-of-range?
+    // We also deposit the current for diagnostics reasons, e.g., in ES simulations that have no CFL
+    // Esirkepov's current deposition algorithm assumes that dt*c does not cross a cell boundary,
+    // let's ensure that.
+    if (WarpX::current_deposition_algo == CurrentDepositionAlgo::Esirkepov) {
```
ax3l, line 373: Do not check on `ElectromagneticSolverAlgo::HybridPIC` / `algo.maxwell_solver = hybrid`, ... https://warpx.readthedocs.io/en/latest/usage/parameters.html#maxwell-solver

Technically, only for EM sims with Yee, CKC, PSATD, ...

PR #4233, opened 2023-08-25
Python: Clean Out Old Bindings

ax3l:
Remove old Python binding code.

Follow-up to #3474

PR #4366, opened 2023-10-12
Add pyflakes check in Github actions

RemiLehe:


RemiLehe:
Needs #4233 to fix `pyflakes` errors.

ax3l:
@RemiLehe can we add this to `.pre-commit-config.yaml` instead, ideally if it supports auto-fixes?

ax3l:
Ah, we found no pre-commit for it.
Since pyflake only lints Python code but has no auto-fixes, we will add it as currently proposed :)

ax3l (approved):

ax3l:
Ah no, I found a way to keep it in pre-commit (w/o auto fixes): https://github.com/AMReX-Codes/pyamrex/pull/208

.github/workflows/source.yml
```diff
@@ -13,8 +13,16 @@ concurrency:
...
 jobs:
-  style:
+  lint:
+    runs-on: ubuntu-22.04
+
+    steps:
+    - name: check WarpX python code
+      run: |
+        python3 -m pip install pyflakes
+        pyflakes .
```
ax3l:
```suggestion
        python3 -m pyflakes .
```

.github/workflows/source.yml
```diff
@@ -13,8 +13,16 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  style:
+  lint:
+    runs-on: ubuntu-22.04
+
+    steps:
+    - name: check WarpX python code
```
ax3l, line 21: I forgot to get the source code:
```suggestion
    - uses: actions/checkout@v3
    - name: check WarpX python code
```

ax3l:
Let use use `ruff` as in ImpactX & pyAMReX - it has the same job but is very fast :)
https://github.com/ECP-WarpX/impactx/blob/4676a26d4d14a7cc099ad63abbd703f9b2297726/.pre-commit-config.yaml#L95-L102

PR #4367, opened 2023-10-12
Restore `get_particle_boundary_buffer_structs` functionality

roelof-groenewald:
Restore `get_particle_boundary_buffer_structs` functionality from before pyamrex transition.
As before, in order to access particle struct data for the boundary buffer one would do, for example:
```
boundary_buffer = particle_containers.ParticleBoundaryBufferWrapper()
parts_structs = boundary_buffer.get_particle_boundary_buffer_structs(ions.name, boundary, 0)
for struct in parts_structs:
    print(struct['x'])
```

Also includes a minor code clean-up in `get_particle_boundary_buffer()`.


ax3l (approved):
Thank you, this looks great! :tada:

PR #4370, opened 2023-10-13
Python: use valid cells for fields getitem

dpgrote:
This PR fixes the `getitem` in fields.py when include_ghosts is True so that internal to the domain, it only uses valid cells (instead of values from the ghost cells). It only uses ghost cells on the outer edge of the domain.


ax3l:
Added in https://github.com/AMReX-Codes/pyamrex/pull/370

dpgrote:
This PR has been superseded by PR #5407

PR #4371, opened 2023-10-13
Add filters for boundary scraping diagnostics

kale-j:
Implementation of filtering to optionally reduce the number of particles written by boundary scraping diagnostics

Motivation:
Writing out all the particles leaving the simulation domain or crossing an embedded boundary is often unnecessary and can be very costly. An example of this is the common case wherein one wishes to only diagnose the hot tail of the electron distribution. For a large 3D simulation, writing all of the electrons leaving the domain to file can significantly complicate the analysis due to large file sizes/slow post-processing when only a subset is actually desired.

About the implementation:
I took the approach of applying particle filtering at the time the file is written (not the time the particles encounter the boundary). While this incurs the same buffer memory burden as not applying any filters, it significantly simplifies the implementation relative to filtering the particles at the time they encounter the boundary -- very few code changes were needed. Also, when memory is not a problem, this is *probably* the more cost effective way to do things (if someone wants to change this, I would suggest based on past experience with epoch's "particle probe" feature that they test whether this would affect run time). The one unintended and potentially undesirable consequence of this approach is that "t" in plot_filter_function is the time the file is written rather than the time the particle encountered the boundary.

Changes made:
WarpXOpenPMD.cpp: 
- Separated out how the particles are written for the back transformed diagnostic and the "other" pinned_pc case (presumably, the "other" pinned_pc case is only for BoundaryScrapingDiagnostics)
- Copy/pasted the way filters are handled for FullDiagnostics
Documentation:
- Updated section on BoundaryScrapingDiagnostics to add options for filtering
- Added note about the implications of applying the filtering at the write time of the file rather than the collection time of the particles
Regression tests:
- Added new test scraping_filter, copied from scraping, with the restriction that only particles with z > 0 are written (i.e., half of the number of particles). In the analysis script (also copied from scraping), the check on the total number of particles now uses twice the number of particles written to the BoundaryScrapingDiagnostic.

Additional testing done (behind the scenes):
I used the modified code for a mid-scale (64 gpu nodes) production run on Perlmutter


ax3l:
Thank you and congratulations on your first PR to WarpX, Kale! :sparkles: 

We'll take a look in the coming days! Have a nice weekend!

Examples/Tests/scraping/analysis_rz_filter.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python
+
+# Copyright 2019-2021 Yinjian Zhao
```
roelof-groenewald: This can be updated with your name and 2023.

Examples/Tests/scraping/analysis_rz_filter.py
```diff
@@ -0,0 +1,65 @@
...
+    w, = ts_full.get_particle(['w'], iteration=iteration)
+    return len(w)
+def n_scraped_particles( iteration ):
+    timestamp = ts_scraping.get_particle( ['timestamp'], iteration=ts_scraping.iterations[0] )
+    return (timestamp <= iteration).sum()
+n_remaining = np.array([ n_remaining_particles(iteration) for iteration in ts_full.iterations ])
+n_scraped = np.array([ n_scraped_particles(iteration) for iteration in ts_full.iterations ])
+n_total = n_remaining[0]
+assert np.all( 2*n_scraped+n_remaining == n_total)
+
```
roelof-groenewald: Since you are filtering for electrons scraped at `z > 0` it would be good to have a check in here that all electrons in the boundary diagnostic were scraped at z > 0.

Examples/Tests/scraping/analysis_rz_filter.py
```diff
@@ -0,0 +1,65 @@
...
+# Checksum test
+test_name = os.path.split(os.getcwd())[1]
+checksumAPI.evaluate_checksum(test_name, fn, do_particles=False)
```
roelof-groenewald: I don't think this checksum comparison is necessary since you are not evolving the fields anyway, and not comparing particles in the simulation.

Regression/Checksum/benchmarks_json/scraping_filter.json
```diff
@@ -0,0 +1,8 @@
+{
```
roelof-groenewald: If you remove the checksum comparison this file can be removed.

roelof-groenewald (changes requested):
Hi Kale. Thanks for this PR!

I left just a few comments on things that can improve the CI test you added.

Btw, it would in principle be possible to apply the filter to the time a particle was scraped since the particles in the buffer record the timestep at which they were scraped. But this would require some changes to how the filter is applied which can be left to a future PR.

ax3l:
@kale-j when you have the time, let us know if you have any questions on @roelof-groenewald 's feedback :)

RemiLehe:
Note that this PR might be superseded by https://github.com/ECP-WarpX/WarpX/pull/4420. (But the automated tests that @kale-j implemented are definely worth keeping!)

As discussed with @RevathiJambunathan, in #4420, we will apply particle selection in the particle functor directly. This will in particular avoid having to duplicate the code for particle selection in different format-specific file (e.g. for openPMD, plotfiles, etc.)

kale-j:
Hi Roelof, Axel, and Remi,
Great suggestions. I've now made the changes that you requested. Sorry for the long delay -- I've been out sick for the last 3 week. 
Let me know if there's anything else. Even if this gets superseded by a more recent PR, I'll be happy if you keep the functionality.

Examples/Tests/scraping/analysis_rz_filter.py
kale-j: Done

Examples/Tests/scraping/analysis_rz_filter.py
kale-j: Removed

Regression/Checksum/benchmarks_json/scraping_filter.json
kale-j: Removed

Examples/Tests/scraping/analysis_rz_filter.py
kale-j: I added a test for this (it checks that no particles have z <= 0)

RemiLehe (approved):

roelof-groenewald (approved):
Thanks for updating.

PR #4372, opened 2023-10-13
Properly flush metadata of incomplete BTD, at the end of simulation

RemiLehe:
This PR makes sure that we write openPMD metadata (esp. constant record) for all files at the end of the simulation. In particular, BTD snapshots that are incomplete (i.e. do not have all slices written yet) also have their metadata written.


RemiLehe:
I just tested this in production simulations and it does work as expected.

ax3l (approved):
Thank you for this, looks great!

ax3l:
Looks like @roelof-groenewald sees a regression of the form
```
ValueError: [Fri Nov 03 13:06:44 2023] [ADIOS2 EXCEPTION] <Helper> <adiosType> <CheckForNullPtr> : found null pointer in call to IO::VariableType
```
on Perlmutter.

PR #4377, opened 2023-10-16
Updated proton-boron cross section 

PhysicsDan:
Hello,

## To do

- [ ] Add ability to choose between Nevins and Tentori cross section fits

I have updated the proton-Boron cross section to use the analytical fit described in [A. Tentori & F. Belloni, Nuclear Fusion, 63, 086001 (2023)](https://doi.org/10.1088/1741-4326/acda4b) for the [Sikora & Wellar cross section dataset](https://doi.org/10.1007/s10894-016-0069-y). This fit is valid for E<=9.76 MeV.

To ensure the cross section properly decreases at higher energies, the same power law previously used for E>3.5MeV is used for E>9.76 MeV.

Below is a plot of the analytical cross-section, up to an energy of 50 MeV, for both Tentori (2023) included in this pull request and Nevins (2000) which is currently implemented in WarpX. Also highlighted is the region calculated using a simple power law (for E>3.5 MeV when using Nevins fit and for E>9.76 MeV when using the Tentori fit). The cross section above 9.76 MeV is basically unchanged by this PR.
Tentori2023

Thanks,
Daniel


RemiLehe:
Thanks again for this PR!

It looks like the tests are failing because of the comparison with checksum results.
(See https://warpx.readthedocs.io/en/latest/developers/checksum.html for more explanation.)
Could you reset the corresponding checksum files as described here:
https://warpx.readthedocs.io/en/latest/developers/checksum.html#reset-a-benchmark-with-new-values-that-you-know-are-correct

PhysicsDan:
No problem. I updated the checksums in 38ad9473a3b87dedf4b5510df912c14a083148ef. It looks like all tests bar `cartesian3d` were successful. However, it has failed with the message.
```
ERROR! Cmake configuration failed for Langmuir_multi_psatd_Vay_deposition_nodal 
Check /tmp/ci/rt-WarpX/WarpX-tests/2023-10-18/Langmuir_multi_psatd_Vay_deposition_nodal.cmake.log for more information.
```
Which should be unrelated to anything that has been changed in the PR. Is there an option to rerun without an extra empty commit?

Examples/Tests/nuclear_fusion/analysis_proton_boron_fusion.py
```diff
@@ -285,54 +285,40 @@ def astrophysical_factor_midE(E):
...
-    E0 = 581.3
-    E1 = 1083.
-    E2 = 2405.
-    E3 = 3344.
-    dE0 = 85.7
-    dE1 = 234.
-    dE2 = 138.
-    dE3 = 309.
-    B = 4.38
+    ## 668 keV < E < 3500 keV described in equation (5) of A. Tentori and F. Belloni,
```
NeilZaim:
```suggestion
    ## 668 keV < E < 9760 keV described in equation (5) of A. Tentori and F. Belloni,
```

NeilZaim (approved):
Thanks for this Pull Request! I am not an expert to judge whether the new cross-sections (the Sikora data) are better but the code looks good to me. And it's good that we use a published fit that extends to ~10 MeV.

PhysicsDan:
> Thanks for this Pull Request! I am not an expert to judge whether the new cross-sections (the Sikora data) are better but the code looks good to me. And it's good that we use a published fit that extends to ~10 MeV.

Yes you're correct, we are making the assumption that the newer measurements are more 'correct'.

More importantly you have just highlighted a good point. This fit only extends to ~10 MeV. This was not something I had considered. In that case I should "re-add" the exponential fit from Buck et al., Nuclear Physics A, 398(2), 189-202 (1983) for energies >9.76 MeV to ensure the cross section correctly decreases.

Examples/Tests/nuclear_fusion/analysis_proton_boron_fusion.py
```diff
@@ -316,20 +316,20 @@ def astrophysical_factor(E):
...
     return cross_section_start_fit*(E/E_start_fit)**slope_fit
 
 def pb_cross_section(E):
     ## E is in keV
-    ## Returns cross section in b using the fits described in W.M. Nevins and R. Swain,
-    ## Nuclear Fusion, 40, 865 (2000) for E < 3.5 MeV and a power law fit of the data presented in
-    ## Buck et al., Nuclear Physics A, 398(2), 189-202 (1983) for E > 3.5 MeV.
+    ## Returns cross section in b using the fits described in A. Tentori and F. Belloni,
+    ## Nucl. Fusion, 63, 086001 (2023) for E < 9.76 MeV otherwise returns a power law fit
+    ## decribed in Buck et al., Nuclear Physics A, 398(2), 189-202 (1983)
```
NeilZaim:
```suggestion
    ## of the data in Buck et al., Nuclear Physics A, 398(2), 189-202 (1983)
```

The fit itself is not described in the paper, it's just a fit of the data I had done when implementing the pB fusion module.

NeilZaim:
Ah right, I did not see that the new fit didn't include the highest energy Buck datapoints. In which case I agree with your last change. I'm assuming you made it so that Tentori and Buck give the same cross-section at 9.76 MeV?

PhysicsDan:
Thanks for your help. Yes the power law for E > 9.76 MeV and the from Tentori return the same value at 9.76 MeV. For reference I have extended the figure in the initial comment to 50 MeV. Above 9.760 MeV the cross section we be basically unchanged with this pull request.

ax3l (approved):
LGTM. You could also have kept both models in there, which might be better for reproducibility/benchmarks.

But if there is consensus that the newer data is better then let's ship it :+1:

ax3l:
I noticed our parameter https://warpx.readthedocs.io/en/latest/usage/parameters.html and also theory section are a bit sparse on the models we use for fusion cross-sections.

Feel free to add a follow-up PR that links your figures @PhysicsDan into the user-facing manual. (Please use external links to figures in the GitHub PR description instead of adding them in Git, which keeps `git clone` times of WarpX small.)

Maybe as a subsection of https://warpx.readthedocs.io/en/latest/theory/collisions.html or as a new section in it? :)

PhysicsDan:
Hi @ax3l - That is a good idea to keep both cross-sections in and include some option in the input. This is probably the correct way to proceed. I am busy for a month or so with experiments but can have a look at this at the end of December. I am also happy to add some details to the documentation at the same time.

Cheers,
Daniel

roelof-groenewald:
I don't think we need to worry about maintaining support for the Nevins cross-section since the consensus in the community is that the Tentori one is more accurate (this is based on reports from the recent WANDA meeting on fusion research needs).

I'll sort out the conflicts with checksum values so we can merge this in.

Thanks for the contribution @PhysicsDan!

roelof-groenewald (approved):

PhysicsDan:
Hi @roelof-groenewald, Thanks for merging this and fixing the checksums. I have been preoccupied with experiments. I agree with the decision; the data the Nevins cross-section is based on is a bit outdated.

PR #4379, opened 2023-10-16
Fix: CMake `Python_EXECUTABLE`

ax3l:
Replace the hard-coded `python3` with the found/set Python executable of the current build environment.


roelof-groenewald (approved):
I'm guessing there is a very particular situation in which this is needed, but the changes look fine.

ax3l:
Thanks! Yes, in package managers this is important. E.g., https://github.com/conda-forge/warpx-feedstock/pull/72

PR #4380, opened 2023-10-17
Fix: Fails in Python Callbacks Abort

ax3l:
Make sure that errors in Python callbacks abort the simulation. Before, MPI-parallel simulations would hang.
Fix #4375

Tested with this input file: [PICMI_inputs_3d.py.txt](https://github.com/ECP-WarpX/WarpX/files/12923292/PICMI_inputs_3d.py.txt)
```console
mpirun -n 4 Examples/Physics_applications/laser_acceleration/PICMI_inputs_3d.py
```
```
STEP 1 starts ...
Python callback 'beforestep' failed!
AttributeError: module 'mpi4py.MPI' has no attribute 'oopsi'

At:
  /home/axel/src/warpx/Examples/Physics_applications/laser_acceleration/PICMI_inputs_3d.py(169): mycallback
  /home/axel/micromamba/envs/warpx-openmp-dev/lib/python3.11/site-packages/pywarpx/callbacks.py(247): callfuncsinlist
  /home/axel/micromamba/envs/warpx-openmp-dev/lib/python3.11/site-packages/pywarpx/callbacks.py(98): __call__
  /home/axel/micromamba/envs/warpx-openmp-dev/lib/python3.11/site-packages/pywarpx/WarpX.py(101): evolve
  /home/axel/micromamba/envs/warpx-openmp-dev/lib/python3.11/site-packages/pywarpx/picmi.py(1930): step
  /home/axel/src/warpx/Examples/Physics_applications/laser_acceleration/PICMI_inputs_3d.py(175): <module>
```
```
# simulation returns
```

PR #4381, opened 2023-10-17
Python: Rename Callback Files

ax3l:
Renamed after thinning out everything else.

- [x] rebase & merge after the bug fix in #4380 is in

### DIFF
--- a/Source/Python/callbacks.cpp
+++ b/Source/Python/callbacks.cpp
@@ -2,11 +2,11 @@
  *
  * This file is part of WarpX.
  *
- * Authors: David Grote, Maxence Thevenet, Weiqun Zhang, Roelof Groenewald
+ * Authors: David Grote, Maxence Thevenet, Weiqun Zhang, Roelof Groenewald, Axel Huebl
  *
  * License: BSD-3-Clause-LBNL
  */
-#include "WarpX_py.H"
+#include "callbacks.H"
 
 #include <cstdlib>
 #include <exception>
--- a/Source/Evolve/WarpXEvolve.cpp
+++ b/Source/Evolve/WarpXEvolve.cpp
@@ -27,7 +27,7 @@
 #include "Fluids/MultiFluidContainer.H"
 #include "Fluids/WarpXFluidContainer.H"
 #include "Particles/ParticleBoundaryBuffer.H"
-#include "Python/WarpX_py.H"
+#include "Python/callbacks.H"
 #include "Utils/TextMsg.H"
 #include "Utils/WarpXAlgorithmSelection.H"
 #include "Utils/WarpXUtil.H"
--- a/Source/FieldSolver/ElectrostaticSolver.cpp
+++ b/Source/FieldSolver/ElectrostaticSolver.cpp
@@ -12,7 +12,7 @@
 #include "Parallelization/GuardCellManager.H"
 #include "Particles/MultiParticleContainer.H"
 #include "Particles/WarpXParticleContainer.H"
-#include "Python/WarpX_py.H"
+#include "Python/callbacks.H"
 #include "Utils/Parser/ParserUtils.H"
 #include "Utils/WarpXAlgorithmSelection.H"
 #include "Utils/WarpXConst.H"
--- a/Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
+++ b/Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
@@ -10,7 +10,7 @@
 #include "Parallelization/GuardCellManager.H"
 #include "Particles/MultiParticleContainer.H"
 #include "Particles/WarpXParticleContainer.H"
-#include "Python/WarpX_py.H"
+#include "Python/callbacks.H"
 #include "Utils/WarpXAlgorithmSelection.H"
 #include "Utils/WarpXConst.H"
 #include "Utils/TextMsg.H"
--- a/Source/Initialization/WarpXInitData.cpp
+++ b/Source/Initialization/WarpXInitData.cpp
@@ -29,7 +29,7 @@
 #include "Utils/WarpXConst.H"
 #include "Utils/WarpXProfilerWrapper.H"
 #include "Utils/WarpXUtil.H"
-#include "Python/WarpX_py.H"
+#include "Python/callbacks.H"
 
 #include <ablastr/parallelization/MPIInitHelpers.H>
 #include <ablastr/utils/Communication.H>
--- a/Source/Python/CMakeLists.txt
+++ b/Source/Python/CMakeLists.txt
@@ -7,7 +7,7 @@ foreach(D IN LISTS WarpX_DIMS)
     target_sources(lib_${SD}
       PRIVATE
         # callback hooks
-        WarpX_py.cpp
+        callbacks.cpp
     )
     if(WarpX_PYTHON)
         target_sources(pyWarpX_${SD}
--- a/Source/Python/Make.package
+++ b/Source/Python/Make.package
@@ -1,3 +1,3 @@
-CEXE_sources += WarpX_py.cpp
+CEXE_sources += callbacks.cpp
 
 VPATH_LOCATIONS   += $(WARPX_HOME)/Source/Python
--- a/Source/Python/callbacks.H
+++ b/Source/Python/callbacks.H
@@ -2,12 +2,12 @@
  *
  * This file is part of WarpX.
  *
- * Authors: David Grote, Maxence Thevenet, Weiqun Zhang, Roelof Groenewald
+ * Authors: David Grote, Maxence Thevenet, Weiqun Zhang, Roelof Groenewald, Axel Huebl
  *
  * License: BSD-3-Clause-LBNL
  */
-#ifndef WARPX_PY_H_
-#define WARPX_PY_H_
+#ifndef WARPX_PY_CALLBACKS_H_
+#define WARPX_PY_CALLBACKS_H_
 
 #include "Utils/export.H"
 #include "Utils/WarpXProfilerWrapper.H"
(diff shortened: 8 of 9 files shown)


Source/Python/callbacks.cpp
```diff
@@ -0,0 +1,52 @@
...
+ * This file is part of WarpX.
+ *
+ * Authors: David Grote, Maxence Thevenet, Weiqun Zhang, Roelof Groenewald, Axel Huebl
+ *
+ * License: BSD-3-Clause-LBNL
+ */
+#include "callbacks.H"
+
+#include <cstdlib>
+#include <exception>
```
ax3l: for `std::cerr`:
```suggestion
#include <cstdlib>
#include <exception>
#include <iostream>
```

Source/Python/callbacks.cpp
```diff
@@ -0,0 +1,52 @@
...
+            std::cerr << "Python callback '" << name << "' failed!" << std::endl;
+            std::cerr << e.what() << std::endl;
+            std::exit(3);  // note: NOT amrex::Abort(), to avoid hangs with MPI
+
+            // future note:
+            // if we want to rethrow/raise exceptions from Python callbacks through here (C++) and
+            // back the managing Python interpreter, we first need to discard and clear
+            // out the Python error in py::error_already_set. Otherwise, MPI-runs will hang
+            // (and Python will be in continued error state).
+            // https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html#handling-unraisable-exceptions
```
ax3l: I played a bit more with this and the result is not trivial:
Imagine we install into `installbeforestep` and a Python exception is raised (for one or even all ranks). If we rethrow it to Python, we will not finish the step at all. If only some ranks raise, we might deadlock in (MPI) collectives that follow the callback.

The only other potentially useful option I see is to ignore Python errors (w/o rethrow) if we want to. That way, we finish the C++ code that is after the callback.

ax3l:
```diff
diff --git a/Examples/Physics_applications/laser_acceleration/PICMI_inputs_3d.py b/Examples/Physics_applications/laser_acceleration/PICMI_inputs_3d.py
index ac3398e43..b044a8801 100755
--- a/Examples/Physics_applications/laser_acceleration/PICMI_inputs_3d.py
+++ b/Examples/Physics_applications/laser_acceleration/PICMI_inputs_3d.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python3
 
 from pywarpx import picmi
+from pywarpx.callbacks import installbeforestep
+from mpi4py import MPI as mpi
 
 # Physical constants
 c = picmi.constants.c
@@ -159,5 +161,18 @@ sim.write_input_file(file_name = 'inputs_3d_picmi')
 sim.initialize_inputs()
 sim.initialize_warpx()
 
+def mycallback():
+    comm = mpi.COMM_WORLD
+    rank = comm.Get_rank()
+    print(f"I am called ({rank})!")
+    if rank == 0:
+        mpi.oopsi
+
+# callback
+installbeforestep(mycallback)
+
 # Advance simulation until last time step
-sim.step(max_steps)
+try:
+    sim.step(max_steps)
+except RuntimeError:
+    print("ahhh")
diff --git a/Source/Python/callbacks.H b/Source/Python/callbacks.H
index f0c98caf6..fe8c247c5 100644
--- a/Source/Python/callbacks.H
+++ b/Source/Python/callbacks.H
@@ -25,6 +25,7 @@
  * list of currently supported callback names.
 */
 extern WARPX_EXPORT std::map< std::string, std::function<void()> > warpx_callback_py_map;
+extern WARPX_EXPORT std::function<void(std::function<void()>&, std::string const &)> warpx_callback_exec_handler;
 
 /**
  * \brief Function to install the given name and function in warpx_callback_py_map
@@ -39,7 +40,7 @@ bool IsPythonCallbackInstalled ( std::string name );
 /**
  * \brief Function to look for and execute Python callbacks
  */
-void ExecutePythonCallback ( std::string name );
+void ExecutePythonCallback ( std::string const & name );
 
 /**
  * \brief Function to clear the given callback name from warpx_callback_py_map
diff --git a/Source/Python/callbacks.cpp b/Source/Python/callbacks.cpp
index a163f4ac1..f809a7e16 100644
--- a/Source/Python/callbacks.cpp
+++ b/Source/Python/callbacks.cpp
@@ -10,9 +10,31 @@
 
 #include <cstdlib>
 #include <exception>
+#include <iostream>
 
 
 std::map< std::string, std::function<void()> > warpx_callback_py_map;
+std::function<void(std::function<void()>&, std::string const &)> warpx_callback_exec_handler;
+
+namespace {
+    void default_callback_exec_handler (std::function<void()>& callback, std::string const & name)
+    {
+        try {
+            callback();
+        } catch (std::exception &e) {
+            std::cerr << "Python callback '" << name << "' failed!" << std::endl;
+            std::cerr << e.what() << std::endl;
+            std::exit(3);  // note: NOT amrex::Abort(), to avoid hangs with MPI
+
+            // future note:
+            // if we want to rethrow/raise exceptions from Python callbacks through here (C++) and
+            // back the managing Python interpreter, we first need to discard and clear
+            // out the Python error in py::error_already_set. Otherwise, MPI-runs will hang
+            // (and Python will be in continued error state).
+            // https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html#handling-unraisable-exceptions
+        }
+    }
+}
 
 void InstallPythonCallback ( std::string name, std::function<void()> callback )
 {
@@ -25,24 +47,14 @@ bool IsPythonCallbackInstalled ( std::string name )
 }
 
 // Execute Python callbacks of the type given by the input string
-void ExecutePythonCallback ( std::string name )
+void ExecutePythonCallback ( std::string const & name )
 {
     if ( IsPythonCallbackInstalled(name) ) {
         WARPX_PROFILE("warpx_py_" + name);
-        try {
-            warpx_callback_py_map[name]();
-        } catch (std::exception &e) {
-            std::cerr << "Python callback '" << name << "' failed!" << std::endl;
-            std::cerr << e.what() << std::endl;
-            std::exit(3);  // note: NOT amrex::Abort(), to avoid hangs with MPI
-
-            // future note:
-            // if we want to rethrow/raise exceptions from Python callbacks through here (C++) and
-            // back the managing Python interpreter, we first need to discard and clear
-            // out the Python error in py::error_already_set. Otherwise, MPI-runs will hang
-            // (and Python will be in continued error state).
-            // https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html#handling-unraisable-exceptions
-        }
+        if (warpx_callback_exec_handler)
+            warpx_callback_exec_handler(warpx_callback_py_map[name], name);
+        else
+            default_callback_exec_handler(warpx_callback_py_map[name], name);
     }
 }
 
diff --git a/Source/Python/pyWarpX.cpp b/Source/Python/pyWarpX.cpp
index 26f4c7750..405a64dbf 100644
--- a/Source/Python/pyWarpX.cpp
+++ b/Source/Python/pyWarpX.cpp
@@ -10,6 +10,10 @@
 #include <Utils/WarpXUtil.H>  // todo: move to its own Python/Utils.cpp
 #include <Utils/WarpXVersion.H>
 #include <Initialization/WarpXAMReXInit.H>
+#include <Python/callbacks.H>
+
+#include <exception>
+#include <iostream>
 
 #define STRINGIFY(x) #x
 #define MACRO_STRINGIFY(x) STRINGIFY(x)
@@ -28,6 +32,39 @@
 
 //using namespace warpx;
 
+namespace {
+    void pybind_callback_exec_handler (std::function<void()>& callback, std::string const & name)
+    {
+        try {
+            callback();
+        } catch (py::error_already_set &eas) {
+            // if we want to rethrow/raise exceptions from Python callbacks through here (C++) and
+            // back the managing Python interpreter, we first need to discard and clear
+            // out the Python error in py::error_already_set. Otherwise, MPI-runs will hang
+            // (and Python will be in continued error state).
+            // https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html#handling-unraisable-exceptions
+
+            // Python exceptions: print
+            std::cerr << "Python callback '" << name << "' failed!" << std::endl;
+            std::cerr << eas.what() << std::endl;
+
+            // Discard the Python error using Python APIs
+            eas.discard_as_unraisable(__func__);
+
+            // raise as a regular runtime error, which can be caught from Python
+            //   downside: C++ code that is written after the callback hook will not be executed
+            // throw std::runtime_error("Python callback '" + name + "' failed!");
+
+            // do NOT abort but return to Python
+        } catch (std::exception &e) {
+            // C++ and all other exceptions: print and discard
+            std::cerr << "Python callback '" << name << "' failed!" << std::endl;
+            std::cerr << e.what() << std::endl;
+
+            // do NOT abort but return to Python
+        }
+    }
+}
 
 // forward declarations of exposed classes
 void init_BoundaryBufferParIter (py::module&);
@@ -58,6 +95,8 @@ PYBIND11_MODULE(PYWARPX_MODULE_NAME, m) {
                WarpX
     )pbdoc";
 
+    warpx_callback_exec_handler = pybind_callback_exec_handler;
+
     // note: order from parent to child classes
     init_PinnedMemoryParticleContainer(m);
     init_WarpXParticleContainer(m);
```

roelof-groenewald:
Looks like just the corresponding update to `Source/Python/Make.package` is still needed and this should be good to go.

ax3l:
The only thing I do not like about this yet: this should not be placed in `Python/` but the general sources.
The callbacks are general C++ hooks and compiled with the core library, not the pybind module library (i.e., one cannot use pybind11 in this file).

roelof-groenewald:
> The only thing I do not like about this yet: this should not be placed in `Python/` but the general sources. The callbacks are general C++ hooks and compiled with the core library, not the pybind module library (i.e., one cannot use pybind11 in this file).

That makes sense to me too. So should we wait to merge this until the files can be moved out of `Python`, or are you thinking that will be a follow-up PR?

ax3l:
Let's merge it for now, I think. Will be a small change to move it out and maybe change some doc strings.

roelof-groenewald (approved):

PR #4383, opened 2023-10-17
[WIP] Add PICMI interface for BoundaryScrapingDiagnostic

RemiLehe:


Python/pywarpx/picmi.py
```diff
@@ -2315,6 +2315,120 @@ def initialize_inputs(self):
...
+
+    warpx_uniform_stride: integer, optional
+        Stride to down select to the particles to include in the diagnostic
+
+    warpx_plot_filter_function: string, optional
+        Analytic expression to down select the particles to in the diagnostic
+    """
+    def init(self, kw):
+
+        self.format = kw.pop('warpx_format', 'plotfile')
```
RemiLehe, line 2349: These parts should be in the base class, as they are used across many diagnostics.

Python/pywarpx/picmi.py
```diff
@@ -2315,6 +2315,120 @@ def initialize_inputs(self):
...
+    def initialize_inputs(self):
+
+        self.add_diagnostic()
+
+        self.diagnostic.diag_type = 'BoundaryScraping'
+        self.diagnostic.format = self.format
+        self.diagnostic.openpmd_backend = self.openpmd_backend
+        self.diagnostic.file_min_digits = self.file_min_digits
+        self.diagnostic.intervals = self.period
+        self.diagnostic.particles_at_eb = 1
```
RemiLehe, line 2377: This should be in a parent PICMI class

Python/pywarpx/picmi.py
```diff
@@ -2315,6 +2315,120 @@ def initialize_inputs(self):
...
+        self.set_write_dir()
+
+        # --- Use a set to ensure that fields don't get repeated.
+        variables = set()
+
```
RemiLehe, line 2382: The code below should be in the WarpXDiagnostic parent class, since it is used across many diagnostics

RemiLehe:
This is superseded by PR #5109 ("Add scraping boundary diagnostics with PICMI"), which introduced the`ParticleBoundaryScrapingDiagnostic` class.

PR #4386, opened 2023-10-20
GetAndSetPosition: General Particle PIdx

ax3l:
Generalize the `SetParticlePosition` & `GetParticlePosition` classes to work with arbitrary particle index enumerators. That way, we can reuse them between different particle containers.

Follow-up to #4057 making the `FieldProbe` implementation way more robust to changes.
https://www.youtube.com/watch?v=QRVExJZKIT8


roelof-groenewald (approved):
Nice! Very elegant solution!

PR #4389, opened 2023-10-24
Machines: HDF5 & ADIOS2 Tools in PATH

ax3l:
Ensure tools such as `bpls`, `bpdump` and `h5ls` are available to users if we self-compile them.


dpgrote (approved):
This looks good. Thanks!

bzdjordje (commented):
Looks good!

PR #4390, opened 2023-10-26
Fix External AMReX Builds

ax3l:
- print version
- make AMReX external if pyAMReX is external
- new AMReX 23.11+ installed CMake module paths for CMake scripts
- `TINYP` is not really required for external packages and thus should not fail if configured not to be used

### DIFF
--- a/cmake/dependencies/AMReX.cmake
+++ b/cmake/dependencies/AMReX.cmake
@@ -1,4 +1,10 @@
 macro(find_amrex)
+    # if pyAMReX is external, AMReX must be as well
+    if(DEFINED WarpX_pyamrex_internal AND NOT WarpX_pyamrex_internal)
+        set(WarpX_amrex_internal OFF CACHE BOOL
+            "Download & build AMReX" FORCE)
+    endif()
+
     if(WarpX_amrex_src)
         message(STATUS "Compiling local AMReX ...")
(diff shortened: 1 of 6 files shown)


cmake/dependencies/AMReX.cmake
```diff
@@ -1,4 +1,10 @@
 macro(find_amrex)
+    # if pyAMReX is external, AMReX must be as well
+    if(DEFINED WarpX_pyamrex_internal AND NOT WarpX_pyamrex_internal)
```
dpgrote, line 3: Is this a typo? Shouldn't this be
```
if(DEFINED WarpX_amrex_internal AND NOT WarpX_pyamrex_internal)
```

ax3l: No, that is intentional:
`if(NOT WarpX_pyamrex_internal)` is true if the variable is truly false-valued _or_ undefined. Thus, I first check if its defined, then check if the value is negative.

Since we include the pyAMReX logic that defines `WarpX_pyamrex_internal` after including `AMReX.cmake`, this slightly awkward logic is needed here to catch the first pass.

Ways that it were defined is either if its set from CLI or a superbuild (e.g., ImpactX) or by reconfiguring an existing build directory (e.g., with ccmake). But the very first pass on a fresh build dir of WarpX itself, it is undefined when the code gets here.

PR #4391, opened 2023-10-26
Doc: Python Backtrace in CXX Segfaults

ax3l:
Document the workflow to find out which last Python line was called if the code crashes in the CXX part for PICMI runs.


dpgrote (approved):
This looks good. I didn't know about faulthandler - it looks useful.

PR #4392, opened 2023-10-27
Python 3.12: Refactor out Distutils

### LINKED ISSUE #2821: Refactor out Distutils

ax3l:
Fix #2821 for Python 3.12

Refs.:
- https://peps.python.org/pep-0632/#migration-advice
- https://github.com/AMReX-Codes/pyamrex/pull/180

### DIFF
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -18,10 +18,6 @@ jobs:
     env:
       CXXFLAGS: "-Werror"
       CMAKE_GENERATOR: Ninja
-      # setuptools/mp4py work-around, see
-      #   https://github.com/mpi4py/mpi4py/pull/159
-      #   https://github.com/mpi4py/mpi4py/issues/157#issuecomment-1001022274
-      SETUPTOOLS_USE_DISTUTILS: stdlib
     steps:
     - uses: actions/checkout@v3
     - uses: actions/setup-python@v4
@@ -75,7 +71,8 @@ jobs:
           -DAMReX_CUDA_ERROR_CAPTURE_THIS=ON
         cmake --build build_sp -j 2
 
-        python3 -m pip install --upgrade pip setuptools wheel
+        python3 -m pip install --upgrade pip
+        python3 -m pip install --upgrade build packaging setuptools wheel
         export WARPX_MPI=ON
         export PYWARPX_LIB_DIR=$PWD/build_sp/lib/site-packages/pywarpx/
         python3 -m pip wheel .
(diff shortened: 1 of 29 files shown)


setup.py
```diff
@@ -9,6 +6,7 @@
 import sys
 
 from setuptools import Extension, setup
+import setuptools.command.build
```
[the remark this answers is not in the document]
ax3l: outdated

.github/workflows/cuda.yml
```diff
@@ -18,16 +18,12 @@ jobs:
...
-      #   https://github.com/mpi4py/mpi4py/pull/159
-      #   https://github.com/mpi4py/mpi4py/issues/157#issuecomment-1001022274
-      SETUPTOOLS_USE_DISTUTILS: stdlib
     steps:
     - uses: actions/checkout@v3
     - uses: actions/setup-python@v4
       name: Install Python
       with:
-        python-version: '3.10'
+        python-version: '3.x'
```
dpgrote, line 26: What does this do?

ax3l: Unconstrain the upper Python version that gets installed. I constrained it in https://github.com/ECP-WarpX/WarpX/commit/ed13486ac0495b049958989022a9c76e8ef8f042 to unbreak CI by avoiding Python 3.12 for a few days.

I want it unconstrained again, so I see if there are breaking changes once new Python versions are released.

Ref.: https://github.com/actions/setup-python#setup-python

Tools/machines/lassen-llnl/install_v100_ml.sh
```diff
@@ -64,6 +64,7 @@ rm -rf build
 cd -
 
 # optional: optimas dependencies (based on libEnsemble & ax->botorch->gpytorch->pytorch)
+TODO: scikit-learn needs a BLAS hint
```
roelof-groenewald: Is this okay to not be a comment?
```suggestion
# TODO: scikit-learn needs a BLAS hint
```

roelof-groenewald (approved):
Looks good. Just one minor comment.

Tools/machines/lassen-llnl/install_v100_ml.sh
ax3l: Oh no, added accidentally

PR #4393, opened 2023-10-28
Set `amrex.omp_threads = "nosmt"`

ax3l:
Improve the performance on Intel and AMD CPUs by default: Avoid oversubscribing the physical number of CPU threads, as is default in OpenMP.

https://amrex-codes.github.io/amrex/docs_html/InputsComputeBackends.html

- [x] requires https://github.com/AMReX-Codes/amrex/pull/3607 and https://github.com/AMReX-Codes/amrex/pull/3647
- [x] test again to check that the overwrite here worked


ax3l:
Huh, was `.azure-pipelines.yml`'s `OMP_NUM_THREADS=1` never propagated to our tests...? :eyes: 
That would explain a lot o.0

Update: ah no, little bug https://github.com/AMReX-Codes/amrex/pull/3647

WeiqunZhang (approved):

PR #4397, opened 2023-10-30
Allow negative mean velocity with NFluxPerCell

dpgrote:
This updates the rejection method used with NFluxPerCell with a drifting Maxwellian allowing a negative drift velocity. Modifications were made to the formulas implemented in PR #3514.

The CI test was updated adding cases with negative drift velocity.

### DIFF
--- a/Source/Initialization/InjectorMomentum.H
+++ b/Source/Initialization/InjectorMomentum.H
@@ -106,50 +106,15 @@ namespace {
 
         if (u_th == 0._rt) {
             u = u_m; // Trivial case ; avoids division by 0 in the rest of the code below
-        } else if (u_m < -u_th*0.6) {
-            // Mean velocity is greater than negative thermal velocity
-            // Use the distribution exp(-(u-u_m+u_th**2/u_m)**2/(2*u_th**2)) as an approximation
+        } else if (std::abs(u_m) < 0.6*u_th) {
+            // Mean velocity magnitude is less than thermal velocity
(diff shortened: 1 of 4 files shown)


Source/Initialization/InjectorMomentum.H
```diff
@@ -106,35 +106,36 @@ namespace {
 
         if (u_th == 0._rt) {
             u = u_m; // Trivial case ; avoids division by 0 in the rest of the code below
-        } else if (u_m < 0.6*u_th) {
-            // Mean velocity is lower than thermal velocity
-            // Use the distribution u*exp(-u**2*(1-u_m/u_th)/(2*u_th**2)) as an approximation
+        } else if (std::abs(u_m) < 0.6*u_th) {
```
ax3l: Since `std::abs(u_m)` is used so often, it might be wort storing it in a constant `abs_u_m`?

ax3l:
Thank you for this! Just a minor comment :)

Source/Initialization/InjectorMomentum.H
dpgrote: I can make the change, but is this something that the compiler would do anyway, common expression elimination?

ax3l (approved):

Source/Initialization/InjectorMomentum.H
ax3l: That's a good question, one would have to test. But it's not too weird to make it a var here.

PR #4398, opened 2023-10-30
Docs: Add Particles-in-PML Paper

ax3l:
Add paper: "Absorption of charged particles in perfectly matched layers by optimal damping of the deposited current" (2022).


lucafedeli88 (approved):
Thanks for this PR!

PR #107, opened 2019-04-29
Runtime boosted frame particles

atmyers:
This removes the need to recompile when using the boosted frame particles. 

I tested this using the `Examples/Physics_applications/laser_acceleration/inputs.2d.boost` inputs file, and the particle output seems to be the same. 

Note that this PR also includes some bug fixes. In particular, the `old` particle positions and momenta were not being initialized properly for some of the particle types. Because of this, the answers were slightly different in the 6th decimal place or so. 

If someone else could test this using a different inputs file, that would be great.

PR #4400, opened 2023-11-01
Add gaussian_parse_momentum_function

dpgrote:
This adds a plasma initialization where the mean momentum and the standard deviation are specified with parsed expressions, functions of position. The new momentum distribution is called `gaussian_parse_momentum_function`.

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -908,9 +908,12 @@ Particle initialization
       ``<species_name>.momentum_function_uy(x,y,z)`` and ``<species_name>.momentum_function_uz(x,y,z)``,
       which gives the distribution of each component of the momentum as a function of space.
 
-    * ``gaussian_parse_momentum_function``: Gaussian momentum distribution where the mean normalized momentum :math:`u = (u_{x},u_{y},u_{z})=(\gamma v_{x}/c,\gamma v_{y}/c,\gamma v_{z}/c)` and the standard deviation are given by functions of position in the input file.
-      It requires the following arguments which specify the distribution of each component of the mean normalized momentum
-      and standard deviation as a function of space.
+    * ``gaussian_parse_momentum_function``: Gaussian momentum distribution where the mean and the standard deviation are given by functions of position in the input file.
+      Both are assumed to be non-relativistic.
+      The mean is the normalized momentum, :math:`u_m = \gamma v_m/c`.
+      The standard deviation is normalized, :math:`u_th = v_th/c`.
+      For example, this might be `u_th = sqrt(T*q_e/mass)/clight` given the temperature (in eV) and mass.
+      It requires the following arguments:
 
       * ``<species_name>.momentum_function_ux_m(x,y,z)``: mean :math:`u_{x}`
       * ``<species_name>.momentum_function_uy_m(x,y,z)``: mean :math:`u_{y}`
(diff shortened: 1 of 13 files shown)


Examples/Tests/initial_distribution/analysis_distribution.py
```diff
@@ -261,6 +261,34 @@
...
+#============================================
+
+# load data
+bin_value_ux, bin_data_ux = read_reduced_diags_histogram("h9x.txt")[2:]
+bin_value_uy, bin_data_uy = read_reduced_diags_histogram("h9y.txt")[2:]
+bin_value_uz, bin_data_uz = read_reduced_diags_histogram("h9z.txt")[2:]
+
+def Gaussian(mean, sigma, u):
+    V = 8.0 # volume in m^3
+    n = 1.0e21 # number density in 1/m^3
```
aeriforme: Could these be parsed from the input file using `Tools/Parser/input_file_parser.py?`

Docs/source/usage/parameters.rst
```diff
@@ -908,6 +908,17 @@ Particle initialization
       ``<species_name>.momentum_function_uy(x,y,z)`` and ``<species_name>.momentum_function_uz(x,y,z)``,
       which gives the distribution of each component of the momentum as a function of space.
 
+    * ``gaussian_parse_momentum_function``: Gaussian momentum distribution where the mean momentum :math:`u = (u_{x},u_{y},u_{z})=(\gamma v_{x}/c,\gamma v_{y}/c,\gamma v_{z}/c)` and the standard deviation are given by functions of position in the input file.
```
aeriforme: I would call it the mean _normalized_ momentum.

Docs/source/usage/parameters.rst
```diff
@@ -908,6 +908,17 @@ Particle initialization
...
+      It requires the following arguments which specify the distribution of each component of the mean momentum
```
aeriforme: Same as above: mean _normalized_ momentum.

aeriforme (commented):
Two main questions:
* Are there any issues or caveats for relativistic species? 
* Does the initialization happen in the lab or rest frame?

Examples/Tests/initial_distribution/analysis_distribution.py
dpgrote: Yes, they could, but I did it the same way as the other tests in this file, just making the parameters explicit. I agree that it is better to get the parameters from the input file so that they are written in only one place.

dpgrote:
> Two main questions:
> 
> * Are there any issues or caveats for relativistic species?
> * Does the initialization happen in the lab or rest frame?

The assumption is non-relativistic, since it loads a v_drift + v_gaussian. The input parameters are in the lab frame. I can add a comment in the doc indicating that this is valid only when both `u_m` and `u_th` are non-relativistic.

aeriforme (approved):

Python/pywarpx/picmi.py
```diff
@@ -458,6 +464,20 @@ def initialize_inputs(self, species_number, layout, species, density_scale):
 
 
 class AnalyticDistribution(picmistandard.PICMI_AnalyticDistribution, DensityDistributionBase):
+    """
```
dpgrote, line 467: Note the once PR #107 from PICMI is merged, this code can be removed since the WarpX specific argument is no longer needed.

lucafedeli88 (approved):
Looks good to me! Thanks for this!

PR #4404, opened 2023-11-03
Make the static variable "authors" of the WarpX class a private member of the WarpX class 

lucafedeli88:
This PR is part of the effort to reduce the use of static functions and variables in WarpX.


ax3l (approved):
Thank you! :)

PR #4405, opened 2023-11-03
Add external current handling to Ohm's law solver

roelof-groenewald:
It is useful to be able to specify an external current in order to simulate magnetic control of a plasma as well as other interesting problems such as Alfven wave heating.

### DIFF
--- a/Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
+++ b/Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
@@ -263,12 +263,6 @@ void HybridPICModel::GetCurrentExternal (
     auto t = warpx.gett_new(lev);
 
     auto dx_lev = warpx.Geom(lev).CellSizeArray();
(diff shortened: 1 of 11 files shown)


Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -42,6 +42,11 @@ void HybridPICModel::ReadParameters ()
 
     // convert electron temperature from eV to J
     m_elec_temp *= PhysConst::q_e;
+
+    // external currents
+    pp_hybrid.query("Jx_external_grid_function(x,y,z,t)", m_Jx_ext_grid_function);
```
RemiLehe, line 47: Could you add documentation for this on this page: https://warpx.readthedocs.io/en/latest/usage/parameters.html
In particular, could you state there that this is only used for the hybrid model?

roelof-groenewald: Thanks for the comment! I added it to the documentation.

RemiLehe (approved):

PR #4409, opened 2023-11-07
In doGatherShapeN, for RZ use temporaries for Er and Etheta

dpgrote:
This avoids a problem with the gathering of the fields when RZ is used. The routine doGatherShapeN calculated the fields and converted them from (r, theta) to (x, y). Unfortunately, that conversion would affect the values of the fields that might have been calculated before doGatherShapeN was called.

This PR fixes the issue by calculating Er and Etheta in temporaries first and then converting the temporaries, so that the values of Ex and Ey passed in are not modified.

This issue was first noted in PR #4094 where a fix was implemented in Ionization.H. The fix in this PR completely removes the issue.

### DIFF
--- a/Source/Particles/Gather/FieldGather.H
+++ b/Source/Particles/Gather/FieldGather.H
@@ -252,7 +252,7 @@ void doGatherShapeN (const amrex::ParticleReal xp,
             by_arr(lo.x+l_by+iz, 0, 0, 0);
     }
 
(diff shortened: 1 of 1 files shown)


ax3l (approved):
LGTM - @RemiLehe any comments? :)

lucafedeli88 (approved):
Looks good to me!

RemiLehe (approved):

PR #4420, opened 2023-11-14
[WIP] Refactor particle diags for filters

RemiLehe:
The aim of this PR is to perform particle selection in particle functors, for all types of diagnostics (not just BTD), instead of doing particle selection in format-specific files (openPMD, plotfiles).

- All particle diagnostics (i.e. Full, BoundaryScraping, BackTransformed, Checkpoint) now initialize particle buffers in pinned memory. (Previously, only the BackTransformed diagnostics did) 
(`InitializeParticleBuffer` is not a virtual function anymore ; it's general definition in the parent class `Diagnostics` is [here](https://github.com/ECP-WarpX/WarpX/pull/4420/files#diff-f03035e5d437bc992ba6a971c6af5a35dc78e7a2d20f8d01eb5071d84347c5d1R456) ; I removed the definitions in the children classes)

Minor  changes:
- I had to modify `GetParticleContainerFromName` so that it works with laser species too

TODO:

- [ ] PrepareFunctorData: make virtual function
- [ ] Write a functor for the other diagnostic (`GenericParticleDiagFunctor`), templated on the type of particle it operates on: it should select particles and move them to the pinned memory buffer
- [ ] Do not pass WarpXParticleContainer anymore to ParticleDiagnostic
- [ ] Always call getPinnedParticleContainer when flushing ; remove use_pinned_pc argument
- [ ] For `phi` diagnostic: add a function that allocate phi on the particle buffer and gathers it


Source/Particles/MultiParticleContainer.cpp
```diff
@@ -388,11 +388,26 @@ MultiParticleContainer::ReadParameters ()
 WarpXParticleContainer&
 MultiParticleContainer::GetParticleContainerFromName (const std::string& name) const
 {
```
RemiLehe, line 390: This generalizes the function `GetParticleContainerFromName` so that it also works with laser particles

Source/Diagnostics/Diagnostics.cpp
```diff
@@ -452,6 +452,45 @@ Diagnostics::InitData ()
     }
 }
 
+void
+Diagnostics::InitializeParticleBuffer ()
```
RemiLehe, line 456: This function has the combined logic from the old `BTDiagnostic::InitializeParticleBuffer` and `BoundaryScrapingDiagnostics::InitializeParticleBuffer`

PR #4425, opened 2023-11-16
In .github/workflows/macos.yml, clean up pip packages

dpgrote:
This is a bit of a hack but it does seem to fix the failing CI tests for MacOS. The errors that were appearing seem to be related to doing pip installs of packages that are already installed. The fix proposed here is to do a uninstall of all of the pip packages first before doing the later installs. Yes, this is an ugly brute force approach, but at least it seems to work.


dpgrote:
It working initially, but now fails.

PR #4437, opened 2023-11-22
add start/end moving step

jiecaix:
Add 'warpx_start_moving_window_step' and 'warpx_end_moving_window_step' for class CylindricalGrid/Cartesian1DGrid to make python interface could stop the moving window

### DIFF
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -584,7 +584,7 @@ class CylindricalGrid(picmistandard.PICMI_CylindricalGrid):
 
     warpx_start_moving_window_step: int, default=0
         The timestep at which the moving window starts
-        
+
     warpx_end_moving_window_step: int, default=-1
         The timestep at which the moving window ends. If -1, the moving window
     will continue until the end of the simulation.
@@ -688,7 +688,7 @@ class Cartesian1DGrid(picmistandard.PICMI_Cartesian1DGrid):
 
     warpx_start_moving_window_step: int, default=0
         The timestep at which the moving window starts
-        
+
     warpx_end_moving_window_step: int, default=-1
         The timestep at which the moving window ends. If -1, the moving window
     will continue until the end of the simulation.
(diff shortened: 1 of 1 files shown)


Python/pywarpx/picmi.py
```diff
@@ -598,6 +598,9 @@ def init(self, kw):
         self.potential_zmax = kw.pop('warpx_potential_hi_z', None)
         self.reflect_all_velocities = kw.pop('warpx_reflect_all_velocities', None)
 
+        self.warpx_start_moving_window_step = kw.pop('warpx_start_moving_window_step', 0)
```
dpgrote: The default values here should be `None`, as is done for the other parameters. The default values of these parameters (i.e. 0 and -1) are already set in the c++ and shouldn't be duplicated here.

dpgrote:
Thanks for this PR! It looks good, though I have the comment on the default values. Also, can you add the two new parameters to the documentation by including them in the list of parameters in the triple quoted doc strings?

Python/pywarpx/picmi.py
jiecaix: Thanks for your comment, and i had change them to be None. Also, i add the description of two new parameters to the triple quoted doc strings.

Python/pywarpx/picmi.py
```diff
@@ -581,6 +581,13 @@ class CylindricalGrid(picmistandard.PICMI_CylindricalGrid):
     warpx_reflect_all_velocities: bool default=False
         Whether the sign of all of the particle velocities are changed upon
         reflection on a boundary, or only the velocity normal to the surface
+
+    warpx_start_moving_window_step: int, default=0
+        The timestep at which the moving window starts
+
+    warpx_end_moving_window_step: int, default=-1
+        The timestep at which the moving window ends. If -1, the moving window
+    will continue until the end of the simulation.
```
dpgrote: One last detail - can you fix the indentation?

jiecaix: Thanks for this check, the indentation has been fixed.

dpgrote (approved):
Looks good! Thanks!

PR #4441, opened 2023-11-23
Move external field parameters out of WarpX class

lucafedeli88:
This PR is part of the efforts to clean the WarpX class.

Parameters related to external fields are now stored together in a struct defined in `Source/Initialization/ExternalField.H` . This header file also defines a function to read these parameters (implemented in  `Source/Initialization/ExternalField.cpp`). The WarpX class has a pointer to a struct holding all the parameters related to the external field: `m_p_ext_field_params` .


ax3l:
Attn: #4339

ax3l (approved):
Thank you, that's a nice clean up.

I let @dpgrote finalize the review.

Source/Initialization/ExternalField.cpp
```diff
@@ -0,0 +1,179 @@
...
+        else{
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(s != "parse_e_ext_grid_function",
+                "parse_E_ext_grid_function can be used only for E_ext_grid_init_style");
+        }
+
+        return ExternalFieldType::default_zero;
+    }
+}
+
+std::unique_ptr<ExternalFieldParams> ReadExternalFieldParams(const amrex::ParmParse& pp_warpx)
```
dpgrote: Instead of a separate routine, can this be the constructor of `ExternalFieldParams`?

Source/Initialization/ExternalField.cpp
```diff
@@ -0,0 +1,179 @@
...
+    pp_warpx.query("E_ext_grid_init_style", E_ext_grid_s);
+    p_external_field->E_ext_grid_type = string_to_external_field_type<EMFieldType::E>(E_ext_grid_s);
+
+    //
+    //  Constant external field
+    //
+
+    // if the input string is "constant", the values for the
+    // external grid must be provided in the input.
+    auto v_B = std::vector<amrex::Real>(3);
```
dpgrote, line 78: Is there a reason for the temporary vector, instead of doing
```utils::parser::getArrWithParser(pp_warpx, "B_external_grid", p_external_field->B_external_grid);```

File: Source/Initialization/ExternalField_fwd.H [hunk 3]
```diff
@@ -0,0 +1,20 @@
...
+ *
+ *
+ * This file is part of WarpX.
+ *
+ * License: BSD-3-Clause-LBNL
+ */
+#ifndef EXTERNAL_FIELD_FWD_H_
+#define EXTERNAL_FIELD_FWD_H_
+
+#include <AMReX_Parser.H>
```
dpgrote: Are these includes needed?

dpgrote:
Thanks for the PR! I like the idea, but have the various comments.

Source/Initialization/ExternalField_fwd.H
lucafedeli88: You are totally right! That's definitively a mistake! I will remove them immediately.

Source/Initialization/ExternalField.cpp
lucafedeli88: Yes. The reason is that `getArrWithParser` requires a `std::vector` as an argument, while I want to store the field value into a GpuArray.

Source/Initialization/ExternalField.cpp
lucafedeli88: yes. That's a good idea.

lucafedeli88:
@dpgrote , do the latest changes address your points about this this PR?

Source/Initialization/ExternalField.cpp
```diff
@@ -0,0 +1,175 @@
...
+            return ExternalFieldType::constant;
+        else if ( s == "parse_b_ext_grid_function" || s == "parse_e_ext_grid_function")
+            return ExternalFieldType::parse_ext_grid_function;
+        else if ( s == "read_from_file")
+            return ExternalFieldType::read_from_file;
+        else
+            WARPX_ABORT_WITH_MESSAGE(
+                "'" + s + "' is an unknown external field type!");
+
+        if constexpr (T == EMFieldType::E){
```
dpgrote, line 29: The logic here doesn't look correct. Since the above `if` blocks call `return`, it will only get here when `s` is an unknown field type. Should these checks go in the `if` block at line 34 above?

Source/Initialization/ExternalField.cpp
```diff
@@ -0,0 +1,175 @@
...
+{
+
+    enum class EMFieldType{E, B};
+
+    template <EMFieldType T>
+    ExternalFieldType string_to_external_field_type(std::string s)
+    {
+        std::transform(s.begin(), s.end(), s.begin(), ::tolower);
+
+        if ( s.empty() || s == "default")
```
dpgrote: It has been suggested as a standard to always use curly braces and not have single line `if` blocks.

lucafedeli88: I am not sure that we follow this rule systematically in WarpX, but it is certainly a good practice! I've added the curly braces.

Source/Initialization/ExternalField.cpp
lucafedeli88: I've moved the check before this block. As you remarked, the logic was wrong and the second if block was ignored.

dpgrote (approved):
Looks good! Thanks!

PR #4444, opened 2023-11-26
Change shiftMF from a WarpX public static function to a public member function of the WarpX class

lucafedeli88:
This PR is part of the effort to simplify the WarpX class.


Source/WarpX.H
```diff
@@ -1137,6 +1132,11 @@ public:
 
     FiniteDifferenceSolver * get_pointer_fdtd_solver_fp (int lev) { return m_fdtd_solver_fp[lev].get(); }
 
+    void shiftMF (amrex::MultiFab& mf, const amrex::Geometry& geometry,
```
ax3l, line 1135: Let's add a Doxygen string?

ax3l (commented):
I do not understand how it simplifies the class yet o.0

It is not static anymore, but still part of the public class interface and still using static member variables.

ax3l:
ping @lucafedeli88 :)

lucafedeli88:
I will close this PR for the moment, since it is not a particularly useful refactoring of the code.

PR #2252, opened 2021-08-30
Interpolate mu to B-location when computing H

RevathiJambunathan:
H = B/mu is computed when taking curl of H to update E in the macroscopic solver.
However, B and mu are not defined at the same location in the grid - for yee , B is on the face-center and mu is at the cell-center.
This PR computes macroparameters at a given location (x,y,z) instead of storing in a multifab with some predefined staggering.

Many thanks to Wei Hou Tan for bringing this to my attention in an offline conversation.

As tested in a previous PR #1016 
5 tests were performed to verify the implementation with on-the-fly computation of the macroparameters.

**Case 1 : Skin depth poor conductor**
input file :
[inputs_sigma1000.txt](https://github.com/ECP-WarpX/WarpX/files/7188883/inputs_sigma1000.txt)
 Variation of skin-depth is as expected.
skindepth_sigma1000

**Case 2  : Perfect conductor**
Input file : 
[inputs_perfectconductor.txt](https://github.com/ECP-WarpX/WarpX/files/7188886/inputs_perfectconductor.txt)
standing wave generated : 
good_conductor

**Case 3 : Varying epsilon** 
epsilon_r = 9 : reflection coefficient = 0.5, transmission coefficient = 0.5
Input file :
[inputs_epsilon9.txt](https://github.com/ECP-WarpX/WarpX/files/7188889/inputs_epsilon9.txt)
epsilon9
 
Case 4 : Varying mu and epsilon:
epsilon_r = 9, mu_r = 9 : transmission coefficient = 1
[inputs_eps9_mu9.txt](https://github.com/ECP-WarpX/WarpX/files/7188893/inputs_eps9_mu9.txt)
eps9_mu9

Case 5 : Varying only mu
mu_r = 40000: reflection coefficient = 1
input file : 
[inputs_mu40000.txt](https://github.com/ECP-WarpX/WarpX/files/7188897/inputs_mu40000.txt)
only_mu

PR #4448, opened 2023-11-27
Avoid divide by `epsilon` in macroscopic algorithms

eebasso:
# Summary

This PR makes the calculation of `alpha` and `beta` in the algorithms of MacroscopicProperties.H more efficient and avoids dividing by the relative permittivity `epsilon`. The check on `epsilon` upon initialization is relaxed to require non-negativity instead of strict positivity.

# Additional Information

The changes allow for the possibility of `epsilon` being zero while reducing the numbers of operations to calculate the coefficients. In particular, the number of division operations gets cut from two to one. In the case of `beta`, an additional multiplication operation is cut. This could be significant since these calculations occur inside the for loops to evolve the E-field.


ax3l (approved):
Awesome :+1:

RevathiJambunathan (approved):
Thanks for this PR! Looks good to me! 
Were you able to test this, @eebasso with any of the inputs in [this PR ](https://github.com/ECP-WarpX/WarpX/pull/2252) or any other inputs?

ax3l:
@eebasso just curious: for which (meta?) material is epsilon=0?

eebasso:
> Thanks for this PR! Looks good to me! Were you able to test this, @eebasso with any of the inputs in [this PR ](https://github.com/ECP-WarpX/WarpX/pull/2252) or any other inputs?

I have not yet tested it. I'll try it soon.

> @eebasso just curious: for which (meta?) material is epsilon=0?

According to "N. Garcia, E. V. Ponizovskaya, John Q. Xiao; Zero permittivity materials: Band gaps at the visible. Appl. Phys. Lett. 18 February 2002; 80 (7): 1120–1122", one possible realization would be a dielectric medium with embedded metallic nanoparticles and nanowires. They state
> "In the limit of long wavelengths, these composites will behave like a homogeneous medium with zero permittivity that will completely reflect electromagnetic waves."

Furthermore, there are materials such as superconductors that can have negative permittivity at radio frequencies.

RevathiJambunathan:
@eebasso would you like to add your test result in the PR description just to double check ? Thank you!

PR #4452, opened 2023-11-28
Contributing style and conventions additions

eebasso:
This PR modifies contributing.rst in the following ways:
- New convention about class definitions: use “class” on same line as class name, e.g., “class MyClass”.
- New convention to always use curly brackets for single statement blocks.
- Reword function definition convention to be more concise and only use one bullet point.

Additionally, documentation.rst is changed to conform to class snake-case naming convention.


ax3l (approved):
Thank you, LGTM :)

PR #4453, opened 2023-11-28
Clang-Tidy Cache

WeiqunZhang:
Use ccache in the clang-tidy CI. This allows us to skip clang-tidy checks on files that are not changed as reported by ccache.

Without cache, the clang-tidy CI takes 2h 30m. With everything in cache, it takes 5m in total and 3m of it is on installing dependency (e.g., blaspp and lapackpp).

### DIFF
--- a/.github/workflows/clang_tidy.yml
+++ b/.github/workflows/clang_tidy.yml
@@ -16,28 +16,26 @@ jobs:
     - name: install dependencies
       run: |
         .github/workflows/dependencies/clang14.sh
-    - name: build WarpX using clang-tidy
+    - name: set up cache
+      uses: actions/cache@v3
+      with:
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
Will push a fix for the isort issue after the clang-tidy CI finishes. (Need to add a blank line. No kidding.)

ax3l (approved):
This is awesome, thank you so much!

ax3l:
@WeiqunZhang do we also need your improved ccache PR logic? (as in https://github.com/AMReX-Codes/pyamrex/pull/188)

WeiqunZhang:
Yes, I think we can improve the cache logic. I can start with say the CUDA CI. Then we can see the difference it can make.

PR #4454, opened 2023-11-29
[WIP] Fix the bug for diag and fieldprobe

jiecaix:
The diag_lo and diag_hi in Diag didn't be calculated correctly, and in this PR it fixed. And fieldprobe will leave the window because x_probe don't move. By the way, when level > 1 at the fieldprobe will get wrong results so i rewritten it and add a new kind of sort method of the number id.


ax3l:
Thank you for the pull request!

Can you please separate the interpolation issue for MR that you describe and the move/window issue in independent PRs?

Please also do not leave commented old code in the PR :)

According to our CI, there is a test crashing now, can you take a look why it crashes with this change?

jiecaix:
Thank you for help.
I separate the Diagnostics boundary issue for "fix Diagnostics.cpp for diag_lo and diag_hi #4457", and plan to made the FieldProbe to be anothor pull request.
But i'm get a little tough for the "ECP-WarpX.CI (cartesian1d)" check, the Azure Pipelines report a error in the log information, it says that "LaserAccelerateMR CRASHED". But i don't know the details and how to debug in the Azure Pipelines.

PR #4456, opened 2023-11-30
New Cache Strategy

WeiqunZhang:
If a cache is updated during a job, GitHub does not store the new cache unless it's not found with the primary key. To work around that, we use the git hash in the cache name. When the cache action tries to load the cache using the primary key, it's always a miss. It will then try to load a cache using the restore-keys. If the cache is updated during CI, the new version will be stored under the primary key's name.

In this new strategy, a new PR will use the cache associated with the latest development branch. The updated cache will be stored and become private to that PR. Any updates to the PR will also create new updated caches. But we don't need to keep all the copies.  We only need to keep the latest version of the cache associated with the PR. So we use a workflow (clean-cache.yml) to clean up the old caches.

After a PR is closed, we use clean-cache-postpr.yml to clean up all the caches associated with the PR. There is no easy way to get the PR number of the PR that triggers the cleanup workflow. Thus, we store the PR number as an artifact for the cleanup workflow.

It turns out ccache encounters internal errors in some CIs on Ubuntu 20.04. Switching to the latest ccache solves the issues.

CIs on Windows are not updated, because ccache does not work there.

### DIFF
--- a/.github/workflows/clang_tidy.yml
+++ b/.github/workflows/clang_tidy.yml
@@ -20,9 +20,9 @@ jobs:
       uses: actions/cache@v3
       with:
         path: ~/.cache/ccache
-        key: ccache-clang-tidy-${{ github.job }}-git-${{ github.sha }}
+        key: ccache-${{ github.workflow }}-${{ github.job }}-git-${{ github.sha }}
         restore-keys: |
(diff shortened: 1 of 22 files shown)


ax3l (approved):
Thank you, that is awesome :rocket: :sparkles:

PR #4460, opened 2023-12-01
Use "class" on the same line as the name of the class

eebasso:
Adjusts a few class definitions so that the word "class" appears on the same line as the name of the class.


ax3l (approved):
LGTM :+1:

PR #4461, opened 2023-12-01
Use RK4 to integrate the B-field in time in the hybrid-PIC algorithm

roelof-groenewald:
This PR provides a performance improvement in the hybrid-PIC algorithm by using a Runge-Kutta scheme to advance the B-field rather than a simple forward differencing scheme. This allows a lower number of `substeps` to be used in the field advance which improves the algorithm performance.

This should be merged after #4405 as these changes were branched off that PR.

Todo:
- [x] Complete RK4 transition
- [x] Rerun benchmarks to confirm physics accuracy
- [x] Update CI tests
- [x] Change default to 10 sub-steps


aveksler1:
**Performance Improvement**
The performance improvement from switching from a first order substep procedure to update the B-field (left) to a 4th order RK substep scheme (right) is shown below. ~1.6x faster!
image
Caption: Parallel scaling of the WarpX hybrid-PIC algorithm on Perlmutter GPU nodes with fixed gris size of 148x148x592, while particle count increases proportionally to the compute resources, _i.e._, a mixed strong and weak scaling test.

**Ensuring correct physics**
We reran some EM modes and Landau damping test in 1D, 2D, 3D, and RZ as physics tests to make sure the transition to RK4 didn't break anything. The number of substeps was fixed at 20 (a 10 to 50-fold reduction compared to the previous # of substeps required for these tests for numerical stability).

**Parallel EM Modes**
**1d:**
spectrum_par_1d_20_substeps_1e-07_eta

**3d:**
spectrum_par_3d_20_substeps_1e-07_eta

**Perpendicular EM Modes**

**2d:**
spectrum_perp_2d_100_substeps_1e-05_eta

**Normal Modes of cylindrical plasma (RZ)**:
normal_modes_disp

**Landau Damping**
**1d:**
ion_Landau_damping_1d_T_ratio_0 3333333333333333

**2d:**
ion_Landau_damping_2d_T_ratio_0 3333333333333333

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -534,3 +534,156 @@ void HybridPICModel::FillElectronPressureMF (
...
+
+        // Overwrite the Bfield with the Runge-Kutta sum:
+        // B_new = B_old + 1/3 * dt * (0.5 * K0 + K1 + K2 + 0.5 * K3).
+        MultiFab::LinComb(
+            *Bfield[lev][ii], 1.0, B_old[ii], 0, 1.0/3.0, K[ii], 0, 0, 1, ng
+        );
+    }
+}
+
+void HybridPICModel::BfieldPush (
```
clarkse: This function also updates E as well as B.  Perhaps just call it FieldPush?  It is not a show stopper, but could be a bit unclear when looking at the top level routine if the Ohm's law E solve was done.

clarkse (approved):
I think this looks good and I am excited to use it as well since the previous second order integration scheme needed substantial subcycling to remain stable.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
roelof-groenewald: Good point. Thanks for raising it. I called it `BfieldPush` since the E-field is not really "pushed" but rather recalculated. But I do see your point about it looking from the function name that only the B field is updated. I'll change it when I push updates for other requested changes from other reviewers.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -534,3 +534,156 @@ void HybridPICModel::FillElectronPressureMF (
...
+    }
+}
+
+void HybridPICModel::BfieldEvolveRK (
+        amrex::Vector<std::array< std::unique_ptr<amrex::MultiFab>, 3>>& Bfield,
+        amrex::Vector<std::array< std::unique_ptr<amrex::MultiFab>, 3>>& Efield,
+        amrex::Vector<std::array< std::unique_ptr<amrex::MultiFab>, 3>> const& Jfield,
+        amrex::Vector<std::unique_ptr<amrex::MultiFab>> const& rhofield,
+        amrex::Vector<std::array< std::unique_ptr<amrex::MultiFab>, 3>> const& edge_lengths,
+        amrex::Real const dt, int lev, DtType dt_type,
```
ax3l: Avoid `const` for fundamental types that are passed by-value:
```suggestion
        amrex::Real dt, int lev, DtType dt_type,
```

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -534,3 +534,156 @@ void HybridPICModel::FillElectronPressureMF (
...
     }
 }
+
+void HybridPICModel::BfieldEvolveRK (
+        amrex::Vector<std::array< std::unique_ptr<amrex::MultiFab>, 3>>& Bfield,
+        amrex::Vector<std::array< std::unique_ptr<amrex::MultiFab>, 3>>& Efield,
+        amrex::Vector<std::array< std::unique_ptr<amrex::MultiFab>, 3>> const& Jfield,
+        amrex::Vector<std::unique_ptr<amrex::MultiFab>> const& rhofield,
+        amrex::Vector<std::array< std::unique_ptr<amrex::MultiFab>, 3>> const& edge_lengths,
+        amrex::Real const dt, DtType dt_type,
```
ax3l:
```suggestion
        amrex::Real dt, DtType dt_type,
```

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -534,3 +534,156 @@ void HybridPICModel::FillElectronPressureMF (
...
+    }
+}
+
+void HybridPICModel::BfieldPush (
+        amrex::Vector<std::array< std::unique_ptr<amrex::MultiFab>, 3>>& Bfield,
+        amrex::Vector<std::array< std::unique_ptr<amrex::MultiFab>, 3>>& Efield,
+        amrex::Vector<std::array< std::unique_ptr<amrex::MultiFab>, 3>> const& Jfield,
+        amrex::Vector<std::unique_ptr<amrex::MultiFab>> const& rhofield,
+        amrex::Vector<std::array< std::unique_ptr<amrex::MultiFab>, 3>> const& edge_lengths,
+        amrex::Real const dt, DtType dt_type,
```
ax3l:
```suggestion
        amrex::Real dt, DtType dt_type,
```

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -534,3 +534,156 @@ void HybridPICModel::FillElectronPressureMF (
...
+    std::array< MultiFab, 3 > K;
+    for (int ii = 0; ii < 3; ii++)
+    {
+        B_old[ii] = MultiFab(
+            Bfield[lev][ii]->boxArray(), Bfield[lev][ii]->DistributionMap(), 1,
+            Bfield[lev][ii]->nGrowVect()
+        );
+        MultiFab::Copy(B_old[ii], *Bfield[lev][ii], 0, 0, 1, ng);
+
+        K[ii] = MultiFab(
```
ax3l, line 579: If I am not mistaken, then you can reduce the memory footprint of the currently allocated MultiFabs if you destruct the prior stored on before constructing the next one... cc @WeiqunZhang

WeiqunZhang: `K[ii]` is empty before the assignment. In some other situations, yes.

roelof-groenewald: @ax3l Just to make sure I understand, you mean if `K[ii]` already held a MultiFab it would be better to destruct it before overwriting with a new MultiFab?

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.H
```diff
@@ -107,6 +107,33 @@ public:
...
+        amrex::IntVect ng, std::optional<bool> nodal_sync);
+
+    void FieldPush (
+        amrex::Vector<std::array< std::unique_ptr<amrex::MultiFab>, 3>>& Bfield,
+        amrex::Vector<std::array< std::unique_ptr<amrex::MultiFab>, 3>>& Efield,
+        amrex::Vector<std::array< std::unique_ptr<amrex::MultiFab>, 3>> const& Jfield,
+        amrex::Vector<std::unique_ptr<amrex::MultiFab>> const& rhofield,
+        amrex::Vector<std::array< std::unique_ptr<amrex::MultiFab>, 3>> const& edge_lengths,
+        amrex::Real dt, DtType dt_type,
+        amrex::IntVect ng, std::optional<bool> nodal_sync);
```
RemiLehe, line 135: Could you add docstrings for these functions (esp. for the parameter `dt_type`, which may be the least intuitive)?

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -534,3 +534,156 @@ void HybridPICModel::FillElectronPressureMF (
...
+        // Add 2 * 0.5 * dt * K1 to index 0 of K.
+        MultiFab::LinComb(
+            K[ii], 1.0, K[ii], 0, 2.0, K[ii], 1, 0, 1, ng
+        );
+
+        // Overwrite the Bfield with the Runge-Kutta sum:
+        // B_new = B_old + 1/3 * dt * (0.5 * K0 + K1 + K2 + 0.5 * K3).
+        MultiFab::LinComb(
+            *Bfield[lev][ii], 1.0, B_old[ii], 0, 1.0/3.0, K[ii], 0, 0, 1, ng
+        );
```
RemiLehe, line 666: Would it be worth to write all these operations in a single `ParallelFor` kernel, both for readability and to avoid kernel-launch overheads from multiple kernels?

RemiLehe (approved):
Looks good to me! Thanks for this PR!
I added a few comments that could be addressed in follow-up PRs.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
roelof-groenewald: Thanks for the suggestion. I also thought about that as a good follow-up performance improvement but haven't gotten around to it yet since with this new scheme the field solver takes a small amount of runtime in the simulations we are commonly running (so it is not high on the priority list). But of course it would be a good thing to make the code more performant. My current idea is to save this for a time when we have a new team member for which this would be a good task as a way to learn the hybrid-PIC algorithm as well as how to work with WarpX/AMReX.

PR #4464, opened 2023-12-01
Correct on-axis field boundary for RZ

RemiLehe:
This extracts the changes that @dpgrote made to the RZ boundary as part of PR #4071 

This sets the E and B field in the guard cells at `r<0` (i.e. below the axis), by copying the field values at `r>0`, with proper sign flip depending on the symmetry of the field (e.g. `Er`, `Ez`, ...) and of the mode (`m=0`, `m=1`, ...)

Setting the values of these guard cells is important since particles use these values during field gather.


dpgrote (approved):
Thanks!

PR #4465, opened 2023-12-01
Fix preprocessor definition checks for 1D in two locations

archermarx:
In two locations in the code, we had the line

```cpp
#ifdef WARPX_DIM_1D
    // do something
#endif
```

but the actual preprocessor variable is ```WARPX_DIM_1D_Z```. This PR fixes this issue in these two locations.


RemiLehe:
Great, thanks a lot for this PR!

RemiLehe (approved):

PR #4466, opened 2023-12-01
Bugfix in `fields.py` `mesh()` with ghost cells included

roelof-groenewald:
There is an issue in `fields.py`'s `mesh()` function in 1d when ghost cells are included due to the fact that 
```
        ilo = min_box.small_end[idir]
        ihi = min_box.big_end[idir]
```
returns `int`s not lists.
Firstly, the call `ilo = list(ilo)` fails with `TypeError: 'int' object is not iterable`.
Secondly (after fixing the first point), the `lo + np.arange(ilo,ihi+1)*dd + shift` fails due to the fact that `lo` is a float: `TypeError: can only concatenate list (not "int") to list`.

### DIFF
--- a/Python/pywarpx/fields.py
+++ b/Python/pywarpx/fields.py
@@ -160,8 +160,8 @@ class _MultiFABWrapper(object):
         if self.include_ghosts:
             # The ghost cells are added to the upper and lower end of the global domain.
             nghosts = self.mf.n_grow_vect()
-            ilo = list(ilo)
-            ihi = list(ihi)
+            ilo = [ilo] if self.dim == 1 else list(ilo)
+            ihi = [ihi] if self.dim == 1 else list(ihi)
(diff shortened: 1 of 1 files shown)


dpgrote:
Clearly, this had not been tested. It is broken for any dimensionality. It should be
```
        if self.include_ghosts:
            # The ghost cells are added to the upper and lower end of the global domain.
            nghosts = self.mf.n_grow_vect()
            ilo -= nghosts[idir]
            ihi += nghosts[idir]

```

roelof-groenewald:
> Clearly, this had not been tested. It is broken for any dimensionality.

Oh, I assumed it was tested at some point but just not for 1d, which is why I assumed the issue is particular to 1d. I'll update accordingly.

dpgrote (approved):
Thanks for fixing! This code was working in the past - the broken code was from a recent change and that hadn't been tested.

PR #4468, opened 2023-12-04
[WIP] Update install documentation for Brew

aveksler1:
I updated the installation documentation for `Brew`. 
It was previously "coming soon...", so I updated it to reflect the brew dependency installation as outlined in the `macos.yml` file used in github workflows.
Users are then pointed to install using `CMake`. 

Here is a preview of the docs:
image

### DIFF
--- a/Docs/source/install/users.rst
+++ b/Docs/source/install/users.rst
@@ -156,7 +156,7 @@ Using the Brew Package
    python3 -m pip install --upgrade build packaging setuptools wheel
    python3 -m pip install --upgrade mpi4py
 
-WarpX can now be installed from source using CMake, outline below.
+WarpX can now be installed from source using CMake, outlined below.
 
 .. _install-cmake:
 
--- a/Docs/source/install/users.rst
+++ b/Docs/source/install/users.rst
@@ -129,24 +129,15 @@ In the future, will publish pre-compiled binary packages on `PyPI <https://pypi.
 Using the Brew Package
 ----------------------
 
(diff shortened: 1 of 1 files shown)


ax3l:
Hi @aveksler1, thanks a lot for the PR!

The docs for dependencies for from-source install are linked here: https://warpx.readthedocs.io/en/latest/install/dependencies.html#brew-macos-linux

What I envisioned for the doc you currently updated is a package like  this:
- user-facing doc: https://openpmd-api.readthedocs.io/en/0.15.2/install/install.html#using-the-brew-package
- code details for the package in brew: https://github.com/openPMD/homebrew-openPMD

aveksler1:
Ah sorry for misunderstanding. How much work do you think creating a homewbrew formula would be? If it's not a huge undertaking, I'd be interested in doing this, but otherwise I'll just close this.

ax3l:
Hi @aveksler1,

since we can build WarpX now for all four dimensions at once, this should be easy :)

I started a new repo: https://github.com/ECP-WarpX/homebrew-blast

One can tap it via
```bash
brew update
brew tap ecp-warpx/blast
```

and after that install formulas of the kind:
```bash
brew install warpx
```

Take a look at https://github.com/openPMD/homebrew-openPMD/blob/master/Formula/openpmd-api.rb if you like and fill the `Formula/warpx.rb` accordingly with it?
I added a quick draft here: https://github.com/ECP-WarpX/homebrew-blast/blob/main/Formula/warpx.rb

PR #4470, opened 2023-12-04
Remove brew cache

WeiqunZhang:
The size of the brew cache is huge. It can be as big as more than 6 GB. Since we only have 10 GB GitHub cache space, the brew cache is disabled in the macOS CI. It should not increase the time for the macOS CI too much. Without any cache, it usually takes less than 10 minutes to install all the brew packages.


ax3l (approved):
LGTM, thank you!

PR #4472, opened 2023-12-05
Doc: Streamline Python Input

ax3l:
Starting to streamline the Python user-facing sections :)


Python/pywarpx/callbacks.py
```diff
@@ -6,6 +6,7 @@
 #
 # License: BSD-3-Clause-LBNL
 
+# add this in the docs
```
ax3l: This string should show up in the docs. Not sure what exactly it currently is attached to, maybe the file it is in.

ax3l: yep, that's it

RemiLehe (approved):

PR #4475, opened 2023-12-05
macOS CI: export CCACHE_DEPEND=1

WeiqunZhang:
Without it, the hit rate of ccache is very low.


ax3l:
Very interesting
https://ccache.dev/manual/4.0.html#_the_depend_mode

ax3l (approved):

PR #4477, opened 2023-12-05
Add updated instructions for WarpX on Karolina

berceanu:
Use spack for installing all dependencies, and the new slurm scheduler.


berceanu:
This is done from my side. @ax3l, what do you think?

Tools/machines/karolina-it4i/karolina_warpx.profile.example
```diff
@@ -0,0 +1,81 @@
...
+fi
+
+# download and activate spack
+if [ ! -d "$WORK/spack" ]
+then
+    # load required modules
+    module purge
+    module load OpenMPI/4.1.4-GCC-11.3.0-CUDA-11.7.0
+
+    git clone -c feature.manyFiles=true https://github.com/spack/spack.git $WORK/spack
```
ax3l: I would recommend to use a stable release of Spack, with `-b v0.21.0`. This will avoid surprising breakage if something changes on `develop` of Spack and you can in a controlled way bump the version in PRs later on.

Tools/machines/karolina-it4i/karolina_warpx.profile.example
```diff
@@ -0,0 +1,81 @@
...
+# clone warpx
+# you can also clone your own fork here, eg git@github.com:<user>/WarpX.git
+if [ ! -d "$WORK/src/warpx" ]
+then
+    git clone https://github.com/ECP-WarpX/WarpX.git $WORK/src/warpx
+fi
+
+# download and activate spack
+if [ ! -d "$WORK/spack" ]
+then
```
ax3l, line 31: Having this in an extra `install_..._dependencies.sh` script is intentional:
- it allows you to clean up prior installs
- it gives the user control to wait hours for the first install (`spack install`) when they are ready

ax3l (commented):
This looks great, thank you! Minor suggestions inline.

Tools/machines/karolina-it4i/karolina_warpx.profile.example
berceanu: Yup, good point. Ideally we should also pin the dependency versions in the spack env file, as well as the pip packages.

Tools/machines/karolina-it4i/karolina_warpx.profile.example
berceanu: I moved the spack install to the `install_dependencies.sh` script

berceanu:
ping

Tools/machines/karolina-it4i/install_dependencies.sh
```diff
@@ -0,0 +1,47 @@
...
+# this might take about ~ 1 hour
+if [ ! -d "$WORK/spack" ]
+then
+    git clone -c feature.manyFiles=true -b v0.21.0 https://github.com/spack/spack.git $WORK/spack
+    source $WORK/spack/share/spack/setup-env.sh
+
+    # create and activate the spack environment
+    spack env create warpx-karolina-cuda $WORK/src/warpx/Tools/machines/karolina-it4i/spack-karolina-cuda.yaml
+    spack env activate warpx-karolina-cuda
+    spack install
```
ax3l, line 33:
```suggestion
    spack install
fi
```

I would also add an `else` branch that checks out the `v0.21.0` branch if the directory exists, removes the `warpx-karolina-cuda` environment, if it exists, and does the same install.

The idea about this script is, that you can run it multiple times and it will "refresh" your install if it should be needed, e.g., if you updated the docs and tell a user "please reinstall everything". In that case, you do not want to instruct how to "deinstall" the existing deps, because it gets confusing, instead automate it here.

ax3l (commented):
Thank you!

I think there is a small bug left and a suggestion. Please also run again the final workflow to make sure there are not bugs left :)

ax3l:
@berceanu ping :)

ax3l (approved):

PR #4478, opened 2023-12-05
Set alpha mass directly from NIST data

eebasso:
NIST data provides the mass of the alpha particle directly with a high degree of certainty. We might as well use that instead of approximating it by subtracting the mass of two electrons from the mass of helium4.


eebasso:
Some regression tests might have to be recomputed with the new alpha mass. Any help with that would be appreciated.

RemiLehe (approved):

PR #4479, opened 2023-12-05
Update descriptions of hybrid-PIC examples

roelof-groenewald:
Here is a first set of updates for the hybrid-PIC examples, switching to the new example layout.

### DIFF
--- a/Docs/source/usage/examples.rst
+++ b/Docs/source/usage/examples.rst
@@ -113,9 +113,16 @@ Coming soon:
 Kinetic-fluid Hybrid Models
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Several examples and benchmarks of the kinetic-fluid hybrid model are provided below.
-A few examples are replications of the verification tests described in :cite:t:`ex-MUNOZ2018`.
-The hybrid-PIC model was added to WarpX in `PR #3665 <https://github.com/ECP-WarpX/WarpX/pull/3665>`_ - the figures in the examples below were generated at that time.
+WarpX includes a reduced plasma model in which electrons are treated as a massless
+fluid while ions are kinetically evolved, and Ohm's law is used to calculate
+the electric field. This model is appropriate for problems in which ion kinetics
+dominate (ion cyclotron waves, for instance). See the
+:ref:`theory section <theory-kinetic-fluid-hybrid-model>` for more details. Several
+examples and benchmarks of this kinetic-fluid hybrid model are provided below.
+A few of the examples are replications of the verification tests described in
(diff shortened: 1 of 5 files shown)


Examples/Tests/ohm_solver_EM_modes/README.rst
```diff
@@ -1,11 +1,56 @@
 .. _examples-ohm-solver-em-modes:
 
-Ohm Solver: Electromagnetic modes
-=================================
+Electromagnetic modes
+=====================
```
ax3l: I made the title a bit more specific because it is not the default loop (EM PIC)... is that worth keeping? Let me know if I misunderstood something.

Docs/source/usage/examples.rst
```diff
@@ -113,8 +113,8 @@ Coming soon:
 Kinetic-fluid Hybrid Models
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Several examples and benchmarks of the kinetic-fluid hybrid model are shown below.
-The first few examples are replications of the verification tests described in :cite:t:`ex-MUNOZ2018`.
+Several examples and benchmarks of the kinetic-fluid hybrid model are provided below.
```
ax3l: As a general thought I had, can we potentially rework this to use a even more descriptive name than `Kinetic-fluid Hybrid`?
I am a bit concerened about mismatches with other hybrid "solvers" we have and other fluid-hybrid models like https://warpx.readthedocs.io/en/latest/theory/cold_fluid_model.html

ax3l (commented):
This is excellent, thank you! Two questions inline, mods look great! :)

Examples/Tests/ohm_solver_EM_modes/README.rst
roelof-groenewald: The reason I changed it is because this page is a child of the "Kinetic-fluid Hybrid Models" examples page:
image
so my thinking was that it is redundant to specify "Ohm solver" again. But I definitely can see an argument that the "Ohm solver" is a type of fluid-hybrid solver and therefore needs the extra specificity.

Docs/source/usage/examples.rst
roelof-groenewald: Confusion with other hybrid models is definitely a concern (also note that in the input script the Ohm solver hybrid scheme takes input as `hybrid_pic_model.xxx`).
I will change the wording in the section you linked to specify "inertialess electron fluid, kinetic ion" model. And with this consideration I will change the section headings in the previous comment back to specifying that those examples use the "Ohm's law based solver".

ax3l (approved):
Awesome, thank you!

PR #4480, opened 2023-12-05
Use more consistent species types in fusion module

RemiLehe:
This changes the code, so that the user has to use `hydrogen1` and `helium4` instead of `proton` and `alpha` in the fusion modules. I makes the code more consistent across all fusion modules, and makes sure that the mass of the electron is included in all species involved (instead of having some species that include the mass of the electrons while others don't)


eebasso:
One possibility is that we allow for both helium4 and alpha. Is alpha going to be retired soon?

eebasso (changes requested):
I like the changes as they make the terminology more consistent and less confusing. Perhaps we should delete `PhysicalSpecies::proton` and `PhysicalSpecies::alpha` from the enum struct `Physical Species` to avoid future confusion.

roelof-groenewald (approved):
Looks good to me. I agree just having consistency between the different fusion reactions will avoid confusion. 
In my opinion it is preferred to keep `PhysicalSpecies::proton` since it is also used to set up electron-proton plasmas that do not involve the fusion module. I don't have a strong opinion about keeping or deleting `PhysicalSpecies::alpha`.

PR #4481, opened 2023-12-06
Add `latex_theory/allbibs.bib` to bibliography list and delete duplicates

eebasso:
## Summary

This PR modifies `conf.py` to also use the BibTeX file `allbibs.bib` in the latex_theory folder, thus allowing for many more references without having to add to `refs.bib`.

## Additional Information

Many of the entries in `refs.bib` were redundant to those in `allbibs.bib`. Those entries have been deleted, and the citations have been renamed when necessary. Furthermore, the prefix "i-" is used in the theory introduction `intro.rst` to prevent unused references from appearing.


ax3l:
That sounds good.

Some history on the folder: originally, those files were auto-converted from LaTeX to Sphinx .rst on the fly, and there are a few more resources in there we should translate over. Unfortunately, pandoc conversion needs some extra manual interaction to integrate well in formatting with our docs, so we should just convert all that is left in latex once, remove the files, and fix the converted files for style and formatting.

ax3l (approved):

PR #4486, opened 2023-12-06
[WIP]Add a thermal plasma flux

dpgrote:
This implements a method of flux injection where the flux of injected particles is set dynamically so that the grid cells at the plane of injection will always have at least the specified number of particles. The purpose is to provide a stable boundary condition representing an extend thermal plasma.

Todo:
- [x] CI test
- [x] Add to PICMI interface


archermarx:
This is a really nice feature. Looking forward to seeing it merged!

Docs/source/usage/python.rst
archermarx, line 1: Is there a reason for the capitalization of `PLasmaFluxDistribution` (vs `PlasmaFluxDistribution`)?

PR #4490, opened 2023-12-09
sqrt -> std::sqrt

WeiqunZhang:
Unlike std::sqrt that has overloads for double and float, sqrt is a C function that takes a double. If the argument is a float, it will have to be promoted to double first.


ax3l (approved):
Thanks, good catch!

PR #4491, opened 2023-12-09
Clang-Tidy CI: Keep Going after Errors

WeiqunZhang:
Add `--keep-going` to the make command in Clang-Tidy CI. With that, the job will keep going and show all the clang-tidy check errors instead of stopping on the first error.


lucafedeli88 (approved):
Thanks, @WeiqunZhang

PR #4492, opened 2023-12-09
Clang-Tidy: Update performance and readability

WeiqunZhang:
For performance and readability, instead of including checks explicitly, switch to excluding checks explicitly. These checks are excluded because currently WarpX fails at them.

Add portability and diagnostic checks.


lucafedeli88 (approved):
Thanks for this PR, @WeiqunZhang .

PR #4494, opened 2023-12-09
macOS CI: Fix brew installation by explicitly specifing Python version

WeiqunZhang:
Force install python 3.12. This is likely to be faster than `brew upgrade` in #4493.


WeiqunZhang:
Unfortunately it did not work. But somehow it worked on the push to my fork.

PR #4498, opened 2023-12-11
[WIP] Shared Memory Esirkpov Deposition

atmyers:
This implements a shared memory option for Esirkepov deposition. This includes work by @hklion and optimizations made by @mhaseeb123.


roelof-groenewald:
There are some outdated changes in this PR due to #4978. In particular the arguments passed to the deposition kernels were changed (specifically the variables that used to be `dx` and `xyzmin`).
If this branch is rebased on `development` the build failures would be seen.

thmihnea:
Hi,

We wanted to test the Esirkepov current deposition method's speed and seing whether it is worth it to use shared memory or not. We ran multiple simulations:

1. Esirkepov current deposition with shared memory activated (using 1, 2 & 4 nodes; 1 node = 4 GPUs)
2. Esirkepov current deposition without shared memory activated (using 1, 2 & 4 nodes)
3. Esirkepov current deposition with shared memory activated (using 2 nodes only), sweeping over the NPPC variable (using 64, 128, 256 and 512 particles)
4. Esirkepov current deposition without shared memory activated (2 nodes only), sweeping over the NPPC variable (as previously).

The results are shown below.

warpx_esirkepov_NPPC_sweep_metrics
warpx_esirkepov_no_shared_NPPC_sweep_metrics
warpx_esirkepov_no_shared_sweep_metrics
warpx_esirkepov_node_sweep_metrics

As can be seen from the above plots (these plot the time it took for the simulation to run the WarpXParticleContainer::DepositCurrent::CurrentDeposition method), running the Esirkepov current deposition scheme using shared memory results in a much slower runtime (it takes ~20x more time to run the node sweeps and ~40x more time to run the NPPC sweeps).

Do you know what could cause these issues?

Thanks.
Mihnea

mhaseeb123:
> 3. Esirkepov current deposition with shared memory activated (using 2 nodes only), sweeping over the NPPC variable (using 64, 128, 256 and 512 particles)
> 4. Esirkepov current deposition without shared memory activated (2 nodes only), sweeping over the NPPC variable (as previously).

@thmihnea are the figures 3 and 4 swapped?

thmihnea:
Figures 1 and 2 are NPPC sweeps, 3 and 4 are node sweeps. I have also attached the simulation types in the figure titles.

atmyers:
@thmihnea Could you share the inputs files you used to generate these results? When I do this comparison at 64 ppc using a standard uniform plasma benchmark, I get very similar timings between the two approaches.

thmihnea:
@atmyers Yes - I used the script Examples/Tests/ohm_solver_EM_modes/PICMI_inputs.py. For the simulations where we swept across the number of nodes, we kept NPPC[2] at 256 and submitted SLURM simulations using -N 1, -N 2, -N 4. For the case where we swept across NPPC, we set NPPC[2] to 64, 128, 256, 512. 

I could also provide you with the entire .py scripts if you wish.

PR #4505, opened 2023-12-12
Implement correction to ADK for high fields

MaxThevenet:
This PR supersedes the old #3011. ADK is valid for an external field smaller than the critical field, but we sometimes needs to simulate higher fields. Article [Zhang et al., PRA 90, 043410 (2014)](https://journals.aps.org/pra/pdf/10.1103/PhysRevA.90.043410) proposes some correction to ADK for high fields, which can yield some small difference. This PR proposes to implement this correction, for Hydrogen only for the moment.


ax3l:
@MaxThevenet thank you for this PR! Can you add input documentation for it? :)

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -2967,10 +2967,15 @@ PhysicalParticleContainer::InitIonizationModule ()
...
     }
+    utils::parser::queryWithParser(pp_species_name, "do_adk_correction", do_adk_correction);
+
     utils::parser::queryWithParser(
         pp_species_name, "ionization_initial_level", ionization_initial_level);
     pp_species_name.get("ionization_product_species", ionization_product_name);
     pp_species_name.get("physical_element", physical_element);
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+        physical_element == "H" || !do_adk_correction,
+        "Correction to ADK by Zhang et al., PRA 90, 043410 (2014) only workds with Hydrogen");
```
ax3l:
```suggestion
        "Correction to ADK by Zhang et al., PRA 90, 043410 (2014) only works with Hydrogen");
```

Source/Particles/ElementaryProcess/Ionization.H
```diff
@@ -82,8 +84,10 @@ struct IonizationFilterFunc
                           const amrex::Real* AMREX_RESTRICT a_adk_prefactor,
                           const amrex::Real* AMREX_RESTRICT a_adk_exp_prefactor,
                           const amrex::Real* AMREX_RESTRICT a_adk_power,
+                          const amrex::Real* const AMREX_RESTRICT a_adk_correction_factors,
```
ax3l:
```suggestion
                          const amrex::Real* AMREX_RESTRICT a_adk_correction_factors,
```

ax3l (approved):
Thank you! :sparkles:

PR #4506, opened 2023-12-12
Bibliography: make author family names appear last

eebasso:
This is a simple follow up on the custom class that disables the 'lastfirst' option. That option makes the family names of authors appear before the given names. Disabling it reverts to the default of family names appearing after given names. This leads to less commas overall, which can be considered more aesthetically pleasing. Example: "Jean-Luc Vay" will go from "Vay, J.-L." to "J.-L. Vay"


ax3l (approved):
Thanks!

PR #4507, opened 2023-12-12
Clang-Tidy: readability-qualified-auto

WeiqunZhang:


Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -1026,7 +1026,7 @@ WarpXOpenPMDPlot::SetConstParticleRecordsEDPIC (
         amrex::ParticleReal const mass)
 {
     auto realType = openPMD::Dataset(openPMD::determineDatatype<amrex::ParticleReal>(), {np});
-    auto const scalar = openPMD::RecordComponent::SCALAR;
+    const auto *const scalar = openPMD::RecordComponent::SCALAR;
```
WeiqunZhang, line 1029: @ax3l Do you prefer the change here or using `NOLINT(readability-qualified-auto)`?

I didn't use `NOLINT` because so far we have not used it in WarpX. But I think it would actually be better to use `NOLINT`. First, `openPMD::RecordComponent::SCALAR` may not be a pointer in the future. More importantly, knowing it's a pointer does not give us any useful information, given how it is used.

ax3l: Thanks. Technically this is a constant char array right now. I am ok to not NOLINT it and fix it if we change/break it some day.

ax3l (approved):

PR #4510, opened 2023-12-13
Clang-tidy: enforce clang-tidy_readability_make_member_function_const check

lucafedeli88:
TODO


Source/AcceleratorLattice/AcceleratorLattice.cpp
```diff
@@ -88,7 +88,7 @@ AcceleratorLattice::InitElementFinder (int const lev, amrex::BoxArray const & ba
 }
 
 void
-AcceleratorLattice::UpdateElementFinder (int const lev)
```
WeiqunZhang, line 91: We could consider NOLINT here. Technically this could be a const function. However, it's only because `m_element_finder` is a `std::unique_ptr`. But that's an implementation detail. To the user, it actually makes more sense to not mark this as const.

WeiqunZhang: https://isocpp.org/wiki/faq/const-correctness#logical-vs-physical-const

WeiqunZhang (approved):

ax3l:
@WeiqunZhang I realized that with the current caching logic, we do not run clang tidy over all files when introducing new rules (as in this PR).

WeiqunZhang:
We have `export CCACHE_EXTRAFILES=${{ github.workspace }}/.clang-tidy`. https://ccache.dev/manual/latest.html#config_extra_files_to_hash

WeiqunZhang:
So any modification to `.clang-tidy` will invalidate the cache.

ax3l (approved):

PR #4515, opened 2023-12-14
Clang-Tidy: readability-isolate-declaration

WeiqunZhang:


WeiqunZhang:
I am actually not sure we should use this check. It seems unnecessary. I did this PR only because it could be fixed automatically by clang-tidy. So if we agree, we could close this PR.

ax3l:
Thanks for the details. Yeah, it think it's ok to have some compact declarations here and there. Did this fix a bug?

WeiqunZhang:
No, there were no bugs.

WeiqunZhang:
Will not use this check.

PR #4156, opened 2023-08-04
[WIP] CI: checksum regression analysis for openPMD output

EZoni:
Milestone-related task: add support for openPMD output in our checksum regression analysis.

**Goal:**
Perform checksum regression analysis on both plotfile and openPMD output for all tests.

**Notes:**
- The particle momentum values are different between plotfile and openPMD (probably different units/normalization)
- The field values at MR level > 0 seem to be different (see example of `nci_correctorMR`, unclear to me at this point)
- Note to reviewers: please select "hide whitespace" for better display of code changes.

**To-do:**
- [x] Discuss/modify goal and strategy
- [x] Test auxiliary functions (e.g., reset)
- [ ] Update remaining CI tests (currently failing)

**Workflow to add openPMD diagnostics output in a given test:**
- add openPMD diagnostics output named `diag2` to input file
- update calls to checksum analysis in analysis file
- rename original checksum file by adding `_plotfile` suffix
- copy plotfile checksum file into openPMD checksum file, with `_openpmd` suffix
- reset openPMD checksum file after running the test locally with `run_test.sh`

**Follow-up PR(s):**
- Use openPMD output in each test's analysis

PR #4519, opened 2023-12-14
Add ability to run checksum on openPMD files

RemiLehe:
This PR extracts the changes implemented in https://github.com/ECP-WarpX/WarpX/pull/4156 to the checksum infrastructure (`checksum.py`, `checksumAPI.py`), so that it can read both plotfiles and openPMD files, and compute the checksum value. 

Moreover, instead of updating **all** of our examples to use openPMD (as was done in https://github.com/ECP-WarpX/WarpX/pull/4156), this PR only switches the file format (from plotfile to openPMD) for **one** of the examples (`laserAcceleration`), in which case it calls a new script `analysis_default_openpmd_regression.py` (which is almost identical to `analysis_default_regression.py`). Note that the checksum did not need to be reset.


ax3l (approved):
:rocket:

ax3l:
@RemiLehe it looks like
`LaserAcceleration_single_precision_comms` failed

PR #4520, opened 2023-12-15
Doc: Extend a Simulation Workflow

ax3l:
Rework and move the workflow on how to extend simulations from Python.

### DIFF
--- a/Docs/source/usage/python.rst
+++ b/Docs/source/usage/python.rst
@@ -25,7 +25,7 @@ Once the simulation is fully configured, it can be used in one of two modes.
 
       :py:meth:`~pywarpx.picmi.Simulation.write_input_file`: create an :ref:`inputs file for a WarpX executable <running-cpp-parameters>`
 
-When run directly from Python, once can also extend WarpX with further custom user logic.
+When run directly from Python, one can also extend WarpX with further custom user logic.
 See the :ref:`detailed workflow page <usage-python-extend>` on how to extend WarpX from Python.
 
 
--- a/Docs/source/usage/python.rst
+++ b/Docs/source/usage/python.rst
@@ -13,7 +13,7 @@ In the input file, instances of classes are created defining the various aspects
 A variable of type :py:class:`pywarpx.picmi.Simulation` is the central object to which all other options are passed, defining the simulation time, field solver, registered species, etc.
 
 Once the simulation is fully configured, it can be used in one of two modes.
-**Interactive** use is the most common and can be :ref:`extended with custom runtime functionality <usage-picmi-extend>`:
+**Interactive** use is the most common and can be :ref:`extended with custom runtime functionality <usage-python-extend>`:
 
 .. tab-set::
 
@@ -24,6 +24,9 @@ Once the simulation is fully configured, it can be used in one of two modes.
    .. tab-item:: Preprocessor
 
       :py:meth:`~pywarpx.picmi.Simulation.write_input_file`: create an :ref:`inputs file for a WarpX executable <running-cpp-parameters>`
+
+When run directly from Python, once can also extend WarpX with further custom user logic.
+See the :ref:`detailed workflow page <usage-python-extend>` on how to extend WarpX from Python.
 
 
 .. _usage-picmi-parameters:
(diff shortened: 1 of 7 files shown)


Docs/source/usage/python.rst
```diff
@@ -25,6 +25,9 @@ Once the simulation is fully configured, it can be used in one of two modes.
 
       :py:meth:`~pywarpx.picmi.Simulation.write_input_file`: create an :ref:`inputs file for a WarpX executable <running-cpp-parameters>`
 
+When run directly from Python, once can also extend WarpX with further custom user logic.
```
dpgrote: Typo? "one" instead of "once"

Docs/source/usage/workflows.rst
```diff
@@ -8,11 +8,12 @@ This section collects typical user workflows and best practices for WarpX.
 .. toctree::
    :maxdepth: 2
 
+   workflows/extend
```
dpgrote: To be more precise, perhaps the file should be called something like "extend_with_python.rst". Who knows, perhaps in the future someone might implement "extend_with_julia.rst".

Docs/source/usage/workflows/extend.rst
```diff
@@ -0,0 +1,229 @@
+.. _usage-python-extend:
+
+Extend a Simulation
```
dpgrote:
```suggestion
Extend a Simulation with Python
```

Docs/source/usage/workflows/extend.rst
```diff
@@ -0,0 +1,229 @@
...
+
+* `amr.ParallelDescriptor <https://pyamrex.readthedocs.io/en/latest/usage/api.html#amrex.space3d.ParallelDescriptor.IOProcessor>`__: MPI-parallel rank information
+* `amr.MultiFab <https://pyamrex.readthedocs.io/en/latest/usage/api.html#amrex.space3d.MultiFab>`__: MPI-parallel field data
+* `amr.ParticleContainer_* <https://pyamrex.readthedocs.io/en/latest/usage/api.html#amrex.space3d.ParticleContainer_1_1_2_1_default>`: MPI-parallel particle data for a particle species
+
+
+Data Access
+-----------
+
+While the simulation is running, callbacks can read and write access the WarpX simulation data *in situ*.
```
dpgrote:
```suggestion
While the simulation is running, callbacks can have read and write access the WarpX simulation data *in situ*.
```

Docs/source/usage/python.rst
ax3l:
```suggestion
When run directly from Python, one can also extend WarpX with further custom user logic.
```

Docs/source/usage/workflows/extend.rst
ax3l: I thought to keep it simple for now, because there is mostly one way, but agreed - technically we can also extend WarpX from C/C++ right now (e.g., with Cling) :sweat_smile:

Docs/source/usage/workflows/extend.rst
```diff
@@ -0,0 +1,229 @@
...
+
+.. code-block:: python3
+
+   # Preparation: set up the simulation
+   #   sim = picmi.Simulation(...)
+   #   ...
+
+
+   @callfromafterstep
+   def sum_rho_electrons():
```
roelof-groenewald: This function name should probably be changed to something more descriptive of the actual function, perhaps `get_max_E_x()`

Docs/source/usage/workflows/extend.rst
```diff
@@ -0,0 +1,229 @@
...
+
+      The physical fields in WarpX have the following naming:
+
+      - ``_fp`` are the "fine" patches, the regular resolution of a current mesh-refinement level
+      - ``_aux`` are temporary (auxiliar) patches at the same resolution as ``_fp``.
+        They usually include contributions from other levels and can be interpolated for gather routines of particles.
+      - ``_cp`` are "coarse" patches, at the same resolution (but not necessary values) as the ``_fp`` of ``level - 1``
+        (only for level 1 and higher).
+
+   .. py:method:: multi_particle_container
```
dpgrote:
```suggestion
   .. py:method:: multi_particle_container()
```
Add empty parenthesis to make it clear that this is a method.

Docs/source/usage/workflows/extend.rst
```diff
@@ -0,0 +1,229 @@
...
+
+.. code-block:: python
+
+   from pywarpx import particle_containers, picmi
+
+   # ...
+
+   electron_wrapper = particle_containers.ParticleContainerWrapper("electrons")
+
+
```
roelof-groenewald: This is missing the actual example of adding particles:
```
electron_wrapper.add_particles(x=x, y=y, z=z, ux=ux, uy=uy, uz=uz, w=weight)

where x, y, z, ux, uy, uz, weight are either single values or numpy arrays of particle coordinates.
```

roelof-groenewald:
I think it would be good to (somewhere in this section) point out the example in `Examples/Physics_applications/capacitive_discharge/PICMI_inputs_2d.py` where the MLMG field solver is replaced by a python function that uses superLU to directly solve the Poisson equation.

Docs/source/usage/workflows/extend.rst
ax3l: The empty parenthesis are removed for no-arg methods - this otherwise causes a warning in Spinx.

Docs/source/usage/workflows/extend.rst
```diff
@@ -0,0 +1,229 @@
...
+   #   ...
+
+
+   @callfromafterstep
+   def sum_rho_electrons():
+       warpx = sim.extension.warpx
+
+       # data access
+       E_x_mf = warpx.multifab(f"Efield_fp[x][l=0]")
+       E_x_np = E_x_mf.to_numpy(copy=True)  # list of arrays
```
dpgrote: This doesn't seem like a good example since it is making a copy of the arrays, which would include transferring the data from the GPU to CPU. Is there a reason it is done this way instead of using the `E_x_mf.max()` method or showing an example of an `mfiter` loop?

ax3l: > This doesn't seem like a good example since it is making a copy of the arrays, which would include transferring the data from the GPU to CPU. 

This is intentional currently, because this will work on both CPU and GPU runs. The link below the example points to more general GPU examples.

I think we need to add a `.to_xp()` method to make it more agnostics, otherwise we need to add a couple of if loops.

> Is there a reason it is done this way instead of using the E_x_mf.max() method or showing an example of an mfiter loop?

Yes, I wanted to show a small example that can run in mfiter, but maybe `max` is not the best idea because we have an intrinsic for it :) Do you have another idea? Maybe write to the array or calculate pointwise `E²`?

Docs/source/usage/workflows/extend.rst
ax3l: The
```
.. autoclass:: pywarpx.particle_containers.ParticleContainerWrapper
   :members:
```
should contain the method:
image

Docs/source/usage/workflows/extend.rst
ax3l: Reference:
https://www.sphinx-doc.org/en/master/usage/domains/python.html#role-py-func

image


(func and method are aliases)

Python/pywarpx/callbacks.py
```diff
@@ -46,27 +43,33 @@
...
-       ppzx()
+       # do something here
+
+   installcallback('afterstep', myplots)
 
-This is equivalent to the following:
+   # run simulation
+   sim.step(nsteps=100)
+
+The install can be done using a `Python decorator <https://docs.python.org/3/glossary.html#term-decorator>`__, which has the prefix ``callfrom``.
```
dpgrote:
```suggestion
The install can also be done using a `Python decorator <https://docs.python.org/3/glossary.html#term-decorator>`__, which has the prefix ``callfrom``.
```

Docs/source/usage/workflows/extend.rst
roelof-groenewald: Ah, okay. I thought that was perhaps the intention but wasn't sure if you also meant to have a code snippet that included it. Thanks for clarifying!

Docs/source/usage/workflows/extend.rst
ax3l: Note that this will still *render* the `()` to the user in the manual, so there is no confusion for the reader :)

Docs/source/usage/workflows/extend.rst
ax3l: Yes, if we want a code snippet we should add it as an example in that doc string :) But maybe it is quite clear (we could add maybe 1-2 more sentences if we need to to the doc string, too).

Docs/source/usage/workflows/extend.rst
ax3l: Updated to calculate `max(E²_x)` instead :)

Docs/source/usage/workflows/python_extend.rst
```diff
@@ -0,0 +1,238 @@
...
+
+
+.. autoclass:: pywarpx.particle_containers.ParticleBoundaryBufferWrapper
+   :members:
+
+
+Modify Solvers
+--------------
+
+From Python, once can also replace numerical solvers in the PIC loop or add new physical processes into the time step loop.
```
roelof-groenewald:
```suggestion
From Python, one can also replace numerical solvers in the PIC loop or add new physical processes into the time step loop.
```

roelof-groenewald (approved):
Looks good! Thanks for adding the `superLU` solve example :)

Docs/source/usage/workflows/extend.rst
ax3l: Discussed: let's do an mfiter loop here:
- easier to extend to multiple fields
- directly usable if indices are needed

ax3l: Done :)

ax3l:
All tested now :tada: :)

Docs/source/usage/workflows/python_extend.rst
```diff
@@ -0,0 +1,277 @@
...
+               # additional compile-time and runtime attributes in SoA format
+               soa = pti.soa().to_numpy()
+
+               # notes:
+               # Only the next lines are the "HOT LOOP" of the computation.
+               # For efficiency, use numpy array operation for speed on CPUs.
+               # For GPUs use .to_cupy() above and compute with cupy or numba.
+
+               # write to all particles in the chunk
+               aos[()]["x"] = 0.30
```
ax3l, line 215:
```suggestion
               # note: careful, if you change particle positions, you need to 
               #       redistribute particles before continuing the simulation step
               aos[()]["x"] = 0.30
```

Docs/source/usage/workflows/python_extend.rst
```diff
@@ -0,0 +1,277 @@
...
+               aos[()]["y"] = 0.35
+               aos[()]["z"] = 0.40
```
ax3l:
```suggestion
               # write to all particles in the chunk
               # note: careful, if you change particle positions, you need to 
               #       redistribute particles before continuing the simulation step
               # aos[()]["x"] = 0.30
               # aos[()]["y"] = 0.35
               # aos[()]["z"] = 0.40
```

PR #4527, opened 2023-12-15
Fix CI for ROCm 6.0

WeiqunZhang:
Need to explicitly install hiprand package in CI because it's now a standalone project, not a submodule for rocRand according to the release notes.


ax3l (approved):
Thank you, looks great :+1:

PR #4528, opened 2023-12-15
Documentation: numbered reference formatting

eebasso:
## Summary

This PR adds options for numbered equation, figure, and other references, and overcomes a bug in the Sphinx RTD Theme related to equation numbering.

## Additional Information

Figures can now be referenced simply as

> In :numref:\`fig_label_example\`, we see that...

which will appear as

> In Fig. 3, we see that...

where `Fig. 3` is a link to the figure labeled `fig_label_example`. Similarly, equations labeled using the `:label:` directive can be referenced via

>Using Eq. (:eq:\`eq_label_example\`) yields...

and will appear as

> Using Eq. (4) yields...

where `4` is a link to the equation labeled as `eq_label_example`.

Additionally, the file `custom.css` is added in the `_static` folder to overcome a bug in the Sphinx RTD Theme wherein the equation numbers appear above and not to the right of the displayed equation.


Docs/source/_static/custom.css
```diff
@@ -0,0 +1,6 @@
+.eqno {
+    /* As of 2023 Dec, sphinx_rtd_theme has a bug where equation numbers appear above the equations instead of on the right */
+    /* The following is a make-shift fix, but comes at the cost of "making the header link invisible," though I'm not sure what that means */
+    /* Copied from https://github.com/readthedocs/sphinx_rtd_theme/issues/301#issuecomment-1205755904 */
+    float: right;
```
ax3l, line 5: Looks good. This is a long-standing bug and I used this in the past for formatting:
```css
/* fix for position of equation numbers
 * https://github.com/rtfd/sphinx_rtd_theme/issues/301
 */
.eqno {
    margin-left: 5px;
    float: right;
}
.math .headerlink {
    display: none;
    visibility: hidden;
}
.math:hover .headerlink {
    display: inline-block;
    visibility: visible;
    margin-right: -0.7em;
}
```

ax3l (approved):

PR #4537, opened 2023-12-16
Docs: rename and spruce up boundary conditions section

eebasso:
Rename `PML.rst` to `boundary_conditions.rst` and fix citations and other references.


Docs/source/theory/boundary_conditions.rst
```diff
@@ -0,0 +1,303 @@
+.. _theory-bc:
```
ax3l, line 1: Don't forget to update the table of contents with the new file name

Docs/source/theory/boundary_conditions.rst
```diff
@@ -0,0 +1,303 @@
...
+
+Boundary conditions
```
ax3l, line 3: Title case?
```suggestion
Boundary Conditions
```

eebasso: Not sure. Seems like multiple case conventions are used. I personally prefer lower case for all words except the first one, proper names such as "Lorentz", and other such exceptions.

Docs/source/theory/boundary_conditions.rst
ax3l: fixed

ax3l (approved):

Docs/source/theory/boundary_conditions.rst
ax3l: I like title case, but I also miss the diverse spelling of words from German :D

PR #4545, opened 2023-12-19
Add updated instructions for WarpX on Karolina

berceanu:
Use spack for installing all dependencies, and the new slurm scheduler.

As suggested in https://github.com/ECP-WarpX/WarpX/pull/4477#discussion_r1425742388, I
added an else branch that checks out the v0.21.0 branch if the directory exists, removes the warpx-karolina-cuda environment, if it exists, and does the same install.

Also added the SRUN_CPUS_PER_TASK var in the slurm script, according to https://slurm.schedmd.com/sbatch.html#OPT_cpus-per-task

Finally, the profile file was somewhat simplified.

### DIFF
--- a/Tools/machines/karolina-it4i/install_dependencies.sh
+++ b/Tools/machines/karolina-it4i/install_dependencies.sh
@@ -26,12 +26,24 @@ if [ ! -d "$WORK/spack" ]
 then
     git clone -c feature.manyFiles=true -b v0.21.0 https://github.com/spack/spack.git $WORK/spack
     source $WORK/spack/share/spack/setup-env.sh
+else
+    # If the directory exists, checkout v0.21.0 branch
+    cd $WORK/spack
+    git checkout v0.21.0
+    git pull origin v0.21.0
+    source $WORK/spack/share/spack/setup-env.sh
 
-    # create and activate the spack environment
-    spack env create warpx-karolina-cuda $WORK/src/warpx/Tools/machines/karolina-it4i/spack-karolina-cuda.yaml
-    spack env activate warpx-karolina-cuda
-    spack install
+    # Delete spack env if present
+    if spack env list | grep -q warpx-karolina-cuda; then
+        spack env deactivate
+        spack env rm -y warpx-karolina-cuda
+    fi
 fi
+
+# create and activate the spack environment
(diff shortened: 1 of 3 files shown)


Tools/machines/karolina-it4i/install_dependencies.sh
```diff
@@ -26,13 +26,25 @@ if [ ! -d "$WORK/spack" ]
...
 
-    # create and activate the spack environment
-    spack env create warpx-karolina-cuda $WORK/src/warpx/Tools/machines/karolina-it4i/spack-karolina-cuda.yaml
-    spack env activate warpx-karolina-cuda
-    spack install
+    # Delete spack env if present
+    if spack env list | grep -q warpx-karolina-cuda; then
+        spack env deactivate
+        spack env rm -y warpx-karolina-cuda
+    fi
```
ax3l: This block can fail and abort the whole script, because we run the script with `set -eu -o pipefail`, as is best practice.

Thus, a more robust way to write this is:
```suggestion
    spack env deactivate || true
    spack env rm -y warpx-karolina-cuda || true

    cd -
```

Tools/machines/karolina-it4i/install_dependencies.sh
```diff
@@ -26,13 +26,25 @@ if [ ! -d "$WORK/spack" ]
 then
     git clone -c feature.manyFiles=true -b v0.21.0 https://github.com/spack/spack.git $WORK/spack
     source $WORK/spack/share/spack/setup-env.sh
+else
+    # If the directory exists, checkout v0.21.0 branch
+    cd $WORK/spack
```
ax3l, line 31: `cd -` is missing.

Added to suggested `diff` suggestion in the lines below.

ax3l (commented):
Thank you!

I added a few suggestions to fix bugs. Please apply and test again with it :)

Tools/machines/karolina-it4i/install_dependencies.sh
berceanu: Nice catch!

berceanu:
> Thank you!
> 
> I added a few suggestions to fix bugs. Please apply and test again with it :)

Applied the changes. Should we also use `set -eu -o pipefail` in the profile script?

ax3l:
No, the profile script is sourced and would change the current user environment.

ax3l (approved):

PR #4550, opened 2023-12-20
Adding user defined keywords in Hybrid PIC Python class to allow for …

clarkse:
…addition of constants when defining functions for resistivity and current.


roelof-groenewald (approved):
Thanks for adding!

PR #4554, opened 2023-12-20
Disambiguate the routine name initialize_inputs in picmi.py

dpgrote:
This PR improves code readability in the picmi interface. Previously, each class had a  method called `initialize_inputs`, but with all classes having a routine of the same name, it made it hard to find specific versions of the method. This PR adds prefixes to the name, based on the input type of the class. For instance, in the diagnostic classes, the routine is called `diagnostic_initialize_inputs`.

### DIFF
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -219,10 +219,10 @@ class Species(picmistandard.PICMI_Species):
         self.resampling_trigger_intervals = kw.pop('warpx_resampling_trigger_intervals', None)
         self.resampling_triggering_max_avg_ppc = kw.pop('warpx_resampling_trigger_max_avg_ppc', None)
 
-    def initialize_inputs(self, layout,
-                          initialize_self_fields = False,
-                          injection_plane_position = None,
-                          injection_plane_normal_vector = None):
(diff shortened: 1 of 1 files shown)


roelof-groenewald:
@ax3l had the idea that instead of prefixing the function name by the class type, the argument types for each individual `initialize_inputs` function could be specified (see https://docs.python.org/3/library/typing.html) which would similarly allow for the identification of the function definition.
That might be a cleaner solution, what do you think @dpgrote?

dpgrote:
> @ax3l had the idea that instead of prefixing the function name by the class type, the argument types for each individual `initialize_inputs` function could be specified (see https://docs.python.org/3/library/typing.html) which would similarly allow for the identification of the function definition. That might be a cleaner solution, what do you think @dpgrote?

@roelof-groenewald Unfortunately, that doesn't solve the issue that when editing the file, I have to search around to find the right routine I'm looking for, since I have to scan over all of the routines of the same name. Also, some of the routines from different categories don't have any arguments.

roelof-groenewald (approved):
As discussed we will merge this now and in a follow-up PR (at some later time) add type hints as well.

Issue #4373, opened 2023-10-13
Particle selection is ignored in BTD diagnostics


RemiLehe:
`<diag_name>.<species_name>.plot_filter_function(t,x,y,z,ux,uy,uz)` is ignored in BTD diagnostics:
https://github.com/ECP-WarpX/WarpX/blob/development/Source/Diagnostics/WarpXOpenPMD.cpp#L589

We should fix this in the future.

RemiLehe:
Ah, actually I think that filters are taken into account for BTD, but it is in a different part of the code:
https://github.com/ECP-WarpX/WarpX/blob/development/Source/Diagnostics/ComputeDiagFunctors/BackTransformParticleFunctor.cpp#L140

RemiLehe:
OK, after more careful consideration, the particle "selection" at https://github.com/ECP-WarpX/WarpX/blob/development/Source/Diagnostics/ComputeDiagFunctors/BackTransformParticleFunctor.cpp#L140 is only performing the selection associated with the boosted-to-lab frame transform (i.e. selecting particles at a specific `z` slice that corresponds to the lab-frame snapshot that is to be written in the output files).

It does not apply filters like `plot_filter_function` and random sampling.

PR #4570, opened 2024-01-03
Apply particle selection to BTDiagnostics

RemiLehe:
This PR addresses the fact that particles selection (e.g. through `plot_filter_function`, or `random_fraction`) is not being applied in the BTDiagnostics (see #4373).

I added corresponding test, whereby beam particles are sub-selected (using `random_fraction`). The test should fail on the `dev` branch, but passes with this PR.


ax3l (approved):
Looks good to me, thank you!

(Took a look further up the function that now checks `pinned_pc->isDefined`, so this must have been outdated.)

Did you test this on a GPU run? :)

RemiLehe:
Thanks for reviewing. 
Yes, I just ran the new automated test on GPU, and it does pass.

ax3l:
Memo for ourselves, we forgot to update `m_totalParticles_flushed_already`, which is done outside of the filter routines.

Tracking in  #4604

PR #4572, opened 2024-01-03
Update dsmc and fusion binary collisions

dpgrote:
This PR makes two updates:

For the DSMC test cast Python_dsmc_1d, this now sets the numpy random number seed, to ensure consistent results.

For DSMC and fusion, when calculating the probability, the code now uses `std::expm1`, which ensures accurate results even for small exponents. This is a simplification of the code.


RemiLehe (approved):

roelof-groenewald (approved):
Thanks for fixing! I'm surprised the checksum values do not need to be updated. Or did you somehow set the random seed to be the same as what it was before?

PR #4573, opened 2024-01-03
Perlmutter: Load `gpu` Module

ax3l:
If a user loaded the `cpu` module, they might lack the CUDA modules. This makes the load a bit more robust, in lieu of a working `module purge` on Cray/HPE systems.


kngott:
Definitely don't `module purge`. :)

And, yes, cpu will turn a lot of stuff off (`module show cpu` to see the details).

I think this'll be more robust, but it won't cover all cases, so a bit of band-aid. Just be aware.

kngott (approved):

PR #4575, opened 2024-01-04
Scale fields by a dimensionless number between 0 and 1

RemiLehe:
It seems that, for rigidly injected particles, fields are being scaled by a time (in SI units, i.e. seconds), instead of a ratio of times (i.e. the ratio of the time spent to the right of the injecting plane, divided by the timestep).

This PR fixes this issue.

### DIFF
--- a/Source/Particles/Gather/ScaleFields.H
+++ b/Source/Particles/Gather/ScaleFields.H
@@ -47,7 +47,7 @@ struct ScaleFields
         // This only approximates what should be happening. The particles
         // should by advanced a fraction of a time step instead.
         // Scaling the fields is much easier and may be good enough.
-        const amrex::Real dtscale = 1 - (m_z_plane_previous - zp)/(m_vz_ave_boosted + m_v_boost)/m_dt;
+        const amrex::Real dtscale = 1._rt - (m_z_plane_previous - zp)/(m_vz_ave_boosted + m_v_boost)/m_dt;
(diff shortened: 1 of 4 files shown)


Source/Particles/Gather/ScaleFields.H
```diff
@@ -47,8 +47,8 @@ struct ScaleFields
         // This only approximates what should be happening. The particles
         // should by advanced a fraction of a time step instead.
         // Scaling the fields is much easier and may be good enough.
-        const amrex::Real dtscale = m_dt - (m_z_plane_previous - zp)/(m_vz_ave_boosted + m_v_boost);
-        if (0._rt < dtscale && dtscale < m_dt)
+        const amrex::Real dtscale = 1 - (m_z_plane_previous - zp)/(m_vz_ave_boosted + m_v_boost)/m_dt;
```
RemiLehe:
```suggestion
        const amrex::Real dtscale = 1._rt - (m_z_plane_previous - zp)/(m_vz_ave_boosted + m_v_boost)/m_dt;
```

dpgrote (approved):
Thanks for catching and fixing this!

jlvay:
Here is an input script that was used to identify the issue as well as a couple of plots that shows the effect.
[inputs-ref.txt](https://github.com/ECP-WarpX/WarpX/files/13835881/inputs-ref.txt)
- Before the fix
Before_fix
- After the fix
After_fix

PR #4576, opened 2024-01-05
Reduce tolerance for DSMC CI test

roelof-groenewald:
After #4572 updated the DSMC test to have a fixed random seed we can reduce the tolerance of that CI test so that it is more sensitive to breaking changes.

- [x] Update benchmark with new Azure values.


Examples/Physics_applications/capacitive_discharge/analysis_dsmc.py
```diff
@@ -57,4 +55,4 @@
 
 density_data = np.load( 'ion_density_case_1.npy' )
 print(repr(density_data))
-assert np.allclose(density_data, ref_density, rtol=0.01)
+assert np.allclose(density_data, ref_density, rtol=1e-3)
```
dpgrote: How small can `rtol` be? The density should be exactly reproducible.

roelof-groenewald: I'll leave it to the default (1e-5) and we can see if it that causes any issues.

dpgrote (approved):
Looks good!

PR #4577, opened 2024-01-05
Optimizing GPU acceleration of binary collision algorithms.

mhaseeb123:
[Edit by @RemiLehe]: The aim of this PR is to speed up binary collisions on GPU by exposing more parallelism: instead of looping with one GPU thread per cell, we loop with one GPU thread per "number of independent pair" (i.e. pairs that do not touch the same macroparticles, so that there is no race condition), where the number of independent pairs is determined by the lower number of macroparticle of either species, within each cell.

 [Updated]: This PR includes optimized GPU implementation for the two particle collision algorithms (Coulomb and Nuclear). 

## Average Performance Improvement: 4x

Results for a specific collisions-heavy use-case from Dave [inputs_1d_H1_lassen2.txt](https://github.com/ECP-WarpX/WarpX/files/14030903/inputs_1d_H1_lassen2.txt)

Built with: `cmake -DWarpX_DIMS=1 -DWarpX_COMPUTE=CUDA` on Perlmutter

New kernel:
```bash
MultiParticleContainer::doCollisions()                   100      11.89      11.89      11.89  81.47%
Avg. per step = 0.1375590809 s
Total Time                     : 14.58777054
```
WarpX dev branch:
```bash
MultiParticleContainer::doCollisions()                   100      67.96      67.96      67.96  96.49%
Avg. per step = 0.6979252147
Total Time                     : 70.43095279
```

## Error Rate 
Slightly higher for a few Azure tests leading to failure of CI pipelines. All good otherwise.

TODO (@RemiLehe):
- [ ] Add some more comments
- [ ] Benchmark on CPU + possibly use `#ifdef` to use the new code only with GPU


RemiLehe:
Thanks for this PR!
I tried to run the `Deuterium_Tritium_Fusion_3D` locally (on a Macbook), with:
```
./run_test.sh Deuterium_Tritium_Fusion_3D
```
but I got the following error:
```
--- INFO    : Writing plotfile Deuterium_Tritium_Fusion_3D_plt000000
STEP 1 starts ...

   WARNING: Test stderr:
SIGILL Invalid, privileged, or ill-formed instruction
SIGILL Invalid, privileged, or ill-formed instruction
See Backtrace.1.0 file for details
Abort(4) on node 0 (rank 0 in comm 496): application called MPI_Abort(comm=0x84000002, 4) - process 0
Abort(4) on node 1 (rank 1 in comm 496): application called MPI_Abort(comm=0x84000001, 4) - process 1
```

Note that this error only appears when using `amrex.fpe_trap_invalid=1` (which is automatically used with the `run_test.sh` utility)

RemiLehe:
Nevermind: it looks like the above also occurs on the `development` branch.

.gitignore
```diff
@@ -5,13 +5,15 @@ d/
 f/
 o/
 build/
+diag*/
```
RemiLehe:
```suggestion
```

.gitignore
```diff
@@ -5,13 +5,15 @@ d/
...
 tmp_build_dir/
 test_dir
 test_dir/
 AMReX_buildInfo.cpp
 Backtrace.*
 *.vth
 *.sw[opqrs]
+warpx_used_inputs*
```
RemiLehe:
```suggestion
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -306,6 +306,31 @@ public:
                                                     p_n_pairs_in_each_cell, pair_offsets.data());
             index_type* AMREX_RESTRICT p_pair_offsets = pair_offsets.dataPtr();
 
+            /*
+               preprocessing calculations for efficient gpu acceleration
+            */
+            amrex::Gpu::DeviceVector<index_type> n_ind_pairs_in_each_cell(n_cells+1);
```
RemiLehe, line 328:
```suggestion
            // Compute the number of independent pairs in each cell
            // (i.e. number of pairs that can be processed in parallel without race conditions)
            amrex::Gpu::DeviceVector<index_type> n_ind_pairs_in_each_cell(n_cells+1);
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -306,6 +306,31 @@ public:
                                                     p_n_pairs_in_each_cell, pair_offsets.data());
             index_type* AMREX_RESTRICT p_pair_offsets = pair_offsets.dataPtr();
```
RemiLehe, line 326: Are the structures `p_n_pairs_in_each_cell` still used?

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -306,6 +306,31 @@ public:
...
+
+            // start indices of independent collisions.
+            amrex::Gpu::DeviceVector<index_type> coll_offsets(n_cells+1);
+            // number of total independent collision pairs
+            const auto n_independent_pairs =  (int) amrex::Scan::ExclusiveSum(n_cells+1,
+                                                    p_n_ind_pairs_in_each_cell, coll_offsets.data(), amrex::Scan::RetSum{true});
+            index_type* AMREX_RESTRICT p_coll_offsets = coll_offsets.dataPtr();
+            /*
+               end of preprocessing calculations
+            */
```
```suggestion
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -306,6 +306,31 @@ public:
                                                     p_n_pairs_in_each_cell, pair_offsets.data());
             index_type* AMREX_RESTRICT p_pair_offsets = pair_offsets.dataPtr();
 
+            /*
+               preprocessing calculations for efficient gpu acceleration
+            */
```
```suggestion
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -324,7 +349,6 @@ public:
               End of calculations only required when creating product particles
             */
 
-
             // Loop over cells
```
RemiLehe, line 365:
```suggestion
            // Loop over cells and shuffle the particle order within each cell
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -335,31 +359,98 @@ public:
...
                     // Do not collide if there is only one particle in the cell
                     if ( cell_stop_1 - cell_start_1 <= 1 ) { return; }
 
                     // shuffle
                     ShuffleFisherYates(
                         indices_1, cell_start_1, cell_half_1, engine );
+                }
+            );
+
+            // Loop over cells
```
```suggestion
            // Loop over independent pairs
            // (Pairs that can be processed in parallel without race conditions)
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -335,31 +359,98 @@ public:
                     index_type const cell_stop_1  = cell_offsets_1[i_cell+1];
                     index_type const cell_half_1 = (cell_start_1+cell_stop_1)/2;
 
-                    // Same but for the pairs
-                    index_type const cell_start_pair = have_product_species?
-                                                           p_pair_offsets[i_cell] : 0;
-
                     // Do not collide if there is only one particle in the cell
```
RemiLehe, line 375:
```suggestion
                    // Do not shuffle if there is only one particle in the cell
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -335,31 +359,98 @@ public:
...
+            );
+
+            // Loop over cells
+            amrex::ParallelForRNG( n_independent_pairs,
+                [=] AMREX_GPU_DEVICE (int i_coll, amrex::RandomEngine const& engine) noexcept
+                {
+                    // to avoid type mismatch errors
+                    auto ui_coll = (index_type)i_coll;
+
+                    // compute i_cell = upper_bound(p_coll_offsets[0:n_independent_
```
```suggestion
                    // Use a bisection algorithm to find the index of the cell in which this pair is located
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -335,31 +359,98 @@ public:
...
+                    int i_cell = 0;
+                    {
+                        int left = 0;
+                        int right = n_cells;
+
+                        for (;;)
```
RemiLehe: Could this be written with a `while` loop, which is slightly more readable?
Alternatively, could we use `amrex`'s implementation of a binary search.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -335,31 +359,98 @@ public:
...
+                        }
+                    }
+
+                    // The particles from species1 that are in the cell `i_cell` are
+                    // given by the `indices_1[cell_start_1:cell_stop_1]`
+                    index_type const cell_start_1 = cell_offsets_1[i_cell];
+                    index_type const cell_stop_1  = cell_offsets_1[i_cell+1];
+                    index_type const cell_half_1 = (cell_start_1+cell_stop_1)/2;
+
+                    // collision number of the cell
```
RemiLehe, line 406:
```suggestion
                    // collision index within the cell
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -335,31 +359,98 @@ public:
...
+                            q1, q1, m1, m1, dt, dV, coll_idx,
+                            cell_start_pair, p_mask, p_pair_indices_1, p_pair_indices_2,
+                            p_pair_reaction_weight, engine);
+                    }
+                    else
+                    {
+                        // Call the function in order to perform collisions
+                        // If there are product species, mask, p_pair_indices_1/2, and
+                        // p_pair_reaction_weight are filled here
+                        binary_collision_functor( // NOLINT(readability-suspicious-call-argument)
```
RemiLehe: Could you explain why the `NOLINT` is needed here?

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -335,31 +359,98 @@ public:
...
+                    // The particles from species1 that are in the cell `i_cell` are
+                    // given by the `indices_1[cell_start_1:cell_stop_1]`
+                    index_type const cell_start_1 = cell_offsets_1[i_cell];
+                    index_type const cell_stop_1  = cell_offsets_1[i_cell+1];
+                    index_type const cell_half_1 = (cell_start_1+cell_stop_1)/2;
+
+                    // collision number of the cell
+                    const index_type coll_idx = ui_coll - p_coll_offsets[i_cell];
+
+                    // Same but for the pairs
```
RemiLehe, line 409:
```suggestion
                    // Starting index for the product species, for this cell
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -335,31 +359,98 @@ public:
...
-                        cell_start_1, cell_half_1,
-                        cell_half_1, cell_stop_1,
-                        indices_1, indices_1,
-                        soa_1, soa_1, get_position_1, get_position_1,
-                        q1, q1, m1, m1, dt, dV,
-                        cell_start_pair, p_mask, p_pair_indices_1, p_pair_indices_2,
-                        p_pair_reaction_weight, engine );
+                    // we will fix species_1 to be the larger set. i.e., max(NI1, NI2)
+                    // to reduce redundant computations and extra params in the functor
+                    if ((cell_half_1 - cell_start_1) > (cell_stop_1 - cell_half_1))
```
RemiLehe: My understanding is that, because `cell_half_1 = (cell_start_1+cell_stop_1)/2`, we have are always in the `else` branch of this `if` condition. Could we therefore get rid of the `if` condition here?

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
mhaseeb123: I think I saw one of the clang-format checks to fail on this so I put it here. If it is passing, it can be removed.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
mhaseeb123: Yes of course, if `amrex` implements the same binary search logic, it can definitely be replaced. Yes, it can be replaced with a `while(left < right)` loop or the same condition could be added here as well.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
mhaseeb123: In the tests that I ran, the number of pair-particles is zero so a bunch of code doesn't execute at all. Can't say if there aren't any cases where this will be valid.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -458,6 +522,46 @@ public:
                                                     p_n_pairs_in_each_cell, pair_offsets.data());
             index_type* AMREX_RESTRICT p_pair_offsets = pair_offsets.dataPtr();
 
+            /*
+               preprocessing calculations for efficient gpu acceleration
+            */
```
```suggestion
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -458,6 +522,46 @@ public:
...
+
+            // start indices of independent collisions.
+            amrex::Gpu::DeviceVector<index_type> coll_offsets(n_cells+1);
+            // number of total independent collision pairs
+            const auto n_independent_pairs = (int) amrex::Scan::ExclusiveSum(n_cells+1,
+                                                    p_n_ind_pairs_in_each_cell, coll_offsets.data(), amrex::Scan::RetSum{true});
+            index_type* AMREX_RESTRICT p_coll_offsets = coll_offsets.dataPtr();
+            /*
+               end of preprocessing calculations
+            */
```
```suggestion
```

Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
```diff
@@ -52,10 +53,14 @@ void ElasticCollisionPerez (
     T_PR const  T1, T_PR const  T2,
     T_R const  dt, T_PR const   L, T_R const dV,
     amrex::RandomEngine const& engine,
-    bool const isSameSpecies)
+    bool const isSameSpecies, T_index coll_idx)
```
RemiLehe, line 56: Maybe this variable could be called `ind_pair_idx`

Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
```diff
@@ -126,10 +131,17 @@ void ElasticCollisionPerez (
...
-      {
+      T_index i1 = I1s + coll_idx;
+      const T_index i2 = I2s + coll_idx;
 
+      // NI2 is always the smaller set
+      const auto stride = NI2;
+
+      // we will start from collision number = coll_idx and then add
+      // stride (smaller set size) until we do all NI1 (laarger set size)
+      // collisions as needed
```
RemiLehe: Maybe we can add comment to explain the method and the fact that this avoids race conditions

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
mhaseeb123: That does make sense and logically could be removed to avoid redundant code. Would recommend adding an explanation in comments for future purposes!

Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
```diff
@@ -52,10 +53,14 @@ void ElasticCollisionPerez (
...
     amrex::RandomEngine const& engine,
-    bool const isSameSpecies)
+    bool const isSameSpecies, T_index coll_idx)
 {
     const T_index NI1 = I1e - I1s;
     const T_index NI2 = I2e - I2s;
+    const T_index max_N = amrex::max(NI1,NI2);
+
+    // ensure this
+    AMREX_ASSERT(max_N == NI1);
```
RemiLehe: Maybe we don't need the `ASSERT` here: instead, we could also define `min_N`, and then use it instead of `NI2` everywhere below.

Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
```diff
@@ -126,10 +131,17 @@ void ElasticCollisionPerez (
...
     // call UpdateMomentumPerezElastic()
     {
-      T_index i1 = I1s; T_index i2 = I2s;
-      for (T_index k = 0; k < amrex::max(NI1,NI2); ++k)
-      {
+      T_index i1 = I1s + coll_idx;
+      const T_index i2 = I2s + coll_idx;
 
+      // NI2 is always the smaller set
+      const auto stride = NI2;
```
RemiLehe: This could be `stride = min_N`

Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
```diff
@@ -159,8 +171,8 @@ void ElasticCollisionPerez (
           u1y[I1[i1]] = u1xbuf_new*std::sin(-theta) + u1y[I1[i1]]*std::cos(-theta);
 #endif
 
-          ++i1; if ( i1 == I1e ) { i1 = I1s; }
-          ++i2; if ( i2 == I2e ) { i2 = I2s; }
+          // update the i1
+          i1 += stride;
```
RemiLehe: This could be: 
```
if (max_N == NI1) i1 += min_N;
if (max_N == NI2) i2 += min_N;
```

RemiLehe:
Out of curiosity: did the code become slower on CPU, as a consequence of these changes?

Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
mhaseeb123: I think this is ok because we only need to update `i1` since `i2` is fixed. `NI1` will always be the larger set.

mhaseeb123: Note that doing this change will also lead to updating i2/i1 which was declared as const a few lines above

roelof-groenewald:
These changes are awesome! Thanks for the effort! 🎉 👍 🚀 

I updated the DSMC module to incorporate these improvements (see the PR mentioned above and https://github.com/mhaseeb123/WarpX/pull/1) and compared the time spent on binary collisions before and after the update for different cases of the Turner (2013) benchmark (see docs [here](https://warpx.readthedocs.io/en/latest/usage/examples/capacitive_discharge/README.html#visualize)).

|         |  NPPC  | #steps | Before this PR (s) | With this PR (s)  | Speed-up (%) |
| :---    | :----: | :----: |    :----:    |    :----:    |   :----:   |
| Case 1 |  512   | 2560   |    16.995    |    4.277     |   397     |
| Case 2 |  256   | 5120   |    33.181    |    9.061     |   366     |
| Case 3 |  128   | 10240  |    36.470    |   12.595     |   290     |

Unsurprisingly the speed-up is better for more particles-per-cell. With the higher NPPC cases we see the same ~4x improvement seen before for Coulomb collisions. These NPPC ranges are in-line with what we generally run at so we can expect a very nice speed-up for "standard" simulations.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -337,16 +383,38 @@ public:
...
+
+            // Loop over cells
+            amrex::ParallelForRNG( n_independent_pairs,
+                [=] AMREX_GPU_DEVICE (int i_coll, amrex::RandomEngine const& engine) noexcept
+                {
+                    // to avoid type mismatch errors
+                    auto ui_coll = (index_type)i_coll;
+
+                    // Use a bisection algorithm to find the index of the cell in which this pair is located
+                    int i_cell = amrex::bisect( p_coll_offsets, 0, n_cells, ui_coll );
```
roelof-groenewald: To fix `clang-tidy` error:
```suggestion
                    const int i_cell = amrex::bisect( p_coll_offsets, 0, n_cells, ui_coll );
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -505,6 +609,38 @@ public:
...
+
+            // Loop over independent particle pairs
+            amrex::ParallelForRNG( n_independent_pairs,
+                [=] AMREX_GPU_DEVICE (int i_coll, amrex::RandomEngine const& engine) noexcept
+                {
+                    // to avoid type mismatch errors
+                    auto ui_coll = (index_type)i_coll;
+
+                    // Use a bisection algorithm to find the index of the cell in which this pair is located
+                    int i_cell = amrex::bisect( p_coll_offsets, 0, n_cells, ui_coll );
```
roelof-groenewald: To fix `clang-tidy` error:
```suggestion
                    const int i_cell = amrex::bisect( p_coll_offsets, 0, n_cells, ui_coll );
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -460,6 +527,46 @@ public:
...
+
+            amrex::ParallelFor( n_cells+1,
+                [=] AMREX_GPU_DEVICE (int i_cell) noexcept
+                {
+                    if (i_cell < n_cells)
+                    {
+                        const auto n_part_in_cell_1 = cell_offsets_1[i_cell+1] - cell_offsets_1[i_cell];
+                        const auto n_part_in_cell_2 = cell_offsets_2[i_cell+1] - cell_offsets_2[i_cell];
+                        // Particular case: no pair if a species has no particle in that cell
+                        if (n_part_in_cell_1 == 0 || n_part_in_cell_2 == 0)
```
dpgrote: Is this `if` needed, since it would get the same result using the `min`?

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -199,6 +217,10 @@ public:
                     amrex::HostDevice::Atomic::Add( &(*cost)[mfi.index()], wt);
                 }
             }
+
+            // Call redistribute to remove particles with negative ids
+            species1.Redistribute(lev, lev, 0, true, true);
```
dpgrote: Is this always needed? We should know whether collisions might remove particles.

RemiLehe: As discussed with @dpgrote, we could add a flag (known at compilation type) checking whether this type of collision can remove particles. `Redistribute` would only need to be called in that case.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
RemiLehe: Agreed.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
roelof-groenewald: Good point! Any collision that has product species would be capable of removing particles, so we can simply wrap this in a 
```
if (m_have_product_species) {
    ...
}
```

Btw, this change came in with #4744 so it might be better to review (and merge) that PR before reviewing this one too much further.

roelof-groenewald: I updated this in the other PR.

Source/Particles/Collision/BinaryCollision/NuclearFusion/NuclearFusionFunc.H
```diff
@@ -174,8 +174,12 @@ public:
             amrex::ParticleReal * const AMREX_RESTRICT theta1 = soa_1.m_rdata[PIdx::theta];
             amrex::ParticleReal * const AMREX_RESTRICT theta2 = soa_2.m_rdata[PIdx::theta];
 #endif
-
-            for (index_type k = 0; k < max_N; ++k)
+            index_type i1 = I1s + coll_idx;
+            index_type i2 = I2s + coll_idx;
+            // we will start from collision number = coll_idx and then add
+            // stride (smaller set size) until we do all NI1 (laarger set size)
+            // collisions as needed
```
RemiLehe:
```suggestion
            // we will start from collision number = coll_idx and then add
            // stride (smaller set size) until we do all collisions (larger set size)
```

Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.H
```diff
@@ -135,8 +135,12 @@ public:
             amrex::ParticleReal * const AMREX_RESTRICT theta1 = soa_1.m_rdata[PIdx::theta];
             amrex::ParticleReal * const AMREX_RESTRICT theta2 = soa_2.m_rdata[PIdx::theta];
 #endif
-
-            for (index_type k = 0; k < max_N; ++k)
+            index_type i1 = I1s + coll_idx;
+            index_type i2 = I2s + coll_idx;
+            // we will start from collision number = coll_idx and then add
+            // stride (smaller set size) until we do all NI1 (laarger set size)
+            // collisions as needed
```
RemiLehe:
```suggestion
            // we will start from collision number = coll_idx and then add
            // stride (smaller set size) until we do all collisions (larger set size)
```

Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
```diff
@@ -126,10 +129,14 @@ void ElasticCollisionPerez (
...
     {
-      T_index i1 = I1s; T_index i2 = I2s;
-      for (T_index k = 0; k < amrex::max(NI1,NI2); ++k)
-      {
+      T_index i1 = I1s + coll_idx;
+      T_index i2 = I2s + coll_idx;
 
+      // we will start from collision number = coll_idx and then add
+      // stride (smaller set size) until we do all NI1 (laarger set size)
+      // collisions as needed
```
RemiLehe:
```suggestion
      // we will start from collision number = coll_idx and then add
      // stride (smaller set size) until we do all collisions (larger set size)
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -522,6 +584,38 @@ public:
                     // shuffle
                     ShuffleFisherYates(indices_1, cell_start_1, cell_stop_1, engine);
                     ShuffleFisherYates(indices_2, cell_start_2, cell_stop_2, engine);
+                }
+            );
+
+            // Loop over independent particle pairs
```
RemiLehe, line 596:
```suggestion
            // Loop over independent particle pairs
            // To speed up binary collisions on GPU, we try to expose as much parallelism 
            // as possible (while avoiding race conditions): Instead of looping with one GPU 
            // thread per cell, we loop with one GPU thread per "independent pairs" (i.e. pairs 
            // that do not touch the same macroparticles, so that there is no race condition), 
            // where the number of independent pairs is determined by the lower number of 
            // macroparticles of either species, within each cell.
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -354,16 +372,38 @@ public:
...
                     // Do not collide if there is only one particle in the cell
                     if ( cell_stop_1 - cell_start_1 <= 1 ) { return; }
 
                     // shuffle
                     ShuffleFisherYates(
                         indices_1, cell_start_1, cell_half_1, engine );
+                }
+            );
+
+            // Loop over cells
```
```suggestion
            // Loop over independent particle pairs
            // To speed up binary collisions on GPU, we try to expose as much parallelism 
            // as possible (while avoiding race conditions): Instead of looping with one GPU 
            // thread per cell, we loop with one GPU thread per "independent pairs" (i.e. pairs 
            // that do not touch the same macroparticles, so that there is no race condition), 
            // where the number of independent pairs is determined by the lower number of 
            // macroparticles of either species, within each cell.
```

roelof-groenewald:
@RemiLehe I ran the CPU performance tests as we discussed (using cases 1 - 3 of the Turner benchmarks). The short and good news is that the results matched expectation and performance was NOT worse on CPU due to these changes. So, as discussed, this PR can be merged 🎉 

Here are the inclusive timings (note that `MultiParticleContainer::doCollisions()` include both MCC and DSMC collisions involved in these tests, but we don't have a timer for just DSMC):
* test 1:
  * devel: `MultiParticleContainer::doCollisions()     512000      3.723      52.39      145.4  13.71%`
  * this PR: `MultiParticleContainer::doCollisions()   512000      3.155      35.78      124.8  12.19%`

* test 2:
  * devel: `MultiParticleContainer::doCollisions()     4096000      390.6      636.2        796  13.30%`
  * this PR: `MultiParticleContainer::doCollisions()   4096000      275.8        523      746.5  12.48%`

* test 3:
  * devel: `MultiParticleContainer::doCollisions()      8192000      721.1       2407       3471  19.19%`
  * this PR: `MultiParticleContainer::doCollisions()    8192000        497       2099       3287  17.91%`

roelof-groenewald:
Here are also the physics results (for posterity) with the changes in this PR included:
image

RemiLehe:
@roelof-groenewald Thanks a lot for doing these tests! This is great news!

RemiLehe (approved):

PR #4546, opened 2023-12-20
Beam-beam collision example

aeriforme:
This PR adds a beam-beam collision example to the documentation and to the `Physics_applications` directory. 

Comparisons with the literature are also shown. 

beambeamcollexample

PR #4578, opened 2024-01-05
Beam-beam collision example follow-up

aeriforme:
This updates and refines the beam-beam collision example #4546.


RemiLehe (approved):
Thanks for updating this!

ax3l:
@aeriforme can you please add this update as well in a cleanup? :)
https://github.com/ECP-WarpX/WarpX/pull/4546#discussion_r1443395780

PR #4587, opened 2024-01-08
[WIP] Replace sin(theta), cos(theta) with amrex::Math::sincos(theta) when appropriate

lucafedeli88:
This PR mirrors what was done for ImpactX in https://github.com/ECP-WarpX/impactx/pull/493 by @ax3l


ax3l (dismissed):
Thanks, LGTM! :+1: 

Note that as linked in the other PR, and tested when I added the function to AMReX, most optimizing compilers should find this automatically. But this way it is done also in all optimization levels, and cleaner if the results are needed multiple times.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1815,11 +1816,10 @@ PhysicalParticleContainer::AddPlasmaFlux (PlasmaInjector const& plasma_injector,
...
                 // choose it randomly.
-                const Real theta = (nmodes == 1 && rz_random_theta)?
-                    (2._prt*MathConst::pi*amrex::Random(engine)):
-                    (2._prt*MathConst::pi*r.y);
-                Real const cos_theta = std::cos(theta);
-                Real const sin_theta = std::sin(theta);
+                const Real tt = (nmodes == 1 && rz_random_theta)?
+                    (2._prt*amrex::Random(engine)):
+                    (2._prt*r.y);
+                auto const [sin_theta, cos_theta] = amrex::Math::sincospi(tt);
```
ax3l:
```suggestion
                const Real theta = (nmodes == 1 && rz_random_theta)?
                    (2._prt*amrex::Random(engine)):
                    (2._prt*r.y);
                auto const [sin_theta, cos_theta] = amrex::Math::sincospi(theta);
```

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1276,11 +1276,12 @@ PhysicalParticleContainer::AddPlasma (PlasmaInjector const& plasma_injector, int
...
                 // choose it randomly.
-                const Real theta = (nmodes == 1 && rz_random_theta)?
-                    (2._rt*MathConst::pi*amrex::Random(engine)):
-                    (2._rt*MathConst::pi*r.y + theta_offset);
-                pos.x = xb*std::cos(theta);
-                pos.y = xb*std::sin(theta);
+                const Real tt = (nmodes == 1 && rz_random_theta)?
+                    (2._rt*amrex::Random(engine)):
+                    (2._rt*r.y + tt_offset);
+                auto const [sin_theta, cos_theta] = amrex::Math::sincospi(tt);
```
ax3l:
```suggestion
                const Real theta = (nmodes == 1 && rz_random_theta)?
                    (2._rt*amrex::Random(engine)):
                    (2._rt*r.y + tt_offset);
                auto const [sin_theta, cos_theta] = amrex::Math::sincospi(theta);
```

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1277,10 +1277,11 @@ PhysicalParticleContainer::AddPlasma (PlasmaInjector const& plasma_injector, int
                 // With only 1 mode, the angle doesn't matter so
                 // choose it randomly.
                 const Real theta = (nmodes == 1 && rz_random_theta)?
-                    (2._rt*MathConst::pi*amrex::Random(engine)):
-                    (2._rt*MathConst::pi*r.y + theta_offset);
-                pos.x = xb*std::cos(theta);
-                pos.y = xb*std::sin(theta);
+                    (2._rt*amrex::Random(engine)):
+                    (2._rt*r.y + tt_offset);
```
ax3l, line 1332: Where did the pi go?

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1204,8 +1204,8 @@ PhysicalParticleContainer::AddPlasma (PlasmaInjector const& plasma_injector, int
             const IntVect iv = IntVect(AMREX_D_DECL(i, j, k));
             const auto index = overlap_box.index(iv);
 #ifdef WARPX_DIM_RZ
-            Real theta_offset = 0._rt;
-            if (rz_random_theta) { theta_offset = amrex::Random(engine) * 2._rt * MathConst::pi; }
+            Real tt_offset = 0._rt;
+            if (rz_random_theta) { tt_offset = amrex::Random(engine) * 2._rt; }
```
ax3l, line 1261: Where did the pi go?

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1816,10 +1817,9 @@ PhysicalParticleContainer::AddPlasmaFlux (PlasmaInjector const& plasma_injector,
                 // With only 1 mode, the angle doesn't matter so
                 // choose it randomly.
                 const Real theta = (nmodes == 1 && rz_random_theta)?
-                    (2._prt*MathConst::pi*amrex::Random(engine)):
-                    (2._prt*MathConst::pi*r.y);
-                Real const cos_theta = std::cos(theta);
-                Real const sin_theta = std::sin(theta);
+                    (2._prt*amrex::Random(engine)):
+                    (2._prt*r.y);
```
ax3l, line 1876: Where did the pi go?

Source/Particles/PhysicalParticleContainer.cpp
lucafedeli88: I am using `amrex::Math::sincospi` instead of `amrex::Math::sincos`, so the `pi` is already taken into account in the function. That's why I wanted to use `tt` instead of `theta` as the name of the variable, in order to avoid confusion. However, I can add back the `pi` and use `amrex::Math::sincos` (it may be more readable).

Source/Particles/PhysicalParticleContainer.cpp
lucafedeli88: (see comment above)

Source/Particles/PhysicalParticleContainer.cpp
lucafedeli88: (see comment above)

Source/Particles/PhysicalParticleContainer.cpp
ax3l: Thanks, oh how subtle, I overlooked that.

ax3l:
Restarting CI, all Azure tests failed...?

ax3l (approved):

ax3l:
CI still not happy, huh. Can you please rebase?

Source/Diagnostics/ReducedDiags/BeamRelevant.cpp
```diff
@@ -220,8 +221,9 @@ void BeamRelevant::ComputeDiags (int step)
                 const ParticleReal p_y_mean = p_pos1*p_w;
 #elif defined(WARPX_DIM_RZ)
                 const ParticleReal p_theta = p.rdata(PIdx::theta);
-                const ParticleReal p_x_mean = p_pos0*std::cos(p_theta)*p_w;
-                const ParticleReal p_y_mean = p_pos0*std::sin(p_theta)*p_w;
+                auto const [sin_p_theta, cos_p_theta] = amrex::Math::sincos(p_theta);
+                const ParticleReal p_x_mean = p_pos0*cos_p_theta*p_w;
+                const ParticleReal p_y_mean = p_pos0*sin_p_theta*p_w;
```
RevathiJambunathan: @lucafedeli88 whats the advantage of doing '[sin,cos]=amrex::Math::sincos` and using that to compute px,py versus 
`p_x_mean = p_pos0 * amrex::Math::sin(p_theta) * p_w` ?

RevathiJambunathan (approved):
cool! Thanks for this PR, @lucafedeli88

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -30,6 +30,7 @@
 #include <AMReX_FabArray.H>
 #include <AMReX_GpuQualifiers.H>
 #include <AMReX_IntVect.H>
+#include <AMReX_Math.H>
```
```suggestion
```

Source/Diagnostics/ReducedDiags/BeamRelevant.cpp
lucafedeli88: @RevathiJambunathan, theoretically you could save some time by computing sine and cosine together if you have to use them both. In practice most compilers are able to recognize this optimization by themselves
(see also https://github.com/ECP-WarpX/impactx/pull/493)

ax3l:
Looks like CI is not happy.

Can you rebase @lucafedeli88 ?

ax3l:
There seem to be percent-level changes for momenta in checksum files :thinking:

ax3l:
@lucafedeli88 I think there might be a bug here if the results change.

Potentially, one could systematically pick one test and bisect which file change introduces the differences?

ax3l:
@lucafedeli88 can you rebase this?
Potentially, it is worth submitting this piece wise if it is hard to spot which line changes cause the checksum changes.

lucafedeli88:
> @lucafedeli88 can you rebase this? Potentially, it is worth submitting this piece wise if it is hard to spot which line changes cause the checksum changes.

@ax3l , I've merged the newest version of WarpX into this branch. I'll try to understand which substitution is causing the bug.

lucafedeli88:
> > @lucafedeli88 can you rebase this? Potentially, it is worth submitting this piece wise if it is hard to spot which line changes cause the checksum changes.
> 
> @ax3l , I've merged the newest version of WarpX into this branch. I'll try to understand which substitution is causing the bug.

I am testing the hypothesis that the small discrepancies come from using `sincospi` instead of `sincos(pi*arg)`. The former should be more precise (`sincospi` does not use `pi` internally), while `sincos(pi*arg)` uses `pi` in single or double precision. This could explain discrepancies at roughly machine precision. In case this hypothesis turns out to be correct, I think that we should switch to `sincospi`, since it is more precise and more performant.

lucafedeli88:
Actually not all the discrepancies are small, so maybe there's something else.

lucafedeli88:
Implementing this feature without introducing discrepancies seems to be very difficult, and I am not sure that I will have time to work on that on the foreseeable future. Since the priority of this feature is very low, I will close the PR.

PR #4588, opened 2024-01-08
Fix documentation for diags : coarsening ratio

RevathiJambunathan:
The coarsening ratio implementation averages N (or N+1) points in each direction if the selected coarsening ratio is N.
This is now correctly reflected in the documentation.


ax3l (approved):
Thx! :sparkles:

PR #4597, opened 2024-01-09
Fix plasma injection from file for parallel runs

WeiqunZhang:
The use of amrex::ParallelDescriptor::ReduceBoolOr was incorrect. The logic is also simplified, even though it may not the most efficient way.


dpgrote (approved):

PR #4599, opened 2024-01-10
Output the potential phi on the macroparticles

RemiLehe:
This adds the option for the user to output the field `phi` on the macroparticles.

The fields are gathered **after** particles have been filtered for output. As a consequence, the particles are in pinned memory, when the gather kernel is called, which could be slow if this diagnostic is done frequently and/or on a large number of particles.

The automated test looks at the case of an electrostatic sphere exploding under its own space charge (Coulomb explosition). In this process, electrostatic potential energy (space charge) is converted into kinetic energy (explosion velocity). The test uses `phi` on the particles to compute the potential energy.

TODO in follow-up PRs:
- [ ] Generalize to any field (on a staggered grid)
- [ ] Implement arbitrary shape factors


Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -548,48 +548,6 @@ for (unsigned i = 0, n = particle_diags.size(); i < n; ++i) {
         pinned_pc->make_alike<amrex::PinnedArenaAllocator>() :
         pc->make_alike<amrex::PinnedArenaAllocator>();
 
-    // names of amrex::Real and int particle attributes in SoA data
```
RemiLehe, line 551: I had to move this code further down, so that it comes after `tmp.AddRealComp("phi");`

(This is because the code below reads `tmp.getParticleRuntimeComps();` and sets the output flags accordingly.)

Source/Diagnostics/ParticleDiag/ParticleDiag.H
```diff
@@ -23,6 +23,7 @@ public:
     [[nodiscard]] PinnedMemoryParticleContainer* getPinnedParticleContainer() const { return m_pinned_pc; }
     [[nodiscard]] std::string getSpeciesName() const { return m_name; }
     amrex::Vector<int> m_plot_flags;
+    bool m_plot_phi = false; // Whether to plot the potential phi on the particles
```
RemiLehe:
```suggestion
    bool m_plot_phi = false; // Whether to output the potential phi on the particles
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -631,6 +589,77 @@ for (unsigned i = 0, n = particle_diags.size(); i < n; ++i) {
...
+        // Using pinned PC indicates that the particles are not written at the same physical time (i.e. PIC iteration)
+        // that they were collected. This happens for diagnostics that use buffering (e.g. BackTransformed, BoundaryScraping).
+        // Here `phi` is gathered at the iteration when particles are written (not collected) and is thus mismatched.
+        // To avoid confusion, we raise an error in this case.
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+            use_pinned_pc == false,
+            "Output of the electrostatic potential (phi) on the particles was requested, "
+            "but this is only available with `diag_type = Full`.");
+        tmp.AddRealComp("phi");
+        tmp.defineAllParticleTiles();
```
ax3l: This line should not be needed anymore as of AMReX 24.04 :)

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -631,6 +589,77 @@ for (unsigned i = 0, n = particle_diags.size(); i < n; ++i) {
         particlesConvertUnits(ConvertDirection::SI_to_WarpX, pc, mass);
     }
 
+    // Gather the electrostatic potential (phi) on the macroparticles
+    if ( particle_diags[i].m_plot_phi ) {
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
```
ax3l, line 594: Can you make this if block a function call please so it does separate out nicely and does not make this sub-function longer? I would call it `add_phi_component` or so and also declare it in `ParticleIO.H`?

Source/Particles/ParticleIO.H
```diff
@@ -8,6 +8,8 @@
 #define PARTICLEIO_H_
 
 #include "Particles/WarpXParticleContainer.H"
+#include "Particles/Pusher/GetAndSetPosition.H"
+#include "ablastr/particles/NodalFieldGather.H"
```
ax3l:
```suggestion

#include <ablastr/particles/NodalFieldGather.H>
```

ax3l (commented):
LGTM - just a small separation of the functional block added into a function maybe? :)

Source/Diagnostics/WarpXOpenPMD.cpp
RemiLehe:
```suggestion
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -631,6 +589,76 @@ for (unsigned i = 0, n = particle_diags.size(); i < n; ++i) {
...
+            use_pinned_pc == false,
+            "Output of the electrostatic potential (phi) on the particles was requested, "
+            "but this is only available with `diag_type = Full`.");
+        tmp.AddRealComp("phi");
+        int const phi_index = tmp.getParticleComps().at("phi");
+        auto& warpx = WarpX::GetInstance();
+#ifdef AMREX_USE_OMP
+#pragma omp parallel if (amrex::Gpu::notInLaunchRegion())
+#endif
+        for (int lev=0; lev<=warpx.finestLevel(); lev++) {
```
ax3l, line 613:
```suggestion
        for (int lev=0; lev<=tmp.finestLevel(); lev++) {
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -631,6 +589,76 @@ for (unsigned i = 0, n = particle_diags.size(); i < n; ++i) {
...
+            const amrex::Geometry& geom = warpx.Geom(lev);
+            auto plo = geom.ProbLoArray();
+            auto dxi = geom.InvCellSizeArray();
+            amrex::MultiFab const& phi = warpx.getField( FieldType::phi_fp, lev, 0 );
```
ax3l, line 617: For less overhead, we should add a helper that returns an mf on all levels. -> follow-up pr

ax3l (approved):

PR #4600, opened 2024-01-10
Allow filtering with implicit and semi-implicit solver

dpgrote:
This PR adds a call to filter the E-fields that are applied to the particles when filtering is turned on. The current is already filtered in the call to `SyncCurrentAndRho`. With the filtering of E, the energy conservation is maintained.

Note that it is using the `ApplyFilterJ` routine to apply the filtering to the E-field. Perhaps that routine could be renamed to `ApplyFilterMF` since it is not specific to J.


dpgrote:
Replaced by PR #5086

PR #4601, opened 2024-01-10
Add warpx.nguards input parameter

dpgrote:
This adds the input parameter `warpx.nguards_minimum` which allows overriding the default value of the number of guard cells for the MultiFab arrays.


Docs/source/usage/parameters.rst
```diff
@@ -2037,6 +2037,12 @@ Particle push, charge and current deposition, field gathering
 
     Note that this input parameter is not optional and must always be set in all input files provided that there is at least one particle species (set in input as ``particles.species_names``) or one laser species (set in input as ``lasers.names``) in the simulation. No default value is provided automatically.
 
+* ``warpx.nguards`` (`integers, one per space dimension`, optional)
```
ax3l: Should this be called `warpx.deposition_nguards`, because it only overwrites the deposition arrays?

Or should it also influence the other multi-fabs?

Source/Parallelization/GuardCellManager.cpp
```diff
@@ -139,24 +139,30 @@ guardCellManager::Init (
...
-                    dt_J = 2._rt*dt;
-                }
-                // With multi_J but without time averaging, particles can move during dt per PIC
-                // cycle for the current deposition as well.
-                else {
-                    dt_J = dt;
+        const ParmParse pp_warpx("warpx");
+        bool const nguards_specified = utils::parser::queryWithParser(pp_warpx, "nguards", ng_alloc_Rho);
+        if (nguards_specified) {
+            ng_alloc_J = ng_alloc_Rho;
```
ax3l: If a users sets this parameter *smaller* than the auto-calculcated value, should we throw a warning?
I see users having this parameter in their input and silently corrupting their memory.

Can you slightly restructure for this?

Docs/source/usage/parameters.rst
JustinRayAngus: This is also needed for the fields used in the gather.

dpgrote:
@axel @JustinRayAngus I changed this to now be `nguards_minimum` and applies to all of the MultiFabs. This should be more robust and avoids possible user errors.

EZoni:
Thanks, @dpgrote.

If I understand correctly, this new parameter overrides also the values set with `psatd.nx_guard`, `psatd.ny_guard`, `psatd.nz_guard`. If the parameter is intended to be used only with the implicit solver, should we enforce that in the implementation, i.e., use it only when the implicit solver is also used and ignore it otherwise? Wouldn't this help avoid confusion with other parameters, e.g., the `psatd` ones, when the implicit solver is not used?

Source/Parallelization/GuardCellManager.cpp
```diff
@@ -341,4 +341,24 @@ guardCellManager::Init (
...
+        ng_alloc_G.max(nguards_minimum);
+        ng_FieldSolver.max(nguards_minimum);
+        ng_FieldSolverF.max(nguards_minimum);
+        ng_FieldSolverG.max(nguards_minimum);
+        ng_FieldGather.max(nguards_minimum);
+        ng_UpdateAux.max(nguards_minimum);
+        ng_MovingWindow.max(nguards_minimum);
+        ng_afterPushPSATD.max(nguards_minimum);
+        ng_depos_J.max(nguards_minimum);
+        ng_depos_rho.max(nguards_minimum);
```
JustinRayAngus, line 362: I recently ran into a bug in PR #5768 related to incompatible guard cells. The solution was to set the guard cells for the mass matrices based on current_fp rather than Efield_fp. For the implicit solvers, these should be the same. However, it seems that the guard cells for current_fp is based on speed of light while Efield_fp is based on the gather stencil. For the 1D planar pinch problem, this results in the guard cells for J being 11 for the largest time step considered. That is much larger than needed. 

Would it be possible to set the guard cells for E/B/J (and everything else for that matter) to just be equal to the specified nguards rather than using max()?

dpgrote: See the comment from Axel above. The concern is that the user could set a value smaller than the calculated minimum value leading to out of memory access. Perhaps a separate input parameter could be added that sets nguards as you want, but with warnings that it could lead to code crashes if too small.

Is there a problem that you are seeing with the too many guards cells in your case?

JustinRayAngus: I'm not sure about a problem. Let's say I'm using 4 cells wide for the valid domain. Are there issues that can arise when you set the guard cells to be much larger than the valid cell region? In that case, exchange would need to communicate not just with neighboring procs but others as well. I can imagine there also might be inefficiency with exchange with such a large number of guard cells.

Once the mass matrices become more mature, setting their guard cells will also depend on the particular deposition/gather routine, dimensionality, and the number of cell crossings permitted. 

Maybe we make this new flag implicit specific. What if the user defines the number of cell crossings permitted? Then we could compute the guard cells for E/B/J based on that?

dpgrote: For this case, that would be a good way to do this, have an input parameter setting the number of cell crossings permitted. The number of guard cells would then be set taking this in to account as well as what you listed, the deposition/gather stencils, and dimensionality.

On the first point, the number of guard cells must be smaller than the number of valid cells. AMReX does not handle that case of multiple overlapping guard cells.

dpgrote:
This PR was replaced with the better method implemented in PR #5866

PR #4603, opened 2024-01-10
Doc: Fast, Local Compilation

ax3l:
Workflows for compiling without or with slow internet access or when working on WarpX and its dependencies at the same time.

### DIFF
--- a/Docs/source/developers/local_compile.rst
+++ b/Docs/source/developers/local_compile.rst
@@ -13,7 +13,7 @@ For simplicity, WarpX :ref:`compilation with CMake <building-cmake>` by default 
 
 on-the-fly, which is called a *superbuild*.
 
-In some scenarios, e.g., when compiling without or with slow internet access or when working on WarpX and its dependencies, other strategies might be preferable.
+In some scenarios, e.g., when compiling without internet, with slow internet access, or when working on WarpX and its dependencies, other strategies might be preferable.
 In the below workflows, you as the developer need to make sure to use compatible versions of the dependencies you provide.
 
 
@@ -26,7 +26,7 @@ This workflow is best for developers that make changes to WarpX, AMReX, PICSAR, 
 For instance, use this if you add a feature in AMReX and want to try it in WarpX before it is proposed as a pull request for inclusion in AMReX.
 
 Instead of downloading the source code of the above dependencies, one can also use an already cloned source copy.
-For instance, clone these three dependencies to ``$HOME/src``:
+For instance, clone these dependencies to ``$HOME/src``:
 
 .. code-block:: bash
 
@@ -64,7 +64,7 @@ When building WarpX, :ref:`the following CMake flags <building-cmake-options>` w
 Compiling With Pre-Compiled Dependencies
 ----------------------------------------
 
-This workflow is best the fastest to compile WarpX, when you just want to change code in WarpX and have the above central dependencies already made available *in the right configurations* (e.g., w/ or w/o MPI or GPU support) from a :ref:`module system <install-hpc>` or :ref:`package manager <install-dependencies>`.
+This workflow is the best and fastest to compile WarpX, when you just want to change code in WarpX and have the above central dependencies already made available *in the right configurations* (e.g., w/ or w/o MPI or GPU support) from a :ref:`module system <install-hpc>` or :ref:`package manager <install-dependencies>`.
 
 Instead of downloading the source code of the above central dependencies, or using a local copy of their source, we can compile and install those once and instruct CMake to `find their install locations and configurations <https://hsf-training.github.io/hsf-training-cmake-webpage/09-findingpackages/index.html>`__.
 
--- a/Docs/source/developers/local_compile.rst
+++ b/Docs/source/developers/local_compile.rst
@@ -57,6 +57,7 @@ When building WarpX, :ref:`the following CMake flags <building-cmake-options>` w
      -DWarpX_pybind11_src=$HOME/src/pybind11
 
    cmake --build build -j 8
+   cmake --build build -j 8 --target pip_install
 
 
 .. _developers-local-compile-findpackage:
@@ -86,6 +87,7 @@ WarpX supports this with :ref:`the following CMake flags <building-cmake-options
      -DWarpX_pybind11_internal=OFF
 
    cmake --build build -j 8
+   cmake --build build -j 8 --target pip_install
 
 As a background, this is also the workflow how WarpX is built in :ref:`package managers such as Spack and Conda-Forge <install-dependencies>`.
 
--- a/Docs/source/developers/local_compile.rst
+++ b/Docs/source/developers/local_compile.rst
@@ -87,3 +87,38 @@ WarpX supports this with :ref:`the following CMake flags <building-cmake-options
    cmake --build build -j 8
 
 As a background, this is also the workflow how WarpX is built in :ref:`package managers such as Spack and Conda-Forge <install-dependencies>`.
+
+
+.. _developers-local-compile-pylto:
+
+Faster Python Builds
+--------------------
+
+The Python bindings of WarpX and AMReX (pyAMReX) use `pybind11 <https://pybind11.readthedocs.io>`__.
+Since pybind11 relies heavily on `C++ metaprogramming <https://pybind11.readthedocs.io/en/stable/faq.html#how-can-i-create-smaller-binaries>`__, speeding up the generated binding code requires that we perform a `link-time optimization (LTO) <https://pybind11.readthedocs.io/en/stable/compiling.html#pybind11-add-module>`__ step, also known as `interprocedural optimization (IPO) <https://en.wikipedia.org/wiki/Interprocedural_optimization>`__.
+
+For fast local development cycles, one can skip LTO/IPO with the following flags:
+
+.. code-block:: bash
+
+   cd src/warpx
+
+   cmake -S . -B build       \
+     -DWarpX_PYTHON=ON       \
+     -DWarpX_PYTHON_IPO=OFF  \
+     -DpyAMReX_IPO=OFF
+
+   cmake --build build -j 8 --target pip_install
+
+.. note::
+
+   We might transition to `nanobind <https://github.com/wjakob/nanobind>`__ in the future, which `does not rely on LTO/IPO <https://nanobind.readthedocs.io/en/latest/benchmark.html>`__ for optimal binaries.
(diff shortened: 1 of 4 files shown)


Docs/source/developers/local_compile.rst
```diff
@@ -0,0 +1,89 @@
...
+
+   cmake --build build -j 8
+
+
+.. _developers-local-compile-findpackage:
+
+Compiling With Pre-Compiled Dependencies
+----------------------------------------
+
+This workflow is best the fastest to compile WarpX, when you just want to change code in WarpX and have the above central dependencies already made available *in the right configurations* (e.g., w/ or w/o MPI or GPU support) from a :ref:`module system <install-hpc>` or :ref:`package manager <install-dependencies>`.
```
roelof-groenewald: Looks like there is a typo here. Maybe
```suggestion
This workflow is the best and fastest to compile WarpX, when you just want to change code in WarpX and have the above central dependencies already made available *in the right configurations* (e.g., w/ or w/o MPI or GPU support) from a :ref:`module system <install-hpc>` or :ref:`package manager <install-dependencies>`.
```

Docs/source/developers/local_compile.rst
```diff
@@ -0,0 +1,89 @@
...
+
+* `AMReX <https://amrex-codes.github.io>`__
+* `PICSAR <https://github.com/ECP-WarpX/picsar>`__
+* `openPMD-api <https://github.com/openPMD/openPMD-api>`__
+* `pyAMReX <https://github.com/AMReX-Codes/pyamrex>`__
+* `pybind11 <https://github.com/pybind/pybind11>`__
+
+on-the-fly, which is called a *superbuild*.
+
+In some scenarios, e.g., when compiling without or with slow internet access or when working on WarpX and its dependencies, other strategies might be preferable.
```
dpgrote: I would expand this phrase a bit
```suggestion
In some scenarios, e.g., when compiling without internet, with slow internet access, or when working on WarpX and its dependencies, other strategies might be preferable.
```

Docs/source/developers/local_compile.rst
```diff
@@ -0,0 +1,89 @@
...
+.. _developers-local-compile-findpackage:
+
+Compiling With Pre-Compiled Dependencies
+----------------------------------------
+
+This workflow is best the fastest to compile WarpX, when you just want to change code in WarpX and have the above central dependencies already made available *in the right configurations* (e.g., w/ or w/o MPI or GPU support) from a :ref:`module system <install-hpc>` or :ref:`package manager <install-dependencies>`.
+
+Instead of downloading the source code of the above central dependencies, or using a local copy of their source, we can compile and install those once and instruct CMake to `find their install locations and configurations <https://hsf-training.github.io/hsf-training-cmake-webpage/09-findingpackages/index.html>`__.
+
+WarpX supports this with :ref:`the following CMake flags <building-cmake-options>`:
```
dpgrote, line 73: Can you give an example of how to tell cmake where to find a preinstalled amrex?

Docs/source/developers/local_compile.rst
```diff
@@ -0,0 +1,89 @@
...
+.. _developers-local-compile-src:
+
+Compiling From Local Sources
+----------------------------
+
+This workflow is best for developers that make changes to WarpX, AMReX, PICSAR, openPMD-api and/or pyAMReX at the same time.
+For instance, use this if you add a feature in AMReX and want to try it in WarpX before it is proposed as a pull request for inclusion in AMReX.
+
+Instead of downloading the source code of the above dependencies, one can also use an already cloned source copy.
+For instance, clone these three dependencies to ``$HOME/src``:
```
lucafedeli88:
```suggestion
For instance, clone these dependencies to ``$HOME/src``:
```
they are more than three!

Docs/source/install/cmake.rst
```diff
@@ -147,10 +147,12 @@ Relative paths are also supported, e.g. ``-DWarpX_amrex_src=../amrex``.
 
 Or build against an AMReX feature branch of a colleague.
 Assuming your colleague pushed AMReX to ``https://github.com/WeiqunZhang/amrex/`` in a branch ``new-feature`` then pass to ``cmake`` the arguments: ``-DWarpX_amrex_repo=https://github.com/WeiqunZhang/amrex.git -DWarpX_amrex_branch=new-feature``.
```
lucafedeli88, line 149:
```suggestion
Assuming your colleague pushed AMReX to ``https://github.com/YourColleagueUserName/amrex/`` in a branch ``new-feature`` then pass to ``cmake`` the arguments: ``-DWarpX_amrex_repo=https://github.com/YourColleagueUserName/amrex.git -DWarpX_amrex_branch=new-feature``.
```
I would suggest this change

lucafedeli88 (commented):
I've left a couple of comments

Docs/source/install/cmake.rst
ax3l: But @WeiqunZhang is the most likely one :D

Docs/source/developers/local_compile.rst
ax3l: Yes, but the link one line above references how to do this:
https://hsf-training.github.io/hsf-training-cmake-webpage/09-findingpackages/index.html

Added one more link now.

ax3l: Explicit examples that would also show how to install the dependencies exactly are very painful here (overly concrete or too abstract) due to the many variants that we maintain in a superbuild to match AMReX-pyAMReX-WarpX properly. This is why we recommend to use a package manager or modules for providing them (or simply not doing it and staying with a superbuild, local source and relying on CCache).

roelof-groenewald (approved):
Looks good to me but I'll hold off on merging since others had comments as well.

Docs/source/developers/local_compile.rst
```diff
@@ -0,0 +1,136 @@
...
+When building WarpX, :ref:`the following CMake flags <building-cmake-options>` will use the respective local sources:
+
+.. code-block:: bash
+
+   cd src/warpx
+
+   rm -rf build
+
+   cmake -S . -B build  \
+     -DWarpX_PYTHON=ON  \
```
dpgrote, line 52: This could be potentially confusing since the Python option is not needed because of local compilation. Also, the build comment below is not consistent since it doesn't have the `--target pip_install` option. Maybe leave out the Python related options for simplicity in this example?

Docs/source/developers/local_compile.rst
```diff
@@ -0,0 +1,136 @@
...
+The Python bindings of WarpX and AMReX (pyAMReX) use `pybind11 <https://pybind11.readthedocs.io>`__.
+Since pybind11 relies heavily on `C++ metaprogramming <https://pybind11.readthedocs.io/en/stable/faq.html#how-can-i-create-smaller-binaries>`__, speeding up the generated binding code requires that we perform a `link-time optimization (LTO) <https://pybind11.readthedocs.io/en/stable/compiling.html#pybind11-add-module>`__ step, also known as `interprocedural optimization (IPO) <https://en.wikipedia.org/wiki/Interprocedural_optimization>`__.
+
+For fast local development cycles, one can skip LTO/IPO with the following flags:
+
+.. code-block:: bash
+
+   cd src/warpx
+
+   cmake -S . -B build       \
```
dpgrote, line 109: For completeness here, maybe include the setting of the `_src` or `_internal` options here?

Docs/source/developers/local_compile.rst
```diff
@@ -0,0 +1,136 @@
...
+   cmake --build build -j 8 --target pip_install
+
+.. note::
+
+   We might transition to `nanobind <https://github.com/wjakob/nanobind>`__ in the future, which `does not rely on LTO/IPO <https://nanobind.readthedocs.io/en/latest/benchmark.html>`__ for optimal binaries.
+   You can contribute to `this pyAMReX pull request <https://github.com/AMReX-Codes/pyamrex/pull/127>`__ to help exploring this library (and if it works for the HPC/GPU compilers that we need to support).
+
+For robustness, our ``pip_install`` target performs a regular ``wheel`` build and then installs it with ``pip``.
+This step will check every time of WarpX dependencies are properly installed, to avoid broken installations.
+When developing without internet or after the first ``pip_install`` succeeded in repeated installations in rapid development cycles, this check of ``pip`` can be skipped by using the ``pip_install_nodeps`` target instead:
```
dpgrote, line 123: Maybe also include this comment about `pip_install_nodeps` in `cmake.rst` as well?

Docs/source/install/cmake.rst
lucafedeli88: As you prefer, then ;-)

lucafedeli88 (approved):

Docs/source/developers/local_compile.rst
```diff
@@ -0,0 +1,136 @@
...
+
+   cmake -S . -B build  \
+     -DWarpX_PYTHON=ON  \
+     -DWarpX_amrex_src=$HOME/src/amrex          \
+     -DWarpX_openpmd_src=$HOME/src/openPMD-api  \
+     -DWarpX_picsar_src=$HOME/src/picsar        \
+     -DWarpX_pyamrex_src=$HOME/src/pyamrex      \
+     -DWarpX_pybind11_src=$HOME/src/pybind11
+
+   cmake --build build -j 8
```
ax3l, line 59:
```suggestion
   cmake --build build -j 8
   cmake --build build -j 8 --target pip_install
```

Docs/source/developers/local_compile.rst
ax3l: Good catch, adding `--target pip_install`.

I would show with Python, since we want to show this by default. Also, that enables me to show how to use a local pybind11 and pyamrex.

Developers understand that they can turn it off when they read this part.

Docs/source/developers/local_compile.rst
ax3l: I would not do that - it is either-or, and people can combine with one of the two options above. Also, the hints here also work perfectly fine and fast with the superbuild, too.

Docs/source/developers/local_compile.rst
ax3l: Good point, let me find a good location for this.

Generally, I want to reduce the details in `cmake.rst` that are a workflow and have them only here. Listing options should be complete though (with a brief explanation).

ax3l: Added :)

Docs/source/developers/local_compile.rst
ax3l: Target added :)

Docs/source/developers/local_compile.rst
ax3l: Added :)

dpgrote (approved):
Looks good!

PR #4607, opened 2024-01-12
Fix input of galerkin_scheme

dpgrote:
Move input of galerkin_scheme upward to be before the check of it that is done for the implicit solver.


lucafedeli88 (approved):
Thanks for this fix, @dpgrote

PR #4608, opened 2024-01-12
atomicAdd -> lockAdd

WeiqunZhang:
For WarpX's Gordon Bell runs on Fugaku, Stephan Jaure of ATOS optimized atomicAdd using pthread spin locks. The optimized approach has been implemented in amrex::BaseFab::lockAdd, which is now used in WarpX.

### DIFF
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -115,7 +115,7 @@ jobs:
         which nvcc || echo "nvcc not in PATH!"
 
         git clone https://github.com/AMReX-Codes/amrex.git ../amrex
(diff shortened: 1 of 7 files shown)


lucafedeli88 (approved):
Thanks a lot, @WeiqunZhang

ax3l:
Quick check: did you test this at runtime in debug mode?

I see in ImpactX on OMP devices, when built in `RelWithDebInfo` mode, now double free caused segfaults... (Could be an existing issue that is now elevated...?)

ax3l:
Ah, sorry - false alarm. Mixed the old and new particle layout branches I think.

PR #4614, opened 2024-01-16
Fix undefined behavior in meanParticleVelocity

RemiLehe:
Since the variable `mean_v` was not initialized, the code resulted in undefined behavior in the case `ntot = 0`.

This PR fixes the issue.


WeiqunZhang (approved):

lucafedeli88 (approved):
Thanks for having spotted this!

PR #4620, opened 2024-01-18
Perlmutter (NERSC): January 2024 Update

### LINKED ISSUE #4606: January 17, 2024 NERSC Perlmutter Update
We need to update our NERSC Perlmutter documentation in a week from now, and users/developers will need to recompile their dependencies after this system upgrade:

> Dear NERSC Users:
>
> Updates to the user environment during the next Perlmutter system maintenance, which will mark the allocation year boundary on January 17, 2024, may impact some users. During this maintenance, the default Cray Programming Environment (CPE) will be updated to [23.12](https://github.com/PE-Cray/cpe-changelog/blob/main/ex/cpe-23.12-sles15-sp5-FullReleaseNotes.txt). As a consequence of this update: 
> 
>     The default cudatoolkit module version will be updated to cudatoolkit/12.2. The new version of Cray MPICH in the new CPE requires CUDA 12 for GPU-aware MPI. 

>     The gcc compiler will now be provided by the underlying SLES OS. This means that the native system gcc at /usr/bin/gcc will be updated to version 12.3 and PrgEnv-gnu will now offer a gcc-native module instead of a gcc module.
> 
> 
> We expect most users should not need to re-compile their code, though we recommend testing existing applications in the new environment. 
> 
>     For applications that require CUDA 11 or are incompatible in some other way with the new default environment,  module load cpe/23.03 will restore most of the prior environment.
> 
>     Conda environments with packages from the Anaconda defaults/main channel and locally built packages that are linked with libraries in the CPE may experience compatibility issues with gcc libraries. We recommend using the conda-forge channel which provides a more recent gcc library that is compatible with the new CPE. The conda and python modules will be updated to include the conda-forge channel by default.
> 
>     The nvcc compiler provided by cudatoolkit 11.X modules is incompatible with the updated system gcc, which may be a problem in non-GNU programming environments such as PrgEnv-nvidia. In such cases, it may be helpful to load the gcc-mixed/11.2.0 module to provide a compatible gcc for nvcc to use other than the system gcc.  We expect this will impact only a handful of users.
> 
> 
> Thanks for being a NERSC user, and don’t hesitate to [submit a ticket](https://help.nersc.gov/) if you have any questions!

ax3l:
Update modules of Perlmutter after the system upgrade.

Since the default compilers changed, please run the install of dependencies scripts again and rebuild your WarpX executables, too. Essentially, the order documented here from the top: https://warpx.readthedocs.io/en/latest/install/hpc/perlmutter.html

GPU update example:
```bash
source $HOME/perlmutter_gpu_warpx.profile  # updates in this PR!!

bash $HOME/src/warpx/Tools/machines/perlmutter-nersc/install_gpu_dependencies.sh
source ${CFS}/${proj%_g}/${USER}/sw/perlmutter/gpu/venvs/warpx/bin/activate

cd ~/src/warpx
rm -rf build
...
```

Close #4606

## Details

- System default C/CXX compiler is now GNU 12.3.0
- System default CUDA compiler is now NVCC 12.2.91 (from NVHPC 23.9)
- Both appear to be [compatible](https://gist.github.com/ax3l/9489132).

## Check List

- [x] GPU compiles
- [x] GPU runs
- [x] CPU compiles
- [x] CPU runs

### DIFF
--- a/Tools/machines/perlmutter-nersc/perlmutter_cpu_warpx.profile.example
+++ b/Tools/machines/perlmutter-nersc/perlmutter_cpu_warpx.profile.example
@@ -7,14 +7,14 @@ if [ -z ${proj-} ]; then echo "WARNING: The 'proj' variable is not yet set in yo
 
 # required dependencies
 module load cpu
-module load cmake/3.22.0
-module load cray-fftw/3.3.10.3
+module load cmake/3.24.3
+module load cray-fftw/3.3.10.6
 
 # optional: for QED support with detailed tables
 export BOOST_ROOT=/global/common/software/spackecp/perlmutter/e4s-23.05/default/spack/opt/spack/linux-sles15-zen3/gcc-11.2.0/boost-1.82.0-ow5r5qrgslcwu33grygouajmuluzuzv3
 
 # optional: for openPMD and PSATD+RZ support
-module load cray-hdf5-parallel/1.12.2.7
+module load cray-hdf5-parallel/1.12.2.9
 export CMAKE_PREFIX_PATH=${CFS}/${proj}/${USER}/sw/perlmutter/cpu/c-blosc-1.21.1:$CMAKE_PREFIX_PATH
 export CMAKE_PREFIX_PATH=${CFS}/${proj}/${USER}/sw/perlmutter/cpu/adios2-2.8.3:$CMAKE_PREFIX_PATH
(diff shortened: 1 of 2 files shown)


ax3l:
Follow-up needed to update cupy:
https://github.com/ECP-WarpX/WarpX/pull/4621

PR #4622, opened 2024-01-22
Update `picmi.ElectrostaticSolver` child class in external Poisson solver test

roelof-groenewald:
This is a follow-up to #4554. The external solver tests create child classes of `picmi.ElectrostaticSolver`. The overwrite functions for `initialize_inputs` should be updated to be `solver_initialize_inputs`. This did not break those tests since without the overwrite the standard ES solver was simply used.


dpgrote (approved):
Looks good. Thanks for fixing and sorry about the oversight.

ax3l (approved):

PR #4623, opened 2024-01-22
Implemented Villasenor and Buneman deposition routine for implicit solver

JustinRayAngus:
Implemented Villasenor and Buneman current deposition for implicit scheme. Also implemented the associated gather for energy conservation. This is a charge-conserving deposition. The difference from Esirkepov is that the deposit is done segment by segment, where the segments are determined by cell crossings. In general, this results in a tighter stencil. The implementation is valid for an arbitrary number of cell crossings.


Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -1533,6 +1533,668 @@ void doChargeConservingDepositionShapeNImplicit (const amrex::ParticleReal * con
...
 }
 
+/**
+ * \brief Villasenor and Buneman Current Deposition for thread thread_num for implicit scheme.
+ *        The specifics for the implicit scheme are in how gamma is determined. This is a charge-
+ *        conserving deposition. The difference from Esirkepov is that the deposit is done segment
+ *        by segment, where the segments are determined by cell crossings. In general, this results
+ *        in a tighter stencil. The implementation is valid for an arbitrary number of cell crossings.
+ *
+ * \param depos_order  deposition order
```
dpgrote, line 1545: The list of parameters here should be made consistent with the subroutine arguments.

Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -1533,6 +1533,668 @@ void doChargeConservingDepositionShapeNImplicit (const amrex::ParticleReal * con
...
+                    dxp_seg = x0_new - x0_old;
+                    dyp_seg = y0_new - y0_old;
+                    dzp_seg = z0_new - z0_old;
+
+                    if ( (dyp == 0. || std::abs(dxp_seg) < std::abs(dxp/dyp*dyp_seg))
+                      && (dzp == 0. || std::abs(dxp_seg) < std::abs(dxp/dzp*dzp_seg)) ) {
+                        Xcell = x0_new;
+                        dyp_seg = dyp/dxp*dxp_seg;
+                        dzp_seg = dzp/dxp*dxp_seg;
+                        y0_new = y0_old + dyp_seg;
```
dpgrote, line 1820: Just curious - is there a reason that this was done this way instead of setting the `x0_new` etc after the `if` statement?

dpgrote (approved):
This looks good to me, excepting for the minor comments. I approve, but this should be examined by others before merging.

Source/Particles/Deposition/CurrentDeposition.H
JustinRayAngus: It was a stylistic choice. The first six lines inside the else{} statement mirror that inside the if() statement above.

Source/Particles/Deposition/CurrentDeposition.H
JustinRayAngus: This part was directly copied from the doChargeConservingDepositionShapeNImplicit() routine, which has the same parameter list issue. Should I go ahead and change it in both places?

dpgrote: I'll submit a separate PR to make the other fix. It is generally better to not make changes elsewhere in the code not directly related to the PR. This is PR #4658

Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -1533,6 +1533,670 @@ void doChargeConservingDepositionShapeNImplicit (const amrex::ParticleReal * con
...
+            amrex::Real const rp_new = std::sqrt(xp_new*xp_new + yp_new*yp_new);
+            amrex::Real const rp_old = std::sqrt(xp_old*xp_old + yp_old*yp_old);
+            amrex::Real const rp_mid = (rp_new + rp_old)/2._rt;
+            amrex::Real costheta_mid, sintheta_mid;
+            if (rp_mid > 0._rt) {
+                costheta_mid = xp_mid/rp_mid;
+                sintheta_mid = yp_mid/rp_mid;
+            } else {
+                costheta_mid = 1._rt;
+                sintheta_mid = 0._rt;
```
ax3l, line 1698: Confused by the variable name: does this use a small angle approximation for the `sin()`?

Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -1533,6 +1533,670 @@ void doChargeConservingDepositionShapeNImplicit (const amrex::ParticleReal * con
...
+
+            // compute cell crossings in Z-direction
+            const auto k_old = static_cast<int>(z_old-shift);
+            const auto k_new = static_cast<int>(z_new-shift);
+            int cell_crossings_z = std::abs(k_new-k_old);
+            num_segments += cell_crossings_z;
+
+            // need to assert that the number of cell crossings in each direction
+            // is within the range permitted by the number of guard cells
+            // e.g., if (num_segments > 3) ...
```
ax3l: Debug-mode only assert would be:
```suggestion
            // e.g., if (num_segments > 3) ...
            AMREX_ASSERT_WITH_MESSAGE(
                num_segments < 3,
                "VB Deposition: Detected particle with too many Cell Crossings!"
            );
```

Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -1533,6 +1533,670 @@ void doChargeConservingDepositionShapeNImplicit (const amrex::ParticleReal * con
...
+            // locations in each direction used to find the position at cell crossings.
+            const double dxp = x_new - x_old;
+            const double dzp = z_new - z_old;
+            const auto dirX_sign = static_cast<int>(dxp < 0. ? -1 : 1);
+            const auto dirZ_sign = static_cast<int>(dzp < 0. ? -1 : 1);
+            double Xcell = 0., Zcell = 0.;
+            if (num_segments > 1) {
+                Xcell = i_old + shift + 0.5_rt*(1-dirX_sign);
+                Zcell = k_old + shift + 0.5_rt*(1-dirZ_sign);
+            }
```
ax3l: Idea/optional operator, but I would explicitly cast:
```suggestion
            double Xcell = num_segments > 1 ?
                static_cast<double>(i_old) + shift + 0.5*(1.0-dirX_sign) : 0.;
            double Zcell = num_segments > 1 ?
                static_cast<double>(k_old) + shift + 0.5*(1.0-dirZ_sign) : 0.;
```

Do you want to use `amrex::ParticleReal` (`_prt`) here, which can be `float` and `double`, or do you require `double` here?


Note that `shift` is currently defined as a `Real` (float or double, depending on compile).

Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -1533,6 +1533,670 @@ void doChargeConservingDepositionShapeNImplicit (const amrex::ParticleReal * con
...
+                        Zcell = z0_new;
+                        dxp_seg = dxp/dzp*dzp_seg;
+                        x0_new = x0_old + dxp_seg;
+                    }
+
+                }
+
+                // compute the segment factors (each equal to dt_seg/dt for nonzero dxp, or dzp)
+                const auto seg_factor_x = static_cast<amrex::Real>(dxp == 0. ? 1. : dxp_seg/dxp);
+                const auto seg_factor_z = static_cast<amrex::Real>(dzp == 0. ? 1. : dzp_seg/dzp);
```
ax3l: We usually use `amrex::ParticleReal` for particle quanitites and `amrex::Real` for field quantities.

For depositions, there is a natural shift from one to another, just need to pick where.

Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -1533,6 +1533,670 @@ void doChargeConservingDepositionShapeNImplicit (const amrex::ParticleReal * con
...
+            num_segments += cell_crossings_z;
+
+            // need to assert that the number of cell crossings in each direction
+            // is within the range permitted by the number of guard cells
+            // e.g., if (num_segments > 2) ...
+
+            // compute dzp and the initial cell location used to find the cell crossings.
+            double const dzp = z_new - z_old;
+            const auto dirZ_sign = static_cast<int>(dzp < 0. ? -1 : 1);
+            double Zcell = k_old + shift + 0.5_rt*(1-dirZ_sign);
```
ax3l: as above, here I would explicitly cast the int to avoid surprises:
```suggestion
            double Zcell = static_cast<double>(k_old) + shift + 0.5*(1.0-dirZ_sign);
```

also here: consistently Real/ParticleReal or double?

Source/Particles/Gather/FieldGather.H
```diff
@@ -880,6 +880,681 @@ void doGatherShapeNEsirkepovStencilImplicit (
...
+    const double dxp = x_new - x_old;
+    const double dyp = y_new - y_old;
+    const double dzp = z_new - z_old;
+    const auto dirX_sign = static_cast<int>(dxp < 0. ? -1 : 1);
+    const auto dirY_sign = static_cast<int>(dyp < 0. ? -1 : 1);
+    const auto dirZ_sign = static_cast<int>(dzp < 0. ? -1 : 1);
+    double Xcell = 0., Ycell = 0., Zcell = 0.;
+    if (num_segments > 1) {
+        Xcell = i_old + shift + 0.5_rt*(1-dirX_sign);
+        Ycell = j_old + shift + 0.5_rt*(1-dirY_sign);
```
ax3l: As above: explicit case for the ints would be safer and deciding if `amre::Real/ParticleReal` or enforce `double` is the best choice.

Source/Particles/Gather/FieldGather.H
```diff
@@ -880,6 +880,681 @@ void doGatherShapeNEsirkepovStencilImplicit (
...
+    // compute total change in particle position and the initial cell
+    // locations in each direction used to find the position at cell crossings.
+    const double dxp = x_new - x_old;
+    const double dzp = z_new - z_old;
+    const auto dirX_sign = static_cast<int>(dxp < 0. ? -1 : 1);
+    const auto dirZ_sign = static_cast<int>(dzp < 0. ? -1 : 1);
+    double Xcell = 0., Zcell = 0.;
+    if (num_segments > 1) {
+        Xcell = i_old + shift + 0.5_rt*(1-dirX_sign);
+        Zcell = k_old + shift + 0.5_rt*(1-dirZ_sign);
```
ax3l: Implicit int cast, mixed double types.

Source/Particles/ShapeFactors.H
```diff
@@ -64,6 +64,19 @@ struct Compute_shape_factor
             // index of the leftmost cell where particle deposits
             return j-1;
         }
+        else if constexpr (depos_order == 4){
```
ax3l, line 67: yay, glad this worked out :) :)

ax3l (commented):
This looks really good, thank you!

I added minor comments on our precision system:
Note that we can control field and particle precision independently, by using `amrex::Real` (fields) and `amrex::ParticleReal` (particles) to set `float` or `double` independently.

If we do not use them, then we enforce a precision (e.g., `double`).

We usually set `Real` and `ParticleReal` to the same values, but mixing is generally possible.

Two things to keep in mind:
- position deltas used in depositions are usually our source of precision issues in our current unit system, this is where we currently promote to `double` until we normalize cell edges in a clever way in the future.
- converting (casting) from/to float/double/int is a surprisingly **expensive** operation at runtime, especially on GPUs. So we want to minimize them.

Source/Particles/Deposition/CurrentDeposition.H
JustinRayAngus: This was copied from Dave Grote's implicit deposition routine. The appended "_mid" for variable names means it is at time n + 1/2. There is no small angle approximation made. If you are confused about why costheta_mid and sintheta_mid are defined as they are, it is because this definition is needed to keep exact charge conservation. It will be described in a forthcoming paper, but I can share with you now.

Source/Particles/Deposition/CurrentDeposition.H
JustinRayAngus: Will keep this in mind. It really should be based on the number of guard cells rather than a hardwired value. So I think I will defer this change to the future when the implicit method is more mature.

Source/Particles/Deposition/CurrentDeposition.H
JustinRayAngus: Will adopt the static cast. I'm still a little unclear on when to use the different types, but I think we should keep double here since it is used to find cell crossings and we want to minimize roundoff error.

JustinRayAngus: I will change shift to a double as well to be consistent

Source/Particles/Deposition/CurrentDeposition.H
JustinRayAngus: Will do.

Source/Particles/Gather/FieldGather.H
JustinRayAngus: Got it.

Source/Particles/Gather/FieldGather.H
JustinRayAngus: Got it

Source/Particles/ShapeFactors.H
JustinRayAngus: Yes. Glad you asked about it.

Source/Particles/Deposition/CurrentDeposition.H
ax3l: Sounds good! We have two kinds of asserts: the one here is only added in debug mode - the `WARPX_ALWAYS_ASSERT` is always run and should not be used in hot loops like GPU kernels.

Source/Particles/Deposition/CurrentDeposition.H
ax3l: Thanks, ok! Yes happy to read that up.

Source/Particles/Deposition/CurrentDeposition.H
ax3l: By default we actually set `amrex::Real` and `amrex::ParticleReal` to `double`.

Long-term, it would be ideal if we can get WarpX to generally run with `float`, by normalizing well and implementing necessary routines where needed (e.g., special roundoff treatments in long sums).

We can chat about this next time to clarify questions and strategy more :)

Source/Particles/Deposition/CurrentDeposition.H
JustinRayAngus: Still a bit confused on types, but it looks like from previous deposition schemes that anything to do with computing the shape factor is kept double. I see the seg_factor terms as part of the shape factor, so I will make them all double.

ax3l (approved):
Thanks a lot, looks great!

Source/Particles/Deposition/CurrentDeposition.H
ax3l: Yes, we have not carefully debugged this yet and promoted some types based on empirical tests we did.

I think more stable would be to make sure the cell edges and position changes dx, dy and dz are in kernels scaled to 0:1 or -1:1 and then we scale up again at the current deposition atomic.

(We definitely had a working _single precision_ PIC loop with VB and Esirkepov in PIConGPU using such normalization. But that's an orthogonal effort that we just have not done yet in WarpX.)

Source/Particles/Deposition/CurrentDeposition.H
JustinRayAngus: Good to know. Thanks.

PR #4624, opened 2024-01-23
Apply filtering for the charge density in full diagnostics RZ PSATD 

oshapoval:
Apply filtering for the charge density in full diagnostics when using RZ PSATD.

To do: check charge conservation in the following CI tests: 

- [x] galilean_rz_psatd_current_correction,
- [x] galilean_rz_psatd_current_correction_psb.

Screenshot 2024-01-23 at 1 14 08 AM


RemiLehe:
As a side-note: with this PR, it looks like, when using PSATD, `rho` and `J` are unfiltered in the diagnostics, because:

- When using PSATD, we turn off the regular filter and use k-space filtering instead:
https://github.com/ECP-WarpX/WarpX/blob/4ab88c5909ac2a0dd300d8ed111bfa04280b3e76/Source/WarpX.cpp#L804

- k-space filtering is only applied in spectral space, but it is not applied (e.g. with a backward transform) to the original real-space data:
https://github.com/ECP-WarpX/WarpX/blob/4ab88c5909ac2a0dd300d8ed111bfa04280b3e76/Source/FieldSolver/WarpXPushFieldsEM.cpp#L313
https://github.com/ECP-WarpX/WarpX/blob/4ab88c5909ac2a0dd300d8ed111bfa04280b3e76/Source/FieldSolver/WarpXPushFieldsEM.cpp#L391
(This happens even when we do current correction: https://github.com/ECP-WarpX/WarpX/blob/development/Source/FieldSolver/WarpXPushFieldsEM.cpp#L708)

Source/Diagnostics/FullDiagnostics.H
```diff
@@ -25,6 +25,8 @@ private:
      * before writing the diagnostic.
      */
     bool m_solver_deposits_current = true;
+    /** Whether to apply k-space filtering of the charge density */
+    bool m_apply_filter_rho = false;
```
RemiLehe: I don't think that storing this attribute is needed. (In practice, we always use `true` for the `FullDiagnostic`, as shown below in the `.cpp` file.) 
```suggestion
```

Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -1,5 +1,8 @@
 #include "FullDiagnostics.H"
-
+#if (defined WARPX_DIM_RZ) && (defined WARPX_USE_PSATD)
+    #include "FieldSolver/SpectralSolver/SpectralFieldData.H"
+    #include "FieldSolver/SpectralSolver/SpectralSolverRZ.H"
+#endif
```
RemiLehe: Do we need to add this here?

Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -46,6 +49,9 @@ FullDiagnostics::FullDiagnostics (int i, std::string name):
     m_solver_deposits_current{
         !(WarpX::electromagnetic_solver_id == ElectromagneticSolverAlgo::None &&
         WarpX::electrostatic_solver_id != ElectrostaticSolverAlgo::LabFrameElectroMagnetostatic)}
+    #if (defined WARPX_DIM_RZ) && (defined WARPX_USE_PSATD)
+    , m_apply_filter_rho{true}
+    #endif
```
RemiLehe: If we remove `m_apply_filter_rho` in `FullDiagnostics.H` as I suggested, then it would also be removed here.
```suggestion
```

Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -268,14 +274,14 @@ FullDiagnostics::InitializeFieldFunctorsRZopenPMD (int lev)
             }
         } else if ( m_varnames_fields[comp] == "rho" ){
             // Initialize rho functor to dump total rho
-            m_all_field_functors[lev][comp] = std::make_unique<RhoFunctor>(lev, m_crse_ratio, -1,
+            m_all_field_functors[lev][comp] = std::make_unique<RhoFunctor>(lev, m_crse_ratio, m_apply_filter_rho, -1,
```
```suggestion
            m_all_field_functors[lev][comp] = std::make_unique<RhoFunctor>(lev, m_crse_ratio, true, -1,
```

Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -268,14 +274,14 @@ FullDiagnostics::InitializeFieldFunctorsRZopenPMD (int lev)
...
                                                         false, ncomp);
             if (update_varnames) {
                 AddRZModesToOutputNames(std::string("rho"), ncomp);
             }
         } else if ( m_varnames_fields[comp].rfind("rho_", 0) == 0 ){
             // Initialize rho functor to dump rho per species
-            m_all_field_functors[lev][comp] = std::make_unique<RhoFunctor>(lev, m_crse_ratio, m_rho_per_species_index[i],
+            m_all_field_functors[lev][comp] = std::make_unique<RhoFunctor>(lev, m_crse_ratio, m_apply_filter_rho, m_rho_per_species_index[i],
```
```suggestion
            m_all_field_functors[lev][comp] = std::make_unique<RhoFunctor>(lev, m_crse_ratio, true, m_rho_per_species_index[i],
```

Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -443,7 +449,7 @@ FullDiagnostics::AddRZModesToDiags (int lev)
     }
     // rho
     if (rho_requested) {
-        m_all_field_functors[lev][icomp] = std::make_unique<RhoFunctor>(lev, m_crse_ratio, -1, false, ncomp_multimodefab);
+        m_all_field_functors[lev][icomp] = std::make_unique<RhoFunctor>(lev, m_crse_ratio, m_apply_filter_rho, -1, false, ncomp_multimodefab);
```
```suggestion
        m_all_field_functors[lev][icomp] = std::make_unique<RhoFunctor>(lev, m_crse_ratio, true, -1, false, ncomp_multimodefab);
```

Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -640,10 +646,10 @@ FullDiagnostics::InitializeFieldFunctors (int lev)
             m_all_field_functors[lev][comp] = std::make_unique<CellCenterFunctor>(warpx.get_pointer_vector_potential_fp(lev, 2), lev, m_crse_ratio);
         } else if ( m_varnames[comp] == "rho" ){
             // Initialize rho functor to dump total rho
-            m_all_field_functors[lev][comp] = std::make_unique<RhoFunctor>(lev, m_crse_ratio);
+            m_all_field_functors[lev][comp] = std::make_unique<RhoFunctor>(lev, m_crse_ratio, m_apply_filter_rho);
```
```suggestion
            m_all_field_functors[lev][comp] = std::make_unique<RhoFunctor>(lev, m_crse_ratio, true);
```

Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -640,10 +646,10 @@ FullDiagnostics::InitializeFieldFunctors (int lev)
...
         } else if ( m_varnames[comp].rfind("rho_", 0) == 0 ){
             // Initialize rho functor to dump rho per species
-            m_all_field_functors[lev][comp] = std::make_unique<RhoFunctor>(lev, m_crse_ratio, m_rho_per_species_index[i]);
+            m_all_field_functors[lev][comp] = std::make_unique<RhoFunctor>(lev, m_crse_ratio, m_apply_filter_rho, m_rho_per_species_index[i]);
```
```suggestion
            m_all_field_functors[lev][comp] = std::make_unique<RhoFunctor>(lev, m_crse_ratio, true, m_rho_per_species_index[i]);
```

Source/Diagnostics/ComputeDiagFunctors/RhoFunctor.cpp
```diff
@@ -20,11 +20,13 @@
 
 RhoFunctor::RhoFunctor (const int lev,
                         const amrex::IntVect crse_ratio,
+                        bool apply_filter_rho,
```
RemiLehe: Could we rename this variable to e.g. `apply_rz_psatd_filter`, to convey more precisely what this variable does?

RemiLehe (changes requested):
Thanks for this PR! I added a few suggestions.

Source/Diagnostics/FullDiagnostics.H
oshapoval: oops, it was a leftover from my previous draft implementation. Thank you for spotting this! Fixed in 14dd38d705.

Source/Diagnostics/FullDiagnostics.cpp
oshapoval: Removed in e0ded03d013475d99d.

Source/Diagnostics/ComputeDiagFunctors/RhoFunctor.cpp
oshapoval: I agree, fixed in 044d20727910.

RemiLehe (approved):

PR #4627, opened 2024-01-23
Temporarily deactivate `Python_background_mcc_1d`

RemiLehe:
This test mysteriously started failing about 1 week ago.

While we debug this, we should probably deactivate it, so as to avoid confusion in the CI tests of other PRs.

### DIFF
--- a/Regression/WarpX-tests.ini
+++ b/Regression/WarpX-tests.ini
@@ -3021,25 +3021,6 @@ doVis = 0
 compareParticles = 0
 analysisRoutine = Examples/Physics_applications/capacitive_discharge/analysis_1d.py
 
-[Python_background_mcc_1d_tridiag]
-buildDir = .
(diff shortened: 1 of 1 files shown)


ax3l:
@roelof-groenewald sorry, saw Remi also pushed one in the morning :sweat_smile: 
x-ref #4628

ax3l (approved):
Thx!

roelof-groenewald:
Thanks! But maybe we should merge mine since it doesn't delete the test, just comments it out?

ax3l:
Looks like this one now hangs in `Python_dsmc_1d`

roelof-groenewald:
> Looks like this one now hangs in `Python_dsmc_1d`

Hmm. That is interesting since all three those tests use the same input script. So why is only the first instance of the test completing...

PR #4628, opened 2024-01-23
Temporary fix: disable hanging CI test (`Python_background_mcc_1d_tridiag`)

roelof-groenewald:
There is a 1d CI test that hangs for unknown reasons. In this temporary fix the test is disabled so that the other tests will run. The test should be reactivated once we figure out why it hangs.


RemiLehe (approved):

ax3l:
Thank you!

Issue #4604, opened 2024-01-10
Backtransform diagnostics crash when filters remove all particles


RemiLehe:
PR #4570 added particle filters to the Backtransformed diagnostics. 
However, when the filters are such that they eliminate all particles, the code crashes with the following error:
```
A Record can not be written without any contained RecordComponents: id
```

Reproducer:
[inputs_3d.txt](https://github.com/ECP-WarpX/WarpX/files/13894175/inputs_3d.txt)
(identical to [this test](https://github.com/ECP-WarpX/WarpX/blob/development/Examples/Tests/boosted_diags/inputs_3d), but with `diag2.beam.random_fraction = 0` to remove all particles)

RemiLehe:
Interestingly, the same issue does not occur for `BoundaryScraping` diagnostics or `Full` diagnostics (i.e. the code does not crash when the filters eliminate all particles)

RemiLehe:
@ax3l @guj Any idea why this is happening?

guj:
In openPMD code, this is the line where exception is thrown:
"if (!this->written() && this->empty() && !this->datasetDefined())"
so looks like there is an attempt to write empty dataset without defining it.  

Guess it can be reproduced with the given input file and  PR 4570? I will give it a try

guj:
OK I can reproduce and will put up a fix

guj:
So we ran into a new scenario from BTD:   
particle_diags.size() = 3
num particles [0, 0, 0]
particle offset [0,0,1000]

I want to double check first that this is all correct.  (1000 is a bit suspicious here as (1) it should be 0, (2) particles in the previous flushes are double digits: 92, 16, 18)

ax3l:
I think this is a corner case, where we usually declare an openPMD record as constant if there are zero particles ever written to a step.

For BTD, due to multiple writes (or non-writes in this case), tracking this is a bit intricate. We probably missed a corner case and need to add it with @RemiLehe's reproducer.

ax3l:
Hm, if the particle offset is non-zero, that means there were particles flushed before to the lab station step.

guj:
> Hm, if the particle offset is non-zero, that means there were particles flushed before to the lab station step.

Remi's input has random_fraction = 0, that's why the offset is 1000 (which is right), and there is no particle to store.

ax3l:
@guj thanks for taking this on!

PR #4631, opened 2024-01-23
added a workaround for issue 4604

guj:
Guard against BTD case:
 - 0  particles are written at the last flush step  (due to fraction = 0)
 - offset > 0 (particles do exist although not selected)

See issue #4604

### DIFF
--- a/Source/Diagnostics/WarpXOpenPMD.cpp
+++ b/Source/Diagnostics/WarpXOpenPMD.cpp
@@ -709,6 +709,7 @@ WarpXOpenPMDPlot::DumpToFile (ParticleContainer* pc,
     }
 
     if (is_last_flush_to_step) {
+        SetupPos(currSpecies, NewParticleVectorSize, isBTD); // workaround for issue 4604
         SetConstParticleRecordsEDPIC(currSpecies, NewParticleVectorSize, charge, mass);
     }
 


Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -709,6 +709,7 @@ WarpXOpenPMDPlot::DumpToFile (ParticleContainer* pc,
     }
 
     if (is_last_flush_to_step) {
+        SetupPos(currSpecies, NewParticleVectorSize, isBTD); // workaround for issue 4604
         SetConstParticleRecordsEDPIC(currSpecies, NewParticleVectorSize, charge, mass);
     }
```
ax3l, line 714: I am not sure this should go exactly like this here.

What do you think about this?
```suggestion
    if (is_last_flush_and_never_particles) {
        SetupPos(currSpecies, NewParticleVectorSize, isBTD);
    }
    if (is_last_flush_to_step) {
        SetConstParticleRecordsEDPIC(currSpecies, NewParticleVectorSize, charge, mass);
    }
```

ax3l: Admittedly, this is exactly what is done a few lines earlier, isn't it? Why is that not called, I wonder.

ax3l: What are the values of `num_dump_particles` and `ParticleFlushOffset` for this problem?

guj: num_dump_particles=0
ParticleFlushOffset=1000

so  is_last_flush_and_never_particles (= is_last_flush_to_step && num_dump_particles == 0 && ParticleFlushOffset == 0) will be false.

guj:
Not sure why it fails during "copying files to run directory..."

Source/Diagnostics/WarpXOpenPMD.cpp
ax3l: Thanks, got it.

If `ParticleFlushOffset > 0` that means at some point, at least one rank had particles to write.

We need to check if maybe in the call where we write particles (earlier), we have a logic mistake that forgets to call `SetupPos` collectively.

ax3l: We should printf debug for this run the `ParticleFlushOffset` and `num_dump_particles` variables. It is also possible that the book-keeping of `ParticleFlushOffset` is wrong.

guj: ParticleFlushOffset = 1000  (which is correct without the num_fracion option)
num_dump_particles = 0  (which is correct _with_ "num_fraction = 0" ) 

I don't know how easy it is to adjust ParticleFlushOffset with "num_fraction" on in the diags.

ax3l: Hm, maybe we filter twice and forget `num_fraction` in one case? Let's take a look where `num_fraction` is applied and if the counter is done correctly afterwards.

guj: My understanding is that the offsets are set in the BTD object, and when it calls flush(), the  filters are then applied in the underlying flush formats during copyParticles().  Unlike other filters,  the random filter affected the offsets of the caller BTD object. 
I think the workaround is the easy and harmless. Otherwise, have to collect the particles written from all processes and update the offset in BTD.

ax3l: >  I think the workaround is the easy and harmless.

I would rather like to fix the underlying issue in the BTD offset counter. We need consistent meta data.

cc @RevathiJambunathan :)

guj: If it makes sense,  can apply the random filter before BTD calls flush().  It will make things cleaning and easier.

ax3l: Yes, we probably need to add this to:
- full diags
- BTD
- scraping

ax3l:
Fixing the broken meta-data in #4657 :)

PR #4632, opened 2024-01-24
Adding support for pml in domain for multiple patches

RevathiJambunathan:
Thanks to @WeiqunZhang for the discussion and prototype!!!

The current implementation for `pml_in_domain` only works for a single patch 
This PR adds capability to support 'pml_in_domain' for multiple distinct mesh-refinement patches 

without the support for this feature : (PML in domain only on the left most edge of the left patch and right edge of the right patch) - i.e., the interiors edges dont apply PML correctly
Screenshot 2024-01-23 at 10 16 36 PM

With this PR : 
Screenshot 2024-01-23 at 10 15 32 PM


Source/BoundaryConditions/PML.cpp
```diff
@@ -760,39 +767,43 @@ PML::PML (const int lev, const BoxArray& grid_ba, const DistributionMapping& gri
...
-            BoxArray(grid_cba.boxList().intersect(cdomain)) : grid_cba;
+        Box const cdomain = grid_cba_reduced.minimalBox();
 
         const IntVect cncells = IntVect(ncell)/ref_ratio;
         const IntVect cdelta = IntVect(delta)/ref_ratio;
 
         // Assuming that refinement ratio is equal in all dimensions
         const BoxArray& cba = MakeBoxArray(is_single_box_domain, cdomain, *cgeom, grid_cba_reduced,
                                            cncells, do_pml_in_domain, do_pml_Lo, do_pml_Hi);
+        amrex::Print() << " cba : " << cba << "\n";
```
RevathiJambunathan:
```suggestion
```

Source/BoundaryConditions/PML.cpp
```diff
@@ -867,8 +878,10 @@ PML::MakeBoxArray (bool is_single_box_domain, const amrex::Box& regular_domain,
                    const amrex::IntVect& do_pml_Lo, const amrex::IntVect& do_pml_Hi)
 {
     if (is_single_box_domain) {
+        amrex::Print() << " calling MakeBoxArraySingle \n";
```
RevathiJambunathan:
```suggestion
```

Source/BoundaryConditions/PML.cpp
```diff
@@ -867,8 +878,10 @@ PML::MakeBoxArray (bool is_single_box_domain, const amrex::Box& regular_domain,
...
         return MakeBoxArray_single(regular_domain, grid_ba, ncell, do_pml_Lo, do_pml_Hi);
     } else { // the union of the regular grids is *not* a single rectangular domain
+        amrex::Print() << " calling MakeBoxArrayMultiple \n";
```
RevathiJambunathan:
```suggestion
```

Source/BoundaryConditions/PML.cpp
```diff
@@ -566,34 +566,41 @@ PML::PML (const int lev, const BoxArray& grid_ba, const DistributionMapping& gri
     // and passed to `MakeBoxArray`, which surrounds it by PML boxes
     // (thus creating the PML boxes at the right position, where they overlap with the original domain)
     // minimalBox provides the bounding box around grid_ba for level, lev.
```
RevathiJambunathan:
```suggestion
```

Source/BoundaryConditions/PML.cpp
```diff
@@ -565,35 +565,43 @@ PML::PML (const int lev, const BoxArray& grid_ba, const DistributionMapping& gri
...
+
+    BoxArray grid_ba_reduced = grid_ba;
     if (do_pml_in_domain) {
-        for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
-            if (do_pml_Lo[idim]){
-                domain0.growLo(idim, -ncell);
-            }
-            if (do_pml_Hi[idim]){
-                domain0.growHi(idim, -ncell);
+        BoxList bl = grid_ba.boxList();
```
WeiqunZhang, line 571: We could also use `simplified_list()` instead of `boxList` here. That way, it's less likely to hit the assert failure below.

Source/BoundaryConditions/PML.cpp
```diff
@@ -760,24 +768,28 @@ PML::PML (const int lev, const BoxArray& grid_ba, const DistributionMapping& gri
...
-                if (do_pml_Lo[idim]){
-                    // ncell is divided by refinement ratio to ensure that the
-                    // physical width of the PML region is equal in fine and coarse patch
-                    cdomain.growLo(idim, -ncell/ref_ratio[idim]);
-                }
-                if (do_pml_Hi[idim]){
-                    // ncell is divided by refinement ratio to ensure that the
-                    // physical width of the PML region is equal in fine and coarse patch
-                    cdomain.growHi(idim, -ncell/ref_ratio[idim]);
+            BoxList bl = grid_cba.boxList();
```
WeiqunZhang, line 781: Here we can use simplified_list as well.

Source/BoundaryConditions/PML.cpp
RevathiJambunathan: Thank you Weiqun for the suggestions! 
Done

Source/BoundaryConditions/PML.cpp
RevathiJambunathan: done

WeiqunZhang (approved):

Source/BoundaryConditions/PML.cpp
RevathiJambunathan: discussed offline, switched back to boxList due to failing CI

Source/BoundaryConditions/PML.cpp
```diff
@@ -565,35 +565,43 @@ PML::PML (const int lev, const BoxArray& grid_ba, const DistributionMapping& gri
     // In order to implement this, a reduced domain is created here (decreased by ncells in all direction)
```
RemiLehe: Maybe the term "reduced domain" is not appropriate anymore in this comment, since it is not a single contiguous domain anymore.

Source/BoundaryConditions/PML.cpp
```diff
@@ -565,35 +565,43 @@ PML::PML (const int lev, const BoxArray& grid_ba, const DistributionMapping& gri
...
-            if (do_pml_Lo[idim]){
-                domain0.growLo(idim, -ncell);
-            }
-            if (do_pml_Hi[idim]){
-                domain0.growHi(idim, -ncell);
+        BoxList bl = grid_ba.boxList();
+        for (auto& b : bl) {
+            for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
+                if (do_pml_Lo[idim]) {
+                    Box const& bb = amrex::adjCellLo(b, idim);
```
RemiLehe, line 580: Could you add a comment to explains what this code (the call to `adjCellLo` and the subsequent `intersects`) does?

Source/BoundaryConditions/PML.cpp
```diff
@@ -565,35 +565,43 @@ PML::PML (const int lev, const BoxArray& grid_ba, const DistributionMapping& gri
...
+    BoxArray grid_ba_reduced = grid_ba;
     if (do_pml_in_domain) {
-        for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
-            if (do_pml_Lo[idim]){
-                domain0.growLo(idim, -ncell);
-            }
-            if (do_pml_Hi[idim]){
-                domain0.growHi(idim, -ncell);
+        BoxList bl = grid_ba.boxList();
+        for (auto& b : bl) {
```
RemiLehe, line 574: Could you add a comment to explain what this `for` loop does?
(In my understanding, it loops over the box at this level, and whenever a box touches a PML, it reduces the box by `ncell`, since this is where the surrounding "in-domain" PML will be created.)

Source/BoundaryConditions/PML.cpp
```diff
@@ -760,24 +768,28 @@ PML::PML (const int lev, const BoxArray& grid_ba, const DistributionMapping& gri
         BoxArray grid_cba = grid_ba;
         grid_cba.coarsen(ref_ratio);
 
-        // assuming that the bounding box around grid_cba is a single patch, and not disjoint patches, similar to fine patch.
-        amrex::Box cdomain = grid_cba.minimalBox();
+        BoxArray grid_cba_reduced = grid_cba;
```
RemiLehe, line 779: The code below duplicates the above code. 
Maybe it would be good, in a separate follow-up PR, to create a separate function for this.

RemiLehe (changes requested):
Thanks for this PR! 
Could you add an automated test for this?

Also: could you confirm that this feature still work nicely with `do_similar_dm_pml`?

Source/BoundaryConditions/PML.cpp
RevathiJambunathan: added comments to clarify

Source/BoundaryConditions/PML.cpp
RevathiJambunathan: added comments to clarify

Source/BoundaryConditions/PML.cpp
RevathiJambunathan: yes. I ll do this in a separate PR

RevathiJambunathan:
Added a CI test by modifying existing 2D mr particles_in_pml test to include two patches 
The dev branch shows spurious fields on the inner boundaries of the patch 
Screenshot 2024-02-27 at 3 57 32 PM
The current PR applies the pml inside the domain correctly, fixing this spurious effect
Screenshot 2024-02-27 at 3 58 28 PM

RemiLehe (approved):

PR #4634, opened 2024-01-24
Update profile and job script for LUMI supercomputer

lucafedeli88:
This PR updates the profile and the job script for the LUMI supercomputer.


tmsclark2 (approved):
It works well, thanks

ax3l (approved):
Thank you both, excellent :+1:

PR #4637, opened 2024-01-25
Add WarpX_CCACHE Option

haampie:
WarpX autodetects `ccache` and uses it, and there's nothing you can do
about it. This PR disables it by default, and lets developers enable it
through `-DWarpX_CCACHE:BOOL=ON`.

The reason for this is mostly related to compiler wrappers, of which
there are many :)

If `g++` is a compiler wrapper, then `ccache g++` cannot see the
effective flags passed to underlying real `g++`. That leads eventually
to false positive cache hits when the wrapper changes, which is a pain to
debug.

For compiler wrappers you want the wrapper to invoke the call to `ccache <real`
`g++ invocation>` to fix that, which is for example how Spack handles it.

Further, if you use Spack with `ccache` enabled, the defaults of WarpX
cause `ccache` to be invoked twice (inner & outer), doubling the cache
requirements.

Also, compiler wrappers that handle ccache themselves may set further ccache
options / flags that WarpX does not set, such as disabling hashing of the build
dir -- w/o that option cache may be useless.

Finally, it may just not be very nice to populate people's caches with WarpX objects if
they are not developers and have no intention to build WarpX multiple times.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,14 +38,6 @@ endif()
 set_cxx17_superbuild()
 
 
-# CCache Support ##############################################################
-#
-# this is an optional tool that stores compiled object files; allows fast
-# re-builds even with "make clean" in between. Mainly used to store AMReX
-# objects
-set_ccache()
-
-
 # Output Directories ##########################################################
 #
 # temporary build directories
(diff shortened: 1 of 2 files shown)


ax3l:
Hi @haampie, thanks for this!

I am ok adding an option, but defaulting it to ON :grin: 

> and there's nothing you can do about it

Admittedly, `-DCCACHE_PROGRAM=NO` should do the same, but a clean option is nicer.

I will think about your reasoning for another default a bit more... ;-)

haampie:
Feel free to push to this branch or close and make your own ;)

ax3l (approved):

ax3l (approved):

PR #4646, opened 2024-01-29
Update DSMC picmi class initialization function

roelof-groenewald:
Update the DSMC picmi class in accordance with #4554.

I'm hoping this fix will allow the DSMC test to be activated again.


dpgrote (approved):
Thanks! Looks good. Somehow I missed one.

roelof-groenewald:
> Thanks! Looks good. Somehow I missed one.

Thanks! I think this was just due to the sequence of PR merges. Both your PR and the DSMC PR were merged around the same time.

PR #4648, opened 2024-01-30
Implement Poisson solver based on integrated Green functions 

aeriforme:
This uses the method from:
https://journals.aps.org/prab/abstract/10.1103/PhysRevSTAB.9.044204
https://journals.aps.org/prab/pdf/10.1103/PhysRevSTAB.10.129901

Because this uses the same FFT infrastructure as PSATD, right now, the code needs to be compiled with:
```-DWarpX_PSATD=ON```.

Also: since the solver uses a global FFT, right now only one MPI rank participates in the FFT solve (but this MPI rank can still use a full GPU) ; this could be improved later by using heFFTe. The rest of the PIC operations (charge deposition, gather, etc.) are still performed by the different MPI ranks in parallel.

Comparison with [Basseti-Erskine formula](https://cds.cern.ch/record/122227/files/198005132.pdf) (see the test script in this PR):
296619769-50a1d730-84f8-475e-8888-b76c24c4daf1

## TODO

- [x] Test on GPU
- [x] Add back the relativistic factor in dx
- [x] Overcome limitation with single MPI rank (use ```ParallelCopy```)
- [x] Fix bug in the multi-GPU implementation, in the case ```grid_type = collocated``` (this could be done with the regular Poisson solver instead of the relativistic one, so that we can output ```phi```)
- [x] Add timers to the FFT and the calculation of the IGF
- [x] Finalize the automated test
- [x] Decide on the user interface: introduce a new parameter ```warpx.poisson_solver``` that can be set to ```multigrid```, ```fft_based```, etc. + how to handle compatibility with boundary conditions?
if electromagnetic simulation with PML + ```initialize_self_fields```, then use IGF which implies open boundaries 
IGF only compatible with ```open``` and viceversa
multigrid not compatible with ```open```
- [ ] This should follow the same logic as the Maxwell solver. In the PICMI interface, there is already a “method” parameter for the Poisson solver.

more
- add new boundary condition “open” that is only valid for Poisson-like field solvers + organize documentation to group boundaries in similar types

more
- [x] ~Abort if someone tries to use mesh refinement~ If mesh refinement is on and the user has selected the IGF solver, then use IGF only on the coarsest level (`lev = 0`) and MLMG in the refined patches
- [x] Properly set guard cells of ```phi``` outside of the global domain (useful for ```grid_type = collocated```)
- [ ] ~Save the FFT plans in a permanent object, so that we do not need to recompute them.~ Probably not needed. From the timers, it seems that the creation and destruction of FFT plan does not take any significant time.
- [x] Clean-up ; move code to a different file, maybe ```IntegratedGreenFunctionSolver.H```, in ```ablastr```
- [x] Handle dimensionality that are not 3D, abort if we are not in 3D
- [x] Address warning
```
/Users/rlehe/Documents/codes/warpx_directory/warpx/Source/ablastr/fields/PoissonSolver.H:69:1: warning: function 'computePhiIGF' is not needed and will not be emitted [-Wunneeded-internal-declaration]
computePhiIGF ( amrex::MultiFab const & rho,
```
By calling the function in ```VectorPoissonSolver.H``` (for the magnetostatic solver)

- [ ] Fix ```ParallelCopy``` error that appears in ```DEBUG``` mode:
```
Assertion `boxArray().ixType() == src.boxArray().ixType()' failed, file "/global/u2/a/aforment/src/war
```

- [x] Documentation

### CODE REVIEW

**Comment by ax3l** [comment]
@aeriforme small merge conflict in `Source/FieldSolver/ElectrostaticSolver.cpp` :pray:

File: Source/ablastr/fields/PoissonSolver.H [hunk 1]
```diff
@@ -11,8 +11,12 @@
 #include <ablastr/utils/Communication.H>
 #include <ablastr/utils/TextMsg.H>
 #include <ablastr/warn_manager/WarnManager.H>
+#include <ablastr/math/fft/AnyFFT.H>
 #include <ablastr/fields/Interpolate.H>
 
+#if defined(WARPX_USE_PSATD)  && defined(WARPX_DIM_3D)
```
ax3l: Macros are ok. We  have yet to introduce separate FFT macros for ABLASTR (will do another time, already not done for other FFT functions in ABLASTR).

Source/ablastr/fields/IntegratedGreenFunctionSolver.H
```diff
@@ -0,0 +1,225 @@
+/* Copyright 2019-2022 Axel Huebl, Remi Lehe
```
ax3l:
```suggestion
/* Copyright 2019-2024 Arianna Formenti, Remi Lehe
```

Source/ablastr/fields/IntegratedGreenFunctionSolver.H
```diff
@@ -0,0 +1,225 @@
...
+ *
+ * This file is part of WarpX.
```
ax3l:
```suggestion
 * This file is part of ABLASTR.
```

Source/ablastr/fields/IntegratedGreenFunctionSolver.H
```diff
@@ -0,0 +1,225 @@
...
+
+#include <array>
+#include <optional>
+
+
+namespace {
+
+using namespace amrex::literals;
+
+// Implements equation 2 in https://journals.aps.org/prab/pdf/10.1103/PhysRevSTAB.10.129901 with some modification to symmetrize the function
```
ax3l: More robust to rely on DOI :)
```suggestion
// Implements equation 2 in https://doi.org/10.1103/PhysRevSTAB.10.129901 with some modification to symmetrize the function
```

Maybe expand the comment slightly on the modification?

Source/ablastr/fields/IntegratedGreenFunctionSolver.H
```diff
@@ -0,0 +1,225 @@
...
+#include <AMReX_MLLinOp.H>
+#include <AMReX_MultiFab.H>
+#include <AMReX_Parser.H>
+#include <AMReX_REAL.H>
+#include <AMReX_SPACE.H>
+#include <AMReX_Vector.H>
+
+
+#include <array>
+#include <optional>
```
ax3l:
```suggestion
#include <cmath>
#include <optional>
```

Source/ablastr/fields/IntegratedGreenFunctionSolver.H
```diff
@@ -0,0 +1,225 @@
...
+        + x*y*std::asinh( z/std::sqrt(x*x + y*y) );
+    return G;
+}
+
+inline void
+computePhiIGF ( amrex::MultiFab const & rho,
+                amrex::MultiFab & phi,
+                std::array<amrex::Real, 3> const cell_size,
+                amrex::BoxArray const ba ) {
+
```
ax3l: Looking at how this is used, I do not think it benefits from `inline`/inlining, does it?

I would propose, in the sense of improving compile time, to move the body `{}` of this function into a `.cpp` file.

Source/ablastr/fields/PoissonSolver.H
```diff
@@ -109,6 +113,8 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
 {
     using namespace amrex::literals;
 
+    WARPX_PROFILE("computePhi");
```
ax3l: Inside ABLASTR we use:
```suggestion
    BL_PROFILE("computePhi");
```
or
```suggestion
    ABLASTR_PROFILE("computePhi", false);
```
to avoid access to the `WarpX::` global variable space - which for instance ImpactX has no access to.

Source/ablastr/fields/PoissonSolver.H
```diff
@@ -121,7 +127,8 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
     amrex::Real max_norm_b = 0.0;
     for (int lev=0; lev<=finest_level; lev++) {
         using namespace ablastr::constant::SI;
-        rho[lev]->mult(-1._rt/ep0);  // TODO: when do we "un-multiply" this? We need to document this side-effect!
+
```
ax3l: unintentional change?
```suggestion
```

Source/ablastr/fields/PoissonSolver.H
```diff
@@ -153,6 +160,23 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
                 {{ beta[0], beta[1], beta[2] }};
 #endif
 
+#if (defined(WARPX_USE_PSATD)  && defined(WARPX_DIM_3D))
+        // Use the Integrated Green Function solver (fft-based) on the coarsest level if it was selected
+        if(WarpX::poisson_solver_id==PoissonSolverAlgo::IntegratedGreenFunction && lev==0){
```
ax3l: Instead of relying on global variables of `WarpX::`, please pass parameters explicit in ABLASTR.

Add a new parameter or enum parameter to `computePhi` that selects the algorithm.

When you call `computePhi` inside WarpX you can use `WarpX::poisson_solver_id==PoissonSolverAlgo::IntegratedGreenFunction`, but not inside ABLASTR itself, because ImpactX will not have access to `WarpX::` globals.

Source/ablastr/fields/PoissonSolver.H
```diff
@@ -153,6 +160,23 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
...
+            rho[lev]->mult(-PhysConst::ep0);
```
ax3l: With this we first divide and then multiply rho by `-eps0`, do we?
Can this be integrated above to avoid the necessary scaling for the IGF solver option?

Source/ablastr/math/fft/WrapCuFFT.cpp
```diff
@@ -69,10 +71,12 @@ namespace ablastr::math::anyfft
 
     void DestroyPlan(FFTplan& fft_plan)
     {
+        ABLASTR_PROFILE("ablastr::math::anyfft::DestroyPlan", true);
         cufftDestroy( fft_plan.m_plan );
     }
 
     void Execute(FFTplan& fft_plan){
+        ABLASTR_PROFILE("ablastr::math::anyfft::Execute", true);
```
ax3l: Please use `false` for device synchronize. We luckily have a new AMReX runtime option for this and can thus avoid unnecessary device syncs (costly).

Source/ablastr/math/fft/WrapCuFFT.cpp
```diff
@@ -30,6 +31,7 @@ namespace ablastr::math::anyfft
                        Complex * const complex_array, const direction dir, const int dim)
     {
         FFTplan fft_plan;
+        ABLASTR_PROFILE("ablastr::math::anyfft::CreatePlan", true);
```
ax3l: See below:
```suggestion
        ABLASTR_PROFILE("ablastr::math::anyfft::CreatePlan", false);
```

ax3l (commented):
Thank you, that is awesome!

A few structural comments to make this work with ImpactX inline :)

Source/ablastr/fields/IntegratedGreenFunctionSolver.H
aeriforme: It may be too complex to be just a comment in the code. 
I will consider adding this info to the documentation.

Source/ablastr/fields/PoissonSolver.H
aeriforme: Is this about the empty line? If so, yes

Source/FieldSolver/ElectrostaticSolver.cpp
```diff
@@ -352,6 +356,11 @@ WarpX::computePhi (const amrex::Vector<std::unique_ptr<amrex::MultiFab> >& rho,
     const std::optional<amrex::Vector<amrex::FArrayBoxFactory const *> > eb_farray_box_factory;
 #endif
 
+    bool is_solver_multigrid = true;
+    if(WarpX::poisson_solver_id == PoissonSolverAlgo::IntegratedGreenFunction){
+        is_solver_multigrid = false;
+    }
```
ax3l: More compact, but stylistic choice:
```suggestion
    bool const is_solver_multigrid =
        WarpX::poisson_solver_id == PoissonSolverAlgo::IntegratedGreenFunction ?
        false : true;
```

Source/FieldSolver/ElectrostaticSolver.cpp
```diff
@@ -1001,6 +1011,7 @@ void ElectrostaticSolver::PoissonBoundaryHandler::definePhiBCs (const amrex::Geo
     amrex::ignore_unused(geom);
 #endif
     for (int idim=dim_start; idim<AMREX_SPACEDIM; idim++){
+    if(WarpX::poisson_solver_id == PoissonSolverAlgo::Multigrid){
```
ax3l:
```suggestion
    if (WarpX::poisson_solver_id == PoissonSolverAlgo::Multigrid){
```

Source/FieldSolver/ElectrostaticSolver.cpp
```diff
@@ -1036,10 +1047,38 @@ void ElectrostaticSolver::PoissonBoundaryHandler::definePhiBCs (const amrex::Geo
...
         }
+
+                WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                    (WarpX::field_boundary_lo[idim] != FieldBoundaryType::Open &&
+                    WarpX::field_boundary_hi[idim] != FieldBoundaryType::Open &&
+                    WarpX::field_boundary_lo[idim] != FieldBoundaryType::PML &&
+                    WarpX::field_boundary_hi[idim] != FieldBoundaryType::PML) ,
+                    "Open and PML field boundary conditions only work with "
+                    "warpx.poisson_solver = fft-based."
+                );
```
ax3l: Unusual indent?

Source/FieldSolver/ElectrostaticSolver.cpp
```diff
@@ -1036,10 +1047,38 @@ void ElectrostaticSolver::PoissonBoundaryHandler::definePhiBCs (const amrex::Geo
...
+    }
+    else if (WarpX::poisson_solver_id == PoissonSolverAlgo::IntegratedGreenFunction){
+            if (WarpX::electrostatic_solver_id != ElectrostaticSolverAlgo::None){
+                WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                    (WarpX::field_boundary_lo[idim] == FieldBoundaryType::Open &&
+                    WarpX::field_boundary_hi[idim] == FieldBoundaryType::Open),
+                    "The fft-based Poisson solver only works with field open boundary conditions "
```
ax3l:
```suggestion
                    "The FFT-based Poisson solver only works with field open boundary conditions "
```

Source/FieldSolver/ElectrostaticSolver.cpp
```diff
@@ -1036,10 +1047,38 @@ void ElectrostaticSolver::PoissonBoundaryHandler::definePhiBCs (const amrex::Geo
...
+                    "in electrostatic mode."
+                );
+            }
+            else{ // if electromagnetic mode on with species.initialize_self_fields = 1
+                WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                    (WarpX::field_boundary_lo[idim] == FieldBoundaryType::PML &&
+                    WarpX::field_boundary_hi[idim] == FieldBoundaryType::PML),
+                    "The fft-based Poisson solver only works with field PML boundary conditions "
```
ax3l:
```suggestion
                    "The FFT-based Poisson solver only works with field PML boundary conditions "
```

Source/Initialization/WarpXInitData.cpp
```diff
@@ -213,11 +213,20 @@ WarpX::PrintMainPICparameters ()
...
-      }
+    }
     else if ((em_solver_medium == MediumForEM::Macroscopic) &&
             (WarpX::macroscopic_solver_algo == MacroscopicSolverAlgo::BackwardEuler)){
       amrex::Print() << "                      |  - Backward Euler algorithm\n";
-      }
+    }
+    if(electrostatic_solver_id != ElectrostaticSolverAlgo::None){
+        if(poisson_solver_id == PoissonSolverAlgo::IntegratedGreenFunction){
+            amrex::Print() << "Poisson solver:       | fft-based" << "\n";
```
```suggestion
            amrex::Print() << "Poisson solver:       | FFT-based" << "\n";
```

Source/Utils/WarpXAlgorithmSelection.cpp
```diff
@@ -55,6 +55,12 @@ const std::map<std::string, int> electrostatic_solver_algo_to_int = {
     {"default", ElectrostaticSolverAlgo::None }
 };
 
+const std::map<std::string, int> poisson_solver_algo_to_int = {
+    {"multigrid", PoissonSolverAlgo::Multigrid},
+    {"fft-based", PoissonSolverAlgo::IntegratedGreenFunction},
```
ax3l: Just curious: Why do we call this FFT-based and not IntegratedGreenFunction/IGF?

Source/Utils/WarpXUtil.cpp
```diff
@@ -469,6 +470,14 @@ void ReadBCParams ()
             ),
             "PEC boundary not implemented for PSATD, yet!"
         );
+
+        if(WarpX::field_boundary_lo[idim] == FieldBoundaryType::Open &&
+           WarpX::field_boundary_hi[idim] == FieldBoundaryType::Open){
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                poisson_solver_id == PoissonSolverAlgo::IntegratedGreenFunction,
+                "Field open boundary conditions are only implemented for the fft-based Poisson solver"
```
ax3l:
```suggestion
                "Field open boundary conditions are only implemented for the FFT-based Poisson solver"
```

Source/WarpX.cpp
```diff
@@ -763,6 +764,27 @@ WarpX::ReadParameters ()
...
+        poisson_solver_id!=PoissonSolverAlgo::IntegratedGreenFunction,
+        "To use the fft-based Poisson solver, compile with WARPX_USE_PSATD=ON.");
+#endif
+
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+        (
+            electrostatic_solver_id!=ElectrostaticSolverAlgo::LabFrameElectroMagnetostatic ||
+            poisson_solver_id!=PoissonSolverAlgo::IntegratedGreenFunction
+        ),
+        "The fft-based Poisson solver is not implemented in labframe-electromagnetostatic mode yet."
```
```suggestion
        "The FFT-based Poisson solver is not implemented in labframe-electromagnetostatic mode yet."
```

Source/ablastr/fields/IntegratedGreenFunctionSolver.H
ax3l: ping @aeriforme

Source/ablastr/fields/PoissonSolver.H
```diff
@@ -100,6 +104,7 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
             amrex::Vector<amrex::DistributionMapping> const dmap,
             amrex::Vector<amrex::BoxArray> const grids,
             T_BoundaryHandler const boundary_handler,
+            [[maybe_unused]] bool is_solver_multigrid,
```
ax3l, line 108: Please add to doxygen string.

Source/ablastr/fields/IntegratedGreenFunctionSolver.H
aeriforme: `inline` was requested by Clang-Tidy. 
Should I create a `.cpp` for the body of this function only then? 
Wouldn't it be less readable?

Source/Utils/WarpXAlgorithmSelection.cpp
aeriforme: I think "Integrated Green Function" is not very common. 
On the other hand, "FFT-based" conveys the most important feature: that is a spectral electrostatic solver. 
If we used "IGF" I believe the users would have to look it up, while with FFT-based they would already have an idea of how the solver works.

RemiLehe: I agree with @aeriforme.
In addition, it fits better with the PICMI interface for the electrostatic solver:
https://picmi-standard.github.io/standard/field.html#electrostatic
Btw, actually, @aeriforme, could we use the exact same names as in PICMI ? (i.e. `Multigrid` and `FFT`)

Source/FieldSolver/ElectrostaticSolver.cpp
```diff
@@ -352,6 +356,10 @@ WarpX::computePhi (const amrex::Vector<std::unique_ptr<amrex::MultiFab> >& rho,
     const std::optional<amrex::Vector<amrex::FArrayBoxFactory const *> > eb_farray_box_factory;
 #endif
 
+    bool const is_solver_multigrid =
+        WarpX::poisson_solver_id == PoissonSolverAlgo::IntegratedGreenFunction ?
+        false : true;
```
ax3l: Clang tidy suggestion:
```suggestion
    bool const is_solver_multigrid =
        WarpX::poisson_solver_id != PoissonSolverAlgo::IntegratedGreenFunction;
```

Source/Utils/WarpXAlgorithmSelection.cpp
ax3l: Sounds good to me, that would be consistent and intuitive.

Docs/source/refs.bib
```diff
@@ -393,3 +393,32 @@ @article{GrismayerNJP2021
...
+  issue = {4},
+  pages = {044204},
+  numpages = {10},
+  year = {2006},
+  month = {Apr},
+  publisher = {American Physical Society},
+  doi = {10.1103/PhysRevSTAB.9.044204},
+  url = {https://link.aps.org/doi/10.1103/PhysRevSTAB.9.044204}
+}
+@article{QiangPhysRevSTAB2006err,
```
ax3l, line 411: From looking at the style of existing entries:

```suggestion

@article{QiangPhysRevSTAB2006err,
```

Docs/source/usage/parameters.rst
```diff
@@ -157,16 +157,28 @@ Overall simulation parameters
...
-    use the solution max norm when checking convergence. If there is no charge
-    density, the MLMG solver will switch to using the initial guess max norm
-    error when evaluating convergence and an absolute error tolerance of
-    :math:`10^{-6}` :math:`\mathrm{V/m}^2` will be used (unless a different
-    non-zero value is specified by the user via
-    ``warpx.self_fields_absolute_tolerance``).
+* ``warpx.poisson_solver`` (`string`) optional (default `multigrid`)
+
+    * ``multigrid``: Poisson's equation is solved using an iterative multigrid (MLMG) solver.
+        See the `AMReX documentation <https://amrex-codes.github.io/amrex/docs_html/LinearSolvers.html#>`_
```
ax3l: Anonymous link (because the link text _AMReX documentation_ is generic and will collide, which causes warnings):
```suggestion
        See the `AMReX documentation <https://amrex-codes.github.io/amrex/docs_html/LinearSolvers.html#>`__
```

Docs/source/usage/parameters.rst
```diff
@@ -157,16 +157,28 @@ Overall simulation parameters
...
+        simulations). The default behavior of the code is to check whether there is
+        non-zero charge density in the system and if so force the MLMG solver to
+        use the solution max norm when checking convergence. If there is no charge
+        density, the MLMG solver will switch to using the initial guess max norm
+        error when evaluating convergence and an absolute error tolerance of
+        :math:`10^{-6}` :math:`\mathrm{V/m}^2` will be used (unless a different
+        non-zero value is specified by the user via
+        ``warpx.self_fields_absolute_tolerance``).
+
+    * ``fft-based``: Poisson's equation is solved using an Integrated Green Function method (which requires FFT calculations).
```
ax3l: As in PICMI?
```suggestion
    * ``fft``: Poisson's equation is solved using an Integrated Green Function method (which requires FFT calculations).
```

Examples/Tests/openbc_poisson_solver/inputs_3d
```diff
@@ -0,0 +1,46 @@
...
+amr.max_level = 0
+amr.max_grid_size = 128
+geometry.dims = 3
+geometry.prob_lo = -5*sigmax -5*sigmay -5*sigmaz
+geometry.prob_hi = +5*sigmax +5*sigmay +5*sigmaz
+boundary.field_lo = open open open
+boundary.field_hi = open open open
+warpx.const_dt = 1e-14
+warpx.do_electrostatic = relativistic
+warpx.poisson_solver = fft-based
```
ax3l: See above:
```suggestion
warpx.poisson_solver = fft
```

Source/FieldSolver/ElectrostaticSolver.cpp
```diff
@@ -249,7 +253,7 @@ WarpX::AddSpaceChargeFieldLabFrame ()
         // Use the tridiag solver with 1D
         computePhiTriDiagonal(rho_fp, phi_fp);
 #else
-        // Use the AMREX MLMG solver otherwise
+        // Use the AMREX MLMG or the IGF solver otherwise
```
ax3l:
```suggestion
        // Use the AMREX MLMG or the FFT (IGF) solver otherwise
```

Source/ablastr/fields/IntegratedGreenFunctionSolver.H
ax3l: Yes, please move the body of the function to a `.cpp` file and only leave the definition in the header (without `inline`).

This is about compile-time - if we unnecessarily put implementations in `.H` files then our compilation times for WarpX and ImpactX will go up over time. `.cpp` files help us to compile the same logic _only once_ and to parallelize by compiling _various_ `.cpp` files _in parallel_.

ax3l (approved):
Awesome :rocket: :sparkles:

Source/Utils/WarpXAlgorithmSelection.cpp
```diff
@@ -55,6 +55,12 @@ const std::map<std::string, int> electrostatic_solver_algo_to_int = {
     {"default", ElectrostaticSolverAlgo::None }
 };
 
+const std::map<std::string, int> poisson_solver_algo_to_int = {
+    {"Multigrid", PoissonSolverAlgo::Multigrid},
+    {"FFT", PoissonSolverAlgo::IntegratedGreenFunction},
```
RemiLehe:
```suggestion
    {"multigrid", PoissonSolverAlgo::Multigrid},
    {"fft", PoissonSolverAlgo::IntegratedGreenFunction},
```

Source/FieldSolver/ElectrostaticSolver.cpp
```diff
@@ -1036,10 +1045,38 @@ void ElectrostaticSolver::PoissonBoundaryHandler::definePhiBCs (const amrex::Geo
...
             }
         }
+
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+            (WarpX::field_boundary_lo[idim] != FieldBoundaryType::Open &&
+            WarpX::field_boundary_hi[idim] != FieldBoundaryType::Open &&
+            WarpX::field_boundary_lo[idim] != FieldBoundaryType::PML &&
+            WarpX::field_boundary_hi[idim] != FieldBoundaryType::PML) ,
+            "Open and PML field boundary conditions only work with "
+            "warpx.poisson_solver = FFT."
```
RemiLehe:
```suggestion
            "warpx.poisson_solver = fft."
```

RemiLehe (approved):

RemiLehe:
@aeriforme Are you able to reproduce the build the Sphinx documentation on your local computer? I wonder why it fails here and not on other PRs.

ax3l:
The `openbc_poisson_solver` test fails with a checksum issue right now.

> Are you able to reproduce the build the Sphinx documentation on your local computer? I wonder why it fails here and not on other PRs.

Found it :tada: Just a small typo, the bib file has invalid syntax. Will fix.

Docs/source/refs.bib
```diff
@@ -394,11 +394,40 @@ @article{GrismayerNJP2021
...
+  journal = {Phys. Rev. ST Accel. Beams},
+  volume = {9},
+  issue = {4},
+  pages = {044204},
+  numpages = {10},
+  year = {2006},
+  month = {Apr},
+  publisher = {American Physical Society},
+  doi = {10.1103/PhysRevSTAB.9.044204},
+  url = {https://link.aps.org/doi/10.1103/PhysRevSTAB.9.044204}
```
ax3l: Not needed:
```suggestion
  doi = {10.1103/PhysRevSTAB.9.044204}
```

Docs/source/refs.bib
```diff
@@ -394,11 +394,40 @@ @article{GrismayerNJP2021
...
+  journal = {Phys. Rev. ST Accel. Beams},
+  volume = {10},
+  issue = {12},
+  pages = {129901},
+  numpages = {2},
+  year = {2007},
+  month = {Dec},
+  publisher = {American Physical Society},
+  doi = {10.1103/PhysRevSTAB.10.129901},
+  url = {https://link.aps.org/doi/10.1103/PhysRevSTAB.10.129901}
```
ax3l: missing closing bracket, url not needed w/ DOI present.
```suggestion
  doi = {10.1103/PhysRevSTAB.10.129901}
}
```

Docs/source/refs.bib
```diff
@@ -394,11 +394,40 @@ @article{GrismayerNJP2021
...
-isbn = {978-3-319-18317-6},
-publisher = {Springer Cham},
-title = {{Particle Accelerator Physics}},
-doi = {https://doi.org/10.1007/978-3-319-18317-6},
-year = {2015}
+  author = {Wiedemann, H.},
+  isbn = {978-3-319-18317-6},
+  publisher = {Springer Cham},
+  title = {{Particle Accelerator Physics}},
+  doi = {https://doi.org/10.1007/978-3-319-18317-6},
```
ax3l: DOI:
```suggestion
  doi = {10.1007/978-3-319-18317-6},
```

Regression/Checksum/benchmarks_json/openbc_poisson_solver.json
```diff
@@ -0,0 +1,20 @@
...
+  "electron": {
+    "particle_momentum_x": 6.37807350553507e-19,
+    "particle_momentum_y": 3.838951101280745e-19,
+    "particle_momentum_z": 1.1454327682972433e-10,
+    "particle_position_x": 21.64260864000591,
+    "particle_position_y": 0.3229614080035443,
+    "particle_position_z": 12582.912000000004,
+    "particle_weight": 19972794743.418434
+  }
+}
```
RemiLehe:
```suggestion
{
  "lev=0": {
    "Bx": 67876594.52390265,
    "By": 112770930.05280717,
    "Bz": 1.1060167389533003e-13,
    "Ex": 3.380787637240387e+16,
    "Ey": 2.034889235345363e+16,
    "Ez": 2601470.9380206033,
    "rho": 5630115646680.221
  },
  "electron": {
    "particle_momentum_x": 6.378073505535043e-19,
    "particle_momentum_y": 3.838951101276526e-19,
    "particle_momentum_z": 1.1454327682972433e-10,
    "particle_position_x": 21.64260864000591,
    "particle_position_y": 0.3229614080035443,
    "particle_position_z": 12582.912000000004,
    "particle_weight": 19972794743.418434
  }
}
```

RemiLehe:
I updated the checksum. It was a negligible change in `Bz`. (`Bz` is supposed to be almost `0` for a relativistic beam propagating along `z`.)

Regression/WarpX-tests.ini
```diff
@@ -4737,3 +4737,21 @@ compileTest = 0
...
+[openbc_poisson_solver]
+buildDir = .
+inputFile = Examples/Tests/openbc_poisson_solver/inputs_3d
+runtime_params = warpx.abort_on_warning_threshold = high
+dim = 3
+addToCompileString = USE_OPENPMD=TRUE USE_PSATD=TRUE
+cmakeSetupOpts = -DWarpX_DIMS=3 -DWarpX_PSATD=ON -DWarpX_OPENPMD=ON
+restartTest = 0
+useMPI = 1
+numprocs = 2
```
ax3l, line 4750: @aeriforme this number (2 MPI ranks) contradicts the limitation that it only works with 1 MPI rank.

Is there an assert in the code to check this is run definitely only with 1 MPI rank?
Is there another assumption we need to check, e.g., only one box?

RemiLehe: The code works for more than 1 MPI rank:
- All operations (e.g., charge deposition and field gather) are done in parallel, using MPI as usual, **except** the Poisson solve
- For the Poisson solve, `rho` is gathered from all MPI rank to a single MPI rank (`ParallalCopy`), which solves the Poisson equation using non-MPI-parallelized FFT, and sends the result back to the other ranks.

RemiLehe: See the PR description above:
```
Also: since the solver uses a global FFT, right now only one MPI rank participates in the FFT solve (but this MPI rank can still use a full GPU) ; this could be improved later by using heFFTe. The rest of the PIC operations (charge deposition, gather, etc.) are still performed by the different MPI ranks in parallel.
```

ax3l: Thanks for clarifying!
That makes sense, I overlooked that.

Source/ablastr/fields/PoissonSolver.H
```diff
@@ -153,6 +159,27 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
                 {{ beta[0], beta[1], beta[2] }};
 #endif
 
+#if (defined(WARPX_USE_PSATD)  && defined(WARPX_DIM_3D))
+        // Use the Integrated Green Function solver (FFT) on the coarsest level if it was selected
```
ax3l, line 163: Small update I would suggest:

- If FFTs were not compiled in, but `is_solver_multigrid=false` is passed to this function, then this silently falls back to a MLGM solve for the lowest level. This should raise a runtime error, because we cannot fulfill the API/user request.
  - An alternative, throwing a warning and "loud" fallback to MLMG, is not ideal in practice, because boundaries/padding & resolution would be quite different.
- I realized the `is_solver_multigrid` name is a but confusing, because we do the IGF only on level 0 and an MLMG on all others in MR. Maybe we find a good alternative name, e.g., `is_solver_igf_on_lev0` (most explicit/good?) or `is_mlmg_all_levels` (kinda like now).

aeriforme: The first point does not really apply to WarpX, I believe. 
There's an assert that checks if FFTs are compiled when the IGF solver is selected. 
It's [here](https://github.com/ECP-WarpX/WarpX/pull/4648/files#diff-1450fa5afa66d2e4dde0bd48fe21eb084d339554ad937d3604376713e053e3c4R774-R778). 
Anyway, I added a similar assert in ablastr in the follow-up PR #4945. 
Let me know if this solution is ok for you.

ax3l: Thanks! Yes, we need the check in ABLASTR directly because that is a library that we also call from ImpactX :pray:

PR #4655, opened 2024-01-31
Docs: update instructions for Adastra supercomputer (CINES, France)

lucafedeli88:
This PR updates instructions and scripts to compile and use WarpX on the Adastra supercomputer (CINES, France).
A peculiarity of this machine is that, by default, there is a  low  per project inode quota. For this reason the instructions have been re-written so that WarpX is installed in a folder shared by all the users of a given project.


ax3l (approved):
That looks good to me.

Do you want to change write permissions, so other users can reuse it but not overwrite by accident your SW dir?

Or even add a per-user directory?

Otherwise, sharing and reusing a single install on the machine is great!

PR #4656, opened 2024-01-31
Docs: update instructions for Fugaku supercomputer (Riken, Japan)

lucafedeli88:
This PR updates the scripts and the instructions to compile WarpX for the Fugaku supercomputer.
The PR introduces also a small change to `BinaryCollision.H` . The reason is that the Fujitsu compiler in clang mode is based on an old version of clang and sees this code as an error (see [here](https://stackoverflow.com/questions/29483120/program-with-noexcept-constructor-accepted-by-gcc-rejected-by-clang), code snippet copied from [here](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86583) ):
```
struct T {
    T() noexcept(false) { }
    ~T() noexcept(false) { }
};

struct A
{
    A() noexcept = default;
    ~A() noexcept = default;
    T t;
};
```
 while more modern compilers silently delete the noexcept methods.
 
 I've also removed `polly` from the compilation flags since otherwise I observe a crash of the Fujitsu compiler. Removing polyhedral optimizations does not seem to affect performances.


ax3l (approved):
Thank you! :sparkles:

PR #4657, opened 2024-01-31
Fix BTD/Scrape Flush Count with Filters

### LINKED ISSUE #4604: Backtransform diagnostics crash when filters remove all particles

ax3l:
Move the counting of already flushed particles for writers that call the I/O backends multiple time per data set, e.g., BTD and boundary scraping, into the I/O backend.

Currently (until #4420), filtering is done as the first step in I/O backends and thus the previous count outside of the I/O backends was over-counting particles that might still get filtered out.

Regression to #4570.

Fix #4604.

### DIFF
--- a/Source/Diagnostics/BTDiagnostics.H
+++ b/Source/Diagnostics/BTDiagnostics.H
@@ -399,8 +399,6 @@ private:
         lab-frame data. */
     void InitializeParticleFunctors () override;
 
-    /** Update total number of particles flushed for all species for ith snapshot */
-    void UpdateTotalParticlesFlushed(int i_buffer);
     /** Reset total number of particles in the particle buffer to 0 for ith snapshot */
     void ResetTotalParticlesInBuffer(int i_buffer);
     /** Clear particle data stored in the particle buffer */
--- a/Source/Diagnostics/BTDiagnostics.cpp
+++ b/Source/Diagnostics/BTDiagnostics.cpp
@@ -1065,12 +1065,13 @@ BTDiagnostics::Flush (int i_buffer, bool force_flush)
         }
     }
     m_flush_format->WriteToFile(
-        m_varnames, m_mf_output[i_buffer], m_geom_output[i_buffer], warpx.getistep(),
-        labtime, m_output_species[i_buffer], nlev_output, file_name, m_file_min_digits,
+        m_varnames, m_mf_output.at(i_buffer), m_geom_output.at(i_buffer), warpx.getistep(),
+        labtime,
+        m_totalParticles_flushed_already.at(i_buffer),
+        m_output_species.at(i_buffer), nlev_output, file_name, m_file_min_digits,
         m_plot_raw_fields, m_plot_raw_fields_guards,
-        use_pinned_pc, isBTD, i_buffer, m_buffer_flush_counter[i_buffer],
-        m_max_buffer_multifabs[i_buffer], m_geom_snapshot[i_buffer][0], isLastBTDFlush,
-        m_totalParticles_flushed_already[i_buffer]);
+        use_pinned_pc, isBTD, i_buffer, m_buffer_flush_counter.at(i_buffer),
+        m_max_buffer_multifabs.at(i_buffer), m_geom_snapshot.at(i_buffer).at(0), isLastBTDFlush);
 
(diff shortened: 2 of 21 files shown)


Source/Diagnostics/BTDiagnostics.H
```diff
@@ -399,8 +399,6 @@ private:
         lab-frame data. */
     void InitializeParticleFunctors () override;
 
-    /** Update total number of particles flushed for all species for ith snapshot */
-    void UpdateTotalParticlesFlushed(int i_buffer);
```
ax3l, line 403: Removed and delegated to the I/O backend, which currently does the filtering and thus knows what number of particles is truly written.

ax3l:
cc @roelof-groenewald just FYI for scraping with filters

ax3l:
Hm, why do
- `scraping_filter`
- `scraping`
- `qed_breit_wheeler_2d_opmd`
- `qed_breit_wheeler_3d_opmd`

fail now...

guj:
Just checked,  the input file in issue 4604 works!

guj:
I noticed orders of inputs are shuffled in several functions. Why is that?

ax3l:
> I noticed orders of inputs are shuffled in several functions. Why is that?

I pass the offsets as references now, thus I cannot use default parameters anymore.

guj:
> > I noticed orders of inputs are shuffled in several functions. Why is that?
> 
> I pass the offsets as references now, thus I cannot use default parameters anymore.

Yes, I saw that. But orders of inputs are different.   For example totalParticlesFlushedAlready was placed a few lines up in FlushFormat.H.  Just curious.

ax3l:
Simplified it further to get rid of state that we will forget during checkpoint-restart :) Too much back and forth over the frontend-backend APIs.

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -1271,10 +1269,10 @@ void BTDiagnostics::MergeBuffersForPlotfile (int i_snapshot)
                 InterleaveSpeciesHeader(recent_species_Header,snapshot_species_Header,
                                         m_output_species_names[i], m_buffer_flush_counter[i_snapshot]);
                 if (BufferSpeciesHeader.m_total_particles == 0) { continue; }
-                if (m_totalParticles_flushed_already[i_snapshot][i]==0) {
-                WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
-                    std::rename(recent_ParticleHdrFilename.c_str(), snapshot_ParticleHdrFilename.c_str()) == 0,
-                    std::string("Renaming ").append(recent_ParticleHdrFilename).append(" to ").append(snapshot_ParticleHdrFilename).append(" has failed"));
+                if (!amrex::FileExists(snapshot_ParticleHdrFilename)) {
```
ax3l, line 1272: @RevathiJambunathan @atmyers highlighting this: I want to remove `m_totalParticles_flushed_already` because it is brittle state.

Let me know if you would add another check for this line.

RevathiJambunathan: I think this is good!

Source/Diagnostics/BTDiagnostics.H
RevathiJambunathan: is this also done for plotfile?

RevathiJambunathan: ah I see - its not needed for plotfile - because we just merge headers and interleave data . this was needed mainly for openpmd. 
all good!

RevathiJambunathan (approved):

RevathiJambunathan:
Looks good to me! @ax3l :) Thanks for this PR

RemiLehe (approved):

PR #4659, opened 2024-02-01
ParticleHistogram2D change openPMD access type

n01r:
Changed openPMD access type for file creation in ParticleHistogram2D from `APPEND` to `CREATE` to avoid warnings on execution.

Additionally:
- make paths compatible with Windows
- allow to set the runtime option `file_min_digits`

### DIFF
--- a/Source/Diagnostics/ReducedDiags/ParticleHistogram2D.cpp
+++ b/Source/Diagnostics/ReducedDiags/ParticleHistogram2D.cpp
@@ -60,6 +60,7 @@ ParticleHistogram2D::ParticleHistogram2D (std::string rd_name)
     ParmParse pp_rd_name(rd_name);
 
     pp_rd_name.query("openpmd_backend", m_openpmd_backend);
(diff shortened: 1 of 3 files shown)


Source/Diagnostics/ReducedDiags/ParticleHistogram2D.cpp
```diff
@@ -60,6 +60,7 @@ ParticleHistogram2D::ParticleHistogram2D (std::string rd_name)
     ParmParse pp_rd_name(rd_name);
 
     pp_rd_name.query("openpmd_backend", m_openpmd_backend);
+    pp_rd_name.query("file_min_digits", m_file_min_digits);
```
ax3l, line 63: Awesome, please add to `parameters.rst` in the docs :)

n01r: Ah, right, forgot that.

Thanks, added :)

ax3l (approved):
Nice, thx! :)

PR #4661, opened 2024-02-02
Add hybrid resistivity current term

clarkse:
This PR adds a current magnitude term in the hybrid resistivity to allow for scaling of the resistivity as a function of the magnitude of total J.


Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -589,6 +589,12 @@ void FiniteDifferenceSolver::HybridPICSolveECylindrical (
                 // Interpolate to get the appropriate charge density in space
                 Real rho_val = Interp(rho, nodal, Er_stag, coarsen, i, j, 0, 0);
 
+                // Interpolate current to appropriate staggering to match E field
+                Real jr_val = Interp(Jr, Jr_stag, Er_stag, coarsen, i, j, 0, 0);
+                Real jt_val = Interp(Jt, Jt_stag, Er_stag, coarsen, i, j, 0, 0);
+                Real jz_val = Interp(Jz, Jz_stag, Er_stag, coarsen, i, j, 0, 0);
+                Real jtot_val = sqrt(pow(jr_val, 2) + pow(jt_val, 2) + pow(jz_val,2));
```
roelof-groenewald: Could we wrap this in an `if` statement based on `m_resistivity_has_J_dependence`? Maybe you are already planning that.

clarkse: I just pushed this change to the PR.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -907,6 +954,15 @@ void FiniteDifferenceSolver::HybridPICSolveECartesian (
                 // Interpolate to get the appropriate charge density in space
                 Real rho_val = Interp(rho, nodal, Ez_stag, coarsen, i, j, k, 0);
 
+                // Interpolate current to appropriate staggering to match E field
+                Real jtot_val = 0.;
+                if (include_resistivity_term && resistivity_has_J_dependence) {
+                    Real jx_val = Interp(Jx, Jx_stag, Ex_stag, coarsen, i, j, k, 0);
+                    Real jy_val = Interp(Jy, Jy_stag, Ex_stag, coarsen, i, j, k, 0);
+                    Real jz_val = Interp(Jz, Jz_stag, Ex_stag, coarsen, i, j, k, 0);
```
roelof-groenewald:
```suggestion
                    Real jx_val = Interp(Jx, Jx_stag, Ez_stag, coarsen, i, j, k, 0);
                    Real jy_val = Interp(Jy, Jy_stag, Ez_stag, coarsen, i, j, k, 0);
                    Real jz_val = Interp(Jz, Jz_stag, Ez_stag, coarsen, i, j, k, 0);
```

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -589,6 +590,15 @@ void FiniteDifferenceSolver::HybridPICSolveECylindrical (
                 // Interpolate to get the appropriate charge density in space
                 Real rho_val = Interp(rho, nodal, Er_stag, coarsen, i, j, 0, 0);
 
+                // Interpolate current to appropriate staggering to match E field
+                Real jtot_val = 0.;
+                if (include_resistivity_term && resistivity_has_J_dependence) {
+                    Real jr_val = Interp(Jr, Jr_stag, Er_stag, coarsen, i, j, 0, 0);
+                    Real jt_val = Interp(Jt, Jt_stag, Er_stag, coarsen, i, j, 0, 0);
+                    Real jz_val = Interp(Jz, Jz_stag, Er_stag, coarsen, i, j, 0, 0);
+                    jtot_val = sqrt(pow(jr_val, 2) + pow(jt_val, 2) + pow(jz_val,2));
```
roelof-groenewald: I think the preferred way to calculate it would be:
```suggestion
                    jtot_val = sqrt(jr_val*jr_val + jt_val*jt_val + jz_val*jz_val);
```
and same for the other directions and Cartesian case.

roelof-groenewald (commented):
Thanks for the check to only calculate `j_tot` if it will be used.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
clarkse: Thanks for catching the staggering bug.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
clarkse: I have dropped using the pow call.  Just curious though, with compiler optimization wouldn't pow(,2) be inlined anyway?

roelof-groenewald: Yes, with new enough compilers it should be but the reasoning is that with a simple square, the `x*x` approach will never be less efficient than `pow(x, 2)` whereas the latter could be less efficient with sub-optimal compiler flags.

roelof-groenewald: So, actually with the change I don't think you need the `static_cast<Real>` anymore if you use `std::sqrt`, even in a lambda function now (https://github.com/ECP-WarpX/WarpX/pull/4490).

roelof-groenewald (approved):
Thanks! I'll be curious to hear how you are using the current dependent resistivity.

PR #4636, opened 2024-01-25
Add install instructions for ALCF's Polaris

roelof-groenewald:
Adding instructions for Polaris. These instructions are **mostly** copied from the Perlmutter instructions.
I'll make a couple comments directly on the changes for some issues I'm still having.

I've only tested GPU installation which is why I left the CPU instructions as *Under construction*.

PR #4664, opened 2024-02-03
Update Polaris install instructions

roelof-groenewald:
Follow-up to #4636.

### DIFF
--- a/Tools/machines/polaris-alcf/polaris_gpu_warpx.profile.example
+++ b/Tools/machines/polaris-alcf/polaris_gpu_warpx.profile.example
@@ -1,12 +1,16 @@
 # Set the project name
 export proj=""  # change me!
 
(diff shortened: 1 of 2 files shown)


Tools/machines/polaris-alcf/polaris_gpu_warpx.profile.example
```diff
@@ -1,12 +1,16 @@
 # Set the project name
 export proj=""  # change me!
 
+# swap to GNU programming environment (with gcc 11.2)
+module swap PrgEnv-nvhpc PrgEnv-gnu
+module swap gcc/12.2.0 gcc/11.2.0
```
ax3l, line 6: Checks out, CUDA 11.8 only works with GCC 11:
https://gist.github.com/ax3l/9489132

ax3l (approved):

PR #4667, opened 2024-02-04
RZ/RCYLINDER/RSPHERE: Rename `PIdx:x` to `PIdx:r`

ax3l:
Update `PIdx`:
- RZ/RCYLINDER/RSPHERE: rename first component from `x` to `r`.

Follow-ups could be:
- RZ: move `theta` after `r`, `z`.
- All: move `w` (weight) after momenta.


ax3l:
In the checksums, the only values that really change are
```
Benchmark: [driver,particle_theta] 1.570790043609588e+06
Test file: [driver,particle_theta] 2.306055207176163e+06
```
and sometimes `Br`. Could be from different (random) `theta` values in init?

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -220,7 +220,9 @@ namespace
         ) noexcept
     {
         pa[PIdx::z][ip] = 0._rt;
-#if (AMREX_SPACEDIM >= 2)
+#ifdef WARPX_DIM_RZ
+        pa[PIdx::r][ip] = 0._rt;
+#elif (AMREX_SPACEDIM >= 2)
```
ax3l: style:
```suggestion
#elif !defined(WARPX_DIM_1D_Z)
```

Source/Particles/NamedComponentParticleContainer.H
```diff
@@ -78,7 +82,9 @@ public:
     NamedComponentParticleContainer (amrex::AmrParGDB* amr_pgdb)
     : amrex::ParticleContainerPureSoA<PIdx::nattribs, 0, T_Allocator>(amr_pgdb) {
         // build up the map of string names to particle component numbers
-#if !defined (WARPX_DIM_1D_Z)
+#ifdef WARPX_DIM_RZ
+        particle_comps["r"]  = PIdx::r;
```
ax3l: Change `particle_container.py` with respect to `get_particle_r`.

ax3l: Maybe add to any of the RZ PICMI tests an access to positions to cover it.

ax3l:
@RemiLehe I have this rename of the `r` component in the code another push, let's see if we can get it merged :)

EZoni:
@ax3l @RemiLehe @dpgrote @JustinRayAngus 

While going through long-standing PRs that have no reason to be stalled besides resources, I came across this one, which I do think improves the readability of the code (in that it avoids things like `x = std::sqrt(xp*xp + yp*yp)` and similar situations where `x` on the left side means something geometrically different than `x` on the right side). 

Therefore, I merged `development` and fixed the remaining build errors. Let's see if we can finalize it eventually.

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.cpp
```diff
@@ -178,16 +178,16 @@ void GenerateVirtualPhotons (MultiParticleContainer* mypc){
                     const amrex::ParticleReal x  = soa.m_rdata[PIdx::x][i];
                     const amrex::ParticleReal z  = soa.m_rdata[PIdx::z][i];
 #elif defined (WARPX_DIM_RZ)
-                    const amrex::ParticleReal x  = soa.m_rdata[PIdx::x][i];
+                    const amrex::ParticleReal x  = soa.m_rdata[PIdx::r][i];
```
dpgrote: It looks like this can be
```
const amrex::ParticleReal r  = soa.m_rdata[PIdx::r][I];
```
and changing the corresponding line below.

Source/Particles/Pusher/GetAndSetPosition.H
dpgrote, line 82: You can add `m_r` to here and use it below to avoid setting `m_x` from `PIdx::r`

Source/Particles/ParticleBoundaryBuffer.cpp
```diff
@@ -164,7 +164,7 @@ struct FindEmbeddedBoundaryIntersection {
         dst.m_runtime_rdata[m_normal_index+1][dst_i] = 0.0;
         dst.m_runtime_rdata[m_normal_index+2][dst_i] = 0.0;
 #elif (defined WARPX_DIM_RCYLINDER) || (defined WARPX_DIM_RSPHERE)
-        dst.m_rdata[PIdx::x][dst_i] = x_temp;
+        dst.m_rdata[PIdx::r][dst_i] = x_temp;
```
dpgrote: This looks like a bug - it should be `std::sqrt(x_temp*x_temp + y_temp*y_temp)` for RCYLINDER and `std::sqrt(x_temp*x_temp + y_temp*y_temp + z_temp*z_temp)` for RSPHERE. Though, this is basically moot since the embedded boundary is not supported in 1D.

EZoni: Applied in e6ba5bde9afa9abe3b244b24762ba6b5a0a0b0db.

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -344,11 +344,11 @@ public:
 #elif defined(WARPX_DIM_XZ)
                     constexpr std::array<int, 2> position_indices = {PIdx::x, PIdx::z};
 #elif defined(WARPX_DIM_RZ)
-                    constexpr std::array<int, 3> position_indices = {PIdx::x, PIdx::z, PIdx::theta};
+                    constexpr std::array<int, 3> position_indices = {PIdx::r, PIdx::z, PIdx::theta};
```
JustinRayAngus, line 347: Maybe minor, but does it make sense to do r,theta,z, which is in cyclic order for a right-handed coordinate system?

JustinRayAngus (commented):
Looks good. I left a more comment.

JustinRayAngus (approved):

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.cpp
EZoni: Applied in [1b2dbef](https://github.com/BLAST-WarpX/warpx/pull/4667/commits/1b2dbef97fc931d5a12df3280437f90cd3bb0f97).

Source/Particles/Pusher/GetAndSetPosition.H
EZoni: This would require a few more changes in GetAndSetPosition.H, so I would maybe leave it for a separate PR and keep this one focused on the PIdx:x to PIdx:r renaming as much as possible, if that's okay with you? For example, I see inconsistencies in what `x` means in `operator()` vs. `AsStored()`.

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
EZoni: Similarly to my reply to @dpgrote's https://github.com/BLAST-WarpX/warpx/pull/4667#discussion_r3327363854, I agree with the suggestion but I would implement it in a separate PR, just in case it brings more changes than expected, so we keep this PR focused on its original goal.

dpgrote: This makes sense, however AMReX needs the positions relative to grid geometry first and in the correct order. The `r` and `z` are used by AMReX to keep particles associated with the grid patches. So the ordering in memory will be `r`, `z`, `theta`.

EZoni:
@dpgrote 

Does this PR look good to you now, taking into account the latest comments and replies? If so, please consider approving. Then I'll ping the author and assignee to re-review and finalize.

dpgrote (approved):
Looks good!

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
ax3l: Jup, might cause more confusion than good if we mix the index and SoA order (even if technically separate).

PR #4677, opened 2024-02-08
Fix docs for checksumAPI terminal call

n01r:
Changed `--format` to `--output-format` like the call signature demands


roelof-groenewald (approved):
Thanks!

PR #4681, opened 2024-02-08
[WIP] Add particle thermal boundary condition

RemiLehe:
TODO:
- [ ] Prevent laser particles from calling the boundary condition routines (they should just be absorbed)
- [ ] Test with a laser pulse

### DIFF
--- a/Source/Particles/ParticleBoundaries_K.H
+++ b/Source/Particles/ParticleBoundaries_K.H
@@ -210,14 +210,14 @@ namespace {
             uy = ur*std::sin(y) + ut*std::cos(y);
         }
 #else
-        ux = amrex::RandomNormal(0, boundaries.u_th, engine);
-        uy = amrex::RandomNormal(0, boundaries.u_th, engine);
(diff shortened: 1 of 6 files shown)


Source/Particles/ParticleBoundaries_K.H
```diff
@@ -64,6 +64,76 @@ namespace ApplyParticleBoundaries {
         }
     }
 
+
+namespace {
```
RemiLehe, line 68: This function should be in a separate `.H` file.

archermarx:
Neat! My understanding is this is effectively a diffuse thermal reflection BC, so particles that hit the boundary have their velocities resampled according to a provided thermal velocity?

RemiLehe:
@archermarx Yes, that's correct.
It is supposed to model what happens if the simulation box was embedded in an infinite plasma with a finite temperature.
The PR also includes some changes for the field boundary conditions, which I will later extract in a separate PR to clean things up.

PR #4684, opened 2024-02-09
Clang tidy CI test: add several clang-analyzer-* checks to clang tidy CI test

lucafedeli88:
This PR adds several `clang-analyzer-*` checks to `clang-tidy` CI test:

- [clang-analyzer-core.DynamicTypePropagation](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/clang-analyzer-core.DynamicTypePropagation.html)
- [clang-analyzer-core.uninitialized.CapturedBlockVariable](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/clang-analyzer-core.uninitialized.CapturedBlockVariable.html)
- [clang-analyzer-cplusplus.InnerPointer](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/clang-analyzer-cplusplus.InnerPointer.html)
-~~[clang-analyzer-optin.mpi.MPI-Checker](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/clang-analyzer-optin.mpi.MPI-Checker.html)~~
- [clang-analyzer-nullability.NullableReturnedFromNonnull](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/clang-analyzer-nullability.NullableReturnedFromNonnull.html)
- [clang-analyzer-optin.performance.GCDAntipattern](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/clang-analyzer-optin.performance.GCDAntipattern.html)
- ~~[clang-analyzer-optin.performance.Padding](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/clang-analyzer-optin.performance.Padding.html)~~
- [clang-analyzer-optin.portability.UnixAPI](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/clang-analyzer-optin.portability.UnixAPI.html)
- [clang-analyzer-valist.CopyToSelf](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/clang-analyzer-valist.CopyToSelf.html)
- [clang-analyzer-valist.Uninitialized](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/clang-analyzer-valist.Uninitialized.html)
- [clang-analyzer-valist.Unterminated](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/clang-analyzer-valist.Unterminated.html)

The following checks are explicitly excluded (`osx` checks are not relevant for WarpX):
- clang-analyzer-osx.*
- clang-analyzer-optin.osx.*

Also this check is excluded, since it generates a non-relevant warning:
- [clang-analyzer-optin.mpi.MPI-Checker](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/clang-analyzer-optin.mpi.MPI-Checker.html)

Finally, padding check is excluded because it generates warning for lambda functions, which require a verbose `NOLINT` to be silenced:
- [clang-analyzer-optin.performance.Padding](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/clang-analyzer-optin.performance.Padding.html)

___

### Use new `AMREX_ASSUME` feature to silence warnings

In WarpX we can enforce a given condition by using `WARPX_ALWAYS_ASSERT_WITH_MESSAGE` . However, the compiler can't see through  `WARPX_ALWAYS_ASSERT_WITH_MESSAGE` and `clang-tidy` may therefore generate warnings for scenarios that can't actually take place. 

A possible solution is to use a newly introduced feature in AMReX: `AMREX_ASSUME` :
```
#include <AMReX_Extension.H>
//...
WARPX_ALWAYS_ASSERT_WITH_MESSAGE(m_mf_src != nullptr, "m_mf_src can't be a nullptr.");
AMREX_ASSUME(m_mf_src != nullptr);
```

After `AMREX_ASSUME(m_mf_src != nullptr);` most compilers (including `gcc`, `clang`, and `nvcc`) should be able to assume that `m_mf_src != nullptr`.

Using this feature requires to update the AMReX version.

### Change getters of fields and field pointers in WarpX

As discussed with @ax3l (see https://github.com/ECP-WarpX/WarpX/pull/4684#discussion_r1495358308), this PR also changes the getters of the fields and field pointers in the WarpX class:

```
enum struct FieldType : int
{
    Efield_aux,
    Bfield_aux,
    Efield_fp,
    Bfield_fp,
    current_fp,
    current_fp_nodal,
    rho_fp,
    F_fp,
    G_fp,
    phi_fp,
    vector_potential_fp,
    Efield_cp,
    Bfield_cp,
    current_cp,
    rho_cp,
    F_cp,
    G_cp,
    edge_lengths,
    face_areas,
    Efield_avg_fp,
    Bfield_avg_fp,
    Efield_avg_cp,
    Bfield_avg_cp
};

//...

[[nodiscard]] bool
isFieldInitialized (FieldType field_type, int lev, int direction = 0) const;

[[nodiscard]] amrex::MultiFab*
getFieldPointer (FieldType field_type, int lev, int direction = 0) const;

[[nodiscard]] std::array<const amrex::MultiFab* const, 3>
getFieldPointerArray (FieldType field_type, int lev) const;

const amrex::MultiFab&
getField(FieldType field_type, int lev, int direction = 0) const;

//...
private:

[[nodiscard]] amrex::MultiFab*
getFieldPointerUnchecked (FieldType field_type, int lev, int direction = 0) const;

```

`getFieldPointerUnchecked` does **not** check if the requested pointer is ` != nullptr` . This function is conceived to be used internally by the other getters, which  check if the pointer is valid and, in case it isn't, they abort.


Source/Diagnostics/ComputeDiagFunctors/JFunctor.cpp
```diff
@@ -54,6 +54,12 @@ JFunctor::operator() (amrex::MultiFab& mf_dst, int dcomp, const int /*i_buffer*/
         mypc.DepositCurrent(current_fp_temp, warpx.getdt(m_lev), 0.0);
     }
 
-    InterpolateMFForDiag(mf_dst, *m_mf_src, dcomp, warpx.DistributionMap(m_lev),
-                         m_convertRZmodes2cartesian);
+    if (m_mf_src != nullptr){
```
ax3l: The function `get_pointer_current_fp` should throw instead of having to check the return value in user code.

Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -407,44 +407,39 @@ FullDiagnostics::AddRZModesToDiags (int lev)
     // E
     for (int dim=0; dim<3; dim++){
         // 3 components, r theta z
-        m_all_field_functors[lev][icomp] =
+        m_all_field_functors[lev][icomp++] =
```
ax3l: Although formally correct, this can be quite confusing for our average developer.

Most people will not immediately know/see that `icomp` is increased after this line is executed (`icomp++` vs `++icomp`), which can lead to subtle bugs at the tail of the loop over time when people assume a different value for `icomp`, don't you think?

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1202,7 +1202,7 @@ PhysicalParticleContainer::AddPlasma (PlasmaInjector const& plasma_injector, int
         const bool rz_random_theta = m_rz_random_theta;
 #endif
         amrex::ParallelForRNG(overlap_box,
-        [=] AMREX_GPU_DEVICE (int i, int j, int k, amrex::RandomEngine const& engine) noexcept
+        [=] AMREX_GPU_DEVICE (int i, int j, int k, amrex::RandomEngine const& engine) noexcept //NOLINT(clang-analyzer-optin.performance.Padding)
```
ax3l: Does this belong in AMReX or at least an AMReX issue instead?

Source/ablastr/utils/SignalHandling.cpp
```diff
@@ -191,7 +191,7 @@ SignalHandling::WaitSignals ()
     }
 
 #if defined(AMREX_USE_MPI)
-    BL_MPI_REQUIRE(MPI_Wait(&signal_mpi_ibcast_request, MPI_STATUS_IGNORE));
+    BL_MPI_REQUIRE(MPI_Wait(&signal_mpi_ibcast_request, MPI_STATUS_IGNORE)); // NOLINT(clang-analyzer-optin.mpi.MPI-Checker)
```
ax3l: This seems like a common pattern that could need a lot of silencing down the road. Would not add this test.

Source/ablastr/parallelization/MPIInitHelpers.H
```diff
@@ -15,7 +15,7 @@ namespace ablastr::parallelization
      *
      * @return the MPI_THREAD_* level required for MPI_Init_thread
      */
-    int
+    constexpr int
```
ax3l, line 18: Sure, possible given the implementation.

In practice, never compile-time because the `.cpp` file is not seen in other TUs.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -2825,11 +2825,11 @@ PhysicalParticleContainer::PushPX (WarpXParIter& pti,
...
-    amrex::ParallelFor(TypeList<CompileTimeOptions<no_exteb,has_exteb>,
-                                CompileTimeOptions<no_qed  ,has_qed>>{},
-                       {exteb_runtime_flag, qed_runtime_flag},
-                       np_to_push, [=] AMREX_GPU_DEVICE (long ip, auto exteb_control,
-                                                         auto qed_control)
+    amrex::ParallelFor(
+        TypeList<CompileTimeOptions<no_exteb,has_exteb>, CompileTimeOptions<no_qed  ,has_qed>>{},
+        {exteb_runtime_flag, qed_runtime_flag},
+        np_to_push,
+        [=] AMREX_GPU_DEVICE (long ip, auto exteb_control, auto qed_control) //NOLINT(clang-analyzer-optin.performance.Padding)
```
ax3l: This is so verbose that I recommend not adding the test to CI.

Source/Parallelization/WarpXComm_K.H
```diff
@@ -237,20 +231,18 @@ void warpx_interp (int j, int k, int l,
...
-    wl = 1.0_rt;
     for         (int jj = 0; jj < nj; jj++) {
         for     (int kk = 0; kk < nk; kk++) {
             for (int ll = 0; ll < nl; ll++) {
-                wj = (rj - amrex::Math::abs(jn - (jc + jj) * rj)) / static_cast<amrex::Real>(rj);
+                auto c = arr_coarse_zeropad(jc+jj,kc+kk,lc+ll);
+                c *= (rj - amrex::Math::abs(jn - (jc + jj) * rj)) / static_cast<amrex::Real>(rj);
 #if (AMREX_SPACEDIM >= 2)
-                wk = (rk - amrex::Math::abs(kn - (kc + kk) * rk)) / static_cast<amrex::Real>(rk);
+                c *= (rk - amrex::Math::abs(kn - (kc + kk) * rk)) / static_cast<amrex::Real>(rk);
```
ax3l, line 240: Ok but what is the difference? I would assume any optimizing compiler to do the same (with fastmath enabled).

I like the change, but I am not sure which test triggered it and if we want it in CI :)

Source/Diagnostics/ComputeDiagFunctors/JFunctor.cpp
lucafedeli88: In order to address this issue, would it be okay to restructure this part of `WarpX.H`
```
    [[nodiscard]] amrex::MultiFab * get_pointer_Efield_aux  (int lev, int direction) const { return Efield_aux[lev][direction].get(); }
    [[nodiscard]] amrex::MultiFab * get_pointer_Bfield_aux  (int lev, int direction) const { return Bfield_aux[lev][direction].get(); }

    [[nodiscard]] amrex::MultiFab * get_pointer_Efield_fp  (int lev, int direction) const { return Efield_fp[lev][direction].get(); }
    [[nodiscard]] amrex::MultiFab * get_pointer_Bfield_fp  (int lev, int direction) const { return Bfield_fp[lev][direction].get(); }
    [[nodiscard]] amrex::MultiFab * get_pointer_current_fp  (int lev, int direction) const { return current_fp[lev][direction].get(); }
    [[nodiscard]] amrex::MultiFab * get_pointer_current_fp_nodal  (int lev, int direction) const { return current_fp_nodal[lev][direction].get(); }
    [[nodiscard]] amrex::MultiFab * get_pointer_rho_fp  (int lev) const { return rho_fp[lev].get(); }
    [[nodiscard]] amrex::MultiFab * get_pointer_F_fp  (int lev) const { return F_fp[lev].get(); }
    [[nodiscard]] amrex::MultiFab * get_pointer_G_fp  (int lev) const { return G_fp[lev].get(); }
    [[nodiscard]] amrex::MultiFab * get_pointer_phi_fp  (int lev) const { return phi_fp[lev].get(); }
    [[nodiscard]] amrex::MultiFab * get_pointer_vector_potential_fp  (int lev, int direction) const { return vector_potential_fp_nodal[lev][direction].get(); }

    [[nodiscard]] amrex::MultiFab * get_pointer_Efield_cp  (int lev, int direction) const { return Efield_cp[lev][direction].get(); }
    [[nodiscard]] amrex::MultiFab * get_pointer_Bfield_cp  (int lev, int direction) const { return Bfield_cp[lev][direction].get(); }
    [[nodiscard]] amrex::MultiFab * get_pointer_current_cp  (int lev, int direction) const { return current_cp[lev][direction].get(); }
    [[nodiscard]] amrex::MultiFab * get_pointer_rho_cp  (int lev) const { return rho_cp[lev].get(); }
    [[nodiscard]] amrex::MultiFab * get_pointer_F_cp  (int lev) const { return F_cp[lev].get(); }
    [[nodiscard]] amrex::MultiFab * get_pointer_G_cp  (int lev) const { return G_cp[lev].get(); }

    [[nodiscard]] amrex::MultiFab * get_pointer_edge_lengths  (int lev, int direction) const { return m_edge_lengths[lev][direction].get(); }
    [[nodiscard]] amrex::MultiFab * get_pointer_face_areas  (int lev, int direction) const { return m_face_areas[lev][direction].get(); }
```
with something like:

```
enum class FieldType
{
    Efield_aux,
    Bfield_aux
    //  ...
};



    template <FieldType Type>
    [[nodiscard]] amrex::MultiFab * get_pointer (int lev, int direction)
    {
        if (auto* ptr = m_multifabs[lev][direction][Type].get(); ptr != nullptr)
            return ptr; 
        else
            WARPX_ABORT_WITH_MESSAGE("...");
    }
```
?

Source/Diagnostics/ComputeDiagFunctors/CellCenterFunctor.cpp
```diff
@@ -17,6 +17,12 @@ void
 CellCenterFunctor::operator()(amrex::MultiFab& mf_dst, int dcomp, const int /*i_buffer*/) const
 {
     auto& warpx = WarpX::GetInstance();
-    InterpolateMFForDiag(mf_dst, *m_mf_src, dcomp, warpx.DistributionMap(m_lev),
-                         m_convertRZmodes2cartesian);
+    if (m_mf_src != nullptr){
```
ax3l: I like the defensive programming here. Small suggestion: Instead of the `if` I would write an always assert macro on function entry here. This can go together with other prerequisite checks of the functions.

Source/Diagnostics/ComputeDiagFunctors/BackTransformFunctor.cpp
```diff
@@ -60,13 +60,20 @@ BackTransformFunctor::operator ()(amrex::MultiFab& mf_dst, int /*dcomp*/, const
...
         // Generate slice of the cell-centered multifab containing boosted-frame field-data
         // at current z-boost location for the ith buffer
-        slice = amrex::get_slice_data(moving_window_dir,
-                                     m_current_z_boost[i_buffer],
-                                     *m_mf_src,
-                                     geom,
-                                     scomp,
-                                     m_mf_src->nComp(),
-                                     interpolate);
+        if (m_mf_src != nullptr){
```
ax3l: I like the defensive programming here. Small suggestion: Instead of the if I would write an always assert macro here. This can go together with other prerequisite checks of the functions.

Source/Particles/PhysicalParticleContainer.cpp
lucafedeli88: I think that it is something that can happen with any lambda function (e.g., you can have a look here: https://godbolt.org/z/fK6PhvdsG). I suppose it depends on the order in which the variables captured by the lambda function are actually used.

Source/ablastr/utils/SignalHandling.cpp
lucafedeli88: It's the only check of the `clang-analyzer-*` group not enabled in AMReX. It could be OK to disable it also for WarpX.

Source/Parallelization/WarpXComm_K.H
lucafedeli88: The test that triggered this is `clang-analyzer-deadcode.DeadStores`.  This test flags as issues scenarios like:
```
int a = 1;
# ifdef FLAG
    a = 2;
#endif

do_stuff(a);
```
if  `FLAG` is defined the value stored initially in `a` is never used.

lucafedeli88: I think that in general it is a good test, although here it forces us to change a perfectly valid code. I do prefer the new version though, because  I sometimes find code like
```
int a = 1;
# ifdef FLAG
    a = 2;
#endif
```
a bit confusing.

lucafedeli88: Of course there is no difference in terms of performances.

Source/Diagnostics/FullDiagnostics.cpp
lucafedeli88: What about removing `icomp` completely and using `push_back` ?

Source/ablastr/utils/SignalHandling.cpp
lucafedeli88: Done. The check has been disabled and `//NOLINT` has been removed.

Source/ablastr/parallelization/MPIInitHelpers.H
lucafedeli88: `mpi_thread_required()` is called only twice in the code, inside `MPIInitHelpers.cpp` .
With my latest commit I've changed 
` int thread_required = mpi_thread_required();` into `constexpr int thread_required = mpi_thread_required();`, to make sure that the function is computed at compile-time.

Source/Particles/PhysicalParticleContainer.cpp
lucafedeli88: I've removed the `clang-analyzer-optin.performance.Padding` check and the corresponding `//NOLINT` directives.

Source/Diagnostics/ComputeDiagFunctors/CellCenterFunctor.cpp
lucafedeli88: I would agree in principle. However, unfortunately, `clang-tidy` can't see through `WARPX_ALWAYS_ASSERT_WITH_MESSAGE`  :-(  :

 ```
/home/luca/Projects/WarpX/Source/Diagnostics/ComputeDiagFunctors/CellCenterFunctor.cpp:22:7: warning: Forming reference to null pointer [clang-analyzer-core.NonNullParamChecker]
      InterpolateMFForDiag(
      ^
/home/luca/Projects/WarpX/Source/Diagnostics/ComputeDiagFunctors/CellCenterFunctor.cpp:20:5: note: '?' condition is false
    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(m_mf_src != nullptr,
    ^
/home/luca/Projects/WarpX/Source/Utils/TextMsg.H:13:50: note: expanded from macro 'WARPX_ALWAYS_ASSERT_WITH_MESSAGE'
#define WARPX_ALWAYS_ASSERT_WITH_MESSAGE(EX,MSG) (EX)?((void)0) : ablastr::utils::TextMsg::Assert( # EX , __FILE__, __LINE__ , MSG)
                        
 ```
 I am not sure if we can do something about that. That is anyway the reason why I used `WARPX_ABORT_WITH_MESSAGE`

Source/Diagnostics/ComputeDiagFunctors/BackTransformFunctor.cpp
lucafedeli88: Same as above. Unfortunately using `WARPX_ALWAYS_ASSERT_WITH_MESSAGE` does not work here....

WeiqunZhang:
Maybe adding a throw to the end of the assert function could help.

lucafedeli88:
> WARPX_ALWAYS_ASSERT_WITH_MESSAGE

Very indirectly there is an assert: `WARPX_ALWAYS_ASSERT_WITH_MESSAGE` calls `ablastr::utils::TextMsg::Assert` which calls `amrex::Assert` which finally calls `assert`. However, `ablastr::utils::TextMsg::Assert` is in a `.cpp` file, so the compiler may not be able to figure that out. Should I move it to the corresponding `.H` file and force inlining?

WeiqunZhang:
I don't think it will help, especially when MPI is on, we call MPI_Abort eventually.

lucafedeli88:
> I don't think it will help, especially when MPI is on, we call MPI_Abort eventually.

I  see... Then, I  am not sure about what could be the ideal solution among the following:
- we drop the check from CI
- we use `NOLINT`
- we keep the `if (m_mf_src != nullptr)`
- we add something like `AMREX_ASSUME(...)` in AMReX so that it uses under the hood `__builtin_assume(...)` (clang), or `__attribute__((assume(...)))` (GNU compiler), or `[[assume( ... )]] ` (C++23) and hope that `AMREX_ASSUME(m_mf_src != nullptr)` after `WARPX_ALWAYS_ASSERT_WITH_MESSAGE` is enough.
- We add an additional `assert` inside `WARPX_ALWAYS_ASSERT_WITH_MESSAGE`

Source/Diagnostics/ComputeDiagFunctors/JFunctor.cpp
ax3l: Yes, maybe we can call it `get_mf_pointer` or `get_field_pointer`?

Not sure the `FieldType` must be a template - this is not used in hot loops so it can be a regular argument, no?

Source/Diagnostics/FullDiagnostics.cpp
ax3l: That works well :+1:

Source/Particles/PhysicalParticleContainer.cpp
ax3l: Let's open this as an issue in AMReX, could be interesting for @WeiqunZhang.

Source/ablastr/parallelization/MPIInitHelpers.H
ax3l: Ah, inside `MPIInitHelpers.cpp` the implementation is know. Makes sense.

Source/Parallelization/WarpXComm_K.H
ax3l: Agreed, new looks better.

ax3l:
`AMREX_ASSUME` to drop in `[[assume(same-expression == true)]]` in _after_ an `ALWAYS_ASSERT` is a very good idea :+1: 

We have many ways to abort / except asserts: we can exit, we can MPI exit, we can throw and recover. We can even call a hook and do something custom that the compiler cannot see yet.

lucafedeli88:
> `AMREX_ASSUME` to drop in `[[assume(same-expression == true)]]` in _after_ an `ALWAYS_ASSERT` is a very good idea 👍
> 
> We have many ways to abort / except asserts: we can exit, we can MPI exit, we can throw and recover. We can even call a hook and do something custom that the compiler cannot see yet.

I've opened a PR to implement `AMREX_ASSUME` in AMReX: https://github.com/AMReX-Codes/amrex/pull/3770

Source/Diagnostics/ComputeDiagFunctors/CellCenterFunctor.cpp
lucafedeli88: It should be fixed now!

ax3l:
@lucafedeli88 when you rebase against the latest development, your new AMReX assume should be available now. There is also a small merge conflict right now.

lucafedeli88:
> @lucafedeli88 when you rebase against the latest development, your new AMReX assume should be available now. There is also a small merge conflict right now.

@ax3l , I should have fixed the merge conflict

lucafedeli88:
ping, @ax3l  :-) !

Source/WarpX.H
```diff
@@ -69,19 +69,45 @@
...
-#include <map>
-
 
 enum struct PatchType : int
 {
     fine,
     coarse
 };
 
+enum struct FieldType : int
```
ax3l, line 84: This is ok for now, but we should make it a runtime construct so we can add more multifabs later on at runtime and also easily name them.

ax3l (approved):

PR #811, opened 2020-03-17
Add some WarpXUtilMsg::AlwaysAssert in particle containers

lucafedeli88:
This mini-PR replaces some of the `AMREX_ALWAYS_ASSERT_WITH_MESSAGE` macros with the new ` WarpXUtilMsg::AlwaysAssert` function, where this might result in more meaningful error messages for the user.

PR #4696, opened 2024-02-13
Removed const modifier from LPInfo object on stack that was causing a…

clarkse:
… warning from not being initialized properly.

### DIFF
--- a/Source/ablastr/fields/PoissonSolver.H
+++ b/Source/ablastr/fields/PoissonSolver.H
@@ -170,7 +170,7 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
         );
     }
 
-    const amrex::LPInfo& info = amrex::LPInfo();
+    amrex::LPInfo info;
     for (int lev=0; lev<=finest_level; lev++) {
         // Set the value of beta
         amrex::Array<amrex::Real,AMREX_SPACEDIM> beta_solver =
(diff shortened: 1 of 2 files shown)


lucafedeli88 (commented):
Hi @clarkse ,
thank you for your contribution.
Could you please provide some more details on the configuration in which you saw a warning about this variable? In principle, the `amrex::MLEBNodeFDLaplacian` constructor takes `info` as a constant reference, so it should be OK to have `const amrex::LPInfo info;` here.

clarkse:
This is the error I am getting building with GCC 9.4.0:

/home/st247c@HE.local/src/WarpX/Source/ablastr/fields/VectorPoissonSolver.H(133): warning #811-D: const variable "info" requires an initializer -- class "amrex::LPInfo" has no user-provided default constructor
      const amrex::LPInfo info;

I checked in PoissonSolver.H and it appears that the const is not in front of the amrex::LPInfo instantiation in that case.  I traced the addition of the const modifier to:
https://github.com/ECP-WarpX/WarpX/pull/3921

In this PR it was added to VectorPoissonSolver, but not PoissonSolver. I was just getting annoyed at seeing the compile time warning, so if we want to use the const then the default constructor for amrex::LPInfo should have a default constructor.

clarkse:
@lucafedeli88 does this work for you?  It appears consistent with usage in AMReX.

lucafedeli88 (approved):
Looks OK to me. Thanks!

PR #4704, opened 2024-02-19
[WIP] re-organizing metadata for incomplete BTD flush to close cleanly without nans

RevathiJambunathan:
The example in `Examples/Physics_applications/laser_acceleration/inputs_2d_boost` 
will produce nans in openpmd output if BTD flush is incomplete (i.e., run with fewer timesteps)

In this PR, this is fixed by changing the meta data for the snapshot box and geometry, so its consistent with all the previous buffers that were flushed and the number of cells in the current flush. 
Now the same example, when run with fewer timesteps (i.e., scenario where we have an incomplete BTD flush), has consistent meta data and therefore will not show nans for the missing cell data in the OpenPMD output.

This PR fixes the meta-data for the fields 

To do :
1. Update OpenPMD metadata so that the geometry and new number of boxes are corrected wrt to that of the terminated output. @ax3l Is there a way to  reset the openpmd metadata (that was set at first buffer flush) if there is an abrupt force flush and the btd flush is incomplete? 
It wont change the way we do offsets, except for the last incomplete flush, wher the offset will be zero, because the index of the geometry is updated to be consistent with the incomplete buffer.
Screenshot 2024-02-19 at 4 47 31 PM

2. meta-data for particles

### DIFF
--- a/Source/Diagnostics/BTDiagnostics.cpp
+++ b/Source/Diagnostics/BTDiagnostics.cpp
@@ -1018,12 +1018,12 @@ BTDiagnostics::Flush (int i_buffer, bool force_flush)
     amrex::Vector< amrex::MultiFab > out;
     out.resize(nlev_output);
     for(int lev = 0; lev < nlev_output; ++lev) {
-        if (!force_flush) {
-        out[lev] = amrex::MultiFab(m_mf_output[i_buffer][lev].boxArray(),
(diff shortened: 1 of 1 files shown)


RevathiJambunathan:
Discussed at development meeting : 
@ax3l is adding support for openPMD meta-data to be updated.

PR #4695, opened 2024-02-13
Obtain exact time (real) when particles hit boundaries + test

EyaDammak:
This PR is adding a new **real attribute** for the time called **"time_scraped"** in the buffer of the particles absorbed in the boundaries. This new attribute calculates the real time when the particle hits the embedded boundary at the exact point of contact. 
The PIC iteration at which each particle was absorbed at the boundary remains saved as an integer but it is called now **"step_scraped"** rather than "timestamp". 

_**Note 1:** Note that the attributes are accessed in the Python function ts.get_particle() respectively by **'timeScraped' and 'stepScraped'**. For the Python function buffer.get_particle_boundary_buffer(), the names remain 'time_scraped' and 'step_scraped'_

_**Note 2:** We changed how to read in the dictionary in the file Source/Python/Particles/PinnedMemoryParticleContainer.cpp_

_**Note 3:** It still needs to calculate the exact point of contact + exact time for the regular boundaries. The real time added for the regular boundaries is temporary_

PR #4702, opened 2024-02-17
Save the normal components in the particle buffer when there is EB

EyaDammak:
This PR completes the PR #4695. It saves in the particle buffer the normal components when there are embedded boundaries. The calculation of the normal components were also modified: we take now in account the weights of the cell-centered nodes.

Note: new option for the function compute_weights() --> nodal=0 or 1 (1 by default)
nodal=0: we calculate the cell-centered indices
nodal=1: we calculate the nodes indices

PR #4710, opened 2024-02-20
Update of the time saved for scraped particles on the boundaries

EyaDammak:
Update of the PR #4695
From the buffer of scraped particles, we can have access now to:
-the integer "stepScraped" which corresponds to the PIC iterations just before each particle hits boundaries (EB + regular boundaries)
-The real "deltaTimeScraped" which corresponds to the fraction of time between the time associated with the last step and the exact time when each particle hits an EB (for the regular boundaries, it is equal to 0 --> we still need to calculate it). This real is between 0 and dt.

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -1539,7 +1539,7 @@ are applied to the grid directly. In particular, these fields can be seen in the
     ``warpx.Bz_external_grid_function(x,y,z)`` to initialize the external
     magnetic field for each of the three components on the grid.
     Constants required in the expression can be set using ``my_constants``.
-    For example, if ``warpx.Bx_external_grid_function(x,y,z)=Bo*x + *(y + z)``
+    For example, if ``warpx.Bx_external_grid_function(x,y,z)=Bo*x + delta*(y + z)``
     then the constants `Bo` and `` required in the above equation
     can be set using ``my_constants.Bo=`` and ``my_constants.delta=`` in the
     input file. For a two-dimensional simulation, it is assumed that the first dimension
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -416,9 +416,9 @@ Domain Boundary Conditions
 
 * ``boundary.verboncoeur_axis_correction`` (`bool`) optional (default `true`)
     Whether to apply the Verboncoeur correction on the charge and current density on axis when using RZ.
-    For nodal values (rho and Jz), the cell volume for values on axis is calculated as :math:`\pi*\Delta r^2/4`.
+    For nodal values (rho and Jz), the cell volume for values on axis is calculated as :math:`\pi*\ r^2/4`.
     In :cite:t:`param-VerboncoeurJCP2001`, it is shown that using
-    :math:`\pi*\Delta r^2/3` instead will give a uniform density if the particle density is uniform.
+    :math:`\pi*\ r^2/3` instead will give a uniform density if the particle density is uniform.
 
 Additional PML parameters
 -------------------------
@@ -430,7 +430,7 @@ Additional PML parameters
     Whether or not to use an amrex::DistributionMapping for the PML grids that is `similar` to the mother grids, meaning that the
     mapping will be computed to minimize the communication costs between the PML and the mother grids.
 
-* ``warpx.pml_delta`` (`int`; default: 10)
+* ``warpx.pml_`` (`int`; default: 10)
     The characteristic depth, in number of cells, over which
     the absorption coefficients of the PML increases.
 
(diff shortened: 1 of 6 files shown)


Source/Particles/ParticleBoundaryBuffer.cpp
```diff
@@ -101,7 +101,7 @@ struct FindEmbeddedBoundaryIntersection {
 
         // Also record the real time on the destination
         dst.m_runtime_idata[m_step_index][dst_i] = m_step;
-        dst.m_runtime_rdata[m_time_index][dst_i] = m_step*m_dt + (1- dt_fraction)*m_dt;
+        dst.m_runtime_rdata[m_delta_index][dst_i] = (1- dt_fraction)*m_dt;
```
RevathiJambunathan:
```suggestion
        dst.m_runtime_rdata[m_delta_index][dst_i] = (1._rt- dt_fraction)*m_dt;
```

Source/Particles/ParticleBoundaryBuffer.cpp
```diff
@@ -151,7 +151,7 @@ struct CopyAndTimestamp {
             dst.m_runtime_idata[j][dst_i] = src.m_runtime_idata[j][src_i];
         }
         dst.m_runtime_idata[m_step_index][dst_i] = m_step;
-        dst.m_runtime_rdata[m_time_index][dst_i] = m_step*m_dt;
+        dst.m_runtime_idata[m_delta_index][dst_i] = 0*m_dt; //delta_fraction is putted as zero, we need to calculate it
```
RevathiJambunathan:
```suggestion
        dst.m_runtime_idata[m_delta_index][dst_i] = 0._rt; //delta_fraction is initialized to zero
```
Just checking . if its set to 0., we dont need to multiple with `m_dt`? 
also this is where its getting initialized and will be computed later in `FinedEmbeddedBoundaryIntersection`

EyaDammak: Yes exactly, for now we don't have yet the deltaTimeScraped for regular boundaries. We just assume that it is equal to 0 temporarily. I just did 0*dt to remember that we have to calculate the dt_fraction for the regular boundaries. 
Yes, we can just put it equal to 0 for now!
Thank you for the review Revathi!!

RevathiJambunathan:
Note : Merge this PR after #4702 due to conflits in the doc for set_Step, time_scraped in #4702

Docs/source/usage/parameters.rst
```diff
@@ -2752,10 +2752,10 @@ The data collected at each boundary is written out to a subdirectory of the diag
 By default, all of the collected particle data is written out at the end of the simulation. Optionally, the ``<diag_name>.intervals`` parameter can be given to specify writing out the data more often.
 This can be important if a large number of particles are lost, avoiding filling up memory with the accumulated lost particle data.
 
-In addition to their usual attributes, the saved particles have an integer attribute ``step_scraped``, which
+In addition to their usual attributes, the saved particles have an integer attribute ``stepScraped``, which
 indicates the PIC iteration at which each particle was absorbed at the boundary,
-and a real attribute ``time_scraped``, which indicates the exact time calculated when each
-particle hits the EB.
+and a real attribute ``deltaTimeScraped``, which indicates the fraction of time between the time associated to the last step
+and the exact time when each particle hits the boundary. ``deltaTimeScraped`` is a real between 0 and dt.
```
RemiLehe: Thanks for adding this in the documentation.
Could you also add these attributes in the table here:
https://github.com/ECP-WarpX/WarpX/blob/development/Docs/source/developers/particles.rst#particle-attributes
In the `Notes`, you can add that this is only saved for particles collected at the boundaries (in the boundary buffers, and the corresponding `BoundaryScraping` Diagnostic)

```suggestion
and the exact time when each particle hits the boundary. ``deltaTimeScraped`` is a real between ``0`` and ``dt``.
```

Source/Particles/ParticleBoundaryBuffer.cpp
```diff
@@ -151,7 +152,7 @@ struct CopyAndTimestamp {
             dst.m_runtime_idata[j][dst_i] = src.m_runtime_idata[j][src_i];
         }
         dst.m_runtime_idata[m_step_index][dst_i] = m_step;
-        dst.m_runtime_rdata[m_time_index][dst_i] = m_step*m_dt;
+        dst.m_runtime_rdata[m_delta_index][dst_i] = 0._rt; //delta_fraction is initialized to zero
```
RemiLehe, line 186: Could you open an issue (https://github.com/ECP-WarpX/WarpX/issues) so that we remember to fix this?
(You could also add that we need to save the normal in this case.)

RemiLehe (changes requested):
Thanks! I added a few comments.

Source/Particles/ParticleBoundaryBuffer.cpp
EyaDammak: https://github.com/ECP-WarpX/WarpX/issues/4718

Docs/source/usage/parameters.rst
EyaDammak: I opened a new PR for that:  #4719!

RemiLehe (approved):

PR #4720, opened 2024-02-22
Enable different types of diagnostics for ParticleDiagnostic() in Python

EyaDammak:
This PR enables to have the file of ``ParticleDiagnostic`` with different type ("Full", "BoundaryScraping" or "BackTransformed").
This is possible in adding the option `warpx_diag_type` in ``picmi.ParticleDiagnostic()``
The option warpx_diag_type='Full' is set by default.

### DIFF
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -2404,7 +2404,7 @@ class ParticleDiagnostic(picmistandard.PICMI_ParticleDiagnostic, WarpXDiagnostic
     def init(self, kw):
 
         self.format = kw.pop('warpx_format', 'plotfile')
-        self.type = kw.pop('warpx_diag_type', None)
+        self.type = kw.pop('warpx_diag_type', 'Full')
         self.openpmd_backend = kw.pop('warpx_openpmd_backend', None)
         self.file_prefix = kw.pop('warpx_file_prefix', None)
         self.file_min_digits = kw.pop('warpx_file_min_digits', None)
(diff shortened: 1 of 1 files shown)


Python/pywarpx/picmi.py
```diff
@@ -2404,6 +2404,7 @@ class ParticleDiagnostic(picmistandard.PICMI_ParticleDiagnostic, WarpXDiagnostic
     def init(self, kw):
 
         self.format = kw.pop('warpx_format', 'plotfile')
+        self.type = kw.pop('warpx_diag_type', 'Full')
```
ax3l, line 2410: Awesome, great work! 

Can you please just a few lines above, in the documentation string, add an entry for the new `warpx_diag_type` option? That way it will be shown in our manual automatically and users will find it.
https://warpx.readthedocs.io/en/latest/usage/python.html#diagnostics

Python/pywarpx/picmi.py
```diff
@@ -2383,6 +2383,9 @@ class ParticleDiagnostic(picmistandard.PICMI_ParticleDiagnostic, WarpXDiagnostic
     warpx_format: {plotfile, checkpoint, openpmd, ascent, sensei}, optional
         Diagnostic file format
 
+    warpx_diag_type: {Full, BoundaryScraping, BackTransformed}, 'Full' by default, optional
```
RemiLehe, line 2386: Hm, allowing `BackTransformed` makes it confusing, since this seems to duplicate the capability of `LabFrameParticleDiagnostic`... We can discuss this in more detail.

PR #4721, opened 2024-02-22
Position keyword now adds position variables in PICMI ParticleDiagnostic

n01r:
As our [documentation](https://warpx.readthedocs.io/en/latest/usage/python.html#pywarpx.picmi.ParticleDiagnostic) states, adding the string `'position'` to the `data_list` parameter of PICMI's `ParticleDiagnostic` should add position variables. This was true by default until recently but now positions can actually be left out and are not written automatically, anymore.

This PR also symmetrizes the list of accepted variables between the `LabFrameParticleDiagnostic` and the `ParticleDiagnostic` classes while adding the former to the documentation.


Python/pywarpx/picmi.py
```diff
@@ -2457,7 +2458,7 @@ def diagnostic_initialize_inputs(self):
                     variables.add('Bx')
                     variables.add('By')
                     variables.add('Bz')
-                elif dataname in ['ux', 'uy', 'uz', 'Ex', 'Ey', 'Ez', 'Bx', 'By', 'Bz']:
+                elif dataname in ['x', 'y', 'z', 'ux', 'uy', 'uz', 'Ex', 'Ey', 'Ez', 'Bx', 'By', 'Bz', 'Er', 'Et', 'Br', 'Bt']:
```
n01r: What about the current variables `j`, though?
I don't quite see why they are missing.

n01r:
Opening and closing because CI test for HIP failed with a timeout while installing dependencies

ax3l (approved):

Python/pywarpx/picmi.py
ax3l: We do not anymore have an E or B field on the particle itself.
(For j we never had it.)

PR #4722, opened 2024-02-22
PICMI: Fix unresolved reference for species

n01r:
The variable `species` was an unresolved reference and needed to accessed via `self.species` instead.

Fixed also a word `specie` where `species` was meant.


ax3l (approved):
Thank you for this!! :sparkles:

Python/pywarpx/picmi.py
```diff
@@ -2696,9 +2696,9 @@ def diagnostic_initialize_inputs(self):
         if self.species is None:
             species_names = pywarpx.particles.species_names
         elif np.iterable(self.species):
-            species_names = [specie.name for specie in self.species]
+            species_names = [species.name for species in self.species]
         else:
-            species_names = [species.name]
+            species_names = [self.species.name]
```
ax3l, line 2701: This is the fix, right? The other parts above are just renaming, right?

PR #4723, opened 2024-02-23
Compute electrostatic fields at the beginning of EM simulations, if potential is specified

RemiLehe:
This modifies the behavior of WarpX, such that, if the user uses a Maxwell solver and specifies the potential at the boundaries, an initial Poisson solve will be performed at ``t=0``.


roelof-groenewald:
@RemiLehe this shouldn't be a problem for the Ohm's law solver since it completely overwrites the E-field, but it would be good to raise a warning if a user tries to specify boundary potentials since they won't actually be taken into account.

Examples/Tests/point_of_contact_EB/inputs_3d
```diff
@@ -12,12 +12,6 @@ geometry.prob_hi     = 0.26 0.26 0.26
 
 boundary.field_lo = pec pec pec
 boundary.field_hi = pec pec pec
```
RemiLehe, line 14: I removed these lines in order to avoid triggering the electrostatic solver, for this particular test.

Source/WarpX.cpp
```diff
@@ -750,7 +750,13 @@ WarpX::ReadParameters ()
...
         }
 
-        if (electrostatic_solver_id == ElectrostaticSolverAlgo::LabFrame ||
+#if defined(AMREX_USE_EB) && defined(WARPX_DIM_RZ)
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+        electromagnetic_solver_id==ElectromagneticSolverAlgo::None,
+        "Currently, the embedded boundary in RZ only works for electrostatic solvers (or no solver).");
+#endif
+
+      if (electrostatic_solver_id == ElectrostaticSolverAlgo::LabFrame ||
```
```suggestion
      if (electrostatic_solver_id == ElectrostaticSolverAlgo::LabFrame ||
```

Source/WarpX.cpp
```diff
@@ -750,7 +750,7 @@ WarpX::ReadParameters ()
             electromagnetic_solver_id = ElectromagneticSolverAlgo::None;
         }
 
-        if (electrostatic_solver_id == ElectrostaticSolverAlgo::LabFrame ||
+      if (electrostatic_solver_id == ElectrostaticSolverAlgo::LabFrame ||
```
```suggestion
        if (electrostatic_solver_id == ElectrostaticSolverAlgo::LabFrame ||
```

Source/WarpX.cpp
```diff
@@ -765,13 +765,27 @@ WarpX::ReadParameters ()
...
+        m_boundary_potential_specified = potential_specified;
+        if (potential_specified & (WarpX::electromagnetic_solver_id == ElectromagneticSolverAlgo::HybridPIC)) {
+            ablastr::warn_manager::WMRecordWarning(
+                "Algorithms",
+                "The input script specifies the electric potential (phi) at the boundary, but \
+                also uses the hybrid PIC solver based on Ohm’s law. When using this solver, the \
+                electric potential does not have any impact on the simulation.",
+                ablastr::warn_manager::WarnPriority::low);
+        }
+
```
roelof-groenewald, line 795: I think it would be good to add a print statement here to say that an initial Poisson solve will be performed since, as you already pointed out, if someone has the boundary potentials specified in their input scripts the code will now behave differently (and could even crash if they use `pml`/`damped`/`none` boundaries or if for some reason the Poisson solve does not converge for their specific setup). It could just be a message though, doesn't have to be a warning.

```suggestion
        else if (potential_specified & (WarpX::electromagnetic_solver_id != ElectromagneticSolverAlgo::None)) {
            ablastr::warn_manager::WMRecordWarning(
                "Algorithms",
                "The input script specifies the electric potential (phi) at the boundary so \
                an initial Poisson solve will be performed.",
                ablastr::warn_manager::WarnPriority::low);
        }
        
```

roelof-groenewald (commented):
I think this can go in as is (possibly with the suggestion for a warning that the code might behave differently now for old input scripts).

But one thing that should be fixed is that currently if a user specifies the boundary potential (on purpose or due to an outdated script) and is using `damped`, `pml` or `none` for one of the boundaries the simulation will crash from
https://github.com/ECP-WarpX/WarpX/blob/7317aafb9d99b74d077c5bcc1bae3af29f31d9fd/Source/FieldSolver/ElectrostaticSolver.cpp#L1022

A solution would be to set the Poisson solve to use Neumann BCs if any of the above boundary conditions are used.

roelof-groenewald (approved):

PR #4726, opened 2024-02-23
[WIP] Split of momentum push for explicit solve with insertion of collision…

jlvay:
…s in the middle of the momentum push.


ax3l:
X-ref: @EZoni works on a production implementation of this feature in #6275 :tada:

PR #4733, opened 2024-02-28
Update test that checks absorption of particles in PML

RemiLehe:
This removes duplication between different files.


RevathiJambunathan (approved):
Thanks @RemiLehe  for this PR. Looks good to me

PR #4734, opened 2024-02-29
Fix: Windows add_dll_directory Expand

ax3l:
Expand paths in `PATH` environment variable before adding them and make them absolute paths.

See https://github.com/ECP-WarpX/impactx/issues/534


lucafedeli88 (approved):
Thanks, @ax3l

PR #4729, opened 2024-02-27
New Python test: Particle-Boundary interaction 

EyaDammak:
This PR adds a new Python test using a simple case of reflection of one electron on a sphere. 
When the electron hits the sphere, it adds a new electron from the exact position of impact with a velocity antisymetrical to the normal of the EB.

PR #4738, opened 2024-02-29
[WIP] Add test for particle boundary interaction

RemiLehe:
The code is the same as in this PR: https://github.com/ECP-WarpX/WarpX/pull/4729
 
The aim is to see whether the strange CI bug would still show up.


RemiLehe:
It also hangs in CI at the same point as https://github.com/ECP-WarpX/WarpX/pull/4729 ...

PR #4739, opened 2024-03-01
Doc: Python Bld in Separate Dir

ax3l:
Python builds require shared AMReX and WarpX libraries. To avoid confusion with users that also build the executables, but do not use the full installer logic, we recommend to use two separate build directories, so the executables can use a static build of their dependendencies to stay relocatable (as in: can be copied without setting environment hints to dependent shared libs).


lucafedeli88 (approved):
Thanks, @ax3l !

PR #4740, opened 2024-03-01
Add function to set domain boundary potentials from Python

### LINKED ISSUE #4715: Request: update domain boundary potentials during simulation execution
To reproduce the error, build WarpX in the following way

```
warpx=WarpX-24.02
echo "compile $warpx"
cmake -S $HOME/$warpx -B $HOME/$warpx/build -DWarpX_DIMS=RZ \
  -DWarpX_COMPUTE=CUDA \
  -DWARX_MPI=ON \
  -DWarpX_QED=OFF \
  -DWarpX_OPENPMD=ON \
  -DWarpX_PYTHON=ON

echo "build warpx and do pip install"
cmake --build $HOME/$warpx/build --target pip_install -j 16
```

Then use the example input [spacecraft_charging/PICMI_inputs_rz.py](https://github.com/ECP-WarpX/WarpX/blob/development/Examples/Physics_applications/spacecraft_charging/PICMI_inputs_rz.py).

Here is the output file and Backtrace file
[output.log](https://github.com/ECP-WarpX/WarpX/files/14375623/output.log)
[Backtrace.0.txt](https://github.com/ECP-WarpX/WarpX/files/14375628/Backtrace.0.txt)

I also tried my own input script, it runs fine as long as we don't access fields 
```
phi_wrapper = fields.PhiFPWrapper()
phi_wrapper[Ellipsis] # Segfault occurs when accessing fields data
```

Edit:

According the discussion, it seems there is no simple way to access fields data in GPU memory. I wonder if WarpX can implement this feature to read/write fields data in GPU?

roelof-groenewald:
Fixes #4715.

Functionality is added to allow a python user to update the potentials on the domain boundaries, similar to how one can already update the potential on an embedded boundary.

### DIFF
--- a/Docs/source/usage/workflows/python_extend.rst
+++ b/Docs/source/usage/workflows/python_extend.rst
@@ -90,6 +90,12 @@ This object is the Python equivalent to the C++ ``WarpX`` simulation class.
 
    .. py:method:: get_particle_boundary_buffer
 
+   .. py:method:: set_potential_on_domain_boundary(potential_[lo/hi]_[x/y/z]: str)
+
+      The potential on the domain boundaries can be modified when using the electrostatic solver.
+      This function updates the strings and function parsers which set the domain
+      boundary potentials during the Poisson solve.
+
    .. py:method:: set_potential_on_eb(potential: str)
 
       The embedded boundary (EB) conditions can be modified when using the electrostatic solver.
--- a/Source/Python/WarpX.cpp
+++ b/Source/Python/WarpX.cpp
@@ -170,6 +170,34 @@ The physical fields in WarpX have the following naming:
             "Get the current physical time step size on mesh-refinement level ``lev``."
         )
 
+        .def("set_potential_on_domain_boundary",
+            [](WarpX& wx,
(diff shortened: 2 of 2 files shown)


HuntFeng (commented):
Instead of using `potential_zlo`, `potential_zhi`, maybe be using `potential_zmin`, `potential_zmax` (or `potential_lo_z`, `potential_hi_z`) is better because it sticks with the naming convention of WarpX. 

Edit: This implementation works well. Should we document its usage somewhere?

roelof-groenewald:
> Instead of using `potential_zlo`, `potential_zhi`, maybe be using `potential_zmin`, `potential_zmax` (or `potential_lo_z`, `potential_hi_z`) is better because it sticks with the naming convention of WarpX.

Thanks for the suggestions! I'll update the keywords to `potential_[dir]_[lo/hi]` as suggested to match the conventions in the input scripts. I'll also add the function to the documentation as suggested.
Thanks for testing it out!

HuntFeng (approved):
Awesome, I like them. Thanks for your hard work!

RemiLehe (approved):

PR #4744, opened 2024-03-03
Use `BinaryCollision` class for DSMC collisions

### LINKED ISSUE #4727: Use `BinaryCollision` instance for DSMC with custom collision functor
Currently Coulomb collisions and nuclear fusion are handled as instances of the `BinaryCollision` class with special `CollisionFunctorType`s and `CopyTransformFunctorType`s used to initialize the `BinaryCollision` class in each collision instance (different fusion reactions for example). In contrast, DSMC is handled through an instance of the special class `DSMC`.

It would be good to have the overlapping functionality used for all binary collisions.

The "issue" we currently face in having DSMC use the same approach as the other binary collisions is the following. The way `BinaryCollision` is currently set up it assumes two distinct types of collisions:
1. Collisions that do not produce any product species but simply alter the momentum of the colliding particles (ex. Coulomb collisions).
2. Collisions that consume the colliding particles and create products (ex. p + B -> 3 \alpha).

The DSMC collisions generally land in a third category where the mismatch between weights of the colliding particles require the splitting of the higher weight particle, and therefore, DSMC collisions require the copying of at least one of the incoming particles even though that particle is not a product of the collision.

We can deal with this in a number of ways.
1. Merge the "new" particle with the "old" one immediately after scattering thereby avoiding the need to copy one of the colliding particles.
2. Expand the `BinaryCollision` class to handle splitting of uneven weight colliding pairs.
3. ...

roelof-groenewald:
This PR unifies (somewhat) the manner in which binary collisions are performed by switching the DSMC scheme to also use the `BinaryCollision` class with special functors for the actual collision process.

Fixes #4727.

Still to do:
- [x] check that new implementation still gives correct physics results

Also note that there is still significant overlap in code between the DSMC module and nuclear fusion. Specifically the following files/routines are almost the same (but different in ways that requires significant work to avoid the code duplication):

- operators for `NuclearFusionFunc` and `DSMCFunc`
- `ParticleCreationFunc` and `SplitAndScatterFunc`


Source/Particles/Collision/ScatteringProcess.H
```diff
@@ -42,10 +42,10 @@ public:
...
     ~ScatteringProcess() = default;
 
-    ScatteringProcess (ScatteringProcess const&)            = delete;
-    ScatteringProcess& operator= (ScatteringProcess const&) = delete;
-    ScatteringProcess (ScatteringProcess &&)                = default;
-    ScatteringProcess& operator= (ScatteringProcess &&)     = default;
+    // ScatteringProcess (ScatteringProcess const&)            = delete;
+    // ScatteringProcess& operator= (ScatteringProcess const&) = delete;
+    // ScatteringProcess (ScatteringProcess &&)                = default;
+    // ScatteringProcess& operator= (ScatteringProcess &&)     = default;
```
roelof-groenewald: It's not great to allow `ScatteringProcess` to be copied since it can hold substantial data from the cross-section files. But this was the quickest solution to avoid the compiler complaining about the copy of `DSMCFunc` at line https://github.com/ECP-WarpX/WarpX/blob/b0ae812c1aa09b954b9c2fa5ba8b0f22a652696f/Source/Particles/Collision/BinaryCollision/BinaryCollision.H#L227

@ax3l how much of a problem do you think this is?

roelof-groenewald: Looks like it is actually a big problem since all the MCC tests fail now.

ax3l: Can we try this first?
```suggestion
    ScatteringProcess (ScatteringProcess &&)                = default;
    ScatteringProcess& operator= (ScatteringProcess &&)     = default;
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -224,7 +247,7 @@ public:
         using namespace ParticleUtils;
         using namespace amrex::literals;
 
-        CollisionFunctorType binary_collision_functor = m_binary_collision_functor;
+        CollisionFunctor binary_collision_functor = m_binary_collision_functor;
```
ax3l: first thought: We could use an alias/reference here (or a pointer)
```suggestion
        CollisionFunctor & binary_collision_functor = m_binary_collision_functor;
```

roelof-groenewald: Somehow using a reference wasn't enough and I still got compiler complaints about trying to copy `ScatteringProcess` instances. I was able to work around this issue by only keeping the `ScatteringProcess::Executor` structs instead of the entire class but it requires that I move the cross-section data from each `ScatteringProcess` class to the `DSMCFunc` instance.

roelof-groenewald:
I could use some help getting past the (hopefully) final hurdle for this PR. I'm getting the message:
```
./Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.H(252): error: calling a __host__ function("amrex::PODVector< ::ScatteringProcess::Executor,  ::amrex::ArenaAllocator< ::ScatteringProcess::Executor> > ::~PODVector()") from a __host__ __device__ function("DSMCFunc::~DSMCFunc") is not allowed
```

Update: Okay, I think I'm starting to understand what the problem is here.`DSMCFunc::operator()` is specified to be a `AMREX_GPU_HOST_DEVICE` function, so this forces the entire class to be compiled for both host and device. But that is not possible for `amrex::Gpu::DeviceVector` which causes the compilation failure. Does that sound right?
A solution would be to define a new `struct` inside `DSMCFunc` that contains the function executed within the `ParallelFor` from `BinaryCollision`. This struct would then not contain any `amrex::Gpu::DeviceVector` member variables and will therefore be fine for device code. In order to then have `BinaryCollision` work for Coulomb collisions, nuclear fusion and DSMC, I will need to update the former two's collision filter classes to have the same struct-approach.
@ax3l @RemiLehe any thoughts on that?

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
roelof-groenewald: This issue is solved by using the internal `struct` approach 🎉

roelof-groenewald:
I reran the first 2 cases of the Turner benchmark with the updated code, and the results look good.
image

Considering also that the CI test benchmark values did not need to be changed, we can confidently say that the changes did not affect the DSMC module accuracy.

After this PR is merged it should be straightforward to extend #4577 for the DSMC case.

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -9,173 +9,287 @@
...
+        {
+            // How many particles of product species i are created.
+            const index_type num_added = total * m_num_products_host[i];
+            num_added_vec[i] = static_cast<int>(num_added);
+            tile_products[i]->resize(products_np[i] + num_added);
+        }
+
+        // this works for DSMC since the colliding particles are also products
+        const auto soa_1 = tile_products[0]->getParticleTileData();
+        const auto soa_2 = tile_products[1]->getParticleTileData();
```
roelof-groenewald, line 98: I wanted to highlight this bit of code. For some reason that I don't understand it doesn't work to set the id of colliding particles to invalid using the `soa_1` object passed into this function (particles are not actually removed from the simulation in that case). However, using the tiles to get the SoA values as is done here does produce the expected result. Based on the similarity between this code and the fusion module, I expect the same issue to exist there. I will submit a follow-up PR once this is merged to change the fusion approach as well.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -199,6 +217,10 @@ public:
                     amrex::HostDevice::Atomic::Add( &(*cost)[mfi.index()], wt);
                 }
             }
+
+            // Call redistribute to remove particles with negative ids
+            species1.Redistribute(lev, lev, 0, true, true);
+            species2.Redistribute(lev, lev, 0, true, true);
```
roelof-groenewald: Address https://github.com/ECP-WarpX/WarpX/pull/4577#discussion_r1531167214
```suggestion
            if (m_have_product_species) {
                // Call redistribute to remove particles with negative ids
                species1.Redistribute(lev, lev, 0, true, true);
                species2.Redistribute(lev, lev, 0, true, true);
            }
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -199,6 +217,12 @@ public:
                     amrex::HostDevice::Atomic::Add( &(*cost)[mfi.index()], wt);
                 }
             }
+
+            if (m_have_product_species) {
+                // Call redistribute to remove particles with negative ids
+                species1.Redistribute(lev, lev, 0, true, true);
+                species2.Redistribute(lev, lev, 0, true, true);
```
roelof-groenewald: The `local` parameter sets how many cells particles could have moved since the previous `Redistribute` call. Therefore, a "local" redistribute has `local=0`.
```suggestion
                species1.Redistribute(lev, lev, 0, 0, true);
                species2.Redistribute(lev, lev, 0, 0, true);
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -199,6 +217,12 @@ public:
                     amrex::HostDevice::Atomic::Add( &(*cost)[mfi.index()], wt);
                 }
             }
+
+            if (m_have_product_species) {
+                // Call redistribute to remove particles with negative ids
+                species1.Redistribute(lev, lev, 0, 0, true);
+                species2.Redistribute(lev, lev, 0, 0, true);
```
roelof-groenewald: The `local` flag should actually be `1`. See discussion at https://github.com/ECP-WarpX/WarpX/pull/4831#discussion_r1554498964
```suggestion
                species1.Redistribute(lev, lev, 0, 1, true);
                species2.Redistribute(lev, lev, 0, 1, true);
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -92,26 +93,43 @@ public:
...
         const amrex::ParmParse pp_collision_name(collision_name);
         pp_collision_name.queryarr("product_species", m_product_species);
+
+        // if DSMC the colliding species are also product species
+        if (collision_type == CollisionType::DSMC) {
+            m_product_species.push_back(m_species_names[0]);
+            if (!m_isSameSpecies) {
+                m_product_species.push_back(m_species_names[1]);
+            }
+            // rotate to have the colliding species first
```
RemiLehe: I don't think I understand why this is needed.
Also, wouldn't it be simpler to do:
```
            
            if (m_isSameSpecies) {
                m_product_species.push_back(m_species_names[0]);
            } else {
                // have the colliding species first
                m_product_species.push_back(m_species_names[1]);
                m_product_species.push_back(m_species_names[0]);
            }
```

RemiLehe: without a call to `rotate`

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -199,6 +217,12 @@ public:
                     amrex::HostDevice::Atomic::Add( &(*cost)[mfi.index()], wt);
                 }
             }
+
+            if (m_have_product_species) {
```
RemiLehe, line 212:
```suggestion
            if (m_have_product_species) {
                // The fact that there are product species indicates that particles of
                // the colliding species (`species1` and `species2`) may be removed 
                // (i.e., marked as invalid) in the process of creating new product particles.
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -58,14 +59,14 @@
 /**
  * \brief This class performs generic binary collisions.
  *
- * \tparam CollisionFunctorType the type of the specific binary collision functor that acts on a
+ * \tparam CollisionFunctor the the specific binary collision functor that acts on a
```
```suggestion
 * \tparam CollisionFunctor the specific binary collision functor that acts on a
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -58,14 +59,14 @@
...
  *         single cell
- * \tparam CopyTransformFunctorType the type of the second functor used in the case of
+ * \tparam CopyTransformFunctor the the second functor used in the case of
```
```suggestion
 * \tparam CopyTransformFunctor the second functor used in the case of
```

Source/Particles/Collision/BinaryCollision/BinaryCollisionUtils.cpp
```diff
@@ -17,123 +17,129 @@
 
 namespace BinaryCollisionUtils{
 
+    CollisionType get_collision_type (const std::string& collision_name,
+                                      MultiParticleContainer const * const mypc)
+    {
+        const amrex::ParmParse pp_collision_name(collision_name);
+        // For legacy, pairwisecoulomb is the default
```
RemiLehe, line 24: Maybe we can introduce a breaking change by removing the default, in a follow-up PR.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
roelof-groenewald: Basically, since the DSMC scheme splits particles during collision (to account for differences in weight) the colliding species are also considered product species. That is why the colliding species are added to the `m_product_species` vector. The `SplitAndScatterFunc` operator is written such that it assumes `m_species_names[0]` is also the 0'th element of `m_product_species` (and similarly for `m_species_names[1]` if more than 1 species is involved). The `rotate` is needed since there could be other product species in future DSMC processes (ionization for example).
Of course the whole thing could be changed so that `SplitAndScatterFunc` uses the last (or last two) element(s), but I don't think it would make a difference.

Source/Particles/Collision/BinaryCollision/DSMC/CollisionFilterFunc.H
```diff
@@ -34,43 +34,51 @@
  */
```
RemiLehe, line 36: Could you add a docstring for the new arguments that you added?

RemiLehe (changes requested):

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
RemiLehe: Ah, I get it now, thanks.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -92,26 +93,43 @@ public:
...
+                m_product_species.push_back(m_species_names[1]);
+            }
+            // rotate to have the colliding species first
+            std::rotate(
+                m_product_species.rbegin(),
+                m_product_species.rbegin() + 1 + (!m_isSameSpecies),
+                m_product_species.rend()
+            );
+        }
+
```
```suggestion
        // if DSMC the colliding species are also product species
        // Therefore, we insert the colliding species (`m_species_names`) at the beginning of `m_product_species`
        if (collision_type == CollisionType::DSMC) {
            m_product_species.insert( m_product_species.begin(), m_species_names.begin(), m_species_names.end() );
        }
```

RemiLehe: Does this work? It would make the code simpler, no?

roelof-groenewald: Thanks! That is much simpler!

RemiLehe (approved):

PR #4749, opened 2024-03-05
Enable `WarpX::ComputeEdgeLengths` in RZ

roelof-groenewald:
As far as I can tell, `WarpX::ComputeEdgeLengths` was excluded in RZ simulations out of an abundance of caution. But initial testing doesn't show any problems with handling RZ the same as 2D. I'm specifically interested in using this with the hybrid-PIC solver so I haven't done any testing with the regular electromagnetic solvers (which is why I only request to allow the hybrid-solver to use this for now).

As a simple sanity check of whether this works as intended, I ran a modified case of the normal modes in a cylinder test by covering the domain at r > R_max/2 with an embedded boundary. The plot below shows the normal modes dispersion relations for a cylinder with radius equal to R_max in white and with radius equal to R_max/2 in blue. The simulation with embedded boundary over r > R_max/2 is seen to follow the blue dispersion and therefore passes the simple check.

image

### DIFF
--- a/Source/WarpX.cpp
+++ b/Source/WarpX.cpp
@@ -754,7 +754,7 @@ WarpX::ReadParameters ()
         WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
         electromagnetic_solver_id==ElectromagneticSolverAlgo::None
         || electromagnetic_solver_id==ElectromagneticSolverAlgo::HybridPIC,
-        "Currently, the embedded boundary in RZ only works for electrostatic solvers (or no solver).");
+        "Currently, the embedded boundary in RZ only works for electrostatic solvers, the Ohm's law solver or with no solver installed.");
 #endif
 
         if (electrostatic_solver_id == ElectrostaticSolverAlgo::LabFrame ||
--- a/Source/WarpX.cpp
+++ b/Source/WarpX.cpp
@@ -752,7 +752,8 @@ WarpX::ReadParameters ()
 
 #if defined(AMREX_USE_EB) && defined(WARPX_DIM_RZ)
         WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
-        electromagnetic_solver_id==ElectromagneticSolverAlgo::None,
+        electromagnetic_solver_id==ElectromagneticSolverAlgo::None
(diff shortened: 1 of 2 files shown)


roelof-groenewald:
@WeiqunZhang or @atmyers anything jump out on why it might be invalid to handle the EB in RZ the same as in 2D? Do you happen to know why these functions were excluded from RZ builds in the first place?

RemiLehe:
Thanks @roelof-groenewald! 
This makes sense to me. I think that this is related to this PR:
https://github.com/ECP-WarpX/WarpX/pull/2707/
for which I have not finalized yet the automated tests.

roelof-groenewald:
Thanks for looking over it @RemiLehe. I wasn't aware of your draft PR but that is good to know about now and to see that you had a similar approach to enabling `ComputeEdgeLengths`. I don't have `ComputeFaceAreas` activated for RZ though as you do in your PR (and all the changes to the electromagnetic solver). I guess since your PR is a bit outdated it would make sense to merge this one and update yours accordingly, or what do you think?

RemiLehe:
Yes, absolutely, I agree that this PR (4749) should be merged first.
The `ComputeFaceAreas` may indeed be obsolete and not been needed anymore.

Source/WarpX.cpp
```diff
@@ -752,7 +752,8 @@ WarpX::ReadParameters ()
 
 #if defined(AMREX_USE_EB) && defined(WARPX_DIM_RZ)
         WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
-        electromagnetic_solver_id==ElectromagneticSolverAlgo::None,
+        electromagnetic_solver_id==ElectromagneticSolverAlgo::None
+        || electromagnetic_solver_id==ElectromagneticSolverAlgo::HybridPIC,
         "Currently, the embedded boundary in RZ only works for electrostatic solvers (or no solver).");
```
RemiLehe: Could you update this message to include the hybrid solver in the text?

roelof-groenewald:
```suggestion
        "Currently, the embedded boundary in RZ only works for electrostatic solvers, the Ohm's law solver or with no solver installed.");
```

RemiLehe (approved):

PR #4750, opened 2024-03-05
Bugfix in `fields.py` for GPU run without `cupy`

roelof-groenewald:
Currently if `libwarpx.libwarpx_so.Config.have_gpu == True` but `cupy` is not installed (i.e. `cp == None`) an error occurs if a user tries to set field values:
```
AttributeError: 'NoneType' object has no attribute 'asarray'
```

This PR avoids that failure and allows for the possibility that managed memory is turned on. If managed memory is not turned on an error will still occur but before the error a message will be printed based on `cupy_status`, which for CUDA builds without cupy will read:
```
"Warning: GPU found but cupy not available! Trying managed memory in numpy..."
```

### DIFF
--- a/Python/pywarpx/fields.py
+++ b/Python/pywarpx/fields.py
@@ -47,6 +47,7 @@ try:
     import cupy as cp
 except ImportError:
     cp = None
+from .LoadThirdParty import load_cupy
 
 try:
     from mpi4py import MPI as mpi
@@ -525,8 +526,11 @@ class _MultiFABWrapper(object):
                 if isinstance(value, np.ndarray):
                     slice_value = value3d[global_slices]
                     if libwarpx.libwarpx_so.Config.have_gpu:
+                        xp, cupy_status = load_cupy()
                         # Copy data from host to device
-                        slice_value = cp.asarray(slice_value)
+                        slice_value = xp.asarray(slice_value)
+                        if cupy_status is not None:
+                            libwarpx.amr.Print(cupy_status)
                     mf_arr[block_slices] = slice_value
                 else:
                     mf_arr[block_slices] = value
--- a/Python/pywarpx/fields.py
+++ b/Python/pywarpx/fields.py
@@ -517,6 +517,12 @@ class _MultiFABWrapper(object):
             if not isinstance(ic   , slice) or len(global_shape) < 4: global_shape[3:3] = [1]
             value3d.shape = global_shape
 
+            if libwarpx.libwarpx_so.Config.have_gpu:
+                # check if cupy is available for use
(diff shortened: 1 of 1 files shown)


roelof-groenewald:
It would be nice if we had a way to query whether managed memory is turned on in AMReX, in which case we could even exit gracefully with a message that the attempted functionality requires managed memory (or cupy) instead of ending in a segfault (I assume that is what will happen without managed memory turned on).

ax3l:
There are two ways to query if managed memory is used in AMReX today: check the parameter parser
```py
pp_amrex = amr.ParmParse("amrex")
is_managed = pp_amrex.query_int("the_arena_is_managed", 0)
```

or you can ask the Arena of the data container you work with
https://github.com/AMReX-Codes/pyamrex/pull/269

Python/pywarpx/fields.py
```diff
@@ -525,8 +526,11 @@ def __setitem__(self, index, value):
                 if isinstance(value, np.ndarray):
                     slice_value = value3d[global_slices]
                     if libwarpx.libwarpx_so.Config.have_gpu:
+                        xp, cupy_status = load_cupy()
```
ax3l: Two suggestions:
- move the load_copy outside of the loop, inside a `if libwarpx.libwarpx_so.Config.have_gpu and isinstance(value, np.ndarray):` to avoid repeated prints
- check `cupy_status` and print immediately after, otherwise missing managed memory will crash before you see that the issue is a missing cupy

roelof-groenewald: Sounds good on avoiding the repeated prints. 
Just want to point out that currently it would print before crashing since it won't crash until line 534 😉

roelof-groenewald:
> There are two ways to query if managed memory is used in AMReX today: check the parameter parser
> ...
> or you can ask the Arena of the data container you work with.

Thanks! In a follow-up PR we can check the Arena and abort if on GPU and `isinstance(value, np.ndarray)` and `cupy` is not available.

Python/pywarpx/fields.py
```diff
@@ -516,6 +517,12 @@ def __setitem__(self, index, value):
             if not isinstance(ic   , slice) or len(global_shape) < 4: global_shape[3:3] = [1]
             value3d.shape = global_shape
 
+            if libwarpx.libwarpx_so.Config.have_gpu:
```
ax3l, line 520: I think only needed for  `ndarray`s:
```suggestion
            if libwarpx.libwarpx_so.Config.have_gpu and isinstance(value, np.ndarray)::
```

ax3l (approved):

Issue #3684, opened 2023-02-08
PICMI: Support Python Function or Array for Density Init


shefalys:
Hi @ax3l , Hi @dpgrote , 

thanks a lot for your kind help. I have a request (as a user) to make the plasma density profile more flexible and even useful for experimentally measured density.

Please see the pdf file in the attachment. 

If such features will be included it might be quite useful to simulate conditions more close to experiments. 

Thank you.

[Feature_request.pdf](https://github.com/ECP-WarpX/WarpX/files/10681581/Feature_request.pdf)

ax3l:
Hi @shefalys ,

Slide 2: This is a great request to load a density function via a Python hook.
We could add support for this relatively well soon with our new #3474 bindings, but you will need to use a more general writeup if this shall be runnable on GPU, e.g., using cupy https://docs.cupy.dev/en/stable/user_guide/basic.html#how-to-write-cpu-gpu-agnostic-code

I think a simpler intermediate step would be to support reading and interpolating from an array, as you have on slide 2.

shefalys:
Hi @ax3l , thank you for considering the request. 

I agree with your suggestion and will try to take care of it. 

I am looking forward to the new release with the additional functional option for density etc.

Thanks a lot.

PR #4754, opened 2024-03-06
[WIP] Add ReadDensityFromFile capabilities - Test 1 (3D only)

### LINKED ISSUE #618: Initialize plasma density from file
Hi,
I would like to propose to implement a new option for the initialization of the plasma density: using a (possibly very long) list of geometric primitives (boxes, cylinders, spheres) from a binary file. 
The use case I have in mind is the simulation of laser interaction with very complex nanostructured targets (e.g. https://doi.org/10.1038/s41598-018-22147-6 ), where the target cannot be simply described as a density function.
This is not a top priority for me at the moment, but I would like to discuss if adding this feature could be interesting for WarpX.

### LINKED ISSUE #3948: Alternative method of particle initialisation from a file
Hello,
I have a feature suggestion. I have initialised macro particles from an external openPMD file so that I can use the output of a hydrodynamic input as my starting conditions. This method has worked well however requires conversion of particle number density mesh into macroparticles.

An alternative to this could be supplying the number density as an openPMD mesh and then specifying a variable such as total_macro_particle_number to initialise the particles or a second mesh containing the particle per cell number.

This may be a simpler method for users to create custom targets that can't be easily created using an analytical function.

Thanks,
Daniel

sevicky:
Adding the capability to read a density from openpmd file to use in particle deposition. The current version has only 3D capabilities and is not quite correct with the indexing, which I am continuing to work on but am putting this version in the pull request. Edited files are InjectorDensity.H, InjectorDensity.cpp, and SpeciesUtils.cpp, with most of the changes in the former two. I have tested it using the uniform_plasma example script to give a rho (density) file which is then read in by another input file. 

Close #618
Close #3948

Added by @RemiLehe: TODO:
- [ ] Add automated test: read an openPMD file with multiple MPI ranks ; check that the density of particles corresponds to the expected density.
- [ ] Fix error on GPU
- [ ] Avoid out-of-bounds when reading the array.
- [ ] Handle the case of the moving window, whereby we need to periodically read the file.


sevicky:
The current version in this pull request works for to read a 3D openpmd file and use the density for another simulation. There is no parallelization and further user input options to be implemented.

sevicky:
Linked here are the compilation script, input files, and Jupyter notebook that was used in development.

Compilation script: 
[compile_script.txt](https://github.com/ECP-WarpX/WarpX/files/15135703/compile_script.txt)

Files to create the data used in the tests: 
[plasma_test.txt](https://github.com/ECP-WarpX/WarpX/files/15135704/plasma_test.txt)
[plasma_test_gradient.txt](https://github.com/ECP-WarpX/WarpX/files/15135705/plasma_test_gradient.txt)

input files run to test the parabolic and asymmetric cases respectively: 
[file_input.txt](https://github.com/ECP-WarpX/WarpX/files/15135707/file_input.txt)
[file_asymm.txt](https://github.com/ECP-WarpX/WarpX/files/15135709/file_asymm.txt)

Jupyter Notebook: (actual version is in my Google Drive - 2024-Sarah_Vickers/misc.)
[openpmd_read.pdf](https://github.com/ECP-WarpX/WarpX/files/15135761/openpmd_read.pdf)

aodhanmci:
@ax3l @sevicky is it possible to input a density in a PICMI input file using a function? e.g. if I pass the function x, y, z, it will return rho(x, y, z) for the grid? this means I can add complicated density profiles directly in the input file e.g. micro/nanostructures that are limited to the functions which can be parsed by amrex

ax3l:
Hi @aodhanmci,

Yes, this is already possible for the string based parser syntax and does not require the feature in this pull request.

We support loading density profiles with functions following this syntax:
- top link in https://warpx.readthedocs.io/en/latest/usage/parameters.html
- supports outside-of-function, [pre-defined and user-added constants](https://warpx.readthedocs.io/en/latest/usage/parameters.html#running-cpp-parameters-parser)
- supports a [variety of basic math and logic functions](https://amrex-codes.github.io/amrex/docs_html/Basics.html#parser), e.g., `exp`, `sqrt`, `pow`, `if`, `sin`, `cos`, ...
- support temporary variables, e.g., `"r=sqrt(x*x+y*y+z*z); if(r>0, ..., ...else...)"`
- supports complex multi-line functions and piecewise functions

Here is a laser-ion acceleration example that uses this from PICMI, see `density_expression_str`: https://warpx.readthedocs.io/en/latest/usage/examples/laser_ion/README.html

You can go pretty complex in those expressions as multi-line strings, but you cannot pass Python functions (`def ...(...):` #3684). (Added in DM: support such as random number generators are not in the AMReX parser functions but needed for Aodhan's workflow.) I'll check with others if they have already done the latter, otherwise once the PR here is ready it could be a good basis to support that.

RemiLehe:
It looks like the current code does work (including in cases with multiples boxes), if the openPMD data is C-ordered.

Here are the files that I used to create a C-ordered openPMD file, and then to run the 
simulation.

[file_asymm.txt](https://github.com/user-attachments/files/15962969/file_asymm.txt)
[openpmd_create_3d.py.txt](https://github.com/user-attachments/files/15962971/openpmd_create_3d.py.txt)

~~While the simulation produces correct results, it segfaults later on.~~ It does not segfault anymore.
TODO: 
- [x] Understand current SegFault
- [ ] Reuse function `ReadExternalField` to avoid code duplication

Source/WarpX.H
```diff
@@ -1047,7 +1047,7 @@ public:
      * \brief Load field values from a user-specified openPMD file
      * for a specific field (specified by `F_name`)
      */
-    void ReadExternalFieldFromFile (
+    static void ReadExternalFieldFromFile (
```
RemiLehe, line 1057: We could make this a free function, and move it to the `Utils` folder.

Source/Initialization/InjectorDensity.H
```diff
@@ -118,12 +164,60 @@ private:
...
+        if (x < lo0 or x > hi0 or y < lo1 or y > hi1 or z < lo2 or z > hi2){
+            return 0;
+        }
+
+        int ix = std::floor((x-lo0)/m_cell_size[0]);
+        int iy = std::floor((y-lo1)/m_cell_size[1]);
+        int iz = std::floor((z-lo2)/m_cell_size[2]);
+
+
+        return m_rho_array4(ix,iy,iz);
```
RemiLehe, line 160: We need to make sure that we are not doing out-of-bound access here.

aodhanmci:
im unable to compile these commits (error below). Wonder if anyone can help?

/global/homes/a/aodhan/src/warpx/Source/Initialization/PlasmaInjector.cpp(56): error: static assertion failed with "InjectorDensity must be trivially copyable"
      static_assert(std::is_trivially_copyable<InjectorDensity>::value,
      ^

1 error detected in the compilation of "/global/homes/a/aodhan/src/warpx/Source/Initialization/PlasmaInjector.cpp".
gmake[2]: *** [CMakeFiles/lib_2d.dir/build.make:1588: CMakeFiles/lib_2d.dir/Source/Initialization/PlasmaInjector.cpp.o] Error 2
gmake[2]: *** Waiting for unfinished jobs....
gmake[1]: *** [CMakeFiles/Makefile2:1500: CMakeFiles/lib_rz.dir/all] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:1388: CMakeFiles/lib_2d.dir/all] Error 2

following the 'ordinary' Perlmutter instructions, wonder if anyone can help?

RemiLehe:
Thanks for reporting this @aodhanmci.
This is also one of the errors that is reported in the automated tests (specifically for compilation on GPU). I am not sure that there is a quick fix. We will need to take some dedicated time to finalize this PR.

The code might work on CPUs though.

RemiLehe:
Superseded by #6062

PR #4758, opened 2024-03-08
Replace links to learn git

RemiLehe:
The links that we had for `git` where not ideal:
- One of them was using Bitbucket, which is not what we use in practice
- The other one is more of a cheatsheet, rather than something that introduce `git` to someone that does not know it.

This PR replaces these links with the Software Carpentry tutorial (which does use Github)

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -10,10 +10,7 @@ Git workflow
 ------------
 
 The WarpX project uses `git <https://git-scm.com>`_ for version control.
-If you are new to git, you can follow one of these tutorials:
-
-- `Learn git with bitbucket <https://www.atlassian.com/git/tutorials/learn-git-with-bitbucket-cloud>`_
-- `git - the simple guide <http://rogerdudler.github.io/git-guide/>`_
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -10,10 +10,7 @@ Git workflow
 ------------
 
 The WarpX project uses `git <https://git-scm.com>`_ for version control.
-If you are new to git, you can follow one of these tutorials:
-
-- `Learn git with bitbucket <https://www.atlassian.com/git/tutorials/learn-git-with-bitbucket-cloud>`_
-- `git - the simple guide <http://rogerdudler.github.io/git-guide/>`_
+If you are new to git, you can follow `this tutorial <https://swcarpentry.github.io/git-novice/>`_.
```
ax3l: Anonymous link because link text is quite generic:
```suggestion
If you are new to git, you can follow `this tutorial <https://swcarpentry.github.io/git-novice/>`__.
```

ax3l (approved):
That's way better, thx!

PR #4769, opened 2024-03-13
CI: `isort` + `black` compatibility

ax3l:
Avoid that these tools follow slightly different rules.

### DIFF
--- a/Examples/Tests/rigid_injection/analysis_rigid_injection_BoostedFrame.py
+++ b/Examples/Tests/rigid_injection/analysis_rigid_injection_BoostedFrame.py
@@ -26,7 +26,6 @@ import sys
 import numpy as np
 import openpmd_api as io
 import yt
-from scipy.constants import m_e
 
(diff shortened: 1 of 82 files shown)


Examples/Tests/rigid_injection/analysis_rigid_injection_BoostedFrame.py
```diff
@@ -25,8 +25,8 @@
 
 import numpy as np
 import openpmd_api as io
-from scipy.constants import m_e
 import yt
+from scipy.constants import m_e
```
lucafedeli88:
```suggestion
```

I think that with just this minor change the PR would be ready to be merged.

lucafedeli88 (changes requested):

lucafedeli88 (approved):
Thanks for this PR, @ax3l .


`clang-tidy` CI test is currently failing but this other PR should fix the issue:
https://github.com/ECP-WarpX/WarpX/pull/4781

PR #4774, opened 2024-03-13
Species supercycling

archermarx:
This PR adds the ability to advance different species at different multiples of the provided `const_dt`. It introduces two new parameters, `<species>.do_supercycling` and `<species>.supercycling_interval`, which determine whether the provided species is to be supercycled, and the interval (in terms of iterations) at which they are to be advanced. It also updates the PICMI species interface to support these parameters. This is useful in cases where the dynamics of interest are dominated by a lighter species which must be advanced at a much smaller timestep (i.e. electrons in an electron-ion plasma), while allowing ions or other heavier species to advance at a much larger timestep.

This is my first PR, so let me know if I'm missing anything critical or if there's a better way to handle this than how I implemented it!


Source/Particles/MultiParticleContainer.cpp
```diff
@@ -468,9 +468,21 @@ MultiParticleContainer::Evolve (int lev,
         if (rho) { rho->setVal(0.0); }
         if (crho) { crho->setVal(0.0); }
     }
+
+    // Get current iteration for use in supercycling
+    auto& warpx = WarpX::GetInstance();
+    int iter = warpx.getistep(0);
```
dpgrote: Since this is called from WarpX, the `iter` could be passed in as an argument (the same way the `t` and `dt` are).

Docs/source/usage/parameters.rst
```diff
@@ -1259,6 +1259,20 @@ Particle initialization
     Resampling is performed everytime the number of macroparticles per cell of the species
     averaged over the whole simulation domain exceeds this parameter.
 
+* ``<species>.do_supercycling`` (`0` or `1`) optional (default `0`)
```
dpgrote: This flag is not really needed as a separate input argument. Giving `supercycling_interval` a value other than 1 would be sufficient to turn on the super cycling.

dpgrote:
Overall, this looks good. Is this intended for electrostatic or electromagnetic?
Note that for electrostatic, the charge is deposited elsewhere, so as you have it, the charge for all particles will still be deposited every step. For electromagnetic, this does skip the current deposition for the slower moving species, so the field advance each step will have only the current of the faster moving species. Is this what is intended?

archermarx:
> Overall, this looks good. Is this intended for electrostatic or electromagnetic? Note that for electrostatic, the charge is deposited elsewhere, so as you have it, the charge for all particles will still be deposited every step. For electromagnetic, this does skip the current deposition for the slower moving species, so the field advance each step will have only the current of the faster moving species. Is this what is intended?

Thanks for taking a look! This is currently intended only for electrostatic. I added a check in `ComputeDt` to verify that a constant dt has been passed, but perhaps I should also update that to make sure we're using the electrostatic solver.

archermarx:
>Note that for electrostatic, the charge is deposited elsewhere, so as you have it, the charge for all particles will still be deposited every step. 

Thanks for pointing this out. I thought that was handled in `MultiParticleContainer->Evolve` but evidently that isn't correct. I'll see if i can fix this because we would ideally only deposit charge from the slower species on the iterations at which their positions actually change.

archermarx:
If I'm understanding the charge deposition for the electrostatic solver correctly (looking at `WarpX::AddSpaceChargeField` in `ElectrostaticSolver.cpp`), It looks like the space charge fields are accumulated onto a common value which is reset every iteration. This means that we have to deposit charge from all species at each timestep, regardless of supercycling, or else the fields would be massively wrong on the iterations where the slower species aren't pushed. It doesn't look like there's an easy way to store the computed fields/charge density from a previous iteration for use on the supercycled iterations. 

If that's correct, then I think this PR is fine without that, and perhaps a future PR could address that part.

Source/Particles/MultiParticleContainer.cpp
archermarx: Good idea, done!

Docs/source/usage/parameters.rst
archermarx: Done!

Source/Evolve/WarpXComputeDt.cpp
```diff
@@ -32,15 +33,31 @@
 void
 WarpX::ComputeDt ()
 {
+    // Check if any species is using supercycling
+    bool do_supercycling = false;
+    int nSpecies = mypc->nSpecies();
```
lucafedeli88:
```suggestion
    const int nSpecies = mypc->nSpecies();
```
Hi @archermarx ! I think that the absence of `const` here is causing the `clang-tidy` CI test to fail.

archermarx: Thanks! Fixed.

archermarx:
@dpgrote  @lucafedeli88 I think I've addressed all the comments. Let me know if there's anything else you'd like me to change!

archermarx:
Looks like clang-tidy is still failing, but I don't think it's anything I've done, the error seems to be 

```
/home/runner/work/WarpX/WarpX/Source/Utils/WarpXAlgorithmSelection.cpp:252:17: error: redundant string initialization [readability-redundant-string-init,-warnings-as-errors]
    std::string boundary_name = "";
Suppressed 120982 warnings (120917 in non-user code, 65 NOLINT).
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
1 warning treated as error
                ^~~~~~~~~~~~~~~~~~
                boundary_name
```

Looks like this was introduced in [this PR](https://github.com/ECP-WarpX/WarpX/pull/4778). I've fixed it in this one.

ax3l:
Comment by @jlvay in the developer meeting:
One can save even more by depositing the supercycled species (in a separate array) for the `supercycling_interval` forward in time and then interpolating every step.

PR #4778, opened 2024-03-14
Better error message when Poisson solver does not have supported boundary

RemiLehe:
The error message now shows the boundary that the user is currently using.


roelof-groenewald (approved):
I think this will definitely help users understand why their input files fail when using inappropriate boundary conditions. Thanks!

Source/Utils/WarpXAlgorithmSelection.cpp
```diff
@@ -246,3 +246,15 @@ GetParticleBCTypeInteger( std::string BCType ){
     // return ParticleBCType_algo_to_enum[BCType]; // This operator cannot be used for a const map
     return ParticleBCType_algo_to_enum.at(BCType);
 }
+
+std::string
+GetFieldBCTypeString( int fb_type ) {
+    std::string boundary_name = "";
+    for (const auto &valid_pair : FieldBCType_algo_to_int) {
+        if ((valid_pair.second == fb_type)&(valid_pair.first != "default")){
```
RemiLehe:
```suggestion
        if ((valid_pair.second == fb_type)&&(valid_pair.first != "default")){
```

PR #4779, opened 2024-03-15
Fix ParticleHistogram2D

pordyna:
Without `m_h_data_2D.clear();` the host memory usage continues growing over the simulation until the process gets killed.
(Observed in `24.3` and `24.2`)

`i.close();` and `series.close();` were my first suspicion. It didn't help, but it is still probably a good idea to explicitly close the iteration and series.

### DIFF
--- a/Source/Diagnostics/ReducedDiags/ParticleHistogram2D.cpp
+++ b/Source/Diagnostics/ReducedDiags/ParticleHistogram2D.cpp
@@ -138,6 +138,7 @@ void ParticleHistogram2D::ComputeDiags (int step)
     Array<int,2> tlo{0,0}; // lower bounds
     Array<int,2> thi{m_bin_num_abs-1, m_bin_num_ord-1}; // inclusive upper bounds
     amrex::TableData<amrex::Real,2> d_data_2D(tlo, thi);
+    m_h_data_2D.clear();
     m_h_data_2D = amrex::TableData<amrex::Real,2> (tlo, thi, The_Pinned_Arena());
     auto const& h_table_data = m_h_data_2D.table();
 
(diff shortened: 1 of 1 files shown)


Source/Diagnostics/ReducedDiags/ParticleHistogram2D.cpp
```diff
@@ -138,6 +138,7 @@ void ParticleHistogram2D::ComputeDiags (int step)
     Array<int,2> tlo{0,0}; // lower bounds
     Array<int,2> thi{m_bin_num_abs-1, m_bin_num_ord-1}; // inclusive upper bounds
     amrex::TableData<amrex::Real,2> d_data_2D(tlo, thi);
+    m_h_data_2D.clear();
```
ax3l: @WeiqunZhang do you see why we leaked (host) memory here?

WeiqunZhang: It's an amrex bug. :blush:  https://github.com/AMReX-Codes/amrex/pull/3807

Source/Diagnostics/ReducedDiags/ParticleHistogram2D.cpp
```diff
@@ -138,6 +138,7 @@ void ParticleHistogram2D::ComputeDiags (int step)
...
     m_h_data_2D = amrex::TableData<amrex::Real,2> (tlo, thi, The_Pinned_Arena());
```
WeiqunZhang:
```suggestion
    m_h_data_2D.resize(tlo, thi, The_Pinned_Arena());
```
This is more efficient.

WeiqunZhang:
For the initialization of `h_table_data`, could we switch the order of the two for loops? Because TableData has column major ordering, we want to make the i-loop the innermost loop.

ax3l:
#4800 should have brought in the fix for https://github.com/AMReX-Codes/amrex/pull/3807

Source/Diagnostics/ReducedDiags/ParticleHistogram2D.cpp
ax3l: WarpX PR #4800 should have brought in the fix from AMReX now.

ax3l (approved):
Thank you! :)

PR #4780, opened 2024-03-15
Define class function `defineAllParticleTiles` for `NamedComponentParticleContainer`

RemiLehe:
The function `defineAllParticleTiles` is useful (esp. after creating runtime-components), but is currently only defined for the `WarpXParticleContainer`. 
This PR defines it instead for the `WarpXParticleContainer`'s parent class (i.e. `NamedComponentParticleComponent`).


Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -1517,6 +1517,10 @@ WarpXParticleContainer::PushX (int lev, amrex::Real dt)
 // without runtime component).
 void WarpXParticleContainer::defineAllParticleTiles () noexcept
 {
+    // Call the parent class's method
+    NamedComponentParticleContainer<amrex::DefaultAllocator>::defineAllParticleTiles();
+
+    // Resize the tmp_particle_data (no present in parent class)
     tmp_particle_data.resize(finestLevel()+1);
```
RemiLehe, line 1524: @atmyers Out of curiosity, do you remember what the `tmp_particle_data` is used for?

atmyers: It's used in the back transformed diagnostics, IIRC to store the old particle positions.

atmyers (approved):

PR #4781, opened 2024-03-15
Fix redundant string init in `WarpXAlgorithmSelection.cpp`

lucafedeli88:
This PR fixes a minor issue found with `clang-tidy` that currently causes CI tests to fail.


RemiLehe (approved):

ax3l:
Thanks @lucafedeli88 !

PR #4783, opened 2024-03-18
Add UB sanitizer CI test

lucafedeli88:
This PR adds an Undefined Behavior sanitizer test in CI (see https://releases.llvm.org/15.0.0/tools/clang/docs/UndefinedBehaviorSanitizer.html for more details).

The PR also fixes a bug found with the UB sanitizer.

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -258,7 +258,7 @@ jobs:
           -DWarpX_PRECISION=SINGLE          \
           -DWarpX_PARTICLE_PRECISION=SINGLE \
           -DWarpX_QED_TABLE_GEN=ON
-        cmake --build build -j 4 VERBOSE=TRUE
+        cmake --build build -j 4
 
         ccache -s
         du -hs ~/.cache/ccache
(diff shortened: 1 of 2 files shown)


roelof-groenewald (commented):
It's cool that the new check found a possible bug in execution here. I wonder, though, if it wouldn't make more sense to add
```
if (np == 0) { return; }
```
at line 79 of `FieldProbeParticleContainer.cpp` instead of skipping over the part that creates new particles.

lucafedeli88:
> It's cool that the new check found a possible bug in execution here. I wonder, though, if it wouldn't make more sense to add
> 
> ```
> if (np == 0) { return; }
> ```
> 
> at line 79 of `FieldProbeParticleContainer.cpp` instead of skipping over the part that creates new particles.

I agree with you. However, we still need to call `Redistribute();`, since it is a collective operation.
I've added these lines:

```
    if (np <= 0){
        Redistribute();
        return;
    }
```

roelof-groenewald (approved):
Thanks for updating!

ax3l:
Cool, thx! :)

PR #4784, opened 2024-03-18
Add callbacks after B and E field pushes

roelof-groenewald:
I need to insert some logic to modify the B-field right after it is advanced (akin to applying a special boundary condition), which the `afterBpush` callback facilitates. I added the `afterEpush` callback for symmetry.


lucafedeli88 (approved):
Looks good to me!

PR #4789, opened 2024-03-19
[WIP] Removed assert and allowed bilinear filtering along r with RZ FDTD

oshapoval:
- To do:   we might want to remove the charge buffer from WarpX within a separate PR, which should also fix one of the failing CI tests.

This PR removes an abort message which prevented from using bilinear filtering for charge and currents on the mesh, along `r` with RZ FDTD. Previously this assert message was added because we had observed strange results in RZ FDTD when filtering was ON ( see [#1943](https://github.com/ECP-WarpX/WarpX/issues/1943) ) that we do not observe anymore.

Now, the default is `warpx.use_filter=1` in all cases, including the simulations with RZ FDTD solver. Note, with the RZ PSATD solver, the filtering is done `k`-space. 
One can also set the number of passes along each direction `warpx.filter_npass_each_dir` (`3 int`), and default is `1 1 1`.

Screenshot 2024-03-25 at 2 23 24 PM
Here is the corresponding input script: 

[inputs_rz.txt](https://github.com/ECP-WarpX/WarpX/files/14749699/inputs_rz.txt)


Examples/Physics_applications/laser_acceleration/inputs_rz
```diff
@@ -20,8 +20,7 @@ boundary.field_hi = pec pec
 #################################
 warpx.verbose = 1
 warpx.do_dive_cleaning = 0
-warpx.use_filter = 1
-warpx.filter_npass_each_dir = 0 1
+#warpx.use_filter = 1
```
RemiLehe: Wouldn't it be better to use `use_filter = 1` here, to test the new capability?

oshapoval: Yes, here we want to use and set `warpx.use_filter = 1` explicitly, still needed to clean up after testing. Thank you for spotting this. Fixed in 6d87779fb2c2d345007.

ax3l:
@oshapoval can you locally take a look why `ElectrostaticSphereEB_RZ_MR` crashes pls?

PR #4790, opened 2024-03-20
Add thermalBC for particles

RevathiJambunathan:
redoing #4681 
@RemiLehe added a PR to demonstrate thermal BC 
This PR does the same, with some restructuring, and changes to API

1. Tested locally and ensured that the velocity distribution is Maxwellian (thermalized) even at the end of the simulation.

2. CI test is added to ensure that the particle energy does not increase by more than 2% 
    (if particle boundary is reflecting or thermal without "folding of the current density and charge", then the particle energy increase by more than 10% )


Source/Initialization/SampleGaussianDistribution.H
```diff
@@ -0,0 +1,81 @@
+/* Copyright 2024 Remi Lehe, Revathi Jambunathan
```
RemiLehe: Could we call this file `SampleGaussianFluxDistribution` instead of `SampleGaussianDistribution`?
This is because the distribution below is consistently called "gaussian-flux" in WarpX, to distinguish it from a regular Gaussian distribution.

Source/Particles/ParticleBoundaries_K.H
```diff
@@ -61,16 +66,38 @@ namespace ApplyParticleBoundaries {
...
+     *        with thermal velocity uth
+     */
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    void thermalize_boundary_particle (amrex::ParticleReal& u_norm, amrex::ParticleReal& u_tang1,
+                                       amrex::ParticleReal& u_tang2, amrex::Real uth,
+                                       amrex::RandomEngine const& engine)
+    {
+        u_tang1 = (uth > 0._rt) ? PhysConst::c * amrex::RandomNormal(0._rt, uth, engine) : 0._rt;
+        u_tang2 = (uth > 0._rt) ? PhysConst::c * amrex::RandomNormal(0._rt, uth, engine) : 0._rt;
+        u_norm = -1._rt * PhysConst::c * generateGaussianFluxDist(0._rt, uth, engine) * ((u_norm>0)-(u_norm<0));
```
dpgrote: This would be the more standard usage
```suggestion
        u_norm = std::copysign(1._prt, -u_norm) * PhysConst::c * generateGaussianFluxDist(0._rt, uth, engine);
```

Source/Particles/ParticleBoundaries_K.H
```diff
@@ -101,24 +128,45 @@ namespace ApplyParticleBoundaries {
...
+        amrex::ignore_unused(rethermalize_y);
 #endif
-        apply_boundary(z, zmin, zmax, change_sign_uz, particle_lost,
+        apply_boundary(z, zmin, zmax, change_sign_uz, rethermalize_z, particle_lost,
                        boundaries.zmin_bc, boundaries.zmax_bc,
                        boundaries.reflection_model_zlo(-uz), boundaries.reflection_model_zhi(uz),
                        engine);
 
+#ifndef WARPX_DIM_1D_Z
+        if (rethermalize_x) {
```
dpgrote: This code might be cleaner if this was put in the same `ifdef/endif` block as the associated call to apply_boundary. Also including the declaration of the boolean would allow removing of the `ignore_unused` lines.

Source/Particles/ParticleBoundaries_K.H
RevathiJambunathan: nice! Thanks!

Source/Particles/ParticleBoundaries_K.H
RevathiJambunathan: done

Examples/Tests/particle_thermal_boundary/analysis_2d.py
```diff
@@ -0,0 +1,34 @@
...
+beyond 2% in the time that it takes all particles to cross the domain boundary
+"""
+
+
+import sys
+
+import numpy as np
+
+sys.path.insert(1,'../../../../warpx/Regression/Checksum/')
+import checksumAPI
```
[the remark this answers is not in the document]
RevathiJambunathan: I am not sure yet if we should include/exclude checksum for this test, because its very noisy.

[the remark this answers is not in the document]
dpgrote: I would say include the checksum. In the CI test, is will always use the same random numbers and so should always get the same result.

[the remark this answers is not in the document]
RevathiJambunathan: I agree. 
What about the random differences between the json file I generate locally, and that generated in CI? 
I will still add this checksum back and lets see the differences, and may be add a larger tolerance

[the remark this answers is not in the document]
RemiLehe: Instead of using larger tolerances, it is best to directly use the benchmark values generated by the CI.
(See https://warpx.readthedocs.io/en/latest/developers/checksum.html#reset-a-benchmark-from-the-azure-pipeline-output-on-github)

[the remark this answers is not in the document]
RevathiJambunathan: great!!! thanks!

Examples/Tests/particle_thermal_boundary/analysis_2d.py
```diff
@@ -0,0 +1,37 @@
...
+init_Fenergy = np.loadtxt(FE_rdiag)[1,2]
+final_Fenergy = np.loadtxt(FE_rdiag)[-1,2]
+assert(final_Fenergy/init_Fenergy < 40)
+assert(final_Fenergy < 5.e-5)
+
+PE_rdiag = './diags/reducedfiles/EN.txt'
+init_Penergy = np.loadtxt(PE_rdiag)[0,2]
+final_Penergy = np.loadtxt(PE_rdiag)[-1,2]
+assert( abs(final_Penergy - init_Penergy)/init_Penergy < 0.02)
+
```
RemiLehe:
```suggestion
filename = sys.argv[1]
```

Source/Particles/ParticleBoundaries_K.H
```diff
@@ -61,16 +66,38 @@ namespace ApplyParticleBoundaries {
...
+
+    /* \brief Applies absorbing, reflecting or thermal boundary condition to the input particles, along all axis.
      *        For reflecting boundaries, the position of the particle is changed appropriately and
      *        the sign of the velocity is changed (depending on the reflect_all_velocities flag).
      *        For absorbing, a flag is set whether the particle has been lost (it is up to the calling
      *        code to take appropriate action to remove any lost particles). Absorbing boundaries can
      *        be given a reflection coefficient for stochastic reflection of particles, this
      *        coefficient is zero by default.
+     *        For thermal boundaries, the particle is first reflected and the positrion of the particle
+     *        is change appropriately. The sign of the ve
```
RemiLehe, line 100:
```suggestion
     *        is change appropriately.
```

RemiLehe:
```suggestion
     *        For thermal boundaries, the particle is first reflected and the position of the particle
     *        is changed appropriately.
```

RemiLehe (approved):

Regression/Checksum/benchmarks_json/particle_thermal_boundary.json
```diff
@@ -0,0 +1,12 @@
...
+  "lev=0": {
+    "Bx": 229.28366410112736,
+    "By": 242.40221902487008,
+    "Bz": 164.70543954019666,
+    "Ex": 789234700854.9321,
+    "Ey": 15892271613.51144,
+    "Ez": 693631132513.1528,
+    "divE": 5.441439446237374e+19,
+    "rho": 331367728.98455656
+  }
```
RemiLehe:
```suggestion
  "lev=0": {
    "Bx": 167.49421217759414,
    "By": 269.6793406439766,
    "Bz": 258.3282607081799,
    "Ex": 864657516595.9229,
    "Ey": 24072965021.088013,
    "Ez": 702786324880.7338,
    "divE": 5.137699003981513e+19,
    "rho": 357057125.6699786
  }
```

PR #4736, opened 2024-02-29
New Implicit Solver interface with options to select Picard or Newton (JFNK) for the nonlinear solver.

JustinRayAngus:
There are two new folders: `Source/FieldSolver/ImplicitSolvers/` and `Source/NonlinearSolvers/`. All of the containers needed by the implicit time solver and corresponding nonlinear/linear solvers are now owned by the Implicit Solver and nonlinear/ linear solver class rather than by WarpX. WarpX owns a pointer to ImplicitSolver class object. The base functions in the implicit solver class are those needed to interact with WarpX ( such as OneStep ) and with the nonlinear/linear solvers. The current choices for implicit time solvers are SemiImplicitEM and ThetaImplicitEM. The current nonlinear solver choices are Picard and Newton. The Newton solver is coupled to AMReX_GMRES for solving the linear system needed to obtain the Newton step at each Newton iteration in the standard JFNK method. Both the semi-implicit and theta-implicit algorithms can use particle-suppressed (PS)-JFNK for the nonlinear solver. However, it is not yet setup to use mass matrices or preconditioning, and thus it is not yet efficient.

image002

PR #4791, opened 2024-03-20
Evolve Loop: Simplify

ax3l:
Introduce new functions to simplify the number of calls in our main evolve loop.

I will probably do another follow-up PR on the sub-stepping calls, but try to avoid colliding too much with @JustinRayAngus's PRs (e.g., #4736).

### DIFF
--- a/Source/Evolve/WarpXEvolve.cpp
+++ b/Source/Evolve/WarpXEvolve.cpp
@@ -70,9 +70,6 @@ WarpX::Evolve (int numsteps)
     // Note that the default argument is numsteps = -1
     const int numsteps_max = (numsteps < 0)?(max_step):(istep[0] + numsteps);
 
-    bool early_params_checked = false; // check typos in inputs after step 1
-    bool exit_loop_due_to_interrupt_signal = false;
-
     static Real evolve_time = 0;
 
     const int step_begin = istep[0];
@@ -81,7 +78,7 @@ WarpX::Evolve (int numsteps)
         WARPX_PROFILE("WarpX::Evolve::step");
         const auto evolve_time_beg_step = static_cast<Real>(amrex::second());
 
-        //Check and clear signal flags and asynchronously broadcast them from process 0
+        // Check and clear signal flags and asynchronously broadcast them from process 0
         SignalHandling::CheckSignals();
 
         multi_diags->NewIteration();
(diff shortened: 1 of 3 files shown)


Source/Evolve/WarpXEvolve.cpp
```diff
@@ -70,9 +70,6 @@ WarpX::Evolve (int numsteps)
     // Note that the default argument is numsteps = -1
     const int numsteps_max = (numsteps < 0)?(max_step):(istep[0] + numsteps);
 
-    bool early_params_checked = false; // check typos in inputs after step 1
```
dpgrote, line 73: My own opinion is that since these flags are only used within this routine, making them class attributes is excessive and adds to bloat of the WarpX class.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -377,15 +273,7 @@ WarpX::Evolve (int numsteps)
         ExecutePythonCallback("afterdiagnostics");
 
         // inputs: unused parameters (e.g. typos) check after step 1 has finished
-        if (!early_params_checked) {
-            amrex::Print() << "\n"; // better: conditional \n based on return value
```
dpgrote, line 381: I would put only lines 381 - 386 into a subroutine (keeping the flag a local variable).
```
if (!early_params_checked) {
    checkEarlyUnusedParams();
    early_params_checked = true;
}
```

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -284,49 +222,7 @@ WarpX::Evolve (int numsteps)
             }
         }
 
-        mypc->ContinuousFluxInjection(cur_time, dt[0]);
```
dpgrote, line 287: I'm not sure that it is helpful to put this single line calls into a separate routine. For me, this would make it more difficult to follow what the evolve step is doing since these would be hidden.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -284,49 +222,7 @@ WarpX::Evolve (int numsteps)
...
-
-        // interact the particles with EB walls (if present)
-#ifdef AMREX_USE_EB
-        mypc->ScrapeParticles(amrex::GetVecOfConstPtrs(m_distance_to_eb));
-#endif
-
-        m_particle_boundary_buffer->gatherParticles(*mypc, amrex::GetVecOfConstPtrs(m_distance_to_eb));
-
-        // Non-Maxwell solver: particles can move by an arbitrary number of cells
-        if( electromagnetic_solver_id == ElectromagneticSolverAlgo::None ||
```
dpgrote, line 299: My preference would be to have only this `if` statement wrapped in a subroutine.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -284,49 +222,7 @@ WarpX::Evolve (int numsteps)
...
-                    num_redistribute_ghost += 1;
-                }
-                mypc->RedistributeLocal(num_redistribute_ghost);
-            }
-            else {
-                mypc->Redistribute();
-            }
-        }
-
-        if (sort_intervals.contains(step+1)) {
```
dpgrote, line 324: This `if` check could be wrapped into the `mypc->SortParticlesByBin` routine if the sorting related variables are moved to the `MultiParticleContainer` class. That should be done anyway since that is the only place where `sort_interval` and `sort_bin_size` are used.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -401,20 +289,18 @@ WarpX::Evolve (int numsteps)
                       << " s; Avg. per step = " << evolve_time/(step-step_begin+1) << " s\n\n";
         }
 
-        exit_loop_due_to_interrupt_signal = SignalHandling::TestAndResetActionRequestFlag(SignalHandling::SIGNAL_REQUESTS_BREAK);
```
dpgrote, line 404: I would keep this here since it makes it clearer what is happening (and keeps that flag local).

Source/Evolve/WarpXEvolve.cpp
ax3l: I thought so too, but here it might be justified because our exit logic is non-trivial (and outlives the routine that checks). We can explore if a return type is easier.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -92,90 +89,29 @@ WarpX::Evolve (int numsteps)
...
-                    if (WarpX::electromagnetic_solver_id != ElectromagneticSolverAlgo::PSATD) {
-                        FillBoundaryAux(guard_cells.ng_UpdateAux);
-                    }
-                }
-                UpdateAuxilaryData();
-                FillBoundaryAux(guard_cells.ng_UpdateAux);
-            }
+        if (evolve_scheme == EvolveScheme::Explicit)
+        {
+            ExplicitEBUpdate();
```
dpgrote: Maybe make this routine name more descriptive. Maybe `ExplicitFillBoundaryEBUpdateAux`?

Source/Evolve/WarpXEvolve.cpp
ax3l: I think you might have misread, `InjectRedistributeScapeParticles` includes:
- injection,
- particle BC
- scraping
- redistribute

for particles. From the doc string: `Inject, communicate, scape and sort particles`

Source/Evolve/WarpXEvolve.cpp
ax3l: see above. Moves what belongs together for particle updates into a subroutine (Inject, communicate, scape and sort particles)

Source/Evolve/WarpXEvolve.cpp
ax3l: see above.

Could be further substructured into the MPC class (or free standing functions) if we like :)

Source/Evolve/WarpXEvolve.cpp
ax3l: I find the whole if condition a bit verbose, moving it (and further logic on how to stop a simulation) into a separate call helps simplify it, I find.
I do not think that signal handling logic belongs explicitly here, it is just one out of many ways to end a sim.

Source/Evolve/WarpXEvolve.cpp
ax3l: Good idea, updating.

ax3l: There is a general design question we can discuss: do we want to have slim calls to sub-functions or a log of the call logic in the call site. The former is a bit easier to expose to Python (less verbose), the latter needs less member variable state.

Source/Evolve/WarpXEvolve.cpp
ax3l: Updated as suggested.

Source/Evolve/WarpXEvolve.cpp
ax3l: Renamed, thanks for the suggestion!

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -526,6 +418,124 @@ WarpX::OneStep_nosub (Real cur_time)
...
+bool WarpX::checkStopSimulation (amrex::Real cur_time)
+{
+    m_exit_loop_due_to_interrupt_signal = SignalHandling::TestAndResetActionRequestFlag(SignalHandling::SIGNAL_REQUESTS_BREAK);
+    if (cur_time >= stop_time - 1.e-3*dt[0] || m_exit_loop_due_to_interrupt_signal) {
+        return true;
+    }
+    else
+    {
+        return false;
+    }
```
lucafedeli88:
```suggestion
    return (cur_time >= stop_time - 1.e-3*dt[0])  ||
        m_exit_loop_due_to_interrupt_signal;
```
`clang-tidy` suggests doing this. l think that's a good idea: it makes this section less verbose.

Source/Evolve/WarpXEvolve.cpp
dpgrote: I understand what this is doing - my comment is just that having the calls here makes it easier to following what is happening without having to jump to some other routine to see it.

All of these actions are handling particles at the boundaries. Perhaps call the routine `HandleParticlesAtBoundaries` (or something less clunky).

Further, almost all of these are using `mypc` so this new function could be put into that class. And also put `m_particle_boundary_buffer` into that class as well since it is particle related.

Source/Evolve/WarpXEvolve.cpp
dpgrote: Ok. I would suggest a comment here that `checkStopSimulation` has the side effect of setting `m_exit_loop_due_to_interrupt_signal`, otherwise the behavior might seem mysterious.
BTW, I had submitted a PR to clean this up, but it was never merged, PR #3881. That would simply this so that the extra routine may not be needed.

lucafedeli88 (approved):
Thanks for this PR, @ax3l

Source/Evolve/WarpXEvolve.cpp
ax3l: Rename sounds great - applied.

Moving to the other class is a good idea, let's do in a follow-up to keep this simple here.

Source/WarpX.H
```diff
@@ -1722,8 +1735,41 @@ private:
...
+    /** Print Unused Parameter Warnings after Step 1
+     *
+     * Instead of waiting for a simulation to end, we already do an early "unused parameter check"
+     * after step 1 to inform users early of potential issues with their simulation setup.
+     */
+    void checkEarlyUnusedParams ();
+
+    /** Perform essential particle house keeping at boundaries
+     *
+     * Inject, communicate, scape and sort particles.
```
dpgrote: Fixes a typo
```suggestion
     * Inject, communicate, scrape and sort particles.
```

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -186,6 +125,7 @@ WarpX::Evolve (int numsteps)
 
         ExecutePythonCallback("particleinjection");
 
+        // TOOD: move out
```
roelof-groenewald: Typo: "TODO"

roelof-groenewald (approved):
Looks like the `embedded_circle` test crashed, but this seems to happen sporadically (not related to this PR). I noticed one typo which can be used as an excuse to push another commit and hopefully that test won't crash again 😄

Source/Evolve/WarpXEvolve.cpp
ax3l:
```suggestion
        // TODO: move out
```

PR #4792, opened 2024-03-20
Update spacecraft charging test

RemiLehe:
The spacecraft charging test was not properly correctly the electric field in the simulation with callback functions.

The purpose of the callback functions here is to correct the electric field, so that the charge on the spacecraft computed using Gauss law (from the electric field) matches the actual charge (calculated from collecting particles on the embedded boundary).

However, the current script corrects a **copy** of the electric field. The actual electric field in the simulation was not corrected. This can be seen from the `print` statements when the code is running. (The `print` statements print the charge from Gauss law and the actual charge, and they did not match with the current version of the script.)

### DIFF
--- a/Examples/Physics_applications/spacecraft_charging/PICMI_inputs_rz.py
+++ b/Examples/Physics_applications/spacecraft_charging/PICMI_inputs_rz.py
@@ -25,7 +25,8 @@ from pywarpx.particle_containers import ParticleBoundaryBufferWrapper
 class SpaceChargeFieldCorrector(object):
     """
     Class used by the callback functions to calculate the
-    correct charge on the spacecraft at each initialisation.
+    correct field around the spacecraft, at each timestep
+    (taking into account the charge that has been collected on the spacecraft)
     """
     def __init__(self):
         self.saved_first_iteration_fields = False
@@ -46,11 +47,11 @@ class SpaceChargeFieldCorrector(object):
             q = compute_actual_charge_on_spacecraft()
 
         # Correct fields so as to recover the actual charge
-        Er = ExWrapper(include_ghosts=True)[:,:]
+        Er = ExWrapper(include_ghosts=True)
         Er[...] = Er[...]+(q - q_v)*self.normalized_Er[...]
-        Ez = EzWrapper(include_ghosts=True)[:,:]
+        Ez = EzWrapper(include_ghosts=True)
         Ez[...]  += (q - q_v)*self.normalized_Ez[...]
-        phi = PhiFPWrapper(include_ghosts=True)[:,:]
+        phi = PhiFPWrapper(include_ghosts=True)
         phi[...]  += (q - q_v)*self.normalized_phi[...]
         self.spacecraft_potential += (q - q_v)*self.spacecraft_capacitance
         sim.extension.warpx.set_potential_on_eb( "%f" %self.spacecraft_potential )
--- a/Examples/Physics_applications/spacecraft_charging/PICMI_inputs_rz.py
+++ b/Examples/Physics_applications/spacecraft_charging/PICMI_inputs_rz.py
@@ -48,11 +48,11 @@ class SpaceChargeFieldCorrector(object):
 
         # Correct fields so as to recover the actual charge
         Er = ExWrapper(include_ghosts=True)
-        Er[...] += (q - q_v)*self.normalized_Er[...]
+        Er[...] += (q - q_v)*self.normalized_Er
         Ez = EzWrapper(include_ghosts=True)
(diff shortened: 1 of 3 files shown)


Examples/Physics_applications/spacecraft_charging/PICMI_inputs_rz.py
```diff
@@ -46,11 +47,11 @@ def correct_space_charge_fields(self, q=None):
             q = compute_actual_charge_on_spacecraft()
 
         # Correct fields so as to recover the actual charge
-        Er = ExWrapper(include_ghosts=True)[:,:]
+        Er = ExWrapper(include_ghosts=True)
         Er[...] = Er[...]+(q - q_v)*self.normalized_Er[...]
```
RemiLehe:
```suggestion
        Er[...] += (q - q_v)*self.normalized_Er[...]
```

Examples/Physics_applications/spacecraft_charging/PICMI_inputs_rz.py
```diff
@@ -46,11 +47,11 @@ def correct_space_charge_fields(self, q=None):
             q = compute_actual_charge_on_spacecraft()
 
         # Correct fields so as to recover the actual charge
-        Er = ExWrapper(include_ghosts=True)[:,:]
```
RemiLehe, line 49: Using `[:,:]` creates a copy, as @dpgrote pointed out to me.

Examples/Physics_applications/spacecraft_charging/PICMI_inputs_rz.py
```diff
@@ -46,11 +47,11 @@ def correct_space_charge_fields(self, q=None):
...
-        Er[...] = Er[...]+(q - q_v)*self.normalized_Er[...]
-        Ez = EzWrapper(include_ghosts=True)[:,:]
+        Er = ExWrapper(include_ghosts=True)
+        Er[...] += (q - q_v)*self.normalized_Er[...]
```
dpgrote: This could be cleaned up a bit more. Since `self.normalized_Er` is an already a numpy array, the `[...]` is not needed. And the same for `Ez` and `phi` below.

RemiLehe:
```suggestion
        Er[...] += (q - q_v)*self.normalized_Er
```

Examples/Physics_applications/spacecraft_charging/PICMI_inputs_rz.py
```diff
@@ -46,11 +47,11 @@ def correct_space_charge_fields(self, q=None):
...
+        Ez = EzWrapper(include_ghosts=True)
         Ez[...]  += (q - q_v)*self.normalized_Ez[...]
```
RemiLehe:
```suggestion
        Ez[...]  += (q - q_v)*self.normalized_Ez
```

Examples/Physics_applications/spacecraft_charging/PICMI_inputs_rz.py
```diff
@@ -46,11 +47,11 @@ def correct_space_charge_fields(self, q=None):
...
-        phi = PhiFPWrapper(include_ghosts=True)[:,:]
+        phi = PhiFPWrapper(include_ghosts=True)
         phi[...]  += (q - q_v)*self.normalized_phi[...]
```
RemiLehe:
```suggestion
        phi[...]  += (q - q_v)*self.normalized_phi
```

dpgrote (approved):
This looks good.

I would comment for the record that using MFIter type loops would be better here (since it would avoid global copies and would run on GPU), if there is ever time to update this example.

RemiLehe:
Yes, agreed. 
I think I could do a follow-up PR in the next week or so, using `pyamrex` to do the above.

Examples/Physics_applications/spacecraft_charging/PICMI_inputs_rz.py
```diff
@@ -111,10 +112,7 @@ def compute_virtual_charge_on_spacecraft():
 
     # Compute integral of rho over volume of the domain
     # (i.e. total charge of the plasma particles)
-    rho_integral = 0.0
-    for k in range(1, nz-1):
-        for i in range(1, nr-1):
-            rho_integral += rho[i,k] * r[i] * dr * dz
+    rho_integral = (rho[1:nr-1,1:nz-1] * r[1:nr-1,np.new_axis]).sum()*dr*dz
```
RemiLehe:
```suggestion
    rho_integral = (rho[1:nr-1,1:nz-1] * r[1:nr-1,np.newaxis]).sum()*dr*dz
```

PR #4793, opened 2024-03-20
fix variable add in picmi lab particle diagnostics

RTSandberg:
This PR corrects an inadvertent indentation change in #4721 that broke PICMI lab particle diagnostics.


dpgrote (approved):
This looks good. Thanks for the fix!

PR #4798, opened 2024-03-22
Clean up BeamRelevant with TypeMultiplier

n01r:
The `BeamRelevant` reduced diagnostics had a lot of code where `amrex::ReduceOps` were constructed from lists of types. This PR cleans it up with the `TypeMultiplier` utility that makes it more legible.

Credit goes to @AlexanderSinn's work in https://github.com/Hi-PACE/hipace/pull/1052 and https://github.com/AMReX-Codes/amrex/pull/3718.

### DIFF
--- a/Source/Diagnostics/ReducedDiags/BeamRelevant.cpp
+++ b/Source/Diagnostics/ReducedDiags/BeamRelevant.cpp
@@ -200,7 +200,7 @@ void BeamRelevant::ComputeDiags (int step)
         using PType = typename WarpXParticleContainer::SuperParticleType;
 
         // number of reduction operations in first concurrent batch
-        const int num_red_ops_1 = 8;
+        constexpr size_t num_red_ops_1 = 8;
         TypeMultiplier<amrex::ReduceOps, ReduceOpSum[num_red_ops_1]> reduce_ops_1;
         using ReducedDataT1 = TypeMultiplier<amrex::ReduceData, ParticleReal[num_red_ops_1]>;
 
@@ -276,7 +276,7 @@ void BeamRelevant::ComputeDiags (int step)
         }
 
         // number of reduction operations in second concurrent batch
-        const int num_red_ops_2 = 11;
+        constexpr size_t num_red_ops_2 = 11;
 
         TypeMultiplier<amrex::ReduceOps, ReduceOpSum[num_red_ops_2]> reduce_ops2;
         using ReducedDataT2 = TypeMultiplier<amrex::ReduceData, ParticleReal[num_red_ops_2]>;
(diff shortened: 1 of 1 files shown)


Source/Diagnostics/ReducedDiags/BeamRelevant.cpp
```diff
@@ -198,14 +199,14 @@ void BeamRelevant::ComputeDiags (int step)
 
         using PType = typename WarpXParticleContainer::SuperParticleType;
 
-        amrex::ReduceOps<ReduceOpSum,ReduceOpSum,ReduceOpSum,ReduceOpSum,ReduceOpSum,
-        ReduceOpSum,ReduceOpSum,ReduceOpSum> reduce_ops;
-        auto r = amrex::ParticleReduce<amrex::ReduceData<ParticleReal,ParticleReal,
-        ParticleReal,ParticleReal,ParticleReal,ParticleReal,ParticleReal,ParticleReal>>(
+        // number of reduction operations in first concurrent batch
+        const int num_red_ops_1 = 8;
```
AlexanderSinn:
```suggestion
        constexpr std::size_t num_red_ops_1 = 8;
```
Looks great! I would write constexpr explicitly here (and in the other places including in ImpactX). Currently, constexpr is implied by the const; for the code to function constexpr is necessary.

n01r: Thanks, @AlexanderSinn!

ax3l (approved):
Very nice :)

And maybe as a follow-up, we can make this a one-time pass reduction :)

Source/Diagnostics/ReducedDiags/BeamRelevant.cpp
```diff
@@ -242,18 +243,17 @@ void BeamRelevant::ComputeDiags (int step)
...
+            reduce_ops_1);
+
+        std::vector<ParticleReal> values_per_rank_1st(num_red_ops_1);
+
+        /* contains in this order:
+         * w, x_mean, y_mean, z_mean
+         * ux_mean, uy_mean, uz_mean, gm_mean
+         */
+        amrex::constexpr_for<0, num_red_ops_1> ([&](auto i) {
+            values_per_rank_1st[i] = amrex::get<i>(r1);
```
ax3l, line 255: For some reason, @AlexanderSinn uses this syntax in HiPACE++:
```suggestion
            values_per_rank_1st[i.value] = amrex::get<i.value>(r1);
```

Source/Diagnostics/ReducedDiags/BeamRelevant.cpp
```diff
@@ -336,19 +335,18 @@ void BeamRelevant::ComputeDiags (int step)
...
+
+        /* contains in this order:
+         * x_ms, y_ms, z_ms
+         * ux_ms, uy_ms, uz_ms,
+         * gm_ms
+         * xux, yuy, zuz,
+         * charge
+         */
+        amrex::constexpr_for<0, num_red_ops_2> ([&](auto i) {
+            values_per_rank_2nd[i] = amrex::get<i>(r2);
```
ax3l, line 348: see above

Issue #4593, opened 2024-01-09
'out_of_range' error depending to OMP_NUM_THREADS


EyaDammak:
On the local computer, when I run some input codes (inputs_rz from the scraping test for examples, https://github.com/ECP-WarpX/WarpX/blob/development/Examples/Tests/scraping/inputs_rz), I have this error: [[error_message.txt](https://github.com/ECP-WarpX/WarpX/files/13867498/error_message.txt)] + this backtrace file: [Backtrace.0.1.txt](https://github.com/ECP-WarpX/WarpX/files/13867619/Backtrace.0.1.txt)

Here the commands that I did:
/cmake -S . -B build -DWarpX_DIMS="RZ" -DWarpX_EB=ON -DWarpX_MPI=OFF -DCMAKE_BUILD_TYPE=Debug
/cmake --build build -j 4

Nonetheless, when I impose `export OMP_NUM_THREADS=1`, the inputs script runs. With `export OMP_NUM_THREADS=2`, sometimes it runs and sometimes not. And for `OMP_NUM_THREADS' > 2, the input script doesn't run.

I have similar problems with python scripts: https://github.com/EyaDammak/WarpX/blob/spacecraft_charging_automated_test/Examples/Physics_applications/spacecraft_charging/PICMI_inputs_rz.py
Using these commands:
/cmake -S . -B build -DWarpX_DIMS="RZ;3" -DWarpX_EB=ON -DWarpX_PYTHON=ON
/cmake --build build --target pip_install -j 4

ax3l:
Interesting, this points to `ComputeDivB`:
https://github.com/ECP-WarpX/WarpX/blob/4ab88c5909ac2a0dd300d8ed111bfa04280b3e76/Source/WarpX.cpp#L2931

with `algo.maxwell_solver = none`...

ax3l:
It might be possible that you posted the `Backtrace.0.1.txt` from another run (a Python run)?

ax3l:
If it is a race condition #4799, then the backtrace hint to `ParticleBoundaryBuffer::gatherParticles(MultiParticleContainer&, amrex::Vector<amrex::MultiFab const*, std::allocator<amrex::MultiFab const*> > const&) [clone ._omp_fn.1] at /home/eyadammak/src/warpx/Source/Particles/ParticleBoundaryBuffer.cpp:326` might be valid...?

PR #4799, opened 2024-03-22
Add thread sanitizer test in CI

lucafedeli88:
This PR adds a thread sanitizer test in CI.

Specifically:

- a new `clang_sanitizers.yml` workflow file is created
- the pre-existing UB sanitizers test is moved from `ubuntu.yml` to `clang_sanitizers.yml` workflow
- a thread sanitizer test is added. This test needs `clang-17`. Therefore, for this test,  we use a `Ubuntu 23.10` container and we install `clang-17` with the instructions added to a new dependencies file (`clang17.sh`).
- the test consists in compiling WarpX with the thread sanitizer and in running the `laser_acceleration` example in 1D,2D,3D, and RZ, as well as the `embedded_circle` example in 2D. In all cases the serialization of  initial conditions is deactivated, although it does not seem to affect the thread sanitizer.  Note that WarpX is compiled in double precision, since the `embedded_circle` example is not  tested in single precision and seems to be numerically unstable unless double precision is used. Note also that WarpX is compiled without EB support for the 4  `laser_acceleration` tests because there seems to be bugs when WarpX with EB support is used to run  `laser_acceleration` in 2D (this is currently under investigation). 
-  A fix proposed by @WeiqunZhang in https://github.com/AMReX-Codes/amrex/pull/3951 is implemented to fix a race condition in `ParticleBoundaryBuffer.cpp`
- Following @RemiLehe 's suggestion, `address` and `pointer-compare` sanitizer checks have been added to UB sanitizer test.
- The PR changes where `current_correction` is set to `true` (the default behavior) to fix an issue found by the UB sanitizer.

The thread sanitizer test uses `libarcher`, which should theoretically avoid false positives with `OpenMP` (see https://github.com/llvm/llvm-project/tree/main/openmp/tools/archer)

Some additional comments:
- `export PMIX_MCA_gds=hash` I am not sure why this setting is required to be honest (I am still trying to understand). However, if I run without this setting I get an error message like: `The gds/shmem component attempted to attach to a shared-memory segment at a particular base address, but was given a different one`, followed by a suggestion to disable the `gds/shmem` component. It seems to be a relatively common issue.
- `export TSAN_OPTIONS='ignore_noninstrumented_modules=1'` should avoid false positives with libraries not compiled with the thread sanitizer. Archer documentation suggests using this setting.
- `export ARCHER_OPTIONS="verbose=1"` makes WarpX print the message:
`Archer detected OpenMP application with TSan, supplying OpenMP synchronization semantics`, which confirms that `libarcher` is in use.
- `export OMPI_ALLOW_RUN_AS_ROOT=1` and `export OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1` are needed because we are running the test in a docker container.
- `export "ASAN_OPTIONS=detect_leaks=0"` is required to disable leak checks: MPI implementations leak memory


.github/workflows/clang_sanitizers.yml
```diff
@@ -0,0 +1,149 @@
...
+        export OMPI_ALLOW_RUN_AS_ROOT=1
+        export OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1
+
+        export OMP_NUM_THREADS=2
+        mpirun -n 2 ./build/bin/warpx.rz Examples/Physics_applications/laser_acceleration/inputs_rz warpx.serialize_initial_conditions = 0
+        mpirun -n 2 ./build/bin/warpx.1d Examples/Physics_applications/laser_acceleration/inputs_1d warpx.serialize_initial_conditions = 0
+        mpirun -n 2 ./build/bin/warpx.2d Examples/Physics_applications/laser_acceleration/inputs_2d warpx.serialize_initial_conditions = 0
+        mpirun -n 2 ./build/bin/warpx.3d Examples/Physics_applications/laser_acceleration/inputs_3d warpx.serialize_initial_conditions = 0
+
+        git clone https://github.com/ECP-WarpX/warpx-data
```
lucafedeli88:
```suggestion
        git clone https://github.com/ECP-WarpX/warpx-data ../warpx-data
```
Fix position of `warpx-data`

ax3l:
Uh, is there a race condition using OpenMP threads in `ParticleBoundaryBuffer::gatherParticlesFromDomainBoundaries`?

RemiLehe:
This could be related to:
https://github.com/ECP-WarpX/WarpX/issues/4593

.github/workflows/clang_sanitizers.yml
```diff
@@ -0,0 +1,152 @@
...
+        cmake -S . -B build                 \
+          -GNinja                           \
+          -DCMAKE_VERBOSE_MAKEFILE=ON       \
+          -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+          -DWarpX_DIMS="RZ;1;2;3"           \
+          -DWarpX_FFT=ON                    \
+          -DWarpX_QED=ON                    \
+          -DWarpX_QED_TABLE_GEN=ON          \
+          -DWarpX_OPENPMD=ON                \
+          -WarpX_EB=ON                      \
```
RemiLehe: @lucafedeli88 Although this is set to `ON`, the CI test logs suggest that WarpX is being compiled with EB turned off. 
As a consequence, the `embedded_circle` test later fails.

.github/workflows/clang_sanitizers.yml
```diff
@@ -0,0 +1,152 @@
...
+        export CCACHE_COMPRESS=1
+        export CCACHE_COMPRESSLEVEL=10
+        export CCACHE_MAXSIZE=100M
+        ccache -z
+
+        export CXX=$(which clang++-17)
+        export CC=$(which clang-17)
+        export CXXFLAGS="-fsanitize=thread"
+
+        cmake -S . -B build                 \
```
RemiLehe, line 104:
```suggestion
        rm -rf build
        cmake -S . -B build                 \
```

.github/workflows/clang_sanitizers.yml
RemiLehe: This is maybe because the `build` folder was not cleared...

WeiqunZhang: Missing `D` in `-WarpX_EB`.

.github/workflows/clang_sanitizers.yml
```diff
@@ -0,0 +1,153 @@
...
+          -GNinja                           \
+          -DCMAKE_VERBOSE_MAKEFILE=ON       \
+          -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+          -DWarpX_DIMS="RZ;1;2;3"           \
+          -DWarpX_FFT=ON                    \
+          -DWarpX_QED=ON                    \
+          -DWarpX_QED_TABLE_GEN=ON          \
+          -DWarpX_OPENPMD=ON                \
+          -WarpX_EB=ON                      \
```
RemiLehe:
```suggestion
          -DWarpX_EB=ON                      \
```

.github/workflows/clang_sanitizers.yml
```diff
@@ -0,0 +1,153 @@
...
+        export CCACHE_COMPRESS=1
+        export CCACHE_COMPRESSLEVEL=10
+        export CCACHE_MAXSIZE=100M
+        ccache -z
+
+        export CXX=$(which clang++-17)
+        export CC=$(which clang-17)
+        export CXXFLAGS="-fsanitize=thread"
+
+        rm -rf build
```
RemiLehe:
```suggestion
```

.github/workflows/clang_sanitizers.yml
```diff
@@ -0,0 +1,152 @@
...
+    - name: build WarpX
+      run: |
+        export CCACHE_COMPRESS=1
+        export CCACHE_COMPRESSLEVEL=10
+        export CCACHE_MAXSIZE=100M
+        ccache -z
+
+        export CXX=$(which clang++-15)
+        export CC=$(which clang-15)
+        export CXXFLAGS="-fsanitize=undefined -fno-sanitize-recover=all"
```
RemiLehe: Should we add more options, like `-fsanitize=address`, `-fsanitize=builtin`, `-fsanitize=pointer-compare`?

.github/workflows/clang_sanitizers.yml
lucafedeli88: Thanks for spotting this! I've reverted the `warpx.self_fields_required_precision` change

.github/workflows/clang_sanitizers.yml
lucafedeli88: So, `-fsanitize=undefined` should already include  `fsanitize=builtin', but we can certainly add the other two

lucafedeli88: Done

lucafedeli88: I had to disable leak checks though, because MPI implementations typically leak some memory on purpose.

lucafedeli88:
Ping, @ax3l  and @RemiLehe ;-) !

ax3l (approved):
Great work, thank you! :)

We can run this over more of the fast tests once CTest is merged #5068

PR #4801, opened 2024-03-22
Fix a number of compilation warnings

roelof-groenewald:
Fixes warnings:

- WarpX/Source/EmbeddedBoundary/ParticleScraper.H(192): warning 550-D: variable "nodal" was set but never used
- WarpX/Source/Diagnostics/ComputeDiagFunctors/JdispFunctor.cpp:33:35: warning: ‘m_mf_j_external’ may be used uninitialized in this function [-Wmaybe-uninitialized]

And single precision build "error" (error on my Mac but only a warning on CI builds I guess)
- WarpX/Source/Particles/PhysicalParticleContainer.cpp:151:31: error: non-constant-expression cannot be narrowed from type 'amrex::Real' (aka 'double') to 'amrex::ParticleReal' (aka 'float') in initializer list [-Wc++11-narrowing]

### DIFF
--- a/Source/Diagnostics/ComputeDiagFunctors/JdispFunctor.cpp
+++ b/Source/Diagnostics/ComputeDiagFunctors/JdispFunctor.cpp
@@ -28,15 +28,15 @@ JdispFunctor::operator() (amrex::MultiFab& mf_dst, int dcomp, const int /*i_buff
     auto* hybrid_pic_model = warpx.get_pointer_HybridPICModel();
 
     /** pointer to total simulation current (J) multifab */
-    amrex::MultiFab* m_mf_j = warpx.get_pointer_current_fp(m_lev, m_dir);
-    amrex::MultiFab* m_mf_curlB;
+    amrex::MultiFab* mf_j = warpx.get_pointer_current_fp(m_lev, m_dir);
+    amrex::MultiFab* mf_curlB;
     if (hybrid_pic_model) {
         /** pointer to current calculated from Ampere's Law (Jamp) multifab */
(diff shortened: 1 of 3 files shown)


Source/Diagnostics/ComputeDiagFunctors/JdispFunctor.cpp
```diff
@@ -59,8 +56,11 @@ JdispFunctor::operator() (amrex::MultiFab& mf_dst, int dcomp, const int /*i_buff
         -1, *m_mf_j, 0, 0, 1, Jdisp.nGrowVect()
     );
 
-   // Subtract the interpolated j_external value from j_displacement.
     if (hybrid_pic_model) {
+        // Subtract the interpolated j_external value from j_displacement.
+        /** pointer to external currents (Jext) multifab */
+        amrex::MultiFab* m_mf_j_external = hybrid_pic_model->get_pointer_current_fp_external(m_lev, m_dir);
```
lucafedeli88: Thanks for this PR, @roelof-groenewald ! Maybe we could change the name of this variable to `mf_j_external`, since it is not a member variable.

lucafedeli88 (changes requested):
Thanks for this PR, @roelof-groenewald

Source/Diagnostics/ComputeDiagFunctors/JdispFunctor.cpp
roelof-groenewald: Good point. I'll change that now. Thanks!

lucafedeli88 (approved):
Thanks!

PR #4806, opened 2024-03-27
Add clang-tidy workflow

lucafedeli88:
This PR adds a script (and a corresponding entry in the documentation) to run `clang-tidy` locally.


Tools/Linter/runClangTidy.sh
```diff
@@ -0,0 +1,157 @@
...
+# Install LAPACK++
+echo
+echo "Installing LAPACK++"
+echo
+if [ -d ${REPO_DIR}/src/lapackpp ]
+then
+  cd ${REPO_DIR}/src/lapackpp
+  git fetch --prune
+  git checkout master
+  git pull
```
ax3l: I was thinking users would run this script in their current developer environment, e.g., https://warpx.readthedocs.io/en/latest/install/dependencies.html#conda-linux-macos-windows, and do not install dependencies again.

Tools/Linter/runClangTidy.sh
```diff
@@ -0,0 +1,157 @@
...
+  git clone https://github.com/icl-utk-edu/lapackpp.git ${REPO_DIR}/src/lapackpp
+fi
+rm -rf ${REPO_DIR}/src/lapackpp-build
+cmake -S ${REPO_DIR}/src/lapackpp -B ${REPO_DIR}/src/lapackpp-build -DCMAKE_CXX_STANDARD=17 -Dbuild_tests=OFF -DCMAKE_INSTALL_RPATH_USE_LINK_PATH=ON -DCMAKE_INSTALL_PREFIX=${REPO_DIR}/lapackpp-install
+cmake --build ${REPO_DIR}/src/lapackpp-build --target install --parallel ${PARALLEL}
+rm -rf ${REPO_DIR}/src/lapackpp-build
+export CMAKE_PREFIX_PATH=${REPO_DIR}/lapackpp-install:${CMAKE_PREFIX_PATH}
+echo "_____________________________________________"
+
+# Prepare clang-tidy wrapper
```
ax3l, line 72: This and below would be the main thing we need to document then:

Tools/Linter/runClangTidy.sh
lucafedeli88: You are right. I have removed it. Both the script and the documentation now recommends to ensure that all the dependencies of WarpX are met.

Tools/Linter/runClangTidy.sh
lucafedeli88: I've added a few more details in the documentation.

ax3l (approved):
Thank you very much, that is great! :)

PR #4808, opened 2024-03-28
Update profile for Fugaku supercomputer

lucafedeli88:
This mini PR is an update of the profile file to use WarpX on the Fugaku supercomputer


ax3l (approved):
Thank you! :)

PR #4810, opened 2024-03-29
Doc: LUMI (CSC) Update

ax3l:
Update the documentation for LUMI (CSC).

Follows some breakage and updates I saw when documenting ImpactX based on this:
https://impactx.readthedocs.io/en/latest/install/hpc/lumi.html

- [x] finalize updates
- [x] test

### DIFF
--- a/Tools/machines/lumi-csc/lumi_warpx.profile.example
+++ b/Tools/machines/lumi-csc/lumi_warpx.profile.example
@@ -1,5 +1,5 @@
 # please set your project account
-#export proj=<yourProject>
+#export proj="project_..."
 
 # required dependencies
 module load LUMI/23.09  partition/G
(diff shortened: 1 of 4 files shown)


Tools/machines/lumi-csc/lumi_warpx.profile.example
```diff
@@ -50,4 +57,4 @@ export CXX=$(which amdclang++)
 export FC=$(which amdflang)
 export CFLAGS="-I${ROCM_PATH}/include"
 export CXXFLAGS="-I${ROCM_PATH}/include -Wno-pass-failed"
-export LDFLAGS="-L${ROCM_PATH}/lib -lamdhip64"
+export LDFLAGS="-L${ROCM_PATH}/lib -lamdhip64 ${PE_MPICH_GTL_DIR_amd_gfx90a} -lmpi_gtl_hsa"
```
lucafedeli88, line 60: Do they ask to add `-lmpi_gtl_hsa` in the LUMI documentation or did you find out that leaving this out leads to issues?

lucafedeli88 (commented):
Thanks for this PR, @ax3l!
I left just a minor comment

Tools/machines/lumi-csc/lumi_warpx.profile.example
ax3l: I did not yet test this with WarpX, but I had issues starting ImpactX without it when GPU-aware MPI is enabled.

The flags are from my earlier scripts on Spock/Crusher/Frontier and are usually needed when _not_ using `CC` directly (as we do here with `amdclang`), usually with an old CMake before my [CC patches landed](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/6264).

LUMI has `cmake version 3.27.7`, but I still saw the issues when using `amdclang` instead of Cray's `CC` wrappers. I did not dig deeper yesterday and just fixed it with the extra flags.

ax3l:
@lucafedeli88 just checking: did you write somewhere that you tested this and it worked?
Otherwise I can do a comprehensive test again today :)

lucafedeli88:
Hi @ax3l ,

I am very sorry for my late review. 
I've tested your script on LUMI and I've found an issue, which is most likely **not** related to your PR.

After having compiled WarpX on LUMI, if I do, for instance, `ldd warpx.3d | grep "not found"` I get :
```
	libomp.so => not found
``` 
I think that this comes from building WarpX with lookup-tables generation enabled. 
Indeed `PICSAR.cmake` contains this logic:
```cmake
        # Enable or disable QED lookup tables generation
        if(WarpX_QED_TABLE_GEN OR WarpX_QED_TOOLS)
            set(PXRMP_QED_TABLEGEN ON CACHE INTERNAL "")
        else()
            set(PXRMP_QED_TABLEGEN OFF CACHE INTERNAL "")
        endif()

        # Enable or disable OpenMP for lookup tables generation
        if(PXRMP_QED_TABLEGEN)
            if(WarpX_QED_TABLES_GEN_OMP STREQUAL AUTO)
                find_package(OpenMP REQUIRED CXX)
                if(OpenMP_CXX_FOUND)
                    set(PXRMP_QED_OMP ON CACHE INTERNAL "")
                else()
                    set(PXRMP_QED_OMP OFF CACHE INTERNAL "")
                endif()
            elseif(WarpX_QED_TABLES_GEN_OMP STREQUAL ON)
                    set(PXRMP_QED_OMP ON CACHE INTERNAL "")
            else()
                    set(PXRMP_QED_OMP OFF CACHE INTERNAL "")
            endif()
        else()
            set(PXRMP_QED_OMP OFF CACHE INTERNAL "")
        endif()
``` 

A quick workaround would be to set `WarpX_QED_TABLES_GEN_OMP=OFF`.  However, we may want to reconsider the logic above entirely.

ax3l:
Thanks! Does the logic above mix up a bit the `find_packages`?
I would write
```cmake
        # Enable or disable OpenMP for lookup tables generation
        if(PXRMP_QED_TABLEGEN)
            if(WarpX_QED_TABLES_GEN_OMP STREQUAL AUTO)
                find_package(OpenMP COMPONENTS CXX)
                if(OpenMP_CXX_FOUND)
                    set(PXRMP_QED_OMP ON CACHE INTERNAL "")
                else()
                    set(PXRMP_QED_OMP OFF CACHE INTERNAL "")
                endif()
            elseif(WarpX_QED_TABLES_GEN_OMP STREQUAL ON)
                find_package(OpenMP REQUIRED COMPONENTS CXX)
                set(PXRMP_QED_OMP ON CACHE INTERNAL "")
            else()
                set(PXRMP_QED_OMP OFF CACHE INTERNAL "")
            endif()
        else()
            set(PXRMP_QED_OMP OFF CACHE INTERNAL "")
        endif()
```

lucafedeli88:
> Thanks! Does the logic above mix up a bit the `find_packages`? I would write
> 
> ```cmake
>         # Enable or disable OpenMP for lookup tables generation
>         if(PXRMP_QED_TABLEGEN)
>             if(WarpX_QED_TABLES_GEN_OMP STREQUAL AUTO)
>                 find_package(OpenMP COMPONENTS CXX)
>                 if(OpenMP_CXX_FOUND)
>                     set(PXRMP_QED_OMP ON CACHE INTERNAL "")
>                 else()
>                     set(PXRMP_QED_OMP OFF CACHE INTERNAL "")
>                 endif()
>             elseif(WarpX_QED_TABLES_GEN_OMP STREQUAL ON)
>                 find_package(OpenMP REQUIRED COMPONENTS CXX)
>                 set(PXRMP_QED_OMP ON CACHE INTERNAL "")
>             else()
>                 set(PXRMP_QED_OMP OFF CACHE INTERNAL "")
>             endif()
>         else()
>             set(PXRMP_QED_OMP OFF CACHE INTERNAL "")
>         endif()
> ```

Unfortunately, this rewriting results in the same error.
Maybe, for the moment, we can switch `WarpX_QED_TABLES_GEN_OMP` OFF. However, I don't understand why before I didn't see any issue

lucafedeli88 (approved):
It works now!

PR #4263, opened 2023-09-01
[WIP] Implement PSATD with arbitrary temporal dependency of `J` and `rho`.

oshapoval:
This PR replaces #4261.
It also closes #3506.

In this PR we implement second-order PSATD solver with arbitrary temporal dependencies of densities J and rho.
Both J and rho can be either constant or linear or quadratic. Now, fields update equaitons are implemented in `/WarpX/Source/FieldSolver/SpectralSolver/SpectralAlgorithms/PsatdAlgorithmJRm.cpp` for any PSATD solver except of Galilean PSATD.

To do:

- [x] Debug CI tests that fail (e.g., standard PSATD): issue seems related to the option `psatd.update_with_rho`
- [ ] Changes from #3506 need to be ported here, or viceversa
- [x] Rename coefficient `X2`, since it does not match the coefficient `X2` of the standard PSATD equations
- [x] move Galilean PSATD equations from `PsatdAlgorithmJConstantInTime` to a separate class, as in #3506
- [x] add CI tests for PSATD-QQ1(nodal), PSATD-LQ1(nodal), PSATD-CC2(nodal), PASTD-LL1(nodal) and PASTD-LL1(staggered).

@EZoni Equations for E,B,F,G and averaged E_avg, B_avg fields and corresponding coefficients are given in the paper draft: Eq.(8)-(9) and Eqs.(11)-(12).

PR #4815, opened 2024-04-01
Printing out the time dependencies of the charge and current densitie…

oshapoval:
This PR prints out the information about the time dependencies of the charge and current densities over one time step when used with multi-J PSATD: `warpx.do_multi_J = 1`. It's especially will be useful once [PR 4263](https://github.com/ECP-WarpX/WarpX/pull/4263) is merged, where we have arbitrary time dependencies for J and rho (up to the second order).


lucafedeli88 (approved):
Thanks!

PR #4820, opened 2024-04-03
Implement grid based merging algorithm

roelof-groenewald:
A new resampling algorithm (really a merging algorithm) is added in this PR. The approach is as follows:

- Particles are ordered in their spatial cells (using `ParticleUtils::findParticlesInEachCell`)
- loop over cells (GPU loop)
    - index each particle based on its momentum cell (momentum divisions `delta_ur`, `n_theta` and `n_phi` are given as inputs)
    - perform `argsort` on the momentum cell labeling so that particle indices for particles in the same momentum cell are next to each other
    - loop over particles
        - check if a transition to a new momentum cell has occurred
        - if yes, merge particles if more than 2 particles are present in the same momentum cell. Merging is done into 2 particles  to conserve momentum and energy.
        - accumulate momentum cell quantities (momentum, energy, etc) 
 
I tested the implementation on a simple 1d case where I set `do_not_push=true` and only loaded an ion species. The simulation was run for 3 steps with merging occurring on the first step. About 500k particles were removed during the merging (~80% of the total). The plot below shows density and current before (blue) and after (orange) merging. The test used particle shape 1. As expected, the charge density did not change at all (the merged particles are placed in the cell so as to preserve deposited charge density), and the current changed only minimally (consequence of energy conserving merge).

image

Using the particle energy reduced diagnostic, the change in species kinetic energy due to merging was as ~~4.775e-06%~~ 1.759e-12% (updated after including relativistic corrections).


Source/Particles/Resampling/GridBasedMerging.H
```diff
@@ -0,0 +1,128 @@
...
+     */
+    void operator() (WarpXParIter& pti, int lev, WarpXParticleContainer* pc) const final;
+
+    /**
+     * \brief This merging routine requires functionality to sort a GPU vector
+     * based on another GPU vector. The heap-sort functions below where obtained
+     * from https://www.geeksforgeeks.org/heap-sort/ and modified for the
+     * current purpose.
+    */
+    struct HeapSort {
```
roelof-groenewald: It'd be cool if this functionality is already available in AMReX. But I'm not aware if it is.

Source/Particles/Resampling/GridBasedMerging.cpp
```diff
@@ -0,0 +1,303 @@
...
+#endif
+#if defined(WARPX_DIM_3D)
+                cluster_y += w[part_idx]*y[part_idx];
+#endif
+                cluster_z += w[part_idx]*z[part_idx];
+                cluster_ux += w[part_idx]*ux[part_idx];
+                cluster_uy += w[part_idx]*uy[part_idx];
+                cluster_uz += w[part_idx]*uz[part_idx];
+                total_weight += w[part_idx];
+                total_energy += 0.5_prt * mass * w[part_idx] * (
```
dpgrote: Does this algorithm only work with non-relativistic energy? If not, you can use the routine `Algorithms::KineticEnergy`.

Docs/source/usage/parameters.rst
```diff
@@ -1269,25 +1269,38 @@ Particle initialization
...
+
+        * ``<species>.resampling_algorithm_n_theta`` (`int`) optional (default `120`)
+            The number of cell divisions to use in the :math:`\theta` direction
+            when clustering the particle velocities.
+
+        * ``<species>.resampling_algorithm_n_phi`` (`int`) optional (default `60`)
+            The number of cell divisions to use in the :math:`\phi` direction
+            when clustering the particle velocities.
+
+* ``<species>.resampling_min_ppc`` (`int`) optional (default `1`)
```
dpgrote, line 1302: This name change is breaking the API. Is the name change needed?

dpgrote:
This looks like a nice idea! I have two comments (in addition to my inline comments).
I would prefer a less generic name than `GridBasedMerging`. How about something like `VelocityCoincidenceThinning`?

Since this is binning in spherical velocity space, it will have the bias that particles with larger velocities will be merged with particles with larger velocity differences. Will that matter? Also, if there is any sort of beaming of the particles, they might be aggressively merged since the velocities will be in a narrow angular region.

Source/Particles/Resampling/GridBasedMerging.cpp
roelof-groenewald: The algorithm could definitely work for relativistic velocities as well, but the current implementation does not account for such corrections. I will switch the energy calculation to use `Algorithms::KineticEnergy` since that calculation will be correct for non-relativistic particles anyway, and it won't produce any more problems in the later parts where relativistic corrections should be added (I think just in the calculation of the product species' velocity magnitude).

Docs/source/usage/parameters.rst
roelof-groenewald: I will bring this up at the Tuesday meeting for general discussion. I think it is worth breaking the API for this since `resampling_min_ppc` is a general parameter for resampling, not an algorithmic specific option.

roelof-groenewald:
Thanks for the review @dpgrote. I will change the name to `VelocityCoincidenceThinning` as you requested.

The reason I opted for spherical binning in velocity space is because it avoids the need to know maximum and minimum velocity values - `theta` and `phi` are naturally bounded and the number of velocity cells in `|u|` never enters the calculation (the maximum cell id is just unbounded). The issue of allowing larger velocity differences for faster particles might a problem (testing in real applications is needed to find out). On the other hand, the spherical binning means that merging will happen for 2 particles with some max value of |u1 - u2|/|u1| - basically allowing a difference in velocity as a fraction of the involved particles' velocities.
In regards to aggressively binning beam particles, I don't think this is a problem since the `resampling_min_ppc` parameter is meant to limit merging. The user presumably will have a required number of beam particles per cell and can simply set `resampling_min_ppc` equal to that number. If there is no source of new beam particles there will be no actual merging happening.

roelof-groenewald:
As pointed out by @anecas1979, another way we can limit the merging intensity is to always only merge 3 particles into 2 i.e. in contrast to the current approach where all the particles in a phase space cell are merged into 2, we can do successive merging in each cell to reduce the number of particles by ~33% (within the phase space cell).

dpgrote:
Good point about the spherical not needing the velocity maxima. Though, since this is using `int` for the indexing, that puts a limit of ~1600 cells in each direction (cube root of 2**32), giving a practical range of `+-800*du`. You could rely on this to do the indexing in Cartesian velocity space.

Source/Particles/Resampling/VelocityCoincidenceThinning.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+                        // perform merging of momentum bin particles
+                        auto u_perp2 = cluster_ux*cluster_ux + cluster_uy*cluster_uy;
+                        auto u_perp = std::sqrt(u_perp2);
+                        auto cluster_u_mag2 = u_perp2 + cluster_uz*cluster_uz;
+                        auto cluster_u_mag = std::sqrt(cluster_u_mag2);
+
+                        // calculate required velocity magnitude to achieve
+                        // energy conservation
+                        auto v_mag2 = 2._prt * total_energy / (total_weight * mass);
```
roelof-groenewald: Include relativistic corrections in energy calculation for the product particles:
```suggestion
                        auto v_mag2 = total_energy / total_weight * (
                            (total_energy / total_weight + 2._prt * mass * c2 )
                            / (mass * mass * c2)
                        );
```

Source/Particles/Resampling/GridBasedMerging.cpp
roelof-groenewald: I now also updated the calculation for the product species' velocity magnitude to include relativistic effects. I was surprised to see this actually made a substantial difference in energy conservation even though I'm testing with 100 eV hydrogen ions.

roelof-groenewald:
@dpgrote testing of this approach on our use case has shown positive results - I get macroparticle inventory stabilization with merging parameters that do not seem to affect the physics. I would still like to test with a Cartesian momentum grid rather than the spherical one since it could provide another confirmation that the merging details do not impact the physics. But what do you think about merging this as is and then I'll open a follow-up PR to add the Cartesian momentum grid option once I implement it?

dpgrote:
Merging as is sounds good.
I would be interested to see test cases that include a beam (perhaps an LPA example) and/or a case with a drifting plasma.

Source/Particles/Resampling/VelocityCoincidenceThinning.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+    utils::parser::queryWithParser(
+        pp_species_name, "resampling_min_ppc", m_min_ppc
+    );
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+        m_min_ppc >= 1,
+        "Resampling min_ppc should be greater than or equal to 1"
+    );
+
+    utils::parser::queryWithParser(
```
dpgrote: These parameters will be very much application dependent. Perhaps make these required inputs to avoid surprises for users. This also forces the user to have some consideration of what is happening. The documentation can give guidelines for what values to use.

roelof-groenewald: Sounds good! I agree, default values here could be problematic. I left the default `min_ppc=1` since that is also used by `LevelingThinning`.

Source/Particles/Resampling/VelocityCoincidenceThinning.H
```diff
@@ -0,0 +1,115 @@
+/* Copyright 2024 The WarpX Community
+ *
+ * This file is part of WarpX.
+ *
+ * Authors: Roelof Groenewald (TAE Technologies)
+ *
+ * License: BSD-3-Clause-LBNL
+ */
+#ifndef WARPX_GRID_BASED_MERGING_H_
```
dpgrote: This could be updated to match the file name.

Source/Particles/Resampling/VelocityCoincidenceThinning.cpp
```diff
@@ -0,0 +1,284 @@
...
+    const auto n_cells = static_cast<int>(bins.numBins());
+    auto *const indices = bins.permutationPtr();
+    auto *const cell_offsets = bins.offsetsPtr();
+
+    const auto min_ppc = m_min_ppc;
+
+    const auto mass = pc->getMass();
+
+    // create a GPU vector to hold the momentum cluster index for each particle
+    amrex::Gpu::DeviceVector<int> momentum_bin_number;
```
dpgrote: This could be simplified
```suggestion
    amrex::Gpu::DeviceVector<int> momentum_bin_number(bins.numItems());
```

Source/Particles/Resampling/VelocityCoincidenceThinning.cpp
```diff
@@ -0,0 +1,284 @@
...
+    sorted_indices.resize(bins.numItems());
+    auto* sorted_indices_data = sorted_indices.data();
+
+    const auto Ntheta = m_ntheta;
+    const auto Nphi = m_nphi;
+
+    auto dr = m_delta_ur;
+    auto dtheta = 2.0_prt * MathConst::pi / Ntheta;
+    auto dphi = MathConst::pi / Nphi;
+    auto c2 = PhysConst::c * PhysConst::c;
```
dpgrote: This could be a `constexpr`, and the ones above `const`.

Source/Particles/Resampling/VelocityCoincidenceThinning.cpp
```diff
@@ -0,0 +1,284 @@
...
+#endif
+#if defined(WARPX_DIM_3D)
+            amrex::ParticleReal cluster_y = 0._prt;
+#endif
+            amrex::ParticleReal cluster_z = 0._prt;
+            amrex::ParticleReal cluster_ux = 0._prt, cluster_uy = 0._prt, cluster_uz = 0._prt;
+
+            // Finally, loop through the particles in the cell and merge
+            // ones in the same momentum bin
+            for (int i = cell_start; i < cell_stop-1; ++i)
```
dpgrote: This loop is skipping the last particle in each cell. It looks like that last particle would never get merged, even if it is in the same velocity bin as the previous particles.
If you switch the checks below, this loop can include the last particle because the `if` check would be short circuited and `sorted_indices_data[i + 1]]` won't be evaluated.
```
if ((i == cell_stop - 1) ||
                    (momentum_bin_number_data[sorted_indices_data[i]] != momentum_bin_number_data[sorted_indices_data[i + 1]]))
```

Source/Particles/Resampling/VelocityCoincidenceThinning.cpp
```diff
@@ -0,0 +1,284 @@
...
+                        auto vy = v_perp * std::sin(phi);
+
+                        // calculate rotation angles to parallel coord. frame
+                        auto cos_theta = cluster_uz / cluster_u_mag;
+                        auto sin_theta = u_perp / cluster_u_mag;
+                        auto cos_phi = cluster_ux / u_perp;
+                        auto sin_phi = cluster_uy / u_perp;
+
+                        // handle edge case with stationary clusters or particles
+                        if (std::abs(u_perp) <= std::numeric_limits<amrex::ParticleReal>::min()) {
```
dpgrote: Is there a reason to use the numeric limit min rather than the more simple `std::abs(u_perp) == 0.`?

Source/Particles/Resampling/VelocityCoincidenceThinning.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+                        // choose random angle for new velocity vector
+                        auto phi = amrex::Random(engine) * MathConst::pi;
+
+                        // set new velocity components based on chosen phi
+                        auto vx = v_perp * std::cos(phi);
+                        auto vy = v_perp * std::sin(phi);
+
+                        // calculate rotation angles to parallel coord. frame
+                        auto cos_theta = cluster_uz / cluster_u_mag;
```
dpgrote: The divides should be protected in case the denominators are zero.

Source/Particles/Resampling/VelocityCoincidenceThinning.cpp
roelof-groenewald: Good point! Thanks for the suggestion.

Source/Particles/Resampling/VelocityCoincidenceThinning.cpp
roelof-groenewald: Just to make the protection fully capture the numerical resolution. Basically, checking whether `u_perp` is closer to zero than the numerically smallest value of `ParticleReal`.

Source/Particles/Resampling/VelocityCoincidenceThinning.cpp
roelof-groenewald: The logic in the next few lines provide that protection. Basically instead of doing:
```
ParticleReal cos_theta;
if cluster_u_mag != 0 {
    cos_theta = cluster_uz / cluster_u_mag;
} else {
    cos_theta = 0;
}
```
I allow cos_theta to be initialized as a `nan` but then correct it if needed.

dpgrote: Ok. Though this nan would be caught if the user sets `amrex.fpe_trap_invalid = 1`. It is better to avoid nans if possible.

dpgrote: This is Ok, but it looks like `u_perp` will be of type `ParticleReal`, so checking for `== 0.` should be sufficient.

Source/Particles/Resampling/VelocityCoincidenceThinning.cpp
dpgrote: This is Ok, but it looks like `u_perp` will be of type `ParticleReal`, so checking for `== 0.` should be sufficient.

roelof-groenewald: I avoided the issue of comparing floats for equality by changing the logic to check if `u_perp > 0`.

Source/Particles/Resampling/VelocityCoincidenceThinning.cpp
roelof-groenewald: This should be taken care of now with the new assignment strategy.

Source/Particles/Resampling/VelocityCoincidenceThinning.cpp
```diff
@@ -0,0 +1,272 @@
...
+                        auto u_perp2 = cluster_ux*cluster_ux + cluster_uy*cluster_uy;
+                        auto u_perp = std::sqrt(u_perp2);
+                        auto cluster_u_mag2 = u_perp2 + cluster_uz*cluster_uz;
+                        auto cluster_u_mag = std::sqrt(cluster_u_mag2);
+
+                        // calculate required velocity magnitude to achieve
+                        // energy conservation
+                        auto v_mag2 = total_energy / total_weight * (
+                            (total_energy / total_weight + 2._prt * mass * c2 )
+                            / (mass * mass * c2)
```
dpgrote, line 202: What happens here with photons where `mass = 0`?

Source/Particles/Resampling/VelocityCoincidenceThinning.cpp
```diff
@@ -0,0 +1,272 @@
...
+            if (cell_numparts < min_ppc) {
+                return;
+            }
+
+            // Loop over particles and label them with the appropriate momentum bin
+            // number. Also assign initial ordering to the sorted_indices array.
+            for (int i = cell_start; i < cell_stop; ++i)
+            {
+                // get polar components of the velocity vector
+                auto u_mag = std::sqrt(
```
dpgrote, line 120: This is ok and doesn't need to the changed, but elsewhere in WarpX, `auto` is typically only used when the type is explicitly specified in the RHS (or would be a long type definition).
Also, this and other variables here could be declared `const`.

Source/Particles/Resampling/VelocityCoincidenceThinning.cpp
roelof-groenewald: Yeah, this is not set up to handle photons. The paper that Remi mentioned before (https://www.sciencedirect.com/science/article/abs/pii/S0010465515000405) discusses the proper handling of photons but we can leave that to a future PR.

dpgrote: Ok. Until then, can an assert be added to check for that?

Docs/source/usage/parameters.rst
dpgrote: This is the last issue - did anyone one else have an opinion? @axel ? @RemiLehe ?

ax3l: We have the `BackwardCompatibility` method - @roelof-groenewald can you check if the old parameter is set and if yes, error out with a clear migration guide.

(I do not think it is used a lot, thus we can make it simple.)

dpgrote (approved):
Looks good now!
Though presumably you'll need to merge in development to get past the CI failures.

PR #4823, opened 2024-04-04
Filter `rho`, when using relativistic electrostatic solver

### LINKED ISSUE #4705: incorrect residual net field after self-field initialization
Hi,

I encountered a problem that an unphysical net field is generated after the self-field initialization. By net field I mean **E**+c***B**, or the net force exerted on relativistic particle. The following is the simulation in folder WarpX\Examples\Tests\space_charge_initialization, and I let simulation run longer to test the problem I found.

I found after the z-propagating beam initializes its field, there is unphysical field staying and frozen in its initial position z=0. The left picture is the net transverse field, Ex-c*By. The right plot is just Ex, it shows the beam is too far to influence z=0(yet we can also notice there is some unphysical Ex frozen at z=0).
examp

Looking at this residual unphysical net field, its strength is 0.8, big enough to affect simulation considering the space charge E field is 5. Especially for a long beam, this field can cause unphysical trifurcation issue like this density plot shown below: all the density near the edge are pushed outward
image

I did a few more tests, and I noticed this frozen field is associated with the current deposition method I chose. When I change from default "esirkepov" to "direct" with setting "do_dive_cleaning=1" , the residual frozen field issue is gone. 

So I want to raise some concern about the problem above and also ask a question: Is "direct" current deposition and "do_dive_cleaning=1" compatible with FDTD simulation? 

The input deck I used is also attached below. 
[inputs_3d.zip](https://github.com/ECP-WarpX/WarpX/files/14336254/inputs_3d.zip)

RemiLehe:
When using the relativistic electrostatic solver, (`warpx.do_electrostatic = relativistic`, or when initializing the space charge fields of a species at the beginning of an electromagnetic simulation), `rho` is deposited on the grid for each species, but never filtered.

This PR fixes the issue by calling `SyncRho` (which performs filtering under the hood). 
The corresponding code (in `AddSpaceChargeField`) is now much more similar to the code in `AddSpaceChargeFieldLabFrame` (which is used for `warpx.do_electrostatic = labframe`), since `AddSpaceChargeFieldLabFrame` already uses `SyncRho`: https://github.com/ECP-WarpX/WarpX/blob/development/Source/FieldSolver/ElectrostaticSolver.cpp#L229

The main difference is that, in `AddSpaceChargeFieldLabFrame`, we use the `WarpX` member variables `rho_fp` and `rho_cp`, while in `AddSpaceChargeField` we cannot assume that these arrays are allocated. This is because `AddSpaceChargeField` could be called at the beginning of an electromagnetic simulation (to initialized the space charge field at t=0), for which `rho_fp`/`rho_cp` and not used and therefore not allocated.

Changes to the checksums are due to the fact that filtering is now taken into account.
In some cases, I had to increase the resolution of some of the tests in order, since filtering tends to smooth the distribution of charge for low-resolution and was therefore failing in the physics tests.

Close #4705


oshapoval:
Thanks for this PR @RemiLehe. It looks great on relativistic space charge initialization test case.
image

Examples/Tests/electrostatic_sphere/inputs_3d
```diff
@@ -10,6 +10,7 @@ boundary.field_lo = pec pec pec
 boundary.field_hi = pec pec pec
 warpx.const_dt = 1e-6
 warpx.do_electrostatic = relativistic
+warpx.use_filter = 0
```
RemiLehe:
```suggestion
```

Examples/Tests/relativistic_space_charge_initialization/inputs_3d
```diff
@@ -10,6 +10,7 @@ geometry.prob_lo     = -50.e-6   -50.e-6    -50.e-6
 geometry.prob_hi     =  50.e-6    50.e-6     50.e-6
 
 warpx.cfl = 1.e-5
+warpx.use_filter = 0
```
RemiLehe:
```suggestion
```

Examples/Tests/space_charge_initialization/inputs_3d
```diff
@@ -10,6 +10,7 @@ geometry.prob_lo     = -50.e-6   -50.e-6    -50.e-6
 geometry.prob_hi     =  50.e-6    50.e-6     50.e-6
 
 warpx.cfl = 1.e-3
+warpx.use_filter = 0
```
RemiLehe:
```suggestion
```

Examples/Physics_applications/beam-beam_collision/inputs
```diff
@@ -37,6 +37,7 @@ my_constants.dt = sigmaz/clight/10.
 # DIAGS
 my_constants.every_red = 1.
 warpx.used_inputs_file = warpx_used_inputs.txt
+warpx.use_filter = 0
```
RemiLehe:
```suggestion
```

Examples/Tests/space_charge_initialization/inputs_3d
```diff
@@ -1,5 +1,5 @@
 max_step = 1
-amr.n_cell =  64 64 64
+amr.n_cell =  128 128 128
```
RemiLehe, line 2: The resolution had to be increased in order to mitigate the effects of filtering.
(The previous resolution was very low, with the beam being only covered with ~2-3 cells.)

Examples/Tests/openbc_poisson_solver/inputs_3d
```diff
@@ -8,17 +8,17 @@ amr.n_cell = 128 128 128
 amr.max_level = 0
 amr.max_grid_size = 128
 geometry.dims = 3
-geometry.prob_lo = -5*sigmax -5*sigmay -5*sigmaz
-geometry.prob_hi = +5*sigmax +5*sigmay +5*sigmaz
+geometry.prob_lo = -4*sigmax -4*sigmay -4*sigmaz
+geometry.prob_hi = +4*sigmax +4*sigmay +4*sigmaz
```
RemiLehe, line 12: Here the resolution was effectively increased by reduced the physical size of the box.

Examples/Tests/openbc_poisson_solver/inputs_3d
```diff
@@ -8,17 +8,17 @@ amr.n_cell = 128 128 128
...
 boundary.field_lo = open open open
 boundary.field_hi = open open open
 warpx.const_dt = 1e-14
 warpx.do_electrostatic = relativistic
 warpx.poisson_solver = fft
+warpx.use_filter = 1
```
RemiLehe:
```suggestion
```

Regression/Checksum/benchmarks_json/ElectrostaticSphere.json
```diff
@@ -1,17 +1,17 @@
 {
   "lev=0": {
-    "Ex": 6.504803536004636,
-    "Ey": 6.504803536004637,
-    "Ez": 6.504803536004637,
+    "Ex": 6.504631859049922,
```
roelof-groenewald, line 3: I notice this file and `ElectrostaticSphereRelNodal.json` match exactly. In a different PR we could make a single analysis script that would use the same json file for both tests.

Regression/Checksum/benchmarks_json/BeamBeamCollision.json
```diff
@@ -1,96 +1,96 @@
...
+    "Bz": 50291310.73170665,
+    "Ex": 2.873993867215236e+20,
+    "Ey": 2.8740263458334176e+20,
+    "Ez": 1.3469564521662371e+17,
+    "rho_beam1": 7.96926761425663e+16,
+    "rho_beam2": 7.969234189119718e+16,
+    "rho_ele1": 325562788515568.7,
+    "rho_ele2": 301373974746269.7,
+    "rho_pos1": 314013278169396.75,
+    "rho_pos2": 311298336003435.56
```
roelof-groenewald, line 14: These values changed quite drastically. Is that within your expectation?

roelof-groenewald (approved):
Looks good, I just had one question about the amount of change in the `rho` benchmark values. If that level of change is reasonable to you based on what you saw during testing / implementation we can merge.

Regression/Checksum/benchmarks_json/BeamBeamCollision.json
roelof-groenewald: Other tests with changes this large had resolution changes as well.

RemiLehe: Yes, I think that this makes sense, because `ele1`, `ele2` and `pos1`, `pos2` come from QED pair creation processes that occur when the two beams are crossing. The rates for these QED processes are quite sensitive to the magnitude of the local E/B field, and thus it is quite possible that filtering alters the corresponding E/B in a way that already significantly modifies these QED rates.

I think that this indicates that, in order to reach numerical convergence, this simulation should be run at higher resolution.

PR #4830, opened 2024-04-05
Fix out of bounds memory access in `HybridPICModel::GetCurrentExternal`

roelof-groenewald:
Running in debug mode revealed a bug in the `HybridPICModel::GetCurrentExternal` function:
```
amrex::Abort::0:: (-4,-4,140,0) is out of bound (-1:76,-1:77,143:193,0:0) !!!
SIGABRT
```
This PR fixes the out of bounds access and reduces the memory requirement of the Ohm's law solver by setting the number of ghost cells for the external current multifab equal to zero. This can be done since no boundary values ever need to be exchanged between tiles of this multifab, all values are local.


ax3l (approved):
Good catch. Thanks! :)

PR #4831, opened 2024-04-05
Fix out of bounds access to `distance_to_eb`

roelof-groenewald:
While running in debug mode I found that there are out of bounds accesses to the `distance_to_eb` multifab. Tracing these I discovered that they come from `AddNParticles()` due to the fact that particles are scraped before they are redistributed. This PR fixes the issue.

### DIFF
--- a/Source/Particles/WarpXParticleContainer.cpp
+++ b/Source/Particles/WarpXParticleContainer.cpp
@@ -302,10 +302,9 @@ WarpXParticleContainer::AddNParticles (int lev, long n,
 #ifdef AMREX_USE_EB
     auto & distance_to_eb = WarpX::GetInstance().GetDistanceToEB();
     scrapeParticles( *this, amrex::GetVecOfConstPtrs(distance_to_eb), ParticleBoundaryProcess::Absorb());
-#endif
-
     // Call (local) redistribute again to remove particles with invalid ids
     Redistribute(lev, lev, 0, 1, true);
+#endif
 }
 
 /* \brief Current Deposition for thread thread_num
--- a/Source/Particles/WarpXParticleContainer.cpp
+++ b/Source/Particles/WarpXParticleContainer.cpp
@@ -305,7 +305,7 @@ WarpXParticleContainer::AddNParticles (int lev, long n,
 #endif
 
     // Call (local) redistribute again to remove particles with invalid ids
-    Redistribute(lev, lev, 1, true, true);
+    Redistribute(lev, lev, 0, 0, true);
 }
 
 /* \brief Current Deposition for thread thread_num
--- a/Source/Particles/WarpXParticleContainer.cpp
+++ b/Source/Particles/WarpXParticleContainer.cpp
@@ -158,7 +158,7 @@ WarpXParticleContainer::AllocData ()
 }
 
 void
-WarpXParticleContainer::AddNParticles (int /*lev*/, long n,
+WarpXParticleContainer::AddNParticles (int lev, long n,
                                        amrex::Vector<amrex::ParticleReal> const & x,
                                        amrex::Vector<amrex::ParticleReal> const & y,
                                        amrex::Vector<amrex::ParticleReal> const & z,
@@ -305,7 +305,7 @@ WarpXParticleContainer::AddNParticles (int /*lev*/, long n,
 #endif
 
     // Call (local) redistribute again to remove particles with invalid ids
-    Redistribute(0, -1, 0, true, true);
+    Redistribute(lev, lev, 1, true, true);
 }
 
 /* \brief Current Deposition for thread thread_num
--- a/Source/Particles/WarpXParticleContainer.cpp
+++ b/Source/Particles/WarpXParticleContainer.cpp
@@ -295,13 +295,17 @@ WarpXParticleContainer::AddNParticles (int /*lev*/, long n,
         );
     }
 
+    // Move particles to their appropriate tiles
+    Redistribute();
+
     // Remove particles that are inside the embedded boundaries
 #ifdef AMREX_USE_EB
     auto & distance_to_eb = WarpX::GetInstance().GetDistanceToEB();
     scrapeParticles( *this, amrex::GetVecOfConstPtrs(distance_to_eb), ParticleBoundaryProcess::Absorb());
 #endif
 
-    Redistribute();
+    // Call (local) redistribute again to remove particles with invalid ids
+    Redistribute(0, -1, 0, 1, 1);
 }
 
 /* \brief Current Deposition for thread thread_num
--- a/Source/Particles/WarpXParticleContainer.cpp
+++ b/Source/Particles/WarpXParticleContainer.cpp
@@ -305,7 +305,7 @@ WarpXParticleContainer::AddNParticles (int /*lev*/, long n,
 #endif
 
     // Call (local) redistribute again to remove particles with invalid ids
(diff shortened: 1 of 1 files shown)


Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -295,13 +295,17 @@ WarpXParticleContainer::AddNParticles (int /*lev*/, long n,
...
+
     // Remove particles that are inside the embedded boundaries
 #ifdef AMREX_USE_EB
     auto & distance_to_eb = WarpX::GetInstance().GetDistanceToEB();
     scrapeParticles( *this, amrex::GetVecOfConstPtrs(distance_to_eb), ParticleBoundaryProcess::Absorb());
 #endif
 
-    Redistribute();
+    // Call (local) redistribute again to remove particles with invalid ids
+    Redistribute(0, -1, 0, 1, 1);
```
roelof-groenewald:
```suggestion
    Redistribute(0, -1, 0, true, true);
```

dpgrote: I think I see. You specify the arguments to set `local = 1` to avoid global communication. Correct? Perhaps setting `nGrow=1` also would avoid all communication?
The last argument should be a boolean, `true`, to match the argument types of `Redistribute`.

dpgrote: Ok, I see you changed to a bool. Though `local` should still be an `int`.

roelof-groenewald: Yes, that is correct about `local = 1` (I've actually seen from the collision module that using `true` is okay even though it takes an `int`. I think passing `true` makes more sense, as in `local=true`).
I don't really understand what `nGrow` does. Wouldn't that include a 1 cell wide region around the tile, which would therefore include particles not on the current tile?

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -295,13 +295,17 @@ WarpXParticleContainer::AddNParticles (int /*lev*/, long n,
...
+
     // Remove particles that are inside the embedded boundaries
 #ifdef AMREX_USE_EB
     auto & distance_to_eb = WarpX::GetInstance().GetDistanceToEB();
     scrapeParticles( *this, amrex::GetVecOfConstPtrs(distance_to_eb), ParticleBoundaryProcess::Absorb());
 #endif
 
-    Redistribute();
+    // Call (local) redistribute again to remove particles with invalid ids
+    Redistribute(lev, lev, 1, true, true);
```
dpgrote: I still think that the `local` argument should be 1, since it is expecting an `int`. The description for local says "If > 0, this is the maximum number of cells a particle can have moved since the last Redistribute() call." It is not a flag.
I was thinking that if nGrow=1 and local=1, the routine might be smart enough to know that no communication is needed. No communication is needed since it is only removing invalid particles. Is there maybe  a separate routine that does just that? This would need a comment from @WeiqunZhang

roelof-groenewald: Oh, I see. I definitely interpreted `local` incorrectly. I guess it should actually be `local=0` then, since the particles will have moved at all since the previous time `Redistribute()` was called.
Looking at the code in `AMReX_ParticleContainerI.H`, `nGrow` is used in the `locateParticle()` function ([here](https://github.com/AMReX-Codes/amrex/blob/2cd72a3bbaeb9e44475225e4acb02b35084e19da/Src/Particle/AMReX_ParticleContainerI.H#L316)) only after the particle could not be located within the `ParticleLocData` with `nGrow=0` - basically it expands the search area with `nGrow`. Therefore, seeing as the first `Redistribute` call happens with `nGrow=0`, it shouldn't make any difference to have the second one use `nGrow=1`.

There was some mention recently of creating a function that will only remove invalid particles, but as far as I know that has not been done yet.
```suggestion
    Redistribute(lev, lev, 0, 0, true);
```

roelof-groenewald: This confuses me though since the default value for `local` is `0`, meaning usually when we call `Redistribute` it is as if we specify that the particles have not moved.

dpgrote: Local=0 means a global communication will be done as if particles may have moved across multiple processors. I think the correct value is 1.

roelof-groenewald: I see. But then looking at the actual redistribute function code, it has [this](https://github.com/AMReX-Codes/amrex/blob/2cd72a3bbaeb9e44475225e4acb02b35084e19da/Src/Particle/AMReX_ParticleContainerI.H#L1433):
```
    if (local > 0) { BuildRedistributeMask(0, local); }
```
While I don't fully understand what exactly is done with `BuildRedistributeMask`, it seems to me that it determines the neighboring ranks to which particles could be passed / received from.
Is it correct then to say that even with `local=1` there would still be some unnecessary overhead but that overhead will be minimized compared to the case where particles could be globally redistributed?

This really makes a case for having a function that simply, locally, removes invalid particles.

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -295,13 +295,17 @@ WarpXParticleContainer::AddNParticles (int /*lev*/, long n,
...
+
     // Remove particles that are inside the embedded boundaries
 #ifdef AMREX_USE_EB
     auto & distance_to_eb = WarpX::GetInstance().GetDistanceToEB();
     scrapeParticles( *this, amrex::GetVecOfConstPtrs(distance_to_eb), ParticleBoundaryProcess::Absorb());
 #endif
 
-    Redistribute();
+    // Call (local) redistribute again to remove particles with invalid ids
+    Redistribute(lev, lev, 0, 0, true);
```
roelof-groenewald:
```suggestion
    Redistribute(lev, lev, 0, 1, true);
```

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -295,13 +295,17 @@ WarpXParticleContainer::AddNParticles (int /*lev*/, long n,
...
+
     // Remove particles that are inside the embedded boundaries
 #ifdef AMREX_USE_EB
     auto & distance_to_eb = WarpX::GetInstance().GetDistanceToEB();
     scrapeParticles( *this, amrex::GetVecOfConstPtrs(distance_to_eb), ParticleBoundaryProcess::Absorb());
 #endif
 
-    Redistribute();
+    // Call (local) redistribute again to remove particles with invalid ids
+    Redistribute(lev, lev, 0, 1, true);
```
RemiLehe: Shouldn't this be moved inside the #ifdef:

```
#ifdef AMREX_USE_EB
    auto & distance_to_eb = WarpX::GetInstance().GetDistanceToEB();
    scrapeParticles( *this, amrex::GetVecOfConstPtrs(distance_to_eb), ParticleBoundaryProcess::Absorb());
    // Call (local) redistribute again to remove particles with invalid ids
    Redistribute(lev, lev, 0, 1, true);
#endif
```

RemiLehe (changes requested):

Source/Particles/WarpXParticleContainer.cpp
roelof-groenewald: Yeah, that makes sense for sure. I'll update now.

RemiLehe (approved):

PR #4839, opened 2024-04-09
Refactor output of phi on the particles

RemiLehe:
This is a follow-up on https://github.com/ECP-WarpX/WarpX/pull/4599. As suggested in review comments, the function has been put in a separate `.cpp` file.


ax3l (approved):
Thank you!

ax3l:
Azure was hanging

PR #4846, opened 2024-04-10
ablastr::particles::compute_weights : implement 1D and use template parameter to specify if field is nodal

lucafedeli88:
Currently WarpX does not compile on Frontier because the host function
`ABLASTR_ABORT_WITH_MESSAGE("Error: compute_weights not yet implemented in 1D");` is called inside the host device `ablastr::particles::compute_weights`. This PR fixes the issue by implementing `compute_weights` also for the 1D case.
In addition, an enum template parameter is used to tell the function if the field is nodal or cell-centered, instead of relying on an integer function parameter.


Source/ablastr/particles/NodalFieldGather.H
```diff
@@ -17,11 +17,17 @@
 
 namespace ablastr::particles
 {
+
+    enum class FieldType{
+        nodal,
+        cell_centered
+    };
```
ax3l: Can we use `amrex::IndexType::NODE`/`amrex::IndexType::CELL` for this?

lucafedeli88: Good idea!

lucafedeli88:
ping @ax3l, @RemiLehe ;-)

Source/ablastr/particles/NodalFieldGather.H
```diff
@@ -30,24 +31,30 @@ namespace ablastr::particles
  * \param W        2D array of weights to store each neighbouring node (or cell-centered node)
  * \param nodal    Int that tells if the weights are calculated in respect to the nodes (nodal=1) of the cell-centered nodes (nodal=0)
  */
+template <amrex::IndexType::CellIndex IdxType>
```
ax3l, line 36: Should this file `#include <AMReX_IndexType.H>` for robustness now?

ax3l: Added :)

Source/ablastr/particles/NodalFieldGather.H
```diff
@@ -12,16 +12,19 @@
...
+
 namespace ablastr::particles
 {
+
 /**
  * \brief Compute weight of each surrounding node (or cell-centered nodes) in interpolating a nodal ((or a cell-centered node) field
- *        to the given coordinates. If nodal=1, then the calculations will be done with respect to the nodes (default). If nodal=0, then the calculations will be done with respect to the cell-centered nodal)
- *
- * This currently only does linear order.
+ *        to the given coordinates. If template parameter IdxType is amrex::IndexType::CellIndex::NODE, then the calculations will be done with respect to the nodes (default).
```
ax3l: Is the documented default set?
```suggestion
 *        to the given coordinates. If template parameter IdxType is amrex::IndexType::CellIndex::NODE, then the calculations will be done with respect to the nodes.
```

ax3l (approved):
This looks great, thank you! :sparkles:

PR #4848, opened 2024-04-10
Add ohms law hybrid multifab redistribution

clarkse:
This PR adds redistribution of Ohm's Law Hybrid solver multifabs to allow for load balancing when not using explicit processor decomposition.

### DIFF
--- a/Source/Parallelization/WarpXRegrid.cpp
+++ b/Source/Parallelization/WarpXRegrid.cpp
@@ -235,7 +235,7 @@ WarpX::RemakeLevel (int lev, Real /*time*/, const BoxArray& ba, const Distributi
         RemakeMultiFab(phi_fp[lev], dm, true ,lev);
 
         if (WarpX::electromagnetic_solver_id == ElectromagneticSolverAlgo::HybridPIC) {
-            RemakeMultiFab(m_hybrid_pic_model->rho_fp_temp[lev], dm, false, lev);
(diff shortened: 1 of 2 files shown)


Source/Parallelization/WarpXRegrid.cpp
```diff
@@ -206,6 +207,11 @@ WarpX::RemakeLevel (int lev, Real /*time*/, const BoxArray& ba, const Distributi
                 RemakeMultiFab(Efield_avg_fp[lev][idim], dm, true ,lev);
                 RemakeMultiFab(Bfield_avg_fp[lev][idim], dm, true ,lev);
             }
+            if (WarpX::electromagnetic_solver_id == ElectromagneticSolverAlgo::HybridPIC) {
+                RemakeMultiFab(m_hybrid_pic_model->current_fp_temp[lev][idim], dm, false, lev);
```
roelof-groenewald: Is it to correct to say that the second to last argument sets whether the current values of the multifab will be kept after the remake?
If so, we should keep the values in `current_fp_temp` and `rho_fp_temp` since they are used in the next step to interpolate to intermediate time values.

roelof-groenewald (changes requested):
Thanks for this PR! 
There were some changes in the fix for `AddNParticles` that are not reflected here. Could you please rebase to `development`? I also had one other comment.

roelof-groenewald (approved):
Looks good. Thanks a lot!

PR #4849, opened 2024-04-10
Use function to remove invalid particles without MPI exchange

RemiLehe:
The new function (introduced by @AlexanderSinn in `amrex`) is a local operation (no MPI communication) and is thus preferable over `ReDistribute` (which also removes invalid particles, but involves MPI communications at the same time).

### DIFF
--- a/Source/Particles/WarpXParticleContainer.cpp
+++ b/Source/Particles/WarpXParticleContainer.cpp
@@ -302,7 +302,11 @@ WarpXParticleContainer::AddNParticles (int /*lev*/, long n,
 #ifdef AMREX_USE_EB
     auto & distance_to_eb = WarpX::GetInstance().GetDistanceToEB();
     scrapeParticles( *this, amrex::GetVecOfConstPtrs(distance_to_eb), ParticleBoundaryProcess::Absorb());
-    // Remove particles with invalid ids
+    deleteInvalidParticles();
+}
+
+void
+WarpXParticleContainer::deleteInvalidParticles () {
     const int nLevels = finestLevel();
     for (int lev = 0; lev <= nLevels; ++lev) {
 #ifdef AMREX_USE_OMP
--- a/Source/Particles/WarpXParticleContainer.cpp
+++ b/Source/Particles/WarpXParticleContainer.cpp
@@ -309,7 +309,8 @@ WarpXParticleContainer::AddNParticles (int /*lev*/, long n,
 #pragma omp parallel
 #endif
         for (WarpXParIter pti(*this, lev); pti.isValid(); ++pti) {
-            removeInvalidParticles( pti );
+            ParticleTileType& ptile = ParticlesAt(lev, pti);
+            removeInvalidParticles( ptile );
         }
     }
 #endif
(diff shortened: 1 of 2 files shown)


RemiLehe:
@roelof-groenewald Do you know whether we have existing CI tests that are sensitive to this part of the code and would confirm that this function is working as expected?

roelof-groenewald:
> Do you know whether we have existing CI tests that are sensitive to this part of the code and would confirm that this function is working as expected?

Probably not since the CI tests were not affected by #4831 and the collision routines have all invalid particles' weights set to 0 (so they don't affect checksum values).
But I can test it locally using the new merging routine since I'm tracking particle numbers in my tests.

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -302,8 +302,21 @@ WarpXParticleContainer::AddNParticles (int /*lev*/, long n,
 #ifdef AMREX_USE_EB
     auto & distance_to_eb = WarpX::GetInstance().GetDistanceToEB();
     scrapeParticles( *this, amrex::GetVecOfConstPtrs(distance_to_eb), ParticleBoundaryProcess::Absorb());
-    // Call (local) redistribute again to remove particles with invalid ids
-    Redistribute(0, -1, 0, 1, true);
+    deleteInvalidParticles();
```
roelof-groenewald, line 305: The `#endif` currently on line 320 should be directly below this line. Currently the `WarpXParticleContainer::deleteInvalidParticles` function is only defined for EB=ON compilations.

roelof-groenewald:
@RemiLehe I can confirm that this function works as intended (with the above suggested change included).
I tested a problem where I merge a number of particles on step 2. The case was run with 2 procs. Immediately after merging I print the number of ions using `particle_container.total_number_of_particles(True, True)` (i.e only counting valid particles). I also grab the particle ids and print the length of those arrays as well as the number of invalid entries in the array (i.e. w < 0). The following output is without a `Redistribute` or `deleteInvalidParticles` call:
```
STEP 2 starts ...
--- INFO    : Resampled hydrogen at step 2: macroparticle count decreased by
             516885
Num ions: 123115; len(id) = 320000; invalid ids # 258445
Num ions: 123115; len(id) = 320000; invalid ids # 258440
```
The following output is with a `deleteInvalidParticles` call between merging and the output:
```
STEP 2 starts ...
--- INFO    : Resampled hydrogen at step 2: macroparticle count decreased by
             516885
Num ions: 123115; len(id) = 61555; invalid ids # 0
Num ions: 123115; len(id) = 61560; invalid ids # 0
```

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -302,8 +302,21 @@ WarpXParticleContainer::AddNParticles (int /*lev*/, long n,
...
+    for (int lev = 0; lev <= nLevels; ++lev) {
+#ifdef AMREX_USE_OMP
+#pragma omp parallel
+#endif
+        for (WarpXParIter pti(*this, lev); pti.isValid(); ++pti) {
+            ParticleTileType& ptile = ParticlesAt(lev, pti);
+            removeInvalidParticles( ptile );
+        }
+    }
 #endif
```
RemiLehe:
```suggestion
```

Source/Particles/WarpXParticleContainer.cpp
RemiLehe, line 305:
```suggestion
    deleteInvalidParticles();
#endif
```

RemiLehe: Thanks! Done.

roelof-groenewald (approved):

PR #4850, opened 2024-04-10
Add thermal boundary specification to picmi

roelof-groenewald:
- [x] need to also include for 1d, 2d and 3d cartesian grids


dpgrote (approved):
This looks good.
I was trying to come up with a different way of specifying the input, but couldn't come up with anything nicer.

PR #4851, opened 2024-04-11
Use local deleteInvalidParticles (instead of Redistribute) in binary collisions

RemiLehe:
This should be more efficient since it avoids MPI communications.


roelof-groenewald (approved):
I made the same change in #4577 😃

PR #4852, opened 2024-04-11
Reorder particle boundary handling and call to `Redistribute`

roelof-groenewald:
Based on the bugfix from #4831, wouldn't the same risk currently exist in the `WarpX::HandleParticlesAtBoundaries()` function?
This PR reorganizes that function to first redistribute particles to their appropriate tiles before determining whether the particle is overlapping with an embedded boundary. It then uses the new `deleteInvalidParticles()` function to remove particles that were determined to be invalid during scraping.

### DIFF
--- a/Source/Evolve/WarpXEvolve.cpp
+++ b/Source/Evolve/WarpXEvolve.cpp
@@ -488,6 +488,9 @@ void WarpX::HandleParticlesAtBoundaries (int step, amrex::Real cur_time, int num
 {
     mypc->ContinuousFluxInjection(cur_time, dt[0]);
 
+    mypc->ApplyBoundaryConditions();
+    m_particle_boundary_buffer->gatherParticlesFromDomainBoundaries(*mypc);
+
     // Non-Maxwell solver: particles can move by an arbitrary number of cells
     if( electromagnetic_solver_id == ElectromagneticSolverAlgo::None ||
         electromagnetic_solver_id == ElectromagneticSolverAlgo::HybridPIC )
@@ -514,14 +517,11 @@ void WarpX::HandleParticlesAtBoundaries (int step, amrex::Real cur_time, int num
         }
     }
 
-    mypc->ApplyBoundaryConditions();
-
     // interact the particles with EB walls (if present)
 #ifdef AMREX_USE_EB
     mypc->ScrapeParticles(amrex::GetVecOfConstPtrs(m_distance_to_eb));
+    m_particle_boundary_buffer->gatherParticlesFromEmbeddedBoundaries(*mypc, amrex::GetVecOfConstPtrs(m_distance_to_eb));
 #endif
-
-    m_particle_boundary_buffer->gatherParticles(*mypc, amrex::GetVecOfConstPtrs(m_distance_to_eb));
 
     mypc->deleteInvalidParticles();
 
--- a/Source/Evolve/WarpXEvolve.cpp
+++ b/Source/Evolve/WarpXEvolve.cpp
@@ -519,11 +519,10 @@ void WarpX::HandleParticlesAtBoundaries (int step, amrex::Real cur_time, int num
 
     // interact the particles with EB walls (if present)
 #ifdef AMREX_USE_EB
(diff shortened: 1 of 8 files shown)


Source/Evolve/WarpXEvolve.cpp
```diff
@@ -488,15 +488,6 @@ void WarpX::HandleParticlesAtBoundaries (int step, amrex::Real cur_time, int num
 {
     mypc->ContinuousFluxInjection(cur_time, dt[0]);
 
-    mypc->ApplyBoundaryConditions();
```
ax3l, line 491: Potentially, adding a test similar as in current deposition, that checks travel distance is needed here (for EB).
Essentially, particles that shall undergo EB handling need to be checked if they are within the ngrow of the iMultiFab of the EBs.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -523,6 +514,17 @@ void WarpX::HandleParticlesAtBoundaries (int step, amrex::Real cur_time, int num
         }
     }
 
+    mypc->ApplyBoundaryConditions();
```
dpgrote: I'm not sure that this is correct. If I understand it, `Redistribute` not only sends particles to their appropriate processor, but it also deletes any particles outside of the domain. You can see this in the routine `locateParticle` in amrex/Src/Particle/AMReX_ParticleContainerI.H. So, `ApplyBoundaryConditions` would have to be called before `Redistribute` to handle the reflecting and thermal boundary conditions.

roelof-groenewald:
From 4/16/24 development meeting:
New logic:
- Deal with outer boundaries (ApplyBoundaries currently includes scraping)
- Collect particles that end up in outer boundaries, put them into boundary buffers.
- Redistribute (and delete invalid particles).
- Flag particles that are in the EB, copy them in the boundary buffer, mark them as invalid in their source species. - Small possible update: merge flagging src and copying into dst into one call.
- Delete invalid particles in source species.

roelof-groenewald:
@ax3l and @dpgrote this latest version should achieve the desired logic as discussed. Let me know your thoughts.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -523,6 +517,14 @@ void WarpX::HandleParticlesAtBoundaries (int step, amrex::Real cur_time, int num
         }
     }
 
+    // interact the particles with EB walls (if present)
+#ifdef AMREX_USE_EB
+    mypc->ScrapeParticles(amrex::GetVecOfConstPtrs(m_distance_to_eb));
```
dpgrote: While these changes are being made, this routine could be renamed to be more explicit, something like `ScrapeParticlesAtEB`.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -523,6 +517,14 @@ void WarpX::HandleParticlesAtBoundaries (int step, amrex::Real cur_time, int num
...
+    m_particle_boundary_buffer->gatherParticlesFromEmbeddedBoundaries(*mypc, amrex::GetVecOfConstPtrs(m_distance_to_eb));
+#endif
+
+    mypc->deleteInvalidParticles();
```
dpgrote, line 524: This should be inside the EB block above since it is only needed there. The Redistribute above will have already deleted invalid particles outside of the domain.

Source/Evolve/WarpXEvolve.cpp
roelof-groenewald: Sounds good!

Source/Evolve/WarpXEvolve.cpp
roelof-groenewald: Good point! I'll move it.

Source/Evolve/WarpXEvolve.cpp
dpgrote: The problem I see is if a particles passes through an EB and then goes outside the domain. I don't see an easy way to deal with this case, associating that lost particle to the EB instead of the domain boundary. It would be even worse if the particle went beyond any guard cells.

dpgrote (approved):
Looks good to me! Thanks!

PR #4653, opened 2024-01-30
Particle Container to Pure SoA Again

ax3l:
Transition to new, purely SoA particle containers.

This was originally merged in #3850 and reverted in #4652, since we discovered issues loosing particles & laser particles on GPU.

- [x] update AMReX again with Jan 30/31 changes #4654
- [x] improve performance of id handling via constants and https://github.com/AMReX-Codes/amrex/pull/3735
- [x] update docs to make sure AoS is not mentioned anymore and particle id handling is guided
- [x] run test suite on CUDA devices with high checksum tolerance

## Fun Mini-Benchmarks on CPU, DP

Hardware: 12th Gen Intel(R) Core(TM) i9-12900H

```bash
export OMP_NUM_THREADS=1
./warpx.3d ../../Examples/Tests/performance_tests/automated_test_1_uniform_rest_32ppc amr.max_grid_size=64 amr.n_cell=64 64 64 max_step=5 &
taskset -cp 6 $!
```

[cpu_legacy.txt](https://github.com/ECP-WarpX/WarpX/files/14149010/cpu_legacy.txt), [cpu_soa.txt](https://github.com/ECP-WarpX/WarpX/files/14149011/cpu_soa.txt)

Overall speed: similar to noise level of repeated runs (as expected).

Few noteworthy details in top 10 functions by runtime (Excl.):
- `ParticleContainer::RedistributeCPU`: 8% slower :eyes: -> https://github.com/AMReX-Codes/amrex/issues/3744
- `WarpXParticleContainer::ApplyBoundaryConditions`: 5% faster
- `WarpX::OneStep_nosub`: 2% slower :eyes: 

## Fun Mini-Benchmarks on A100 GPU, DP

Hardware: Perlmutter (NERSC) A100 GPU

```bash
./warpx.3d ../../Examples/Tests/performance_tests/automated_test_1_uniform_rest_32ppc amr.max_grid_size=256 amr.n_cell=256 256 256 max_step=5
```

Overall speed: 1.4% faster

Few noteworthy details in top 10 functions by runtime (Excl.):
- `GatherAndPush`: 1.2% faster
- `Redistribute_partition`: 4% faster
- `AddPlasma`: 2.6% faster
- `ApplyBoundaryConditions`: 1% faster
- `SortParticlesForDeposition`: 231% faster :rocket: :rocket: :sparkles:
- `PermutationForDeposition`: 3% faster
- `InitData`: 15% faster :rocket:
- rest in TOP10 is the same as before

PR #4686, opened 2024-02-09
Fix: openPMD RZ Positions

ax3l:
Fix the particle position write in openPMD with SoA particle layout.

For the openPMD I/O backend, we transform the r,z,theta particle components back to x,y,z.

Missing in #4653.
Fix #4676.

PR #4855, opened 2024-04-12
Always write the particles' positions

### LINKED ISSUE #4841: Particle diagnostic silently writes zero position if x y z is not explicitly specified
One of the changes in the last few months (some time between December and March) has affected the default behavior of the output for the particle data diagnostic. The new behavior seems unexpected and is not documented.

In a Full diagnostic, consider
<diag_name>.<species_name>.variables = w

Previous behavior:
Particle position is written in addition to the weight
-- this is what the documentation indicates should happen

Current behavior:
Zeros are written for the particle position regardless of where the particles actually are
The actual particle position is only written if the desired positions are explicitly specified, i.e.,
<diag_name>.<species_name>.variables = w x y z

MWE is attached

[inputs_3d.txt](https://github.com/ECP-WarpX/WarpX/files/14924231/inputs_3d.txt)

RemiLehe:
As part of our restructuring from AoS positions to SoA (https://github.com/ECP-WarpX/WarpX/pull/4653), it seems that the particles' positions are not written out by default anymore. (But they probably should, since, in my understanding, the particles' positions and ID are [required attributes](https://github.com/openPMD/openPMD-standard/blob/latest/STANDARD.md#records-for-each-particle-species) in the openPMD standard.)

This PR writes out the particle positions by default.

Fixes #4841

### DIFF
--- a/Source/Diagnostics/ParticleDiag/ParticleDiag.cpp
+++ b/Source/Diagnostics/ParticleDiag/ParticleDiag.cpp
@@ -50,8 +50,8 @@ ParticleDiag::ParticleDiag(
                 }
             }
             // Always write the position
-            // Note: for WARPX_DIM_RZ, no need to set the flag for x and y: they are always written
-#if defined (WARPX_DIM_2D) || defined(WARPX_DIM_3Z)
+            // Note: for WARPX_DIM_RZ, no need to set the flag y: it is always written
+#if !defined (WARPX_DIM_1D_Z)
             m_plot_flags[pc->getParticleComps().at("x")] = 1;
 #endif
 #if defined (WARPX_DIM_3D)
--- a/Source/Diagnostics/ParticleDiag/ParticleDiag.cpp
+++ b/Source/Diagnostics/ParticleDiag/ParticleDiag.cpp
@@ -61,6 +61,13 @@ ParticleDiag::ParticleDiag(
         }
     }
 
+#ifdef WARPX_DIM_RZ
+    // Always write out theta, whether or not it's requested,
+    // to be consistent with always writing out r and z.
+    // TODO: openPMD does a reconstruction to Cartesian, so we can now skip force-writing this
+    m_plot_flags[pc->getParticleComps().at("theta")] = 1;
+#endif
+
     // build filter functors
     m_do_random_filter = utils::parser::queryWithParser(
         pp_diag_name_species_name, "random_fraction", m_random_fraction);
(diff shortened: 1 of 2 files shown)


Source/Diagnostics/ParticleDiag/ParticleDiag.cpp
```diff
@@ -41,6 +41,7 @@ ParticleDiag::ParticleDiag(
                     // Therefore, this case needs to be treated specifically.
                     m_plot_phi = true;
                 } else {
+                    amrex::Print() << var << std::endl;
```
RemiLehe:
```suggestion
```

Source/Diagnostics/ParticleDiag/ParticleDiag.cpp
```diff
@@ -51,13 +51,10 @@ ParticleDiag::ParticleDiag(
             }
         }
     }
-
-#ifdef WARPX_DIM_RZ
-    // Always write out theta, whether or not it's requested,
-    // to be consistent with always writing out r and z.
-    // TODO: openPMD does a reconstruction to Cartesian, so we can now skip force-writing this
-    m_plot_flags[pc->getParticleComps().at("theta")] = 1;
-#endif
```
RemiLehe: My impression is that `theta` is not useful, since the particles are dumped in Cartesian coordinates anyway. In fact, I think that this flag is not doing anything, since `theta` is actually not written anymore as of:
https://github.com/ECP-WarpX/WarpX/pull/4686

lucafedeli88 (approved):
Thanks!

Source/Diagnostics/ParticleDiag/ParticleDiag.cpp
RemiLehe: Hm, it seems that this is still needed when writing `plotfiles`.

Source/Diagnostics/ParticleDiag/ParticleDiag.cpp
```diff
@@ -49,6 +49,15 @@ ParticleDiag::ParticleDiag(
                     m_plot_flags[existing_variable_names.at(var)] = 1;
                 }
             }
+            // Always write the position
+            // Note: for WARPX_DIM_RZ, no need to set the flag y: it is always written
```
RemiLehe, line 53: In RZ, I observed empirically that: 
- I **need** to set `m_plot_flags[pc->getParticleComps().at("x")] = 1;` (otherwise the Cartesian `x` position is not written correctly with openPMD)
- I **cannot** set `m_plot_flags[pc->getParticleComps().at("y")] = 1;` (otherwise I get an error in `at`, saying that this property does not exist), but the Cartesian `y` is automatically reconstructed and written anyway.

@ax3l Do you understand why `y` is automatically written, but not `x`? (i.e. why do I **need** to set the flag for `x`, but not for `y`)

ax3l:
Replaced by #4914

PR #4859, opened 2024-04-16
DepositCharge: Component 1 Allocated?

ax3l:
It looks like in the EM solver, with `rho` field diagnostics enabled, we try to write into component 1 of rho even if it is not allocated.

Spotted by @sevicky 

- [x] add asserts
- [ ] add a test if needed
- [ ] fix bug

### DIFF
--- a/Source/Particles/PhysicalParticleContainer.cpp
+++ b/Source/Particles/PhysicalParticleContainer.cpp
@@ -2227,10 +2227,9 @@ PhysicalParticleContainer::Evolve (int lev,
             if (rho && ! skip_deposition && ! do_not_deposit) {
                 // Deposit charge after particle push, in component 1 of MultiFab rho.
                 // (Skipped for electrostatic solver, as this may lead to out-of-bounds)
-                WARPX_ALWAYS_ASSERT_WITH_MESSAGE(rho->nComp() >= 2,
-                    "Cannot deposit charge in rho component 1: only component 0 is allocated!");
-
                 if (WarpX::electrostatic_solver_id == ElectrostaticSolverAlgo::None) {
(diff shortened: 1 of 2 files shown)


Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -2227,6 +2227,9 @@ PhysicalParticleContainer::Evolve (int lev,
             if (rho && ! skip_deposition && ! do_not_deposit) {
                 // Deposit charge after particle push, in component 1 of MultiFab rho.
                 // (Skipped for electrostatic solver, as this may lead to out-of-bounds)
+                WARPX_ALWAYS_ASSERT_WITH_MESSAGE(rho->nComp() >= 2,
+                    "Cannot deposit charge in rho component 1: only component 0 is allocated!");
+
                 if (WarpX::electrostatic_solver_id == ElectrostaticSolverAlgo::None) {
```
roelof-groenewald: Seeing as the deposit is skipped for electrostatic solvers wouldn't it be better to put the assert inside the `if` statement?
```suggestion
                if (WarpX::electrostatic_solver_id == ElectrostaticSolverAlgo::None) {
                    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(rho->nComp() >= 2,
                        "Cannot deposit charge in rho component 1: only component 0 is allocated!");
```

roelof-groenewald (changes requested):

ax3l:
I see that `Examples/Physics_applications/laser_acceleration/inputs_3d` outputs `rho`, so @sevicky might use an outdated branch where this can fail.

roelof-groenewald (approved):
LGTM. even without immediate failures it's good to have these asserts to protect against future issues.

Issue #4807, opened 2024-03-27
ApplyStencil: Shared Memory & Fixed Stencils


ax3l:
The `ApplyStencil` routines for filters are very slow right now, mainly because of many reads to global memory (e.g., reading the same field N*8 times) and runtime array accesses.

What I propose to fix it:
- add specialization with compile-time stencils for the default values we use
- add shared memory to avoid reading the same values ~8x times from global memory

botszhuang:
I am interested in working on this. 
I have experience with OpenCL and GPU kernel optimization. 
Is there any specific branch or existing work I should be aware of?

ax3l:
That is awesome! @RemiLehe just opened #7032 that we should be aware off.
We should finish that PR (better memory access patterns) first, then a follow-up could do shared memory.

Remi already did a small experiment with compile-time stencils, and adding those did not pay off in his experiment (on GPU) because of the low flop-to-byte ratio of the stencil operations that make the kernels memory-bound. One could revisit this as a last step and also compare to how it influences (auto-)vectorization on CPU, but it can be lower priority.

arcusbuilds:
I'd like to work on the shared-memory part of this. Two questions on sequencing:

#7032 has been WIP since July. Should I rebase and finish that first (separable 1D passes), then do shared memory as a second PR, or is it fine to build the shared-memory tiling on top of it as it stands?

Also @botszhuang expressed interest in March. Are you still working on this?

I'll skip the compile-time stencils given Remi's finding that they didn't pay off on GPU.

PR #4865, opened 2024-04-16
EBs: Compiled by Default, Controlled at Runtime

### LINKED ISSUE #4863: EBs: Runtime Option
We should add a runtime option that explicitly enables embedded boundaries.

That way, we can always build AMReX and WarpX with EBs on, ship a feature-complete binary in package managers, and let users opt in to the algorithms when needed.

If disabled at runtime (default), no additional memory or runtime costs must be incurred.

## Suggested Name

Following existing options https://warpx.readthedocs.io/en/latest/usage/parameters.html#embedded-boundary-conditions
```ini
warpx.eb_enabled = 0
```

Update: we will, after all, derive this enable/disable from existing EB runtime options.

ax3l:
Enable embedded boundary compilation by default.
Control via the existing runtime options.

Close #4863

## Action Items

- [x] build system & doc default updates
- [x] introduce runtime control, update runtime logic
- [x] run a non-EB test, ensure it is as fast and as low high-water mark in memory consumption as with EB compile-time disabled

## Follow-up

We could require that AMReX dependency is always built with `-DAMReX_EB=ON` and with that remove all the safety macros `AMREX_USE_EB` that we currently have to enable that WarpX can also build if AMReX has no EB support. Since EB support in AMReX does not trigger extra dependencies and only moderately increases code complexity for compile (as does that WarpX always requires `-DAMReX_PARTICLES=ON`), this might be sensible.

To make the compile-time impact more mild, @WeiqunZhang proposed to add another compile-time control to AMReX that disables the compile of a few large linear EB solvers that are not used/needed for WarpX.


dpgrote:
This is a good thing to do. Though, is a separate input flag needed? The code can check if `warpx.eb_implicit_function` has been specified, or `eb2.geom_type`. Elsewhere, enable type input parameters have been removed, relying on specification of the controlling parameters.

ax3l:
Thanks, Dave!

I was wondering as well, but in the end we have at least three ways to turn it on:
https://warpx.readthedocs.io/en/latest/usage/parameters.html#embedded-boundary-conditions
- `warpx.eb_implicit_function`
- `warpx.eb_potential(x,y,z,t)`
- and, currently undocumented, a way to load via STL file in AMReX

We could a) fix the lack of docs and b) write a helper function that checks them or so? Open to suggestions cc @RemiLehe @roelof-groenewald

roelof-groenewald:
The `warpx.eb_potential(x,y,z,t)` parameter would not be good to use since it is perfectly fine to leave that out - it is only used by the electrostatic solver and by default the embedded boundary potential is just 0.
I think it would be reasonable to then check if either `warpx.eb_implicit_function` or the STL file input is specified. We don't expect to have a lot more ways of loading EBs in the future, right? So it shouldn't be hard to maintain a list of `or` statements to determine the runtime parameter.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -95,11 +95,13 @@ void FiniteDifferenceSolver::CalculateCurrentAmpereCylindrical (
...
-        amrex::Array4<amrex::Real> const& lt = edge_lengths[1]->array(mfi);
-        amrex::Array4<amrex::Real> const& lz = edge_lengths[2]->array(mfi);
-#endif
+        amrex::Array4<amrex::Real> lr, lt, lz;
+
+        if (eb_enabled) {
+            lr = edge_lengths[0]->array(mfi);
+            lt = edge_lengths[1]->array(mfi);
+            lz = edge_lengths[2]->array(mfi);
+        }
```
roelof-groenewald: It would be nice if we can always have a well defined EBFactory which just doesn't cover any of the domain when it is not wanted. Then the `edge_lengths` and similar multifabs could be properly defined in all cases and we wouldn't have to have the `if (eb_enabled)` block. Similarly lime 131 below could simply be:
```
if (lr(i, j, 0) <= 0) return;
```
and the same for all the other similar occurrences.

WeiqunZhang:
A separate bool ParmParse parameter is not needed unless we want to the one flag to rule them all. If we want a bool variable to be used in the code for convenience, we can add a variable to WarpX class. EB is initialized in WarpX::InitEB. There we can set it to true or false.

There are basically two ways to enable EB in WarpXInitEB, (1) `warpx.eb_implicit_function` (2) AMReX's `eb2.geom_type`. The STL way is the second way documented here https://amrex-codes.github.io/amrex/docs_html/EB.html?highlight=stl.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
WeiqunZhang: That would waste memory.

ax3l:
Thank you for all the ideas!

> we can add a variable to WarpX class. EB is initialized in WarpX::InitEB. There we can set it to true or false.

In ImpactX and also more in WarpX, I try to avoid using global variables in the central sim class and instead use globals from the ParmParser. That way, we do not need to keep them in sync when changing options at runtime (e.g., from Python) and do not need to include a heavy central class everywhere.
But I also just introduced `WarpX::m_eb_enabled`... :sweat_smile: 

I would for now write a free standing helper function instead that we can use :)

WeiqunZhang:
Maybe a free function in WarpX like this
```
bool hasEB ()
{
#if defined(AMREX_USE_EB)
    return ! amrex::EB2::IndexSpace::empty();
#else
    return false;
#endif
}
```

WeiqunZhang:
Note that amrex::IndexSpace has a static member, a vector of amrex::IndexSpace. Somehow this comes into my mind by your mention of global variable. https://en.wiktionary.org/wiki/%E6%88%91%E4%B8%8D%E5%85%A5%E5%9C%B0%E7%8D%84%EF%BC%8C%E8%AA%B0%E5%85%A5%E5%9C%B0%E7%8D%84

WeiqunZhang:
Note that `amrex::EB2::IndexSpace::empty()` is true before WarpX::InitEB is called and will be false if EB is initialized in that function.

Source/EmbeddedBoundary/WarpXFaceExtensions.cpp
```diff
@@ -545,7 +556,11 @@
 
 
 void
-WarpX::ComputeEightWaysExtensions() {
+WarpX::ComputeEightWaysExtensions ()
```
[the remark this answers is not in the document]
ax3l: I agree 😁

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
ax3l: Unless we would implement an `Array4` variant that returns 0 outside of the covered domain, which we could use elsewhere as well (e.g. #4807). But that then probably leads to register bloat (unless we implement as in `mdspan` a variant that have optionally known compile-time sizes).

Source/BoundaryConditions/WarpXEvolvePML.cpp
```diff
@@ -299,29 +302,24 @@ WarpX::DampJPML (int lev, PatchType patch_type)
             int const zs_lo = sigba[mfi].sigma_star_cumsum_fac[1].lo();
 #endif
 
+            bool const eb_enabled = m_eb_enabled;
             amrex::ParallelFor( tjx, tjy, tjz,
                 [=] AMREX_GPU_DEVICE (int i, int j, int k) {
-#ifdef AMREX_USE_EB
-                    if(pml_lxfab(i, j, k) <= 0) return;
-#endif
+                    if (eb_enabled && pml_lxfab(i, j, k) <= 0) return;
```
ax3l: `amrex::Array4` has a cast to `bool` implemented, which enables this compact alternative:
```suggestion
                    if (pml_lxfab && pml_lxfab(i, j, k) <= 0) { return; }
```

ax3l:
Cool, looks like most tests start to pass, Ohm solver and Nodal ES solver still fail.

EZoni:
Just a note: 
I would try to merge this PR before #5068 (still work-in-progress at this time). This way we can avoid having separate Azure matrix elements for builds with/without EB. These unique builds, with EB always ON, are already implemented in #5068 and, hopefully, merging this PR will resolve the failures due to EB builds of non-EB tests that we currently observe in #5068.

@ax3l Please let me know if I can be of any help to finalize this PR before #5068.

ax3l:
Discussed today in person: it is ok to merge CTest first and keep the extra build, because it has higher priority than this update here.

ax3l:
@WeiqunZhang for the error in `Examples/Tests/electrostatic_sphere/inputs_rz` I am not sure if this might not be an AMReX error?
```
0::Assertion `!m_instance.empty()' failed, file "/home/axel/src/warpx/build/_deps/fetchedamrex-src/Src/EB/AMReX_EB2.H", line 48, Msg: "Have you forgot to call EB2::build? It's required even if the geometry is all regular." !!!
```
it happens in a totally unrelated `ParallelFor` of `warpx_computedivb`.

Source/BoundaryConditions/WarpXEvolvePML.cpp
ax3l: done

WeiqunZhang:
`amrex::MLEBNodeFDLaplacian` used `AMREX_USE_EB` as the flag for having EB nor not. We could change that.

ax3l:
@RemiLehe @WeiqunZhang can you help me review the change to the Poisson solvers? This PR is mostly done, but I see a few small changes in the Poisson solve around 1e-8 relative changes in checksums of ES tests.

Source/BoundaryConditions/PML.cpp
```diff
@@ -673,9 +678,11 @@ PML::PML (const int lev, const BoxArray& grid_ba,
...
 
 #ifdef AMREX_USE_EB
-    pml_field_factory = amrex::makeEBFabFactory(*geom, ba, dm,
-                                              {max_guard_EB, max_guard_EB, max_guard_EB},
-                                              amrex::EBSupport::full);
+    if (eb_enabled) {
+        pml_field_factory = amrex::makeEBFabFactory(*geom, ba, dm,
+                                                  {max_guard_EB, max_guard_EB, max_guard_EB},
+                                                  amrex::EBSupport::full);
+    }
```
WeiqunZhang: I think this should be
```
#ifdef AMREX_USE_EB
    if (eb_enabled) {
        pml_field_factory = amrex::makeEBFabFactory(*geom, ba, dm,
                                                  {max_guard_EB, max_guard_EB, max_guard_EB},
                                                  amrex::EBSupport::full);
    } else
#endif
    {
        amrex::ignore_unused(max_guard_EB);
        pml_field_factory = std::make_unique<FArrayBoxFactory>();
    }
```

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -495,7 +495,7 @@ void WarpX::HandleParticlesAtBoundaries (int step, amrex::Real cur_time, int num
     mypc->ContinuousFluxInjection(cur_time, dt[0]);
 
     mypc->ApplyBoundaryConditions();
-    m_particle_boundary_buffer->gatherParticlesFromDomainBoundaries(*mypc);
+    if (m_eb_enabled) { m_particle_boundary_buffer->gatherParticlesFromDomainBoundaries(*mypc); }
```
WeiqunZhang: This was not inside #ifdef AMREX_USE_EB previously. So does it make sense to add `if (m_eb_enabled)`?

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -627,11 +608,10 @@ void FiniteDifferenceSolver::HybridPICSolveECylindrical (
 
             // Et calculation
             [=] AMREX_GPU_DEVICE (int i, int j, int /*k*/){
-#ifdef AMREX_USE_EB
                 // In RZ Et is associated with a mesh node, so we need to check if the mesh node is covered
                 amrex::ignore_unused(lt);
```
WeiqunZhang: We should remove this. `lt` is never used. We could remove it entirely this function. But even if we want to keep it just in case for future extension, we should move `amrex::ignore_unused` outside the device lambda so that it's not captured.

Source/FieldSolver/FiniteDifferenceSolver/MacroscopicEvolveE.cpp
```diff
@@ -193,15 +194,14 @@ void FiniteDifferenceSolver::MacroscopicEvolveECartesian (
             },
 
             [=] AMREX_GPU_DEVICE (int i, int j, int k){
-#ifdef AMREX_USE_EB
 #ifdef WARPX_DIM_3D
-                if (ly(i,j,k) <= 0) return;
+                if (ly && ly(i,j,k) <= 0) { return; }
 #elif defined(WARPX_DIM_XZ)
                 //In XZ Ey is associated with a mesh node, so we need to check if the mesh node is covered
                 amrex::ignore_unused(ly);
```
WeiqunZhang: Could move this line outside the device lambda.

WeiqunZhang:
Currently in Enable.H,
```
#ifndef WARPX_EB_ENABLED_H_
#define WARPX_EB_ENABLED_H_

#include <AMReX_ParmParse.H>

namespace EB
{
    /** Are embedded boundaries enabled? */
    bool enabled ();

} // namespace EB                                                                                                                                                            
#endif // WARPX_EB_ENABLED_H_
```
We probably should remove `#include <AMReX_ParmParse.H>`, because one day clang-tidy or some other tools might complain about it. If we do that we should add `#include <AMReX_Config.H>` to Enabled.cpp for `AMREX_USE_EB`. Now Enable.cpp gets it indirectly via ParmParse.H.

WeiqunZhang:
Could we merge the latest development branch into this?

ax3l:
> Enable.H include

Oh yes, outdated before the split in .H and .cpp - fixed the include now.

Source/BoundaryConditions/PML.cpp
ax3l: Good catch, thx. Changed.

Source/Evolve/WarpXEvolve.cpp
ax3l: I think this only made sense if eb is enabled and does some unnecessary work otherwise that results in zero filtered particles to gather.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
ax3l: good catch. will remove and add an inline comment above.

Source/FieldSolver/FiniteDifferenceSolver/MacroscopicEvolveE.cpp
ax3l: done (in `#ifdef WARPX_DIM_XZ`)

Source/Evolve/WarpXEvolve.cpp
ax3l: @roelof-groenewald in `Python_particle_reflection`, does that test scrape particles from a (domain) boundary without using EBs?

Maybe I should remove this then here again...

ax3l:
@roelof-groenewald @RemiLehe @WeiqunZhang the checksum changes that we see in
- 2D: `Python_prev_positions`
- 2D: `RepellingParticles`
- 2D: `background_mcc`
- 2D: `background_mcc_dp_psp`
- 3D: `NodalElectrostaticSolver`
- 3D: `relativistic_space_charge_initialization`
- 3D: `Python_projection_divb_cleaner_3d` -- unclear

appear very small - at the level of the MLMG precision (besides numbers that are close to zero, as usual).

The notable difference is now that they are compiled with EB, but `eb_enabled` is off at runtime.

Do you have time to check them with me, are they safe to reset?

Source/Evolve/WarpXEvolve.cpp
ax3l: Indeed that was a bug

ax3l:
If I run the `RepellingParticles` test in `development` and this PR, then I see in the tiny profiler that `FillBoundary_finish()` was called 3967x in `development` but only 3785x in this PR.

The method `MLNodeTensorLaplacian::Fsmooth()` was called 960x and now only 800x.
`amrex::Copy()` was 841x now is up to 897x calls.

I see that this PR now calls into `MLCGSolver::bicgstab` while `development` did not (or was in `Other`?).

[dev-branch.txt](https://github.com/user-attachments/files/16822444/dev-branch.txt)
[eb-branch.txt](https://github.com/user-attachments/files/16822445/eb-branch.txt)

ax3l:
@WeiqunZhang I think the `_USE_EB` macro usage in these AMReX files is not yet runtime-controlled:
- https://github.com/AMReX-Codes/amrex/blob/development/Src/LinearSolvers/MLMG/AMReX_MLCGSolver.H
- https://github.com/AMReX-Codes/amrex/blob/development/Src/LinearSolvers/MLMG/AMReX_MLMG.H

WeiqunZhang:
@ax3l Could you give this a try? https://github.com/AMReX-Codes/amrex/pull/4129

ax3l:
Awesome, that resolves the differences :tada: 
No more failing azure integration tests :sparkles:

ax3l:
Correctness is now ensured (see above & passing CI).

For performance, I benchmarked 3D runs on my laptop and an exclusive Perlmutter CPU node.
Somehow both machines were _very noisy today_, likely due to random overheads from MPI startups (`srun`) in the PM case and due to insufficient pinning on my laptop (and other background desktop noise).

From what I can see under the provided noise, there is now no significant performance drawback beyond the noise level for runs that compile with EB but do not use it.

## Data & Notebook

[2024-09-06_EB-runtime.zip](https://github.com/user-attachments/files/16905398/2024-09-06_EB-runtime.zip)

## Perlmutter CPU Node

Compiled and ran in [local](https://docs.nersc.gov/performance/io/dev-shm/) in-memory file system `/dev/shm/`.
Allocated via: `salloc --exclusive --nodes 1 --qos interactive --time 02:00:00 --constraint cpu --account=m4272 --ntasks-per-node=2`
Compiled via: `cmake --fresh -S . -B /dev/shm/build_cpu_rteb -DWarpX_COMPUTE=OMP -DWarpX_FFT=ON -DWarpX_DIMS="3" -DMPIEXEC_PREFLAGS="--cpu-bind=cores"`
Ran via: `ctest --test-dir /dev/shm/build_cpu_rteb -E AMReX -R run`

image
image

## Laptop

image
image

.github/workflows/source/check_inputs.py
```diff
@@ -27,7 +27,7 @@
                     # skip lines related to other function arguments
                     # NOTE: update range call to reflect changes
                     #       in the interface of 'add_warpx_test'
-                    for _ in range(3):
+                    for _ in range(2):  # skip over: dims, numprocs
```
EZoni, line 30: Technically this would be
```suggestion
                    for _ in range(2):  # skip over: dims, nprocs
```
so that the inline comment matches the correct argument name. 

I can include this in one of the CTest follow-up PRs if we do not want to rerun all CI tests again on this PR.

EZoni (approved):

ax3l:
## Memory footprint evaluation

I ran `Examples/Physics_applications/laser_acceleration/inputs_base_3d`, which does not use EBs at runtime, both on `development` (using `WarpX_EB=OFF`) and on this PR (using the new  default, `WarpX_EB=ON`).
I used 1 MPI process and one OMP process to avoid fluctuations.

### Development Memory Footprint

```
Pinned Memory Usage:
-------------------------------------------------------------------------
Name                                    Nalloc  Nfree    AvgMem    MaxMem
-------------------------------------------------------------------------
The_Pinned_Arena::Initialize()               1      1  1948   B  8192 KiB
ParticleContainer::addParticles           1984   1984  2411   B  5144 KiB
WarpX::InitData()                           52     52     8   B   131 KiB
main()                                      75     75   799   B   800   B
PhysicalParticleContainer::AddPlasma()       2      2   159   B   160   B
-------------------------------------------------------------------------

Cpu Memory Usage:
---------------------------------------------------------------------------
Name                                      Nalloc  Nfree    AvgMem    MaxMem
---------------------------------------------------------------------------
WarpX::InitData()                             40     40    54 MiB    54 MiB
Diagnostics::FilterComputePackFlush()         50     50   295 KiB  9448 KiB
ParticleContainer::BuildRedistributeMask      16     16  5395 KiB  7892 KiB
FabArray<FAB>::SumBoundary_nowait()         1200   1200   249 KiB  4275 KiB
WarpX::SyncCurrent()                        1200   1200   665 KiB  4275 KiB
WarpX::shiftMF()                            2364   2364  1274 KiB  4275 KiB
LaserParticleContainer::Evolve()               3      3   260 KiB   264 KiB
PhysicalParticleContainer::Evolve()            3      3   260 KiB   264 KiB
Filter::ApplyStencil(MultiFab)               709    709    12 KiB   106 KiB
---------------------------------------------------------------------------
```

### PR Memory Footprint

```
Pinned Memory Usage:
-------------------------------------------------------------------------
Name                                    Nalloc  Nfree    AvgMem    MaxMem
-------------------------------------------------------------------------
The_Pinned_Arena::Initialize()               1      1  1542   B  8192 KiB
ParticleContainer::addParticles           1984   1984  2337   B  5144 KiB
WarpX::InitData()                           52     52    11   B   131 KiB
main()                                      75     75   799   B   800   B
PhysicalParticleContainer::AddPlasma()       2      2   159   B   160   B
-------------------------------------------------------------------------

Cpu Memory Usage:
---------------------------------------------------------------------------
Name                                      Nalloc  Nfree    AvgMem    MaxMem
---------------------------------------------------------------------------
WarpX::InitData()                             40     40    54 MiB    54 MiB
Diagnostics::FilterComputePackFlush()         50     50   292 KiB  9448 KiB
ParticleContainer::BuildRedistributeMask      16     16  5392 KiB  7892 KiB
FabArray<FAB>::SumBoundary_nowait()         1200   1200   241 KiB  4275 KiB
WarpX::SyncCurrent()                        1200   1200   653 KiB  4275 KiB
WarpX::shiftMF()                            2364   2364  1253 KiB  4275 KiB
LaserParticleContainer::Evolve()               3      3   260 KiB   264 KiB
PhysicalParticleContainer::Evolve()            3      3   260 KiB   264 KiB
Filter::ApplyStencil(MultiFab)               709    709    12 KiB   106 KiB
---------------------------------------------------------------------------
```

## Summary

The maximums do not change, this looks good to me and we seem to effectively prevent allocating EB-related fields if we do not use them.

.github/workflows/source/check_inputs.py
ax3l: Thanks! Yes let's put in the follow-up to avoid rerunning the CI here.

WeiqunZhang (approved):

Source/BoundaryConditions/WarpXEvolvePML.cpp
```diff
@@ -301,27 +304,21 @@ WarpX::DampJPML (int lev, PatchType patch_type)
 
             amrex::ParallelFor( tjx, tjy, tjz,
                 [=] AMREX_GPU_DEVICE (int i, int j, int k) {
-#ifdef AMREX_USE_EB
-                    if(pml_lxfab(i, j, k) <= 0) return;
-#endif
+                    if (pml_lxfab && pml_lxfab(i, j, k) <= 0) { return; }
```
RemiLehe, line 307: Could we rename `pml_l[xyz]fab` as `eb_l[xyz]fab`?
This is because, with the `ifdef` not here anymore, it may not be clear to someone reading the code that this is related to embedded boundaries.
(We could also add a comment: `// Skip the field update if this gridpoint is inside the embedded boundary`.)

Source/Diagnostics/ReducedDiags/ChargeOnEB.cpp
```diff
@@ -87,6 +94,9 @@ void ChargeOnEB::ComputeDiags (const int step)
     // Judge whether the diags should be done
     if (!m_intervals.contains(step+1)) { return; }
 
+    if (!EB::enabled()) {
+        throw std::runtime_error("ComputeDiags only works when EBs are enabled at runtime");
```
RemiLehe, line 98:
```suggestion
        throw std::runtime_error("ChargeOnEB::ComputeDiags only works when EBs are enabled at runtime");
```

Source/FieldSolver/ElectrostaticSolver.cpp
```diff
@@ -323,64 +322,66 @@ WarpX::computePhi (const amrex::Vector<std::unique_ptr<amrex::MultiFab> >& rho,
...
-#if defined(AMREX_USE_EB)
-
     std::optional<ElectrostaticSolver::EBCalcEfromPhiPerLevel> post_phi_calculation;
-
-    // EB: use AMReX to directly calculate the electric field since with EB's the
-    // simple finite difference scheme in WarpX::computeE sometimes fails
-    if (electrostatic_solver_id == ElectrostaticSolverAlgo::LabFrame ||
-        electrostatic_solver_id == ElectrostaticSolverAlgo::LabFrameElectroMagnetostatic)
+#ifdef AMREX_USE_EB
+    // TODO: double check no overhead occurs on "m_eb_enabled == false"
```
RemiLehe, line 336: Is this comment obsolete? Does the variable `m_eb_enabled` still exist?

Source/FieldSolver/FiniteDifferenceSolver/MacroscopicEvolveE.cpp
```diff
@@ -174,10 +179,9 @@ void FiniteDifferenceSolver::MacroscopicEvolveECartesian (
         // Loop over the cells and update the fields
         amrex::ParallelFor(tex, tey, tez,
             [=] AMREX_GPU_DEVICE (int i, int j, int k){
-#ifdef AMREX_USE_EB
                 // Skip field push if this cell is fully covered by embedded boundaries
-                if (lx(i, j, k) <= 0) return;
-#endif
+                if (lx && lx(i, j, k) <= 0) { return; }
```
RemiLehe, line 183: Same comment as further up: we could rename this as `eb_lx` to make it clear that it is related to the embedded boundary.

Source/FieldSolver/ElectrostaticSolver.cpp
RemiLehe, line 336: Should we update this?

RemiLehe (commented):
Looks good, thanks for these changes!

Source/FieldSolver/ElectrostaticSolver.cpp
ax3l: Oh lol, yes we did check that :D

.github/workflows/source/check_inputs.py
ax3l: added in #5223

Issue #4853, opened 2024-04-11
Inquiry on Implementing Time Delay for Particle Injection in WarpX


1661145410:
I am currently working with the WarpX simulation software and have encountered an issue related to particle injection. The current implementation of particle injection in WarpX initializes all plasma components, including the initial plasma and the electron beams that are scheduled to be injected, at the zero moment after program initialization.

However, a limitation I have noticed is the inability to specify a time delay for subsequent particle injections. This becomes problematic when attempting to simulate scenarios where a laser pulse and an electron beam need to propagate in the same direction. Since the laser is initialized using an antenna approach, the electrons interact with the pulse in the opposite direction, altering their distribution when they enter the plasma region.

I am seeking guidance on how to address this issue. Is there a way to introduce a time delay for the injection of the electron beam in WarpX? This would enable a more accurate simulation of the interaction between the laser pulse and the electron beam, ensuring they propagate together as intended.

Please let me know if there are any existing features or workarounds that could help achieve this, or if you have any suggestions on how to approach this problem. Your assistance in this matter would be greatly appreciated.

Thank you for your time and support.

ax3l:
Hi @1661145410, thanks for reaching out!

We can in WarpX, e.g., we feed in neutral plasma in moving window simulations, or from surfaces, or via Python, inject new particles at any time. But I think we need to understand your scenario a bit better to guide you:
- Is this an **electromagnetic** wakefield simulation you are describing? (Then we might need to solve a Poisson equation when adding new particles)
- What is the geometry you try to model? (Co- or counter-propagating laser pulses and beams?)
  - What are the delays and spatial separations / sizes you try to model?

1661145410:
Hi @ax3l ,

Thank you for your prompt response and guidance. I appreciate your willingness to help with our simulation scenario.

Regarding our simulation, we are indeed using two crossing laser beams interacting with a plasma, where the wakefields play a significant role. Specifically, the electron beam in our simulation is incident forty optical cycles later than the lasers.

Here are part parameters for the plasma and lasers in our setup:

geometry.dims = 2					# simulation dimention
geometry.prob_lo     = -20.e-6   -45.e-6		# physical domain,lower corner
geometry.prob_hi     =  20.e-6    115.e-6		# upper corner

#################################
############ PLASMA #############
#################################
my_constants.dens = 1.21e27				#
particles.species_names = plasma_e plasma_p beam	#	

plasma_e.species_type = electron			#
[...]

PR #4868, opened 2024-04-17
Add function to set value of `do_not_push`

roelof-groenewald:
I have a use case where I want to not evolve a certain species for a specific amount of time. This PR allows me to turn on the evolution of that species at a desired time.

This functionality (with Python callbacks) can also be used to achieve the desired effect from #4853 (might need to add a similar function for `do_not_deposit`).

### DIFF
--- a/Source/Particles/WarpXParticleContainer.H
+++ b/Source/Particles/WarpXParticleContainer.H
@@ -426,6 +426,8 @@ public:
     virtual QuantumSynchrotronEngine* get_quantum_sync_engine_ptr () const {return nullptr;}
 #endif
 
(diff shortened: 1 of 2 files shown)


Source/Particles/WarpXParticleContainer.H
```diff
@@ -442,7 +444,7 @@ protected:
     //! instead of gathering fields from the finest patch level, gather from the coarsest
     bool m_gather_from_main_grid = false;
 
-    int do_not_push = 0;
+    bool do_not_push = 0;
```
roelof-groenewald:
```suggestion
    bool do_not_push = false;
```

lucafedeli88 (approved):
Thanks!

PR #4872, opened 2024-04-19
Remove calls to WarpX::GetInstance() from MacroscopicProperties.cpp

lucafedeli88:
This PR removes calls to `WarpX::GetInstance()` from `MacroscopicProperties.cpp`


ax3l (approved):
Thank you! :sparkles:

PR #4873, opened 2024-04-19
Remove call to WarpX::GetInstance() from FilterCreateTransformFromFAB.H

lucafedeli88:
This PR removes a call to `WarpX::GetInstance()` from `FilterCreateTransformFromFAB.H` .


ax3l (approved):
That looks very good!

PR #4874, opened 2024-04-19
Remove "friend class PML" from WarpX class

lucafedeli88:
After this PR, PML class and WarpX class won't be friends anymore ;-).
Jokes apart, there's no need for the PML class to be a friend class of the WarpX class (i.e., a class able to access the private members of the other). I believe that it only complicates the relationship between the two classes, which are already a bit intricate.


ax3l (approved):
:broken_heart: :laughing:

PR #4875, opened 2024-04-19
Move FieldType enum from WarpX.H into a separate header

lucafedeli88:
This PR moves the `FieldType` enum struct from `WarpX.H` to `FieldSolver/Fields.H`, inside a newly created `warpx::fields` namespace. All the other changes (inclusion of the new header, `using namespace warpx::fields` ...) are a consequence of this change.

The immediate advantage of this PR is a slight simplification of the `WarpX.H` header field. However, the main interest of this PR is that it can be considered as a first step towards a more modular approach for fields, similarly to what is done for particles. The final goal (to be discussed) could be to have a sort of MultiFieldContainer member of the WarpX class.

### DIFF
--- a/Source/BoundaryConditions/PML.cpp
+++ b/Source/BoundaryConditions/PML.cpp
@@ -10,6 +10,7 @@
 
 #include "BoundaryConditions/PML.H"
 #include "BoundaryConditions/PMLComponent.H"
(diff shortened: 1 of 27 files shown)


ax3l (approved):
Thank you, that's great!

ax3l:
Follow-up: #4890

PR #4876, opened 2024-04-19
Make sure that include guard names start with WARPX_

lucafedeli88:
In order to avoid possible collisions with other libraries, this PR changes  some names used for include guards in such a way that they all start with `WARPX_`  (with the exception of those that are contained inside `ablastr`)


ax3l (approved):
Thx!

PR #4880, opened 2024-04-22
[WIP] Use singleton to store warpx profiler parameter "do_device_synchronize"

lucafedeli88:
Right now `Source/Utils/WarpXProfilerWrapper.H` includes the huge header `WarpX.H`, just because we need to pass `WarpX::do_device_synchronize` to `ABLASTR_PROFILE`.

This PR avoids this issue by introducing a new singleton class `warpx::profiler::ProfileSettings` (singletons make sense for loggers and profilers in my opinion) that stores this parameter (it might store additional parameters in the future). So, during the initialization, the WarpX class simply calls the function `warpx::profiler::ProfileSettings::InitProfileSettings(do_device_synchronize);` , and the parameter `do_device_synchronize` is stored inside `ProfileSettings`. This allows us to avoid the transitive inclusion of `WarpX.H` every time the profiler is used. 
In some cases, including `WarpX.H` (or other headers included by `WarpX.H`) is now required because `Utils/WarpXProfilerWrapper.H"` no longer includes such libraries.

### DIFF
--- a/Source/Diagnostics/ParticleIO.cpp
+++ b/Source/Diagnostics/ParticleIO.cpp
@@ -17,6 +17,7 @@
 #include "Utils/TextMsg.H"
 #include "Utils/WarpXConst.H"
 #include "Utils/WarpXProfilerWrapper.H"
+#include "WarpX.H"
 
 #include <ablastr/utils/text/StreamUtils.H>
 
--- a/Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.H
+++ b/Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.H
@@ -22,6 +22,8 @@
 #include <AMReX_Array.H>
 #include <AMReX_REAL.H>
 
+#include <optional>
+
 /**
  * \brief This class contains the parameters needed to evaluate hybrid field
  * solutions (kinetic ions with fluid electrons).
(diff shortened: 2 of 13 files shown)


AlexanderSinn:
A while ago, I fixed the synchronizes in amrex::TinyProfiler itself to be ordered correctly with respect to collecting the time https://github.com/AMReX-Codes/amrex/pull/3763 . This means that instead of using the custom added syncs in ABLASTR the ones build in to amrex::TinyProfiler can be used now (Note that a small difference with this is that profiles in amrex would also get syncronizes). So most of the profiler wrapper stuff in WarpX/ABLASTR can be replaced with a small function that passes the input parameter from  `warpx.do_device_synchronize = 1` to `tiny_profiler.device_synchronize_around_region = 1`.

lucafedeli88:
> A while ago, I fixed the synchronizes in amrex::TinyProfiler itself to be ordered correctly with respect to collecting the time [AMReX-Codes/amrex#3763](https://github.com/AMReX-Codes/amrex/pull/3763) . This means that instead of using the custom added syncs in ABLASTR the ones build in to amrex::TinyProfiler can be used now (Note that a small difference with this is that profiles in amrex would also get syncronizes). So most of the profiler wrapper stuff in WarpX/ABLASTR can be replaced with a small function that passes the input parameter from  `warpx.do_device_synchronize = 1` to `tiny_profiler.device_synchronize_around_region = 1`.

Thanks for the information, @AlexanderSinn ! I was not aware of that, sorry... I think that if we can safely remove a wrapper layer we should do that. @ax3l , is the profiler in ABLASTR used also in ImpactX? If it is the case we may need to coordinate on this.

PR #4881, opened 2024-04-22
AMReX/pyAMReX/PICSAR: Weekly Update

ax3l:
Weekly update to latest AMReX.
Weekly update to latest pyAMReX (no changes).
Weekly update to latest PICSAR (no changes).

```console
./Tools/Release/updateAMReX.py
./Tools/Release/updatepyAMReX.py
./Tools/Release/updatePICSAR.py
```

### DIFF
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -115,7 +115,7 @@ jobs:
         which nvcc || echo "nvcc not in PATH!"
 
         git clone https://github.com/AMReX-Codes/amrex.git ../amrex
-        cd ../amrex && git checkout --detach 7813bb7d6bc6cea658edac1f2bde046c7d174168 && cd -
+        cd ../amrex && git checkout --detach b752027c1aebdfb4be339b1e30932b4108286a7a && cd -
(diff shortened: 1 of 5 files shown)


lucafedeli88:
Tests are failing due to a recent change in the `regtest.py` script (https://github.com/AMReX-Codes/regression_testing/blame/c17887e4afabf83a56723d30899e13264439ba8b/regtest.py#L1213).  The new logic checks for files matching the pattern `re.match(f"{test.name}.*_(plt|chk)[0-9]+", pfile))` , and throws an error if no file is found. However, in WarpX tests, we don't always generate such files. In some cases the output consists in just reduced diagnostics, while in others it consists in OpenPMD files.

ax3l:
Thanks, Luca!

@zingale is there a way we could control this via an option?

roelof-groenewald (commented):
We should probably get the CI test failures sorted before merging this to make sure all the tests would actually still pass.

roelof-groenewald:
It's interesting that the MacOS build isn't failing in this PR as it is for other PRs.

ax3l:
Might be a half-propagated homebrew update that is still being rolled out.

roelof-groenewald (approved):
All the tests are passing so we're good to go 🎉

PR #4882, opened 2024-04-22
Downgrade Reg Tests

ax3l:
Downgrade the regression testing scripts to work around
  https://github.com/AMReX-Codes/regression_testing/issues/136


roelof-groenewald (approved):
I'll approve this since @lucafedeli88 is probably done for the day and this will be good to get into development sooner rather than later 😄

PR #4884, opened 2024-04-23
Further downgrade regression_testing

lucafedeli88:
Downgrade the regression testing scripts to work around https://github.com/AMReX-Codes/regression_testing/issues/136 .
In https://github.com/ECP-WarpX/WarpX/pull/4882 we should have downgraded the script by one further commit.


roelof-groenewald (approved):
Thanks @lucafedeli88!
We need @ax3l to force merge this since the MacOS build is still failing.

ax3l (approved):

PR #4887, opened 2024-04-23
Fix macOS CI failure

roelof-groenewald:
Just a quick experiment to see if this fixes the maxOS build failure.

Basically the error seen appears to say that `pip install ...` cannot be called from outside a virtual environment on macOS anymore.


ax3l (approved):
Thank you! :)

PR #4888, opened 2024-04-24
[WIP] Use more compile time options in PushPX kernel

lucafedeli88:


AlexanderSinn:
Looks good! In HiPACE++ we use kernels with up to 48 compile-time variants, so I agree WarpX can use more. I would like to suggest two small changes:
* Use `CompileTimeOptions<false, true>` or  `CompileTimeOptions<0, 1>` instead of the enums. The compile time options use int internally, so the enums provide no extra type safety. To me, it looks like the enums add extra confusion, although this could be debatable.
* Use `if` instead of`if constexpr`, but still with the `..._control` check inside it. In my testing, this gives still all the performance benefits since compilers can easily optimize out `if (false) {...}` while avoiding the lambda campture issues that arise with NVCC.

lucafedeli88:
I will close this for now (no plan to work on the topic soon on my side)

PR #4889, opened 2024-04-24
Profiler: use amrex logic for device synchronization instead of custom solution

lucafedeli88:
This PR is a follow-up of the discussion with @AlexanderSinn  in https://github.com/ECP-WarpX/WarpX/pull/4880 .

Currently we implement in `ablastr`  some logic to ensure device synchronization for the `TinyProfiler`. 
However, after @AlexanderSinn 's PR https://github.com/AMReX-Codes/amrex/pull/3763  , we can rely on what is already implemented in AMReX in order to achieve that.

Therefore, this PR does the following:

- The AMReX logic is now used for device synchronization of the profiler. This is achieved by doing

```
 //https://github.com/AMReX-Codes/amrex/pull/3763
#ifdef AMREX_USE_GPU
        bool warpx_do_device_synchronize = true;
#else
        bool warpx_do_device_synchronize = false;
#endif
        pp_warpx.query("do_device_synchronize", warpx_do_device_synchronize);
        bool do_device_synchronize = warpx_do_device_synchronize;
        amrex::ParmParse pp_tiny_profiler("tiny_profiler");
        if (pp_tiny_profiler.queryAdd("device_synchronize_around_region", do_device_synchronize) )
        {
                do_device_synchronize == warpx_do_device_synchronize,
                "tiny_profiler.device_synchronize_around_region overrides warpx.do_device_synchronize.");
        }
``` 
inside WarpXAMReXInit.cpp`.

- `WarpXProfilerWrapper.H`  and `ablastr/profiler/ProfilerWrapper.H` are now ultra-thin wrappers around the macros provided by AMReX. 

- `WarpX::do_device_synchronize`  static variable is eliminated (with some minor changes to functions that required this variable as a parameter).

- All the other changes are inclusions of dependencies that were previously included transitively via `#include "WarpX.H"` inside `WarpXProfilerWrapper.H`.

### DIFF
--- a/Source/Initialization/WarpXAMReXInit.cpp
+++ b/Source/Initialization/WarpXAMReXInit.cpp
@@ -47,7 +47,7 @@ namespace {
             pp_amr.add("blocking_factor", 1);
         }
 
-        //https://github.com/AMReX-Codes/amrex/pull/3763
+        //See https://github.com/AMReX-Codes/amrex/pull/3763
 #ifdef AMREX_USE_GPU
         bool warpx_do_device_synchronize = true;
 #else
--- a/Source/Initialization/WarpXAMReXInit.cpp
+++ b/Source/Initialization/WarpXAMReXInit.cpp
@@ -7,9 +7,12 @@
 
 #include "Initialization/WarpXAMReXInit.H"
 
+#include "Utils/TextMsg.H"
+
 #include <AMReX.H>
 #include <AMReX_ccse-mpi.H>
 #include <AMReX_ParmParse.H>
+#include <AMReX_TinyProfiler.H>
 
 namespace {
     /** Overwrite defaults in AMReX Inputs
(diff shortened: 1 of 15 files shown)


AlexanderSinn:
Looks Great!

ax3l:
@lucafedeli88 thank you for this! Can you rebase pls? :)

lucafedeli88:
> @lucafedeli88 thank you for this! Can you rebase pls? :)

that's done, @ax3l  !

lucafedeli88:
ping, @AlexanderSinn and @ax3l  ;-) !

Source/Initialization/WarpXAMReXInit.cpp
```diff
@@ -44,6 +47,22 @@ namespace {
...
+        //See https://github.com/AMReX-Codes/amrex/pull/3763
+#ifdef AMREX_USE_GPU
+        bool warpx_do_device_synchronize = true;
+#else
+        bool warpx_do_device_synchronize = false;
+#endif
+        pp_warpx.query("do_device_synchronize", warpx_do_device_synchronize);
+        bool do_device_synchronize = warpx_do_device_synchronize;
+        amrex::ParmParse pp_tiny_profiler("tiny_profiler");
+        if (pp_tiny_profiler.queryAdd("device_synchronize_around_region", do_device_synchronize) )
```
AlexanderSinn, line 59: If I understand this correctly, this would error out every time someone wants to run a simulation using a used input parameters file.

AlexanderSinn (approved):

ax3l (approved):
LGTM. This new default might hide a couple of race conditions that we introduce in the future :sweat_smile: 

@lucafedeli88 @AlexanderSinn , did you have a chance to double check performance stays the same?

PR #4895, opened 2024-04-25
Add two missing includes (found with iwyu)

lucafedeli88:
This PR adds two missing include directives.


ax3l (approved):
Thank you :)

Source/Diagnostics/OpenPMDHelpFunction.cpp
```diff
@@ -6,8 +6,11 @@
  * License: BSD-3-Clause-LBNL
  */
 #include "OpenPMDHelpFunction.H"
+
 #include "Utils/TextMsg.H"
 
+#include <map>
```
ax3l, line 12: Oops, `<string>`, too.

PR #4902, opened 2024-04-29
`isort`: First Party Modules

ax3l:
Declare `amrex`, `picmistandard` and `pywarpx` our first party modules, even if they are not installed. This resolves inconsistencies when sorting imports in developer or remote runs (pre-commit.com CI).


lucafedeli88 (approved):
Thanks for this PR!

PR #4907, opened 2024-04-30
Fix Python installation instructions on Perlmutter

RemiLehe:
The name of the Python virtual environments was incorrect.


ax3l (approved):
Thank you! :pray:

PR #4909, opened 2024-05-03
warpx.roundrobin_sfc: A runtime parameter to control distribution mapping

WeiqunZhang:
The default is false. If it's true, AMReX's RRSFC strategy will be used to override the default SFC strategy used by amrex::AmrCore. The motivation for this is that this might mitigate the load imbalance issue during initialization by avoiding putting neighboring boxes on the same process.


Source/WarpX.cpp
```diff
@@ -1157,6 +1159,8 @@ WarpX::ReadParameters ()
             maxLevel() == 0 || !do_current_centering,
             "Finite-order centering of currents is not implemented with mesh refinement"
         );
+
+        pp_warpx.query("roundrobin_sfc", roundrobin_sfc);
```
ax3l: This might need a user-facing entry in `parameters.rst`.

Source/WarpX.cpp
```diff
@@ -219,6 +219,8 @@ bool WarpX::do_device_synchronize = true;
 bool WarpX::do_device_synchronize = false;
 #endif
 
+bool WarpX::roundrobin_sfc = false;
```
ax3l: We need to add control for this from PICMI, too.

Source/WarpX.cpp
WeiqunZhang: done

Source/WarpX.cpp
WeiqunZhang: done

dpgrote:
Looks good. Would his affect the runtime, slowing the MPI communication since neighboring boxes will be far away from each other?

ax3l:
@WeiqunZhang sorry for the delay, can you please rebase?

Source/WarpX.cpp
```diff
@@ -3515,3 +3519,17 @@ WarpX::getField(FieldType field_type, const int lev, const int direction) const
 {
     return *getFieldPointer(field_type, lev, direction);
 }
+
+amrex::DistributionMapping
+WarpX::MakeDistributionMap (int lev, amrex::BoxArray const& ba)
+{
+    if (roundrobin_sfc) {
```
ax3l, line 3605: I would just query the `ParmParse` here again instead of creating a static member variable.

Source/WarpX.cpp
```diff
@@ -219,6 +219,8 @@ bool WarpX::do_device_synchronize = true;
 bool WarpX::do_device_synchronize = false;
 #endif
 
+bool WarpX::roundrobin_sfc = false;
+
```
```suggestion
```

Source/WarpX.H
```diff
@@ -1212,6 +1212,20 @@ protected:
...
+    //! If this is true, AMReX's RRSFC strategy is used to make
+    //! DistributionMapping in the virtual function MakeDistributionMap. The
+    //! default is false. Note that the DistributionMapping made by the
+    //! MakeDistributionMap function could still be overridden by load
+    //! balancing. In the RRSFC strategy, the Round robin method is used to
+    //! distribute Boxes orderd by the space filling curve. This might help
+    //! avoid some processes running out of memory due to having too many
+    //! particles during initialization.
+    static bool roundrobin_sfc;
+
```
ax3l: Let's move this comment block as the doxygen block after the title of `MakeDistributionMap` or inside its body before the `if`?

I would just query the `ParmParse` here again instead of creating a static member variable.
```suggestion
```

n01r:
Thank you, @WeiqunZhang!
I will test/use this down the line. :)

ax3l:
Thank you so much!

After checking back with our collaborators, we currently do not have a reproducer at hand, but I'll include it in the coming release so we can try it when we should find one again.

@n01r might have one we can try this on, too.

ax3l (approved):
Thank you, @WeiqunZhang! :sparkles:

ax3l:
@dpgrote 
> Looks good. Would his affect the runtime, slowing the MPI communication since neighboring boxes will be far away from each other?

The idea is to overcome out-of-memory in the init routines (e.g., for LPI sims with a lot of particles in close-by ranks) and then apply load balancing based on a user-defined cost (and distribution) model, as usual.

PR #4911, opened 2024-05-03
[WIP] PICMI Usage in HPC Batch Scripts

ax3l:
Try to incorporate the same level of details for PICMI as for executable usage in HPC batch scripts.

- [ ] discuss best approach
- [ ] apply to all machines

### DIFF
--- a/Tools/machines/perlmutter-nersc/perlmutter_gpu.sbatch
+++ b/Tools/machines/perlmutter-nersc/perlmutter_gpu.sbatch
@@ -24,9 +24,21 @@
 #SBATCH -o WarpX.o%j
 #SBATCH -e WarpX.e%j
 
-# executable & inputs file or python interpreter & PICMI script here
-EXE=./warpx
-INPUTS=inputs
+# PICMI Python script or executable?
+USE_PICMI=true
+
+if [[ "${USE_PICMI}" = true ]]
+then
+    EXE=python3
+    INPUTS=PICMI_script.py
+    #   for GPU-aware MPI in PICMI, set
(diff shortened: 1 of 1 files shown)


Tools/machines/perlmutter-nersc/perlmutter_gpu.sbatch
```diff
@@ -24,9 +24,21 @@
 #SBATCH -o WarpX.o%j
 #SBATCH -e WarpX.e%j
 
-# executable & inputs file or python interpreter & PICMI script here
-EXE=./warpx
-INPUTS=inputs
+# PICMI Python script or executable?
+USE_PICMI=true
```
dpgrote, line 28: This is a good idea. Maybe set `USE_PICMI=false` here so the default behavior is the same as previously.

Tools/machines/perlmutter-nersc/perlmutter_gpu.sbatch
```diff
@@ -24,9 +24,21 @@
...
+
+if [[ "${USE_PICMI}" = true ]]
+then
+    EXE=python3
+    INPUTS=PICMI_script.py
+    #   for GPU-aware MPI in PICMI, set
+    #   Simulation(..., warpx_amrex_use_gpu_aware_mpi=True)
```
dpgrote, line 35: Something I've thought about is allowing setting of input parameters from the command line when using Python. Something like
```python3 PICMI_script.py --inputs amrex.use_gpu_aware_mpi=1```
But I'm not sure it's a good idea. For one, it would conflict with any argument parsing that the input script might do.

PR #4914, opened 2024-05-06
Diagnostics: Fix Position Selection

### LINKED ISSUE #4841: Particle diagnostic silently writes zero position if x y z is not explicitly specified

ax3l:
- [x] If a user specifies `<diag_name>.<species_name>.variables` without including particle position arguments we now warn.
- [x] Update missing documentation.
- [x] Close #4841: Do not write zero-positions for deselected position components. - [SetupPos](https://github.com/ECP-WarpX/WarpX/blob/development/Source/Diagnostics/WarpXOpenPMD.cpp#L711) if not needed
  - [x] Test: all positions, some positions, no positions, `none`

Follow-up to #4653


kale-j (approved):
This is now much clearer!
I tested in 2D and 3D -- seems to work as expected

RemiLehe:
Thanks @ax3l ! Would you have time to check the `clang-tidy` error?

ax3l:
@kale-j thank you for testing! Glad this works well! Thanks again for reporting this :sparkles: 

@RemiLehe absolutely - fixed now :)

ax3l:
@RemiLehe when you have a second, can you approve this for the release? :)

ax3l:
Merging, so we can do the release 24.05 :)
It is tested by @kale-j and me and only small fixes where left after Remi saw it :)

Follow-ups that we found during testing are documented in #4917

Source/Diagnostics/ParticleDiag/ParticleDiag.cpp
```diff
@@ -28,12 +30,17 @@ ParticleDiag::ParticleDiag(
...
         // If only specific variables have been specified, fill m_plot_flags with zero and only set
         // requested variables to one
         std::fill(m_plot_flags.begin(), m_plot_flags.end(), 0);
+        bool contains_positions = false;
         if (variables[0] != "none"){
-            const std::map<std::string, int> existing_variable_names = pc->getParticleComps();
+            std::map<std::string, int> existing_variable_names = pc->getParticleComps();
+#ifdef WARPX_DIM_RZ
+            // we reconstruct to Cartesian x,y,z for RZ particle output
+            existing_variable_names["y"] = PIdx::theta;
```
ax3l, line 42: This is a hack to have `y` enabled and de-selecting was already broken before.
#4917 will do a general overhaul of all this logic and remove this assignment.

PR #4916, opened 2024-05-07
Resetting charge buffers to zero.

oshapoval:
This PR sets charge buffer to zero at each time iteration inside calls of `AddSpaceCharge()` and `AddSpaceChargeFieldLabFrame()`. Otherwise, if MR is on at every level `lev < finest_lev` fields will be accumulated (see plots below).

To do:

- [x] add CI test.

Screenshot 2024-05-06 at 5 13 46 PM

Thanks @RemiLehe for helping with this PR.


Examples/Tests/electrostatic_sphere/inputs_3d
oshapoval, line 1: Note, that those changes need to be reverted, since they are already reflected in `ElectrostaticSphereLabFrame_MR_emass_10` CI test set up as runtime parameters.

RemiLehe:
Thanks for this PR @oshapoval. Could you fix the failing automated test?

ax3l (approved):
Thank you both, looks good! :sparkles:

ax3l:
For the HIP CI issue, please see #4927

PR #4919, opened 2024-05-09
Add CI test for velocity coincidence thinning

roelof-groenewald:
This is a follow-up to https://github.com/ECP-WarpX/WarpX/pull/4886#pullrequestreview-2036309010.

It also includes a fix to `picmi.py` where currently the `resampling_algorithm_delta_u` input parameter is set to `[None, None, None]` by default for all species instead of simply `None`.


dpgrote (approved):
Thanks for adding this!

As an aside, my own preference is to add new input files rather than modify input parameters in WarpX-tests.ini. I see the examples as demonstrations for users of how to specify input parameters. Anything in WarpX-tests.ini is essentially hidden from the users.

dpgrote (approved):
Thanks!

PR #4920, opened 2024-05-09
Add particle boundary none

dpgrote:
This adds "None" as a particle boundary option, to correspond to the same option for the fields. This now must be used for the lower particle boundary in radius when it is specified when using RZ.


roelof-groenewald:
This is a good change since I've seen users be confused by what the lower boundary specification in RZ does. Looks like the RZ tests/examples just need to be updated still.

roelof-groenewald (approved):

PR #4921, opened 2024-05-09
Code clean-up for binary collisions

roelof-groenewald:
Added new function `BinaryCollisionUtils::remove_weight_from_colliding_particle` to reduce code duplication.

Switched to passing particle tiles to functors rather than SoA since for some reason (?) the weight of the particles are not correctly decreased when passing SoAs.


ax3l (approved):
Looks good!

> Switched to passing particle tiles to functors rather than SoA since for some reason (?) the weight of the particles are not correctly decreased when passing SoAs.

We might want to dig this, there might be an accidental copy somewhere here.

Source/Particles/Collision/BinaryCollision/Coulomb/PairWiseCoulombCollisionFunc.H
```diff
@@ -84,7 +85,7 @@ public:
             index_type const I2s, index_type const I2e,
             index_type const* AMREX_RESTRICT I1,
             index_type const* AMREX_RESTRICT I2,
-            ParticleTileType ptile1, ParticleTileType ptile2,
```
ax3l: This was the issue in the old code. `ParticleTileType` compared to `ParticleTileDataType` does copy indeed.

PR #4922, opened 2024-05-10
Fix bug with ES solver and MR: `E_aux=E_fp` in `UpdateAuxilaryData`

oshapoval:
This PR fixes a bug with the ES solver and MR. This issues was initially observed by @aeriforme. Worked on the issue with @RemiLehe, @EZoni and @WeiqunZhang .

Problem: observed doubling of the E field for `lev>0`.

Solution: when ES solver is used we have `E_cp[levl] =0` and we should update the auxiliary data as `E_aux[lev] = E_fp[lev]` (inside `WarpX::UpdateAuxilaryDataStagToNodal()` and `WarpX::UpdateAuxilaryDataSameType()` rather than `E_aux[lev] = E_fp[lev] - E_cp[lev] + E_aux[lev-1]` in contrast to the EM solver. Otherwise it would result in doubling of E field we observed.

To do:
- [x] do the same in `WarpX::UpdateAuxilaryDataStagToNodal ()`
- [x] add CI test with `amr.max_level = 1` (reused CI test `electrostatic_sphere_eb_mr`)
- [x] fix scraping tests

Results after the fix:
- With MR patch occupying the whole simulation box:
Figure 32-4
- With MR patch smaller than the simulation box:
Figure 31-6


Source/Parallelization/WarpXComm.cpp
```diff
@@ -673,11 +791,6 @@ WarpX::FillBoundaryE_avg (int lev, PatchType patch_type, IntVect ng)
 {
     if (patch_type == PatchType::fine)
     {
-        if (do_pml && pml[lev]->ok())
-        {
-            WARPX_ABORT_WITH_MESSAGE("Averaged Galilean PSATD with PML is not yet implemented");
-        }
-
```
EZoni: Just double checking, are this and the following similar changes (I counted four of them) related to this PR? The PR is a bug fix for the electrostatic solver with mesh refinement, while these changes are removing abort statements related to averaged Galilean PSATD with PMLs. Were they maybe left over from a different branch and/or should they be implemented in a dedicated PR?

Examples/Tests/electrostatic_sphere_eb/inputs_rz_mr
```diff
@@ -30,3 +30,4 @@ diagnostics.diags_names = diag1
 diag1.intervals = 1
 diag1.diag_type = Full
 diag1.fields_to_plot = Er phi
+diag1.format = openpnd
```
EZoni: Typo:
```suggestion
diag1.format = openpmd
```

Source/ablastr/fields/PoissonSolver.H
```diff
@@ -292,8 +302,8 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
                 0,
                 0,
                 1,
-                ng,
-                ng,
+                amrex::IntVect(0),
+                amrex::IntVect(1),
```
EZoni: I noticed that resetting the first argument (corresponding to `src_nghost` in the definition of `ParallelCopy`) to `amrex::IntVect(1)`, as it was before, fixes the "erroneous arithmetic operation" error that was occurring in the tests `ElectrostaticSphereLabFrame_MR_emass_10` and `ElectrostaticSphereEB_RZ_MR`.

What was the reason for deciding to change the code here and pass 0 ghost cells for the source MultiFab, as opposed to 1 ghost cell for the destination MultiFab?

```suggestion
                amrex::IntVect(1),
                amrex::IntVect(1),
```

EZoni (commented):
I found three small issues, detailed below, that were causing the failure of two CI tests. I will push a commit to fix those and see if this is enough to make all CI tests pass successfully.

Update: bug fixes pushed in [e5f0de6](https://github.com/ECP-WarpX/WarpX/pull/4922/commits/e5f0de68abed339940155f5a11a8f9b102c65660).

Examples/Tests/electrostatic_sphere_eb/analysis_rz.py
```diff
@@ -20,50 +20,93 @@
...
+    print('max error of Er = ', errmax_Er)
+    print('tolerance = ', tolerance)
+
+ts = OpenPMDTimeSeries('./diags/diag1')
+for level in [0,1]:
+    print('At level =', level, ':')
+    get_error_per_lev(ts,level)
 
 test_name = os.path.split(os.getcwd())[1]
 checksumAPI.evaluate_checksum(test_name, fn, do_particles=False)
```
EZoni: This is the third issue. I believe here we need to pass the argument `output_format=openpmd` in order for the checksums to be evaluated correctly.

EZoni:
I'm investigating the failure of the scraping CI tests (which appears as "erroneous arithmetic operation") and what I can report so far is that it probably due to uninitialized values in the MultiFabs `Efield_cax` and `Bfield_cax`. 

It's not clear to me yet why we are leaving some of those values untouched, hence uninitialized, with the code changes here. 

Regardless of this, we might want to think whether we would like to initialize those MultiFabs to 0.0 as we do for the corresponding MultiFabs on level 0.

Merged the code changes in #5049 to verify that they are sufficient to fix the issue. Will rebase once more after #5049 is merged.

EZoni:
@oshapoval @RemiLehe 
The scraping tests seem fixed now, thanks to #5049, which is a separate bug fix that we should merge first.
This said, before merging this PR, I will do another round of review and possibly clean up some code.

Examples/Tests/electrostatic_sphere_eb/analysis_rz_mr.py
```diff
@@ -0,0 +1,112 @@
...
+# This file is part of WarpX.
+#
+# License: BSD-3-Clause-LBNL
+
+# This script tests the embedded boundary in RZ.
+# A cylindrical surface (r=0.1) has a fixed potential 1 V.
+# The outer surface has 0 V fixed.
+# Thus the analytical solution has the form:
+# phi(r) = A+B*log(r), Er(r) = -B/r.
+
```
RemiLehe, line 14: Could you add a quick summary explaining how this differs from `analysis_rz.py` (i.e. what aspects of MR does this check?)

RemiLehe: Also, could you confirm that this test fails with the current `development` version of WarpX?

EZoni:
Note that in [eb3d193](https://github.com/ECP-WarpX/WarpX/pull/4922/commits/eb3d19398306b8f102ab6429a5a92f9b8ba09a4b) I replaced two `MFIter`/`ParallelFor` loops with simple calls to `MultiFab::Copy`, which I think should be equivalent. This way we don't have to go through a trivial `interp` function that simply copies data from input to output.

@WeiqunZhang 
Do you agree with the code changes in the commit above?

Examples/Tests/electrostatic_sphere_eb/analysis_rz_mr.py
EZoni: When I started working on this PR I separated the analysis script into two because it had been modified to process the openpmd results of the new test (which uses openpmd only), thus failing on the plotfile results of the old test (which uses plotfile only). I can try to see if it is possible to re-merge them together.

EZoni: And yes, I just ran the test with the latest `development` branch and the analysis fails as the error on the radial electric field on level 1 is larger than the tolerance.
```
At level = 0 :
max error of phi =  0.004078274696046572
max error of Er =  0.0025379720096999625
tolerance =  0.0041
At level = 1 :
max error of phi =  0.00028625211446039984
max error of Er =  1.0028845814970835
tolerance =  0.0041
  File "/tmp/ci-gymyMO8XZo/rt-WarpX/WarpX-tests/2024-07-17/ElectrostaticSphereEB_RZ_MR/./analysis_rz_mr.py", line 109, in <module>
    get_error_per_lev(ts,level)
  File "/tmp/ci-gymyMO8XZo/rt-WarpX/WarpX-tests/2024-07-17/ElectrostaticSphereEB_RZ_MR/./analysis_rz_mr.py", line 104, in get_error_per_lev
    assert(errmax_Er < tolerance)
AssertionError
```

Examples/Tests/electrostatic_sphere_eb/analysis_rz_mr.py
```diff
@@ -0,0 +1,112 @@
...
+                errmax_Er = err
+
+    print('max error of phi = ', errmax_phi)
+    print('max error of Er = ', errmax_Er)
+    print('tolerance = ', tolerance)
+    assert(errmax_phi < tolerance)
+    assert(errmax_Er < tolerance)
+
+ts = OpenPMDTimeSeries(fn)
+for level in [0,1]:
```
EZoni: It should be possible to rewrite this in a general way such that one does not need to know the number of mesh refinement levels in the data:
```suggestion
level_fields = [field for field in ts.avail_fields if 'lvl' in field]
nlevels = 0 if level_fields == [] else int(level_fields[-1][-1])
for level in range(nlevels+1):
```
Inspired by what we do here:
https://github.com/ECP-WarpX/WarpX/blob/a13a11d5541bc245634cdc793bf997bfa12fbcc1/Regression/Checksum/checksum.py#L128-L131.

Examples/Tests/electrostatic_sphere_eb/analysis_rz_mr.py
```diff
@@ -0,0 +1,112 @@
...
+        er, info = ts.get_field('E', 'r', iteration=0, plot=False)
+        phi, info = ts.get_field('phi', iteration=0, plot=False)
+
+    return er, phi, info
+
+def get_error_per_lev(ts,level):
+    er, phi, info = get_fields(ts, level)
+
+    nj = len(er[0,:])//2
+    dr = info.r[1] - info.r[0]
```
EZoni: `dr` can be accessed directly from `info`:
```suggestion
    dr = info.dr
```

Examples/Tests/electrostatic_sphere_eb/analysis_rz_mr.py
```diff
@@ -0,0 +1,112 @@
...
+    er = []
+    phi = []
+    info = []
+    if level == 1:
+        er, info = ts.get_field('E_lvl1', 'r', iteration=0, plot=False)
+        phi, info = ts.get_field('phi_lvl1', iteration=0, plot=False)
+
+    elif level==0:
+        er, info = ts.get_field('E', 'r', iteration=0, plot=False)
+        phi, info = ts.get_field('phi', iteration=0, plot=False)
```
EZoni: This can be generalized with respect to mesh refinement levels, too:
```suggestion
    if level == 0:
        er, info = ts.get_field('E', 'r', iteration=0)
        phi, info = ts.get_field('phi', iteration=0)
    else:
        er, info = ts.get_field(f'E_lvl{level}', 'r', iteration=0)
        phi, info = ts.get_field(f'phi_lvl{level}', iteration=0)
```

Examples/Tests/electrostatic_sphere_eb/analysis_rz_mr.py
```diff
@@ -0,0 +1,112 @@
...
+
+    return er, phi, info
+
+def get_error_per_lev(ts,level):
+    er, phi, info = get_fields(ts, level)
+
+    nj = len(er[0,:])//2
```
EZoni: Here we can extract the number of points from `info.r` directly, avoiding confusion with other dimensions/indices:
```suggestion
    nj = info.r.shape[0] // 2
```

Examples/Tests/electrostatic_sphere_eb/analysis_rz_mr.py
```diff
@@ -0,0 +1,113 @@
...
+
+def find_first_non_zero_from_upper_right(matrix):
+    nj = len(matrix[0,:])
+    ni = len(matrix[:,0])
+
+    for i in range(ni-1, -1, -1):
+        for j in range(nj-1, -1, -1):
+            if (matrix[i][j] != 0) and (matrix[i][j] != np.nan):
+                return (i, j)
+    return i,j
```
EZoni: @prkkumar 
Based on your work with mesh refinement simulations in WarpX, do you have experience with finding the min/max coordinates of a mesh refinement patch starting from the data available through an openPMD time series? I see that it is done manually here (index search where a given field has a valid value), but I wonder if there is a built-in way to do this.

EZoni:
Discussed offline with @oshapoval:
We will try to run the code at [c46b6da](https://github.com/ECP-WarpX/WarpX/pull/4922/commits/c46b6dad495b2c079a139f3134920b020052adea), without and with the single additional code change in [e5f0de6](https://github.com/ECP-WarpX/WarpX/pull/4922/commits/e5f0de68abed339940155f5a11a8f9b102c65660), on an input script formerly used to verify the correctness and accuracy of the bug fix.

Source/ablastr/fields/PoissonSolver.H
EZoni: @oshapoval 
Based on the last commit of this PR, I tried changing `1` to `0` in the first of these arguments and I confirm that all four CI tests that are currently passing would crash due to erroneous arithmetic operation if we do so. If this turns out to be the one single change that makes the other case work correctly, we would need to find a fix that is good for both that case and all the CI tests here.

Source/Parallelization/WarpXComm.cpp
```diff
@@ -171,136 +171,191 @@ WarpX::UpdateAuxilaryDataStagToNodal ()
...
+                    amrex::ParallelFor(bx,
+                    [=] AMREX_GPU_DEVICE (int j, int k, int l) noexcept
+                    {
+                        warpx_interp(j, k, l, ex_aux, ex_fp, ex_cp, ex_c, Ex_fp_stag, Ex_cp_stag, refinement_ratio);
+                        warpx_interp(j, k, l, ey_aux, ey_fp, ey_cp, ey_c, Ey_fp_stag, Ey_cp_stag, refinement_ratio);
+                        warpx_interp(j, k, l, ez_aux, ez_fp, ez_cp, ez_c, Ez_fp_stag, Ez_cp_stag, refinement_ratio);
+                    });
+                }
+            }
+            else {
```
RemiLehe: We could add a comment emphasizing that this s for electrostatic

Source/Parallelization/WarpXComm.cpp
```diff
@@ -171,136 +171,191 @@ WarpX::UpdateAuxilaryDataStagToNodal ()
 
         // Bfield
         {
```
RemiLehe, line 173: I will add some more comments later on this file.

Source/Parallelization/WarpXComm.cpp
```diff
@@ -341,141 +396,158 @@ WarpX::UpdateAuxilaryDataSameType ()
...
+                    },
+                    [=] AMREX_GPU_DEVICE (int j, int k, int l) noexcept
+                    {
+                        warpx_interp(j, k, l, bz_aux, bz_fp, bz_c, Bz_stag, refinement_ratio);
+                    });
+                }
+            }
+            else // electrostatic
+            {
+                MultiFab::Copy(*Bfield_aux[lev][0], *Bfield_fp[lev][0], 0, 0, Bfield_aux[lev][0]->nComp(), Bfield_aux[lev][0]->nGrowVect());
```
RemiLehe, line 469: We may need to check whether this needs to use `Bfield_aux[lev][0]->nGrowVect()` or `ng_gather`...

Source/Parallelization/WarpXComm.cpp
RemiLehe: As a follow-up PR: we could avoid the duplication for E and B, and instead have a function that is generic.

Source/Parallelization/WarpXComm_K.H
```diff
@@ -285,6 +285,94 @@ void warpx_interp (int j, int k, int l,
     // Final result
     arr_aux(j,k,l) = tmp + (fine - coarse);
 }
```
RemiLehe, line 287: This is needed by the fix in the `WarpXComm.cpp`, we need a function that interpolates between two `MultiFabs`, instead of existing functions that use 3 `MultiFabs`.

RemiLehe: @EZoni will check if this is functionality already exists somewhere else in WarpX.

Source/ablastr/fields/PoissonSolver.H
```diff
@@ -283,19 +294,18 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
...
-                0,
-                1,
-                ng,
-                ng,
+                phi_cp,                    // dst
+                *phi[lev],                 // src
+                0,                         // src_comp
+                0,                         // dst_comp
+                1,                         // num_comp
+                amrex::IntVect(1),         // src_nghost
```
EZoni: @WeiqunZhang 
I think if we change this line to  
```
amrex::IntVect(0),         // src_nghost
```
the following tests crash: `ElectrostaticSphereLabFrame_MR_emass_10`, `ElectrostaticSphereEB_RZ_MR`, `scraping` and `scraping_filter`.

Source/Parallelization/WarpXComm_K.H
EZoni: It looks to me like we might be able to use the `Interp` function here (maybe for a follow-up PR):
https://github.com/ECP-WarpX/WarpX/blob/b58343a9cf31ad2e2e0c47b2c5f0b15cc122a961/Source/ablastr/coarsen/sample.H#L30-L58

oshapoval (commented):
@WeiqunZhang thanks for this fix.

EZoni:
@oshapoval @WeiqunZhang 
It seems like the test `ElectrostaticSphereLabFrame_MR_emass_10`, one among those mentioned in https://github.com/ECP-WarpX/WarpX/pull/4922#discussion_r1683540102, is still crashing. So the fix might not be complete yet. Will try to see if tweaking the number of ghost cells passed for the `ParallelCopy` is still responsible for the crash.

Source/ablastr/fields/PoissonSolver.H
```diff
@@ -308,7 +316,7 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
 
                 details::PoissonInterpCPtoFP const interp(phi_fp_arr, phi_cp_arr, refratio);
 
-                amrex::Box const b = mfi.tilebox(phi[lev + 1]->ixType().toIntVect());
+                amrex::Box const& b = mfi.growntilebox();
```
WeiqunZhang:
```suggestion
                amrex::Box const& b = mfi.growntilebox(ng);
```

Source/ablastr/fields/PoissonSolver.H
```diff
@@ -280,10 +285,13 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
             const amrex::IntVect& refratio = rel_ref_ratio.value()[lev];
             ba.coarsen(refratio);
             const int ncomp = linop.getNComp();
-            amrex::MultiFab phi_cp(ba, phi[lev+1]->DistributionMap(), ncomp, 1);
+            const int ng = (WarpX::grid_type == GridType::Collocated) ? 1 : 0;
+            amrex::MultiFab phi_cp(ba, phi[lev+1]->DistributionMap(), ncomp, ng);
+            if (ng > 0) {
+                phi_cp.setDomainBndry(0.0_rt, geom[lev]);
+            }
```
EZoni: Adding a comment (copied from AMReX's docstring for `setDomainBndry`):
```suggestion
            if (ng > 0) {
                // Set all values outside the domain to zero
                phi_cp.setDomainBndry(0.0_rt, geom[lev]);
            }
```

Source/Parallelization/WarpXComm.cpp
EZoni: @WeiqunZhang 
Speaking of ghost cells, do you think it is correct to use this number of ghost cells here? We are calling `Copy` because we essentially need to copy the `_fp` arrays into the `_aux` ones. What happens, for instance, if the source MultiFab has less ghost cells than the ones that we pass here, taken from the destination MultiFab?

Source/Parallelization/WarpXComm.cpp
```diff
@@ -171,136 +171,191 @@ WarpX::UpdateAuxilaryDataStagToNodal ()
...
+                    amrex::ParallelFor(bx,
+                    [=] AMREX_GPU_DEVICE (int j, int k, int l) noexcept
+                    {
+                        warpx_interp(j, k, l, bx_aux, bx_fp, bx_cp, bx_c, Bx_fp_stag, Bx_cp_stag, refinement_ratio);
+                        warpx_interp(j, k, l, by_aux, by_fp, by_cp, by_c, By_fp_stag, By_cp_stag, refinement_ratio);
+                        warpx_interp(j, k, l, bz_aux, bz_fp, bz_cp, bz_c, Bz_fp_stag, Bz_cp_stag, refinement_ratio);
+                    });
+                }
+            }
+            else {
```
EZoni: Adding a comment:
```suggestion
            else { // electrostatic
```

Source/Parallelization/WarpXComm.cpp
EZoni: Adding a comment:
```suggestion
            else { // electrostatic
```

Source/Parallelization/WarpXComm.cpp
```diff
@@ -171,136 +171,191 @@ WarpX::UpdateAuxilaryDataStagToNodal ()
...
+                for (MFIter mfi(*Bfield_aux[lev][0], TilingIfNotGPU()); mfi.isValid(); ++mfi)
+                {
+                    Array4<Real> const& bx_aux = Bfield_aux[lev][0]->array(mfi);
+                    Array4<Real> const& by_aux = Bfield_aux[lev][1]->array(mfi);
+                    Array4<Real> const& bz_aux = Bfield_aux[lev][2]->array(mfi);
+                    Array4<Real const> const& bx_fp = Bfield_fp[lev][0]->const_array(mfi);
+                    Array4<Real const> const& by_fp = Bfield_fp[lev][1]->const_array(mfi);
+                    Array4<Real const> const& bz_fp = Bfield_fp[lev][2]->const_array(mfi);
+
+                    const Box& bx = mfi.growntilebox();
```
EZoni, line 255: @WeiqunZhang 
Another question for you. It seems like here, for the B field, we use `growntilebox`, while below, for the E field, line 350, we use `fabbox`. I would expect that there is no reason for using different functions for B and E. Which one do you think is the most correct here, and why?

Source/Parallelization/WarpXComm.cpp
EZoni: Ported to #5078, we will fix this in a separate PR.

Source/Parallelization/WarpXComm.cpp
WeiqunZhang: That would be an error, if `_aux` has more ghost cells than `_fp`.

Source/Parallelization/WarpXComm.cpp
WeiqunZhang: I think it's better to use `growntilebox` and add `TilingIfNotGPU()` for the E field loop. For GPU, the two loops are the same. But for CPU, the tiling version should be better due to cache blocking and more OpenMP parallelism.

Source/Parallelization/WarpXComm.cpp
EZoni: Thanks!

In such a case, what would be your advice if one is not sure about the exact number of ghost cells required by the specific algorithm implemented here? 

Would it be more robust to call `Copy` by passing a number of ghost cells that is the minimum between the destination and source ghost cells?

Or would it be more robust to leave it as is, relying on AMReX to flag (with either an assertion error or a segfault - depending on what is implemented on the AMReX side) if the destination ghost cells are more than the source ghost cells, thus pointing out an underlying bug on the WarpX side?

RemiLehe (approved):

Source/ablastr/fields/PoissonSolver.H
```diff
@@ -263,10 +263,15 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
         mlmg.setVerbose(verbosity);
         mlmg.setMaxIter(max_iters);
         mlmg.setAlwaysUseBNorm(always_use_bnorm);
+        if (WarpX::grid_type == GridType::Collocated) {
```
ax3l, line 266: Oops, this slipped through. We cannot use WarpX objects in the ABLASTR library. WarpX depends on ABLASTR but not the other way around. Using this here breaks ImpactX.

I will post a PR to fix this.

ax3l: https://github.com/ECP-WarpX/WarpX/pull/5106

PR #4927, opened 2024-05-12
Update hip installation script for CI

WeiqunZhang:
It appears that "apt/debian ubuntu main" does not work anymore. We now use "apt/latest ${UBUNTU_CODENAME} main" instead.

### DIFF
--- a/.github/workflows/dependencies/hip.sh
+++ b/.github/workflows/dependencies/hip.sh
@@ -12,12 +12,24 @@ set -eu -o pipefail
 #   failed files the given number of times.
 echo 'Acquire::Retries "3";' | sudo tee /etc/apt/apt.conf.d/80-retries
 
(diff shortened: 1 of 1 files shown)


ax3l (approved):
:sparkles: 

Thanks a lot! Did you have a chance to report this upstream with AMD already as an issue?

WeiqunZhang:
No I didn't.

I made the change because of error messages like
```
W: Conflicting distribution: https://repo.radeon.com/rocm/apt/debian ubuntu InRelease (expected ubuntu but got focal)
```
So it seems that it's more reliable to specify the ubuntu code name.

ax3l:
Thanks a lot! I sent an email to get this long-term fixed with you in CC. Will merge here in the meantime.

Thanks for debugging this and adding a work-around!

ax3l:
Oh right, actually it's documented now to use the Ubuntu code name explicitly.
  https://rocm.docs.amd.com/projects/install-on-linux/en/latest/how-to/native-install/ubuntu.html

PR #4929, opened 2024-05-12
LICENSE.txt: Wrap to 80 Characters

ax3l:
Use modern formatting as we have in newer codes to wrap the `LICENSE.txt` file for readability to 80 characters per line.

No content change.


lucafedeli88 (approved):
Thanks!

PR #4930, opened 2024-05-13
`ElectrostaticSphereLabFrame_MR_emass_10`: +Positions

ax3l:
Follow-up to #4914 + #4916, which ran independently (success) but have colliding changes (new test + change in output checks).


ax3l:
Merging to unbreak CI :)

PR #4938, opened 2024-05-16
Ensure compilation with double precision particles for DSMC collisions

aveksler1:
DSMC collisions are not properly captured with single precision for typical plasma collision cross sections. This adds an assert to ensure compilation with double precision. 

I could imagine a scenario where a user might still want to use single precision for some cross sections that would be captured, so this could we a warning too rather than an assert.

### DIFF
--- a/Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
+++ b/Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
@@ -23,7 +23,7 @@ DSMCFunc::DSMCFunc (
 {
     using namespace amrex::literals;
 
-    WARPX_ALWAYS_ASSERT_WITH_MESSAGE( !(std::is_same<amrex::ParticleReal, float>::value),
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE( (std::is_same<amrex::ParticleReal, double>::value),
     "Particle precision must be double for DSMC collisions.");
 
     const amrex::ParmParse pp_collision_name(collision_name);
--- a/Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
+++ b/Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
@@ -7,6 +7,7 @@
  * License: BSD-3-Clause-LBNL
  */
 #include "DSMCFunc.H"
+#include "Utils/TextMsg.H"
 
 /**
  * \brief Constructor of the DSMCFunc class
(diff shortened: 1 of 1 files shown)


archermarx:
Is there some way we can scale the cross sections to make them work numerically with single-precision particles?

Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
```diff
@@ -22,6 +23,9 @@ DSMCFunc::DSMCFunc (
 {
     using namespace amrex::literals;
 
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE( !(std::is_same<amrex::ParticleReal, float>::value),
```
roelof-groenewald: Instead of checking that `ParticleReal` precision is not single, I'd say it is better to check that is is `double`.
```suggestion
    WARPX_ALWAYS_ASSERT_WITH_MESSAGE( (std::is_same<amrex::ParticleReal, double>::value),
```

roelof-groenewald:
> Is there some way we can scale the cross sections to make them work numerically with single-precision particles?

@archermarx we should be able to make single precision work for the collisions. I expect the problem is more likely in the calculation of the collision energy than the cross-sections themselves since I'm fairly certain the MCC routine works for single precision (which uses similar cross-sections). We can open an issue to remind us to sort out this issue in the future since I won't have time this week to look into this deeper.

Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
ax3l: Agreed.

Technically even `double` or `long double`, but we do not expose the latter in build systems yet and using it would be really slow.
https://en.cppreference.com/w/cpp/language/types

ax3l:
LGTM - note: if you do need SP runs of WarpX where DSMC is just a small part, you could promote the DSMC logic to be all `double` until you have found a way to renormalize.

This would be a bigger change, so erroring out at runtime is fine for me. Please double check the constructor where this is placed does not throw when DSMC is not used.

roelof-groenewald (approved):
Thanks!

aveksler1:
> LGTM - note: if you do need SP runs of WarpX where DSMC is just a small part, you could promote the DSMC logic to be all `double` until you have found a way to renormalize.
> 
> This would be a bigger change, so erroring out at runtime is fine for me. Please double check the constructor where this is placed does not throw when DSMC is not used.

Just double checked, it's working as intended.

PR #4942, opened 2024-05-21
Add temperature diagnostic

dpgrote:
This PR adds a grid based temperature diagnostic. This is specified by adding `T_<species>` to the `fields_to_plot` input parameter.

The averages are calculated in two steps, first the average velocity `<u>`, then the average velocity squared `<u - <u>>**2`. This method is more robust than the single step using `<u**2> - <u>**2` when `<u> >> u_rms`.


roelof-groenewald:
This is awesome! Thanks for adding. It will be a much nicer way to get the temperature over my current approach of using the `ParticleFieldDiagnostic`. Could you explain a bit more why you say the one step approach with `<u**2> - <u>**2` is less robust than making 2 deposition steps? Do you know where the numerical error comes from in the former?

Also, have you considered outputting the temperature in each direction (i.e. as a vector) rather than averaging? I imagine it won't be long until a user asks for individual temperature components.

dpgrote:
@roelof-groenewald I had done numerical tests comparing the one-step and two-step methods. I found that when the average of `u` was greater than ~1.e6*u_rms, the calculation of the rms using the one step method began to fail, producing wrong results. I believe it is due to round off, as sum(u**2) gets large enough that the digits needed to calculate the rms get truncated. As evidence, that factor of 1.e6 gets smaller with larger numbers of particles.

I was considering saving the temperature in each direction. Should this be an option or should it always do it? My concern about always doing it is the extra data being saved when it is not needed. If an option, how to request it?

roelof-groenewald:
Thanks for that explanation, @dpgrote.

In terms of the temperature in each direction; it could make sense to always treat it as a vector in a similar way as how the current density is handled. That is, by default output the three components separately unless only specific components are requested. That way, in a case where a user knows their velocity distribution will be isotropic they can choose to output only one component (which would give them the same temperature as the average if the velocities really are isotropic), but it also allows for the more general case with non-isotropic velocity distributions. And since all components will be output on the same grid it would be easy to average them in post-processing if a single temperature is desired. What do you think about that?

roelof-groenewald:
@dpgrote if you prefer we can merge this as is and turn my previous comment into an issue for a future PR.

dpgrote:
> @dpgrote if you prefer we can merge this as is and turn my previous comment into an issue for a future PR.

That sounds good. It can wait until someone requests it. Though, I thought about your comment some - perhaps in addition to `T_<species>` diagnostics, there could also be `Tx_<species>` etc. This would provide the flexibility.

Source/Diagnostics/ComputeDiagFunctors/TemperatureFunctor.cpp
```diff
@@ -0,0 +1,165 @@
...
+                ii = static_cast<int>(amrex::Math::floor(lx));
+#if defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
+                const amrex::Real lz = (zp - plo[1]) * dxi[1];
+                jj = static_cast<int>(amrex::Math::floor(lz));
+#elif defined(WARPX_DIM_3D)
+                const amrex::Real ly = (yp - plo[1]) * dxi[1];
+                jj = static_cast<int>(amrex::Math::floor(ly));
+                const amrex::Real lz = (zp - plo[2]) * dxi[2];
+                kk = static_cast<int>(amrex::Math::floor(lz));
+#endif
```
roelof-groenewald, line 133: It would be useful to have a function (maybe in `ParticleUtils.cpp`) that returns the cell indices for a given particle. That would simplify the code here and from lines 60-71.

I'll make an issue out of this comment for follow-up.

roelof-groenewald (approved):

Source/Diagnostics/ComputeDiagFunctors/TemperatureFunctor.cpp
dpgrote: I would agree. This routine would be useful elsewhere, for example in the deposition and gather routines.

Thanks for merging!

PR #4944, opened 2024-05-21
Add feature projection div cleaner

### LINKED ISSUE #4355: Initial B-field loading gives non divergence free field
When using an openPMD file to load an initial B-field, WarpX interpolates from the openPMD file's grid to the WarpX grid. This can lead to loss of the divergence free aspect of the B-field as demonstrated in the plot below where an FRC is loaded into WarpX:
image

Instead, the vector potential could be interpolated onto the WarpX grid (with staggering matching the E-field) and the B-field calculated as the curl of that field which ensures a divergence free B-field (up to machine precision):
image

Another alternative would be to perform some divergence cleaning of the B-field after interpolation.

### LINKED ISSUE #4356: Add warning when interpolating initial B-field
As a short-term remedy for #4355 a warning could be added that will alert the user of known issues with interpolating B-field values between non-matching grids.

clarkse:
Divergence cleaner class that computes the divergence at cell centers, then uses that as a source for a Poisson solve using MLMG that can be used to remove divergence components from the B-fields after initialization. This is useful for scrubbing interpolated B fields that are loaded into WarpX either through the openPMD loader or when using the pyWarpX fields interface.


clarkse:
Fixes #4355 and #4356.

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
```diff
@@ -0,0 +1,262 @@
...
+            auto const n_coefs_z = static_cast<int>(m_stencil_coefs_z.size());
+
+            amrex::Array4<Real> const& src_arr = m_source[ilev].get()->array(mfi);
+
+            amrex::ParallelFor(mfi.tilebox(),
+            [=] AMREX_GPU_DEVICE (int i, int j, int k)
+            {
+                src_arr(i,j,k) -= CartesianYeeAlgorithm::DownwardDx(Bx_arr, coefs_x, n_coefs_x, i+1, j, k);
+                src_arr(i,j,k) -= CartesianYeeAlgorithm::DownwardDy(By_arr, coefs_y, n_coefs_y, i, j+1, k);
+                src_arr(i,j,k) -= CartesianYeeAlgorithm::DownwardDz(Bz_arr, coefs_z, n_coefs_z, i, j, k+1);
```
roelof-groenewald: Wouldn't the index stepping here cause issues in 1d/2d? Is there a reason not to use `UpwardD{x/y/z}` instead of `DownwardD{x/y/z}`?

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
```diff
@@ -0,0 +1,262 @@
...
+        // Synchronize the ghost cells, do halo exchange
+        ablastr::utils::communication::FillBoundary(*(m_solution[ilev].get()),
+                                                m_solution[ilev]->nGrowVect(),
+                                                WarpX::do_single_precision_comms,
+                                                geom[ilev].periodicity());
+    }
+}
+
+void
+ProjectionDivCleaner::setSourceFromBfield ()
```
roelof-groenewald, line 202: A cleaner approach could be to add a `ComputeDivB.cpp` file (analogous to https://github.com/ECP-WarpX/WarpX/blob/development/Source/FieldSolver/FiniteDifferenceSolver/ComputeDivE.cpp) that can be called from the `run_ProjectionDivCleaner` function to populate the source. What do you think?

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
clarkse: I had some issue with the prototype of UpwardD. I will see if I can adjust it so I don't need to explicitly index.

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
clarkse: I changed this to use the existing static ComputeDivB method from the WarpX class.

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
clarkse: This should now be moot as it has been removed in favor of using existing ComputeDivB.

Source/Initialization/DivCleaner/ProjectionDivCleaner.H
```diff
@@ -0,0 +1,80 @@
+/* Copyright 2024 Helion Energy, Inc.
+ *
+ * Authors: S. Eric Clark
+ * License: None
+ */
+#pragma once
```
roelof-groenewald: Nice! We might want to start a coordinated effort to switch all WarpX header files to this approach instead of the old, lengthier 
```
#ifndef _MYHEADER_H_
#define _MYHEADER_H_
...
#endif
```

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+ProjectionDivCleaner::ProjectionDivCleaner()
+{
+    // Initialize tolerance based on field precision
+    if constexpr (std::is_same<Real, float>::value) {
+        // Error out of divergence cleaner
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(false,
+            "Single Precision Divergence Cleaner has convergence problems. "
+            "Please compile with WarpX_PRECISION=DOUBLE."
+        );
```
roelof-groenewald: Do you still plan on switching this to a warning in order to not break code that was previously running?

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
```diff
@@ -0,0 +1,336 @@
...
+    info.setConsolidation(m_consolidation);
+    info.setMaxCoarseningLevel(m_max_coarsening_level);
+
+    for (int ilev = 0; ilev < m_levels; ++ilev)
+    {
+        MLPoisson mlpoisson({geom[ilev]}, {ba[ilev]}, {dmap[ilev]}, info);
+
+        mlpoisson.setMaxOrder(m_linop_maxorder);
+
+        // This is a 3d problem with Dirichlet BC
```
roelof-groenewald: This comment is outdated, right? Since the `lobc` and `hibc` are not necessarily Dirichlet.

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
```diff
@@ -0,0 +1,336 @@
...
+}
+
+void
+ProjectionDivCleaner::correctBfield ()
+{
+    // Get WarpX object
+    auto & warpx = WarpX::GetInstance();
+    const auto& geom = warpx.Geom();
+
+    // This function will compute divB and add to the source multifab
```
roelof-groenewald: This comment also looks outdated. The function adds the gradient of the solution multifab to the simulation B-field, right?

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1371,6 +1372,17 @@ WarpX::LoadExternalFieldsFromFile (int const lev)
         ReadExternalFieldFromFile(m_p_ext_field_params->external_fields_path, Bfield_fp_external[lev][1].get(), "B", "y");
         ReadExternalFieldFromFile(m_p_ext_field_params->external_fields_path, Bfield_fp_external[lev][2].get(), "B", "z");
 #endif
+        if constexpr (!std::is_same<Real, double>::value) {
+            ablastr::warn_manager::WMRecordWarning("External Fields Loader",
+                "Field Precision is SINGLE, so divB not cleaned. Interpolation may lead to non-zero B field divergence.",
+                ablastr::warn_manager::WarnPriority::medium);
+        } else if (grid_type -= GridType::Collocated) {
```
roelof-groenewald: Typo:
```suggestion
        } else if (grid_type == GridType::Collocated) {
```

Source/Initialization/DivCleaner/ProjectionDivCleaner.H
dpgrote: The `#pragma once` looks like a nice feature but is not standard. This should be checked if it's appropriate here. Also, the header should follow the same pattern as other files, specifically "License: BSD-3-Clause-LBNL".

Source/Python/pyWarpX.cpp
```diff
@@ -67,6 +69,8 @@ PYBIND11_MODULE(PYWARPX_MODULE_NAME, m) {
     init_MultiParticleContainer(m);
     init_WarpX(m);
 
+    m.def("run_div_cleaner", &run_ProjectionDivCleaner);
```
roelof-groenewald: Adding the Python binding here is a bit different from how similar things are done currently. The more standard approach would be that `run_ProjectionDivCleaner()` be a member function of `WarpX`, and the Python binding would then go in `Source/Python/WarpX.cpp`. Do you think it is worth making the change to follow that standard?

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
roelof-groenewald: Oh I see you added the warning in `WarpXInitData.cpp`.

roelof-groenewald (commented):
Looks good! I think the namespace in which to put the calling function is the only remaining issue.

Source/Python/pyWarpX.cpp
clarkse: Done

Source/Initialization/DivCleaner/ProjectionDivCleaner.H
clarkse: I think the pragma once is pretty widely supported. I switched over to include guards to be consistent with the rest of the header files.

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
```diff
@@ -0,0 +1,348 @@
...
+}
+
+void
+ProjectionDivCleaner::setSourceFromBfield ()
+{
+    // Get WarpX object
+    auto & warpx = WarpX::GetInstance();
+    const auto& geom = warpx.Geom();
+
+    // This function will compute -divB and sotre in the source multifab
```
roelof-groenewald:
```suggestion
    // This function will compute -divB and store it in the source multifab
```

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
```diff
@@ -0,0 +1,348 @@
...
+            }
+        }
+
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+            (WarpX::field_boundary_lo[idim] != FieldBoundaryType::Open &&
+            WarpX::field_boundary_hi[idim] != FieldBoundaryType::Open &&
+            WarpX::field_boundary_lo[idim] != FieldBoundaryType::PML &&
+            WarpX::field_boundary_hi[idim] != FieldBoundaryType::PML) ,
+            "Open and PML field boundary conditions only work with "
+            "warpx.poisson_solver = fft."
```
roelof-groenewald: Currently, there is no check that avoids the div B cleaner from being executed if the fft solver is used, right? In that case this error message would be confusing, but it would be good to simply not perform the div B cleaning when the fft solver is used, and print a message warning about possible B-field divergence (as is done for the other cases).

clarkse: Doesn't the PSATD solver have its own flavor of div cleaning that diffuses it away over time? I guess since the BCs could be incompatible it should not be executed if using the PSATD. I will add another check for that.

clarkse: I tightened this check further to keep it from being executed with incompatible solvers.

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
```diff
@@ -0,0 +1,389 @@
...
+    amrex::Array<LinOpBCType,AMREX_SPACEDIM> hibc({AMREX_D_DECL(LinOpBCType::bogus,
+                                                                LinOpBCType::bogus,
+                                                                LinOpBCType::bogus)});
+
+#ifdef WARPX_DIM_RZ
+    if (geom[0].ProbLo(0) == 0){
+        lobc[0] = LinOpBCType::Neumann;
+
+        // handle the r_max boundary explicitly
+        if (WarpX::field_boundary_hi[0] == FieldBoundaryType::PEC) {
```
dpgrote: This line is only reached when `ProbLo(0) == 0`. What happens otherwise?

This whole section of code could be greatly simplified using a map.

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
```diff
@@ -0,0 +1,389 @@
...
+            dmap, ncomps, ng, tag1);
+        const auto tag2 = amrex::MFInfo().SetTag("div_cleaner_source");
+        m_source[lev] = std::make_unique<MultiFab>(amrex::convert(ba, IntVect::TheCellVector()),
+            dmap, ncomps, ng, tag2);
+
+        m_solution[lev]->setVal(0.0, ng);
+        m_source[lev]->setVal(0.0, ng);
+    }
+
+    m_h_stencil_coefs_x.resize(1);
```
dpgrote: Could this routine be used here? `InitializeStencilCoefficients`

Python/pywarpx/picmi.py
```diff
@@ -1448,12 +1449,34 @@ def laser_antenna_initialize_inputs(self, laser):
 
 
 class LoadInitialField(picmistandard.PICMI_LoadGriddedField):
+    def init(self, kw):
+        self.do_divb_cleaning_external = kw.pop('do_divb_cleaning_external', True)
```
dpgrote: This should be `self.do_divb_cleaning_external = kw.pop('warpx_do_divb_cleaning_external', None)` so that it has the same default as the compiled code and has the `warpx` prefix.

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
clarkse: I think it could be used here. I will attempt to do a refactor.

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
clarkse: I will take a look at it. We should probably also refactor the BC logic in the static solvers the same way. I grabbed similar logic from there and modified.  I think a separate PR would appropriate to clean up the BC logic in several spots.

roelof-groenewald: I think this probably needs something like
```
if (geom[0].ProbLo(0) == 0){
    lobc[0] = LinOpBCType::Neumann;
} else {
    // handle the r_min boundary explicitly
    if (WarpX::field_boundary_lo[0] == FieldBoundaryType::PEC) {
        lobc[0] = LinOpBCType::Dirichlet;
    }
    else if (WarpX::field_boundary_lo[0] == FieldBoundaryType::Neumann) {
        lobc[0] = LinOpBCType::Neumann;
    }
    else {
        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(false,
            "Field boundary condition at the inner radius must be either PEC or neumann "
            "when using the Projection Divergence Cleaner"
        );
    }
}
```
at lines 123-124 with the rest of this block moved out of the `if` statement.

Source/WarpX.H
```diff
@@ -264,6 +264,10 @@ public:
     //! Solve additional Maxwell equation for G in order to control errors in magnetic Gauss' law
     static bool do_divb_cleaning;
 
+    //! Solve Poisson equation when loading a an external magnetic field to clean divergence
```
roelof-groenewald:
```suggestion
    //! Solve Poisson equation when loading an external magnetic field to clean divergence
```

Source/Initialization/WarpXInitData.cpp
```diff
@@ -542,12 +549,14 @@ void
 WarpX::AddExternalFields () {
     for (int lev = 0; lev <= finest_level; ++lev) {
         // FIXME: RZ multimode has more than one component for all these
-        if (m_p_ext_field_params->E_ext_grid_type == ExternalFieldType::read_from_file) {
+        if (m_p_ext_field_params->E_ext_grid_type == ExternalFieldType::read_from_file
+            || m_p_ext_field_params->E_ext_grid_type == ExternalFieldType::load_from_python) {
```
roelof-groenewald: I might be missing something here but it seems like this is a bug in `development`, no? The external field should always be added to the fine patch fields, even if it is constant or parsed. So shouldn't this just be:
```suggestion
        if (m_p_ext_field_params->E_ext_grid_type != ExternalFieldType::default_zero) {
```

Assuming that is correct, I don't think the `load_from_python` type is needed anymore.

Examples/Tests/projection_divb_cleaner/inputs_rz
```diff
@@ -0,0 +1,51 @@
+warpx.serialize_initial_conditions = 0
```
roelof-groenewald, line 1: I think for this and the `PICMI_inputs_3d` test in the same folder it would be better to modify the original tests instead of creating new ones. I say this because we know the existing tests are strictly less accurate than these new ones - not having divergence free B-fields are purely bad and therefore I don't see a good reason to keep both sets of tests.

roelof-groenewald (changes requested):
A few minor comments.

I also think there is a bug in `development` for dealing with external grid fields. I will submit a PR to fix that which can then be updated in this PR as well.

Source/Initialization/WarpXInitData.cpp
clarkse: I believe the external fields are only allocated if the read_from_file, and now in this case the load_from_python flags are set for the external field grid type.

Examples/Tests/projection_divb_cleaner/inputs_rz
clarkse: So to be clear you want to merge the external field loader from file tests and have a separate one for the python loader?

Source/Initialization/WarpXInitData.cpp
roelof-groenewald: This can be updated once #5104 is merged.

Examples/Tests/projection_divb_cleaner/inputs_rz
roelof-groenewald: Yes, that sounds good to me.

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
clarkse: I updated this to use a std::map hash table.

Source/Initialization/WarpXInitData.cpp
clarkse: This has been merged from PR #5104

Examples/Tests/projection_divb_cleaner/inputs_rz
clarkse: @dpgrote When we talked earlier you thought that we should leave the default to off for the clean_divb_external flag. I think we should have a quick conversation between you, me, and @roelof-groenewald about it. I set it up to default to off. If we want to default it to on, then some of the other external field unit tests likely will need to have updated benchmarks. What behavior do we desire here?

clarkse: @roelof-groenewald Since the existing tests are for collocated grids, the cleaner with its default of being on will log a warning. The new CI tests added are on a Yee mesh, which will use the cleaner. I think covering all the options is best. I need to probably do some additional testing to make sure it is working properly with the collocated grids, but that could be another PR.

Python/pywarpx/picmi.py
```diff
@@ -1656,12 +1657,52 @@ def laser_antenna_initialize_inputs(self, laser):
...
         if self.load_B:
             pywarpx.warpx.B_ext_grid_init_style = 'read_from_file'
+            pywarpx.warpx.do_divb_cleaning_external = True
+
+class LoadInitialFieldFromPython:
+    """
+    Field Initializer that takes a function handle to be registered as a callback.
+    The function is expected to write the E and/or B fields into the
+    fields.Bx/y/zFPExternalWrapper() multifab. The callback is installed
+    in the beforeInitEsolve hook. This should operate identically to laoding from
```
```suggestion
    in the beforeInitEsolve hook. This should operate identically to loading from
```

Python/pywarpx/picmi.py
```diff
@@ -1656,12 +1657,52 @@ def laser_antenna_initialize_inputs(self, laser):
...
+        Flag that controls whether or not to execute the Projection based B-field divergence cleaner.
+
+    load_E: bool, default=True
+        E field is expected to be loaded in the registered callback.
+
+    load_B: bool, default=True
+        B field is expected to be loaded in the registered callback.
+    """
+    def __init__(self, **kw):
+        self.do_divb_cleaning_external = kw.pop('warpx_do_divb_cleaning_external', True)
```
roelof-groenewald: I'm wondering about the default `True` here and in `LoadInitialField`. Usually we keep these default to `None` so that no parameter is passed to WarpX and therefore the C++ defaults are used (instead of needing to work at maintaining the same default settings in picmi and in C++ separately). Is it needed to have this default to `True`?

Python/pywarpx/picmi.py
```diff
@@ -1656,12 +1657,52 @@ def laser_antenna_initialize_inputs(self, laser):
...
+    def init(self, kw):
+        self.do_divb_cleaning_external = kw.pop('warpx_do_divb_cleaning_external', True)
+
     def applied_field_initialize_inputs(self):
         pywarpx.warpx.read_fields_from_path = self.read_fields_from_path
         if self.load_E:
             pywarpx.warpx.E_ext_grid_init_style = 'read_from_file'
         if self.load_B:
             pywarpx.warpx.B_ext_grid_init_style = 'read_from_file'
+            pywarpx.warpx.do_divb_cleaning_external = True
```
```suggestion
            pywarpx.warpx.do_divb_cleaning_external = self.do_divb_cleaning_external
```

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
```diff
@@ -0,0 +1,335 @@
...
+        if (itlo == bcmap.end() || ithi == bcmap.end()) {
+            WARPX_ABORT_WITH_MESSAGE(
+                "Field boundary conditions have to be either periodic, PEC or neumann "
+                "when using the MLMG projection based divergence cleaner solver."
+            );
+        }
+
+        lobc[idim] = bcmap[WarpX::field_boundary_lo[idim]];
+        hibc[idim] = bcmap[WarpX::field_boundary_hi[idim]];
+    }
```
roelof-groenewald, line 160: This is a nice and clean solution!

Python/pywarpx/picmi.py
roelof-groenewald: It seems fine to set this to `None` by default since the C++ code will default it back to `True`.
```suggestion
        self.do_divb_cleaning_external = kw.pop('warpx_do_divb_cleaning_external', None)
```

Python/pywarpx/picmi.py
```diff
@@ -1656,12 +1657,52 @@ def laser_antenna_initialize_inputs(self, laser):
 
 
 class LoadInitialField(picmistandard.PICMI_LoadGriddedField):
+    def init(self, kw):
+        self.do_divb_cleaning_external = kw.pop('warpx_do_divb_cleaning_external', True)
```
roelof-groenewald: Same comment as below.
```suggestion
        self.do_divb_cleaning_external = kw.pop('warpx_do_divb_cleaning_external', None)
```

roelof-groenewald (commented):
I think just a last couple minor defaulting things to sort out and we should be good to go!

Source/Initialization/WarpXInitData.cpp
```diff
@@ -573,6 +574,12 @@ WarpX::InitData ()
         // Loop through species and calculate their space-charge field
         bool const reset_fields = false; // Do not erase previous user-specified values on the grid
         ExecutePythonCallback("beforeInitEsolve");
+
+        // Run div cleaner here on the loaded fields
+        if (WarpX::do_divb_cleaning_external) {
+            WarpX::ProjectionCleanDivB();
+        }
```
roelof-groenewald: After we merge #5162, these lines could be moved to `WarpX::LoadExternalFields`.

Source/Python/WarpX.cpp
```diff
@@ -199,6 +199,10 @@ The physical fields in WarpX have the following naming:
             py::arg("potential"),
             "Sets the EB potential string and updates the function parser."
         )
+        .def_static("run_div_cleaner",
+            [] () { WarpX::ProjectionCleanDivB(); },
+            "Executes projection based divergence cleaner on loaded B_aux."
```
roelof-groenewald:
```suggestion
            "Executes projection based divergence cleaner on loaded Bfield_fp_external."
```

Source/Initialization/WarpXInitData.cpp
clarkse: Sounds good. I will wait for that to be merged.

Regression/WarpX-tests.ini
```diff
@@ -2352,6 +2352,23 @@ useOMP = 1
...
+addToCompileString = USE_RZ=TRUE
+cmakeSetupOpts = -DWarpX_DIMS=RZ -DWarpX_OPENPMD=ON
+restartTest = 0
+useMPI = 1
+numprocs = 1
+useOMP = 1
+numthreads = 1
+compileTest = 0
+doVis = 0
+compareParticles = 0
```
roelof-groenewald: These unused parameters were removed in #5043. 
```suggestion
```

Regression/WarpX-tests.ini
```diff
@@ -3024,6 +3041,44 @@ useOMP = 1
...
+cmakeSetupOpts = -DWarpX_DIMS=3 -DWarpX_APP=OFF -DWarpX_PYTHON=ON -DWarpX_OPENPMD=ON
+target = pip_install
+restartTest = 0
+useMPI = 1
+numprocs = 1
+useOMP = 1
+numthreads = 1
+compileTest = 0
+doVis = 0
+compareParticles = 0
```
roelof-groenewald:
```suggestion
```

Regression/WarpX-tests.ini
```diff
@@ -3024,6 +3041,44 @@ useOMP = 1
...
```
roelof-groenewald:
```suggestion
```

Source/Initialization/WarpXInitData.cpp
clarkse: I moved this out of the check for checkpoint restart so it will run for new starts and restarts, however there were uninitialized values when trying to run from the LoadExternalFields routine, so I was unable to move to that location. I think having it run and cleaning it up prior to running initial solutions is a good spot.

roelof-groenewald (approved):
Thanks so much for adding this! Looks good to me.

PR #4945, opened 2024-05-22
FFT-based Poisson solver follow-up

aeriforme:
This follows-up PR #4648 as per @ax3l's requests [here](https://github.com/ECP-WarpX/WarpX/pull/4648#discussion_r1595624093).


Source/ablastr/fields/PoissonSolver.H
```diff
@@ -106,7 +106,7 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
             amrex::Vector<amrex::DistributionMapping> const& dmap,
             amrex::Vector<amrex::BoxArray> const& grids,
             T_BoundaryHandler const boundary_handler,
-            [[maybe_unused]] bool is_solver_multigrid,
+            [[maybe_unused]] bool is_solver_igf_on_lev0,
```
ax3l: This should now always be checked:
```suggestion
            bool is_solver_igf_on_lev0,
```

Source/ablastr/fields/PoissonSolver.H
```diff
@@ -160,9 +160,14 @@ computePhi (amrex::Vector<amrex::MultiFab*> const & rho,
                 {{ beta[0], beta[1], beta[2] }};
 #endif
 
+#if !defined(WARPX_USE_PSATD)
+        ABLASTR_ALWAYS_ASSERT_WITH_MESSAGE( !is_solver_igf_on_lev0,
+        "Must compile with -DWarpX_PSATD=ON to use the FFT solver!");
+#endif
+
 #if (defined(WARPX_USE_PSATD)  && defined(WARPX_DIM_3D))
```
ax3l: There is one more runtime path we need to check.

Currently, if the user passes `is_solver_igf_on_lev0=true` but the simulation is not 3D, it silently falls back to MLMG on level 0. Instead, this should also assert with a clear error message.

```suggestion
#if !defined(WARPX_DIM_3D)
        ABLASTR_ALWAYS_ASSERT_WITH_MESSAGE( !is_solver_igf_on_lev0,
        "The FFT Poisson solver is currently only implemented for 3D!");
#endif

#if (defined(WARPX_USE_PSATD) && defined(WARPX_DIM_3D))
```

ax3l (approved):
Thank you, that's perfect!

PR #4947, opened 2024-05-23
Make RemakeMultiFab a lambda function inside WarpX::RemakeLevel

lucafedeli88:
Since `RemakeMultiFab` is used only inside `WarpX::RemakeLevel`, this PR makes it a lambda function inside that function. The main advantage is that it can capture the WarpX class by value and use the method `AllocInitMultiFab` without calling it as `WarpX::AllocInitMultiFab` . This is not particularly useful now, but it will be useful when static methods of the WarpX class will be turned into non-static member functions.


ax3l (approved):
Ok :)

PR #4951, opened 2024-05-24
Picmi: allow different `random_fraction` for different species

roelof-groenewald:
Currently all species included in a `ParticleDiagnostic` will be output with the same stride or random fraction. This PR allows a list of strides to be given so that each species can use a different stride.

### DIFF
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -2605,24 +2605,21 @@ class ParticleDiagnostic(picmistandard.PICMI_ParticleDiagnostic, WarpXDiagnostic
         else:
             species_names = [self.species.name]
 
-        # check if random fraction is specified and whether a value is given for each species
+        # check if random fraction is specified and whether a value is given per species
         random_fraction = {}
-        for species_name in species_names:
-            if isinstance(self.random_fraction, dict):
-                random_fraction[species_name] = self.random_fraction.get(species_name, 1)
-            else:
-                # use the same random fraction for all species
-                random_fraction[species_name] = self.random_fraction
+        random_fraction_default = self.random_fraction
+        if isinstance(self.random_fraction, dict):
+            random_fraction_default = 1.0
+            for key, val in self.random_fraction.items():
+                random_fraction[key.name] = val
 
-        # check if uniform stride is specified and whether a value is given for each species
+        # check if uniform stride is specified and whether a value is given per species
         uniform_stride = {}
-        for species_name in species_names:
-            if isinstance(self.uniform_stride, dict):
-                uniform_stride[species_name] = self.uniform_stride.get(species_name, 1)
-            else:
-                # use the same stride for all species
-                uniform_stride[species_name] = self.uniform_stride
-
+        uniform_stride_default = self.uniform_stride
+        if isinstance(self.uniform_stride, dict):
+            uniform_stride_default = 1
+            for key, val in self.uniform_stride.items():
+                uniform_stride[key.name] = val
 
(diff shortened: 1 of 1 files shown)


Python/pywarpx/picmi.py
```diff
@@ -2599,16 +2599,35 @@ def diagnostic_initialize_inputs(self):
         else:
             species_names = [self.species.name]
 
+        # check if a different random fraction is given for each species
```
dpgrote: Can the input parameter be a dictionary instead of a list? Something like `warpx_random_fraction = {electrons:0.4, ions:0.7}`.  The order of a list would depend on the order in which species are added which is something the could easily be modified in a complex input file, leaving this list inconsistent. Also, this would allow using default values - any species not specified in the dictionary would default to 1.

Python/pywarpx/picmi.py
```diff
@@ -2599,16 +2599,35 @@ def diagnostic_initialize_inputs(self):
...
+        if np.iterable(self.random_fraction):
+            if len(self.random_fraction) != len(species_names):
+                raise AttributeError(
+                    "Species list and random fraction list do not have equal length"
+                )
+        else:
+            self.random_fraction = [self.random_fraction] * len(species_names)
+
+        # check if a different stride is given for each species
```
dpgrote: The same here - can a dictionary be used?

Python/pywarpx/picmi.py
dpgrote: Also, update the documentation above.

roelof-groenewald: Good idea on using a dictionary! I updated accordingly.

Python/pywarpx/picmi.py
```diff
@@ -2497,11 +2497,17 @@ class ParticleDiagnostic(picmistandard.PICMI_ParticleDiagnostic, WarpXDiagnostic
...
 
-    warpx_random_fraction: float, optional
-        Random fraction of particles to include in the diagnostic
-
-    warpx_uniform_stride: integer, optional
-        Stride to down select to the particles to include in the diagnostic
+    warpx_random_fraction: float or dict, optional
+        Random fraction of particles to include in the diagnostic. If a float
+        is given the same fraction will be used for all species, if a dictionary
+        is given the keys should be species names with the value specifying
```
dpgrote: It's a small detail, but it might be better to use species instances for the keys instead of the names. This would be consistent with the `species` input argument. Otherwise, everything looks good.

roelof-groenewald: I considered that but then opted for the name since we use
```
        # species list
        if self.species is None:
            species_names = pywarpx.particles.species_names
        elif np.iterable(self.species):
            species_names = [species.name for species in self.species]
        else:
            species_names = [self.species.name]
```
so if no species is specified we don't actually have a list of species objects, only the names. Or can I get the list of species objects from `pywarpx.particles`?

dpgrote: Good point. But the code could be changed to something like this
```
random_fraction = {}
if isinstance(self.random_fraction, dict):
    random_fraction_default = 1.
    for k, v in self.random_fraction.items():
        random_fraction[k.name] = v
else:
    random_fraction_default = self.random_fraction
```
Then below, in the call to `pywarpx.Bucket.Bucket`,
```
                                         random_fraction = random_fraction.get(name, random_fraction_default),
```

roelof-groenewald: Nice idea! Thanks!

dpgrote (approved):
Looks great! Thanks!

PR #4955, opened 2024-05-28
Make particle IDs valid in BoundaryScraping

RemiLehe:
@jcarlsson-aed recently [reported](https://gist.github.com/jcarlsson-aed/6a71bfcc7aac22b59aad4699a064e444) that the particle IDs of particles in `BoundaryScraping` do not match the particle IDs of the same particles when they are in the simulation volume. 

This PR fixes the issue by calling `make_valid` in the functor that copies scraped particles to the `ParticleBoundaryBuffer` (I was able to confirm this using @jcarlsson-aed's example.)

Many thanks to @ax3l for suggesting this fix!

TODO: 
- [x] Add a CI test.


ax3l (approved):
Excellent, thank you! :sparkles:

PR #4966, opened 2024-06-03
Make WarpX class final and make destructor private

lucafedeli88:
Since the WarpX class is (sadly ;-)) still a singleton, the WarpX instance should not be directly destructible (`WarpX::Finalize` should take care of that). This PR, by making the constructor private enforces this.

Moreover, the WarpX class is declared `final`, as we don't (currently) want to inherit the WarpX class.

### DIFF
--- a/Source/WarpX.H
+++ b/Source/WarpX.H
@@ -78,7 +78,7 @@
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


lucafedeli88:
Ah, we can't do that because we don't use WarpX as a singleton in the python interface!

PR #4967, opened 2024-06-03
Take Version and PicsarVersion out of the WarpX class

lucafedeli88:
This PR takes Version and PicsarVersion out of the WarpX class and puts them in a `warpx` namespace.
The goal is twofold: 1) reducing the complexity of the WarpX class 2) possibly avoid in the future the inclusion of the huge WarpX class where we only need the version of WarpX and Picsar.

With the occasion, the two functions are made `noexcept` and `[[nodiscard]]`.


ax3l (approved):
ok :)

ax3l:
Restarting CI

ax3l:
Restarting CI

ax3l:
@lucafedeli88 I think the GNU makefile needs a touch-up for this to pass.

lucafedeli88:
> @lucafedeli88 I think the GNU makefile needs a touch-up for this to pass.

@ax3l , I had a look at the GNU makefile. It seems to me that when we compile with the GNU makefile there is an AMReX tool that takes care of generating `Utils/WarpXVersion.H.in` .

The solution I chose is to leave `Utils/WarpXVersion.H.in` untouched and implement `warpx::Version` and `warpx::PicsarVersion` inside  `Utils/WarpXGetVersion.H/cpp`. It seems to work on my computer.

ax3l:
Uhoh, CI still said :(

lucafedeli88:
This PR will soon be obsolete: picsar_qed is being integrated directly into ablastr.

PR #4968, opened 2024-06-03
Slightly increase tolerance for embedded_circle test

lucafedeli88:
This test fails quite often due to small numerical errors. It is possible that the required tolerance for the self fields is a bit too small. This PR increases it from 1e-6 to 2e-6.


ax3l (approved):

ax3l:
Discussed with @roelof-groenewald and looks good to us. Thanks!

PR #4969, opened 2024-06-04
Add upper bound for weight of product particles from particle resampling

aveksler1:
A user parameter for an upper bound on particle weight is added for the `VelocityCoincidenceThinning` resampling algorithm. This is to avoid over-merging many particles into a few, heavy particles. Particles with a weight greater than the specific upper bound will be ignored during resampling. This has been an issue when resampling too frequently when trying to avoid exponentially increasing particle counts.


Python/pywarpx/picmi.py
```diff
@@ -137,6 +137,10 @@ class Species(picmistandard.PICMI_Species):
         Cells with fewer particles than this number will be
         skipped during resampling.
 
+    warpx_resampling_algorithm_max_w: float
```
roelof-groenewald: For the sake of clarity I think it would be better to name this `warpx_resampling_algorithm_max_weight`.

Source/Particles/Resampling/VelocityCoincidenceThinning.H
```diff
@@ -141,8 +142,15 @@ public:
                 const int ii = static_cast<int>(u_theta / dutheta);
                 const int jj = static_cast<int>(u_phi / duphi);
                 const int kk = static_cast<int>(u_mag / dur);
+                amrex::ParticleReal w_i = w[indices[i]];
 
-                bin_array[i] = ii + jj * n1 + kk * n1 * n2;
+                // Assign 0th index of bin_array to particles with weight too low to be merged
```
roelof-groenewald: For clarity
```suggestion
                // bin number 0 is used for particles that will not participate in merging
```

Source/Particles/Resampling/VelocityCoincidenceThinning.H
```diff
@@ -160,27 +169,35 @@ public:
                 const int ii = static_cast<int>((ux[indices[i]] - ux_min) / dux);
                 const int jj = static_cast<int>((uy[indices[i]] - uy_min) / duy);
                 const int kk = static_cast<int>((uz[indices[i]] - uz_min) / duz);
+                amrex::ParticleReal w_i = w[indices[i]];
 
-                bin_array[i] = ii + jj * n1 + kk * n1 * n2;
+                // Assign 0th index of bin_array to particles with weight too low to be merged
```
roelof-groenewald: Same updated comment here.

Source/Particles/Resampling/VelocityCoincidenceThinning.H
```diff
@@ -191,6 +208,7 @@ public:
         amrex::ParticleReal dur, dutheta, duphi;
         amrex::ParticleReal dux, duy, duz;
         amrex::ParticleReal ux_min, uy_min, uz_min, ux_max, uy_max;
+        amrex::ParticleReal m_max_w;
```
roelof-groenewald: Keeping with convention of the other member variables (and since `velocityBincalculator` is a `struct`, this might instead be
```suggestion
        amrex::ParticleReal max_weight;
```

Source/Particles/Resampling/VelocityCoincidenceThinning.cpp
```diff
@@ -24,6 +24,10 @@ VelocityCoincidenceThinning::VelocityCoincidenceThinning (const std::string& spe
         "Resampling min_ppc should be greater than or equal to 1"
     );
 
+    utils::parser::queryWithParser(
+        pp_species_name, "resampling_algorithm_max_w", m_max_w
```
roelof-groenewald:
```suggestion
        pp_species_name, "resampling_algorithm_max_weight", m_max_w
```

roelof-groenewald (changes requested):

Source/Particles/Resampling/VelocityCoincidenceThinning.H
```diff
@@ -160,27 +169,35 @@ public:
                 const int ii = static_cast<int>((ux[indices[i]] - ux_min) / dux);
                 const int jj = static_cast<int>((uy[indices[i]] - uy_min) / duy);
                 const int kk = static_cast<int>((uz[indices[i]] - uz_min) / duz);
+                amrex::ParticleReal w_i = w[indices[i]];
```
lucafedeli88:
```suggestion
                const amrex::ParticleReal w_i = w[indices[i]];
```
Hi @aveksler1 ! This should fix the clang-tidy CI test (the test enforces const-correctness, among many other things).

Source/Particles/Resampling/VelocityCoincidenceThinning.H
```diff
@@ -200,5 +218,6 @@ private:
     int m_ntheta, m_nphi;
     amrex::ParticleReal m_delta_ur;
     amrex::Vector<amrex::ParticleReal> m_delta_u;
+    amrex::ParticleReal m_max_w = 0;
```
dpgrote: Could this default to `std::numeric_limits<amrex::ParticleReal>::max()`? This would simplify the checks of the weight.

Source/Particles/Resampling/VelocityCoincidenceThinning.cpp
```diff
@@ -189,6 +195,10 @@ void VelocityCoincidenceThinning::operator() (WarpXParIter& pti, const int lev,
             // ones in the same momentum bin
             for (int i = cell_start; i < cell_stop; ++i)
             {
+                // skip merging if the momentum bin is 0
+                if (momentum_bin_number_data[sorted_indices_data[i]] == 0) {
+                    continue;
+                }
                 particles_in_bin += 1;
                 const auto part_idx = indices[sorted_indices_data[i]];
```
dpgrote: In the `if` below at line 211, could an additional check be added `total_weight > m_max_w`? This would limit the merging of particles to the same maximum weight. Doing it this way would also remove the need for the checks above, setting the bin number to zero.

roelof-groenewald: That is a great idea! Basically instead of setting a maximum weight that particles cannot be merged beyond, it would set a target weight for each merging cluster. But I think it would be important to **not** include the particle that pushed the `total_weight` over the edge. So the `if` statement would be:
```
if (
    (i == cell_stop - 1)
    || (momentum_bin_number_data[sorted_indices_data[i]] != momentum_bin_number_data[sorted_indices_data[i + 1]])
    || (total_weight + w[indices[sorted_indices_data[i + 1]] > m_target_weight * 2.0)
) {
```
where the new `m_target_weight` parameter is introduced. Note the times 2 since half the total weight is given to each product particle.

Source/Particles/Resampling/VelocityCoincidenceThinning.cpp
```diff
@@ -189,6 +196,10 @@ void VelocityCoincidenceThinning::operator() (WarpXParIter& pti, const int lev,
             // ones in the same momentum bin
             for (int i = cell_start; i < cell_stop; ++i)
             {
+                // skip merging if the momentum bin is 0
+                if (momentum_bin_number_data[sorted_indices_data[i]] == 0) {
+                    continue;
+                }
```
roelof-groenewald: No longer needed.
```suggestion
```

Python/pywarpx/picmi.py
```diff
@@ -137,6 +137,9 @@ class Species(picmistandard.PICMI_Species):
         Cells with fewer particles than this number will be
         skipped during resampling.
 
+    warpx_resampling_algorithm_target_weight: float
+        Weight that particles undergoing resampling will not exceed.
```
aveksler1:
```suggestion
        Weight that the product particles from resampling will not exceed.
```

roelof-groenewald (approved):
Looks good! Thanks for improvement!

Source/Particles/Resampling/VelocityCoincidenceThinning.cpp
```diff
@@ -24,6 +24,13 @@ VelocityCoincidenceThinning::VelocityCoincidenceThinning (const std::string& spe
         "Resampling min_ppc should be greater than or equal to 1"
     );
 
+    amrex::ParticleReal target_weight = 0;
+    if (utils::parser::queryWithParser(
+        pp_species_name, "resampling_algorithm_target_weight", target_weight
+    )) {
+        m_cluster_weight = target_weight * 2.0_prt;
```
dpgrote, line 32: It might be worth adding a comment here that the factor of 2 is needed because each cluster is reduced to 2 particles.

dpgrote (approved):
This looks good! Thanks! (Though I do have a minor suggestion.)

dpgrote (approved):

PR #1316, opened 2020-09-11
Adding Particle Extrema Reduced Diagnostics

Yin-YinjianZhao:
This PR adds a reduced diagnostic that outputs the max/min values of some particle information.
This PR is motivated by issue #1311 .

The outputs are:
```
xmin, xmax, ymin, ymax, zmin, zmax, pxmin, pxmax, pymin, pymax, pzmin, pzmax, gmin, gmax, wmin, wmax.
```
for regular species.

When `XZ` dimension is used, `ymin, ymax` are zeros.
When `RZ` dimension is used, `x = r*cos(theta), y = r*sin(theta)` are used to still obtain `xmin, xmax, ymin, ymax`.

PR #4973, opened 2024-06-05
Cleanup diagnostic dimension macros

dpgrote:
This PR cleans up some code in the diagnostics that was unnecessarily using the `WARPX_DIMS` macros. One specific change is making use of the `get_particle_position` routine instead of duplicating the particle transformations.

In `ParticleExtrema.cpp`, this PR also combines the reduction routines, as was recommended in PR #1316.

### DIFF
--- a/Source/Diagnostics/ReducedDiags/FieldEnergy.cpp
+++ b/Source/Diagnostics/ReducedDiags/FieldEnergy.cpp
@@ -102,8 +102,6 @@ void FieldEnergy::ComputeDiags (int step)
         const std::array<Real, 3> &dx = WarpX::CellSize(lev);
         const amrex::Real dV = dx[0]*dx[1]*dx[2];
 
-        Geometry const & geom = warpx.Geom(lev);
-
 #if defined(WARPX_DIM_RZ)
         amrex::Real const tmpEx = ComputeNorm2RZ(Ex, lev);
         amrex::Real const tmpEy = ComputeNorm2RZ(Ey, lev);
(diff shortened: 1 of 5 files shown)


Source/Diagnostics/ReducedDiags/FieldEnergy.cpp
```diff
@@ -100,13 +100,7 @@ void FieldEnergy::ComputeDiags (int step)
...
         // get cell size
         Geometry const & geom = warpx.Geom(lev);
-#if defined(WARPX_DIM_1D_Z)
-        auto dV = geom.CellSize(0);
-#elif defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
-        auto dV = geom.CellSize(0) * geom.CellSize(1);
-#elif defined(WARPX_DIM_3D)
-        auto dV = geom.CellSize(0) * geom.CellSize(1) * geom.CellSize(2);
-#endif
+        const amrex::Real dV = AMREX_D_TERM(geom.CellSize(0), *geom.CellSize(1), *geom.CellSize(2));
```
roelof-groenewald: While this is great for reducing code, I think a comment about what happens here would be good since on a cursory read `*geom.CellSize(1)` looks like it dereferences a pointer and not like it builds a product.

roelof-groenewald (approved):
Thanks a lot for this PR! It's great to reduce unnecessary code in this way. These changes are especially helpful in the RZ geometry case where `PIdx::x` will soon be replaced by `PIdx::r` (see #4667) and this PR will ensure that the code touched here does not break due to that upcoming change.

Source/Diagnostics/ReducedDiags/FieldEnergy.cpp
dpgrote: I'll rewrite this using warpx.CellSize which will remove the need for the `AMREX_D_TERM`.

roelof-groenewald: Nice! Yeah that is much more clear. Thanks!

PR #4978, opened 2024-06-08
Cleanup deposition and gather routines

dpgrote:
This PR cleans up the use of macros in the deposition and gather routines. There are many places where macros were being used that were not really necessary. The changes here reduce the code length by several hundred lines, and hopefully make it more readable and maintainable.

Most of the changes involve using XDim3 instead of length three arrays. This removed many checks for dimensionality, though does add a small number of extra calculations to unneeded quantities (outside of any loops). For example, the routines `LowerCorner` and `UpperCorner` now return XDim3 objects.

Several CI tests were updated because of changes in round-off.


Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -2559,6 +2559,7 @@ PhysicalParticleContainer::PushP (int lev, Real dt,
     if (do_not_push) { return; }
 
     const std::array<amrex::Real,3>& dx = WarpX::CellSize(std::max(lev,0));
+    const amrex::XDim3 dinv = amrex::XDim3{1._rt/dx[0], 1._rt/dx[1], 1._rt/dx[2]};
```
roelof-groenewald: These lines are repeated quite a lot. Would it make sense to directly grab `dinv` from `WarpX` instead of every time needing both lines?
```
    const std::array<amrex::Real,3>& dx = WarpX::CellSize(std::max(lev,0));
    const amrex::XDim3 dinv = amrex::XDim3{1._rt/dx[0], 1._rt/dx[1], 1._rt/dx[2]};
```

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -2591,16 +2592,14 @@ PhysicalParticleContainer::PushP (int lev, Real dt,
             const amrex::ParticleReal Bz_external_particle = m_B_external_particle[2];
 
             const std::array<amrex::Real,3>& xyzmin = WarpX::LowerCorner(box, lev, 0._rt);
+            const amrex::XDim3 gridmins = amrex::XDim3{xyzmin[0], xyzmin[1], xyzmin[2]};
```
roelof-groenewald: Same comment here as above, instead of doing
```
            const std::array<amrex::Real,3>& xyzmin = WarpX::LowerCorner(box, lev, 0._rt);
            const amrex::XDim3 gridmins = amrex::XDim3{xyzmin[0], xyzmin[1], xyzmin[2]};
```
in many places, can it be replaced with
```
            const amrex::XDim3 gridmins = WarpX::GridMins(box, lev, 0._rt);
```
or something similar?

roelof-groenewald (commented):
Thanks for this nice clean-up PR!

I haven't looked through all the changes yet, but here are two initial comments.

Source/Particles/PhysicalParticleContainer.cpp
dpgrote: Yes, definitely. I could add a routine `InvCellSize` that returns an `XDim3`.

Source/Particles/PhysicalParticleContainer.cpp
dpgrote: Yes. I've considered changing `LowerCorner` to return an `XDim3`.

Source/Particles/PhysicalParticleContainer.cpp
roelof-groenewald: Thanks!

Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -367,9 +327,7 @@ void doDepositionShapeN (const GetParticlePosition<PIdx>& GetPosition,
 
             doDepositionShapeNKernel<depos_order>(xp, yp, zp, wq, vx, vy, vz, jx_arr, jy_arr, jz_arr,
                                                   jx_type, jy_type, jz_type,
-                                                  relative_time,
-                                                  AMREX_D_DECL(dzi, dxi, dyi),
```
roelof-groenewald, line 371: The switch to `amrex::XDim3` is particularly nice here to avoid this confusing order of `z, x, y`.

roelof-groenewald (commented):
This looks good to me. A really nice improvement to the code readability. 

My only question before approving is: do we know whether these changes, specifically the change from using `std::array<amrex::Real, 3>` to `amrex::XDim3`, will impact the code performance?

dpgrote:
> My only question before approving is: do we know whether these changes, specifically the change from using `std::array<amrex::Real, 3>` to `amrex::XDim3`, will impact the code performance?

That's a good question. I would expect that it would not. There are some extra calculations, but they are outside the loops. I wouldn't expect a timing different between accessing an array element versus a struct attribute (for a simple struct). I haven't done any timings. It is something I could check. Comments from others would be helpful @ax3l  @RemiLehe @WeiqunZhang .

ax3l (approved):
Cool, very nice!

ax3l:
Yes, I see no performance pitfalls here.

ax3l:
Discussed with @roelof-groenewald: @dpgrote , can you please reset the checksums of the failing tests that are nearly the same (`PEC_particle`, `BTD_rz`)?

dpgrote:
> Discussed with @roelof-groenewald: @dpgrote , can you please reset the checksums of the failing tests that are nearly the same (`PEC_particle`, a BTD test in RZ, ...)?

Yes, will do. Not sure why that one test failed.

roelof-groenewald (approved):

PR #4983, opened 2024-06-11
Fix FieldProbe Check: Particle Shape

### LINKED ISSUE #4958: FieldProbe Point: Diagnostics Errors for Shape/Interpolation
When using input files for point diagnostics, I set ```algo.particle_shape = 3``` and ```interp_order = 1```. During execution, I encountered the following error:
```
1::Assertion `interp_order <= WarpX::nox' failed, file "/public/home/ljz_gpu/warpx/Source/Diagnostics/ReducedDiags/FieldProbe.cpp", line 155, Msg:
### ERROR   : Field probe interp_order should be less than or equal to
#            algo.particle_shape
!!!
SIGABRT
0::Assertion `interp_order <= WarpX::nox' failed, file "/public/home/ljz_gpu/warpx/Source/Diagnostics/ReducedDiags/FieldProbe.cpp", line 155, Msg:
### ERROR   : Field probe interp_order should be less than or equal to
#            algo.particle_shape
```
The backtrace file is as follows.
[Backtrace.0.txt](https://github.com/ECP-WarpX/WarpX/files/15492358/Backtrace.0.txt)

I tried modifying the values of the two variables, and the program only runs when ```interp_order = 0```, but the data from the point diagnostics are all zero.

Here is my input file：
[inputs.txt](https://github.com/ECP-WarpX/WarpX/files/15492380/inputs.txt)

Additionally, if I want to track the electromagnetic fields and current density at fixed grid points, as well as the velocities of fixed macroparticles, should I use Reduced Diagnostics?

ax3l:
The constructor of FieldProbe might be called earlier than the WarpX class parameter init. That could lead to relying on an uninitialized particle shape static. Use the parser instead, similar to our general efforts to reduce the static members stored in the WarpX class.

Fix #4958

### DIFF
--- a/Source/Diagnostics/ReducedDiags/FieldProbe.cpp
+++ b/Source/Diagnostics/ReducedDiags/FieldProbe.cpp
@@ -154,7 +154,12 @@ FieldProbe::FieldProbe (const std::string& rd_name)
             ablastr::warn_manager::WarnPriority::low);
     }
 
-    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(interp_order <= WarpX::nox ,
+    // ensure assumption holds: we read the fields in the interpolation kernel as they are,
+    // without further communication of guard/ghost/halo regions
+    int particle_shape;
+    const ParmParse pp_algo("algo");
+    utils::parser::getWithParser(pp_algo, "particle_shape", particle_shape);
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(interp_order <= particle_shape ,
                                      "Field probe interp_order should be less than or equal to algo.particle_shape");
     if (ParallelDescriptor::IOProcessor())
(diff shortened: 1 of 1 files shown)


roelof-groenewald:
Thanks @ax3l. Is there still a separate problem with the data being all zero or is that not actually an issue?

> I tried modifying the values of the two variables, and the program only runs when interp_order = 0, but the data from the point diagnostics are all zero.

ax3l:
Yes, this PR fixes that and now `interp_order` can go up to the particle shape again.
I think the object `FieldProbe`  is constructed too early.

That the values are zero for exactly `interp_order = 0` seems to be a separate issue, we need to check if our NGP interpolation properly works...

roelof-groenewald (approved):

PR #4985, opened 2024-06-11
Cleanup stencils for filtering

dpgrote:
This PR simplifies the setup of the stencils for the different dimensionalities. Now it relies only on AMREX_SPACEDIM and is independent of the type of geometry (whether XZ or RZ for example).

Also, this fixes what was almost certainly a bug in the 1D filter on the GPU, where the wrong index was being used when getting the stencil value in the loop.


Source/Filter/BilinearFilter.cpp
```diff
@@ -64,34 +64,25 @@ void BilinearFilter::ComputeStencils(){
...
-    compute_stencil(stencil_z, npass_each_dir[0]);
 #endif
+
     slen = stencil_length_each_dir.dim3();
-#if defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
+#if AMREX_SPACEDIM < 3
     slen.z = 1;
 #endif
-#if defined(WARPX_DIM_1D_Z)
+#if AMREX_SPACEDIM < 2
```
ax3l: For the logic here, can we express this also as `or`-concatenated branches of the `WARPX_` macros?

For clarity reasons (e.g., is RZ truly 2D or something else), we factored out all (most?) use of the `AMREX_SPACEDIM` macro in WarpX a while ago :)

dpgrote: Yes, I can use the `WARPX_DIM` macros. But I was thinking of this differently. The code is generic to the geometry - it is only the spatial dimensions of the grid that matter. So RZ and XZ are both 2D. I thought it was nice to have the code written without the `WARPX_DIM` macros emphasizing that the code is only dependent on the number of dimensions (and particularly, does not depend on the ordering of dimensions).

Source/Filter/Filter.cpp
```diff
@@ -148,7 +148,7 @@ void Filter::DoFilter (const Box& tbx,
         for         (int iz=0; iz < slen_local.z; ++iz){
             for     (int iy=0; iy < slen_local.y; ++iy){
                 for (int ix=0; ix < slen_local.x; ++ix){
-                    Real sss = sx[ix]*sz[iy];
+                    Real sss = sx[ix]*sy[iy];
```
ax3l: Is this what we want? Isn't this labelled `sz` in WarpX?

Source/Filter/Filter.cpp
```diff
@@ -174,7 +174,7 @@ void Filter::DoFilter (const Box& tbx,
         for         (int iz=0; iz < slen_local.z; ++iz){
             for     (int iy=0; iy < slen_local.y; ++iy){
                 for (int ix=0; ix < slen_local.x; ++ix){
-                    Real sss = sz[iy];
+                    Real sss = sx[ix];
```
ax3l: Is this what we want? Isn't this labelled `sz` in WarpX?

Source/Filter/Filter.cpp
```diff
@@ -298,19 +298,19 @@ void Filter::DoFilter (const Box& tbx,
...
+#if AMREX_SPACEDIM == 3
                     const Real sss = sx[ix]*sy[iy]*sz[iz];
-#elif defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
-                    const Real sss = sx[ix]*sz[iy];
-#else
-                    const Real sss = sz[ix];
+#elif AMREX_SPACEDIM >= 2
+                    const Real sss = sx[ix]*sy[iy];
+#elif AMREX_SPACEDIM == 1
+                    const Real sss = sx[ix];
```
ax3l: Is this what we want? Isn't this labelled `sz` in WarpX?

Source/Filter/Filter.cpp
dpgrote: This is intended to be independent of what the coordinates are and so work for any system that is 2d. Perhaps I can change the names to use `1` and `2` instead of `x` and `y`,

Source/Filter/Filter.cpp
dpgrote: Same here, maybe call these `i1` and `s1`. This will not always be `z` - these changes are here as a step to introduce 1D radial cylindrical and spherical coordinates.

ax3l (approved):

PR #4964, opened 2024-06-02
Simplify main file

lucafedeli88:
This PR simplifies the main file of WarpX by:
- grouping together the initialization of MPI, FFT, and AMReX (now `WarpXInit.H/cpp`)
-  moving `amrex::Print() << ablastr::warn_manager::GetWMInstance().PrintGlobalWarnings("THE END");` to the end of `WarpX.Evolve()`
- moving `WarpX::Finalize();` before the stopping of the timer
- changing empty lines to show clearly the "warpx block" and the two "timer blocks"

PR #4990, opened 2024-06-13
Revert "group initialization and cleanup of external libraries (#4964)"

ax3l:
This reverts commit c7b203e5670f57e6b9955db8c3f80d718f070b1c.

Try if reverting #4964 unbreaks the failing `collisionXYZ` test.

We can try to reintroduce the PR again later.


ax3l:
Nope, not causing it... maybe parallelism changed on the runner and we did not nail it down in the job script

PR #4981, opened 2024-06-10
[WIP] Add ability to load external particle fields from file

RemiLehe:
We currently have the ability to load an external **grid** field from a file (thanks to https://github.com/ECP-WarpX/WarpX/pull/3584). 

This PR provide a similar ability for the external **particle** fields.

The external **grid** and **particle** fields differ e.g. in the case of the EM solver:
- the external **grid** fields are applied to the grid at `t=0` and serve as initial condition to the Maxwell solver.
- by contrast, the external **particle** fields are applied to the particles directly and are not affected by the Maxwell solver.

TODO:
- [ ] Add explanation of how we are doing this: we add fields to `aux`. This means that this will automatically work with all other modules that gather the fields: particle push, ionization routine, etc.
- [ ] Make it work with momentum conserving (in follow-up PR?)
- [ ] Check MR
- [ ] Make the data structures for external grid fields and external particle fields more similar.

PR #4997, opened 2024-06-17
Add ability to load external particle fields from file

oshapoval:
Description of  PR [#4981](https://github.com/ECP-WarpX/WarpX/pull/4981) @RemiLehe:

We currently have the ability to load an external grid field from a file (thanks to https://github.com/ECP-WarpX/WarpX/pull/3584).

This PR provides a similar ability for the external particle fields.

The external grid and particle fields differ e.g. in the case of the EM solver:

- The external grid fields are applied to the grid at t=0 and serve as initial condition to the Maxwell solver.
- By contrast, the external particle fields are applied to the particles directly and are not affected by the Maxwell solver.

In this PR, support for external particle fields is achieved by reusing many of the structures that are used for the external grid fields. For instance, we store the fields from file into the field `Efield_fp_external` and `Bfield_fp_external` in both cases. (As a result, using external particle fields and external grid fields from file **at the same time** is not possible.) However:
- in the case of the external grid fields, `Efield_external` is added to the grid fields `Efield_fp` directly.
- in the case of the external particle fields, `Efield_external` is instead added to the fields from which the particle gather `Efield_aux`. In some case (e.g. energy-conserving gather), `Efield_fp` and `Efield_aux` used to be aliases of each other. This is no longer the case if the external particle fields from file are used.


roelof-groenewald:
It would be good to include the divergence cleaning from #4944 to these particle fields as well.

Source/Initialization/ExternalField.cpp
```diff
@@ -172,11 +172,6 @@ ExternalFieldParams::ExternalFieldParams(const amrex::ParmParse& pp_warpx)
     //
     //  External fields from file
     //
-
-    if (E_ext_grid_type == ExternalFieldType::read_from_file ||
-        B_ext_grid_type == ExternalFieldType::read_from_file){
-            const std::string read_fields_from_path="./";
-            pp_warpx.query("read_fields_from_path", external_fields_path);
-    }
```
RemiLehe: TODO: add a comment saying that this is used both for external grid fields and external particles fields.

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1383,6 +1398,41 @@ WarpX::LoadExternalFieldsFromFile (int const lev)
         ReadExternalFieldFromFile(m_p_ext_field_params->external_fields_path, Efield_fp_external[lev][0].get(), "E", "x");
         ReadExternalFieldFromFile(m_p_ext_field_params->external_fields_path, Efield_fp_external[lev][1].get(), "E", "y");
         ReadExternalFieldFromFile(m_p_ext_field_params->external_fields_path, Efield_fp_external[lev][2].get(), "E", "z");
+#endif
+    }
+
+    // External particle fields
+
+    if (mypc->m_B_ext_particle_s == "read_from_file") {
+        std::string external_fields_path = "";
```
RemiLehe:
```suggestion
        std::string external_fields_path;
```

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1383,6 +1398,41 @@ WarpX::LoadExternalFieldsFromFile (int const lev)
...
+        ReadExternalFieldFromFile(external_fields_path, B_external_particle_field[lev][1].get(), "B", "t");
+        ReadExternalFieldFromFile(external_fields_path, B_external_particle_field[lev][2].get(), "B", "z");
+#else
+        ReadExternalFieldFromFile(external_fields_path, B_external_particle_field[lev][0].get(), "B", "x");
+        ReadExternalFieldFromFile(external_fields_path, B_external_particle_field[lev][1].get(), "B", "y");
+        ReadExternalFieldFromFile(external_fields_path, B_external_particle_field[lev][2].get(), "B", "z");
+#endif
+    }
+    if (mypc->m_E_ext_particle_s == "read_from_file") {
+        std::string external_fields_path = "";
```
RemiLehe:
```suggestion
        std::string external_fields_path;
```

RemiLehe (approved):

RemiLehe (approved):

PR #4986, opened 2024-06-12
CMake: heFFTe Support

ax3l:
Add optional `-DWarpX_HEFFTE` support to WarpX CMake build logic. Document how to provide the heFFTe dependency locally and on Perlmutter (NERSC).

- [x] CMake
- [x] Fix upstream issues
  - [x] macOS build error: https://github.com/icl-utk-edu/heffte/pull/49
- [x] Update build docs (tables)
- [x] Update Perlmutter Docs
- [ ] Conda-Forge Package: https://github.com/conda-forge/staged-recipes/pull/26633
- [x] Update Spack envs
- [x] tested

PR #5006, opened 2024-06-19
Fix: Perlmutter heFFTe CPU

ax3l:
Follow-up to #4986


jwestern:
Perlmutter compilation still fails for me, same error messages as in the original issue (doing GPU build). Also fails, same messages, for release 24.06 (commit a3cd47d). Will continue going to earlier releases to hopefully avoid the error.

ax3l:
X-ref: #5012

PR #5007, opened 2024-06-19
Fix ccache for macos

WeiqunZhang:
Need to ignore time macros. Brew seems to inject them from some dependent packages.


ax3l (approved):
Cool, did not know it can ignore that.

Do you want to add this to further CI tests?

PR #5009, opened 2024-06-21
Enable general weight scaling for radial particle distributions

dpgrote:
This PR adds the new input parameter `radial_numpercell_power` which specifies the radial power for the number of particles per cell when using cylindrical and spherical geometry. The default value of 0 gives uniform particles per cell with increasing weight with the radius. The value 1 gives uniformly weighted particles when cylindrical with uniform density.

Note that this removes the parameter `radially_weighted` which is no longer used and could not be changed from its default value anyway.

Several CI benchmarks are updated due to changes in round-off.

Note that this replaces PR #4342


Examples/Tests/electrostatic_sphere/analysis_electrostatic_sphere.py
```diff
@@ -142,6 +142,7 @@ def calculate_error(E_axis, xmin, dx, nx):
 print("L2 error along x-axis = %s" %L2_error_x)
 print("L2 error along y-axis = %s" %L2_error_y)
 print("L2 error along z-axis = %s" %L2_error_z)
+print("L2 error tolerance = %s" %l2_tolerance)
```
dpgrote: These print statements were added to provide more output in cases when the checks fail. This was helpful in getting the new CI test working.

Examples/Tests/electrostatic_sphere/inputs_rz_uniform_weighting
```diff
@@ -0,0 +1,45 @@
...
+# Order of particle shape factors
+algo.particle_shape = 1
+
+my_constants.n0 = 1.49e6
+my_constants.R0 = 0.1
+
+electron.charge = -q_e
+electron.mass = m_e
+electron.injection_style = "NUniformPerCell"
+electron.num_particles_per_cell_each_dim = 6 2 2
```
dpgrote: Note here that more particles are needed in this test compared to the non-uniformly weighted case to get to the same tolerance. This is because the error is being checked on axis which has fewer particles when doing uniform weighting.

Examples/Tests/langmuir/PICMI_inputs_rz.py
```diff
@@ -51,7 +51,7 @@
 ##########################
 
 uniform_plasma = picmi.UniformDistribution(density = density,
-                                           upper_bound = [+18e-6, +18e-6, None],
+                                           upper_bound = [+18e-6, None, +40e-6],
```
dpgrote: This was a bug found in this input file that made this test fail with the updated code.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1168,8 +1163,21 @@ PhysicalParticleContainer::AddPlasma (PlasmaInjector const& plasma_injector, int
                 const Real theta = (nmodes == 1 && rz_random_theta)?
                     (2._rt*MathConst::pi*amrex::Random(engine)):
                     (2._rt*MathConst::pi*r.y + theta_offset);
+
+                // Adjust the particle radius to produce the correct distribution.
+                // Note that this may shift particles outside of the current tile,
+                // but this is Ok since particles will be redistributed afterwards.
+                // The tile_realbox.contains check above ensures
+                // that the "logical" space is uniformly filled.
+                amrex::Real const xb = std::pow(pos.x/rmax, 1._rt/(2._rt - radial_weight_power))*rmax;
```
atmyers: I see where the weights and particle radii are adjusted for `radial_weight_power`, but where does the number of particles per cell get changed? Doesn't this need to go up if `radial_weight_power = 0`?

dpgrote: This modifies the particle radii, ending up with different numbers of particles per cell depending on the radius if the power is not equal to 1. This is relying on the call to `Redistribute` afterwards to put the particles in the correct blocks.

Docs/source/usage/parameters.rst
```diff
@@ -1154,9 +1154,16 @@ Particle initialization
...
 
-* ``<species_name>.radially_weighted`` (`bool`) optional (default `true`)
-    Whether particle's weight is varied with their radius. This only applies to cylindrical geometry.
-    The only valid value is true.
+* ``<species_name>.radial_weight_power`` (`float`) optional (default `1.`)
+    With cylindrical geometry, specifies the radial power of the particle weight.
+    The particle weights will be proportional to :math:`(r/rmax)^p`, where :math:`r` is the particle initial radius,
+    :math:`rmax` the maximum radius that particles are loaded, and :math:`p` is the specified value.
+    This must be less than 2.
+    When the value is 1, the weight will increase with radius, producing a uniform number of particles per cell.
```
JustinRayAngus: When the value is 1, the weight will increase with radius, producing a uniform number of particles per cell _for a homogeneous density profile_.

What happens when the density profile is non-uniform?

Source/Particles/PhysicalParticleContainer.cpp
JustinRayAngus: I'm not sure I understand what's going on here. This implementation is different than what is used in PICNIC. In PICNIC, the number of particle per cell is scaled by (r/r0)^p with r0 the minimum cell centered radius of all cells included in the density profile being created. Within each cell, the weight of each particle is the same for all values of p. The distribution of the particles is always uniform in the cell.

It looks like in this implementation that even particles within a cell can have varying weights, but I still don't understand why the particle positions need to be varied based on how the weight scaling is done.

atmyers: Ah okay. So the total number of particles is the same regardless of `radial_weight_power`, they are just distributed differently.

atmyers (approved):

JustinRayAngus:
It seems there may be an issue when specifying xmin for the initial plasma profile.
I tested simulations of a uniform density in RZ using xmax = 1.54e-2 and xmin = 0.2e-2 and varying the radial weight power from 1, to 0.5 ,to 0. The output result at t = 0 is shown below. The lower bound is not correct for radial weight power < 1.

radial_power_xmin

JustinRayAngus:
There may also be a problem with non-uniform densities. I used a gaussian density profile with radial_weight_power = 0 and the particle weights are not uniform. 

Also, there seems to be a slight variation of the initial density profile with Z, despite the fact that the specified initial density profile only depends on x (r?)

`electrons.profile = "parse_density_function"`
`electrons.density_function(x,y,z) = if(x < R0, n0*exp(-x^2/(1.e-2)^2), 0.0)`

JustinRayAngus (changes requested):
See comments above for items to address.

Docs/source/usage/parameters.rst
dpgrote: It should reproduce a nonuniform density. The shifting of the particles first produces a uniform density and then the particle weights are multiplied by the user input density.

dpgrote:
> It seems there may be an issue when specifying xmin for the initial plasma profile. I tested simulations of a uniform density in RZ using xmax = 1.54e-2 and xmin = 0.2e-2 and varying the radial weight power from 1, to 0.5 ,to 0. The output result at t = 0 is shown below. The lower bound is not correct for radial weight power < 1.
> 

Yes, this needs correcting, allowing a nonzero min. This was on my todo list but slipped.

dpgrote:
> There may also be a problem with non-uniform densities. I used a gaussian density profile with radial_weight_power = 0 and the particle weights are not uniform.
> 
> Also, there seems to be a slight variation of the initial density profile with Z, despite the fact that the specified initial density profile only depends on x (r?)
> 
> `electrons.profile = "parse_density_function"` `electrons.density_function(x,y,z) = if(x < R0, n0*exp(-x^2/(1.e-2)^2), 0.0)`

The power modifyies the layout of the particles, so with power = 0, the number of particles per cell will increase with the radius. The weights are adjusted to produce the desired density.

I don't know why there would be a variation in z. I can look into it.

JustinRayAngus:
> > There may also be a problem with non-uniform densities. I used a gaussian density profile with radial_weight_power = 0 and the particle weights are not uniform.
> > Also, there seems to be a slight variation of the initial density profile with Z, despite the fact that the specified initial density profile only depends on x (r?)
> > `electrons.profile = "parse_density_function"` `electrons.density_function(x,y,z) = if(x < R0, n0*exp(-x^2/(1.e-2)^2), 0.0)`
> 
> The power modifying the layout of the particles, so with power = 0, the number of particles per cell will increase with the radius. The weights are adjusted to produce the desired density.
> 
> I don't know why there would be a variation in z. I can look into it.

The z-variation shows up in the density diagnostic. It occurs for all values of the radial_weight_power.

JustinRayAngus:
> The power modifying the layout of the particles, so with power = 0, the number of particles per cell will increase with the radius. The weights are adjusted to produce the desired density.

A desired capability is the ability to use uniform weight particles even for non-uniform density profiles. Perhaps this capability is a different PR. In PICNIC, we use a uniform_particle_weight flag to enforce this. The specified particles per cell then applies to the location on the grid with the smallest non-zero value. This cell is used to compute the particle weights, which is then fixed and used to set the number of particles in a cell for all other cells based on the density profile.

dpgrote:
> > > There may also be a problem with non-uniform densities. I used a gaussian density profile with radial_weight_power = 0 and the particle weights are not uniform.
> > > Also, there seems to be a slight variation of the initial density profile with Z, despite the fact that the specified initial density profile only depends on x (r?)
> > > `electrons.profile = "parse_density_function"` `electrons.density_function(x,y,z) = if(x < R0, n0*exp(-x^2/(1.e-2)^2), 0.0)`
> > 
> > 
> > The power modifying the layout of the particles, so with power = 0, the number of particles per cell will increase with the radius. The weights are adjusted to produce the desired density.
> > I don't know why there would be a variation in z. I can look into it.
> 
> The z-variation shows up in the density diagnostic. It occurs for all values of the radial_weight_power.

Can you give the input file that show this?

dpgrote:
> > The power modifying the layout of the particles, so with power = 0, the number of particles per cell will increase with the radius. The weights are adjusted to produce the desired density.
> 
> A desired capability is the ability to use uniform weight particles even for non-uniform density profiles. Perhaps this capability is a different PR. In PICNIC, we use a uniform_particle_weight flag to enforce this. The specified particles per cell then applies to the location on the grid with the smallest non-zero value. This cell is used to compute the particle weights, which is then fixed and used to set the number of particles in a cell for all other cells based on the density profile.

Ok. This would be a different PR. In Warp I had a method where the total number of particles is specified and the particles are loaded to match the density profile. Is this the capability needed? Maybe then this PR is less important?

JustinRayAngus:
> Ok. This would be a different PR. In Warp I had a method where the total number of particles is specified and the particles are loaded to match the density profile. Is this the capability needed? Maybe then this PR is less important?

The typical way that we load particles for cylindrical and spherical sims is to use a combination of equal macro particles per cell and equal macro particle weights. We typically use equal weights for a certain region near axis up to some distance r1. Then we use equal macro particles per cell for r1 to r2, and then equal weights again from r2 to r3. 

I think this PR would still be useful, as it provides an easy means to vary the weights as desired for uniform density profiles. In PICNIC, we scale with the Jacobian, so that in spherical coordinates the scaling is with r^{2p}.

JustinRayAngus:
> Also, there seems to be a slight variation of the initial density profile with Z, despite the fact that the specified initial density profile only depends on x (r?)
> 
> `electrons.profile = "parse_density_function"` `electrons.density_function(x,y,z) = if(x < R0, n0*exp(-x^2/(1.e-2)^2), 0.0)`

This issue is due to a bug in the input script. Replacing x^2 with x^2 + y^2 in the parse_density function fixes the random variation in Z observed.

dpgrote:
> It seems there may be an issue when specifying xmin for the initial plasma profile. I tested simulations of a uniform density in RZ using xmax = 1.54e-2 and xmin = 0.2e-2 and varying the radial weight power from 1, to 0.5 ,to 0. The output result at t = 0 is shown below. The lower bound is not correct for radial weight power < 1.

This should now be fixed.

JustinRayAngus (approved):
Thanks Dave,

This looks good now. Just need to fix the CI tests and then this PR can be merged.

PR #5011, opened 2024-06-24
Fix gatherParticlesFromEmbeddedBoundaries 

lucafedeli88:
This PR is an attempt at fixing issues in `gatherParticlesFromEmbeddedBoundaries` that seem to cause the `embedded_circle` test to fail frequently.
Essentially the fix  consist in touching the particles' tiles before the OMP loop.


RemiLehe:
@atmyers Could you confirm that calling `DefineAndReturnParticleTile` on each tile, after adding runtime components, is the recommended way to fix the issue?

atmyers (approved):

PR #5013, opened 2024-06-24
Doc: Latest Commit of BLAS++/LAPACK++

ax3l:
For BLAS++ and LAPACK++, switch the documented HPC dependency install logic from using the latest commit on `master` to the latest release `2024.05.31`. Using tags instead of living branches increases robustness and reproducibility with regards to future changes.


dpgrote:
This looks good. Is there a script to update all of the files to a newer version?

ax3l:
Thanks!
No script, modern IDEs (Clion, VS Code, Atom, etc.) support features such as "replace in all files".

dpgrote:
I was thinking a Python script that could loop over the files and make the replacements. It could do the replacements directly or perhaps do something like running `sed` on each of the files. This would be fairly trivial to write and would be much better than updating those 30 files each time by hand. It could live in `Tools/DevUtils`.

Or, something like this
```
git grep -l blaspp | xargs sed -i 's/blaspp-master/blaspp-2024.05.31/g'
git grep -l lapackpp | xargs sed -i 's/lapackpp-master/lapackpp-2024.05.31/g'
```
Maybe you did something like this originally?

RemiLehe (approved):

PR #5015, opened 2024-06-24
[WIP] Split Particle Container Init Logic

ax3l:
Currently, the particle init is done very early in WarpX, which complicates new logic for when we want to query large, distributed fields to initialize particles from field values (external files of density and temperature maps, etc.).

This PR minimizes the logic done in the particle container constructor. We now init data explicitly in `InitData()` calls. This makes it easier to create particle container variables and meshes and then initialize particles based on initialized meshes at a later point in the logic.

Isolated from #4754 for separate testing & merging.


ax3l:
I am back from vacation working on this again now.

ax3l:
Back from parental leave (part 1) and working on this again.

ax3l:
Luckily this is not needed after all for field loading from files #6062. Thanks @WeiqunZhang !

PR #5016, opened 2024-06-25
Simplify VelocityProperties.cpp

lucafedeli88:
This PR simplifies the logic to choose the bulk velocity direction in `Simplify VelocityProperties.cpp`
It also adds a check to ensure that the string obtained from the parser is not empty.


RemiLehe (approved):
Thanks for cleaning up the code!

PR #5021, opened 2024-06-28
obvious bug fix in ElasticCollisionPerez

JustinRayAngus:
min_N in the ElasticCollisionPerez function is defined erroneously using amrex::max(). This bug can lead to significantly incorrect scattering physics when the number of particles in a cell for two species interacting via Coulomb collisions is much different. 

This PR fixes this issue by replacing amrex::max() with amrex::min().


RemiLehe:
Thanks a lot for this PR! 
This is indeed quite a jarring bug. We should update our CI tests in the future ; they should have caught it!

RemiLehe:
Luckily, we have the correct code in the other binary collision modules:
https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/Collision/BinaryCollision/NuclearFusion/NuclearFusionFunc.H#L156
https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.H#L121

RemiLehe (approved):

JustinRayAngus:
> Luckily, we have the correct code in the other binary collision modules:
> 
> https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/Collision/BinaryCollision/NuclearFusion/NuclearFusionFunc.H#L156
> 
> https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.H#L121

Yes. I checked those too after finding this bug.

JustinRayAngus:
Looks like a few CI tests need to be re-baselined. @dpgrote are you able do this Dave?

dpgrote:
> Looks like a few CI tests need to be re-baselined. @dpgrote are you able do this Dave?

If you follow the links to get to the page that is showing the error, it prints out the updated benchmark values. This can be copied and pasted into the appropriate file in the `Regression/Checksum/benchmarks_json` directory.

PR #5022, opened 2024-06-28
Add random seed to Examples/Tests/collision/inputs_3d

dpgrote:
This adds an explicit random seed to the input file for the CI test `collisionXYZ`.

This should fix the random failures of this test.


JustinRayAngus:
Looks like this will have conflicts with the bug fix PR just before this PR.

RemiLehe (approved):

PR #5024, opened 2024-07-02
Update default behavior for gathering with direct deposition

RemiLehe:
In order for the PIC algorithm to be approximately energy-conserving (i.e. energy-conserving in the limit of small timesteps) with direct deposition, the galerkin gathering needs to be turned off with direct deposition.


RemiLehe:
The checksum failures make sense: they occur either for PSATD tests (which uses direct deposition by default) or for the Python Langmuir tests (which, for some reason, force the use of direct deposition: https://github.com/ECP-WarpX/WarpX/blob/development/Examples/Tests/langmuir/PICMI_inputs_3d.py#L86)

EZoni:
Benchmarks updated in [42eb259](https://github.com/ECP-WarpX/WarpX/pull/5024/commits/42eb259e2d08ded60e4fb8727e68995baee79d52), everything looks good.

EZoni (approved):

PR #5028, opened 2024-07-04
C++17 mordernization: replace std::is_same<T1, T2>::value with std::is_same_v<T1, T2>

lucafedeli88:
This PR modernizes the use of `std::is_same<T1, T2>::value` by replacing it with `std::is_same_v<T1, T2>`
(see https://en.cppreference.com/w/cpp/types/is_same)


dpgrote (approved):
Looks good. Thanks!

PR #3991, opened 2023-06-14
Add fluids in WarpX

johnson452:
This PR adds fluid species in WarpX. 

The fluid species evolve according to the Lorentz force ; details have been added in the documentation.

Fluid_Loop
[Flood_loop_editable.pptx](https://github.com/ECP-WarpX/WarpX/files/12772121/Flood_loop_editable.pptx)
[WarpX_Cold_Rel_Fluids.pdf](https://github.com/ECP-WarpX/WarpX/files/12886437/WarpX_Cold_Rel_Fluids.pdf)

PR #5031, opened 2024-07-05
Remove debug build type from Langmuir 2D fluid test

EZoni:
While working on another PR, I noticed that the CI test `Langmuir_fluid_2D` was compiled in debug mode.

The test was added originally in #3991 and it might be that the debug build type was left over unintentionally.

In general, I think we avoid running CI tests in debug mode, in order to keep the runtime of the tests as low as possible (the current runtime of this test in debug mode is around 80 seconds). 

However, we might have changed policy in the last months or there might have been specific reasons for running this test in debug mode and I might not be up-to-date, so feel free to let me know and close this PR without merging if it is not relevant.

### DIFF
--- a/Regression/WarpX-tests.ini
+++ b/Regression/WarpX-tests.ini
@@ -1173,7 +1173,7 @@ inputFile = Examples/Tests/langmuir_fluids/inputs_2d
 runtime_params =
 dim = 2
 addToCompileString =
-cmakeSetupOpts = -DWarpX_DIMS=2 -DCMAKE_BUILD_TYPE=Debug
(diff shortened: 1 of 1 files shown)


EZoni:
Note that I happened to open this PR directly from GitHub, so the head branch `EZoni-langmuir-fluid-2d-test-debug` is on `ECP-WarpX/WarpX` and not on my own fork. I will delete the branch once the PR is either merged or closed. If that's not okay, I can also close this PR and open an identical one more properly from my own fork.

RemiLehe (approved):

PR #5034, opened 2024-07-08
Remove work-around in regression testing script

EZoni:
Based on the comment https://github.com/AMReX-Codes/regression_testing/issues/136#issuecomment-2180755627, the issue reported in https://github.com/AMReX-Codes/regression_testing/issues/136 should now be fixed thanks to https://github.com/AMReX-Codes/regression_testing/pull/141.

If that's the case, it should be possible to remove the work-around that had been introduced in https://github.com/ECP-WarpX/WarpX/pull/4882.

### DIFF
--- a/run_test.sh
+++ b/run_test.sh
@@ -80,8 +80,6 @@ cd -
 
 # Clone the AMReX regression test utility
 git clone https://github.com/AMReX-Codes/regression_testing.git
-# FIXME: https://github.com/AMReX-Codes/regression_testing/issues/136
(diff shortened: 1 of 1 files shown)


RemiLehe (approved):

RemiLehe:
Hm, it seems that one of the CI tests is failing.

EZoni:
@RemiLehe 
We might need to rebase after #5037, although I'm not sure why this failure showed up here specifically.

PR #5037, opened 2024-07-09
CI: fix dimensionality of test `Python_ohms_law_solver_EM_modes_rz`

EZoni:
This CI failure came up in #5034, although I think it is not related to the changes in that PR.

This seems to be an RZ test, but `dim = 1` was set in WarpX-tests.ini. I think it should be replaced with `dim = 2`.

@roelof-groenewald 
I think the test was originally introduced in #4161. Would you be able to confirm that this change makes sense?

Not sure how this test could build and pass so far. Maybe it works when the 2D build is available thanks to other tests, but it doesn't work if the test is run alone?


RemiLehe (approved):

roelof-groenewald:
Thanks for fixing @EZoni. I'm also not sure why this would be a problem since in WarpX `rz` is different from `2d`. I'd expect that all RZ tests are collected and the `dim` parameter is simply ignored. But perhaps that is not the case. I'm glad the test is passing again.

PR #5040, opened 2024-07-10
generalizing temperature calc in Coulomb scattering method to weighted particles

JustinRayAngus:
The ComputeTemperature function called from the Coulomb scattering method currently assumes the particles have equal weights. This PR generalizes the temperature calculation to particles of arbitrary weights.

### DIFF
--- a/Source/Particles/Collision/BinaryCollision/Coulomb/ComputeTemperature.H
+++ b/Source/Particles/Collision/BinaryCollision/Coulomb/ComputeTemperature.H
@@ -14,6 +14,7 @@ template <typename T_index, typename T_R>
 AMREX_GPU_HOST_DEVICE
 T_R ComputeTemperature (
     T_index const Is, T_index const Ie, T_index const * AMREX_RESTRICT I,
+    T_R const * AMREX_RESTRICT w,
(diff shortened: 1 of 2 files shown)


dpgrote:
This looks good. Something to consider is doing the calculation in two loops, first calculating the average then calculating the sum of (v - vbar)**2. When vbar >> vrms, the one loop method can fail (though it only matters when the vbar/vrms is > 1.e6).

JustinRayAngus:
Agreed. I can make the change here if requested, but maybe it should be a different PR?

dpgrote:
> Agreed. I can make the change here if requested, but maybe it should be a different PR?

This would be changing the same code so I think it would be fine to have that change in the same PR.

JustinRayAngus:
@dpgrote I'm now thinking it might be best to leave it as is. The other way will involve more flops, but I'm not sure the gain in accuracy in the cold beam limit is worth the extra flops. This temperature is only used for the Coulomb logarithm. The net gain in precision will be minimal I think since the natural log of temperature is all that is used. 

Also, if you are strongly in the cold beam limit, the Coulomb log term is physically inaccurate anyway. See discussion in NRL formulary.

RemiLehe (approved):

PR #5042, opened 2024-07-10
Reset collision checksums and add particle data

RemiLehe:
Recent PRs consistently show a new value of the checksum for the test `collisionXYZ`, whereby the new temperatures are:
```
    "T_electron": 353384.4616461907,
    "T_ion": 348277.2774202612
```
While this unexpected recent change is being investigated, it might be good to update this value, so as to allow other PRs to pass.

Since the recent change in the checksum value is unexpected, it is quite possible to that this checksum value will change again in the future. In order to get a sense of whether this change is coming from the temperature diagnostic itself, or from a change in the underlying particle data, this PR now also adds checksum values for the particle data.


EZoni (approved):

dpgrote (approved):
Thanks! I had the same thought, adding the checks of the particles.

PR #5045, opened 2024-07-11
shuffling the full list of particles for intra-species binary collisions

JustinRayAngus:
For binary-pairing collision methods to work properly, each particle in a cell must have an equal opportunity of being paired with every other particle in the cell so that each particle can sample the full distribution over many scattering cycles. The shuffling routine was only applied to the first half of the list of particles in a cell for intra-species binary collisions. This means that each particle is only permitted to sample half of the distribution. This may work fine for isotropic distributions, but it does not produce the correct physics for general anisotropic cases.

The test for this issue is test T1a from JCP 413 2023 (https://doi.org/10.1016/j.jcp.2020.109450). The current version of WarpX fails this test. This PR fixes this issue.

T1a_picnic_warpx_shuffle_fix

This test uses equally-weighted particles. The implementation in WarpX still fails this test for weighted particles. That issue will be fixed in a separate PR.


RemiLehe:
Thanks a lot for this PR!

Could you confirm that this also addresses/fixes this issue:
https://github.com/ECP-WarpX/WarpX/discussions/4548

RemiLehe:
Also: could you add the above test as an automated CI test as part of this PR.
(The test should fail with the current `development` branch but it should pass with the changes of this PR)

JustinRayAngus:
> Thanks a lot for this PR!
> 
> Could you confirm that this also addresses/fixes this issue: #4548

Yes. This will fix that issue.

RemiLehe (approved):

PR #5047, opened 2024-07-13
Binary-pairing Coulomb collisions: improvements and optimization

JustinRayAngus:
**Description updated by @RemiLehe, using text by @JustinRayAngus:** 

This PR now just makes some improvements on the relativistic implementation of Coulomb scattering method by writing the formula for the normalized mean free path (s12) in a way that is directly analogous to the well-known non-relativistic formulation. This save flops by removing redundant calculations and improves on the readability of the Coulomb collision algorithm.

The expression for the normalized scattering path in UpdateMomentumPerez.H is written as:
`s12 = sigma_eff * n12 * dt * vrelst * g1sg2s/(g1g2)`
with `sigma_eff = pi*b0^2 * lnLmd`

The proof that this expression is equivalent to Eq. 9 of Perez et al 2012 (https://doi.org/10.1063/1.4742167) is given here 
[CoulombScattering_s12.pdf](https://github.com/user-attachments/files/16663359/CoulombScattering_s12.pdf)

**Previous description:**

This PR originally addressed several issues with the binary-pairing Coulomb collision method in WarpX.
1) Optimization. See PR #5066

2) Weighted-particle scattering physics. See PR #5091. The method for doing weighted-particle scattering for Coulomb collisions prior to PR #5091 is from Perez et al 2012 (https://doi.org/10.1063/1.4742167) which follows from Nanbu and Yonemura 1998 (http://dx.doi.org/10.1006/jcph.1998.6049). That method has been shown to result in incorrect relaxation rates for certain problems by Higginson et al 2020 (https://doi.org/10.1016/j.jcp.2020.109450). PR #5091 fixes this issue using a method similar to that by Higginson 2020 that produces correct relaxation rates with weighted particles. Below are results from Tests T1a-T1d from Higginson 2020 (but with 10X less particles). The physical setup is the same for each of tests T1a-T1d, but they use different particle weighting setups for the different populations. In contrast to results obtained using the development branch, the results obtained using this PR show identical relaxation results for the various particle weight combinations.

Using this PR:
T1abcd_binaryOpt_branch

Using the development branch (prior to PR #5091):
T1abcd_develop_branch_new

### DIFF
--- a/Examples/Tests/collision/analysis_collision_1d.py
+++ b/Examples/Tests/collision/analysis_collision_1d.py
@@ -39,6 +39,8 @@ mass = 1.992100316897910e-26
 # by sorting based on weight
 sorted_indices = data['ions','particle_weight'].argsort()
 sorted_wp = data['ions', 'particle_weight'][sorted_indices].value
+sorted_px = data['ions', 'particle_momentum_x'][sorted_indices].value
+sorted_py = data['ions', 'particle_momentum_y'][sorted_indices].value
 sorted_pz = data['ions', 'particle_momentum_z'][sorted_indices].value
 
 # Find the index 'Npmin' that separates macroparticles from group A and group B
--- a/Examples/Tests/collision/analysis_collision_1d.py
+++ b/Examples/Tests/collision/analysis_collision_1d.py
@@ -37,11 +37,9 @@ mass = 1.992100316897910e-26
 
 # Separate macroparticles from group A (low weight) and group B (high weight)
 # by sorting based on weight
-sorted_indices = data["ions", "particle_weight"].argsort()
-sorted_wp = data["ions", "particle_weight"][sorted_indices].value
-sorted_px = data["ions", "particle_momentum_x"][sorted_indices].value
-sorted_py = data["ions", "particle_momentum_y"][sorted_indices].value
-sorted_pz = data["ions", "particle_momentum_z"][sorted_indices].value
+sorted_indices = data['ions','particle_weight'].argsort()
+sorted_wp = data['ions', 'particle_weight'][sorted_indices].value
+sorted_pz = data['ions', 'particle_momentum_z'][sorted_indices].value
 
 # Find the index 'Npmin' that separates macroparticles from group A and group B
 Np = len(sorted_wp)
--- a/Examples/Tests/collision/analysis_collision_1d.py
+++ b/Examples/Tests/collision/analysis_collision_1d.py
@@ -37,11 +37,11 @@ mass = 1.992100316897910e-26
 
 # Separate macroparticles from group A (low weight) and group B (high weight)
 # by sorting based on weight
-sorted_indices = data['ions','particle_weight'].argsort()
-sorted_wp = data['ions', 'particle_weight'][sorted_indices].value
-sorted_px = data['ions', 'particle_momentum_x'][sorted_indices].value
-sorted_py = data['ions', 'particle_momentum_y'][sorted_indices].value
-sorted_pz = data['ions', 'particle_momentum_z'][sorted_indices].value
+sorted_indices = data["ions", "particle_weight"].argsort()
+sorted_wp = data["ions", "particle_weight"][sorted_indices].value
+sorted_px = data["ions", "particle_momentum_x"][sorted_indices].value
+sorted_py = data["ions", "particle_momentum_y"][sorted_indices].value
+sorted_pz = data["ions", "particle_momentum_z"][sorted_indices].value
 
 # Find the index 'Npmin' that separates macroparticles from group A and group B
 Np = len(sorted_wp)
--- a/Examples/Tests/collision/inputs_1d
+++ b/Examples/Tests/collision/inputs_1d
@@ -20,7 +20,10 @@ my_constants.m_c12 = 12.*m_u - 6.*m_e
 #################################
 max_step = 600
 amr.n_cell = 180
+<<<<<<< HEAD
 amr.max_level = 0
+=======
+>>>>>>> 1c9ba63f0 (updating CI input/analysis)
 amr.blocking_factor = 4
 geometry.dims = 1
 geometry.prob_lo = 0.
--- a/Examples/Tests/collision/inputs_1d
+++ b/Examples/Tests/collision/inputs_1d
@@ -20,10 +20,7 @@ my_constants.m_c12 = 12.*m_u - 6.*m_e
 #################################
 max_step = 600
 amr.n_cell = 180
-<<<<<<< HEAD
 amr.max_level = 0
-=======
->>>>>>> 1c9ba63f0 (updating CI input/analysis)
 amr.blocking_factor = 4
 geometry.dims = 1
 geometry.prob_lo = 0.
--- a/Regression/Checksum/benchmarks_json/collisionXYZ.json
+++ b/Regression/Checksum/benchmarks_json/collisionXYZ.json
@@ -6,7 +6,6 @@
     "Ex": 0.0,
     "Ey": 0.0,
     "Ez": 0.0,
-<<<<<<< HEAD
     "T_electron": 351188.7347045234,
     "T_ion": 350097.5453711827
   },
@@ -26,27 +25,6 @@
     "particle_position_x": 21233610.538310427,
     "particle_position_y": 21280892.516510233,
     "particle_position_z": 21213150.945697505,
-=======
-    "T_electron": 351982.01692182414,
-    "T_ion": 349599.6939052666
-  },
-  "electron": {
-    "particle_momentum_x": 8.359982321196844e-19,
-    "particle_momentum_y": 8.192841151167734e-19,
-    "particle_momentum_z": 8.182985690701254e-19,
-    "particle_position_x": 21255110.08090506,
-    "particle_position_y": 21303488.624262597,
-    "particle_position_z": 21238676.122703437,
-    "particle_weight": 7.168263344048695e+28
-  },
-  "ion": {
-    "particle_momentum_x": 2.00348302409669e-18,
-    "particle_momentum_y": 1.8323959076577174e-18,
-    "particle_momentum_z": 1.8279532308286334e-18,
-    "particle_position_x": 21246214.748882484,
-    "particle_position_y": 21280709.710960116,
-    "particle_position_z": 21206153.002106402,
->>>>>>> b6390f0ea (updating checksums)
     "particle_weight": 7.168263344048695e+28
   }
 }
--- a/Regression/Checksum/benchmarks_json/collisionXYZ.json
+++ b/Regression/Checksum/benchmarks_json/collisionXYZ.json
@@ -6,6 +6,7 @@
     "Ex": 0.0,
     "Ey": 0.0,
     "Ez": 0.0,
+<<<<<<< HEAD
     "T_electron": 351188.7347045234,
     "T_ion": 350097.5453711827
   },
@@ -25,6 +26,27 @@
     "particle_position_x": 21233610.538310427,
     "particle_position_y": 21280892.516510233,
     "particle_position_z": 21213150.945697505,
+=======
+    "T_electron": 351982.01692182414,
+    "T_ion": 349599.6939052666
+  },
+  "electron": {
+    "particle_momentum_x": 8.359982321196844e-19,
+    "particle_momentum_y": 8.192841151167734e-19,
+    "particle_momentum_z": 8.182985690701254e-19,
+    "particle_position_x": 21255110.08090506,
+    "particle_position_y": 21303488.624262597,
+    "particle_position_z": 21238676.122703437,
+    "particle_weight": 7.168263344048695e+28
+  },
+  "ion": {
+    "particle_momentum_x": 2.00348302409669e-18,
+    "particle_momentum_y": 1.8323959076577174e-18,
+    "particle_momentum_z": 1.8279532308286334e-18,
+    "particle_position_x": 21246214.748882484,
+    "particle_position_y": 21280709.710960116,
+    "particle_position_z": 21206153.002106402,
+>>>>>>> b6390f0ea (updating checksums)
     "particle_weight": 7.168263344048695e+28
   }
 }
--- a/Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
+++ b/Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
@@ -119,7 +119,7 @@ void ElasticCollisionPerez (
           // scattering path s12 in UpdateMomentumPerezElastic().
           // s12 is defined such that the expected value of the change in particle
           // velocity is equal to that from the full NxN pairing method, as described
-          // here https://arxiv.org/submit/5758216/view. This method is a direct extension
+          // here https://arxiv.org/abs/2407.19151. This method is a direct extension
           // of the original method by Takizuka and Abe JCP 25 (1977) to weighted particles.
           T_PR n12;
           const T_PR wpmax = amrex::max(w1[ I1[i1] ],w2[ I2[i2] ]);
--- a/Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
+++ b/Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
@@ -32,7 +32,9 @@
  * @param[in] dt is the time step length between two collision calls.
  * @param[in] L is the Coulomb log and will be used if greater than zero,
  *            otherwise will be computed.
+ * @param[in] dV is the volume of the corresponding cell.
  * @param[in] engine the random number generator state & factory
+ * @param[in] isSameSpecies whether this is an intra-species collision process
  * @param[in] coll_idx is the collision index offset.
 */
 
--- a/Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
+++ b/Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
@@ -69,7 +69,7 @@ void ElasticCollisionPerez (
 
     // compute Debye length lmdD (if not using a fixed L = Coulomb log)
     T_PR lmdD = T_PR(-1.0);
-    if ( L <= T_PR(0.0) ) {
+    if ( L < T_PR(0.0) ) {
         lmdD = T_PR(1.0)/std::sqrt( n1*q1*q1/(T1*PhysConst::ep0) +
                                     n2*q2*q2/(T2*PhysConst::ep0) );
     }
@@ -80,6 +80,9 @@ void ElasticCollisionPerez (
 
     // bmax (screening length) cannot be smaller than atomic spacing
     const T_PR bmax = amrex::max(lmdD, rmin);
+
+    // set max cross section based on mfp = atomic spacing
+    const T_PR sigma_max = T_PR(1.0) / (maxn * rmin);
 
 #if (defined WARPX_DIM_RZ)
     T_PR * const AMREX_RESTRICT theta1 = soa_1.m_rdata[PIdx::theta];
@@ -124,10 +127,8 @@ void ElasticCollisionPerez (
           UpdateMomentumPerezElastic(
               u1x[ I1[i1] ], u1y[ I1[i1] ], u1z[ I1[i1] ],
               u2x[ I2[i2] ], u2y[ I2[i2] ], u2z[ I2[i2] ],
-              n1, n2, n12,
               q1, m1, w1[ I1[i1] ], q2, m2, w2[ I2[i2] ],
-              dt, L, bmax,
-              engine);
+              n12, sigma_max, L, bmax, dt, engine );
 
 #if (defined WARPX_DIM_RZ)
           T_PR const u1xbuf_new = u1x[I1[i1]];
--- a/Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
+++ b/Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
@@ -71,7 +71,7 @@ void ElasticCollisionPerez (
 
     // compute Debye length lmdD (if not using a fixed L = Coulomb log)
     T_PR lmdD = T_PR(-1.0);
-    if ( L < T_PR(0.0) ) {
+    if ( L <= T_PR(0.0) ) {
         lmdD = T_PR(1.0)/std::sqrt( n1*q1*q1/(T1*PhysConst::ep0) +
                                     n2*q2*q2/(T2*PhysConst::ep0) );
     }
--- a/Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
+++ b/Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
@@ -23,7 +23,7 @@
  *        otherwise L will be calculated based on the algorithm.
  *        @param[in] n12 = max(w1,w2)*min(N1,N2)/dV is the effective density used for s12
  *        @param[in] sigma_max is the maximum cross section based on mfp = atomic spacing
- *        used for the normalized scattering length s12
+ *        used for the normalized scattering length s12 (see Sec. II.C of Perez et al.)
  *        To see if there are nan or inf updated velocities,
  *        compile with USE_ASSERTION=TRUE.
  *
--- a/Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
+++ b/Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
@@ -151,6 +151,9 @@ void UpdateMomentumPerezElastic (
         }
 
         // Compute s12 with sigma limited by sigma_max where mfp = atomic spacing
+        // See https://github.com/user-attachments/files/16555064/CoulombScattering_s12.pdf
+        // for a proof that this expression for s12 is the same as Eq. 9 of Perez 2012 when
+        // sigma_eff = pi*b0^2*lnLmd
         const T_PR sigma_eff = amrex::min(T_PR(MathConst::pi)*b0*b0*lnLmd,sigma_max);
         s12 = sigma_eff * n12 * dt * vrelst * g1s*g2s/(g1*g2);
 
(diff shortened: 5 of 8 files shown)


RemiLehe:
Thanks a lot for this PR! For clarity and because this PR will probably serve as reference in the future by other WarpX developers, would it make sense to split this into 2 independent PR?
- One PR that fixes the performance issue, by computing the density and temperature ahead of time.
- One PR that fixes the weighting issue, by introducing this code https://github.com/ECP-WarpX/WarpX/pull/5047/files#diff-9a3d3b7442d673878318b7ebdf440a46b42570bb3afae0766af6c2301e910dcfR118 and the automated test

JustinRayAngus:
I understand and agree that your suggestion is ideal. It was my original intent to do them separately. 

The issue is the n12 parameter computed in lines 99-108 of ElasticCollisionPerez.H. That value is an order N calc done for each binary pair (net cost is order N squared), but it's also not used at all in the improved method that produces the correct scattering. It seemed like a lot of unnecessary work to pull this parameter out to being computed at the cell level and then passing it down the chain of calls when it is not needed in the first place and all these added lines of code would just be removed in the follow on PR.

Unless you see a way to avoid the issue I described above, I think we should not split this PR into two. There will be a few more minor PRs related to Coulomb scattering following this one. 

What do you think? Perhaps I can update the PR title to better reflect both parts of this PR?

RemiLehe:
OK, thanks a lot for clarifying.

How about switching the order of the PRs, i.e. doing a separate PR that implements only:

- One PR that fixes the weighting issue, by introducing this code https://github.com/ECP-WarpX/WarpX/pull/5047/files#diff-9a3d3b7442d673878318b7ebdf440a46b42570bb3afae0766af6c2301e910dcfR118 and the automated test

(that way, the calculation of `n12` is not needed in the code anymore)

and then once the above is merged, we could merge the current PR (i.e. #5047), which, at this point, will not show any changes for https://github.com/ECP-WarpX/WarpX/pull/5047/files#diff-9a3d3b7442d673878318b7ebdf440a46b42570bb3afae0766af6c2301e910dcfR118 since that part will already have been merged as part of the first PR.

Does that make sense, and does it sound realistic? If not, I agree we can also simply merge this PR (i.e. #5047) as a single combined PR.

JustinRayAngus:
Splitting that way would be less work, but to be safe I should probably re-run all the tests (T1a-T1d), which will take a while because of the N^2 cost. If the aim of splitting the PRs in two is to be able to go back to the Nanbu method at some point for making comparisons or something, then I think it would make more sense to actually do it the first way you propose where the first PR is just about the optimization. That way also doesn't require re-running any of the new tests, since the code will fail for them anyhow.

For the sake of time and moving forward, I would advocate for simply merging this PR as is. This is my biased opinion. I'll let you have the final say of what to do.

ax3l:
I like that split, that way we have a clear bug fix PR and a clear performance optimization PR. This will make also future potential regressions easier to spot and address (or reverts).

Our idea is not to go back to Nanbu in the future. It is more about having PRs that are easier to understand, to review and to go back to in the future.

Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
```diff
@@ -108,13 +115,21 @@ void ElasticCollisionPerez (
...
 
+          // compute the effective density used for s12 in UpdateMomentumPerezElastic()
+          // The effective density used for s12 is defined such that the average value of s12
+          // after many scattering cycles is the same as that for the full NxN pairing method.
+          // The approach here is similar to that in JCP 413 (2020) by Higginson et al., but
+          // does not use a unique duplication factor for each macro-particle pair.
+          T_PR n12;
+          const T_PR wpmax = amrex::max(w1[ I1[i1] ],w2[ I2[i2] ]);
+          if (isSameSpecies) { n12 = wpmax*static_cast<T_PR>(min_N+max_N-1)/dV; }
+          else { n12 = wpmax*static_cast<T_PR>(min_N)/dV; }
```
RemiLehe: The factor `min_N` is not explained in Higginson et al. (since Higginson's paper is duplicating particles instead, which results in a factor `max_N` instead). 

My understanding is that, for a given species (say species 1), this factor should correspond to the number of macroparticles in the other species (`N2` for species 1), in order to account for the fact that a macroparticle from species 1 will only collide with one or (a few) macroparticles of species 2, instead of all of the `N2` macroparticles. 
If species 1 is the species with the larger number of macroparticles, this automatically works (`min_N = N2`).
If species 2 is the species with the lowest number of macroparticles, then its particles will participate, on average, in `max_N/min_N` pairs, so it is as if they see a factor `max_N/min_N x min_N = max_N = N2`, so it works in this case.

Is the above correct? If yes, could you add an additional comment explaining the factor `min_N` in the code (if possible, in a more concise and clearer way than the explanation I gave above).

RemiLehe: Note also that the other binary collisions in WarpX use a duplication method similar to Higginsons' paper:

https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.H#L147
https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.H#L173

https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/Collision/BinaryCollision/NuclearFusion/NuclearFusionFunc.H#L186
https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/Collision/BinaryCollision/NuclearFusion/NuclearFusionFunc.H#L173

Thus, it might be more consistent to use Higginson's duplication method in the Coulomb module too.

Are there advantages to the proposed method without duplication, aside from the slightly simpler implementation?

Examples/Tests/collision/inputs_1d
```diff
@@ -0,0 +1,93 @@
+#################################
+########## CONSTANTS ############
+#################################
+
+my_constants.epsilon = 1.0e-12
```
RemiLehe: Why is this `epsilon` needed? Is this in the test described by Higginson?

Examples/Tests/collision/inputs_1d
```diff
@@ -0,0 +1,93 @@
...
+my_constants.q_c12 = 6.*q_e
+my_constants.m_c12 = 12.*m_u - 6.*m_e
+
+#################################
+####### GENERAL PARAMETERS ######
+#################################
+max_step = 600
+amr.n_cell = 180
+amr.max_grid_size = 4
+amr.blocking_factor = 4
```
RemiLehe:
```suggestion
```
Since the CI tests run only on 2 cores, it might be more efficient to use the default `max_grid_size`.

Examples/Tests/collision/analysis_collision_1d.py
```diff
@@ -0,0 +1,154 @@
...
+sorted_px_sum = np.abs(sorted_px).sum();
+sorted_py_sum = np.abs(sorted_py).sum();
+sorted_pz_sum = np.abs(sorted_pz).sum();
+sorted_wp_sum = np.abs(sorted_wp).sum();
+
+#print(f"sum(sorted_px)  =  {sorted_px_sum}")
+#print(f"sum(sorted_py)  =  {sorted_py_sum}")
+#print(f"sum(sorted_pz)  =  {sorted_pz_sum}")
+#print(f"sum(sorted_wp)  =  {sorted_wp_sum}")
+
```
RemiLehe:
```suggestion
```

Examples/Tests/collision/analysis_collision_1d.py
```diff
@@ -0,0 +1,154 @@
...
+    wpA = wpmax;
+    NpBs = 0
+    NpBe = Npmin
+    NpAs = Npmin
+    NpAe = Np
+
+#############
+
+#print(f"Np = {Np}, NpA = {NpA}, NpB = {NpB}")
+#print(f"wpA = {wpA}, wpB = {wpB}")
```
RemiLehe:
```suggestion
```

Examples/Tests/collision/analysis_collision_1d.py
```diff
@@ -0,0 +1,154 @@
...
+for i in range(len(sorted_wp)):
+    if sorted_wp[i] > wpmin:
+        Npmin = i
+        break
+
+test_tag = 'T1b'
+print(f"test_tag = {test_tag}")
+
+### use below for tests T1b and T1c, where NpA = Npmin
+if test_tag == 'T1b' or test_tag == 'T1c':
```
RemiLehe: I understand that this makes the script generic, but for a CI test, I think it is good to simplify it and keep only the code that will actually be run in the CI test, i.e. removing `test_tag` and the corresponding `if` conditions.

Examples/Tests/collision/analysis_collision_1d.py
```diff
@@ -0,0 +1,154 @@
...
+
+# this will be the name of the plot file
+fn = sys.argv[1]
+ds = yt.load(fn)
+data = ds.covering_grid(level = 0, left_edge = ds.domain_left_edge, dims = ds.domain_dimensions)
+
+# carbon 12 ion (mass = 12*amu - 6*me)
+mass = 1.992100316897910e-26
+
+# compute sum of ion values
```
RemiLehe:
```suggestion
# Separate macroparticles from group A (low weight) and group B (high weight)
# by sorting based on the weight
```

Examples/Tests/collision/analysis_collision_1d.py
```diff
@@ -0,0 +1,154 @@
...
+sorted_wp = data['ions', 'particle_weight'].value
+sorted_wp.sort();
+sorted_indices = data['ions','particle_weight'].argsort()
+sorted_px = data['ions', 'particle_momentum_x'][sorted_indices].value
+sorted_py = data['ions', 'particle_momentum_y'][sorted_indices].value
+sorted_pz = data['ions', 'particle_momentum_z'][sorted_indices].value
+
```
RemiLehe:
```suggestion
# Find the index `Npmin` that separates macroparticles from group A and group B
# in the sorted arrays.
```

Examples/Tests/collision/analysis_collision_1d.py
```diff
@@ -0,0 +1,154 @@
...
+ds = yt.load(fn)
+data = ds.covering_grid(level = 0, left_edge = ds.domain_left_edge, dims = ds.domain_dimensions)
+
+# carbon 12 ion (mass = 12*amu - 6*me)
+mass = 1.992100316897910e-26
+
+# compute sum of ion values
+sorted_wp = data['ions', 'particle_weight'].value
+sorted_wp.sort();
+sorted_indices = data['ions','particle_weight'].argsort()
```
RemiLehe: Slight reformulation:
```suggestion
sorted_indices = data['ions','particle_weight'].argsort()
sorted_wp = data['ions','particle_weight'][sorted_indices].value
```

Examples/Tests/collision/analysis_collision_1d.py
```diff
@@ -0,0 +1,154 @@
...
+### use below for tests T1a and T1d, where NpB = Npmin
+if test_tag == 'T1a' or test_tag == 'T1d':
+    NpB = Npmin
+    wpB = wpmin;
+    NpA = Np - Npmin
+    wpA = wpmax;
+    NpBs = 0
+    NpBe = Npmin
+    NpAs = Npmin
+    NpAe = Np
```
RemiLehe:
```suggestion
```

Examples/Tests/collision/analysis_collision_1d.py
```diff
@@ -0,0 +1,154 @@
...
+TA = (TAx + TAy + TAz)/3.0
+
+TBpar = TBz
+TBperp = (TBx + TBy)/2.0
+TB = (TBx + TBy + TBz)/3.0
+
+#print(f"uAx = {uAx/1e3:.{5}f}; uAy = {uAy/1e3:.{5}f}; uAz = {uAz/1e3:.{5}f} [km/s]")
+#print(f"uBx = {uBx/1e3:.{5}f}; uBy = {uBy/1e3:.{5}f}; uBz = {uBz/1e3:.{5}f} [km/s]")
+#print(f"TA = {TA:.{5}f}; TAperp = {TAperp:.{5}f}; TApar = {TApar:.{5}f} [eV]")
+#print(f"TB = {TB:.{5}f}; TBperp = {TBperp:.{5}f}; TBpar = {TBpar:.{5}f} [eV]")
```
RemiLehe:
```suggestion
```

Examples/Tests/collision/analysis_collision_1d.py
```diff
@@ -0,0 +1,154 @@
...
+
+TApar_30ps_soln = 6.15e3 # TA parallel solution at t = 30 ps
+error = np.abs(TApar-TApar_30ps_soln)/TApar_30ps_soln
+tolerance = 0.02
+print('TApar at 30ps error = ', error);
+print('tolerance = ', tolerance);
```
RemiLehe: Could you confirm that this fails when using the WarpX version from `development` branch (i.e. without this PR) ?
Also: would it make sense to also add an `assert` checking the other quantities (`TAx`, `TAy`, as well as the `TB`s, `uA`s and `uB`s)?
And would it be possible to run the simulation a bit longer (e.g. 100 ps) so that we can actually use the equilibrium values of the `T`s and `u`s, which can be computed analytically?

Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
```diff
@@ -102,86 +105,87 @@ void UpdateMomentumPerezElastic (
...
-        const auto coeff = static_cast<T_PR>(
-            std::pow(4.0*MathConst::pi/3.0,1.0/3.0));
-        T_PR const vrel = mass_g*p1sm/(m1*g1s*m2*g2s*gc);
-        T_PR const sp = coeff * n1*n2/n12 * dt * vrel * (m1+m2) /
-            amrex::max( m1*cbrt_n1*cbrt_n1,
-                        m2*cbrt_n2*cbrt_n2);
-
-        // Determine s
-        s = amrex::min(s,sp);
+        // Compute s12 with sigma limited by sigma_max where mfp = atomic spacing
```
RemiLehe, line 155: Could add a reference in the comment? (Is it the same reference as above?)

Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
```diff
@@ -102,86 +105,87 @@ void UpdateMomentumPerezElastic (
...
-            // Compute the minimal impact parameter
-            constexpr T_PR hbar_pi = static_cast<T_PR>(PhysConst::hbar*MathConst::pi);
-            const T_PR bmin = amrex::max(hbar_pi/p1sm, b0);
+
+            // Compute the minimum impact parameter from quantum: bqm = lDB/(4*pi) = hbar/(2*p1sm)
+            // See NRL formulary. Also see "An introduction to the physics of the Coulomb logarithm,
+            // with emphasis on quantum-mechanical effects", J. Plasma Phys. vol. 85 (2019). by J.A. Krommes.
+            // Note: The formula in Perez 2012 and in Lee and More 1984 uses h rather than
+            // hbar for bqm. If this is used, then the transition energy where bmin goes from classical
+            // to quantum is only 2.5 eV for electrons; compared to 100 eV when using hbar.
```
RemiLehe, line 144: Given that we cite the above paper, I think that this is not needed.
```suggestion
```

Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
```diff
@@ -102,86 +105,87 @@ void UpdateMomentumPerezElastic (
     T_PR const g1s = ( T_PR(1.0) - vcDv1*inv_c2 )*gc*g1;
     T_PR const g2s = ( T_PR(1.0) - vcDv2*inv_c2 )*gc*g2;
 
-    // Compute s
-    T_PR s = 0;
+    // Compute variant relative velocity in cm frame
```
RemiLehe:
```suggestion
    // Compute variant relative velocity in center-of-mass frame
```
Also: What is meant by `variant` here?

Examples/Tests/collision/inputs_1d
JustinRayAngus: That is there so that one can distinguish the two population of particles in the output file for test T1a when using equally-weighted particles. I didn't know how else to do it other than having one population with a slightly different weight than the other.

Examples/Tests/collision/analysis_collision_1d.py
JustinRayAngus: I chose this time because this is where one sees the largest discrepency between the new corrected implementation and the original implementation. If you ran until equilibrium, then the development branch will not fail the check. It is the relaxation to the equilibrium that is incorrect. Not the steady state solution.

JustinRayAngus: And yes. The development branch fails this test.

Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
JustinRayAngus: It is to distinguish it from the invariant relative velocity defintion that is frame independent.

JustinRayAngus: Will make comment change.

Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
JustinRayAngus: I felt that since the formulation here largely follows Perez 2012, that it may be useful to make explicit comments where things differ. But if you feel it is not needed then I will take it out.

Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
JustinRayAngus: Using sigma_max is the same as using sp from Perez 2012. Did you get a chance to look at the tech-note I sent you and Axel where I went over the different limiting models used by Perez?

Examples/Tests/collision/analysis_collision_1d.py
JustinRayAngus: Ok

Examples/Tests/collision/analysis_collision_1d.py
JustinRayAngus: Ok

Examples/Tests/collision/analysis_collision_1d.py
JustinRayAngus: Ok

Examples/Tests/collision/analysis_collision_1d.py
JustinRayAngus: Ok

Examples/Tests/collision/analysis_collision_1d.py
JustinRayAngus: Ok

Examples/Tests/collision/analysis_collision_1d.py
JustinRayAngus: Ok

Examples/Tests/collision/analysis_collision_1d.py
JustinRayAngus: Ok

Examples/Tests/collision/analysis_collision_1d.py
JustinRayAngus: Ok

Examples/Tests/collision/inputs_1d
JustinRayAngus: Ok

Examples/Tests/collision/inputs_1d
JustinRayAngus: Since the CI test does not do test T1a. I can remove the epsilon.

Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
JustinRayAngus: It is not just Nmax vs Nmin that is the difference. In the Higginson method, one uses (wmax_prime)*Nmax, where wmax_prime = max(wi/df,wj/df) is the max modified weight of the two particles with dp the duplication factor. Note that the average duplication factor per pair is Nmax/Nmin, which means on average (wmax_prime) *Nmax = wmax *Nmin.

Using Nmin is certainly simpler because you don't have to used a modified weight that takes into account a pair-specific duplication factor. It is also the most direct extension of the original TA77 method, which uses Nmin.

JustinRayAngus: > The factor `min_N` is not explained in Higginson et al. (since Higginson's paper is duplicating particles instead, which results in a factor `max_N` instead).
> 
> My understanding is that, for a given species (say species 1), this factor should correspond to the number of macroparticles in the other species (`N2` for species 1), in order to account for the fact that a macroparticle from species 1 will only collide with one or (a few) macroparticles of species 2, instead of all of the `N2` macroparticles. If species 1 is the species with the larger number of macroparticles, this automatically works (`min_N = N2`). If species 2 is the species with the lowest number of macroparticles, then its particles will participate, on average, in `max_N/min_N` pairs, so it is as if they see a factor `max_N/min_N x min_N = max_N = N2`, so it works in this case.
> 
> Is the above correct? If yes, could you add an additional comment explaining the factor `min_N` in the code (if possible, in a more concise and clearer way than the explanation I gave above).

Your interpretation is correct. It is also exactly the same interpretation of the original method for equally-weighted particles by Takizuka and Abe 1977.

Examples/Tests/collision/inputs_1d
JustinRayAngus: Removing amr.blocking_factor leads to crash on run time.

Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
JustinRayAngus: You asked about advantages of the proposed method.

As you stated, one advantage is that it is simpler. Rather than adding un-necessary lines of code here to maintain quasi-consistency with DSMC and Fusion methods, I would suggest considering adopting the simplified approach here in those modules. In addition to being simpler, there is in fact another advantage when using it for reactions that result in product species such as Fusion. I’ll try to concisely explain an issue with the current approach used in the fusion module below.

For simplicity, consider inter-species fusion between two species with equal weight particles w and using fusion_multiplier = 1. Using N_max, the reaction probability will be too large by N_max/N_min on average. However, because the reaction weight used is w_min = N_min/N_max*w (on average), the total number of fusion products produced is correct on average. So, the current method essentially uses a fusion_multiplier of N_max/N_min even when the specified fusion_multiplier is 1. Thus, it is not possible to do a simulation with equally weighted particles as the particles will no longer be equally weighted after a single fusion event when N1 and N2 are not equal. This approach can also result in the avalanching production of small weight particles described in PR #4976 

The alternative approach has an advantage in that it allows one to maintain equally-weighted particles and it will not have the issue described in PR #4976.

I'm happy to discuss further off line if needed.

JustinRayAngus:
Looks like the checksum values for the new CI test are slightly different. Would the removal of max_grid_size line cause this?

Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H
roelof-groenewald: I agree we should switch the DSMC and fusion modules to also use this new method. Doing that for DSMC is on my to-do list.

JustinRayAngus:
[CoulombScattering_s12.pdf](https://github.com/user-attachments/files/16555064/CoulombScattering_s12.pdf)

Regression/Checksum/benchmarks_json/collisionXYZ.json
```diff
@@ -6,25 +6,25 @@
     "Ex": 0.0,
     "Ey": 0.0,
     "Ez": 0.0,
-    "T_electron": 351982.0169218243,
-    "T_ion": 349599.6939052666
+    "T_electron": 381957.7394223898,
+    "T_ion": 319565.6269784763
   },
```
JustinRayAngus: The large difference in these temperatures is a bit concerning. I may need to inspect this test further.

JustinRayAngus: Looks like this relatively large change happened on PR #5110.

Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
ax3l: I am ok to leave it, explicit is good where things differ.

Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
ax3l: This is great. If you like, the PDF of the simplification proof linked and the tech note could also be part of our [theory sections](https://warpx.readthedocs.io/en/latest/theory/intro.html). (Relatively quick to translate from LaTeX to `.rst`.)

ax3l (approved):
Thank you, this looks great to me.

Feel free to move some of the PDFs to the theory section (as .rst with `.. math::` blocks). That can be a follow-up, of course.

Source/Particles/Collision/BinaryCollision/Coulomb/UpdateMomentumPerezElastic.H
```diff
@@ -22,6 +22,8 @@
  *        @param[in] L is the Coulomb log. A fixed L will be used if L > 0,
  *        otherwise L will be calculated based on the algorithm.
  *        @param[in] n12 = max(w1,w2)*min(N1,N2)/dV is the effective density used for s12
+ *        @param[in] sigma_max is the maximum cross section based on mfp = atomic spacing
+ *        used for the normalized scattering length s12
```
RemiLehe: Could you add a reference (e.g. pointing to a specific section of the Perez paper), so that future developers that read this understand where this comes from?

RemiLehe (approved):

PR #5049, opened 2024-07-15
Initialize with value all `aux`, `cax` fields

EZoni:
The fact that these `aux` and `cax` fields were not initialized (to zero, like other similar fields), seems to be the cause of errors that show up in #4922 for two CI tests, `scraping` and `scraping_filter`.


dpgrote (approved):
Thanks! I would say that this is a good thing to do anyway.

PR #5053, opened 2024-07-16
Remove calls to notInLaunchRegion

RemiLehe:
Since we can never compile WarpX with both OpenMP and GPU support, checking `notInLaunchRegion` is not useful, and might be confusing for new developers.

### DIFF
--- a/Source/Particles/MultiParticleContainer.H
+++ b/Source/Particles/MultiParticleContainer.H
@@ -390,7 +390,7 @@ protected:
 
         MFItInfoCheckTiling(pc_src, pc_dsts...);
 
-        if (WarpXParticleContainer::do_tiling && amrex::Gpu::notInLaunchRegion()) {
+        if (WarpXParticleContainer::do_tiling) {
             info.EnableTiling(WarpXParticleContainer::tile_size);
         }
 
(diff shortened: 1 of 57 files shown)


Source/Particles/MultiParticleContainer.H
```diff
@@ -390,7 +390,7 @@ protected:
 
         MFItInfoCheckTiling(pc_src, pc_dsts...);
 
-        if (WarpXParticleContainer::do_tiling && amrex::Gpu::notInLaunchRegion()) {
+        if (WarpXParticleContainer::do_tiling) {
```
AlexanderSinn, line 393: When using GPUs, is `WarpXParticleContainer::do_tiling` always off?

RemiLehe: Hm, actually no: https://github.com/ECP-WarpX/WarpX/blob/development/Source/Initialization/WarpXAMReXInit.cpp#L76 (although I don't think that someone ever overrode the default `false` in practice, for WarpX)
Would you recommend keeping `&& amrex::Gpu::notInLaunchRegion()` here?

AlexanderSinn: Yes. Either keep `&& amrex::Gpu::notInLaunchRegion()` or add `#ifndef AMREX_USE_GPU` around that part.

AlexanderSinn:
I found a `Gpu::inLaunchRegion()` in `meanParticleVelocity`. With how that's written, maybe It's fine to keep it, but I noticed that `meanParticleVelocity` does not take into account particle weight.

Issue #4896, opened 2024-04-25
Binary collision performance


roelof-groenewald:
The recent improvements to the binary collision parallelization strategy (https://github.com/ECP-WarpX/WarpX/pull/4577) has shown excellent speed-up for high particle-per-cell cases where particle density is not uniform. However, recent testing with uniform plasmas showed that the new scheme performs somewhat worse than the previous one.
From @RemiLehe:

> The new implementation is exposing more parallelism, but introduces new overheads.
In cases where there is already enough parallelism in the old implementation (large number of cells), there is no gain from exposing more parallelism, and maybe then the overheads dominate.
The next step would probably be to take a closer look at the overheads and their relative cost.

The input file below produced the following performance trends on Perlmutter GPU nodes ("loop over collision pairs" is the new scheme):
![image](https://github.com/ECP-WarpX/WarpX/assets/40245517/65bcd358-4224-4b3a-8a84-401694bd2cba)
![image](https://github.com/ECP-WarpX/WarpX/assets/40245517/d61c0794-dcb9-4312-9cd3-cb954e41e7dc)
![image](https://github.com/ECP-WarpX/WarpX/assets/40245517/931363e8-0410-46b6-912f-c6cfa32b26f4)

```
#################################
####### GENERAL PARAMETERS ######
#################################
max_step = 100
amr.n_cell =  8 8 8
amr.max_grid_size_x = 32
amr.max_grid_size_y = 32
amr.max_grid_size = 64
amr.blocking_factor = 8
amr.max_level = 0
geometry.dims = 3
geometry.prob_lo = -20.e-6   -20.e-6   -20.e-6    # physical domain
geometry.prob_hi =  20.e-6    20.e-6    20.e-6

#################################
####### Boundary condition ######
#################################
boundary.field_lo = periodic periodic periodic
boundary.field_hi = periodic periodic periodic

#################################
############ NUMERICS ###########
#################################
warpx.serialize_initial_conditions = 1
warpx.verbose = 1
warpx.cfl = 1.0

# Order of particle shape factors
algo.particle_shape = 1

#################################
############ PLASMA #############
#################################
particles.species_names = electrons ions

electrons.species_type = electron
electrons.injection_style = "nrandompercell"
electrons.num_particles_per_cell = ppc
electrons.profile = constant
electrons.density = 1.e25  # number of electrons per m^3
electrons.momentum_distribution_type = "gaussian"
[...]

PR #5066, opened 2024-07-19
Avoid N² Operation in Binary-Pairing Coulomb Collsions

JustinRayAngus:
Moved all order(N) calculations called at each binary pairing for the Coulomb collision method to being called once per cell. The net cost of the binary Coulomb scattering method now scales linear with N rather than scaling like N^2. This fixes Issue #4896 . This PR is the first part of PR #5047 , which is being split into two PRs.


ax3l (approved):
Thank you for this fix, looks great :rocket: :sparkles:

PR #5073, opened 2024-07-22
Make quantum_xi_c2 a non-static member variable of the WarpX class

lucafedeli88:
This PR is part of the effort to make the WarpX class less static.

### DIFF
--- a/Source/WarpX.H
+++ b/Source/WarpX.H
@@ -665,7 +665,7 @@ public:
      */
     void Hybrid_QED_Push (int lev, PatchType patch_type, amrex::Real dt);
 
-    static amrex::Real quantum_xi_c2;
+    amrex::Real quantum_xi_c2;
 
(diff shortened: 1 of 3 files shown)


Source/WarpX.H
```diff
@@ -665,7 +665,7 @@ public:
      */
     void Hybrid_QED_Push (int lev, PatchType patch_type, amrex::Real dt);
 
-    static amrex::Real quantum_xi_c2;
+    amrex::Real quantum_xi_c2;
```
ax3l: Name this `m_...`?
```suggestion
    amrex::Real m_quantum_xi_c2;
```

lucafedeli88: ah, yes, that's a good point... I will rename the variable everywhere.

lucafedeli88: Done

ax3l (approved):

PR #5075, opened 2024-07-22
PICMI (Bucket): NumPy 2.0 Compatibility

### LINKED ISSUE #5071: NumPy 2.0 Compatibility
With the latest NumPy 2.0 release, we need to make sure all our examples and Python scripts still work.

Seeing some tests fail in Conda-Forge https://github.com/conda-forge/warpx-feedstock/pull/86

e.g., running the `laser_acceleration/PICMI_inputs_3d.py` test:
```
+ /home/conda/feedstock_root/build_artifacts/warpx_1721445803976/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_/bin/python Examples/Physics_applications/laser_acceleration/PICMI_inputs_3d.py
Initializing AMReX (24.07)...
OMP initialized with 2 OMP threads
AMReX (24.07) initialized
  File "/home/conda/feedstock_root/build_artifacts/warpx_1721445803976/test_tmp/Examples/Physics_applications/laser_acceleration/PICMI_inputs_3d.py", line 162, in <module>
    sim.initialize_warpx()
  File "/home/conda/feedstock_root/build_artifacts/warpx_1721445803976/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_/lib/python3.10/site-packages/pywarpx/picmi.py", line 2213, in initialize_warpx
    pywarpx.warpx.init(mpi_comm, max_step=self.max_steps, stop_time=self.max_time)
  File "/home/conda/feedstock_root/build_artifacts/warpx_1721445803976/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_/lib/python3.10/site-packages/pywarpx/WarpX.py", line 98, in init
    libwarpx.initialize(argv, mpi_comm=mpi_comm)
  File "/home/conda/feedstock_root/build_artifacts/warpx_1721445803976/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_/lib/python3.10/site-packages/pywarpx/_libwarpx.py", line 131, in initialize
    self.warpx = self.libwarpx_so.get_instance()
RuntimeError: Unknown character . in Parser expression "np.float64(3e-14)"
```

ax3l:
`repr` returns `'np.float64(value)'` instead of `'value'` in NumPy 2.0.

Fix #5071

### DIFF
--- a/Python/pywarpx/Bucket.py
+++ b/Python/pywarpx/Bucket.py
@@ -59,9 +59,6 @@ class Bucket(object):
         for attr, value in self.argvattrs.items():
             if value is None:
                 continue
-            # --- str (and repr) crop some digits for floats. TODO: we should format
-            #     floating point numbers & numpy scalars to the significant digits of the
-            #     precision of amrex::Real/ParticleReal
             # --- The strip of "'" is then needed when value is a string.
             if isinstance(value, str):
                 if value.find('=') > -1:
--- a/Python/pywarpx/Bucket.py
+++ b/Python/pywarpx/Bucket.py
@@ -59,7 +59,9 @@ class Bucket(object):
         for attr, value in self.argvattrs.items():
             if value is None:
                 continue
-            # --- repr is applied to value so that for floats, all of the digits are included.
+            # --- str (and repr) crop some digits for floats. TODO: we should format
+            #     floating point numbers & numpy scalars to the significant digits of the
+            #     precision of amrex::Real/ParticleReal
             # --- The strip of "'" is then needed when value is a string.
             if isinstance(value, str):
                 if value.find('=') > -1:
@@ -73,11 +75,11 @@ class Bucket(object):
                     continue
                 # --- For lists, tuples, and arrays make a space delimited string of the values.
                 # --- The lambda is needed in case this is a list of strings.
-                rhs = ' '.join(map(lambda s : repr(s).strip("'"), value))
+                rhs = ' '.join(map(lambda s : str(s).strip("'"), value))
             elif isinstance(value, bool):
                 rhs = 1 if value else 0
             else:
                 rhs = value
-            attrstring = '{0}.{1} = {2}'.format(self.instancename, attr, repr(rhs).strip("'"))
+            attrstring = '{0}.{1} = {2}'.format(self.instancename, attr, str(rhs).strip("'"))
             result += [attrstring]
         return result
--- a/Python/pywarpx/Bucket.py
+++ b/Python/pywarpx/Bucket.py
@@ -75,11 +75,11 @@ class Bucket(object):
                     continue
                 # --- For lists, tuples, and arrays make a space delimited string of the values.
                 # --- The lambda is needed in case this is a list of strings.
-                rhs = ' '.join(map(lambda s : str(s).strip("'"), value))
+                rhs = ' '.join(map(lambda s : f'{s}', value))
             elif isinstance(value, bool):
                 rhs = 1 if value else 0
             else:
                 rhs = value
-            attrstring = '{0}.{1} = {2}'.format(self.instancename, attr, str(rhs).strip("'"))
+            attrstring = f'{self.instancename}.{attr} = {rhs}'
             result += [attrstring]
         return result
--- a/Python/pywarpx/Bucket.py
+++ b/Python/pywarpx/Bucket.py
@@ -59,7 +59,6 @@ class Bucket(object):
         for attr, value in self.argvattrs.items():
             if value is None:
                 continue
-            # --- The strip of "'" is then needed when value is a string.
             if isinstance(value, str):
                 if value.find('=') > -1:
                     # --- Expressions with temporary variables need to be inside quotes
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 # basic dependencies
-numpy~=1.15
+numpy>=1.15
 periodictable~=1.5
 
 # PICMI


Python/pywarpx/Bucket.py
```diff
@@ -59,7 +59,9 @@ def attrlist(self):
         for attr, value in self.argvattrs.items():
             if value is None:
                 continue
-            # --- repr is applied to value so that for floats, all of the digits are included.
+            # --- str (and repr) crop some digits for floats. TODO: we should format
```
dpgrote: Doing some quick tests show that both repr and str will include all digits, as will f-strings.
```
>>> repr(np.pi)
'3.141592653589793'
>>> str(np.pi)
'3.141592653589793'
>>> f'{np.pi}'
'3.141592653589793'
>>> np.pi - float(f'{np.pi}')
0.0
>>> 
```
So, maybe use the f-string to avoid the explicit call to `str`?

Python/pywarpx/Bucket.py
```diff
@@ -73,11 +75,11 @@ def attrlist(self):
                     continue
                 # --- For lists, tuples, and arrays make a space delimited string of the values.
                 # --- The lambda is needed in case this is a list of strings.
-                rhs = ' '.join(map(lambda s : repr(s).strip("'"), value))
+                rhs = ' '.join(map(lambda s : str(s).strip("'"), value))
```
dpgrote:
```suggestion
                rhs = ' '.join(map(lambda s : f'{s}', value))
```
Note that with the f-string or str, the strip is not needed.

Python/pywarpx/Bucket.py
```diff
@@ -73,11 +75,11 @@ def attrlist(self):
...
             elif isinstance(value, bool):
                 rhs = 1 if value else 0
             else:
                 rhs = value
-            attrstring = '{0}.{1} = {2}'.format(self.instancename, attr, repr(rhs).strip("'"))
+            attrstring = '{0}.{1} = {2}'.format(self.instancename, attr, str(rhs).strip("'"))
```
dpgrote:
```suggestion
            attrstring = f'{self.instancename}.{attr} = {rhs}'
```

Python/pywarpx/Bucket.py
```diff
@@ -59,7 +59,9 @@ def attrlist(self):
         for attr, value in self.argvattrs.items():
             if value is None:
                 continue
-            # --- repr is applied to value so that for floats, all of the digits are included.
+            # --- str (and repr) crop some digits for floats. TODO: we should format
+            #     floating point numbers & numpy scalars to the significant digits of the
+            #     precision of amrex::Real/ParticleReal
```
dpgrote: I don't think that this would be needed. The values will be truncated anyway when read in. Though, a question to @WeiqunZhang is what happens when an input parameter has more digits than the type being read into? Is it rounded or truncated?

ax3l: On the AMReX/C++ side, extra digits above double will most certainly just be dropped.

ax3l:
```suggestion
```

WeiqunZhang: In principle, we need to format it to have `std::numeric_limits<T>::max_digits10` digits. For `double` that's 17. Python's `str` seems to only print 16 digits.

In AMReX, we use `std::sstream` to read numbers. It's significant up to `std::numeric_limits<T>::max_digits10`.

Python/pywarpx/Bucket.py
```diff
@@ -59,7 +59,6 @@ def attrlist(self):
         for attr, value in self.argvattrs.items():
             if value is None:
                 continue
-            # --- repr is applied to value so that for floats, all of the digits are included.
             # --- The strip of "'" is then needed when value is a string.
```
dpgrote: One more thing - this comment can be deleted.

ax3l: The comment on `'` stripping?
```suggestion
```

Python/pywarpx/Bucket.py
dpgrote: I think str gives all the digits. I've tested this `x - float(f'{x}')` on random numbers and it always gives zero.

dpgrote (approved):
Looks good! Thanks!

Python/pywarpx/Bucket.py
WeiqunZhang:
```
        std::cout << "digits10: " << std::numeric_limits<double>::digits10 << '\n'
                  << "max_digits10: " << std::numeric_limits<double>::max_digits10 << '\n';
        std::string str("1.2345678901234567890");
        for (int n = 15; n <= 17; ++n) {
            std::stringstream ss(str.substr(0,n+1)); // +1 because of `.`.                                                                                                   
            double num;
            ss >> num;
            std::cout << "With " << n << " digits, num == 1.2345678901234567890: " << (num == 1.2345678901234567890) << '\n';
        }
```
produces
```
digits10: 15
max_digits10: 17
With 15 digits, num == 1.2345678901234567890: 0
With 16 digits, num == 1.2345678901234567890: 0
With 17 digits, num == 1.2345678901234567890: 1
```

WeiqunZhang: What said about python was wrong. I happened to use a number whose 17th digit is small.
```
In [8]: str(1.23456789012345631234)
Out[8]: '1.2345678901234562'
```
It does print 17 digits.

dpgrote:
```
>>> x = np.float64(1.2345678901234567890)
>>> x - float(f'{x}')
np.float64(0.0)
```

EZoni:
Here as much as in other recent PRs, the test `collisionXYZ` seems to fail again, due to checksum differences.

As noted in a comment on Slack, this seems to be the second occurence of what happened a few weeks ago, which prompted PR #5042 by @RemiLehe. 

A similar, separate change might be needed now, although we don't fully understand why this is happening. 

Note that we ran the test in DEBUG mode with several `-fsanitize` options and did not find any evident issues.

ax3l:
Thanks for restarting CI!

We could increase the tolerance for this test, I assume. Note that Justin also merged a few fixes to collision algorithms recently and one more is coming. We could check if this persists after those landed.

PR #5076, opened 2024-07-23
Update script to compile WarpX on Frontier (OLCF)

lucafedeli88:
After the recent update of the software environment on Frontier, we need to update the scripts to compile WarpX on that machine.


ax3l (approved):
Thank you very much!

Did you have a chance to verify the Cray compiler does not slow down the current deposition in the latest version? Hope the AMD ROCm fixes finally made it upstream to LLVM and downstream to Cray again.

PR #5079, opened 2024-07-23
Add Dan Barnes' effective potential (semi-implicit) Poisson solver

roelof-groenewald:
This PR adds a new electrostatic solver based on the semi-implicit scheme developed by [Barnes, Journal of Comp. Phys., 424 (2021), 109852](https://www.sciencedirect.com/science/article/pii/S0021999120306264?via%3Dihub) (see Appendix A of reference). 

The implementation was tested through the simulation of a Langmuir probe placed inside a uniform plasma. To this end a 2d simulation was performed with an initial uniform plasma with a conducting disk inserted in the center of the domain. The disk is kept at 0 V while Neumann boundary conditions are used for the domain boundary. Electrons and hydrogen ions were injected from all the domain boundaries using the `UniformFluxDistribution` in order to simulate the particle flux from a uniform plasma. Thermal boundaries were used for the particles with the same temperatures as the initial plasma particles. The expected outcome of this simulation is that a sheath develops around the "Langmuir probe" with value
$V_s = 0.5T_e\ln\left(2\pi\frac{m_e}{2m_p}(1 + \frac{T_i}{T_e})\right) $.
A pre-sheath of roughly $0.7T_e$ is also expected to form but the exact value of this depends on the domain size since the pre-sheath extends for many Debye lengths.
The figure below shows an example outcome from a simulation as described above in which the semi-implicit factor was set to $C_{SI} = 10$.
image

Required before merging:

- [x] Merging of https://github.com/AMReX-Codes/amrex/pull/3968
- [x] Add example to CI tests
- [x] Update documentation


archermarx:
Neat! I think the paper reference is off, though. That one points to 

[Improved C1 shape functions for simplex meshes](https://www.sciencedirect.com/science/article/pii/S0021999120306264?via%3Dihub)

Is the paper you're referring to this one? (Barnes, Journal of Computational Physics., (2022) 111151)

[Time-explicit Darwin PIC algorithm](https://www.sciencedirect.com/science/article/pii/S0021999122002133)

roelof-groenewald:
I updated the description to specify Appendix A of the reference (see https://www.osti.gov/servlets/purl/1670766) where the semi-implicit method is described. I believe this reference was earlier than the Darwin model paper which is why I opted for it, but it is nice to have the other paper referenced in the PR as well. Thanks @archermarx!

roelof-groenewald:
For the sake of simplicity I opted for a different example problem. I now use a simulation of a plasma column expanding inside a conducting cylinder. This problem is useful to show the strength (and accuracy) of the semi-implicit scheme. The expansion of the plasma is tracked at fixed radii. The result below compare the semi-implicit results to those from a fully resolved, explicit simulation as well as a under-resolved explicit simulation.
comparison
Note that the resolved explicit results and the semi-implicit results agree very well even though the semi-implicit simulation was 4x faster.

archermarx:
I've installed this branch on my cluster and am testing it out on a thruster simulation benchmark. If it works this semi-implicit method will be very helpful for our research. Let me know if there's anything I need to watch out for.

Docs/source/usage/parameters.rst
```diff
@@ -228,6 +228,16 @@ Overall simulation parameters
         \boldsymbol{\nabla}^2 \phi = - \rho/\epsilon_0 \qquad \boldsymbol{E} = - \boldsymbol{\nabla}\phi \\
         \boldsymbol{\nabla}^2 \boldsymbol{A} = - \mu_0 \boldsymbol{j} \qquad \boldsymbol{B} = \boldsymbol{\nabla}\times\boldsymbol{A}
 
+    * ``labframe-semi-implicit``: Poisson's equation is solved with a modified dielectric function
+      to create a semi-implicit scheme in which grid heating effects due to unresolved
```
RemiLehe: Could you add a reference to the appendix A of Dan Barnes' paper?
Also: maybe we should discuss in more detail what we mean by "unresolved" plasma modes here: we could state that this suppresses the instability that happens when $\omega_p \Delta t > 2$, and allows to take larger timesteps.

RemiLehe: Also, is there any guidance for how to set `C_{SI}`? And could you define "marginally stable"?

Examples/Tests/semi_implicit_electrostatic/analysis.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+
+sys.path.insert(1, "../../../../warpx/Regression/Checksum/")
+import checksumAPI
+
+filename = sys.argv[1]
```
RemiLehe: Could you add physics tests that checks that the expansion of the cylinder is as expected?
(and that would fail e.g. when using the regular explicit labframe solver instead of the semi-implicit labframe solver)

Docs/source/usage/parameters.rst
RemiLehe: Also: I wonder if there is a risk of confusion with `algo.evolve_scheme = semi_implicit_em`, which is not directly related.

One clear difference is the `semi_implicit_em` is EM, while this PR adds an ES algorithm.
However, there is potentially a risk that users would think that `labframe-semi-implicit` is the ES version of `semi_implicit_em`, esp. since both algorithms allow `w_p*dt > 2`. There are some key differences though:
- the ES version of `semi_implicit_em` (not implemented in WarpX) iterates the particle pusher, and would be exactly energy conserving.
- `labframe-semi-implicit` does not need to iterate the particle pusher, and is not exactly energy conserving.

After some offline discussion, it seems that the term "direct implicit" would be appropriate here to avoid confusion with "semi-implicit". @roelof-groenewald said he will follow up on checking any potential difference between direct implicit and the method proposed by D. Barnes.

Examples/Tests/semi_implicit_electrostatic/inputs_test_2d_semi_implicit_electrostatic_picmi.py
```diff
@@ -0,0 +1,243 @@
+#!/usr/bin/env python3
+#
+# --- Test script for the semi-implicit Poisson solver in which a plasma column
+# --- expands inside a conducting cylinder.
```
RemiLehe: Could you add a bit more text regarding what this test is checking, how to know whether things are behaving as expected, etc.

Docs/source/usage/parameters.rst
RemiLehe: ~~Could you also include how the documentation for how to set the C_SI from the input script?~~ I did not see it, but it is there.

Source/Diagnostics/Diagnostics.cpp
```diff
@@ -77,7 +77,8 @@ Diagnostics::BaseReadParameters ()
     if (utils::algorithms::is_in(m_varnames_fields, "phi")){
         WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
             warpx.electrostatic_solver_id==ElectrostaticSolverAlgo::LabFrame ||
-            warpx.electrostatic_solver_id==ElectrostaticSolverAlgo::LabFrameElectroMagnetostatic,
+            warpx.electrostatic_solver_id==ElectrostaticSolverAlgo::LabFrameElectroMagnetostatic ||
+            warpx.electrostatic_solver_id==ElectrostaticSolverAlgo::LabFrameSemiImplicit,
             "plot phi only works if do_electrostatic = labframe or do_electrostatic = labframe-electromagnetostatic");
```
RemiLehe:
```suggestion
            "plot phi only works if do_electrostatic = labframe, labframe-electromagnetostatic, or labframe-semi-implicit");
```

Docs/source/usage/parameters.rst
```diff
@@ -228,6 +228,16 @@ Overall simulation parameters
...
+      plasma modes are suppressed. If this option is used the additional parameter
+      ``warpx.semi_implicit_factor`` can also be specified to set the value of :math:`C_{SI}`.
```
RemiLehe: We can mention the default value for `semi_implicit_factor` here.

The simulation will be stable if `C_{SI} \geq 1` + mention disadvantages of setting it too high.

archermarx:
Just got some results on my problem (2D x-z). Switching from the normal electrostatic sim to the semi-implicit solver (with factor 2) seems to result in no fields whatsoever being computed:

Here's the expected field at time 0:
E_ind=0000_time=0 00us

Here's what the semi-implicit solver predicts with the exact same timestep, grid size, and everything:
E_ind=0000_time=0 00us

I've attached my script. I ran it with these args:
`--semi-implicit --semi-implicit-factor=2 --timestep-factor=1 --grid-factor=1 --mlmg-precision=1e-3`

For the standard version, I just removed the semi-implicit args.

[picmi_hall.py.txt](https://github.com/user-attachments/files/17479155/picmi_hall.py.txt)

roelof-groenewald:
Thanks @archermarx! I'll use your script to debug and let you know what I find. I won't have time to get to it until early next week however. Sorry for the delay.

archermarx:
No worries! Thanks for writing this feature!

Examples/Tests/semi_implicit_electrostatic/inputs_test_2d_semi_implicit_electrostatic_picmi.py
roelof-groenewald: I added more detail in the new test. See description below.

roelof-groenewald:
The CI test was updated to more-or-less follow benchmark C from https://doi.org/10.1109/TPS.2021.3072353, but using different parameters in order to exploit the semi-implicit value. Nonetheless, the plasma expansion matches the analytic solution from that paper (dashed lines show analytic result at each time point):
image

roelof-groenewald:
@archermarx I found the reason why you didn't get any E-field - the E-field was only updated if EBs were used. I fixed that issue now. Would you please pull the latest changes and try your case again?

archermarx:
Yes, I'm compiling and will test shortly.

archermarx:
Can confirm, this particular issue is fixed. I'm going to try out longer runs now to see how the physics are affected.

image

roelof-groenewald:
> Can confirm, this particular issue is fixed. I'm going to try out longer runs now to see how the physics are affected.

Awesome! Thanks for checking. I'd recommend moving slowly towards larger cell sizes and larger timesteps. Maybe first using the same settings as you use explicitly. It is important to keep the electron streaming factor below 1.

archermarx:
The results look pretty good to me, I ran at 2x grid size and 2x timestep with a factor of 2 and got results quite close to the explicit results with a total speedup of ~4x. I will run more tests but overall this seems to work great!

roelof-groenewald:
> The results look pretty good to me, I ran at 2x grid size and 2x timestep with a factor of 2 and got results quite close to the explicit results with a total speedup of ~4x. I will run more tests but overall this seems to work great!

Thanks for following up! I'm glad to hear it is working for your use case so far. I'd be curious to hear how far you can push it before the physics start to suffer. We've run it using $\Delta t \omega_p$ up to ~100 without losing accuracy on the effects we were interested in but the ceiling will really depend on the specifics of the problem.

archermarx:
Interesting! I'm running the explicit version with larger dt and dx to start as a point of comparison, but I'll try 4x now. We're quite close to the streaming limit already so we can only really increase dt by increasing dx. The main question for us is how the larger effective plasma frequency and grid size affect the growth and propagation of the instability we care about, its effects on the plasma, and any numerical heating we might see.

archermarx:
Do you have any guidance for choosing the implicit parameter? i know it needs to be greater than 1, and that higher values are more stable, but any other rules of thumb?

roelof-groenewald:
> Do you have any guidance for choosing the implicit parameter? i know it needs to be greater than 1, and that higher values are more stable, but any other rules of thumb?

At a given $\Delta t$ the higher you make $C_{SI}$ the more you suppress the plasma frequency, or alternatively, at a fixed $C_{SI}$ the larger your $\Delta t$ the more you suppress the plasma frequency. Using $C_{SI} = 4$ we get something like this:
image

 So generally I try to run with the lowest $C_{SI}$ I can get away with to minimize the possibility of the plasma mode hybridizing with some mode I care about and changing the results. Luckily for me I usually care about ion modes which are so far removed from the plasma mode that there is a long way the plasma mode can be lowered before it gets close to frequencies I care about. Hope that helps!

archermarx:
Appreciate it, that's useful.

roelof-groenewald:
@RemiLehe and @archermarx I spoke with Dan Barnes and he suggested we use "effective potential" as the name for this method (giving EP-PIC) in order to address Remi's concern of possible confusion with the semi-implicit method of Chen et al.
I likely missed some spots in the code where the method is referred to as semi-implicit so please let me know if you spot any such occurrences.

Source/FieldSolver/ElectrostaticSolvers/EffectivePotentialES.H
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * This file is part of WarpX.
+ *
+ * Authors: Roelof Groenewald (TAE Technologies)
+ *
+ * License: BSD-3-Clause-LBNL
+ */
+
+#ifndef WARPX_SEMIIMPLICITES_H_
+#define WARPX_SEMIIMPLICITES_H_
```
archermarx:
```suggestion
#ifndef WARPX_EFFECTIVEPOTENTIALES_H_
#define WARPX_EFFECTIVEPOTENTIALES_H_
```

Source/FieldSolver/ElectrostaticSolvers/EffectivePotentialES.H
```diff
@@ -0,0 +1,71 @@
...
+    void ComputeSpaceChargeField (
+        ablastr::fields::MultiFabRegister& fields,
+        MultiParticleContainer& mpc,
+        MultiFluidContainer* mfl,
+        int max_level) override;
+
+    void ComputeSigma ( amrex::MultiFab& sigma ) const;
+
+    /**
+     * Compute the potential `phi` by solving the semi-implicit Poisson equation
```
archermarx:
```suggestion
     * Compute the potential `phi` by solving the semi-implicit Poisson equation using the Effective Potential method
```

Source/FieldSolver/ElectrostaticSolvers/EffectivePotentialES.H
```diff
@@ -0,0 +1,71 @@
...
+        amrex::MultiFab const& sigma,
+        amrex::Real required_precision,
+        amrex::Real absolute_tolerance,
+        int max_iters,
+        int verbosity
+    ) const;
+
+};
+
+#endif // WARPX_SEMIIMPLICITES_H_
```
archermarx:
```suggestion
#endif // WARPX_EFFECTIVEPOTENTIALES_H_
```

archermarx (commented):
Think i found any remaining references to semi-implicit on its own

Source/FieldSolver/ElectrostaticSolvers/EffectivePotentialES.H
archermarx: This one is probably not necessary to change

roelof-groenewald:
> Think i found any remaining references to semi-implicit on its own

Thanks so much!

archermarx:
Just checking in here, are there any remaining obstacles to merging this?

RemiLehe:
Sorry: I was supposed to review this PR, but got side-tracked by other deadlines. I can try to review it today.

Source/ablastr/fields/EffectivePotentialPoissonSolver.H
```diff
@@ -0,0 +1,274 @@
...
+ * \f[
+ *   \nabla \cdot \sigma \nabla \phi = - \rho/\epsilon_0
+ * \f]
+ *
+ * \tparam T_PostPhiCalculationFunctor a calculation per level directly after phi was calculated
+ * \tparam T_BoundaryHandler handler for boundary conditions, for example @see ElectrostaticSolver::PoissonBoundaryHandler
+ * \tparam T_FArrayBoxFactory usually nothing or an amrex::EBFArrayBoxFactory (EB ONLY)
+ * \param[in] rho The charge density a given species
+ * \param[out] phi The potential to be computed by this function
+ * \param[in] sigma The semi-implicit matrix
```
RemiLehe: Could you update this to remove the mention of `semi-implicit`?

roelof-groenewald:
```suggestion
 * \param[in] sigma The matrix representing the mass operator used to lower the local plasma frequency 
```

Examples/Tests/effective_potential_electrostatic/analysis.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+
+import dill
+import matplotlib.pyplot as plt
+import numpy as np
+from openpmd_viewer import OpenPMDTimeSeries
+from scipy.interpolate import RegularGridInterpolator
```
RemiLehe, line 17: Could you add a high-level text description that summarizes what this Python checks (e.g. density after the expansion of the plasma)?

Source/FieldSolver/ElectrostaticSolvers/EffectivePotentialES.cpp
```diff
@@ -0,0 +1,257 @@
...
+    if (!EB::enabled()) { computeE( Efield_fp, phi_fp, beta ); }
+}
+
+void EffectivePotentialES::computePhi (
+    ablastr::fields::MultiLevelScalarField const& rho,
+    ablastr::fields::MultiLevelScalarField const& phi) const
+{
+    // Calculate the mass enhancement factor:
+    // The "sigma" multifab stores the dressing of the Poisson equation. It
+    // is a cell-centered multifab.
```
RemiLehe, line 80: Could you again add a reference to Dan Barnes paper here, just to remind whoever reads the code where this comes from?

RemiLehe (approved):

PR #5080, opened 2024-07-23
Fix electrostatic FFT solver when called from python

### LINKED ISSUE #5070: Cannot get FFT method to work for electrostatic solver
The electrostatic solver has two methods to choose from, "Multigrid" and "FFT." I want to use FFT. I follow the documentation, which says that FFT only works:

1. in 3D
2. with "open" boundary conditions for the field
3. when WarpX is compiled with the flag `-DWarpX_FFT=ON`

I am using GPU execution, and no mesh refinement. Yet my runs quietly try to use the multigrid method instead of FFT, and then crash with an error complaining that the boundary conditions aren't suitable for multigrid:

```
### ERROR   : Field boundary conditions have to be either periodic, PEC or
#            neumann when using the electrostatic Multigrid solver, but they
#            are pml
#            (/global/homes/j/jwestern/src/warpx/Source/FieldSolver/ElectrostaticSolver.cpp:1057)
```

Does anyone know what's going on?

Thanks!

archermarx:
Fixes https://github.com/ECP-WarpX/WarpX/issues/5070

The problem was twofold. Firstly, the python interface never actually set the `pywarpx.warpx.poisson_solver` variable. Secondly, the poisson solver requires open boundaries. However, PICMI maps `open` to `pml`, meaning that you were unable to actually assign open boundaries from the python interface. This PR addresses the first issue with a one line change to the `picmi.py` file. The second change is a bit trickier. I propose renaming the PICMI boundaries such that `open` now refers to the expected `FieldBoundaryType::Open`, and introducing the `pml` option to refer to perfectly-matched layers, which is already standard in the AMReX parmparse input.

I can make a companion PR to PICMI to update the docs if this is amenable to people.


dpgrote:
Unfortunately, changing the `pml` boundary type would break the existing API. I can suggest a different way of fixing this. In `picmi.py`, in the class `ElectrostaticSolver`, just before the call to `self.grid.grid_initialize_inputs()`, the `BC_map` could be updated with the correct name of the open boundary for the electrostatic solver.
```
BC_map['open'] = 'open'
```
This is a bit of a kludge but should be ok.

archermarx:
OK, I can make that change instead.

PR #5082, opened 2024-07-24
[WIP] Recompute the macroscopic properties everytime the moving window moves.

### LINKED ISSUE #4952: macroscopic.epsilon_function(x,y,z) and sigma_function issue: breaks if z dependent
As discussed with Jean-Luc and Axel at IPAC, for epsilon and sigma functions that have a z dependence, the structure "disappears" and is essentially replaced by vacuum. These functions work for simple x-y geometries. I have attached a simplified example that just has the epsilon function, but the same issue arises with the sigma function. The weird thing is that if I am simulating a dielectric structure surrounded by a conductor, if either the epsilon OR the sigma function has z dependence, all structures "disappear". The following are the files attached here with quick descriptions:
- zshift_error_ex.txt - WarpX script detailing a rectangular dielectric tube offset in z by 5 mm. Structure is 100 mm long. Also includes a commented out epsilon function that has only x and y dependence that is known to work. I tried to upload the distribution script that is part of this WarpX script, but it won't take the h5 file. Let me know if there's a way you want me to send that or just use the gaussian bunch created by WarpX as the error will still appear and the difference between distributions should not be too much to recognize that there's an issue here.
- plotscript.txt - Python script to plot the y=0 slice of the transverse force (`np.sqrt(Fx**2+Fy**2)*kV for Fx=-(Exslice-cms*Byslice`) and `Fy=-(Eyslice+cms*Bxslice)`)
- good_diel.png (first image) and bad_diel.png (second image) - good_diel is transverse force plot for a dielectric tube surrounded by a conductor using WarpX functions that do not have z dependence; bad_diel is transverse force plot for dielectric tube without WarpX sigma_function, but PEC boundary in x and y (PML in z), like the zshift_error_ex.3d WarpX script attached.

I'm currently running a simulation without the moving window and while I can't tell exactly quite yet, it does seem like the epsilon function worked for that case. I will update within 24 hours if turning off the moving window works for sure. Thanks!

[zshift_error_ex.txt](https://github.com/ECP-WarpX/WarpX/files/15449134/zshift_error_ex.txt)
[plotscript.txt](https://github.com/ECP-WarpX/WarpX/files/15449135/plotscript.txt)

good_diel
bad_diel

RemiLehe:
When using the macroscopic properties in WarpX, `epsilon`, `mu` and `sigma` are computed on the grid at t=0, but were never updated as the moving window moves. As a consequence, the simulation will always see the `epsilon`, `mu` and `sigma` that were initially in the box, even once the moving window has moved far past the its initial positions.

This PR fixes this by calling `MacroscopicProperties::InitData` everytime the moving window moves. In the process, I had to separate the allocation of the arrays, in a new member function `AllocateLevelMFs` (which is only called at `t=0`) and the filling out of the data, in the function `InitData` (which is called at `t=0`, as well as everytime the moving window moves).

Fixes #4952 

TODO:
- [ ] Add an automated test

### DIFF
--- a/Source/FieldSolver/FiniteDifferenceSolver/MacroscopicProperties/MacroscopicProperties.cpp
+++ b/Source/FieldSolver/FiniteDifferenceSolver/MacroscopicProperties/MacroscopicProperties.cpp
@@ -128,8 +128,6 @@ MacroscopicProperties::InitData (
     const amrex::IntVect& Ey_stag,
     const amrex::IntVect& Ez_stag)
 {
-    amrex::Print() << Utils::TextMsg::Info("we are in init data of macro");
-
     // Define material property multifabs using ba and dmap from WarpX instance
     // sigma is cell-centered MultiFab
     m_sigma_mf = std::make_unique<amrex::MultiFab>(ba, dmap, 1, ng_EB_alloc);
--- a/Source/FieldSolver/FiniteDifferenceSolver/MacroscopicProperties/MacroscopicProperties.cpp
+++ b/Source/FieldSolver/FiniteDifferenceSolver/MacroscopicProperties/MacroscopicProperties.cpp
@@ -119,22 +119,26 @@ MacroscopicProperties::ReadParameters ()
 }
 
 void
-MacroscopicProperties::InitData (
+MacroscopicProperties::AllocateLevelMFs (
     const amrex::BoxArray& ba,
     const amrex::DistributionMapping& dmap,
-    const amrex::IntVect& ng_EB_alloc,
-    const amrex::Geometry& geom,
-    const amrex::IntVect& Ex_stag,
-    const amrex::IntVect& Ey_stag,
-    const amrex::IntVect& Ez_stag)
+    const amrex::IntVect& ng_EB_alloc )
 {
-    // Define material property multifabs using ba and dmap from WarpX instance
     // sigma is cell-centered MultiFab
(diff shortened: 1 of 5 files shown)


Source/Initialization/WarpXInitData.cpp
```diff
@@ -474,7 +474,7 @@ WarpX::InitData ()
 
     BuildBufferMasks();
 
-    if (WarpX::em_solver_medium==1) {
+    if (WarpX::em_solver_medium == MediumForEM::Macroscopic) {
```
RemiLehe, line 477: Minor cleanup, this is not key to the bug fix.

Source/FieldSolver/FiniteDifferenceSolver/MacroscopicProperties/MacroscopicProperties.cpp
```diff
@@ -128,8 +128,6 @@ MacroscopicProperties::InitData (
     const amrex::IntVect& Ey_stag,
     const amrex::IntVect& Ez_stag)
 {
-    amrex::Print() << Utils::TextMsg::Info("we are in init data of macro");
```
RemiLehe, line 131: Minor cleanup, this is not key to the bug fix.

ax3l (approved):
Excellent, thank you!

ax3l:
Ooops, yes we need a test as you indicated. Let's do in a follow-up PR :)

PR #5086, opened 2024-07-26
Implicit add filtering

dpgrote:
This adds filtering to the implicit solver, replacing PR #4600.

It is a simple change. All that is needed is adding a call to filter the `Efield_fp` just before the particles are pushed. The current density is already filtered in `SyncCurrentAndRho`.

The name of the routine `ApplyFilterJ` was changed to `ApplyFilterMF` since it now has a more general usage.


Source/Evolve/WarpXEvolve.cpp
```diff
@@ -569,7 +569,7 @@ void WarpX::SyncCurrentAndRho ()
             {
                 // TODO This works only without mesh refinement
                 const int lev = 0;
-                if (use_filter) { ApplyFilterJ(current_fp_vay, lev); }
+                if (use_filter) { ApplyFilterMF(current_fp_vay, lev); }
             }
```
JustinRayAngus: Should ApplyFilterJ() be different than applying a filter for E or B when using curvilinear geometries? 

I believe the filter should be applied to the current before dividing by the volume element to convert to current density. Perhaps this is being done somewhere else?

dpgrote: You are probably correct, but filtering doesn't work for RZ as it is now (at least for FDTD). As it is now, the inverse volume scaling is done before the filtering. That order could be changed but that would need to be a separate PR.

JustinRayAngus:
This looks good to me.

We could also add filtering to the magnetic field as an option. That should probably be a separate PR though.

JustinRayAngus (approved):
Thanks Dave for this PR!

PR #4937, opened 2024-05-16
Integrated Green's Function Poisson Solver using heFFTe

Haavaan:
This PR adds distributed FFTs in the Integrated Green's Function (IGF) Poisson Solver, which enhances its computational efficiency. It is useful for improving the performance of the IGF Poisson Solver.

PR #5089, opened 2024-07-27
Add quasi-3D Integrated Green Functions solver 

aeriforme:
This PR implements a quasi-3D Poisson solver based on the Integrated Green's functions.
It solves the 2D Poisson equation on the `(x,y)` plane for every slice `z`. 
It is useful for beam-beam simulations. 

See PR #4648 for the full 3D solver.  

- [x] AMReX FFT - works with https://github.com/AMReX-Codes/amrex/pull/4255
- [x] CI test
- [x] documentation

:hocho: :fish:


Docs/source/usage/parameters.rst
```diff
@@ -265,6 +265,16 @@ Overall simulation parameters
         In electromagnetic mode, this solver can be used to initialize the species' self fields
         (``<species_name>.initialize_self_fields=1``) provided that the field BCs are PML (``boundary.field_lo,hi = PML``).
 
+          * warpx.use_2d_slices_fft_solver (`bool`, default: 0): Select the type of Integrated Green Function solver.
+            If 0, solve Poisson equation in full 3D geometry.
+            If 1, solve Poisson equation in a quasi 3D geometry. In this case, the code performes many 2D Poisson solves
+            on all :math:`(x,y)` slices, each slice at a given :math:`z`.
```
RemiLehe: Maybe a more useful description would be: "This solver neglects the z derivatives in the Laplacian of the Poisson equation. This is often a good approximation for ultra-relativistic beams propagating along the z direction, with the `relativistic` solver. As a consequence, this solver does not need to do an FFT along the z direction, and instead uses only transverse FFTs (along x and y) at each z position (or z "slice")."

Docs/source/usage/parameters.rst
```diff
@@ -265,6 +265,16 @@ Overall simulation parameters
...
+
+          * warpx.use_distributed_3d_fft_solver (`bool`, default: 0): Choose whether the 3D FFTs performed in the
+            full 3D Integrated Green Function solver are distributed.
+            If 0, the FFTs are performed on a single process (the rest of the code is still fully parallel).
```
```suggestion
            If 0, the FFTs are performed on a single MPI rank (the rest of the code is still fully parallel).
```

Regression/Checksum/benchmarks_json/test_3d_open_bc_poisson_solver_heffte.json
```diff
@@ -0,0 +1,20 @@
+{
```
RemiLehe: Interesting. Didn't we already have a checksum file for this test before?

Source/FieldSolver/ElectrostaticSolvers/LabFrameExplicitES.H
RemiLehe, line 1: Is this file used? Was it committed by accident?

Source/FieldSolver/ElectrostaticSolvers/RelativisticExplicitES.H
RemiLehe, line 1: Is this file used? Was it committed by accident?

Source/ablastr/fields/IntegratedGreenFunctionSolver.H
```diff
@@ -49,32 +47,75 @@ namespace ablastr::fields
...
+          - IntegratedPotential3D( x+0.5_rt*dx, y+0.5_rt*dy, z-0.5_rt*dz )
+          + IntegratedPotential3D( x-0.5_rt*dx, y+0.5_rt*dy, z-0.5_rt*dz )
+          + IntegratedPotential3D( x+0.5_rt*dx, y-0.5_rt*dy, z-0.5_rt*dz )
+          - IntegratedPotential3D( x-0.5_rt*dx, y-0.5_rt*dy, z-0.5_rt*dz )
+        );
+        return G_value;
+    }
 
 
+    /** @brief Implements equation 2 in https://doi.org/10.1103/PhysRevSTAB.10.129901
```
RemiLehe: I believe that this reference is incorrect (it points to the 3D solver).
Could you add the reference for the 2D solver?

Source/ablastr/fields/IntegratedGreenFunctionSolver.cpp
```diff
@@ -38,16 +39,23 @@ void
...
     BL_PROFILE_VAR_NS("ablastr::fields::computePhiIGF: FFTs", timer_ffts);
     BL_PROFILE_VAR_NS("ablastr::fields::computePhiIGF: FFT plans", timer_plans);
     BL_PROFILE_VAR_NS("ablastr::fields::computePhiIGF: parallel copies", timer_pcopies);
+    BL_PROFILE_VAR_NS("ablastr::fields::computePhiIGF: Green functions", timer_Gfunc);
 
     BL_PROFILE("ablastr::fields::computePhiIGF");
 
+    // Check that the user-defined inputs are consistent
+    ABLASTR_ALWAYS_ASSERT_WITH_MESSAGE(!(is_2d_slices && is_3d_distributed), \
+    "Must choose between the quasi-3D and the fully 3D solver.");
```
RemiLehe:
```suggestion
    "Must choose between the 2D slice solver and the fully 3D solver.");
```

Source/ablastr/fields/IntegratedGreenFunctionSolver.cpp
```diff
@@ -38,16 +39,23 @@ void
...
+                bool const is_3d_distributed)
 {
     using namespace amrex::literals;
 
     BL_PROFILE_VAR_NS("ablastr::fields::computePhiIGF: FFTs", timer_ffts);
     BL_PROFILE_VAR_NS("ablastr::fields::computePhiIGF: FFT plans", timer_plans);
     BL_PROFILE_VAR_NS("ablastr::fields::computePhiIGF: parallel copies", timer_pcopies);
+    BL_PROFILE_VAR_NS("ablastr::fields::computePhiIGF: Green functions", timer_Gfunc);
 
     BL_PROFILE("ablastr::fields::computePhiIGF");
```
RemiLehe: Should we add a check that raises an error if `is_3d_distributed` is selected, but HEFFTE is not available?

Source/ablastr/fields/IntegratedGreenFunctionSolver.cpp
```diff
@@ -124,17 +133,10 @@ computePhiIGF ( amrex::MultiFab const & rho,
...
-#if !defined(ABLASTR_USE_HEFFTE)
-    // Without distributed FFTs (i.e. without heFFTe):
-    // We loop over the original box (not the 2x wider one), and the other quadrants by periodicity
-    amrex::BoxArray const& igf_compute_box = amrex::BoxArray( domain );
-#else
-    // With distributed FFTs (i.e. with heFFTe):
-    // We loop over the full 2x wider box, since 1 MPI rank does not necessarily own the data for the other quadrants
-    amrex::BoxArray const& igf_compute_box = tmp_G.boxArray();
-#endif
+    amrex::BoxArray const& igf_compute_box = (is_2d_slices || is_3d_distributed) ? tmp_G.boxArray() : amrex::BoxArray(domain);
```
RemiLehe: Could you reintroduce some of the comments (or modified versions thereof) to explain what this line does?
(People reading this code might not be sure what its purpose is.)

Source/ablastr/fields/IntegratedGreenFunctionSolver.cpp
```diff
@@ -145,57 +147,71 @@ computePhiIGF ( amrex::MultiFab const & rho,
...
+
+                if (is_2d_slices) {
+                    // 2D
+                    if ((i0< nx)&&(j0< ny)) { tmp_G_arr(i,j,k) = SumOfIntegratedPotential2D(x,      y,      dx, dy); }
+                    if ((i0< nx)&&(j0> ny)) { tmp_G_arr(i,j,k) = SumOfIntegratedPotential2D(x,      y_hi-y, dx, dy); }
+                    if ((i0> nx)&&(j0> ny)) { tmp_G_arr(i,j,k) = SumOfIntegratedPotential2D(x_hi-x, y_hi-y, dx, dy); }
+                    if ((i0> nx)&&(j0< ny)) { tmp_G_arr(i,j,k) = SumOfIntegratedPotential2D(x_hi-x, y,      dx, dy); }
+                } else {
+                    int const k0 = k - lo[2];
+                    amrex::Real const z = k0*dz;
```
RemiLehe: Could we introduce `z_hi` at this point?

Source/ablastr/fields/IntegratedGreenFunctionSolver.cpp
```diff
@@ -145,57 +147,71 @@ computePhiIGF ( amrex::MultiFab const & rho,
...
 
-        // Fill values of the Green function
         amrex::Real const dx = cell_size[0];
         amrex::Real const dy = cell_size[1];
         amrex::Real const dz = cell_size[2];
 
+        amrex::Real const x_hi = dx*(hi[0]+2);
+        amrex::Real const y_hi = dy*(hi[1]+2);
+        amrex::Real const z_hi = dz*(hi[2]+2);
+        amrex::ignore_unused(z_hi);
```
RemiLehe:
```suggestion
```

(See further below)

Source/ablastr/fields/IntegratedGreenFunctionSolver.cpp
```diff
@@ -145,57 +147,71 @@ computePhiIGF ( amrex::MultiFab const & rho,
...
+                    // 2D
+                    if ((i0< nx)&&(j0< ny)) { tmp_G_arr(i,j,k) = SumOfIntegratedPotential2D(x,      y,      dx, dy); }
+                    if ((i0< nx)&&(j0> ny)) { tmp_G_arr(i,j,k) = SumOfIntegratedPotential2D(x,      y_hi-y, dx, dy); }
+                    if ((i0> nx)&&(j0> ny)) { tmp_G_arr(i,j,k) = SumOfIntegratedPotential2D(x_hi-x, y_hi-y, dx, dy); }
+                    if ((i0> nx)&&(j0< ny)) { tmp_G_arr(i,j,k) = SumOfIntegratedPotential2D(x_hi-x, y,      dx, dy); }
+                } else {
+                    int const k0 = k - lo[2];
+                    amrex::Real const z = k0*dz;
+#if defined(ABLASTR_USE_HEFFTE)
+                    if (is_3d_distributed) {
```
RemiLehe:
```suggestion
                    if (is_3d_distributed) {
                                amrex::Real const z_hi = dz*(hi[2]+2);
```

Source/ablastr/fields/IntegratedGreenFunctionSolver.cpp
```diff
@@ -212,81 +228,156 @@ computePhiIGF ( amrex::MultiFab const & rho,
...
+        const int nry = local_box.length(1);
+        const int nrz = local_box.length(2);
+
+        // Number of cells in spectral space
+        const int nsx = c_local_box.length(0);
+        const int nsy = c_local_box.length(1);
+        const int nsz = c_local_box.length(2);
+
+        if(is_2d_slices){
+                        // Create many plans at once
```
RemiLehe: Could you fix the indentation?

Source/ablastr/fields/IntegratedGreenFunctionSolver.cpp
```diff
@@ -212,81 +228,156 @@ computePhiIGF ( amrex::MultiFab const & rho,
...
+            BL_PROFILE_VAR_START(timer_ffts);
+            ablastr::math::anyfft::Execute(backward_plan);
+            BL_PROFILE_VAR_STOP(timer_ffts);
+
+            // Remove plans
+            ablastr::math::anyfft::DestroyPlan(forward_plan_G);
+            ablastr::math::anyfft::DestroyPlan(forward_plan_rho);
+            ablastr::math::anyfft::DestroyPlan(backward_plan);
+
+            // Normalize, since (FFT + inverse FFT) results in a factor N but in 2d
```
RemiLehe:
```suggestion
            // Normalize, since (FFT + inverse FFT) results in a factor N
```

RemiLehe: It wasn't clear to me what "but in 2d" meant here.

Source/ablastr/fields/PoissonSolver.H
```diff
@@ -17,7 +17,7 @@
 #include <ablastr/fields/MultiFabRegister.H>
 #include <ablastr/profiler/ProfilerWrapper.H>
 
-#if defined(ABLASTR_USE_FFT) && defined(WARPX_DIM_3D)
+#if (defined(ABLASTR_USE_FFT) || defined(ABLASTR_USE_HEFFTE)) && defined(WARPX_DIM_3D)
```
RemiLehe: Hm, now that `is_igf_3d` is a runtime option, I think that `ABLASTR_USE_FFT` is always required, correct?

Source/ablastr/fields/PoissonSolver.H
```diff
@@ -232,16 +234,22 @@ computePhi (
                                        "Embedded boundary solve requested but not compiled in");
 #endif
 
-#if !defined(ABLASTR_USE_FFT)
+#if (!defined(ABLASTR_USE_FFT)) && (!defined(ABLASTR_USE_HEFFTE))
```
RemiLehe: Same comment as above: since `is_igf_3d` is a run-time option, `ABLASTR_USE_FFT` is always required.

Source/ablastr/math/fft/AnyFFT.H
```diff
@@ -156,6 +156,19 @@ namespace ablastr::math::anyfft
     FFTplan CreatePlan(const amrex::IntVect& real_size, amrex::Real* real_array,
                        Complex* complex_array, direction dir, int dim);
 
+    /** \brief create FFT plan for the backend FFT library.
+     * \param[in] real_size Size of the real array, along each dimension.
+     *                      Only the first dim elements are used.
+     * \param[out] real_array Real array from/to where R2C/C2R FFT is performed
+     * \param[out] complex_array Complex array to/from where R2C/C2R FFT is performed
+     * \param[in] dir direction, either R2C or C2R
+     * \param[in] dim direction, number of dimensions of the arrays. Must be <= AMREX_SPACEDIM.
```
RemiLehe: Could you add documentation for the other parameters as well?

Source/ablastr/fields/IntegratedGreenFunctionSolver.cpp
```diff
@@ -212,81 +228,156 @@ computePhiIGF ( amrex::MultiFab const & rho,
...
+
+        if(is_2d_slices){
+                        // Create many plans at once
+            int fft_size[] = {nry, nrx};
+            BL_PROFILE_VAR_START(timer_plans);
+            ablastr::math::anyfft::FFTplan forward_plan_rho = ablastr::math::anyfft::CreatePlanMany(
+                fft_size, tmp_rho[local_boxid].dataPtr(),
+                reinterpret_cast<ablastr::math::anyfft::Complex*>(tmp_rho_fft.dataPtr()),
+                ablastr::math::anyfft::direction::R2C, AMREX_SPACEDIM-1,
+                nrz, nullptr, 1, nrx*nry, nullptr, 1, nsx*nsy);
```
RemiLehe: Could we modify the interface of `CreatePlanMany` so that the user does not need to pass `nullptr, 1` ?

Source/ablastr/fields/IntegratedGreenFunctionSolver.cpp
```diff
@@ -212,81 +228,156 @@ computePhiIGF ( amrex::MultiFab const & rho,
...
+        const int nsz = c_local_box.length(2);
+
+        if(is_2d_slices){
+                        // Create many plans at once
+            int fft_size[] = {nry, nrx};
+            BL_PROFILE_VAR_START(timer_plans);
+            ablastr::math::anyfft::FFTplan forward_plan_rho = ablastr::math::anyfft::CreatePlanMany(
+                fft_size, tmp_rho[local_boxid].dataPtr(),
+                reinterpret_cast<ablastr::math::anyfft::Complex*>(tmp_rho_fft.dataPtr()),
+                ablastr::math::anyfft::direction::R2C, AMREX_SPACEDIM-1,
```
RemiLehe: Could you add a comment to explain why this is `AMREX_SPACEDIM-1`  and not `AMREX_SPACEDIM`?
(i.e. we are doing the FFT only in the transverse direction.)

Source/ablastr/math/fft/WrapRocFFT.cpp
```diff
@@ -66,6 +66,69 @@ namespace ablastr::math::anyfft
...
+    FFTplan CreatePlanMany(int * real_size, amrex::Real * real_array,
+                           Complex * complex_array, const direction dir, const int dim,
+                           int howmany, int * inembed, int istride, int idist,
+                           int * onembed, int ostride, int odist)
+    {
+        FFTplan fft_plan;
+
+        const std::size_t lengths[] = {AMREX_D_DECL(std::size_t(real_size[0]),
+                                                    std::size_t(real_size[1]),
+                                                    std::size_t(real_size[2]))};
```
RemiLehe: In this particular application, `real_size[2]` is not defined.
Maybe instead of using `AMREX_D_DECL`, we should do a loop over `dim-1` elements.

Source/ablastr/math/fft/WrapRocFFT.cpp
```diff
@@ -66,6 +66,69 @@ namespace ablastr::math::anyfft
...
+                                    ? rocfft_transform_type_real_forward
+                                    : rocfft_transform_type_real_inverse,
+#ifdef AMREX_USE_FLOAT
+                                    rocfft_precision_single,
+#else
+                                    rocfft_precision_double,
+#endif
+                                    dim, lengths,
+                                    howmany, // number of transforms,
+                                    desc);
```
RemiLehe: What happens if we don't pass a `desc` at all?

Source/ablastr/math/fft/WrapRocFFT.cpp
```diff
@@ -66,6 +66,69 @@ namespace ablastr::math::anyfft
...
+        const std::size_t roc_istride[] = {AMREX_D_DECL( (dir == direction::R2C) ? std::size_t(istride * real_size[0]) : std::size_t(istride * (real_size[0]/2+1)),
+                                                          std::size_t(istride * real_size[1]),
+                                                          std::size_t(istride * real_size[2]))};
+
+        const std::size_t roc_ostride[] = {AMREX_D_DECL( (dir == direction::R2C) ? std::size_t(ostride * (real_size[0]/2+1)) : std::size_t(ostride * real_size[0]),
+                                                          std::size_t(ostride*real_size[1]),
+                                                          std::size_t(ostride*real_size[2]))};
+
+        rocfft_plan_description desc = nullptr;
+        rocfft_status result = rocfft_plan_description_set_data_layout(
```
RemiLehe: The documentation for rocFFT seems to indicate that `rocfft_plan_description_create` should be called first.

Source/FieldSolver/ElectrostaticSolvers/RelativisticExplicitES.H
aeriforme: No and yes, thanks for catching this!

Source/FieldSolver/ElectrostaticSolvers/LabFrameExplicitES.H
aeriforme: No and yes, thank you!

Regression/Checksum/benchmarks_json/test_3d_open_bc_poisson_solver_heffte.json
aeriforme: I thought so too, but there isn't one in development right now.
Something must have happened when we moved to `ctest`, maybe?
Anyway, the main problem here is that the `heffte` test never runs in CI, so that's why we didn't notice.

aeriforme: Actually, from [here](https://github.com/ECP-WarpX/WarpX/pull/4937/files#diff-9d054a23083a17396f4b4a929f708e44643af3b2e24b492b15e1f2047138426a) it seems like that the checksums were saved in the wrong file, overwriting the ones for the test without `heffte`. 

That also explains why I had to change `Regression/Checksum/benchmarks_json/test_3d_open_bc_poisson_solver.json` too.

Source/ablastr/fields/IntegratedGreenFunctionSolver.H
RemiLehe: We could cite equation 58 in:
https://www.sciencedirect.com/science/article/pii/S0021999104000282

cemitch99: You can refer to equation (58) in https://doi.org/10.1016/j.jcp.2004.01.008

Source/ablastr/math/fft/WrapRocFFT.cpp
aeriforme: No need of the plan description anymore. 
However, the rocFFT wrapper is currently less flexible than that of FFTW and cuFFT.

Source/ablastr/math/fft/WrapRocFFT.cpp
aeriforme: No need of the plan description anymore.
However, the rocFFT wrapper is currently less flexible than that of FFTW and cuFFT.

Source/ablastr/fields/IntegratedGreenFunctionSolver.cpp
aeriforme: We could, but I wanted to keep the wrapper as flexible and general as possible.
What if one day anyone will need batched ffts that use those additional parameters?
That was my idea, but I can make the `CreatePlanMany` less flexible, if you think it's best.

Source/ablastr/fields/IntegratedGreenFunctionSolver.cpp
aeriforme: Yes, but it would be inside a loop over `i,j,k`. 
I thought that it would be better to define it beforehand, just once, since it does not depend on either `i`, `j` or `k`.

Source/ablastr/fields/IntegratedGreenFunctionSolver.cpp
aeriforme: I answered [here](https://github.com/ECP-WarpX/WarpX/pull/5089#discussion_r1826383099).

Source/ablastr/fields/IntegratedGreenFunctionSolver.cpp
aeriforme: See [this](https://github.com/ECP-WarpX/WarpX/pull/5089#discussion_r1826383099) comment.

Source/ablastr/fields/IntegratedGreenFunctionSolver.cpp
aeriforme: What about increasing verbosity? 
```
// Normalize, since (FFT + inverse FFT) results in a factor N
// Note that we use nrx*nry instead of realspace_box.numPts() 
// because here FFTs are in 2D along x and y
```

Source/ablastr/fields/IntegratedGreenFunctionSolver.cpp
aeriforme: How about: 
```
// The 2D slice solver performes, for every z, a 2D transform in the transverse plane (x,y).
// Hence, we use AMREX_SPACEDIM-1=2 and not AMREX_SPACEDIM=3 in the batched FFT calculations.
```
?

Docs/source/usage/parameters.rst
```diff
@@ -290,6 +290,19 @@ Overall simulation parameters
...
+            If 1, solve Poisson equation in a quasi 3D geometry, neglecting the :math:`z` derivatives in the Laplacian of the Poisson equation.
+            In practice, in this case, the code performes many 2D Poisson solves on all :math:`(x,y)` slices, each slice at a given :math:`z`.
+            This is often a good approximation for ultra-relativistic beams propagating along the :math:`z` direction, with the relativistic solver.
+            As a consequence, this solver does not need to do an FFT along the :math:`z` direction,
+            and instead uses only transverse FFTs (along :math:`x` and :math:`y`) at each :math:`z` position (or :math:`z` "slice").
+
+          * warpx.use_distributed_3d_fft_solver (`bool`, default: 0): Choose whether the 3D FFTs performed in the
+            full 3D Integrated Green Function solver are distributed.
+            If 0, the FFTs are performed on a single MPI rank (the rest of the code is still fully parallel).
+            If 1, the FFTs are distributed using the heFFTe library. The code must be compiled with `-DWarpX_HEFFTE=ON`.
```
RemiLehe, line 304: This last line should be updated, since we are not using heFFTe anymore.

Docs/source/usage/parameters.rst
```diff
@@ -290,6 +290,19 @@ Overall simulation parameters
         In electromagnetic mode, this solver can be used to initialize the species' self fields
         (``<species_name>.initialize_self_fields=1``) provided that the field BCs are PML (``boundary.field_lo,hi = PML``).
 
+          * warpx.use_2d_slices_fft_solver (`bool`, default: 0): Select the type of Integrated Green Function solver.
```
```suggestion
          * ``warpx.use_2d_slices_fft_solver`` (`bool`, default: 0): Select the type of Integrated Green Function solver.
```

Regression/Checksum/benchmarks_json/test_3d_open_bc_poisson_solver_heffte.json
RemiLehe: Could you remove this `.json` file? It seems that it is no longer needed/used.

Docs/source/usage/parameters.rst
RemiLehe: Could you remove this? It seems that it is not actually used in the code.

Source/ablastr/fields/IntegratedGreenFunctionSolver.H
```diff
@@ -30,54 +31,92 @@ namespace ablastr::fields
...
+                              - 0.5_rt * y*y * std::atan( x*z/(y*r) )
+                              - 0.5_rt * x*x * std::atan( y*z/(x*r) )
+                              + y*z*std::asinh( x/std::sqrt(y*y + z*z) )
+                              + x*z*std::asinh( y/std::sqrt(x*x + z*z) )
+                              + x*y*std::asinh( z/std::sqrt(x*x + y*y) );
+        return G;
+    }
+
+
+    /** @brief Implements minus equation 58 in https://doi.org/10.1016/j.jcp.2004.01.008
```
RemiLehe:
```suggestion
    /** @brief Implements equation 58 in https://doi.org/10.1016/j.jcp.2004.01.008
```
I think I understand what is meant here, but the term "minus equation" might be confusing. I think it is fine to simply say "equation 58".

Source/ablastr/fields/IntegratedGreenFunctionSolver.H
```diff
@@ -90,7 +129,8 @@ namespace ablastr::fields
     computePhiIGF (amrex::MultiFab const & rho,
                    amrex::MultiFab & phi,
                    std::array<amrex::Real, 3> const & cell_size,
-                   amrex::BoxArray const & ba);
+                   amrex::BoxArray const & ba,
+                   bool do_2d_slices);
```
RemiLehe: Would it make sense to use the same name (`is_igf_2d_slices`) as in the other functions - instead of `do_2d_slices`?

RemiLehe (changes requested):
Thanks for this PR! I left a few comments.

Source/ablastr/fields/IntegratedGreenFunctionSolver.H
aeriforme: Yes, sure!

RemiLehe (approved):

Source/ablastr/fields/IntegratedGreenFunctionSolver.cpp
```diff
@@ -71,7 +69,9 @@ computePhiIGF ( amrex::MultiFab const & rho,
     amrex::Real const dy = cell_size[1];
     amrex::Real const dz = cell_size[2];
 
-    obc_solver->setGreensFunction(
+    if (!is_igf_2d_slices){
+        // 2D sliced solver
```
WeiqunZhang, line 73: This comment here and the one in the else part are incorrect.

cemitch99 (approved):

Source/ablastr/fields/IntegratedGreenFunctionSolver.cpp
aeriforme: oops! I will fix it

PR #5091, opened 2024-07-27
Improving Coulomb collision method for weighted-particles 

JustinRayAngus:
Separating part 2 of PR #5047 into two parts. One that fixes the weighted scattering issue (this PR) and another that improves on the Perez 2012 method for relativistic Coulomb scattering.

Description edited by @RemiLehe:

The current method for doing weighted-particle scattering for Coulomb collisions is from Perez et al 2012 (https://doi.org/10.1063/1.4742167) which follows from Nanbu and Yonemura 1998 (http://dx.doi.org/10.1006/jcph.1998.6049). That method has been shown to result in incorrect relaxation rates for certain problems by Higginson et al 2020 (https://doi.org/10.1016/j.jcp.2020.109450). The present PR fixes this issue using a method similar to that by Higginson 2020 that produces correct relaxation rates with weighted particles. Below are results from Tests T1a-T1d from Higginson 2020 (but with 10X less particles). The physical setup is the same for each of tests T1a-T1d, but they use different particle weighting setups for the different populations. In contrast to results obtained using the development branch, the results obtained using this PR show identical relaxation results for the various particle weight combinations.

Using this PR
T1abcd_binaryOpt_branch

Using the development branch:
T1abcd_develop_branch_new

### DIFF
--- a/Examples/Tests/collision/analysis_collision_1d.py
+++ b/Examples/Tests/collision/analysis_collision_1d.py
@@ -1,0 +1,126 @@
+#!/usr/bin/env python3
+
+# Copyright 2024 Justin Angus
+#
+#
+# This file is part of WarpX.
+#
+# License: BSD-3-Clause-LBNL
+#
+# This is a script that analyses the simulation results from the script `inputs_1d`.
+# run locally: python analysis_vandb_1d.py diags/diag1000600/
+#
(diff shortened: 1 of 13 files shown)


RemiLehe (approved):

RemiLehe:
Thanks a lot for this PR!

PR #5096, opened 2024-07-29
Enable tiling in some `MFIter` loops

EZoni:
Based on @WeiqunZhang's [comment](https://github.com/ECP-WarpX/WarpX/pull/4922#discussion_r1694338972) in #4922,

> I think it's better to use `growntilebox` and add `TilingIfNotGPU()` for the E field loop. For GPU, the two loops are the same. But for CPU, the tiling version should be better due to cache blocking and more OpenMP parallelism.

I think there are a couple of loops where we can enable tiling, by using `TilingIfNotGPU` when we set the `MFIter` iterator and `growntilebox` in place of `fabbox` for the subsequent `ParallelFor` loops.


lucafedeli88 (approved):
It looks good to me! Do you know which is the impact on performances of this kernel?

EZoni:
> It looks good to me! Do you know which is the impact on performances of this kernel?

Hi @lucafedeli88, I don't know the performance impact of this kernel. What triggered this PR was not a performance issue but simply the comment reported above, which suggested that, whenever possible, it is preferred to enable tiling in order to have better performance on CPU, in general. Since it is common to take existing code as example/prototype when implementing new code, I think it is easy for not-optimal implementations (in this case, loops without tiling rather than with tiling) to propagate through the code due to simple "copy/paste". So I figured it would be better to follow the preferred implementation whenever possible.

PR #5100, opened 2024-07-30
Expose more diagnostics options in PICMI

RemiLehe:
This exposes additional options in PICMI:
- `openpmd_encoding`
- `dump_last_timestep`


ax3l (approved):
Thank you: :tada:

PR #5101, opened 2024-07-30
PICMI interface for implicit solvers

dpgrote:
This adds a PICMI interface to define and specify the input parameters for the implicit solvers. See the example `Examples/Tests/Implicit/PICMI_inputs_vandb_jfnk_2d.py`.

Note that this PR also adds a new way in the PICMI implementation to specify groups of input parameters. This creates the new method `warpx.get_bucket` that generates and returns a `bucket` for setting input parameters. Then during the initialization phase, the `warpx` instance then loops over the defined buckets to fetch the input parameters. This is done instead of creating a file for each group (for example `Python/pywarpx/HybridPICModel.py`). Using this method will simplify the PICMI interface structure and also clean up the top level `pywarpx` name space (there would no long be the group instances). Note that a number of the existing parameter groups could be converted to this method (in later PRs).


RemiLehe:
Thanks a lot for this PR!
Could you add the new Python classes that you introduced to [the WarpX ReadTheDocs documentation?](https://warpx.readthedocs.io/en/latest/usage/python.html)

Examples/Tests/Implicit/PICMI_inputs_vandb_jfnk_2d.py
```diff
@@ -0,0 +1,149 @@
...
+                                    grid = grid,
+                                    period = diagnostic_intervals,
+                                    data_list = ['Ex', 'Ey', 'Ez', 'Bx', 'By', 'Bz', 'Jx', 'Jy', 'Jz', 'rho', 'divE'],
+                                    write_dir = '.',
+                                    warpx_file_prefix = 'ThetaImplicitJFNK_VandB_2d_PICMI')
+
+part_diag1 = picmi.ParticleDiagnostic(name = 'diag1',
+                                      period = diagnostic_intervals,
+                                      species = [electrons, protons],
+                                      data_list = ['weighting', 'position', 'momentum'])
```
ax3l: Does this diag also have to list `write_dir` and `warpx_file_prefix`?

Examples/Tests/Implicit/PICMI_inputs_vandb_jfnk_2d.py
```diff
@@ -0,0 +1,149 @@
...
+##########################
+# diagnostics
+##########################
+
+field_diag1 = picmi.FieldDiagnostic(name = 'diag1',
+                                    grid = grid,
+                                    period = diagnostic_intervals,
+                                    data_list = ['Ex', 'Ey', 'Ez', 'Bx', 'By', 'Bz', 'Jx', 'Jy', 'Jz', 'rho', 'divE'],
+                                    write_dir = '.',
+                                    warpx_file_prefix = 'ThetaImplicitJFNK_VandB_2d_PICMI')
```
RemiLehe:
```suggestion
                                    warpx_file_prefix = 'ThetaImplicitJFNK_VandB_2d_PICMI_plt')
```

Examples/Tests/Implicit/PICMI_inputs_vandb_jfnk_2d.py
RemiLehe:
```suggestion
                                      data_list = ['weighting', 'position', 'momentum'],
                                      write_dir = '.',
                                      warpx_file_prefix = 'ThetaImplicitJFNK_VandB_2d_PICMI_plt')
```

RemiLehe (approved):

PR #5103, opened 2024-07-30
EB: Assert that its used

ax3l:
Currently, users MUST use EB if it was compiled in, otherwise several locations of the code show undefined behavior. This will change with https://github.com/ECP-WarpX/WarpX/pull/4865 .

First reported by @lucafedeli88.

This assert logic already uses the logic we will use in #4865 for compatibility.

### DIFF
--- a/Source/EmbeddedBoundary/Enabled.cpp
+++ b/Source/EmbeddedBoundary/Enabled.cpp
@@ -32,12 +32,6 @@ namespace EB
         std::string eb_stl;
         eb_enabled |= pp_eb2.query("geom_type", eb_stl);
 
-#if defined(WARPX_DIM_RZ)
-        if (eb_enabled) {
-            throw std::runtime_error("RZ Geometry does not yet support EBs, but EBs are enabled in runtime inputs.");
-        }
-#endif
-
         return eb_enabled;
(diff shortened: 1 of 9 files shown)


ax3l:
Found with @RemiLehe: PR #3637 added the test that fails: We tried to ensure the code runs when EB is compiled but not running. There is more to do, so we will remove this test now.

Source/EmbeddedBoundary/Enabled.cpp
```diff
@@ -0,0 +1,45 @@
...
+        // https://amrex-codes.github.io/amrex/docs_html/EB.html
+        std::string eb_stl;
+        eb_enabled |= pp_eb2.query("geom_type", eb_stl);
+
+#if defined(WARPX_DIM_RZ)
+        if (eb_enabled) {
+            throw std::runtime_error("RZ Geometry does not yet support EBs, but EBs are enabled in runtime inputs.");
+        }
+#endif
+
```
ax3l: @RemiLehe told me on Slack that we have RZ support for EBs with test in ES and EM. So this can go :tada: 
```suggestion
```

RemiLehe (approved):

PR #4547, opened 2023-12-20
Read external fields from file in restarted simulation

RemiLehe:
When restarting a simulation, the grid external field from file were not read. 
As a result, a restarted simulation (e.g. with the electrostatic solver, were the grid external fields are applied at every iteration) would suddenly not feature these fields (without any error or warning message).

- [x] Cleanup code: maybe make separate function to avoid code duplication
- [x] Add a docstring to the new function
- [x] Documentation: explicitly state that the grid external fields are applied at every iteration, with the electrostatic solver
- [x] Add a restart automated test, based on https://github.com/ECP-WarpX/WarpX/tree/development/Examples/Tests/LoadExternalField

PR #5104, opened 2024-07-31
Treat external field types consistently

roelof-groenewald:
This PR fixes what seems to be a bug in the way external fields are handled - external fields specified as constants or through parsable expressions are handled differently from external fields specified with a file. In particular for electrostatic simulations the external fields are added back on to the `fp` fields when they are specified from a file but not when they are set by constants or from the parser. The changes in this PR updates the code to treat all manners of specifying the external fields equivalently.

### DIFF
--- a/Source/Evolve/WarpXEvolve.cpp
+++ b/Source/Evolve/WarpXEvolve.cpp
@@ -248,7 +248,7 @@ WarpX::Evolve (int numsteps)
                 }
                 // Since the fields were reset above, the external fields are added
                 // back on to the fine patch fields. This make it so that the net fields
-                // are the composite of the field solution and any external field.
+                // are the sum of the field solution and any external field.
                 for (int lev = 0; lev <= max_level; ++lev) {
                     AddExternalFields(lev);
                 }
--- a/Source/Evolve/WarpXEvolve.cpp
+++ b/Source/Evolve/WarpXEvolve.cpp
@@ -249,7 +249,9 @@ WarpX::Evolve (int numsteps)
                 // Since the fields were reset above, the external fields are added
                 // back on to the fine patch fields. This make it so that the net fields
                 // are the composite of the field solution and any external field.
-                AddExternalFields();
+                for (int lev = 0; lev <= max_level; ++lev) {
+                    AddExternalFields(lev);
+                }
             } else if (electromagnetic_solver_id == ElectromagneticSolverAlgo::HybridPIC) {
                 // Hybrid-PIC case:
                 // The particles are now at p^{n+1/2} and x^{n+1}. The fields
--- a/Source/Evolve/WarpXEvolve.cpp
+++ b/Source/Evolve/WarpXEvolve.cpp
@@ -246,6 +246,9 @@ WarpX::Evolve (int numsteps)
                     // This is currently a lab frame calculation.
                     ComputeMagnetostaticField();
                 }
+                // Since the fields were reset above, the external fields are added
+                // back on to the fine patch fields. This make it so that the net fields
+                // are the composite of the field solution and any external field.
                 AddExternalFields();
             } else if (electromagnetic_solver_id == ElectromagneticSolverAlgo::HybridPIC) {
                 // Hybrid-PIC case:
--- a/Source/Initialization/WarpXInitData.cpp
+++ b/Source/Initialization/WarpXInitData.cpp
@@ -533,32 +533,30 @@ WarpX::InitData ()
 }
 
 void
-WarpX::AddExternalFields () {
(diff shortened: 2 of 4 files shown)


dpgrote:
This looks Ok, but I have a question. If you have an external field that is a constant or an expression, why not apply it directly to the particles using `particles.E_ext_particle_init_style`?

roelof-groenewald:
> This looks Ok, but I have a question. If you have an external field that is a constant or an expression, why not apply it directly to the particles using `particles.E_ext_particle_init_style`?

One reason we have seen in the past is that since the particle fields are not written out it is hard to confirm that the fields being applied are the desired ones.  With this fix an alternative would be to use the external grid fields, which can be visualized and inspected. This could even just be done in a test of the simulation setup and once the fields are confirmed to be correct the input could be switched back to the particle fields depending on the problem at hand. Nonetheless, since we have the option to also pass constants or expressions for the external grid field, we should be consistent in how those fields are handled.

Another thing I will note is that there is some confusion that comes up sporadically with the "external" fields overall which should be kept in mind too. The particle fields are applied to the particles at every step (and can have `t` dependence). The external grid fields instead function as initial conditions for the fields and therefore evolves based on the field solver (it might be worth changing the input parameter names in a subsequent PR but that would break input scripts). This distinction is particularly important for the Ohm's law based solver where the B-field is used to calculate `J` meaning that different sources of electromagnetic fields cannot be cleanly separated as in Maxwell's equations, or in other words, the external particle fields should not be used with the Ohm's law based solver.

clarkse:
@roelof-groenewald Would we also get rid of the read_from_file type? In general, I am in favor of this treatment as I thought before it was strange that the constant or parsed external fields weren't treated the same as gridded. I assumed it was a memory issue to try and reduce the memory footprint when not reading in external gridded data. It would also make the exposure of the external fields though PICMI more useful for diagnosing issues with the other methods of invoking external fields.

RemiLehe:
Just checking: does this PR also make sure that the external fields defined as constant or a parser are loaded into the external MultiFab during a restart, similar to what this PR does for fields from file:
https://github.com/ECP-WarpX/WarpX/pull/4547

roelof-groenewald:
> Would we also get rid of the read_from_file type? 

I think we still need to keep that type to know when to call `InitializeExternalFieldsOnGridUsingParser` versus `ReadExternalFieldFromFile`. Or do you have something in mind to get around that?

roelof-groenewald:
@RemiLehe I updated `LoadExternalFieldsFromFile` (renamed to `LoadExternalFields`) to also read the parser fields into the external fields multifabs so in all cases a restart should now give the proper fields.

clarkse:
> > Would we also get rid of the read_from_file type?
> 
> I think we still need to keep that type to know when to call `InitializeExternalFieldsOnGridUsingParser` versus `ReadExternalFieldFromFile`. Or do you have something in mind to get around that?

It appears that the particle external fields are controlled via the string passed in through the parser instead of an enumerated type. I would vote for adding an enum type for the particle fields instead of a string compare, but I suppose it doesn't really matter.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -246,7 +246,12 @@ WarpX::Evolve (int numsteps)
                     // This is currently a lab frame calculation.
                     ComputeMagnetostaticField();
                 }
-                AddExternalFields();
+                // Since the fields were reset above, the external fields are added
+                // back on to the fine patch fields. This make it so that the net fields
+                // are the composite of the field solution and any external field.
```
RemiLehe: Would `sum` be a less ambiguous term?
```suggestion
                // are the sum of the field solution and any external field.
```

Source/Initialization/WarpXInitData.cpp
```diff
@@ -906,19 +895,6 @@ WarpX::InitLevelData (int lev, Real /*time*/)
     if ((m_p_ext_field_params->E_ext_grid_type == ExternalFieldType::parse_ext_grid_function)
         && (lev <= maxlevel_extEMfield_init)) {
```
RemiLehe, line 896: This part of the code (for `E`) does not seem to be doing the same thing as the above code (for `B`), or did I miss something?

roelof-groenewald: The difference is just that the external E-field is used to update `ECTRhofield` (if EB's are ON and the ECT solver is used). But the B-field does not have an equivalent field that needs to be updated.

Source/Evolve/WarpXEvolve.cpp
roelof-groenewald: Sure :)

RemiLehe (approved):

PR #1875, opened 2021-04-07
[WIP] run_test.sh: mass-reset benchmarks

ax3l:
Resetting benchmarks is too manual to be enjoyable. Workflow:
- Manual re-compile for 2D, 3D, RZ, SP, ...
- Fiddle out overwritten compile in .ini file
- Fiddle out OMP/MPI options from .ini file
- Fiddle out overwritten runtime options from .ini file over inputs file
- Run
- Reset
- Commit
- Repeat for all compile combinations & run tests that are affected

This adds functionality to reset a couple of tests via:

```bash
CHECKSUM_RESET=TRUE ./run_test.sh test1 test2 test3
```

PR #5068, opened 2024-07-20
Set up CTest

EZoni:
Work-in-progress PR to set up CTest. Apply the same logic as in ImpactX and adjust/advance for WarpX.

- Example of current input files organization:
  - `inputs_base_2d`
    (base input file for shared input parameters named as `inputs_base_<1d;2d;3d;rz>`)
  - `inputs_test_2d_langmuir_multi_nodal`,
    `inputs_test_2d_langmuir_multi_psatd`,
    ...
    (test input files for test-specific input parameters named as `inputs_test_<1d;2d;3d;rz>_...`, read base input file through the `FILE` parameter)
  - test names must start with `test_<1d;2d;3d;rz>_...` to guarantee consistency and make maintenance easier

- New Azure CI matrix elements (list coming soon).

- Run all MPI tests with 2 MPI processes and 1 OpenMP thread. Run single-process tests without MPI.

To-do:
- [x] `WarpX_TEST_CLEANUP` to remove test directories
- [x] `WarpX_TEST_FPETRAP` for runtime fpe-trapping
- [x] Resolve `FIXME` comments or redirect to new issues
- [x] Avoid copying inputs files https://github.com/AMReX-Codes/amrex/pull/4126
- [x] Enforce new naming convention through CI style check
- [x] Remove obsolete CI style check scripts
- [x] Remove obsolete regression testing scripts: #5220
- [x] Update docs
- [x] Remove files after each test: new build flag `WarpX_CI`
- [x] Make sure we don't miss any build variant:
  - [x] single-process Python
  - [x] embedded boundaries
  - single- & mixed-precision (new issue #5205)
- [x] Add missing checksums (new issue #5206):
  - `dirichlet_bc`
  - `dirichlet_bc_picmi`
  - `field_probe`
  - `embedded_boundary_picmi`
  - `particle_absorption`
  - `particle_reflection_picmi`
  - [x] `id_cpu_read_picmi`
  - `runtime_components_picmi`
  - `laser_acceleration_rz_opmd`
- [x] Support restart tests
  - [x] Separate input file based on original input file (read through the `FILE` parameter)
  - [x] Call default regression analysis against original checksum file, with small tolerance
  - [x] Support restart from PICMI (can we set `FILE` through PICMI?)
  - [x] Update `analysis_default_restart.py`
- [x] Support custom run commands (e.g., custom PICMI scripts using lists vs strings)
- Add stricter dependency between `.analysis` test and `.run` test:
[...]

PR #5105, opened 2024-07-31
Checksum Reset via Env Variable

ax3l:
In preparation for CTest adoption, we want to make it easier to quickly run and reset checksums locally. This adds the feature to reset instead of evaluate in regular test runs.

- [x] tested: This will work with CTest (#5068) but does not work with the old `run_tests.sh` logic, because of an extra copy of the whole source directory

Close #1875

### DIFF
--- a/Docs/source/developers/checksum.rst
+++ b/Docs/source/developers/checksum.rst
@@ -68,7 +68,7 @@ Since this will automatically change the JSON file stored on the repo, make a se
 Automated reset of a list of test benchmarks
 --------------------------------------------
 
-You can set the environment variable ``export CHECKSUM_RESET=ON`` before running analysis scripts / test that verify checksums - in that case, the analysis will instead reset to the new values.
+If you set the environment variable ``export CHECKSUM_RESET=ON`` before running tests that are compared against existing benchmarks, the test analysis will reset the benchmarks to the new values, skipping the comparison.
 
 With `CTest <https://cmake.org/cmake/help/latest/manual/ctest.1.html>`__ (coming soon), select the test(s) to reset by `name <https://cmake.org/cmake/help/latest/manual/ctest.1.html#run-tests>`__ or `label <https://cmake.org/cmake/help/latest/manual/ctest.1.html#label-matching>`__.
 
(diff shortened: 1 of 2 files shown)


Regression/Checksum/checksumAPI.py
```diff
@@ -42,6 +42,8 @@ def evaluate_checksum(test_name, output_file, output_format='plotfile', rtol=1.e
     Read checksum from output file, read benchmark
     corresponding to test_name, and assert their equality.
 
+    If the environment variable CHECKSUM_RESET is set while ...
```
ax3l: To complete.

Regression/Checksum/checksumAPI.py
```diff
@@ -65,9 +67,16 @@ def evaluate_checksum(test_name, output_file, output_format='plotfile', rtol=1.e
...
     """
-    test_checksum = Checksum(test_name, output_file, output_format,
-                             do_fields=do_fields, do_particles=do_particles)
-    test_checksum.evaluate(rtol=rtol, atol=atol)
+    # Reset benchmark?
+    reset = ( os.getenv('CHECKSUM_RESET', 'False').lower() in
+              ['true', '1', 't', 'y', 'yes', 'on'] )
+
+    if reset:
+        reset_benchmark(test_name, output_file, output_format, do_fields, do_particles)
```
ax3l, line 77: Maybe could print a line here that `CHECKSUM_RESET` is set, too.

Docs/source/developers/checksum.rst
```diff
@@ -65,6 +65,21 @@ Since this will automatically change the JSON file stored on the repo, make a se
    git add <test name>.json
    git commit -m "reset benchmark for <test name> because ..." --author="Tools <warpx@lbl.gov>"
 
+Automated reset of a list of test benchmarks
+--------------------------------------------
+
+You can set the environment variable ``export CHECKSUM_RESET=ON`` before running analysis scripts / test that verify checksums - in that case, the analysis will instead reset to the new values.
```
EZoni: Possible alternative phrasing:
```suggestion
If you set the environment variable ``export CHECKSUM_RESET=ON`` before running tests that are compared against existing benchmarks, the test analysis will reset the benchmarks to the new values, skipping the comparison.
```

EZoni (approved):

PR #5106, opened 2024-07-31
ABLASTR: Pass `GridType` instead of Global

ax3l:
We cannot use global variables from WarpX in the ABLASTR library, otherwise it will not build on its own or in ImpactX.

Follow-up to #4922


Source/ablastr/utils/Enums.H
```diff
@@ -0,0 +1,38 @@
...
+    enum struct GridType {
+        Collocated = 0,
+        Staggered = 1,
+        Hybrid = 2
+    };
+
+    /** Mesh-refinement patch
+     *
+     * The current (fine) or lower-level (coarse) resolution.
+     */
```
ax3l: @EZoni @RemiLehe @oshapoval pls let me know if these doc strings are reasonable or if you would change something :pray:

Source/ablastr/utils/Enums.H
```diff
@@ -0,0 +1,34 @@
...
+    enum struct GridType {
+        Collocated = 0,
+        Staggered = 1,
+        Hybrid = 2
+    };
+
+    /** Mesh-refinement patch
+     *
+     * The current (fine) or lower-level (coarse) resolution.
```
RemiLehe:
```suggestion
     * The fine or coarse patch (in terms of spatial resolution) on the same MR level.
     * https://warpx.readthedocs.io/en/latest/theory/amr.html
```

RemiLehe (approved):

EZoni (commented):
Looks good to me, thanks! Can you just confirm that moving `PatchType` to ABLASTR does not require any additional changes (wherever `PatchType` is used in the code) at this time?

ax3l:
Correct, I moved it over because it categorically is general enough for ABLASTR when introducing this new Enums header. I added a `using namespace` in WarpX to avoid any changes at this point caused by it

EZoni (approved):

PR #5108, opened 2024-08-04
Add density min and max arguments to `picmi.AnalyticDistribution`

roelof-groenewald:
The arguments `<species>.density_min` and `<species>.density_max` was added to avoid zero weight particles from being loaded in regions where the analytic density function is zero. This PR adds those arguments to `picmi.AnalyticDistribution`.
Note that this functionality is used in the new example from #5079.


dpgrote (approved):
This looks good, thanks!

These arguments could be added to the picmi standard.

PR #5109, opened 2024-08-06
Add scraping boundary diagnostics with PICMI

oshapoval:
Description to be added. 
Depend on PICMI branch: `add_boundary_scraping_diags_picmi.`


EZoni:
This needs to be rebased again on `development` after the latest CI fix.

RemiLehe (approved):

Examples/Physics_applications/spacecraft_charging/PICMI_inputs_rz.py
```diff
@@ -275,6 +275,15 @@ def compute_actual_charge_on_spacecraft():
     warpx_file_prefix="spacecraft_charging_plt",
 )
 
+part_scraping_boundary_diag = picmi.ParticleBoundaryScrapingDiagnostic(
+    name="diag2",
+    period=diagnostic_interval,
+    species=[electrons, protons],
+    warpx_format="openpmd",
+    write_dir=".",
+    warpx_file_prefix="spacecraft_charging_plt",
```
RemiLehe, line 284:
```suggestion
```

RemiLehe:
```suggestion
```

Examples/Physics_applications/spacecraft_charging/PICMI_inputs_rz.py
```diff
@@ -275,6 +275,13 @@ def compute_actual_charge_on_spacecraft():
     warpx_file_prefix="spacecraft_charging_plt",
 )
 
+part_scraping_boundary_diag = picmi.ParticleBoundaryScrapingDiagnostic(
```
ax3l, line 278: Comment by @roelof-groenewald:
Careful, the sim here uses the boundary buffer to accumulate but the diag cleans out the buffer on dump.

Could either
- use another example or
- only do the dump at the end or
- add an option to not clean out the buffer on dump (should _not_ be default-on)

ax3l: @RemiLehe suggests adding this in the magnetostatic w/ boundary example.
Maybe we can also add a warning when a situation like the one here is detected? At least we need to document clearly the cleaning logic in `ParticleBoundaryScrapingDiagnostic`.

Python/pywarpx/picmi.py
```diff
@@ -3968,3 +3968,178 @@ def diagnostic_initialize_inputs(self):
...
                 else:
                     self.diagnostic.__setattr__(key, value)
+
+
+class ParticleBoundaryScrapingDiagnostic(
+    picmistandard.PICMI_ParticleBoundaryScrapingDiagnostic, WarpXDiagnosticBase
+):
+    """
+    See `Input Parameters <https://warpx.readthedocs.io/en/latest/usage/parameters.html>`__ for more information.
+
```
ax3l, line 3979: Let's add a note here, that this diags cleanes out the particle boundary buffer.

Examples/Physics_applications/spacecraft_charging/PICMI_inputs_rz.py
```diff
@@ -275,6 +275,13 @@ def compute_actual_charge_on_spacecraft():
...
+    name="diag2",
+    period=diagnostic_interval,
```
ax3l:
```suggestion
    period=-1,  # only at the end, because we also use the buffers in this test
```

PR #5112, opened 2024-08-06
AnyFFT: 1D Support

ax3l:
Add 1D support for AnyFFT to FFTW and CuFFT. RocFFT already was written in a generic way.

Needed for https://github.com/ECP-WarpX/impactx/pull/627

### DIFF
--- a/Source/ablastr/math/fft/WrapFFTW.cpp
+++ b/Source/ablastr/math/fft/WrapFFTW.cpp
@@ -25,11 +25,15 @@ namespace ablastr::math::anyfft
     const auto VendorCreatePlanC2R3D = fftwf_plan_dft_c2r_3d;
     const auto VendorCreatePlanR2C2D = fftwf_plan_dft_r2c_2d;
     const auto VendorCreatePlanC2R2D = fftwf_plan_dft_c2r_2d;
+    const auto VendorCreatePlanR2C1D = fftwf_plan_dft_r2c_1d;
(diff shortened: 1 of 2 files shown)


Source/ablastr/math/fft/WrapFFTW.cpp
```diff
@@ -67,9 +74,12 @@ namespace ablastr::math::anyfft
             } else if (dim == 2) {
                 fft_plan.m_plan = VendorCreatePlanC2R2D(
                     real_size[1], real_size[0], complex_array, real_array, FFTW_ESTIMATE);
+            } else if (dim == 1) {
+                fft_plan.m_plan = VendorCreatePlanC2R1D(
+                    real_size[0], complex_array, real_array, FFTW_ESTIMATE);
             } else {
                 ABLASTR_ABORT_WITH_MESSAGE(
-                    "only dim=2 and dim=3 have been implemented. Should be easy to add dim=1.");
+                    "only dim=1 and dim=2 and dim=3 have been implemented.");
```
EZoni, line 82: Do we want to keep this general for possible future scenarios with `dim` > 3? If not, maybe we could also remove these last `else` blocks.

ax3l: If we remove the last `else` block, then we will silently do nothing on other dims than 1,2,3 passed by a user (could be by accident). Thus, I would keep it to explicitly handle all cases.

EZoni (approved):

PR #5115, opened 2024-08-07
Use only openPMD output for openbc_poisson_solver test

RemiLehe:
The aim of this PR is to bypass the current CI failure with plotfiles for the openbc_poisson test.


RemiLehe:
While `openbc_poisson_solver` does not fail anymore with this PR, other tests started failing, with disk access errors as well. 

This shows that the error is not specific to `openbc_poisson_solver`. We should revert https://github.com/ECP-WarpX/WarpX/pull/5117

ax3l:
@RemiLehe feel free to rebase this, using openPMD for the test is a good idea :)

PR #5116, opened 2024-08-07
[WIP] Revert to previous version of lasy

RemiLehe:
Another attempt to understand what caused the recent file-system CI failures.

### DIFF
--- a/Regression/requirements.txt
+++ b/Regression/requirements.txt
@@ -1,5 +1,5 @@
 dill
-lasy
+lasy==0.4.0
 matplotlib
 mpi4py
 numpy


ax3l:
I think this should be good to be closed. Only `openbc_poisson_solver` continues to fail, which is a dormant bug (UB) that now shows with a change in the underlying CI base system.

RemiLehe:
Conclusion: `openbc_poisson_solver` still fails with this PR.
This means that the new `lasy` release on Monday was **not** the reason why `openbc_poisson_solver` (and later other tests) started failing at the beginning of this week.

PR #5118, opened 2024-08-07
Function to get particle cell index.

### LINKED ISSUE #4977: Function to get particle cell index.
It would be useful to have a function (maybe in `ParticleUtils.cpp`) that returns the cell indices for a given particle. This would allow cleaning up code in several places (for example in `TemperateFunctor.cpp`.

_Originally posted by @roelof-groenewald in https://github.com/ECP-WarpX/WarpX/pull/4942#discussion_r1631548274_

archermarx:
Close #4977

This PR adds a function to get a particle's cell index to `ParticleUtils.H`, and uses it in both `ParticleUtils.cpp` and `TemperatureFunctor.cpp`.


ax3l:
Thanks for the PR, @archermarx!
We just mitigated a CI issue, can you rebase against the latest `development` and force push (or merge `development`)? Thanks!

archermarx:
yes, done!

ax3l (approved):
That looks splendid, thank you so much for this!

archermarx:
Double checking, I had to cast a few parameters to `void` in order to avoid unused variable warnings due to y and/or z positions not being needed in 1D and 2D simulations. Is this ok, or is there a cleaner way to do this?

Source/Diagnostics/ComputeDiagFunctors/TemperatureFunctor.cpp
```diff
@@ -53,22 +54,10 @@ TemperatureFunctor::operator() (amrex::MultiFab& mf_dst, const int dcomp, const
...
-                const amrex::ParticleReal y = p.pos(1);
-                const amrex::Real ly = (y - plo[1]) * dxi[1];
-                jj = static_cast<int>(amrex::Math::floor(ly));
-#endif
-#if defined(WARPX_DIM_3D)
-                const amrex::ParticleReal z = p.pos(2);
-                const amrex::Real lz = (z - plo[2]) * dxi[2];
-                kk = static_cast<int>(amrex::Math::floor(lz));
-#endif
+                const auto cellIndex = ParticleUtils::getParticleCellIndex(p.pos(0), p.pos(1), p.pos(2), plo, dxi, amrex::Dim3{0, 0, 0}).dim3();
```
roelof-groenewald: Here you could simply do the following, right?
```suggestion
                const auto cellIndex = ParticleUtils::getParticleCellIndex(p, plo, dxi, amrex::Dim3{0, 0, 0}).dim3();
```
Which is a bit cleaner.

Source/Utils/ParticleUtils.H
```diff
@@ -17,6 +17,58 @@
...
+     */
+    AMREX_GPU_HOST_DEVICE AMREX_INLINE
+    amrex::IntVect getParticleCellIndex (amrex::ParticleReal xp,
+                                         amrex::ParticleReal yp,
+                                         amrex::ParticleReal zp,
+                                         amrex::GpuArray<amrex::Real,AMREX_SPACEDIM> const& plo,
+                                         amrex::GpuArray<amrex::Real,AMREX_SPACEDIM> const& dxi,
+                                         amrex::Dim3 lo) {
+        int ii = 0, jj = 0, kk = 0;
+        (void)yp, (void)zp, (void)jj, (void)kk;
```
roelof-groenewald: Instead of casting to `void` we usually do
```suggestion
        amrex::ignore_unused(yp, zp, jj, kk);
```
or you can add `[[maybe_unused]]` before the variable definition, for example, `[[maybe_unused]] int jj`.

roelof-groenewald (commented):
Thanks for this cleanup! Just a couple small comments.

Source/Utils/ParticleUtils.H
ax3l: Yep, `[[maybe_unused]]` in the API is the modern way.
We added `amrex::ignore_unused` when we did not yet compile for C++17.

Source/Diagnostics/ComputeDiagFunctors/TemperatureFunctor.cpp
```diff
@@ -53,22 +54,10 @@ TemperatureFunctor::operator() (amrex::MultiFab& mf_dst, const int dcomp, const
...
+                int ii = cellIndex.x;
+                int jj = cellIndex.y;
+                int kk = cellIndex.z;
```
ax3l: Clang tidy recommends to declare those as `const int`.

Source/Diagnostics/ComputeDiagFunctors/TemperatureFunctor.cpp
```diff
@@ -119,18 +108,10 @@ TemperatureFunctor::operator() (amrex::MultiFab& mf_dst, const int dcomp, const
...
-#elif defined(WARPX_DIM_3D)
-                const amrex::Real ly = (yp - plo[1]) * dxi[1];
-                jj = static_cast<int>(amrex::Math::floor(ly));
-                const amrex::Real lz = (zp - plo[2]) * dxi[2];
-                kk = static_cast<int>(amrex::Math::floor(lz));
-#endif
+                const auto cellIndex = ParticleUtils::getParticleCellIndex(xp, yp, zp, plo, dxi, amrex::Dim3{0, 0, 0}).dim3();
+                int ii = cellIndex.x;
+                int jj = cellIndex.y;
+                int kk = cellIndex.z;
```
ax3l: As above: `const int`

Source/Diagnostics/ComputeDiagFunctors/TemperatureFunctor.cpp
archermarx: Here, there was an issue in that in `TemperatureFunctor`, `p` is a `WarpXParticleContainer::SuperParticleType` but the function expects `ParticleType`.  The opposite is true in `ParticleUtils.cpp`, where a `ParticleType` is used. I couldn't find definitions for these types, but clang warned me there was no way to cast between them.

Source/Utils/ParticleUtils.H
archermarx: Thanks! That's exactly what I was looking for.

Source/Diagnostics/ComputeDiagFunctors/TemperatureFunctor.cpp
roelof-groenewald: Oh, I see. Thanks for clarifying!

Source/Diagnostics/ComputeDiagFunctors/TemperatureFunctor.cpp
archermarx: fixed both

Source/Diagnostics/ComputeDiagFunctors/TemperatureFunctor.cpp
archermarx: I could either 1) remove the overload for `ParticleType` so we just pass in the coords in both locations or 2) add an overload for `SuperParticleType` if you think that would be valuable

roelof-groenewald: Adding another overload for `SuperParticleType` can't hurt, so I vote for that :)
Thanks!

archermarx:
Ok, I added an overload for the function when called on a `SuperParticleType` and incorporated it into `TemperatureFunctor.cpp`

Source/Utils/ParticleUtils.cpp
```diff
@@ -57,12 +57,9 @@ namespace ParticleUtils
...
         bins.build(np, ptd, cbx,
             // Pass lambda function that returns the cell index
-            [=] AMREX_GPU_DEVICE (ParticleType const & p) noexcept -> amrex::IntVect
+            [=] AMREX_GPU_DEVICE (ParticleType const & p) noexcept -> IntVect
             {
-                return IntVect{AMREX_D_DECL(
-                                   static_cast<int>((p.pos(0)-plo[0])*dxi[0] - lo.x),
-                                   static_cast<int>((p.pos(1)-plo[1])*dxi[1] - lo.y),
-                                   static_cast<int>((p.pos(2)-plo[2])*dxi[2] - lo.z))};
+                return getParticleCellIndex(p, plo, dxi, lo);
```
roelof-groenewald: It seems this change is causing some crashes in code that uses the particle binning. I'm thinking it is due to the dimension specific functionality in `getParticleCellIndex` that is absent here.
I think this can still be simplified though, using
```suggestion
                return getParticleCell(p, plo, dxi, cbx);
```
and then the `lo` variable would not be needed anymore.
The `amrex` function `getParticleCell` does not have dimension specific logic.

ax3l:
CI is fixed for good now. Once adding the update from @roelof-groenewald , please rebase once more @archermarx .

archermarx:
Ok, made the requested changes and rebased

Source/Utils/ParticleUtils.cpp
```diff
@@ -57,12 +56,9 @@ namespace ParticleUtils
...
         bins.build(np, ptd, cbx,
             // Pass lambda function that returns the cell index
-            [=] AMREX_GPU_DEVICE (ParticleType const & p) noexcept -> amrex::IntVect
+            [=] AMREX_GPU_DEVICE (ParticleType const & p) noexcept -> IntVect
             {
-                return IntVect{AMREX_D_DECL(
-                                   static_cast<int>((p.pos(0)-plo[0])*dxi[0] - lo.x),
-                                   static_cast<int>((p.pos(1)-plo[1])*dxi[1] - lo.y),
-                                   static_cast<int>((p.pos(2)-plo[2])*dxi[2] - lo.z))};
+                return getParticleCell(p, plo, dxi, cbx);
```
roelof-groenewald: I see most (maybe all) of the examples that involve sorting the particles per cell are still failing. Looking closer, I see that the current `development` function here calculates the cell indices with 
```
ii = (p.pos(0) - plo[0]) * dxi[0] - lo.x
```
whereas the AMReX function uses
```
ii = (p.pos(0) - plo[0]) * dxi[0] + cbx.smallend[0]
```
So the difference is that the `development` version calculates a cell index starting from 0 for each tile whereas the AMReX function calculates a global cell index for the entire domain. It's not clear to me why this breaks the code though since the binning should only care about which particles have the same cell index, not about what the value of the indices are.

roelof-groenewald:
I'll look more into the current reason for the test failures but I also wanted to point out that the new utility function can be used in https://github.com/ECP-WarpX/WarpX/blob/development/Source/Diagnostics/ComputeDiagFunctors/ParticleReductionFunctor.cpp in a similar way as in the temperature diagnostic.

archermarx:
OK, I can add it in there too

ax3l:
The method might actually be general enough to not place in WarpX but ABLASTR if you like (aka: does not use any WarpX globals or helpers and is standalone/modular enough): https://github.com/ECP-WarpX/WarpX/tree/development/Source/ablastr

Source/Utils/ParticleUtils.H
```diff
@@ -17,6 +17,75 @@
 
 namespace ParticleUtils {
 
+    /**
+     * \brief Determine which cell a particle lies in.
```
ax3l: Title+body notation is what I would prefer here, use like this:
```suggestion
    /**  Determine which cell a particle lies in.
     *
```

Source/Utils/ParticleUtils.H
```diff
@@ -17,6 +17,75 @@
...
+#if defined(WARPX_DIM_3D)
+        const amrex::Real lz = (zp - plo[2]) * dxi[2] - lo.z;
+        kk = static_cast<int>(amrex::Math::floor(lz));
+#endif
+
+        return amrex::IntVect(AMREX_D_DECL(ii, jj, kk));
+    }
+
+    /**
+     * \brief Determine which cell a particle lies in.
```
ax3l:
```suggestion
    /** Determine which cell a particle lies in.
     *
```

Source/Utils/ParticleUtils.H
```diff
@@ -17,6 +17,75 @@
...
+    AMREX_GPU_HOST_DEVICE AMREX_INLINE
+    amrex::IntVect getParticleCellIndex (const WarpXParticleContainer::ParticleType& p,
+                                         amrex::GpuArray<amrex::Real,AMREX_SPACEDIM> const& plo,
+                                         amrex::GpuArray<amrex::Real,AMREX_SPACEDIM> const& dxi,
+                                         amrex::Dim3 lo) {
+        return getParticleCellIndex(p.pos(0), p.pos(1), p.pos(2), plo, dxi, lo);
+    }
+
+    /**
+     * \brief Determine which cell a particle lies in.
```
ax3l:
```suggestion
    /** Determine which cell a particle lies in.
     *
```

archermarx:
> The method might actually be general enough to not place in WarpX but ABLASTR if you like (aka: does not use any WarpX globals or helpers and is standalone/modular enough): https://github.com/ECP-WarpX/WarpX/tree/development/Source/ablastr

Sure, that sounds reasonable. I'll move it there and add the extra use of it in ParticleReductionFunctor tomorrow

archermarx:
I added a use of the function to `ParticleReductionFunctor.cpp.` Before I move anything to `ablastr`, however, let's make sure we can get CI to pass.

archermarx:
I reverted the use of the new function in `ParticleUtils.cpp` to see if we can get CI passing, and then go from there.

archermarx:
Looks like `collisionXZ` and `collisionXYZ` tests are still failing, but all other CI tests pass.

archermarx:
No clue why any failures at all should be happening. The code in `get_particle_cell_index` is literally copy-pasted from the code in `TemperatureFunctor.cpp` and `ParticleReductionFunctor.cpp`

roelof-groenewald:
> Looks like `collisionXZ` and `collisionXYZ` tests are still failing, but all other CI tests pass.

Great! Those tests have been problematic for a while now (see #5132). After that PR is merged you can rebase and they should pass.

archermarx:
Oh great, that's good to know!

roelof-groenewald:
I would be okay to merge this since it already cleans up the code quite a bit, but we should be able to replace 
```
    amrex::IntVect getParticleCellIndex (const WarpXParticleContainer::SuperParticleType& p,
                                         amrex::GpuArray<amrex::Real,AMREX_SPACEDIM> const& plo,
                                         amrex::GpuArray<amrex::Real,AMREX_SPACEDIM> const& dxi) {
        return getParticleCellIndex(p.pos(0), p.pos(1), p.pos(2), plo, dxi);
    }
```
with something like
```
    amrex::IntVect getParticleCellIndex (const WarpXParticleContainer::SuperParticleType& p,
                                         amrex::GpuArray<amrex::Real,AMREX_SPACEDIM> const& plo,
                                         amrex::GpuArray<amrex::Real,AMREX_SPACEDIM> const& dxi) {
        // create zero_low_bound_box
        return amrex::getParticleCell(p, plo, dxi, zero_low_bound_box);
    }
```
where `zero_low_bound_box` is an `amrex::Box` that has the property `zero_low_bound_box.smallEnd() == {0, 0, 0}`. This would allow us to directly use the AMReX function and not need all the extra logic in WarpX repeated. @ax3l could you help us out with how to create `zero_low_bound_box`?
Or alternatively, we can maybe add an overload to AMReX for `amrex::getParticleCell` that does not take the 4th argument and in that case does not shift the cell index based on the global box index.

EDIT: On second thought, I like that second option. I'll go ahead and submit a PR to AMReX.

archermarx:
That sounds good, and would let us eliminate a lot of code. For the version that takes positions instead of a particle type, is there a way we can delegate to `getParticleCell` as well? That version is only called in the `WarpXParIter` loop in `TemperatureFunctor.cpp`. Can we create a `WarpxParticleType` from the elements of that iterator?

archermarx:
I just pushed a version that uses your overload of `getParticleCell`. Let me know what you think and if you think we can further simplify

archermarx:
Ok, looks like that got merged into AMReX. Once we update AMReX for the week, this should work, I think

roelof-groenewald:
@archermarx would you mind rebasing on development now that the AMReX version has been updated?

archermarx:
Yes, will do in a moment

archermarx:
To fix some further redundancy, I have eliminated the new function entirely and just call the new overload of `amrex::GetParticleCell`. To do that, I had to make a small modification to `TemperatureFunctor.cpp` to grab a `ParticleType` instead of using the `GetPosition` API. Let me know if there's a cleaner way to do this than the way I'm doing it.

archermarx:
CI should pass now. I tried replacing the particle cell finding in `ParticleUtils.cpp` with a use of `amrex::getParticleCell` but CI failed for some reason. Not sure why.

Source/Diagnostics/ComputeDiagFunctors/TemperatureFunctor.cpp
```diff
@@ -5,6 +5,7 @@
 #include "Particles/MultiParticleContainer.H"
 #include "Particles/WarpXParticleContainer.H"
 #include "Utils/Parser/ParserUtils.H"
+#include "Utils/ParticleUtils.H"
```
roelof-groenewald: Is this still needed?
```suggestion
```

archermarx: Great catch. No!

roelof-groenewald (approved):
Thanks for this PR! I think PRs that remove this many lines of code without altering functionality (or obscuring it) are really great!

Source/Diagnostics/ComputeDiagFunctors/ParticleReductionFunctor.cpp
```diff
@@ -69,23 +69,7 @@ ParticleReductionFunctor::operator() (amrex::MultiFab& mf_dst, const int dcomp,
...
-                const amrex::ParticleReal y = p.pos(1);
-                const amrex::Real ly = (y - plo[1]) * dxi[1];
-                jj = static_cast<int>(amrex::Math::floor(ly));
-#endif
-#if defined(WARPX_DIM_3D)
-                const amrex::ParticleReal z = p.pos(2);
-                const amrex::Real lz = (z - plo[2]) * dxi[2];
-                kk = static_cast<int>(amrex::Math::floor(lz));
-#endif
+                const auto [ii, jj, kk] = amrex::getParticleCell(p, plo, dxi).dim3();
```
ax3l, line 72: Ah, there is another branch, `m_do_average`, below that can use the same functionality.

ax3l: Follow-up in #5557

PR #5120, opened 2024-08-07
CI: clean test dirs, reset `collisionXZ` checksums

EZoni:
Trying temporary work-arounds to fix current CI failures. Based on the temporary branch `EZoni_rm_testdir` in my own fork of the `regression_testing` repo, see https://github.com/AMReX-Codes/regression_testing/pull/146.

To-do:
- [x] ~~Fix local runs (if `-c` is not passed, we do not parse all test names correctly as is)~~

### DIFF
--- a/run_test.sh
+++ b/run_test.sh
@@ -23,10 +23,25 @@
 
 set -eu -o pipefail
 
-# Parse command line arguments: if test names are given as command line arguments,
+# Parse command line arguments:
+# - if -c option is given, set clean=1 and call regtest.py with option --rm_testdir
+# in order to remove all subdirectories from each test directory, after each test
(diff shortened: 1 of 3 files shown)


run_test.sh
```diff
@@ -23,10 +23,25 @@
...
 # store them in variable tests_arg and define new command line argument to call
 # regtest.py with option --tests (works also for single test)
-tests_arg=$*
+clean=0
+while getopts c name
+do
+    case ${name} in
+    c)    clean=1;;
+    esac
+done
```
EZoni: Use environment variable instead.

ax3l: done :)

ax3l (approved):

PR #5121, opened 2024-08-09
CI: heFFTe for tests

ax3l:
Add heFFTe to Azure runners to enable runtime tests that need it.

This uses my usual [cmake-easyinstall](https://github.com/ax3l/cmake-easyinstall) helper script.

Needed for #4937

- [x] waiting for confirmation which runners will use this (1D/2D/3D/other) to optimize setup time
- [x] check if other build-only CI should have a heFFTe build, too, e.g., CUDA and ROCm (and OneAPI, later - local FFT first)


aeriforme:
@Haavaan and I have one 3D test only on our side.

ax3l:
That's great, then I keep it optimized for those for now and save compile time for the other tests :)

RemiLehe (approved):

PR #5123, opened 2024-08-10
CI: Ruff

ax3l:
[Ruff](https://docs.astral.sh/ruff/) is an extremely fast Python linter and code formatter. It replaces tools such as autoflake, black, flake8, pyflakes, or pylint.

Close #4366

We already use `ruff` in [ImpactX](https://github.com/ECP-WarpX/impactx/pull/522) and pyAMReX, this adds it to WarpX using the same fast check & Python code formatter methods.

**Reviewer note: This pull request is best reviewed commit-by-commit.**

- [x] merge #5124 first, then rebase
- [x] address additional issues


dpgrote:
I was looking over the changes in the python scripts and see a some places where the changes will break things. I'll add comments for those that I see. But this looks great otherwise!

Examples/Tests/nuclear_fusion/analysis_two_product_fusion.py
```diff
@@ -405,7 +407,7 @@ def main():
     ds_start = yt.load(filename_start)
     ad_end = ds_end.all_data()
     ad_start = ds_start.all_data()
-    dt = float(ds_end.current_time - ds_start.current_time)
+    float(ds_end.current_time - ds_start.current_time)
```
dpgrote: The `eval` below needs dt. The `eval` is kind of hacky anyway. Something like this would be better
```
if i == 1:
    specific_check1(data, dt)
elif i == 2:
    specific_check2(data, dt)
```

Examples/Tests/photon_pusher/analysis_photon_pusher.py
```diff
@@ -152,7 +152,7 @@ def generate():
             f.write("{}.single_particle_u = {} {} {}\n".
                 format(name, velx, vely, velz))
             f.write("{}.single_particle_weight = 1.0\n".format(name))
-            f.write("\n".format(name))
+            f.write("\n".format())
```
dpgrote: This should be just `f.write("\n")`

Tools/PerformanceTests/run_alltests_1node.py
```diff
@@ -11,7 +11,14 @@
 import shutil
 import time
 
-from functions_perftest import *
+import pandas as pd
+from functions_perftest import (
```
dpgrote: It would probably be better to do something like `import functions_perftest as perftest` and call the routines with that prefix, e.g. `perftest.extract_dataframe`.

Tools/PerformanceTests/run_alltests.py
```diff
@@ -12,7 +12,14 @@
 import shutil
 import time
 
-from functions_perftest import *
+import pandas as pd
```
dpgrote: See comment below.

Examples/Tests/nuclear_fusion/analysis_two_product_fusion.py
ax3l: Fully agreed, the linter overlooked this because of the `eval` call, which is generally unsafe/hacky. Let's add this to #5124 where I pulled those out (ruff in this PR finds even more than I could fix so far).

Tools/PerformanceTests/run_alltests.py
ax3l: Thx! Added to #5124, will rebase out here later.

Tools/PerformanceTests/run_alltests_1node.py
ax3l: Thx! Added to #5124, will rebase out here later.

Examples/Tests/photon_pusher/analysis_photon_pusher.py
ax3l: Thx! Added to #5124, will rebase out here later.

Examples/Tests/nci_psatd_stability/analysis_galilean.py
```diff
@@ -29,40 +31,43 @@
...
 periodic_single_box = False
-warpx_used_inputs = open('./warpx_used_inputs', 'r').read()
-if re.search('geometry.dims\s*=\s*2', warpx_used_inputs):
-    dims = '2D'
-elif re.search('geometry.dims\s*=\s*RZ', warpx_used_inputs):
-    dims = 'RZ'
-elif re.search('geometry.dims\s*=\s*3', warpx_used_inputs):
-    dims = '3D'
-if re.search('psatd.current_correction\s*=\s*1', warpx_used_inputs):
+warpx_used_inputs = open("./warpx_used_inputs", "r").read()
```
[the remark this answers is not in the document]
ax3l: fixed

Examples/Tests/nuclear_fusion/analysis_two_product_fusion.py
```diff
@@ -46,46 +46,48 @@
...
-MeV_to_Joule = scc.e*1e6
-barn_to_square_meter = 1.e-28
+keV_to_Joule = scc.e * 1e3
+MeV_to_Joule = scc.e * 1e6
+barn_to_square_meter = 1.0e-28
 
 ## Checks whether this is the 2D or the 3D test
-warpx_used_inputs = open('./warpx_used_inputs', 'r').read()
-if re.search('geometry.dims = RZ', warpx_used_inputs):
+warpx_used_inputs = open("./warpx_used_inputs", "r").read()
```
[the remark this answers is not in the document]
ax3l: fixed

.github/workflows/source/makeMakefileForClangTidy.py
```diff
@@ -14,40 +14,49 @@
...
 def makeMakefileForClangTidy(argv):
     parser = argparse.ArgumentParser()
-    parser.add_argument("--input",
-                        help="Ccache log file",
-                        default="ccache.log.txt")
-    parser.add_argument("--identifier",
-                        help="Unique identifier for finding compilation line in the log file",
-                        default="WarpX/Source")
+    parser.add_argument("--input", help="Ccache log file", default="ccache.log.txt")
+    parser.add_argument(
```
dpgrote, line 18: If I were formatting this, I would do this, which I think is easier to read (maybe I'm just more used to it)
```
parser.add_argument("--identifier",
                    help = "Unique identifier for finding compilation line in the log file",
                    default = "WarpX/Source")
```
The arguments aligned with the routine name, and a space before and after the `=`.

.github/workflows/source/makeMakefileForClangTidy.py
```diff
@@ -14,40 +14,49 @@
...
+        "override CLANG_TIDY_ARGS += --extra-arg=-Wno-unknown-warning-option --extra-arg-before=--driver-mode=g++\n"
+    )
     fout.write("\n")
 
     fout.write(".SECONDEXPANSION:\n")
     fout.write("clang-tidy: $$(all_targets)\n")
     fout.write("\t@echo SUCCESS\n\n")
 
-    exe_re = re.compile(r" Executing .*? (-.*{}.*) -c .* -o .* (\S*)".format(args.identifier))
+    exe_re = re.compile(
```
dpgrote, line 46: My own preference is allowing longer lines. Keeping this as one line is more succinct and easier to read (for me at least).

Docs/source/conf.py
```diff
@@ -34,7 +34,12 @@
 import sphinx_rtd_theme  # noqa
 from pybtex.style.formatting.unsrt import Style as UnsrtStyle
 
-sys.path.insert(0, os.path.join(os.path.dirname(os.path.abspath(__file__)), '../../Regression/Checksum'))
```
dpgrote, line 37: In this case, this would be cleaner if it used temporaries, like this
```
module_path = os.path.dirname(os.path.abspath(__file__))
checksum_path = os.path.join(module_path, "../../Regression/Checksum")
sys.path.insert(0, checksum_path)
```
But of course, ruff couldn't do this so this would have to be a separate PR.

Docs/source/conf.py
```diff
@@ -131,14 +138,16 @@ def __init__(self, *args, **kwargs):
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = 'sphinx_rtd_theme'
+html_theme = "sphinx_rtd_theme"
 
 numfig = True
 math_eqref_format = "{number}"
-numfig_format = {'figure': 'Fig. %s',
```
dpgrote, line 138: I like this formatting better. Is there an option with ruff to use this style?

Examples/Physics_applications/capacitive_discharge/PICMI_inputs_1d.py
```diff
@@ -65,7 +67,7 @@ def solver_initialize_inputs(self):
         self.nxguardphi = 1
         self.nzguardphi = 1
 
-        self.phi = np.zeros(self.nz + 1 + 2*self.nzguardphi)
```
dpgrote, line 69: There's probably not an option for this, but I much prefer no spaces around `*` (and `/`) since it makes it look more like standard math expressions, with the grouping of multiplied terms. Compare these two expressions
```
x = (3 + 2 * a) * (b * c + 55 * d)
x = (3 + 2*a)*(b*c + 55*d)
```
Which is easier to understand?

Examples/Physics_applications/laser_acceleration/PICMI_inputs_3d.py
```diff
@@ -48,114 +49,114 @@
 plasma_ymax = 20e-06
 plasma_zmax = None
 uniform_distribution = picmi.UniformDistribution(
-    density = plasma_density,
```
dpgrote, line 51: I prefer spaces around `=` since it makes the text less dense and easier to read.

.github/workflows/source/makeMakefileForClangTidy.py
ax3l: Thanks. I think the reason that this is broken to newlines in black is that it will reduce the `diff` size when adding more or removing arguments in a later PR, which makes code easier to review (less files changed).

This is auto-formatted using [black style](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html) ([FAQ](https://black.readthedocs.io/en/stable/faq.html)), as we do in all other projects already:
https://docs.astral.sh/ruff/formatter/

Developers can install a [pre-commit](https://github.com/ECP-WarpX/WarpX/blob/development/.pre-commit-config.yaml):
```bash
python -m pip install pre-commit
pre-commit install
```
to get auto-formatting, otherwise the PR auto-bot does it for them.

.github/workflows/source/makeMakefileForClangTidy.py
ax3l: We usually do not mingle with the default, which are currently around 100 characters (soft limitish with some heuristics). The bot auto-breaks them. This makes `diffs` on github easier to read.

We can increase that limit manually, but since a bot formats them we did keep the defaults on all other BLAST projects so far.

Looking at line 38, the limit is still pretty long, so I would not bump it further.

Docs/source/conf.py
ax3l: I agree, we can apply this manually here :+1:

Docs/source/conf.py
ax3l: Nope, it just uses [black style](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html). As far as I read the style is intentional, the extra break reduces the amount of lines changed when adding/removing new arguments in PRs to dictionaries. (Note also the trailing `,`.)
> Black aims for consistency, generality, readability and reducing git diffs.

The prior formatting would often change the first and last line on updates to the dict.

Examples/Physics_applications/capacitive_discharge/PICMI_inputs_1d.py
ax3l: Spaces around operators is a pretty standard thing these days in C++ and Python. PEP8 requires it even.

For long and block formulas, we can always [disable formatting](https://docs.astral.sh/ruff/formatter/#format-suppression), so we can have nice alignments when we need it, e.g.,
```py
# fmt: off
x = a * ( 42 * b
         + 3 * c
         - 7 * d
        ) 
# fmt: on

```

Examples/Physics_applications/laser_acceleration/PICMI_inputs_3d.py
ax3l: I think PEP8 requires for arguments no spaces and spaces only for assignments. Independent of PEP8, [black style](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html) seems to do the same.

Docs/source/conf.py
```diff
@@ -34,7 +34,12 @@
 import sphinx_rtd_theme  # noqa
 from pybtex.style.formatting.unsrt import Style as UnsrtStyle
 
-sys.path.insert(0, os.path.join(os.path.dirname(os.path.abspath(__file__)), '../../Regression/Checksum'))
+sys.path.insert(
+    0,
+    os.path.join(
+        os.path.dirname(os.path.abspath(__file__)), "../../Regression/Checksum"
+    ),
+)
```
ax3l: @dpgrote suggested a better way via temporaries:
```suggestion
module_path = os.path.dirname(os.path.abspath(__file__))
checksum_path = os.path.join(module_path, "../../Regression/Checksum")
sys.path.insert(0, checksum_path)
```

Docs/source/conf.py
ax3l: Applied :)

Examples/Physics_applications/capacitive_discharge/PICMI_inputs_1d.py
dpgrote: Oh yes, I'm well aware that the spaces around operators is standard, but I never liked it for `*` and `/`. I've never understood why something that is so much harder to read and more prone to typos would be the standard.
BTW, PEP8 actually says to use white space around the lower priority operators, so my example would be
```
x = (3 + 2*a) * (b*c + 55*d)
```
similar to what I would do.

ax3l: I know :sweat_smile: That is how I type as well but I don't mind the auto-formatter adding spaces by now anymore :D

In the end, there are more precedences than I have spaces for, e.g., C++:
https://en.cppreference.com/w/cpp/language/operator_precedence

Examples/Physics_applications/capacitive_discharge/analysis_1d.py
```diff
@@ -4,42 +4,140 @@
...
-       2.56564415e+14, 2.56824494e+14, 2.56459673e+14, 2.56106618e+14,
-       2.56442627e+14, 2.57106509e+14, 2.57468066e+14, 2.57155708e+14,
-       2.56501993e+14, 2.56062276e+14, 2.56109738e+14, 2.56284272e+14,
-       2.56320557e+14, 2.56253594e+14, 2.56155365e+14, 2.56044920e+14,
-       2.55994560e+14, 2.56166612e+14, 2.56327633e+14, 2.56390898e+14,
-       2.56474891e+14, 2.56669566e+14, 2.56646041e+14, 2.56264156e+14,
-       2.56041610e+14, 2.56041551e+14, 2.56088641e+14, 2.23853646e+14,
-       1.27580207e+14
-])
+ref_density = np.array(
```
roelof-groenewald: Can we skip this? The reason is because the `print(repr(density_data))` line below provides the format as it currently is. I guess updating values could be done as a two step process with first updating using the given format and then letting the pre-commit change it, but I think the original is better.

Examples/Physics_applications/capacitive_discharge/analysis_dsmc.py
```diff
@@ -17,42 +17,140 @@
...
-    2.56250819e+14, 2.56045278e+14, 2.56366075e+14, 2.56693733e+14,
-    2.56618530e+14, 2.56580918e+14, 2.56812781e+14, 2.56754216e+14,
-    2.56444736e+14, 2.56473391e+14, 2.56538398e+14, 2.56626551e+14,
-    2.56471950e+14, 2.56274969e+14, 2.56489423e+14, 2.56645266e+14,
-    2.56611124e+14, 2.56344324e+14, 2.56244156e+14, 2.24183727e+14,
-    1.27909856e+14
-])
-
-density_data = np.load( 'ion_density_case_1.npy' )
+ref_density = np.array(
```
roelof-groenewald: Same comment here as above.

Examples/Physics_applications/laser_acceleration/analysis_1d_fluids_boosted.py
```diff
@@ -51,17 +53,22 @@
 # ODE Function
 def odefcn(phi, xi, kp, a0, c, tau, xi_0, lambda_laser):
     phi1, phi2 = phi
-    a_sq = a0**2 * np.exp(-2 * (xi - xi_0)**2 / (c**2 * tau**2))*np.sin(2*np.pi*(xi - xi_0)/lambda_laser)**2
+    a_sq = (
+        a0**2
+        * np.exp(-2 * (xi - xi_0) ** 2 / (c**2 * tau**2))
```
roelof-groenewald, line 58: It's a little unfortunate that the standard is inconsistent with powers:
```
(xi - xi_0) ** 2
```
but
```
c**2
```
Is there a way to have this be consistent, i.e., use `()**2` instead of `() ** 2`?

Python/pywarpx/callbacks.py
```diff
@@ -159,19 +161,19 @@ def callbackfunclist(self):
...
+                    print("For %s" % self.name)
                 print("Only callable objects can be installed.")
                 print("It is possible that the callable's name has been overwritten")
                 print("by something not callable. This can happen during restart")
                 print("if a function name had later been used as a variable name.")
                 print(self.name)
-                if isinstance(f,str):
-                    print("The name of the call back is %s"%f)
+                if isinstance(f, str):
+                    print("The name of the call back is %s" % f)
```
roelof-groenewald: I guess another consistency might be to use `f-strings` everywhere that substitution is made. Could that be added to the Ruff checks?

roelof-groenewald (approved):
This is an awesome change to enforce consistency in Python code! :tada:

I left two comments on things I think would be nice to change / add (formatting of powers in arithmetic expressions and forcing `f-strings`). But I'll approve this PR so long and those can be changed in future PRs. 
If you agree, @ax3l, it would be nice to undo the changes to the capacitive_discharge analysis files before merging.

Examples/Physics_applications/laser_acceleration/PICMI_inputs_3d.py
dpgrote: PEP8 does say no space around `=` for arguments, but only shows this when the arguments are all on one line. This is fine. But for multiline function calls, I find it much more readable to have spaces around `=`. It is inconsistent to require spaces around all operators (to spread things out) except in this one case where it would in fact make the code easier to read.

Examples/Physics_applications/capacitive_discharge/PICMI_inputs_1d.py
dpgrote: "I don't mind the auto-formatter adding spaces" meaning you've gotten used to the bad formatting decision :)

Python/pywarpx/callbacks.py
ax3l: I am not sure if we can enforce it, but we can use it manually here :)
```suggestion
                    print(f"The name of the call back is {f}")
```

Examples/Physics_applications/laser_acceleration/analysis_1d_fluids_boosted.py
ax3l: I think here are the rules on powers: https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-breaks-binary-operators

Examples/Physics_applications/capacitive_discharge/analysis_1d.py
ax3l: If you update the block with what `print(repr(density_data))` outputs, the pre-commit will auto-format it :)

otherwise, if you prefer we can wrap the block in
```py
# fmt: off
...
# fmt: on
```

roelof-groenewald: For the sake of moving forward we can leave this as is. Next time I have to update the values due to a change in the collision routines I'll add the `# fmt: off/on` block.

Python/pywarpx/callbacks.py
roelof-groenewald: If it can't be enforced, we also don't have to change this. There are several places where old replacement strings are still used.

ax3l: Let's start somewhere :D

Examples/Physics_applications/capacitive_discharge/analysis_1d.py
ax3l: already done :D

Examples/Physics_applications/laser_acceleration/PICMI_inputs_3d.py
ax3l: You are right, I think it's a pragmatic choice in black. They do spaces for assignments and assignments in definitions (default values) but not in call site arguments. I kinda like it because I like my call sites compact, but as all style choices it might come down to taste and what one is used to :)

jwestern:
I just want to raise awareness that style changes can have costs for some users, which might not be obvious. Particularly users working with custom extensions of WarpX. This commit created (a lot of) trivial auto-merge conflicts in our workflow, which took a while to track down, and I'm still looking for a workaround.

ax3l:
Hi @jwestern,

Thank you for raising this.

Fully understand. This style change is intended to be a one-time change, which is once a bit painful (e.g., forks and open PRs) but will actually make all _follow-up work_ easier to [maintain, review, pull and merge](https://docs.astral.sh/ruff/formatter/#philosophy) - _if_, you also use ruff in your PRs (automatic) and forks (please set up). Sorry for this one-time change, we try to keep those overall reformats to a minimum.

As a general recommendation for extending WarpX, we do move fast indeed. As people in other projects (e.g., the Cling people on LLVM), maintaining a true fork can be challenging contrary to _extending_ a project.
This is why, we recommend to do the following if you like to extend WarpX:
- plan which parts are generally needed/useful or enabling for your workflow and contribute them upstream
- add additional callbacks/hooks and options you might need upstream here, and only
- extend WarpX via our Python bindings (can still couple in more C++, e.g., via pybind11)
  - you can also use WarpX and extend it as a pure C++ library, in fact both our app executable and Python bindings link against exactly a libwarpx, but it is more productive to go via Python as glue code

That way, you would not have to maintain a full fork and only would have to update breaking API changes, which change less often.

Examples of a fork of WarpX that has same challenge as you: https://github.com/AMReX-Microelectronics/artemis

Examples of companies extending WarpX and contributing common/core functionality: Modern Electron (a few years back), TAE, Helion, etc.

PR #5127, opened 2024-08-11
Add oneMKL FFT to AnyFFT wrapper for SYCL backend

WeiqunZhang:
Status: I did a 2d test Examples/Tests/langmuir/inputs_2d. ~~Unfortunately the results are wrong. The issue might be in the new FFT wrapper. But it could also be in somewhere else because we have not run much on Intel GPUs.~~ It works now after explicitly setting strides in the forward domain.


Source/ablastr/math/fft/WrapMklFFT.cpp
```diff
@@ -0,0 +1,84 @@
...
+        sycl::event r;
+        if (fft_plan.m_dir == direction::R2C) {
+            r = oneapi::mkl::dft::compute_forward(
+                *fft_plan.m_plan,
+                fft_plan.m_real_array,
+                reinterpret_cast<amrex::Real*>(fft_plan.m_complex_array));
+        } else {
+            r = oneapi::mkl::dft::compute_backward(
+                *fft_plan.m_plan,
+                reinterpret_cast<amrex::Real*>(fft_plan.m_complex_array),
```
ax3l: It looks to me like input & output types are distinct types in the API:
https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-dpcpp/2024-1/compute-forward-typename-descriptor-type-typename.html

Is there a need to cast the complex pointers to `Real*`?

WeiqunZhang: Yes, it does not compile with `amrex::GpuComplex<Real>*`. Either `Real*` or `std::complex<Real>*` works.

ax3l (approved):
Yay! :tada:

Thank you!

Can you run a 1D and 3D PSATD or ES IGF test as well just to check? :)

Source/ablastr/math/fft/WrapMklFFT.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+namespace ablastr::math::anyfft
+{
+
+    void setup(){/*nothing to do*/}
+
+    void cleanup(){/*nothing to do*/}
+
+    FFTplan CreatePlan(const amrex::IntVect& real_size, amrex::Real * const real_array,
+                       Complex * const complex_array, const direction dir, const int dim)
```
ax3l: Optional AMReX-ify:
```suggestion
    void setup () {/*nothing to do*/}

    void cleanup () {/*nothing to do*/}

    FFTplan CreatePlan (const amrex::IntVect& real_size, amrex::Real * const real_array,
                        Complex * const complex_array, const direction dir, const int dim)
```

Source/ablastr/math/fft/WrapMklFFT.cpp
```diff
@@ -0,0 +1,96 @@
...
+        fft_plan.m_real_array = real_array;
+        fft_plan.m_complex_array = complex_array;
+        fft_plan.m_dir = dir;
+        fft_plan.m_dim = dim;
+        fft_plan.m_stream = amrex::Gpu::gpuStream();
+
+        return fft_plan;
+    }
+
+    void DestroyPlan(FFTplan& fft_plan)
```
ax3l:
```suggestion
    void DestroyPlan (FFTplan& fft_plan)
```

ax3l (approved):

PR #5129, opened 2024-08-12
CI: update option name to clean test dirs

EZoni:
The option `rm_testdir` was renamed to `clean_testdir` in https://github.com/AMReX-Codes/regression_testing/pull/146.

### DIFF
--- a/run_test.sh
+++ b/run_test.sh
@@ -83,7 +83,7 @@ curl -sOL https://github.com/openPMD/openPMD-example-datasets/raw/4ba1d257c5b489
 cd -
 
 # Clone the AMReX regression test utility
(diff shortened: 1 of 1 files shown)


EZoni:
@ax3l 
PR https://github.com/AMReX-Codes/regression_testing/pull/146 on the AMReX regression testing side was just merged and the corresponding branch deleted. Thus the current CI setup where we clone that branch and use the old name of the option will not work anymore. This PR should fix that.

Update: I just restored the branch of https://github.com/AMReX-Codes/regression_testing/pull/146 because this might cause issues given that the previous implementation was included in the latest release. It should be safe to delete my branch once the next release is issued.

ax3l (approved):

PR #5126, opened 2024-08-11
AMReX/pyAMReX/PICSAR: Weekly Update

ax3l:
Weekly update to latest AMReX.
Weekly update to latest pyAMReX.
Weekly update to latest PICSAR (no changes).

```console
./Tools/Release/updateAMReX.py
./Tools/Release/updatepyAMReX.py
./Tools/Release/updatePICSAR.py
```

- [x] needs updates from https://github.com/AMReX-Codes/pyamrex/pull/352
- [x] breaking changes from https://github.com/AMReX-Codes/amrex/pull/3684
- [x] breaking change in `ref_ratios.size()`
- [x] breaking change of `amr.max_grid_size` default from 32 to 64

PR #5132, opened 2024-08-13
CI: fix fluctuations in `collisionXZ` checksums

EZoni:
Trying to see if we can get under control the fluctuations observed in the checksums of the `collisionXZ` test.

### DIFF
--- a/Examples/Tests/collision/analysis_collision_2d.py
+++ b/Examples/Tests/collision/analysis_collision_2d.py
@@ -114,4 +114,4 @@ post_processing_utils.check_random_filter(last_fn, random_filter_fn, random_frac
                                           dim, species_name)
 
 test_name = os.path.split(os.getcwd())[1]
-checksumAPI.evaluate_checksum(test_name, fn)
+checksumAPI.evaluate_checksum(test_name, fn, rtol=1e-1)
--- a/Examples/Tests/collision/analysis_collision_3d.py
+++ b/Examples/Tests/collision/analysis_collision_3d.py
@@ -111,4 +111,4 @@ post_processing_utils.check_random_filter(last_fn, random_filter_fn, random_frac
                                           dim, species_name)
 
 test_name = os.path.split(os.getcwd())[1]
-checksumAPI.evaluate_checksum(test_name, fn)
+checksumAPI.evaluate_checksum(test_name, fn, rtol=1e-1)


ax3l:
That parameter should be set here already:
https://github.com/ECP-WarpX/WarpX/blob/502e5ea31c1334f0083bed64eb8f8202b1100e81/Regression/prepare_file_ci.py#L86

EZoni:
> That parameter should be set here already:
> 
> https://github.com/ECP-WarpX/WarpX/blob/502e5ea31c1334f0083bed64eb8f8202b1100e81/Regression/prepare_file_ci.py#L86

That's right, thanks for pointing it out. I always forget that we set input parameters in three different places at the moment. For now I'm just keeping this PR open (now it simply increases the relative tolerance for the checksums) until the next meeting where this issue will be discussed.

ax3l:
Since `development` is failing, can you set a tolerance _just_ high enough for it to pass and we merge it?

EZoni:
I'm afraid `1e-1` might be the one that is just high enough. At least on my computer, the test fails with `1e-2`. I think this is consistent with the "10%" error that @RemiLehe mentioned offline. I could check if this is the case also on Azure, but it is for sure on my computer. This holds for both `collisionXZ` and `collisionXYZ`.

ax3l (approved):
OK for me for now.

@RemiLehe @JustinRayAngus what do you think, shall we use this for now?

roelof-groenewald (approved):
I agree this is good to do until the underlying issue with the collision tests can be sorted out.

JustinRayAngus (commented):
Approve. What's different between these two tests and the other tests that use Coulomb collisions?

ax3l:
We don't understand that yet... Maybe less particles per cell or per simulation and thus worse statistics?

ax3l:
In #5126 I see that `collisionZ` fluctuates a well.
```
./analysis_collision_1d.py  collisionZ_plt000600
TApar at 30ps error =  0.0025701770198334937
tolerance =  0.02
ERROR: Benchmark and output file checksum have different value for key [ions,particle_momentum_x]
Benchmark: [ions,particle_momentum_x] 3.425400072687143e-16
Test file: [ions,particle_momentum_x] 3.424633029669351e-16
Absolute error: 7.67e-20
Relative error: 2.24e-04
ERROR: Benchmark and output file checksum have different value for key [ions,particle_momentum_y]
Benchmark: [ions,particle_momentum_y] 3.421937133999805e-16
Test file: [ions,particle_momentum_y] 3.429026824477811e-16
Absolute error: 7.09e-19
Relative error: 2.07e-03
ERROR: Benchmark and output file checksum have different value for key [ions,particle_momentum_z]
Benchmark: [ions,particle_momentum_z] 5.522701882677923e-16
Test file: [ions,particle_momentum_z] 5.528396735566589e-16
Absolute error: 5.69e-19
Relative error: 1.03e-03
ERROR: Benchmark and output file checksum have different value for key [ions,particle_position_x]
Benchmark: [ions,particle_position_x] 7.200011611411148e+02
Test file: [ions,particle_position_x] 7.200708684755696e+02
Absolute error: 6.97e-02
Relative error: 9.68e-05
```

Notable differences in runners:
- `Azure Pipelines 9` (newer) in PR vs. `Azure Pipelines 8` on development
- CPUs: `Xeon(R) Platinum 8272CL` "Cascade Lake" (~2019) on PR vs. `Xeon(R) CPU E5-2673 v4` "Broadwell" (~2016) on development

PR #5133, opened 2024-08-13
Nuclear fusion algorithm modifications

JustinRayAngus:
This PR makes  changes to the Nuclear Fusion algorithm that allows for better control of the particle weights.

The current implentation uses an effective fusion_multiplier equal to max(NA,NB)/min(NA,NB), where NA and NB are, respectively, the number of macro-particles in a cell for species A and species B. The probability of a collision is enhanced by Nmax/Nmin, and the weight of the fusion products is decreased by the same factor to keep the physical number of particles created correct. This implementation does not permit maintaining equally-weighted particles for the incident species.

The new implementation introduced in this PR only uses the fusion_multiplier specified in the input file. If that value is set to 1, and the particles are equally-weighted, then the particles of the colliding species will remain equally-weighted.

Below are results from a mono-energetic Deuterium beam incident on a cold stationary Tritium background. The density of the beam is 10X less than the stationary Tritium background. The number of particles per cell for the deuterium beam is 160 and that for the tritium background is 1600. Thus, the particles have equal weights initially. The total weight of deuterium particles decays exponentially for this problem.

Results using this PR (fusion branch) with fusion_multiplier = 1 is shown in the top right panel. It agrees with the analytic solution until the number of macro particle becomes close to zero. Results from the new branch with fusion_multiplier = 10 (lower left) are identical to results obtained using the development branch (upper left) with fusion_multiplier = 1. 

The total number of macro particles in the simulation, shown in the lower right panel of the figure, remains roughly fixed using this PR with fusion_multiplier = 1. The only reason it is not strictly fixed is because the alpha particle produced in the fusion reaction is split into two to maintain local charge conservation.

development_vs_fusion_4x4

### DIFF
--- a/Source/Particles/Collision/BinaryCollision/NuclearFusion/NuclearFusionFunc.H
+++ b/Source/Particles/Collision/BinaryCollision/NuclearFusion/NuclearFusionFunc.H
@@ -181,59 +181,54 @@ public:
                 if (idcpu1[ I1[i1] ]==amrex::ParticleIdCpus::Invalid ||
                     idcpu2[ I2[i2] ]==amrex::ParticleIdCpus::Invalid ) {
                     p_mask[pair_index] = false;
-                    if (max_N == NI1) { i1 += min_N; }
-                    if (max_N == NI2) { i2 += min_N; }
-                    pair_index += min_N;
-                    continue;
                 }
+                else {
 
 #if (defined WARPX_DIM_RZ)
-                /* In RZ geometry, macroparticles can collide with other macroparticles
-                 * in the same *cylindrical* cell. For this reason, collisions between macroparticles
-                 * are actually not local in space. In this case, the underlying assumption is that
-                 * particles within the same cylindrical cell represent a cylindrically-symmetry
-                 * momentum distribution function. Therefore, here, we temporarily rotate the
-                 * momentum of one of the macroparticles in agreement with this cylindrical symmetry.
-                 * (This is technically only valid if we use only the m=0 azimuthal mode in the simulation;
-                 * there is a corresponding assert statement at initialization.) */
-                amrex::ParticleReal const theta = theta2[I2[i2]]-theta1[I1[i1]];
-                amrex::ParticleReal const u1xbuf = u1x[I1[i1]];
-                u1x[I1[i1]] = u1xbuf*std::cos(theta) - u1y[I1[i1]]*std::sin(theta);
-                u1y[I1[i1]] = u1xbuf*std::sin(theta) + u1y[I1[i1]]*std::cos(theta);
+                    /* In RZ geometry, macroparticles can collide with other macroparticles
+                     * in the same *cylindrical* cell. For this reason, collisions between macroparticles
+                     * are actually not local in space. In this case, the underlying assumption is that
+                     * particles within the same cylindrical cell represent a cylindrically-symmetry
+                     * momentum distribution function. Therefore, here, we temporarily rotate the
+                     * momentum of one of the macroparticles in agreement with this cylindrical symmetry.
+                     * (This is technically only valid if we use only the m=0 azimuthal mode in the simulation;
+                     * there is a corresponding assert statement at initialization.) */
+                    amrex::ParticleReal const theta = theta2[I2[i2]]-theta1[I1[i1]];
+                    amrex::ParticleReal const u1xbuf = u1x[I1[i1]];
+                    u1x[I1[i1]] = u1xbuf*std::cos(theta) - u1y[I1[i1]]*std::sin(theta);
+                    u1y[I1[i1]] = u1xbuf*std::sin(theta) + u1y[I1[i1]]*std::cos(theta);
 #endif
 
-                SingleNuclearFusionEvent(
-                    u1x[ I1[i1] ], u1y[ I1[i1] ], u1z[ I1[i1] ],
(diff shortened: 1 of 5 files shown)


Source/Particles/Collision/BinaryCollision/NuclearFusion/NuclearFusionFunc.H
```diff
@@ -182,10 +175,17 @@ public:
...
+                // do not check for collision if a particle's weight was
+                // reduced to zero from a previous collision
+                if (idcpu1[ I1[i1] ]==amrex::ParticleIdCpus::Invalid ||
+                    idcpu2[ I2[i2] ]==amrex::ParticleIdCpus::Invalid ) {
+                    p_mask[pair_index] = false;
+                    if (max_N == NI1) { i1 += min_N; }
+                    if (max_N == NI2) { i2 += min_N; }
+                    pair_index += min_N;
+                    continue;
+                }
```
roelof-groenewald: Moving the reduction in weight to this file rather than doing it during the actual collision is very clever! This also gives an easy solution to the main issue I discussed with @RemiLehe about changing the DSMC to use your new collision scheme. I'll make a PR to update DSMC later tonight following the template in this PR.

roelof-groenewald:
Besides the checksum values the pB11 test checks several aspects of the fusion products against theory, so the checksum values can be reset without worry and if the rest of the analysis for that test passes it would be further motivation that the collision statistics are correct.

Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H
```diff
@@ -198,10 +198,12 @@ public:
...
-                BinaryCollisionUtils::remove_weight_from_colliding_particle(
-                    w1[p_pair_indices_1[i]], idcpu1[p_pair_indices_1[i]], p_pair_reaction_weight[i]);
-                BinaryCollisionUtils::remove_weight_from_colliding_particle(
-                    w2[p_pair_indices_2[i]], idcpu2[p_pair_indices_2[i]], p_pair_reaction_weight[i]);
+                if (t_collision_type == CollisionType::DSMC) { // performed locally for fusion
+                    BinaryCollisionUtils::remove_weight_from_colliding_particle(
+                        w1[p_pair_indices_1[i]], idcpu1[p_pair_indices_1[i]], p_pair_reaction_weight[i]);
+                    BinaryCollisionUtils::remove_weight_from_colliding_particle(
+                        w2[p_pair_indices_2[i]], idcpu2[p_pair_indices_2[i]], p_pair_reaction_weight[i]);
+                }
```
roelof-groenewald: The DSMC routine doesn't use `ParticleCreationFunc.H`. This is just some unfortunate file organization and naming since the nuclear fusion routine was added before DSMC. For DSMC the file `SplitAndScatterFunc.H` inside the DSMC folder handles particle splitting and scattering.
It would be a good topic to add for the hackathon to move this into the NuclearFusion folder.

The inline comment above this line can also be removed but I can't do that as part of the suggestion.
```suggestion
```

Source/Particles/Collision/BinaryCollision/NuclearFusion/NuclearFusionFunc.H
roelof-groenewald: I did have a question though: 
Consider a case where there is 1 low weight particle and 5 high weight particles in a cell. In principle the 1 low weight particle should be considered for a collision with each of the 5 high weight particles (with low collision probability for each), right? So in theory the low weight particle could be scattered multiple times in a single step. However, with the above approach the low weight particle is removed from consideration after the first time it collides, while the new particle created in it's place is not considered for collisions either. Do you not think this discrepancy will lead to problems?

JustinRayAngus: Good thought exercise. 

I think the short answer is that one cannot expect correct physics if you only have 1 particle. For example, see the upper right panel of the figure in the PR description. The physics is poorly captured when the number of particles in the deuterium beam approaches zero. A longer answer is given below.

There are two main things to consider. 1) The total probability that the low weight particle has a collision (sum of probability that it collides with each of the 5 particles) should be small. So, the chance that it has two collisions should be really small = small squared, and the chance that it has three collisions should be super small = small cubed, and so forth. 2) The physics is only ever right on average when using MC methods. To get the right physics you would ideally want something like 10-100 similar low weight particles (assuming we are using fusion_multiplier = 1), in which case that part of the distribution function would properly sample the part of the distribution function that the 5 high weight particles represent. 

If you really do just have 1 particle, and you want better statistics, you can always set the fusion_multiplier to some value larger than 1. Additionally, it would be simple to add a flag that would scale the fusion_multiplier locally by Nmax/Nmin, in which case one could resort to something like the original method. This can be useful if one doesn’t care about managing particle number and weights but wants better product species statistics when the number of particles in a cell for one the colliding species is small.

As for not considering collisions of the particles created during the scattering cycle, I think that is also fine. You can just consider it as a strictly explicit treatment. These particles will begin being included in the scattering on the next scattering cycle.

roelof-groenewald: Thanks for the explanation! The MCC routine (particles colliding with background fluid) also explicitly assumes only one collision per particle per timestep, so I agree this would be okay under the same conditions (sufficiently low collision frequency relative to the collision interval).

roelof-groenewald (approved):
Looks great! Thanks for the algorithm upgrades!

PR #5135, opened 2024-08-14
Update DSMC reaction weight calculation

roelof-groenewald:
This PR updates the DSMC scheme to use the updated reaction weight calculation described by @JustinRayAngus in #5047.

Still to do:
- [x] rerun benchmark calculations to confirm implementation accuracy
- [x] reset checksum values
- add input parameter that sets minimum weight ratio for which particles will be split and recombine particles from the higher weight particle if that value is not exceeded (do this in #4976 after this PR is merged)


Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.H
```diff
@@ -143,56 +140,64 @@ public:
...
             {
-                // c1k : how many times the current particle of species 1 is paired with a particle
-                // of species 2. Same for c2k.
-                const index_type c1k = (k%NI1 < max_N%NI1) ? c1 + 1: c1;
-                const index_type c2k = (k%NI2 < max_N%NI2) ? c2 + 1: c2;
+                // do not check for collision if a particle's weight was
+                // reduced to zero from a previous collision
+                if (idcpu1[ I1[i1] ]==amrex::ParticleIdCpus::Invalid ||
+                    idcpu2[ I2[i2] ]==amrex::ParticleIdCpus::Invalid ) {
+                    p_mask[pair_index] = false;
```
roelof-groenewald, line 147: @JustinRayAngus I've opted to have an `else` bracket below this instead of 
```
                    if (max_N == NI1) { i1 += min_N; }
                    if (max_N == NI2) { i2 += min_N; }
                    pair_index += min_N;
                    continue;
```
in order to avoid code duplication. Just wanted to point that out in case you wanted to make the same switch in the fusion modification PR.

roelof-groenewald:
I reran the physics benchmark used to originally validate the DSMC implementation (comparing against the [Turner 2013](https://pubs.aip.org/aip/pop/article-abstract/20/1/013507/1017414/Simulation-benchmarks-for-low-pressure-plasmas?redirectedFrom=fulltext) MCC benchmark). The plots below show the results for the 1st case using different numbers of kinetic neutral particles:
image

The results match expectation unless the number of neutral particles is too low (neutral density is much higher than plasma density in this problem so good statistics would require more neutral macro-particles than for ions).

roelof-groenewald:
Here are the results of running the benchmark simulations for cases 2 & 3 as well. The results match expectation in all cases.
image

JustinRayAngus (approved):
This all looks good to me.

ax3l (approved):

Issue #5113, opened 2024-08-06
CFL limitation for electrostatic simulations


archermarx:
Hi all,

Following on some issues encountered in [this issue](https://github.com/ECP-WarpX/WarpX/issues/5036), I was wondering if it would be reasonable to develop an option for CFL-limited timesteps in electrostatic simulations. The basic idea would be to limit the timestep such that the fastest particle in the domain cannot cross more than one cell in a single timestep. Failing that, it would be good to at least warn the user if the chosen timestep and particle velocities lead to CFL condition violations. 

The benefits of this would be twofold. First, it would prevent issues like [this one](https://github.com/ECP-WarpX/WarpX/issues/5065) where particles cannot be brought back into the domain with a single periodic shift, leading to segfaults. A more informative error would help significantly, but even better would be to prevent such an issue from arising in the first place. Second, it would potentially accelerate some simulations. In my work, I often have to use overly-pessimistic timesteps to avoid CFL condition violations in case a fast tail of electrons develops or unexpected heating occurs. 

If this seems like a good idea, I can take a stab at implementing this early next week.

ax3l:
Interesting that it does not hit this safe guard,
https://github.com/ECP-WarpX/WarpX/blob/502e5ea31c1334f0083bed64eb8f8202b1100e81/Source/ablastr/particles/DepositCharge.H#L128-L130
which is supposed to give us a cleaner error message than what you encountered
https://github.com/ECP-WarpX/WarpX/issues/5065#issue-2419886602

Maybe we need to add this at another location, too.

ax3l:
The `if (cur_time + dt[0] >= stop_time - 1.e-3*dt[0] || step == numsteps_max-1) {` is the exit criteria, where we just do a half-step push to synchronize times for diagnsotics. You could try to do a `Redistribute()` at the beginning and end of this `if` block to make sure particles are in the right boxes?

ax3l:
The call you might want to use is `HandleParticlesAtBoundaries` https://github.com/ECP-WarpX/WarpX/blob/502e5ea31c1334f0083bed64eb8f8202b1100e81/Source/Evolve/WarpXEvolve.cpp#L226

archermarx:
> Interesting that it does not hit this safe guard,
> 
> https://github.com/ECP-WarpX/WarpX/blob/502e5ea31c1334f0083bed64eb8f8202b1100e81/Source/ablastr/particles/DepositCharge.H#L128-L130
> 
[...]

Issue #5065, opened 2024-07-19
Bug: Invalid memory access occurred during AMReX::GpuDevice::streamSynchronize


archermarx:
Hi,

When running electrostatic simulations on NVIDIA H100 nodes (further details described in https://github.com/ECP-WarpX/WarpX/issues/5036), I encounter the following error during the final simulation step:
```
amrex::Abort::3::CUDA error 700 in file /home/marksta/src/warpx/build/_deps/fetchedamrex-src/Src/Base/AMReX_GpuDevice.cpp
line 648: an illegal memory access was encountered !!!
SIGABRT
```

This points to this section of code
#### `Amrex/Src/Base/AMReX_GpuDevice.cpp`::648

```cpp
void
Device::streamSynchronize () noexcept
{
#ifdef AMREX_USE_SYCL
    auto& q = streamQueue();
    try {
        q.wait_and_throw();
    } catch (sycl::exception const& ex) {
        amrex::Abort(std::string("streamSynchronize: ")+ex.what()+"!!!!!");
    }
#else
    AMREX_HIP_OR_CUDA( AMREX_HIP_SAFE_CALL(hipStreamSynchronize(gpuStream()));,
                       AMREX_CUDA_SAFE_CALL(cudaStreamSynchronize(gpuStream())); ) // error is occurring here I think
#endif
}
```

Looking at the [attached backtrace file](https://github.com/user-attachments/files/16316195/backtrace.txt), I can see that this call ultimately originates from the final half velocity push in `WarpXEvolve.cpp`

#### `WarpXEvolve.cpp::177`
```cpp
  if (cur_time + dt[0] >= stop_time - 1.e-3*dt[0] || step == numsteps_max-1) {
    // At the end of last step, push p by 0.5*dt to synchronize
    FillBoundaryE(guard_cells.ng_FieldGather);
    FillBoundaryB(guard_cells.ng_FieldGather);
    if (fft_do_time_averaging)
    {
        FillBoundaryE_avg(guard_cells.ng_FieldGather);
        FillBoundaryB_avg(guard_cells.ng_FieldGather);
    }
    UpdateAuxilaryData();
    FillBoundaryAux(guard_cells.ng_UpdateAux);
    for (int lev = 0; lev <= finest_level; ++lev) {
        mypc->PushP(lev, 0.5_rt*dt[lev],   // error starts here
                    *Efield_aux[lev][0],*Efield_aux[lev][1],
                    *Efield_aux[lev][2],
[...]

PR #1751, opened 2021-03-02
Synchronize velocity for diagnostics

dpgrote:
This PR allows the synchronization in time of the particle velocities and positions when generating diagnostics. Without this option, the particle velocities will lag behind the position by a half time step. This adds the boolean input parameter `warpx.synchronize_velocity_for_diagnostics` to turn on this option, defaulting to false.

There are several pieces to this PR:
- Changes to `MultiDiagnostic` and `MultiReducedDiags` adding routines to check if any diagnostics will be done
- Adds a call to `PushP` to just before the diagnostics are done (to get the updated fields from the electrostatic calculation)
- Add the appropriate documentation

What `Evolve` does is if the synchronization is to be done, advance the velocity a half step just before the diagnostics and sets `is_synchronized=true`. Then at the start of the next step, if `is_synchronized` is true, push the velocities back a half step to be ready for the full leap frog advance.

Comments:
- Is the documentation in the correct place in parameters.rst?
- The reduced diagnostics could perhaps use the new DoDiags method instead of accessing `m_intervals` in its ComputeDiags.
- This PR leaves the original PushP unchanged, even though it is possibly buggy. That PushP fetches the fields, but uses the particle positions before the particle boundary conditions have been applied, leading to a possible out of bounds reference. Also, that PushP may not be consistent with the backwards PushP since the fields may had changed. Comments are added to the code to note this potential problem. I avoided changing this since it breaks many CI tests.

PR #5136, opened 2024-08-14
Add call to HandleParticlesAtBoundaries in WarpXEvolve.cpp to prevent oob segfaults

archermarx:
Fixes bug found here https://github.com/ECP-WarpX/WarpX/issues/5065 and partially addresses https://github.com/ECP-WarpX/WarpX/issues/5113

This PR just adds a call to HandleParticlesAtBoundaries to the synchronization step part of WarpXEvolve.cpp to prevent unhelpful memory access errors when particles travel too far in one step.

Before, this error would be thrown:
```
amrex::Abort::3::CUDA error 700 in file /home/marksta/src/warpx/build/_deps/fetchedamrex-src/Src/Base/AMReX_GpuDevice.cpp
line 648: an illegal memory access was encountered !!!
SIGABRT
```

Now, no error is thrown and the simulation finishes successfully.

Ideally, we would still like to warn users about CFL violations in ES simulations, or adapt the timestep to account for a user-specified CFL number (which may be greater than 1 for semi-implicit schemes). I'll keep looking at that.

### DIFF
--- a/Source/Evolve/WarpXEvolve.cpp
+++ b/Source/Evolve/WarpXEvolve.cpp
@@ -172,32 +172,6 @@ WarpX::Evolve (int numsteps)
             // B : guard cells are NOT up-to-date
         }
 
-        // TODO: move out
-        if (evolve_scheme == EvolveScheme::Explicit) {
-            if (cur_time + dt[0] >= stop_time - 1.e-3*dt[0] || step == numsteps_max-1) {
-                // At the end of last step, push p by 0.5*dt to synchronize
-                const int num_moved = MoveWindow(step+1, is_synchronized);
-                HandleParticlesAtBoundaries(step, cur_time, num_moved);
-                FillBoundaryE(guard_cells.ng_FieldGather);
-                FillBoundaryB(guard_cells.ng_FieldGather);
-                if (fft_do_time_averaging)
-                {
-                    FillBoundaryE_avg(guard_cells.ng_FieldGather);
-                    FillBoundaryB_avg(guard_cells.ng_FieldGather);
-                }
-                UpdateAuxilaryData();
-                FillBoundaryAux(guard_cells.ng_UpdateAux);
-                for (int lev = 0; lev <= finest_level; ++lev) {
-                    mypc->PushP(lev, 0.5_rt*dt[lev],
-                                *Efield_aux[lev][0],*Efield_aux[lev][1],
-                                *Efield_aux[lev][2],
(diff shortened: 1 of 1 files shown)


archermarx:
Looks like the `qed_schwinger` and `Point_of_contact_EB_3d FAILED` tests fail, but only by a bit (errors on order of 0.01 relative tolerance vs the requested 0.001). I could wrap this line in `#if not defined(WARPX_QED) || not defined (WARPX_EB)`, perhaps

Also, the `dsmc_1d` test fails now, with relative errors of O(1e-8) as opposed to 1e-9.

ax3l:
I think one could reset these checksums. The test changes the order of particles that go into the checksum by redistributing them. Long floating point sums are not commutative.

archermarx:
How could I go about doing that?

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -176,6 +176,8 @@ WarpX::Evolve (int numsteps)
         if (evolve_scheme == EvolveScheme::Explicit) {
```
dpgrote: I would suggest a more correct way of fixing this is to move this whole block of code to the end of the time step after the electrostatic field calculation (and before the diagnostics). This is suggested in the ancient PR #1751. This fixes two things. At that point the particle boundary conditions would have been handled anyway, and also the fields updated to the correct time level for the velocity push.

archermarx: Ok, that seems reasonable.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -262,6 +238,29 @@ WarpX::Evolve (int numsteps)
             ExecutePythonCallback("afterEsolve");
         }
 
+        if (evolve_scheme == EvolveScheme::Explicit) {
+            if (cur_time + dt[0] >= stop_time - 1.e-3*dt[0] || step == numsteps_max-1) {
```
dpgrote:
```suggestion
            if (cur_time >= stop_time - 1.e-3*dt[0] || step == numsteps_max-1) {
```
Since `cur_time` is already updated.

Source/Evolve/WarpXEvolve.cpp
dpgrote: Unfortunately, this causes many CI tests to fail. I can work through the tests and update the benchmarks, making sure that nothing is actually broken. This will take some time.

archermarx:
Closing this for now. May create a fresh PR in the future.

PR #5141, opened 2024-08-15
Containerfile

sinha-r:
Containerfile to run WarpX at NERSC 

# To Do
- [x] GPU-aware MPI is needed: we need to use `shifter` instead of podman
- [x] heFFTe has been disabled as of now due to compilation errors (this can be skipped for now and added in a follow-up PR later on, we do not use heFFTe extensively yet)

### DIFF
--- a/Tools/machines/perlmutter-nersc/Containerfile
+++ b/Tools/machines/perlmutter-nersc/Containerfile
@@ -1,0 +1,151 @@
+# Base System and Essential Tools Installation
+FROM nvidia/cuda:12.6.0-devel-ubuntu22.04 AS base
+
+# Set up environment variables
+ENV DEBIAN_FRONTEND=noninteractive \
+    SW_DIR=/opt/software \
+    FORCE_UNSAFE_CONFIGURE=1
+
+# Install essential system dependencies including MPI libraries
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    autoconf \
+    build-essential \
+    ca-certificates \
+    coreutils \
+    curl \
+    environment-modules \
+    gfortran \
+    git \
+    openssh-server \
+    python3 \
+    python3-pip \
+    python3-dev \
+    python3-venv \
+    unzip \
+    vim \
+    libmpich-dev \
+    cmake \
+    libblas-dev \
+    liblapack-dev \
+    g++ \
+    cuda-toolkit-12-2 \
+    pkg-config \
+    libbz2-dev \
+    zlib1g-dev \
+    libpng-dev \
+    libzstd-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install c-blosc from source
+FROM base AS c-blosc
+
+RUN git clone -b v1.21.1 https://github.com/Blosc/c-blosc.git /tmp/c-blosc && \
+    cd /tmp/c-blosc && \
+    mkdir build && cd build && \
+    cmake -DCMAKE_INSTALL_PREFIX=/usr .. && \
+    make -j$(nproc) && \
+    make install && \
+    rm -rf /tmp/c-blosc
+
+# Install ADIOS2 from source
+FROM base AS adios2
+
+# Ensure c-blosc is installed before ADIOS2
+COPY --from=c-blosc /usr /usr
+
+# Verify the location of Blosc library
+RUN find /usr -name 'libblosc*'
+
+# Ensure Blosc library paths are correctly configured
+ENV BLOSC_LIBRARY=/usr/lib/libblosc.so.1.21.1
+ENV BLOSC_INCLUDE_DIR=/usr/include
+
+# Install ADIOS2
+RUN git clone -b v2.8.3 https://github.com/ornladios/ADIOS2.git /tmp/adios2 && \
+    cd /tmp/adios2 && \
+    mkdir build && cd build && \
+    cmake -DADIOS2_USE_Blosc=ON \
+          -DBLOSC_LIBRARY=${BLOSC_LIBRARY} \
+          -DBLOSC_INCLUDE_DIR=${BLOSC_INCLUDE_DIR} \
+          -DADIOS2_USE_Fortran=OFF \
+          -DADIOS2_USE_Python=OFF \
+          -DADIOS2_USE_ZeroMQ=OFF \
+          -DADIOS2_USE_BZip2=ON \
+          -DADIOS2_USE_ZFP=OFF \
+          -DADIOS2_USE_SZ=OFF \
+          -DADIOS2_USE_MGARD=OFF \
+  -DADIOS2_USE_PNG=ON \
+          -DCMAKE_INSTALL_PREFIX=/usr .. && \
+    make -j$(nproc) && \
+    make install && \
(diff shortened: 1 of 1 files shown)


Tools/machines/perlmutter-nersc/Containerfile
```diff
@@ -0,0 +1,151 @@
...
+    gfortran \
+    git \
+    openssh-server \
+    python3 \
+    python3-pip \
+    python3-dev \
+    python3-venv \
+    unzip \
+    vim \
+    libmpich-dev \
```
ax3l: Do we need to do anything to get GPU-Aware MPI on Perlmutter to run?
Vanilla without containers, I have to set this to tell the Cray compiler wrappers to get it right:
```
# necessary to use CUDA-aware MPI and run a job
export CRAY_ACCEL_TARGET=nvidia80
```
Since we compile with GCC here, I am not sure how to do it.

Tools/machines/perlmutter-nersc/Containerfile
```diff
@@ -0,0 +1,151 @@
+# Base System and Essential Tools Installation
+FROM nvidia/cuda:12.6.0-devel-ubuntu22.04 AS base
+
+# Set up environment variables
+ENV DEBIAN_FRONTEND=noninteractive \
+    SW_DIR=/opt/software \
+    FORCE_UNSAFE_CONFIGURE=1
```
ax3l, line 21:
```suggestion
    FORCE_UNSAFE_CONFIGURE=1

# Perlmutter A100 compilation optimization
ENV AMREX_CUDA_ARCH=8.0 \
    CUDAARCHS=80 \
    CXXFLAGS="-march=znver3" \
    CFLAGS="-march=znver3"
```

Tools/machines/perlmutter-nersc/Containerfile
```diff
@@ -0,0 +1,151 @@
...
+# Install c-blosc from source
+FROM base AS c-blosc
+
+RUN git clone -b v1.21.1 https://github.com/Blosc/c-blosc.git /tmp/c-blosc && \
+    cd /tmp/c-blosc && \
+    mkdir build && cd build && \
+    cmake -DCMAKE_INSTALL_PREFIX=/usr .. && \
+    make -j$(nproc) && \
+    make install && \
+    rm -rf /tmp/c-blosc
```
ax3l: Shorter:
```suggestion
    cmake -S /tmp/c-blosc -B /tmp/c-blosc-build \
      -DCMAKE_INSTALL_PREFIX=/usr .. && \
    cmake --build /tmp/c-blosc-build --target install -j$(nproc) && \
    rm -rf /tmp/c-blosc*
```

Tools/machines/perlmutter-nersc/Containerfile
```diff
@@ -0,0 +1,151 @@
+# Base System and Essential Tools Installation
```
ax3l, line 11: I think traditionally we called this file `Dockerfile` not `Containerfile`, is that still the case?

Tools/machines/perlmutter-nersc/Containerfile
```diff
@@ -0,0 +1,151 @@
...
+    setuptools
+
+# Set up the environment for the virtual environment
+ENV PATH="/opt/venv/bin:${PATH}"
+
+# Set up entrypoint
+ENTRYPOINT ["/bin/bash", "-c"]
+
+# Default command
+CMD ["/bin/bash"]
```
ax3l, line 167: I think we still need to `activate` the venv on entry, do we?

Tools/machines/perlmutter-nersc/Containerfile
```diff
@@ -0,0 +1,151 @@
+# Base System and Essential Tools Installation
+FROM nvidia/cuda:12.6.0-devel-ubuntu22.04 AS base
```
ax3l, line 12: To test: we could use `-devel` to build all dependencies but then copy their artifacts in the `-runtime` image. To check if those are much smaller or not.

Tools/machines/perlmutter-nersc/Containerfile
ax3l: Ah, convention we use because we build with `podman`.

Tools/machines/perlmutter-nersc/Containerfile
```diff
@@ -0,0 +1,151 @@
...
+
+RUN git clone -b v2.4.0 https://github.com/icl-utk-edu/heffte.git /tmp/heffte #&& \
+   # cd /tmp/heffte && \
+    #mkdir build && cd build && \
+    # Disable LTO for CUDA builds
+ #   CXXFLAGS="-fno-lto" CFLAGS="-fno-lto"
+  #  cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_STANDARD=17 -DHeffte_ENABLE_CUDA=ON -DCMAKE_INSTALL_PREFIX=/usr .. && \
+   # make -j$(nproc) && \
+    #make install && \
+       #rm -rf /tmp/heffte
```
ax3l: This is what I do for Perlmutter GPUs:
https://github.com/ECP-WarpX/WarpX/blob/dfc89a4af43a774efa828ad965498fe6421239d2/Tools/machines/perlmutter-nersc/install_gpu_dependencies.sh#L110-L146

Tools/machines/perlmutter-nersc/Containerfile
```diff
@@ -0,0 +1,151 @@
...
+    rm -rf /tmp/lapackpp
+
+# Install heFFTe
+FROM base AS heffte
+
+RUN git clone -b v2.4.0 https://github.com/icl-utk-edu/heffte.git /tmp/heffte #&& \
+   # cd /tmp/heffte && \
+    #mkdir build && cd build && \
+    # Disable LTO for CUDA builds
+ #   CXXFLAGS="-fno-lto" CFLAGS="-fno-lto"
```
ax3l: This comment breaks the `RUN` command,
I would do:
```suggestion
    CXXFLAGS="-fno-lto" CFLAGS="-fno-lto" \
```

Tools/machines/perlmutter-nersc/Containerfile
ax3l: With `ENV` above that could be sufficient, because `python`, `pip` et al. will then point to the venv. Otherwise if we encounter issues down the road we could run an activation script that sources `/opt/venv/bin/activate`.

ax3l (approved):
This is a developer container for now.

Will expand in follow-up PRs to create also a user-image from it that can be used right away without compiling WarpX inside it  first.

Tools/machines/perlmutter-nersc/Containerfile
ax3l: Resolved: heFFTe not used anymore.

Tools/machines/perlmutter-nersc/Containerfile
ax3l: NERSC info: Podman GPU-aware runs need to be based on CUDA 11.7 at this point.

PR #5143, opened 2024-08-16
AnyFFT: real/imag helpers

ax3l:
Add helpers to access real and imaginary parts of library-dependent complex types.
Use like this:
```c++
using namespace ablastr::math::anyfft;

Complex c = ...;
real(c) = 1;
imag(c) = -1;
```

Needed for GPU support in https://github.com/ECP-WarpX/impactx/pull/627


ax3l:
I actually just need a `mult()` function #5146

PR #5145, opened 2024-08-16
[WIP] FFTW: Compatible with `std::complex`

ax3l:
FFTW's complex numbers are compatible with C99 are compatible with https://en.cppreference.com/w/cpp/numeric/complex:
https://www.fftw.org/fftw3_doc/Complex-numbers.html

Use this to have more default defined intrinsic operations on those types.


ax3l:
`SpectralFieldDataRZ.cpp` uses unabstracted cuFFT/FFTW/etc calls.

PR #5148, opened 2024-08-16
potential fix to varying results for collisionXZ and collisionXYZ CI tests

JustinRayAngus:
The file passed to checksumAPI.evaluate_checksum() at the end of the analysis scripts for the collisionXZ and collisionXYZ CI tests currently uses a file name (fn) that comes from the end of a loop over filenames. This does not always return the same name. The intended file here is the last time step, which is given by last_fn = sys.argv[1]. 

This PR changes it such that last_fn is used in checksumAPI.evaluate_checksum() for these two CI tests. This may fix the random failures of these tests experienced recently.

- [x] also update collisionZ and collisionRZ

### DIFF
--- a/Examples/Tests/collision/analysis_collision_1d.py
+++ b/Examples/Tests/collision/analysis_collision_1d.py
@@ -26,8 +26,8 @@ sys.path.insert(1, '../../../../warpx/Regression/Checksum/')
 import checksumAPI
 
 # this will be the name of the plot file
-fn = sys.argv[1]
-ds = yt.load(fn)
+last_fn = sys.argv[1]
+ds = yt.load(last_fn)
 data = ds.covering_grid(level = 0, left_edge = ds.domain_left_edge, dims = ds.domain_dimensions)
 
 # carbon 12 ion (mass = 12*amu - 6*me)
@@ -123,4 +123,4 @@ print('tolerance = ', tolerance);
 assert error < tolerance
 
 test_name = os.path.split(os.getcwd())[1]
-checksumAPI.evaluate_checksum(test_name, fn, rtol=2.5e-3)
+checksumAPI.evaluate_checksum(test_name, last_fn)
--- a/Examples/Tests/collision/analysis_collision_2d.py
+++ b/Examples/Tests/collision/analysis_collision_2d.py
@@ -114,4 +114,4 @@ post_processing_utils.check_random_filter(last_fn, random_filter_fn, random_frac
                                           dim, species_name)
 
 test_name = os.path.split(os.getcwd())[1]
-checksumAPI.evaluate_checksum(test_name, last_fn, rtol=1e-1)
+checksumAPI.evaluate_checksum(test_name, last_fn)
--- a/Examples/Tests/collision/analysis_collision_2d.py
+++ b/Examples/Tests/collision/analysis_collision_2d.py
@@ -26,7 +26,6 @@
 import glob
 import math
 import os
-import re
 import sys
 
 import numpy
--- a/Examples/Tests/collision/analysis_collision_2d.py
+++ b/Examples/Tests/collision/analysis_collision_2d.py
@@ -55,12 +55,10 @@ a =  0.04330638981264072
 b = -0.11588277796546632
 
 last_fn = sys.argv[1]
-# Remove trailing '/' from file name, if necessary
-last_fn.rstrip('/')
-# Find last iteration in file name, such as 'test_name_plt000001' (last_it = '000001')
-last_it = re.search('\d+', last_fn).group()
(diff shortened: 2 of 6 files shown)


RemiLehe:
Thanks for this PR!

I think this makes sense: as you pointed out, we create a list of files using `glob.glob` and we then loop over this list and run the checksum on the last element of the list.
The [documentation for `glob.glob`](https://docs.python.org/3/library/glob.html#glob.glob) suggests that the order of the files in the list is not necessarily deterministic:
```
Whether or not the results are sorted depends on the file system.
```

JustinRayAngus:
Even on the same file system, it seems to not be deterministic. I ran collisionXYZ twice on LLNL's Ruby cluster and printed out the index in the loop for each run. Below are results

**test0:**
j =  150
j =  20
j =  120
j =  50
j =  140
j =  30
j =  130
j =  40
j =  100
j =  70
j =  0
j =  80
j =  110
j =  60
j =  90
j =  10
fn =  diags/diag1000010

**test1:**
j =  70
j =  100
j =  0
j =  80
j =  60
j =  110
j =  90
j =  10
j =  20
j =  150
j =  50
j =  120
j =  30
j =  140
j =  40
j =  130
fn =  diags/diag1000130

The results are the same if I run the analysis script repeated times on the same output data. The order is unique for a given simulation, but that order varies for different simulations.

ax3l:
Good find!!
Can you also update the collisionZ and collisionRZ analysis scripts with this PR pls? :)

EZoni:
For `collisionZ` it would be necessary to rebase on `development`, since I merged a tolerance change just before I saw this.

EZoni (approved):
Looks good to me. Some of this might change again anyways in the near future, because we might manage the diag files handling a bit differently when we migrate to CTest. We might need to reset the checksums again, if they were reset earlier based on the wrong data, but let's see once the CI checks finish.

RemiLehe:
Btw, as a way to estimate how to set the tolerance for these tests: here are some checksum results for `collisionXZ` that I obtained with different random seeds. (In this case, with a MacBook, using only 1 OpenMP thread and 1 MPI rank. Also: I used the current branch, so as to make sure that the checksum is always calculated from the same file.)

- **warpx.random_seed = 1**
```
  "electron": {
    "particle_momentum_x": 1.0462672745456273e-19,
    "particle_momentum_y": 1.0203237631739066e-19,
    "particle_momentum_z": 1.018305242305984e-19,
    "particle_position_x": 2655181.8485983927,
    "particle_position_y": 2658393.0923468275,
    "particle_weight": 1.7256099431746894e+26
  },
  "ion": {
    "particle_momentum_x": 2.479805850018366e-19,
    "particle_momentum_y": 2.2519951548539823e-19,
    "particle_momentum_z": 2.292851189671145e-19,
    "particle_position_x": 2670787.727335387,
    "particle_position_y": 2656231.3310583937,
    "particle_weight": 1.7256099431746894e+26
  }
```

- **warpx.random_seed=2**
```
  "electron": {
    "particle_momentum_x": 1.0318101035585444e-19,
    "particle_momentum_y": 1.0113937180250944e-19,
    "particle_momentum_z": 1.0216943211513149e-19,
    "particle_position_x": 2671915.795842967,
    "particle_position_y": 2649400.7099330146,
    "particle_weight": 1.7256099431746894e+26
  },
  "ion": {
    "particle_momentum_x": 2.5183516479604383e-19,
    "particle_momentum_y": 2.2879384214291196e-19,
    "particle_momentum_z": 2.26935784287823e-19,
    "particle_position_x": 2671733.2592584365,
    "particle_position_y": 2646929.9006488123,
    "particle_weight": 1.7256099431746894e+26
  }
```

- **warpx.random_seed=3**
```
  "electron": {
    "particle_momentum_x": 1.0525675338367951e-19,
    "particle_momentum_y": 1.0416982914156779e-19,
    "particle_momentum_z": 1.0251397034510792e-19,
    "particle_position_x": 2681379.754814725,
    "particle_position_y": 2661579.067538915,
    "particle_weight": 1.7256099431746894e+26
  },
  "ion": {
    "particle_momentum_x": 2.5065080898247363e-19,
    "particle_momentum_y": 2.2659107634599083e-19,
    "particle_momentum_z": 2.298035507996226e-19,
    "particle_position_x": 2670953.5604058728,
    "particle_position_y": 2666640.3124145865,
    "particle_weight": 1.7256099431746894e+26
  }
```

The fluctuations with different random seeds seem to be at the 1-3 % level.

ax3l (approved):

JustinRayAngus:
What does this mean:
ECP-WarpX.CI Expected — Waiting for status to be reported

RemiLehe:
@JustinRayAngus  I think that it's a random hiccup in Github Actions. I merged now.

PR #5158, opened 2024-08-21
Fix: heFFTe on Perlmutter GPU

ax3l:
heFFTe 2.4.0 usage with NVCC 12.2 on Perlmutter fails to compile WarpX once its public header is included, due to unguarded include of a CPU AVX intrincics header in its stock backend.

We do not need a fast stock CPU backend if we compile for CUDA, so we patch it out now to avoid the issue. A future (post v2.4.0) release of heFFTe will ship our fix for this, until then we post-install patch heFFTe on Perlmutter for GPU builds: https://github.com/icl-utk-edu/heffte/pull/54


aeriforme (approved):
I wholeheartedly approve :heart:

PR #5161, opened 2024-08-21
Add differential luminosity reduced diagnostic

aeriforme:
This PR add a reduce diagnostic that computes the differential luminosity (or luminosity spectrum) in the center of mass energy, $d\mathcal{L}/dE_{com}$, between two species (assumed to be two colliding beams). 

For each cell of the spatial grid, we loop through all the possible pair of particles $(p_1, p_2)$ from the first and second species, respectively. For each of these pairs, we compute the center of mass energy: 

$E_{com} = \sqrt{m_1^2 + m_2^2 +2 m_1 m_2 \gamma_1 \gamma_2 (1 - \boldsymbol{\beta}_1  \cdot \boldsymbol{\beta}_2 )}c^2$

and the contribution to the luminosity per crossing:
$d\mathcal{L} = \int c\sqrt{(\boldsymbol{\beta}_1 - \boldsymbol{\beta}_2)^2 - (\boldsymbol{\beta}_1 \times \boldsymbol{\beta}_2 )^2} f_1 f_2 d\boldsymbol{x} d\boldsymbol{v} dt$

divided by $dE_{com}$.
The result is a histogram $dL/dE_{com}$ as a function of $E_{com}$. 

In the above formulas: 
* $m_1, m_2$ mass of the species
* $f_1(\boldsymbol{x}, \boldsymbol{v}, t), f_2(\boldsymbol{x}, \boldsymbol{v}, t)$ the distribution functions of the two particles 
* $\boldsymbol{\beta}_1, \boldsymbol{\beta}_2$ the velocities of the two particles normalized to $c$
* $\gamma_1, \gamma_2$ the Lorentz factors of the two particles 

TO DO:

- [x] check formulas above. Note from @RemiLehe : I added a factor `c` in the above formula. The formulas look consistent with e.g. this paper: https://www.osti.gov/biblio/836235
- [x] debug: something is wrong, the values are way too big
- [x] documentation
- [x] finalize CI test, compare with theoretical formula
- [ ] ~~check normalization: should we divide by the total luminosity at every timestep?~~
- [x] save bins in scientific format
- [x] fix comments in the code
- [ ] compare with GUINEA-PIG
- [x] Make sure this works on GPU


Source/Diagnostics/ReducedDiags/DifferentialLuminosity.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+                        Real u1_cross_u2_x = u1y[j_1]*u2z[j_2] - u1z[j_1]*u2y[j_2];
+                        Real u1_cross_u2_y = u1z[j_1]*u2x[j_2] - u1x[j_1]*u2z[j_2];
+                        Real u1_cross_u2_z = u1x[j_1]*u2y[j_2] - u1y[j_1]*u2x[j_2];
+
+                        Real v1_cross_v2_square = (u1_cross_u2_x*u1_cross_u2_x + u1_cross_u2_y*u1_cross_u2_y + u1_cross_u2_z*u1_cross_u2_z) / (gamma1*gamma1*gamma2*gamma2);
+
+                        Real radicand = v1_minus_v2_square - v1_cross_v2_square / c2;
+
+                        Real dL_dEcom = std::sqrt( radicand ) * w1[j_1] * w2[j_2] / dV / bin_size; // s^-1 m^-2 J^-1
```
RemiLehe: Given the expression of the luminosity:
$d\mathcal{L} = \int \sqrt{(\boldsymbol{\beta}_1 - \boldsymbol{\beta}_2)^2 - (\boldsymbol{\beta}_1 \times \boldsymbol{\beta}_2 )^2} f_1 f_2 d\boldsymbol{x} d\boldsymbol{v} dt$
I think that this should be multiplied by `dt`.

Accordingly, the results should be in `m^-2 . J^-1`, instead of `s^-1 m^-2 J^-1`.
It only becomes `s^-1 m^-2 J^-1` once we multiply by the repetition rate of the collider (i.e. number of collisions per second).

Source/Diagnostics/ReducedDiags/DifferentialLuminosity.cpp
```diff
@@ -0,0 +1,287 @@
...
+    const ParticleReal m2 = species_2.getMass();
+
+    // zero-out old data on the host
+    //std::fill(m_data.begin(), m_data.end(), amrex::Real(0.0));
+    //amrex::Gpu::DeviceVector< amrex::Real > d_data( m_data.size(), 0.0 );
+    //amrex::Real* const AMREX_RESTRICT dptr_data = d_data.dataPtr();
+
+    amrex::Gpu::DeviceVector< amrex::Real > d_data( m_data.size() );
+    amrex::Gpu::copy(amrex::Gpu::hostToDevice,
+        m_data.begin(), m_data.end(), d_data.begin());
```
RemiLehe: We should move this to the constructor

aeriforme: Uhm, why? This copies `m_data` to the GPU, without zeroing it out, right?
The initialization to zero is only in the constructor, where:
`m_data.resize(m_bin_num,0.0_rt);`

RemiLehe: This is what I meant:
https://github.com/ECP-WarpX/WarpX/pull/5161/commits/1402ba2124da0cdfb9068488658953e2069e1aa1

RemiLehe (approved):

Source/Diagnostics/ReducedDiags/DifferentialLuminosity.cpp
```diff
@@ -0,0 +1,272 @@
...
+                        Real const u1_cross_u2_y = u1z[j_1]*u2x[j_2] - u1x[j_1]*u2z[j_2];
+                        Real const u1_cross_u2_z = u1x[j_1]*u2y[j_2] - u1y[j_1]*u2x[j_2];
+
+                        Real const v1_cross_v2_square = (u1_cross_u2_x*u1_cross_u2_x + u1_cross_u2_y*u1_cross_u2_y + u1_cross_u2_z*u1_cross_u2_z) / (gamma1*gamma1*gamma2*gamma2);
+
+                        Real const radicand = v1_minus_v2_square - v1_cross_v2_square / c2;
+
+                        Real const dL_dEcom = std::sqrt( radicand ) * w1[j_1] * w2[j_2] / dV / bin_size * dt; // m^-2 J^-1
+
+                        amrex::HostDevice::Atomic::Add(&dptr_data[bin], dL_dEcom);
```
ax3l, line 249: This works and is safe.

We can probably make it faster on CPUs for the OMP loop (where this now uses an `omp atomic update`) by using a copy of `d_data` per thread (block/tile), doing non-atomic adds into the histogram per particle, and reducing (sum) the copies after the OMP loop. @atmyers do you have a good construct for this? (We need something similar in ImpactX right now :D)

atmyers: The atomic-free version of this is implemented for amrex::DenseBins here: https://github.com/AMReX-Codes/amrex/blob/development/Src/Particle/AMReX_DenseBins.H#L342

Maybe with a little refactoring we could reuse this here?

ax3l: Yes, and we could maybe also use it here for the 1D charge binning:
https://github.com/ECP-WarpX/impactx/pull/627

PR #5162, opened 2024-08-21
Add callback to write values to external field multifabs

roelof-groenewald:
Currently a standard practice for writing initial or external fields using Python is to use the `beforeInitEsolve` callback. However, this creates a problem when using electrostatic solvers since the external fields (`B/Efield_fp_external`) are then not added to the `B/Efield_fp` multifabs until after the first step (in the regular PIC loop) so the first step particle pushing does not use the external fields. This PR fixes that problem by truly handling external fields loaded from Python (through the new callback) equivalently to external fields loaded from file or from the parser.

### DIFF
--- a/Python/pywarpx/callbacks.py
+++ b/Python/pywarpx/callbacks.py
@@ -25,6 +25,7 @@ extra reference to the method's object is saved.
 
 Functions can be called at the following times:
 
+* ``loadExternalFields``: during ``WarpX::LoadExternalFields`` to write ``B/Efield_fp_external`` values
 * ``beforeInitEsolve``: before the initial solve for the E fields (i.e. before the PIC loop starts)
 * ``afterinit``: immediately after the init is complete
 * ``beforeEsolve``: before the solve for E fields
@@ -276,6 +277,7 @@ class CallbackFunctions(object):
 # =============================================================================
 
 callback_instances = {
+    "loadExternalFields": {},
     "beforeInitEsolve": {},
     "afterInitEsolve": {},
     "afterinit": {},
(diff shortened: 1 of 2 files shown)


dpgrote:
I see in an older version of WarpX, that the routine `AddExternalFields` was called right after the init Esolve, but it is not there now. Is there a reason that was removed? This would also fix the issue.

clarkse (approved):
This seems like a reasonable place to have a callback when loading external fields. It looks fine to me.

roelof-groenewald:
@dpgrote that call was removed in #5104 - which did actually create this problem but also solved another as described in that PR. With this PR now I believe things will be fully fixed (at least until we discover the next bug).

dpgrote:
This is good and I can approve, but I have a suggestion as a possible alternative. Could the call to `AddExternalFields` that is now made in `InitLevelData` be moved to after the init E solve in `InitData`? This should fix this issue and not break what was fixed in PR #5104. This would also make it consistent with what is done in `Evolve`.

roelof-groenewald:
> Could the call to `AddExternalFields` that is now made in `InitLevelData` be moved to after the init E solve in `InitData`? This would also make it consistent with what is done in `Evolve`.

I agree having the consistency with `Evolve` is good, so I moved the `AddExternalFields` call as you suggested. 

I still think it is cleaner to have a new callback in the same function where external fields specified through other means are read (`LoadExternalFields`) rather than using `beforeInitESolve`. Besides just being more organized it is also needed for restart consistency since `beforeInitESolve` is not executed on restarts but `LoadExternalFields` is.

dpgrote (approved):
Looks good!

PR #5169, opened 2024-08-26
Improvements to maxParticleVelocity

archermarx:
This PR improves the functionality of `WarpXParticleContainer::maxParticleVelocity` in two ways.

1. The function can now make use of the GPU for the reduction stage, just like `meanParticleVelocity`
2. The function now actually returns velocities, wheras before it returned momenta.

This PR also adds a version to `MultiParticleContainer`, so the maximum velocity among all particle species can be obtained.

As far as I can tell, this function is not used anywhere at the moment. I am however working on another PR which will make use of it, so I thought i would go ahead and send this off for feedback now.


archermarx:
A question for those more experienced. Is the a reason we need to call OpenMP manually when we're not using GPU? My understanding is that AMReX should be able to use OpenMP just fine within the `ParticleReduce` function, but I'm basing a lot of this code on the `getMeanParticleVelocity` function above this, which adopts this strategy.

archermarx:
Not sure why CUDA is failing to compile. Can compile with NVCC on my machine without errors.

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -1420,27 +1420,26 @@ std::array<ParticleReal, 3> WarpXParticleContainer::meanParticleVelocity(bool lo
...
+    using PType = typename WarpXParticleContainer::SuperParticleType;
+    amrex::ReduceOps<amrex::ReduceOpMax> reduce_ops;
+    const auto reduce_res = amrex::ParticleReduce<amrex::ReduceData<amrex::ParticleReal>>(
+            *this,
+            [=] AMREX_GPU_HOST_DEVICE(const PType &p) noexcept -> amrex::GpuTuple<amrex::ParticleReal>
+            {
+                const auto ux = p.rdata(PIdx::ux);
+                const auto uy = p.rdata(PIdx::uy);
+                const auto uz = p.rdata(PIdx::uz);
+                return {(ux*ux + uy*uy + uz*uz) * inv_clight_sq,};
```
ax3l: NVCC issue: We have to double check if you want to use a different syntax for the return line or skip the training return type to make it happy.

Maybe try to use `amrex::makeTuple(...)`?

ax3l: @RemiLehe also spotted an overload of this without a tuple (for scalars) - will post

archermarx:
Thanks for the refactor. It seems all tests are green. Anything else needed before merge? @RemiLehe @ax3l

RemiLehe (approved):

RemiLehe:
Thanks for the PR @archermarx

PR #5170, opened 2024-08-26
minor adjustments to comments in implicit solver classes

JustinRayAngus:
Using consistent notation for fields Eg and Bg in comments.


dpgrote (approved):
Looks good!

PR #5171, opened 2024-08-26
Generalization of WarpXSolverVec class used by implicit solvers

JustinRayAngus:
The WarpXSovlerVec class is a wrapper around a Vector of pointers to MutliFabs that contains basic math operators and functionality needed to interact with nonlinear solvers in WarpX and linear solvers in AMReX, such as GMRES. The size of the Vector is the number of amr levels. The original implementation of this class was hard-wired to only work for the electric field. The PR generalizes this class to work for any arbitrary FieldType as listed in Fields.H. In addition to vector fields (array size 3), this class can now also be used for scalar fields, such as the electrostatic potential phi. In general, a WarpXSolver vector can consist of one array size 3 of Multifabs (for vector fields such as E, B, and A) and one scalar field. For example, the array size 3 field can be used for the vector potential A and the scalar field can be used for the scalar potential phi, which is the full state of unknowns for a Darwin electromagnetic model.

This PR does add some bloat to the main WarpX class. A logical next step to this PR would be to create a WarpXFields class and to transfer much of what is in the WarpX class related to the fields to the WarpXFields class.

### DIFF
--- a/Source/FieldSolver/Fields.H
+++ b/Source/FieldSolver/Fields.H
@@ -42,15 +42,14 @@ namespace warpx::fields
     inline bool
     isFieldArray (const FieldType field_type)
     {
-        if (field_type == FieldType::Efield_aux || field_type == FieldType::Bfield_aux ||
-            field_type == FieldType::Efield_fp  || field_type == FieldType::Bfield_fp ||
-            field_type == FieldType::current_fp || field_type == FieldType::current_fp_nodal ||
-            field_type == FieldType::vector_potential_fp ||
-            field_type == FieldType::Efield_cp  || field_type == FieldType::Bfield_cp ||
-            field_type == FieldType::current_cp ||
-            field_type == FieldType::Efield_avg_fp || field_type == FieldType::Bfield_avg_fp ||
-            field_type == FieldType::Efield_avg_cp || field_type == FieldType::Bfield_avg_cp) { return true; }
-        else { return false; }
+        return (field_type == FieldType::Efield_aux || field_type == FieldType::Bfield_aux ||
+                field_type == FieldType::Efield_fp  || field_type == FieldType::Bfield_fp ||
+                field_type == FieldType::current_fp || field_type == FieldType::current_fp_nodal ||
+                field_type == FieldType::vector_potential_fp ||
+                field_type == FieldType::Efield_cp  || field_type == FieldType::Bfield_cp ||
+                field_type == FieldType::current_cp ||
+                field_type == FieldType::Efield_avg_fp || field_type == FieldType::Bfield_avg_fp ||
+                field_type == FieldType::Efield_avg_cp || field_type == FieldType::Bfield_avg_cp);
     }
 }
 
--- a/Source/FieldSolver/Fields.H
+++ b/Source/FieldSolver/Fields.H
@@ -8,6 +8,7 @@
 #define WARPX_FIELDS_H_
 
 #include <algorithm>
+#include <iterator>
 
 namespace warpx::fields
 {
--- a/Source/FieldSolver/Fields.H
+++ b/Source/FieldSolver/Fields.H
@@ -11,6 +11,7 @@ namespace warpx::fields
 {
     enum struct FieldType : int
     {
+        None,
         Efield_aux,
         Bfield_aux,
         Efield_fp,
--- a/Source/FieldSolver/Fields.H
+++ b/Source/FieldSolver/Fields.H
@@ -38,6 +38,20 @@ namespace warpx::fields
         Efield_avg_cp,
         Bfield_avg_cp
     };
+
+    inline bool
+    isFieldArray (const FieldType field_type)
+    {
+        if (field_type == FieldType::Efield_aux || field_type == FieldType::Bfield_aux ||
+            field_type == FieldType::Efield_fp  || field_type == FieldType::Bfield_fp ||
+            field_type == FieldType::current_fp || field_type == FieldType::current_fp_nodal ||
+            field_type == FieldType::vector_potential_fp ||
+            field_type == FieldType::Efield_cp  || field_type == FieldType::Bfield_cp ||
+            field_type == FieldType::current_cp ||
+            field_type == FieldType::Efield_avg_fp || field_type == FieldType::Bfield_avg_fp ||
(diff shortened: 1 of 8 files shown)


ax3l (approved):
This looks great to me, thank you.

To reduce the data bloat in WarpX and to make init/finalize more controllable, one task we thought we can do in the hackathon is to move all AMReX data into a member variable of WarpX, as ImpactX is doing:
https://github.com/ECP-WarpX/impactx/blob/24.08/src/ImpactX.H#L109-L113
([header](https://github.com/ECP-WarpX/impactx/blob/24.08/src/initialization/AmrCoreData.H))

ax3l:
@JustinRayAngus oopsi, can you please update/rebase? :)

JustinRayAngus:
> @JustinRayAngus oopsi, can you please update/rebase? :)

oopsi?

I rebased just now.

Source/FieldSolver/Fields.H
```diff
@@ -35,6 +36,19 @@ namespace warpx::fields
...
+    {
+        return (field_type == FieldType::Efield_aux || field_type == FieldType::Bfield_aux ||
+                field_type == FieldType::Efield_fp  || field_type == FieldType::Bfield_fp ||
+                field_type == FieldType::current_fp || field_type == FieldType::current_fp_nodal ||
+                field_type == FieldType::vector_potential_fp ||
+                field_type == FieldType::Efield_cp  || field_type == FieldType::Bfield_cp ||
+                field_type == FieldType::current_cp ||
+                field_type == FieldType::Efield_avg_fp || field_type == FieldType::Bfield_avg_fp ||
+                field_type == FieldType::Efield_avg_cp || field_type == FieldType::Bfield_avg_cp);
+    }
```
lucafedeli88:
```suggestion

        constexpr FieldType ArrayFieldTypes[] = {
            FieldType::Efield_aux, FieldType::Bfield_aux, FieldType::Efield_fp, FieldType::Bfield_fp,
            FieldType::current_fp, FieldType::current_fp_nodal, FieldType::vector_potential_fp,
            FieldType::Efield_cp, FieldType::Bfield_cp, FieldType::current_cp,
            FieldType::Efield_avg_fp, FieldType::Bfield_avg_fp, FieldType::Efield_avg_cp, FieldType::Bfield_avg_cp};

    inline bool
    isFieldArray (const FieldType field_type)
    {
        return std::find(
            std::begin(ArrayFieldTypes), std::end(ArrayFieldTypes),
            field_type) != std::end(ArrayFieldTypes);
    }
```
I would suggest this optional change in order to improve readability.

Source/FieldSolver/ImplicitSolvers/WarpXSolverVec.H
```diff
@@ -32,18 +31,25 @@
 #include <ostream>
 #include <vector>
 
+#include "FieldSolver/Fields.H"
```
lucafedeli88: In WarpX we typically follow this convention for the order of `#include` directives:

- WarpX headers
- ablastr headers
- AMReX headers
- STL headers

We also try to order the headers in each category alphabetically (although this is not rigorously respected).

Would it be possible to include `FieldSolver/Fields.H` before `Utils/TextMsg.H` ?

lucafedeli88 (commented):
Hi @JustinRayAngus ! Thanks a lot for this PR! I've left a couple of minor comments. I'll try to finish my review by tomorrow.

Source/FieldSolver/Fields.H
JustinRayAngus: Thanks for the suggestion. I ran into compile issues with this usage of std::find() that I couldn't readily figure out. I settled on a slightly different version that does not use std::find(). Let me know what you think.

`constexpr bool
   isFieldArray (const FieldType field_type)
   {
        for (const auto& f : ArrayFieldTypes) {
            if (f == field_type) { return true; }
        }
        return false;
 }`

JustinRayAngus: Clang tidy complained. **error: replace loop by 'std::any_of()'**

It wants me to replace the for loop with std::any_of().

lucafedeli88: Yes,  the reason is that we enable this check in clang-tidy:

- [readability/use-anyofallof](https://releases.llvm.org/15.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/readability/use-anyofallof.html)

In this case I personally think that the final solution is quite elegant!

However, in general, if you find some `clang-tidy` suggestions too pedantic, please don't hesitate to add a point in the agenda of the developers' meeting: we may decide to remove a check or, in some cases, to tune how the check works.

lucafedeli88 (approved):
Thanks for this PR, @JustinRayAngus !
I think that it can be merged.

Source/FieldSolver/Fields.H
JustinRayAngus: I think the clang-tidy suggestion here is good.

Source/FieldSolver/ImplicitSolvers/WarpXSolverVec.H
```diff
@@ -33,17 +33,22 @@
...
+ *  A WarpXSolverVec can consist of an array size 3 of MultiFabs (for vector fields
+ *  such as E, B, and A) or of a single MultiFab for scalar fields. Both the array
+ *  size 3 and scalar fields must be of type warpx::fields::FieldType.
+ *  Additionally, a WarpXSolverVec can in general contain both an array size 3 field and a
+ *  scalar field. For example, the array size 3 field can be used for the vector potential A
+ *  and the scalar field can be used for the scalar potential phi, which is the full state of
+ *  unknowns for a Darwin electromagnetic model.
  */
 
+class WarpX;
```
ax3l: Doxygen shall not apply to `WarpX`, so let's move it up.
```suggestion
```

Source/FieldSolver/ImplicitSolvers/WarpXSolverVec.H
```diff
@@ -33,17 +33,22 @@
 #include <vector>
 
 /**
```
ax3l, line 39:
```suggestion

// forward declaration
class WarpX;

/**
```

ax3l (approved):

PR #5175, opened 2024-08-27
[WIP] Magnetostatic solver refactor and bug fixes

roelof-groenewald:
This PR introduces a major refactor of the magnetostatic solver primarily with three goals:

1. Allow the magnetostatic solver to be used with either the explicit or semi-implicit electrostatic potential solvers.
2. Fix a bug in the calculation of B from A. Currently the AMReX gradient function is used to get $\nabla A$ on a grid matching the E-field staggering; this field is then interpolated to match the B-field staggering for each component needed in calculating $B = \nabla \times A$. However, this interpolation causes significant non-zero B-field divergence as is shown in the plot below. Instead, now $A$ is interpolated to match the E-field staggering and B is calculated directly from that A-field using the standard FD curl calculation.

image
_Example of non-zero B-field divergence introduced with the current magnetostatic solver implementation._

image
_Same simulation as above but with the fixes from this PR included (note the colorbar range)_

3. Add the inductive electric field term ($E_{ind} = -\frac{dA}{dt}$) during ES+MS evolution.

This PR builds on and requires first merging:

- [x]  #5079.


Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.H
```diff
@@ -2,34 +2,169 @@
...
+                if (prob_lo == 0){
+                    lobc[adim][0] = LinOpBCType::Neumann;
+                    dirichlet_flag[adim][0] = false;
+                    dim_start = 1;
+
+                    // handle the r_max boundary explicitly
+                    if (field_boundary_hi[0] == FieldBoundaryType::PEC) {
+                        if (adim == 0) {
+                            hibc[adim][0] = LinOpBCType::Neumann;
+                            dirichlet_flag[adim][1] = false;
```
clarkse, line 96: This can be changed to use an STL map data structure as in the ProjectionDivCleaner. Seesm like a good time to do this refactoring.

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -249,9 +269,9 @@ void
 WarpX::setVectorPotentialBC ( amrex::Vector<amrex::Array<std::unique_ptr<amrex::MultiFab>,3>>& A ) const
```
clarkse, line 326: Should this have some sort of warning about the inaccuracy of the field near an EB surface? This is now effetively going to be a staircase EB instead of cut cell correct?

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
```diff
@@ -326,7 +326,7 @@ WarpX::ProjectionCleanDivB() {
     } else if ( WarpX::electromagnetic_solver_id == ElectromagneticSolverAlgo::Yee
             ||  WarpX::electromagnetic_solver_id == ElectromagneticSolverAlgo::HybridPIC
             ||  ( (WarpX::electrostatic_solver_id == ElectrostaticSolverAlgo::LabFrame
-                || WarpX::electrostatic_solver_id == ElectrostaticSolverAlgo::LabFrameElectroMagnetostatic)
+                || WarpX::electrostatic_solver_id == ElectrostaticSolverAlgo::LabFrameSemiImplicit)
```
clarkse, line 329: Is the explicit version of the solver going away completely? I.e. the LabFrameElectroMagnetostatic Algo?

clarkse (changes requested):
This looks good @roelof-groenewald thank you for shaking this out.

Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.H
roelof-groenewald: Great idea! I'll add that to my tasks for the upcoming hackathon (also changing it for the current ES solver).

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
roelof-groenewald: No, the magnetostatic solver is just changed to function as an add-on to an electrostatic solver. So instead of defining it as it's own solver (explicit ES + MS) the type of ES solver is set in the input and the extra flag `do_magnetostatic` sets whether the MS solve will be added on to the ES solve.

PR #5176, opened 2024-08-27
CFL-limited adaptive timestepping for electrostatic solver

archermarx:
This PR implements CFL-limited adaptive timestepping for the electrostatic and magnetostatic solvers. This relies on #5169 (edit: now merged!) and would help fix a lot of the problems in #5113.

~~The algorithm is pretty simple. At the end of each simulation step, the timestep is updated to be CFL * min(dx,dy,dz) / max_v, where max_v is the maximum speed of any particle in the simulation. For accuracy using leapfrog-like methods where position and velocity are stored at different times, we store both the current and next timestep in `dt` and `dt_next`, respectively. During a particle push, the position is pushed by `dt` and the momentum by `dt/2 + dt_next/2`. Then, dt is set to dt_next and dt_next is computed using the formula above. I found this approach demonstrated in [Chen et al](https://www-sciencedirect-com.proxy.lib.umich.edu/science/article/pii/S0010465523000590), which included a helpful diagram, although their time centering is offset from ours~~

image

**EDIT**: After some feedback from @roelof-groenewald, we now set the timestep at user-defined intervals as opposed to every step. When the timestep is updated, we first synchronize position and velocity. We then set the timestep according to CFL * min(dx,dy,dz) / max_v, where max_v is the maximum speed of any particle in the simulation. We then push the velocity backward one half step and continue using the new timestep. 

For PICMI, the cfl number for electrostatic simulations is now controlled by the `warpx_cfl` parameter in the `ElectrostaticSolver` class. Users may also specify a `warpx.max_dt` and `warpx.dt_update_interval`. The former parameter specifies the maximum allowable timestep and additionally sets the initial timestep. The latter determines how often the timestep is updated. Adaptive timestepping is disabled if this parameter is <= 0.

## TODO
- [x] Determine if this would work for the hybrid solver (no, CFL is not main limiter)
- [x] Add timestep diagnostic
- [x] Add version of ES sphere test


archermarx:
Marking as ready for review. Ran a long simulation over the weekend using this feature at two different CFL numbers and encountered no issues.

Python/pywarpx/picmi.py
```diff
@@ -1886,13 +1886,17 @@ def init(self, kw):
...
         self.magnetostatic = kw.pop("warpx_magnetostatic", False)
+        self.cfl = kw.pop("warpx_cfl", None)
 
     def solver_initialize_inputs(self):
         # Open BC means FieldBoundaryType::Open for electrostatic sims, rather than perfectly-matched layer
         BC_map["open"] = "open"
 
         self.grid.grid_initialize_inputs()
 
+        if self.cfl is not None:
```
dpgrote: This check `self.cfl is not None` is not needed. When the parameters are processed, there is already a check, so if `pywarpx.warpx.cfl` is None, it is ignored.

Source/Evolve/WarpXComputeDt.cpp
```diff
@@ -27,29 +28,33 @@
 #include <algorithm>
 #include <memory>
 
+AMREX_FORCE_INLINE amrex::Real
+minDim (const amrex::Real* x)
+{
+#if defined(WARPX_DIM_1D_Z)
```
dpgrote: I learned a new trick today. This can be simplified to
```
std::min({AMREX_D_DECL(x[0], x[1], x[2])});
```
So this separate routine is not needed.

archermarx: Oh, neat!

Source/Evolve/WarpXComputeDt.cpp
```diff
@@ -27,29 +28,33 @@
...
+#endif
+}
+
 /**
  * Determine the timestep of the simulation. */
 void
 WarpX::ComputeDt ()
 {
     // Handle cases where the timestep is not limited by the speed of light
-    if (electromagnetic_solver_id == ElectromagneticSolverAlgo::None ||
```
dpgrote, line 36: A comment here, is that when `electromagnetic_solver_id == ElectromagneticSolverAlgo::None` it does not necessarily mean that the electrostatic solver is being used. There are cases when no solver is being used. But that should be fine since the new CFL limit should work for those cases as well.

Would this CFL limit also work for the hybrid solver? @roelof-groenewald

Source/Evolve/WarpXComputeDt.cpp
```diff
@@ -27,29 +28,33 @@
...
-        electromagnetic_solver_id == ElectromagneticSolverAlgo::HybridPIC) {
-
+    // and no constant timestep is provided
+    if (electromagnetic_solver_id == ElectromagneticSolverAlgo::HybridPIC) {
         std::stringstream errorMsg;
-        if (electrostatic_solver_id != ElectrostaticSolverAlgo::None) {
-            errorMsg << "warpx.const_dt must be specified with the electrostatic solver.";
-        } else if (electromagnetic_solver_id == ElectromagneticSolverAlgo::HybridPIC) {
+        if (electromagnetic_solver_id == ElectromagneticSolverAlgo::HybridPIC) {
```
dpgrote: This can be cleaned up since the `if` check is repeated, and the `else` clause would never be reached.

Source/Evolve/WarpXComputeDt.cpp
```diff
@@ -91,10 +91,41 @@ WarpX::ComputeDt ()
...
+    const amrex::ParticleReal max_v = mypc->maxParticleVelocity();
+    const amrex::Real deltat_new = cfl * dx_min / max_v;
+
+    // Set present dt to previous next dt
+    dt[max_level] = dt_next[max_level];
+    dt_next[max_level] = deltat_new;
+
+    for (int lev = max_level-1; lev >= 0; --lev) {
+        if (do_subcycling) {
+            dt[lev] = dt[lev+1] * refRatio(lev)[0];
```
dpgrote: Could this be `dt[lev] = dt_next[lev];` and taken out of the `if` blocks?

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -298,6 +298,12 @@ WarpX::Evolve (int numsteps)
                       << " s; Avg. per step = " << evolve_time/(step-step_begin+1) << " s\n\n";
         }
 
+        // Update timestep for electrostatic solver if a constant dt is not provided
+        if (electrostatic_solver_id != ElectrostaticSolverAlgo::None &&
+            !m_const_dt.has_value()) {
```
dpgrote: This could be
```
if (electromagnetic_solver_id == ElectromagneticSolverAlgo::None && !m_const_dt.has_value()) {
```
since this update could be done for any case that does not have the speed of light CFL.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -2931,7 +2933,7 @@ PhysicalParticleContainer::PushPX (WarpXParIter& pti,
 #endif
                                       dt);
 
-            UpdatePosition(xp, yp, zp, ux[ip], uy[ip], uz[ip], dt);
+            UpdatePosition(xp, yp, zp, ux[ip], uy[ip], uz[ip], 0.5 * (dt + dt_next));
```
dpgrote: This should be `0.5_rt`.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -2949,7 +2951,7 @@ PhysicalParticleContainer::PushPX (WarpXParIter& pti,
                                           t_chi_max,
                                           dt);
 
-                UpdatePosition(xp, yp, zp, ux[ip], uy[ip], uz[ip], dt);
+                UpdatePosition(xp, yp, zp, ux[ip], uy[ip], uz[ip], 0.5 * (dt + dt_next));
```
dpgrote: This should be `0.5_rt`.

Source/Evolve/WarpXComputeDt.cpp
archermarx: This should work for the hybrid solvers used in my field (plasma propulsion), as the condition I impose is typically used to set a constant global timestep. As long as the way the leapfrog algorithm interacts with the adaptive timestep maintains its accuracy and consistency, I don't think there should be any problems. The only issues would stem from how (if at all) the Ohm's law solve interacts with the timestep.

Source/Evolve/WarpXComputeDt.cpp
archermarx: Good catch, thanks. And if it looks like the hybrid solver works with this timestep, we can remove this check entirely

Source/Evolve/WarpXComputeDt.cpp
archermarx: Yes, I think so. That would give 
```cpp
for (int lev = max_level-1; lev >= 0; --lev) {
    dt[lev] = dt_next[lev];
    if (do_subcycling) {
        dt_next[lev] = dt_next[lev+1] * refRatio(lev)[0];
    } else {
        dt_next[lev] = dt_next[lev+1];
    }
}
```

Source/Evolve/WarpXEvolve.cpp
archermarx: Yes, that's a bit cleaner, if we think the hybrid solver will work with this.

Source/Particles/PhysicalParticleContainer.cpp
archermarx: Thanks, fixed on both of these

Python/pywarpx/picmi.py
archermarx: Thanks, fixed.

archermarx:
Addressed most of the comments. Remaining open issues concern the hybrid solver. If we think that will work OK with this timestepping method, then we can use it for all non-EM solver cases. I also want to add a test case or two. Would modifying one of the existing ES test cases to not use a constant dt be sufficient, or should I add an extra test?

Source/Evolve/WarpXComputeDt.cpp
```diff
@@ -58,16 +63,11 @@ WarpX::ComputeDt ()
...
         // Computation of dt for spectral algorithm
         // (determined by the minimum cell size in all directions)
-#if defined(WARPX_DIM_1D_Z)
-        deltat = cfl * dx[0] / PhysConst::c;
-#elif defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
-        deltat = cfl * std::min(dx[0], dx[1]) / PhysConst::c;
-#else
-        deltat = cfl * std::min(dx[0], std::min(dx[1], dx[2])) / PhysConst::c;
-#endif
+        deltat = cfl / PhysConst::c * minDim(dx);
```
dpgrote: Since this is using `c`, this `dt` for the first time step would typically be much smaller than needed. Could `UpdateDtFromParticleSpeeds` be called from `WarpX::InitData` after the particles have been initialized to give an appropriate `dt` for the first step?

archermarx: My thinking was that in most real simulations, you use enough timesteps that a small first timestep really won't matter, and that I wanted to have all of the timestep-setting logic in one place in the code. Would it break much to instead move the call to `ComputeDt` to after the particles have been initialized?

dpgrote:
I would add a new version of one of the ES test cases to use the CFL, perhaps `Tests/electrostatic_sphere/inputs_3d`.

Source/Evolve/WarpXComputeDt.cpp
dpgrote: I was thinking the same thing. Though the `InitPML` in `InitFromScratch` look likes it depend on `dt`.

archermarx: I am testing moving the timestep update to the start of the `Evolve` loop. Previously we would actually get two very small timesteps as `dt_next` wouldn't be updated until the end of the first iteration. A new problem has arisen though -- what to do when particles are at rest? The `ElectrostaticSphere` case starts with particles with zero velocity, so the computed adaptive timestep is infinite. I'm testing just setting it to a maximum value for now but that isn't especially principled.

archermarx: It might be good to add a user-specified maximum timestep as a fix for this

Python/pywarpx/picmi.py
```diff
@@ -2739,6 +2745,9 @@ class Simulation(picmistandard.PICMI_Simulation):
 
     warpx_used_inputs_file: string, optional
         The name of the text file that the used input parameters is written to,
+
+    warpx_max_dt: float, optional
+        The maximum allowable timestep when using adaptive timestepping (only available for non-electromagnetic solvers)
```
roelof-groenewald: Seeing as this is specific to the Poisson solver I think it would make more sense to put this argument in `ElectrostaticSolver`.

roelof-groenewald (commented):
This is really cool! I do worry a bit that the need to pass `dt_next` to so many functions for which it usually is not needed tangles things up in a way that could make it hard to maintain in the future, so I wonder if there is a way to avoid this. Said another way, I wonder if this new algorithm can be added without touching as many functions that are used by other algorithms as well. 
For example, to get adaptive time stepping in the ES scheme one could run a simulation for a small number of steps and create a checkpoint (during which particle positions and velocities are synchronized), and then restart that simulation with a different timestep. Could we do this in a way that doesn't require actually checkpointing and restarting? The WarpX evolve loop includes a check for whether positions and velocities are synchronized (`is_synchronized`) and moves the momentum back half a timestep if it is. Could this algorithm instead re-synchronize the positions and velocities every time the timestep is changed (setting `is_synchronized = True`)? At the next loop the momentum would then be moved back based on the new timestep and the push should work as usual. Does that make sense? What do you think?

Python/pywarpx/picmi.py
archermarx: Sure, that makes sense

archermarx:
> This is really cool! I do worry a bit that the need to pass `dt_next` to so many functions for which it usually is not needed tangles things up in a way that could make it hard to maintain in the future, so I wonder if there is a way to avoid this. Said another way, I wonder if this new algorithm can be added without touching as many functions that are used by other algorithms as well. For example, to get adaptive time stepping in the ES scheme one could run a simulation for a small number of steps and create a checkpoint (during which particle positions and velocities are synchronized), and then restart that simulation with a different timestep. Could we do this in a way that doesn't require actually checkpointing and restarting? The WarpX evolve loop includes a check for whether positions and velocities are synchronized (`is_synchronized`) and moves the momentum back half a timestep if it is. Could this algorithm instead re-synchronize the positions and velocities every time the timestep is changed (setting `is_synchronized = True`)? At the next loop the momentum would then be moved back based on the new timestep and the push should work as usual. Does that make sense? What do you think?

That's an interesting idea, and would probably work well. Another thought I had while plumbing `dt_next` through the code was that we could instead have a struct that contains timestepping details, so `dt`, `dt_next`, the current iteration, and maybe even the push type. That way if any future changes are made to the timestepping algorithms, we don't need to update as many files again

Python/pywarpx/picmi.py
roelof-groenewald: I don't think the adaptive time stepping would be particularly useful for the hybrid code since the allowable timestep in that model is set by the frequency of Whistler waves at the smallest wavelength the grid can support rather than a particle CFL condition.

archermarx: That makes sense!

archermarx:
> This is really cool! I do worry a bit that the need to pass `dt_next` to so many functions for which it usually is not needed tangles things up in a way that could make it hard to maintain in the future, so I wonder if there is a way to avoid this. Said another way, I wonder if this new algorithm can be added without touching as many functions that are used by other algorithms as well. For example, to get adaptive time stepping in the ES scheme one could run a simulation for a small number of steps and create a checkpoint (during which particle positions and velocities are synchronized), and then restart that simulation with a different timestep. Could we do this in a way that doesn't require actually checkpointing and restarting? The WarpX evolve loop includes a check for whether positions and velocities are synchronized (`is_synchronized`) and moves the momentum back half a timestep if it is. Could this algorithm instead re-synchronize the positions and velocities every time the timestep is changed (setting `is_synchronized = True`)? At the next loop the momentum would then be moved back based on the new timestep and the push should work as usual. Does that make sense? What do you think?

That's a neat idea! I'll see if I can implement that instead.

archermarx:
Ok, I refactored according to your suggestion. I think it's much cleaner now.

archermarx:
I could use a bit of help with the last failing test. It's this one, in cartesian2D:

`194 - test_2d_pml_x_psatd_restart.analysis (Failed)`

I don't do anything to the restarts at all, so I'm unsure why this would be off. Here's the test output:

```

tolerance = 1e-12

field: ('boxlib', 'Bx'); error = 0.0
field: ('boxlib', 'divE'); error = 0.0
field: ('boxlib', 'rho'); error = 0.0

ERROR: Benchmark and output file checksum have different value for key [lev=0,divE]
Benchmark: [lev=0,divE] 1.891054630780910e+05
Test file: [lev=0,divE] 1.891054630778933e+05
Absolute error: 1.98e-07
Relative error: 1.05e-12
ERROR: Benchmark and output file checksum have different value for key [lev=0,rho]
Benchmark: [lev=0,rho] 1.673495200001335e-06
Test file: [lev=0,rho] 1.673495199999600e-06
Absolute error: 1.73e-18
Relative error: 1.04e-12

----------------
New file for test_2d_pml_x_psatd:
{
  "lev=0": {
    "Bx": 1.1599355635529139e-08,
    "By": 1.51580110019918e-08,
    "Bz": 8.995705405878174e-09,
    "Ex": 3.936315546012292,
    "Ey": 4.13448272729307,
    "Ez": 3.307082741814466,
    "divE": 189105.46307789328,
    "rho": 1.6734951999995999e-06
  }
}
```

Source/Evolve/WarpXComputeDt.cpp
```diff
@@ -27,29 +28,30 @@
...
         WARPX_ALWAYS_ASSERT_WITH_MESSAGE(m_const_dt.has_value(), errorMsg.str());
-
-        for (int lev=0; lev<=max_level; lev++) {
-            dt[lev] = m_const_dt.value();
-        }
-        return;
+    } else if (electromagnetic_solver_id == ElectromagneticSolverAlgo::None) {
+        std::stringstream errorMsg;
+        errorMsg << "warpx.const_dt must be specified with the electrostatic solver, or the warpx.dt_update_interval must be > 0.";
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(m_const_dt.has_value() || dt_update_interval.isActivated(), errorMsg.str());
```
roelof-groenewald: Usually we write this as:
```suggestion
        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
            m_const_dt.has_value() || dt_update_interval.isActivated(), 
            "warpx.const_dt must be specified with the electrostatic solver, or the warpx.dt_update_interval must be > 0."
        );
```
In this case the errorMsg was changed depending on the solver while the assert itself was the same, which is why the `std::stringstream` was used. I don't mind keeping this as is, just wanted to point it out since it currently follows a different practice from other places in WarpX.

Source/Evolve/WarpXComputeDt.cpp
```diff
@@ -58,16 +60,15 @@ WarpX::ComputeDt ()
...
-#elif defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
-        deltat = cfl * std::min(dx[0], dx[1]) / PhysConst::c;
-#else
-        deltat = cfl * std::min(dx[0], std::min(dx[1], dx[2])) / PhysConst::c;
-#endif
+        if (m_max_dt.has_value() && electromagnetic_solver_id == ElectromagneticSolverAlgo::None) {
+            deltat = m_max_dt.value();
+        } else {
+            deltat = cfl / PhysConst::c * minDim(dx);
+        }
```
roelof-groenewald: I think it would be more readable to split this into two `if`-blocks, i.e., something like:
```
} else if (electrostatic_solver_id != ElectrostaticSolverAlgo::None) {
    // Set dt for electrostatic algorithm
    if (m_max_dt.has_value()) {
        deltat = m_max_dt.value();
    } else {
        deltat = cfl / PhysConst::c * minDim(dx);
    }
} else if (electromagnetic_solver_id == ElectromagneticSolverAlgo::PSATD) {
    // Computation of dt for spectral algorithm
    deltat = cfl / PhysConst::c * minDim(dx);
}
```
It does repeat one line of code but I think the logic is much easier to read.

Source/WarpX.H
```diff
@@ -116,6 +116,7 @@ public:
     void InitData ();
 
     void Evolve (int numsteps = -1);
+    void Synchronize ();
```
roelof-groenewald, line 123: Can you please add a doc-string for what this function does (we're trying to improve the WarpX inline commenting to make things easier to understand 😃 )

roelof-groenewald (commented):
Thanks for updating the implementation! I just have a few minor comments.

As for the test that is suddenly failing, I'd have to guess it is due to the timestep in that simulation having slightly changed since the logic to set it is subtly (and slightly) different. Before it would have been:
```
deltat = cfl * std::min(dx[0], dx[1]) / PhysConst::c;
```
whereas now it is
```
deltat = cfl / PhysConst::c * std::min({AMREX_D_DECL(x[0], x[1], x[2])});
```
I'm guessing something in that, probably the order of operations, caused a floating point change in `deltat` which broke the test.

Source/Evolve/WarpXComputeDt.cpp
archermarx: Thanks, good to know!

archermarx:
> Thanks for updating the implementation! I just have a few minor comments.
> 
> As for the test that is suddenly failing, I'd have to guess it is due to the timestep in that simulation having slightly changed since the logic to set it is subtly (and slightly) different. Before it would have been:
> 
> ```
> deltat = cfl * std::min(dx[0], dx[1]) / PhysConst::c;
> ```
> 
> whereas now it is
> 
> ```
> deltat = cfl / PhysConst::c * std::min({AMREX_D_DECL(x[0], x[1], x[2])});
> ```
> 
> I'm guessing something in that, probably the order of operations, caused a floating point change in `deltat` which broke the test.

That's what I was thinking, but the weird part is it's only the restart that fails. The regular `test_2d_pml_x_psatd` test passes normally. I'm not sure how to fix that, since they both use the same checksum. I'll try switching the order of operations and see what happens.

archermarx:
Ok, looks like tests pass now, yay! Also addressed the most recent round of comments.

roelof-groenewald (approved):
Looks good! Thanks for adding this feature!

PR #5181, opened 2024-08-28
UB sanitizer CI test: bump clang version from 15 to 17

lucafedeli88:
With this PR, `clang-17` instead of `clang-15` is used for UB sanitizer test in CI .

Note: Ubuntu 24.04 seems to be in beta for GitHub-hosted runners (see https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#choosing-github-hosted-runners ), so in the future we won't need to use the Ubuntu 23.10 container.


ax3l (approved):
LGTM, thx

PR #5184, opened 2024-08-28
Modernize usage of type traits in WarpX

lucafedeli88:
This PR implements C++14 style type templates  and C++17 style variable templates, as suggested by the 
- [[modernize-type-traits]](https://clang.llvm.org/extra/clang-tidy/checks/modernize/type-traits.html)

clang-tidy check.


ax3l (approved):
LGTM :+1:

PR #5185, opened 2024-08-28
ABLASTR: make deposit_charge function non-static

lucafedeli88:
I think that the `deposit_charge` function here does not need to be static.

### DIFF
--- a/Source/ablastr/particles/DepositCharge.H
+++ b/Source/ablastr/particles/DepositCharge.H
@@ -46,7 +46,7 @@ namespace ablastr::particles
  * \param nc number of components to deposit
  */
 template< typename T_PC >
-static void
+void
 deposit_charge (typename T_PC::ParIterType& pti,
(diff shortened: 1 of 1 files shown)


ax3l (changes requested):
I think it has to stay static, since it is header only.

Otherwise we will get multiple definition errors if we use it in multiple TUs.
https://stackoverflow.com/questions/780730/c-c-static-function-in-header-file-what-does-it-mean

lucafedeli88:
In this specific case the function is templated (thus implicitly `inline`), so it should be fine without `static`. 
If I understand correctly, using `static` each TU has an independent copy of the function (resulting in a larger executable), while with just `inline` only one compiled version of the function exists in the executable.

ax3l:
> In this specific case the function is templated (thus implicitly inline), so it should be fine without static.

Aha, good point:
https://stackoverflow.com/a/11527505/2719194

ax3l (approved):

PR #5188, opened 2024-08-28
remove unnecessary limit check in UpdateMomentumPerezElastic

JustinRayAngus:
Third PR of several related to cleaning/streamlining UpdateMomentumPerezElastic.


ax3l:
Context: we introduced this to prevent NaNs in #4304

It looks like this could still apply. If it does, let us add an inline comment on the `if` condition.

JustinRayAngus:
Thanks for the comment @ax3l. There is no division in these lines by vcms. So it is safe to remove the limit check.

RemiLehe (approved):

PR #5189, opened 2024-08-28
Add 1d cylinder and sphere

dpgrote:
This large PR adds 1d cylindrical and spherical geometries.

The `x` dimension is used as the radial dimension. Note that this is in contrast to the 1-d Cartesian which uses the `z` axis.

This touches many files, every place that has a dependency on the dimensionality. One complication is that since there is no z-axis, so some parts of the code (the moving window, boosted frame, and BTD diagnostics) are not applicable, and require changes there to avoid compilation problems.

Current status:
 - cylindrical works, producing correct results in Langmuir test
 - spherical works, producing correct results in Langmuir test

@dpgrote @RemiLehe 
Noting down here ideas for follow-up PRs:
- Improve names of functions like `UpwardDrr_over_r`, etc. (https://github.com/BLAST-WarpX/warpx/pull/5189#discussion_r2085568533)
- Improve build options like `WarpX_DIMS`, `WarpX_GEOM`, etc. (https://github.com/BLAST-WarpX/warpx/pull/5189#discussion_r2085109261)
- Improve names of CI tests like `test_1d_cylindrical_`, etc. (https://github.com/BLAST-WarpX/warpx/pull/5189#discussion_r2085103311)
- Improve names of $\theta$ and $\phi$ field components like `Bt`, `Bp`, etc. (https://github.com/BLAST-WarpX/warpx/pull/5189#discussion_r2085466295)
- Fix pairwise collisions in 1D cylindrical and spherical coordinates
- Add Jacobian $J$ (and metric tensor $g$...?) somewhere to make geometric calculations (e.g., express differential operator in a certain basis) more robust, as opposed to retyping the calculation every time, now that we have three different curvilinear coordinate systems (1D cylindrical, 1D spherical, 2D cylindrical)?


CMakeLists.txt
```diff
@@ -97,8 +97,8 @@ if(WarpX_BACKTRACE_INFO)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g1")
 endif()
 
-set(WarpX_DIMS_VALUES 1 2 3 RZ)
-set(WarpX_DIMS 3 CACHE STRING "Simulation dimensionality <1;2;3;RZ>")
+set(WarpX_DIMS_VALUES 1 2 3 RZ RCYLINDER RSPHERE)
```
JustinRayAngus, line 100: Perhaps the shorted naming convention of RCYL and RSPH are descriptive enough? Just a suggestion to consider.

EZoni:
@dpgrote 

Thanks for the PR! Could you maybe add a few more details about the implementation in the PR description (e.g., most relevant changes, design logic), just to make the review process a bit easier to start?

Source/Diagnostics/ReducedDiags/FieldPoyntingFlux.cpp
```diff
@@ -76,6 +76,10 @@ FieldPoyntingFlux::FieldPoyntingFlux (const std::string& rd_name)
             std::vector<std::string> space_coords = {"z"};
 #elif defined(WARPX_DIM_RZ)
             std::vector<std::string> space_coords = {"r", "z"};
+#elif defined(WARPX_DIM_RCYLINDER)
+            std::vector<std::string> space_coords = {"r"};
+#elif defined(WARPX_DIM_RSPHERE)
+            std::vector<std::string> space_coords = {"r"};
 #endif
```
JustinRayAngus: Maybe these lines could use || logic to combine RCYL and RSPH?

Source/Diagnostics/ReducedDiags/FieldEnergy.cpp
```diff
@@ -185,15 +185,25 @@ FieldEnergy::ComputeNorm2(amrex::MultiFab const& field, [[maybe_unused]]int lev)
...
+#endif
             amrex::Real const theta_integral = (n == 0 ? 1._rt : 0.5_rt);
             return MathConst::pi*v_factor*theta_integral;
+#elif defined(WARPX_DIM_RSPHERE)
+            amrex::Real const r = rmin + (i - tb_lo[0])*dr;
+            // Volume is 4/3*pi*((r + dr/2)**3 - (r - dr/2)**3)
+            amrex::Real const r1 = (i == tb_lo[0] && is_nodal[0] ? r : r - dr/2.0_rt);
+            amrex::Real const r2 = (i == tb_hi[0] && is_nodal[0] ? r : r + dr/2.0_rt);
+            amrex::Real const v_factor = 4.0_rt/3.0_rt*MathConst::pi*(r2*r2*r2 - r1*r1*r1);
+            // Note that dr is already included in dV above so divide it out here.
```
JustinRayAngus: I agree that this is the exact definition of the volume element for spherical geometry, but it does not seem to be the same as the discrete definition that is consistent with both exact energy and charge conservation in the implicit schemes. The method here gives a volume element proportional to the following:
image

The definition of the volume element (other than axis) used in our implicit method is defined in the text below Eq. 15 in this reference: https://doi.org/10.1016/j.jcp.2024.113427. The d^3/4 term is the difference. Maybe there is a way to have it all (exact energy and charge conservation) with the exact volume definition too, but it needs to be shown.

Note that the discrete volume element used for the CYL geometry that is consistent with exact energy and charge conservation is the same as the exact definition. So, there is no problem there.

CMakeLists.txt
dpgrote: For now, I'd rather keep the longer names. I generally prefer to have things spelled out.

Source/Diagnostics/ReducedDiags/FieldEnergy.cpp
dpgrote: This has been addressed to use the correct expressions for the volume. The code has been tested in both cylindrical and spherical with the implicit scheme and gives exact energy conservation.

.github/workflows/source/check_inputs.py
```diff
@@ -5,7 +5,14 @@
...
 # mandatory prefixes for test names
-testname_prefix = ["test_1d_", "test_2d_", "test_3d_", "test_rz_"]
+testname_prefix = [
+    "test_1d_",
+    "test_2d_",
+    "test_3d_",
+    "test_rz_",
+    "test_rcylinder_",
+    "test_rsphere_",
+]
```
EZoni, line 15: I wonder if we should be more consistent/descriptive with the test naming conventions, e.g.,
- `"test_1d_cartesian_"`
- `"test_2d_cartesian_"`
- `"test_3d_cartesian_"`
- `"test_1d_cylindrical_"`
- `"test_2d_cylindrical_"`
- `"test_1d_spherical_"`

If there is consensus on this, I could take care of it in a follow-up PR, once this is merged.

CMakeLists.txt
EZoni: In principle, we could consider something along the lines of https://github.com/BLAST-WarpX/warpx/pull/5189/files#r2085103311 here, too.

EZoni: In other words, we could move from a `WarpX_DIMS` to more of a `WarpX_GEOM` build option, e.g.,
- `WarpX_GEOM=1D_CARTESIAN`
- `WarpX_GEOM=1D_CYLINDRICAL`
- `WarpX_GEOM=1D_SPHERICAL`
- ...

from which the number of dimensions could be extracted automatically. 

Or, we could have two different build options, 
- `WarpX_DIMS=1` and `WarpX_GEOM=CARTESIAN`
- `WarpX_DIMS=1` and `WarpX_GEOM=CYLINDRICAL`
- `WarpX_DIMS=1` and `WarpX_GEOM=SPHERICAL`
- ...

Examples/Tests/langmuir/analysis_r1d.py
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env python3
+
+# Copyright 2019 David Grote, Maxence Thevenet
+#
+# This file is part of WarpX.
+#
+# License: BSD-3-Clause-LBNL
+#
+# This is a script that analyses the simulation results from
+# the script `inputs_rcylinder`. This simulates a RCylinder plasma wave.
```
RemiLehe: Could you update the description to indicate that this Python script is used to check both the 1D cylindrical (`inputs_test_rcylinder_langmuir`) and 1D spherical (`inputs_test_rsphere_langmuir`)? 
(and that the analytical solution has the same form in both cases.)

Examples/Tests/langmuir/analysis_r1d.py
```diff
@@ -0,0 +1,112 @@
...
+# The electric field in the simulation is given (in theory) by:
+# $$ E_r = -\partial_r \phi = \epsilon \,\frac{mc^2}{e}\frac{2\,r}{w_0^2} \exp\left(-\frac{r^2}{w_0^2}\right) \sin(\omega_p t)
+# Unrelated to the Langmuir waves, we also test the plotfile particle filter function in this
```
RemiLehe:
```suggestion
```

Examples/Tests/ohm_solver_em_modes/analysis_rz.py
```diff
@@ -177,7 +177,7 @@ def process(it):
 
     # check if power spectrum sampling match earlier results
     amps = np.abs(F_kw[2, 1, len(kz) // 2 - 2 : len(kz) // 2 + 2])
-    print("Amplitude sample: ", amps)
+    print("Amplitude sample: ", repr(amps))
     assert np.allclose(
-        amps, np.array([80.98876727, 27.9975239, 215.76237021, 24.82014683])
+        amps, np.array([79.98544592, 27.2373895, 213.02920946, 24.12065146])
```
RemiLehe: @dpgrote Could you explain why your PR needed to update this?

Regression/Checksum/benchmarks_json/test_rz_laser_injection_from_RZ_lasy_file.json
```diff
@@ -1,12 +1,12 @@
 {
```
RemiLehe: @dpgrote Could you explain why this needed to be changed?

Regression/Checksum/benchmarks_json/test_rz_laser_injection_from_lasy_file.json
```diff
@@ -1,13 +1,12 @@
 {
   "lev=0": {
```
RemiLehe: @dpgrote Same question.

cmake/WarpXFunctions.cmake
```diff
@@ -223,12 +223,16 @@ endfunction()
 
 # Set the suffix for targets and binaries depending on dimension
 #
-# User specify 1;2;RZ;D in WarpX_DIMS.
-# We append to CMake targets and binaries the suffix "Nd" for 1,2,3 and "rz" for RZ.
+# User specify 1;2;RZ;RCYLINDER;RSPHERE;D in WarpX_DIMS.
```
EZoni: Should this be `3` instead of `D`?
```suggestion
# User specify 1;2;RZ;RCYLINDER;RSPHERE;3 in WarpX_DIMS.
```

Source/BoundaryConditions/PEC_Insulator.cpp
```diff
@@ -142,10 +142,16 @@ namespace
                                 amrex::Real const rmirror = ijk_mirror[idim] + 0.5_rt*(1._rt - is_nodal[idim]);
                                 // Calculate radial scale factor
                                 sign *= rmirror/rguard;
+#if defined(WARPX_DIM_RSPHERE)
```
RemiLehe: Could you add a comment here, explaining why this needs to be done, for spherical geometry?

Source/BoundaryConditions/PEC_Insulator.cpp
```diff
@@ -142,10 +142,16 @@ namespace
...
+                                sign *= rmirror/rguard;
+#endif
                             }
                             if (isInsulatorBoundary) {
                                 // Apply radial scale factor
                                 field_value *= dom_hi[idim]/rguard;
+#if defined(WARPX_DIM_RSPHERE)
+                                field_value *= dom_hi[idim]/rguard;
```
RemiLehe: Could you add a comment here, explaining why this needs to be done, for spherical geometry?

Source/ablastr/particles/NodalFieldGather.H
```diff
@@ -91,7 +91,31 @@ void compute_weights (const amrex::ParticleReal xp,
...
+    const amrex::Real r = (std::sqrt(xp*xp+yp*yp+zp*zp) - plo[0]) * dxi[0] - half_if_cell_centered;
+    i = static_cast<int>(amrex::Math::floor(r));
+    W[0][1] = r - i;
+
+    W[0][0] = 1.0_rt - W[0][1];
+
+    j = 0;
+    k = 0;
+
+#elif (defined WARPX_DIM_1D_Z)
```
EZoni, line 118: Maybe this could be rewritten as follows, not sure which one is clearer?
```suggestion

#elif defined(WARPX_DIM_RCYLINDER) || defined(WARPX_DIM_RSPHERE)

#if defined(WARPX_DIM_RCYLINDER)
    amrex::ignore_unused(zp);
    const amrex::Real r = (std::sqrt(xp*xp+yp*yp) - plo[0]) * dxi[0] - half_if_cell_centered;
#elif defined(WARPX_DIM_RSPHERE)
    const amrex::Real r = (std::sqrt(xp*xp+yp*yp+zp*zp) - plo[0]) * dxi[0] - half_if_cell_centered;
#endif

    i = static_cast<int>(amrex::Math::floor(r));
    j = 0;
    k = 0;
    
    W[0][1] = r - i;
    W[0][0] = 1.0_rt - W[0][1];

#elif (defined WARPX_DIM_1D_Z)
```

Source/BoundaryConditions/WarpXFieldBoundaries.cpp
```diff
@@ -294,8 +294,8 @@ void WarpX::ApplyJfieldBoundary (const int lev, amrex::MultiFab* Jx,
     }
 }
 
-#ifdef WARPX_DIM_RZ
-// Applies the boundary conditions that are specific to the axis when in RZ.
+#if defined(WARPX_DIM_RZ) || defined(WARPX_DIM_RCYLINDER) || defined(WARPX_DIM_RSPHERE)
+// Applies the boundary conditions that are specific to the axis when in cylinderical or spherical
```
RemiLehe:
```suggestion
// Applies the boundary conditions that are specific to the axis when in cylindrical or spherical
```

Examples/Tests/langmuir/analysis_r1d.py
```diff
@@ -0,0 +1,113 @@
...
+
+# Copyright 2019 David Grote, Maxence Thevenet
+#
+# This file is part of WarpX.
+#
+# License: BSD-3-Clause-LBNL
+#
+# This is a script that analyses the simulation results from
+# the script `inputs_test_rcylinder_langmuir_multi` and `inputs_test_rspherical_langmuir_multi`.
+# This simulates a 1d plasma wave using either cylinderical or spherical coordinates.
```
RemiLehe:
```suggestion
# This simulates a 1d plasma wave using either cylindrical or spherical coordinates.
```

Examples/Tests/langmuir/analysis_r1d.py
dpgrote: Change make

Examples/Tests/langmuir/analysis_r1d.py
```diff
@@ -0,0 +1,108 @@
...
+matplotlib.use("Agg")
+import matplotlib.pyplot as plt
+import yt
+
+yt.funcs.mylog.setLevel(50)
+
+import numpy as np
+from scipy.constants import c, e, epsilon_0, m_e
+
+sys.path.insert(1, "../../../../warpx/Regression/Checksum/")
```
EZoni: It should be possible to remove this:
```suggestion
```

Examples/Tests/langmuir/inputs_test_rcylinder_langmuir_multi
EZoni, line 1: If there is a lot of overlap between this input file and the one for the spherical test, we could set the common input parameters in a base input file, e.g., `inputs_base_1d_r` or any other name that makes sense, and then use that by adding the following at the top of the two specialized input files:
```
# base input parameters
FILE = inputs_base_1d_r

# test-specific parameters here
# ...
```

Source/BoundaryConditions/PML.cpp
```diff
@@ -664,6 +664,8 @@ PML::PML (const int lev, const BoxArray& grid_ba,
         auto ngFFT = IntVect(ngFFt_x, ngFFt_z);
 #elif defined(WARPX_DIM_1D_Z)
         auto ngFFT = IntVect(ngFFt_z);
+#elif defined(WARPX_DIM_RCYLINDER) || defined(WARPX_DIM_RSPHERE)
+        auto ngFFT = IntVect(ngFFt_x);
 #endif
```
EZoni, line 669: Just double checking, does `ngFFT` here need to be set for internal consistency even though the FFT-based solver is not supported in the new geometries?

Source/BoundaryConditions/WarpXFieldBoundaries.cpp
```diff
@@ -365,12 +367,18 @@ WarpX::ApplyFieldBoundaryOnAxis (amrex::MultiFab* Er, amrex::MultiFab* Et, amrex
         },
         [=] AMREX_GPU_DEVICE (int i, int j, int /*k*/)
         {
+#if defined(WARPX_DIM_RSPHERE)
+            // Ephi is anti-symmetric
+            Ez_arr(i,j,0,0) = -Ez_arr(-i-ishift_z,j,0,0);
```
EZoni, line 372: What does `ishift_z` represent in spherical geometry?

Source/BoundaryConditions/PEC_Insulator.cpp
dpgrote: The scaling factor is proportional to the surface area, which is proportional to the radius squared. Note that this block of code is unneeded and has been removed.

Source/BoundaryConditions/PEC_Insulator.cpp
dpgrote: As above, the scaling if proportional the surface area which varies as r-squared.

Source/ablastr/particles/NodalFieldGather.H
dpgrote: It is clearer to have the code for each dimensionality separated out.

Examples/Tests/langmuir/inputs_test_rcylinder_langmuir_multi
dpgrote: There is enough differences to keep the two files separate.

Source/BoundaryConditions/WarpXFieldBoundaries.cpp
dpgrote: This flags whether the `Ez` is nodal or cell centered in the radial dimension. `Ez` will the `Ephi`.

Source/Diagnostics/ReducedDiags/FieldMaximum.cpp
```diff
@@ -62,6 +62,14 @@ FieldMaximum::FieldMaximum (const std::string& rd_name)
...
     m_data.resize(noutputs*nLevel, 0.0_rt);
 
+#if defined(WARPX_DIM_RCYLINDER)
+    std::vector<std::string> field_names = {"r", "t", "z"};
+#elif defined(WARPX_DIM_RSPHERE)
+    std::vector<std::string> field_names = {"r", "t", "p"};
+#else
+    std::vector<std::string> field_names = {"x", "y", "z"};
+#endif
+
```
EZoni, line 72: Are these names also mandatory or can they be customized without changing too much code? I'm wondering if it would be possible to replace `t` with `theta` and `p` with `phi`.

Source/BoundaryConditions/PML.cpp
dpgrote: `ngFFT` is only set here to allow the code to compile. The aim was to have the changes as small as possible since this is already a large PR. The alternative is to wrap all of the PML code in macros so it it skipped when using the 1d versions, but that is more complicated that this change. Note that PML are not supported here.

Source/Diagnostics/ReducedDiags/FieldMaximum.cpp
dpgrote: This is consistent with what had already been done with the RZ version. This could be changed, but probably should be a separate PR so that the RZ is updated in the same way.

Source/FieldSolver/FiniteDifferenceSolver/ComputeDivE.cpp
```diff
@@ -195,4 +203,55 @@ void FiniteDifferenceSolver::ComputeDivECylindrical (
...
+        Real const rmin = m_rmin;
+
+        // Extract tileboxes for which to loop
+        Box const& tdive  = mfi.tilebox(divEfield.ixType().toIntVect());
+
+        // Loop over the cells and update the fields
+        amrex::ParallelFor(tdive,
+
+            [=] AMREX_GPU_DEVICE (int i, int j, int /*k*/){
+                Real const r = rmin + i*dr; // r on a nodal grid (F is nodal in r)
```
EZoni: Should `F` be `divE` in this comment?

Source/Diagnostics/ReducedDiags/FieldMaximum.cpp
```diff
@@ -62,6 +62,14 @@ FieldMaximum::FieldMaximum (const std::string& rd_name)
     // resize data array
     m_data.resize(noutputs*nLevel, 0.0_rt);
```
RemiLehe, line 64: Hm, in this file, for `WARPX_DIM_RZ`, we do not do anything specific (and only raise an error message). Should we do the same here for `WARPX_DIM_RCYLINDER` and `WARPX_DIM_RSPHERE`?

RemiLehe: Similar to what is done in `Source/Diagnostics/ReducedDiags/FieldMomentum.cpp`

Regression/Checksum/benchmarks_json/test_rz_laser_injection_from_RZ_lasy_file.json
dpgrote: Because of the change in the boundary condition. This has been moved to PR #5886

Regression/Checksum/benchmarks_json/test_rz_laser_injection_from_lasy_file.json
dpgrote: Because of the change in the boundary condition. This has been moved to PR https://github.com/BLAST-WarpX/warpx/pull/5886

Source/FieldSolver/FiniteDifferenceSolver/ComputeDivE.cpp
dpgrote: No, this was left over from code copied. Now removed.

Source/FieldSolver/FiniteDifferenceSolver/FiniteDifferenceAlgorithms/SphericalYeeAlgorithm.H
```diff
@@ -0,0 +1,163 @@
...
+        // The spherical solver requires one guard cell in each dimension
+        return amrex::IntVect{AMREX_D_DECL(1,1,1)};
+    }
+
+    /** Applies the differential operator `1/r * d(rF)/dr`,
+     * where `F` is on a *nodal* grid in `r`
+     * and the differential operator is evaluated on a *cell-centered* grid.
+     * The input parameter `r` is given at the cell-centered position */
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    static amrex::Real UpwardDrr_over_r (
```
EZoni, line 63: What is the reason behind choosing `UpwardDrr_over_r` for this function's name (as well as similar subsequent functions)?

The name is using a combination of `PascalCase` and `snake_case`, and `Drr` to denote the operator `d(rF)/dr`.

Maybe we could try to use one style only and also make `Drr` more descriptive, e.g., `upward_radial_derivative_rF_over_r` or `upward_radial_grad_rF_over_r` or similar (although `F` may not be the best choice since it is related to div(E) cleaning).

I know this naming is used elsewhere, so this could a follow-up cleanup as well.

Source/Diagnostics/ReducedDiags/FieldMaximum.cpp
dpgrote: The code should be correct to calculate the field maximum for the 1d cases, so can be kept. The field momentum would need to be modified to account for the spatially dependent volume element.

Source/FieldSolver/FiniteDifferenceSolver/FiniteDifferenceAlgorithms/SphericalYeeAlgorithm.H
dpgrote: This could be addressed in a separate PR, since it will also affect RZ.

Source/FieldSolver/FiniteDifferenceSolver/ApplySilverMuellerBoundary.cpp
```diff
@@ -53,40 +57,52 @@ void FiniteDifferenceSolver::ApplySilverMuellerBoundary (
...
     amrex::Real const coef1_r = (1._rt - cdt_over_dr)/(1._rt + cdt_over_dr);
     amrex::Real const coef2_r = 2._rt*cdt_over_dr/(1._rt + cdt_over_dr) / PhysConst::c;
     amrex::Real const coef3_r = cdt/(1._rt + cdt_over_dr) / PhysConst::c;
+#if defined(WARPX_DIM_RZ)
     amrex::Real const cdt_over_dz = cdt*m_h_stencil_coefs_z[0];
     amrex::Real const coef1_z = (1._rt - cdt_over_dz)/(1._rt + cdt_over_dz);
     amrex::Real const coef2_z = 2._rt*cdt_over_dz/(1._rt + cdt_over_dz) / PhysConst::c;
+#endif
 
+#if !defined(WARPX_DIM_RSPHERE)
```
RemiLehe, line 73: Should this be
```suggestion
#if !defined(WARPX_DIM_RSPHERE) && !defined(WARPX_DIM_RCYLINDER)
```

Source/FieldSolver/FiniteDifferenceSolver/ApplySilverMuellerBoundary.cpp
```diff
@@ -53,40 +57,52 @@ void FiniteDifferenceSolver::ApplySilverMuellerBoundary (
...
     // Extract stencil coefficients
     Real const * const AMREX_RESTRICT coefs_z = m_stencil_coefs_z.dataPtr();
     auto const n_coefs_z = static_cast<int>(m_h_stencil_coefs_z.size());
+#endif
 
     // Extract cylindrical specific parameters
     Real const dr = m_dr;
-    int const nmodes = m_nmodes;
     Real const rmin = m_rmin;
+#if !defined(WARPX_DIM_RSPHERE)
```
RemiLehe: Should this be
```suggestion
#if !defined(WARPX_DIM_RSPHERE) && !defined(WARPX_DIM_RCYLINDER)
```

Examples/Tests/langmuir/inputs_test_rsphere_langmuir_multi
```diff
@@ -0,0 +1,85 @@
...
+algo.current_deposition = esirkepov
+warpx.use_filter = 0
+
+# Order of particle shape factors
+algo.particle_shape = 1
+
+# CFL
+# The cfl value was chosen to give better adhoc agreement with theory for the cell
+# nearest the axis where the largest error occurs
+warpx.cfl = 0.78
```
RemiLehe:
```suggestion
warpx.cfl = 1
```

Source/FieldSolver/FiniteDifferenceSolver/FiniteDifferenceAlgorithms/SphericalYeeAlgorithm.H
```diff
@@ -0,0 +1,163 @@
...
+    /** Compute the maximum, CFL-stable timestep
+     *
+     * Compute the maximum timestep, for which the scheme remains stable
+     * under the Courant-Friedrichs-Levy limit.
+     */
+    static amrex::Real ComputeMaxDt (amrex::Real const * const dx) {
+        using namespace amrex::literals;
+        // Check this value! This was determined empirically (for one case)
+        const amrex::Real alpha = 0.001_rt;
+        const amrex::Real delta_t = 1._rt / ( std::sqrt( (1._rt + alpha) / (dx[0]*dx[0])) * PhysConst::c );
```
RemiLehe:
```suggestion
        const amrex::Real delta_t = 0.78._rt * dx[0] / PhysConst::c;
```

Source/FieldSolver/FiniteDifferenceSolver/ApplySilverMuellerBoundary.cpp
dpgrote: This was changed to ```#if defined(WARPX_DIM_RZ)``` and changes were more below to only include the modes when RZ.

Source/FieldSolver/FiniteDifferenceSolver/ComputeCurlA.cpp
```diff
@@ -198,6 +205,101 @@ void FiniteDifferenceSolver::ComputeCurlACylindrical (
...
+            },
+
+            // Bt calculation
+            [=] AMREX_GPU_DEVICE (int i, int j, int /*k*/){
+                // Skip field update in the embedded boundaries
+                if (update_Bt_arr && update_Bt_arr(i, j, 0) == 0) { return; }
+
+                Bt(i, j, 0, 0) = - (
+                    T_Algo::UpwardDr(Ap, coefs_r, n_coefs_r, i, j, 0, 0));
+            },
```
RemiLehe: Should this be:
```suggestion
                Real const r = rmin + (i + 0.5_rt)*dr; // r on a cell-centered grid (Bt is cell-centered in r)
                Bt(i, j, 0, 0) =  -T_Algo::UpwardDrr_over_r(Ap, r, dr, coefs_r, n_coefs_r, i, j, 0, 0);
```
See https://library.psfc.mit.edu/catalog/online_pubs/NRL_FORMULARY_13.pdf, page 8

Source/FieldSolver/FiniteDifferenceSolver/EvolveB.cpp
```diff
@@ -499,4 +505,73 @@ void FiniteDifferenceSolver::EvolveBCylindrical (
...
+            },
+
+            [=] AMREX_GPU_DEVICE (int i, int /*j*/, int /*k*/){
+                Real const r = rmin + (i + 0.5_rt)*dr; // r on a cell-centered grid (Bp is cell-centered in r)
+                Bt(i, 0, 0, 0) += dt*( + T_Algo::UpwardDrr_over_r(Ep, r, dr, coefs_r, n_coefs_r, i, 0, 0, 0));
+            },
+
+            [=] AMREX_GPU_DEVICE (int i, int /*j*/, int /*k*/){
+                Real const r = rmin + (i + 0.5_rt)*dr; // r on a cell-centered grid (Bt is cell-centered in r)
+                Bp(i, 0, 0, 0) += dt*( - T_Algo::UpwardDrr_over_r(Et, r, dr, coefs_r, n_coefs_r, i, 0, 0, 0));
```
EZoni: The PSATD RZ solver is an example where `Bp` is used to mean something different, namely $B^+$ instead of $B^\phi$ in that case. Regardless of whether in this PR or in a follow-up PR, I would personally be in favor of not using the same naming convention to mean different things, unless there is a clear benefit. For example, we could spell out both variables as `Bplus` and `Bphi` or `B_plus` and `B_phi` (the latter with the understanding that `_` does not mean "subscript" in the covariant vs contravariant sense).

Source/Diagnostics/ReducedDiags/FieldMaximum.cpp
JustinRayAngus: I think theta and phi is more explicit. Or "th" "ph". 

"t" makes me think of time and "p" makes me think of particle.

EZoni: And the example in https://github.com/BLAST-WarpX/warpx/pull/5189#discussion_r2085721616 is another example of a case where `p` can be interpreted differently.

Source/FieldSolver/FiniteDifferenceSolver/FiniteDifferenceAlgorithms/CylindricalYeeAlgorithm.H
```diff
@@ -57,6 +58,11 @@ struct CylindricalYeeAlgorithm {
                                      (1._rt + multimode_alpha) / (dx[0]*dx[0])
                                     + 1._rt / (dx[1]*dx[1])
                               ) * PhysConst::c );
+#elif defined(WARPX_DIM_RCYLINDER)
+        // A slightly higher value of alpha is needed compared to RZ
```
RemiLehe:
```suggestion
        // The numerical value of the max Dt below was determined empirically (for one case)
```

Source/FieldSolver/FiniteDifferenceSolver/FiniteDifferenceAlgorithms/SphericalYeeAlgorithm.H
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /** Compute the maximum, CFL-stable timestep
+     *
+     * Compute the maximum timestep, for which the scheme remains stable
+     * under the Courant-Friedrichs-Levy limit.
+     */
+    static amrex::Real ComputeMaxDt (amrex::Real const * const dx) {
+        using namespace amrex::literals;
+        // Check this value! This was determined empirically (for one case)
```
RemiLehe:
```suggestion
        // The numerical value of the max Dt below was determined empirically (for one case)
```

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1737,6 +1833,25 @@ PhysicalParticleContainer::AddPlasmaFlux (PlasmaInjector const& plasma_injector,
...
+                // the radius ; thus, the calculation is finalized here
+                Real t_weight = flux * scale_fac * dt;
+                if (loc_flux_normal_axis == 0) {
+                    if (radially_weighted) {
+                         t_weight *= 4._rt/3._rt*MathConst::pi*radial_position*radial_position;
+                    } else {
+                         // This is not correct since it might shift the particle
+                         // out of the local grid
+                         ppos.x = std::cbrt(radial_position*rmax);
+                         t_weight *= dx[0];
```
EZoni, line 1851: Would there be a "correct" implementation instead?

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1687,6 +1759,30 @@ PhysicalParticleContainer::AddPlasmaFlux (PlasmaInjector const& plasma_injector,
...
+                // This because, when the flux direction is e.g. "r"
+                // the `inj_mom` objects generates a v*Gaussian distribution
+                // along the Cartesian "x" direction by default. This
+                // needs to be rotated along "r".
+                amrex::Real const ur = pu.x;
+                amrex::Real const ut = pu.y;
+                amrex::Real const up = pu.z;
+                pu.x = sin_phi*cos_theta*ur - sin_theta*ut + cos_phi*cos_theta*up;
+                pu.y = sin_phi*sin_theta*ur + cos_theta*ut + cos_phi*sin_theta*up;
+                pu.z = cos_phi*ur - sin_phi*up;;
```
EZoni: Typo:
```suggestion
                pu.z = cos_phi*ur - sin_phi*up;
```

Source/FieldSolver/FiniteDifferenceSolver/FiniteDifferenceAlgorithms/SphericalYeeAlgorithm.H
```diff
@@ -0,0 +1,162 @@
...
+        amrex::Real const * const coefs_r, int const /*n_coefs_r*/,
+        int const i, int const j, int const k, int const comp ) {
+
+        using namespace amrex::literals;
+
+        amrex::Real const inv_dr2 = coefs_r[0]*coefs_r[0];
+        return 1._rt/r * inv_dr2*( (r+0.5_rt*dr)*(F(i+1,j,k,comp) - F(i,j,k,comp))
+                                 - (r-0.5_rt*dr)*(F(i,j,k,comp) - F(i-1,j,k,comp)) );
+    }
+
```
RemiLehe: It looks like this is not currently used in the code ; therefore I would remove it for now.
```suggestion
```

Source/Particles/Gather/FieldGather.H
```diff
@@ -623,6 +683,87 @@ void doGatherShapeN ([[maybe_unused]] const amrex::ParticleReal xp,
...
+    // Gather field on particle Bthetap from field on grid by_arr
+    for (int ix=0; ix<=depos_order-galerkin_interpolation; ix++){
+        Bthetap += sx_by[ix]*by_arr(lo.x+j_by+ix, 0, 0, 0);
+    }
+
+    const amrex::Real rpxy = std::sqrt(xp*xp + yp*yp);
+    amrex::Real const costheta = (rpxy > 0. ? xp/rpxy : 1.);;
+    amrex::Real const sintheta = (rpxy > 0. ? yp/rpxy : 0.);;
+    amrex::Real const cosphi = (rp > 0. ? zp/rp : 1.);;
+    amrex::Real const sinphi = (rp > 0. ? rpxy/rp : 0.);;
```
EZoni: Typos:
```suggestion
    const amrex::Real rpxy = std::sqrt(xp*xp + yp*yp);
    amrex::Real const costheta = (rpxy > 0. ? xp/rpxy : 1.);
    amrex::Real const sintheta = (rpxy > 0. ? yp/rpxy : 0.);
    amrex::Real const cosphi = (rp > 0. ? zp/rp : 1.);
    amrex::Real const sinphi = (rp > 0. ? rpxy/rp : 0.);
```

Source/Initialization/InjectorPosition.H
```diff
@@ -51,8 +51,8 @@ struct InjectorPositionRandomPlane
         if (dir == 0) { return amrex::XDim3{0._rt, amrex::Random(engine), 0._rt}; }
         if (dir == 1) { return amrex::XDim3{amrex::Random(engine), amrex::Random(engine), 0._rt}; }
         else          { return amrex::XDim3{amrex::Random(engine), 0._rt, 0._rt }; }
-#elif (defined(WARPX_DIM_1D_Z))
-        // In 2D, the first components of the `XDim3` vector below correspond to z
+#elif (defined(WARPX_DIM_1D_Z) || defined(WARPX_DIM_RCYLINDER) || defined(WARPX_DIM_RSPHERE))
+        // In 1D, the first components of the `XDim3` vector below correspond to the first dimension
```
RemiLehe:
```suggestion
        // The first components of the `XDim3` vector below correspond to z for 1D Cartesian, and r for 1D cylindrical/spherical.
```

Source/Initialization/PlasmaInjector.cpp
```diff
@@ -388,24 +399,23 @@ void PlasmaInjector::setupNFluxPerCell (amrex::ParmParse const& pp_species)
 void PlasmaInjector::setupNuniformPerCell (amrex::ParmParse const& pp_species)
 {
     // Note that for RZ, three numbers are expected, r, theta, and z.
+    // For RCYLINDER, two numbers are expected, r, theta.
+    // For RSPHERE, three numbers are expected, r, theta, and phi
     // For 2D, only two are expected. The third is overwritten with 1.
     // For 1D, only one is expected. The second and third are overwritten with 1.
 #if defined(WARPX_DIM_1D_Z)
```
RemiLehe, line 406: Should it be this?
```suggestion
#if defined(WARPX_DIM_1D_Z) || defined(WARPX_DIM_RCYLINDER)
```

Source/Initialization/PlasmaInjector.cpp
```diff
@@ -388,24 +399,23 @@ void PlasmaInjector::setupNFluxPerCell (amrex::ParmParse const& pp_species)
...
     constexpr int num_required_ppc_each_dim = 1;
-#elif defined(WARPX_DIM_XZ)
+#elif defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RCYLINDER)
```
RemiLehe, line 408:
```suggestion
#elif defined(WARPX_DIM_XZ)
```

Source/Initialization/PlasmaInjector.cpp
```diff
@@ -388,24 +399,23 @@ void PlasmaInjector::setupNFluxPerCell (amrex::ParmParse const& pp_species)
 void PlasmaInjector::setupNuniformPerCell (amrex::ParmParse const& pp_species)
 {
     // Note that for RZ, three numbers are expected, r, theta, and z.
+    // For RCYLINDER, two numbers are expected, r, theta.
+    // For RSPHERE, three numbers are expected, r, theta, and phi
```
RemiLehe, line 398: Could you add this in the online documentation for the input parameters too?

Source/Initialization/WarpXInitData.cpp
```diff
@@ -103,6 +103,8 @@ namespace
             amrex::Print() << "Level " << lev << ": dt = " << dt[lev]
     #if defined(WARPX_DIM_1D_Z)
                            << " ; dz = " << dx_lev[0] << '\n';
+    #elif defined(WARPX_DIM_RCYLINDER) || defined(WARPX_DIM_RSPHERE)
+                           << " ; dx = " << dx_lev[0] << '\n';
     #elif defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
```
RemiLehe: This goes slightly beyond this PR, since it also modifies the behavior for RZ, but I think it's worth it.
```suggestion
    #elif defined(WARPX_DIM_RCYLINDER) || defined(WARPX_DIM_RSPHERE)
                           << " ; dr = " << dx_lev[0] << '\n';
    #elif defined(WARPX_DIM_RZ)
                           << " ; dr = " << dx_lev[0]
                           << " ; dz = " << dx_lev[1] << '\n';
    #elif defined(WARPX_DIM_XZ)
```

Source/WarpX.cpp
```diff
@@ -732,7 +744,7 @@ WarpX::ReadParameters ()
         );
 #endif
 
-#ifdef WARPX_DIM_RZ
+#if defined(WARPX_DIM_RZ) || defined(WARPX_DIM_RCYLINDER) || defined(WARPX_DIM_RSPHERE)
         const ParmParse pp_boundary("boundary");
         pp_boundary.query("verboncoeur_axis_correction", m_verboncoeur_axis_correction);
```
RemiLehe, line 749: Could you also update the online WarpX documentation for this parameter?

Source/WarpX.cpp
```diff
@@ -3067,6 +3108,9 @@ WarpX::UpperCorner(const Box& bx, const int lev, const amrex::Real time_shift_de
 #elif defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
     return { grid_max[0] + galilean_shift[0], std::numeric_limits<Real>::max(), grid_max[1] + galilean_shift[1] };
 
+#elif defined(WARPX_DIM_RCYLINDER) || defined(WARPX_DIM_RSPHERE)
+    return { grid_max[0] + galilean_shift[0], std::numeric_limits<Real>::max(), std::numeric_limits<Real>::max() };
```
RemiLehe: Should we raise an exception here if `galilean_shift[0]` is different than 0?
```suggestion
    return { grid_max[0], std::numeric_limits<Real>::max(), std::numeric_limits<Real>::max() };
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -308,23 +308,38 @@ public:
...
+            // This needs double checking
+            int const ri = i_cell;
+            // rr is radius at the cell center
+            amrex::ParticleReal const rr = (ri + 0.5_prt)*dr;
+            return 4.0_prt*MathConst::pi*rr*rr;
 #else
             // No factor is needed for Cartesian
             amrex::ignore_unused(i_cell);
-            return 1._prt;
+            return 1.0_prt;
```
RemiLehe, line 342: In addition to the above changes, the particle's velocity vector should be mapped:
https://github.com/BLAST-WarpX/warpx/blob/development/Source/Particles/Collision/BinaryCollision/Coulomb/ElasticCollisionPerez.H#L109

Source/Particles/PhysicalParticleContainer.cpp
dpgrote: This is implemented in PR #5009.

Source/Particles/PhysicalParticleContainer.cpp
dpgrote: This is fixed.

Source/WarpX.cpp
dpgrote: Fixed. Now raises an error if boosted frame is specified.

Source/FieldSolver/FiniteDifferenceSolver/EvolveB.cpp
dpgrote: These were changed to Bphi etc

Docs/source/usage/parameters.rst
```diff
@@ -1072,11 +1073,12 @@ Particle initialization
 
     * ``none``: Do not inject macro-particles (for example, in a simulation that starts with neutral, ionizable atoms, one may want to create the electrons species -- where ionized electrons can be stored later on -- without injecting electron macro-particles).
 
-* ``<species_name>.num_particles_per_cell_each_dim`` (`3 integers in 3D and RZ, 2 integers in 2D`)
+* ``<species_name>.num_particles_per_cell_each_dim`` (`3 integers in 3D, RZ, RSPHERE, 2 integers in 2D and RCYLNDER`)
```
```suggestion
* ``<species_name>.num_particles_per_cell_each_dim`` (`3 integers in 3D, RZ, RSPHERE, 2 integers in 2D and RCYLINDER`)
```

Regression/Checksum/benchmarks_json/test_rsphere_langmuir_multi.json
```diff
@@ -1,28 +1,27 @@
 {
   "lev=0": {
-    "Bt": 4.955190029855237e-15,
-    "Er": 16181760914.279827,
-    "jr": 863095696601.4752,
-    "rho_electrons": 18595449.23275246,
-    "rho_ions": 18494177.078754093
+    "Bt": 4.521366416357738e-15,
```
EZoni, line 3: Would it be problematic (e.g., for post-processing tools) to update also this name from `Bt` to `Btheta` (as well as others, e.g., `Bp`, `Et`, `Ep`, etc.)?

Docs/source/developers/dimensionality.rst
```diff
@@ -15,9 +15,11 @@ Dimensions  CMake Option
 **2D3V**    ``WarpX_DIMS=2``
 **1D3V**    ``WarpX_DIMS=1``
 **RZ**      ``WarpX_DIMS=RZ``
+**RCYLINDER**   ``WarpX_DIMS=RCYLINDER``
+**RSPHERE**     ``WarpX_DIMS=RSPHERE``
```
RemiLehe:
```suggestion
**RZ3V**      ``WarpX_DIMS=RZ``
**RCYLINDER3V**   ``WarpX_DIMS=RCYLINDER``
**RSPHERE3V**     ``WarpX_DIMS=RSPHERE``
```

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -300,49 +300,49 @@ void FiniteDifferenceSolver::CalculateCurrentAmpereSpherical (
 
         // Extract tileboxes for which to loop
         Box const& tjr  = mfi.tilebox(Jfield[0]->ixType().toIntVect());
-        Box const& tjt  = mfi.tilebox(Jfield[1]->ixType().toIntVect());
+        Box const& tjtheta  = mfi.tilebox(Jfield[1]->ixType().toIntVect());
         Box const& tjp  = mfi.tilebox(Jfield[2]->ixType().toIntVect());
```
EZoni: Just noting that here `tjp` has not been renamed to `tjphi` yet.

Docs/source/usage/parameters.rst
```diff
@@ -432,13 +432,16 @@ Setting up the field mesh
...
-    For ``RZ``, we apply an azimuthal mode decomposition, with ``warpx.n_rz_azimuthal_modes`` providing further control.
+    Supported values are ``1``, ``2``, ``3``, ``RZ``, ``RCYLINDER``, and ``RSPHERE``.
 
-    Note that this value has to match the :ref:`WarpX_DIMS <building-cmake-options>` compile-time option.
+    * For ``3``, a cartesian geometry of ``x``, ``y``, ``z`` is modeled.
+    * For ``2``, the axes are ``x`` and ``z`` and all physics in ``y`` is assumed to be translation symmetric.
+    * For ``1``, the only axis is ``z`` and the dimensions ``x`` and ``y`` are translation symmetric.
+    * For ``RZ``, we apply an azimuthal mode decomposition, with ``warpx.n_rz_azimuthal_modes`` providing further control.
+    * For ``RCYLINDER``, one dimensional cylinder, radius
+    * For ``RSPHERE``, one dimensional sphere, radius
```
EZoni: Should we say something more verbose like in the entries above?
```suggestion
    * For ``RCYLINDER``, the only axis is `r` and all physics is invariant under rotations in `theta` and translations in `z`.
    * For ``RSPHERE``, the only axis is `r` and all physics is invariant under rotations in `theta` and `phi`.
```
Not sure if this suggestion is the most accurate way of describing the two geometries.

Regression/Checksum/benchmarks_json/test_rsphere_langmuir_multi.json
dpgrote: For plotfiles this would be no problem. Not sure about OpenPMD. Note that this could be a separate PR since the same change could be made to the theta component in RZ.

Docs/source/usage/parameters.rst
```diff
@@ -2977,7 +2981,7 @@ In-situ capabilities can be used by turning on Sensei or Ascent (provided they a
 
 * ``<diag_name>.<species_name>.variables`` (list of `strings` separated by spaces, optional)
     List of particle quantities to write to output.
-    Choices are ``x``, ``y``, ``z`` for the particle positions (3D and RZ), ``x`` & ``z`` in 2D, ``z`` in 1D,
+    Choices are ``x``, ``y``, ``z`` for the particle positions (3D, RZ, RSPERE), ``x`` and ``z`` in 2D, ``z`` in 1D, ``x`` and ``y`` for RCYLINDER,
```
```suggestion
    Choices are ``x``, ``y``, ``z`` for the particle positions (3D, RZ, RSPHERE), ``x`` and ``z`` in 2D, ``z`` in 1D, ``x`` and ``y`` for RCYLINDER,
```

Docs/source/usage/parameters.rst
dpgrote: @EZoni I've updated this. To be consistent, I changed the description for all geometries. How is this?

RemiLehe (approved):

CMakeLists.txt
ax3l: This would be rough to type @EZoni , we need something that is developer command-line friendly (short) like `1;2;RZ;3` are.
(For all other cases, I am also fully for spelling things out.)

We had the double-option for a while in the past and it is confusing to use/less clear than single options.

EZoni: Yes, I'm aware. I think I moved the conversation from here to #5946 a while ago. Recentely @aeriforme commented on that, too.

EZoni: This was the final suggestion that came up in that issue:
```
WarpX_GEOM=1D_CAR
WarpX_GEOM=2D_CAR
WarpX_GEOM=3D_CAR
WarpX_GEOM=1D_CYL
WarpX_GEOM=1D_SPH
WarpX_GEOM=2D_CYL
```

PR #5196, opened 2024-08-30
Set t_old

dpgrote:
The `t_old` variable wasn't being updated. This fixes that, updating it during evolve.

But note that `t_old` is never used so it could possibly be deleted. However, it is written out to the `WarpXHeader` file so deleting it would change the format of that file. This matters since that file is used for checkpoint/restart, so changing its format would break old checkpoints. So, probably best to just keep it around since it doesn't cost much. It might be useful in the Python interface.


lucafedeli88 (approved):
Thanks for this PR!

PR #5197, opened 2024-08-30
Express differential luminosity diagnostics in eV

RemiLehe:
Given the applications for this reduced diagnostic, I think it makes sense to express the center-of-mass energy in eV rather than in Joules.


ax3l (approved):
Makes sense :+1:

PR #5198, opened 2024-08-30
Adding path to cmake for lapackpp so it can find installed blaspp on …

clarkse:
…perlmutter.

Error during dependency reinstall on perlmutter showed up as:

```
-- Check for BLAS++
CMake Warning at CMakeLists.txt:710 (find_package):
  By not providing "Findblaspp.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "blaspp", but
  CMake did not find one.

  Could not find a package configuration file provided by "blaspp" with any
  of the following names:

    blasppConfig.cmake
    blaspp-config.cmake

  Add the installation prefix of "blaspp" to CMAKE_PREFIX_PATH or set
  "blaspp_DIR" to a directory containing one of the above files.  If "blaspp"
  provides a separate development package or SDK, be sure it has been
  installed.


CMake Error at CMakeLists.txt:714 (message):
  BLAS++ not found.  LAPACK++ requires BLAS++ to be installed first.


-- Configuring incomplete, errors occurred!

```

Adding the blaspp install directory fixed this issue.

### DIFF
--- a/Tools/machines/perlmutter-nersc/install_gpu_dependencies.sh
+++ b/Tools/machines/perlmutter-nersc/install_gpu_dependencies.sh
@@ -103,7 +103,7 @@ else
   git clone -b v2024.05.31 https://github.com/icl-utk-edu/lapackpp.git $HOME/src/lapackpp
 fi
 rm -rf $HOME/src/lapackpp-pm-gpu-build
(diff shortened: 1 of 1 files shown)


ax3l:
Hi @clarkse ,

Thanks for the PR!

I think this might not be needed because there is an env hint already in the profile:
https://github.com/ECP-WarpX/WarpX/blob/80bef35bf142098c960f8e8bcd6248f4d78492ee/Tools/machines/perlmutter-nersc/perlmutter_gpu_warpx.profile.example#L24

Did you update the profile to this updated path already?

clarkse:
Well that would do it. Thanks for looking @ax3l

PR #5201, opened 2024-09-02
Fix directional comparisons in assertions

bnara:
Lorentz boost direction and laser propagation direction always evaluate as being in the same direction, assuming both are normalized.  For example, m_nvec = {0, 0, -1} and boost_direction = {0, 0, 1} failed to trigger the assertion.

Replaced faulty comparisons with idiom used elsewhere:
https://github.com/ECP-WarpX/WarpX/blob/cc6e7ea1074c0d9e97cdf60363d772c119cc1dea/Source/Initialization/WarpXInitData.cpp#L803

### DIFF
--- a/Source/Particles/LaserParticleContainer.cpp
+++ b/Source/Particles/LaserParticleContainer.cpp
@@ -181,10 +181,11 @@ LaserParticleContainer::LaserParticleContainer (AmrCore* amr_core, int ispecies,
 
     if (WarpX::gamma_boost > 1.) {
         // Check that the laser direction is equal to the boost direction
-        AMREX_ALWAYS_ASSERT_WITH_MESSAGE(  m_nvec[0]*WarpX::boost_direction[0]
-                                         + m_nvec[1]*WarpX::boost_direction[1]
-                                         + m_nvec[2]*WarpX::boost_direction[2] - 1. < 1.e-12,
-                                           "The Lorentz boost should be in the same direction as the laser propagation");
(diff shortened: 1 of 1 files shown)


ax3l (approved):
Thank you for your first PR to WarpX :tada:, good catch.
Any negative (in z) laser normal, including non-parallel for boost, passed.

ax3l:
Did you check if initializing a laser _perfectly anti-parallel_ to the boost was working correctly? In that case, we could relax the assert & message to parallel and anti-parallel to the boost.

Is that a science case you want to study in the boosted frame @bnara (e.g., some counter-propagating laser pulse)?

bnara:
Here is a comparison of laser direction along boost (direction = 0 0 1) versus laser direction against boost (direction = 0 0 -1):
<video src="https://github.com/user-attachments/assets/4ff137e4-6b9f-414a-aa6f-9570a2df3a35" loop controls muted></video>

In both cases, beta0=0.4 is the boosted frame's speed and the moving window's speed, with respect to the unboosted lab frame.

When the laser direction is along the boost, the antenna correctly stays fixed in the lab frame.  When the laser direction is against the boost, however, the antenna is translating in the wrong direction in the moving window, corresponding to motion in the lab frame.

@ax3l, yes, I'm interested in looking at counterpropagating laser pulses for SFQED studies.  Because the antenna radiates both upstream and downstream, it looks like setting the laser direction to be along the boost gives me the properly blue-shifted laser in the counter-propagating direction!

[warpx.txt](https://github.com/user-attachments/files/16899182/warpx.txt)

PR #5203, opened 2024-09-03
[WIP] CI: 30-40sec install w/ pixi

ax3l:
Update our unit tests to use:
- Ubuntu 22.04 (from 20.04)
- GCC 12 (from GCC 9)

This speeds up dependency installs significantly.

- [ ] many test are too unstable to produce the same checksums and their tolerances need to be bumped
- [x] rebase after #5068 was merged
- [x] some analysis scripts might fail with the SciPy update (i.e., see #5661)

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -1,7 +1,7 @@
 # -*- mode: yaml -*-
 
 pool:
-  vmImage: 'ubuntu-20.04'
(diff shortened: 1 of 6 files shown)


RemiLehe:
The CI has since moved directly to Ubuntu 24.04 (without pixi), entirely superseding this PR.

PR #5208, opened 2024-09-03
Implement injection of particles from the embedded boundary

RemiLehe:
# Overview

This PR implements flux injection of particles from the embedded boundary. 

It also adds a test that emits particles from a sphere in 3D as represented here:
movie
as well as RZ and 2D versions of this test. (In 2D, the particles are emitted from a cylinder.)

As can be seen in the above movie, particles are emitted from a single point within each cell (the centroid of the EB), instead of being emitted uniformly on the surface of the EB within the cell. This could be improved in a future PR.

The implementation as well as the user interface largely re-use the infrastructure for the flux injection from a plane. However, as a result, the user interface is perhaps not very intuitive. In particular, when specify the velocity distribution, `uz` represents the direction normal to the EB while `ux`, `uy` represent the tangential directions. This again will be improved in follow-up PR.

# Follow-up PRs

- [ ] Change the interface of `gaussianflux` so as to specify the tangential and normal distribution. In other words, instead of:
```
electron.momentum_distribution_type = gaussianflux
electron.ux_th = 0.01
electron.uy_th = 0.01
electron.uz_th = 0.1
electron.uz_m = 0.07
```
we would do:
```
electron.momentum_distribution_type = gaussianflux
electron.u_tangential_th = 0.01  # Tangential to the emitting surface
electron.u_normal_th = 0.1   # Normal to the emitting surface
electron.u_normal_m = 0.07
```

- [ ] Change the interface so that the user does not need to specify the number of macroparticles per cell (which is problematic for EB, since difference cell contain different EB surface, and should in general emit different numbers of macroparticles). Instead, we would specify the weight of macroparticles, i.e. instead of 
```
electron.injection_style = NFluxPerCell
electron.num_particles_per_cell = 100   
electron.flux_function(x,y,z,t) = "1.”
```
we would do
```
electron.injection_style = NFluxPerCell
electron.flux_macroweight = 200   # Number of physical particles per macroparticle
electron.flux_function(x,y,z,t) = "4e12” # Number of physical particles emitted per unit time and surface  
```

- [ ] Add a way for the user to specify the total flux across the whole emitting surface
Example:
```
electron.flux_function(x,y,z,t) = "(x>-1)*(x<1)"
electron.total_flux = 4e12 # physical particle / second (not per unit area)
```
(In that case, `flux_function` would be rescaled internally by WarpX so as to emit the right number of particles.)

- [ ] Add PICMI interface
- [ ] Emit the particles uniformly from the surface of the EB within one cell


Examples/Tests/flux_injection/inputs_test_3d_flux_injection_from_eb
```diff
@@ -0,0 +1,52 @@
...
+geometry.prob_hi     =  4  4  4
+
+# Deactivate Maxwell solver
+algo.maxwell_solver = none
+warpx.const_dt = 7e-9
+
+# Boundary condition
+boundary.field_lo = periodic periodic periodic
+boundary.field_hi = periodic periodic periodic
+warpx.eb_implicit_function = "-(x**2+y**2+z**2-0.1**2)"
```
RemiLehe, line 29:
```suggestion
warpx.eb_implicit_function = "-(x**2+y**2+z**2-0.1**2)"
```
```suggestion
warpx.eb_implicit_function = "-(x**2+y**2+z**2-2**2)"
```
```suggestion
warpx.eb_implicit_function = "-(x**2+y**2+z**2-0.1**2)"
```

RemiLehe:
```suggestion
warpx.eb_implicit_function = "-(x**2+y**2+z**2-2**2)"
```

Source/Particles/AddPlasmaUtilities.H
```diff
@@ -22,6 +22,29 @@
 #include <AMReX_REAL.H>
 #include <AMReX_RealBox.H>
 
+struct PDim3 {
```
RemiLehe, line 25: I had to move this object from `PhysicalParticleContainer.cpp` to `AddPlasmaUtilities.H` as part of this PR.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -147,29 +147,6 @@ namespace
...
-        AMREX_GPU_HOST_DEVICE
-        PDim3(PDim3 const &)            = default;
-        AMREX_GPU_HOST_DEVICE
-        PDim3& operator=(PDim3 const &) = default;
-        AMREX_GPU_HOST_DEVICE
-        PDim3(PDim3&&)                  = default;
-        AMREX_GPU_HOST_DEVICE
-        PDim3& operator=(PDim3&&)       = default;
-    };
-
```
RemiLehe, line 172: I had to move this object from `PhysicalParticleContainer.cpp` to `AddPlasmaUtilities.H` as part of this PR.

Source/Particles/AddPlasmaUtilities.H
```diff
@@ -71,12 +94,110 @@ int compute_area_weights (const amrex::IntVect& iv, const int normal_axis) {
     return r;
 }
 
+
+#ifdef AMREX_USE_EB
+/*
+  This computes the scale_fac (used for setting the particle weights) on a on area basis
```
dpgrote: This should use the standard documentation format, like
```
    /*
     * \brief This computes...
     * 
     * \param[in] dx
     * /
```
Though I do notice that no other docs in this file match this format.

Source/Particles/AddPlasmaUtilities.H
```diff
@@ -71,12 +94,110 @@ int compute_area_weights (const amrex::IntVect& iv, const int normal_axis) {
...
+    // The minus sign below takes into account the fact that eb_bnd_normal_arr
+    // points towards the covered region, while particles are to be emitted
+    // *away* from the covered region.
+    amrex::Real const nx = -eb_bnd_normal_arr(i,j,k,0);
+    amrex::Real const ny = -eb_bnd_normal_arr(i,j,k,1);
+    amrex::Real const nz = -eb_bnd_normal_arr(i,j,k,2);
+
+    // Rotate the momentum along the theta direction (in the z-x plane)
+    amrex::Real const cos_theta = nz;
+    amrex::Real const sin_theta = std::sqrt(1-nz*nz);
```
dpgrote: 1._rt

To simplify this, could the theta and phi rotations be combined into one set of expressions, a 3D rotation?

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1371,6 +1348,22 @@ PhysicalParticleContainer::AddPlasmaFlux (PlasmaInjector const& plasma_injector,
     const auto dx = geom.CellSizeArray();
     const auto problo = geom.ProbLoArray();
 
+#ifdef AMREX_USE_EB
+    bool inject_from_eb = plasma_injector.m_inject_from_eb; // whether to inject from EB or from a plane
```
dpgrote: `bool const`

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1450,24 +1454,57 @@ PhysicalParticleContainer::AddPlasmaFlux (PlasmaInjector const& plasma_injector,
...
-                }
-                const int num_ppc_int = static_cast<int>(num_ppc_real*r + amrex::Random(engine));
-                pcounts[index] = num_ppc_int;
+                // Injection from a plane
+                auto lo = getCellCoords(overlap_corner, dx, {0._rt, 0._rt, 0._rt}, iv);
+                auto hi = getCellCoords(overlap_corner, dx, {1._rt, 1._rt, 1._rt}, iv);
+                // Skip cells that do not overlap with the plane
+                if (!flux_pos->overlapsWith(lo, hi)) { return; }
+            }
+            const int num_ppc_int = static_cast<int>(num_ppc_real_in_this_cell + amrex::Random(engine));
```
dpgrote: This looses the change made in PR #5272. The calculation of `num_ppc_int` should happen after the `r` factor is included.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1659,6 +1736,7 @@ PhysicalParticleContainer::AddPlasmaFlux (PlasmaInjector const& plasma_injector,
                 // For cylindrical emission (flux_normal_axis==0
                 // or flux_normal_axis==2), the emission surface depends on
                 // the radius ; thus, the calculation is finalized here
+
```
dpgrote: Can this blank line be delete? It's an unrelated format change.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1548,13 +1593,32 @@ PhysicalParticleContainer::AddPlasmaFlux (PlasmaInjector const& plasma_injector,
...
-                  // Otherwise: use 1 as the refinement ratio
-                  flux_pos->getPositionUnitBox(i_part, amrex::IntVect::TheUnitVector(), engine);
-                auto pos = getCellCoords(overlap_corner, dx, r, iv);
+                // Determine the position of the particle within the cell
+                XDim3 pos;
+                XDim3 r;
+#ifdef AMREX_USE_EB
+                if (inject_from_eb) {
+#if defined(WARPX_DIM_3D)
+                    pos.x = overlap_corner[0] + (iv[0] + 0.5_rt + eb_bnd_cent_arr(i,j,k,0))*dx[0];
```
dpgrote, line 1602: Is this putting all of the particles at the same location, the EB boundary center? Is there a way to distribute the particles around on the EB surface?

RemiLehe: Yes, that's absolutely correct: all the particles are emitted from a single point. (See also the PR description.) 
We should indeed distribute the particles over the EB surface. This should be doable but will take a bit more time. This will be added in a follow-up PR.

dpgrote: I should have read the whole PR description :)

dpgrote (approved):
Looks good! Thanks!

Issue #5192, opened 2024-08-29
Save an evaluated inputs file on start


titoiride:
Hi all,
talking with @ax3l  and @n01r we thought it would be cool and useful to have WarpX dump an "evaluated" version of the used inputs on start, particularly when using text files as input.
It currently returns the "used input" file, but that only lists the input as they were requested in the file.
If some of the parameters are expressed using constants or mathematical expressions, the file returns the expression as it is.
Perhaps in addition to the file already dumped it would be useful to have one where constants and mathematical expressions are return evaluated so that it's possible to catch bugs related to their definition.

ax3l:
@WeiqunZhang, do you think we could create a file similar to `warpx_used_inputs` that has `amrex::ParmParse` constants evaluated in it? Something like `warpx_evaluated_inputs`.

ax3l:
Example:
https://github.com/ECP-WarpX/WarpX/blob/ce7f5cc28cdfa0710f4199912caef4c4d9d2f161/Examples/Tests/Implicit/inputs_1d#L1-L24

Regarding replaced for `amr.n_cell` and `geometry.prob_hi`.

ax3l:
@WeiqunZhang mentioned that currently we will see in `warpx_used_inputs`:
```
amr.n_cell = nz
amr.n_cell = 40
geometry.prob_hi = 10.*de0 
geometry.prob_hi = <single-number>
```
(related to  https://github.com/AMReX-Codes/amrex/issues/4087 ). We recently updated the sorting, before that the lines were a bit more apart from each other, this was pulled into ` development` last week: https://github.com/AMReX-Codes/amrex/pull/4101

WeiqunZhang:
@titoiride Do you want a separate file or something like below?
```
amr.n_cell = nz # 40
```

@ax3l @RemiLehe

titoiride:
Personally I have no strong preference, I like this commented mode.

PR #5216, opened 2024-09-04
Use amrex::ParmParse::prettyPrintTable

WeiqunZhang:
Instead of `amrex::ParmParse::dumpTable`, we use `prettyPrintTable` that removes duplicates.

Related to #5192 and https://github.com/AMReX-Codes/amrex/issues/4087
Improves #3132


ax3l (approved):
Nice, thank you!

lucafedeli88 (approved):

Issue #5199, opened 2024-08-31
Optimized PushPX 


liuyuchuncn:
**Kernel : PushPX general  local memory , We want to analyze the causes of local memory and how to optimize local memory**
As shown in the figure below, the array produces local memory `amrex::Real sx_node_galerkin[depos_order + 1 - galerkin_interpolation] = {0._rt}; ``amrex::Real sx_cell_galerkin[depos_order + 1 - galerkin_interpolation] = {0._rt};`

![image](https://github.com/user-attachments/assets/ee74a95f-a2e5-415f-8161-cb5b45d0f46c)

and `amrex::Real sx_node_galerkin[depos_order + 1 - galerkin_interpolation] = {0._rt}; ``amrex::Real sx_cell_galerkin[depos_order + 1 - galerkin_interpolation] = {0._rt};` used to compute ShapeFactor 
 https://github.com/ECP-WarpX/WarpX/blob/5b34b84dfc588c8f6cf8088e94cd884e850344f7/Source/Particles/ShapeFactors.H#L55-L70

liuyuchuncn:
```
  // Gather field on particle Exp from field on grid ex_arr
 for (int iz=0; iz<=depos_order; iz++){
        for (int iy=0; iy<=depos_order; iy++){
            for (int ix=0; ix<= depos_order - galerkin_interpolation; ix++){
                Exp += sx_ex[ix]*sy_ex[iy]*sz_ex[iz]*
                    ex_arr(lo.x+j_ex+ix, lo.y+k_ex+iy, lo.z+l_ex+iz);
            }
        }
    }
```
#### ex_arr 3D Index calculations are complex
 ex_arr(lo.x+j_ex+ix, lo.y+k_ex+iy, lo.z+l_ex+iz)       **Resolves to**
 p[(lo.x+j_ex+ix-begin.x)+(lo.y+k_ex+iy-begin.y)*jstride+(lo.z+l_ex+iz-begin.z)*kstride];
#### Partial assembly code
![image](https://github.com/user-attachments/assets/5529c988-c331-4ef3-903e-5bab76ffdac6)


**so . I think the bottleneck is additive calculations**

ax3l:
Hi @liuyuchuncn,

Thank you for profiling this kernel and for aiming to improve its performance, this is super welcome! :rocket: :sparkles: 

[...]

PR #5217, opened 2024-09-04
GatherAndPush: Use CTO ParallelFor

WeiqunZhang:
The GPU kernel in GatherAndPush contains Parsers for computing external E and B fields. It's known that even if the Parsers are not used at run time, the occupancy on AMD GPUs are still affected. This is probably reason why this kernel is 50% slower on MI200 than on A100. Using CTO ParallelFor should improve the performance.

Related to #5199

### DIFF
--- a/Source/Fluids/WarpXFluidContainer.cpp
+++ b/Source/Fluids/WarpXFluidContainer.cpp
@@ -1072,7 +1072,8 @@ void WarpXFluidContainer::GatherAndPush (
 
 #ifdef AMREX_USE_CUDA
                     amrex::ignore_unused(Exfield_parser, Eyfield_parser, Ezfield_parser,
-                                         Bxfield_parser, Byfield_parser, Bzfield_parser, gamma_boost);
+                                         Bxfield_parser, Byfield_parser, Bzfield_parser,
+                                         gamma_boost, problo, dx, t, beta_boost);
 #endif
 
                     if constexpr (boost_control == has_gamma_boost) { // Lorentz transform fields due to moving frame
--- a/Source/Fluids/WarpXFluidContainer.cpp
+++ b/Source/Fluids/WarpXFluidContainer.cpp
@@ -1055,6 +1055,7 @@ void WarpXFluidContainer::GatherAndPush (
                 if (N_arr(i,j,k)>0.0) {
 
                     // Interpolate fields from tmp to Nodal points
+                    // NOLINTBEGIN(misc-const-correctness)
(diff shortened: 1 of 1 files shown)


Source/Fluids/WarpXFluidContainer.cpp
```diff
@@ -1055,9 +1068,15 @@ void WarpXFluidContainer::GatherAndPush (
...
+
+#ifdef AMREX_USE_CUDA
+                    amrex::ignore_unused(Exfield_parser, Eyfield_parser, Ezfield_parser,
+                                         Bxfield_parser, Byfield_parser, Bzfield_parser, gamma_boost);
+#endif
 
-                    if (gamma_boost > 1._rt) { // Lorentz transform fields due to moving frame
-                        if ( ( external_b_fields ) || ( external_e_fields ) ){
+                    if constexpr (boost_control == has_gamma_boost) { // Lorentz transform fields due to moving frame
+                        if constexpr (exte_control == has_exte || extb_control == has_extb) {
```
ax3l, line 1080: Wow, NVCC is picky again and needs a few temporaries before we go in the `if constexpr`:
```
error: An extended __device__ lambda cannot first-capture variable in constexpr-if context
```
for
- `problo`
- `dx`
- `t`
- `beta_boost`
- ...

ax3l:
@WeiqunZhang While this file for the fluid solver definitely needs an update as well, I think the kernel benchmarked in #5199 is called from `PhysicalParticleContainer::Evolve` and might refer to `PhysicalParticleContainer::PushPX` / `PhysicalParticleContainer::PushP` ( / `PhysicalParticleContainer::ImplicitPushXP`)?

Source/Fluids/WarpXFluidContainer.cpp
WeiqunZhang: Oh, right. We can add them to amrex::ignore_unused.

liuyuchuncn:
Sorry ！！！ @WeiqunZhang @ax3l ， The title causes you to misunderstand ,so I modified the Title
Kernel : PushPX refer to `PhysicalParticleContainer::Evolve` `PhysicalParticleContainer::PushPX`

ax3l:
That makes sense. I think our `BL_Profile` (TinyProfiler entry) for these kernels still reads `GatherAndPush` as well, which caused the confusion.

Funny enough, the kernel here needs the same work, so I will merge this performance improvement for the fluid solver :)

ax3l (approved):

PR #5220, opened 2024-09-06
CTest follow-up: remove obsolete files/scripts

EZoni:
Follow-up on #5068. Since CTest is the new regression testing framework, it should be safe to remove all files and scripts related to the old regression testing framework.


ax3l (approved):
Nice :sparkling_heart:

PR #5226, opened 2024-09-08
General moving-window transformations in boosted-frame simulations

bnara:
During initialization in boosted-frame simulations, the moving window, of length $L$ given in lab-frame coordinates, is resized along the direction of the boost.  Currently, the code assumes that the moving window is traveling along the direction of the boost at the speed-of-light.

This PR generalizes the transformation of the moving window from the lab-frame to the boosted-frame, taking into account the direction and speed of the moving window (c.f., Eq. 1 in Ref. [1]).  As an example, let's say the boosted-frame moves with speed $\beta = 0.6$ (`gamma_boost = 1.25`).   If the moving window also travels at speed $\beta$, then its length in the boosted frame is $L \gamma = 1.25 L$ rather than $2 L$, as is currently calculated.

Special cases for negative values of `moving_window_v` and moving windows not traveling along the boost direction were both checked.  In order to have access to the moving window parameters, there needed to be some rearrangement of initialization order.

[1] J.-L. Vay, Phys. Rev. Lett. **98**, 130405 (2007)


bnara:
The diagnostic field output window, specified via `diag_lo` and `diag_hi`, could be transformed in the same way:
https://github.com/ECP-WarpX/WarpX/blob/78ff8bf92b9cfebbca19ffdb1202cd04f7703b35/Source/Diagnostics/Diagnostics.cpp#L229

RemiLehe (approved):

Source/Utils/WarpXUtil.cpp
```diff
@@ -194,8 +231,11 @@ void ConvertLabParamsToBoost()
     {
         if (boost_direction[dim_map[idim]]) {
             amrex::Real convert_factor;
-            // Assume that the window travels with speed +c
-            convert_factor = 1._rt/( gamma_boost * ( 1 - beta_boost ) );
+            amrex::Real beta_window = beta_boost;
```
RemiLehe, line 234: Shouldn't this be 
```
amrex::Real beta_window = 0;
```
i.e. if `do_moving_window` is false, the velocity of the moving window is effectively 0?

RemiLehe:
Thanks for this PR!
Some of the tests were failing because they produce different results. (Since the box size is now different.)
I udpated the input script so that, with the new code, it produces the same results as the old input script with the old code.

I also had a question (see inline comment), but have not modified the code according to what I was suggesting in this question.

Source/Utils/WarpXUtil.cpp
RemiLehe: Apologies, I got confused. After more thought, I agree that this should be `amrex::Real beta_window = beta_boost;` in the absence of an explicit moving window.

PR #5227, opened 2024-09-08
[WIP] ABLASTR: IntervalsParser & SafeCast

ax3l:
Refactor and move out the `SafeCast` and `IntervalsParser` from WarpX to ABLASTR.


Source/ablastr/utils/IntervalsParser.cpp
```diff
@@ -0,0 +1,176 @@
...
+        instr, m_separator, true);
+
+    if(insplit.size() == 1){ // no colon in input string. The input is the period.
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(!m_isBTD, "must specify interval stop for BTD");
+        m_period = parseStringtoInt(insplit[0], "interval period");}
+    else if(insplit.size() == 2) // 1 colon in input string. The input is start:stop
+    {
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(!m_isBTD || !insplit[1].empty(), "must specify interval stop for BTD");
+        if (!insplit[0].empty()){
+            m_start = parseStringtoInt(insplit[0], "interval start");}
```
ax3l, line 36: `parseStringtoInt`: @WeiqunZhang can I use an AMReX intrinsic for this now?

Source/ablastr/utils/IntervalsParser.cpp
```diff
@@ -0,0 +1,176 @@
...
+#include "Utils/TextMsg.H"
+
+#include <AMReX_Utility.H>
+
+#include <algorithm>
+
+
+ablastr::utils::parser::SliceParser::SliceParser (
+    const std::string& instr,
+    const bool isBTD
```
ax3l, line 21: Urgh, such a hack: The `isBTD` should never have made it into this interface and needs to be removed. `BTDIntervalsParser` can handle this.

Source/ablastr/utils/IntervalsParser.cpp
WeiqunZhang: I looked at the implementation of that function. It use amrex::Parser to parse it as double precision numbers and then check if it can be safely cast to int. There is amrex::IParser we can use. It parses the numbers as long long. So the chance of overflowing during parsing is tiny. The safety checking for casting to int would also be much simpler.

ax3l:
new attempt in #6933

PR #5232, opened 2024-09-09
Reduce time in CI tests

oshapoval:
Reduced time of several CI tests by reducing the resolution:

- test_3d_beam_beam_collision;
- inputs_test_rz_multiJ_psatd.


aeriforme:
We're waiting to update the checksums, right?
Also, we might need to update the documentation of the beam-beam collision example.

ax3l:
@aeriforme is this good to go from your end or does it need docs updates?

ax3l (approved):

ax3l:
ping @aeriforme @EZoni

EZoni:
> ping @aeriforme @EZoni

I'm not familiar with the beam-beam collision example, so I'm not sure what would need to be changed in the README/docs of the example, after the changes here. I would wait for more instructions from @aeriforme about this.

oshapoval:
Note: in the input, we already mention the resolution needed for full-scale simulation: nx, ny, nz = 512, 512, 1024.

PR #5246, opened 2024-09-10
[Hackathon] Update Source/Parallelization/WarpXComm_K.H

dpgrote:
This cleans up `warpx_interp` by using inline macros. For example, changing
```
    const int rj = rr[0];
#if defined(WARPX_DIM_1D_Z)
    constexpr int rk = 1;
    constexpr int rl = 1;
    const     int rk = rr[1];
    constexpr int rl = 1;
#else
    const int rk = rr[1];
    const int rl = rr[2];
#endif
```
to
```
    const int rj = rr[0];
    const int rk = (AMREX_SPACEDIM > 1) ? rr[1] : 1;
    const int rl = (AMREX_SPACEDIM > 2) ? rr[2] : 1;
```

### DIFF
--- a/Source/Parallelization/WarpXComm_K.H
+++ b/Source/Parallelization/WarpXComm_K.H
@@ -44,18 +44,18 @@ void warpx_interp (int j, int k, int l,
 
     // Refinement ratio
     const int rj = rr[0];
-    const int rk = (AMREX_SPACEDIM == 1) ? 1 : rr[1];
-    const int rl = (AMREX_SPACEDIM <= 2) ? 1 : rr[2];
+    const int rk = (AMREX_SPACEDIM > 1) ? rr[1] : 1;
+    const int rl = (AMREX_SPACEDIM > 2) ? rr[2] : 1;
 
     // Staggering (0: cell-centered; 1: nodal)
     const int sj = arr_stag[0];
-    const int sk = (AMREX_SPACEDIM == 1) ? 0 : arr_stag[1];
-    const int sl = (AMREX_SPACEDIM <= 2) ? 0 : arr_stag[2];
+    const int sk = (AMREX_SPACEDIM > 1) ? arr_stag[1] : 1;
+    const int sl = (AMREX_SPACEDIM > 2) ? arr_stag[2] : 1;
 
     // Number of points used for interpolation from coarse grid to fine grid
     const int nj = 2;
-    const int nk = 2;
-    const int nl = 2;
+    const int nk = (AMREX_SPACEDIM > 1) ? 2 : 1;
+    const int nl = (AMREX_SPACEDIM > 2) ? 2 : 1;
 
     const int jc = (sj == 0) ? amrex::coarsen(j - rj/2, rj) : amrex::coarsen(j, rj);
     const int kc = (sk == 0) ? amrex::coarsen(k - rk/2, rk) : amrex::coarsen(k, rk);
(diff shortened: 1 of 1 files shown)


AlexanderSinn:
Another way to write
```
    const int rj = rr[0];
    const int rk = (AMREX_SPACEDIM > 1) ? rr[1] : 1;
    const int rl = (AMREX_SPACEDIM > 2) ? rr[2] : 1;
```
would be
```
    const amrex::IntVectND<3> rr_ex = amrex::IntVectExpand<3>(rr, 1);
    const auto [rj, rk, rl] = rr_ex;
```
or 
```
    const auto [rj, rk, rl] = amrex::IntVectExpand<3>(rr, 1);
```
or 
```
    const auto [rj, rk, rl] = rr.dim3(1);
```
(untested)

Source/Parallelization/WarpXComm_K.H
```diff
@@ -44,18 +44,18 @@ void warpx_interp (int j, int k, int l,
...
-    const int rl = (AMREX_SPACEDIM <= 2) ? 1 : rr[2];
+    const int rk = (AMREX_SPACEDIM > 1) ? rr[1] : 1;
+    const int rl = (AMREX_SPACEDIM > 2) ? rr[2] : 1;
 
     // Staggering (0: cell-centered; 1: nodal)
     const int sj = arr_stag[0];
-    const int sk = (AMREX_SPACEDIM == 1) ? 0 : arr_stag[1];
-    const int sl = (AMREX_SPACEDIM <= 2) ? 0 : arr_stag[2];
+    const int sk = (AMREX_SPACEDIM > 1) ? arr_stag[1] : 1;
+    const int sl = (AMREX_SPACEDIM > 2) ? arr_stag[2] : 1;
```
lucafedeli88, line 54: This seems to change the behavior of the code.
For instance, if  `AMREX_SPACEDIM` is 1, in the previous version `sk` would be 0, while now it would be `1`. Is this intended?

Source/Parallelization/WarpXComm_K.H
```diff
@@ -44,18 +44,18 @@ void warpx_interp (int j, int k, int l,
...
 
     // Number of points used for interpolation from coarse grid to fine grid
     const int nj = 2;
-    const int nk = 2;
-    const int nl = 2;
+    const int nk = (AMREX_SPACEDIM > 1) ? 2 : 1;
+    const int nl = (AMREX_SPACEDIM > 2) ? 2 : 1;
```
lucafedeli88, line 59: This changes the behavior of the code. Is this intended?

Source/Parallelization/WarpXComm_K.H
```diff
@@ -133,38 +133,18 @@ void warpx_interp (int j, int k, int l,
...
-    constexpr int sl_cp = 0;
-#elif defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
-    const     int sk_cp = arr_coarse_stag[1];
-    constexpr int sl_cp = 0;
-#else
-    const int sk_cp = arr_coarse_stag[1];
-    const int sl_cp = arr_coarse_stag[2];
-#endif
+    const int sk_cp = (AMREX_SPACEDIM > 1) ? arr_coarse_stag[1] : 1;
+    const int sl_cp = (AMREX_SPACEDIM > 2) ? arr_coarse_stag[2] : 1;
```
lucafedeli88, line 150: This seems to change the behavior of the code. For instance, if `WARPX_DIM_1D_Z` is defined `sk_cp` would be 1 instead of 0.  Is this intended?

lucafedeli88 (commented):
@dpgrote , nice cleaning! I have just a couple of questions because I did not fully understand a couple of changes.

Source/Parallelization/WarpXComm_K.H
dpgrote: That was intended. If you follow through the original code, for the unused dimensions, the value of `sk` doesn't matter, it produces the same result whether it is 0 or 1. I thought it made it clearer what is happening to set `sk` to 1 since in some sense the data is nodal in that direction.

Source/Parallelization/WarpXComm_K.H
dpgrote: Yes, this is also intended. In the original version is was setting the `nk = 2` which meant that it would have a loop of length 2 over the unused direction. Following the math, the value added to `res` was zero for the second iteration of the loop so it was wasted computation.

lucafedeli88 (approved):
The PR looks good to me!

PR #5248, opened 2024-09-10
[Hackathon] Clean dimension macros in particle routines

dpgrote:
This makes various changes reducing the usage of dimension macros.


lucafedeli88 (approved):
Thanks for this PR, @dpgrote . It looks good to me.

PR #5252, opened 2024-09-11
[Hackathon] Use AMREX_ENUM for algorithm selection

WeiqunZhang:
AMREX_ENUM is strongly typed enum (i.e., C++ scoped enum). It also supports reflection (e.g., converting from enumerator to string and vice versa). This provides more safety checks at compile time and simplifies the algorithm selection code by avoiding duplication of the mapping between enumerator and string.


ax3l (approved):
Nice!! Thank you!

PR #5254, opened 2024-09-12
[Hackathon] Refactor ParserUtils and IntervalsParser

WeiqunZhang:
Refactor ParserUtils and IntervalsParser by using the new Parser capability in amrex::Parser. It's also okay now to use amrex::ParmParse directly because we have added warpx constants to ParmParse for Parser to use (see #5751).


Source/Utils/Parser/ParserUtils.H
```diff
@@ -221,31 +137,21 @@ namespace utils::parser
...
-                    val[i] = safeCastToInt(
-                        static_cast<amrex::Real>(std::round(parser.compileHost<0>()())), str);
-                }
-                else {
-                    val[i] = static_cast<T>(parser.compileHost<0>()());
-                }
+            val.erase(val.begin(), val.begin()+start_ix);
+            if (num_val != amrex::ParmParse::LAST) {
+                if (num_val > int(val.size())) { return 0; }
+                val.resize(num_val);
```
EZoni, line 146: Could you give a couple of examples of which use cases this logic would cover? For example, what is a user input such that `num_val != amrex::ParmParse::LAST`?

EZoni (commented):
@WeiqunZhang 

I saw that my review had been requested on this PR, so I went through the code and left just one question.

More generally, my understanding is that this PR is ready for review and merge (once approved) on your end, is that correct?

I see @dpgrote is a reviewer and assignee. Was there someone else who requested this feature in particular and who should be involved in the review process?

Source/Utils/Parser/ParserUtils.H
WeiqunZhang: Yes, it's ready for review and merge.

The spec of this function says
```
* \param[in] num_val number of input values to use (optional with arrays, default is
* amrex::ParmParse::LAST for reading until the last input value)
 ```
So if `num_val != LAST`, we should read exactly `num_val` values. But the `queryArrWithParser` function always reads all values. That's why we need to call `erase` first and then resize if needed.

**Review by EZoni — APPROVED** [review]

**Comment by EZoni** [comment]
The last CI checks here ran two months ago. Should we merge `development` (even if there are no apparent conflicts) and rerun CI just in case, before we merge?

**Comment by EZoni** [comment]
Also, the PR looks good to me. Does anybody else have additional comments, e.g., @dpgrote (assignee) or others?

**Comment by dpgrote** [comment]
I remember that there was an issue with how integer values are handled. In the WarpX parsing, when expressions are evaluated, it is done with floating point and then converted to integer. If I remember, the amrex parser evaluates expressions as integers, which could give differing results. Is that correct @WeiqunZhang?

**Comment by WeiqunZhang** [comment]
@dpgrote There is a function called queryAsDouble in ParmParse. It evaluates expressions as doubles and then converts to integers. We use that function to parse integer values.

File: Source/Initialization/WarpXAMReXInit.cpp [hunk 2]
```diff
@@ -111,6 +137,7 @@ namespace {
         apply_workaround_for_warpx_numprocs();
         set_device_synchronization();
         override_default_tiling_option_for_particles();
+        add_constants();
```
dpgrote: Maybe call this first before the other routines. It is unlikely, but those read in some parameters that could possibly depend on the constants.

EZoni:
@dpgrote 

It seems like Weiqun addressed all your comments. I will request a new review from you so you can approve the PR, if it looks good to you.

dpgrote (approved):
Thanks! Looks good to me.

PR #5257, opened 2024-09-12
CTest: more docs on `-R` regular expression filtering

EZoni:
Add docs on common use case for `-R` regular expression filtering with CTest.


Docs/source/developers/testing.rst
```diff
@@ -73,13 +73,25 @@ For easier debugging, it can be convenient to run the tests on your local machin
 
   .. code-block:: sh
 
-       ctest --test-dir build -R laser_acceleration
+       ctest --test-dir build -R "laser_acceleration"
```
dpgrote: I don't think the quotes are needed here. Same for the `-E` option.

ax3l: Correct, only needed when `"actual.*regular\.expressions"` are used.

ax3l:
```suggestion
       ctest --test-dir build -R laser_acceleration
```

Docs/source/developers/testing.rst
```diff
@@ -73,13 +73,25 @@ For easier debugging, it can be convenient to run the tests on your local machin
...
 
 * Run tests except those matching the regular expression ``laser_acceleration``:
 
   .. code-block:: sh
 
-       ctest --test-dir build -E laser_acceleration
+       ctest --test-dir build -E "laser_acceleration"
```
ax3l:
```suggestion
       ctest --test-dir build -E laser_acceleration
```

Docs/source/developers/testing.rst
```diff
@@ -73,13 +73,25 @@ For easier debugging, it can be convenient to run the tests on your local machin
...
+
+* Sometimes two or more tests share a large number of input parameters and differ by a small set of options.
+  Such tests typically also share a base string in their names.
+  For example, you can find three different tests named ``test_3d_langmuir_multi``, ``test_3d_langmuir_multi_nodal`` and ``test_3d_langmuir_multi_picmi``.
+  In such a case, if you wish to run the test ``test_3d_langmuir_multi`` only, this can be done again with the ``-R`` regular expression filter via
```
ax3l:
```suggestion
  In such a case, if you wish to run the test ``test_3d_langmuir_multi`` only, this can be done again with the ``-R`` regular `expression filter <https://regex101.com>`__ via
```

Docs/source/developers/testing.rst
```diff
@@ -73,13 +73,25 @@ For easier debugging, it can be convenient to run the tests on your local machin
...
+
+  .. code-block:: sh
+
+       ctest --test-dir build -R "test_3d_langmuir_multi\..*"
+
+  Note that filtering with ``-R "test_3d_langmuir_multi"`` would include the other existing tests and would not be sufficient.
```
ax3l:
```suggestion
  Note that filtering with ``-R "test_3d_langmuir_multi"`` would include the additional tests that have the same string in them and can be sufficient if one aims to isolate a single test.
```

Docs/source/developers/testing.rst
EZoni: Yes, I know it's not needed here, but I thought it was a good idea to add it here too for consistency, to sort of signal that one can use regular expressions here. I thought it could be okay to keep it in our examples in the docs.

ax3l (approved):

ax3l (approved):

Docs/source/developers/testing.rst
```diff
@@ -81,6 +81,18 @@ For easier debugging, it can be convenient to run the tests on your local machin
...
+* Sometimes two or more tests share a large number of input parameters and differ by a small set of options.
+  Such tests typically also share a base string in their names.
+  For example, you can find three different tests named ``test_3d_langmuir_multi``, ``test_3d_langmuir_multi_nodal`` and ``test_3d_langmuir_multi_picmi``.
+  In such a case, if you wish to run the test ``test_3d_langmuir_multi`` only, this can be done again with the ``-R`` regular `expression filter <https://regex101.com>`__ via
+
+  .. code-block:: sh
+
+       ctest --test-dir build -R "test_3d_langmuir_multi\..*"
+
+  Note that filtering with ``-R "test_3d_langmuir_multi"`` would include the additional tests that have the same string in them and can be sufficient if one aims to isolate a single test.
```
EZoni: Could we maybe rephrase like this?
```suggestion
  Note that filtering with ``-R "test_3d_langmuir_multi"`` would include the additional tests that have the same string in the name and can be sufficient if one aims to isolate a single test.
```

ax3l: How about `their` name?

ax3l: and `substring`

ax3l:
```suggestion
  Note that filtering with ``-R "test_3d_langmuir_multi"`` would include the additional tests that have the same substring in their name and can be sufficient if one aims to isolate a single test.
```

EZoni: Wait, this should be "would not be sufficient to isolate a single test", no?

Docs/source/developers/testing.rst
```diff
@@ -81,6 +81,18 @@ For easier debugging, it can be convenient to run the tests on your local machin
...
+* Sometimes two or more tests share a large number of input parameters and differ by a small set of options.
+  Such tests typically also share a base string in their names.
+  For example, you can find three different tests named ``test_3d_langmuir_multi``, ``test_3d_langmuir_multi_nodal`` and ``test_3d_langmuir_multi_picmi``.
+  In such a case, if you wish to run the test ``test_3d_langmuir_multi`` only, this can be done again with the ``-R`` regular `expression filter <https://regex101.com>`__ via
+
+  .. code-block:: sh
+
+       ctest --test-dir build -R "test_3d_langmuir_multi\..*"
+
+  Note that filtering with ``-R "test_3d_langmuir_multi"`` would include the additional tests that have the same substring in their name and can be sufficient if one aims to isolate a single test.
```
EZoni: That's what I meant:
```suggestion
  Note that filtering with ``-R "test_3d_langmuir_multi"`` would include the additional tests that have the same substring in their name and would not be sufficient to isolate a single test.
```

ax3l: Ok, but I used could because this is just an example - not all example names are 5x extended :)

EZoni: Ah I see. Here I was using the negative form because I was referring to the specific example in the text, of the three Langmuir tests with shared substring. So I meant to connect this sentence with that example, since it's in the same list item. That's why I got confused.

PR #5155, opened 2024-08-20
Catalyst2 support

ChristosT:
This PR completes the work of https://github.com/ECP-WarpX/WarpX/pull/4121 by updating the cmake files where appropriate and adding a CI job specifically for the catalyst integration as well as updating existing tests.

This supersedes and closes https://github.com/ECP-WarpX/WarpX/pull/4121.

cc: @c-wetterer-nelson 

Examples form the generated images (thanks to @andrewcombs !):

https://github.com/user-attachments/assets/6c073c74-7ce1-43cc-b7e8-7438c5163dbb

https://github.com/user-attachments/assets/94c6ec13-8c8b-4acf-a5b2-e8b16cb033dc

PR #5264, opened 2024-09-13
docs: provove build and usage instructions of catalyst

ChristosT:
This takes care of the comments in https://github.com/ECP-WarpX/WarpX/pull/5155


ax3l (approved):
Thank you, that looks great!

PR #5266, opened 2024-09-13
[Hackathon] Over-the-Barrier ionization equation fix

johvandewetering:
Added the minus sign in the argument of the exponent in Eq. 8 from Zhang et al., PRA 90, 043410 (2014).


lucafedeli88 (approved):
Thanks for having spotted this bug!

johvandewetering:
The paper is actually wrong, previous code without minus sign is correct. I have edited the ionization documentation in a separate PR to reflect this.

RemiLehe:
Thanks for opening a follow-up PR.

PR #5268, opened 2024-09-13
Fix C++20 issue

lucafedeli88:
This PR fixes the following warning, observed with `g++ 14`:

```
warning: template-id not allowed for constructor in C++20
```

See for instance https://stackoverflow.com/questions/71978335/class-templates-constructor-declaration-doesnt-compile-for-c20-but-compiles


ax3l (approved):
Yay!!

PR #5273, opened 2024-09-15
Use only plasma current in `HybridPICSolveE` 

roelof-groenewald:
This PR fixes a bug in the Ohm's law solver. Currently the total current ($J = J_e + J_i + J_{ext}$) is used in the resistivity term in Ohm's law but actually only the plasma current ($J_{plasma} = J_e + J_i = \nabla\times B/\mu_0 - J_{ext}$) should be used. 
Correspondingly the external current `multifabs` are changed to have staggering that matches the `current_fp` `multifabs`.


marcoacc95 (commented):
Thanks for fixing this Roelof, I have a question regarding the J_ext

Does J_ext still need to be a current through the plasma? or can it be used as a "coil" current too?

roelof-groenewald:
> Does J_ext still need to be a current through the plasma? or can it be used as a "coil" current too?

With the current implementation of Ohm's law - assuming massless electrons - it does have to a current through the plasma. However, once we can extend Ohm's law to include electron inertia one would in principle be able to use `J_ext` to specify coil currents. However, I would like to move towards being generally able to separately handle the magnetic field applied from external coils versus the field generated by the plasma. I'm still trying to work out how exactly to do that in a consistent manner though. Maybe you guys have some thoughts on that?

marcoacc95:
For now we are trying to do that by splitting the fields between the self-generated field and the external one for the particle pusher, it is work in progress but maybe @clarkse can comment on that?

clarkse:
@roelof-groenewald I have a WIP PR for adding external fields here through the analytical particle field interface. I do plan to add time dependent solenoids into the accelerator lattice that could be used to specify coil geometries.

https://github.com/ECP-WarpX/WarpX/pull/5275

Source/Fields.H
```diff
@@ -40,11 +40,11 @@ namespace warpx::fields
...
-        hybrid_electron_pressure_fp,
-        hybrid_rho_fp_temp,
-        hybrid_current_fp_temp,
-        hybrid_current_fp_ampere,
-        hybrid_current_fp_external,
+        hybrid_electron_pressure_fp, //!< Used with Ohm's law solver. Stores the electron pressure
+        hybrid_rho_fp_temp,          //!< Used with Ohm's law solver. Stores the time interpolated/extrapolated charge density
+        hybrid_current_fp_temp,      //!< Used with Ohm's law solver. Stores the time interpolated/extrapolated current density
+        hybrid_current_fp_plasma,    //!< Used with Ohm's law solver. Stores plasma current calculated as J_plasma = curl x B / mu0 - J_ext
+        hybrid_current_fp_external,  //!< Used with Ohm's law solver. Stores external current
```
roelof-groenewald: @ax3l I went ahead and added the docstrings in this PR as you requested.

ax3l: Thank you, Roelof, appreciate it! :)

Source/Fields.H
```diff
@@ -45,11 +45,11 @@ namespace warpx::fields
...
```
ax3l: Need to use the inline format for the comment, because of the macro used:
```suggestion
        hybrid_electron_pressure_fp, /**< Used with Ohm's law solver. Stores the electron pressure */
        hybrid_rho_fp_temp,          /**< Used with Ohm's law solver. Stores the time interpolated/extrapolated charge density */
        hybrid_current_fp_temp,      /**< Used with Ohm's law solver. Stores the time interpolated/extrapolated current density */
        hybrid_current_fp_plasma,    /**< Used with Ohm's law solver. Stores plasma current calculated as J_plasma = curl x B / mu0 - J_ext */
        hybrid_current_fp_external,  /**< Used with Ohm's law solver. Stores external current */
```

ax3l (approved):
Thank you for the PR and thanks for the reviews!

PR #5274, opened 2024-09-15
Minor refactoring of refined injection for AddPlasmaFlux

RemiLehe:
In `AddPlasmaFlux`, the code associated with the refined injection can be simplified a bit in a few places.

### DIFF
--- a/Source/Particles/PhysicalParticleContainer.cpp
+++ b/Source/Particles/PhysicalParticleContainer.cpp
@@ -1456,7 +1456,7 @@ PhysicalParticleContainer::AddPlasmaFlux (PlasmaInjector const& plasma_injector,
             auto hi = getCellCoords(overlap_corner, dx, {1._rt, 1._rt, 1._rt}, iv);
 
             // Skip cells that do not overlap with the injection plane
-            if (!flux_pos->overlapsWith(lo, hi)) return;
+            if (!flux_pos->overlapsWith(lo, hi)) { return; }
 
             // Determine how many particles to inject in this cell
             auto index = overlap_box.index(iv);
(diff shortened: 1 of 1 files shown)


Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1533,24 +1533,18 @@ PhysicalParticleContainer::AddPlasmaFlux (PlasmaInjector const& plasma_injector,
...
             const IntVect iv = IntVect(AMREX_D_DECL(i, j, k));
+            amrex::ignore_unused(j,k);
             const auto index = overlap_box.index(iv);
 
             Real scale_fac = compute_scale_fac_area(dx, num_ppc_real, flux_normal_axis);
 
             auto lo = getCellCoords(overlap_corner, dx, {0._rt, 0._rt, 0._rt}, iv);
             auto hi = getCellCoords(overlap_corner, dx, {1._rt, 1._rt, 1._rt}, iv);
 
-            if (flux_pos->overlapsWith(lo, hi))
```
RemiLehe, line 1542: This `if` condition is not key here, since the `pcounts[index]` is `0` anyway for cells that do not overlap.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1450,22 +1450,21 @@ PhysicalParticleContainer::AddPlasmaFlux (PlasmaInjector const& plasma_injector,
...
             auto lo = getCellCoords(overlap_corner, dx, {0._rt, 0._rt, 0._rt}, iv);
             auto hi = getCellCoords(overlap_corner, dx, {1._rt, 1._rt, 1._rt}, iv);
 
             if (flux_pos->overlapsWith(lo, hi))
             {
                 auto index = overlap_box.index(iv);
-                int r;
+                int r = 1;
                 if (fine_overlap_box.ok() && fine_overlap_box.contains(iv)) {
                     r = compute_area_weights(lrrfac, flux_normal_axis);
```
dpgrote: There is another small clean up that can be done. The `lrrfac` is not necessary since `rrfac` is already locally declared within the routine. Similarly for `AddPlasma`.

atmyers (approved):

Source/Particles/PhysicalParticleContainer.cpp
RemiLehe: Sounds good, thanks. I added this in my latest commit.

PR #5275, opened 2024-09-16
Add external particle fields ohms law hybrid

clarkse:
This PR allows for the addition of external fields through the particle fields analytical interface. This is useful for field splitting external vs. self fields in the hybrid ohm's law solver.


Python/pywarpx/picmi.py
```diff
@@ -1782,6 +1782,9 @@ class HybridPICSolver(picmistandard.base._ClassWithInit):
 
     Jx/y/z_external_function: str
         Function of space and time specifying external (non-plasma) currents.
+
+    Ex/y/z_external_function: str
```
```suggestion
    Ax/y/z_external_function: str
```

Python/pywarpx/picmi.py
```diff
@@ -1782,6 +1782,9 @@ class HybridPICSolver(picmistandard.base._ClassWithInit):
...
+        Function of space and time specifying external (non-plasma) E-fields.
```
```suggestion
        Function of space and time specifying external (non-plasma) vector potential fields.
```

Source/FieldSolver/FiniteDifferenceSolver/ComputeCurlA.cpp
```diff
@@ -0,0 +1,289 @@
...
+void FiniteDifferenceSolver::ComputeCurlACylindrical (
+    ablastr::fields::VectorField& Bfield,
+    ablastr::fields::VectorField const& Afield,
+    int lev
+)
+{
+    // for the profiler
+    amrex::LayoutData<amrex::Real>* cost = WarpX::getCosts(lev);
+
+    // reset Jfield
```
roelof-groenewald:
```suggestion
    // reset Bfield
```

Source/FieldSolver/FiniteDifferenceSolver/ComputeCurlA.cpp
```diff
@@ -0,0 +1,289 @@
...
+                        Br(i, j, 0, 2*m  ) = - (
+                            T_Algo::UpwardDz(At, coefs_z, n_coefs_z, i, j, 0, 2*m  )
+                            + m * Az(i, j, 0, 2*m-1)/r ); // Imaginary part
+                    }
+                } else { // r==0: On-axis corrections
+                    // Ensure that Br remains 0 on axis (except for m=1)
+                    Br(i, j, 0, 0) = 0.; // Mode m=0
+                    for (int m=1; m<nmodes; m++) { // Higher-order modes
+                        if (m == 1){
+                            // For m==1, Ez is linear in r, for small r
```
roelof-groenewald:
```suggestion
                            // For m==1, Bz is linear in r, for small r
```

Source/FieldSolver/FiniteDifferenceSolver/ComputeCurlA.cpp
```diff
@@ -0,0 +1,289 @@
...
+
+                By(i, j, k) = (
+                    - T_Algo::UpwardDx(Az, coefs_x, n_coefs_x, i, j, k)
+                    + T_Algo::UpwardDz(Ax, coefs_z, n_coefs_z, i, j, k)
+                );
+            },
+
+            // Bz calculation
+            [=] AMREX_GPU_DEVICE (int i, int j, int k){
+                if (cov_ptr.isCovered(2, EB::CoverTopology::face, i, j, k)) { return; }
```
roelof-groenewald: Just to keep things consistent
```suggestion
                // Skip if this cell is fully covered by embedded boundaries
                if (cov_ptr.isCovered(2, EB::CoverTopology::face, i, j, k)) { return; }
```

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -679,10 +676,14 @@ void FiniteDifferenceSolver::HybridPICSolveECylindrical (
                 // Add resistivity only if E field value is used to update B
                 if (solve_for_Faraday) { Ez(i, j, 0) += eta(rho_val, jtot_val) * Jz(i, j, 0); }
 
-                if (include_hyper_resistivity_term) {
+                if (include_hyper_resistivity_term && solve_for_Faraday) {
```
roelof-groenewald: This is not needed since we set:
```
const bool include_hyper_resistivity_term = (eta_h > 0.0) && solve_for_Faraday;
```
```suggestion
                if (include_hyper_resistivity_term) {
```

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -842,13 +861,15 @@ void FiniteDifferenceSolver::HybridPICSolveECartesian (
...
-        amrex::Array4<amrex::Real> lx, ly, lz;
-        if (EB::enabled()) {
-            lx = edge_lengths[0]->array(mfi);
-            ly = edge_lengths[1]->array(mfi);
-            lz = edge_lengths[2]->array(mfi);
+        Array4<Real> Ex_ext, Ey_ext, Ez_ext;
+        if (include_external_fields) {
+        Ex_ext = Efield_external[0]->array(mfi);
+        Ey_ext = Efield_external[1]->array(mfi);
+        Ez_ext = Efield_external[2]->array(mfi);
```
roelof-groenewald:
```suggestion
            Ex_ext = Efield_external[0]->array(mfi);
            Ey_ext = Efield_external[1]->array(mfi);
            Ez_ext = Efield_external[2]->array(mfi);
```

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -368,21 +346,23 @@ void FiniteDifferenceSolver::HybridPICSolveE (
...
+            lev, hybrid_model, solve_for_Faraday
         );
 
 #endif
     } else {
         amrex::Abort(Utils::TextMsg::Err(
             "HybridSolveE: The hybrid-PIC electromagnetic solver algorithm must be used"));
     }
+    auto& warpx = WarpX::GetInstance();
+    warpx.ApplyEfieldBoundary(lev, PatchType::fine);
```
roelof-groenewald: Why is it necessary to move this here?

Source/WarpX.cpp
```diff
@@ -738,8 +738,8 @@ WarpX::ReadParameters ()
         {
             // Filter currently not working with FDTD solver in RZ geometry along R
             // (see https://github.com/ECP-WarpX/WarpX/issues/1943)
-            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(!use_filter || filter_npass_each_dir[0] == 0,
-                "In RZ geometry with FDTD, filtering can only be apply along z. This can be controlled by setting warpx.filter_npass_each_dir");
+            //WARPX_ALWAYS_ASSERT_WITH_MESSAGE(!use_filter || filter_npass_each_dir[0] == 0,
+            //    "In RZ geometry with FDTD, filtering can only be apply along z. This can be controlled by setting warpx.filter_npass_each_dir");
```
roelof-groenewald: Is this change intentional?

roelof-groenewald (commented):
Thanks for this PR!

I like the approach in that it mostly keeps the augmented functionality separate from the Ohm's law solving.
One thing though, this really includes 2 PRs in one
1. Introduction of the EB "Covered" functionality
2. The external fields in the hybrid algorithm

Do you think it would be feasible to pull the "covered" stuff into a separate PR? That would help a lot in reviewing and merging.

Source/WarpX.cpp
clarkse: This inclusion is not meant to be part of this PR. I will remove it.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
clarkse: I was looking around to fix some boundary issues. I will double check on this.

ax3l:
Oh, this needs a merge with `development` and conflict resolution.

Docs/source/refs.bib
```diff
@@ -507,3 +507,13 @@ @article{Rhee1987
...
 }
+
+@misc{holmstrom2013handlingvacuumregionshybrid,
+      title={Handling vacuum regions in a hybrid plasma solver},
+      author={M. Holmstrom},
+      year={2013},
+      eprint={1301.0272},
+      archivePrefix={arXiv},
+      primaryClass={physics.space-ph},
+      url={https://arxiv.org/abs/1301.0272},
```
roelof-groenewald, line 518: I just quickly checked and couldn't find a journal reference for this article, was it never actually peer-reviewed?

Docs/source/usage/parameters.rst
```diff
@@ -2535,6 +2535,27 @@ Maxwell solver: kinetic-fluid hybrid
...
+    If ``algo.maxwell_solver`` is set to ``hybrid``, this sets the vacuum region handling of the generalized Ohm's Law to suppress vacuum fluctuations. :cite:t:`param-holmstrom2013handlingvacuumregionshybrid`.
+
+* ``hybid_pic_model.add_external_fields`` (`bool`) optional (default ``false``)
+    If ``algo.maxwell_solver`` is set to ``hybrid``, this sets the hybrid solver to use split external fields defined in external_vector_potential inputs.
+
+* ``external_vector_potential.fields`` (list of `str`) optional (default ``empty``)
+    If ``hybid_pic_model.add_external_fields`` is set to ``true``, this adds a list names for external time varying vector potentials to be added to hybrid solver.
+
+* ``external_vector_potential.<field name>.read_from_file`` (`bool`) optional (default ``false``)
+    If ``hybid_pic_model.add_external_fields`` is set to ``true``, this flag determines whether to load an external field or use an implcit function to evaluate teh time varying field.
```
```suggestion
    If ``hybid_pic_model.add_external_fields`` is set to ``true``, this flag determines whether to load an external field or use an implcit function to evaluate the time varying field.
```

Examples/Tests/ohm_solver_cylinder_compression/inputs_test_3d_ohm_solver_cylinder_compression_picmi.py
```diff
@@ -0,0 +1,393 @@
...
+
+    B0 = np.sqrt(2 * constants.mu0 * p0)  # Initial magnetic field strength (T)
+
+    # Do a 2x uniform B-field compression
+    dB = B0
+
+    # Flux Conserver radius
+    R_c = 0.5
+
+    # Plasma Radius (These values match GS solution in gs_psi.csv)
```
roelof-groenewald: Is `gs_psi.csv` used in the test / available to the public?

Examples/Tests/ohm_solver_cylinder_compression/inputs_test_3d_ohm_solver_cylinder_compression_picmi.py
```diff
@@ -0,0 +1,393 @@
...
+            zvec = np.linspace(-self.LZ, self.LZ, num=2 * self.NZ)
+            XM, YM, ZM = np.meshgrid(xvec, yvec, zvec, indexing="ij")
+
+            RM = np.sqrt(XM**2 + YM**2)
+
+            Ax_data = -0.5 * YM * self.dB
+            Ay_data = 0.5 * XM * self.dB
+            Az_data = np.zeros_like(RM)
+
+            # Write vector potential to file to exercise field loading via OPenPMD
```
roelof-groenewald:
```suggestion
            # Write vector potential to file to exercise field loading via OpenPMD
```

Python/pywarpx/fields.py
```diff
@@ -578,6 +578,24 @@ def norm0(self, *args):
...
+        mf_name=mf_name, idir=1, level=level, include_ghosts=include_ghosts
+    )
+
+
+def CustomNamedzWrapper(mf_name, level=0, include_ghosts=False):
+    return _MultiFABWrapper(
+        mf_name=mf_name, idir=2, level=level, include_ghosts=include_ghosts
+    )
+
+
```
roelof-groenewald, line 598: I wonder if these are really necessary. I see the other wrappers here as useful since the user might not know exactly what the `multifab` name is on the C++ side (for example E-field -> `Efield_aux`). But in this case the user would have to know the name, so they might as well call the `_MultiFABWrapper` directly.

Python/pywarpx/picmi.py
```diff
@@ -1853,8 +1853,37 @@ class HybridPICSolver(picmistandard.base._ClassWithInit):
     substeps: int, default=100
         Number of substeps to take when updating the B-field.
 
+    holmstrom_vacuum_region: bool, default=False
+        Flag to determine handling of vacuum region. Setting to True will solve the simplified Generalized Ohm's Law dropping the Hall and pressure terms.
```
```suggestion
        Flag to determine handling of vacuum region. Setting to True will solve the simplified Generalized Ohm's Law dropping the Hall and pressure terms in the vacuum region.
```

roelof-groenewald: It might also be useful here to point out that using Holstrom reduces Ohm's law to a diffusion equation for the B-field.

Python/pywarpx/picmi.py
```diff
@@ -1853,8 +1853,37 @@ class HybridPICSolver(picmistandard.base._ClassWithInit):
...
+    A_external: dict
+        Function of space and time specifying external (non-plasma) vector potential fields.
+        It is expected that a nested dicitonary will be passed
+        into picmi for each field that has different timings
+        e.g.
+        A_external = {
+            '<field_name1>': {
+                'Ax_external_function': <implicit function with (x,y,z) dependence>,
+                'Ax_external_function': <implicit function with (x,y,z) dependence>,
+                'Ax_external_function': <implicit function with (x,y,z) dependence>,
```
roelof-groenewald: Should these be?
```suggestion
                'Ay_external_function': <implicit function with (x,y,z) dependence>,
                'Az_external_function': <implicit function with (x,y,z) dependence>,
```

Python/pywarpx/picmi.py
```diff
@@ -1936,6 +1976,50 @@ def solver_initialize_inputs(self):
                 self.Jz_external_function, self.mangle_dict
             ),
         )
+        pywarpx.hybridpicmodel.add_external_fields = self.add_external_fields
+        if self.add_external_fields:
```
roelof-groenewald: It might be a little cleaner to remove `self.add_external_fields` in the constructor and here do:
```suggestion
        if self.A_external is not None:
                pywarpx.hybridpicmodel.add_external_fields = True
```
It all should work fine then as long as you use `query` and not `get` for the runtime parameter `add_external_fields` with the parser.

Python/pywarpx/picmi.py
```diff
@@ -1936,6 +1976,50 @@ def solver_initialize_inputs(self):
...
+                "fields",
+                pywarpx.my_constants.mangle_expression(
+                    list(self.A_external.keys()), self.mangle_dict
+                ),
+            )
+            for field_name, field_dict in self.A_external.items():
+                if (
+                    "read_from_file" in field_dict.keys()
+                    and field_dict["read_from_file"]
+                ):
```
```suggestion
                if field_dict.get("read_from_file", False):
```

Source/FieldSolver/FiniteDifferenceSolver/ComputeCurlA.cpp
```diff
@@ -0,0 +1,308 @@
...
+#ifdef WARPX_DIM_RZ
+#   include "FiniteDifferenceAlgorithms/CylindricalYeeAlgorithm.H"
+#else
+#   include "FiniteDifferenceAlgorithms/CartesianYeeAlgorithm.H"
+#endif
+
+#include "Utils/TextMsg.H"
+#include "WarpX.H"
+
+#include <ablastr/coarsen/sample.H>
```
roelof-groenewald: I don't think this is used anywhere is it?
```suggestion
```

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/ExternalVectorPotential.cpp
```diff
@@ -0,0 +1,376 @@
...
+                    m_A_external[i][0],
+                    m_A_external[i][1],
+                    m_A_external[i][2],
+                    lev, PatchType::fine,
+                    warpx.GetEBUpdateEFlag(),
+                    false);
+
+                for (int idir = 0; idir < 3; ++idir) {
+                    warpx.m_fields.get(Aext_field, Direction{idir}, lev)->
+                        FillBoundary(warpx.Geom(lev).periodicity());
```
roelof-groenewald, line 206: I just very quickly checked so I might have missed it, but does `ReadExternalFieldFromFile` call `FillBoundary` internally? Is that why this is only done for the parser fields?

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/ExternalVectorPotential.cpp
```diff
@@ -0,0 +1,376 @@
...
+    }
+}
+
+void
+ExternalVectorPotential::CalculateExternalCurlA (std::string& coil_name)
+{
+    using ablastr::fields::Direction;
+    auto & warpx = WarpX::GetInstance();
+
+    // Compute the curl of at at max and store
```
roelof-groenewald: Should this be?
```suggestion
    // Compute the curl of A at t_max and store
```

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/ExternalVectorPotential.cpp
```diff
@@ -0,0 +1,376 @@
...
+        for (int idir = 0; idir < 3; ++idir) {
+            warpx.m_fields.get(curlAext_field, Direction{idir}, lev)->
+                FillBoundary(warpx.Geom(lev).periodicity());
+        }
+    }
+}
+
+AMREX_FORCE_INLINE
+void
+ExternalVectorPotential::ZeroFieldinEB (
```
roelof-groenewald: I think there is likely an AMReX function that achieves this. We should ask on Slack.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/ExternalVectorPotential.cpp
```diff
@@ -0,0 +1,376 @@
...
+                    0, 1, 0);
+            }
+
+            for (int idir = 0; idir < 3; ++idir) {
+                E_ext[lev][Direction{idir}]->FillBoundary(warpx.Geom(lev).periodicity());
+                B_ext[lev][Direction{idir}]->FillBoundary(warpx.Geom(lev).periodicity());
+            }
+
+            if (EB::enabled()) {
+                ZeroFieldinEB(B_ext[lev], warpx.GetEBUpdateBFlag()[lev]);
```
roelof-groenewald: This should not be necessary since `ComputeCurlA` does not update inside the EB.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/ExternalVectorPotential.cpp
```diff
@@ -0,0 +1,376 @@
...
+    ablastr::fields::MultiLevelVectorField E_ext =
+        warpx.m_fields.get_mr_levels_alldirs(FieldType::hybrid_E_fp_external, warpx.finestLevel());
+
+    // Zero E and B external fields
+    for (int lev = 0; lev <= warpx.finestLevel(); ++lev) {
+        for (int idir = 0; idir < 3; ++idir) {
+            B_ext[lev][Direction{idir}]->setVal(0.0_rt);
+            E_ext[lev][Direction{idir}]->setVal(0.0_rt);
+        }
+    }
```
roelof-groenewald: You can save these cycles by setting the coefficient for each of these fields to 0 in the `amrex::MultiFab::LinComb` further down. It would be good to just make comment at that location of why it is done that way.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -114,6 +135,16 @@ void HybridPICModel::AllocateLevelMFs (ablastr::fields::MultiFabRegister & field
         lev, amrex::convert(ba, jz_nodal_flag),
         dm, ncomps, IntVect(1), 0.0_rt);
 
+    if (m_add_external_fields) {
+        m_external_vector_potential->AllocateLevelMFs(
+            fields,
+            lev, ba, dm,
+            ncomps, ngEB,
+            Ex_nodal_flag, Ey_nodal_flag, Ez_nodal_flag,
+            Bx_nodal_flag, By_nodal_flag, Bz_nodal_flag
```
roelof-groenewald, line 144: I think the idea is to move away from passing these arguments around and I believe there is a method now that will create a new multifab using an existing one as template, such that in `AllocateLevelMFs` you would grab E from the field register, and then just make the new multifab you want off it. I'll search for that function tomorrow or we can ask on Slack.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -552,13 +588,29 @@ void HybridPICModel::FieldPush (
...
     // Calculate the E-field from Ohm's law
     HybridPICSolveE(Efield, Jfield, Bfield, rhofield, eb_update_E, true);
     warpx.FillBoundaryE(ng, nodal_sync);
+    warpx.ApplyEfieldBoundary(0, PatchType::fine, t_old);
+
     // Push forward the B-field using Faraday's law
-    amrex::Real const t_old = warpx.gett_old(0);
     warpx.EvolveB(dt, dt_type, t_old);
     warpx.FillBoundaryB(ng, nodal_sync);
+    warpx.ApplyBfieldBoundary(0, PatchType::fine, dt_type, t_old);
```
roelof-groenewald: These changes should not be needed since the E-field boundary is applied at
https://github.com/ECP-WarpX/WarpX/blob/2cc4fd2c3ad5be96e1aa5811d72a8e1018d925c4/Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp#L353

And the B-field boundary at 
https://github.com/ECP-WarpX/WarpX/blob/2cc4fd2c3ad5be96e1aa5811d72a8e1018d925c4/Source/FieldSolver/WarpXPushFieldsEM.cpp#L932

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -552,13 +588,29 @@ void HybridPICModel::FieldPush (
...
+}
+
+void
+WarpX::CalculateExternalCurlA() {
```
roelof-groenewald: I still need to review the rest to understand why this has to be a member function of `WarpX`, but in any case it would probably be good to move it to a different file, perhaps `WarpXPushFieldsHybridPIC.cpp`.

roelof-groenewald (commented):
I've made some progress but have not fully reviewed all the changes yet. Here are some comments I've recorded so far.

Docs/source/refs.bib
clarkse: I could not find a peer reviewed version of it.

Examples/Tests/ohm_solver_cylinder_compression/inputs_test_3d_ohm_solver_cylinder_compression_picmi.py
clarkse: I ended up using an analytical form of the GS solution instead of reading it in. The solution is in the Bz function. I will update the comment.

Python/pywarpx/fields.py
clarkse: I am using MF names that are unregistered at compilation to hold all of the external A arrays, so this is useful to query those or just call via the string name post registration and allocation. I agree we can drop these and just grab the _MultiFABWrapper, but usually the convention is that anything with a leading _ should be an internal or private call only. I don't see a problem with having a helper function to call it.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/ExternalVectorPotential.cpp
clarkse: I had to add this to make sure there is no integration errors at the EB since I load in the external field on all grid points. I can try without it again now that I switched how the vacuum region is handled in the B-field integration.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
clarkse: I will double check this again. I was having some issues with the boundaries, so I will remove and test.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
clarkse: This is a hook that I added to expose CalculateExternalCurlA as a static function to be called in a python callback. It is used to load fields dynamically in the MFs then this is called at the end of the callback to populate the external E and B arrays from A. It is more of an advanced use case and I used it for initial debugging, but could probably be removed if you think it is too messy to keep it around. I will likely use it again when integrating lattice elements into this framework.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/ExternalVectorPotential.cpp
clarkse: So this is a bug that was reported in https://github.com/ECP-WarpX/WarpX/issues/5588 where the external field from file will populate ghost cells via interpolation. So all cells are filled with valid values.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/ExternalVectorPotential.cpp
clarkse: I will update this comment to be more verbose. It should read something like computes the curl of A that is not-scaled by the time dependent function.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/ExternalVectorPotential.cpp
clarkse: Perhaps. Let me see with the latest version of this if I need to run this anymore. I think it was more of a band-aid solution to just make things work.

clarkse:
@roelof-groenewald thanks for the partial review. I will start making some of the non-trivial changes and rechecking the tests. Very thorough review as usual. This will make for a better PR.

Python/pywarpx/fields.py
roelof-groenewald: Sounds good. I agree there is no downside to having it.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
roelof-groenewald: I think it is fine to have, I just think it should live in a file that generally defines `WarpX` member functions, so I think moving it to `WarpXPushFieldsHybridPIC.cpp` would be a fine change.

Source/Fields.H
```diff
@@ -50,19 +50,21 @@ namespace warpx::fields
...
+        F_cp,   //!< Only used with MR. Used for divE cleaning, on the coarse patch of each level
+        G_cp,   //!< Only used with MR. Used for divB cleaning, on the coarse patch of each level
+        Efield_cax, //!< Only used with MR. Particles that are close to the edge of the MR patch (i.e. in the gather buffer) gather from this field
+        Bfield_cax, //!< Only used with MR. Particles that are close to the edge of the MR patch (i.e. in the gather buffer) gather from this field
+        E_external_particle_field, //!< Stores external particle fields provided by the user as  through an openPMD file
+        B_external_particle_field, //!< Stores external particle fields provided by the user as  through an openPMD file
+        distance_to_eb, //!< Only used with embedded boundaries (EB). Stores the distance to the nearest EB
+        edge_lengths,   //!< Only used with the ECT solver. Indicates the length of the cell edge that is covered by the EB, in SI units
+        face_areas,     //!< Only used with the ECT solver. Indicates the area of the cell face that is covered by the EB, in SI units
```
roelof-groenewald: Do you know why the comment notation was changed for these lines? Maybe from a IDE setting?
Looks like it changed `/**< ... */ to `//!<`. I guess for the sake of consistency it would be good to change back. Sorry for that hassle.

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1308,8 +1375,10 @@ void WarpX::CheckKnownIssues()
             mypc->m_B_ext_particle_s != "none" || mypc->m_E_ext_particle_s != "none"
         );
         WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
-            !external_particle_field_used,
-            "The hybrid-PIC algorithm does not work with external fields "
+            (!external_particle_field_used
+            || mypc->m_B_ext_particle_s == "parse_b_ext_particle_function"
+            || mypc->m_E_ext_particle_s == "parse_e_ext_particle_function"),
+            "The hybrid-PIC algorithm only works with analytical external E/B fields "
```
roelof-groenewald: Is this change still necessary or is it left over from earlier versions of this implementation when it used the particle fields?

```suggestion
            !external_particle_field_used,
            "The hybrid-PIC algorithm does not work with external fields "
```

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1048,30 +1048,35 @@ WarpX::InitLevelData (int lev, Real /*time*/)
     }
 }
 
-void WarpX::ComputeExternalFieldOnGridUsingParser (
-    warpx::fields::FieldType field,
+template<typename T>
+void ComputeExternalFieldOnGridUsingParser_template (
```
roelof-groenewald, line 1052: This is a nice improvement!

Source/WarpX.cpp
```diff
@@ -768,7 +768,7 @@ WarpX::ReadParameters ()
             use_kspace_filter = use_filter;
             use_filter = false;
         }
-        else // FDTD
+        else if (WarpX::electromagnetic_solver_id != ElectromagneticSolverAlgo::HybridPIC)
```
roelof-groenewald: Could you please briefly explain why r-filtering in RZ should be allowed for the hybrid-solver but not the Maxwell FDTD solver?

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/ExternalVectorPotential.cpp
```diff
@@ -0,0 +1,376 @@
...
+        }
+
+        amrex::Gpu::streamSynchronize();
+
+        CalculateExternalCurlA(m_field_names[i]);
+
+        // Generate parser for time function
+        m_A_external_time_parser[i] = std::make_unique<amrex::Parser>(
+            utils::parser::makeParser(m_A_ext_time_function[i],{"t",}));
+        m_A_time_scale[i] = m_A_external_time_parser[i]->compile<1>();
```
roelof-groenewald, line 218: It occurred to me that it would be good to add a bit of logic here that will check whether a time dependence was specified, and if so do a divergence cleaning on the specified vector potential, since $E = \partial A/\partial t$ requires the Coulomb gauge.
What do you think? If not we should probably at least print a warning to users that it is expected that A is divergence free.

Source/Fields.H
clarkse: I did not change it, may have been a pre-commit thing? I will take a look

Source/WarpX.cpp
clarkse: I was using it to help with source smoothing. It currently does not use radial weights in R, so there is some non-conservation of J and rho in r when using a straight bilinear filter radially on the density fields. I am planning to try to fix this to add the geometric factor, but this will be a follow-up PR. I can add a warning stating this and to use at your own risk, but I did not want to change the behavior of the FDTD solver until I can work with Remi on it. Issue here: https://github.com/ECP-WarpX/WarpX/issues/1943

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/ExternalVectorPotential.cpp
clarkse: I will add a warning here and then plug in the div cleaner as a follow up PR. I will need to test it to make sure it still works as expected with the offset staggering of A vs B.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/ExternalVectorPotential.cpp
clarkse: I refactored this to eliminate the LinComb and just set up the assignment loops with the EB mask.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
clarkse: This is following what the HybridPICModel is doing. Can we do this as a follow up clean up PR for both objects?

Python/pywarpx/picmi.py
```diff
@@ -1884,10 +1915,18 @@ def __init__(
...
         self.Jx_external_function = Jx_external_function
         self.Jy_external_function = Jy_external_function
         self.Jz_external_function = Jz_external_function
 
+        self.add_external_fields = None
+        self.A_external = A_external
+
+        if A_external is not None:
+            self.add_external_fields = True
+
```
roelof-groenewald: From the earlier changes you don't need `self.add_external_fields` anymore (and it is not used anywhere anymore).
```suggestion
        self.A_external = A_external

```

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
roelof-groenewald: Sounds good!

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -174,6 +210,26 @@ void WarpX::HybridPICEvolveFields ()
         m_fields.get_mr_levels(FieldType::rho_fp, finest_level),
         m_eb_update_E, false);
     FillBoundaryE(guard_cells.ng_FieldSolver, WarpX::sync_nodal_points);
+    // ApplyEfieldBoundary(0, PatchType::fine);
```
```suggestion
```

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1303,15 +1370,6 @@ void WarpX::CheckKnownIssues()
...
         }
-
-        const bool external_particle_field_used = (
-            mypc->m_B_ext_particle_s != "none" || mypc->m_E_ext_particle_s != "none"
-        );
-        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
-            !external_particle_field_used,
-            "The hybrid-PIC algorithm does not work with external fields "
-            "applied directly to particles."
-        );
```
roelof-groenewald: I don't think we should remove this. External fields directly applied to particles are still not supported in the hybrid-solver (those are the ones where the field multifabs are not updated, the Lorentz force is just augmented for each particle).

Source/WarpX.cpp
```diff
@@ -768,12 +768,22 @@ WarpX::ReadParameters ()
...
         {
-            // Filter currently not working with FDTD solver in RZ geometry along R
-            // (see https://github.com/ECP-WarpX/WarpX/issues/1943)
-            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(!use_filter || filter_npass_each_dir[0] == 0,
-                "In RZ geometry with FDTD, filtering can only be apply along z. This can be controlled by setting warpx.filter_npass_each_dir");
+            if (WarpX::electromagnetic_solver_id != ElectromagneticSolverAlgo::HybridPIC) {
+                // Filter currently not working with FDTD solver in RZ geometry along R
+                // (see https://github.com/ECP-WarpX/WarpX/issues/1943)
+                WARPX_ALWAYS_ASSERT_WITH_MESSAGE(!use_filter || filter_npass_each_dir[0] == 0,
+                    "In RZ geometry with FDTD, filtering can only be apply along z. This can be controlled by setting warpx.filter_npass_each_dir");
```
roelof-groenewald:
```suggestion
                    "In RZ geometry with FDTD, filtering can only be applied along z. This can be controlled by setting warpx.filter_npass_each_dir");
```

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -617,6 +655,10 @@ void FiniteDifferenceSolver::HybridPICSolveECylindrical (
                         + T_Algo::Dzz(Jr, coefs_z, n_coefs_z, i, j, 0, 0) - jr_val/(r*r);
                     Er(i, j, 0) -= eta_h * nabla2Jr;
                 }
+
+                if (include_external_fields && (rho_val >= rho_floor)) {
+                    Er(i, j, 0) -= Er_ext(i, j, 0);
```
roelof-groenewald, line 660: Why is the external field subtracted rather than added?

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -604,7 +638,11 @@ void FiniteDifferenceSolver::HybridPICSolveECylindrical (
                 // interpolate the nodal neE values to the Yee grid
                 auto enE_r = Interp(enE, nodal, Er_stag, coarsen, i, j, 0, 0);
 
-                Er(i, j, 0) = (enE_r - grad_Pe) / rho_val;
+                if (rho_val < rho_floor && holmstrom_vacuum_region) {
```
roelof-groenewald, line 641: It's curious to me what the clang-tidy tests focus on. I would think it would prefer this to be
```
if (holmstrom_vacuum_region && (rho_val < rho_floor))
```
due to the possibility to short-circuit and avoid the comparison.

roelof-groenewald (commented):
Overall it looks good! I just had a final few small comments and a somewhat bigger question:
In the solve loop $E_{ext} = -\partial A/\partial t$ is subtracted from the E-field (so in total $E = E_{plasma} - E_{ext}$ which is then used to update the B-field, giving $$B_{new} = B_{old} -\nabla \times E_{plasma} dt + \nabla \times E_{ext} dt.$$
But $\nabla \times E_{ext} dt = -\partial/\partial t \nabla \times A dt = -\partial/\partial t B_{ext} dt$, so in effect
$$B_{new} = B_{old} -\nabla \times E_{plasma} dt - B_{ext}^{t+1}.$$
Then in the next half of the field push, $B_{ext}^{t+1}$ is added to $B_{new}$, cancelling the effect of subtracting $E_{ext}$ from the E-field in the first place.
And then at the very end of the field-solve $B_{ext}$ is added to the B-field again. It seems to me it should be sufficient to have $E_{ext}$ added to the E-field at every evaluation of E, and that it would then carry the change in $B_{ext}$ through to Faraday's law (the B-field update). Maybe we can chat about this quickly tomorrow.

Here's what I mean in slightly easier to read font:
image

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
clarkse: So it is set up this way since the external fields are what they are, they operate in the soak through limit. So because of this since it is not solving A via a Poisson solve with totally self-consistent boundaries. I initially tried to just use the numerical integration to integrate the total B field, but because if the user is not extra careful to make sure the external field is self-consistent with the boundaries this approach was taken. I am doing a subtraction at the beginning of the Hypric PIC loop to remove the external component from the total field solution, then the E and B fields are the plasma response fields only moving through the integration loop. This allows for the Hybrid loop to proceed with integrating the plasma response into B
$\vec{E_{plasma}} + \vec{E_{ext}} = \frac{(\vec{J_{plasma}}-\vec{J_i}) \times (\vec{B_{plasma} + \vec{B_{ext}}})}{ne} - \frac{\nabla p_e}{ne} + \eta \vec{J_{plasma}} - \eta_h \nabla^2 \vec{J_{plasma}} $ 

Since the E and B fields at this point are the plasma fields in the Ohm's Law solver, however the solution is for a total field in order to get the plasma response only the E_ext needs to be subtracted from both sides of the equation.

In the vacuum region it gets a bit hairier. If the resistivity in that region is high enough or if the holmstrom flag is used then the Generalized Ohms Law asymptotes to:
$\vec{E_{resistive}}  \sim \eta \vec{J} $
This can be linearly added to the vacuum field to get the total field since the assumption is that it is a vacuum and not dense enough to shield an external field.
This resistive electric field gets integrated into the B field, then the resistivity term is dropped because of the cancellation of the drag term in the ion momentum update for the particle push.
Since the plasma response in the vacuum region is not well defined the algorithm switches to using this form of Ohm's law to have a smoothly varying E field around the density cutoff to limit vacuum fluctuations integrating into the plasma field. Since in the vacuum field the plasma E field should be zero except for some damping electric fields that show up near the edges of the plasma, then this gets integrated into B with the boundary conditions. E_ext is not subtracted off in this case as the plasma shielding response is just not accurate in the vacuum region. This really cuts down on the vacuum fluctuations for me and makes the simulations more stable.

roelof-groenewald: Sorry to get stuck on this more, but looking at it again a few hours after we talked I'm again unsure about the following:
You say the E-field from Ohm's law is the total field (i.e. $E_{plasma} + E_{ext}$), is there a clear argument for why that is the case? I agree Ohm's law of course includes induction (from $v_e \times B$, where $B$ here includes the external field), but it seems it would still only be the E-field due to the plasma response. Shouldn't there be an **additional** inductive E-field from $-\partial A/\partial t$?
In such a case we'd have:

$$E_{total} = E_{Ohm} + E_{ext} = E_{Ohm} - \partial A/\partial t$$

such that the full Faraday equation would give:

$$\frac{\partial}{\partial t} B_{total} = -\nabla\times E_{total}$$

$$\frac{\partial}{\partial t} (B_{plasma} + B_{ext}) = -\nabla\times E_{Ohm} + \frac{\partial}{\partial t} \nabla \times A$$

giving

$\frac{\partial}{\partial t} B_{plasma} = -\nabla\times E_{Ohm}$ and $\frac{\partial}{\partial t} B_{ext} = \frac{\partial}{\partial t} B_{ext}$.

If the Ohm's law calculated E field includes $-\partial A/\partial t$, there would not be a way to have Faraday's law operate on the full E-field that would result in a B-field with both the plasma and external B-field included. I must still be missing something...

roelof-groenewald: In terms of implementation, it seems to me that if the Ohm's law E-field does not already include $E_{ext}$, the formalism here would be that a time varying external field can be mathematically included by replacing the E-field calculation with $E = E_{Ohm} + E_{ext}$ where $E_{ext} = -\partial A / \partial t$. In that case the B-field integration from Faraday would itself add the external B-field. (As you mentioned earlier, the numerics could be complicated by boundary handling, in which case treating $B_{plasma}$ and $B_{ext}$ separately is understandable). 

But the point of difference with the current implementation is, for clarity:
Is $E_{total} = E_{Ohm} = E_{plasma} + E_{ext}$ or $E_{total} = E_{Ohm} + E_{ext} = E_{plasma} + E_{ext} $?

clarkse: $E_{ohm} = E_{total} = E_{plasma} + E_{external}$. I agree that it would be nicer if the integration worked out, but the boundary handling of the external field is finicky and not consistent with having a PEC at the EB or the domain edge. If we did a vector potential formulation of the integration I could do this all fully self-consistent. I would also like to be able to get rid of one of the sets of MFs. I cannot fix the boundary properly with having a soak through external field that ignores the boundaries and the high-frequency PEC response in this formulation.

I am exploring options to make this better, but I think it would require either having a displacement solver that has slowed down EM waves, or the vector potential formulation with infinite speed propagation. This is more of a solution to quickly add fields to try out. I think this is better suited to deal with in a follow up PR. I think I need to do a bit of work on the Magnetostatic solver, then I can address doing a self-consistent magneto-inductive version of the hybrid algorithm.

roelof-groenewald: Why do you say $E_{ohm} = E_{plasma} + E_{external}$? That is just the part I'm stuck on. It seems to me it should be the case that $E_{ohm} = E_{plasma}$. 

I think the implementation of the model as you propose is good, so I'll approve this PR and merge now. But I think we should continue a discussion of the accuracy of the underlying physics model (and adjust the implementation in future PRs if we agree it is needed).

roelof-groenewald (approved):
The code changes look good! Thanks for the all the work to get this implemented!

yanghf263:
Dear @clarkse , I'm conducting a 2D turbulence simulation using the hybrid PIC method, with a constant background magnetic field By = B0. Under periodic boundary conditions, how can this background field be introduced as an external field? It seems that A_external = (0, 0, -x*B0) does not satisfy periodic boundary conditions and thus cannot be used. #5960 . I'd be grateful for your insightful advice.

clarkse:
> Dear @clarkse , I'm conducting a 2D turbulence simulation using the hybrid PIC method, with a constant background magnetic field By = B0. Under periodic boundary conditions, how can this background field be introduced as an external field? It seems that A_external = (0, 0, -x*B0) does not satisfy periodic boundary conditions and thus cannot be used. #5960 . I'd be grateful for your insightful advice.

I believe it should work since the E/B fields are what it generates and does not enforce BCs on analytical A. So long as curlA is periodic you will be OK. I would note though that this is primarily used for time varying fields, so if it is present at t=0, the initial conditions for the plasma and initial fields still needs to be self-consistent. Typically for simulations with uniform background the field is included in the initial field condition. I guess I would do something along the lines of doing an initial magnetoelectrostatic solve with ions and electrons collocated to be space charge neutral, then solve for the self-consistent field using the magnetostatic solver and write the field out to an openPMD file. Then you can kick off a hybrid run with the ions only and the magnetostatic solution to run forward with the hybrid solver.

There is one caveat that there is some divergence errors present in the existing magnetostatic solver due to interpolation issues, however the div cleaner should be able to clean it up well enough for the hybrid solver.

yanghf263:
> > Dear @clarkse , I'm conducting a 2D turbulence simulation using the hybrid PIC method, with a constant background magnetic field By = B0. Under periodic boundary conditions, how can this background field be introduced as an external field? It seems that A_external = (0, 0, -x*B0) does not satisfy periodic boundary conditions and thus cannot be used. #5960 . I'd be grateful for your insightful advice.
> 
> I believe it should work since the E/B fields are what it generates and does not enforce BCs on analytical A. So long as curlA is periodic you will be OK. I would note though that this is primarily used for time varying fields, so if it is present at t=0, the initial conditions for the plasma and initial fields still needs to be self-consistent. Typically for simulations with uniform background the field is included in the initial field condition. I guess I would do something along the lines of doing an initial magnetoelectrostatic solve with ions and electrons collocated to be space charge neutral, then solve for the self-consistent field using the magnetostatic solver and write the field out to an openPMD file. Then you can kick off a hybrid run with the ions only and the magnetostatic solution to run forward with the hybrid solver.
> 
> There is one caveat that there is some divergence errors present in the existing magnetostatic solver due to interpolation issues, however the div cleaner should be able to clean it up well enough for the hybrid solver.

Dear @clarkse , thank you very much for your detailed and insightful reply. From your response, I've gathered the following key points:

1. WarpX does not require `A_external` to be a periodic function. As long as the magnetic field generated from `A_external` (i.e., **curl A**) is periodic, it should be acceptable.  
2. It is generally recommended to initialize the background field directly using methods like `AnalyticInitialField`, as this approach is more physically self-consistent.  
3. When using `A_external`, it’s important to ensure that the initial plasma state and field configuration are self-consistent.

I greatly appreciate these suggestions and completely agree with them.

However, I’ve encountered a few issues during my 2D simulations that I’m still trying to understand:

1. When I set only `A_external = (0, 0, -x*B0)`, the code reports an error during initialization: `Efield_fp[lev][idim]->is_finite()' failed`. Could this be due to an inconsistency between the initial plasma and the field configuration?
2. I've found that increasing the number of substeps can eliminate this error. Is this the recommended approach?
3. Under the same plasma conditions, why does initializing B0 using `AnalyticInitialField` not lead to consistency issues, while doing so via `A_external` does?

Thank you once again for your valuable guidance!

clarkse:
> > > Dear @clarkse , I'm conducting a 2D turbulence simulation using the hybrid PIC method, with a constant background magnetic field By = B0. Under periodic boundary conditions, how can this background field be introduced as an external field? It seems that A_external = (0, 0, -x*B0) does not satisfy periodic boundary conditions and thus cannot be used. #5960 . I'd be grateful for your insightful advice.
> > 
> > 
> > I believe it should work since the E/B fields are what it generates and does not enforce BCs on analytical A. So long as curlA is periodic you will be OK. I would note though that this is primarily used for time varying fields, so if it is present at t=0, the initial conditions for the plasma and initial fields still needs to be self-consistent. Typically for simulations with uniform background the field is included in the initial field condition. I guess I would do something along the lines of doing an initial magnetoelectrostatic solve with ions and electrons collocated to be space charge neutral, then solve for the self-consistent field using the magnetostatic solver and write the field out to an openPMD file. Then you can kick off a hybrid run with the ions only and the magnetostatic solution to run forward with the hybrid solver.
> > There is one caveat that there is some divergence errors present in the existing magnetostatic solver due to interpolation issues, however the div cleaner should be able to clean it up well enough for the hybrid solver.
> 
> Dear @clarkse , thank you very much for your detailed and insightful reply. From your response, I've gathered the following key points:
> 
> 1. WarpX does not require `A_external` to be a periodic function. As long as the magnetic field generated from `A_external` (i.e., **curl A**) is periodic, it should be acceptable.
> 2. It is generally recommended to initialize the background field directly using methods like `AnalyticInitialField`, as this approach is more physically self-consistent.
> 3. When using `A_external`, it’s important to ensure that the initial plasma state and field configuration are self-consistent.
> 
> I greatly appreciate these suggestions and completely agree with them.
> 
> However, I’ve encountered a few issues during my 2D simulations that I’m still trying to understand:
> 
> 1. When I set only `A_external = (0, 0, -x*B0)`, the code reports an error during initialization: `Efield_fp[lev][idim]->is_finite()' failed`. Could this be due to an inconsistency between the initial plasma and the field configuration?
> 2. I've found that increasing the number of substeps can eliminate this error. Is this the recommended approach?
> 3. Under the same plasma conditions, why does initializing B0 using `AnalyticInitialField` not lead to consistency issues, while doing so via `A_external` does?
> 
> Thank you once again for your valuable guidance!

I think I would need to see the problem. Could you open an issue with a reproducer input file exemplifying the differences and tag me in it? It is possible there may be a bug as I have not exercised these routines much in XZ slab geometry.  I usually go one of two ways when I am getting numerical instability, increase substeps or edge up the hyper-resistivity if there are strong currents to suppress whistler noise. There may be some inconsistency at the boundary that needs to be worked out in your particular geometry.

PR #5279, opened 2024-09-17
Docs: how to use `add_subdirectory` for new tests

EZoni:
@oshapoval noticed that instructions on how to use `add_subdirectory` for tests added to new directories were missing.


lucafedeli88 (approved):
Thanks!  It's a good idea to explain this in the documentation.

PR #5280, opened 2024-09-17
[WIP] Increase coverage of clang sanitizer tests

lucafedeli88:
In CI tests based on clang sanitizers (UB and thread), most of the time (~ 1h) is spent in compiling the code, while just a few minutes are spent actually running some simulations. This means that we can increase the coverage of the test by adding some more simulations to the tests with a negligible increase of the total runtime for this CI test.
This PR does just that.

Note: we can probably add even more tests.


lucafedeli88:
This PR is no longer needed. https://github.com/ECP-WarpX/WarpX/pull/5597 adds more tests for the UB sanitizer and a new PR will be created to add more cases for the thread sanitizer as soon as the issues are fixed.

PR #5282, opened 2024-09-17
[DoNotMerge] Revert "Doc: Conda-Forge w/ heFFTe (#4989)"

RemiLehe:
The installation instructions include `heffte` in the dependencies that are installed by default for WarpX (when compiling by hand). However, this does not work on MacOS. (The `conda` installation fails in this case.) 
Since we are not yet using heFFTe in WarpX anyway, I would recommend to temporarily remove heffte from the installation instructions.

This reverts commit a4fbb137583372d7d632d5317845f7a15fdc162d.

### DIFF
--- a/Docs/source/install/dependencies.rst
+++ b/Docs/source/install/dependencies.rst
@@ -81,7 +81,7 @@ Conda (Linux/macOS/Windows)
 
       .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


ax3l:
Thanks for checking this!

This PR
https://github.com/conda-forge/heffte-feedstock/pull/6
will add Apple aarch64 support. It should be available in about an hr.

RemiLehe:
The installation issue on MacOS is fixed.

Issue #5205, opened 2024-09-03
CTest follow-up: add single-precision tests


EZoni:
Follow-up on #5068. Run all (if not, a subset of) tests in single-precision, compare against existing double-precision checksum files with a larger tolerance.

ax3l:
To generalize the analysis scripts, we could change the APIs of the tolerance argument to the checksum API, and pass a dictionary (double and single as keys) for the tolerances. Or have two arguments, one double tolerance one float (with a dictionary, we keep it open to have other types later, but I do not think this is timely.)

Inside the checksum API, we could do a test for WarpX config used or type of a data file entry read to determine if something was a SP or DP run.

PR #5285, opened 2024-09-18
Add Time-Averaged Field Diagnostics

### LINKED ISSUE #5165: Diagnostics: Time-Averaged Fields
For LPI simulations, we would like a diagnostics that time-averages (over N subsequent steps) the output fields.

This is relatively easy to implement if we enforce the assumption that the output period P is greater than the averaging period N. For many simulations, that is sufficient and true.

n01r:
This PR adds time-averaged field diagnostics to the WarpX output

- [x] code
- [x] docs
- [x] tests
- [x] example
- meta-data $\Longrightarrow$ Follow-up PR
- make compatible with adaptive time stepping $\Longrightarrow$ Follow-up PR

Currently, compiling for GPU creates a linking error. We would like to see the CI output to possibly find the issue.
Hence, we do not flag this as WIP right now.

This PR is based on work performed during the *2024 WarpX Refactoring Hackathon* and was created together with @RevathiJambunathan. :slightly_smiling_face: 

Successfully merging this pull request may close #5165.

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -2634,10 +2634,10 @@ In-situ visualization
 ^^^^^^^^^^^^^^^^^^^^^
 
 WarpX has five types of diagnostics:
-``FullDiagnostics`` consist in dumps of fields and particles at given iterations,
-``TimeAveragedDiagnostics`` only allow field data which they output after averaging over a period of time,
-``BackTransformedDiagnostics`` are used when running a simulation in a boosted frame, to reconstruct output data to the lab frame,
-``BoundaryScrapingDiagnostics`` are used to collect the particles that are absorbed at the boundary, throughout the simulation, and
+``Full`` diagnostics consist in dumps of fields and particles at given iterations,
+``TimeAveraged`` diagnostics only allow field data, which they output after averaging over a period of time,
+``BackTransformed`` diagnostics are used when running a simulation in a boosted frame, to reconstruct output data to the lab frame,
+``BoundaryScraping`` diagnostics are used to collect the particles that are absorbed at the boundary, throughout the simulation, and
 ``ReducedDiags`` allow the user to compute some reduced quantity (particle temperature, max of a field) and write a small amount of data to text files.
 Similar to what is done for physical species, WarpX has a class Diagnostics that allows users to initialize different diagnostics, each of them with different fields, resolution and period.
 This currently applies to standard diagnostics, but should be extended to back-transformed diagnostics and reduced diagnostics (and others) in a near future.
@@ -2889,13 +2889,13 @@ In-situ capabilities can be used by turning on Sensei or Ascent (provided they a
 Time-Averaged Diagnostics
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
-``TimeAveraged`` diagnostics are a special type of ``FullDiagnostics`` that allows for the output of time-averaged field data.
+``TimeAveraged`` diagnostics are a special type of ``Full`` diagnostics that allows for the output of time-averaged field data.
 This type of diagnostics can be created using ``<diag_name>.diag_type = TimeAveraged``.
 We support only field data and related options from the list at `Full Diagnostics`_.
 
 .. note::
 
-    As with ``FullDiagnostics``, ``TimeAveraged`` diagnostics output the initial **instantaneous** conditions of the selected fields on step 0 (unless more specific output intervals exclude output for step 0).
+    As with ``Full`` diagnostics, ``TimeAveraged`` diagnostics output the initial **instantaneous** conditions of the selected fields on step 0 (unless more specific output intervals exclude output for step 0).
 
 In addition, ``TimeAveraged`` diagnostic options include:
 
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -2934,6 +2934,7 @@ BackTransformed Diagnostics
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 ``BackTransformed`` diag type are used when running a simulation in a boosted frame, to reconstruct output data to the lab frame. This option can be set using ``<diag_name>.diag_type = BackTransformed``. We support the following list of options from `Full Diagnostics`_
+
     ``<diag_name>.format``, ``<diag_name>.openpmd_backend``, ``<diag_name>.dump_rz_modes``, ``<diag_name>.file_prefix``, ``<diag_name>.diag_lo``, ``<diag_name>.diag_hi``, ``<diag_name>.write_species``, ``<diag_name>.species``.
 
     Additional options for this diagnostic include:
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -2633,8 +2633,9 @@ Diagnostics and output
 In-situ visualization
 ^^^^^^^^^^^^^^^^^^^^^
 
-WarpX has four types of diagnostics:
+WarpX has five types of diagnostics:
 ``FullDiagnostics`` consist in dumps of fields and particles at given iterations,
+``TimeAveragedDiagnostics`` only allow field data which they output after averaging over a period of time,
 ``BackTransformedDiagnostics`` are used when running a simulation in a boosted frame, to reconstruct output data to the lab frame,
 ``BoundaryScrapingDiagnostics`` are used to collect the particles that are absorbed at the boundary, throughout the simulation, and
 ``ReducedDiags`` allow the user to compute some reduced quantity (particle temperature, max of a field) and write a small amount of data to text files.
@@ -2881,6 +2882,47 @@ In-situ capabilities can be used by turning on Sensei or Ascent (provided they a
 
 * ``warpx.mffile_nstreams`` (`int`) optional (default `4`)
     Limit the number of concurrent readers per file.
+
+
+.. _running-cpp-parameters-diagnostics-timeavg:
+
+Time-Averaged Diagnostics
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``TimeAveraged`` diagnostics are a special type of ``FullDiagnostics`` that allows for the output of time-averaged field data.
+This type of diagnostics can be created using ``<diag_name>.diag_type = TimeAveraged``.
+We support only field data and related options from the list at `Full Diagnostics`_.
+
+    In addition, `TimeAveraged` diagnostic options include:
+
+* ``<diag_name>.time_average_mode`` (`string`, default `none`)
+    Describes the operating mode for time averaged field output.
+
+    * ``none`` for no averaging (instantaneous fields)
+
+    * ``fixed_start`` for a diagnostic that averages all fields between the current output step and a fixed point in time
+
+    * ``dynamic_start`` for a constant averaging period and output at different points in time (non-overlapping)
+
+    .. note::
+
+        To enable time-averaged field output with intervals tightly spaced enough for overlapping averaging periods,
+        please create additional instances of ``TimeAveraged`` diagnostics.
+
+* ``<diag_name>.average_period_steps`` (`int`)
+    Configures the number of time steps in an averaging period.
+    Set this only in the ``dynamic_start`` mode and only if ``average_period_time`` has not already been set.
+    Will be ignored in the ``fixed_start`` mode (with warning).
+
+* ``<diag_name>.average_period_time`` (`float`, in seconds)
+    Configures the time (SI units) in an averaging period.
+    Set this only in the ``dynamic_start`` mode and only if ``average_period_steps`` has not already been set.
+    Will be ignored in the ``fixed_start`` mode (with warning).
+
+* ``<diag_name>.average_start_step`` (`int`)
+    Configures the time step at which time-averaging begins.
+    Set this only in the ``fixed_start`` mode.
+    Will be ignored in the ``dynamic_start`` mode (with warning).
 
 .. _running-cpp-parameters-diagnostics-btd:
 
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -2893,7 +2893,11 @@ Time-Averaged Diagnostics
 This type of diagnostics can be created using ``<diag_name>.diag_type = TimeAveraged``.
 We support only field data and related options from the list at `Full Diagnostics`_.
 
-    In addition, `TimeAveraged` diagnostic options include:
+.. note::
+
+    As with ``FullDiagnostics``, ``TimeAveraged`` diagnostics output the initial **instantaneous** conditions of the selected fields on step 0 (unless more specific output intervals exclude output for step 0).
+
+In addition, ``TimeAveraged`` diagnostic options include:
 
 * ``<diag_name>.time_average_mode`` (`string`, default `none`)
     Describes the operating mode for time averaged field output.
--- a/Examples/Physics_applications/laser_ion/CMakeLists.txt
+++ b/Examples/Physics_applications/laser_ion/CMakeLists.txt
@@ -6,19 +6,9 @@ add_warpx_test(
     2  # dims
     2  # nprocs
     inputs_test_2d_laser_ion_acc  # inputs
-    analysis_default_openpmd_regression.py  # analysis
+    analysis_test_laser_ion.py  # analysis
     diags/diagInst/  # output
     OFF  # dependency
-)
-
-add_warpx_test(
-    test_2d_laser_ion_acc_time_avg  # name
-    2  # dims
-    2  # nprocs
-    inputs_test_2d_laser_ion_acc  # inputs; will not be used b/c dependent test
-    analysis_time_avg_diags.py  # analysis
-    diags/diagTimeAvg/  # output
-    test_2d_laser_ion_acc  # dependency
 )
 
 add_warpx_test(
@@ -26,18 +16,7 @@ add_warpx_test(
     2  # dims
     2  # nprocs
     inputs_test_2d_laser_ion_acc_picmi.py  # inputs
-    analysis_default_openpmd_regression.py  # analysis
+    analysis_test_laser_ion.py  # analysis
     diags/diagInst/  # output
     OFF  # dependency
 )
-
-# We do not have an intervals parser for PICMI yet which would accept more than a single integer for the output period
-#add_warpx_test(
-#    test_2d_laser_ion_acc_time_avg_picmi  # name
-#    2  # dims
-#    2  # nprocs
-#    inputs_test_2d_laser_ion_acc_picmi.py  # inputs
-#    analysis_time_avg_diags.py  # analysis
-#    diags/diagTimeAvg/  # output
-#    test_2d_laser_ion_acc_picmi  # dependency
-#)
--- a/Examples/Physics_applications/laser_ion/CMakeLists.txt
+++ b/Examples/Physics_applications/laser_ion/CMakeLists.txt
@@ -12,6 +12,16 @@ add_warpx_test(
 )
 
 add_warpx_test(
+    test_2d_laser_ion_acc_time_avg  # name
+    2  # dims
+    2  # nprocs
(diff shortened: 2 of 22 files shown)


archermarx:
Thanks for making this PR! One complication is that with https://github.com/ECP-WarpX/WarpX/pull/5176, adaptive timestepping will be possible. This means that the timestep length can change, so simply summing the multifabs and dividing by the period won't be enough for a proper time average. I don't think addressing this will be too hard, however.

n01r:
Thanks, @archermarx! I wasn't aware of that yet. :slightly_smiling_face:  We can do a follow-up and catch if `dt_update_interval` is set. In that case, we could create a cumulative sum weighted with the time step length and also add up the total time as we go. During the flush, we then divide by the total time.

archermarx:
> Thanks, @archermarx! I wasn't aware of that yet. 🙂 We can do a follow-up and catch if `dt_update_interval` is set. In that case, we could create a cumulative sum weighted with the time step length and also add up the total time as we go. During the flush, we then divide by the total time.

Yep, my thought exactly. The adaptive timestepping PR was just merged, too.

Source/Diagnostics/MultiDiagnostics.cpp
```diff
@@ -22,11 +22,13 @@ MultiDiagnostics::MultiDiagnostics ()
     alldiags.resize( ndiags );
     for (int i=0; i<ndiags; i++){
         if ( diags_types[i] == DiagTypes::Full ){
-            alldiags[i] = std::make_unique<FullDiagnostics>(i, diags_names[i]);
+            alldiags[i] = std::make_unique<FullDiagnostics>(i, diags_names[i], diags_types[i]);
+        } else if ( diags_types[i] == DiagTypes::TimeAveraged ){
+            alldiags[i] = std::make_unique<FullDiagnostics>(i, diags_names[i], diags_types[i]);
```
EZoni, line 25: Should this be `TimeAveragedDiagnostics` instead? I think this should fix the clang-tidy error "repeated branch in conditional chain [bugprone-branch-clone,-warnings-as-errors]":
```suggestion
            alldiags[i] = std::make_unique<TimeAveragedDiagnostics>(i, diags_names[i], diags_types[i]);
```

Source/Diagnostics/Diagnostics.cpp
```diff
@@ -574,6 +579,15 @@ Diagnostics::ComputeAndPack ()
             // Check that the proper number of components of mf_avg were updated.
             AMREX_ALWAYS_ASSERT( icomp_dst == m_varnames.size() );
 
+            if (m_diag_type == DiagTypes::TimeAveraged) {
+
+                amrex::Real real_a = 1.0;
```
EZoni: This should fix the clang-tidy error "variable 'real_a' of type 'amrex::Real' (aka 'float') can be declared 'const' [misc-const-correctness,-warnings-as-errors]":
```suggestion
                const amrex::Real real_a = 1.0;
```

Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -136,11 +192,27 @@ FullDiagnostics::Flush ( int i_buffer, bool /* force_flush */ )
     // is supported for BackTransformed Diagnostics, in BTDiagnostics class.
     auto & warpx = WarpX::GetInstance();
 
+    if (m_diag_type == DiagTypes::TimeAveraged) {
+        if (m_time_average_type == TimeAverageType::Static || m_time_average_type == TimeAverageType::Dynamic) {
+            // Loop over the output levels and divide by the number of steps in the averaging period
+            for (int lev = 0; lev < nlev_output; ++lev) {
+                m_sum_mf_output.at(i_buffer).at(lev).mult(1./m_average_period_steps);
```
EZoni: I wonder if this could fix the clang-tidy error "narrowing conversion from 'double' to 'amrex::Real' (aka 'float') [bugprone-narrowing-conversions,-warnings-as-errors]":
```suggestion
                m_sum_mf_output.at(i_buffer).at(lev).mult(1._rt/static_cast<amrex::Real>(m_average_period_steps));
```

Source/Diagnostics/MultiDiagnostics.cpp
n01r: We don't actually have a full class `TimeAveragedDiagnostics` because our diagnostics are basically `FullDiagnostics` with the addition of time averaging and without particles. 

I am already having new changes ready to address the clang-tidy errors.
What I was thinking was to just do
```C++
if ( diags_types[i] == DiagTypes::Full || diags_types[i] == DiagTypes::TimeAveraged ){
    alldiags[i] = std::make_unique<FullDiagnostics>(i, diags_names[i], diags_types[i]);
```

EZoni:
I think #5296 should fix the clang-tidy errors that remain, which don't seem related to this PR.

ax3l:
>  We can do a follow-up and catch if dt_update_interval is set
...
> Yep, my thought exactly. The adaptive timestepping PR was just merged, too.

Thanks for the catch! Please add a runtime assert in this PR now.

n01r:
Added commit that will cause the sim to abort when `dt_update_interval` is set.

The last thing for this PR that remains from our list is the CI tests.

@ax3l, should we also use the recently merged `MultiFabRegister`?

n01r:
Just empirically, I tried this in our [Laser-Ion Acceleration](https://warpx.readthedocs.io/en/24.08/usage/examples/laser_ion/README.html) example. At step 1000, I compared between the electron density output from the in-situ generated averaged output and a post-processed output that I created from 25 instantaneous outputs. The relative difference is on the order of 1e-16. :tada: 
Figure 33

Created with the following input script:
[inputs_test_2d_laser_ion_acc.txt](https://github.com/user-attachments/files/17236190/inputs_test_2d_laser_ion_acc.txt)

n01r:
And comparing instantaneous output with averaged output for, e.g. Ez shows the much clearer structure one can see with this diagnostic.
image

EZoni:
@n01r @RevathiJambunathan 
After resolving the merge conflicts, do you think the feature added in this PR is complete and ready to go, pending review?

n01r:
@EZoni, the CI tests are still missing. I was thinking of adding these today since Perlmutter is under maintenance anyway.

Examples/Physics_applications/laser_ion/CMakeLists.txt
```diff
@@ -20,3 +30,14 @@ add_warpx_test(
     diags/diagInst/  # output
     OFF  # dependency
 )
+
+# We do not have an intervals parser for PICMI yet which would accept more than a single integer for the output period
```
n01r: We should think about implementing an intervals parser for our PICMI diagnostics which can handle more than a single integer. It should be able to do the same as our app input: https://warpx.readthedocs.io/en/latest/usage/parameters.html#intervals-parser

Examples/Physics_applications/laser_ion/CMakeLists.txt
```diff
@@ -11,6 +11,16 @@ add_warpx_test(
...
 )
 
+add_warpx_test(
+    test_2d_laser_ion_acc_time_avg  # name
+    2  # dims
+    2  # nprocs
+    inputs_test_2d_laser_ion_acc  # inputs; will not be used b/c dependent test
+    analysis_time_avg_diags.py  # analysis
+    diags/diagTimeAvg/  # output
+    test_2d_laser_ion_acc  # dependency
```
n01r: I had hoped that it would not run the full WarpX simulation if I specified the dependency. But I guess that is not true. In principle, it is enough to run the analysis script on the data of the main test.
Is there a quick way to do this or should we just run this twice and write double the amount of data?

EZoni: I'm not sure if I understand this correctly, you are saying that the dependent test `test_2d_laser_ion_acc_time_avg` is not supposed to run WarpX but only an analysis script on the result produced by another test? In that case I would just add the new analysis to the existing analysis script that is already run for the other test. 

The dependency is typically used for restart tests or for preparatory scripts (e.g., LASY scripts that produce a LASY file), meaning that the dependent test always runs WarpX.

n01r: I see. Yes, in principle, I just need to run a script on the data that already `test_2d_laser_ion_acc` produces.
I will just combine the normal regression test with the time averaged diagnostics test, then. :)

n01r: I made the necessary changes.

n01r:
@EZoni, feel free to review :)

Examples/Physics_applications/laser_ion/analysis_test_laser_ion.py
```diff
@@ -0,0 +1,101 @@
...
+            rtol=2e-6,
+        )
+    else:
+        evaluate_checksum(
+            test_name=test_name,
+            output_file=output_file,
+            output_format="openpmd",
+        )
+
+
```
EZoni: We want to change the way we handle single-precision tests anyways (see #5205), so if this test is not a single-precision test, I think we can remove this function and simply call `evaluate_checksum` below, consistently with the other analysis scripts in our test suite (since #5297):
```suggestion
```

Examples/Physics_applications/laser_ion/analysis_test_laser_ion.py
```diff
@@ -0,0 +1,101 @@
...
+    #   NOTE: works only in the example directory due to relative path import
+    analyze_openpmd_regression(inst_output_file)
+
+    # TODO: implement intervals parser for PICMI that allows more complex output periods
+    if "picmi" not in test_name:
+        # Functionality test for TimeAveragedDiagnostics
+        compare_time_avg_with_instantaneous_diags(
+            dir_inst=sys.argv[1],
+            dir_avg="diags/diagTimeAvg/",
+        )
```
EZoni: Given the comment above, I would rewrite it like this:
```suggestion
    # NOTE: works only in the example directory due to relative path import
    # compare checksums
    evaluate_checksum(
        test_name=os.path.split(os.getcwd())[1],
        output_file=sys.argv[1],
    )

    # TODO: implement intervals parser for PICMI that allows more complex output periods
    test_name = os.path.split(os.getcwd())[1]
    if "picmi" not in test_name:
        # Functionality test for TimeAveragedDiagnostics
        compare_time_avg_with_instantaneous_diags(
            dir_inst=sys.argv[1],
            dir_avg="diags/diagTimeAvg/",
        )
```

If possible, I would leave the `evaluate_checksum` call as in this suggestion (instead of passing, for example, `test_name=test_name` with a `test_name` variable defined before) because it matches the format of this "unit code block" across the test suite and will make it easier to implement some upcoming changes in a more automatic way.

EZoni:
@n01r 
If it's okay with you, I will push a fix for the CI tests (bug introduced by the changes I had suggested in my last review), rebase on the latest `development`, and then proceed with a final review.

n01r:
Sounds good to me, @EZoni :)

EZoni:
@n01r 

I guess this (in the PR description) isn't a problem anymore, right?

> Currently, compiling for GPU creates a linking error. We would like to see the CI output to possibly find the issue.
Hence, we do not flag this as WIP right now.

Docs/source/usage/parameters.rst
```diff
@@ -2633,8 +2633,9 @@ Diagnostics and output
 In-situ visualization
 ^^^^^^^^^^^^^^^^^^^^^
 
-WarpX has four types of diagnostics:
+WarpX has five types of diagnostics:
 ``FullDiagnostics`` consist in dumps of fields and particles at given iterations,
+``TimeAveragedDiagnostics`` only allow field data which they output after averaging over a period of time,
 ``BackTransformedDiagnostics`` are used when running a simulation in a boosted frame, to reconstruct output data to the lab frame,
 ``BoundaryScrapingDiagnostics`` are used to collect the particles that are absorbed at the boundary, throughout the simulation, and
```
EZoni: Maybe an opportunity to change the names exposed here:
```suggestion
``Full`` diagnostics consist in dumps of fields and particles at given iterations,
``TimeAveraged`` diagnostics only allow field data, which they output after averaging over a period of time,
``BackTransformed`` diagnostics are used when running a simulation in a boosted frame, to reconstruct output data to the lab frame,
``BoundaryScraping`` diagnostics are used to collect the particles that are absorbed at the boundary, throughout the simulation, and
```
I think they do not match the actual input names, do they?

Docs/source/usage/parameters.rst
```diff
@@ -2882,12 +2883,58 @@ In-situ capabilities can be used by turning on Sensei or Ascent (provided they a
 * ``warpx.mffile_nstreams`` (`int`) optional (default `4`)
     Limit the number of concurrent readers per file.
 
+
+.. _running-cpp-parameters-diagnostics-timeavg:
+
+Time-Averaged Diagnostics
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``TimeAveraged`` diagnostics are a special type of ``FullDiagnostics`` that allows for the output of time-averaged field data.
```
```suggestion
``TimeAveraged`` diagnostics are a special type of ``Full`` diagnostics that allows for the output of time-averaged field data.
```

Docs/source/usage/parameters.rst
```diff
@@ -2882,12 +2883,58 @@ In-situ capabilities can be used by turning on Sensei or Ascent (provided they a
...
+This type of diagnostics can be created using ``<diag_name>.diag_type = TimeAveraged``.
+We support only field data and related options from the list at `Full Diagnostics`_.
+
+.. note::
+
+    As with ``FullDiagnostics``, ``TimeAveraged`` diagnostics output the initial **instantaneous** conditions of the selected fields on step 0 (unless more specific output intervals exclude output for step 0).
```
```suggestion
    As with ``Full`` diagnostics, ``TimeAveraged`` diagnostics output the initial **instantaneous** conditions of the selected fields on step 0 (unless more specific output intervals exclude output for step 0).
```

Examples/Physics_applications/laser_ion/analysis_test_laser_ion.py
```diff
@@ -0,0 +1,85 @@
...
+
+    for i in np.arange(avg_output_step - avg_period_steps + 1, avg_output_step + 1):
+        data_inst += load_field_from_iteration(si, i, field, coord)
+
+    data_inst = data_inst / avg_period_steps
+
+    data_avg = load_field_from_iteration(sa, avg_output_step, field, coord)
+
+    # Compare the data
+    if np.allclose(data_inst, data_avg):
```
EZoni: I think the default tolerances for NumPy's `allclose` are `rtol=1e-05` and `atol=1e-08` (seen at https://numpy.org/doc/stable/reference/generated/numpy.allclose.html). Shall we use stricter tolerances for this comparison?

Source/Diagnostics/Diagnostics.H
```diff
@@ -266,6 +270,12 @@ protected:
      *  The second vector is loops over the total number of levels.
      */
     amrex::Vector< amrex::Vector< amrex::MultiFab > >  m_mf_output;
+    /** summation multifab, where all fields (computed, cell-centered, and stacked)
+     *  are summed for every step in a time averaging period.
+     *  The first vector is for total number of snapshots. (=1 for FullDiagnostics)
```
EZoni, line 275: Just checking, `=1` in this comment means something like "only one element"? If so (that is, if `1` does not refer to the index to be used to access that vector layer), a more verbose description might be clearer.

Source/Diagnostics/Diagnostics.cpp
```diff
@@ -575,6 +583,15 @@ Diagnostics::ComputeAndPack ()
             // Check that the proper number of components of mf_avg were updated.
             AMREX_ALWAYS_ASSERT( icomp_dst == m_varnames.size() );
 
+            if (m_diag_type == DiagTypes::TimeAveraged) {
+
+                const amrex::Real real_a = 1.0;
+                // call amrex sax operation to do the following
+                amrex::MultiFab::Saxpy(
+                        m_sum_mf_output[i_buffer][lev], real_a, m_mf_output[i_buffer][lev],
+                        0, 0, m_mf_output[i_buffer][lev].nComp(), m_mf_output[i_buffer][lev].nGrowVect());
```
EZoni: We could maybe write what AMReX's `Saxpy` does, as the name isn't that self-explanatory (is it?):
```suggestion
                // Compute m_sum_mf_output += real_a*m_mf_output
                amrex::MultiFab::Saxpy(
                        m_sum_mf_output[i_buffer][lev], real_a, m_mf_output[i_buffer][lev],
                        0, 0, m_mf_output[i_buffer][lev].nComp(), m_mf_output[i_buffer][lev].nGrowVect());
```

EZoni (commented):
Thanks for this PR, looks great to me, and it's being thoroughly tested with your simulations. I left only a few minor comments.

n01r:
> @n01r
> 
> I guess this (in the PR description) isn't a problem anymore, right?
> 
> > Currently, compiling for GPU creates a linking error. We would like to see the CI output to possibly find the issue.
> > Hence, we do not flag this as WIP right now.

Oh, yea. That is not a thing anymore.
The CI also does not complain. I could check if locally I can compile for GPU in 3D because I think this was where I had some problems. But that may be my setup.

Examples/Physics_applications/laser_ion/analysis_test_laser_ion.py
n01r: Uh, I don't know. I think they are pretty strict already because the absolute values are on the order of `1e8`.

Source/Diagnostics/Diagnostics.H
n01r: Unsure. I think @RevathiJambunathan may have added/suggested this comment.

Examples/Physics_applications/laser_ion/analysis_test_laser_ion.py
EZoni: I see, I think I'm mostly concernced about the relative tolerance, where the order of the absolute values in the calculation of things like $|x_{avg}-x_{inst}|/|x_{inst}|$ should not play a role as it's factored out.

n01r: Right, that's fair. But what should we set it to?
Weren't some other recent tests failing because of too strict tolerance limits?

EZoni: We definitely had that issue with checksums, but I thought this could be more controlled. Sometimes we do expect close results, for example when we check restarts.

I guess I would be interested in making the tolerance here as strict as possible, just for the sake of making sure that we don't overlook future bugs.

I think `np.allclose` must be used with care. They have some info in the docs, e.g., about the fact that `atol` must be set with care otherwise the function won't work for small numbers. For instance, `np.allclose(a=1e-9, b=2e-9)` returns `True`, which one probably would not want, because of the default values of `rtol` and `atol` (in this case the issue is with `atol`).

In our case I think the default `atol` is okay because the order of magnitude of the values is $10^8$, but we can maybe try to lower `rtol` as much as possible and see if it causes instabilities when we start running this test regularly.

I think we could try `rtol=1e-12` like for the restart tests (the checksums part in that case).

Source/Diagnostics/Diagnostics.H
n01r: Actually, that was copy-pasted and then modified from the comment of the `m_mf_output` above.
We could consider making this in all the places where it appears in this file.
But a separate PR would probably be better for this.

Examples/Physics_applications/laser_ion/analysis_test_laser_ion.py
EZoni:
```suggestion
    if np.allclose(data_inst, data_avg, rtol=1e-12):
```

EZoni (approved):

EZoni:
@n01r 

Thanks! If possible, I would suggest that we track the work planned for follow-up PRs in an issue, just so we don't forget.

n01r:
Agreed, @EZoni. Would you recommend continuing to use #5165 or opening new ones for individual tasks?

EZoni:
Thanks, as you prefer. I think #5165 will be closed automatically because this PR is linked. You can either reopen that one after the PR is merged or open a new issue for follow-up work, as you prefer.

PR #5308, opened 2024-09-23
CI: avoid duplicate runs for secondary branches on main repo

EZoni:
When `pre-commit` gets auto-updated, we typically see a PR from the branch `pre-commit-ci-update-config`. 

This branch is created directly in the main fork (unlike the individual branches that WarpX contributors create from their own forks) and all CI checks run twice:
- once for the activity on the PR that `pre-commit` automatically opens (these CI checks are labeled "PR automated")
- once for the activity on the branch `pre-commit-ci-update-config` (these CI checks are labeled "individual CI")

Here's an example:
Screenshot from 2024-09-23 15-54-57

On top of this, once the PR is merged, CI runs a third time, because the merge is counted as activity on the branch `development` (again "individual CI").

We should be able to safely skip "individual CI" for the activity on the branch `pre-commit-ci-update-config`. This PR should do the trick, although it's good to double check the syntax for GitHub Actions and Azure pipelines.

My understanding is that the cleanup-cache and post-PR workflows don't need to be updated, also to be double checked.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -6,6 +6,10 @@ pool:
 pr:
   autoCancel: true
   drafts: false
+  branches:
+    include:
+    - development
+    - nightly
 
 jobs:
 - job:
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -9,7 +9,6 @@ pr:
   branches:
     include:
     - development
-    - nightly
 
 jobs:
 - job:
--- a/.github/workflows/clang_sanitizers.yml
+++ b/.github/workflows/clang_sanitizers.yml
@@ -1,6 +1,11 @@
 name: 🧴 clang sanitizers
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - "development"
+      - "nightly"
+  pull_request:
 
 concurrency:
   group: ${{ github.ref }}-${{ github.head_ref }}-clangsanitizers
--- a/.github/workflows/clang_sanitizers.yml
+++ b/.github/workflows/clang_sanitizers.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - "development"
-      - "nightly"
   pull_request:
 
 concurrency:
--- a/.github/workflows/clang_tidy.yml
+++ b/.github/workflows/clang_tidy.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - "development"
-      - "nightly"
   pull_request:
 
 concurrency:
--- a/.github/workflows/clang_tidy.yml
+++ b/.github/workflows/clang_tidy.yml
@@ -1,6 +1,11 @@
 name: 🧹 clang-tidy
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - "development"
+      - "nightly"
+  pull_request:
 
 concurrency:
   group: ${{ github.ref }}-${{ github.head_ref }}-clangtidy
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -1,6 +1,11 @@
 name: 🐧 CUDA
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - "development"
+      - "nightly"
+  pull_request:
 
 concurrency:
   group: ${{ github.ref }}-${{ github.head_ref }}-cuda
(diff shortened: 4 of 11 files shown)


.azure-pipelines.yml
```diff
@@ -6,6 +6,10 @@ pool:
 pr:
   autoCancel: true
   drafts: false
+  branches:
+    include:
+    - development
+    - nightly
```
ax3l:
```suggestion
```

.github/workflows/clang_sanitizers.yml
```diff
@@ -1,6 +1,11 @@
 name: 🧴 clang sanitizers
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - "development"
+      - "nightly"
```
ax3l:
```suggestion
```

.github/workflows/clang_tidy.yml
```diff
@@ -1,6 +1,11 @@
 name: 🧹 clang-tidy
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - "development"
+      - "nightly"
```
ax3l:
```suggestion
```

.github/workflows/cuda.yml
```diff
@@ -1,6 +1,11 @@
 name: 🐧 CUDA
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - "development"
+      - "nightly"
```
ax3l:
```suggestion
```

.github/workflows/hip.yml
```diff
@@ -1,6 +1,11 @@
 name: 🐧 HIP
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - "development"
+      - "nightly"
```
ax3l:
```suggestion
```

.github/workflows/insitu.yml
```diff
@@ -1,6 +1,11 @@
 name: 🐧 In Situ Vis
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - "development"
+      - "nightly"
```
ax3l:
```suggestion
```

.github/workflows/intel.yml
```diff
@@ -1,6 +1,11 @@
 name: 🐧 Intel
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - "development"
+      - "nightly"
```
ax3l:
```suggestion
```

.github/workflows/macos.yml
```diff
@@ -1,6 +1,11 @@
 name: 🍏 macOS
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - "development"
+      - "nightly"
```
ax3l:
```suggestion
```

.github/workflows/source.yml
```diff
@@ -6,7 +6,12 @@
 
 name: 📜 Source
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - "development"
+      - "nightly"
```
ax3l:
```suggestion
```

.github/workflows/ubuntu.yml
```diff
@@ -1,6 +1,11 @@
 name: 🐧 OpenMP
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - "development"
+      - "nightly"
```
ax3l:
```suggestion
```

.github/workflows/windows.yml
```diff
@@ -1,6 +1,11 @@
 name: 🪟 Windows
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - "development"
+      - "nightly"
```
ax3l:
```suggestion
```

ax3l (commented):
Thanks! The `nightly` branch actually has its own CI scripts (and tries to install WarpX using various methods we document on various platforms)

ax3l (approved):
Thank you! :sparkles:

EZoni:
> Thanks! The `nightly` branch actually has its own CI scripts (and tries to install WarpX using various methods we document on various platforms)

Ah, I see. Does it have to be handled in a branch separate from `development` or is it just a legacy implementation? I'm asking because I realize that I never look at the `nightly` branch, so I don't perform any maintenance on that branch on my end. I would expect that we should be able to set up and maintain nightly workflows in the `development` branch as well. Would there be arguments against doing this?

PR #5314, opened 2024-09-24
[WIP] Rename current_fp/current_cp to j_fp/j_cp

RemiLehe:
This makes the name shorter and more consistent with the rest of the code.

Note that this will break the Python access API.


EZoni:
Just a thought, as we use upper-case `E` and `B` for the fields, should we use upper-case `J` for the current as well? I think it's typically written as `J`, not `j`, in textbook formulas. Also, in code, `j` is easy to confuse with a count index.

PR #5230, opened 2024-09-09
Add: `MultiFabRegister`

ax3l:
Add a central location to store all `MultiFab`s in. The `ablastr::fields::MultiFabRegister` provides us with a central location to allocate, initialize and exchange fields now.

This is useful to add new fields at runtime (even Python!), access them from anywhere for read & write (incl. Python), alias them from anywhere, redistribute them on load balancing by default (unless you say: no, don't do this, during allocation), and to simplify function calls that access a lot of fields (e.g., solvers just need to pass the `ablastr::fields::MultiFabRegister & WarpX::m_fields` around instead of including `WarpX.H`).

### New Headers

New ABLASTR header:
```C++
#include <ablastr/fields/MultiFabRegister.H>
```

### New Type Aliases

```C++
using ScalarField = amrex::MultiFab *;
using VectorField = std::array<amrex::MultiFab *, 3>;
using MultiLevelScalarField = amrex::Vector<ScalarField>;
```
and const variants.

### New Functions

`MultiFabRegister` provides calls to allocate & initialize (`alloc_init`, `alias_init`), test (`has`), access (`get`, `get_mr_levels`, `get_alldirs`, `get_mr_levels_alldirs`), list (`list`), remove (`erase`, `clear_level`), and remake on AMR/load balance (`remake_level`).
Please see the [Doxygen](https://warpx.readthedocs.io/en/latest/_static/doxyhtml/index.html) page for all details.

These functions can be used both with `warpx::fields::FieldType` enums (preferred) as well as `std::string` (e.g., for purely runtime logic or Python hackery).

### TODO

- [x] make sure all test pass
- [x] write more extended PR description
- [x] `Fields.H`: `warpx::fields::FieldType` add per-enum-entry inline doc string
- [ ] use `throw` logic on `get` with missing keys
- [x] cleanup

PR #5321, opened 2024-09-24
Python: `warpx.multifab` legacy signature

ax3l:
Keep the legacy signature of the fully qualified name a bit longer to avoid breakage. Throw warnings for users to migrate.

Follow-up to #5230

### DIFF
--- a/Docs/source/usage/workflows/python_extend.rst
+++ b/Docs/source/usage/workflows/python_extend.rst
@@ -135,9 +135,9 @@ This example accesses the :math:`E_x(x,y,z)` field at level 0 after every time s
 
        # data access
        #   vector field E, component x, on the fine patch of MR level 0
-       E_x_mf = warpx.multifab(f"Efield_fp", dir=0, level=0)
+       E_x_mf = warpx.multifab("Efield_fp", dir=0, level=0)
        #   scalar field rho, on the fine patch of MR level 0
-       rho_mf = warpx.multifab(f"rho_fp", level=0)
+       rho_mf = warpx.multifab("rho_fp", level=0)
 
        # compute on E_x_mf
        # iterate over mesh-refinement levels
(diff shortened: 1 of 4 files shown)


Docs/source/usage/workflows/python_extend.rst
```diff
@@ -134,9 +134,12 @@ This example accesses the :math:`E_x(x,y,z)` field at level 0 after every time s
        warpx = sim.extension.warpx
 
        # data access
-       E_x_mf = warpx.multifab(f"Efield_fp[x][level=0]")
+       #   vector field E, component x, on the fine patch of MR level 0
+       E_x_mf = warpx.multifab(f"Efield_fp", dir=0, level=0)
```
dpgrote: A quick comment - the `f` prefix is not needed on the strings.

ax3l: True! not sure how this slipped in, maybe a more general example.

RemiLehe (approved):

Source/Python/WarpX.cpp
```diff
@@ -120,17 +120,27 @@ void init_WarpX (py::module& m)
...
-                if (wx.m_fields.has(multifab_name, level)) {
-                    return wx.m_fields.get(multifab_name, level);
+             [](WarpX & wx, std::string internal_name) {
+                 if (wx.m_fields.internal_has(internal_name)) {
+                     return wx.m_fields.internal_get(internal_name);
+                 } else {
+                     throw std::runtime_error("MultiFab '" + internal_name + "' is unknown or is not allocated!");
+                 }
+             },
+             py::arg("internal_name")
```
ax3l, line 130: Forgot a
```suggestion
             py::arg("internal_name"),
             py::return_value_policy::reference_internal
```

ax3l: Fix in #5325

PR #5322, opened 2024-09-24
Remove m_cell_centered_data from multifab map

dpgrote:
The back transformed data diagnostics generates a MultiFab, `m_cell_centered_data`, to hold cell centered data when the output data is being generated. This was being added to the old version of the MultiFab map. Since the data is temporary and since there could be multiple copies of the MultiFab (if there are multiple BTD diagnostics), it doesn't make sense to have the MultiFab in the registry. This PR changes the code so that the MultiFab is directly allocated.

With this PR, the last MultiFab is removed from the old MultiFab map, so it can be removed in a subsequent PR.

Follow-up to #5230

### DIFF
--- a/Source/Diagnostics/BTDiagnostics.H
+++ b/Source/Diagnostics/BTDiagnostics.H
@@ -241,7 +241,7 @@ private:
      *  will be used by all snapshots to obtain lab-frame data at the respective
      *  z slice location.
      */
-    amrex::Vector<std::unique_ptr<amrex::MultiFab> > m_cell_centered_data;
+    std::string const cell_centered_data_name;
     /** Vector of pointers to compute cell-centered data, per level, per component
      *  using the coarsening-ratio provided by the user.
      */
@@ -346,7 +346,7 @@ private:
      *  \param[in] i_buffer snapshot index
      */
     void SetSnapshotFullStatus (int i_buffer);
-    /** Vector of field-data stored in the cell-centered multifab, m_cell_centered_data.
+    /** Vector of field-data stored in the cell-centered multifab, cell_centered_data.
      *  All the fields are stored regardless of the specific fields to plot selected
      *  by the user.
      */
--- a/Source/Diagnostics/BTDiagnostics.H
+++ b/Source/Diagnostics/BTDiagnostics.H
@@ -241,7 +241,7 @@ private:
      *  will be used by all snapshots to obtain lab-frame data at the respective
      *  z slice location.
      */
-    std::string const cell_centered_data_name;
+    std::string const m_cell_centered_data_name;
     /** Vector of pointers to compute cell-centered data, per level, per component
      *  using the coarsening-ratio provided by the user.
      */
@@ -346,7 +346,7 @@ private:
      *  \param[in] i_buffer snapshot index
      */
     void SetSnapshotFullStatus (int i_buffer);
-    /** Vector of field-data stored in the cell-centered multifab, cell_centered_data.
+    /** Vector of field-data stored in the cell-centered MultiFab.
      *  All the fields are stored regardless of the specific fields to plot selected
      *  by the user.
      */
--- a/Source/Diagnostics/BTDiagnostics.cpp
+++ b/Source/Diagnostics/BTDiagnostics.cpp
@@ -57,7 +57,7 @@ namespace
 
 BTDiagnostics::BTDiagnostics (int i, const std::string& name)
     : Diagnostics{i, name},
-      cell_centered_data_name("BTD_cell_centered_data_"+name)
+      m_cell_centered_data_name("BTD_cell_centered_data_" + name)
 {
     ReadParameters();
 }
@@ -521,7 +521,7 @@ BTDiagnostics::DefineCellCenteredMultiFab(int lev)
 #endif
     bool const remake = false;
     bool const redistribute_on_remake = false;
-    warpx.m_fields.alloc_init(cell_centered_data_name, lev, ba, dmap, ncomps, amrex::IntVect(ngrow), 0.0_rt,
+    warpx.m_fields.alloc_init(m_cell_centered_data_name, lev, ba, dmap, ncomps, amrex::IntVect(ngrow), 0.0_rt,
                               remake, redistribute_on_remake);
 
 }
@@ -548,7 +548,7 @@ BTDiagnostics::InitializeFieldFunctors (int lev)
     // to the correct field-data pointers
     m_all_field_functors[lev].clear();
     // For back-transformed data, all the components are cell-centered and stored
-    // in a single multifab, cell_centered_data.
+    // in a single multifab.
     // Therefore, size of functors at all levels is 1.
     const int num_BT_functors = 1;
     m_all_field_functors[lev].resize(num_BT_functors);
@@ -557,11 +557,11 @@ BTDiagnostics::InitializeFieldFunctors (int lev)
     // Create an object of class BackTransformFunctor
     for (int i = 0; i < num_BT_functors; ++i)
     {
-        // coarsening ratio is not provided since the source MultiFab, cell_centered_data
+        // coarsening ratio is not provided since the source MultiFab
         // is coarsened based on the user-defined m_crse_ratio
         const int nvars = static_cast<int>(m_varnames.size());
         m_all_field_functors[lev][i] = std::make_unique<BackTransformFunctor>(
-                  warpx.m_fields.get(cell_centered_data_name, lev), lev,
+                  warpx.m_fields.get(m_cell_centered_data_name, lev), lev,
                   nvars, m_num_buffers, m_varnames, m_varnames_fields);
     }
 
@@ -673,14 +673,14 @@ BTDiagnostics::InitializeFieldFunctorsRZopenPMD (int lev)
     // to the correct field-data pointers
     m_all_field_functors[lev].clear();
     // For back-transformed data, all the components are cell-centered and stored
-    // in a single multifab, cell_centered_data.
+    // in a single MultiFab.
     // Therefore, size of functors at all levels is 1
     const int num_BT_functors = 1;
     m_all_field_functors[lev].resize(num_BT_functors);
     for (int i = 0; i < num_BT_functors; ++i) {
         const int nvars = static_cast<int>(m_varnames.size());
         m_all_field_functors[lev][i] = std::make_unique<BackTransformFunctor>(
-                                       warpx.m_fields.get(cell_centered_data_name, lev), lev,
+                                       warpx.m_fields.get(m_cell_centered_data_name, lev), lev,
                                        nvars, m_num_buffers, m_varnames,
                                        m_varnames_fields);
     }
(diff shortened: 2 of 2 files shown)


Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -519,7 +519,8 @@ BTDiagnostics::DefineCellCenteredMultiFab(int lev)
 #else
     const int ncomps = static_cast<int>(m_cellcenter_varnames.size());
 #endif
-    WarpX::AllocInitMultiFab(m_cell_centered_data[lev], ba, dmap, ncomps, amrex::IntVect(ngrow), lev, "cellcentered_BTD", 0._rt);
+    m_cell_centered_data[lev] = std::make_unique<amrex::MultiFab>(ba, dmap, ncomps, amrex::IntVect(ngrow));
```
ax3l: Why not use  `MultiFabRegister::alloc_init` and `::erase`?

ax3l: So far, we deallocated `m_cell_centered_data[lev]` when the `BTDiagnostics` is deallocated. We do not want to change that behavior here in terms of how long the lifetime goes.

Deallocation of each `BTDiagnostics` variable currently probably (to verify with @RevathiJambunathan) happens only at the end of the simulation. If that is true, we do not need even need to call `::erase` when a station is done. If there is an earlier deallocation of `BTDiagnostics` (proably should be) or more logic in `~BTDiagnostics` or when a station is fully done, then we should call `::erase` at those points.

For the name we put in the MultiFabRegister, we should create a unique name by combining:
- a clear BTD prefix and
- the current BTD diagnostics name (from the inputs) and
- the station it works on (usually called `i_buffer`, if I am not mistaken).

This will be nice, because we can now more easily track the overhead of BTD diags.

dpgrote: The lifetime of the diagnostics is the same as the lifetime of the WarpX class so the erase is not needed.

Source/Diagnostics/BTDiagnostics.H
```diff
@@ -241,7 +241,7 @@ private:
      *  will be used by all snapshots to obtain lab-frame data at the respective
      *  z slice location.
      */
-    amrex::Vector<std::unique_ptr<amrex::MultiFab> > m_cell_centered_data;
+    std::string const cell_centered_data_name;
```
ax3l:
```suggestion
    std::string const m_cell_centered_data_name;
```

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -56,7 +56,8 @@ namespace
 }
 
 BTDiagnostics::BTDiagnostics (int i, const std::string& name)
-    : Diagnostics{i, name}
+    : Diagnostics{i, name},
+      cell_centered_data_name("BTD_cell_centered_data_"+name)
```
ax3l:
```suggestion
      m_cell_centered_data_name("BTD_cell_centered_data_" + name)
```

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -519,7 +519,10 @@ BTDiagnostics::DefineCellCenteredMultiFab(int lev)
 #else
     const int ncomps = static_cast<int>(m_cellcenter_varnames.size());
 #endif
-    WarpX::AllocInitMultiFab(m_cell_centered_data[lev], ba, dmap, ncomps, amrex::IntVect(ngrow), lev, "cellcentered_BTD", 0._rt);
+    bool const remake = false;
+    bool const redistribute_on_remake = false;
+    warpx.m_fields.alloc_init(cell_centered_data_name, lev, ba, dmap, ncomps, amrex::IntVect(ngrow), 0.0_rt,
```
ax3l:
```suggestion
    warpx.m_fields.alloc_init(m_cell_centered_data_name, lev, ba, dmap, ncomps, amrex::IntVect(ngrow), 0.0_rt,
```

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -554,11 +557,11 @@ BTDiagnostics::InitializeFieldFunctors (int lev)
     // Create an object of class BackTransformFunctor
     for (int i = 0; i < num_BT_functors; ++i)
     {
-        // coarsening ratio is not provided since the source MultiFab, m_cell_centered_data
+        // coarsening ratio is not provided since the source MultiFab, cell_centered_data
         // is coarsened based on the user-defined m_crse_ratio
         const int nvars = static_cast<int>(m_varnames.size());
         m_all_field_functors[lev][i] = std::make_unique<BackTransformFunctor>(
-                  m_cell_centered_data[lev].get(), lev,
+                  warpx.m_fields.get(cell_centered_data_name, lev), lev,
```
ax3l:
```suggestion
                  warpx.m_fields.get(m_cell_centered_data_name, lev), lev,
```

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -554,11 +557,11 @@ BTDiagnostics::InitializeFieldFunctors (int lev)
     // Create an object of class BackTransformFunctor
     for (int i = 0; i < num_BT_functors; ++i)
     {
-        // coarsening ratio is not provided since the source MultiFab, m_cell_centered_data
+        // coarsening ratio is not provided since the source MultiFab, cell_centered_data
```
ax3l:
```suggestion
        // coarsening ratio is not provided since the source MultiFab
```

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -545,7 +548,7 @@ BTDiagnostics::InitializeFieldFunctors (int lev)
     // to the correct field-data pointers
     m_all_field_functors[lev].clear();
     // For back-transformed data, all the components are cell-centered and stored
-    // in a single multifab, m_cell_centered_data.
+    // in a single multifab, cell_centered_data.
```
ax3l:
```suggestion
    // in a single multifab.
```

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -670,14 +673,14 @@ BTDiagnostics::InitializeFieldFunctorsRZopenPMD (int lev)
...
```
ax3l:
```suggestion
    // in a single MultiFab.
```

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -670,14 +673,14 @@ BTDiagnostics::InitializeFieldFunctorsRZopenPMD (int lev)
...
     // Therefore, size of functors at all levels is 1
     const int num_BT_functors = 1;
     m_all_field_functors[lev].resize(num_BT_functors);
     for (int i = 0; i < num_BT_functors; ++i) {
         const int nvars = static_cast<int>(m_varnames.size());
         m_all_field_functors[lev][i] = std::make_unique<BackTransformFunctor>(
-                                       m_cell_centered_data[lev].get(), lev,
+                                       warpx.m_fields.get(cell_centered_data_name, lev), lev,
```
ax3l:
```suggestion
                                       warpx.m_fields.get(m_cell_centered_data_name, lev), lev,
```

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -804,21 +807,23 @@ BTDiagnostics::PrepareFieldDataForOutput ()
         for (int icomp = 0; icomp<m_cell_center_fuctors_at_lev_size; ++icomp) {
             // Call all the cell-center functors in m_cell_center_functors.
             // Each of them computes cell-centered data for a field and
-            // stores it in cell-centered MultiFab, m_cell_centered_data[lev].
-            m_cell_center_functors[lev][icomp]->operator()(*m_cell_centered_data[lev], icomp_dst);
+            // stores it in cell-centered MultiFab, cell_centered_data[lev].
+            m_cell_center_functors[lev][icomp]->operator()(*warpx.m_fields.get(cell_centered_data_name, lev), icomp_dst);
```
ax3l:
```suggestion
            // stores it in cell-centered MultiFab.
            m_cell_center_functors[lev][icomp]->operator()(*warpx.m_fields.get(m_cell_centered_data_name, lev), icomp_dst);
```

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -804,21 +807,23 @@ BTDiagnostics::PrepareFieldDataForOutput ()
...
             icomp_dst += m_cell_center_functors[lev][icomp]->nComp();
         }
         // Check that the proper number of user-requested components are cell-centered
         AMREX_ALWAYS_ASSERT( icomp_dst == m_cellcenter_varnames.size() );
         // fill boundary call is required to average_down (flatten) data to
         // the coarsest level.
-        ablastr::utils::communication::FillBoundary(*m_cell_centered_data[lev], WarpX::do_single_precision_comms,
+        ablastr::utils::communication::FillBoundary(*warpx.m_fields.get(cell_centered_data_name, lev),
```
ax3l:
```suggestion
        ablastr::utils::communication::FillBoundary(*warpx.m_fields.get(m_cell_centered_data_name, lev),
```

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -804,21 +807,23 @@ BTDiagnostics::PrepareFieldDataForOutput ()
...
+                                                    WarpX::do_single_precision_comms,
                                                     warpx.Geom(lev).periodicity());
     }
     // Flattening out MF over levels
 
     for (int lev = warpx.finestLevel(); lev > 0; --lev) {
-        ablastr::coarsen::sample::Coarsen(*m_cell_centered_data[lev - 1], *m_cell_centered_data[lev], 0, 0,
+        ablastr::coarsen::sample::Coarsen(*warpx.m_fields.get(cell_centered_data_name, lev - 1),
+                                          *warpx.m_fields.get(cell_centered_data_name, lev), 0, 0,
```
ax3l:
```suggestion
        ablastr::coarsen::sample::Coarsen(*warpx.m_fields.get(m_cell_centered_data_name, lev - 1),
                                          *warpx.m_fields.get(m_cell_centered_data_name, lev), 0, 0,
```

Source/Diagnostics/BTDiagnostics.H
```diff
@@ -346,7 +346,7 @@ private:
      *  \param[in] i_buffer snapshot index
      */
     void SetSnapshotFullStatus (int i_buffer);
-    /** Vector of field-data stored in the cell-centered multifab, m_cell_centered_data.
+    /** Vector of field-data stored in the cell-centered multifab, cell_centered_data.
```
ax3l:
```suggestion
    /** Vector of field-data stored in the cell-centered MultiFab.
```

ax3l (approved):
Thank you! :)

PR #5324, opened 2024-09-25
Divide rho by epsilon to undo the multiplication before computing phi

RevathiJambunathan:
Just before phi is computed, rho is divided by -epsilon 
but it is not divided back and this in python callbacks, the `RhoFPWrapper` points to rho/(-epsilon)
which is mis-leading and can lead to errors

### DIFF
--- a/Source/ablastr/fields/PoissonSolver.H
+++ b/Source/ablastr/fields/PoissonSolver.H
@@ -275,7 +275,7 @@ computePhi (
 #endif
         // Use the Multigrid (MLMG) solver if selected or on refined patches
         // but first scale rho appropriately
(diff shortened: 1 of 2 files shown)


Source/ablastr/fields/PoissonSolver.H
```diff
@@ -409,6 +409,8 @@ computePhi (
                 post_phi_calculation.value()(mlmg, lev);
             }
         }
+        rho[lev]->mult(-ablastr::constant::SI::ep0);  // Divide rho by epsilon again
```
RemiLehe:
```suggestion
        rho[lev]->mult(-ablastr::constant::SI::ep0);  // Multiply rho by epsilon again
```

RemiLehe (approved):

roelof-groenewald (approved):
Thank you!!

PR #5325, opened 2024-09-25
Fix Legacy Python MF API

ax3l:
Keep ownership on the C++ side.
Forgot to copy-paste the rest of the signature.

Follow-up to #5321


Source/Python/WarpX.cpp
```diff
@@ -127,7 +127,15 @@ void init_WarpX (py::module& m)
                      throw std::runtime_error("MultiFab '" + internal_name + "' is unknown or is not allocated!");
                  }
              },
-             py::arg("internal_name")
+             py::arg("internal_name"),
+             py::return_value_policy::reference_internal,
```
ax3l, line 131: This is the bug fix

RemiLehe (approved):

PR #5326, opened 2024-09-25
Python: Warn old `warpx.multifab` Signature

ax3l:
Warn users that use the old `warpx.multifab("internal_name")` overload to use the new one that only requests a prefix, with dir and level as extra arguments.

Follow-up to #5321


dpgrote (approved):
This looks good, thanks!

EZoni (approved):

PR #5327, opened 2024-09-25
PoissonSolver: Missing Include

ax3l:
Fix the missing include for `MultiLevelRegister`, which includes the MultiLevel MF types.

Seen as an issue in ImpactX.

Follow-up to #5230


ax3l:
Merging to open a PR in ImpactX that has an issue with the current `development` FFT solvers.

PR #5329, opened 2024-09-25
[WIP] Hybrid PIC External Magnetic Field

kli-jfp:
Feature Description: External Magnetic Field in Hybrid Solver

A static curl-free external magnetic field has been integrated into the Hybrid solver, analogous to the external fields applied to particles in the fully kinetic case.

Key Features:

1. Analytical Function Parser for J and B: Parses analytical functions for both current density (J) and magnetic field (B). Issues a warning if the magnetic field is time-dependent (this is future work).
2. File Input for J and B: Allows reading current density (J) and magnetic field (B) from a file. Read-from-file does not support time-dependent fields.

CI test (ohm_solver_guiding_field): This is a copy of the magnetic reconnection test, except that the guiding field has been taken away from the initial magnetic field and set as an external field.


kli-jfp:
Have to fix CI ohm solver tests that have broken. If you have an idea of what is wrong let me know. I also don't see the new CI guiding field test in the azure pipeline.

EZoni:
I have not looked into why some CI tests are failing in this PR yet, but to answer this 

> I also don't see the new CI guiding field test in the azure pipeline.

I think you need to add the line
```cmake
add_subdirectory(ohm_solver_guiding_field)
```

in https://github.com/ECP-WarpX/WarpX/blob/development/Examples/Tests/CMakeLists.txt, because `Examples/Tests/ohm_solver_guiding_field` is a new test directory that did not exist before.

We recently added this to the docs: https://warpx.readthedocs.io/en/latest/developers/testing.html#how-to-add-automated-tests.

This should also allow you to run the test locally through CTest, which might make it easier to debug.

Docs/source/usage/parameters.rst
```diff
@@ -2447,9 +2447,24 @@ Maxwell solver: kinetic-fluid hybrid
 * ``hybrid_pic_model.plasma_hyper_resistivity`` (`float` or `str`) optional (default ``0``)
     If ``algo.maxwell_solver`` is set to ``hybrid``, this sets the plasma hyper-resistivity in :math:`\Omega m^3`.
 
+* ``hybrid_pic_model.[J/B]_ext_grid_init_style`` (`str`) optional (default ``0``)
+    If ``algo.maxwell_solver`` is set to ``hybrid``, this sets the external field style:math:`\Omega m^3`.
```
```suggestion
    If ``algo.maxwell_solver`` is set to ``hybrid``, this sets the external field style.
```

Docs/source/usage/parameters.rst
```diff
@@ -2447,9 +2447,24 @@ Maxwell solver: kinetic-fluid hybrid
...
+
+    * ``read_from_file`` (`str`) optional (default ``0``)
+      If ``algo.maxwell_solver`` is set to ``hybrid``, this sets wether to read from file.
```
```suggestion
      If ``algo.maxwell_solver`` is set to ``hybrid``, this sets whether to read from file.
```

Docs/source/usage/parameters.rst
```diff
@@ -2447,9 +2447,24 @@ Maxwell solver: kinetic-fluid hybrid
...
+
+    * ``parse_ext_grid_function`` (`str`) optional (default ``0``)
+      If ``algo.maxwell_solver`` is set to ``hybrid``, this sets wether to use analytical function.
```
```suggestion
      If ``algo.maxwell_solver`` is set to ``hybrid``, this sets whether to use analytical function.
```

Examples/Tests/ohm_solver_guiding_field/README.rst
```diff
@@ -0,0 +1,49 @@
+.. _examples-ohm-solver-guiding_field:
+
+Ohm Solver: Guiding Field (Magnetic Reconnection)
+=================================
```
roelof-groenewald: The heading line should be the same length as the text for the documentation to render properly.
```suggestion
Ohm Solver: Guiding Field (Magnetic Reconnection)
=======================================
```

Examples/Tests/ohm_solver_guiding_field/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+
+add_warpx_test(
+    test_2d_ohm_solver_guiding_field_picmi  # name
+    2  # dims
+    2  # nprocs
+    "inputs_test_2d_ohm_solver_guiding_field_picmi.py --test"  # inputs
+    analysis.py  # analysis
+    diags/diag1000020  # output
+    OFF  # dependency
+)
```
roelof-groenewald: It would make sense to put this test in the same folder as the existing magnetic reconnection test, then the exact same analysis script can be used (and even the same checksum files). You could even add the new functionality to the existing test and just control which test is run through an additional runtime argument (like the `--test` argument).

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.H
```diff
@@ -184,7 +190,23 @@ public:
...
+    std::array< amrex::ParserExecutor<4>, 3> m_B_external;
+    bool m_external_bfield_has_time_dependence = false;
+
+    // Path of the file where external
+    // fields are stored
+    std::string m_external_j_fields_path = "0.0";
+    std::string m_external_b_fields_path = "0.0";
+
+    std::string m_J_ext_grid_s = "0.0";
+    std::string m_B_ext_grid_s = "0.0";
```
roelof-groenewald: Being a bit more verbose here would help to make the code more readable.
```suggestion
    std::string m_J_ext_grid_style = "0.0";
    std::string m_B_ext_grid_style = "0.0";
```

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -53,6 +53,25 @@ void HybridPICModel::ReadParameters ()
     pp_hybrid.query("Jx_external_grid_function(x,y,z,t)", m_Jx_ext_grid_function);
     pp_hybrid.query("Jy_external_grid_function(x,y,z,t)", m_Jy_ext_grid_function);
     pp_hybrid.query("Jz_external_grid_function(x,y,z,t)", m_Jz_ext_grid_function);
+
+    pp_hybrid.query("J_external_init_style", m_J_ext_grid_s);
+
+    if (m_J_ext_grid_s == "read_from_file"){
+            const std::string read_j_fields_from_path="./";
```
roelof-groenewald: Why is this variable defined?
```suggestion
```

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -53,6 +53,25 @@ void HybridPICModel::ReadParameters ()
...
+
+    // external magnetic field
+    pp_hybrid.query("Bx_external_grid_function(x,y,z,t)", m_Bx_ext_grid_function);
+    pp_hybrid.query("By_external_grid_function(x,y,z,t)", m_By_ext_grid_function);
+    pp_hybrid.query("Bz_external_grid_function(x,y,z,t)", m_Bz_ext_grid_function);
+
+    pp_hybrid.query("B_external_init_style", m_B_ext_grid_s);
+
+    if (m_B_ext_grid_s == "read_from_file"){
+            const std::string read_b_fields_from_path="./";
```
roelof-groenewald: Similar as above.
```suggestion
```

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -53,6 +53,25 @@ void HybridPICModel::ReadParameters ()
     pp_hybrid.query("Jx_external_grid_function(x,y,z,t)", m_Jx_ext_grid_function);
     pp_hybrid.query("Jy_external_grid_function(x,y,z,t)", m_Jy_ext_grid_function);
     pp_hybrid.query("Jz_external_grid_function(x,y,z,t)", m_Jz_ext_grid_function);
+
+    pp_hybrid.query("J_external_init_style", m_J_ext_grid_s);
+
+    if (m_J_ext_grid_s == "read_from_file"){
+            const std::string read_j_fields_from_path="./";
+            pp_hybrid.query("read_j_fields_from_path", m_external_j_fields_path);
```
roelof-groenewald: I would suggest to make this:
```suggestion
            pp_hybrid.get("read_j_fields_from_path", m_external_j_fields_path);
```
that way an exception will be raised if the file path is not provided.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -238,24 +291,62 @@ void HybridPICModel::InitData ()
...
+    for (int lev = 0; lev <= warpx.finestLevel(); ++lev) {
+        GetExternalField<FieldType::hybrid_current_fp_external>(m_J_external, edge_lengths[lev], lev);
     }
 }
 
-
-void HybridPICModel::GetCurrentExternal (
+template <FieldType T>
+void HybridPICModel::GetExternalField (
+    std::array< amrex::ParserExecutor<4>, 3> const& F_external,
```
roelof-groenewald, line 345: This is a nice generalization! With it we can move all of this repeated functionality out of the hybrid-PIC class and instead replace `WarpX::InitializeExternalFieldsOnGridUsingParser` with this more general function and then just call it directly there. Would you mind making that change?

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -269,41 +360,46 @@ void HybridPICModel::GetCurrentExternal (
...
+#endif
 
         amrex::ParallelFor (tbx, tby, tbz,
             [=] AMREX_GPU_DEVICE (int i, int j, int k) {
                 // skip if node is covered by an embedded boundary
-                if (lx && lx(i, j, k) <= 0) { return; }
-
+#ifdef AMREX_USE_EB
+                if (lx(i, j, k) <= 0) return;
+#endif
```
roelof-groenewald: These should stay as
```suggestion
                if (lx && lx(i, j, k) <= 0) { return; }
```
since we are slowly moving towards not having the `AMREX_USE_EB` directives.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -269,41 +360,46 @@ void HybridPICModel::GetCurrentExternal (
...
+#ifdef AMREX_USE_EB
+       amrex::Array4<amrex::Real> const& lx = edge_lengths[0]->array(mfi);
+       amrex::Array4<amrex::Real> const& ly = edge_lengths[1]->array(mfi);
+       amrex::Array4<amrex::Real> const& lz = edge_lengths[2]->array(mfi);
+#if defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
+        amrex::ignore_unused(ly);
+#endif
+#else
+       amrex::ignore_unused(edge_lengths);
+#endif
```
roelof-groenewald: Please replace this with
```suggestion
        amrex::Array4<amrex::Real> lx, ly, lz;
        if (EB::enabled()) {
            lx = edge_lengths[0]->array(mfi);
            ly = edge_lengths[1]->array(mfi);
            lz = edge_lengths[2]->array(mfi);
        }
```
as we are trying to get rid of the compile time option for EB support.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -468,12 +566,13 @@ void HybridPICModel::HybridPICSolveE (
...
     ablastr::fields::VectorField current_fp_ampere = warpx.m_fields.get_alldirs(FieldType::hybrid_current_fp_ampere, lev);
     const ablastr::fields::VectorField current_fp_external = warpx.m_fields.get_alldirs(FieldType::hybrid_current_fp_external, lev);
+    const ablastr::fields::VectorField bfield_fp_external = warpx.m_fields.get_alldirs(FieldType::hybrid_bfield_fp_external, lev);
     const ablastr::fields::ScalarField electron_pressure_fp = warpx.m_fields.get(FieldType::hybrid_electron_pressure_fp, lev);
 
     // Solve E field in regular cells
     warpx.get_pointer_fdtd_solver_fp(lev)->HybridPICSolveE(
         Efield, current_fp_ampere, Jfield, current_fp_external,
-        Bfield, rhofield,
+        Bfield, bfield_fp_external, rhofield,
```
roelof-groenewald, line 563: The idea with creating the `m_fields` container was that functions like this would now be able to just take `m_fields` as an argument and extract themselves all the fields they need. We don't have to make this change as part of this PR but if you were so inclined you could switch this instead to just take `m_fields` and then grab all the needed fields when they are actually used.

roelof-groenewald (commented):
Thanks for this PR! I've left a couple of comments / suggestions below.

I also want to make sure I understand the use case for which this is needed. My reading of the code changes is that it adds functionality for the user to specify an additional B-field which is included in both the Lorentz force applied to the particles and in the calculation of the E-field from Ohm's law, but not in the calculation of the total current from `mu_0 J = curl x B`. Am I missing anything in that?
If not, assuming the additionally applied B-field has zero curl, it wouldn't make any difference to directly apply it to the initial B-field, right? And if it has non-zero curl, the same result can be obtained by setting `J_ext = curl x B_ext / mu0`, right? Considering only the affected term in Ohm's law, the above basically says:

$enE = (\nabla\times B_{sim}/\mu_0 - J_i)\times(B_{sim} + B_{ext}) =  (\nabla\times(B_{sim} + B_{ext})/\mu_0 - J_i - J_{ext})\times(B_{sim} + B_{ext}) $

if $ \nabla\times B_{ext} / \mu_0 = J_{ext} $.

I think it would make more sense to instead create functionality that will calculate J_ext given a B_ext and keep the simulation B-field as the full B? Does that make sense? What do you think? In that case we would allow the user to either specify B_ext **or** J_ext (but not both) and if B_ext is specified the code would calculate J_ext during initialization. This would then make it so that we don't need to add the auxiliary B-field.

Btw, I'm hoping to implement finite electron mass terms in the not too distant future in which case we would add a term $dJ_ext / dt$ which would easily capture inductive effects. Implementing that from a time-varying external B-field would be more trouble (I think).

kli-jfp:
> I think you need to add the line
> 
> ```cmake
> add_subdirectory(ohm_solver_guiding_field)
> ```
Thanks [EZoni](https://github.com/EZoni) for the tip. I can see it now.

kli-jfp:
> I also want to make sure I understand the use case for which this is needed. My reading of the code changes is that it adds functionality for the user to specify an additional B-field which is included in both the Lorentz force applied to the particles and in the calculation of the E-field from Ohm's law, but not in the calculation of the total current from `mu_0 J = curl x B`. Am I missing anything in that?

Yes, that's correct [roelof-groenewald](https://github.com/roelof-groenewald)! This feature assumes a static curl-free B field, which is specifically interesting for magnetically confined fusion where your coils are outside of the vacuum vessel. This allows you to only simulate the plasma region of the device with the PIC algorithm.

I'm concerned that including B_ext in the initial condition would not give the same results as explicitly setting it as a static field throughout the simulation. If I set a high beta plasma in a mirror, and then the plasma escapes, would the resulting total field be B_sim = B_ext? I haven't really understood this issue properly yet :) My current understanding is that if you put the external field as initial conditions then the external field can change, unless the coils and current is included to make sure that a static field is being sustained throughout the simulation.

kli-jfp:
> Btw, I'm hoping to implement finite electron mass terms in the not too distant future in which case we would add a term 
... which would easily capture inductive effects. Implementing that from a time-varying external B-field would be more trouble (I think).

That's great! I think a time-depdent B_ext can be included if Faradays law is changed to
dB_sim/dt = -curl E - dB_ext/dt

kli-jfp:
[roelof-groenewald](https://github.com/roelof-groenewald) I'm trying to figure out why the ohm_solver CTests are failing on azure pipeline with error:

0::1::Assertion `dst.boxArray() == src.boxArray()' failed, file "/home/vsts/work/1/s/build/_deps/fetchedamrex-src/Src/Base/AMReX_MultiFab.cpp", line 161 !!!
SIGABRT
Assertion `dst.boxArray() == src.boxArray()' failed, file "/home/vsts/work/1/s/build/_deps/fetchedamrex-src/Src/Base/AMReX_MultiFab.cpp", line 161 !!!
SIGABRT

I'm not seeing this locally. Perhaps it is how I'm initializing the Bfield_aux in WarpX.cpp. Perhaps you can clarify: is the Bfield_fp in the hybrid solver staggered? If I set B_ext to nodal, then I'm effictively adding a staggered array with a nodal grid. Could this be why the error occurs? If this is the error I'm making then B_ext field should be created on a staggered grid and then interpolated in HybridPICSolveE?

clarkse:
I have a PR working toward time dependent split fields in the Hybrid solver here: https://github.com/ECP-WarpX/WarpX/pull/5275/ It is getting closer to prime time and I think a gridded version with an associated time function could be added fairly easily. Additionally AMReX now can compute complete elliptic functions on GPU for doing analytical current loops and sheets, etc. We should sync up on how we want to proceed with adding this functionality.

roelof-groenewald:
> I'm concerned that including B_ext in the initial condition would not give the same results as explicitly setting it as a static field throughout the simulation. If I set a high beta plasma in a mirror, and then the plasma escapes, would the resulting total field be B_sim = B_ext?

Yes, only the part of the B-field that comes from plasma current will decay. That is to say that you can consider the system to have a total B-field $B = B_{plasma} + B_{external}$ where $B_{plasma}$ is the B-field due to the current given by $J_{plasma} = \nabla\times B / \mu_0 - J_{ext}$ and only $B_{plasma}$ will change during the simulation. This requires, though, that $J_{ext} = \nabla\times B_{external} / \mu_0$. If you consider the coils that generate the external field to be outside of your domain that is totally fine - you just have $J_{ext} = 0$, and that external field still would not change as the simulation advances (this is because in that case $\nabla\times B_{external} = 0$). This is why I suggest we change this PR to be able to calculate $J_{ext} = \nabla\times B_{external} / \mu_0$ given an external field, instead of applying the external field as a separate field - the two approaches are mathematically equivalent but, I think, the former is cleaner.

kli-jfp:
> Yes, only the part of the B-field that comes from plasma current will decay. That is to say that you can consider the system to have a total B-field B = B p l a s m a + B e x t e r n a l where B p l a s m a is the B-field due to the current given by J p l a s m a = ∇ × B / μ 0 − J e x t and only B p l a s m a will change during the simulation. This requires, though, that J e x t = ∇ × B e x t e r n a l / μ 0 . If you consider the coils that generate the external field to be outside of your domain that is totally fine - you just have J e x t = 0 , and that external field still would not change as the simulation advances (this is because in that case ∇ × B e x t e r n a l = 0 ). This is why I suggest we change this PR to be able to calculate J e x t = ∇ × B e x t e r n a l / μ 0 given an external field, instead of applying the external field as a separate field - the two approaches are mathematically equivalent but, I think, the former is cleaner.

If this is true then I 100% agree that B_ext should be in the initial conditions and J_ext should be computed as you said. I will see if I can verify this (if you have any reference let me know I'm curious). If a substantial increase in spatial and temporal resolution is required in order for the B_ext to be static throughout the simulation in that case then perhaps it is still worth to include a split field implementation as a user-defined feature to decrease the computional cost.

I noticed that the fully-kinetic case has a split field implementation. Would the external particle fields be uneccessary there as well? 

P.s. I tried implementing J_ext from external field etc but I failed to get a stable simulation going. A lot of numerical noise occurs around the coils. I will continue to look into this.

kli-jfp:
> I have a PR working toward time dependent split fields in the Hybrid solver here: #5275 It is getting closer to prime time and I think a gridded version with an associated time function could be added fairly easily. Additionally AMReX now can compute complete elliptic functions on GPU for doing analytical current loops and sheets, etc. We should sync up on how we want to proceed with adding this functionality.

That's great! What are your thoughts on using split fields vs using the total field as initial conditions (and using J_ext)?

clarkse:
> > I have a PR working toward time dependent split fields in the Hybrid solver here: #5275 It is getting closer to prime time and I think a gridded version with an associated time function could be added fairly easily. Additionally AMReX now can compute complete elliptic functions on GPU for doing analytical current loops and sheets, etc. We should sync up on how we want to proceed with adding this functionality.
> 
> That's great! What are your thoughts on using split fields vs using the total field as initial conditions (and using J_ext)?

I kind of had to go the route of the split field for a time varying external field. This is due to the improper propagation of the EM wave through the vacuum/low density region in the hybrid code. There are ways to handle this via other methods described in Lipatov's monograph, e.g. magnetoinductive model evolving A with appropriate vacuum operators to handle time varying BCs, or potentially using a solver with electron inertia or displacement currents. With the split field and a slowly varying external field the assumption of instantaneous propagation works out, then the plasma responds to it via the Ohm's law. I think for a static case the split field vs. non-split should be exactly the same. Since the Ohm's law is only valid for the total E and B that the plasma sees, split vs. not for a static case should yield the exact same solution.

Where the split static external field may be useful is when you have a PEC boundary, but you have a B field that is static and has soaked through the metal. I.e. only the fast varying plasma dynamics obey the PEC boundary, which includes effective induced mirror currents and charges. You can do this when the time scale of this simulation is much faster than the timescale of B field diffusion through your metal. Then the external magnets can be present and the internal dynamics obey the BCs.

I currently plan to augment my time varying fields to automatically compute the induced E field instead of having to input it manually, but needs a bit more development work and should be ready by next week.

Would you mind if I cherry pick some of your changes for adding the OpenPMD loader into my PR? I will add an additional input field for a time varying analytical function that could be coupled with a loaded file. This will be useful for scrubbing B field divergence from the input, so I will probably use that route. I think if it accepts a list of external fields and a list of corresponding time values that a series of external fields with different time variations could be added that way.

After this gets pushed through I am planning to add time varying solenoids to the Accelerator Lattice and integrate with the split field handling.

PR #5334, opened 2024-09-26
`MultiFabRegister`: use `has_vector` when possible

### LINKED ISSUE #5305: `MultiFabRegister`: replace `has` with `has_vector` for vectors
Originally posted in https://github.com/ECP-WarpX/WarpX/pull/5230#discussion_r1772068317.

As a follow-up on #5230, we should replace `has` with `has_vector` wherever possible, so in most occurrences where `has` is called to check if a vector field exists in the register, unless we intentionally want to check only a single specific direction.

EZoni:
Follow-up on #5230: use `has_vector` for vector fields when possible, instead of `has` on a single direction.

This PR may fix #5305.

### DIFF
--- a/Source/BoundaryConditions/PML.cpp
+++ b/Source/BoundaryConditions/PML.cpp
@@ -1234,7 +1234,7 @@ PML::CheckPoint (
 {
     using ablastr::fields::Direction;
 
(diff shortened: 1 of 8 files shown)


ax3l (approved):
Awesome, thanks!

PR #5340, opened 2024-09-27
HPC: `SRUN_CPUS_PER_TASK` to `SLURM_`

ax3l:
This environment variable was used for Perlmutter when `--cpus-per-task=N` did not work yet. It was copied around to other templates.

These days, `--cpus-per-task` should work and the name of the env variable was renamed in SLURM to  `SLURM_CPUS_PER_TASK`. https://slurm.schedmd.com/sbatch.html#OPT_SLURM_CPUS_PER_TASK

Thanks to NERSC engineers for reporting this update!


kngott (approved):
Checked the SLURM_CPUS_PER_TASK variable is successfully set when -c is set in the batch flags. It does, so I think this should work. 😃

EZoni (approved):

EZoni:
Merging this as it was approved by @kngott two weeks ago and hasn't received further comments since.

PR #5337, opened 2024-09-26
Add WarpX example for FEL simulation

RemiLehe:
This adds an example for how to run FEL simulations with the boosted-frame technique.
https://warpx--5337.org.readthedocs.build/en/5337/usage/examples/free_electron_laser/README.html

PR #5342, opened 2024-09-29
Avoid interpolating from guard cells in BTD

RemiLehe:
BTD diagnostics sometimes show artifacts at the edge of the range of collected data. (See for instance the red curve below.)

My understanding is that this happens because the BTD collection planes use data from the guard cells outside of the simulation domain, when interpolating fields that are then used for the Lorentz back-transform. The guard cell data may not be physically correct (e.g. it may not have the right cancellation between `E` and `B`), and could thus cause this artifact.

This PR avoids this issue by prevents the collection planes to collect data when it is half a cell from the edge of the simulation domain.

See the example below, taken from https://github.com/ECP-WarpX/WarpX/pull/5337 (plot of the laser field, from the BTD diagnostic)
Figure 40
The BTD diagnostics values are identical with this PR, except for the problematic point appearing at the edge of the domain.


RevathiJambunathan (approved):
Looks great! 
Thank you for the PR!

PR #5349, opened 2024-10-01
General parser function for external fields

kli-jfp:
This feature replaces the InitializeExternalFieldsOnGridUsingParser with a more general function ComputeExternalFieldOnGridUsingParser. The new function takes parsed functions with four dimensions (x,y,z,t), so that it can be used to evaluate functions throughout the simulation, for example time-dependent external fields.

The function ComputeExternalFieldOnGridUsingParser has optional edge length and face areas.

### DIFF
--- a/Source/Initialization/WarpXInitData.cpp
+++ b/Source/Initialization/WarpXInitData.cpp
@@ -1528,7 +1528,7 @@ WarpX::LoadExternalFields (int const lev)
             m_p_ext_field_params->Ezfield_parser->compile<4>(),
             m_fields.get_alldirs(FieldType::edge_lengths, lev),
             m_fields.get_alldirs(FieldType::face_areas, lev),
-            'f', lev, PatchType::fine);
+            'e', lev, PatchType::fine);
     }
     else if (m_p_ext_field_params->E_ext_grid_type == ExternalFieldType::read_from_file) {
 #if defined(WARPX_DIM_RZ)
--- a/Source/Initialization/WarpXInitData.cpp
+++ b/Source/Initialization/WarpXInitData.cpp
@@ -979,7 +979,7 @@ WarpX::InitLevelData (int lev, Real /*time*/)
             m_p_ext_field_params->Bzfield_parser->compile<4>(),
             m_fields.get_alldirs(FieldType::edge_lengths, lev),
             m_fields.get_alldirs(FieldType::face_areas, lev),
-            "face", lev, PatchType::fine);
+            'f', lev, PatchType::fine);
 
         ComputeExternalFieldOnGridUsingParser(
             FieldType::Bfield_cp,
@@ -988,7 +988,7 @@ WarpX::InitLevelData (int lev, Real /*time*/)
             m_p_ext_field_params->Bzfield_parser->compile<4>(),
             m_fields.get_alldirs(FieldType::edge_lengths, lev),
             m_fields.get_mr_levels_alldirs(FieldType::face_areas, max_level)[lev],
-            "face", lev, PatchType::coarse);
+            'f', lev, PatchType::coarse);
     }
 
     // if the input string for the E-field is "parse_e_ext_grid_function",
@@ -1021,7 +1021,7 @@ WarpX::InitLevelData (int lev, Real /*time*/)
                 m_p_ext_field_params->Ezfield_parser->compile<4>(),
                 m_fields.get_alldirs(FieldType::edge_lengths, lev),
                 m_fields.get_alldirs(FieldType::face_areas, lev),
-                "edge", lev, PatchType::fine);
+                'e', lev, PatchType::fine);
 
             ComputeExternalFieldOnGridUsingParser(
                 FieldType::Efield_cp,
@@ -1030,7 +1030,7 @@ WarpX::InitLevelData (int lev, Real /*time*/)
                 m_p_ext_field_params->Ezfield_parser->compile<4>(),
                 m_fields.get_alldirs(FieldType::edge_lengths, lev),
                 m_fields.get_alldirs(FieldType::face_areas, lev),
-                "edge", lev, PatchType::coarse);
+                'e', lev, PatchType::coarse);
 #ifdef AMREX_USE_EB
             if (eb_enabled) {
                 if (WarpX::electromagnetic_solver_id == ElectromagneticSolverAlgo::ECT) {
@@ -1180,7 +1180,7 @@ void WarpX::ComputeExternalFieldOnGridUsingParser (
     amrex::ParserExecutor<4> const& fz_parser,
     ablastr::fields::VectorField const& edge_lengths,
     ablastr::fields::VectorField const& face_areas,
-    [[maybe_unused]] std::string topology,
+    [[maybe_unused]] const char topology,
     int lev, PatchType patch_type)
 {
     auto t = gett_new(lev);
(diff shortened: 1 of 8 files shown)


roelof-groenewald (commented):
Thanks a lot for this PR! It will help to make things more maintainable when all grid parsing is done through the same function.

One question I have, do you think it would be possible to reduce the amount of code duplication in the two overloads of `ComputeExternalFieldOnGridUsingParser`? One suggestion would be to make the `edge_lengths` and `face_areas` arguments `std::optional` with default values of `std::nullopt`. That way if they are not given the `lx`, `Sx`, ... variables could just remain uninitialized and things like `if(lx && ...)` would return `false`.

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1144,7 +1145,7 @@ WarpX::InitializeExternalFieldsOnGridUsingParser (
                 const amrex::Real x = 0._rt;
                 const amrex::Real y = 0._rt;
                 const amrex::Real fac_z = (1._rt - x_nodal_flag[0]) * dx_lev[0] * 0.5_rt;
-                const amrex::Real z = i*dx_lev[0] + real_box.lo(0) + fac_z;
+                const amrex::Real z = j*dx_lev[0] + real_box.lo(0) + fac_z;
```
roelof-groenewald: Is this change (`i` -> `j`) on purpose? I think it should be `i` since in 1d only the first index is used.

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1160,27 +1161,27 @@ WarpX::InitializeExternalFieldsOnGridUsingParser (
...
-                   (field=='B' and Sy(i,j,k)<=0))) { return; }
+                   (topology=='f' and Sy(i,j,k)<=0))) { return; }
 #endif
 #endif
 #if defined(WARPX_DIM_1D_Z)
                 const amrex::Real x = 0._rt;
                 const amrex::Real y = 0._rt;
                 const amrex::Real fac_z = (1._rt - y_nodal_flag[0]) * dx_lev[0] * 0.5_rt;
-                const amrex::Real z = i*dx_lev[0] + real_box.lo(0) + fac_z;
+                const amrex::Real z = j*dx_lev[0] + real_box.lo(0) + fac_z;
```
roelof-groenewald: Same comment as above.

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1196,22 +1197,22 @@ WarpX::InitializeExternalFieldsOnGridUsingParser (
...
-                if(lz && ((field=='E' and lz(i, j, k)<=0) or (field=='B' and lx(i, j, k)<=0))) { return; }
+                if(lz && ((topology=='e' and lz(i, j, k)<=0) or (topology=='f' and lx(i, j, k)<=0))) { return; }
 #endif
 #endif
 #if defined(WARPX_DIM_1D_Z)
                 const amrex::Real x = 0._rt;
                 const amrex::Real y = 0._rt;
                 const amrex::Real fac_z = (1._rt - z_nodal_flag[0]) * dx_lev[0] * 0.5_rt;
-                const amrex::Real z = i*dx_lev[0] + real_box.lo(0) + fac_z;
+                const amrex::Real z = j*dx_lev[0] + real_box.lo(0) + fac_z;
```
roelof-groenewald: And here as well.

roelof-groenewald (commented):
Looks good to me. I just have a question about the index change for the z-direction in 1d. See below.

Source/Initialization/WarpXInitData.cpp
kli-jfp: Good catch! That's a typo, I'll fix.

kli-jfp:
I'm struggling to get clang-tidy build to compile. Working on it.

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1072,48 +1059,61 @@ WarpX::InitLevelData (int lev, Real /*time*/)
...
 
     auto dx_lev = geom[lev].CellSizeArray();
-    amrex::IntVect refratio = (lev > 0 ) ? WarpX::RefRatio(lev-1) : amrex::IntVect(1);
+    const RealBox& real_box = geom[lev].ProbDomain();
+
+    amrex::IntVect refratio = (lev > 0 ) ? RefRatio(lev-1) : amrex::IntVect(1);
     if (patch_type == PatchType::coarse) {
         for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
-            dx_lev[idim] = dx_lev[idim] * refratio[idim];
+            dx_lev[idim] = dx_lev[idim] + refratio[idim];
```
roelof-groenewald: I'm guessing this is why the CI test failed:
```suggestion
            dx_lev[idim] = dx_lev[idim] * refratio[idim];
```

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1072,48 +1059,61 @@ WarpX::InitLevelData (int lev, Real /*time*/)
...
         amrex::Array4<amrex::Real> lx, ly, lz, Sx, Sy, Sz;
         if (eb_enabled) {
-            lx = edge_lengths[0]->array(mfi);
-            ly = edge_lengths[1]->array(mfi);
-            lz = edge_lengths[2]->array(mfi);
-            Sx = face_areas[0]->array(mfi);
-            Sy = face_areas[1]->array(mfi);
-            Sz = face_areas[2]->array(mfi);
+            if (edge_lengths.has_value()) {
+                auto& edge_lengths_array = edge_lengths.value();
```
roelof-groenewald: `clang-tidy` wants this to be
```suggestion
                const auto& edge_lengths_array = edge_lengths.value();
```

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1072,48 +1059,61 @@ WarpX::InitLevelData (int lev, Real /*time*/)
...
+                lx = edge_lengths_array[0]->array(mfi);
+                ly = edge_lengths_array[1]->array(mfi);
+                lz = edge_lengths_array[2]->array(mfi);
+            }
+            if (face_areas.has_value()) {
+                auto& face_areas_array = face_areas.value();
```
roelof-groenewald: And same here
```suggestion
                const auto& face_areas_array = face_areas.value();
```

roelof-groenewald (commented):
The only remaining `clang-tidy` issues I see are these ones.

Source/Initialization/WarpXInitData.cpp
kli-jfp: Yea typo from trying to fix clang-tidy problems but I haven't gotten it to compile yet. Maybe I'm missing it but I can't see the errors from the azure pipeline.

roelof-groenewald: Yeah the clang-tidy errors are quite hard to spot since they get buried in lots of other output. My trick is to search for the carrot symbol (`^`) in the output since usually the errors have that symbol.

EZoni: Keep in mind that our `clang-tidy` builds run through GitHub Actions, not Azure. If you scroll down to the bottom of the main page of this PR, where all CI checks are listed, you can click on "Details" next to a given `clang-tidy` check (for example, the workflow labeled `clang-tidy / clang-tidy-1D (pull_request)`) and you should be redirected to a page where you can see the output of all the steps of that particular `clang-tidy` workflow, which in turn will display build errors, if any. Do you manage to access the output logs like this?

kli-jfp: [EZoni](https://github.com/EZoni) You are right thanks, I missed that it was GitHub Actions. I have access to it so hopefully things will go much smoother.

roelof-groenewald (approved):
Looks good! Thanks for the contribution!

PR #5351, opened 2024-10-01
Remove multifab_map and associated functions

dpgrote:
Now that all MultiFabs are registered in the new system, the old `multifab_map` can be removed.

Note that the `imultifab_map` is still needed since the new system does not yet work with `iMultiFabs`.

This breaks the Python API since it removes `warpx.field` which was returning the old map.


roelof-groenewald (approved):
Looks good to me!

roelof-groenewald:
I don't think this one really needs to wait to be merged with the other breaking API changes as this is effectively already removed in `development` since all the multifabs have been moved.

lucafedeli88 (approved):
Thanks, @dpgrote !

PR #5352, opened 2024-10-02
[WIP] CI tests: use ubuntu 24.04 for clang tidy and clang sanitizer tests

lucafedeli88:
In this PR I am updating the OS for the clang-tidy and clang sanitizer tests to Ubuntu 24.04, since [it is no longer in beta for github hosted runners](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners). The PR also bumps the version of `clang` used for sanitizer tests from 17 to 18.

The sanitizer test was already using `ubuntu:23.10` via a container, so some changes consists in removing lines that were only necessary to run inside a container.

For the clang-tidy test the final goal is also to bump the version of `clang-tidy` from 15 to 18, but this will be done in a separate PR (it will require to update the clang-tidy configuration file).


EZoni:
I think we did some of this in #5474 when new problems came up, although some is still missing (the clang version update).

As you might have noticed here too, new potential issues came up, which led to #5492. 

Just cross-referencing here to make sure that we share info and rebase when needed.

lucafedeli88:
This PR can be closed. Thanks to https://github.com/ECP-WarpX/WarpX/pull/5575 it is no longer necessary.

PR #5353, opened 2024-10-02
Move isAnyBoundaryPML to Warpx.cpp

lucafedeli88:
`isAnyBoundaryPML` is used only inside `WarpX.cpp`. It does not need to be a member function of the WarpX class and it can be moved into an anonymous namespace inside `WarpX.cpp`

### DIFF
--- a/Source/WarpX.H
+++ b/Source/WarpX.H
@@ -735,7 +735,7 @@ public:
     void DampJPML (int lev, PatchType patch_type);
 
     void CopyJPML ();
-    bool isAnyBoundaryPML();
+
     /** True if any of the particle boundary condition type is Thermal */
     static bool isAnyParticleBoundaryThermal();
 
--- a/Source/WarpX.cpp
+++ b/Source/WarpX.cpp
@@ -196,6 +196,22 @@ amrex::IntVect m_rho_nodal_flag;
 
 WarpX* WarpX::m_instance = nullptr;
 
+namespace
+{
+
+    [[nodiscard]] bool
+    isAnyBoundaryPML(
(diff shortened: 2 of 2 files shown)


EZoni:
Thanks, Luca. What is the difference between having the new function in an anonymous namespace as opposed to having it as a stand-alone function?

lucafedeli88:
Hi @EZoni ,
I think that the main advantage is simplifying the interface of the WarpX class. `isAnyBoundaryPML` is a helper function used only inside `WarpX.ccp`. It is therefore more like an implementation detail of the WarpX class rather than something that needs to be exposed.
One function would not make a difference, but gradually reducing the size of the header of the WarpX class could improve performances at compilation time, since this header is included almost everywhere.

EZoni:
Yes, I think that is in line with what we had discussed in the past. I didn't express myself too well, what I was asking is the following, is having an anonymous namespace to encapsulate the function the only way to take the function out of the class and have it as a free function or are there alternative ways (which I approximately labeled "stand-alone function" in my previous comment) to achieve that? I can't remember if some alternatives had been discussed in the past.

lucafedeli88:
Oh, I see... sorry, I misinterpreted your question!

The main advantage of using anonymous namespaces in the cpp file with respect to just using a simple free function is that  functions inside anonymous namespaces can be used only in the translation unit where the function is defined. This protects against accidental conflicts with functions with the same name declared  in the included header files. Declaring the free function as static in the cpp file would achieve the same effect if we don't want to use anonymous namespaces.

EZoni (approved):

PR #5355, opened 2024-10-02
Expose Synchronize to picmi

budjensen:
Added the ability to synchronize particle velocities and positions inside a python script via simulation.extension.warx.synchronize().

Initiated after discussion with @roelof-groenewald in https://github.com/ECP-WarpX/WarpX/discussions/5331.

Implemented and tested by comparing to Turner's benchmark results (https://doi.org/10.1063/1.4775084).

<p align="center">


roelof-groenewald (approved):
Thanks for adding this!

PR #5356, opened 2024-10-02
MultiFabRegister: `throw` in get

### LINKED ISSUE #5319: Throw error in `MultiFabRegister.get` when a field does not exist
We should uncomment this line:
https://github.com/ECP-WarpX/WarpX/blob/development/Source/ablastr/fields/MultiFabRegister.cpp#L346

ax3l:
Close #5319
Follow-up to #5230

- [x] Throw a runtime exception instead of returning a `nullptr` if a field is requested via the getter.
- [x] update logic to ensure this passes all tests


EZoni:
Discussed with @RemiLehe offline:

The functions `get_mr_levels` and `get_mr_levels_alldirs` fails _by definition_ when we call it on coarse-patch MultiFabs, e.g., `current_cp`, because those are allocated only on levels greater than level `0` while the function attempts to return all levels starting from `0`.

We will try to add to `get_mr_levels` and `get_mr_levels_alldirs` a function argument, e.g., `skip_level_0`, such that the function returns a `nullptr` for level `0` instead of calling the internal functions of the MultiFabRegister and throwing an exception.

Source/ablastr/fields/MultiFabRegister.cpp
```diff
@@ -350,8 +350,7 @@ namespace ablastr::fields
     )
     {
         if (m_mf_register.count(internal_name) == 0) {
-            // FIXME: temporary, throw a std::runtime_error
-            // throw std::runtime_error("MultiFabRegister::get name does not exist in register: " + key);
+            throw std::runtime_error("MultiFabRegister::get name does not exist in register: " + internal_name);
             return nullptr;
```
EZoni: I think we can remove this:
```suggestion
            throw std::runtime_error("MultiFabRegister::get name does not exist in register: " + internal_name);
```

Source/ablastr/fields/MultiFabRegister.cpp
```diff
@@ -365,8 +364,7 @@ namespace ablastr::fields
     ) const
     {
         if (m_mf_register.count(internal_name) == 0) {
-            // FIXME: temporary, throw a std::runtime_error
-            // throw std::runtime_error("MultiFabRegister::get name does not exist in register: " + internal_name);
+            throw std::runtime_error("MultiFabRegister::get name does not exist in register: " + internal_name);
             return nullptr;
```
EZoni: I think we can remove this:
```suggestion
            throw std::runtime_error("MultiFabRegister::get name does not exist in register: " + internal_name);
```

EZoni (changes requested):

Source/WarpX.cpp
```diff
@@ -2593,6 +2593,7 @@ WarpX::AllocLevelMFs (int lev, const BoxArray& ba, const DistributionMapping& dm
         }
 
         // Create the MultiFabs for the current
+        amrex::Print() << "allocate current_cp" << std::endl;
```
EZoni: Debugging info left over by mistake:
```suggestion
```

EZoni:
@ax3l @RemiLehe 

Just a heads-up, I rebased and pushed a few more fixes and this might be almost ready. Let's see the latest CI status after the last commit and move from there.

EZoni:
This is the only CI failure left after [fb0dc3c](https://github.com/ECP-WarpX/WarpX/pull/5356/commits/fb0dc3c83c62b5c799cf8891ce51bcaeb2a18186):
```
The following tests FAILED:
	417 - test_3d_effective_potential_electrostatic_picmi.run (Failed)
	418 - test_3d_effective_potential_electrostatic_picmi.analysis (Failed)
	419 - test_3d_effective_potential_electrostatic_picmi.checksum (Failed)
```
I pushed [57e8e22](https://github.com/ECP-WarpX/WarpX/pull/5356/commits/57e8e22e3ebbb5dee3f51ff2c36b2eb3c94390bb) to try to fix the test.

EZoni:
@ax3l @lucafedeli88 @RemiLehe 

CI is green now and the PR is ready for review.

Source/BoundaryConditions/PML.cpp
```diff
@@ -1298,16 +1298,16 @@ PML::PushPSATD (ablastr::fields::MultiFabRegister& fields, const int lev)
 {
     ablastr::fields::VectorField pml_E_fp = fields.get_alldirs(FieldType::pml_E_fp, lev);
     ablastr::fields::VectorField pml_B_fp = fields.get_alldirs(FieldType::pml_B_fp, lev);
-    ablastr::fields::ScalarField pml_F_fp = fields.get(FieldType::pml_F_fp, lev);
-    ablastr::fields::ScalarField pml_G_fp = fields.get(FieldType::pml_G_fp, lev);
+    ablastr::fields::ScalarField pml_F_fp = (fields.has(FieldType::pml_F_fp, lev)) ? fields.get(FieldType::pml_F_fp, lev) : nullptr;
```
ax3l, line 1301: hehe, that's one way to quickly get the behavior in without changing downstream :D okok, nice idea and we can make this more robust later.

Source/ablastr/fields/MultiFabRegister.H
```diff
@@ -479,24 +479,28 @@ namespace ablastr::fields
         [[nodiscard]] MultiLevelScalarField
         get_mr_levels (
             T name,
-            int finest_level
+            int finest_level,
+            bool skip_level_0=false
```
ax3l, line 484: We need to add the new param `skip_level_0` in `@param` of the doxgen string.

Source/ablastr/fields/MultiFabRegister.H
```diff
@@ -550,24 +554,28 @@ namespace ablastr::fields
         [[nodiscard]] MultiLevelVectorField
         get_mr_levels_alldirs (
             T name,
-            int finest_level
+            int finest_level,
+            bool skip_level_0=false
```
ax3l, line 560: We need to add the new param `skip_level_0` in `@param` of the doxgen string.

Source/ablastr/fields/MultiFabRegister.cpp
```diff
@@ -419,29 +415,45 @@ namespace ablastr::fields
...
     )
     {
         MultiLevelScalarField field_on_level;
         field_on_level.reserve(finest_level+1);
         for (int lvl = 0; lvl <= finest_level; lvl++)
         {
-            field_on_level.push_back(internal_get(name, lvl));
+            if (lvl == 0 && skip_level_0)
+            {
+                field_on_level.push_back(nullptr);
```
ax3l, line 428: Is pushing back a `nullptr` the clean approach or not pushing back anything (`skip` might imply either).
Then on the other hand, I assume if we move the level indexing on `skip` that can cause a lot of confusion, too (one argument to not use a `Vector` for the list of pointers.)

Thus, ok for me to keep as you suggest for now.

ax3l (commented):
Excellent! Just a small Doxygen update needed and a few ideas for later improvements.

Source/ablastr/fields/MultiFabRegister.H
ax3l: done.

Source/ablastr/fields/MultiFabRegister.H
ax3l: done.

EZoni (approved):
@ax3l 

Thanks for reviewing my changes.

Yes, in some corner cases (e.g., F in the PML) I just pushed the issue to the next "layer". Those were changes that date back to last year, I might have been unsure about how to resolve those at the time.

I'll approve on my end and let @roelof-groenewald or @lucafedeli88 chime in further.

I would make sure that people have enough time to look at this PR and double check the changes, since they are not trivial.

roelof-groenewald (commented):
The changes are fine, but I wonder if we can rename the `skip_level_0` variable to something else, maybe `skip_level_0_for_coarse_patch`. The reason I say that is if someone reads the code in the future, seeing just `skip_level_0 = true`, in something like `ComputeSpaceChargeField()` it could seem like the potential is not calculated for level 0 at all. Does that make sense?

EZoni:
Thanks, @roelof-groenewald. I renamed the variable to `skip_lev0_coarse_patch`, just to shorten the name as much as possible without missing any crucial component.

ax3l:
I think renaming in the usage "usage" code constants to `skip_lev0_coarse_patch` made sense.

In the `MultiFabRegister` class it does not make sense to me: this flag works for any MF (any patch, fine or coarse), if you pass it level 0 is skipped. I would thus suggest to keep `skip_level_0` as the argument name in `MultiFabRegister.H`/`.cpp` if that makes sense @EZoni @roelof-groenewald 

(Alternatively, do not add the new arg to the API at all and instead in user-code zero-out the 0-level coarse patch component on your `skip_lev0_coarse_patch` variable or even better, pass `skip_lev0_coarse_patch` to the call that should not process the level to be explicit. I am not sure how universally useful this argument addition in `MultiFabRegister` truly is, looks a bit hacky to me.)

EZoni:
Thanks, @ax3l.

My bad, I went through the renaming too fast. I agree that it would make sense to use two different names.

As for this,

> (Alternatively, do not add the new arg to the API at all and instead in user-code zero-out the 0-level coarse patch component on your `skip_lev0_coarse_patch` variable or even better, pass `skip_lev0_coarse_patch` to the call that should not process the level to be explicit. I am not sure how universally useful this argument addition in MultiFabRegister truly is, looks a bit hacky to me.)

I think I don't understand the suggestion. The reason for introducing the variable was that we have two methods, `get_mr_levels` and `get_mr_levels_alldirs`, that loop over all MR levels, and therefore we cannot call those methods on any coarse-patch MultiFab, e.g., `rho_cp`, because nothing is allocated for those MultiFabs on level 0. I think what I don't understand in the suggestion is what data we would "zero out": my understanding is that there would be no data to zero out because the underlying allocation is missing, since coarse-patch data is not allocated on level 0. I think this was @RemiLehe's and my understanding when we discussed what is summarized in https://github.com/ECP-WarpX/WarpX/pull/5356#issuecomment-2423349551 and came up with the `skip_level_0` workaround.

ax3l:
Perfect.

Per last suggestion: oh yes, you are absolutely right that would not work 😂 ok, yes since this applies actually to all our cp, let's have it in the API (with the name back to the original).

PR #5358, opened 2024-10-02
Update ALCF Polaris profile

WeiqunZhang:
* Explicitly use gcc-12, because gcc points to gcc/7.5.
* Load the non-cray version of hdf5 because the cray version has some cmake issues.


ax3l (approved):
Thank you, @WeiqunZhang!

Thanks for the report @ppiot !

ppiot:
@WeiqunZhang  thank you. The code compiles fine but does not run when the job is submitted. The nodes seems to not have the non-cray hdf5 library module accessible. I submitted a ticket to request help from ALCF. I will update when things are resolved...

PR #5359, opened 2024-10-02
 Fix: newVersion.sh Python

ax3l:
The script was not robust enough for the black/ruff formatting updates we applied. This affected the 24.09 and 24.10 release tags.

This fixes it.


EZoni (approved):
Thanks for the fix. I wonder if it could be possible to replace these bash scripts with Python scripts. I think it could make development and maintenance easier. I might give it a try whenever I have some time.

PR #5361, opened 2024-10-02
Python 3.13 Support, 3.8 EOL

ax3l:
Add support for Python 3.13.
Remove 3.8 because it is EOL as of Oct 2024.

Bump to pybind11 2.13.0+, which add Python 3.13 support in CI.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -307,7 +307,7 @@ setup(
     cmdclass=cmdclass,
     # scripts=['warpx_1d', 'warpx_2d', 'warpx_rz', 'warpx_3d'],
     zip_safe=False,
-    python_requires=">=3.8",
+    python_requires=">=3.8",  # left for CI, truly ">=3.9"
     # tests_require=['pytest'],
     install_requires=install_requires,
     # see: src/bindings/python/cli
@@ -336,10 +336,11 @@ setup(
         "Topic :: Scientific/Engineering :: Physics",
         "Programming Language :: C++",
         "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
         "Programming Language :: Python :: 3.11",
+        "Programming Language :: Python :: 3.12",
+        "Programming Language :: Python :: 3.13",
         (
             "License :: OSI Approved :: " "BSD License"
         ),  # TODO: use real SPDX: BSD-3-Clause-LBNL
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -301,7 +301,7 @@ def build_extension(self, ext):
     cmdclass=cmdclass,
     # scripts=['warpx_1d', 'warpx_2d', 'warpx_rz', 'warpx_3d'],
     zip_safe=False,
-    python_requires=">=3.8",
+    python_requires=">=3.8",  # left for CI, truly ">=3.9"
```
roelof-groenewald, line 310: Does this comment mean the CI uses Python 3.8?

roelof-groenewald (approved):
Looks good to me!

setup.py
ax3l: Yes, a few CI runners still use Python 3.8 and for me that is ok for now.
We can bump this once we update CI runners for 3.9+.

EZoni:
I did a quick grep of `python-version` in our workflows and this is what I see:
```sh
$ git grep "python-version" .github/
.github/workflows/cuda.yml:30:        python-version: '3.x'
.github/workflows/windows.yml:24:        python-version: '3.x'
.github/workflows/windows.yml:75:        python-version: '3.8'
```

I think `3.x` gets the "latest minor release", if I understand correctly what is explained here: https://docs.github.com/en/actions/use-cases-and-examples/building-and-testing/building-and-testing-python#using-a-specific-python-version.

So maybe the only job that uses Python 3.8 is "Clang C++17 w/ OMP w/o MPI" in the "Windows" workflow:
https://github.com/ECP-WarpX/WarpX/blob/d0690e23402c28d624791ab033b5f0d3be289dbd/.github/workflows/windows.yml#L67-L119

Do we want to try to upgrade this Python version directly in this PR, if Python 3.8 is EOL as of October 2024?

ax3l:
There are a few more that use Ubuntu 20.04 without `setup-python` at the moment.

The way to update these would be to bump to 22.04 or, where we intentionally test an old compiler to ensure we can still use it (e.g., old NVCC that we specify as our minimum requirement), use `setup-python`. I can push this here so we have it all together.

ax3l:
Ah yes, and Azure runs on Ubuntu 20.04 and checksums are too strict: #5203
I am not aware of a `setup-python` trick for Azure pipelines.

ax3l:
Simplified as in pyAMReX for now (no hard enforcement of Python 3.9+ during build time until we have all CI updated).

PR #5363, opened 2024-10-03
(WIP) Added code to use joined array 

guj:
The  joined array will be applied if the input diag entry openpmd_backend=bp4 or bp5

TODO:
1. double check that const var support in openpmd is relaxed
2. check file based encoding
3. check BTD/RZ cases


RemiLehe:
superseded by PR #5804

PR #5364, opened 2024-10-03
Disable AMReX_LINEAR_SOLVER_INCFLO by default

WeiqunZhang:
We do not need to compile these linear solvers that are intended for incompressible flow solvers. This should speed up the build process a little bit.

Depends on https://github.com/AMReX-Codes/amrex/pull/4181


WeiqunZhang:
I have tested with the latest CUDA, HIP and SYCL compilers. The compilers seem to have improved a lot. It used to be quite slow to compile the EB tensor and nodal projection solvers. But now I only saw a few percent difference when those solvers were not included.

ax3l (approved):
Thank you! :rocket: :sparkles:

cmake/dependencies/AMReX.cmake
```diff
@@ -92,6 +92,7 @@ macro(find_amrex)
         set(AMReX_PARTICLES ON CACHE INTERNAL "")
         set(AMReX_PROBINIT OFF CACHE INTERNAL "")
         set(AMReX_TINY_PROFILE ON CACHE BOOL "")
+        set(AMReX_LINEAR_SOLVER_INCFLO OFF CACHE INTERNAL "")
```
ax3l, line 96: Explicit on might be best:
```suggestion
        set(AMReX_LINEAR_SOLVERS_EM ON CACHE INTERNAL "")
        set(AMReX_LINEAR_SOLVER_INCFLO OFF CACHE INTERNAL "")
```

GNUmakefile
```diff
@@ -43,5 +43,7 @@ USE_RZ = FALSE
 
 USE_EB = FALSE
 
+USE_LINEAR_SOLVERS_INCFLO = FALSE
```
ax3l, line 47:
```suggestion
USE_LINEAR_SOLVERS_EM = TRUE
USE_LINEAR_SOLVERS_INCFLO = FALSE
```

ax3l (approved):

PR #5365, opened 2024-10-04
Doc: Catalyst Verbatim Formatting

ax3l:
Fix some RST `verbatim`, which needs two backticks compared to Markdown.

@ChristosT @c-wetterer-nelson Somewhat related, I do not yet understand how I could do a Catalyst [Live Visualization](https://docs.paraview.org/en/v5.13.1/Catalyst/getting_started.html#live-visualization). Do I need to dump to disk or can we transport the data over the network into ParaView (e.g., via ADIOS2 SST or openPMD-api using ADIOS2 SST)?

### DIFF
--- a/Docs/source/dataanalysis/catalyst.rst
+++ b/Docs/source/dataanalysis/catalyst.rst
@@ -19,7 +19,7 @@ First, we build
 build `Catalyst 2 <https://catalyst-in-situ.readthedocs.io/en/latest/build_and_install.html>`_
 using the conduit library created in the previous step.
 The latter can be achieved by adding the installation path of conduit to the environmental
-variable `CMAKE_PREFIX_PATH` and setting `CATALYST_WITH_EXTERNAL_CONDUIT=ON` during the configuration step of Catalyst.
+variable ``CMAKE_PREFIX_PATH`` and setting ``CATALYST_WITH_EXTERNAL_CONDUIT=ON`` during the configuration step of Catalyst.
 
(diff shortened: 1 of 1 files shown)


lucafedeli88 (approved):

c-wetterer-nelson:
> Fix some RST `verbatim`, which needs two backticks compared to Markdown.
> 
> @ChristosT @c-wetterer-nelson Somewhat related, I do not yet understand how I could do a Catalyst [Live Visualization](https://docs.paraview.org/en/v5.13.1/Catalyst/getting_started.html#live-visualization). Do I need to dump to disk or can we transport the data over the network into ParaView (e.g., via ADIOS2 SST or openPMD-api using ADIOS2 SST)?

Hi Axel, It's a bit of a curve to get setup to do a Live visualization session the first go around. You will need to enable Live when you export your script from ParaView, this will add some info on ports. Then you need to start a separate pvserver and if your ports are configured correctly, your Catalyst enabled simulation will send that data over to the pvserver. Then you can login with a ParaView client and connect to that pvserver. If you're targeting a remote compute resource, that can be challenging if we don't already have a provided connection script for your platform (I think Perlmutter does have one).

c-wetterer-nelson:
There is also an ADIOS route uses SST to send data between your simulation and an ADIOS consumer that sends data to a pvserver. I haven't used that paradigm myself. https://www.kitware.com/catalyst-adios2-a-new-catalyst2-implementation-for-in-transit-analysis/

ChristosT (approved):

EZoni (approved):

PR #5375, opened 2024-10-08
openPMD-api: 0.16.1+

ax3l:
Update to the latest stable release of openPMD-api, version 0.16.1.


EZoni (approved):

ax3l:
The clang-tidy run for 1D passes in 163min and the other three run into a timeout at 180min.

PR #5384, opened 2024-10-10
SYCL: 1D EB Compile

ax3l:
Attempt to fix 1D SYCL EB compile errors (throw not allowed on device).

X-ref: https://github.com/spack/spack/pull/46765#issuecomment-2403937237


WeiqunZhang (approved):

ax3l:
Yay, it looks like this works on Spack E4S CI.

PR #5385, opened 2024-10-10
Fix CI: CodeQL Setup

ax3l:
Fix broken Python setup in CodeQL CI.


ax3l:
Merging to unbreak CI.

PR #5386, opened 2024-10-10
Docs: Fix reStructuredText syntax for figures

EZoni:
While reviewing some of our documentation PRs, I noticed that we use the `:alt:` (alternate text) syntax in a way that might be neither optimal nor consistent with the `reStructuredText` [figure examples](https://docutils.sourceforge.io/docs/ref/rst/directives.html#figure):

```rst
.. figure:: picture.png
   :scale: 50 %
   :alt: map to buried treasure

   This is the caption of the figure (a simple paragraph).
```

As a reminder, the `reStructuredText` [image docs](https://docutils.sourceforge.io/docs/ref/rst/directives.html#image) describes alternate text as

> Alternate text: a short description of the image, displayed by applications that cannot display images, or spoken by applications for visually impaired users

_Before_ this PR, this is the output when an image is displayed:
Screenshot from 2024-10-10 15-16-25

_Before_ this PR, this is the output when an image is _not_ displayed (e.g., due to a typo in the docs):
Screenshot from 2024-10-10 15-20-20

_After_ this PR, I would propose to have this output when an image is _not_ displayed (e.g., due to a typo in the docs):
Screenshot from 2024-10-10 15-34-00

I think this is a better output for both users that read the docs (avoid reading the caption twice) and visually impaired users (avoid hearing the caption twice - the real caption is spoken too).

Note that in some cases we do not yet provide a caption. I believe we should add a caption to all figures, but I left this out of this PR.

PR #5387, opened 2024-10-10
CI: skip build/test jobs if only `Docs/` is modified

EZoni:
Could we do this to make sure that we run the GitHub Actions and Azure jobs (build, test) only if _at least one file outside the_ `Docs` _directory_ is modified, i.e., skip those jobs if only files in the `Docs` directory are modified?

I think it would be safe to do so (and a bit of a waste of resources to not do so...), but I leave it open for discussion.

If merged, we could test this rebasing #5386 and seeing if the correct CI jobs are skipped.

Note that this PR leaves the other CI jobs untouched, e.g., `source`, `docs`, `CodeQL`, etc.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -9,6 +9,9 @@ pr:
   branches:
     include:
     - development
+  paths:
+    exclude:
+    - Docs
 
 jobs:
 - job:
--- a/.github/workflows/clang_sanitizers.yml
+++ b/.github/workflows/clang_sanitizers.yml
@@ -5,6 +5,8 @@ on:
     branches:
       - "development"
   pull_request:
+    paths-ignore:
+      - "Docs/**"
 
 concurrency:
   group: ${{ github.ref }}-${{ github.head_ref }}-clangsanitizers
--- a/.github/workflows/clang_tidy.yml
+++ b/.github/workflows/clang_tidy.yml
@@ -5,6 +5,8 @@ on:
     branches:
       - "development"
   pull_request:
+    paths-ignore:
+      - "Docs/**"
 
 concurrency:
   group: ${{ github.ref }}-${{ github.head_ref }}-clangtidy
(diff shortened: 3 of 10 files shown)


WeiqunZhang (approved):
Good idea!

lucafedeli88 (approved):
Excellent idea! I fully agree

ax3l (approved):
Nice!! :star_struck:

EZoni:
Note that this is handled by GitHub Actions in a way that is not optimal.

From their docs (https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks):

> Warning: If a workflow is skipped due to [path filtering](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore), [branch filtering](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore) or a [commit message](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs), then checks associated with that workflow will remain in a "Pending" state. A pull request that requires those checks to be successful will be blocked from merging.

The GitHub Actions workflows that we skip will appear as "pending". This should be fine for now, as maintainers should be able to force-merge the PR anyways. Hopefully GitHub will fix this in the future, marking the workflows that are skipped as "skipped" rather than "pending". I can think of other work-arounds in the meantime.

EZoni:
@lucafedeli88 @WeiqunZhang 

I think what I reported in the comment above, https://github.com/ECP-WarpX/WarpX/pull/5387#issuecomment-2412459815, might be more pressing than we think.

If you look at a docs PR, e.g., #5403, I personally do not have a way to merge it now, because GitHub (unlike Azure) is flagging the skipped workflows as "pending" and, since they are required, at least I cannot merge. I think other maintainers can probably merge, I definitely remember people merging PRs with failing or pending CI, but I cannot.

WeiqunZhang:
@EZoni Instead of `path-ignore`, we could try this https://github.com/dorny/paths-filter?tab=readme-ov-file

PR #5388, opened 2024-10-11
[WIP] Add support for electron impact ionization in binary collisions (DSMC module)

RemiLehe:
Note: could we do binary charge exchange in the same way?

TODO:

- [ ] Use the proper energy cost: store the energy cost in the `SplitAndScatter` functor 
- [ ] Add a dedicated test for DSMC ionization
- [ ] Add support for excitation in DSMC (right now, it prevents us from running the capacitive discharge)
- [ ] Test on GPU
- [ ] Check physics, esp. assumptions on the velocity of the indident electron after scattering.


Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
```diff
@@ -52,29 +52,54 @@ DSMCFunc::DSMCFunc (
         WARPX_ALWAYS_ASSERT_WITH_MESSAGE(process.type() != ScatteringProcessType::INVALID,
                                         "Cannot add an unknown scattering process type");
 
-        m_scattering_processes.push_back(std::move(process));
-    }
+        // if the scattering process is ionization get the secondary species
+        // only one ionization process is supported, the vector
+        // m_ionization_processes is only used to make it simple to calculate
+        // the maximum collision frequency with the same function used for
+        // particle conserving processes
```
RemiLehe, line 59: Hm, I copied this from `BackgroundMCC.ccp` but it makes no sense here since we are *not* calculating the maximum collision frequency...

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.cpp
```diff
@@ -15,17 +15,37 @@ SplitAndScatterFunc::SplitAndScatterFunc (const std::string& collision_name,
...
-        // here we can add logic to deal with cases where products are created,
-        // for example with impact ionization
-        m_num_product_species = 2;
-        m_num_products_host.push_back(1);
-        m_num_products_host.push_back(1);
+        if (ionization_flag) {
+            // Product species include the ion
+            // TODO: as an alternative, we could use the runtime attribute `ionization_level` for this species
+            m_num_product_species = 3;
+            m_num_products_host.push_back(2); // electron species:
```
RemiLehe, line 33: I attempted to fit impact ionization within the DSMC framework, but I now realize that this is making the book keeping complicated:

The amount of memory to be allocated (in `SplitAndScatter`) after the first pass (`Executor`, that checks which type of reaction is occurring for each pair) will depend on the type of reaction for each pair. (Because, if a pair does impact ionization, we need to allocate space for the scattered electron, the produced electron and the ionized species, whereas if a pair does one of the other reactions, we allocate space only for the scattered electron and the scattered target.) As a consequence, the offset (in `SplitAndScatter` will depend on the type of reaction.)

To simplify things, here I will always allocate space for the ionized species and the new electron, but if the reaction is not impact ionization, I will set the ID of the ionized species and the new electron to invalid.

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -202,6 +236,26 @@ public:
...
+                    const amrex::ParticleReal up = sqrt(E_out * (E_out + 2.0_prt*mc2) / c2) / m1;
+
+                    // isotropically scatter electrons
+                    // - The incident electron
+                    ParticleUtils::RandomizeVelocity(ux1, uy1, uz1, up, engine);
+                    // - The newly created electron
+                    auto& e_ux1 = soa_products_data[0].m_rdata[PIdx::ux][product1_index+1];
+                    auto& e_uy1 = soa_products_data[0].m_rdata[PIdx::uy][product1_index+1];
+                    auto& e_uz1 = soa_products_data[0].m_rdata[PIdx::uz][product1_index+1];
+                    ParticleUtils::RandomizeVelocity(e_ux1, e_uy1, e_uz1, up, engine);
```
RemiLehe, line 258: Need to do the transform back to labframe for `e_u`

archermarx:
I've written a test script for this feature (attached)
[picmi.py.txt](https://github.com/user-attachments/files/17495531/picmi.py.txt)

It should work with both MCC and DSMC, once everything is implemented 

In brief, this script sets up a simulation of electrons, ions, and neutrals in a periodic box. Ionization is then allowed to happen for a short time, and the resulting electron density and temperature curves are compared to the output of very simple global model, the equations of which are given below:

image

Here, $\epsilon_{iz}$ is the ionization energy and $k_{iz}$ is the ionization rate coefficient, which the script computes by integrating over the provided cross sections (i.e. those from https://github.com/ECP-WarpX/warpx-data/tree/master/MCC_cross_sections/Xe)

---

At present, this segfaults when using DSMC after O(10-20) iterations, but completes when using MCC. When MCC is enabled, the global model neglects the depletion of neutrals. Some results obtained when using Xenon at a few electron temperatures, are given below

## $T_e$ = 10 eV
Untitled

## $T_e$ = 30 eV
Untitled

## $T_e$ = 50 eV
Untitled

---

Overall, the results are close, but seem to be off from the global model in a consistent manner -- namely that WarpX ionizes less than the global model predicts that it should. The relative discrepancy gets smaller as the temperature increases but the absolute value of the temperature difference between the model and WarpX is roughtly constant. I'd appreciate some extra eyes on this to make sure both my physics or my WarpX set up are correct. The script runs fast (~11 seconds on my machine), so hopefully it will be useful for debugging the DSMC implementation. Once energy loss and excitation are added, I can update the script to incorporate those as well.

jwestern:
Could this functionality be extended to do recombination? Even resulting in the deletion of the recombination product (i.e. back into the neutral "background" gas).

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.cpp
```diff
@@ -15,17 +15,36 @@ SplitAndScatterFunc::SplitAndScatterFunc (const std::string& collision_name,
 {
     const amrex::ParmParse pp_collision_name(collision_name);
 
+    // Check if ionization is one of the scattering processes
+    amrex::Vector<std::string> scattering_process_names;
+    pp_collision_name.queryarr("scattering_processes", scattering_process_names);
+    for (const auto& scattering_process : scattering_process_names) {
+        if (scattering_process.find("excitation") != std::string::npos) {
```
RemiLehe:
```suggestion
        if (scattering_process.find("ionization") != std::string::npos) {
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -106,6 +106,21 @@ public:
         // Therefore, we insert the colliding species at the beginning of `m_product_species`
         if (collision_type == CollisionType::DSMC) {
             m_product_species.insert( m_product_species.begin(), m_species_names.begin(), m_species_names.end() );
+            // Check if ionization is one of the scattering processes
+            // TODO: This is reused in several places ; we should put it in a separate function
+            amrex::Vector<std::string> scattering_process_names;
+            bool ionization_flag = false;
+            pp_collision_name.queryarr("scattering_processes", scattering_process_names);
+            for (const auto& scattering_process : scattering_process_names) {
+                if (scattering_process.find("excitation") != std::string::npos) {
```
```suggestion
                if (scattering_process.find("ionization") != std::string::npos) {
```

RemiLehe:
Superseded by https://github.com/ECP-WarpX/WarpX/pull/5524

PR #5389, opened 2024-10-11
Rigid injection: Center field scaling around the v push

RemiLehe:
In the rigid injection, the fields where scale by the fraction of time spent between `n*dt` and `(n+1)*dt` to the right of the injection plane. However, to be consistent with the leap-frog velocity update, this needs to be between `(n-1/2)*dt` and `(n+1/2)*dt` instead.

As a side-effect of this PR, saving and re-setting `u` and `optical_depth` to their original value is not needed anymore since the scaling factor for E and B is 0 for particles to the left of the plane.


Source/Particles/RigidInjectedParticleContainer.cpp
```diff
@@ -252,9 +252,6 @@ RigidInjectedParticleContainer::PushPX (WarpXParIter& pti,
         amrex::ParticleReal* AMREX_RESTRICT x_save = xp_save.dataPtr();
         amrex::ParticleReal* AMREX_RESTRICT y_save = yp_save.dataPtr();
         amrex::ParticleReal* AMREX_RESTRICT z_save = zp_save.dataPtr();
```
RemiLehe, line 220: Saving and re-setting `u` to its original value is not needed anymore since the scaling factor for E and B is 0 for particles to the left of the plane.

Source/Particles/Gather/ScaleFields.H
```diff
@@ -47,8 +47,20 @@ struct ScaleFields
...
+        // The scaling factor corresponds to the fraction of time that
+        // the particles spends to the right of the injection plane,
+        // between (n-1/2)*dt and (n+1/2)*dt, which is the interval
+        // over which the velocity is updated, in the leap-frog velocity push.
+        // (Note that here, `zp` is the particle position at time n*dt)
+        amrex::Real dtscale = 0.5_rt - (m_z_plane_previous - zp)/(m_vz_ave_boosted + m_v_boost)/m_dt;
+        // If the particle stays to the left of the plane during the
+        // whole push, simply set the scaling factor to 0.
+        if (dtscale < 0._rt) {
+            dtscale = 0;
```
dpgrote, line 61: Perhaps add a comment that since the scale factor is zero, the push of the velocity doesn't change the velocity.

dpgrote (approved):
This looks good. I approve, but have one suggestion for a comment to add.

PR #5395, opened 2024-10-15
Add PICMI interface for injecting from embedded boundary

RemiLehe:
This adds the option to inject particles from the embedded boundary with PICMI.


roelof-groenewald (approved):
Looks good to me!

PR #5391, opened 2024-10-14
AMReX/pyAMReX/PICSAR: weekly update

EZoni:
- Weekly update to latest AMReX:
```console
./Tools/Release/updateAMReX.py
```
- Weekly update to latest pyAMReX:
```console
./Tools/Release/updatepyAMReX.py
```
- Weekly update to latest PICSAR (no changes):
```console
./Tools/Release/updatePICSAR.py
```

PR #5401, opened 2024-10-21
Fix ambiguous AddRealComp

dpgrote:
Add boolean argument so that the calls to `AddRealComp` and `AddIntComp` are not ambiguous. This is needed since in AMReX PR https://github.com/AMReX-Codes/amrex/pull/4163 the `AddRealComp` was added to AMReX that takes a string as the first argument. This duplicates the routine in WarpX, giving a compile time error (with gcc on Mac OS) that "call to member function 'AddRealComp' is ambiguous". The second arguments are different (WarpX takes a boolean, AMReX an integer), so adding the explicit boolean to the calls in WarpX makes the calls unambiguous.

Is the long term plan to remove the code from WarpX and rely on the name components capability in AMReX?

### DIFF
--- a/Source/Diagnostics/ParticleIO.cpp
+++ b/Source/Diagnostics/ParticleIO.cpp
@@ -173,7 +173,7 @@ MultiParticleContainer::Restart (const std::string& dir)
                     + " was found in the checkpoint file, but it has not been added yet. "
                     + " Adding it now."
                 );
-                pc->AddRealComp(comp_name);
+                pc->AddRealComp(comp_name, true);
             }
         }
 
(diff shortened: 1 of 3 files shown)


EZoni:
Hi @dpgrote, thanks!

Please check #5391 where we started some of this. The update was supposed to be merged last week, but it got delayed until today. 

As for this question, 
> Is the long term plan to remove the code from WarpX and rely on the name components capability in AMReX?

yes, @atmyers confirmed that this is the plan.

Also, I think https://github.com/AMReX-Codes/amrex/pull/4163 wasn't included in AMReX's 24.10 release, so this should not affect our 24.10 release either, if I'm not mistaken. I think this is why it came up only with last week's weekly update #5391.

dpgrote:
> Also, I think [AMReX-Codes/amrex#4163](https://github.com/AMReX-Codes/amrex/pull/4163) wasn't included in AMReX's 24.10 release, so this should not affect our 24.10 release either, if I'm not mistaken. I think this is why it came up only with last week's weekly update #5391.

I can confirm that using the specified version of AMReX with 24.10 does build correctly on MacOS.

dpgrote:
This is closed since it duplicates the fix that is done in PR #5391.

PR #4817, opened 2024-04-02
Use pyamrex in spacecraft charging example

RemiLehe:
# Overview

This PR modernizes the spacecraft charging example to use the newer `pyamrex` interface for multifabs, instead of the fields wrapper. 

The `pyamrex` interface is in principle more flexible (since it gives access to the AMReX functions that operate on `MultiFabs` and should have all the functionalities that the old field wrappers had. (See also https://github.com/AMReX-Codes/pyamrex/pull/370). This is also the interface that is documented in the [WarpX documentation](https://warpx.readthedocs.io/en/latest/usage/workflows/python_extend.html#fields)

# Resetting the checksum

I had to reset the checksum, as they differed by `~1e-2`, which is not negligible!

Manually looking at the fields seem to indicate that the old interface and new interface differ in how they were handling the guard cells (at least in the way I use them in this test). While I don't fully understand the differences, this PR seems to produce more correct results. 
(In hindsight, it is striking that we missed the unphysical `Ez` field when we implemented this test. The reason might be that we were focusing `phi`, which does not exhibit this unphysical feature in such an obvious way.)

Screenshot 2025-04-18 at 8 55 31 AM

PR #3630, opened 2023-01-19
Add photon-photon collisions (linear Breit-Wheeler)

aeriforme:
This PR adds a module to simulate two-photon collisions according to the linear Breit-Wheeler process. 

Minimal to-do list in arbitrary order:
- [x] add check on product species (must be electron and positrons)
- [x] fix comments 
- [x] analysis tools for tests (develop benchmark)
- [ ] possibly merge [TwoProductFusionInitializeMomentum.H](https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/Collision/BinaryCollision/NuclearFusion/TwoProductFusionInitializeMomentum.H)  with [TwoProductInitializeMomentum.H](https://github.com/aeriforme/WarpX/blob/7a4d862bc541bf01e24376e4e4afe8ce9536aeb1/Source/Particles/Collision/BinaryCollision/PhotonPhoton/TwoProductInitializeMomentum.H) and [TwoProductFusionUtil.H](https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/Collision/BinaryCollision/NuclearFusion/TwoProductFusionUtil.H) with [TwoProductUtil.H](https://github.com/aeriforme/WarpX/blob/7a4d862bc541bf01e24376e4e4afe8ce9536aeb1/Source/Particles/Collision/BinaryCollision/PhotonPhoton/TwoProductUtil.H)
- [ ] consider nomenclature 
- [x] remove masses from inputs e.g. in  [TwoProductInitializeMomentum.H](https://github.com/aeriforme/WarpX/blob/7a4d862bc541bf01e24376e4e4afe8ce9536aeb1/Source/Particles/Collision/BinaryCollision/PhotonPhoton/TwoProductInitializeMomentum.H)
- [x] remove temporary prints
- [x] remove compiler warnings
- [x] documentation

PR #5407, opened 2024-10-22
fields.py relies on MultiFab global indexing

dpgrote:
The global indexing capability for MultiFabs was moved to pyamrex. This PR deletes the now not needed code from WarpX, relying on the capability in pyamrex.

Note that the MultiFab wrapper is maintained since it is persistent across load balancing, while the underlying MultiFabs are deleted and regenerated.

This is an API breaking change since there are differences in how to access the ghost cells when using the global indexing.


Examples/Physics_applications/spacecraft_charging/inputs_test_rz_spacecraft_charging_picmi.py
```diff
@@ -98,8 +98,8 @@ def compute_virtual_charge_on_spacecraft():
     that WarpX thinks there should be on the spacecraft.
     """
     # Get global array for the whole domain (across MPI ranks)
-    phi = PhiFPWrapper(include_ghosts=False)[:, :]
-    rho = RhoFPWrapper(include_ghosts=False)[:, :]
+    phi = PhiFPWrapper()[:, :]
+    rho = RhoFPWrapper()[:, :]
```
EZoni: Just to make sure I understood this PR correctly, these are intentionally using `[:, :]`, instead of `[(), ()]`, because we want to exclude ghost cells, correct?

Docs/source/developers/python.rst
```diff
@@ -123,30 +121,27 @@ This example will return the ``Bz`` field at all points along ``x`` at the speci
...
 
    from pywarpx import fields
    Jy = fields.JyFPWrapper()
    Jy[5,6:20,8:30] = 7.
 
-The code does error checking to ensure that the specified indices are within the bounds of the global domain.
-Note that negative indices are handled differently than with numpy arrays because of the possibility of having ghost cells.
-With ghost cells, the lower ghost cells are accessed using negative indices (since ``0`` is the index of the lower bound of the
-valid cells). Without ghost cells, a negative index will always raise an out of bounds error since there are no ghost cells.
+In this example, seven is added to all of the values along ``x``, including both valid and ghost cells (specified by using the empty tuple), the first ghost cell at the lower boundary in ``y``, and the last valid cell and fist upper guard cell in ``z``.
```
EZoni:
```suggestion
In this example, seven is added to all of the values along ``x``, including both valid and ghost cells (specified by using the empty tuple), the first ghost cell at the lower boundary in ``y``, and the last valid cell and first upper guard cell in ``z``.
```

EZoni:
@dpgrote 

The PR looks good to me in general. There is only a merge conflict to fix. Since this contains an API breaking change, personally I don't remember how we had decided to roll this out. @ax3l and @RemiLehe might want to comment on this.

Examples/Physics_applications/spacecraft_charging/inputs_test_rz_spacecraft_charging_picmi.py
dpgrote: Yes, that is correct.

ax3l: Fancy :D Had to remind myself of that convention again by reading your doc :)
https://pyamrex.readthedocs.io/en/latest/usage/compute.html

Examples/Physics_applications/spacecraft_charging/inputs_test_rz_spacecraft_charging_picmi.py
```diff
@@ -49,12 +49,12 @@ def correct_space_charge_fields(self, q=None):
             q = compute_actual_charge_on_spacecraft()
 
         # Correct fields so as to recover the actual charge
-        Er = ExWrapper(include_ghosts=True)
-        Er[...] += (q - q_v) * self.normalized_Er
-        Ez = EzWrapper(include_ghosts=True)
-        Ez[...] += (q - q_v) * self.normalized_Ez
-        phi = PhiFPWrapper(include_ghosts=True)
-        phi[...] += (q - q_v) * self.normalized_phi
+        Er = ExWrapper()
```
ax3l: attn #4817

ax3l:
> Note that the MultiFab wrapper is maintained since it is persistent across load balancing, while the underlying MultiFabs are deleted and regenerated.

I might need a recap why we want to keep the *named* wrappers (like `ExWrapper`) though.  I thought we want to remove them, because we have to keep their class definitions in sync with all potential multifabs we register, and we might add more at runtime that then have no corresponding wrapper.

Wouldn't it be also just fine to query the fields directly from `warpx.multifab` again when we next access them (potentially after a load balance)? That way we could remove the extra wrappers.

dpgrote:
> > Note that the MultiFab wrapper is maintained since it is persistent across load balancing, while the underlying MultiFabs are deleted and regenerated.
> 
> I might need a recap why we want to keep the _named_ wrappers (like `ExWrapper`) though. I thought we want to remove them, because we have to keep their class definitions in sync with all potential multifabs we register, and we might add more at runtime that then have no corresponding wrapper.
> 
> Wouldn't it be also just fine to query the fields directly from `warpx.multifab` again when we next access them (potentially after a load balance)? That way we could remove the extra wrappers.

Having the named wrappers is a convenience for the users, providing a list of the available MultiFabs in one place without having to search through the code. This also provides a nice way of fetching the MultiFabs without having to know the format of the MultiFab names in the registry. This is not essential - for instance a user could list the keys of the MF registry to get the names. But that is only available at runtime.

Yes, a user could always fetch the MF whenever they want to use it, but this puts on the onus on the user to do this. It would also put an onus on the user to keep track of when load balancing is done, which is likely to be problematic. With the wrapper, the user doesn't have to think about this and doesn't even have to be aware of the issue.

Docs/source/developers/python.rst
```diff
@@ -98,23 +98,21 @@ There are two python modules that provide convenient access to the fields and th
...
-By default, this wraps the MultiFab for level 0. The ``level`` argument can be specified for other levels.
-By default, the wrapper only includes the valid cells. To include the ghost cells, set the argument ``include_ghosts=True``.
+   Ex = fields.ExWrapper(level=0)
 
 The wrapper provides access to the data via global indexing.
-Using standard array indexing (with exceptions) with square brackets, the data can be accessed using indices that are relative to the full domain (across the MultiFab and across processors).
-With multiple processors, the result is broadcast to all processors.
-This example will return the ``Bz`` field at all points along ``x`` at the specified ``y`` and ``z`` indices.
+Using standard array indexing with square brackets, the data can be accessed using indices that are relative to the full domain (across the MultiFab and across processors).
+For indices within the domain, values from valid cells are always returned. The ghost cells at the exterior of the domain are accessed using imagninary numbers, with negative values accessing the lower ghost cells, and positive the upper ghost cells.
```
roelof-groenewald:
```suggestion
For indices within the domain, values from valid cells are always returned. The ghost cells at the exterior of the domain are accessed using imaginary numbers, with negative values accessing the lower ghost cells, and positive the upper ghost cells.
```

Examples/Physics_applications/capacitive_discharge/inputs_test_2d_background_mcc_picmi.py
```diff
@@ -161,14 +161,14 @@ def _run_solve(self):
 
         # get rho from WarpX
         if self.rho_wrapper is None:
-            self.rho_wrapper = fields.RhoFPWrapper(0, True)
-        self.rho_data = self.rho_wrapper[Ellipsis]
+            self.rho_wrapper = fields.RhoFPWrapper(0)
+        self.rho_data = self.rho_wrapper[(), ()]
```
roelof-groenewald, line 165: Seeing as the examples are meant to guide users on how to use the code, it would be good to make a comment here that `[(), ()]` accesses both valid and ghost cells (whereas `[Ellipsis]` would only access valid cells.
Do you know if there is a way to automatically collect all dimensions? The `Ellipsis` is convenient since code can be reused for different dimensions, can the same be done with the `[()]` indexing?

Docs/source/developers/python.rst
```diff
@@ -123,30 +121,27 @@ This example will return the ``Bz`` field at all points along ``x`` at the speci
...
 
    from pywarpx import fields
    Jy = fields.JyFPWrapper()
    Jy[5,6:20,8:30] = 7.
 
-The code does error checking to ensure that the specified indices are within the bounds of the global domain.
-Note that negative indices are handled differently than with numpy arrays because of the possibility of having ghost cells.
-With ghost cells, the lower ghost cells are accessed using negative indices (since ``0`` is the index of the lower bound of the
-valid cells). Without ghost cells, a negative index will always raise an out of bounds error since there are no ghost cells.
+In this example, seven is added to all of the values along ``x``, including both valid and ghost cells (specified by using the empty tuple), the first ghost cell at the lower boundary in ``y``, and the last valid cell and first upper guard cell in ``z``.
```
roelof-groenewald:
```suggestion
In this example, seven is added to all of the values along ``x``, including both valid and ghost cells (specified by using the empty tuple, ``()``), the first ghost cell at the lower boundary in ``y``, and the last valid cell and first upper guard cell in ``z``.
```

roelof-groenewald (commented):
These changes look good to me. It will not be a big task to update existing scripts to use the new indexing notation.

I just a left a few comments on suggested clarifications.

EZoni:
Just one more question, to prepare for more discussions later on.

It's not clear to me if we reached consensus on the fact that this new syntax
```
rho = RhoFPWrapper()[:, :]    # exclude ghost cells
rho = RhoFPWrapper()[(), ()]  # include ghost cells
```
is more user friendly than the `multifab` syntax, and/or than the old syntax for the wrappers, which used an input parameter `include_ghosts`.

My understanding based on the discussion we just had during the developers meeting is that @jlvay, and possibly others, were concerned about keeping a syntax that is user-friendly (e.g., more Warp-like) and claimed that the new interface that relies on `multifab` isn't user-friendly enough.

What makes this new syntax for the old wrappers (example above) more user-friendly than the `multifab` syntax, given that we are advocating for not exposing the `multifab` syntax to the user because of the fact that it is not user-friendly enough (and because of the missing feature of guaranteeing a pointer to valid memory, which however could be probably integrated/added, wherever missing)?

It's possible that I got confused about which arguments and/or objections were raised during the meeting.

Python/pywarpx/fields.py
```diff
@@ -161,568 +184,159 @@ def mesh(self, direction):
...
-        index: integer, or sequence of integers or slices, or Ellipsis
-            Index of the slice to return
-        """
-        # Note that the index can have negative values (which wrap around) and has 1 added to the upper
-        # limit using python style slicing
-        if index == Ellipsis:
-            index = self.dim * [slice(None)]
-        elif isinstance(index, slice):
-            # If only one slice passed in, it was not wrapped in a list
-            index = [index]
```
roelof-groenewald, line 401: Perhaps this functionality is worth keeping? This could also allow a workaround for the question of needing to do `[(), (), ()]` for 3d to allow `[()]` to work.

roelof-groenewald:
@EZoni from my perspective, there's no problem with changing the notation used for indexing (especially since only access to guard cells are changed). Bringing the notation used with `fields.py` wrappers into alignment with what is used directly in `pyamrex` is a good thing. I just objected to removing the `fields.py` wrappers altogether, since that would require a significantly bigger update on our end than only changing how the indexing is done.

Examples/Physics_applications/capacitive_discharge/inputs_test_2d_background_mcc_picmi.py
dpgrote: Yes, this can be done, having `[()]` refer to guard cells in all axis. With this PR, that would be done in pyamrex. I can submit a PR there.

Python/pywarpx/fields.py
dpgrote: Yes, this could be done here, but it's a small enough change in pyamrex, that having a temporary solution here is not needed.

Examples/Physics_applications/capacitive_discharge/inputs_test_2d_background_mcc_picmi.py
roelof-groenewald: Thanks!

roelof-groenewald (approved):
This looks good-to-go. 
I don't think we need to wait for the pyamrex PR before we merge this since it is good to have the examples extra verbose for clarity.

roelof-groenewald:
@RevathiJambunathan and @clarkse have you guys had a chance to look over this PR yet?

Python/pywarpx/fields.py
dpgrote: Here is the pull request https://github.com/AMReX-Codes/pyamrex/pull/438

roelof-groenewald: Thank you!

EZoni:
Before merging, please note that none of the GitHub Actions workflows ran on the last commit, not sure why. It may be worth amending the commit to re-trigger the CI checks.

EZoni:
~On another note, two big PRs will be merged soon:~
~- #3630~
~- #5682~ 

~Since those PRs introduce new CI tests, I would recommend that we merge `development` here, after those PRs are merged, before resetting the checksums in this PR.~

Docs/source/developers/python.rst
```diff
@@ -123,30 +121,31 @@ This example will return the ``Bz`` field at all points along ``x`` at the speci
...
-Under the covers, the wrapper object has a list of numpy arrays that have pointers to the underlying data, one array for each FAB.
-When data is being fetched, it loops over that list to gather the data.
-The result is then gathered among all processors.
-Note that the result is not writeable, in the sense that changing it won’t change the underlying data since it is a copy.
-When the data is set, using the global indexing, a similar process is done where the processors loop over their FABs and set the data at the appropriate indices.
+   from pywarpx import fields
+   Jx = fields.JxFPWrapper()
+   Jx[(),-1j,-1:2j] += 7.
+
+The fetch the data from all of the valid cells of all dimensions, the ellipsis can be used, ``Jx[...]``.
```
roelof-groenewald:
```suggestion
To fetch the data from all of the valid cells of all dimensions, the ellipsis can be used, ``Jx[...]``.
```

roelof-groenewald:
> Since those PRs introduce new CI tests, I would recommend that we merge `development` here, after those PRs are merged, before resetting the checksums in this PR.

@EZoni, this PR does not reset any checksums, so I don't think it is necessary to wait for those other PRs. Or am I missing something?

EZoni:
@roelof-groenewald 

Sorry, I meant to post my message on #5820. Too many PRs opened in the browser...

roelof-groenewald:
I think we should go ahead an merge this to avoid the need for @dpgrote to continually keep it from going stale. 
Any objections @RevathiJambunathan?

Examples/Tests/ohm_solver_magnetic_reconnection/inputs_test_2d_ohm_solver_magnetic_reconnection_picmi.py
```diff
@@ -301,12 +301,12 @@ def check_fields(self):
         if not (step == 1 or step % self.diag_steps == 0):
             return
 
-        rho = fields.RhoFPWrapper(include_ghosts=False)[:, :]
-        Jiy = fields.JyFPWrapper(include_ghosts=False)[...] / self.J0
-        Jy = fields.JyFPPlasmaWrapper(include_ghosts=False)[...] / self.J0
-        Bx = fields.BxFPWrapper(include_ghosts=False)[...] / self.B0
-        By = fields.ByFPWrapper(include_ghosts=False)[...] / self.B0
-        Bz = fields.BzFPWrapper(include_ghosts=False)[...] / self.B0
+        rho = fields.RhoFPWrapper()[:, :]
```
RevathiJambunathan:
```suggestion
        rho = fields.RhoFPWrapper()[...]
```

Examples/Physics_applications/capacitive_discharge/inputs_base_1d_picmi.py
```diff
@@ -98,11 +98,11 @@ def _run_solve(self):
         """Function run on every step to perform the required steps to solve
         Poisson's equation."""
         # get rho from WarpX
-        self.rho_data = fields.RhoFPWrapper(0, False)[...]
+        self.rho_data = fields.RhoFPWrapper(0)[...]
```
EZoni, line 101: I would be in favor of introducing an extra layer that makes the ghost cells access more user-friendly, doing something like
```py
        self.rho_data = fields.RhoFP(0, ghost=False)
```
with `RhoFP` defined somewhere like
```py
        def RhoFP(level, ghost):
            if ghost:
                return RhoFPWrapper(level)[()]
            else:
                return RhoFPWrapper(level)[...]
```
hence possibly hiding the distinction of `[()]` vs `[...]`.


@jlvay @RemiLehe @RevathiJambunathan @roelof-groenewald @dpgrote
Sorry for the repeat, but my question would be, why is this option not considered in the same way as the rest of the discussion about keeping the `Wrapper` interfaces for clarity and ease of use?

Examples/Physics_applications/spacecraft_charging/inputs_test_rz_spacecraft_charging_picmi.py
```diff
@@ -82,49 +82,34 @@ def save_normalized_vacuum_Efields(
...
-            "normalized_Er",
-            0,
-            Er.box_array(),
-            warpx.DistributionMap(0),
-            1,
-            Er.n_grow_vect,
-            0.0,
-            True,
-            True,
+        normalized_Er = fields.MultiFabWrapper(
```
dpgrote, line 94: @RevathiJambunathan I've added a way of adding new MultiFabs to the wrapper and updated this to use it. How does this look? Note that I changed the variable name used for the registry since using `fields` conflicts with the name of the `fields` module where the wrappers are.

clarkse:
I have no issue with the change in syntax. Looks good to me. An announcement after this gets merged would be nice so I know to go update my personal toolkits.

roelof-groenewald:
@dpgrote sorry for the delay in merging this. I take it all interested parties have now had a chance to weigh in on this and everyone is satisfied that we can merge this change. I see there is now one conflict though. Could you please resolve that so we can merge the PR?

PR #5412, opened 2024-10-24
Python: useful error when `initialize_warpx` not called before creating `ParticleContainerWrapper`

archermarx:
Currently, WarpX is initialized when `sim.step` is called, or when the user calls `initialize_warpx`. However, if the user tries to create a `ParticleContainerWrapper` before this point, they get an error along the following lines:

```
 File "/home/marksta/projects/warpx-ionization/picmi.py", line 185, in <module>
    sim.run()
  File "/home/marksta/projects/warpx-ionization/picmi.py", line 179, in run
    self.elec_wrapper = particle_containers.ParticleContainerWrapper(self.electrons.name)
  File "/home/marksta/.local/lib/python3.10/site-packages/pywarpx/particle_containers.py", line 29, in __init__
    mypc = libwarpx.warpx.multi_particle_container()
  File "/home/marksta/.local/lib/python3.10/site-packages/pywarpx/_libwarpx.py", line 46, in __getattr__
    return self.__getattribute__(attribute)
AttributeError: 'LibWarpX' object has no attribute 'warpx'
```
This is confusing. When I got this, I assumed I had maybe installed WarpX wrong. I added a catch for this exception that re-raises it with some additional context that should help direct the user to call `initialize_warpx`


dpgrote:
I would suggest handling this in a different way. Now, the issue is that the `libwarpx.warpx ` is accessed in the `__init__` routine. A change would be to only access `libwarpx.warpx` when it is needed. Perhaps make `particle_container` a property that gets the underlying particle container when needed. The user can then create the ParticleContainerWrapper at any time. The message this PR adds would then go in the `particle_container` property get method. This is how the field wrappers are setup.

archermarx:
Got it, thanks for the feedback!

archermarx:
Ok, I've modified the PR to now lazy initialize the particle container (and boundary buffers) only when they're first needed instead of in `__init__`. Only then is an error thrown.

Python/pywarpx/particle_containers.py
```diff
@@ -24,10 +24,20 @@ class ParticleContainerWrapper(object):
 
     def __init__(self, species_name):
         self.name = species_name
+        self.particle_container = None
+
+    def get_particle_container(self):
```
dpgrote: This is a great idea - I often use this scheme of only defining things when they are needed.

Note that making this a property would be a bit cleaner.
```
    self._particle_container = None

@property
def particle_container(self):
    if self._particle_container is None:
        self._particle_container = ...
    return self._particle_container
```

archermarx: Oh neat, that's new to me. I'll make the changes

archermarx:
Changes made. That's definitely a lot cleaner.

dpgrote (approved):

Python/pywarpx/particle_containers.py
```diff
@@ -24,10 +24,21 @@ class ParticleContainerWrapper(object):
...
+    @property
+    def particle_container(self):
+        if self._particle_container is None:
+            try:
+                mypc = libwarpx.warpx.multi_particle_container()
+                self._particle_container = mypc.get_particle_container_from_name(
+                    self.name
+                )
+            except AttributeError as e:
+                msg = "This is likely caused by attempting to access a ParticleContainerWrapper before initialize_warpx has been called"
```
ax3l, line 38: Thank you for this PR! Sorry for just seeing it now, I was just on parental leave when the PR was opened.

I thought we could also add one more fallback, right in the library loader, where we assign the `libwarpx.warpx` attribute.

In essence, we could assign an initial attribute that throws when accessed with a more distinct error message ("initialize_warpx was not yet called!"), to catch all. More specific exceptions like this one are great to add and keep.

One initialization, we then just overwrite the attribute.

archermarx: Oh neat idea! I can open a new PR shortly!

PR #5417, opened 2024-10-27
Check that all E-field values are finite in Ohm solver

roelof-groenewald:
This assert is meant to help identify the origin of a common problem in hybrid-PIC simulations, wherein unresolved Whistler waves cause runaway E-field values. Currently when this happens, WarpX fails in the current deposition routine.

### DIFF
--- a/Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
+++ b/Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
@@ -189,6 +189,18 @@ void WarpX::HybridPICEvolveFields ()
                            0, 0, 1, current_fp_temp[lev][idim]->nGrowVect());
         }
     }
+
+    // Check that the E-field does not have nan or inf values, otherwise print a clear message
+    ablastr::fields::MultiLevelVectorField Efield_fp = m_fields.get_mr_levels_alldirs(FieldType::Efield_fp, finest_level);
+    for (int lev = 0; lev <= finest_level; ++lev)
+    {
+        for (int idim = 0; idim < 3; ++idim) {
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                Efield_fp[lev][idim]->is_finite(),
+                "Non-finite value detected in E-field; this indicates more substeps should be used in the field solver."
+            );
+        }
+    }
 }
 
 void WarpX::HybridPICDepositInitialRhoAndJ ()
--- a/Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
+++ b/Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
@@ -195,7 +195,7 @@ void WarpX::HybridPICEvolveFields ()
     for (int lev = 0; lev <= finest_level; ++lev)
     {
         for (int idim = 0; idim < 3; ++idim) {
-            if !(Efield_fp[lev][idim]->is_finite(true)) {
+            if (!Efield_fp[lev][idim]->is_finite(true)) {
                 amrex::Abort("Non-finite value detected in E-field; this indicates more substeps should be used in the field solver.");
             }
         }
--- a/Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
+++ b/Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
@@ -195,9 +195,9 @@ void WarpX::HybridPICEvolveFields ()
     for (int lev = 0; lev <= finest_level; ++lev)
     {
         for (int idim = 0; idim < 3; ++idim) {
-            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
-                Efield_fp[lev][idim]->is_finite(),
-                "Non-finite value detected in E-field; this indicates more substeps should be used in the field solver."
+            if !(Efield_fp[lev][idim]->is_finite(true)) {
+                amrex::Abort("Non-finite value detected in E-field; this indicates more substeps should be used in the field solver.");
+            }
             );
         }
     }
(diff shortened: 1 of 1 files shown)


kli-jfp:
Great idea! If this impacts the code’s speed, perhaps a debug flag should be used for this.

EZoni:
@roelof-groenewald 

The Intel oneAPI workflows have been fixed in #5419, please rebase on/merge `development` to fix those workflows here.

roelof-groenewald:
> Great idea! If this impacts the code’s speed, perhaps a debug flag should be used for this.

Thanks @kli-jfp. Local testing didn't show much of any performance penalty due to the new check since the field solver already includes quite a few reduction operations (so adding one more isn't really noticeable). But if you find in your application it does change the performance we can definitely update it to only check in debug mode.

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -189,6 +189,18 @@ void WarpX::HybridPICEvolveFields ()
...
     }
+
+    // Check that the E-field does not have nan or inf values, otherwise print a clear message
+    ablastr::fields::MultiLevelVectorField Efield_fp = m_fields.get_mr_levels_alldirs(FieldType::Efield_fp, finest_level);
+    for (int lev = 0; lev <= finest_level; ++lev)
+    {
+        for (int idim = 0; idim < 3; ++idim) {
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                Efield_fp[lev][idim]->is_finite(),
+                "Non-finite value detected in E-field; this indicates more substeps should be used in the field solver."
```
RemiLehe, line 200: We could:
- Set the flag `local=true` to avoid parallel operations (reduction)
- Use `amrex::Abort` so that it does not hand if only one of the MPI ranks fails.

roelof-groenewald, line 200:
```suggestion
            if !(Efield_fp[lev][idim]->is_finite(true)) {
                amrex::Abort("Non-finite value detected in E-field; this indicates more substeps should be used in the field solver.");
            }
```

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -189,6 +189,18 @@ void WarpX::HybridPICEvolveFields ()
...
+
+    // Check that the E-field does not have nan or inf values, otherwise print a clear message
+    ablastr::fields::MultiLevelVectorField Efield_fp = m_fields.get_mr_levels_alldirs(FieldType::Efield_fp, finest_level);
+    for (int lev = 0; lev <= finest_level; ++lev)
+    {
+        for (int idim = 0; idim < 3; ++idim) {
+            if !(Efield_fp[lev][idim]->is_finite(true)) {
+                amrex::Abort("Non-finite value detected in E-field; this indicates more substeps should be used in the field solver.");
+            }
+            );
```
roelof-groenewald, line 201:
```suggestion
```

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -189,6 +189,17 @@ void WarpX::HybridPICEvolveFields ()
                            0, 0, 1, current_fp_temp[lev][idim]->nGrowVect());
         }
     }
+
+    // Check that the E-field does not have nan or inf values, otherwise print a clear message
+    ablastr::fields::MultiLevelVectorField Efield_fp = m_fields.get_mr_levels_alldirs(FieldType::Efield_fp, finest_level);
+    for (int lev = 0; lev <= finest_level; ++lev)
+    {
+        for (int idim = 0; idim < 3; ++idim) {
+            if !(Efield_fp[lev][idim]->is_finite(true)) {
```
```suggestion
            if (!Efield_fp[lev][idim]->is_finite(true)) {
```

roelof-groenewald:
Closing and reopening to trigger Azure tests.

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
roelof-groenewald: I tested that approach and it still hangs if all processors don't call `Abort`. So I reverted to using `local=false`. I would imagine the way that is implemented is to basically call `is_finite` locally and then reduce the resulting `boolean` for all ranks, which being 1 bit should be very fast.

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -189,6 +189,18 @@ void WarpX::HybridPICEvolveFields ()
...
         }
     }
+
+    // Check that the E-field does not have nan or inf values, otherwise print a clear message
+    ablastr::fields::MultiLevelVectorField Efield_fp = m_fields.get_mr_levels_alldirs(FieldType::Efield_fp, finest_level);
+    for (int lev = 0; lev <= finest_level; ++lev)
+    {
+        for (int idim = 0; idim < 3; ++idim) {
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                Efield_fp[lev][idim]->is_finite(),
```
roelof-groenewald, line 199: @WeiqunZhang this is the PR we mentioned in the developer meeting yesterday. 
I found that if I use `Efield_fp[lev][idim]->is_finite(true)` to avoid the MPI reduction the execution hangs (presumably due to only some processes returning `false`).
I've tested this by modifying the magnetic reconnection test to make it unstable. I attached the input script for the unstable case if you want to play around with it.

[warpx_inputs.txt](https://github.com/user-attachments/files/17651778/warpx_inputs.txt)

RemiLehe (approved):

PR #5419, opened 2024-10-28
CI: fix IntelLLVM builds

EZoni:
The CI checks `Intel / oneAPI ICX SP` and `Intel / oneAPI DPC++ SP` are failing since a few days.

This is likely due to the fact that the GitHub Actions runner is now installing IntelLLVM 2025.0.0 instead of IntelLLVM 2024.2.1, as until a few days ago.

This causes the following issue when building openPMD:
```console
/home/runner/work/WarpX/WarpX/build_sp/_deps/fetchedopenpmd-src/include/openPMD/backend/Container.hpp:263:32: error: no member named 'm_container' in 'Container<T, T_key, T_container>'
  263 |         container().swap(other.m_container);
      |                          ~~~~~ ^
1 error generated.
```

We can try to install the previous version of IntelLLVM manually and see if that fixes the issue.

### DIFF
--- a/.github/workflows/intel.yml
+++ b/.github/workflows/intel.yml
@@ -112,6 +112,7 @@ jobs:
         set +e
         source /opt/intel/oneapi/setvars.sh
         set -e
+        export PATH=$PATH:/opt/intel/oneapi/compiler/2024.2/bin  # FIXME
         export CXX=$(which icpx)
         export CC=$(which icx)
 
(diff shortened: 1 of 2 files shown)


.github/workflows/intel.yml
```diff
@@ -112,6 +112,7 @@ jobs:
         set +e
         source /opt/intel/oneapi/setvars.sh
         set -e
+        export PATH=$PATH:/opt/intel/oneapi/compiler/2024.2/bin  # FIXME
         export CXX=$(which icpx)
         export CC=$(which icx)
```
EZoni, line 117: Exporting `PATH` should be unnecessary here, given that we execute `source /opt/intel/oneapi/setvars.sh` before. However, for some reason the Intel `setvars` script still adds the path to the 2025 version of the Intel compiler.

EZoni:
openPMD reference: https://github.com/openPMD/openPMD-api/pull/1690

PR #5422, opened 2024-10-29
Added AnalyticFluxDistribution class

oshapoval:
Added an AnalyticFluxDistribution class with a parsed `flux_expression`. Depends on https://github.com/picmi-standard/picmi/pull/121


Python/pywarpx/picmi.py
```diff
@@ -739,6 +739,41 @@ def distribution_initialize_inputs(
             species.add_new_group_attr(source_name, "density", density_scale)
 
 
+class AnalyticFluxDistribution(
```
dpgrote: Is there a different way to handle this? This is an exact duplicate of `UniformFluxDistribution` except for the one keyword argument `flux_expression`.
This could be simplified if the picmistandard class took the same argument, `flux`, as the uniform distribution, then one could do "UniformFluxDistribution = AnalyticFluxDistribution" and have only one version of the class.

RemiLehe: Good point. 

Note that we also have two different classes for the volume injection: `UniformDistribution` and `AnalyticDistribution`, so we were trying to mirror the same setup here for `UniformFluxDistribution` and `AnalyticFluxDistribution`. 

Do you think that your comment also applies to `UniformDistribution` and `AnalyticDistribution`?

oshapoval: Thanks for pointing this out, I will refactor accordingly.

dpgrote: The UniformDistribution and AnalyticDistribution are quite a bit different from each other and it's probably too late to change that API.

dpgrote: Is this PR ready? I still see the duplication of 'UniformFluxDistribution' - this is fine if that is the intent. Also, `UniformFluxDistribution` has the `inject_from_embedded_boundary` option - would that also be supported with the analytic distribution?

dpgrote (approved):
This looks good now. Thanks!

PR #5426, opened 2024-10-29
Add `twiss` injection source

bnara:
Add `twiss` injection source.  Our need for this was to inject a beam with independent control of phase-space in all three Cartesian directions, in order to match an asymmetric beam into a plasma ramp.  This gives a flexible Twiss parameterization, including transverse focusing and longitudinal compression and chirp.  Furthermore, arbitrary orientation and location of beams is supported.  Online documentation `parameters.rst` updated.

Implementation details: [twiss-implementation.pdf](https://github.com/user-attachments/files/17808743/twiss-implementation.pdf) (updated Nov. 18)

<video src="https://github.com/user-attachments/assets/f26cd1c3-d390-496c-8e52-cc7a723b7b49" loop controls muted title="test-title"></video>

Three independent focal distances:<br>
<a href="url"></a>

Arbitrary positioning and orientation of beams, including rotation about longitudinal axes: <br>
<a href="url"></a>


bnara:
**Illustration of different cutting methods available in ``twiss``**

During beam initialization, either "planar" or "ellipsoidal" cuts can be made in 6D phase-space.  The user provides six parameters, corresponding to the three Cartesian coordinates and momenta.  The parameters are given in units of their respective focal rms standard deviations.  Specifying the numerical constant ``inf`` effectively disables cutting along the corresponding axis.  An ellipsoidal cut and a planar cut can be simultaneously used.

The ``gaussian_beam`` source provides planar cuts along 3 coordinate directions.  To make the same cut in ``twiss``, use:

```
beam.twiss.planar_cut = 1 1 1 inf inf inf
```

<video src="https://github.com/user-attachments/assets/8a399e7f-43f2-4ded-b19a-6979da1e9331" loop controls muted title="test-title"></video>

To make an ellipsoidal cut in all six coordinates and momenta at one standard deviation, use:

```
beam.twiss.ellipsoidal_cut = 1 1 1 1 1 1
```

<video src="https://github.com/user-attachments/assets/b747e389-c6a8-4ba1-8fd4-8e2bd9019b25" loop controls muted title="test-title"></video>

aeriforme:
Thank you for working on this PR! It's very useful.
Would you be able to add a test? This is required before merging. You can find [instructions here](https://warpx.readthedocs.io/en/latest/developers/how_to_test.html ).

Source/Initialization/WarpXAMReXInit.cpp
aeriforme, line 1: Why is this file being modified? 
It should't be necessary in this PR, given that it only deals with initialization.
Am I wrong?

bnara: Yes, I'd be happy to write a test!

The change I made to this file was adding `inf` to the AMReX parser, so that I could avoid hard-coded arbitrary "large numbers" in my input decks.  For example, I prefer writing
```
beam.twiss.planar_cut = 1 1 1 inf inf inf
```
instead of something like
```
beam.twiss.planar_cut = 1 1 1 1e12 1e12 1e12
```

It is not integral to the operation of the PR code, and it has been handy in other applications as well.  I could break it out into a separate PR.

aeriforme: Uhm ok, I will have to take a closer look.
However, I would advise against introducing `inf`, if not strictly necessary. 
We don't want to facilitate bugs :) 
I will get back to you on this anyway

PR #5427, opened 2024-10-30
Add Novatron paper in documentation

RemiLehe:


EZoni:
I don't remember our policy, do we wait until papers are peer-reviewed and published in a journal or do we add them at the preprint stage?

PR #5434, opened 2024-10-31
Fix dt_update_interval argument to ElectrostaticSolver

### LINKED ISSUE #5432: Documentation error, and possible bug, for Electrostatic solver
I think I have found 2 errors in the documentation for the Electrostatic solver. First, on the Parameters: Python (PICMI) page (https://warpx.readthedocs.io/en/latest/usage/python.html) it says:

> warpx_dt_update_interval (string, optional (default = -1)) – How frequently the timestep is updated. Adaptive timestepping is disabled when this is <= 0.

There seems to be 2 errors here. First, if the default value is -1, then this is not a string. It seems like it should be an integer. Second, there is no "warpx_dt_update_interval" keyword in the picmi.py file. The keyword is "dt_update_interval."

The "string" error is also present in the Parameters: Inputs File documentation (https://warpx.readthedocs.io/en/latest/usage/parameters.html)

> warpx.dt_update_interval (string) optional (default -1)

Finally, I should point out that neither "warpx_dt_update_interval" nor "dt_update_interval" are recognized as keywords when I run a simulation using the WarpX solver. But I am not sure if this is a bug, or that adaptive timestepping is not supported for the Electrostatic solver. If it is the latter, then the documentation should reflect that.

Thanks for your help!

dpgrote:
The `warpx` prefix was left off of this argument.

This addresses issues raised in #5431 and #5432

### DIFF
--- a/Python/pywarpx/picmi.py
+++ b/Python/pywarpx/picmi.py
@@ -1898,10 +1898,10 @@ class ElectrostaticSolver(picmistandard.PICMI_ElectrostaticSolver):
         How frequently the timestep is updated. Adaptive timestepping is disabled when this is <= 0.
 
     warpx_cfl: float, optional
-        Fraction of the CFL condition for particle velocity vs grid size, used to set the timestep when `dt_update_interval > 0`.
+        Fraction of the CFL condition for particle velocity vs grid size, used to set the timestep when `warpx_dt_update_interval > 0`.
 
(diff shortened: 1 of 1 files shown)


archermarx:
Thanks for the catch!

archermarx (approved):

roelof-groenewald (approved):
Thanks for fixing!

PR #5438, opened 2024-11-04
Correct inaccurate comment in IGF code

RemiLehe:
I forgot to update the comment in this PR:
https://github.com/ECP-WarpX/WarpX/pull/5335


aeriforme:
Ok, but why don't we copy the ghost cells? 
From [here](https://github.com/ECP-WarpX/WarpX/commit/36b55449dccb0a8bf60a431341c730c5b74cc5a1) I see that the reason is that otherwise ImpactX wouldn't converge. Do we know why though?

aeriforme (approved):

RemiLehe:
@aeriforme This is still being investigated, but in the meantime, I think we want to keep the comments in sync with the code.

PR #5441, opened 2024-11-04
Remove theta() function from ImplicitSolver base class.

JustinRayAngus:
This PR provides some minor cleanup to the implicit solvers.

1) Added the simulation time step (potentially adaptive) as a member to the ImplicitSolver base class. This removes the need to circularly pass a_dt from the implicit solver to the nonlinear solver and back to functions in the implicit solver class (e.g., ComputeRHS()).

2) The time step passed to m_nlsolver->Solve() inside OneStep() from the Implicit solvers is now the appropriate fractional time step. This removes the need for the theta() function called by the preconditioner. The theta() function is removed from the ImplicitSolver base class.

### DIFF
--- a/Source/FieldSolver/ImplicitSolvers/ImplicitSolver.H
+++ b/Source/FieldSolver/ImplicitSolvers/ImplicitSolver.H
@@ -86,8 +86,6 @@ public:
                               int              a_nl_iter,
                               bool             a_from_jacobian ) = 0;
 
-    [[nodiscard]] virtual amrex::Real theta () const { return 1.0; }
-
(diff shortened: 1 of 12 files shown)


Source/FieldSolver/ImplicitSolvers/ImplicitSolver.H
```diff
@@ -83,12 +83,9 @@ public:
     virtual void ComputeRHS ( WarpXSolverVec&  a_RHS,
```
dpgrote, line 83: I notice that the comments do not have the list of input parameters for the routines. This is not for this PR, but can be added in the future.

dpgrote (approved):
Looks good!

PR #5443, opened 2024-11-05
CI: build tests with `-g1` compile option

EZoni:
Build and run CI tests with more debug information by adding the `-g1` compile option.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -151,12 +151,19 @@ jobs:
       # display disk space usage
       df -h
       # configure
-      export AMReX_CMAKE_FLAGS="-DAMReX_ASSERTIONS=ON -DAMReX_TESTING=ON"
-      cmake -S . -B build       \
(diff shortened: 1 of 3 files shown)


EZoni:
I injected manual crashing and tested that the backtrace files display line numbers as expected. For example, see logs at https://dev.azure.com/ECP-WarpX/WarpX/_build/results?buildId=19698&view=results.

EZoni:
Note that #2979 attempted to do the same a couple of years ago, but ran into problems related to cache size. 

We might be close to the limit here too (see screenshot), but let's see if we can go ahead as is for now.

Screenshot from 2024-11-06 15-41-20

WeiqunZhang:
We could use `-g1`.

EZoni:
@WeiqunZhang 

Thanks, using `-g1` seems to have resolved the cache size issue and it should still print out file names and line numbers in backtrace files. I think this PR is ready to review/merge.

WeiqunZhang (approved):

Issue #5444, opened 2024-11-07
Problems with magnetostatic solver test input file


clarkse:
### Discussed in https://github.com/ECP-WarpX/WarpX/discussions/5425

<div type='discussions-op-text'>

<sup>Originally posted by **BCatGA** October 29, 2024</sup>
Hi,

I would like to use the "warpx_magnetostatic=True" option for the Electrostatic solver so I have been running the WarpX Test input file "inputs_test_rz_magnetostatic_eb_picmi.py" to figure out how it works. In this RZ simulation, there is an electron beam of radius 0.1 m going down a pipe (embedded boundary) of radius 0.2 m. The domain is rmax=0.25 m X zmax=1 m, and there are 128 cells in the both the radial and axial directions. The input file is setup to run the simulation for only 1 time step, so it can calculate the  electric and magnetic fields.

When I run the input file with Python the simulation converges. But I found that convergence is very sensitive to changes in the pipe radius (say, from 0.2 to 0.21 m) when "warpx_magnetostatic=True", but not when "warpx_magnetostatic=False". I figured out that this sensitivity goes away when I make the cells square (reducing nr to 32 for rmax = 0.25). In that case, I can put the pipe radius right on the domain boundary and the simulation will converge, as long as I make the boundary conditions at rmax the same as the pipe boundary condition (dirichlet,1V). 

But that begs the question, why do I need the pipe as an embedded boundary? Why not just remove it and treat the outer boundary as the pipe by setting dirichlet boundary conditions with a constant voltage? However, when I do that, the multigrid solver converges but I get multiple segmentation faults and the simulation crashes. And again, if I set "warpx_magnetostatic=False" everything runs fine.

So, my questions are:
1) Why is the convergence of the multi-grid solver so sensitive to the pipe radius when the cells aren't square?
2) Why does the simulation generate segmentation faults if the pipe/embedded boundary is removed?
3) Why do these problems occur when warpx_magnetostatic is True but not when it is False?

Thanks for your help,
Brian
</div>

I have reproduced this problem and am converting it to an issue.

RemiLehe:
@clarkse 
From offline discussions with @roelof-groenewald @WeiqunZhang, it seems that https://github.com/AMReX-Codes/amrex/pull/4226 fixes the fact that the magnetostatic solver did not work when EB where not defined. Would it make sense to revert this part of your PR:
[...]

PR #5446, opened 2024-11-08
Fixing some bugs that lead to non-convergence. Relaxing tolerance whi…

clarkse:
…ch is currently hard coded, and adding assert to force EB being enabled in order to avoid a seg fault in AMReX when computing the gradient solution.

This PR partially addresses https://github.com/ECP-WarpX/WarpX/issues/5444. This PR adds semi-coarsening in 3D and then adds an assert to keep the magnetostatic solver from being run without an EB. This is required since in AMReX MLMG->getGradSolution will segfault when not using an EB. 

It should also be noted that https://github.com/ECP-WarpX/WarpX/pull/5175 will use a different scheme around the embedded boundaries to compute gradients, and will likely mitigate these issues.

A work around in RZ to use the outer edge is to enable the embedded boundary and set the boundary radius larger than the outer grid radius. This works like it would without an embedded boundary and can be used until either the refactor or the bugfix in AMReX for getGradSolution.

### DIFF
--- a/Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
+++ b/Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
@@ -68,8 +68,13 @@ WarpX::ComputeMagnetostaticField()
     WARPX_ALWAYS_ASSERT_WITH_MESSAGE(this->max_level == 0,
         "Magnetostatic solver not implemented with mesh refinement.");
 
+#if defined(AMREX_USE_EB)
     WARPX_ALWAYS_ASSERT_WITH_MESSAGE(EB::enabled(),
-        "Magnetostatic Solver currently requires an embedded boundary.");
+        "Magnetostatic Solver currently requires an embedded boundary to be installed for "
+        "compatibility with AMReX when compiling with EB support. "
+        "Current workaround is to install an EB outside of domain or recompile with EB support off."
+        "Workaround for https://github.com/AMReX-Codes/amrex/issues/4223");
+#endif
 
     AddMagnetostaticFieldLabFrame();
 }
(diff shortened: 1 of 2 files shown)


Source/FieldSolver/MagnetostaticSolver/MagnetostaticSolver.cpp
```diff
@@ -65,7 +65,11 @@ WarpX::ComputeMagnetostaticField()
     // Fields have been reset in Electrostatic solver for this time step, these fields
     // are added into the B fields after electrostatic solve
 
-    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(this->max_level == 0, "Magnetostatic solver not implemented with mesh refinement.");
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(this->max_level == 0,
+        "Magnetostatic solver not implemented with mesh refinement.");
+
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(EB::enabled(),
+        "Magnetostatic Solver currently requires an embedded boundary.");
```
roelof-groenewald: `EB::enabled()` only returns `true` if an embedded boundary is actually installed, right? But this fix really only needs EB support to be on when compiling WarpX (which will eventually be enforced). In the meantime though I think this issue would be better fixed by doing:
```suggestion
#ifndef AMREX_USE_EB
    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(false,
        "Magnetostatic Solver currently requires an embedded boundary.");
#endif
```

I know we are trying to move away from using the compiler directives but I think this special circumstance justifies it since I imagine some users are using the magnetostatic solver without an EB installed.

clarkse: I will double check if this impacts 3D cases as well, but I had to actually install an EB and set the boundary outside of the domain to keep it from crashing in AMReX. So I would need EB::enabled() to be true. I guess the message needs to be more verbose though to explain it better.

clarkse: I have updated the assert to reflect that the solver works when EB support is not compiled, but when compiled with EB there are functions in AMReX that have not been updated to work when EB support is compiled but not installed.

Please see AMReX Issue https://github.com/AMReX-Codes/amrex/issues/4223

roelof-groenewald (approved):
Discussed on Slack - currently the issue raised comes up whenever EB support is ON during compile but an EB is not installed.

PR #5447, opened 2024-11-09
[WIP] CTest: split checksum analysis from test analysis, expose arguments

EZoni:
Prototype of implementation to see if this can achieve goals such as:
- run test and analysis (physics, numerics) locally without worrying about checksums
- avoid duplicate code for default regression analysis with custom parameters (e.g., tolerance, output format)
- minimize work needed to implement checksum regression analysis for new tests (with this PR, one-line command)

To-do:
- [ ] Update documentation
- [ ] Prototype example of how to use/extend this in order to enable multiple tolerances (e.g., single portable, double portable, double machine precision)


EZoni:
Closed by mistake, but cannot reopen due to force-pushing. Will open a new PR later.

PR #5451, opened 2024-11-11
Fix PEC-Insulator boundary condition with staggering

dpgrote:
When calculating the location of the fields to determine whether the field is on a PEC or insulator boundary, add the appropriate shift to account for the staggering of the fields.


Examples/Tests/pec/inputs_test_2d_pec_field_insulator
```diff
@@ -15,7 +15,7 @@ geometry.prob_hi = 1.e-2  3.e-2
 
 # Boundary condition
 boundary.field_lo = neumann      periodic
-boundary.field_hi = PECInsulator periodic
+boundary.field_hi = pec_insulator periodic
```
dpgrote, line 18: This is an incidental fix, making the name consistent with what is in the documentation.

Source/BoundaryConditions/PEC_Insulator.cpp
```diff
@@ -471,10 +474,13 @@ PEC_Insulator::ApplyPEC_InsulatortoField (
...
-                amrex::Real const x = (AMREX_SPACEDIM > 1 ? xyzmin_y.x + (iv[0] - lo_y[0])*dx[0] : 0._rt);
-                amrex::Real const y = (AMREX_SPACEDIM == 3 ? xyzmin_y.y + (iv[1] - lo_y[1])*dx[1] : 0._rt);
+                amrex::Real const shiftx = (Fy_nodal[0] ? 0._rt : 0.5_rt);
+                amrex::Real const x = (AMREX_SPACEDIM > 1 ? xyzmin_y.x + (iv[0] - lo_y[0] + shiftx)*dx[0] : 0._rt);
+                amrex::Real const shifty = (Fy_nodal[1] ? 0._rt : 0.5_rt);
+                amrex::Real const y = (AMREX_SPACEDIM == 3 ? xyzmin_y.y + (iv[1] - lo_y[1] + shifty)*dx[1] : 0._rt);
 #if (AMREX_SPACEDIM > 1)
-                amrex::Real const z = xyzmin_y.z + (iv[WARPX_ZINDEX] - lo_y[WARPX_ZINDEX])*dx[2];
+                amrex::Real const shiftz = (Fy_nodal[2] ? 0._rt : 0.5_rt);
+                amrex::Real const z = xyzmin_y.z + (iv[WARPX_ZINDEX] - lo_y[WARPX_ZINDEX] + shiftz)*dx[2];
```
JustinRayAngus, line 483: Fy_nodal[2] should not be a valid call for AMREX_SPACEDIM < 3. I think 2 should be replaced with WARPX_ZINDEX.

JustinRayAngus: same for Fz_nodal[2] on line 530 and Fx_nodal[2] 434.

JustinRayAngus (approved):
Just need to update the checksum values.

PR #5456, opened 2024-11-13
CTest: split checksum analysis from test analysis, expose arguments

EZoni:
Prototype of implementation to see if this can achieve goals such as:
1. run test and analysis locally (possibly without CTest) without worrying about checksums
2. avoid duplicate code for default regression analysis with custom parameters (e.g., tolerance, output format)
3. minimize work needed to implement checksum regression analysis for new tests

This PR replaces #5447, see https://github.com/ECP-WarpX/WarpX/pull/5447#issuecomment-2471636621.

## Old usage
1. Add this to the test analysis script:
```python
import os
import sys
...
...
sys.path.insert(1, "../../../../warpx/Regression/Checksum/")
from checksumAPI import evaluate_checksum
...
...
# compare checksums	
evaluate_checksum(	
    test_name=os.path.split(os.getcwd())[1],	
    output_file=sys.argv[1],	
)	
```
2. Add this to the CMakeLists.txt file:
```cmake
add_warpx_test(
    test_1d_laser_acceleration_fluid_boosted  # name
    1  # dims
    2  # nprocs
    inputs_test_1d_laser_acceleration_fluid_boosted  # inputs
    analysis_1d_fluid_boosted.py  # analysis
    diags/diag1000001  # output
    OFF  # dependency
)
```

## New usage
1. Create a soft link to the default regression analysis script from the test directory (we already do this when there isn't a custom analysis script, see comment in the follow-up section  below):
```console
ln -s ../../analysis_default_regression.py analysis_default_regression.py
```

2. Add this to the CMakeLists.txt file:
```cmake
add_warpx_test(
    test_1d_laser_acceleration_fluid_boosted  # name
    1  # dims
    2  # nprocs
    inputs_test_1d_laser_acceleration_fluid_boosted  # inputs
    "analysis_1d_fluid_boosted.py diags/diag1000001"  # analysis
    "analysis_default_regression.py --path diags/diag1000001"  # checksum
    OFF  # dependency
)
```

## Notes
- The updated default regression analysis script has the following usage:
```console
usage: analysis_default_regression.py [-h] [--path PATH] [--rtol RTOL] [--skip-fields] [--skip-particles]
options:
  -h, --help        show this help message and exit
  --path PATH       path to output file(s)
  --rtol RTOL       relative tolerance to compare checksums
  --skip-fields     skip fields when comparing checksums
  --skip-particles  skip particles when comparing checksums
```

- The checksum files that changed (as opposed to the ones that were added from scratch) changed because the corresponding analysis scripts were not performing any checksum analysis. The files had been added but they were not used by the test analysis, hence they were outdated.

## To-do
- [x] Add missing checksum files or update existing checksum files that were not used
- [x] Update documentation

## Follow-up
- Improve documentation even more:
  - do we need all user-facing features of `checksumAPI` or are some obsolete?
  - can we merge documentation for testing and checksums into one section?
- Check/fix custom tolerances (`git grep "# checksum" Examples/ | grep "rtol"`)
- Add logic to reset tolerances based on environment variables (e.g., to run all tests in single-precision)
- Update custom test analysis scripts so that they do not need to take the output path as argument (would be difficult to maintain as a convention, though, as these scripts are up to each PR author)
- Try (again) to make it work with direct path rather than soft link (did not work when we tried in the past)


EZoni:
Tests that currently use a custom tolerance for checksums:
```console
$ git grep "# checksum" Examples/ | grep "rtol"
Examples/Physics_applications/capacitive_discharge/CMakeLists.txt:51:    "analysis_default_regression.py --path diags/diag1000050 --plotfile --rtol 5e-3"  # checksum
Examples/Physics_applications/uniform_plasma/CMakeLists.txt:30:    "analysis_default_regression.py --path diags/diag1000010 --plotfile --rtol 1e-12"  # checksum
Examples/Tests/diff_lumi_diag/CMakeLists.txt:10:    "analysis_default_regression.py --path diags/diag1000080 --plotfile --rtol 1e-2"  # checksum
Examples/Tests/diff_lumi_diag/CMakeLists.txt:20:    "analysis_default_regression.py --path diags/diag1000080 --plotfile --rtol 1e-2"  # checksum
Examples/Tests/initial_plasma_profile/CMakeLists.txt:10:    "analysis_default_regression.py --path diags/diag1000001 --plotfile --skip-particles --rtol 1e-4"  # checksum
Examples/Tests/nci_psatd_stability/CMakeLists.txt:11:        "analysis_default_regression.py --path diags/diag1000400 --plotfile --rtol 1e-8"  # checksum
Examples/Tests/nci_psatd_stability/CMakeLists.txt:23:        "analysis_default_regression.py --path diags/diag1000400 --plotfile --rtol 1e-8"  # checksum
Examples/Tests/nci_psatd_stability/CMakeLists.txt:47:        "analysis_default_regression.py --path diags/diag1000400 --plotfile --rtol 1e-8"  # checksum
Examples/Tests/nci_psatd_stability/CMakeLists.txt:59:        "analysis_default_regression.py --path diags/diag1000400 --plotfile --rtol 1e-8"  # checksum
Examples/Tests/nci_psatd_stability/CMakeLists.txt:71:        "analysis_default_regression.py --path diags/diag1000400 --plotfile --rtol 1e-8"  # checksum
Examples/Tests/nci_psatd_stability/CMakeLists.txt:95:        "analysis_default_regression.py --path diags/diag1000160 --plotfile --rtol 1e-8"  # checksum
Examples/Tests/nci_psatd_stability/CMakeLists.txt:107:        "analysis_default_regression.py --path diags/diag1000160 --plotfile --rtol 1e-8"  # checksum
Examples/Tests/nci_psatd_stability/CMakeLists.txt:119:        "analysis_default_regression.py --path diags/diag1000300 --plotfile --rtol 1e-8"  # checksum
Examples/Tests/nci_psatd_stability/CMakeLists.txt:131:        "analysis_default_regression.py --path diags/diag1000300 --plotfile --rtol 1e-8"  # checksum
Examples/Tests/nci_psatd_stability/CMakeLists.txt:143:        "analysis_default_regression.py --path diags/diag1000300 --plotfile --rtol 1e-8"  # checksum
Examples/Tests/nci_psatd_stability/CMakeLists.txt:167:        "analysis_default_regression.py --path diags/diag1000400 --plotfile --rtol 1e-8"  # checksum
Examples/Tests/nci_psatd_stability/CMakeLists.txt:179:        "analysis_default_regression.py --path diags/diag1000400 --plotfile --rtol 1e-8"  # checksum
Examples/Tests/nci_psatd_stability/CMakeLists.txt:191:        "analysis_default_regression.py --path diags/diag1000400 --plotfile --rtol 1e-8"  # checksum
Examples/Tests/ohm_solver_em_modes/CMakeLists.txt:20:    "analysis_default_regression.py --path diags/diag1000100 --plotfile --rtol 1e-6"  # checksum
Examples/Tests/open_bc_poisson_solver/CMakeLists.txt:11:        "analysis_default_regression.py --path diags/diag1000001 --plotfile --rtol 1e-2"  # checksum
Examples/Tests/open_bc_poisson_solver/CMakeLists.txt:23:        "analysis_default_regression.py --path diags/diag1000001 --plotfile --rtol 1e-2"  # checksum
Examples/Tests/particle_fields_diags/CMakeLists.txt:21:#    "analysis_default_regression.py --path diags/diag1000200 --plotfile --rtol 1e-3"  # checksum
```

Examples/Tests/collision/CMakeLists.txt
```diff
@@ -56,7 +56,7 @@ add_warpx_test(
     RZ  # dims
     1  # nprocs
     inputs_test_rz_collision  # inputs
-    analysis_collision_rz.py  # analysis
-    diags/diag1000150  # output
+    "analysis_collision_rz.py diags/diag1000150"  # analysis
+    "analysis_default_regression.py --path diags/diag1000150 --plotfile --skip-particle"  # checksum
```
dpgrote: Typo? `--skip-particles`

Examples/CMakeLists.txt
```diff
@@ -72,14 +72,25 @@ function(add_warpx_test
...
+    if(ANALYSIS_LIST_LENGTH GREATER 1)
+        list(SUBLIST ANALYSIS_LIST 1 -1 ANALYSIS_ARGS)
+        list(JOIN ANALYSIS_ARGS " " ANALYSIS_ARGS)
+    else()
+        set(ANALYSIS_ARGS "")
+    endif()
+
+    # get checksum script and optional command-line arguments
+    separate_arguments(CHECKSUM_LIST UNIX_COMMAND "${checksum}")
+    list(GET CHECKSUM_LIST 0 CHECKSUM_FILE)
```
dpgrote, line 85: I'm thinking about ways to make this simpler. Instead of having a soft link each place where the default analysis script is used, can this check if `CHECKSUM_FILE` is equal to `analysis_default_regression.py` and put in the appropriate direct path? Perhaps use some keyword like "DEFAULT" instead of the script name?

Docs/source/developers/testing.rst
```diff
@@ -196,6 +196,27 @@ The shared input parameters can be collected in a "base" input file that can be
...
+
+Here is the help message of the default regression analysis script, including usage and list of available options and arguments:
+
+  .. code-block:: bash
+
+       usage: analysis_default_regression.py [-h] [--path PATH] (--plotfile | --openpmd) [--rtol RTOL] [--skip-fields]
+                                             [--skip-particles]
+       options:
+         -h, --help        show this help message and exit
+         --path PATH       path to output file(s)
```
dpgrote, line 214: Can automatic discovery be used here instead of relying on the arguments? The checksum script could determine whether the output is a plotfile or openpmd. Also, the checksum would presumably always be done on the last diagnostic file and this could be automatically found. Then the path here would only be the top level diag directory.

Examples/CMakeLists.txt
EZoni: I guess that would be nice, but when Axel and I set up ctest in the first place we did not manage to make it work with the direct paths (there were several attempts and then we turned to soft links). Especially, I think Axel found issues with running ctest from within an IDE. We could rivisit once he is back, but I would not try this here.

dpgrote: Thinking further, instead of writing out the full command to do the benchmark check, could the arguments to the analysis script be added as arguments to this function? In most cases, default values could be used so most `CMakeLists.txt` files wouldn't need any extra arguments (especially with the automatic discovery I suggest above). Then the `rtol` and `skip-fields` and `skip-particles` could be added via variable arguments pairs, like this
```
rtol 1.e-5
```

Examples/Tests/collision/CMakeLists.txt
EZoni: Thanks, fixed in [b57c6f1](https://github.com/ECP-WarpX/WarpX/pull/5456/commits/b57c6f19c3c3bdb66f2d180ead16d20a7b277285).

Examples/CMakeLists.txt
EZoni: I think this was another design choice that we made with Axel initially. 

We wanted to keep the number of arguments of `add_warpx_test` as little as possible, especially given that we cannot use keywords like in Python (e.g., `add_warpx_test(..., rtol=1e-6, ...)`). 

Another issue I see is that we would be treating the arguments to the custom analysis script (which cannot be predicted in advance) differently than the arguments to the default analysis script. It might be confusing to see a `rtol` argument and not know for sure whether it refers to the custom analysis script or to the default one, given that the interface of the custom analysis script is up to the developer who added it.

If you feel strongly about this, I would recommend that we wait until Axel is back and discuss with him further.

Docs/source/developers/testing.rst
EZoni: Yes, automating some of this was something I was hoping to do as well. Would you have a suggestion on how to automatically distinguish between plotfile and openpmd output?

Examples/CMakeLists.txt
EZoni: Another comment is that we wanted to display commands that developers could copy/paste to try things locally, whenever that is needed (e.g., for debugging purposes). This came out of discussions with @aeriforme, and I think it would be valuable in general, as opposed to passing arguments to `add_warpx_test` when they are actually arguments read/used by the analysis scripts.

Docs/source/developers/testing.rst
dpgrote: One way would be to check for the `Header` file which would signify a plotfile diagnostic.

Examples/CMakeLists.txt
dpgrote: I don't feel strongly about this, just brain storming ideas. I see that having the command to do the benchmark check written out is convenient for debugging.

BTW, you can imitate keyword arguments though it's a bit kludgy. You can use variable arguments and scan through them. If one has a value of the argument name, then take the next argument as the value.

Docs/source/developers/testing.rst
EZoni: I pushed something to see if this works:
```python
    # set args.format automatically
    try:
        yt.load(args.path)
    except Exception:
        try:
            OpenPMDTimeSeries(args.path)
        except Exception:
            print("Could not open the file as a plotfile or an openPMD time series")
        else:
            args.format = "openpmd"
    else:
        args.format = "plotfile"
```

dpgrote: This looks good.

Examples/analysis_default_regression.py
```diff
@@ -1,25 +1,88 @@
...
+        "--skip-particles",
+        help="skip particles when comparing checksums",
+        action="store_true",
+        dest="skip_particles",
     )
+    # parse arguments
+    args = parser.parse_args()
+    # set args.format automatically
+    try:
+        ds = yt.load(args.path)
```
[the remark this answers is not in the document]
EZoni: Resolved in [fa0c1c6](https://github.com/ECP-WarpX/WarpX/pull/5456/commits/fa0c1c62d72fcb3dbaa794280a8290512bdee12d).

Examples/analysis_default_regression.py
```diff
@@ -1,25 +1,88 @@
...
+    except Exception:
+        try:
+            ts = OpenPMDTimeSeries(args.path)
```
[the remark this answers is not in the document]
EZoni: Resolved in [fa0c1c6](https://github.com/ECP-WarpX/WarpX/pull/5456/commits/fa0c1c62d72fcb3dbaa794280a8290512bdee12d).

EZoni:
@dpgrote 

Hi Dave, as you saw I updated the default analysis script to detect the output file format automatically as you suggested. Personally I would postpone further automation and/or interface changes to follow-up work, due to the comments following https://github.com/ECP-WarpX/WarpX/pull/5456#discussion_r1884410509. Please feel free to leave another review.

EZoni:
@aeriforme 

Thanks to @dpgrote's comments, we made this even easier to use, with automatic detection of the output file format, which does not have to be passed as command line argument anymore. See updated PR description for the latest usage.

EZoni:
@roelof-groenewald @RemiLehe 

I added you as reviewers to this PR as well, since this is now at a mature stage and could be approved and merged soon. Please feel free to have a look, especially at the proposed feature rather then the individual code changes, and provide feedback, if any.

EZoni:
To-do: rebase after #5089 is merged (new CI test added).

aeriforme:
I see that it's possible to turn off the analysis with the `OFF` keyword. 
I think this information is missing in the documentation. 

What would happen if one used `OFF` instead of the checksums script? 
Can we raise an error in that case, so we can ensure that all tests are checksummed?

EZoni:
I think the behavior is the same as for the analysis script (i.e., it's skipped).

I agree that it should be documented, I will update the documentation asap.

I think it's a good idea to throw an error if the checksum analysis is skipped. I just have to double check that this is fine for all existing tests, but I guess it is.

EZoni:
@aeriforme 

Regarding this,

> What would happen if one used OFF instead of the checksums script?
Can we raise an error in that case, so we can ensure that all tests are checksummed?

I think we cannot throw an error when the checksum command is set to OFF because we use that feature specifically for tests that need a preparatory step, such as the LASY tests ("output" here is "checksum" in this PR):
https://github.com/ECP-WarpX/WarpX/blob/2cdcb77668bba4a08b36445f403d77080edeb9c3/Examples/Tests/laser_injection_from_file/CMakeLists.txt#L64-L82

We did not manage to merge the preparatory step and the actual test into one single test when we set up ctest initially, I can't remember why but we ended up with this solution after trying to fix the LASY tests for a while, so I would not change this now.

EZoni:
@aeriforme 

Regarding this,

> I see that it's possible to turn off the analysis with the OFF keyword.
I think this information is missing in the documentation.

I updated the documentation in [874d362](https://github.com/ECP-WarpX/WarpX/pull/5456/commits/874d362f32dc11a1392f157bb8b62025a0f52398).

EZoni:
After discussing with @aeriforme, we added the particle data to the checksums of two RZ scraping tests (not clear why the particle data had been skipped when those tests were added in the first place). 

To be double checked in the future if the first two values will raise issues due to the small order of magnitude:
```
    "particle_momentum_x": 8.802233511708275e-20,
    "particle_momentum_y": 8.865573181381068e-20,
```

aeriforme (approved):
We discussed offline.
It looks good to me now!

PR #5461, opened 2024-11-14
Update license to explicitly list LLNL

dpgrote:
I noticed this oversight in the LICENSE.txt file. Is there some official way of fixing this or is this PR enough?


EZoni (approved):
I think this is enough, but I would ask @jlvay and @RemiLehe to double check if something else needs to be done.

jlvay:
Looks good to me.

jlvay (approved):
Looks good to me.

PR #5464, opened 2024-11-15
Add diagnostic file output for the implicit solvers

dpgrote:
This PR adds the option to write out information about the implicit solvers to a file during the simulation.

Note that this PR uses `std::filesystem` directly. This does not work easily for older gcc compilers, from before version 9.1. As a side effect, this PR removes builds using the older gcc and bumps up the comment in the dependencies documentation. This should not be an issue since all of the installs on HPC systems use newer versions.


JustinRayAngus (approved):
Thanks Dave.

Looks great.

PR #5465, opened 2024-11-15
Docs: update list of TC members

EZoni:
I think we forgot to update the list of TC members available in the GitHub repository and in the online documentation.


JustinRayAngus (commented):
Thanks for this PR. My name is correct.

RemiLehe (approved):

PR #5469, opened 2024-11-18
[WIP] CI: Fix Skipped Required Workflows, Clean Up

EZoni:
Work-in-progress PR to try out various solutions to handle skipped but required CI workflows properly, e.g., when only documentation is changed (trivial one-word change in this PR).

Follow-up to #5387. Originally tried some of these in #5386.


EZoni:
@WeiqunZhang 

With regard to https://github.com/ECP-WarpX/WarpX/pull/5387#issuecomment-2427962637 as well as what was done in #5387 and https://github.com/AMReX-Codes/amrex/pull/4197, I tried several approaches including (i) adding dependencies between workflows through `workflow_run`, (ii) using reusable workflows through `workflow_call`, and others, but all failed at working around GitHub's handling of required checks that are skipped.

The approach implemented here, based on a custom bash script that is called from within each workflow before running the expected checks, seems to be the only one that I got to make work and that doesn't rely on third-party code (hence maybe easier to maintain, debug, etc.). 

Let me know if this seems like a good enough solution to you. 

Note that for the clang tidy checks, organized with a strategy matrix approach, we need to apply the `if` conditions at each step of the job, rather than at the job level, otherwise each workflow defined by the matrix remains pending.

RemiLehe (dismissed):

.github/workflows/scripts/check_diff.sh
```diff
@@ -0,0 +1,42 @@
...
+# Save output of git diff to inspect files changed
+git diff --name-only --diff-filter=ACMRTUXB origin/${base_ref}..fork/${head_ref} > check_diff.txt
+echo "Files changed:"
+cat check_diff.txt
+
+# Set paths to ignore
+paths_ignore=()
+paths_ignore+=("Docs/")
+paths_ignore+=(".github/")
+paths_ignore+=(".azure-pipelines.yml")
```
EZoni: These two paths patterns are here for testing purposes only (since this PR modifies also files that match those patterns). They should be removed before merging the PR.

Docs/source/index.rst
```diff
@@ -24,7 +24,7 @@ Contact us
 
 If you are starting using WarpX, or if you have a user question, please pop in our `discussions page <https://github.com/ECP-WarpX/WarpX/discussions>`__ and get in touch with the community.
 
-The `WarpX GitHub repo <https://github.com/ECP-WarpX/WarpX>`__ is the main communication platform.
+The `WarpX GitHub repository <https://github.com/ECP-WarpX/WarpX>`__ is the main communication platform.
```
EZoni: We can keep this change but it was added initially for testing purposes only.

.azure-pipelines.yml
```diff
@@ -13,6 +13,8 @@ pr:
   paths:
     exclude:
     - Docs
+    - .github/workflows/**
+    - .azure-pipelines.yml
```
EZoni: These two paths patterns are here for testing purposes only (since this PR modifies also files that match those patterns). They should be removed before merging the PR.

EZoni:
Please see the checks for [488add9](https://github.com/ECP-WarpX/WarpX/pull/5469/commits/488add99a62afce72c723603a47da74bfcfed770) or [6636191](https://github.com/ECP-WarpX/WarpX/pull/5469/commits/6636191a818717ab280f2b9c76d56990d1df05c4) to verify how this works when all checks are skipped.

EZoni:
Not clear how the new `skip_checks` job behaves when a given workflow is triggered by a `push` event (e.g., when merging a PR into the main branch), rather than a `pull_request` event (e.g., when pushing commits to an open PR, like here)...

.github/workflows/scripts/check_diff.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# Save output of git diff to inspect files changed
+git diff --name-only --diff-filter=ACMRTUXB origin/${base_ref}..fork/${head_ref} > check_diff.txt
+echo "Files changed:"
+cat check_diff.txt
+
+# Set paths to ignore (please test grep command below when adding new patterns)
+paths_ignore=()
+paths_ignore+=("Docs/")
+paths_ignore+=(".github/")
```
EZoni, line 27: This
```bash
paths_ignore+=(".github/")
```
was added for testing purposes and should be removed before the PR is merged, if approved.

EZoni:
Replaced by the alternative solution implemented in #5889.

PR #5473, opened 2024-11-19
Add multifab method that takes an int for the direction

dpgrote:
This adds a new way to call the `multifab` routine from Python, taking an integer for the direction instead of requiring an instance of the `Direction` class.

This also cleans up the documentation some for the other versions of the routine.


EZoni:
@dpgrote 
You might need to rebase after #5474 to fix the Clang sanitizer checks.

EZoni (approved):

PR #5475, opened 2024-11-20
Add FieldPoyntingFlux reduced diagnostic

dpgrote:
This adds a reduced diagnostic that calculates the Poynting flux on the surfaces of the domain, providing the power flow into and out of the domain. This also includes the time integrated data.

When using the implicit evolve scheme, to get the energy accounting correct, the flux needs to be calculated at the mid step. For this reason, the `ComputeDiagsMidStep` was added which is called directly at the appropriate times.

Because of the time integration, there are two main differences of this reduced diagnostic compared to the others. The first is that it is calculated every time step in order to get the full resolution in time. The intervals parameter still controls how often the diagnostic data is written out. The second is that a facility is added to write out the values of the time integration to a file when a checkpoint is made, so on a restart the integration can continue with the previous values. The facility was written in a general way so that other reduced diagnostics can also do this.

The CI test using the implicit solver is dependent on PR #5498 and PR #5489.


Examples/Tests/pec/inputs_test_2d_pec_field_insulator_implicit
```diff
@@ -0,0 +1,71 @@
...
+boundary.field_lo = neumann      periodic
+boundary.field_hi = pec_insulator periodic
+
+insulator.area_x_hi(y,z) = (2.25e-2 <= z and z <= 2.75e-2)
+insulator.By_x_hi(y,z,t) = min(t/1.0e-12,1)*1.e1*3.3e-4
+
+warpx.serialize_initial_conditions = 1
+
+# Implicit setup
+warpx.const_dt = 7.37079480234276e-13/2.
```
JustinRayAngus, line 28: This time step appears to be 1/2 the CFL value. Perhaps make a comment.

Source/Diagnostics/ReducedDiags/FieldPoyntingFlux.H
```diff
@@ -0,0 +1,62 @@
...
+#ifndef WARPX_DIAGNOSTICS_REDUCEDDIAGS_FIELDPOYTINGFLUX_H_
+#define WARPX_DIAGNOSTICS_REDUCEDDIAGS_FIELDPOYTINGFLUX_H_
+
+#include "ReducedDiags.H"
+
+#include <string>
+
+/**
+ * \brief This class mainly contains a function that computes the field Poynting flux.
+ */
```
JustinRayAngus, line 18: Perhaps add the Poynting flux formula here.

Source/Diagnostics/ReducedDiags/FieldPoyntingFlux.cpp
```diff
@@ -0,0 +1,333 @@
...
+        int const face_dir = face().coordDir();
+
+        if (face().isHigh() && WarpX::field_boundary_hi[face_dir] == FieldBoundaryType::Periodic) {
+            // For upper periodic boundaries, copy the lower value instead of regenerating it.
+            int const iu = int(face());
+            int const il = int(face().flip());
+            m_data[iu] = -m_data[il];
+            m_data[iu + 2*AMREX_SPACEDIM] = -m_data[il + 2*AMREX_SPACEDIM];
+            continue;
+        }
```
JustinRayAngus, line 180: Perhaps another option to consider would be to skip writing for periodic directions all together.

JustinRayAngus:
This looks good to me, pending the minor comments I made.

After a closer look at the 2D CI test, it seems that FieldBoundaryType::Neumann does nothing for the electromagnetic solvers. As we have discussed, this can be addressed in a future PR.

Source/Diagnostics/ReducedDiags/FieldPoyntingFlux.cpp
dpgrote: I would rather leave it in. It could be a useful diagnostic to keep track of flows in the system. Also, skipping the periodic directions would make the format of the output file be dependent on the boundary conditions which would be problematic when reading it.

Examples/Tests/pec/inputs_test_2d_pec_field_insulator_implicit
dpgrote: Done

Source/Diagnostics/ReducedDiags/FieldPoyntingFlux.H
dpgrote: Done

JustinRayAngus (approved):
Looks good to me.

PR #5474, opened 2024-11-20
CI: fix Clang UB sanitizer, disable Clang thread sanitizer

EZoni:
I started seeing this error from the Clang sanitizer workflow in several PRs:
```console
E: The repository 'http://archive.ubuntu.com/ubuntu mantic Release' does not have a Release file.
E: The repository 'http://archive.ubuntu.com/ubuntu mantic-updates Release' does not have a Release file.
E: The repository 'http://archive.ubuntu.com/ubuntu mantic-backports Release' does not have a Release file.
E: The repository 'http://security.ubuntu.com/ubuntu mantic-security Release' does not have a Release file.
```

The Ubuntu 23.10 container was introduced in #5181 when `ubuntu-24.04` was pre-release as GitHub-hosted runner.

I think we can try to move to `ubuntu-24.04` now and see if this is enough to fix the issue above.

PR #5476, opened 2024-11-20
Removing asserts now that AMReX has been updated.

### LINKED ISSUE #5444: Problems with magnetostatic solver test input file

clarkse:
https://github.com/AMReX-Codes/amrex/pull/4226 fixes the issue that the asserts protected against. They have been removed.

This should close out https://github.com/ECP-WarpX/WarpX/issues/5444


roelof-groenewald (commented):
Looks good. Thanks @clarkse!
Once #5474 is merged we can rebase and merge.

EZoni:
@clarkse 

Thanks for cleaning this up! #5474 is merged now, so we would need to rebase on `development`, as suggested by @roelof-groenewald. Do you want to do that or shall @roelof-groenewald or I go ahead and finalize this PR?

roelof-groenewald (approved):
Thanks @clarkse!

PR #5483, opened 2024-11-24
Split Bfield advance in two for SemiImplicit_EM evolve

JustinRayAngus:
This PR splits the BField advance for the SemiImplicit_EM evolve scheme into two half dt advances, consistent with the explicit evolve scheme. This makes it such that the B field at write time is at the same instance in time as the electric field, as stated in the documentation.


dpgrote (approved):
Looks good!

PR #5485, opened 2024-11-25
Setting laser particle positions to be time-centered for implicit solvers

JustinRayAngus:
The current deposition schemes expect the particle positions to be time-centered when using the implicit solvers. This is not currently the case for the laser particles. This PR fixes this issue.

It should be commented that there is really no need for the current deposition schemes to expect the particle positions to be time centered for the implicit solvers. This is somewhat of a legacy thing from PICNIC. This can and should be changed in a future PR.

Here are results of Ey generated from the analysis script for the 1D CI test using the semi-implicit method:
plt_Ey

Here are results of Ey generated from the analysis script for the 2D CI test using the semi-implicit method:
plt_Ey

Here are results generated from the analysis script for the 3D CI test using the semi-implicit method:
laser_analysis


dpgrote (approved):
Looks good to me!

PR #5488, opened 2024-11-26
Fix the time for the insulator boundary field evaluation

dpgrote:
In the application of the insulator boundary conditions, the field in the insulator is specified using an expression that can be a function of time. This PR updates the calls to that routine to pass in the correct time, for example the time at the half step or the end of the step. This requires passing the time down the call chain to the routines that apply the insulator boundaries which makes this a fairly sizable set of changes.

Note that in some places the variable names were changed to be more precise about which time is passed in. For example, using `start_time` whenever the time at the start of the step is expected.


Source/FieldSolver/WarpXPushFieldsEM.cpp
```diff
@@ -714,9 +714,10 @@ WarpX::PSATDScaleAverageFields (const amrex::Real scale_factor)
 #endif // WARPX_USE_FFT
 
 void
-WarpX::PushPSATD ()
+WarpX::PushPSATD (amrex::Real start_time)
```
EZoni, line 717: clang-tidy is complaining about the fact that we use `start_time` in the definition here, but `cur_time` in the declaration in WarpX.H. We need to use one single name to suppress that warning (which we treat as error), not sure which one you think is most appropriate.

Source/FieldSolver/ImplicitSolvers/ThetaImplicitEM.cpp
```diff
@@ -102,69 +102,67 @@ void ThetaImplicitEM::OneStep ( const amrex::Real  a_time,
...
 
     // Update Bfield_fp owned by WarpX
     ablastr::fields::MultiLevelVectorField const& B_old = m_WarpX->m_fields.get_mr_levels_alldirs(FieldType::B_old, 0);
-    m_WarpX->UpdateMagneticFieldAndApplyBCs( B_old, m_theta*m_dt );
+    m_WarpX->UpdateMagneticFieldAndApplyBCs( B_old, m_theta*m_dt, start_time );
 
 }
 
-void ThetaImplicitEM::FinishFieldUpdate ( amrex::Real  a_new_time )
+void ThetaImplicitEM::FinishFieldUpdate ( amrex::Real end_time )
```
EZoni, line 155: Here too, clang-tidy complains about the fact that we use `end_time` in the definition here and `a_new_time` in the declaration in ThetaImplicitEM.H.

Regression/Checksum/benchmarks_json/test_2d_pec_field_insulator.json
EZoni, line 1: Could you confirm that these changes in the checksums are expected? In particular, do you know which call(s) specifically were using the wrong time value and were thus fixed in this PR, leading to these changes in the checksums?

Source/FieldSolver/ImplicitSolvers/ThetaImplicitEM.cpp
dpgrote: Fixed now. Too many changes to keep track of and the local compilation doesn't check this.

Source/FieldSolver/WarpXPushFieldsEM.cpp
dpgrote: Fixed.

EZoni (approved):

Regression/Checksum/benchmarks_json/test_2d_pec_field_insulator.json
dpgrote: These changes were expected. Previously, in each call to `EvolveB` the time used when setting the boundary conditions was the start of the time step. With this PR, the time is the correct time, at the half time step and end time step for the split calls to `EvolveB` in the explicit solver.

PR #5489, opened 2024-12-02
Fix restart for implicit simulations

### LINKED ISSUE #5482: restart with implicit solver
WarpX cannot perform a restart when using the implicit solvers. 

Here is the error message from trying to restart a 2D simulation:
--- INFO    : Runtime real component x_n was found in the checkpoint file, but
             it has not been added yet. Adding it now.
--- INFO    : Runtime real component z_n was found in the checkpoint file, but
             it has not been added yet. Adding it now.
--- INFO    : Runtime real component ux_n was found in the checkpoint file, but
             it has not been added yet. Adding it now.
--- INFO    : Runtime real component uy_n was found in the checkpoint file, but
             it has not been added yet. Adding it now.
--- INFO    : Runtime real component uz_n was found in the checkpoint file, but
             it has not been added yet. Adding it now.
amrex::Abort::444::ParticleContainer::Restart(): nr not the expected value !!!
SIGABRT

The --INFO lines are produced around line 172 of Source/Diagnostics/ParticleIO.cpp. The implicit solver defines these extra variables at runtime. They are written to the checkpoint file, but they are not needed. At the time the restart file is read, these extra variables have not yet been added to the particle container, so these variables are added in the Restart function in ParticleIO.cpp (should they be added here?). 

The Abort comes from the Restart function called in amrex/Src/Particle/AMReX_ParticleIO.H around line 700. There it reads in the number of components (which is 5 more now than it is normally for a 2D simulation), and it fails because it is expected this value to be equal to 
nrc = NStructReal + NumRealComps() - AMREX_SPACEDIM, which I have verified is 5 time less than what nr is equal to.

This needs to be fixed. Question is how to fix it. Which part of this problem is the one that is non-ideal? @RemiLehe  @WeiqunZhang  @dpgrote 

Ideas?

atmyers:
Fix #5482.

Also, with this PR the temporary implicit particle attributes are not written to the checkpoint file, and also are not communicated when doing a redistribute of the particles. Note that this requires AMReX PR #4264.

A CI test of this restart is included in PR #5475.


atmyers:
Hi @JustinRayAngus - just checking, did you get a chance to verify that this fixes the issue for you?

EZoni:
@atmyers 

Shall we merge `development` to re-run the CI checks and then just merge this? If the CI checks pass, I guess we could fix possible residual bugs in a follow-up PR. What do you think?

dpgrote:
> @atmyers
> 
> Shall we merge `development` to re-run the CI checks and then just merge this? If the CI checks pass, I guess we could fix possible residual bugs in a follow-up PR. What do you think?

I can do the merge of development. I've checked the changes in this PR and it works.

EZoni (approved):

dpgrote (approved):

Source/Diagnostics/FlushFormats/FlushFormatCheckpoint.cpp
```diff
@@ -188,31 +188,49 @@ FlushFormatCheckpoint::CheckpointParticles (
...
+        for (auto const& x : runtime_inames) {
+            int const i = x.second + 0;
+            int_names[i] = x.first;
+            write_int_comps[i] = pc->h_redistribute_int_comp[i+AMREX_SPACEDIM];
+        }
 
-        pc->Checkpoint(dir, part_diag.getSpeciesName(), true,
+        pc->Checkpoint(dir, part_diag.getSpeciesName(),
+                       write_real_comps, write_int_comps,
                        real_names, int_names);
```
ax3l, line 234: @JustinRayAngus is experiencing a build issue from this call right now. @dpgrote is taking  a look.

Source/Diagnostics/FlushFormats/FlushFormatCheckpoint.cpp
```diff
@@ -188,31 +188,49 @@ FlushFormatCheckpoint::CheckpointParticles (
...
         auto runtime_inames = pc->getParticleRuntimeiComps();
-        for (auto const& x : runtime_inames) { int_names[x.second+0] = x.first; }
+        for (auto const& x : runtime_inames) {
+            int const i = x.second + 0;
+            int_names[i] = x.first;
+            write_int_comps[i] = pc->h_redistribute_int_comp[i+AMREX_SPACEDIM];
+        }
 
-        pc->Checkpoint(dir, part_diag.getSpeciesName(), true,
+        pc->Checkpoint(dir, part_diag.getSpeciesName(),
```
ax3l, line 232: @JustinRayAngus is experiencing a build issue from this call right now. @dpgrote is taking  a look.

PR #2173, opened 2021-08-09
Gaussian Laser: fix issue related to single precision

lucafedeli88:
`inv_tau2` and `inv_complex_waist_2` can be very large for realistic laser parameters. Indeed, in a recent simulation I had `inv_tau2 = 1.000000062e+28` and  `inv_complex_waist_2 = 4.707382067e+10 + -1.367689789e+11i`. When these parameters were multiplied together to calculate the `stretch_factor`, in single precision the product was `inf - infi`, which finally lead to a crash of WarpX. I've rearranged the calculation so that this issue is much less likely. 

To fix all the issues related to single precision, I think that at some point we should consider performing all the internal calculations using a normalized unit system (while keeping the inputfile and the output in SI units), although I acknowledge that this would require a huge amount of work.

PR #5494, opened 2024-12-04
[mini] Fix stretch factor in Gaussian initialisation w/ spatial and angular chirps

MaxThevenet:
It seems to me that PR https://github.com/ECP-WarpX/WarpX/pull/2173 changed the actual terms of laser initialization in the presence of chirps. This PR proposes to revert back to the previous behavior, which I think is more correct. In this way, I think the advantage of https://github.com/ECP-WarpX/WarpX/pull/2173 regarding single-precision are preserved.


lucafedeli88:
@MaxThevenet , you are totally right! This is a bug that had to be corrected!

lucafedeli88 (approved):
@MaxThevenet , thanks a lot for this bug-fixing! I agree with you!

PR #5498, opened 2024-12-06
Fix FieldEnergy reduced diagnostic

dpgrote:
Fix FieldEnergy reduced diagnostic to account for partial cell volumes on the domain boundaries. For fields that are nodal on the boundary, the integral is only over the fraction of the cell centered about the field value within the domain.

This PR also tidies up the code for the diagnostic.

### DIFF
--- a/Source/Diagnostics/ReducedDiags/FieldEnergy.cpp
+++ b/Source/Diagnostics/ReducedDiags/FieldEnergy.cpp
@@ -159,6 +159,17 @@ FieldEnergy::ComputeNorm2(amrex::MultiFab const& field, int lev)
     amrex::IntVect const domain_lo = matched_domain_box.smallEnd();
     amrex::IntVect const domain_hi = matched_domain_box.bigEnd();
 
+    amrex::IntVect is_periodic;
+    amrex::IntVect half_volume;;
+    for (int idir=0 ; idir < AMREX_SPACEDIM ; idir++) {
+        is_periodic[idir] = (WarpX::field_boundary_hi[idir] == FieldBoundaryType::Periodic);
+        // Only half of the cell volume is within the domain if the field
+        // is nodal but not periodic. For periodic, only the value on the lower
+        // boundary is used, accounting for the half cell at the lower and upper
+        // ends of the domain.
+        half_volume[idir] = (is_nodal[idir] && !is_periodic[idir]);
+    }
+
     amrex::ReduceOps<amrex::ReduceOpSum> reduce_ops;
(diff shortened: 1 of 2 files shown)


Source/Diagnostics/ReducedDiags/FieldEnergy.cpp
```diff
@@ -178,45 +163,63 @@ FieldEnergy::ComputeNorm2RZ(const amrex::MultiFab& field, const int lev)
...
-                return theta_integral*field_arr(i,j,k,n)*field_arr(i,j,k,n)*volume_factor;
+                return field_arr(i,j,k,n)*field_arr(i,j,k,n)*volume_factor(i,j,k,n);
             });
 
     }
 
-    const amrex::Real field_sum = amrex::get<0>(reduce_data.value());
-    const amrex::Real result = MathConst::pi*field_sum;
+    amrex::Real result = amrex::get<0>(reduce_data.value());
+    amrex::ParallelDescriptor::ReduceRealSum(result);
```
roelof-groenewald, line 222: I guess this was generally a bug in the RZ version of `FieldEnergy` since the parallel reduction wasn't done before. Nice to have another bug fixed (besides the main issue fixed in this PR 😄 )

Source/Diagnostics/ReducedDiags/FieldEnergy.cpp
```diff
@@ -178,45 +163,63 @@ FieldEnergy::ComputeNorm2RZ(const amrex::MultiFab& field, const int lev)
...
+        amrex::XDim3 const xyzmin = WarpX::LowerCorner(tilebox, lev, 0._rt);
+        amrex::Real const rmin = xyzmin.x + (is_nodal[0] ? 0._rt : 0.5_rt*dr);
+#endif
 
-        int const ncomp = field.nComp();
+        // On the boundaries, if the grid is nodal, use half of the volume.
+        // This applies to all boundary conditions, and to the overlap of
+        // boxes within the domain.
+        // Previously, the code used MultiFab::norm2, but that does not do
+        // the half-volume scaling for the domain boundaries when not periodic.
```
roelof-groenewald, line 184: Is it worth updating `MultiFab::norm2` in AMReX? Are there use cases in other codes that requires the boundary values to not be scaled by 1/2?

I'll merge this now and if we decide to update AMReX we can go back to using `norm2`.

cc @WeiqunZhang

roelof-groenewald (approved):
Looks good to me!

JustinRayAngus:
Looks good to me Dave.  Thanks.

PR #5499, opened 2024-12-06
Fix the extrapolation for guard cells in the insulator region

dpgrote:
For the nodal guard cells in the insulator region, the extrapolation was being done relative to cells only inside the domain, not using the value on the boundary. This fixes this to do extrapolation from the value on the boundary and the next inward cell.

### DIFF
--- a/Source/BoundaryConditions/PEC_Insulator.cpp
+++ b/Source/BoundaryConditions/PEC_Insulator.cpp
@@ -62,7 +62,8 @@ namespace
                                   amrex::GpuArray<FieldBoundaryType, 3> const fbndry_hi)
     {
         using namespace amrex::literals;
-        amrex::IntVect ijk_boundary = ijk_vec;
+        amrex::IntVect ijk_next = ijk_vec;
+        amrex::IntVect ijk_nextp1 = ijk_vec;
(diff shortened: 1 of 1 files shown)


EZoni (approved):
Thanks, looks good to me. How did you catch this and/or do you think we can change something on the CI end to make our tests more sensitive to errors like this? It seems like this fix does not affect the result of any tests currently.

dpgrote:
> Thanks, looks good to me. How did you catch this and/or do you think we can change something on the CI end to make our tests more sensitive to errors like this? It seems like this fix does not affect the result of any tests currently.

I caught this debugging the changes in the PR #5475. I was seeing wrong values in the guard cells. Adding a CI for this would be a bit of work since it only affects the data in the guard cells. Can the guard cells be included in the benchmark values? It doesn't seem like it. I would have to come up with a case where there are particles near the boundary that are affected by the data in the guard cells because of their shape function, which would be a sizable amount of work.

EZoni:
> Can the guard cells be included in the benchmark values? It doesn't seem like it.

Thanks @dpgrote, not sure about this, but we could look into it, it's a good point.

EZoni (approved):

PR #5503, opened 2024-12-08
Add `verbosity` flag for diagnostic output print statements

roelof-groenewald:
This PR passes the `warpx.verbose` value to diagnostics to set whether a message is output when the diagnostic is active. Specifically, this helps with the time-averaged diagnostic where currently a message is print at every step that averaging is done (commonly every step).


roelof-groenewald:
@n01r and @RevathiJambunathan the time-averaged diagnostic is super nice! Thanks for implementing that!
One thing though is that it prints out a message to the `stdout` at every step in the way I'm using it which makes it hard to see anything else in the output. This PR updates the diagnostics to only print messages to `stdout` if the `verbosity` flag is set to a non-zero value.

n01r (approved):
Looks good to me! Thanks, @roelof-groenewald!
Happy you like the new diags!

EZoni (approved):

PR #5506, opened 2024-12-11
[WIP] Expose electrostatic solver that only accounts for boundary potentials

RemiLehe:
TODO: 
- [ ] Add documentation
- [ ] Add automated test: spacecraft charging? (incl. with restart?) This is tricky since the new function does not affect the potential `phi` ; maybe we should make this an option:
     - if `phi` exists modify it
     - if not, create temporary `phi` (as is done currently)
- [ ] Check that guard cells are properly exchanged
- [ ] Doc: discuss distinction between `Ex_fp` and `Ex_aux`


Source/FieldSolver/ElectrostaticSolvers/ElectrostaticSolver.cpp
```diff
@@ -541,3 +541,39 @@ void ElectrostaticSolver::computeB (
         }
     }
 }
+
+void ElectrostaticSolver::AddBoundaryField (ablastr::fields::MultiLevelVectorField& Efield_fp)
```
RemiLehe:
```suggestion
void ElectrostaticSolver::AddBoundaryField const (ablastr::fields::MultiLevelVectorField& Efield_fp)
```

Source/FieldSolver/ElectrostaticSolvers/ElectrostaticSolver.H
```diff
@@ -136,6 +136,15 @@ public:
         std::array<amrex::Real, 3> beta
     ) const;
 
+    /** Compute the potential `phi` by solving the Poisson equation with the
+       simulation specific boundary conditions and boundary values, then add the
+       E field due to that `phi` to `Efield_fp`.
+    * \param[in] Efield Efield updated to include potential gradient from boundary condition
+    */
+    void AddBoundaryField (
```
RemiLehe, line 145:
```suggestion
    void AddBoundaryField const (
```

PR #5515, opened 2024-12-16
DSMC: make the check of the number of processes more robust

RemiLehe:
The previous code was brittle because it was not using `AMREX_ALWAYS_ASSERT_WITH_MESSAGE` (which indeed cannot be used inside a GPU kernel. In practice, a user could specify e.g. 10 scattering processes and not get an error.

The new code checks the number of processes before calling the GPU kernel, by using `AMREX_ALWAYS_ASSERT_WITH_MESSAGE`.

### DIFF
--- a/Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.H
+++ b/Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.H
@@ -165,7 +165,7 @@ public:
 
                     const int max_process_count = 4; // Pre-defined value, for performance reasons
                     AMREX_ALWAYS_ASSERT_WITH_MESSAGE(
-                        (m_process_count < max_process_count), "Too many scattering processes in DSMC routine."
+                        (m_process_count < max_process_count), "Too many scattering processes in DSMC routine (hardcoded to only allow 4). Update the max_process_count value in source code to allow more scattering processes."
                     );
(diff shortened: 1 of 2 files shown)


Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.H
```diff
@@ -163,7 +163,11 @@ public:
                     u1y[I1[i1]] = u1xbuf*std::sin(theta) + u1y[I1[i1]]*std::cos(theta);
 #endif
 
-                    CollisionPairFilter(
+                    const int max_process_count = 4; // Pre-defined value, for performance reasons
+                    AMREX_ALWAYS_ASSERT_WITH_MESSAGE(
+                        (m_process_count < max_process_count), "Too many scattering processes in DSMC routine."
```
roelof-groenewald: I know this is the same message I wrote originally, but in hind-sight maybe it is better to notify the user that this is not a hard limit. Maybe something like:
```suggestion
                        (m_process_count < max_process_count), "Too many scattering processes in DSMC routine (hardcoded to only allow 4). Update the max_process_count value in source code to allow more scattering processes."
```

roelof-groenewald (approved):
This looks good. I added one suggestion which you can accept if you agree, otherwise feel free to merge without it.

PR #5521, opened 2024-12-18
Injection from EB: do not create particles outside of user-specified bounds

RemiLehe:
When using the injection from embedded boundaries along with `zmin`, `zmax`, `xmin`, `xmax`, etc., we were creating a large number of macroparticles along the EB surface, and then removing the particles that are outside of `zmin`, `zmax`, `xmin`, `xmax`, etc. by setting their ID to an invalid value.

In case where the area defined by `zmin`, `zmax`, `xmin`, `xmax`, etc. is a small fraction of the EB surface, it is much more efficient not to create these particles in the first place. In addition, it avoids having to create a large number of particle IDs, which are eventually not used in the simulation. (In some use cases, this unnecessarily lead to a particle ID overflow, i.e. WarpX ended up having to generate particle IDs that were outside the maximum possible range given by the number of bits.)

This PR removes this issue.


dpgrote (approved):
Looks good. Nice that the fix was fairly small.

PR #5522, opened 2024-12-19
Docs: fix bugs (broken links, missing examples)

EZoni:
Just fixing a few errors and warnings I found while working on another documentation PR. Mostly broken links and missing examples.

Merge #5523 first, see https://github.com/ECP-WarpX/WarpX/pull/5522#issuecomment-2555975093 below.

PR #5523, opened 2024-12-20
CI: ignore all `.rst` files in the repository

EZoni:
Following up on #5387, I think we should also ignore all `.rst` files in the repository when we decide whether or not to run the CI workflows. 

GitHub Actions syntax taken from the examples [here](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#patterns-to-match-file-paths) (see `'**.js'` example). Azure syntax to be tested.

If we merge this before #5522, we can test it (i.e., test that CI is skipped) in #5522 after rebasing there.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -13,6 +13,7 @@ pr:
   paths:
     exclude:
     - Docs
+    - '**/*.rst'
 
 jobs:
 - job:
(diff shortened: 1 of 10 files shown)


ax3l (approved):
LGTM, thx!

PR #5209, opened 2024-09-03
Reduce particle shape when a particle approaches the EB

oshapoval:
**Description edited by @RemiLehe**

# Overview

This PR reduces the particle shape to order 1, when the particle gets closer to the embedded boundary:
Screenshot 2025-01-23 at 8 46 34 AM

This ensures that the particle does not deposit charge in valid cells, at the time when it is removed, which in turn ensures proper charge conservation with the electromagnetic solver.

# Implementation

- This PR allocates and initializes a new mask `eb_reduce_particle_shape` (and `iMultiFab`) that indicates in which cells to reduce the particle shape.
- The deposition kernels have been modified to use this flag. In order to make sure that this PR does not affect the performance of the deposition kernel in the absence of EB, two versions of the deposition kernel are compiled.

# Tests

This PR adds tests similar to the ones introduced in https://github.com/ECP-WarpX/WarpX/pull/5562 to check for charge conservation near the embedded boundary, but with higher-order shape factors:

- The 2D tests fail on `development` for shape 2 and 3 but pass on this PR.
- For some reason, the 3D and RZ tests only fail on `development` for shape 3 ; they do pass for this PR. It is not clear why the tests do not fail on `development` with shape 2.

**Note:** For now, this PR only modifies the current deposition (and only the Esirkepov kernel). A follow-up PR will also modify the charge deposition.

PR #5534, opened 2025-01-03
Modify stair-case approximation to the EB

RemiLehe:
# Overview

This PR changes the definition of the stair-case approximation of the EB (used in the Yee solver), so that the actual EB boundary (where e.g. particles are removed from the simulation) is **inside** the stair-case-approximated boundary - as opposed to the definition used in the current `development` branch, for which the actual EB is **outside** the stair-case-approximated boundary.

This ensures that the algorithm remains charge conserving, when charged particles are absorbed or emitted by the embedded boundary, **for particle shape of order 1**. (Higher-order particle shapes will be addressed in #5209.) As illustrated in the figure below (and as discussed in [this paper](https://www.researchgate.net/publication/318642364_Charge_Conserving_Emission_from_Conformal_Boundaries_in_Electromagnetic_PIC_simulations)), this is fundamentally because the particle does not deposit any charge in the valid cells, at the time when it is removed/emitted.

Screenshot 2025-01-12 at 8 56 17 AM

(The black crosses show the locations where the electric field is not updated, and thus usually remains equal to 0. The red dots show the locations where the particle deposits charge, for particle shape of order 1.)

The better behavior with respect to charge-conservation can be observed in the following animations, which show two particles of opposing charge separating and going into the embedded boundary. (Note that a static error in `divE` remains at the position where the particle was absorbed, with the `development` branch. The propagating errors in `divE` are expected: they are due to electromagnetic waves reflecting on the EB.)

- **development branch**
movie

- **this PR**
movie

Input script: [inputs.txt](https://github.com/user-attachments/files/18428873/inputs.txt)
Analysis script: [openPMD-visualization.ipynb.txt](https://github.com/user-attachments/files/18428878/openPMD-visualization.ipynb.txt)

(An automated tests using a similar configuration has been added in a separate, follow-up PR: https://github.com/ECP-WarpX/WarpX/pull/5562)

Note that, as part of this PR, the above new definition has been adopted for all the finite-difference solvers, except for the ECT solver (which uses a cut-cell representation instead of a stair-case representation).

# Implementation

This PR uses the changes of #5574. It still uses `MarkUpdateECellsECT` and `MarkUpdateBCellsECT` for the ECT sover - which preserve the previous behavior of the embedded boundary for this solver, but now uses `MarkUpdateCellsStairCase` for the other FDTD solvers - which introduce the above-mentioned new stair-case definition.


Source/Parallelization/WarpXRegrid.cpp
```diff
@@ -174,6 +174,14 @@ WarpX::RemakeLevel (int lev, Real /*time*/, const BoxArray& ba, const Distributi
     using ablastr::fields::Direction;
     using warpx::fields::FieldType;
 
+    const auto RemakeMultiFab = [&](auto& mf){
```
RemiLehe: This reintroduces a function that [we removed](https://github.com/ECP-WarpX/WarpX/pull/5230/files#diff-55dfa73c9e66860de4e9da580f1e41bd129ff4e2aab0b7b97e69d26f9adf1b40L172) when introducing the `MultiFabRegister` (#5230). 

It could be removed once we introduce an `iMultiFabRegister`.

Source/EmbeddedBoundary/WarpXInitEB.cpp
```diff
@@ -291,9 +291,215 @@ WarpX::ScaleAreas (ablastr::fields::VectorField& face_areas,
...
+#endif
+            }
+        );
+
+    }
+}
 
 void
-WarpX::MarkCells ()
+WarpX::MarkExtensionCells ()
```
RemiLehe: This function was renamed to avoid confusion with the other `Mark...` function.

Source/FieldSolver/FiniteDifferenceSolver/EvolveE.cpp
```diff
@@ -189,14 +171,9 @@ void FiniteDifferenceSolver::EvolveECartesian (
...
-#ifdef WARPX_DIM_3D
-                if (ly && ly(i,j,k) <= 0) { return; }
-#elif defined(WARPX_DIM_XZ)
-                //In XZ Ey is associated with a mesh node, so we need to check if the mesh node is covered
-                amrex::ignore_unused(ly);
-                if (lx && (lx(i, j, k)<=0 || lx(i-1, j, k)<=0 || lz(i, j-1, k)<=0 || lz(i, j, k)<=0)) { return; }
-#endif
+
+                // Skip field push in the embedded boundaries
+                if (update_Ey_arr && update_Ey_arr(i, j, k) == 0) { return; }
```
RemiLehe: The `if` condition here is now more simple, since the logic that determines whether a cell should be updated has been moved to the function that initializes the `update_E` flag.

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1300,6 +1253,7 @@ void WarpX::InitializeEBGridData (int lev)
 
             auto const eb_fact = fieldEBFactory(lev);
 
+            // TODO: move inside if condition for ECT
```
RemiLehe:
```suggestion
```

Source/WarpX.cpp
```diff
@@ -2301,6 +2303,22 @@ WarpX::AllocLevelMFs (int lev, const BoxArray& ba, const DistributionMapping& dm
...
+                                  guard_cells.ng_FieldSolver, lev, "m_eb_update_E[z]");
+
+                AllocInitMultiFab(m_eb_update_B[lev][0], amrex::convert(ba, Bx_nodal_flag), dm, ncomps,
+                                  guard_cells.ng_FieldSolver, lev, "m_eb_update_B[x]");
+                AllocInitMultiFab(m_eb_update_B[lev][1], amrex::convert(ba, By_nodal_flag), dm, ncomps,
+                                  guard_cells.ng_FieldSolver, lev, "m_eb_update_B[y]");
+                AllocInitMultiFab(m_eb_update_B[lev][2], amrex::convert(ba, Bz_nodal_flag), dm, ncomps,
+                                  guard_cells.ng_FieldSolver, lev, "m_eb_update_B[z]");
+
+                // TODO: do not allocate these arrays anymore with the Yee solver
```
```suggestion
```

Regression/Checksum/benchmarks_json/test_3d_eb_picmi.json
```diff
@@ -1,10 +1,10 @@
...
   "lev=0": {
-    "Bx": 148673.005859208,
-    "By": 148673.00585920806,
-    "Bz": 3371.758117878558,
-    "Ex": 55378581103426.71,
-    "Ey": 55378581103426.72,
-    "Ez": 68412803445328.25
+    "Bx": 144495.08082507108,
+    "By": 144495.08082507114,
+    "Bz": 8481.958724628861,
```
roelof-groenewald, line 5: This change is quite big (more than 2x). Do you understand why this component had such a dramatic change?
I guess the Bz value is small compared to the others so this is probably fine.

Source/EmbeddedBoundary/WarpXInitEB.cpp
```diff
@@ -291,9 +291,215 @@ WarpX::ScaleAreas (ablastr::fields::VectorField& face_areas,
...
+
+            } else { // The box contains a mix of covered and regular cells
+
+                auto const & flag = eb_flag[mfi].array();
+                auto index_type = field[idim]->ixType();
+
+                amrex::ParallelFor(box, [=] AMREX_GPU_DEVICE (int i, int j, int k) {
+
+                    // Stair-case approximation: If neighboring cells are either partially
+                    // or fully covered: do not update field
```
roelof-groenewald: If I'm understanding the logic here correctly, `eb_flag` is defined on the cell centers and returns `false` for `isRegular()` if any part of the cell is touched by the EB. The reason you have to check the cell below the current one when dealing with a nodal MF is because for the same index, nodal MFs are offset by half a cell downward from the cell centered MF. I think it might be worth adding a bit more explanation in the comment to make it easier for future readers to understand why it is only necessary to check the index offset by -1 (but never +1).

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -226,9 +226,8 @@ void HybridPICModel::InitData ()
             m_J_external[0],
             m_J_external[1],
             m_J_external[2],
-            lev, PatchType::fine, 'e',
-            warpx.m_fields.get_alldirs(FieldType::edge_lengths, lev),
-            warpx.m_fields.get_alldirs(FieldType::face_areas, lev));
+            lev, PatchType::fine,
+            warpx.GetEBUpdateEFlag());
```
roelof-groenewald: This is a slight problem in that there will be an inconsistency in the hybrid solver until the follow-up PR is merged that also updates the field update for the hybrid solver. I guess we should just be sure to have both PRs merge in quick succession.

Source/FieldSolver/WarpXPushFieldsEM.cpp
```diff
@@ -1155,7 +1157,7 @@ WarpX::MacroscopicEvolveE (int lev, PatchType patch_type, amrex::Real a_dt, amre
         m_fields.get_alldirs(FieldType::Efield_fp, lev),
         m_fields.get_alldirs(FieldType::Bfield_fp, lev),
         m_fields.get_alldirs(FieldType::current_fp, lev),
-        m_fields.get_alldirs(FieldType::edge_lengths, lev),
+        m_eb_update_E[lev],
```
roelof-groenewald: I'm assuming the plan is to replace these arguments once `iMultiFabRegister` is introduced?

Source/Initialization/WarpXInitData.cpp
```diff
@@ -976,23 +976,21 @@ WarpX::InitLevelData (int lev, Real /*time*/)
     // The default maxlevel_extEMfield_init value is the total number of levels in the simulation
     if ((m_p_ext_field_params->B_ext_grid_type == ExternalFieldType::parse_ext_grid_function)
          && (lev > 0) && (lev <= maxlevel_extEMfield_init)) {
+
+        // TODO: raise error when EB is activated
```
roelof-groenewald: Why is this comment here? Why should an error be raised when EB is activated?

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1104,44 +1095,19 @@ void WarpX::ComputeExternalFieldOnGridUsingParser (
...
         amrex::ParallelFor (tbx, tby, tbz,
             [=] AMREX_GPU_DEVICE (int i, int j, int k) {
-#ifdef AMREX_USE_EB
-#ifdef WARPX_DIM_3D
-                if(lx && ((topology=='e' and lx(i, j, k)<=0) or (topology=='f' and Sx(i, j, k)<=0))) { return; }
-#elif defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
-                //In XZ and RZ Ex is associated with a x-edge, while Bx is associated with a z-edge
-                if(lx && ((topology=='e' and lx(i, j, k)<=0) or (topology=='f' and lz(i, j, k)<=0))) { return; }
-#endif
-#endif
```
roelof-groenewald: This is a great simplification! 🎉

roelof-groenewald (commented):
This looks good to me. I was wondering, though, did you by any chance try running the CI tests using the new `m_eb_update_E` functionality but before moving the staircase approximation? That should be a really clean way to check that there are no hard to see bugs introduced in this PR, since all the CI tests should then pass with their original checksum values.

RemiLehe:
> This looks good to me. I was wondering, though, did you by any chance try running the CI tests using the new `m_eb_update_E` functionality but before moving the staircase approximation? That should be a really clean way to check that there are no hard to see bugs introduced in this PR, since all the CI tests should then pass with their original checksum values.

That's a good point! I have done this in this PR: https://github.com/ECP-WarpX/WarpX/pull/5574
That PR is based on #5534 and uses the new `eb_update_E`/`eb_update_B` framework, but:
- it initializes them with the previous definition of the EB, by using `MarkUpdateECellsECT`/`MarkUpdateBCellsECT` for all solvers (not just the ECT solver):
https://github.com/ECP-WarpX/WarpX/pull/5574/commits/3a3f36efe252fe60645dafe89146f56748f41892
- it does not modify the checksum (compared to the current `development`)

The checksum tests indeed pass for https://github.com/ECP-WarpX/WarpX/pull/5574/.

Regression/Checksum/benchmarks_json/test_3d_eb_picmi.json
RemiLehe: Hm, yes, it is somewhat large. My guess is that this is because this PR changes the actual position of the EB for this test.

Source/EmbeddedBoundary/WarpXInitEB.cpp
RemiLehe: Yes, that's correct. I added more comments ; could you check if that's clearer now?

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
RemiLehe: Sounds good. 
I finalized the hybrid PR https://github.com/ECP-WarpX/WarpX/pull/5558 and the tests pass on that PR, so it is ready to be merged soon after this one.

Source/FieldSolver/WarpXPushFieldsEM.cpp
RemiLehe: Yes, I agree.

Source/Initialization/WarpXInitData.cpp
RemiLehe: This is because I realized that the code is probably incorrect when using MR with the EM solver (i.e. when we use the `cp` version of the fields). 

This is because we should also have `cp` and `fp` versions of the arrays `eb_update_E` / `eb_update_B`.

That being said, the problem already existed before this PR. (i.e. we should have had `cp` and `fp` version of `edge_length` and `face_area`).

For now, I cleaned up this comment and will instead open an issue with more explanation.

Source/Initialization/WarpXInitData.cpp
```diff
@@ -1253,6 +1253,21 @@ void WarpX::InitializeEBGridData (int lev)
...
+                    m_fields.get_alldirs(FieldType::Efield_fp, lev),
+                    eb_fact );
+                // Mark on which grid points B should be updated (stair-case approximation)
+                MarkUpdateCellsStairCase(
+                    m_eb_update_B[lev],
+                    m_fields.get_alldirs(FieldType::Bfield_fp, lev),
+                    eb_fact );
             }
 
             // Mark on which grid points E should be updated
```
roelof-groenewald: The lines below this should now be removed, no?

RemiLehe: Ah yes, sorry.

roelof-groenewald (approved):
Looks good. Thanks!

PR #5536, opened 2025-01-03
PEC: Fix Uninit Var Warning (GCC)

ax3l:
GCC 13 throws a compile-time warning in 3D that in some cases, `is_normal_to_boundary` is used as uninitialized. There are a lot of `if`s in its usage, so I simply changed the init value to `false` to silence it.


dpgrote (approved):
Thanks! I double checked the logic and that variable would have never been used if it was not set. But the fix is good.

ax3l:
Thanks! Yes, it looks a bit like a noisy compiler diagnostics to me, but the default feels sensible to add.

PR #5537, opened 2025-01-03
Implementation of Bremsstrahlung collisions

dpgrote:
This PR adds Bremsstrahlung collisions, producing photon particles. It does pair wise collisions between electron and ion species, produce photons. The algorithm conserves momentum and energy.

It uses the cross section tables from Seltzer and Berger, but now only has the cross sections for Z = 1, 2, 5, and 6.  A future PR can add individual cross section data as needed or change this to read in data from the warpx-data repo to allow any Z.

Todo:
- [x]  Add documentation
- [x]  Check on GPU
- [x]  Add CI tests


Examples/Tests/collision/inputs_test_1d_collision_z_Bremsstrahlung
```diff
@@ -0,0 +1,106 @@
...
+photons.species_type = photon
+photons.injection_style = none
+
+#################################
+############ COLLISION ##########
+#################################
+collisions.collision_names = bremsstrahlung
+bremsstrahlung.type = bremsstrahlung
+bremsstrahlung.species = electrons ions
+bremsstrahlung.Z = 5
```
ax3l, line 88: Is this to overwrite the ionization state or, since this is a particle-particle mode, can we also rely on the charge / ionization state of the ion we collide with?

dpgrote: The `Z` is only used to indicate the atomic number, which is used to select which cross section data to use from the tables.

Examples/Tests/collision/analysis_collision_1d_Bremsstrahlung.py
```diff
@@ -0,0 +1,124 @@
...
+#
+#
+# This file is part of WarpX.
+#
+# License: BSD-3-Clause-LBNL
+#
+# This is a script that analyses the simulation results from the script `inputs_test_1d_collision_z_Bremsstrahlung`.
+# run locally: python analysis_collision_1d_Bremsstrahlung.py diags/diag1000600/
+#
+# This is a 1D Bremsstrahlung collision of electrons on Borons ions
```
atmyers:
```suggestion
# This is a 1D Bremsstrahlung collision of electrons on Boron ions
```

Source/Particles/Collision/BinaryCollision/Bremsstrahlung/BremsstrahlungFunc.H
```diff
@@ -0,0 +1,529 @@
...
+
+            if (KErel_eV < KEcut_eV) {
+                // Ignore low energy electrons which would only produce low energy photons
+                return 0._prt;
+            }
+
+            // find lo-index for koT1 cutoff (will typically be 1)
+            koT1_cut = std::max(KEcut_eV/KErel_eV, m_koT1_cut_default);
+            i0_cut = 0;
+            for (int i=1; i < nkoT1; i++) {
```
atmyers, line 208: Does it make sense to bisect here as well? Or will this normally only require a few iterations?

Source/Particles/Collision/BinaryCollision/Bremsstrahlung/BremsstrahlungFunc.H
```diff
@@ -0,0 +1,529 @@
...
+            if (sigma_total == 0._prt) { return; }
+
+            // determine if the pair collide and then do the collision
+            amrex::ParticleReal fmulti = m_multiplier; // >= 1.0
+            amrex::ParticleReal const gamma_factor = gamma1_rel/(gamma1*gamma2);
+            amrex::ParticleReal const v1 = std::sqrt(u1sq_rel)/gamma1_rel; // magnitude electron velocity
+            amrex::ParticleReal arg = fmulti*v1*sigma_total*n2*dt*gamma_factor;
+            if (arg > 1.0_prt) {
+#ifndef AMREX_USE_GPU
+                amrex::Print() << "Notice: arg = " << arg << " in interSpeciesBremsstrahlung" << "\n";
```
atmyers, line 376: Do you want to remove these? Or maybe use the warning logger?

Source/Particles/Collision/BinaryCollision/Bremsstrahlung/PhotonCreationFunc.H
```diff
@@ -0,0 +1,306 @@
...
+#include <AMReX_GpuAtomic.H>
+#include <AMReX_GpuDevice.H>
+#include <AMReX_GpuContainers.H>
+#include <AMReX_INT.H>
+#include <AMReX_Random.H>
+#include <AMReX_REAL.H>
+#include <AMReX_Vector.H>
+
+/**
+ * \brief This functor creates photons produced from a binary collision and sets their initial
```
atmyers, line 28: Is this functor specific to Bremsstrahlung or could it in principle be reused in PR #5595 or #5878?

Source/Particles/Collision/BinaryCollision/Bremsstrahlung/BremsstrahlungFunc.H
dpgrote: bisect is probably not needed here since this will almost always return 1 since it is unlikely that the wpe will be large compared to the electron energy values of the table.

Source/Particles/Collision/BinaryCollision/Bremsstrahlung/BremsstrahlungFunc.H
dpgrote: Yes, I can remove this and use a warning instead.

dpgrote: Actually, I don't see a way of adding a warning since presumably the warning logger cannot be called inside a kernel and this routine is called from the generic binary collision code.

Source/Particles/Collision/BinaryCollision/Bremsstrahlung/PhotonCreationFunc.H
dpgrote: Good question. Looking at those two PRs there are enough differences that I think it would be hard to write a common routine for adding photons. For example here, this routine is also modifying the momentum of the electron and ion to conserve momentum and energy, which is not done in the other PRs.

Source/Particles/Collision/BinaryCollision/Bremsstrahlung/BremsstrahlungFunc.H
```diff
@@ -0,0 +1,529 @@
...
+        {
+            using namespace amrex::literals;
+            amrex::ParticleReal koT1_grid_im1 = koT1_cut;
+            amrex::ParticleReal kdsigdk_im1 = kdsigdk_cut;
+            amrex::ParticleReal sigma = 0._prt;
+            amrex::ParticleReal kdsigdk_i;
+            amrex::ParticleReal dk;
+            amrex::ParticleReal k_ave;
+            int i;
+            for (i=i0_cut+1; i < nkoT1; i++) {
```
ax3l: I don't think you use the `i` afterwards, so you can write:
```suggestion
            for (int i=i0_cut+1; i < nkoT1; i++) {
```

Source/Particles/Collision/BinaryCollision/Bremsstrahlung/BremsstrahlungFunc.H
```diff
@@ -0,0 +1,529 @@
...
+         */
+        AMREX_GPU_HOST_DEVICE AMREX_INLINE
+        amrex::ParticleReal
+        Photon_energy(int index, int i0_cut, amrex::ParticleReal koT1_cut, amrex::ParticleReal kdsigdk_cut,
+                      amrex::ParticleReal w0, amrex::ParticleReal sigma_total, amrex::ParticleReal random_number) const
+        {
+            using namespace amrex::literals;
+            amrex::ParticleReal koT1_grid_im1 = koT1_cut;
+            amrex::ParticleReal kdsigdk_im1 = kdsigdk_cut;
+            amrex::ParticleReal sigma = 0._prt;
```
ax3l, line 280: Works, too (ints or floats, cast is constexpr):
```suggestion
            amrex::ParticleReal sigma = 0_prt;
```

Source/Particles/Collision/BinaryCollision/Bremsstrahlung/BremsstrahlungFunc.H
```diff
@@ -0,0 +1,529 @@
...
+
+        // relative photon energy grid ( koT1 = Ephoton_eV/KErel_eV )
+        amrex::GpuArray<amrex::ParticleReal, nkoT1> m_koT1_grid = {0., 0.025, 0.05, 0.075, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 0.95, 0.97, 0.99, 1.00};
+
+        // energy grid for electrons in eV
+        amrex::GpuArray<amrex::ParticleReal, nKE> m_KEgrid_eV = {1.0e3, 2.0e3, 5.0e3, 1.0e4, 2.0e4, 5.0e4, 1.0e5, 2.0e5, 5.0e5, 1.0e6, 2.0e6};
+
+        // differential total cross section
+        amrex::GpuArray<amrex::GpuArray<amrex::ParticleReal, nkoT1>, nKE> m_kdsigdk;
+        amrex::GpuArray<amrex::ParticleReal, nKE> m_sigma_total;
```
ax3l: Depending on the size of `nkoT1` and `nKE` this probably spills registers heavily on GPU.
(Currently the values are hard-coded to 11 and 17, so 1.5kB for double.)

One could alternatively implement this in global memory, but as long as it works for now we can roll like this for a first implementation.

For CPU the main difference is stack vs heap storage.

ax3l (approved):
The implementation and documentation looks good to me.

I think the current interpolation probably tries to put a lot (around 1.5kB) of interpolated memory on stack memory and thus on GPU will spill a lot of registers (as in: might be slow). But as long as it runs for now that can also be improved in a follow-up PR.

atmyers (approved):

Source/Particles/Collision/BinaryCollision/Bremsstrahlung/BremsstrahlungFunc.H
dpgrote: I can change these to use `DeviceVector` instead of `GpuArray`. Is there a DeviceVector or equivalent that allows 2D array indexing?

Docs/source/usage/parameters.rst
```diff
@@ -2104,19 +2106,21 @@ Details about the collision models can be found in the :ref:`theory section <mul
       The implementation follows the same numerical algorithm as that of fusion reactions (see. `Higginson et al. (JCP 388, 439-453, 2019) <https://doi.org/10.1016/j.jcp.2019.03.020>`__).
 
 * ``<collision_name>.species`` (`strings`)
-    If using ``dsmc``, ``pairwisecoulomb`` or ``nuclearfusion``, this should be the name(s) of the species,
+    If using ``dsmc``, ``pairwisecoulomb``, ``nuclearfusion``, or ``bremsstrahlung``, this should be the name(s) of the species,
     between which the collision will be considered. (Provide only one name for intra-species collisions.)
+    With ``bremsstrahlung``, the electron species must be given first, followed by the target species.
```
aeriforme, line 2119: Are there any restrictions on the target species?

dpgrote: Essentially no, but for the correct physics, it would need to be consistent with the `Z` input parameter.

Source/Particles/Collision/BinaryCollision/Bremsstrahlung/BremsstrahlungFunc.H
dpgrote: This is now updated. The cross section data is now explicitly copied to the GPU at the start of the calculation rather than through the lambda capture at the launch of each kernel.

PR #5543, opened 2025-01-07
[WIP] Add collisional diagnostic buffers

budjensen:
Stemming out of conversation https://github.com/ECP-WarpX/WarpX/discussions/5381 with @roelof-groenewald, included is initial work on adding `multifab` buffers to track collisional diagnostics in a MCC-type simulation.

The work on two buffers is in progress--one for ionization and one for total collisional energy loss. Both `multifabs` have been included in `Fields.H` and initialized in `WarpX.cpp`. Relevant flags to turn on MCC tracking have been created in `WarpX.H` and initialized in `WarpX.cpp`. The energy loss `multifab` has been added into the collision protocol, and future commits will add the ionization buffer.

In the end, I aim to set up Python access to these `multifabs`, with similar functionality to the `particle_containers.ParticleBoundaryBufferWrapper()` wrappers.


dpgrote:
This looks good. A comment - these new MultiFabs will automatically be accessible from Python since they are stored in the MultiFab map. You can also add convenience wrappers for them by adding the appropriate wrapper routines in `Python/pywarpx/fields.py`.

budjensen:
@dpgrote, thanks for the tip!

Source/Particles/Collision/BackgroundMCC/BackgroundMCCCollision.cpp
```diff
@@ -515,6 +575,36 @@ void BackgroundMCCCollision::doBackgroundIonization
...
+            amrex::ParallelFor(num_added,
+                               [=] AMREX_GPU_HOST_DEVICE (long ip)
+                               {
+                               // store the energy (add E_lost * w[ip] to the buffer of energy sent to the background)
+                               // get the particle's cell index (method used here is similar to TemperatureFunctor.cpp)
+                               const auto p = WarpXParticleContainer::ParticleType(ptd, ip + np_elec);
+                               const auto [ii, jj, kk] = amrex::getParticleCell(p , plo, dxi).dim3();
+
+                               // store energy lost in the buffer
+                               coll_ionization_arr(ii, jj, kk, 0) += w[ip + np_elec];
```
budjensen: I've added everything that I think I'll need for this diagnostic, but am running into the following compilation error:

```
/home/bj8080/src/warpx/Source/Particles/Collision/BackgroundMCC/BackgroundMCCCollision.cpp:603:67: error:
assignment of read-only location 'coll_ionization_arr.amrex::Array4<const double>::operator()<>(((int)ii), ((int)jj), ((int)kk), 0)'
  603 |                                coll_ionization_arr(ii, jj, kk, 0) += w[ip + np_elec];
      |                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
```
since `coll_ionization_tracking` is a constant. When I correct for this by removing `const` from the initialization of `coll_ionization_arr` and the earlier line initializing  `coll_ionization`, I get an error:

```
/home/bj8080/src/warpx/Source/Particles/Collision/BackgroundMCC/BackgroundMCCCollision.cpp:586:67:
error: cannot bind non-const lvalue reference of type 'amrex::Array4<double>&' to an rvalue of type 'amrex::Array4<double>'
  586 |             auto& coll_ionization_arr = coll_ionization[pti].array();
      |                                         ~~~~~~~~~~~~~~~~~~~~~~~~~~^~
```

I think to get around this I should create a `multifab` to sum up the ionization events, and then add that multifab to coll_ionization_arr. How would I do this? Or, is there a better approach?

budjensen: Pinging this with @roelof-groenewald, since we've talked about this feature in discussion [5381](https://github.com/ECP-WarpX/WarpX/discussions/5381). Would you be able to take a look?

Source/Fields.H
```diff
@@ -81,7 +81,9 @@ namespace warpx::fields
         Bfield_avg_cp,
         B_old, /**< Stores the value of B at the beginning of the timestep, for the implicit solver */
         ECTRhofield,
-        Venl
+        Venl,
+        collision_energy_change, /**< Stores collisional energy transfer data for Monte Carlo Collisions */
+        collision_ionization /**< Stores ionization data for Monte Carlo Collisions */
```
roelof-groenewald, line 86: For clarity it would be good to add `mcc` to these multifab names (to make it clear they are tied to the MCC routine). Maybe something like `mcc_collision_energy_change`, etc.

Source/Particles/Collision/BackgroundMCC/BackgroundMCCCollision.cpp
```diff
@@ -414,6 +434,20 @@ void BackgroundMCCCollision::doBackgroundCollisionsWithinTile
...
+                                          ParticleUtils::getEnergy(v_lost2, m, E_lost);
+
+                                          // store the energy (add E_lost * w[ip] to the buffer of energy sent to the background)
+                                          // get the particle's cell index (method used here is similar to TemperatureFunctor.cpp)
+                                          const auto p = WarpXParticleContainer::ParticleType(ptd, ip);
+                                          const auto [ii, jj, kk] = amrex::getParticleCell(p , plo, dxi).dim3();
+
+                                          // store energy lost in the buffer
+                                          coll_E_change_arr(ii, jj, kk, 0) += E_lost * w[ip];
+                                      }
```
roelof-groenewald: It would be nice if this code does not need to be duplicated as it is currently. Perhaps the `break` can be removed, and a `else` used for the non-charge exchange processes?

Source/WarpX.H
```diff
@@ -867,6 +867,10 @@ public:
     static amrex::Real moving_window_v;
     static bool fft_do_time_averaging;
 
+    // Control whether Monte Carlo Collision statistics are collected
+    static bool do_MCC_energy_tracking;
+    static bool do_MCC_ionization_tracking;
+
```
roelof-groenewald, line 873: These should instead be made member variables of the `BackgroundMCCCollision` class.

Source/WarpX.cpp
```diff
@@ -2224,6 +2274,25 @@ WarpX::AllocLevelMFs (int lev, const BoxArray& ba, const DistributionMapping& dm
     m_fields.alloc_init(FieldType::current_fp, Direction{1}, lev, amrex::convert(ba, jy_nodal_flag), dm, ncomps, ngJ, 0.0_rt);
     m_fields.alloc_init(FieldType::current_fp, Direction{2}, lev, amrex::convert(ba, jz_nodal_flag), dm, ncomps, ngJ, 0.0_rt);
 
+    // Allocate extra multifabs needed for MCC collision tracking
+    if (do_MCC_ionization_tracking || do_MCC_energy_tracking) {
```
roelof-groenewald, line 2278: These multifabs could also be initialized in the constructor for `BackgroundMCCCollision`.

Source/Particles/Collision/BackgroundMCC/BackgroundMCCCollision.cpp
```diff
@@ -515,6 +575,36 @@ void BackgroundMCCCollision::doBackgroundIonization
...
+            // get parameters for getParticleCell (method used here is similar to TemperatureFunctor.cpp)
+            // should these lines be removed and we just do ptd = elec_tile.getParticleTileData()?
+            auto& tile = pti.GetParticleTile();
+            auto ptd = tile.getParticleTileData();
+
+            // get multi-fab array for ionization creation tracking
+            const auto& coll_ionization_arr = coll_ionization[pti].array();
+
+            // get Struct-Of-Array particle data
+            // as long as w is set after the copy, will it include newly created particles (be of length np_elec + num_added)?
```
roelof-groenewald, line 593: Yes, at this point the length will include the new particles (since this happens after `setNewParticleIDs`.

Source/Particles/Collision/BackgroundMCC/BackgroundMCCCollision.cpp
```diff
@@ -515,6 +575,36 @@ void BackgroundMCCCollision::doBackgroundIonization
...
+            // add each created particle to tracking buffer
+            amrex::ParallelFor(num_added,
+                               [=] AMREX_GPU_HOST_DEVICE (long ip)
+                               {
+                               // store the energy (add E_lost * w[ip] to the buffer of energy sent to the background)
+                               // get the particle's cell index (method used here is similar to TemperatureFunctor.cpp)
+                               const auto p = WarpXParticleContainer::ParticleType(ptd, ip + np_elec);
+                               const auto [ii, jj, kk] = amrex::getParticleCell(p , plo, dxi).dim3();
+
+                               // store energy lost in the buffer
```
roelof-groenewald: This buffer is tracking number of electrons created, not energy, right?

Source/Particles/Collision/BackgroundMCC/BackgroundMCCCollision.cpp
roelof-groenewald: I think you should do
```
const auto& coll_ionization_arr = coll_ionization.array(pti);
```
Does that help?

roelof-groenewald (commented):
Sorry for the slow response (again). Things have been quite hectic since the start of the year 😃

Source/Particles/Collision/BackgroundMCC/BackgroundMCCCollision.cpp
budjensen: That was part of it. It seems that my latest commit is correctly accessing multifabs. Thanks!

Python/pywarpx/fields.py
```diff
@@ -1030,3 +1030,15 @@ def GCPPMLWrapper(level=0, include_ghosts=False):
...
+def MCCEnergyWrapper(level=0, include_ghosts=False):
+    return _MultiFABWrapper(
+        mf_name="collision_energy_change", level=level, include_ghosts=include_ghosts
+    )
+
+
+def MCCIonizWrapper(level=0, include_ghosts=False):
+    return _MultiFABWrapper(
+        mf_name="collision_ionization", level=level, include_ghosts=include_ghosts
+    )
```
budjensen, line 1044: This may be a question for @dpgrote: Is this line in `fields.py` all I need to do for multifab access in python? When I try the following in a picmi script:

```python
ioniz_tracker = fields.MCCIonizWrapper()
tracking_data = ioniz_tracker[...]
```
I getting the following error:
```bash
  File "/scratch/gpfs/user/warpx/Feb_2025/mcc_diag_check/Turner_inputs.py", line 425, in run_sim
    tracking_data = ioniz_tracker[...]
  File "/home/user/src/warpx/build_MCC_coll/lib/site-packages/pywarpx/fields.py", line 464, in __getitem__
    result_global[global_slices] = f_arr
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^
ValueError: could not broadcast input array from shape (0,1,1,1) into shape (16,1,1,1)
```

Do you know how to fix this error? It seems like the multifab is either not being accessed correctly, or initialized correctly.

In some rudimentary checking, the multifabs seem to be correctly allocated (in this 1D case, 128 cells spread across 8 boxes of 16 for 8 MPI ranks) on the C++ side. It also looks like data is being saved in the multifab arrays correctly, as I can see the multifab array elements increase as ionization events occur (via print statements inserted in BackgroundMCCCollisions.cpp).

PR #5545, opened 2025-01-09
Add reduced diagnostic: 2d differential luminosity 

aeriforme:
Adds a luminosity diagnostic differentiated in the energies of two colliding species, called `DifferentialLuminosity2D`.
It is defined as follows:
```math
\begin{align*}
\frac{d^2\mathcal{L}}{dE_1 dE_2}(E_1, E_2, t)  = \int_0^t dt'\int d\boldsymbol{x}\, & \int d\boldsymbol{p}_1 \int d\boldsymbol{p}_2\; \sqrt{ |\boldsymbol{v}_1 - \boldsymbol{v}_2|^2 - |\boldsymbol{v}_1\times\boldsymbol{v}_2|^2/c^2} \\ 
& f_1(\boldsymbol{x}, \boldsymbol{p}_1, t')f_2(\boldsymbol{x}, \boldsymbol{p}_2, t') \delta(E_1 - E_1(\boldsymbol{p}_1) \delta( E_2 - E_2(\boldsymbol{p}_2))
\end{align*}
```
where:
*  $\boldsymbol{p}_i$ is the momentum of a particle of species $i$
*  $E_i$ is the energy of a particle of species $i$, $E_i (\boldsymbol{p}_i) = \sqrt{m_1^2c^4 + c^2 |\boldsymbol{p}_i|^2}$
*  $f_i$ is the distribution function of species $i$, normalized such that $\int \int f(\boldsymbol{x} \boldsymbol{p}, t )d\boldsymbol{x} d\boldsymbol{p} = N$, the number of particles in species $i$ at time $t$

The 2D differential luminosity is given in units of $\text{m}^{-2} \ \text{eV}^{-2}$.

The user must specify the minimum, maximum, and number of bins to discretize the $E_1$ and $E_2$ axes.
The computation of this diagnostic is similar to that of `ParticleHistogram2D`.
 
The output is a folder containing a set of openPMD files.
The values of the diagnostic are stored in a record labeled `d2L_dE1_dE2`, with axes `E1` and `E2`.


Examples/Tests/diff_lumi_diag/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 # Add tests (alphabetical order) ##############################################
 #
-
+if(WarpX_FFT)
```
RemiLehe, line 3: Is this change intentional? It seems unrelated to the PR.

Examples/Tests/diff_lumi_diag/analysis.py
```diff
@@ -33,21 +38,37 @@
...
-    tol = 6e-2
+    tol1 = 0.02
+    tol2 = 0.003
+
+# Check that the 1D diagnostic and analytical result match
+error1 = abs(dL_dE_sim - dL_dE_th).max() / abs(dL_dE_th).max()
+print("Relative error: ", error1)
+print("Tolerance: ", tol1)
+
+# Check that the 2D and 1D diagnostics match
```
RemiLehe, line 78: Hm, a stronger test would be to compute `d2L_dE1_dE2_th` and to compare the values.

aeriforme: You are right.

Examples/Tests/diff_lumi_diag/CMakeLists.txt
aeriforme: It is intentional. 
This example requires the compilation flag `-DWarpX_FFT=ON`.
Without this, the related `ctest` would fail.

Note that this has not been a problem in CI so far because the FFT compilation flag is on.

RemiLehe (approved):

PR #5549, opened 2025-01-10
python callback at restart

RevathiJambunathan:
This PR adds a python callback option after initialization at restart, enabling users to read user-defined checkpoint vars that could be written using the python call afterdiagnostics that already exists


roelof-groenewald (approved):
Looks good!

PR #5557, opened 2025-01-14
Use `amrex::getParticleCell` More

ax3l:
One leftover occurance in `ParticleReductionFunctor`.

Follow-up to #5118


archermarx (commented):
Looks good to me. What is the `AMReX NGP interpolator`?

ax3l:
I think the comment by @hklion (cc @atmyers) is to use an actual interpolation instead of assigning to a cell index. NGP (next-grid-cell) interpolation and just assigning to the cell a particle is in differs slightly. But in the end, if we use interpolation we probably would like to use at least CIC (cloud-in-cell) for clean results and charge conservation (where needed for analysis in respect to currents).

archermarx:
Gotcha, thanks!

roelof-groenewald (approved):

PR #5560, opened 2025-01-15
Use `self_fields_verbosity` value for magnetostatic solver

### LINKED ISSUE #5530: Control relativistic and magnetostatic electrostatic solver verbosity with warpx_self_fields_verbosity
When using the `pywarpx.picmi.ElectrostaticSolver` class with `warpx_relativistic=True` or `warpx_magnetostatic=True`, the solver verbosity is not controlled by `warpx_self_fields_verbosity`. This results in a lot of undesired stdout for me, because I set verbosity to 0 when my runs are going well.

Technically the documentation says that `warpx_self_fields_verbosity` only controls verbosity for the lab frame solver, so this is technically a feature request rather than a bug. But it might as well be a bug.

roelof-groenewald:
This is another temporary fix (similar to #5517) to close #5530.

### DIFF
--- a/Source/WarpX.cpp
+++ b/Source/WarpX.cpp
@@ -684,6 +684,7 @@ WarpX::ReadParameters ()
         "To use the FFT Poisson solver, compile with WARPX_USE_FFT=ON.");
 #endif
         utils::parser::queryWithParser(pp_warpx, "self_fields_max_iters", magnetostatic_solver_max_iters);
+        utils::parser::queryWithParser(pp_warpx, "self_fields_verbosity", magnetostatic_solver_verbosity);
 
         WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
         (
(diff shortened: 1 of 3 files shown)


Source/WarpX.cpp
```diff
@@ -684,6 +684,7 @@ WarpX::ReadParameters ()
         "To use the FFT Poisson solver, compile with WARPX_USE_FFT=ON.");
 #endif
         utils::parser::queryWithParser(pp_warpx, "self_fields_max_iters", magnetostatic_solver_max_iters);
+        utils::parser::queryWithParser(pp_warpx, "self_fields_verbosity", magnetostatic_solver_verbosity);
```
ax3l, line 687: LGTM, but let's add a documentation in `parameters.rst` and also Python control?

ax3l: Oh, it's the existing `self_fields_verbosity` flag.

ax3l (approved):

PR #5567, opened 2025-01-16
Catch warpx not being initialized in library loader

archermarx:
Follows up on, https://github.com/ECP-WarpX/WarpX/pull/5412, using @ax3l's [suggestion](https://github.com/ECP-WarpX/WarpX/pull/5412#pullrequestreview-2556960357) to catch warpx not being initialized when accessing libwarpx.warpx


ax3l (approved):
Awesome, that is a great usability improvement :+1: Thanks!

PR #5568, opened 2025-01-16
Doc: More Developer Packages

ax3l:
Add missing developer packages to our conda developer environment.

X-ref: https://github.com/ECP-WarpX/impactx/pull/801

### DIFF
--- a/Docs/source/install/dependencies.rst
+++ b/Docs/source/install/dependencies.rst
@@ -80,7 +80,7 @@ Conda (Linux/macOS/Windows)
 
       .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


EZoni (approved):

EZoni:
@ax3l 

FYI, I had enabled auto-merge here, but I forgot that GitHub thinks that _required_ workflows that are skipped (like here, since we change only documentation files) are pending instead of skipped, so I cannot merge PRs where we skip required workflows (i.e., all documentation PRs). I think only maintainers with higher admin access can force the merge, like @ax3l or @RemiLehe.

We saw and looked a bit into this in the last couple of months, see for instance https://github.com/ECP-WarpX/WarpX/pull/5469. This is a known issue (or feature?) on the GitHub end, see their notes at https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/skipping-workflow-runs.

ax3l:
That is right. The workflow to mark a workflow as "skipped" as you planned will not work then. You can though early return and mark as success to speed up CI.

PR #5569, opened 2025-01-16
Docs: configure conda to use only `conda-forge`

EZoni:
Requested by @ax3l:

> We might also need to rework our conda docs, to change the default channel priorities to only use `conda-forge`. By default, many people have the `anaconda` repos in them too and, although we do not use them, this can break the build in this situation: some labs now block them in their firewall due to recent events, and then the installer just aborts instead of trying the next channel in the list. Very easy to do, I think we just need to prepend `conda-forge` in the channels (instead of the CLI) or add a guide to install Conda/Mamba from Conda-Forge/Minimamba instead of Anaconda.

This PR adds instructions to:
1. add `conda-forge` to the top of the channel list;
2. set the channel priority to strict.

Hopefully this should be enough?

### DIFF
--- a/Docs/source/install/dependencies.rst
+++ b/Docs/source/install/dependencies.rst
@@ -74,6 +74,13 @@ Conda (Linux/macOS/Windows)
 
       conda config --set auto_activate_base false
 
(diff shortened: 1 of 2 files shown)


ax3l (approved):
Thank you! :)

PR #5571, opened 2025-01-17
Remove temporary particle container

### LINKED ISSUE #5056: Remove Temporary Particle Container `tmp_pc`
The `WarpXParticleContainer` class has a `tmp_particle_data` (`tmp_pc`) here: https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/WarpXParticleContainer.H#L507, which is used exclusively to save the old positions and momenta used in the boosted-frame diagnostics:
https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/PhysicalParticleContainer.cpp#L2813
https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/PhysicalParticleContainer.cpp#L2813

This is a bit clunky, e.g. as `tmp_particle_data` needs to be resized here:
https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/WarpXParticleContainer.cpp#L1523
Additionally, the name is not very descriptive, and might confuse new developers (it is not clear that this is related to BTD, unless one really search through the code for the places where `tmp_particle_data` is used.)

It **might** be cleaner to remove `tmp_particle_data` and instead use runtime components, as is done currently for the implicit PIC (which also needs to save the old positions and momenta, albeit at a different point in the PIC loop):
https://github.com/ECP-WarpX/WarpX/blob/development/Source/Initialization/WarpXInitData.cpp#L591
https://github.com/ECP-WarpX/WarpX/blob/development/Source/FieldSolver/ImplicitSolvers/WarpXImplicitOps.cpp#L113
https://github.com/ECP-WarpX/WarpX/blob/development/Source/FieldSolver/ImplicitSolvers/WarpXImplicitOps.cpp#L113

We could use similar runtime components for the BTD, but with a different name in order not to conflict with the implicit solver (e.g. `uz_btd_old` instead of `uz_n`). I also seem to remember that we can set a flag on some runtime component to avoid doing MPI exchanges (which would be useful for the runtime components that would replace `tmp_particle_data`).

EZoni:
Close #5056.


Source/Diagnostics/ComputeDiagFunctors/BackTransformParticleFunctor.cpp
```diff
@@ -15,29 +15,37 @@
...
-SelectParticles::SelectParticles (const WarpXParIter& a_pti, TmpParticles& tmp_particle_data,
-                                  amrex::Real current_z_boost, amrex::Real old_z_boost,
-                                  int a_offset)
+SelectParticles::SelectParticles (
+    const WarpXParticleContainer& pc,
+    WarpXParIter& a_pti,
+    amrex::Real current_z_boost,
+    amrex::Real old_z_boost,
+    int a_offset
+)
```
EZoni: @atmyers 

Before I proceed in a similar way to remove `tmp_particle_data` from `CopyParticleAttribs`, do you agree with how I did it here? 

I am now passing the `WarpXParticleContainer` as an argument to this function in order to be able to call `getParticleComps` from it and use the result when calling `GetAttribs` from the particle iterator.

Does that make sense or do you see another/better way to do this?

EZoni:
@atmyers 

The test `test_2d_laser_acceleration_boosted` is one where I still see a segfault, which points to some discrepancy between source and destination runtime components:
```
  what():  Assertion `dst.m_num_runtime_real == src.m_num_runtime_real' failed, file "/home/edoardo/src/warpx/build/_deps/fetchedamrex-src/Src/Particle/AMReX_ParticleTransformation.H", line 35
```
It seems to originate from particle filtering whithin `BackTransformParticleFunctor::operator ()`. I have not figured out yet the bug in my changes.

atmyers:
Note - we need this AMReX PR for this to work: https://github.com/AMReX-Codes/amrex/pull/4333/files

EZoni:
> Note - we need this AMReX PR for this to work: https://github.com/AMReX-Codes/amrex/pull/4333/files

Thanks for the bug fix! We can proceed here further after #5669 is merged.

ax3l:
Needs https://github.com/AMReX-Codes/amrex/pull/4341

ax3l:
Great, only these tests still fail.

### RZ

	 11 - test_rz_electrostatic_sphere.checksum (Failed)
	 38 - test_rz_flux_injection.analysis (Failed)
	 39 - test_rz_flux_injection.checksum (Failed)
	 42 - test_rz_flux_injection_from_eb.analysis (Failed)
	 43 - test_rz_flux_injection_from_eb.checksum (Failed)
	 46 - test_rz_langmuir_multi.analysis (Failed)
	 47 - test_rz_langmuir_multi.checksum (Failed)
	 50 - test_rz_langmuir_multi_picmi.checksum (Failed)
	 53 - test_rz_langmuir_multi_psatd.analysis (Failed)
	 54 - test_rz_langmuir_multi_psatd.checksum (Failed)
	 57 - test_rz_langmuir_multi_psatd_current_correction.analysis (Failed)
	 58 - test_rz_langmuir_multi_psatd_current_correction.checksum (Failed)
	 61 - test_rz_langmuir_multi_psatd_multiJ.analysis (Failed)
	 62 - test_rz_langmuir_multi_psatd_multiJ.checksum (Failed)
	 82 - test_rz_load_external_field_grid.checksum (Failed)
	 86 - test_rz_load_external_field_grid_restart.checksum (Failed)
	 90 - test_rz_load_external_field_particles.checksum (Failed)
	 94 - test_rz_load_external_field_particles_restart.checksum (Failed)
	101 - test_rz_galilean_psatd.checksum (Failed)
	105 - test_rz_galilean_psatd_current_correction.checksum (Failed)
	109 - test_rz_galilean_psatd_current_correction_psb.checksum (Failed)
	112 - test_rz_multiJ_psatd.checksum (Failed)
	115 - test_rz_deuterium_tritium_fusion.analysis (Failed)
	116 - test_rz_deuterium_tritium_fusion.checksum (Failed)
	119 - test_rz_ohm_solver_cylinder_compression_picmi.checksum (Failed)
	123 - test_rz_ohm_solver_em_modes_picmi.checksum (Failed)
	134 - test_rz_point_of_contact_eb.analysis (Failed)
	147 - test_rz_scraping.checksum (Failed)
	151 - test_rz_scraping_filter.checksum (Failed)
	158 - test_rz_laser_acceleration.checksum (Failed)
	165 - test_rz_laser_acceleration_picmi.checksum (Failed)

### 3D

- `test_3d_point_of_contact_eb.analysis`

atmyers:
Now we are down to:

`test_rz_point_of_contact_eb.analysis`
`test_3d_point_of_contact_eb.analysis`

atmyers (approved):

Source/Diagnostics/ComputeDiagFunctors/BackTransformParticleFunctor.cpp
```diff
@@ -15,29 +15,39 @@
...
+      m_gammaboost(WarpX::gamma_boost), m_betaboost(WarpX::beta_boost),
+      m_Phys_c(PhysConst::c), m_inv_c2(amrex::Real(1.0)/(m_Phys_c * m_Phys_c)),
+      m_uzfrm(-m_gammaboost*m_betaboost*m_Phys_c)
 {
     using namespace amrex::literals;
 
-    if (tmp_particle_data.empty()) { return; }
+    // FIXME
+    //if (tmp_particle_data.empty()) { return; }
+
```
EZoni: @atmyers 

How do you think we should handle this `if` condition that is currently commented out?

Source/Particles/Pusher/CopyParticleAttribs.H
```diff
@@ -41,30 +39,34 @@ struct CopyParticleAttribs
...
-    CopyParticleAttribs (const WarpXParIter& a_pti, TmpParticles& tmp_particle_data,
-                         long a_offset = 0) noexcept
+    CopyParticleAttribs (
+        const WarpXParticleContainer& /*pc*/,
+        WarpXParIter& a_pti,
+        long a_offset = 0) noexcept
     {
-        if (tmp_particle_data.empty()) { return; }
+        // FIXME
+        //if (tmp_particle_data.empty()) { return; }
```
EZoni: @atmyers 

How do you think we should handle this `if` condition that is currently commented out?

Source/Diagnostics/ComputeDiagFunctors/BackTransformParticleFunctor.cpp
atmyers: I think we should replace this with an assertion that the BTD components are found in the input particle iterator.

atmyers: Actually it will trigger an assertion failure below, when it tries to access them. So I think this is fine to just remove.

EZoni:
@ax3l 

@atmyers fixed all remaining issues in this PR and approved the changes as well. I was the original author of the PR, so I cannot approve the changes myself, although @atmyers did most of the work here. Please let us know if you would like to leave a review on your end or if you would be okay with us merging the PR as is - either @atmyers or I could do it.

ax3l (approved):
That's a wonderful simplification of the code, thank you two! :rocket: :sparkles:

PR #5573, opened 2025-01-17
Move checkEarlyUnusedParams from WarpX class to anonymous namespace

lucafedeli88:
`checkEarlyUnusedParams` is only used once inside `WarpX::Evolve`.  Therefore, this PR moves the function from the WarpX class to an anonymous namespae inside `WarpXEvolve.cpp` .
The final goal is to reduce the complexity of the WarpX class.


ax3l (approved):
LGTM, thx

PR #5492, opened 2024-12-03
[WIP] CI: debug clang thread sanitizer errors

EZoni:
Debug data race conditions raised by the Clang thread sanitizer CI job that was disabled in #5474.

PR #5575, opened 2025-01-17
CI: install clang dependencies directly from LLVM

EZoni:
I think this was a suggestion by @WeiqunZhang in the context of debugging the clang sanitizer issue currently addressed in #5492. I'm extracting all related changes from #5492 to implement and test the LLVM installation separately here.

This effectively unifies the CI scripts to install clang dependencies into one single script that reads the clang version number from the command line.

I think all CI checks should pass here as a prerequisite for debugging the clang sanitizer issue further in #5492.


.github/workflows/dependencies/clang.sh
```diff
@@ -1,45 +1,50 @@
...
-# `man apt.conf`:
-#   Number of retries to perform. If this is non-zero APT will retry
-#   failed files the given number of times.
+# `man apt.conf`: number of retries to perform (if non-zero,
+# APT will retry failed files the given number of times).
 echo 'Acquire::Retries "3";' | sudo tee /etc/apt/apt.conf.d/80-retries
 
-# This dependency file is currently used within a docker container,
-# which does not come (among others) with wget, xz-utils, curl, git,
-# ccache, and pkg-config pre-installed.
```
EZoni, line 17: I think what was stated in this comment is not valid since we removed the Ubuntu 23.10 container in #5474.

ax3l (approved):
LGTM :)

.github/workflows/dependencies/clang.sh
```diff
@@ -1,45 +1,50 @@
...
-    git                 \
-    ccache              \
-    pkg-config
+    ninja-build
 
-# Use clang 17
-export CXX=$(which clang++-17)
-export CC=$(which clang-17)
+# parse clang version number from command line
+version_number=${1}
```
ax3l, line 26: Saw too late.

You could make the script more robust by checking that the number of arguments are exactly one (or two?) (0 is the program/script name, 1 the first argument) and otherwise throw a helpful error message. In bash, the number of args is `$#`

PR #5576, opened 2025-01-18
Added CI to test secondary ion emission in RZ.

oshapoval:
This PR adds secondary ion emission through a callback function, allowing secondary electrons to be emitted when an ion hits the embedded boundary. In the following CI test, the random seed was fixed to ensure consistent emission of secondary electrons for reproducibility. We used a secondary electron emission yield (SEY) of 0.4.


Examples/Tests/secondary_ion_emission/analysis.py
```diff
@@ -0,0 +1,62 @@
...
+    print("x=%5.5f, y=%5.5f, z=%5.5f" % (x_analytic[i], y_analytic[i], z_analytic[i]))
+
+    diff_x = np.abs((x[i] - x_analytic[i]) / x_analytic[i])
+    diff_y = np.abs((y[i] - y_analytic[i]) / y_analytic[i])
+    diff_z = np.abs((z[i] - z_analytic[i]) / z_analytic[i])
+
+    print("\n")
+    print("percentage error for x = %5.4f %%" % (diff_x * 100))
+    print("percentage error for y = %5.4f %%" % (diff_y * 100))
+    print("percentage error for z = %5.4f %%" % (diff_z * 100))
```
EZoni: I think these can be simplified a bit using f-strings:
```suggestion
    print(f"percentage error for x = {diff_x * 100:5.4f} %")
    print(f"percentage error for y = {diff_y * 100:5.4f} %")
    print(f"percentage error for z = {diff_z * 100:5.4f} %")
```

Examples/Tests/secondary_ion_emission/analysis.py
```diff
@@ -0,0 +1,62 @@
...
+    print(f"Electron # {i}:")
+    print("NUMERICAL coordinates of the emitted electrons:")
+    print("x=%5.5f, y=%5.5f, z=%5.5f" % (x[i], y[i], z[i]))
+    print("\n")
+    print("ANALYTICAL coordinates of the point of contact:")
+    print("x=%5.5f, y=%5.5f, z=%5.5f" % (x_analytic[i], y_analytic[i], z_analytic[i]))
+
+    diff_x = np.abs((x[i] - x_analytic[i]) / x_analytic[i])
+    diff_y = np.abs((y[i] - y_analytic[i]) / y_analytic[i])
+    diff_z = np.abs((z[i] - z_analytic[i]) / z_analytic[i])
```
EZoni: It could be a bit clearer to name these `err_<x,y,z>`, `rel_err_<x,y,z>`, or something along the lines of "error" or "relative error", rather than `diff_<x,y,z>`, which seems to be along the lines of "difference". I would use `diff_<x,y,z>` only for the numerator, which is the actual difference.

Examples/Tests/secondary_ion_emission/analysis.py
```diff
@@ -0,0 +1,62 @@
...
+tolerance = 1e-3
+
+for i in range(0, N_sec_e):
+    print("\n")
+    print(f"Electron # {i}:")
+    print("NUMERICAL coordinates of the emitted electrons:")
+    print("x=%5.5f, y=%5.5f, z=%5.5f" % (x[i], y[i], z[i]))
+    print("\n")
+    print("ANALYTICAL coordinates of the point of contact:")
+    print("x=%5.5f, y=%5.5f, z=%5.5f" % (x_analytic[i], y_analytic[i], z_analytic[i]))
```
EZoni: I think these can be simplified using f-strings:
```suggestion
    print(f"Electron # {i}:")
    print("NUMERICAL coordinates of the emitted electrons:")
    print(f"x={x[i]:5.5f}, y={y[i]:5.5f}, z={z[i]:5.5f}")
    print("\n")
    print("ANALYTICAL coordinates of the point of contact:")
    print(f"x={x_analytic[i]:5.5f}, y={y_analytic[i]:5.5f}, z={z_analytic[i]:5.5f}")
```

Examples/Tests/secondary_ion_emission/inputs_test_rz_secondary_ion_emission_picmi.py
```diff
@@ -0,0 +1,255 @@
...
+        if energy > 0.0:
+            sigma = (
+                delta_H
+                * (E_HMax + 1.0)
+                / (E_HMax * 1.0 + energy)
+                * np.sqrt(energy / 1.0)
+            )
+        else:
+            sigma = 0.0
+        sigma_nascap = np.append(sigma_nascap, sigma)
```
EZoni, line 179: Could you try to see if this one-line list comprehension works and is equivalent to your implementation?
```suggestion
    sigma = np.array([(delta_H * (E_HMax + 1.0) / (E_HMax + energy) * np.sqrt(energy) if energy > 0.0 else 0.0) for energy in energy_kEv])
 ```

**Review by EZoni — CHANGES_REQUESTED** [review]
Thanks, @oshapoval! I left some first few comments.

File: Examples/Tests/secondary_ion_emission/analysis.py (hunk 3 above)
Comment (oshapoval) [reply Examples/Tests/secondary_ion_emission/analysis.py]: Agreed with the above-mentioned comments. I was reusing an existing CI analysis script and will make changes here as well as separately to the initial CI test.

File: Examples/Tests/secondary_ion_emission/inputs_test_rz_secondary_ion_emission_picmi.py (hunk 4 above)
Comment (oshapoval) [reply Examples/Tests/secondary_ion_emission/inputs_test_rz_secondary_ion_emission_picmi.py:179]: The pre-commit hook enforces the repository’s style policy and divides it into multiple lines anyway, so I suggest leaving the code as-is instead of bypassing it.

File: Examples/Tests/secondary_ion_emission/analysis.py [hunk 5]
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env python
+"""
+This script tests the last coordinates of the emitted secondary electrons.
+The EB sphere is centered on O and has a radius of 0.2.
+The proton is initially at: (0,0,-0.25) and moves with a velocity:
+(0.1e6,0,1.5e6) with a time step of dt = 0.000000015.
+The simulation uses a fixed random seed (np.random.seed(10025015))
+to ensure the emission of secodnary electrons.
+An input file inputs_test_rz_secondary_ion_emission_picmi.py is used.
```
RemiLehe:
```suggestion
This script checks that electron secondary emission (implemented by a callback function) works as intended.

In this test, four ions hit a spherical embedded boundary, and produce secondary 
electrons with a probability of `0.4`. We thus expect ~2 electrons to be produced. 
This script tests the number of electrons emitted and checks that their position is 
close to the embedded boundary.
```

Examples/Tests/secondary_ion_emission/analysis.py
```diff
@@ -0,0 +1,70 @@
...
+
+it = ts.iterations
+x, y, z = ts.get_particle(["x", "y", "z"], species="electrons", iteration=it[-1])
+print("x", x)
+print("y", y)
+print("z", z)
+# Analytical results calculated
+# x_analytic = [0.004028, 0.003193]
+# y_analytic = [-0.0001518, -0.0011041]
+# z_analytic = [-0.19967, -0.19926]
```
RemiLehe:
```suggestion
```

Examples/Tests/secondary_ion_emission/analysis.py
```diff
@@ -0,0 +1,70 @@
...
+
+x_analytic = [-0.091696, 0.011599]
+y_analytic = [-0.002282, -0.0111624]
+z_analytic = [-0.200242, -0.201728]
+
+# x_analytic = [-0.09169647,  0.01159922]
+# y_analytic  = [-0.00228188, -0.01116237]
+# z_analytic = [-0.20024175, -0.20172786]
+
```
RemiLehe:
```suggestion
```

RemiLehe (approved):

EZoni (approved):

PR #5577, opened 2025-01-20
storePhiOnParticles: fix misplaced pragma omp parallel if

lucafedeli88:
This PR moves  `#pragma omp parallel if` from an outer loop on levels down to the loop on tiles.
It seems to me that threads should be used for tiles rather than for levels here.


RemiLehe (approved):

RemiLehe:
Thanks for fixing this!

PR #5582, opened 2025-01-21
WarpX.cpp : move free function TagWithLevelSuffix inside WarpX::AllocInitMultiFab

lucafedeli88:
`TagWithLevelSuffix` is used only once inside `WarpX::AllocInitMultiFab`, and it can be replaced with a one-liner. 
For this reason, in order to simplify the WarpX class, I would suggest implementing the functionality  directly inside `WarpX::AllocInitMultiFab`

### DIFF
--- a/Source/WarpX.cpp
+++ b/Source/WarpX.cpp
@@ -3314,14 +3314,6 @@ WarpX::isAnyParticleBoundaryThermal ()
     return false;
 }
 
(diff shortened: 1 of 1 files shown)


dpgrote (approved):
Looks good to me!

EZoni (approved):

PR #5585, opened 2025-01-22
Old MF Wrapper: Warn of no CuPy

ax3l:
There is an old code path that does not warn when it implicitly hopes that managed memory is enabled. This adds a warning.

In general, we should use the new pyAMReX-ported MultiFabWrapper, which is safer to use.

### DIFF
--- a/Python/pywarpx/fields.py
+++ b/Python/pywarpx/fields.py
@@ -303,6 +303,10 @@ class _MultiFABWrapper(object):
         # Note: transposing creates a view and not a copy.
         device_arr4 = self.mf.array(mfi)
         if libwarpx.libwarpx_so.Config.have_gpu:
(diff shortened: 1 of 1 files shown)


roelof-groenewald (commented):
To avoid the (albeit very small) overhead of executing `load_cupy` each time `_get_field` is called, `self.cupy_status` could instead be set during the `_MultiFABWrapper` initialization, and in the block that relies on managed memory we can then simply add
```
libwarpx.amr.Print(self.cupy_status)
```
What do you think?

dpgrote:
Note that this code is deleted in PR #5407.  If that PR is merged, then this PR is not needed.

RemiLehe:
Superseded by https://github.com/BLAST-WarpX/warpx/pull/5407.

PR #5592, opened 2025-01-22
Clang-tidy CI test: bump version from 15 to 16

lucafedeli88:
This PR bumps the version used for `clang-tidy` CI tests from 15 to 16.
It also addresses all the issues found with the upgraded tool.

Most of the issues are related to this newly introduced check:
- [cppcoreguidelines-avoid-const-or-ref-data-members](https://releases.llvm.org/16.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/cppcoreguidelines/avoid-const-or-ref-data-members.html)
The check enforces  [CppCoreGuidelines about constant and reference data members]( https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c12-dont-make-data-members-const-or-references) .
In general, I understand the argument against using constant or reference data members.
 ~~There is however one case in which I am not fully convinced by the suggestion of the tool: in [PML.H](https://github.com/ECP-WarpX/WarpX/pull/5592/files#diff-f1e020ebe3cd2222f399d50ff05769d0c70482f0e12bbe29b498e9ab2d0f4a53) `const amrex::BoxArray& m_grids;` becomes  `amrex::BoxArray m_grids;` and I am wondering if this can be an issue for performances. Maybe we could consider using a (possibly smart) pointer to  the `BoxArray`, instead of making a copy.~~ (we are now using a pointer for `amrex::BoxArray m_grids;`).

Few issues were instead related to these checks:
- [modernize-loop-convert](https://releases.llvm.org/16.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize/loop-convert.html)
This check was already enabled, but `clang-tidy-16` broadens its scope with respect to previous versions.

- [modernize-use-auto](https://releases.llvm.org/16.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize/use-auto.html)
Only one case. I am a bit confused because this should have been found also by version 15 of the tool.

- [misc-use-anonymous-namespace](https://releases.llvm.org/16.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/misc/use-anonymous-namespace.html)
This is a new check. Actually, the issues found with this check were false positives, but they disappeared when I properly set `misc-use-anonymous-namespace.HeaderFileExtensions` in the `.clang-tidy` configuration file to recognize `.H` files as headers.

- [misc-misplaced-const](https://releases.llvm.org/16.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/misc/misplaced-const.html)
Only one case. I am a bit confused because this should have been found also by version 15 of the tool.

- [readability-misleading-indentation](https://releases.llvm.org/16.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/readability/misleading-indentation.html)  [**NOW DISABLED DUE TO FALSE POSITIVES**]
This check was already enabled. However, with this newer version of the tool, it seems to me that it generates some false positives. Therefore, I would like to propose to **disable** it. We may try to re-enable it when we will bump the version from 16 to 17.


Source/BoundaryConditions/PML.H
```diff
@@ -112,7 +112,7 @@ public:
     }
 
 private:
-    const amrex::BoxArray& m_grids;
```
lucafedeli88, line 115: I am not fully convinced about this substitution. Maybe a (possibly smart) pointer would be better, in order to avoid copies.

lucafedeli88: What do you think?

EZoni:
@lucafedeli88 

Is there a specific reason why you want to move to version 16 as opposed to even newer versions?

lucafedeli88:
> @lucafedeli88
> 
> Is there a specific reason why you want to move to version 16 as opposed to even newer versions?

@EZoni ,  the main reason is to split into two PRs the changes that will be required with newer versions of `clang-tidy` .  This is also motivated by the fact that `clang-tidy-17` introduces a fair amount of  [new checks](https://releases.llvm.org/17.0.1/tools/clang/tools/extra/docs/ReleaseNotes.html#new-checks ) !  I have already opened a (WIP) PR to upgrade from 16 to 17 (  https://github.com/ECP-WarpX/WarpX/pull/5600 ). I expect that it will be finished by next week.
However, if you prefer a single PR, I can bump the version to 17 and address the issues that `clang-tidy-17` will find.

Source/BoundaryConditions/PML.H
ax3l: Hm, also should this change when we load balance or was the PML BoxArray special?

cc @WeiqunZhang @atmyers

Source/BoundaryConditions/PML.H
```diff
@@ -112,7 +112,7 @@ public:
...
+    amrex::BoxArray m_grids;
```
ax3l: Why not simply a ref?
```suggestion
    amrex::BoxArray & m_grids;
```

Source/BoundaryConditions/PML.H
WeiqunZhang: I think we can change it to `const amrex::BoxArray*` and we should not make a copy. The cost of copy itself is actually not that expensive because the data inside BoxArray are stored with shared_ptr. But we only need to use the BoxArray when constructing PML. After a PML object is constructed, we no long need that BoxArray. So there is no reason to hold a copy of it.

lucafedeli88: Thanks, @WeiqunZhang . I've followed your suggestion.

Source/BoundaryConditions/PML.H
lucafedeli88: @ax3l , the goal of [avoid-const-or-ref-data-members](
https://releases.llvm.org/16.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/cppcoreguidelines/avoid-const-or-ref-data-members.html) is to avoid using `const` or `ref` data members, on the grounds that
>  Having such members is rarely useful, and makes the class only copy-constructible but not copy-assignable

This recommendation is part of the [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c12-dont-make-data-members-const-or-references), and `clant-tidy` documentation suggests also this [additional reading](https://quuxplusone.github.io/blog/2022/01/23/dont-const-all-the-things/#data-members-never-const) on the topic.

That said, we are not forced to enforce this rule in WarpX (although I think that this rule makes sense in most cases), or we could make an exception specifically for this line. For the moment I've implemented @WeiqunZhang 's suggestion, but I can revert it and just add a `NOLINT`.

ax3l: Agreed, but my suggestion just adds `&` not `const` back :D

ax3l (approved):
LGTM :+1:

PR #5593, opened 2025-01-22
Only allocate `edge_lengths` and `face_areas` when using the ECT solver

RemiLehe:
Following #5558, the MultiFab `edge_lengths` and `face_areas` are not used anymore, expect for the ECT solver. Thus, only allocateing them when using the ECT solver saves a considerable amount of memory.

### DIFF
--- a/Examples/Tests/embedded_boundary_python_api/inputs_test_3d_embedded_boundary_picmi.py
+++ b/Examples/Tests/embedded_boundary_python_api/inputs_test_3d_embedded_boundary_picmi.py
@@ -43,7 +43,7 @@ grid = picmi.Cartesian3DGrid(
 
 flag_correct_div = False
 
-solver = picmi.ElectromagneticSolver(grid=grid, method="Yee", cfl=1.0)
+solver = picmi.ElectromagneticSolver(grid=grid, method="ECT", cfl=1.0)
 
 n_cavity = 30
 L_cavity = n_cavity * unit
--- a/Source/FieldSolver/FiniteDifferenceSolver/FiniteDifferenceSolver.H
+++ b/Source/FieldSolver/FiniteDifferenceSolver/FiniteDifferenceSolver.H
@@ -168,11 +168,7 @@ class FiniteDifferenceSolver
           *
           * \param[out] Jfield  vector of current MultiFabs at a given level
           * \param[in] Bfield   vector of magnetic field MultiFabs at a given level
-<<<<<<< HEAD
           * \param[in] eb_update_E indicate in which cell E should be updated (related to embedded boundaries)
-=======
-          * \param[in] eb_update_E indicate in which cell E should be udpated (related to embedded boundaries)
->>>>>>> 8136934a5 (Remove edge_length and faces)
           * \param[in] lev  level number for the calculation
           */
         void CalculateCurrentAmpere (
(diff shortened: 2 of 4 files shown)


Examples/Tests/embedded_boundary_python_api/inputs_test_3d_embedded_boundary_picmi.py
```diff
@@ -43,7 +43,7 @@
 
 flag_correct_div = False
 
-solver = picmi.ElectromagneticSolver(grid=grid, method="Yee", cfl=1.0)
+solver = picmi.ElectromagneticSolver(grid=grid, method="ECT", cfl=1.0)
```
RemiLehe, line 46: This test fails when using the "Yee" solver, because the MultiFab that are accessed with PICMI do not exist ; not sure how we want to handle this type of error in general.

Source/FieldSolver/FiniteDifferenceSolver/FiniteDifferenceSolver.H
```diff
@@ -147,7 +147,11 @@ class FiniteDifferenceSolver
           * \param[in] Bfield   vector of magnetic field MultiFabs at a given level
           * \param[in] rhofield scalar ion charge density Multifab at a given level
           * \param[in] Pefield  scalar electron pressure MultiFab at a given level
+<<<<<<< HEAD
           * \param[in] eb_update_E indicate in which cell E should be updated (related to embedded boundaries)
+=======
+          * \param[in] eb_update_E indicate in which cell E should be udpated (related to embedded boundaries)
+>>>>>>> 8136934a5 (Remove edge_length and faces)
```
RemiLehe:
```suggestion
          * \param[in] eb_update_E indicate in which cell E should be updated (related to embedded boundaries)
```

Source/FieldSolver/FiniteDifferenceSolver/FiniteDifferenceSolver.H
```diff
@@ -168,7 +168,11 @@ class FiniteDifferenceSolver
           *
           * \param[out] Jfield  vector of current MultiFabs at a given level
           * \param[in] Bfield   vector of magnetic field MultiFabs at a given level
+<<<<<<< HEAD
           * \param[in] eb_update_E indicate in which cell E should be updated (related to embedded boundaries)
+=======
+          * \param[in] eb_update_E indicate in which cell E should be udpated (related to embedded boundaries)
+>>>>>>> 8136934a5 (Remove edge_length and faces)
```
RemiLehe:
```suggestion
          * \param[in] eb_update_E indicate in which cell E should be updated (related to embedded boundaries)
```

Examples/Tests/embedded_boundary_python_api/inputs_test_3d_embedded_boundary_picmi.py
roelof-groenewald: One way to handle this would be to create a python accessible function that would create those multifabs if they don't exist and return them. But that would still require this test to change since in that case Python would have to take ownership of the multifabs themselves, not just wrappers for them. I think it is okay to not support that at this time. If it turns out there are users that rely on those multifabs we can fix it when they come forward. How does that sound?

roelof-groenewald (approved):

Examples/Tests/embedded_boundary_python_api/inputs_test_3d_embedded_boundary_picmi.py
RemiLehe: OK, that works.

PR #5595, opened 2025-01-22
Virtual photons 

aeriforme:
This PR adds the option of creating a virtual photon population associated to a given electron or positron species. 
The virtual photons are generated from scratch at every timestep.
They are initialized in the same position as the parent particle.
Their momenta is sampled from the theoretical analytical spectrum (see equation 99.16 in Berestetskii, Pitaevskii, Lifshitz. Quantum Electrodynamics: Volume 4.)

The virtual photons can be used in linear Breit-Wheeler or linear Compton collisions. This is used to simulate incoherent beam-beam effects at the interaction point of colliders.

To do:
- [x] create a virtual photon species with the correct number of particles and with attributes same as primary particles
- [x] draw the virtual photon energy from the correct probability distribution
- [x] add infrastructure
- [x] add docs and test

Note: As of this PR, the virtual photons are technically not virtual at all, as their virtuality is zero. A more accurate name would be "equivalent quanta". However, follow-up PRs will soon add virtuality calculations.

🧨


Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -0,0 +1,147 @@
...
+            // and fill the corresponding vector
+            amrex::ParallelFor(num,
+             [=] AMREX_GPU_DEVICE (amrex::Long i) noexcept
+            {
+                amrex::ParticleReal ux = soa.m_rdata[PIdx::ux][i];
+                amrex::ParticleReal uy = soa.m_rdata[PIdx::uy][i];
+                amrex::ParticleReal uz = soa.m_rdata[PIdx::uz][i];
+
+                // same as Xsuite and guinea-pig
+                amrex::ParticleReal gamma = std::sqrt( 1.0_rt + ux*ux + uy*uy + uz*uz);
```
RemiLehe:
```suggestion
                amrex::ParticleReal gamma = std::sqrt( 1.0_rt + (ux*ux + uy*uy + uz*uz)*invclightsq );
```
This is admittedly confusing, but WarpX stores `u` in units of `m/s` (i.e. `u` is `gamma*beta*c`), even though the user enters `u` as dimensionless. See for instance:
https://github.com/BLAST-WarpX/warpx/blob/development/Source/Particles/Pusher/UpdateMomentumVay.H#L34

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -0,0 +1,147 @@
...
+
+                amrex::Long n_photons = (amrex::Long)r_photons;
+
+                num_vp_data[i] = n_photons;
+            });
+
+            // compute the offsets vector as the cumulative sum of the elements of num_vp,
+            // excluding the current element
+            amrex::Gpu::DeviceVector<amrex::Long> offsets_vp(num);
+            amrex::Gpu::exclusive_scan(num_vp.begin(), num_vp.end(), offsets_vp.begin());
```
RemiLehe: Do we need this call here? 
It seems that this is redundant with
```
total_num_vp = amrex::Scan::ExclusiveSum(num_vp.size(), num_vp.data(), offsets_vp.data());
```

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -0,0 +1,147 @@
...
+                amrex::ParticleReal y  = soa.m_rdata[PIdx::y][i];
+                amrex::ParticleReal z  = soa.m_rdata[PIdx::z][i];
+                amrex::ParticleReal w  = soa.m_rdata[PIdx::w][i];
+                uint64_t pid = soa.m_idcpu[i];
+
+                for (int j = 0; j < num_vp_data[i]; j++)
+                {
+                    // photon index for the current primary
+                    amrex::Long ip = offset_vp_data[i] + j;
+                    pa_vp[PIdx::ux][ip] = ux;
```
RemiLehe: We need to be careful here because the way in which we store `u` for photons and for other particles is different: for photons, I think we use `u = p/m_e`, where (somewhat arbitrarily) `m_e` is the mass of the electron.

RemiLehe: I guess this part will be replaced anyway by something where we regenerate the momentum of the photons...

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -0,0 +1,147 @@
...
+                    pa_vp[PIdx::uy][ip] = uy;
+                    pa_vp[PIdx::uz][ip] = uz;
+                    pa_vp[PIdx::x][ip] = x;
+                    pa_vp[PIdx::y][ip] = y;
+                    pa_vp[PIdx::z][ip] = z;
+                    pa_vp[PIdx::w][ip] = w;
+                    pid_vp[ip] = pid;
```
RemiLehe: We can discuss this further, but I think it would be better to generate a new ID for the new photons.

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -0,0 +1,147 @@
...
+                }
+            });
+        } // mfi
+    } // lev
+
+    amrex::Gpu::synchronize();
```
RemiLehe: I don't think that this is needed.

dpgrote:
In comparison, I add a routine that generates photons in the PR #5537.

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
aeriforme: Just saw this, I did realize it eventually

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
aeriforme: Discussed offline and agreed

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -0,0 +1,229 @@
...
+
+                    amrex::ParticleReal gamma_primary = std::sqrt( 1.0_rt + (ux_primary*ux_primary + uy_primary*uy_primary + uz_primary*uz_primary)*inv_c2 );
+                    // minimum fractional (wrt primary particle) photon energy
+                    amrex::Real y_min = vphoton_min_energy / (mass * gamma_primary * PhysConst::c * PhysConst::c);
+                    // primary position, weight and id
+                    amrex::ParticleReal x  = soa.m_rdata[PIdx::x][i];
+                    amrex::ParticleReal y  = soa.m_rdata[PIdx::y][i];
+                    amrex::ParticleReal z  = soa.m_rdata[PIdx::z][i];
+                    amrex::ParticleReal w  = soa.m_rdata[PIdx::w][i];
+                    uint64_t pid = soa.m_idcpu[i]; // to be discussed
```
atmyers: Do you need to be able to compute the primary particle id from the id of the virtual particles generated from it, or do we just need to make sure they all get an unique number here?

aeriforme: Not just yet, but yes. I will need to updated the parent particle based on what happened to the virtual photon.

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -0,0 +1,248 @@
...
+        utils::parser::getWithParser(pp_species_name, "virtual_photons_min_energy", vphoton_min_energy);
+
+        // Sampling factor:
+        // the number of virtual photons generated is multiplied by this factor,
+        // the weight of each virtual photon is divided by this factor
+        amrex::Real sampling_factor;
+        utils::parser::getWithParser(pp_species_name, "sampling_factor", sampling_factor);
+
+        // Enable tiling
+        amrex::MFItInfo info;
```
atmyers: Could this be the reason it fails when `particles.do_tiling = 0`? It looks like this loop always has tiling enabled. But those tiles might not exist in the underyling particle containers. I think we only want to enable tiling in this loop if `particles.do_tiling = 1`.

Docs/source/usage/parameters.rst
```diff
@@ -1535,6 +1535,23 @@ Particle initialization
     in each dimension with a matched shape function and filtering used for current deposition.
     This is required when using the electron energy solver with electron-ion temperature relaxation.
 
+* ``<species>.do_qed_virtual_photons`` (`boolean`) optional (default `false`)
+    Create a population of virtual photons associated to ``<species>``.
```
atmyers:
```suggestion
    Create a population of virtual photons associated with ``<species>``.
```

Docs/source/usage/parameters.rst
```diff
@@ -1535,6 +1535,23 @@ Particle initialization
...
+    It only works if ``<species>`` is an electron or a positron species.
+    The virtual photon species has to be created as a regular photon species in the input file.
+    Virtual photons are created from scratch at each timestep in the same position of the parent particle.
```
atmyers:
```suggestion
    Virtual photons are created from scratch at each timestep in the same position as the parent particle.
```

Docs/source/usage/parameters.rst
```diff
@@ -1535,6 +1535,23 @@ Particle initialization
...
+    The energy of the virtual photons is sampled from their spectrum (see :cite:t:`LandauVol4` section 99 for more details).
+    The momentum of the virtual photons is parallel to that of the parent particle.
+    This feature also requires the following input parameters:
+      * ``<species>.qed_virtual_photon_species_name`` (`string`) name of the virtual photon species associated to the current lepton species.
```
atmyers:
```suggestion
      * ``<species>.qed_virtual_photon_species_name`` (`string`) name of the virtual photon species associated with the current lepton species.
```

Docs/source/usage/parameters.rst
```diff
@@ -1535,6 +1535,23 @@ Particle initialization
...
+      * ``<virtual_photon_species>.qed_virtual_photon_min_energy`` (`float`, in Joules) minimum energy of the virtual photons
+      * ``<virtual_photon_species>.sampling_factor`` (`int`), sampling factor for the virtual photons.
+        A sampling factor of ``f`` means that the number of virtual photons is multiplied by ``f``, while their weights are divided by ``f``.
+    The virtual photons can undergo collisions via the linear Breit-Wheeler or linear Compton processes.
+    This is useful to model incoherent beam-beam effects in colliders (e.g. pair generation, radiative Bhabha scattering).
+    This QED feature is separated from the strong-field QED modules (quantum synchrotron and non-linear Breit-Wheeler).
+    It requires to compile with QED=TRUE.
```
atmyers:
```suggestion
    It requires WarpX to be compiled with QED=TRUE.
```

Examples/Tests/virtual_photons/analysis_virtual_photons.py
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env python3
+
+# This test generatees the population of virtual photons
```
atmyers:
```suggestion
# This test generates the population of virtual photons
```

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -0,0 +1,244 @@
...
+#elif defined (WARPX_DIM_RCYLINDER)
+                    amrex::ParticleReal x  = soa.m_rdata[PIdx::x][i];
+                    amrex::ParticleReal theta  = soa.m_rdata[PIdx::theta][i];
+#elif defined(WARPX_DIM_RSPHERE)
+                    amrex::ParticleReal x  = soa.m_rdata[PIdx::x][i];
+                    amrex::ParticleReal theta  = soa.m_rdata[PIdx::theta][i];
+                    amrex::ParticleReal phi  = soa.m_rdata[PIdx::phi][i];
+#endif
+                    amrex::ParticleReal w  = soa.m_rdata[PIdx::w][i];
+                    uint64_t pid = soa.m_idcpu[i]; // TODO: use this to update the parent lepton if needed
```
atmyers: Is this TODO comment still needed? If this has been done, you can remove the comment.

atmyers:
This looks good! I left a few proofreading comments.

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
aeriforme: No, this would be for future possible enhancement.
Mostly a note to remember why the virtual photons don't have their own PIDs.

atmyers (approved):

Source/Particles/MultiParticleContainer.cpp
```diff
@@ -1821,6 +1827,12 @@ void MultiParticleContainer::CheckQEDProductSpecies()
                     AmIA<PhysicalSpecies::photon>(),
                 "ERROR: Quantum Synchrotron product species is of wrong type");
         }
+        if(pc->has_virtual_photons()){
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                allcontainers[pc->m_qed_virtual_photon_species]->
+                    AmIA<PhysicalSpecies::photon>(),
+                "ERROR: virtual photons species has to be a...photon species!");
+        }
```
EZoni, line 1872: Indeed, virtual photons must be photons. Do we envision/plan to add new virtual particles down the road (bosons for things like scattering or fermions for things like one-loop diagrams)?

If not, or not in the near future, I wonder if we should just hard-code the species type in the source code, rather than exposing it to the user through the input parameter `<virtual_photon_species>.species_type` (which I see in the input file of the new test) and then asserting with error. Or is this not possible because of some inheritance situation (i.e., `species_type` comes from a parent class)?

Docs/source/usage/parameters.rst
```diff
@@ -1535,6 +1535,23 @@ Particle initialization
...
+    The energy of the virtual photons is sampled from their spectrum (see :cite:t:`LandauVol4` section 99 for more details).
+    The momentum of the virtual photons is parallel to that of the parent particle.
+    This feature also requires the following input parameters:
+      * ``<species>.qed_virtual_photon_species_name`` (`string`) name of the virtual photon species associated with the current lepton species.
+      * ``<virtual_photon_species>.qed_virtual_photon_min_energy`` (`float`, in Joules) minimum energy of the virtual photons
+      * ``<virtual_photon_species>.sampling_factor`` (`int`), sampling factor for the virtual photons.
+        A sampling factor of ``f`` means that the number of virtual photons is multiplied by ``f``, while their weights are divided by ``f``.
+    The virtual photons can undergo collisions via the linear Breit-Wheeler or linear Compton processes.
+    This is useful to model incoherent beam-beam effects in colliders (e.g. pair generation, radiative Bhabha scattering).
+    This QED feature is separated from the strong-field QED modules (quantum synchrotron and non-linear Breit-Wheeler).
```
EZoni, line 1562: Maybe we have reached enough complexity in the processes that we support that showing the corresponding Feynman diagrams could be helpful. Either directly here or in a referenced section where we list all processes. What do you think? I haven't checked if we already have some in other parts of the documentation.

Docs/source/usage/parameters.rst
```diff
@@ -1535,6 +1535,23 @@ Particle initialization
...
+    It requires WarpX to be compiled with QED=TRUE.
```
EZoni: I think the CMake build option is called `WarpX_QED`, `QED=TRUE` should be the GNU Make syntax:
```suggestion
    It requires WarpX to be compiled with ``WarpX_QED=ON`` (CMake) or ``QED=TRUE`` (GNU Make).
```

Examples/Tests/virtual_photons/analysis_virtual_photons.py
```diff
@@ -0,0 +1,91 @@
...
+###########################
+
+# useful constants
+GeV = 1e9 * eV
+gamma_em = 0.5772156649015  # Euler-Mascheroni
+
+# electron energy
+energy = 125 * GeV
+gamma = energy / (m_e * c**2)
+gamma0 = 125 * GeV / (m_e * c**2)
```
EZoni: If I'm not mistaken `gamma_em` appears only in code that is commented out below (line 63),
```py
A = 0  # log(4) - 2 * gamma_em - 1 (if not neglecting some terms)
```
while `gamma` and `gamma0` (which I believe equal each other) are not used. Could they be removed?

Examples/Tests/virtual_photons/analysis_virtual_photons.py
```diff
@@ -0,0 +1,91 @@
...
+
+# min fractional energy of the virtual photon wrt electron energy
+ymin = hw_min / energy
+
+#############
+### WarpX ###
+#############
+
+series = OpenPMDTimeSeries("./diags/diag1/")
+sampling_factor = 10000000
```
EZoni: I learned recently that this can be written also as `10_000_000`, which may be more readable. `int(1e7)` could be another option (if it has to be an integer).

Examples/Tests/virtual_photons/analysis_virtual_photons.py
```diff
@@ -0,0 +1,91 @@
...
+y = np.geomspace(ymin, 1, 401)
+
+# number of virtual photons per electron obtained with WarpX
+N_wx = np.sum(w_wx) / np.sum(w_primary_wx)
+
+# spectrum of the virtual photons per electron
+H, b = np.histogram(y_wx, bins=y, weights=w_wx)
+db = np.diff(b)
+b = 0.5 * (b[1:] + b[:-1])
+dN_dy_wx = H / db / np.sum(w_primary_wx)
```
EZoni: The following could be alternative names for the variables used here (`old`: `new` syntax):
- `uz`: `uz_virtual_photons`
- `w_wx`: `w_virtual_photons`
- `w_primary_wx`: `w_electrons`
- `np.sum(w_wx)`: `total_virtual_photons`
- `np.sum(w_primary_wx)`: `total_electrons`

It's possible that I'm misunderstanding the meaning of some variables. I also came up with this without fully understanding the current naming conventions, e.g., what the tag `_wx` is meant to denote, etc.

Examples/Tests/virtual_photons/analysis_virtual_photons.py
```diff
@@ -0,0 +1,91 @@
...
+print("Number of virtual photons per electron:")
+print(f"From simulation : {N_wx}")
+print(f"From theory     : {N}")
+print(f"Relative error  : {abs(N_wx - N) / N:.2%}")
+
+print("Spectrum of virtual photons per electron:")
+print(f"Max relative error: {(np.abs(dN_dy_wx - dN_dy) / dN_dy).max()}")
+assert (np.abs(dN_dy_wx - dN_dy) < 0.04 * dN_dy).all()
+
+assert abs(N - N_wx) < 0.02 * N
```
EZoni: Here, instead of retyping the formulas for the relative errors twice (for logging and assertion), it could be handy to assign them to variables, e.g.,
- `number_rel_error = abs(N_wx - N) / N`
- `spectrum_rel_error = np.abs(dN_dy_wx - dN_dy) / dN_dy`

in which case the assertions would simply read 
- `assert (number_rel_error < 0.02`)
- `assert (spectrum_rel_error < 0.04).all()`

P.S. I know I'm being a bit pedantic...

Examples/Tests/virtual_photons/inputs_test_3d_virtual_photons
```diff
@@ -0,0 +1,105 @@
...
+#################################
+######### DIAGNOSTICS ###########
+#################################
+# FULL
+diagnostics.diags_names = diag1
+
+diag1.intervals = -1
+diag1.diag_type = Full
+diag1.write_species = 1
+diag1.dump_last_timestep = 1
```
EZoni: I think `<diag_name>.dump_last_timestep` is true by default, just in case you want to make the input file as minimal as possible:
https://github.com/BLAST-WarpX/warpx/blob/c00dbeb96278393243fa06de275690670646c0e5/Source/Diagnostics/Diagnostics.H#L267-L268

Examples/Tests/virtual_photons/inputs_test_3d_virtual_photons
```diff
@@ -0,0 +1,105 @@
...
+beam.ux_m = 0.0
+beam.ux_th = 0.0
+beam.uy_th = 0.0
+beam.uz_th = 0
+beam.do_not_deposit = 1
+beam.do_not_push = 1
+beam.do_qed_virtual_photons = 1
+beam.qed_virtual_photon_species_name = virtual_photons
+
+virtual_photons.species_type = photon
```
EZoni, line 85: This is the line I referred to in https://github.com/BLAST-WarpX/warpx/pull/5595#discussion_r2369778864.

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -0,0 +1,244 @@
+/* Copyright 2021 Neil Zaim
```
EZoni: I have expressed my preference for removing the list of authors at the top of the files a few times in the past years, lastly in https://github.com/BLAST-WarpX/warpx/pull/6102#discussion_r2294962438 (in case you have an opinion and want to chime in), but since the majority of developers prefer to keep it or don't have a preference, I suggest to fix it in this case:
```suggestion
/* Copyright 2025 Arianna Formenti
```

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
EZoni: @atmyers 

Could you confirm that [ae03cc2](https://github.com/BLAST-WarpX/warpx/pull/5595/commits/ae03cc2b1e3f9ec2be94f6d083bbaff7bde0814e) resolved your request?

Source/Particles/MultiParticleContainer.cpp
aeriforme: Ah yes, in principle we could hard-code the type.
I don't know if this fix is as simple as it seems, though.
I can try it.

In any case, for photon species, we always have to define the type and add it to the photon species list with `particles.photon_species = ...`. 

So, if we hard-code the type, then the virtual photons would be the only photon species that doesn't need the type but still needs to be listed as photon species.

So I wonder if this should be a separate PR that streamlines the way photons are treated altogether.

What do you think?

Docs/source/usage/parameters.rst
aeriforme: Of course, I still need to document everything in detail. 
This is high up in the to-do list, but it's going to be in other separate PRs.
Waiting for the next documentathon...

Examples/Tests/virtual_photons/analysis_virtual_photons.py
aeriforme: Yes, I left it there for future me, just in case, but I can remove it.

Source/Particles/MultiParticleContainer.cpp
EZoni: I agree, I think it could be a separate PR for all types of photons. Personally I would be in favor of the hard-coded species type for cases where only one option is possible. Not sure what others think.

Docs/source/usage/parameters.rst
EZoni: Okay, sounds good.

Examples/Tests/virtual_photons/analysis_virtual_photons.py
aeriforme: It doesn't have to be an integer, so `1e7` is good

Examples/Tests/virtual_photons/analysis_virtual_photons.py
aeriforme: Ah and yes, the gammas are not needed!

Examples/Tests/virtual_photons/analysis_virtual_photons.py
aeriforme: Ah `_wx` is for WarpX, to distinguish it from the theory calculation. I made it more clear

Examples/Tests/virtual_photons/analysis_virtual_photons.py
aeriforme: The choice here was to avoid divisions, but I'm not sure it strictly necessary

EZoni:
Just FYI, I think the CI check `docs/readthedocs.org:warpx` fails because this branch does not include the fix we merged from #6151. You would need to merge `development` into this branch, but I think it's fine to do it later on when we're close to merging the PR.

Examples/Tests/virtual_photons/analysis_virtual_photons.py
EZoni: I see, that's fine by me. But just for my understanding, what is the problem with divisions in this case?

EZoni (approved):
Looks good to me. @RemiLehe, I let you merge this, since I think you wanted to review as well. I am not familiar with the details of the context behind the need for virtual photons at this point, so my review was relatively superficial.

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
aeriforme: I confirm!

Examples/Tests/virtual_photons/analysis_virtual_photons.py
aeriforme: Actually, no problem. It was a problem in previous iterations. 
I updated it.

Docs/source/usage/parameters.rst
```diff
@@ -1544,6 +1544,23 @@ Particle initialization
     in each dimension with a matched shape function and filtering used for current deposition.
     This is required when using the electron energy solver with electron-ion temperature relaxation.
 
+* ``<species>.do_qed_virtual_photons`` (`boolean`) optional (default `false`)
+    Create a population of virtual photons associated with ``<species>``.
+    It only works if ``<species>`` is an electron or a positron species.
+    The virtual photon species has to be created as a regular photon species in the input file.
+    Virtual photons are created from scratch at each timestep in the same position as the parent particle.
```
```suggestion
    Virtual photons are created from scratch at each timestep at the same position as the parent particle.
```

Docs/source/usage/parameters.rst
```diff
@@ -1544,6 +1544,23 @@ Particle initialization
...
+    The energy of the virtual photons is sampled from their spectrum (see :cite:t:`LandauVol4` section 99 for more details).
+    The momentum of the virtual photons is parallel to that of the parent particle.
+    This feature also requires the following input parameters:
+      * ``<species>.qed_virtual_photon_species_name`` (`string`) name of the virtual photon species associated with the current lepton species.
+      * ``<virtual_photon_species>.qed_virtual_photon_min_energy`` (`float`, in Joules) minimum energy of the virtual photons
```
RemiLehe, line 1557: In the example script, this is `virtual_photon_min_energy` instead of `qed_virtual_photon_min_energy`. For consistency, I would vote to keep things as is in the documentation (i.e. `qed_virtual_photon_min_energy`) but modify the example and the parsing code.

Docs/source/usage/parameters.rst
```diff
@@ -1544,6 +1544,23 @@ Particle initialization
...
+      * ``<virtual_photon_species>.sampling_factor`` (`int`), sampling factor for the virtual photons.
```
RemiLehe: Should we make this
```suggestion
      * ``<virtual_photon_species>.qed_virtual_photon_sampling_factor`` (`int`), sampling factor for the virtual photons.
```
for consistency?
Or altenatively, if could be:
```
<species>.qed_virtual_photon_multiplier
```
for consistency with the similar `multiplier` used in fusion, dsmc (not quite the same because it involves collisions, but my impression is that the concept is basically the same)

Examples/Tests/virtual_photons/inputs_test_3d_virtual_photons
```diff
@@ -0,0 +1,104 @@
...
+virtual_photons.injection_style = none
+virtual_photons.do_not_push = 1
+virtual_photons.do_not_gather = 1
+virtual_photons.do_not_deposit = 1
```
RemiLehe, line 89: I am not sure whether we want to keep `do_not_deposit` here.

On one hand, it emphasizes that these photons do not participate in the PIC loop.

On the other hand, photons never deposit: https://github.com/BLAST-WarpX/warpx/blob/development/Source/Particles/PhotonParticleContainer.H#L94

~~Moreover, my understanding is that the virtual photons are anyway destroyed immediately after the collisions, so that they never get to the push/gather/deposit function, so that these flags are in principle superfluous. Is this correct?~~

Source/Particles/PhysicalParticleContainer.H
```diff
@@ -368,6 +368,20 @@ public:
...
+     * Tells if this PhysicalParticleContainer has
+     * virtual photons enabled
+     * @return true if process is enabled
+     */
+    bool has_virtual_photons () const override;
+
+    /**
+     * Gets the index of the PhysicalParticleContainer
+     * that corresponds to the virtual photons
+     * @return index of the
```
RemiLehe: It seems that this is incomplete.

Source/Particles/Collision/CollisionHandler.cpp
```diff
@@ -108,6 +112,11 @@ CollisionHandler::CollisionHandler(MultiParticleContainer const * const mypc)
  */
 void CollisionHandler::doCollisions ( int step, amrex::Real cur_time, amrex::Real dt, MultiParticleContainer* mypc)
 {
+#ifdef WARPX_QED
```
RemiLehe, line 115: Could you add a comment here explaining this code for developers that are not familiar with QED?
e.g. 
```
For QED incoherent processes (e.g. Bethe-Heitler, Landau-Lifschitz), the process is mediated by virtual photons.
The virtual photons are newly generated here and participate in the collisions.
```

Examples/Tests/virtual_photons/inputs_test_3d_virtual_photons
RemiLehe: Ah, I now see that the virtual photons are not immediately destroyed (which allows output of the virtual photons, which the test relies on). But doesn't this mean that the virtual photons will accumulate from timestep to timestep?

In any case, maybe my above comment can be ignored for now, and we could discuss this offline.

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -0,0 +1,244 @@
...
+        if(!primary.has_virtual_photons()){
+            continue;
+        }
+
+        // Get the virtual photon species corresponding to this primary species
+        int vphotons_index = primary.getVirtualPhotonSpeciesIndex();
+        auto& vphotons = mypc->GetParticleContainer(vphotons_index);
+        const amrex::ParmParse pp_species_name(mypc->GetSpeciesNames()[vphotons_index]);
+
+        // Minimum allowed enrgy of the virtual photons
```
RemiLehe:
```suggestion
        // Minimum allowed energy of the virtual photons
```

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -0,0 +1,244 @@
...
+        amrex::Real vphoton_min_energy;
+        utils::parser::getWithParser(pp_species_name, "virtual_photons_min_energy", vphoton_min_energy);
+
+        // Sampling factor:
+        // the number of virtual photons generated is multiplied by this factor,
+        // the weight of each virtual photon is divided by this factor
+        amrex::Real sampling_factor;
+        utils::parser::getWithParser(pp_species_name, "sampling_factor", sampling_factor);
+
+        amrex::Real const alpha_pi = PhysConst::alpha / MathConst::pi;
```
RemiLehe: Could you use `alpha_over_pi` here, to be more explicit.

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -0,0 +1,244 @@
...
+                // First pass: compute the number of virtual photons for each primary particle
+                // and fill the corresponding vector
+                amrex::ParallelForRNG(num,
+                [=] AMREX_GPU_DEVICE (amrex::Long i, amrex::RandomEngine const& engine) noexcept
+                {
+                    amrex::ParticleReal ux = soa.m_rdata[PIdx::ux][i]; // u=v*gamma=p/m_e
+                    amrex::ParticleReal uy = soa.m_rdata[PIdx::uy][i];
+                    amrex::ParticleReal uz = soa.m_rdata[PIdx::uz][i];
+
+                    // Similar to Jackson's formula 15.58, but neglect longitudinal field, assume relativistic velocities, and integrate in energy
```
RemiLehe: Is there a more direct reference for the formula that we use here?

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -0,0 +1,244 @@
...
+                    amrex::ParticleReal lny = std::log( y_min );
+                    // Number of virtual photons per primary particle
+                    amrex::Real r_photons = alpha_pi * lny * lny * sampling_factor;
+                    amrex::Real i_photons; // integer part
+                    amrex::Real f_photons = std::modf(r_photons, &i_photons); // fractional part
+                    amrex::Long n_photons = (amrex::Long)i_photons;
+
+                    // Account for the fractional part of r_photons by
+                    // randomly adding an extra virtual photon
+                    if(amrex::Random(engine) < f_photons){ n_photons += amrex::Long(1);}
```
RemiLehe: I think that this code can be simplified with 
```suggestion
                    // `n_photons` must be an integer, but must average to `r_photons` over many realizations
                    // This is achieved by adding a random number between 0 and 1, and taking the integer part.
                    amrex::Long n_photons = (amrex::Long)( r_photons + amrex::Random(engine) );
```

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -0,0 +1,244 @@
...
+                // i.e., offsets[i] = sum_{j=0}^{i-1} num_vp[j],
+                // and return the total number of virtual photons to be generated in the current tile
+                // (which is the last element of the offsets vector)
+                amrex::Gpu::DeviceVector<amrex::Long> offsets_vp(num);
+                const amrex::Long total_num_vp = amrex::Scan::ExclusiveSum(num_vp.size(), num_vp.data(), offsets_vp.data());
+                auto *const offset_vp_data = offsets_vp.dataPtr();
+
+                // Now we can allocate and build the virtual photon species in the current tile
+                ParticleTileType& ptile_vp = vphotons.ParticlesAt(lev, mfi);
+                ptile_vp.resize(total_num_vp);
```
RemiLehe, line 144: Could you add a comment saying that this will erase/overwrite an virtual photons that may have been generated previously? (e.g. at the previous timestep)

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -0,0 +1,244 @@
...
+#include <cmath>
+
+
+namespace collision::binarycollision::virtualphotons{
+
+using namespace amrex::literals;
+using SoaData_type = typename WarpXParticleContainer::ParticleTileType::ParticleTileDataType;
+
+
+/** Generate virtual photons from primary particles (electrons or positrons)
```
RemiLehe, line 33: Could you add that this also erases/overwrites the virtual photons that we previously generated by `mypc` (e.g. at the previous timestep)

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -0,0 +1,244 @@
...
+ * up to the energy of the primary particle.
+ * The photons are assumed to be emitted in the direction of motion of the primary particle.
+ * The virtual photons are generated in a separate species, which can then be used
+ * as input for other processes, e.g., Breit-Wheeler pair production and Compton scattering.
+ * The virtual photons are generated at the same position as the primary particle,
+ * and inherit the same weight as the primary particle, divided by a sampling factor.
+ * The virtual photons are given the same particle ID as the primary particle.
+ * The virtual photon species is regenerated from scratch at each time step.
+ * In the current implementation, the primary particles are not affected by
+ * the possible interactions that the virtual photons may undergo.
```
RemiLehe, line 50: As a follow-up PR: Could you add the above info in the Theory section?

Source/Particles/Collision/CollisionHandler.cpp
RemiLehe: Also: could we add an indication that this erases any previously generated virtual photons? 
(e.g. should we rename this function to `Regenerate...` instead of `Generate`?)

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
RemiLehe: Btw, it seems that we could therefore have an issue if several species create virtual photons in the same species. (i.e. the virtual photons created by the second species would erase those created by the first species). 

Should we guard against this? Or alternatively should we have a separate function to remove virtual photons?

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -0,0 +1,244 @@
...
+                        // Energy of the virtual photon
+                        amrex::ParticleReal vphoton_energy = frac_energy * gamma_primary * PhysConst::c;
+
+                        // Photon index for the current primary
+                        amrex::Long ip = offset_vp_data[i] + j;
+                        pa_vp[PIdx::ux][ip] = vphoton_energy * nx; // will be multiplied by m_e before dumping the outputs
+                        pa_vp[PIdx::uy][ip] = vphoton_energy * ny; // will be multiplied by m_e before dumping the outputs
+                        pa_vp[PIdx::uz][ip] = vphoton_energy * nz; // will be multiplied by m_e before dumping the outputs
+
+                        // TODO: add beam size effect - displace the photon position
```
RemiLehe, line 240: @atmyers Would it make sense to use a `SmartCopy` object instead here?

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -0,0 +1,244 @@
...
+#elif defined (WARPX_DIM_RCYLINDER)
+                        pa_vp[PIdx::x][ip] = x;
+                        pa_vp[PIdx::theta][ip] = theta;
+#elif defined(WARPX_DIM_RSPHERE)
+                        pa_vp[PIdx::x][ip] = x;
+                        pa_vp[PIdx::theta][ip] = theta;
+                        pa_vp[PIdx::phi][ip] = phi;
+#endif
+                        pa_vp[PIdx::w][ip] = w / sampling_factor;
+                        pid_vp[ip] = pid; // Same pid as the parent lepton - will be useful if we want to update the parent electron
```
RemiLehe:
```suggestion
```
I think I would actually prefer removing this line of code for now (at least until we implement the update of the parent particle) -- but we can add a TODO comment. This is because this makes the ID non-unique among virtual photons, which is in general not expected in WarpX/AMReX.

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -0,0 +1,244 @@
...
+                    // Minimum fractional (wrt primary particle) photon energy
+                    amrex::Real y_min = vphoton_min_energy / (mass * gamma_primary * PhysConst::c * PhysConst::c);
+                    amrex::Real umax = std::log(y_min) * std::log(y_min);
+
+                    for (int j = 0; j < num_vp_data[i]; j++)
+                    {
+                        // Draw a random number between umin and umax
+                        amrex::ParticleReal rnd = (umax - umin) * amrex::Random(engine) + umin ;
+                        // Fractional energy of the photon, often denoted as y (or x)
+                        amrex::ParticleReal frac_energy = std::exp( - std::sqrt(rnd) );
```
RemiLehe, line 230: Could you add a comment that this generates `frac_energy` with a probability distribution proportional to `log(frac_energy)/frac_energy` (inverse CDF method)?

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -0,0 +1,244 @@
...
+                    amrex::ParticleReal x  = soa.m_rdata[PIdx::x][i];
+                    amrex::ParticleReal theta  = soa.m_rdata[PIdx::theta][i];
+                    amrex::ParticleReal phi  = soa.m_rdata[PIdx::phi][i];
+#endif
+                    amrex::ParticleReal w  = soa.m_rdata[PIdx::w][i];
+                    uint64_t pid = soa.m_idcpu[i]; // TODO: use this to update the parent lepton if needed
+
+                    // Minimum fractional (wrt primary particle) photon energy
+                    amrex::Real y_min = vphoton_min_energy / (mass * gamma_primary * PhysConst::c * PhysConst::c);
+                    amrex::Real umax = std::log(y_min) * std::log(y_min);
```
RemiLehe, line 218: Maybe the code that defines `umin` can be here, instead of at the beginning of the function.

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
RemiLehe: Discussed offline, we can give the reference for the non-integrated formula, and then say to integrate it to get this result.

Source/Particles/PhysicalParticleContainer.H
aeriforme: @RemiLehe Why?

EDIT: discussed offline, I misread

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
aeriforme: I added this to the docs. 
I don't think we need a separate function to remove the virtual photons just yet.

Examples/Tests/virtual_photons/inputs_test_3d_virtual_photons
aeriforme: As written in the docs and in the comments, the virtual photons are generated from scratch at each timestep.

RemiLehe (approved):

PR #5597, opened 2025-01-23
Clang UB sanitizer CI test: increase coverage

lucafedeli88:
In CI test based on clang UB sanitizers, most of the time (~ 1h30) is spent in compiling the code, while just a few minutes are spent actually running some simulations. This means that we can increase the coverage of the test by adding some more simulations to the tests with a negligible increase of the total runtime.
This PR does just that: now most of the cases in `Examples/Physics_applications`  are tested with the UB sanitizer.

~~Note that some cases cannot run in double precision (see below). For this reason, the PR also splits the UB sanitizer test into single precision and double precision (in double precision only the cases that cannot run in single precision are tested).~~

### 1)  Issue found while running  inputs_test_3d_beam_beam_collision --> We need to run this case in double precision

The tool has found an issue while running `mpirun -n 2 ./build/bin/warpx.3d Examples/Physics_applications/beam_beam_collision/inputs_test_3d_beam_beam_collision` :
```
STEP 1 starts ...
/home/runner/work/WarpX/WarpX/build/_deps/fetchedpicsar-src/multi_physics/QED/include/picsar_qed/containers/picsar_tables.hpp:310:17: runtime error: -nan is outside the range of representable values of type 'int'
/home/runner/work/WarpX/WarpX/build/_deps/fetchedpicsar-src/multi_physics/QED/include/picsar_qed/containers/picsar_tables.hpp:310:17: runtime error: -nan is outside the range of representable values of type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/runner/work/WarpX/WarpX/build/_deps/fetchedpicsar-src/multi_physics/QED/include/picsar_qed/containers/picsar_tables.hpp:310:17 in 
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/runner/work/WarpX/WarpX/build/_deps/fetchedpicsar-src/multi_physics/QED/include/picsar_qed/containers/picsar_tables.hpp:310:17 in 
```

I've temporarily commented out this case while I investigate the cause. ~~For the moment, I am **not** able to reproduce the issue on my local machine~~.  The issue is using single precision for this specific test case! Specifically, momenta end up being `NaN` and the sanitizer detects the attempt to convert a floating point NaN into an integer. 
I have added an `ABLASTR_ALWAYS_ASSERT_WITH_MESSAGE` in `PoissonSolver.H` so that a more readable error message is provided to the user.

### 2) Issue found while running inputs_test_2d_background_mcc --> We need to run  this case in double precision

`MLMG` does not converge in single precision for this simulation case. We need to run it in double precision. 

### 3) Issue found while running free_electron_laser --> We need to run  this case in double precision

I have observed this issue:
```
 STEP 444 starts ...
0::1::Assertion `m_current_z_lab[i_buffer] >= m_buffer_domain_lab[i_buffer].lo(m_moving_window_dir) and m_current_z_lab[i_buffer] <= m_buffer_domain_lab[i_buffer].hi(m_moving_window_dir)' failed, file "/home/runner/work/WarpX/WarpX/Source/Diagnostics/BTDiagnostics.cpp", line 870, Msg: 
 ### ERROR   : z-slice in lab-frame (0.299976) is outside the buffer domain
#            physical extent (0.299976 to 0.299988).
 !!!
```
which seems to be related to using single precision instead of double precision. Therefore, we need to run this case in double precision. 

### 4) Issue found while running inputs_test_2d_laser_ion_acc --> bugfix in WarpX

`inputs_test_2d_laser_ion_acc` case has the following issue in single precision:

```
--- INFO    : Writing openPMD file diags/openPMDbw000000
  what():  Datatypes of chunk data (FLOAT) and record component (DOUBLE) do not match.
SIGABRT
```

This comes from the fact that the datatype of this dataset in `ParticleHistogram2D.cpp` is hard-coded as double:
```
    auto dataset = io::Dataset(
            io::determineDatatype<double>(),
            {static_cast<unsigned long>(m_bin_num_ord), static_cast<unsigned long>(m_bin_num_abs)});
```
this PR modifies these lines as follows:
```
    auto dataset = io::Dataset(
            io::determineDatatype<amrex::Real>(),
            {static_cast<unsigned long>(m_bin_num_ord), static_cast<unsigned long>(m_bin_num_abs)});
```

### 5) Perform all the simulations in double precision
See https://github.com/BLAST-WarpX/warpx/pull/5936

### 7) In `IntervalsParser` prevent the use of an uninitialized object
This was discovered by adding a new test case.
`temp_slice.getStart() > m_slices[i_slice].getStart() && i_slice < static_cast<int>(m_slices.size())` must be reversed: `i_slice < static_cast<int>(m_slices.size()) && temp_slice.getStart() > m_slices[i_slice].getStart() ` otherwise we call a method of a non-existent object.


Source/Diagnostics/ReducedDiags/ParticleHistogram2D.cpp
```diff
@@ -298,7 +298,7 @@ void ParticleHistogram2D::WriteToFile (int step) const
     data.setPosition<amrex::Real>({0.5, 0.5});
 
     auto dataset = io::Dataset(
-            io::determineDatatype<double>(),
+            io::determineDatatype<amrex::Real>(),
```
ax3l: Wouldn't the properties be in `ParticleReal`?
```suggestion
            io::determineDatatype<amrex::ParticleReal>(),
```

lucafedeli88: Yes, you are right!

lucafedeli88: However, we also have to modify the datatype of `m_h_data_2D` to be coherent (I've done that in my latest commit):  `amrex::TableData<amrex::Real,2> m_h_data_2D` ->   `amrex::TableData<amrex::ParticleReal,2> m_h_data_2D`

lucafedeli88:
ping, @EZoni !

dpgrote (approved):
This looks good, thanks!

I will comment that NaNs always make me nervous that there is something bad lurking in the code. Do you know why they appear in that case with single precision?

lucafedeli88:
> This looks good, thanks!
> 
> I will comment that NaNs always make me nervous that there is something bad lurking in the code. Do you know why they appear in that case with single precision?

Thanks, @dpgrote .    The NaN comes from this line in `PoissonSolver.H`.  `beta_solver[2]` is `1.00059783`  here due to numerical errors, so the  square root is negative. We could indeed try to fix that.
```
geom[lev].CellSize(2)/std::sqrt(1._rt-beta_solver[2]*beta_solver[2]))};
```

lucafedeli88:
> > This looks good, thanks!
> > I will comment that NaNs always make me nervous that there is something bad lurking in the code. Do you know why they appear in that case with single precision?
> 
> Thanks, @dpgrote . The NaN comes from this line in `PoissonSolver.H`. `beta_solver[2]` is `1.00059783` here due to numerical errors, so the square root is negative. We could indeed try to fix that.
> 
> ```
> geom[lev].CellSize(2)/std::sqrt(1._rt-beta_solver[2]*beta_solver[2]))};
> ```

Actually, in this case, we have a 250 GeV electron beam. So we can't represent beta reliably in single precision for such high energies. I think that the best we can do is to add an assert to ensure that beta < 1 .

lucafedeli88:
@dpgrote , I've added this check inside `PoissonSolver.H` :

```
    ABLASTR_ALWAYS_ASSERT_WITH_MESSAGE (
        std::all_of(beta_solver.begin(), beta_solver.end(),
            [](const auto b){return (b<1.0_rt);}),
        "Components of beta_solver must be < 1.");
```
so that at least the user has a more readable error message.

.github/workflows/clang_sanitizers.yml
EZoni, line 1: I'm kind of wondering, instead of running specific input files manually, couldn't we simply run specific tests with `ctest`? 

I would expect this to be equivalent and it is part of what the CTest migration was meant for (I mean, run tests through `ctest` rather than manually).

For example, replacing the `mpirun` commands with
```
ctest --test-dir build --output-on-failure -R "laser_acceleration"
ctest --test-dir build --output-on-failure -R "laser_ion"
...
```

We could also make the `-R` filter more complex to filter specific tests and/or only execute the `.run` tests, if we want to skip analysis, checksums, etc.

What do you think? Would you be okay with trying this?

lucafedeli88: That sounds like a good idea, @EZoni ! Would it be possible also to override some of the inputfile parameters, in order to achieve something similar to what I do here:

```
        cd Examples/Physics_applications/uniform_plasma
        mpirun -n 2 ../../../build/bin/warpx.3d inputs_test_3d_uniform_plasma warpx.cfl=0.98 algo.particle_shape=2
        rm -rf ./diags
        mpirun -n 2 ../../../build/bin/warpx.3d inputs_test_3d_uniform_plasma warpx.cfl=0.98 algo.particle_shape=3
        rm -rf ./diags
        mpirun -n 2 ../../../build/bin/warpx.3d inputs_test_3d_uniform_plasma warpx.cfl=0.98 algo.particle_shape=4
        rm -rf ./diags
        mpirun -n 2 ../../../build/bin/warpx.3d inputs_test_3d_uniform_plasma warpx.cfl=0.98 algo.particle_shape=4 algo.current_deposition=direct
        rm -rf ./diags
        mpirun -n 2 ../../../build/bin/warpx.3d inputs_test_3d_uniform_plasma warpx.cfl=0.98 algo.particle_shape=4 algo.field_gathering=momentum-conserving
        rm -rf ./diags
        mpirun -n 2 ../../../build/bin/warpx.3d inputs_test_3d_uniform_plasma warpx.cfl=0.98 algo.particle_shape=4 algo.particle_pusher=vay
        rm -rf ./diags
        mpirun -n 2 ../../../build/bin/warpx.3d inputs_test_3d_uniform_plasma warpx.cfl=0.98 algo.particle_shape=4 algo.particle_pusher=higuera
        rm -rf ./diags
        cd -
```

lucafedeli88: ( I want to use the same inputfile to test different code paths)

EZoni: I don't think that this is possible at the moment. 

We could modify our setup to support this, but on the other hand we kind of intentionally avoided that, to make sure that all input information is in one place (the input file(s)) rather than in two places (the input file(s) and the command line).

The command line customization used to be available with the AMReX regression testing framework, but we were  intentionally avoiding that with ctest.

The alternative we currently provide to reduce duplication is the so-called "base" input files that can be included in the test-specific input files, e.g.,
https://github.com/BLAST-WarpX/warpx/blob/35ebb3c64d231e5068c141e195ae45c21484167c/Examples/Physics_applications/plasma_acceleration/inputs_test_3d_plasma_acceleration_boosted#L1-L5

I wonder if we can use the test variants that already exist, rather than new custom variants added through the command line?

EZoni (dismissed):
Looks good to me.

.github/workflows/clang_sanitizers.yml
```diff
@@ -71,11 +71,44 @@ jobs:
...
+
+        ctest --test-dir build --output-on-failure -R "3d_thomson_parabola_spectrometer.run"
+
+        ctest --test-dir build --output-on-failure -R "2d_uniform_plasma.run"
+        ctest --test-dir build --output-on-failure -R "3d_uniform_plasma.run"
+        ctest --test-dir build --output-on-failure -R "3d_uniform_plasma_restart.run"
+
+        ctest --test-dir build --output-on-failure -R "2d_theta_implicit_jfnk_vandb.run"
+        ctest --test-dir build --output-on-failure -R "2d_theta_implicit_jfnk_vandb_filtered.run"
+
```
EZoni: Note that 
```
$ ctest --test-dir build/Examples/Physics_applications/ -N -R ".*\.run"
Test project /home/edoardo/src/warpx/build/Examples/Physics_applications
  Test  #1: test_3d_beam_beam_collision.run
  Test  #3: test_1d_background_mcc_picmi.run
  Test  #6: test_1d_dsmc_picmi.run
  Test  #9: test_2d_background_mcc.run
  Test #11: test_2d_background_mcc_picmi.run
  Test #13: test_1d_fel.run
  Test #16: test_1d_laser_acceleration.run
  Test #18: test_1d_laser_acceleration_fluid_boosted.run
  Test #21: test_1d_laser_acceleration_picmi.run
  Test #23: test_2d_laser_acceleration_boosted.run
  Test #25: test_2d_laser_acceleration_mr.run
  Test #27: test_2d_laser_acceleration_mr_picmi.run
  Test #29: test_2d_refined_injection.run
  Test #32: test_3d_laser_acceleration.run
  Test #34: test_3d_laser_acceleration_picmi.run
  Test #36: test_3d_laser_acceleration_single_precision_comms.run
  Test #38: test_rz_laser_acceleration.run
  Test #40: test_rz_laser_acceleration_opmd.run
  Test #43: test_rz_laser_acceleration_picmi.run
  Test #45: test_2d_laser_ion_acc.run
  Test #48: test_2d_laser_ion_acc_picmi.run
  Test #51: test_1d_plasma_acceleration_picmi.run
  Test #53: test_2d_plasma_acceleration_boosted.run
  Test #55: test_2d_plasma_acceleration_mr.run
  Test #57: test_2d_plasma_acceleration_mr_momentum_conserving.run
  Test #59: test_3d_plasma_acceleration_boosted.run
  Test #61: test_3d_plasma_acceleration_boosted_hybrid.run
  Test #63: test_3d_plasma_acceleration_mr_picmi.run
  Test #65: test_3d_plasma_acceleration_picmi.run
  Test #67: test_2d_plasma_mirror.run
  Test #69: test_rz_spacecraft_charging_picmi.run
  Test #72: test_2d_uniform_plasma.run
  Test #74: test_3d_uniform_plasma.run
  Test #76: test_3d_uniform_plasma_restart.run
  Test #79: test_3d_thomson_parabola_spectrometer.run

Total Tests: 35
```

Is there a specific reason why you chose to subselect 25 (if I didn't make mistakes counting) out of these 35?

Otherwise we could run just the command above:
```suggestion
        ctest --test-dir build/Examples/Physics_applications/ -R ".*\.run" --output-on-failure
```

EZoni (commented):
Just one last comment, in case it helps make this more compact and robust.

.github/workflows/clang_sanitizers.yml
EZoni: We could add a `-E "picmi"` if there is any problem with running the PICMI tests in this context.

lucafedeli88: Oh, right ! Let's try doing that then

EZoni (approved):

PR #5600, opened 2025-01-24
Clang-tidy CI test: bump version from 16 to 17 

lucafedeli88:
This PR bumps the version used for `clang-tidy` CI tests from 16 to 17. It also addresses all the issues found with the upgraded tool.

To be merged **after** https://github.com/ECP-WarpX/WarpX/pull/5592 ✅

### The issues found 🧐  and fixed 🛠️ with the upgraded tool are the following  :

- [bugprone-switch-missing-default-case](https://releases.llvm.org/17.0.1/tools/clang/tools/extra/docs/clang-tidy/checks/bugprone/switch-missing-default-case.html)
A newly introduced check to flag `switch` statements without a `default` case (unless the argument is an `enum`)

- [cppcoreguidelines-rvalue-reference-param-not-moved](https://releases.llvm.org/17.0.1/tools/clang/tools/extra/docs/clang-tidy/checks/cppcoreguidelines/rvalue-reference-param-not-moved.html)
A newly introduced check to flag when an rvalue reference argument of a function is never moved inside the function body.
⚠️ **Warning**: in order to have this check compatible with  [performance-move-const-arg](https://releases.llvm.org/17.0.1/tools/clang/tools/extra/docs/clang-tidy/checks/performance/move-const-arg.html)  I had to set `performance-move-const-arg.CheckTriviallyCopyableMove` to `false` (specifically for the three methods in `ablastr::utils::msg_logger` accepting `std::vector<char>::const_iterator&& rit` arguments).

- [misc-header-include-cycle](https://releases.llvm.org/17.0.1/tools/clang/tools/extra/docs/clang-tidy/checks/misc/header-include-cycle.html)
A newly introduced check to prevent cyclic header inclusions.

- [modernize-type-traits](https://releases.llvm.org/17.0.1/tools/clang/tools/extra/docs/clang-tidy/checks/modernize/type-traits.html)
A newly introduced check. The idea is to replace currencies of, e.g., `std::is_integral<T>::value`, with the less verbose alternative  `std::is_integral_v<T>`

- [performance-avoid-endl](https://releases.llvm.org/17.0.1/tools/clang/tools/extra/docs/clang-tidy/checks/performance/avoid-endl.html) 
A newly introduced check. The idea is to replace `<< std::endl` with `\n`, since `endl` also forces a flush of the stream. In few cases flushing the buffer is actually the desired behavior. Typically, this happens when we want to write to `std::cerr`, which is however automatically flushed after each write operation.  In cases where actually flushing to `std::cout` is the desired behavior one can do `<< \n << std::flush `, which is arguably more transparent than `<< std::endl`. 

- [performance-noexcept-swap](https://releases.llvm.org/17.0.1/tools/clang/tools/extra/docs/clang-tidy/checks/performance/noexcept-swap.html) 
For performance reasons it is better if `swap` functions are declared as `noexcept`, in order to allow the compiler to perform more aggressive optimizations. In any case, we can use the AMReX function `amrex::Swap`, which is `noexcept`. 

### 🔄 Re-enabled checks:

- [readability-misleading-indentation](https://releases.llvm.org/17.0.1/tools/clang/tools/extra/docs/clang-tidy/checks/readability/misleading-indentation.html)
This check was already available in v16, but a bug led to false positives. The bug has been corrected in v17 of the tool, so we can re-enable the check.

### ⛔ The PR excludes the following checks :

- [cppcoreguidelines-missing-std-forward](https://releases.llvm.org/17.0.1/tools/clang/tools/extra/docs/clang-tidy/checks/cppcoreguidelines/missing-std-forward.html)
A newly introduced check that warns when a forwarding reference parameter is not forwarded. In order to comply with this check I think that I have to pass some parameters by reference to lambda functions inside `ParallelFor` constructs. However, this leads to issues when we compile for GPUs. Therefore, I think that the best solution is to exclude this check. See an example below (for `PredFunc&& filter` ):
```
amrex::ParallelForRNG(np,
    [=,&filter] AMREX_GPU_DEVICE (int i, amrex::RandomEngine const& engine)
    {
        p_mask[i] = filter(src_data, i, engine);
    });
``` 

- [misc-include-cleaner](https://releases.llvm.org/17.0.1/tools/clang/tools/extra/docs/clang-tidy/checks/misc/include-cleaner.html)
It would be awesome to include this check. However, as it is now implemented, it has no notion of "associated headers". For instance, let's suppose that the header `MyClass.H` has `#include<string>` and that `MyClass.cpp` has `#include "MyClass.H"` and uses `std:string` somewhere. In this case, the check raises a warning stating that you should include `<string>` in  `MyClass.cpp` even if it is transitively included via the associate header `MyClass.H` . For this reason, for the moment, it is better to periodically check headers with the `IWYU` tool.


ax3l:
Ready for rebase, v16 PR is in :)

EZoni (approved):

ax3l (approved):
Thank you for the migration, very clean PR description and fixes!

PR #5604, opened 2025-01-24
WarpX class: make verboncoeur_axis_correction a private member variable

lucafedeli88:
This PR changes `verboncoeur_axis_correction` from a static WarpX class variable to a private member variable (renamed `m_verboncoeur_axis_correction`).
This is a small step towards reducing the use of static variables in the WarpX class.


ax3l:
RZ needs a little update still :)

EZoni (approved):

PR #5605, opened 2025-01-24
WarpX class: make zmax_plasma_to_compute_max_step a private member variable and remove do_compute_max_step_from_zmax

lucafedeli88:
This PR combines the following static variables of the WarpX class:
```
static amrex::Real zmax_plasma_to_compute_max_step;
static bool do_compute_max_step_from_zmax;
```
into a single private member variable:
```
std::optional<amrex::Real> m_zmax_plasma_to_compute_max_step = std::nullopt;
```

This is a small step towards reducing reliance on static class variables.


ax3l (approved):
Nice!

PR #5609, opened 2025-01-27
WarpX class: move shiftMF to anonymous namespace in WarpXMovingWindow.cpp

lucafedeli88:
This PR moves the static function `shiftMF` from the WarpX class to an anonymous namespace in `WarpXMovingWindow.cpp`, where it is actually used.
This is done to simplify the Warpx class.


Source/Utils/WarpXMovingWindow.cpp
```diff
@@ -57,6 +57,184 @@
...
+        amrex::MultiFab& mf, const amrex::Geometry& geom,
+        int num_shift, int dir,
+        bool safe_guard_cells, bool do_single_precision_comms,
+        amrex::LayoutData<amrex::Real>* cost,
+        amrex::Real external_field=0.0, bool useparser = false,
+        amrex::ParserExecutor<3> const& field_parser={},
+        const bool PMLRZ_flag = false)
+    {
+        using namespace amrex::literals;
+        WARPX_PROFILE("::shiftMF()");
```
ax3l: Does this still have a sensible prefix in the BL Profiler?

Source/Utils/WarpXMovingWindow.cpp
```diff
@@ -57,6 +57,184 @@
 
 using namespace amrex;
 
+namespace
+{
+    void shiftMF (
```
ax3l, line 77: Adding a Doxygen string on the way would be really nice :)

Source/Utils/WarpXMovingWindow.cpp
lucafedeli88: It depends on what you mean for sensible ;-) ! It appears like that at the moment:

```
---------------------------------------------------------------------------------------------------
Name                                                NCalls  Excl. Min  Excl. Avg  Excl. Max   Max %
---------------------------------------------------------------------------------------------------
::shiftMF()                                            591     0.8066     0.8066     0.8066  18.74%
WarpX::OneStep_nosub()                                 100     0.6293     0.6293     0.6293  14.62%
PhysicalParticleContainer::Evolve()                    100     0.4915     0.4915     0.4915  11.42%
ParticleContainer::RedistributeCPU()                   304     0.4015     0.4015     0.4015   9.33%
FillBoundary_nowait()                                 1819     0.2607     0.2607     0.2607   6.06%
Filter::ApplyStencil(MultiFab)                         303     0.2414     0.2414     0.2414   5.61%
FlushFormatOpenPMD::WriteToFile()                        2     0.2344     0.2344     0.2344   5.45%

```
Do you have an alternative in mind? Just `shiftMF`  maybe ?

Source/Utils/WarpXMovingWindow.cpp
lucafedeli88: I agree!

lucafedeli88: Done!

Source/Utils/WarpXMovingWindow.cpp
```diff
@@ -57,6 +57,199 @@
...
+        const bool PMLRZ_flag = false)
+    {
+        using namespace amrex::literals;
+        WARPX_PROFILE("::shiftMF()");
+        const amrex::BoxArray& ba = mf.boxArray();
+        const amrex::DistributionMapping& dm = mf.DistributionMap();
+        const int nc = mf.nComp();
+        const amrex::IntVect& ng = mf.nGrowVect();
+
+        AMREX_ALWAYS_ASSERT(ng[dir] >= std::abs(num_shift));
```
lucafedeli88, line 93: Previously, this was `AMREX_ALWAYS_ASSERT(ng[dir] >= num_shift);`
however, it seems to me that, since `num_shift` can be negative, we need an absolute value here. What do you think @ax3l ?

Source/Utils/WarpXMovingWindow.cpp
```diff
@@ -57,6 +57,199 @@
...
+        amrex::MultiFab tmpmf(ba, dm, nc, ng);
+        amrex::MultiFab::Copy(tmpmf, mf, 0, 0, nc, ng);
+
+        if ( safe_guard_cells ) {
+            // Fill guard cells.
+            ablastr::utils::communication::FillBoundary(tmpmf, do_single_precision_comms, geom.periodicity());
+        } else {
+            amrex::IntVect ng_mw = amrex::IntVect::TheUnitVector();
+            // Enough guard cells in the MW direction
+            ng_mw[dir] = std::abs(num_shift);
```
lucafedeli88, line 104: Same as above. This was `ng_mw[dir] = num_shift;`, but I think that the absolute value is necessary here.

Source/Utils/WarpXMovingWindow.cpp
ax3l: maybe `WarpX::`/`warpx::` or just `shiftMF` then? :)

In doubt, prefix with lowercase `warpx::` because technically we should wrap every functionality is this basic namespace in WarpX :)

Source/Utils/WarpXMovingWindow.cpp
ax3l: Wait, we _can_ shift backwards (or just know it is technically possible because `num_shift` is an `int`)? :D 

Either way, LGTM ;)

Source/Utils/WarpXMovingWindow.cpp
ax3l:
```suggestion
        WARPX_PROFILE("warpx::shiftMF()");
```

ax3l (approved):

PR #5611, opened 2025-01-28
CDash: fix variables in CTestConfig.cmake

EZoni:
Try using CTest module variables, instead of CTest script variables, given that we are not using a CTest script, to fix fields displayed in the CDash dashboard.

Examples of guidance from the official documentation (https://cmake.org/cmake/help/v3.24/manual/ctest.1.html):

Screenshot from 2025-01-28 11-04-12

Screenshot from 2025-01-28 11-08-45

However, this seems to be true for all other variables in CTestConfig.cmake. Should we try to update them all?

### DIFF
--- a/CTestConfig.cmake
+++ b/CTestConfig.cmake
@@ -13,6 +13,8 @@ set(CTEST_SUBMIT_URL https://my.cdash.org/submit.php?project=WarpX)
 
 set(CTEST_DROP_SITE_CDASH TRUE)
 
-# Additional settings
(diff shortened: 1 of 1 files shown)


ax3l:
Oh, interesting.

Or, we could use a site-specific [ctest script](https://cmake.org/cmake/help/v3.24/manual/ctest.1.html#ctest-script), that way we can have a script for Azure, one for Perlmutter, etc...

ax3l (approved):

PR #5617, opened 2025-01-29
Remove unused SliceDiagnostic.H/cpp

lucafedeli88:
The functions defined in `SliceDiagnostic.H` and implemented in ` SliceDiagnostic.cpp` are never used in WarpX. Therefore, this PR removes these two source files (and updates `CMakeLists.txt` and `Make.package` accordingly)


ax3l (approved):

ax3l:
LGTM, slicing is a supported feature in the full diagnostics.

PR #5619, opened 2025-01-30
Move Fornberg coefficients calculations from WarpX to ablastr

lucafedeli88:
The calculation of Fornberg stencil coefficients is rather general, and it can be shared with other projects of the BLAST family. Therefore, this PR moves the responsible functions into `ablastr`.

Specifically, the PR does the following:

- 2 new files (`FiniteDifference.H` and `FiniteDifference.cpp`) are created under `ablastr/math` (`CMakeLists.txt` and `Make.package` accordingly)
- the static method of the WarpX class `getFornbergStencilCoefficients` and the `ReorderFornbergCoefficients` function (originally defined in an anonymous namespace in `WarpX.cpp`) are moved to these new files, inside the namespace `ablastr::math`
- the two methods are minimally adapted (e.g., `AMREX_ALWAYS_ASSERT_WITH_MESSAGE` becomes `ABLASTR_ALWAYS_ASSERT_WITH_MESSAGE`)
- `WarpX.cpp` and `SpectralKSpace.cpp` (where the aforementioned functions were called)  are updated

Note that with this PR `SpectralKSpace.cpp` does not need anymore to include the heavy `WarpX.H` header.

### DIFF
--- a/Source/ablastr/math/Make.package
+++ b/Source/ablastr/math/Make.package
@@ -1,3 +1,5 @@
-include $(WARPX_HOME)/Source/ablastr/math/fft/Make.package
+CEXE_sources += FiniteDifference.cpp
 
-VPATH_LOCATIONS   += $(WARPX_HOME)/Source/ablastr
+VPATH_LOCATIONS   += $(WARPX_HOME)/Source/ablastr/math
+
+include $(WARPX_HOME)/Source/ablastr/math/fft/Make.package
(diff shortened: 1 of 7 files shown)


Source/ablastr/math/Make.package
EZoni, line 1: Kind of a tangential question, but do we currently have any CI check that compiles with GNU Make instead of CMake? I wonder if code changes like the ones in this file are actually tested automatically.

EZoni: I can answer myself. Yes, I see at least one CI workflow here that uses GNU Make: https://github.com/ECP-WarpX/WarpX/blob/e0421a1cebbd69be6593145cd1e713000ec5ae46/.github/workflows/cuda.yml#L100-L134.

EZoni (approved):

PR #5621, opened 2025-01-30
Add MCC forward scattering

budjensen:
Added a version of forward scattering suggested in [J F J Janssen et al (2016)](https://doi.org/10.1088/0963-0252/25/5/055026). This process decreases total particle energy by the process' energy threshold. If no energy threshold is given in the input file, this process is equivalent to no collision being carried out (no scattering and no energy change).

Adjusted documentation appropriately and fixed a pre-existing typo.

Feature was tested on my own machine by confirming with python callbacks that the pre-collision and post-collision velocities were equal in the case of no energy cost threshold being supplied, and that the velocities were scaled down by the appropriate amount when a threshold was supplied (it was also checked that particle direction was the same before and after collision). No formal test was added since no current MCC tests exist and adding a framework to access cross sections while executing tests was prohibitive.

### DIFF
--- a/Docs/source/theory/multiphysics/collisions.rst
+++ b/Docs/source/theory/multiphysics/collisions.rst
@@ -128,6 +128,14 @@ Excitation
 
 The process is also the same as for elastic scattering except the excitation energy cost is subtracted from the particle energy. This is done by reducing the velocity before a scattering angle is chosen.
 
+Forward scattering
+^^^^^^^^^^^^^^^^^^
+
+This process operates in two ways:
+
+1. If an excitation energy cost is provided, the energy cost is subtracted from the particle energy and no scattering is performed.
+2. If an excitation energy cost is not provided, the particle is not scattered and the velocity is unchanged (corresponding to a scattering angle of :math:`0` in the elastic scattering process above).
+
 Benchmarks
 ----------
 
--- a/Docs/source/theory/multiphysics/collisions.rst
+++ b/Docs/source/theory/multiphysics/collisions.rst
@@ -136,6 +136,8 @@ This process operates in two ways:
 1. If an excitation energy cost is provided, the energy cost is subtracted from the particle energy and no scattering is performed.
 2. If an excitation energy cost is not provided, the particle is not scattered and the velocity is unchanged (corresponding to a scattering angle of :math:`0` in the elastic scattering process above).
 
+See :cite:t:`b-Janssen2016` for a recommended use of this process.
+
 Benchmarks
 ----------
 
--- a/Docs/source/theory/multiphysics/collisions.rst
+++ b/Docs/source/theory/multiphysics/collisions.rst
@@ -121,7 +121,7 @@ The particle velocity in the COM frame is then isotropically scattered using the
 Back scattering
 ^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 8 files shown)


Docs/source/usage/parameters.rst
```diff
@@ -2168,7 +2168,7 @@ Details about the collision models can be found in the :ref:`theory section <mul
 * ``<collision_name>.scattering_processes`` (`strings` separated by spaces)
     Only for ``dsmc`` and ``background_mcc``. The scattering processes that should be
     included. Available options are ``elastic``, ``back`` & ``charge_exchange``
-    for ions and ``elastic``, ``excitationX`` & ``ionization`` for electrons.
+    for ions and ``elastic``, ``excitationX``, ``ionization`` & ``forward`` for electrons.
```
roelof-groenewald, line 2171: The forward scattering will also be available for ions (and I'm realizing excitation would also in principle work for ions). Would you mind updating this bit of the docs to reflect that?

Docs/source/theory/multiphysics/collisions.rst
```diff
@@ -121,13 +121,21 @@ The particle velocity in the COM frame is then isotropically scattered using the
...
 
-The process is the same as for elastic scattering above expect the scattering angle is fixed at :math:`\pi`, meaning the particle velocity in the COM frame is updated to :math:`-\vec{u}_c`.
+The process is the same as for elastic scattering above except the scattering angle is fixed at :math:`\pi`, meaning the particle velocity in the COM frame is updated to :math:`-\vec{u}_c`.
 
 Excitation
 ^^^^^^^^^^
 
 The process is also the same as for elastic scattering except the excitation energy cost is subtracted from the particle energy. This is done by reducing the velocity before a scattering angle is chosen.
 
+Forward scattering
```
roelof-groenewald, line 131: Could you please add a citation to the paper you mentioned (https://doi.org/10.1088/0963-0252/25/5/055026) in this section? I know we don't have citations for the other processes, but seeing as there is one available for this, we might as well add it.

We usually add references in https://github.com/ECP-WarpX/WarpX/blob/development/Docs/source/refs.bib and then use sphinx commands to cite the appropriate reference.

roelof-groenewald (changes requested):
Thank you very much for the contribution!

The PR looks good. I just have two requests for small changes.

Btw, we do have tests for MCC (https://github.com/ECP-WarpX/WarpX/tree/development/Examples/Physics_applications/capacitive_discharge). The cross-sections for such tests are kept in a separate repo: https://github.com/ECP-WarpX/warpx-data.

Docs/source/usage/parameters.rst
budjensen: Leaving this here for a possible future change. It looks like back scattering and charge exchange can also be chosen for electrons (I did not add these to the docs in my last commit). It may be worth issuing a warning if charge exchange is chosen for electrons. I can't think of a physical process where an electron would lose so much momentum to the background gas.

roelof-groenewald: Good point! I guess some protection we have is that the user would also have to specify a cross-section for that process which would hopefully make them realize they are trying something unrealistic. But it is true that we don't have protection here against an user accidentally specifying `back` when they meant to do `elastic`. Thanks for noting that.

roelof-groenewald (approved):
Looks good! Thanks again for the contribution.

PR #5622, opened 2025-01-31
Fix bug with DSMC collisions in RZ

roelof-groenewald:
I realized there is a bug in the DSMC module for RZ geometry where the velocity vectors for the colliding pair was not rotated so that a proper center-of-momentum calculation could be done. This PR fixes the bug.

To check that the fix in this PR works, I compared the azimuthal velocity distribution for energetic ions created from NBI with finite impact parameter (such that a net rotation should be imparted on the ions), when simulated in 3d versus RZ:

3d result:
image

RZ result:
image


JustinRayAngus (approved):
Good catch!. The changes look good to me.

PR #5229, opened 2024-09-09
[WIP] Rename multi-J PSATD to PSATD-JRhom 

oshapoval:
This PR just renames the `multi-J PSATD` algorithm to `PSATD-JRhom` as it used in the paper [Phys. Rev. E 110, 025206 (2024)](https://journals.aps.org/pre/abstract/10.1103/PhysRevE.110.025206). Another open [PR](https://github.com/ECP-WarpX/WarpX/pull/4263) does actual implementation of `PSATD-JRhom` with an arbitrary temporal dependency of `J` and `rho`.

PR #5623, opened 2025-01-31
WarpX class: do_multi_J and do_multi_J_n_depositions no longer static

lucafedeli88:
This PR turns the static variables of the WarpX class `do_multi_J` and `do_multi_J_n_depositions` into private member variables (also adding the prefix `m_`).
This PR contributes to make the WarpX class less static.


EZoni:
Yes, we are hopefully going to be renaming and simplifying these in #5229. I'll try to see which one goes first.

lucafedeli88:
> Yes, we are hopefully going to be renaming and simplifying these in #5229. I'll try to see which one goes first.

Ok! We can wait and see if this PR will still be relevant after the refactoring.

lucafedeli88:
This PR may not be needed after https://github.com/ECP-WarpX/WarpX/pull/5229

PR #5624, opened 2025-01-31
Embedded Boundary: take ScaleEdges and ScaleAreas out of WarpX class

lucafedeli88:
`ScaleAreas` and `ScaleEdges` are pure functions that can be easily taken out of the WarpX class, in order to make it simpler.

This PR places these two functions under a newly created namespace `warpx::embedded_boundary`, inside the files `EmbeddedBoundray/EmbeddedBoundary.H/cpp` .


lucafedeli88:
I will re-open a more comprehensive PR. I've spotted more opportunities to take out stuff from the WarpX class.

PR #5625, opened 2025-01-31
Embedded Boundary: take some EB-related methods out of WarpX class

lucafedeli88:
`ComputeEdgeLengths`,  `ComputeFaceAreas`, `ScaleAreas`, and `ScaleEdges` are pure functions that can be easily taken out of the WarpX class, in order to make it simpler.

This PR places these two functions under a newly created namespace `warpx::embedded_boundary`, inside the files `EmbeddedBoundray/EmbeddedBoundary.H/cpp` .


RemiLehe (approved):

RemiLehe:
Thanks for this PR! This is great!

Would you be able to do a follow-up PR to do the same treatment for the `Mark...Cells` functions (e.g., `MarkUpdateBCellsECT`)

lucafedeli88:
> Thanks for this PR! This is great!
> 
> Would you be able to do a follow-up PR to do the same treatment for the `Mark...Cells` functions (e.g., `MarkUpdateBCellsECT`)

Sure!

PR #5626, opened 2025-01-31
WarpX class: remove unused static variable

lucafedeli88:
`static bool do_device_synchronize;` is unused. Therefore this PR removes it from the `WarpX.H` header.


ax3l (approved):
Ty! :tada:

PR #5627, opened 2025-01-31
Add reference for new article using WarpX [Tyushev (2025)]

roelof-groenewald:
Just adding a new article that uses WarpX to the Science Highlights section.


ax3l (approved):
Awesome, congratulations on the new paper & thanks for adding! :tada:

PR #5628, opened 2025-02-01
Add PMC boundary conditions

dpgrote:
This is a quick PR that adds the PMC boundary condition. This replaced PR #3724. This reuses the existing code for the PEC BC, but passes E into `ApplyPECtoBfield` and B into `ApplyPECtoEfield`. This also touches a few other places where checks of the BC are done, allowing PMC. Note that this includes calling `ApplyReflectiveBoundarytoJfield` with PMC and setting the correct sign for the J correction.

The `FieldBoundaryType` is updated so that `PMC` and `Neumann` are the same.

Note that there is a fair amount of duplicated code when handing the PEC and PMC boundaries. A subsequent PR could clean this up.


Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
```diff
@@ -151,7 +152,7 @@ ProjectionDivCleaner::solve ()
         auto ithi = bcmap.find(WarpX::field_boundary_hi[idim]);
         if (itlo == bcmap.end() || ithi == bcmap.end()) {
             WARPX_ABORT_WITH_MESSAGE(
-                "Field boundary conditions have to be either periodic, PEC or neumann "
+                "Field boundary conditions have to be either periodic, PEC, PMC, or neumann "
```
JustinRayAngus, line 154: Turns out Neumann does nothing for electromagnetic. Should we make it such that the BC defaults to PMC of if the user specifies Neumann for EM simulation?

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
```diff
@@ -141,6 +141,7 @@ ProjectionDivCleaner::solve ()
 
     std::map<FieldBoundaryType, LinOpBCType> bcmap{
         {FieldBoundaryType::PEC, LinOpBCType::Dirichlet},
+        {FieldBoundaryType::PMC, LinOpBCType::Dirichlet},
```
JustinRayAngus: Should this be Neumann for PMC?

Source/BoundaryConditions/WarpXFieldBoundaries.cpp
```diff
@@ -239,7 +284,8 @@ void WarpX::ApplyJfieldBoundary (const int lev, amrex::MultiFab* Jx,
 {
     if (::isAnyBoundary<ParticleBoundaryType::Reflecting>(particle_boundary_lo, particle_boundary_hi) ||
         ::isAnyBoundary<ParticleBoundaryType::Thermal>(particle_boundary_lo, particle_boundary_hi) ||
-        ::isAnyBoundary<FieldBoundaryType::PEC>(field_boundary_lo, field_boundary_hi))
+        ::isAnyBoundary<FieldBoundaryType::PEC>(field_boundary_lo, field_boundary_hi) ||
+        ::isAnyBoundary<FieldBoundaryType::PMC>(field_boundary_lo, field_boundary_hi))
```
JustinRayAngus, line 288: Just wanted to write down what we already discussed. Confusing naming convention through this entire file, but this will all be cleaned up in a future PR so it is ok.

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
dpgrote: Yes, correct.Pr

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
dpgrote: I would say that Neumann should return an error with electromagnetic. I was thinking that that change would be a separate PR, but it could be added to this one.

JustinRayAngus: I think it could make sense to do in this PR. 

We should also change the neumann BC to pmc in the pec_insulator CI tests, which are what motivated this PR in the first place.

dpgrote: I thought about this some more and changed my mind. Since Neumann is the same thing as PMC, Neumann should be allowed with EM. The question is do we allow both Neumann and PMC as input? It looks like we can have the two different names but with the same value so either could be used. I'll make the changes.

JustinRayAngus:
Looks like the development branch has PMC already present in a few places, though it does not appear to be used. The following comments in WarpX.H need to be updated (occurs twice).

(https://github.com/ECP-WarpX/WarpX/blob/10af74faacdc0c34c3648b780f052f6e9e32394a/Source/WarpX.H#L217)

JustinRayAngus (approved):

PR #5634, opened 2025-02-03
Added support to use adios2's flatten_step

### LINKED ISSUE #1655: file based bp5 writer hang
**Describe the bug**
The recent optimization breaks a MPI use case when in file based mode. A minimal code is included below. One can use 2 ranks to see effect. 
In short, at the second flush, rank 1 has nothing to contribute, so it didn't call BP5 while rank 0 did.  In essence, BP5 write is collective.  So rank 0 hangs because inactivity of rank 1. 
If we use variable based, it looks like a flush to ADIOS is forced  (by openPMD-api? ) on all ranks and so it works. 

**To Reproduce**
c++ example: 
```C++
#include <openPMD/openPMD.hpp>
#include <mpi.h>
#include <iostream>
#include <memory>

using std::cout;
using namespace openPMD;

int main(int argc, char *argv[])
{
    int provided;
    MPI_Init_thread(&argc, &argv, MPI_THREAD_MULTIPLE, &provided);

    int mpi_size;
    int mpi_rank;

    MPI_Comm_size(MPI_COMM_WORLD, &mpi_size);
    MPI_Comm_rank(MPI_COMM_WORLD, &mpi_rank);

    auto const value = float(mpi_size*100+mpi_rank);
    std::vector<float> local_data(10 * 300, value);

    std::string filename = "ptl_%T.bp";
    //std::string filename = "ptl.bp";  //this is variable based and it works                                                                                                        

    Series series = Series(filename, Access::CREATE, MPI_COMM_WORLD);

    Datatype datatype = determineDatatype<float>();

    auto myptl = series.writeIterations()[1].particles["ion"];
    Extent global_ptl = {10ul * mpi_size * 300};
    Dataset dataset_ptl = Dataset(datatype, global_ptl, "{}");
    myptl["charge"].resetDataset(dataset_ptl);

    series.flush();

    if (mpi_rank == 0)     // only rank 0 adds data                                                                                                                                  
        myptl["charge"].storeChunk(local_data, {0}, {3000});

    series.flush(); // hangs here                                                                                                                                                    
    MPI_Finalize();

    return 0;
}
```
 

**Software Environment**
 - version of openPMD-api:  latest
 - machine:  Mac
 - version of ADIOS2: latest

**Additional context**
-  I used OPENPMD_ADIOS2_BP5_TypeAgg=EveryoneWritesSerial, but any choice of aggregation will fail.
-  run with 2 cores is enough to see file based has issue. 
- As far as I was aware, this use case  worked fine not long ago. 
- It does not affect HDF5.

guj:
## New Opt-In Feature for BTD

To enable it, put it in the openPMD option through input file:
e.g. 
```
diag1.openpmd_backend = bp5
diag1.adios2_engine.parameters.FlattenSteps = on
```

This feature is useful for BTD use case. Data can be flushed after each buffered writes of a  snapshot

To check weather this feature is in use, try  "bpls -V your_bp_file_name" 

Note: we still debug performance regressions in writing when `FlattenSteps` is used. This will be addressed in follow-up updates to ADIOS2/openPMD-api.

## Bug Fix

Also fixes https://github.com/openPMD/openPMD-api/issues/1655 for BP5 with file-based encoding, i.e., when some ranks have no particles to contribute to a step.

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -2773,7 +2773,7 @@ In-situ capabilities can be used by turning on Sensei or Ascent (provided they a
     Only read if ``<diag_name>.format = sensei``.
     When 1 lower left corner of the mesh is pinned to 0.,0.,0.
 
-* ``<diag_name>.openpmd_backend`` (``bp``, ``h5`` or ``json``) optional, only used if ``<diag_name>.format = openpmd``
+* ``<diag_name>.openpmd_backend`` (``bp5``, ``bp4``, ``h5`` or ``json``) optional, only used if ``<diag_name>.format = openpmd``
     `I/O backend <https://openpmd-api.readthedocs.io/en/latest/backends/overview.html>`_ for `openPMD <https://www.openPMD.org>`_ data dumps.
     ``bp`` is the `ADIOS I/O library <https://csmd.ornl.gov/adios>`_, ``h5`` is the `HDF5 format <https://www.hdfgroup.org/solutions/hdf5/>`_, and ``json`` is a `simple text format <https://en.wikipedia.org/wiki/JSON>`_.
     ``json`` only works with serial/single-rank jobs.
@@ -2795,19 +2795,26 @@ In-situ capabilities can be used by turning on Sensei or Ascent (provided they a
 
     .. code-block:: text
 
-        <diag_name>.adios2_operator.type = blosc
-        <diag_name>.adios2_operator.parameters.compressor = zstd
-        <diag_name>.adios2_operator.parameters.clevel = 1
-        <diag_name>.adios2_operator.parameters.doshuffle = BLOSC_BITSHUFFLE
-        <diag_name>.adios2_operator.parameters.threshold = 2048
-        <diag_name>.adios2_operator.parameters.nthreads = 6  # per MPI rank (and thus per GPU)
+       <diag_name>.adios2_operator.type = blosc
+       <diag_name>.adios2_operator.parameters.compressor = zstd
+       <diag_name>.adios2_operator.parameters.clevel = 1
+       <diag_name>.adios2_operator.parameters.doshuffle = BLOSC_BITSHUFFLE
+       <diag_name>.adios2_operator.parameters.threshold = 2048
+       <diag_name>.adios2_operator.parameters.nthreads = 6  # per MPI rank (and thus per GPU)
 
     or for the lossy ZFP compressor using very strong compression per scalar:
 
     .. code-block:: text
 
-        <diag_name>.adios2_operator.type = zfp
-        <diag_name>.adios2_operator.parameters.precision = 3
+       <diag_name>.adios2_operator.type = zfp
+       <diag_name>.adios2_operator.parameters.precision = 3
+
+    For back-transformed diagnostics with ADIOS BP5, we are experimenting with a new option for variable-based encoding that "flattens" the output steps, aiming to increase write and read performance:
+
+    .. code-block:: text
+
+       <diag_name>.openpmd_backend = bp5
+       <diag_name>.adios2_engine.parameters.FlattenSteps = on
 
 * ``<diag_name>.adios2_engine.type`` (``bp4``, ``sst``, ``ssc``, ``dataman``) optional,
     `ADIOS2 Engine type <https://openpmd-api.readthedocs.io/en/0.16.1/details/backendconfig.html#adios2>`__ for `openPMD <https://www.openPMD.org>`_ data dumps.
--- a/Source/Diagnostics/WarpXOpenPMD.H
+++ b/Source/Diagnostics/WarpXOpenPMD.H
@@ -176,6 +176,18 @@ private:
     }
   }
 
+  /** Flushing out data of the current openPMD iteration
+   *
+   * @param[in] isBTD if the current diagnostic is BTD
+   *
+   * if isBTD=false, apply the default flush behaviour
+   * if isBTD=true,  advice to use ADIOS Put() instead of PDW for better performance.
+   *
+   * iteration.seriesFlush() is used instead of series.flush()
+   *  because the latter flushes only if data is dirty
+   *  this causes trouble when the underlying writing function is collective (like PDW)
+   *
+   */
   void flushCurrent (bool isBTD) const;
 
   /** This function does initial setup for the fields when interation is newly created
--- a/Source/Diagnostics/WarpXOpenPMD.H
+++ b/Source/Diagnostics/WarpXOpenPMD.H
@@ -176,7 +176,7 @@ private:
     }
   }
 
-  void flushCurrent(bool isBTD) const;
+  void flushCurrent (bool isBTD) const;
 
   /** This function does initial setup for the fields when interation is newly created
    *  @param[in] meshes   The meshes in a series
--- a/Source/Diagnostics/WarpXOpenPMD.H
+++ b/Source/Diagnostics/WarpXOpenPMD.H
@@ -176,6 +176,7 @@ private:
     }
   }
 
+  void flushCurrent(bool isBTD) const;
 
   /** This function does initial setup for the fields when interation is newly created
    *  @param[in] meshes   The meshes in a series
--- a/Source/Diagnostics/WarpXOpenPMD.cpp
+++ b/Source/Diagnostics/WarpXOpenPMD.cpp
@@ -402,18 +402,6 @@ WarpXOpenPMDPlot::~WarpXOpenPMDPlot ()
   }
 }
 
-/*
- * Flushing out data of the current openPMD iteration
- * @param [in]: isBTD    if the current diagnostic is BTD
- *
- * if isBTD=false, apply the default flush behaviour
- * if isBTD=true,  advice to use ADIOS Put() instead of PDW for better performance.
- *
- *    iteration.seriesFlush() is used instead of series.flush()
- *    because the latter flushes only if data is dirty
- *    this causes trouble when the underlying writing function is collective (like PDW)
- *
- */
 void WarpXOpenPMDPlot::flushCurrent (bool isBTD) const
 {
      WARPX_PROFILE("WarpXOpenPMDPlot::flushCurrent");
--- a/Source/Diagnostics/WarpXOpenPMD.cpp
+++ b/Source/Diagnostics/WarpXOpenPMD.cpp
@@ -391,7 +391,7 @@ WarpXOpenPMDPlot::WarpXOpenPMDPlot (
 {
     m_OpenPMDoptions = detail::getSeriesOptions(operator_type, operator_parameters,
                                                 engine_type, engine_parameters);
-    amrex::Print()<<".... openPMD options used: "<<m_OpenPMDoptions<<" \n";
+    amrex::Print() << Utils::TextMsg::Info("openPMD options used: " + m_OpenPMDoptions);
 }
 
 WarpXOpenPMDPlot::~WarpXOpenPMDPlot ()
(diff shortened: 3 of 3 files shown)


Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -402,6 +403,21 @@ WarpXOpenPMDPlot::~WarpXOpenPMDPlot ()
...
+     WARPX_PROFILE("WarpXOpenPMDPlot::flushCurrent");
+     // open files from all processors, in case some will not contribute below
+     // m_Series->flush(); ## NOTE flush only flushes if data is dirty. When the underlying function is collective
+     //                    ## like PDW, there will be traouble.
+     // the change will be use PDW when not BTD, and Put if BTD to avoid slowing down due to  multiple writes of small data
+     openPMD::Iteration currIteration = GetIteration(m_CurrentStep, isBTD);
+     if (isBTD)
+         currIteration.seriesFlush(  "adios2.engine.preferred_flush_target = \"buffer\"" );
+     else
+         currIteration.seriesFlush();
```
```suggestion
     if (isBTD) {
         currIteration.seriesFlush(  "adios2.engine.preferred_flush_target = \"buffer\"" );
     }
     else {
         currIteration.seriesFlush();
     }
```
Could you please add braces here?

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -642,6 +657,16 @@ for (const auto & particle_diag : particle_diags) {
         pc->getCharge(), pc->getMass(),
         isBTD, isLastBTDFlush);
     }
+
+    auto hasOption=m_OpenPMDoptions.find("FlattenSteps");
+    bool flattenSteps = isBTD && (m_Series->backend() == "ADIOS2") && (hasOption != std::string::npos);
```
```suggestion
    const bool flattenSteps = isBTD && (m_Series->backend() == "ADIOS2") && (hasOption != std::string::npos);
```
Could you please declare this as a `const` ?

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -642,6 +657,16 @@ for (const auto & particle_diag : particle_diags) {
...
+
+    if ( flattenSteps)
```
```suggestion
    if (flattenSteps)
```
I would suggest to eliminate this space

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -797,7 +823,8 @@ WarpXOpenPMDPlot::DumpToFile (ParticleContainer* pc,
         }
     }
 
-    m_Series->flush();
+    //m_Series->flush();
```
```suggestion
```
You may want to erase this line or maybe replace it with a comment on why `m_Series->flush();` is no longer necessary.

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -1469,7 +1496,8 @@ WarpXOpenPMDPlot::WriteOpenPMDFieldsAll ( //const std::string& filename,
         amrex::Gpu::streamSynchronize();
 #endif
         // Flush data to disk after looping over all components
-        m_Series->flush();
+        //m_Series->flush();
```
```suggestion
```
same as above

lucafedeli88 (commented):
Thanks for this PR, @guj !
I've left few comments, mainly to make `clang-tidy` and `CodeQL` happy.

Source/Diagnostics/WarpXOpenPMD.H
```diff
@@ -176,6 +176,7 @@ private:
     }
   }
 
+  void flushCurrent(bool isBTD) const;
```
lucafedeli88: Would it be possible to add a doc-string to document this function?

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -391,6 +391,7 @@ WarpXOpenPMDPlot::WarpXOpenPMDPlot (
 {
     m_OpenPMDoptions = detail::getSeriesOptions(operator_type, operator_parameters,
                                                 engine_type, engine_parameters);
+    amrex::Print()<<".... openPMD options used: "<<m_OpenPMDoptions<<" \n";
```
lucafedeli88: It would be ideal to format the output here as an "Info" message for consistency:
`amrex::Print() << Utils::TextMsg::Info(your_string) ;`

(note that you don't need to add `\n` at the end of the message)

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -402,6 +403,23 @@ WarpXOpenPMDPlot::~WarpXOpenPMDPlot ()
   }
 }
 
+
+void WarpXOpenPMDPlot::flushCurrent(bool isBTD) const
+{
+     WARPX_PROFILE("WarpXOpenPMDPlot::flushCurrent");
+     // open files from all processors, in case some will not contribute below
+     // m_Series->flush(); ## NOTE flush only flushes if data is dirty. When the underlying function is collective
+     //                    ## like PDW, there will be traouble.
```
```suggestion
     //                    ## like PDW, there will be trouble.
```
typo

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -402,6 +403,23 @@ WarpXOpenPMDPlot::~WarpXOpenPMDPlot ()
   }
 }
 
+
+void WarpXOpenPMDPlot::flushCurrent(bool isBTD) const
+{
+     WARPX_PROFILE("WarpXOpenPMDPlot::flushCurrent");
+     // open files from all processors, in case some will not contribute below
```
lucafedeli88: These comments can be adapted to write a doc-string for the function declaration

lucafedeli88 (commented):
Thanks again, @guj , for your PR.
I've added few more comments. I'll let @ax3l  comment on the ADIOS2 option that you are adding.

Source/Diagnostics/WarpXOpenPMD.cpp
guj: OK used Info to output  instead

Source/Diagnostics/WarpXOpenPMD.H
guj: I added doc for this function

Source/Diagnostics/WarpXOpenPMD.cpp
guj: Added documentation for this function.

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -391,6 +391,7 @@ WarpXOpenPMDPlot::WarpXOpenPMDPlot (
 {
     m_OpenPMDoptions = detail::getSeriesOptions(operator_type, operator_parameters,
                                                 engine_type, engine_parameters);
+    amrex::Print() << Utils::TextMsg::Info("openPMD options used: " + m_OpenPMDoptions);
```
ax3l:
```suggestion
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -402,6 +403,31 @@ WarpXOpenPMDPlot::~WarpXOpenPMDPlot ()
...
+ *
+ * if isBTD=false, apply the default flush behaviour
+ * if isBTD=true,  advice to use ADIOS Put() instead of PDW for better performance.
+ *
+ *    iteration.seriesFlush() is used instead of series.flush()
+ *    because the latter flushes only if data is dirty
+ *    this causes trouble when the underlying writing function is collective (like PDW)
+ *
+ */
+void WarpXOpenPMDPlot::flushCurrent(bool isBTD) const
```
ax3l:
```suggestion
void WarpXOpenPMDPlot::flushCurrent (bool isBTD) const
```

Source/Diagnostics/WarpXOpenPMD.H
ax3l: Please add the doc in this `.H` file, not the `.cpp` file.

ax3l:
```suggestion
  void flushCurrent (bool isBTD) const;
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -642,6 +667,16 @@ for (const auto & particle_diag : particle_diags) {
         pc->getCharge(), pc->getMass(),
         isBTD, isLastBTDFlush);
     }
+
+    auto hasOption=m_OpenPMDoptions.find("FlattenSteps");
```
ax3l:
```suggestion
    auto hasOption = m_OpenPMDoptions.find("FlattenSteps");
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -402,6 +403,31 @@ WarpXOpenPMDPlot::~WarpXOpenPMDPlot ()
   }
 }
 
+/*
```
ax3l: Please move doxygen strings to `.H` files

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -642,6 +667,16 @@ for (const auto & particle_diag : particle_diags) {
         pc->getCharge(), pc->getMass(),
         isBTD, isLastBTDFlush);
     }
+
+    auto hasOption=m_OpenPMDoptions.find("FlattenSteps");
+    const bool flattenSteps = isBTD && (m_Series->backend() == "ADIOS2") && (hasOption != std::string::npos);
+
+    if (flattenSteps)
+    {
+       // forcing new step so data from each btd batch can be flushed out
```
ax3l:
```suggestion
       // forcing new step so data from each btd batch in
       // preferred_flush_target="buffer" can be flushed out
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -402,6 +403,31 @@ WarpXOpenPMDPlot::~WarpXOpenPMDPlot ()
...
+ *    this causes trouble when the underlying writing function is collective (like PDW)
+ *
+ */
+void WarpXOpenPMDPlot::flushCurrent(bool isBTD) const
+{
+     WARPX_PROFILE("WarpXOpenPMDPlot::flushCurrent");
+
+     openPMD::Iteration currIteration = GetIteration(m_CurrentStep, isBTD);
+     if (isBTD) {
+         currIteration.seriesFlush(  "adios2.engine.preferred_flush_target = \"buffer\"" );
```
ax3l, line 416: Need to add a comment:
```suggestion
         // delayed until all fields and particles are registered for flush
         // and dumped once via flattenSteps
         currIteration.seriesFlush(  "adios2.engine.preferred_flush_target = \"buffer\"" );
```

ax3l (commented):
Formatting & Inline Comments

Source/Diagnostics/WarpXOpenPMD.cpp
ax3l: done.

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -402,6 +402,21 @@ WarpXOpenPMDPlot::~WarpXOpenPMDPlot ()
   }
 }
 
+void WarpXOpenPMDPlot::flushCurrent (bool isBTD) const
+{
+     WARPX_PROFILE("WarpXOpenPMDPlot::flushCurrent");
+
+     openPMD::Iteration currIteration = GetIteration(m_CurrentStep, isBTD);
+     if (isBTD) {
```
ax3l: Also check if `flattenSteps` is used.

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -642,6 +656,17 @@ for (const auto & particle_diag : particle_diags) {
         pc->getCharge(), pc->getMass(),
         isBTD, isLastBTDFlush);
     }
+
+    auto hasOption = m_OpenPMDoptions.find("FlattenSteps");
```
ax3l, line 663: Docs missing in `parameters.rst`

Source/Diagnostics/WarpXOpenPMD.cpp
ax3l: Added

Source/Diagnostics/WarpXOpenPMD.cpp
ax3l: added

Docs/source/usage/parameters.rst
```diff
@@ -2795,19 +2795,26 @@ In-situ capabilities can be used by turning on Sensei or Ascent (provided they a
...
-        <diag_name>.adios2_operator.parameters.precision = 3
+       <diag_name>.adios2_operator.type = zfp
+       <diag_name>.adios2_operator.parameters.precision = 3
+
+    For back-transformed diagnostics with ADIOS BP5, we are experimenting with a new option for variable-based encoding that "flattens" the output steps, aiming to increase write and read performance:
+
+    .. code-block:: text
+
+       <diag_name>.openpmd_backend = bp5
+       <diag_name>.adios2_engine.parameters.FlattenSteps = on
```
ax3l, line 2817: @guj does this only have an effect for variable based encoding or generally (e.g., file based encoding)?

ax3l (approved):

Source/Diagnostics/WarpXOpenPMD.H
ax3l: done

Docs/source/usage/parameters.rst
guj: general

guj: but right now flatten steps  is only applied to BTD

PR #5481, opened 2024-11-22
Remove NamedComponentParticleContainer (Use from AMReX)

atmyers:
This capability has been upstreamed to AMReX.

## Depends on

- [x] https://github.com/AMReX-Codes/amrex/pull/4322 / #5643

PR #1706, opened 2021-02-11
[WIP] CMake Env Var: COMPUTE

ax3l:
Introduce an environment variable that can be used in `warpx.profile` files to pre-select a compute environment on a queue-cluster combination.

### TODO

- [ ] maybe all-caps (as AMReX); but Python control env vars in `setup.py` are currently case sensitive (also make ALL-caps: #2317)
- [ ] use for other cluster docs as well
- [ ] add general doc section for build-time env var control

PR #5637, opened 2025-02-04
Diagnostics : EM fields on particles

grobertdautun:
This PR implements the options `Ex` `Ey` `Ez` `Bx` `By` and `Bz` on the particles diagnostics to output the electromagnetic fields seen by the particles. 

I added a test case `test_2d_particle_EM_diagnostics` with its analysis. This case represents a single particle fixed in a box traversed by a laser pulse, and the analysis verifies that the fields are matching the theoretical ones.

There are two limitations to this diagnostics that should be updated later :
- the external fields are not yet written on the diagnostics as it would require a change in the `getExternalEBField` template
- it only works without mesh refinement


Source/Diagnostics/ParticleIO.cpp
```diff
@@ -293,3 +294,114 @@ storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
...
+
+
+    tmp.NewRealComp("Ex");
+    tmp.NewRealComp("Ey");
+    tmp.NewRealComp("Ez");
+    tmp.NewRealComp("Bx");
+    tmp.NewRealComp("By");
+    tmp.NewRealComp("Bz");
+
+    int const Ex_index = tmp.getParticleComps().at("Ex");
```
ax3l: There will be a minimal change we need to add to this PR once we merged #5481 (simplification of named SoA bookkeeping), but it will be easy to update once your PR is ready/out-of-WIP.

Source/Diagnostics/ParticleIO.cpp
```diff
@@ -293,3 +294,114 @@ storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
...
+        amrex::ParticleReal* Bz_particle_arr = pti.GetStructOfArrays().GetRealData(Bz_index).dataPtr();
+
+        const auto box = pti.tilebox();
+        const amrex::XDim3 xyzmin = WarpX::LowerCorner(box, lev0, 0._rt);
+        const Dim3 lo = lbound(box);
+
+        // Loop over the particles and update their position
+        amrex::ParallelFor( pti.numParticles(),
+            [=] AMREX_GPU_DEVICE (long ip) {
+               // !!!!!!!!!! pas fini
```
lucafedeli88:
```suggestion
```

;-)

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -590,6 +590,9 @@ for (const auto & particle_diag : particle_diags) {
     if ( particle_diag.m_plot_phi ) {
         storePhiOnParticles( tmp, WarpX::electrostatic_solver_id, !use_pinned_pc );
```
lucafedeli88: @RemiLehe , do you know why for `storePhiOnParticles` we pass `!use_pinned_pc` instead of `full_diag` ?

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -590,6 +590,9 @@ for (const auto & particle_diag : particle_diags) {
...
     }
+    if ( particle_diag.m_plot_EM ) {
+        storeEMFieldsOnParticles( tmp, WarpX::electromagnetic_solver_id, !use_pinned_pc ); // !!!!!!!!!!!!! why use_pinned_pc here? and not full_diag ????
```
```suggestion
        storeEMFieldsOnParticles( tmp, WarpX::electromagnetic_solver_id, !use_pinned_pc );
```
The comment must be removed, but we should definitely answer the question!

Source/Diagnostics/ParticleIO.cpp
```diff
@@ -293,3 +294,114 @@ storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
...
+        amrex::ParticleReal* Ez_particle_arr = pti.GetStructOfArrays().GetRealData(Ez_index).dataPtr();
+        amrex::ParticleReal* Bx_particle_arr = pti.GetStructOfArrays().GetRealData(Bx_index).dataPtr();
+        amrex::ParticleReal* By_particle_arr = pti.GetStructOfArrays().GetRealData(By_index).dataPtr();
+        amrex::ParticleReal* Bz_particle_arr = pti.GetStructOfArrays().GetRealData(Bz_index).dataPtr();
+
+        const auto box = pti.tilebox();
+        const amrex::XDim3 xyzmin = WarpX::LowerCorner(box, lev0, 0._rt);
+        const Dim3 lo = lbound(box);
+
+        // Loop over the particles and update their position
```
lucafedeli88:
```suggestion
        // Loop over the particles and compute the EM field using the doGatherShapeN function
```

I would suggest to change this comment

lucafedeli88: I would suggest to change this

Source/Diagnostics/ParticleIO.cpp
```diff
@@ -299,3 +300,113 @@ storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
...
+
+        // Loop over the particles and compute the EM field using the doGatherShapeN function
+        amrex::ParallelFor( pti.numParticles(),
+            [=] AMREX_GPU_DEVICE (long ip) {
+               amrex::ParticleReal xp, yp, zp;
+               getPosition(ip, xp, yp, zp);
+
+                Ex_particle_arr[ip] = 0._rt;
+                Ey_particle_arr[ip] = 0._rt;
+                Ez_particle_arr[ip] = 0._rt;
```
ax3l: alignment is off by one whitespace :sweat_smile:

Tools/PostProcessing/read_variable_based_adios2.py
```diff
@@ -0,0 +1,145 @@
+import numpy as np
+from adios2 import Stream
```
ax3l: As discussed: we could use [openPMD-api](https://openpmd-api.readthedocs.io/en/0.16.1/usage/examples.html#python).

I think for openPMD-viewer support, we rely on this upcoming feature: https://github.com/openPMD/openPMD-api/pull/1706

Current ADIOS2 support status:
https://openpmd-api.readthedocs.io/en/0.16.1/backends/adios2.html

@franzpoeschel, @guj @pnorbert can you remind me if we can do variable-based encoding without steps (i.e., can we already set something during write that allows-us to read variable based encoding with random access)?

Docs/source/usage/parameters.rst
```diff
@@ -2916,9 +2916,9 @@ In-situ capabilities can be used by turning on Sensei or Ascent (provided they a
 * ``<diag_name>.<species_name>.variables`` (list of `strings` separated by spaces, optional)
     List of particle quantities to write to output.
     Choices are ``x``, ``y``, ``z`` for the particle positions (3D and RZ), ``x`` & ``z`` in 2D, ``z`` in 1D,
-    ``w`` for the particle weight and ``ux``, ``uy``, ``uz`` for the particle momenta.
+    ``w`` for the particle weight, ``ux``, ``uy``, ``uz`` for the particle momenta, and ``EM`` for the electromagnetic fields.
```
RemiLehe: Could we instead allow the user to pass ``Ex``, ``Ey``, ``Bx``, etc.

Docs/source/usage/parameters.rst
```diff
@@ -2784,7 +2784,7 @@ In-situ capabilities can be used by turning on Sensei or Ascent (provided they a
 * ``<diag_name>.openpmd_encoding`` (optional, ``v`` (variable based), ``f`` (file based) or ``g`` (group based) ) only read if ``<diag_name>.format = openpmd``.
      openPMD `file output encoding <https://openpmd-api.readthedocs.io/en/0.16.1/usage/concepts.html#iteration-and-series>`__.
      File based: one file per timestep (slower), group/variable based: one file for all steps (faster)).
-     ``variable based`` is an `experimental feature with ADIOS2 <https://openpmd-api.readthedocs.io/en/0.16.1/backends/adios2.html#experimental-new-adios2-schema>`__ and not supported for back-transformed diagnostics.
+     ``variable based`` is an `experimental feature with ADIOS2 <https://openpmd-api.readthedocs.io/en/0.16.1/backends/adios2.html#experimental-new-adios2-schema>`__ and not supported for back-transformed diagnostics. This format is also not supported by OpenPMDTimeSeries at the moment, the script :download:`read_variable_based_adios2.py <../../../Tools/PostProcessing/read_variable_based_adios2.py>` provides basic functions to read variable-based particle diagnostics directly with the ``adios2`` module.
```
RemiLehe: This could go in a separate PR. Could you remove from this one?

Tools/PostProcessing/read_variable_based_adios2.py
```diff
@@ -0,0 +1,145 @@
+import numpy as np
```
RemiLehe: This file could go in a separate PR. Could you remove from this one?

Source/Diagnostics/ParticleIO.cpp
```diff
@@ -299,3 +300,113 @@ storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
...
+                Bx_particle_arr[ip] = 0._rt;
+                By_particle_arr[ip] = 0._rt;
+                Bz_particle_arr[ip] = 0._rt;
```
RemiLehe: Could we add the external fields as part of this call?

Source/Diagnostics/ParticleIO.cpp
```diff
@@ -299,3 +300,113 @@ storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
...
+    );
+
+    constexpr auto lev0=0;
+
+    amrex::MultiFab const& Ex = *warpx.m_fields.get(FieldType::Efield_fp, Dir{0}, lev0);
+    amrex::MultiFab const& Ey = *warpx.m_fields.get(FieldType::Efield_fp, Dir{1}, lev0);
+    amrex::MultiFab const& Ez = *warpx.m_fields.get(FieldType::Efield_fp, Dir{2}, lev0);
+    amrex::MultiFab const& Bx = *warpx.m_fields.get(FieldType::Bfield_fp, Dir{0}, lev0);
+    amrex::MultiFab const& By = *warpx.m_fields.get(FieldType::Bfield_fp, Dir{1}, lev0);
+    amrex::MultiFab const& Bz = *warpx.m_fields.get(FieldType::Bfield_fp, Dir{2}, lev0);
```
RemiLehe:
```suggestion
    amrex::MultiFab const& Ex = *warpx.m_fields.get(FieldType::Efield_aux, Dir{0}, lev0);
    amrex::MultiFab const& Ey = *warpx.m_fields.get(FieldType::Efield_aux, Dir{1}, lev0);
    amrex::MultiFab const& Ez = *warpx.m_fields.get(FieldType::Efield_aux, Dir{2}, lev0);
    amrex::MultiFab const& Bx = *warpx.m_fields.get(FieldType::Bfield_aux, Dir{0}, lev0);
    amrex::MultiFab const& By = *warpx.m_fields.get(FieldType::Bfield_aux, Dir{1}, lev0);
    amrex::MultiFab const& Bz = *warpx.m_fields.get(FieldType::Bfield_aux, Dir{2}, lev0);
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -589,6 +589,9 @@ for (const auto & particle_diag : particle_diags) {
     if ( particle_diag.m_plot_phi ) {
         storePhiOnParticles( tmp, WarpX::electrostatic_solver_id, !use_pinned_pc );
     }
+    if ( particle_diag.m_plot_EM ) {
+        storeEMFieldsOnParticles( tmp, WarpX::electromagnetic_solver_id, !use_pinned_pc );
+    }
```
RemiLehe, line 619:
```suggestion
    // Detect whether this is a FullDiagnostic (use_pinned_pc = 0) or
    // a BoundaryScrapingDiagnostic/BackTransformedDiagnostic (use_pinned_pc=1)
    bool const is_full_diagnostic = !use_pinned_pc;
    if ( particle_diag.m_plot_phi ) {
        storePhiOnParticles( tmp, WarpX::electrostatic_solver_id, is_full_diagnostic);
    }
    if ( particle_diag.m_plot_EM ) {
        storeEMFieldsOnParticles( tmp, WarpX::electromagnetic_solver_id, is_full_diagnostic);
    }
```

Source/Diagnostics/ParticleIO.cpp
RemiLehe: See for instance: https://github.com/ECP-WarpX/WarpX/blob/development/Source/Particles/ElementaryProcess/Ionization.H#L115

Tools/PostProcessing/read_variable_based_adios2.py
franzpoeschel: > (i.e., can we already set something during write that allows-us to read variable based encoding with random access)

Not yet, that is what #1706 (and follow-ups for changing metadata) will bring (setting a flag during write will not be needed though).
Until that point, it's better to continue using file encoding for random access.

Source/Diagnostics/ParticleIO.cpp
```diff
@@ -299,3 +301,164 @@ storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
...
+        const auto ex_type = Ex.ixType();
+        const auto ey_type = Ey.ixType();
+        const auto ez_type = Ez.ixType();
+        const auto bx_type = Bx.ixType();
+        const auto by_type = By.ixType();
+        const auto bz_type = Bz.ixType();
+
+        const auto getPosition = GetParticlePosition<PIdx>(pti);
+        const auto getExternalEB = GetExternalEBField(pti);
+        if (fields_to_plot[0]) amrex::ParticleReal* Ex_particle_arr = pti.GetStructOfArrays().GetRealData(fields_index[0]).dataPtr();
```
lucafedeli88: What about
```
autoamrex::ParticleReal* Ex_particle_arr = (fields_to_plot[0]) ? pti.GetStructOfArrays().GetRealData(fields_index[0]).dataPtr() : nullptr;
```
?

Source/Diagnostics/ParticleIO.cpp
```diff
@@ -299,3 +301,164 @@ storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
...
+                    doDirectGatherVectorField(
+                        xp, yp, zp,
+                        Bx_particle, By_particle, Bz_particle,
+                        Bx_grid, By_grid, Bz_grid,
+                        bx_type, by_type, bz_type,
+                        dinv, xyzmin, lo, n_rz_azimuthal_modes //, nox, galerkin_interpolation
+                    );
+                }
+
+                if (ex_control == doEx) {
```
lucafedeli88: if

Source/Diagnostics/ParticleIO.cpp
```diff
@@ -299,3 +301,164 @@ storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
...
+
+    constexpr auto lev0=0;
+    const amrex::XDim3 dinv = WarpX::InvCellSize(lev0);
+    //const bool galerkin_interpolation = WarpX::galerkin_interpolation;
+    //const int nox = WarpX::nox;
+    const int n_rz_azimuthal_modes = WarpX::n_rz_azimuthal_modes;
+
+    static constexpr const fields_names[6] = {"Ex", "Ey", "Ez", "Bx", "By", "Bz"};
+
+    int fields_index[6];
```
lucafedeli88: I think that it would be better to use an `amrex::Array` here.

Source/Diagnostics/ParticleIO.cpp
```diff
@@ -299,3 +301,188 @@ storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
...
+        "output of the electromagnetic fields on particles only works without mesh refinement"
+    );
+
+    constexpr auto lev0=0;
+    const amrex::XDim3 dinv = WarpX::InvCellSize(lev0);
+    //const bool galerkin_interpolation = WarpX::galerkin_interpolation;
+    //const int nox = WarpX::nox;
+    const int n_rz_azimuthal_modes = WarpX::n_rz_azimuthal_modes;
+
+    static constexpr const char* fields_names[6] = {"Ex", "Ey", "Ez", "Bx", "By", "Bz"};
```
grobertdautun: use a str vector

Source/Diagnostics/ParticleIO.cpp
```diff
@@ -299,3 +301,188 @@ storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
...
+        amrex::ParallelFor(
+            TypeList<CompileTimeOptions<doEx, noEx>, CompileTimeOptions<doEy, noEy>, CompileTimeOptions<doEz, noEz>,
+            CompileTimeOptions<doBx, noBx>, CompileTimeOptions<doBy, noBy>, CompileTimeOptions<doBz, noBz>>{},
+            {Ex_runtime_flag, Ey_runtime_flag, Ez_runtime_flag, Bx_runtime_flag, By_runtime_flag, Bz_runtime_flag},
+            pti.numParticles(),
+            [=] AMREX_GPU_DEVICE (long ip, auto ex_control, auto ey_control, auto ez_control,
+                auto bx_control, auto by_control, auto bz_control)
+                {
+                amrex::ParticleReal xp, yp, zp;
+                [[maybe_unused]] amrex::ParticleReal Ex_particle, Ey_particle, Ez_particle, Bx_particle, By_particle, Bz_particle;
```
grobertdautun: def direct @ 0._rt + 1 per line

Source/Particles/ParticleIO.H
```diff
@@ -92,4 +92,7 @@ void
 storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
     ElectrostaticSolverAlgo electrostatic_solver_id, bool is_full_diagnostic );
 
+void
```
lucafedeli88, line 106: Could you please add a Doxygen comment to describe this function?

Source/Particles/ParticleIO.H
```diff
@@ -92,4 +92,7 @@ void
...
+storeEMFieldsOnParticles (PinnedMemoryParticleContainer& tmp,
+    ElectromagneticSolverAlgo electromagnetic_solver_id, const bool fields_to_plot[], const int depos_order, const bool galerkin_interpolation, bool is_full_diagnostic );
```
lucafedeli88:
```suggestion
    ElectromagneticSolverAlgo electromagnetic_solver_id,  bool fields_to_plot[],  int depos_order,  bool galerkin_interpolation, bool is_full_diagnostic );
```

There is no need for `const` here, since these parameters are passed by value ( `const` is only meaningful in the corresponding `.cpp` file). I would suggest removing it.

Source/Diagnostics/ParticleIO.cpp
```diff
@@ -299,3 +301,217 @@ storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
...
+
+namespace
+{
+// This template is isolated in a separate namespace to simplify the usage of
+// the doDirectGatherVectorField template and avoid branching
+template <int depos_order, bool galerkin_interpolation>
+void
+storeEMFieldsOnParticles_t (PinnedMemoryParticleContainer& tmp,
+                            ElectromagneticSolverAlgo electromagnetic_solver_id,
+                            const bool fields_to_plot[],
```
lucafedeli88: I think that an `amrex::Array<bool, N>` passed as a constant reference could be preferable with respect to a C-style bool array here.

Source/Diagnostics/ParticleIO.cpp
```diff
@@ -299,3 +301,217 @@ storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
         }
     }
 }
+
+namespace
+{
+// This template is isolated in a separate namespace to simplify the usage of
+// the doDirectGatherVectorField template and avoid branching
```
lucafedeli88: I would suggest changing the description to something like the following:

```suggestion
// This template function is called with all the possible combinations 
// of the template parameters by storeEMFieldsOnParticles.
// depos_order and galerkin_interpolation must be template parameters
// because they are template parameters of doDirectGatherVectorField,
// which is called by this function.
```

Source/Diagnostics/ParticleIO.cpp
```diff
@@ -299,3 +301,217 @@ storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
...
+template <int depos_order, bool galerkin_interpolation>
+void
+storeEMFieldsOnParticles_t (PinnedMemoryParticleContainer& tmp,
+                            ElectromagneticSolverAlgo electromagnetic_solver_id,
+                            const bool fields_to_plot[],
+                            bool is_full_diagnostic)
```
lucafedeli88:
```suggestion
                            const bool is_full_diagnostic)
```
`const` can be used here.

Source/Diagnostics/ParticleIO.cpp
```diff
@@ -299,3 +301,217 @@ storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
...
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+        electromagnetic_solver_id != ElectromagneticSolverAlgo::None,
+        "output of the electromagnetic fields on the particles was requested, "
+        "but this is only available with an electromagnetic solver.");
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+        is_full_diagnostic,
+        "Output of the electromagnetic fields on the particles was requested, "
+        "but this is only available with `diag_type = Full`.");
+
+    auto& warpx = WarpX::GetInstance();
```
lucafedeli88: Would it be possible to pass `finestLevel` as a parameter instead of relying on `WarpX::GetInstance();` ?

Source/Diagnostics/ParticleIO.cpp
```diff
@@ -299,3 +301,217 @@ storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
...
+    );
+
+    constexpr auto lev0=0;
+    const amrex::XDim3 dinv = WarpX::InvCellSize(lev0);
+    const int n_rz_azimuthal_modes = WarpX::n_rz_azimuthal_modes;
+
+    auto fields_names = amrex::Array<std::string, 6>{
+        "Ex", "Ey", "Ez", "Bx", "By", "Bz"};
+
+    auto fields_index = amrex::Array<int, 6>{0,0,0,0,0,0};
```
lucafedeli88: Maybe you  could consider setting the default field index to `-1`, so that fields that must not be plotted would also have an invalid index.

Source/Diagnostics/ParticleIO.cpp
```diff
@@ -299,3 +301,217 @@ storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
...
+                getPosition(ip, xp, yp, zp);
+
+                [[maybe_unused]] amrex::ParticleReal Ex_particle = 0._rt;
+                [[maybe_unused]] amrex::ParticleReal Ey_particle = 0._rt;
+                [[maybe_unused]] amrex::ParticleReal Ez_particle = 0._rt;
+                [[maybe_unused]] amrex::ParticleReal Bx_particle = 0._rt;
+                [[maybe_unused]] amrex::ParticleReal By_particle = 0._rt;
+                [[maybe_unused]] amrex::ParticleReal Bz_particle = 0._rt;
+
+                if (ex_control == noEx && ey_control == noEy && ez_control == noEz &&
```
lucafedeli88: This should be `constexpr` right?

Source/Diagnostics/ParticleIO.cpp
```diff
@@ -299,3 +301,217 @@ storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
...
+                    doDirectGatherVectorField<depos_order - galerkin_interpolation, depos_order>(
+                        xp, yp, zp,
+                        Bx_particle, By_particle, Bz_particle,
+                        Bx_grid, By_grid, Bz_grid,
+                        bx_type, by_type, bz_type,
+                        dinv, xyzmin, lo, n_rz_azimuthal_modes
+                    );
+                }
+
+                // first capture of the variables in constexpr is not supported
```
lucafedeli88: I would suggest  adding by which compiler  the feature is not supported and a consideration like: `so we have to define these references here`

Source/Diagnostics/ParticleIO.cpp
```diff
@@ -299,3 +301,217 @@ storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
...
+
+}
+} // namespace
+
+void storeEMFieldsOnParticles(PinnedMemoryParticleContainer& tmp,
+                              ElectromagneticSolverAlgo electromagnetic_solver_id,
+                              const bool fields_to_plot[],
+                              const int depos_order,
+                              const bool galerkin_interpolation,
+                              bool is_full_diagnostic)
```
lucafedeli88:
```suggestion
                              const bool is_full_diagnostic)
```
This can be `const`

Source/Diagnostics/ParticleIO.cpp
```diff
@@ -299,3 +301,217 @@ storePhiOnParticles ( PinnedMemoryParticleContainer& tmp,
...
+{
+    if (galerkin_interpolation) {
```
lucafedeli88: I would suggest adding a check to ensure that `depos_order` is < 5 to generate either an error (diag not supported for depos_order >= 5) or a medium priority warning message (diag does not support depos_order >= 5, 4 is used instead).

Source/Diagnostics/ParticleDiag/ParticleDiag.cpp
```diff
@@ -42,10 +42,16 @@ ParticleDiag::ParticleDiag (
...
 #endif
                 if (var == "phi") {
-                    // User requests phi on particle. This is *not* part of the variables that
-                    // the particle container carries, and is only added to particles during output.
-                    // Therefore, this case needs to be treated specifically.
                     m_plot_phi = true;
+                } else if (var == "Ex" || var == "Ey" || var == "Ez" ||
+                           var == "Bx" || var == "By" || var == "Bz") {
+                    m_plot_EM = true;
+                    if (var == "Ex") m_plot_EM_flags[0] = true;
```
lucafedeli88: I think that in WarpX we prefer adding parentheses after if clauses:

```
if (something) {stuff}
```
even for one-liners.

An alternative would be to write 
`m_plot_EM_flags[0] = (var == "Ex");`

Source/Diagnostics/ParticleDiag/ParticleDiag.H
```diff
@@ -24,6 +24,8 @@ public:
     [[nodiscard]] std::string getSpeciesName() const { return m_name; }
     amrex::Vector<int> m_plot_flags;
     bool m_plot_phi = false; // Whether to output the potential phi on the particles
+    bool m_plot_EM = false; // Whether to output the E and B fields on the particles
+    bool m_plot_EM_flags[6] = {false, false, false, false, false, false}; // E and B fields
```
lucafedeli88: I would suggest using an amrex array of 6 bool here.

Examples/Tests/EM_fields_on_particles/inputs_test_2d_particle_EM_diagnostics
```diff
@@ -0,0 +1,84 @@
...
+diagnostics.diags_names = diag1 diag_checksum
+
+diag1.intervals = 1
+diag1.diag_type = Full
+diag1.format = openpmd
+diag1.openpmd_backend = h5
+diag1.fields_to_plot = none
+diag1.write_species = 1
+diag1.species = electron
+diag1.electron.variables = Ex Ey Ez Bx By Bz # will emit a warning because no position
```
lucafedeli88:
```suggestion
diag1.electron.variables = Ex Ey Ez Bx By Bz # will emit a warning because no position is asked
```
I would suggest something like this.

Examples/Tests/EM_fields_on_particles/analysis_2d.py
```diff
@@ -0,0 +1,158 @@
...
+ex_t = np.squeeze(ex_t)
+ey_t = np.squeeze(ey_t)
+ez_t = np.squeeze(ez_t)
+bx_t = np.squeeze(bx_t)
+by_t = np.squeeze(by_t)
+bz_t = np.squeeze(bz_t)
+
+n_diags = 401
+
+DT = 6.324524234e-17  # @ CFL = 0.99
```
lucafedeli88, line 33: If you want I think that this can be obtained by reading OpenPMD output.

Examples/Tests/EM_fields_on_particles/analysis_2d.py
```diff
@@ -0,0 +1,158 @@
...
+iterations = np.linspace(0, n_diags - 1, n_diags)
+T = (iterations * DT) - DT / 2
```
lucafedeli88, line 35: Also this, in principle, could be read from OpenPMD output, I think.

Examples/Tests/EM_fields_on_particles/analysis_2d.py
```diff
@@ -0,0 +1,158 @@
...
+interp_func_ez = interp1d(T, Ez_th, kind="cubic", fill_value="extrapolate")
+interp_func_by = interp1d(T, By_th, kind="cubic", fill_value="extrapolate")
+interp_func_bz = interp1d(T, Bz_th, kind="cubic", fill_value="extrapolate")
+
+Ey_aligned = interp_func_ey(T + opt_shift)
+Ez_aligned = interp_func_ez(T + opt_shift)
+By_aligned = interp_func_by(T + opt_shift)
+Bz_aligned = interp_func_bz(T + opt_shift)
+
+# verif
```
lucafedeli88: I would suggest a comment such as:
"test cases"

Examples/Tests/EM_fields_on_particles/analysis_2d.py
```diff
@@ -0,0 +1,158 @@
...
+
+ey_p = ey_t[T >= delta_t]
+Ey_p = Ey_aligned[T >= delta_t]
+ey_p += 1e11
+Ey_p += 1e11
+
+ez_p = ez_t[T >= delta_t]
+Ez_p = Ez_aligned[T >= delta_t]
+ez_p += 1e11
+Ez_p += 1e11
```
lucafedeli88, line 112: I think that adding a comment on why you add `1e11` here would be useful.

Examples/Tests/EM_fields_on_particles/analysis_2d.py
```diff
@@ -0,0 +1,158 @@
...
+
+by_p = by_t[T >= delta_t]
+By_p = By_aligned[T >= delta_t]
+by_p += 1e3
+By_p += 1e3
+
+bz_p = bz_t[T >= delta_t]
+Bz_p = Bz_aligned[T >= delta_t]
+bz_p += 1e3
+Bz_p += 1e3
```
lucafedeli88, line 148: I think that adding a comment on why you add 1e3 here could be useful.

Docs/source/usage/parameters.rst
```diff
@@ -2966,11 +2966,14 @@ In-situ capabilities can be used by turning on Sensei or Ascent (provided they a
...
     Choices are ``x``, ``y``, ``z`` for the particle positions (3D and RZ), ``x`` & ``z`` in 2D, ``z`` in 1D,
-    ``w`` for the particle weight and ``ux``, ``uy``, ``uz`` for the particle momenta.
+    ``w`` for the particle weight, ``ux``, ``uy``, ``uz`` for the particle momenta, and ``Ex``, ``Ey``, ``Ez``, ``Bx``, ``By``, ``Bz`` for the electromagnetic fields.
     When using the lab-frame electrostatic solver, ``phi`` (electrostatic potential, on the macroparticles) is also available.
-    By default, all particle quantities (except ``phi``) are written.
+    By default, positions, momenta and particle weight are written.
     If ``<diag_name>.<species_name>.variables = none``, no particle data are written.
 
+    .. note::
+        The electromagnetic fields diagnostics will not display the external fields.
```
lucafedeli88: Maybe you could write something like "will not include" instead of "will not display"

lucafedeli88 (changes requested):
Thanks a lot for this PR, @grobertdautun ! It implements a very useful feature.
I've left a few comments.

Source/Particles/ParticleIO.H
grobertdautun: changing this raises an "invalid `const bool*` to `bool*` conversion"

Source/Diagnostics/ParticleIO.cpp
grobertdautun: This one is weird, it is supposed to be `constexpr`, however when I add it it raises the CUDA error about first capturing variables in constexpr, however it does not raise it when I first capture them with `doDirectGatherVectorField`, which is in a `constexpr`.....

I will just remove the `if` block and `ignore_unused` for every kernel

Source/Diagnostics/ParticleDiag/ParticleDiag.cpp
grobertdautun: I don't think the alternative would work (unless I'm mistaken, either it would overwrite every flag to false except the last or I will need to check `var` to know which flag; which would make the `var =="Ex"` redundant)

Source/Diagnostics/ParticleIO.cpp
```diff
@@ -299,3 +301,218 @@
...
+                {
+                amrex::ParticleReal xp, yp, zp;
+                getPosition(ip, xp, yp, zp);
+
+                [[maybe_unused]] amrex::ParticleReal Ex_particle = 0._rt;
+                [[maybe_unused]] amrex::ParticleReal Ey_particle = 0._rt;
+                [[maybe_unused]] amrex::ParticleReal Ez_particle = 0._rt;
+                [[maybe_unused]] amrex::ParticleReal Bx_particle = 0._rt;
+                [[maybe_unused]] amrex::ParticleReal By_particle = 0._rt;
+                [[maybe_unused]] amrex::ParticleReal Bz_particle = 0._rt;
```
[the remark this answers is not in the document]
lucafedeli88: I will dismiss this because the quantity ` Bz_particle` is marked as `[[maybe_unused]]` .

Source/Diagnostics/ParticleIO.cpp
grobertdautun: is it only the diag or globally in warpx that the order < 5 ?

Source/Diagnostics/ParticleIO.cpp
grobertdautun: from what I see `finestLevel` is not available in `WarpXOpenPMD.cpp` where the function is called

Examples/Tests/EM_fields_on_particles/analysis_2d.py
grobertdautun: I put it like this to completely decouple warpx and the theoretical result

EZoni:
Thank you, @grobertdautun!

We just merged in `development` a big updated of the physical constants as well as an update of the Ubuntu version used in our CI Azure pipelines, see #5661.

Could you please merge `development` in this PR's head branch `diags_em_on_particles_2`, to reflect that change and possibly update any CI test's checksums?

Regression/Checksum/benchmarks_json/test_2d_particle_EM_diagnostics.json
```diff
@@ -0,0 +1,25 @@
...
+  "lev=0": {
+    "Bx": 3864.6903326868182,
+    "By": 239757.36697092294,
+    "Bz": 119878.42982987872,
+    "Ex": 2396689176285.2944,
+    "Ey": 35934018061853.67,
+    "Ez": 71884648325011.86
+  },
+  "electron": {
+    "particle_bx": 3.8909803033142373e-07,
```
lucafedeli88:
```suggestion
    "particle_bx": 3.890980338189904e-07,
```

I would suggest correcting this with what is actually obtained by the github runner (I get the same on my machine, by the way).

aeriforme:
Is anybody planning to work on this PR? 
If not, I can help. Lmk

grobertdautun:
Hi @aeriforme , thanks for offering to work on this PR. I updated it and it is now ready for review. 
2 of the CI tests are failing (PETSc GCC) due to an uninitialised variable, which I fixed in #6254 as it seems unrelated to this PR

EZoni:
@dpgrote 

Can you remind me if this PR was effectively replaced by your #6570 (already merged)?

dpgrote:
> @dpgrote
> 
> Can you remind me if this PR was effectively replaced by your #6570 (already merged)?

Hi @EZoni - Yes, my PR implements the same capability. When I submitted mine, I had not known about this PR.

EZoni:
Thanks, @dpgrote. I'll close this PR then. @grobertdautun, please feel free to reopen this PR later, or open a new one, if you have additional features ready to add.

PR #5638, opened 2025-02-04
Fixing bug in hyper-resistivity calculation which had missing terms i…

clarkse:
…n vector laplacian evaluation. Additioanally fixing a staggering bug for density calculation in RZ.

### DIFF
--- a/Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
+++ b/Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
@@ -586,8 +586,9 @@ void FiniteDifferenceSolver::HybridPICSolveECylindrical (
 
                 // Interpolate current to appropriate staggering to match E field
                 Real jtot_val = 0._rt;
+                Real jr_val = 0._rt;
                 if (solve_for_Faraday && resistivity_has_J_dependence) {
-                    const Real jr_val = Interp(Jr, Jr_stag, Er_stag, coarsen, i, j, 0, 0);
+                    jr_val = Interp(Jr, Jr_stag, Er_stag, coarsen, i, j, 0, 0);
                     const Real jt_val = Interp(Jt, Jt_stag, Er_stag, coarsen, i, j, 0, 0);
                     const Real jz_val = Interp(Jz, Jz_stag, Er_stag, coarsen, i, j, 0, 0);
                     jtot_val = std::sqrt(jr_val*jr_val + jt_val*jt_val + jz_val*jz_val);
@@ -613,7 +614,8 @@ void FiniteDifferenceSolver::HybridPICSolveECylindrical (
                     // r on cell-centered point (Jr is cell-centered in r)
                     Real const r = rmin + (i + 0.5_rt)*dr;
 
-                    auto nabla2Jr = T_Algo::Dr_rDr_over_r(Jr, r, dr, coefs_r, n_coefs_r, i, j, 0, 0);
+                    auto nabla2Jr = T_Algo::Dr_rDr_over_r(Jr, r, dr, coefs_r, n_coefs_r, i, j, 0, 0)
+                        + T_Algo::Dzz(Jr, coefs_z, n_coefs_z, i, j, 0, 0) - jr_val/(r*r);
                     Er(i, j, 0) -= eta_h * nabla2Jr;
                 }
             },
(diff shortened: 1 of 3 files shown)


Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -586,8 +586,9 @@ void FiniteDifferenceSolver::HybridPICSolveECylindrical (
 
                 // Interpolate current to appropriate staggering to match E field
                 Real jtot_val = 0._rt;
+                Real jr_val = 0._rt;
                 if (solve_for_Faraday && resistivity_has_J_dependence) {
-                    const Real jr_val = Interp(Jr, Jr_stag, Er_stag, coarsen, i, j, 0, 0);
+                    jr_val = Interp(Jr, Jr_stag, Er_stag, coarsen, i, j, 0, 0);
```
roelof-groenewald: This will only be evaluated if `resistivity_has_J_dependence == true`, but it needs to be evaluated `if (solve_for_Faraday && include_hyper_resistivity_term)`, and the same for the other components.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -633,13 +635,14 @@ void FiniteDifferenceSolver::HybridPICSolveECylindrical (
                 }
 
                 // Interpolate to get the appropriate charge density in space
-                Real rho_val = Interp(rho, nodal, Er_stag, coarsen, i, j, 0, 0);
+                Real rho_val = Interp(rho, nodal, Et_stag, coarsen, i, j, 0, 0);
```
roelof-groenewald, line 637: Nice catch!

roelof-groenewald (commented):
~~Hmm, the hyper-resistivity term is a **vector** Laplacian, no? I agree there are bugs here, but it looks like you changed it to a sort-of hybrid scalar / vector Laplacian. Or am I missing something?~~

~~For RZ the individual components should be as follows:~~
image

~~But the Cartesian implementation was already correct.~~

Update: Nevermind. I see this is in fact all wrong. Thanks for catching!

roelof-groenewald:
The fixes look good! Just the one comment above about the `resistivity_has_J_dependence` issue and this should be good to go!

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
clarkse: Thanks for catching this.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -697,7 +704,15 @@ void FiniteDifferenceSolver::HybridPICSolveECylindrical (
...
                 if (include_hyper_resistivity_term) {
-                    auto nabla2Jz = T_Algo::Dzz(Jz, coefs_z, n_coefs_z, i, j, 0, 0);
+                    // r on nodal point (Jz is nodal in r)
+                    Real const r = rmin + i*dr;
+
+                    auto nabla2Jz = 0._rt;
+                    if (r > 0.5_rt*dr) {
+                        nabla2Jz = T_Algo::Dr_rDr_over_r(Jz, r, dr, coefs_r, n_coefs_r, i, j, 0, 0)
+                            + T_Algo::Dzz(Jz, coefs_z, n_coefs_z, i, j, 0, 0);
+                    }
```
roelof-groenewald: Do you think the way it is, should be kept, or would this be better as:
```suggestion
                    auto nabla2Jz = T_Algo::Dzz(Jz, coefs_z, n_coefs_z, i, j, 0, 0);
                    if (r > 0.5_rt*dr) {
                        nabla2Jz += T_Algo::Dr_rDr_over_r(Jz, r, dr, coefs_r, n_coefs_r, i, j, 0, 0)
                    }
```

Thereby, essentially just making $r\partial / \partial r \rightarrow 0$ at r = 0.

clarkse: Earlier in this loop there is an exit point if r == 0, so this should never be reached if at origin. Et needs to stay zero at the origin, I removed the check though since it was redundant.

roelof-groenewald (approved):
Thanks so much for fixing!

PR #5641, opened 2025-02-05
WarpX class : current_centering_* variables no longer static

lucafedeli88:
This PR turns the static variables of the WarpX class `current_centering_nox`, `current_centering_noy`, and `current_centering_noz` into private member variables: `m_current_centering_nox`, `m_current_centering_noy`, and `m_current_centering_noz`.


Python/pywarpx/picmi.py
```diff
@@ -3043,10 +3043,10 @@ def initialize_inputs(self):
             pywarpx.warpx.field_centering_noz = self.field_centering_order[-1]
         # warpx.current_centering_no<x,y,z>
         if self.current_centering_order is not None:
-            pywarpx.warpx.current_centering_nox = self.current_centering_order[0]
+            pywarpx.warpx.m_current_centering_nox = self.current_centering_order[0]
```
lucafedeli88: I am not too familiar with `picmi`: is it OK to make this change here?

EZoni: Good question. I'm not 100% sure.

I think these might need to match the names of the input parameters exposed to the users, i.e., the names set here without `m_`:
https://github.com/lucafedeli88/WarpX/blob/ce93182632c5cd43049e1111f3cd8976642e7aa2/Source/WarpX.cpp#L1073-L1078
```cpp
                pp_warpx, "current_centering_nox", m_current_centering_nox);
                pp_warpx, "current_centering_noy", m_current_centering_noy);
                pp_warpx, "current_centering_noz", m_current_centering_noz);
```

@dpgrote 
Can you confirm? I might be wrong, I'm not 100% sure.

EZoni: I think this is the case, and I think it is kind of confirmed by the fact that
```
git grep "pywarpx\.warpx\.m_"
```
returns no results. I will suggest the change in the code review.

EZoni:
```suggestion
            pywarpx.warpx.current_centering_nox = self.current_centering_order[0]
```

Python/pywarpx/picmi.py
```diff
@@ -3043,10 +3043,10 @@ def initialize_inputs(self):
...
             if self.solver.grid.number_of_dimensions == 3:
-                pywarpx.warpx.current_centering_noy = self.current_centering_order[1]
-            pywarpx.warpx.current_centering_noz = self.current_centering_order[-1]
+                pywarpx.warpx.m_current_centering_noy = self.current_centering_order[1]
+            pywarpx.warpx.m_current_centering_noz = self.current_centering_order[-1]
```
EZoni:
```suggestion
                pywarpx.warpx.current_centering_noy = self.current_centering_order[1]
            pywarpx.warpx.current_centering_noz = self.current_centering_order[-1]
```

lucafedeli88: I agree with you!

Python/pywarpx/picmi.py
lucafedeli88: I agree with you!

Python/pywarpx/picmi.py
```diff
@@ -3033,7 +3033,7 @@ def initialize_inputs(self):
 
         self.solver.solver_initialize_inputs()
 
-        # Initialize warpx.field_centering_no<x,y,z> and warpx.current_centering_no<x,y,z>
+        # Initialize warpx.field_centering_no<x,y,z> and warpx.m_current_centering_no<x,y,z>
```
lucafedeli88:
```suggestion
        # Initialize warpx.field_centering_no<x,y,z> and warpx.current_centering_no<x,y,z>
```

I think that, since we are in picmi, reverting this change would be better for the user.

lucafedeli88:
@EZoni , I accepted all your suggestions.

lucafedeli88:
@EZoni , I've fixed the merge conflict

EZoni (approved):

PR #5642, opened 2025-02-05
WarpX class : em_solver_medium no longer a static variable

lucafedeli88:
This PR turns the static variable `em_solver_medium` of the WarpX class into a private non-static member variable : `m_em_solver_medium` . This is done with the aim of reducing the usage of static variables in WarpX.


ax3l (approved):
:+1:

PR #5644, opened 2025-02-06
WarpX class: move SetDotMask to anonymous namespace in WarpX.cpp

lucafedeli88:
`SetDotMask`, a member function of the WarpX class, is only used inside the member function `getFieldDotMaskPointer` . This PR turns it into a pure function and moves it into an anonymous namespace inside `WarpX.cpp`.
This (slightly) simplifies the WarpX class header.


JustinRayAngus:
These changes look fine to me.

Thanks!

JustinRayAngus (approved):

Issue #4913, opened 2024-05-06
Add Yourself to Our Community Map! :sparkles: 


ax3l:
Hi everyone,

We are starting for various upcoming reviews a map of **BLAST users, science applications of our codes and contributors** ([BLAST code list](https://blast.lbl.gov/codes/), which includes **WarpX**).

The map is open and contributable, exactly like the BLAST codes. If you are or did use BLAST codes in your research, studies, benchmarks, ..., please consider [adding yourself to our map](https://github.com/BLAST-codes/blast-communitymap/issues/new).

Thank you for your support! :pray:

* [Repository](https://github.com/BLAST-codes/blast-communitymap?tab=readme-ov-file#contribute-to-the-map)
* [Current Map Data](https://github.com/BLAST-codes/blast-communitymap/blob/gh-pages/data/blast-community.geojson)

> Since BLAST codes are available for everyone, without registration or strings attached, **we likely will not know BLAST is useful for you unless you reach out to us**. BLAST development primarily relies on public funding and reviews conducted by peers and funding agencies. **We rely on your feedback to show the community, diversity and impact of BLAST as an open science ecosystem for beam, plasma and accelerator modeling.**

CC @kli-jfp @Aquios7 @jinze-liu @1120805582 @zhazhajust @tanweihou @Poto45 @ddghosal @Kellysherry @ercoop @moisesa2 @Change72 @wnguyen1312 @berceanu @Byeerendon @sheroytata @zenineasa @GivenChen @yuanjiang-lance @abigailalexjames @spadova-a @floresv299 @TomPaolo @aodhanmci @deSUBHAM @electroflow @wphu @Sbozzolo @bzdjordje @HollyHuddle  @erny123  @ZippoFight  @griffinfarrow  @iamcalvinlau  @OscarJakobssonLPA @OscarJakobssonResearch @rl3418 @1661145410 @chiehjen @ozgurculfa @rupenether @Aleeeeeen @chipopt @archermarx @bstassel @archermarx @denisbertini @DrakeWhu @ycaophysics @Hnair13 @lijiangdong1999

ax3l:
[...]

PR #5646, opened 2025-02-07
Doc: New APL on Magnetic Reconnection

ax3l:
**Magnetic Reconnection: An Alternative Explanation of Radio Emission in Galaxy Clusters** by Subham Ghosh and Pallavi Bhat was just published. https://10.3847/2041-8213/ad9f2d

### DIFF
--- a/Docs/source/highlights.rst
+++ b/Docs/source/highlights.rst
@@ -154,6 +154,11 @@ High Energy Astrophysical Plasma Physics
 
 Scientific works in astrophysical plasma modeling.
 
+#. Ghosh S, Bhat P.
+   **Magnetic Reconnection: An Alternative Explanation of Radio Emission in Galaxy Clusters**.
+   The Astrophysical Journal **979** 1, 2025.
+   `DOI:10.3847/2041-8213/ad9f2d <https://doi.org/10.3847/2041-8213/ad9f2d>`__
+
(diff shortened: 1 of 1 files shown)


ax3l:
@deSUBHAM do you like to be added to our Community Map? ✨  #4913

Docs/source/highlights.rst
```diff
@@ -154,6 +154,11 @@ High Energy Astrophysical Plasma Physics
 
 Scientific works in astrophysical plasma modeling.
 
+#. Ghosh S, Bhat P.
+   **Magnetic Reconnection: An Alternative Explanation of Radio Emission in Galaxy Clusters**.
+   The Astrophysical Journal **979** 1, 2025.
```
EZoni: Thanks for adding this! I think the full name of the journal is The Astrophysical Journal Letters:
```suggestion
   The Astrophysical Journal Letters **979** 1, 2025.
```

ax3l: True, maybe we should fix the entry below, too.

ax3l: Ah no, they are indeed different.

deSUBHAM:
Surely, I would like to be added to the Community Map.

Thank you.
With regards,
Subham

EZoni (approved):

PR #5648, opened 2025-02-08
Perlmutter: SW Install Updates

ax3l:
- [x] profile: avoid repetition, use `SW_DIR` variable as in `install_...` scripts
- [x] move from CFS to PSCRATCH (more stable, faster, where the binary lives); uses an undocumented, purge-exempt location for container images/software
- [x] build our own boost (SW stack consistency)
- [x] get our own CCache (prior one is gone)
- [x] RT tested

### DIFF
--- a/Tools/machines/perlmutter-nersc/install_cpu_dependencies.sh
+++ b/Tools/machines/perlmutter-nersc/install_cpu_dependencies.sh
@@ -31,7 +31,7 @@ fi
 
 # Remove old dependencies #####################################################
 #
-SW_DIR="${PSCRATCH}/storage/sw/perlmutter/cpu"
+SW_DIR="${PSCRATCH}/storage/sw/warpx/perlmutter/cpu"
 rm -rf ${SW_DIR}
 mkdir -p ${SW_DIR}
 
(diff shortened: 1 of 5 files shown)


Tools/machines/perlmutter-nersc/install_cpu_dependencies.sh
```diff
@@ -31,7 +31,7 @@ fi
 
 # Remove old dependencies #####################################################
 #
-SW_DIR="${CFS}/${proj}/${USER}/sw/perlmutter/cpu"
+SW_DIR="${PSCRATCH}/storage/sw/perlmutter/cpu"
```
n01r: There are two things that come to mind:
* Should we add `warpx` to the path because we may want to do this for ImpactX and other BLAST codes as well and would accidentally remove the dependencies of the other codes when rerunning the install dependencies script?
* Should we add `${proj}` because different projects could require different dependencies? (Maybe this one is not necessary because a developer would know to adjust the install script or run with a temporary solution)


Could look like this `SW_DIR="${PSCRATCH}/storage/sw/perlmutter/${proj}/warpx/cpu/"`?

n01r (approved):
Looks good :)

Since PSCRATCH does not distinguish between projects, I would put the project ID into the path for the `SW_DIR` so one can switch between projects/BLAST codes and update dependencies without deleting. But that could also be a follow-up.

Tools/machines/perlmutter-nersc/install_cpu_dependencies.sh
ax3l: Good point, using `.../sw/warpx/perlmutter/...` now :+1:

ax3l:
@n01r @EZoni ready now :)

n01r (approved):
Thanks a lot! :)

n01r (approved):

EZoni (approved):

PR #5649, opened 2025-02-08
Cropping of particles at boundaries for deposition for charge conservation

dpgrote:
This modifies the deposition (and field gather), cropping particle orbits at the boundary of the domain for PEC and insulator boundaries. This is a simple way of achieving charge conservation as particles are absorbed at boundaries.

In input flag `particles.crop_on_PEC_boundary` was added to turn this feature on. It now defaults to `false`.

Only implemented in Villasenor and Buneman and Esirkepov current deposition.


JustinRayAngus:
The cropping should not be applied at PMC (symmetry) boundaries. Only at physical boundaries like conductors and insulators.

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -477,6 +477,21 @@ WarpXParticleContainer::DepositCurrent (WarpXParIter& pti,
...
+
+    amrex::GpuArray<amrex::GpuArray<bool,2>, AMREX_SPACEDIM> is_absorbing;
+    for (int idim=0; idim < AMREX_SPACEDIM; ++idim) {
+        is_absorbing[idim][0] = (tilebox.smallEnd(idim) <= domain_box.smallEnd(idim) &&
+                                 (field_boundary_lo[idim] == FieldBoundaryType::PEC
+                               || field_boundary_lo[idim] == FieldBoundaryType::PMC));
+        is_absorbing[idim][1] = (tilebox.bigEnd(idim) >= domain_box.bigEnd(idim) &&
+                                 (field_boundary_hi[idim] == FieldBoundaryType::PEC
+                               || field_boundary_hi[idim] == FieldBoundaryType::PMC));
+    }
```
JustinRayAngus: PMC ==> PECInsulator

Source/Utils/ParticleUtils.H
```diff
@@ -189,6 +189,19 @@ namespace ParticleUtils {
...
+     * \param[in] xmin The lower boundary of the domain
+     * \param[in] xmax The upper boundary of the domain
+     * \param[in] check Whether to crop at each of the boundaries
+     */
+    AMREX_GPU_HOST_DEVICE AMREX_INLINE
+    amrex::Real crop_at_boundary(amrex::Real x, amrex::Real xmin, amrex::Real xmax, amrex::GpuArray<bool,2> const& check) {
+        if (check[0] && x < xmin) { return xmin; }
+        if (check[1] && x > xmax) { return xmax; }
+        return x;
+    }
```
JustinRayAngus, line 312: I don't think this is sufficient for D>1. You need to crop the full position vector where the particle intersects the boundary. See the truncate_boundaries() routine in picnic/src/particle_tools/MeshInterpChargeConservingF.ChF

Source/Particles/Gather/FieldGather.H
```diff
@@ -928,26 +931,30 @@ void doGatherPicnicShapeN (
...
-    double const y_bar = (yp_nph - xyzmin.y)*dinv.y;
+    double const y_bar = (y_new + y_old)*0.5_rt;
 #endif
+    amrex::Real const zp_np1_c = ParticleUtils::crop_at_boundary(zp_np1, xyzmin.z, xyzmax.z, is_absorbing[WARPX_ZINDEX]);
     // Keep these double to avoid bug in single precision
-    double const z_new = (zp_np1 - xyzmin.z)*dinv.z;
+    double const z_new = (zp_np1_c - xyzmin.z)*dinv.z;
     double const z_old = (zp_n   - xyzmin.z)*dinv.z;
-    double const z_bar = (zp_nph - xyzmin.z)*dinv.z;
+    double const z_bar = (z_new + z_old)*0.5_rt;
```
JustinRayAngus: As implemented, this will not achieved the desired result. We can talk offline about this.

Source/Utils/ParticleUtils.H
```diff
@@ -245,6 +245,71 @@ namespace ParticleUtils {
...
+     * \param[in] xmin The lower boundary of the domain in grid units
+     * \param[in] xmax The upper boundary of the domain in grid units
+     * \param[in] check Whether to crop at each of the boundaries
+     */
+    AMREX_GPU_HOST_DEVICE AMREX_INLINE
+    void crop_at_boundary(double x0, double z0,
+                          double & x1, double & z1,
+                          int xmin, int xmax, amrex::GpuArray<bool,2> const& check) {
+        if (check[0] && x1 < xmin) {
+            double const fx = ((x0 - x1) != 0. ? (x0 - xmin)/(x0 - x1) : 0.);
```
JustinRayAngus, line 278: This is a minor comment, but I think it may look cleaner to write `(xmin - x0)/(x1 - x0)` here so that this line mirrors line 283 with `xbdry = xmin` and `xbdry = xmax`

Source/Utils/ParticleUtils.H
```diff
@@ -245,6 +245,71 @@ namespace ParticleUtils {
...
+     * \param[in] xmin The lower boundary of the domain in grid units
+     * \param[in] xmax The upper boundary of the domain in grid units
+     * \param[in] check Whether to crop at each of the boundaries
+     */
+    AMREX_GPU_HOST_DEVICE AMREX_INLINE
+    void crop_at_boundary(double x0, double y0, double z0,
+                          double & x1, double & y1, double & z1,
+                          int xmin, int xmax, amrex::GpuArray<bool,2> const& check) {
+        if (check[0] && x1 < xmin) {
+            double const fx = ((x0 - x1) != 0. ? (x0 - xmin)/(x0 - x1) : 0.);
```
JustinRayAngus, line 301: Same here. `(xmin - x0)/(x1 - x0)`

Source/Utils/ParticleUtils.H
```diff
@@ -245,6 +245,71 @@ namespace ParticleUtils {
...
+     * \param[in] x0, z0 The initial particle position
+     * \param[in/out] x1, z1 The final particle position, with the returned value being cropped
+     * \param[in] xmin The lower boundary of the domain in grid units
+     * \param[in] xmax The upper boundary of the domain in grid units
+     * \param[in] check Whether to crop at each of the boundaries
+     */
+    AMREX_GPU_HOST_DEVICE AMREX_INLINE
+    void crop_at_boundary(double x0, double z0,
+                          double & x1, double & z1,
+                          int xmin, int xmax, amrex::GpuArray<bool,2> const& check) {
```
JustinRayAngus: I think using `int xmin` and `int xmax` may cause a performance issue on GPUs since you have to cast them to double below. Can we change this to `double xmin` and `double xmax` here? I'll add additional comments about this where this routine is called.

Source/Utils/ParticleUtils.H
```diff
@@ -245,6 +245,71 @@ namespace ParticleUtils {
...
+    /* \brief Crops the position at the specified boundary if check is true, 1D version
+     * \param[in] x0 The initial particle position
+     * \param[in/out] x1 The final particle position, with the returned value being cropped
+     * \param[in] xmin The lower boundary of the domain in grid units
+     * \param[in] xmax The upper boundary of the domain in grid units
+     * \param[in] check Whether to crop at each of the boundaries
+     */
+    AMREX_GPU_HOST_DEVICE AMREX_INLINE
+    void crop_at_boundary(double & x1,
+                          int xmin, int xmax, amrex::GpuArray<bool,2> const& check) {
```
JustinRayAngus: Same comment as for 2D. Can we change `int xmin` and `int xmax` to `double xmin` and `double xmax`?

Source/Utils/ParticleUtils.H
```diff
@@ -245,6 +245,71 @@ namespace ParticleUtils {
...
+     * \param[in] x0, y0, z0 The initial particle position
+     * \param[in/out] x1, y1, z1 The final particle position, with the returned value being cropped
+     * \param[in] xmin The lower boundary of the domain in grid units
+     * \param[in] xmax The upper boundary of the domain in grid units
+     * \param[in] check Whether to crop at each of the boundaries
+     */
+    AMREX_GPU_HOST_DEVICE AMREX_INLINE
+    void crop_at_boundary(double x0, double y0, double z0,
+                          double & x1, double & y1, double & z1,
+                          int xmin, int xmax, amrex::GpuArray<bool,2> const& check) {
```
JustinRayAngus: Same comment as for 2D. Can we change `int xmin` and `int xmax` to `double xmin` and `double xmax`?

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -635,31 +657,31 @@ WarpXParticleContainer::DepositCurrent (WarpXParIter& pti,
...
                 } else if (WarpX::nox == 4){
                     doEsirkepovDepositionShapeN<4>(
                         GetPosition, wp.dataPtr() + offset, uxp.dataPtr() + offset,
                         uyp.dataPtr() + offset, uzp.dataPtr() + offset, ion_lev,
                         jx_arr, jy_arr, jz_arr,
-                        np_to_deposit, dt, relative_time, dinv, xyzmin, lo, q,
+                        np_to_deposit, dt, relative_time, dinv, xyzmin, domain_box, is_absorbing, lo, q,
                         WarpX::n_rz_azimuthal_modes,
                         eb_reduce_particle_shape, EB::enabled() );
                 }
```
JustinRayAngus: I don't think we want to do this (at least not for this PR) inside the standard explicit Esirkepov deposit routines.

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -514,6 +516,26 @@ WarpXParticleContainer::DepositCurrent (WarpXParIter& pti,
...
+        is_absorbing[idim][0] = m_crop_on_PEC_boundary &&
+                                (tilebox.smallEnd(idim) <= domain_box.smallEnd(idim) &&
+                                 (field_boundary_lo[idim] == FieldBoundaryType::PEC
+                               || field_boundary_lo[idim] == FieldBoundaryType::PECInsulator));
+        is_absorbing[idim][1] = m_crop_on_PEC_boundary &&
+                                (tilebox.bigEnd(idim) >= domain_box.bigEnd(idim) &&
+                                 (field_boundary_hi[idim] == FieldBoundaryType::PEC
+                               || field_boundary_hi[idim] == FieldBoundaryType::PECInsulator));
+    }
+
```
JustinRayAngus, line 544: This comment is related to the comments in ParticleUtils.H about changing some variables from `int` to `double`. I think this is a good place to directly compute the domain boundaries in grid units like `xmin` and `xmax`. Just after these lines, we could have something like

```
    amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> domain_lo, domain_hi;
    domain_lo[0] = static_cast<amrex::Real>(domain_box.smallEnd(0) - lo.x);
    domain_hi[0] = static_cast<amrex::Real>(domain_box.bigEnd(0) - lo.x);
#if (AMREX_SPACEDIM >= 2)
    domain_lo[1] = static_cast<amrex::Real>(domain_box.smallEnd(1) - lo.y);
    domain_hi[1] = static_cast<amrex::Real>(domain_box.bigEnd(1) - lo.y);
#endif
#if (AMREX_SPACEDIM == 3)
    domain_lo[1] = static_cast<amrex::Real>(domain_box.smallEnd(2) - lo.z);
    domain_hi[1] = static_cast<amrex::Real>(domain_box.bigEnd(2) - lo.z);
#endif
```

Then instead of passing `domain_box` below to the deposit routines we pass `domain_lo` and `domain_hi`. Then these values are only computed once per box (instead of once per particle) and we can avoid the need to cast these quantites from `int` to `double` inside the particle loop

Source/Utils/ParticleUtils.H
dpgrote: I wrote it this way intentionally, with the idea that the expressions in the parentheses will always be positive. I find this more intuitive.

Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -1106,6 +1107,8 @@ void doEsirkepovDepositionShapeN (const GetParticlePosition<PIdx>& GetPosition,
  * \param dt           Time step for particle level
  * \param dinv         3D cell size inverse
  * \param xyzmin       Physical lower bounds of domain.
+ * \param domain_double The domain boundary indicies cast to double values
```
JustinRayAngus, line 1110: Looks like the name here (`domain_double`) doesn't match the name used in the function (`domain`).

I also noticed that sometimes you between these two names in different places. Maybe just pick one and use that name everywhere?

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -436,9 +439,31 @@ PhysicalParticleContainer::ImplicitPushXP (WarpXParIter & pti,
...
+                               || field_boundary_lo[idim] == FieldBoundaryType::PECInsulator));
+        is_absorbing[idim][1] = m_crop_on_PEC_boundary &&
+                                (box.bigEnd(idim) >= domain_box.bigEnd(idim) &&
+                                 (field_boundary_hi[idim] == FieldBoundaryType::PEC
+                               || field_boundary_hi[idim] == FieldBoundaryType::PECInsulator));
+
+        domain_double[idim][0] = static_cast<double>(domain_box.smallEnd(idim) - box.smallEnd(idim));
+        domain_double[idim][1] = static_cast<double>(domain_box.bigEnd(idim) - box.smallEnd(idim));
+    }
+
```
JustinRayAngus, line 466: Do you think we should wrap this block of code, which appears in three places now, into a new function?

Source/Utils/ParticleUtils.H
```diff
@@ -245,6 +245,71 @@ namespace ParticleUtils {
...
+    }
+
+    /* \brief Crops the position at the specified boundary if do_cropping is true, 2D version
+     * \param[in] x0, z0 The initial particle position
+     * \param[in/out] x1, z1 The final particle position, with the returned value being cropped
+     * \param[in] xmin The lower boundary of the domain in grid units
+     * \param[in] xmax The upper boundary of the domain in grid units
+     * \param[in] do_cropping Whether to crop at each of the boundaries
+     */
+    AMREX_GPU_HOST_DEVICE AMREX_INLINE
```
JustinRayAngus:
```suggestion
    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
```

JustinRayAngus: Seems like we should force inlining since this is called in a hot loop over particles

Source/Utils/ParticleUtils.H
```diff
@@ -245,6 +245,71 @@ namespace ParticleUtils {
...
+    }
+
+    /* \brief Crops the position at the specified boundary if do_cropping is true, 3D version
+     * \param[in] x0, y0, z0 The initial particle position
+     * \param[in/out] x1, y1, z1 The final particle position, with the returned value being cropped
+     * \param[in] xmin The lower boundary of the domain in grid units
+     * \param[in] xmax The upper boundary of the domain in grid units
+     * \param[in] do_cropping Whether to crop at each of the boundaries
+     */
+    AMREX_GPU_HOST_DEVICE AMREX_INLINE
```
JustinRayAngus:
```suggestion
    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
```

Source/Utils/ParticleUtils.H
```diff
@@ -245,6 +245,71 @@ namespace ParticleUtils {
...
     }
 
+    /* \brief Crops the position at the specified boundary if do_cropping is true, 1D version
+     * \param[in] x0 The initial particle position
+     * \param[in/out] x1 The final particle position, with the returned value being cropped
+     * \param[in] xmin The lower boundary of the domain in grid units
+     * \param[in] xmax The upper boundary of the domain in grid units
+     * \param[in] do_cropping Whether to crop at each of the boundaries
+     */
+    AMREX_GPU_HOST_DEVICE AMREX_INLINE
```
JustinRayAngus:
```suggestion
    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
```

JustinRayAngus (approved):
Thanks Dave, 

This is looking good. I left a few final minor tweaks.

Source/Utils/ParticleUtils.H
dpgrote: I can make the change. How much does this matter? There are many other places in WarpX that have `AMREX_INLINE`.

JustinRayAngus: Good question. It depends on the compiler I suppose. From my understanding of the usage of the function, being called inside a hot loop, AMREX_FORCE_INLINE is suggested. Don't see any downside to it.

PR #5524, opened 2024-12-20
[WIP] Add support for electron impact ionization in binary collisions (DSMC module) 

oshapoval:
*Description edited by @RemiLehe *

This adds support for electron impact ionization, within the DSMC module.

Note: could we do binary charge exchange in the same way?

TODO:

- [ ] Use the proper energy cost: store the energy cost in the SplitAndScatter functor
- [ ] Avoid creating a particle of the neutral species
- [ ] Add support for excitation in DSMC (right now, it prevents us from running the capacitive discharge test)
- [ ] Test on GPU
- [x] Add a dedicated CI test for DSMC ionization. Note: for now, setting very high tolerances on purpose. This needs to be modified once all features are implemented in the PR.

PR #5654, opened 2025-02-11
DSMC: Add impact ionization

roelof-groenewald:
This PR extends the work already done by @RemiLehe and @oshapoval in #5524.

Some changes were made to how the ionization process is initialized, for example, the user must now specify the `target_species` (i.e., the species which undergoes ionization) as well as the `product_species`. The product species can include the colliding species (for example electron + neutral -> 2 x electron + ion), but does not have to (for example H$^+$ + D -> H$^+$ + D$^+$ + electron).

The test created by @archermarx is now passing (at least early on):
image

Todo:

- [x] test implementation
- [x] fix scattering process to conserve momentum
- [x] clean up code


roelof-groenewald:
At this point the existing DSMC test (as part of the capactive discharge test) succeeds, indicating that the existing DSMC functionality is not broken.

roelof-groenewald:
At present running the capacitive discharge problem while using DSMC for electron-impact ionization, results in a slightly higher average plasma density than the Turner result:
image

archermarx:
The results on the comparison to the global model look about as good as we were able to get with MCC, so that's good. I'm still curious why we're off from the theory. I can't think of any missing physics that we're not accounting for. Also, it's a bit strange that we get higher densities in DSMC vs MCC, as MCC doesn't permit neutrals to be depleted and should have an overall higher ionization rate.

Source/Particles/Collision/BinaryCollision/DSMC/CollisionFilterFunc.H
archermarx, line 1: Could we extract the max process count into a compile-time constant that the user could specify (i.e. `-DWARPX_MAX_SCATTERING_PROCS=10` or the like?) Or set the number much larger than five? While we're doing this PR it would also be good to actually address this in the constructor like the comment says we should.

archermarx (changes requested):
Thanks for getting this close to the finish line! I think this looks pretty good. In the spirit of cleanliness I think it would be good to go ahead and move the process count check to the constructor. I also think it would be worth exploring increasing the max process count (see prev comment), but I don't have a good enough understanding of what the performance implications would be to say. That could perhaps wait for a second PR, however.

roelof-groenewald:
> The results on the comparison to the global model look about as good as we were able to get with MCC, so that's good. I'm still curious why we're off from the theory. I can't think of any missing physics that we're not accounting for. Also, it's a bit strange that we get higher densities in DSMC vs MCC, as MCC doesn't permit neutrals to be depleted and should have an overall higher ionization rate.

I'm guessing the reason must be that the electron energy distribution function starts to deviate substantially from a Maxwellian. The theory and simulation seem to match up pretty nicely until the plasma density has increased by ~20%, at that point a sizeable fraction of electrons have been involved in a collision, and therefore likely do not follow a Maxwellian disrtibution anymore. I guess we could check this since we output the electron velocities in during the test. An alternative approach could be to use the electron distribution function at every output to predict the ionization rate over the next diagnostic interval.
In terms of why the DSMC density is higher than the MCC density, I'm not sure either. One possibility is that the binary ion-elastic scattering is causing the neutral gas to heat up a bit leading to a higher ionization rate. I can check this by switching the ion collisions back to MCC.

Update: It does seem like the ion-neutral collisions causes a drift in the ionization rate. Here is what I get if I only use DSMC for the ionization collisions:
image

Source/Particles/Collision/BinaryCollision/DSMC/CollisionFilterFunc.H
roelof-groenewald: It looks like a change to template the max number of processes was partly overwritten in this PR. I fixed that now.

archermarx:
> > The results on the comparison to the global model look about as good as we were able to get with MCC, so that's good. I'm still curious why we're off from the theory. I can't think of any missing physics that we're not accounting for. Also, it's a bit strange that we get higher densities in DSMC vs MCC, as MCC doesn't permit neutrals to be depleted and should have an overall higher ionization rate.
> 
> I'm guessing the reason must be that the electron energy distribution function starts to deviate substantially from a Maxwellian. The theory and simulation seem to match up pretty nicely until the plasma density has increased by ~20%, at that point a sizeable fraction of electrons have been involved in a collision, and therefore likely do not follow a Maxwellian disrtibution anymore. I guess we could check this since we output the electron velocities in during the test. An alternative approach could be to use the electron distribution function at every output to predict the ionization rate over the next diagnostic interval. 

That's a nice idea. In that approach, we'd run the simulation to completion, saving electron information at each output step. In the analysis, we'd then compute the empirical ionization rate coefficient (instead of pre-computing it for a Maxwellian over a range of temperatures). 

Alternatively, we could add binary e-e collisions to try and thermalize the electrons a bit more (though at ne = 1e14 they'd be pretty infrequent)

> In terms of why the DSMC density is higher than the MCC density, I'm not sure either. One possibility is that the binary ion-elastic scattering is causing the neutral gas to heat up a bit leading to a higher ionization rate. I can check this by switching the ion collisions back to MCC.
> 
> Update: It does seem like the ion-neutral collisions causes a drift in the ionization rate. Here is what I get if I only use DSMC for the ionization collisions: image

That's interesting, and I'm surprised that the result is so noticeable, but good to see that we can match MCC well if we neglect them.

archermarx (approved):
I think this looks good for now. We agree with the model in early times before any potential deviations from Maxwellian occur. A better test which accounts for that deviation could be added in a followup PR

Examples/Tests/ionization_dsmc/analysis_ionization_dsmc_3d.py
```diff
@@ -0,0 +1,223 @@
...
+    axs[i].set_title(title)
+    axs[i].plot(t_warpx, field_warpx, label="WarpX (" + method + ")")
+    axs[i].plot(t_theory, field_theory, label="theory", color="red", ls="--")
+
+    axs[i].legend()
+plt.tight_layout()
+plt.savefig("ionization_dsmc_density_Te.png", dpi=150)
+
+
+# For now, setting very high tolerances on purpose. This needs to be modified once all features are implemented in the PR.
```
RemiLehe: This seems obsolete.
```suggestion
```

Examples/Tests/ionization_dsmc/analysis_ionization_dsmc_3d.py
```diff
@@ -0,0 +1,223 @@
...
+    # The rate coefficient is given by
+
+    # $$
+    # k(T_e) = \int \sigma(E) E dE
+    # $$
+
+    # where the energy $E$ is drawn from a maxwellian distribution function with zero speed and temperature $T_e$.
+    # We solve this using a quasi-monte carlo approach, by drawing a large number of low-discrepancy samples from
+    # the appropriate distribution and obtaining the average of $\sigma(E) E$.
+    # """
```
RemiLehe: No need for the `#` character in the docstring.
```suggestion
    """Integrate cross sections over maxwellian VDF to obtain reaction rate coefficients
    Given electron energy in eV (`energy_eV`) and reaction cross sections at those energies (`sigma_m2`),
    this function computes the reaction rate coefficient $k(T_e)$ for maxwellian electrons at
    a provided list of electron temperatures `temperatures_eV`.

    The rate coefficient is given by

    $$
    k(T_e) = \int \sigma(E) E dE
    $$

    where the energy $E$ is drawn from a maxwellian distribution function with zero speed and temperature $T_e$.
    We solve this using a quasi-monte carlo approach, by drawing a large number of low-discrepancy samples from
    the appropriate distribution and obtaining the average of $\sigma(E) E$.
    """
```

Examples/Tests/ionization_dsmc/analysis_ionization_dsmc_3d.py
```diff
@@ -0,0 +1,223 @@
...
+    # $$
+
+    # where
+
+    # $$
+    # \dot{n} = n_n n_e k_{iz}(T_e),
+    # $$
+
+    # $k_iz$ is the ionization rate coefficient as a function of electron temperature in eV, and $\epsilon_{iz}$ is the ionization energy cost in eV.
+    # """
```
RemiLehe:
```suggestion
    """ Compute the right-hand side of ODE system that solves the global model described below.
    The global model solves for the evolution of plasma density ($n_e$), neutral density ($n_n),
    and electron temperature ($T_e$) in the presence of ionization.
    The model equations consist of a continuity equation for electrons and neutrals,
    combined with an energy equation for electrons.

    $$
    \frac{\partial n_e}{\partial t} = \dot{n}
    \frac{\partial n_n}{\partial t} = -\dot{n}
    \frac{3}{2}\frac{\partial n_e T_e}{\partial t} = -\dot{n} \epsilon_{iz},
    $$

    where

    $$
    \dot{n} = n_n n_e k_{iz}(T_e),
    $$

    $k_iz$ is the ionization rate coefficient as a function of electron temperature in eV, and $\epsilon_{iz}$ is the ionization energy cost in eV.
    """
```

Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
```diff
@@ -32,48 +32,56 @@ DSMCFunc::DSMCFunc (
...
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                !ionization_flag,
+                "DSMC only supports a single ionization process"
+            );
+            ionization_flag = true;
+
+            // TODO: Add a check that the first species is the electron species
+            // (This should be done for impact ionization with MCC too)
+            // And add a check that the ionization species has the same mass
+            // (and a positive charge), compared to the target species
```
RemiLehe, line 69: Do you want to handle this in a follow-up PR @roelof-groenewald ?

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -67,28 +67,68 @@ public:
...
-            // How many particles of product species i are created.
-            const index_type num_added = total * m_num_products_host[i];
-            num_added_vec[i] = static_cast<int>(num_added);
-            tile_products[i]->resize(products_np[i] + num_added);
+            // Add the number of product producing events to the species involved
+            // in those processes.
+            int num_products = m_num_products_host[i];
+            if ((i < 2) & (num_products == 1)) {
+                num_products = 0;
+            }
```
RemiLehe, line 127: Could you add a comment to explain what this part of the code does?

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -235,6 +313,184 @@ public:
...
+                uy1 -= uCOM_y;
+                uz1 -= uCOM_z;
+                ux_p1 -= uCOM_x;
+                uy_p1 -= uCOM_y;
+                uz_p1 -= uCOM_z;
+                ux_p2 -= uCOM_x;
+                uy_p2 -= uCOM_y;
+                uz_p2 -= uCOM_z;
+
+                if (mask[i] == int(ScatteringProcessType::IONIZATION)) {
```
RemiLehe, line 402: This seems redundant: we are already in an `else if` condition for `(mask[i] == int(ScatteringProcessType::IONIZATION))`

RemiLehe (commented):
Thanks for this PR!
Could you modify the documentation in `parameters.rst` to add ionization to the list of possible `dsmc` processes, and to add `ionization_target_species` in the parameters?

Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
roelof-groenewald: We actually don't need the first part of this anymore (for DSMC at least), but yes, I think we can leave the second part to a follow-up PR.

Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
```diff
@@ -32,48 +32,56 @@ DSMCFunc::DSMCFunc (
...
+        // collision set.
+        if (process.type() == ScatteringProcessType::IONIZATION) {
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                !ionization_flag,
+                "DSMC only supports a single ionization process"
+            );
+            ionization_flag = true;
+
+            // TODO: Add a check that the first species is the electron species
+            // (This should be done for impact ionization with MCC too)
```
roelof-groenewald:
```suggestion
```

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -67,28 +67,68 @@ public:
...
 
-        amrex::Vector<int> num_added_vec(m_num_product_species);
         for (int i = 0; i < m_num_product_species; i++)
         {
-            // How many particles of product species i are created.
-            const index_type num_added = total * m_num_products_host[i];
-            num_added_vec[i] = static_cast<int>(num_added);
-            tile_products[i]->resize(products_np[i] + num_added);
+            // Add the number of product producing events to the species involved
+            // in those processes.
```
```suggestion
            // Add the number of product producing events to the species involved
            // in those processes. For the two colliding particles, if either is set to 
            // have just 1 copy in the products it indicates that that species is not a
            // product of the product producing reaction (instead it is just tracked as
            // an outgoing particle in non-product producing reactions), and therefore 
            // it does not count in the products. 
```

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
roelof-groenewald: I added it in as a forward-looking thing since we could add excitation events (or other product producing events). I know we will need to change the higher if condition for this as well in that case, but it would still clearly show where the logic should split.

RemiLehe (approved):

PR #5660, opened 2025-02-12
Fix plot_distribution_mapping.py for 2D

atmyers:
The box metadata used in this script follows the AMReX conventions. We want "zyx" in 3D and "yx" in 2D.


ax3l (approved):
Thank you! :tada:

PR #5372, opened 2024-10-07
CI: update docs and reset tool for checksums

EZoni:
The major part of this PR is about updating the docs so that it is a bit easier for developers to connect the section on testing with the section on checksums.

Here's a couple of screenshots showing the new content organization for the testing and checksums sections:
<p align="center">
    
    
</p>

As part of this PR, I also update the tool that we have to reset checksums locally based on the Azure output. The update is necessary due to a change in #5297, as noted in https://github.com/ECP-WarpX/WarpX/pull/5297#discussion_r1787006065, given that new checksum files are now displayed as follows:
```

New checksums file test_2d_langmuir_multi.json:
{
  "lev=0": {
    "Bx": 0.0,
    "By": 5.726296856755232,
    "Bz": 0.0,
    "Ex": 3751589134191.326,
    "Ey": 0.0,
    "Ez": 3751589134191.332,
    "jx": 1.0100623329922576e+16,
    "jy": 0.0,
    "jz": 1.0100623329922578e+16
  },
  "electrons": {
    "particle_momentum_x": 5.668407513430198e-20,
    "particle_momentum_y": 0.0,
    "particle_momentum_z": 5.668407513430198e-20,
    "particle_position_x": 0.6553599999999999,
    "particle_position_y": 0.65536,
    "particle_weight": 3200000000000000.5
  },
  "positrons": {
    "particle_momentum_x": 5.668407513430198e-20,
    "particle_momentum_y": 0.0,
    "particle_momentum_z": 5.668407513430198e-20,
    "particle_position_x": 0.6553599999999999,
    "particle_position_y": 0.65536,
    "particle_weight": 3200000000000000.5
  }
}

```
as opposed to the old way
```

----------------
New file for test_2d_langmuir_multi:
{
  "lev=0": {
    "Bx": 0.0,
    "By": 5.726296856755232,
    "Bz": 0.0,
    "Ex": 3751589134191.326,
    "Ey": 0.0,
    "Ez": 3751589134191.332,
    "jx": 1.0100623329922576e+16,
    "jy": 0.0,
    "jz": 1.0100623329922578e+16
  },
  "electrons": {
    "particle_momentum_x": 5.668407513430198e-20,
    "particle_momentum_y": 0.0,
    "particle_momentum_z": 5.668407513430198e-20,
    "particle_position_x": 0.6553599999999999,
    "particle_position_y": 0.65536,
    "particle_weight": 3200000000000000.5
  },
  "positrons": {
[...]

PR #368, opened 2019-09-18
Unit tests for photon pusher

lucafedeli88:
I've added a test which should check if the photon pusher is working properly. 

Photons should keep their initial momentum and should move along straight lines at the speed of light. 
The `inputs` file initializes 16 photon particles with momenta having different directions and magnitudes. Photon particles are then evolved for 50 steps. The python script `check.py` verifies that the final position is close enough to the expected one (tolerance of ~ 1e-14 on the relative error) and that the final momentum is **exactly** equal to the expected one. 

Btw, where does m_e = 9.10938291e-31 (used in WarpX) come from? It is slightly different from the value currently reported by Wikipedia.

PR #272, opened 2019-08-14
For RigidInjection, fixed calculation of done_injecting

dpgrote:
The variable done_injecting was not being calculated correctly if the injection plane started outside of the simulation domain.

PR #5661, opened 2025-02-12
Update constants to use values from CODATA 2022

dpgrote:
The values of the physical constants were from CODATA 2018. These should be updated to the current accepted values as specified in CODATA 2022.

This breaks many CI benchmarks since the checks are at a higher precision than the changes in the constants.

This PR change is also needed since the older version of Ubuntu (which is used for the Azure CI tests) is being deprecated. This PR switches to the newest Ubuntu which includes a newer version of scipy where the constants have also been updated (since version 1.15.0 on January 3).

Note that this PR has fine tuned values of `mu0` and `alpha` so that the relation between the constants is exact. The is means that these values will differ slightly from the values in scipy.constants.

Many of the CI tests were using locally defined constants in the input files and analysis scripts. These were all changed to rely on the built in constants in the parser and `scipy.constants`.


EZoni:
Thanks, Dave. 

If there is a large number of benchmarks that need to be reset, this could be a good opportunity to test again our tool [Tools/DevUtils/update_benchmarks_from_azure_output.py](https://github.com/ECP-WarpX/WarpX/blob/development/Tools/DevUtils/update_benchmarks_from_azure_output.py) and its [instructions](https://warpx.readthedocs.io/en/latest/developers/how_to_test.html#how-to-reset-checksums-locally) in our documentation.

In theory, I updated and tested the tool manually in #5372. However, it is not tested automatically yet.

dpgrote:
> Thanks, Dave.
> 
> If there is a large number of benchmarks that need to be reset, this could be a good opportunity to test again our tool [Tools/DevUtils/update_benchmarks_from_azure_output.py](https://github.com/ECP-WarpX/WarpX/blob/development/Tools/DevUtils/update_benchmarks_from_azure_output.py) and its [instructions](https://warpx.readthedocs.io/en/latest/developers/how_to_test.html#how-to-reset-checksums-locally) in our documentation.
> 
> In theory, I updated and tested the tool manually in #5372. However, it is not tested automatically yet.

Thanks @EZoni ! It worked and was easy to do. BTW, to download the raw log file, I copied the URL from he location bar and pasted it into the `curl` command, "curl https://dev.azure.com/ECP-WarpX/... > raw_log", making it easy to download it.

Note that almost all of the changes in the benchmarks are small as expected, ~1.e-9 or smaller. One exception is the `test_3d_beam_beam_collision` test with errors of order 10% are seen, presumably because it runs a long time allowing the differences to grow.

EZoni:
> Note that almost all of the changes in the benchmarks are small as expected, ~1.e-9 or smaller. One exception is the `test_3d_beam_beam_collision` test with errors of order 10% are seen, presumably because it runs a long time allowing the differences to grow.

I agree. I think that test has relatively large tolerances anyways, if I remember correctly. @aeriforme, what do you think?

EZoni:
> Thanks @EZoni ! It worked and was easy to do. BTW, to download the raw log file, I copied the URL from he location bar and pasted it into the curl command, "curl https://dev.azure.com/ECP-WarpX/... > raw_log", making it easy to download it.

Thanks for pointing this out! I added this hint to our documentation in #5663.

Source/ablastr/constant.H
```diff
@@ -43,28 +43,28 @@ namespace ablastr::constant
         //! vacuum speed of light [m/s]
         static constexpr auto c = 299'792'458._rt;
         //! vacuum permittivity: dielectric permittivity of vacuum [F/m]
-        static constexpr auto ep0 = 8.8541878128e-12_rt;
+        static constexpr auto ep0 = 8.8541878188e-12_rt;
```
lucafedeli88, line 46: Thanks @dpgrote ! I  think that we have to change `Values are the 2018 CODATA recommended values` few lines above with `Values are the 2022 CODATA recommended values` then.

Regression/Checksum/benchmarks_json/test_rz_laser_injection_from_lasy_file.json
```diff
@@ -1,12 +1,12 @@
 {
   "lev=0": {
-    "Br": 100277153.12954171,
-    "Bz": 2508844.115818364,
-    "Er": 3.263693682174799,
-    "Et": 3.004484377567921e+16,
-    "Ez": 0.24219246674343975,
-    "jr": 49.68134418425658,
```
lucafedeli88, line 8: The discrepancy for `jr` here is non-negligible! Do we have an idea of the possible reason?

lucafedeli88 (changes requested):
Thanks @dpgrote ! I think that updating constants to CODATA 2022 is a good idea.

I've reviewed the logs of the tests that have failed.

I agree that issues with `test_3d_beam_beam_collision.checksum`   (and also with `test_2d_collision_xz_picmi.checksum` ) are likely due to the fact that these tests are relatively long. 

We need to investigate a bit better the cases `test_1d_ohm_solver_ion_beam_picmi`, `test_1d_ohm_solver_ion_beam_picmi`, and `test_3d_qed_schwinger_2` show non-negligible discrepancies (I will have a look at the QED-related one).

We need to increase the tolerance of several analysis scrips, since they seem to be a bit too strict :
`test_2d_theta_implicit_jfnk_vandb`, `test_2d_theta_implicit_jfnk_vandb_filtered`, `test_2d_theta_implicit_jfnk_vandb_picmi`, `test_2d_theta_implicit_strang_psatd`, `test_2d_pec_field_insulator_implicit_restart`,`test_3d_particle_boundaries`, `test_3d_load_external_field_grid_picmi`, `test_3d_load_external_field_grid_picmi`,`test_3d_particle_fields_diags` , `test_3d_particle_fields_diags`, `test_3d_reduced_diags`.

lucafedeli88:
> Thanks @dpgrote ! I think that updating constants to CODATA 2022 is a good idea.
> 
> I've reviewed the logs of the tests that have failed.
> 
> I agree that issues with `test_3d_beam_beam_collision.checksum` (and also with `test_2d_collision_xz_picmi.checksum` ) are likely due to the fact that these tests are relatively long.
> 
> We need to investigate a bit better the cases `test_1d_ohm_solver_ion_beam_picmi`, `test_1d_ohm_solver_ion_beam_picmi`, and `test_3d_qed_schwinger_2` show non-negligible discrepancies (I will have a look at the QED-related one).
> 
> We need to increase the tolerance of several analysis scrips, since they seem to be a bit too strict : `test_2d_theta_implicit_jfnk_vandb`, `test_2d_theta_implicit_jfnk_vandb_filtered`, `test_2d_theta_implicit_jfnk_vandb_picmi`, `test_2d_theta_implicit_strang_psatd`, `test_2d_pec_field_insulator_implicit_restart`,`test_3d_particle_boundaries`, `test_3d_load_external_field_grid_picmi`, `test_3d_load_external_field_grid_picmi`,`test_3d_particle_fields_diags` , `test_3d_particle_fields_diags`, `test_3d_reduced_diags`.

I can provide a possible explanation for the QED-related test. Looking at the analysis script I found this comment:
```
elif test_number == "2":
    # Second Schwinger test with stronger EM field. Many pairs are created and a Gaussian
    # distribution is used to get the weights of the particles. This is the most sensitive test
    # because the relative std is extremely low.
    Ex_test = 1.0e18
    Bx_test = 1679288857.0516706
    By_test = 525665014.1557486
    Bz_test = 1836353079.9561853
```

Note that the analysis script uses these constants (CODATA 2014, I think):
```
c = 299792458.
m_e = 9.10938356e-31
e = 1.6021766208e-19
hbar = 1.054571800e-34
```

while  PICSAR-QED uses (CODATA 2018):
```
template<typename RealType = double>
constexpr auto light_speed = RealType(299792458.);
    
template<typename RealType = double>
constexpr auto electron_mass = RealType(9.1093837015e-31);

template<typename RealType = double>
constexpr auto elementary_charge = RealType(1.602176634e-19);
    
template<typename RealType = double>
constexpr auto reduced_plank = RealType(1.054571817e-34);

```

dpgrote:
@lucafedeli88 Thanks for looking over this PR. I think a number of the issues are related to the use of scipy.constants in the analysis scripts, since the constants are inconsistent, i.e. still the CODATA2018 values. Until this issue is resolved, I don't think this PR should be merged. Unfortunately, there is no easy solution. The simplest is probably to use the most recent version of ubuntu for the CI tests, which uses the most recent scipy with updated constants.

A more robust longer term solution would be to create a new light-weight Python module that has the constants with the values consistent with the ones in C++, and then use this everywhere instead of relying on scipy.constants. This would guarantee that the values are always consistent.

lucafedeli88:
> @lucafedeli88 Thanks for looking over this PR. I think a number of the issues are related to the use of scipy.constants in the analysis scripts, since the constants are inconsistent, i.e. still the CODATA2018 values. Until this issue is resolved, I don't think this PR should be merged. Unfortunately, there is no easy solution. The simplest is probably to use the most recent version of ubuntu for the CI tests, which uses the most recent scipy with updated constants.
> 
> A more robust longer term solution would be to create a new light-weight Python module that has the constants with the values consistent with the ones in C++, and then use this everywhere instead of relying on scipy.constants. This would guarantee that the values are always consistent.

We could maybe discuss this in the upcoming developers' meeting. Using Ubuntu 24.04 in CI tests should be rather straightforward.

EZoni:
> We could maybe discuss this in the upcoming developers' meeting. Using Ubuntu 24.04 in CI tests should be rather straightforward.

Yes, we can do that soon. There had been discussions about the "need" to have less strict tolerances for the checksums to be compatible with version upgrades like this one. The work done in #5456 has set up things so that we can do that easily (see point "Add logic to reset tolerances based on environment variables" in the follow-up list of that PR description). This said, Ubuntu LTS version upgrades come once every two years, so I personally think that the tolerance fine tuning is not a real roadblock for this particular update, we could simply upgrade the Ubuntu version and reset the checksums that need to be reset. I will get to this, one way or another, as soon as possible.

EZoni:
To: @dpgrote  @lucafedeli88 
Cc: @ax3l 

Following up on discussions here as well as on Slack, I opened https://github.com/ECP-WarpX/WarpX/pull/5731 to work on the Ubuntu version upgrade in a dedicated PR.

EZoni:
Ideally I thought it could be good to have the Ubuntu upgrade run through first without prior checksums changes and record how tests fail, like I did in https://github.com/ECP-WarpX/WarpX/pull/5731#discussion_r1980509637, to make it easier to discuss if/when a tolerance upgrade could be appropriate. But let's see if we can still assess this despite the pre-existing checksums changes.

EZoni:
As mentioned in of the last comments in #5731, we can try to merge `development` to incorporate #5736 in order to address some of the Python errors that we see here. If that does not work, we need to add the workaround I had added in #5731, before the installation of `Regression/requirements.txt` in `.azure-pipelines.yml`:
```bash
      # (remove system copy of Matplotlib to avoid conflict
      # with version set in the requirements file - see, e.g.,
      # https://github.com/matplotlib/matplotlib/issues/28768)
      sudo apt remove python3-matplotlib
```

EZoni:
If I'm not mistaken, I still see errors like
```
  File "/home/vsts/work/1/s/Examples/Tests/langmuir/analysis_2d.py", line 116, in <module>
    cax1 = make_axes_locatable(ax1).append_axes("right", size="5%", pad="5%")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vsts/.local/lib/python3.12/site-packages/matplotlib/_api/deprecation.py", line 382, in wrapper
    return func(*inner_args, **inner_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/mpl_toolkits/axes_grid1/axes_divider.py", line 525, in append_axes
    ax = create_axes(
         ^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/mpl_toolkits/axes_grid1/axes_divider.py", line 459, in new_horizontal
    ax = self._get_new_axes(**kwargs)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/mpl_toolkits/axes_grid1/axes_divider.py", line 425, in _get_new_axes
    axes_class = axes._axes_class
                 ^^^^^^^^^^^^^^^^
AttributeError: 'Axes' object has no attribute '_axes_class'
```
which occurr only with older versions of Matplotlib.

There is a conflict between the system version (old) coming from `python3-matplotlib` and the version installed via `pip` through the requirements file (newer).

#5736 does not seem to have resolved the issue, so I think we need the workaround in https://github.com/ECP-WarpX/WarpX/pull/5661#issuecomment-2705000459, unless you have other solutions in mind.

Examples/Tests/nuclear_fusion/analysis_proton_boron_fusion.py
```diff
@@ -80,7 +80,7 @@
 
 ## Checks whether this is the 2D or the 3D test
 with open("./warpx_used_inputs") as warpx_used_inputs:
-    is_2D = re.search("geometry.dims\s*=\s*2", warpx_used_inputs.read())
+    is_2D = re.search(r"geometry.dims\s*=\s*2", warpx_used_inputs.read())
```
EZoni, line 83: Here I think it would be also safe to remove the `\s*`, since `warpx_used_inputs` is generated automatically and uses only one white space, as in `key = value`, but up to you. It might make it more readable overall:
```suggestion
    is_2D = re.search("geometry.dims = 2", warpx_used_inputs.read())
```

EZoni:
I just saw this warning on the Azure DevOps website:

> The Ubuntu-20.04 image for the Microsoft-hosted "Azure Pipelines" pool is deprecated and will be retired April 1st. Please review [aka.ms/azdo-ubuntu-20.04](https://aka.ms/azdo-ubuntu-20.04) for brownout schedule and other image news.

It sounds like ideally we should finalize this PR before the end of next week.

@dpgrote 
Let me know if we should touch base on this PR and/or somehow coordinate to finalize it. I'd be happy to contribute beyond reviewing, if needed.

dpgrote:
> I just saw this warning on the Azure DevOps website:
> 
> > The Ubuntu-20.04 image for the Microsoft-hosted "Azure Pipelines" pool is deprecated and will be retired April 1st. Please review [aka.ms/azdo-ubuntu-20.04](https://aka.ms/azdo-ubuntu-20.04) for brownout schedule and other image news.
> 
> It sounds like ideally we should finalize this PR before the end of next week.
> 
> @dpgrote Let me know if we should touch base on this PR and/or somehow coordinate to finalize it. I'd be happy to contribute beyond reviewing, if needed.

@ezoni I'll put some time into finishing this PR this week.

dpgrote:
> > I just saw this warning on the Azure DevOps website:
> > > The Ubuntu-20.04 image for the Microsoft-hosted "Azure Pipelines" pool is deprecated and will be retired April 1st. Please review [aka.ms/azdo-ubuntu-20.04](https://aka.ms/azdo-ubuntu-20.04) for brownout schedule and other image news.
> > 
> > 
> > It sounds like ideally we should finalize this PR before the end of next week.
> > @dpgrote Let me know if we should touch base on this PR and/or somehow coordinate to finalize it. I'd be happy to contribute beyond reviewing, if needed.
> 
> @EZoni I'll put some time into finishing this PR this week.

@EZoni I worked on this some and made progress but there are still tests failing. They are showing odd behaviors and it would be good to have someone else look them over, if you would have the chance.

EZoni:
For the 2D tests, it seems like we might have only round-off errors. What do you think?

- `test_2d_pec_field_insulator_implicit.analysis`:
    It might be enough to increase `tolerance_rel` to from `1e-13` to `1e-12`.
    ```
            Start 268: test_2d_pec_field_insulator_implicit.analysis
    268/402 Test #268: test_2d_pec_field_insulator_implicit.analysis ...........................***Failed    0.75 sec
    energy accounting error = 6.958681222801173e-13
    tolerance_rel = 1e-13
      File "/home/vsts/work/1/s/Examples/Tests/pec/analysis_pec_insulator_implicit.py", line 57, in <module>
        assert energy_difference_fraction < tolerance_rel
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    AssertionError
    ```

- `test_2d_pec_field_insulator_implicit_restart.analysis`:
    It might be enough to increase `tolerance_rel` from `1e-13` to `1e-12`.
    ```
            Start 272: test_2d_pec_field_insulator_implicit_restart.analysis
    271/402 Test #272: test_2d_pec_field_insulator_implicit_restart.analysis ...................***Failed    0.72 sec
    energy accounting error = 6.958681222801173e-13
    tolerance_rel = 1e-13
      File "/home/vsts/work/1/s/Examples/Tests/pec/analysis_pec_insulator_implicit.py", line 57, in <module>
        assert energy_difference_fraction < tolerance_rel
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    AssertionError
    ```

In the analysis script used for both tests, I see this comment:
```py
# The implicit solver is used, converging to machine tolerance. The energy accounting        
# should be exact to machine precision, so that the energy is the system        
# should be the same as the amount of energy pushed in from the boundary.
```
However, in the base input file of both tests, I see this block:
```
implicit_evolve.nonlinear_solver = "picard"
picard.verbose = true
picard.max_iterations = 25
picard.relative_tolerance = 0.0
picard.absolute_tolerance = 0.0
picard.require_convergence = false
```
which seems to suggest that convergence of the implicit solver is not enforced in these tests. This could explain why even the original tolerance. `1e-13`, was already a few orders of magnitude larger than the machine limit for double precision floating point type. If this is indeed the case, I think it should be safe to increase the tolerance from `1e-13` to `1e-12`, as I mentioned above.

EZoni:
For the 3D tests:

- `test_3d_qed_schwinger_2.analysis`:
    @lucafedeli88 Do you think this could be explained by the fact that we now use the 2022 CODATA values in WarpX and the 2018 CODATA values in PICSAR-QED, as you had already hinted in https://github.com/BLAST-WarpX/warpx/pull/5661#issuecomment-2671780452 (although the problem at the time when that comment was posted was that we were using CODATA values in WarpX even older than the ones in PICSAR-QED)?
    ```
            Start 368: test_3d_qed_schwinger_2.analysis
    368/477 Test #368: test_3d_qed_schwinger_2.analysis ........................................***Failed    2.24 sec
    yt : [INFO     ] 2025-03-26 01:22:13,659 Parameters: current_time              = 9.827726403123683e-17
    yt : [INFO     ] 2025-03-26 01:22:13,659 Parameters: domain_dimensions         = [ 8  8 32]
    yt : [INFO     ] 2025-03-26 01:22:13,660 Parameters: domain_left_edge          = [-5.e-07 -5.e-07 -5.e-07]
    yt : [INFO     ] 2025-03-26 01:22:13,660 Parameters: domain_right_edge         = [5.e-07 5.e-07 5.e-07]
    difference between expected and actual number of pairs created: 474339246080.0 dimensionless
    tolerance: 32212137889.04688
      File "/home/vsts/work/1/s/Examples/Tests/qed/analysis_schwinger.py", line 156, in <module>
        do_analysis(Ex_test, Ey_test, Ez_test, Bx_test, By_test, Bz_test)
      File "/home/vsts/work/1/s/Examples/Tests/qed/analysis_schwinger.py", line 153, in do_analysis
        assert error < 5 * std_total_physical_pairs_created
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    AssertionError
    ```

- `test_3d_photon_pusher.analysis`:
    Based on the analysis script, I understand that momentum should be conserved exactly. The relative error that we seem to get here is `1.7217529609352883e-16`. This is lower than `2.220446049250313e-16`, which is what `np.finfo(np.float64).eps` returns as machine limit for double precision floating point type. I would consider this to be equivalent to zero in our discrete computations. In fact, I would replace `tol_mom = 0.0` with `tol_mom = np.finfo(np.float64).eps`.
    @lucafedeli88, @dpgrote What do you think?
    ```
            Start 315: test_3d_photon_pusher.analysis
    315/477 Test #315: test_3d_photon_pusher.analysis ..........................................***Failed    2.96 sec
    yt : [INFO     ] 2025-03-26 01:20:04,679 Parameters: current_time              = 1.5045571887081817e-15
    yt : [INFO     ] 2025-03-26 01:20:04,679 Parameters: domain_dimensions         = [64 64 64]
    yt : [INFO     ] 2025-03-26 01:20:04,679 Parameters: domain_left_edge          = [-5.e-07 -5.e-07 -5.e-07]
    yt : [INFO     ] 2025-03-26 01:20:04,679 Parameters: domain_right_edge         = [5.e-07 5.e-07 5.e-07]
    max(disc_pos) = 6.098637220230956e-16
    tol_pos = 1e-14
    max(disc_mom) = 1.7217529609352883e-16
    tol_mom = 0.0
     File "/home/vsts/work/1/s/Examples/Tests/photon_pusher/analysis.py", line 206, in <module>
        main()
      File "/home/vsts/work/1/s/Examples/Tests/photon_pusher/analysis.py", line 202, in main
        check()
      File "/home/vsts/work/1/s/Examples/Tests/photon_pusher/analysis.py", line 148, in check
        assert (max(disc_pos) <= tol_pos) and (max(disc_mom) <= tol_mom)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    AssertionError
    ```

EZoni:
For the RZ tests:

- `test_rz_ohm_solver_em_modes_picmi.analysis`:
    ```
            Start 122: test_rz_ohm_solver_em_modes_picmi.analysis
    122/170 Test #122: test_rz_ohm_solver_em_modes_picmi.analysis ..............................***Failed    3.10 sec
    Processing iteration 0
    Processing iteration 10
    Processing iteration 20
    Processing iteration 30
    Processing iteration 40
    Processing iteration 50
    Processing iteration 60
    Processing iteration 70
    Processing iteration 80
    Processing iteration 90
    Processing iteration 100
    Amplitude sample:  [101.45696649  42.99404119  36.75019622  37.88382928]
      File "/home/vsts/work/1/s/Examples/Tests/ohm_solver_em_modes/analysis_rz.py", line 181, in <module>
        assert np.allclose(
               ^^^^^^^^^^^^
    AssertionError
    ```
    It seems that the failing call is
    ```
     assert np.allclose(                                                         
         amps, np.array([59.23850009, 19.26746169, 92.65794174, 10.83627164])    
     )
    ```
    with `amps` being as above, namely `[101.45696649  42.99404119  36.75019622  37.88382928]`.
    @roelof-groenewald, @clarkse The values seem strongly off. Do you have any idea why this might be happening, given that the only change in this PR is related to updating the values for physical constants and updating the Ubuntu version for the tests?

roelof-groenewald:
@EZoni I agree that is quite a substantial change. I wouldn't say it necessarily points to a problem since this test relies on PIC noise to excite the normal modes, and small changes in initial setup could affect when / how those modes are excited. That being said, I would appreciate it if the full test (i.e. without the `--test` flag set) can be executed before this PR is merged to confirm that the expected normal modes still emerge.

lucafedeli88:
> For the 3D tests:
>
>    test_3d_qed_schwinger_2.analysis:
>    @lucafedeli88 Do you think this could be explained by the fact that we now use the 2022 CODATA values in WarpX and the 2018 CODATA values in PICSAR-QED, as you had already hinted in 
>
>https://github.com/BLAST-WarpX/warpx/pull/5661#issuecomment-2671780452 (although the problem at the time when that comment was posted was that we were using CODATA values in WarpX even older than the ones in PICSAR-QED)?

@EZoni , I think so.  This specific test is extremely sensitive. I am currently updating the physical constants in PICSAR-QED to see if it makes any difference. 

> test_3d_photon_pusher.analysis:
> Based on the analysis script, I understand that momentum should be conserved exactly. The relative error that we seem to get here is 1.7217529609352883e-16. This is lower than 2.220446049250313e-16, which is what np.finfo(np.float64).eps returns as machine limit for double precision floating point type. I would consider this to be equivalent to zero in our discrete computations. In fact, I would replace tol_mom = 0.0 with tol_mom = np.finfo(np.float64).eps.

What I find puzzling about this result is that the momentum should not change at all... I think that if it changes, even by something at machine precision, understanding why is still worth of investigation

EZoni:
> @EZoni , I think so. This specific test is extremely sensitive. I am currently updating the physical constants in PICSAR-QED to see if it makes any difference.

@lucafedeli88 
Thanks, sounds good. Feel free to ping me when https://github.com/ECP-WarpX/picsar/pull/58 is merged in PICSAR-QED, so we can do an update of the version we fetch in this PR as well.

EZoni:
> @EZoni I agree that is quite a substantial change. I wouldn't say it necessarily points to a problem since this test relies on PIC noise to excite the normal modes, and small changes in initial setup could affect when / how those modes are excited. That being said, I would appreciate it if the full test (i.e. without the --test flag set) can be executed before this PR is merged to confirm that the expected normal modes still emerge.

@roelof-groenewald 
Thanks, understood. Based on what you remember of this test, is it something that we could run relatively quickly with, say, 8 MPI processes locally on a laptop? I'm trying that now and, even though the 8 grids don't seem to be too large (128 x 64), it seems to be taking a significant amount of time to progress, which may mean that I may be doing something wrong.

EZoni:
@roelof-groenewald 

Thanks for our discussion on Slack! Following up on that, this is the result of the test on this branch after 1000 time steps:

Screenshot from 2025-03-27 02-54-05

I will update the plot here as more time steps become available, but it seems like the expected normal modes still emerge.

Do we have your green light to update 
```py
 assert np.allclose(                                                         
     amps, np.array([59.23850009, 19.26746169, 92.65794174, 10.83627164])    
 )
```
with
```py
 assert np.allclose(                                                         
     amps, np.array([101.45696649  42.99404119  36.75019622  37.88382928])    
 )
```
in `Examples/Tests/ohm_solver_em_modes/analysis_rz.py`?

EZoni:
@dpgrote 

https://github.com/ECP-WarpX/picsar/pull/58 was merged. I think we should run `./Tools/Release/updatePICSAR.py` in your branch and update the PICSAR version. In this case I think it makes more sense to do this here, together with the other changes in this PR, rather than separately (as we usually do with the "weekly updates").

roelof-groenewald:
Thanks @EZoni! Those results look good, so I would say it is fine to update the benchmark values.

EZoni:
@dpgrote @lucafedeli88 

This is the PICSAR PR to be reviewed and merged first: https://github.com/ECP-WarpX/picsar/pull/59.

EZoni:
@dpgrote 

@lucafedeli88 merged the PICSAR PRs. I think we should update this PR to fetch commit https://github.com/ECP-WarpX/picsar/commit/91515a5fc8505e18b0bedc48f58f688ef084bc4a (`91515a5fc8505e18b0bedc48f58f688ef084bc4a`) from PICSAR. 

This can be done by running `./Tools/Release/updatePICSAR.py` locally in your branch (recommended) or by updating the commit manually where needed. Would you like to do this or shall I push the change?

```cmake
$ git diff
diff --git a/cmake/dependencies/PICSAR.cmake b/cmake/dependencies/PICSAR.cmake
index d5249b616..518894a13 100644
--- a/cmake/dependencies/PICSAR.cmake
+++ b/cmake/dependencies/PICSAR.cmake
@@ -88,7 +88,7 @@ function(find_picsar)
         #message(STATUS "PICSAR: Using version '${PICSAR_VERSION}'")
     else()
         # not supported by PICSAR (yet)
-        #find_package(PICSAR 25.01 CONFIG REQUIRED QED)
+        #find_package(PICSAR 25.04 CONFIG REQUIRED QED)
         #message(STATUS "PICSAR: Found version '${PICSAR_VERSION}'")
         message(FATAL_ERROR "PICSAR: Cannot be used as externally installed "
             "library yet. "
@@ -109,7 +109,7 @@ if(WarpX_QED)
     set(WarpX_picsar_repo "https://github.com/ECP-WarpX/picsar.git"
         CACHE STRING
         "Repository URI to pull and build PICSAR from if(WarpX_picsar_internal)")
-    set(WarpX_picsar_branch "25.01"
+    set(WarpX_picsar_branch "91515a5fc8505e18b0bedc48f58f688ef084bc4a"
         CACHE STRING
         "Repository branch for WarpX_picsar_repo if(WarpX_picsar_internal)")
```

EZoni:
@dpgrote 

I will go ahead and push the PICSAR change here, just to trigger CI so that we can see what tests remain to be fixed. Ideally we could aim at merging this PR by the end of today.

dpgrote:
> > test_3d_photon_pusher.analysis:
> > Based on the analysis script, I understand that momentum should be conserved exactly. The relative error that we seem to get here is 1.7217529609352883e-16. This is lower than 2.220446049250313e-16, which is what np.finfo(np.float64).eps returns as machine limit for double precision floating point type. I would consider this to be equivalent to zero in our discrete computations. In fact, I would replace tol_mom = 0.0 with tol_mom = np.finfo(np.float64).eps.
> 
> What I find puzzling about this result is that the momentum should not change at all... I think that if it changes, even by something at machine precision, understanding why is still worth of investigation

This test is still failing. I looked into it some and don't understand what is happening. The momentum is never modified but the values somehow change slightly at some point, from 299792458 to 299792458.00000006.

EZoni:
> This test is still failing. I looked into it some and don't understand what is happening. The momentum is never modified but the values somehow change slightly at some point, from 299792458 to 299792458.00000006.

@dpgrote @lucafedeli88 

Should we relax that tolerance to the machine limit for double precision floating point to finalize this PR and then open a new, follow-up PR where someone familiar with the test can reset the tolerance to strict zero (if it is indeed true that the momentum variable is never touched by any part of the code at any point in the test) and debug further?

dpgrote:
> > This test is still failing. I looked into it some and don't understand what is happening. The momentum is never modified but the values somehow change slightly at some point, from 299792458 to 299792458.00000006.
> 
> @dpgrote @lucafedeli88
> 
> Should we relax that tolerance to the machine limit for double precision floating point to finalize this PR and then open a new, follow-up PR where someone familiar with the test can reset the tolerance to strict zero (if it is indeed true that the momentum variable is never touched by any part of the code at any point in the test) and debug further?

To get this finished, I agree that the tolerance should be increased. I'll set it to `np.finfo(np.float64).eps`. There is something in the code causing that last bit to be jiggled, but I couldn't find it.

Future work as examine why the change is happening and reduce to tolerance back to zero when fixed.

dpgrote:
@EZoni  There is still some weirdness. The values for the CI test `test_3d_ionization_ion_dsmc` keep changing. I updated the values after the small adjustment in the constants and after the picsar version was updated but they changed again. Any thoughts?

EZoni:
> @EZoni There is still some weirdness. The values for the CI test `test_3d_ionization_ion_dsmc` keep changing. I updated the values after the small adjustment in the constants and after the picsar version was updated but they changed again. Any thoughts?

@dpgrote 

Kind of difficult to know if this was just some kind of Azure runner glitch or not. I think it could be worth trying to reset the checksums again and see if they still fail. @roelof-groenewald might be familiar with that test as well, not sure if he observed anything like this already in the past.

roelof-groenewald:
> > @EZoni There is still some weirdness. The values for the CI test `test_3d_ionization_ion_dsmc` keep changing. I updated the values after the small adjustment in the constants and after the picsar version was updated but they changed again. Any thoughts?

I don't know why the ion impact ionization test would be impacted but not the electron impact ionization one. They run the same DSMC routine, just with different colliding species. So it is odd for only one to have it's checksum values changed. The ion version of that test is pretty new though, so maybe it is a bit more unstable than the others. We can keep an eye on it with future PRs to see if it fails "randomly".

dpgrote:
> > @EZoni There is still some weirdness. The values for the CI test `test_3d_ionization_ion_dsmc` keep changing. I updated the values after the small adjustment in the constants and after the picsar version was updated but they changed again. Any thoughts?
> 
> @dpgrote
> 
> Kind of difficult to know if this was just some kind of Azure runner glitch or not. I think it could be worth trying to reset the checksums again and see if they still fail. @roelof-groenewald might be familiar with that test as well, not sure if he observed anything like this already in the past.

@EZoni @roelof-groenewald I did update the benchmarks again and it failed again.

EZoni:
> @EZoni @roelof-groenewald I did update the benchmarks again and it failed again.

This is very strange, I don't understand it. Especially, as @roelof-groenewald said, based on the input file, this test seems identical to the electron one, just with an ion species instead. I don't understand why repeating the test on Azure runners changes the checksums of the ion test (and only of that one).

EZoni:
If I repeat the test multiple times locally on my computer, I get the same checksums. Once I reset them the first time, they seem to be okay each other subsequent time.

dpgrote:
> If I repeat the test multiple times locally on my computer, I get the same checksums. Once I reset them the first time, they seem to be okay each other subsequent time.

I'm able to reproduce something like the problem. Running the ctest locally on ubuntu, it fails that checksum. If I update the benchmarks, it will fail again. Now I have a case where I ran ctest and it passed, then ran it again and it failed. I have something now where I can look at what is changing the results.

dpgrote:
@EZoni  @roelof-groenewald The CI test is now fixed. The issue was a long expected bug related to how the synchronization of the velocity is done for the final diagnostic. The fix was to use the new option `warpx.synchronize_velocity_for_diagnostics`, which does a clean synchronization without the risk of an out of memory access. I suspected this issue when I started seeing NaNs appearing in the output and getting arithmetic exception errors sometimes when running the tests.

roelof-groenewald:
Thanks for digging into that test failure, @dpgrote!

Source/ablastr/constant.H
```diff
@@ -43,28 +43,32 @@ namespace ablastr::constant
...
         //! dalton: unified atomic mass unit [kg]
-        static constexpr auto m_u = 1.66053906660e-27_rt;
+        static constexpr auto m_u = 1.66053906892e-27_rt;
 
         //! reduced Planck Constant = h / tau [J*s]
-        static constexpr auto hbar = 1.054571817e-34_rt;
+        static constexpr auto hbar = 1.0545718176461565e-34_rt;
         //! fine-structure constant = mu0/(4*pi)*q_e*q_e*c/hbar [dimensionless]
-        static constexpr auto alpha = 0.007297352573748943_rt;
+        //! The value is calculated from the espression. This differs
```
roelof-groenewald:
```suggestion
        //! The value is calculated from the expression. This differs
```

roelof-groenewald (approved):
Thanks for this really big lift @dpgrote and @EZoni!

lucafedeli88 (approved):
Thanks a  lot for this PR!
One question: if not setting `warpx.synchronize_velocity_for_diagnostics = 1`  can lead to out-of-bounds accesses, shouldn't we set this to 1 by default and maybe warn the user if they set it to 0?

lucafedeli88:
> Thanks a lot for this PR! One question: if not setting `warpx.synchronize_velocity_for_diagnostics = 1` can lead to out-of-bounds accesses, shouldn't we set this to 1 by default and maybe warn the user if they set it to 0?

Ah, yes, we probably should, otherwise @dpgrote  would not have opened this https://github.com/BLAST-WarpX/warpx/pull/5816

EZoni:
Thanks a lot for the fix and for all the work, @dpgrote. 

And thanks all, @lucafedeli88  and @roelof-groenewald, for your work and feedback as well. 

Is it clear why this issue started to appear when we updated the physical constants? Why did it not occur before? 

Anyways, I will approve and merge the PR in a minute.

EZoni (approved):

dpgrote:
> Thanks a lot for the fix and for all the work, @dpgrote.
> 
> And thanks all, @lucafedeli88 and @roelof-groenewald, for your work and feedback as well.
> 
> Is it clear why this issue started to appear when we updated the physical constants? Why did it not occur before?
> 
> Anyways, I will approve and merge the PR in a minute.

Thanks @EZoni ! I don't know why that CI test started failing here - I can only guess that it was chance. I didn't dig into finding exactly what was going wrong.

PR #5666, opened 2025-02-13
WarpX class: move AllocateCenteringCoefficients to anonymous namespace in WarpX.cpp

lucafedeli88:
`AllocateCenteringCoefficients` is a pure function used only inside `WarpX.cpp`. This PR moves it from a private member function of the WarpX class to an anonymous namespace inside `WarpX.cpp`.  The PR also makes the function more compact.
This is done to simplify the `WarpX.H` header.


EZoni:
Thanks, @lucafedeli88! Please note that there is one conflict to resolve.

lucafedeli88:
> Thanks, @lucafedeli88! Please note that there is one conflict to resolve.

You are right, @EZoni ! It should be fixed now!

lucafedeli88:
Sorry, **now** it should be fixed!

EZoni (approved):

PR #5667, opened 2025-02-13
Add PECInsulator to Curl-Curl BC

dpgrote:
This is a temporary fix, setting a boundary condition for the Curl-Curl preconditioner for the implicit solver. This now sets the BC to Dirichlet for the PEC regions. A correct solution would have to be implemented in the curl-curl solver because of the split between the PEC and insulator sections.

### DIFF
--- a/Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
+++ b/Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
@@ -70,7 +70,7 @@ Array<LinOpBCType,AMREX_SPACEDIM> ImplicitSolver::convertFieldBCToLinOpBC (const
             // Also for FieldBoundaryType::PMC
             lbc[i] = LinOpBCType::Neumann;
         } else if (a_fbc[i] == FieldBoundaryType::PECInsulator) {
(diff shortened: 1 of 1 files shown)


Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
```diff
@@ -69,6 +69,8 @@ Array<LinOpBCType,AMREX_SPACEDIM> ImplicitSolver::convertFieldBCToLinOpBC (const
         } else if (a_fbc[i] == FieldBoundaryType::Neumann) {
             // Also for FieldBoundaryType::PMC
             lbc[i] = LinOpBCType::Neumann;
+        } else if (a_fbc[i] == FieldBoundaryType::PECInsulator) {
+            lbc[i] = LinOpBCType::Dirichlet;
```
lucafedeli88: Maybe we could raise a warning here, since this is a temporary fix. 
Something like:
```
    "Implicit solver",
    YOUR_WARNING_MESSAGE,
    ablastr::warn_manager::WarnPriority::medium);
```
What do you think?

lucafedeli88 (approved):
Thanks @dpgrote ! LGTM

PR #5670, opened 2025-02-14
WarpX class: remove declaration of two unimplemented functions

lucafedeli88:
`AverageAndPackFields` and `prepareFields` are not implemented. Therefore, this PR removes their declaration from the WarpX header.


ax3l (approved):
Good catch. I think this moved to diagnostics a while ago.

PR #5683, opened 2025-02-19
Small fix in Perlmutter GPU sbatch script 

aeriforme:
Changes in Perlmutter GPU job script:
from `#SBATCH --cpus-per-task=16` to `#SBATCH --cpus-per-task=32`.

This is to request (v)cores in consecutive blocks.

GPU 3 is closest to CPU cores 0-15, 64-79, 
GPU 2 to CPU cores 16-31, 80-95, 
...

If `--cpus-per-task=16`, MPI ranks 0 and 1 are mapped to cores 0 and 8.
If `--cpus-per-task=32`, MPI ranks 0 and 1 are mapped to cores 0 and 16. 

Visual representation
pm_gpu_vcores_mpi

@ax3l @WeiqunZhang is this correct?


WeiqunZhang:
I think `export OMP_NUM_THREADS=${SLURM_CPUS_PER_TASK}` should be updated too, because `SLURM_CPUS_PER_TASK` is 32 now. We might want to simply set it to 16.

ax3l (approved):

PR #5691, opened 2025-02-20
[WIP] FieldReduction.H : make volume calculation more compact using std::reduce over dx vector

lucafedeli88:
This PR makes the calculation of the `cell_volume` here more readable.

### DIFF
--- a/Source/Diagnostics/ReducedDiags/FieldReduction.H
+++ b/Source/Diagnostics/ReducedDiags/FieldReduction.H
@@ -33,6 +33,7 @@
 #include <AMReX_Tuple.H>
 
 #include <memory>
+#include <numeric>
 #include <string>
 #include <type_traits>
 #include <vector>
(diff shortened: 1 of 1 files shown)


Source/Diagnostics/ReducedDiags/FieldReduction.H
```diff
@@ -225,15 +226,10 @@ public:
...
-#elif defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
-            reduce_value *= dx[0]*dx[1];
-#else
-            reduce_value *= dx[0]*dx[1]*dx[2];
-#endif
+            // If reduction operation is a sum, multiply the value by the cell volume so that the
+            // result is the integral of the function over the simulation domain.
+            const auto cell_volume = std::reduce(std::begin(dx), std::end(dx), 1.0_rt, std::multiplies<>());
+            reduce_value *= cell_volume;
         }
```
dpgrote, line 233: I've done things like this, which is also compact
```
reduce_value *= AMREX_D_TERM(dx[0], *dx[1], *dx[2]);
```
In fact, I already make this change in PR #5189.

lucafedeli88: Oh, I see... sorry! Then, this PR has no reason to exist. I will close it immediately!

dpgrote:
@lucafedeli88 This PR could submit the small change. This would help reduce the size of PR #5189 a little, as it is a very large PR.

PR #5692, opened 2025-02-21
Adding hyper-resistivity expression, function of rho, B

mattobin:
Adding capability to define the hyper-resistivity as a function of the charge density and magnetic field strength.


Python/pywarpx/picmi.py
```diff
@@ -1951,7 +1951,14 @@ def solver_initialize_inputs(self):
                 self.plasma_resistivity, self.mangle_dict
             ),
         )
-        pywarpx.hybridpicmodel.plasma_hyper_resistivity = self.plasma_hyper_resistivity
+        # Enabling hyperresistivity expression - mtobin 20240218
+        # pywarpx.hybridpicmodel.plasma_hyper_resistivity = self.plasma_hyper_resistivity
```
```suggestion
```

Python/pywarpx/picmi.py
```diff
@@ -1951,7 +1951,14 @@ def solver_initialize_inputs(self):
...
+        pywarpx.hybridpicmodel.__setattr__(
+            "plasma_hyperresistivity(rho,B)",
+            pywarpx.my_constants.mangle_expression(
+                self.plasma_hyperresistivity, self.mangle_dict
```
```suggestion
                self.plasma_hyper_resistivity, self.mangle_dict
```

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -50,7 +50,9 @@ void HybridPICModel::ReadParameters ()
     pp_hybrid.query("plasma_resistivity(rho,J)", m_eta_expression);
     utils::parser::queryWithParser(pp_hybrid, "n_floor", m_n_floor);
 
-    utils::parser::queryWithParser(pp_hybrid, "plasma_hyper_resistivity", m_eta_h);
+    // Trying to add hyper-resistivity expression -  mtobin 20240217
+    // utils::parser::queryWithParser(pp_hybrid, "plasma_hyper_resistivity", m_eta_h);
+    pp_hybrid.query("plasma_hyper_resistivity(rho,B)", m_eta_h_expression);
```
roelof-groenewald, line 51: Can you please move this line up to be just below the line for the resistivity (line 50). Since these are now following the same pattern it would make sense to group them closer together.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.H
```diff
@@ -194,7 +194,12 @@ public:
     bool m_resistivity_has_J_dependence = false;
 
     /** Plasma hyper-resisitivity */
-    amrex::Real m_eta_h = 0.0;
+    /** Expanded mtobin 20240218 */
+    // amrex::Real m_eta_h = 0.0;
+    std::string m_eta_h_expression = "0.0";
+    std::unique_ptr<amrex::Parser> m_hyper_resistivity_parser;
+    amrex::ParserExecutor<2> m_eta_h;
+    bool m_hyper_resistivity_has_B_dependence = false;
```
roelof-groenewald, line 201: Let's include another boolean flag here, `m_include_hyper_resistivity_term`, which we can use to track whether the hyper-resistivity should be included at all (since it requires a bit of extra calculation which would be good to avoid if possible).

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -428,8 +428,11 @@ void FiniteDifferenceSolver::HybridPICSolveECylindrical (
     const auto eta_h = hybrid_model->m_eta_h;
     const auto rho_floor = hybrid_model->m_n_floor * PhysConst::q_e;
     const auto resistivity_has_J_dependence = hybrid_model->m_resistivity_has_J_dependence;
+    // Adding hyper-resistivity expression - mtobin 20240217
+    const auto hyper_resistivity_has_B_dependence = hybrid_model->m_hyper_resistivity_has_B_dependence;
 
-    const bool include_hyper_resistivity_term = (eta_h > 0.0) && solve_for_Faraday;
+    // const bool include_hyper_resistivity_term = (eta_h > 0.0) && solve_for_Faraday;
+    const bool include_hyper_resistivity_term = solve_for_Faraday;
```
roelof-groenewald: With the above changes this can become:
```suggestion
    const bool include_hyper_resistivity_term = hybrid_model->m_include_hyper_resistivity_term;
```

Python/pywarpx/picmi.py
roelof-groenewald: The git blame tracks which changes were added by which user, so you can remove these comments.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -160,6 +162,13 @@ void HybridPICModel::InitData ()
     const std::set<std::string> resistivity_symbols = m_resistivity_parser->symbols();
     m_resistivity_has_J_dependence += resistivity_symbols.count("J");
 
+    // Adding hyper-resistivity expression - mtobin 20240217
+    m_hyper_resistivity_parser = std::make_unique<amrex::Parser>(
```
roelof-groenewald, line 164: To track whether to use hyper-resistivity, here we could do:
```suggestion
    m_include_hyper_resistivity_term = (m_eta_h_expression != "0.0");
    m_hyper_resistivity_parser = std::make_unique<amrex::Parser>(
```

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -811,8 +856,11 @@ void FiniteDifferenceSolver::HybridPICSolveECartesian (
     const auto eta_h = hybrid_model->m_eta_h;
     const auto rho_floor = hybrid_model->m_n_floor * PhysConst::q_e;
     const auto resistivity_has_J_dependence = hybrid_model->m_resistivity_has_J_dependence;
+    // For hyper-resistivity expression - mtobin 20240217
+    const auto hyper_resistivity_has_B_dependence = hybrid_model->m_hyper_resistivity_has_B_dependence;
 
-    const bool include_hyper_resistivity_term = (eta_h > 0.) && solve_for_Faraday;
+    // const bool include_hyper_resistivity_term = (eta_h > 0.) && solve_for_Faraday;
+    const bool include_hyper_resistivity_term = solve_for_Faraday;
```
roelof-groenewald: Similar comment here as above.

roelof-groenewald (changes requested):
These changes look great! I just left a couple of comments for small improvements.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.H
roelof-groenewald:
```suggestion
    bool m_include_hyper_resistivity_term = false;
    bool m_hyper_resistivity_has_B_dependence = false;
```

Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp
```diff
@@ -1030,10 +1082,22 @@ void FiniteDifferenceSolver::HybridPICSolveECartesian (
                     Ex(i, j, k) += eta(rho_val, jtot_val) * Jx(i, j, k);
 
                     if (include_hyper_resistivity_term) {
+
+                        // Interpolate B field to appropriate staggering to match E field - mtobin 20240217
+                        Real btot_val = 0._rt;
+                        if (solve_for_Faraday && hyper_resistivity_has_B_dependence) {
```
roelof-groenewald: These could just be:
```suggestion
                        if (hyper_resistivity_has_B_dependence) {
```
since this whole block of code is inside a `if (solve_for_Faraday)` condition.

Python/pywarpx/picmi.py
```diff
@@ -1951,7 +1951,14 @@ def solver_initialize_inputs(self):
                 self.plasma_resistivity, self.mangle_dict
             ),
         )
-        pywarpx.hybridpicmodel.plasma_hyper_resistivity = self.plasma_hyper_resistivity
+        # Enabling hyperresistivity expression - mtobin 20240218
+        # pywarpx.hybridpicmodel.plasma_hyper_resistivity = self.plasma_hyper_resistivity
+        pywarpx.hybridpicmodel.__setattr__(
+            "plasma_hyperresistivity(rho,B)",
```
```suggestion
            "plasma_hyper_resistivity(rho,B)",
```

Python/pywarpx/picmi.py
mattobin: Yeah this is more for my benefit as I'm going through to see where I made changes in case I need to fix things- I'll remove these comments before the merge request

roelof-groenewald (commented):
These changes look good to me! Thanks for the updates.
Once you remove the comments for the old code, we can merge :tada:

mattobin:
@roelof-groenewald ok, I think ready to merge!

roelof-groenewald (approved):
Thanks so much for these changes!

PR #5700, opened 2025-02-24
Add Villasenor and Buneman current deposition with explicit scheme

dpgrote:
The Villasenor and Buneman current deposition was moved to a separate kernel routine and added implicit and explicit callers.

A test case is a 2D uniform plasma, run for a number of plasma periods. The images below show the relative change in the energgy for the parts of the system. The total energy is conversed to roughly 1.e-5. The results are shown for both the Villasenor and the Esirkepov current deposition. The results are essentially the same, with the differences in the two cases below the resolution of the images.
energy_explicit_VandB
energy_explicit_Esirkepov


JustinRayAngus:
This looks good Dave.

Have you tried running with explicit yet? There is a WARPX_ALWAYS_ASSERT_WITH_MESSAGE in WarpX.cpp that prevents using Villasenor with explicit. That can be removed.

JustinRayAngus:
We should also update the docs to reflect this PR.

https://warpx.readthedocs.io/en/latest/usage/parameters.html#particle-push-charge-and-current-deposition-field-gathering

Docs/source/usage/parameters.rst
JustinRayAngus, line 1: Need a minor update here on line 2257 as well to reflect villasenor option below.

JustinRayAngus (approved):
Looks good to me.

Thanks Dave.

dpgrote:
@ax3l You requested a self-review. Did you still want to review this PR?

PR #5702, opened 2025-02-24
Add `WarpX_UNITY_BUILD`

ax3l:
Unity builds combine all `.cpp` files into a single one (a single translation unit, TU). This is useful for performance optimization, computer science experiments, or differentiable programming.

FYI: A CPU unity build of ImpactX is around 18sec on my laptop (https://github.com/BLAST-ImpactX/impactx/pull/861), for WarpX around 3:30min. In both cases, this refers to the `lib*` of ImpactX/WarpX part.


lucafedeli88:
Uh! We have to be careful with anonymous namespaces if we use unity builds!

EZoni:
@ax3l @lucafedeli88 

This comment by Luca,

> Uh! We have to be careful with anonymous namespaces if we use unity builds!

sounds like something we need to discuss properly before we proceed with moving more functions to anonymous namespaces. At this time I don't have enough knowledge of unity builds to have an informed understanding of Luca's statement, but if you do, it is probably a good idea to touch base on this and see what is the best way to move forward.

ax3l:
To clarify: anonymous namespaces are _not_ the problem, identically named global variables repeated with the same name in different `.cpp` files are what one wants to avoid (in or outside global namespaces).

lucafedeli88:
> To clarify: anonymous namespaces are _not_ the problem, identically named global variables repeated with the same name in different `.cpp` files are what one wants to avoid (in or outside global namespaces).

Yes, sorry, I should have explained that better.... My point is that we are using anonymous namespaces more often than before, so the risk of conflicts may increase in the future. Maybe we could consider *not* using anonymous namespaces, but `parent_namespace::details` in the `.cpp` file to reduce the risk of conflicts. What  do you think?

ax3l:
That is possible for sure and clean :)

Anonymous namespaces are generally still great to use, because what they do is they do not create an unnecessary symbol in the final binary and thus keep the binary a bit smaller. (That said, we can use compiler symbol hiding flags when we want to optimize for that, we do that in pybind11 for non-debug builds.)

Another solution I could have done for this one collision here: move the global variable to a header file (in a regular namespace) and define it `inline` or `static`.

Source/Diagnostics/ReducedDiags/DifferentialLuminosity2D.cpp
aeriforme, line 1: We have updated this file in PR #5726. 
If you merge `development`, there would be a few additional synchronizations `Gpu::streamSynchronize();`, which I guess need to be updated into `amrex:: Gpu::streamSynchronize();`

ax3l: Thanks! That will be automatically merged and does not create a merge conflict, as far as I can see.

ax3l:
Let me know if you have any further questions, but this is otherwise ready to go. cc @EZoni

EZoni (commented):
Should we add a CI job that builds WarpX with `-DWarpX_UNITY_BUILD=ON`?

ax3l:
Yes, def!

I added one in ImpactX already but was refraining from one just yet here because it still threw a few (mild) warnings. Would do in a follow-up where I fix those warnings, so we can build the unity build also with `-Werror`.

EZoni:
I restarted the `cartesian_3d` tests since there was a single failure that appeared to be random.

EZoni (approved):

EZoni:
I merge `development` since the latest commit here is a couple of weeks old and then enable auto-merge, so that this work is ready to be picked up again when needed.

PR #5707, opened 2025-02-25
Fix HIP CI

atmyers:
- [x] Update HIP CI workflow to run on Ubuntu 24.04
- [x] Add HIP version to CLI, default 6.3.2
- [x] Fix bug caused by using `std::optional` on device

### DIFF
--- a/Source/Particles/Gather/GetExternalFields.cpp
+++ b/Source/Particles/Gather/GetExternalFields.cpp
@@ -22,7 +22,7 @@ GetExternalEBField::GetExternalEBField (const WarpXParIter& a_pti, long a_offset
 
     AcceleratorLattice const & accelerator_lattice = warpx.get_accelerator_lattice(lev);
     if (accelerator_lattice.m_lattice_defined) {
-        *d_lattice_element_finder = accelerator_lattice.GetFinderDeviceInstance(a_pti, static_cast<int>(a_offset));
+        d_lattice_element_finder = accelerator_lattice.GetFinderDeviceInstance(a_pti, static_cast<int>(a_offset));
     }
 
     m_gamma_boost = WarpX::gamma_boost;
--- a/Source/Particles/Gather/GetExternalFields.cpp
+++ b/Source/Particles/Gather/GetExternalFields.cpp
@@ -22,7 +22,7 @@ GetExternalEBField::GetExternalEBField (const WarpXParIter& a_pti, long a_offset
 
     AcceleratorLattice const & accelerator_lattice = warpx.get_accelerator_lattice(lev);
     if (accelerator_lattice.m_lattice_defined) {
-        d_lattice_element_finder = accelerator_lattice.GetFinderDeviceInstance(a_pti, static_cast<int>(a_offset));
+        *d_lattice_element_finder = accelerator_lattice.GetFinderDeviceInstance(a_pti, static_cast<int>(a_offset));
     }
 
     m_gamma_boost = WarpX::gamma_boost;
--- a/.github/workflows/dependencies/hip.sh
+++ b/.github/workflows/dependencies/hip.sh
@@ -28,7 +28,9 @@ sudo apt-key add rocm.gpg.key
 
 source /etc/os-release # set UBUNTU_CODENAME: focal or jammy or ...
 
-echo "deb [arch=amd64] https://repo.radeon.com/rocm/apt/${1-latest} ${UBUNTU_CODENAME} main" \
+VERSION=${1-6.3.2}
+
+echo "deb [arch=amd64] https://repo.radeon.com/rocm/apt/${VERSION} ${UBUNTU_CODENAME} main" \
   | sudo tee /etc/apt/sources.list.d/rocm.list
 echo 'export PATH=/opt/rocm/llvm/bin:/opt/rocm/bin:/opt/rocm/profiler/bin:/opt/rocm/opencl/bin:$PATH' \
   | sudo tee -a /etc/profile.d/rocm.sh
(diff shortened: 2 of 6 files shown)


EZoni:
Thank you for opening this PR, @atmyers!

This fixed the dependencies issue.

Now it uncovered this build error:
```
In file included from /home/runner/work/WarpX/WarpX/Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp:11:
In file included from /home/runner/work/WarpX/WarpX/Source/Fields.H:11:
In file included from /home/runner/work/WarpX/WarpX/Source/ablastr/fields/MultiFabRegister.H:23:
/usr/lib/gcc/x86_64-linux-gnu/14/../../../../include/c++/14/optional:482:2: error: reference to __host__ function '__glibcxx_assert_fail' in __host__ __device__ function
  482 |         __glibcxx_assert(this->_M_is_engaged());
      |         ^
/usr/lib/gcc/x86_64-linux-gnu/14/../../../../include/x86_64-linux-gnu/c++/14/bits/c++config.h:612:7: note: expanded from macro '__glibcxx_assert'
  612 |         std::__glibcxx_assert_fail();                                   \
      |              ^
/usr/lib/gcc/x86_64-linux-gnu/14/../../../../include/c++/14/optional:968:22: note: called by 'operator*'
  968 |       { return this->_M_get(); }
      |                      ^
/home/runner/work/WarpX/WarpX/Source/Particles/Gather/GetExternalFields.H:77:14: note: called by 'operator()'
   77 |             (*d_lattice_element_finder)(i, field_Ex, field_Ey, field_Ez,
      |              ^
/home/runner/work/WarpX/WarpX/Source/Diagnostics/ReducedDiags/ColliderRelevant.cpp:511:34: note: called by 'operator()'
  511 |                     getExternalEB(i, ex, ey, ez, bx, by, bz);
      |                                  ^
/home/runner/work/WarpX/WarpX/build_2d/_deps/fetchedamrex-src/Src/Base/AMReX_Reduce.H:647:27: note: called by 'operator()'
  647 |                 auto pr = f(i);
      |                           ^
/home/runner/work/WarpX/WarpX/build_2d/_deps/fetchedamrex-src/Src/Base/AMReX_GpuLaunchFunctsG.H:722:60: note: called by 'operator()'
  722 |                         [=] AMREX_GPU_DEVICE () noexcept { f(); });
      |                                                            ^
/home/runner/work/WarpX/WarpX/build_2d/_deps/fetchedamrex-src/Src/Base/AMReX_GpuLaunchGlobal.H:21:50: note: called by 'launch_global<256, (lambda at /home/runner/work/WarpX/WarpX/build_2d/_deps/fetchedamrex-src/Src/Base/AMReX_GpuLaunchFunctsG.H:721:5)>'
   21 |     AMREX_GPU_GLOBAL void launch_global (L f0) { f0(); }
      |                                                  ^
/usr/lib/gcc/x86_64-linux-gnu/14/../../../../include/x86_64-linux-gnu/c++/14/bits/c++config.h:605:3: note: '__glibcxx_assert_fail' declared here
  605 |   __glibcxx_assert_fail()
      |   ^
1 error generated when compiling for gfx900.
```

ax3l:
Looks like HIP does not support `std::optional<LatticeElementFinderDevice> d_lattice_element_finder;`  on device. We could make this a pointer for the `operator()`. cc @dpgrote

dpgrote:
> Looks like HIP does not support `std::optional<LatticeElementFinderDevice> d_lattice_element_finder;` on device. We could make this a pointer for the `operator()`. cc @dpgrote

This is apparently related to bug in the recent versions of gcc and/or libstdc++11, though it looks like it has been fixed in the latest version. One place reports this problem occurring with the use of `optional`. The solution would be to use a pointer for the `d_lattice_element_finder` attribute instead of `optional`.

EZoni:
FYI, trying @ax3l's and @dpgrote's suggestion in [c09e0f2](https://github.com/ECP-WarpX/WarpX/pull/5707/commits/c09e0f2e28cb92a37c4d38437acf9f9182e07adc).

Source/Particles/Gather/GetExternalFields.cpp
```diff
@@ -22,7 +22,7 @@ GetExternalEBField::GetExternalEBField (const WarpXParIter& a_pti, long a_offset
 
     AcceleratorLattice const & accelerator_lattice = warpx.get_accelerator_lattice(lev);
     if (accelerator_lattice.m_lattice_defined) {
-        d_lattice_element_finder = accelerator_lattice.GetFinderDeviceInstance(a_pti, static_cast<int>(a_offset));
+        *d_lattice_element_finder = accelerator_lattice.GetFinderDeviceInstance(a_pti, static_cast<int>(a_offset));
```
dpgrote: Thanks for these changes! In completeness, does there need to be something to delete the instances created here? Could this be a `unique_ptr`(which should take care of the deletion)?

EZoni: Initially I tried with `unique_ptr`, but several compiler errors appeared wherever we were calling `GetExternalEBField` inside a lambda function.

The compiler was complaining that we were trying to use a deleted function: `error: use of deleted function`. Probably because the capture by value `[=]` attempts a copy, although I would not be 100% sure how to fix that. I also suspect that this may be why the other pointers in this same file are also raw pointers rather than `unique_ptr`.

EZoni: @dpgrote 

Do you see what could be causing the CI segfaults in the current implementation? Where should the allocation of the raw pointer be?

dpgrote: It looks like the pointer can't work since the object is passed to the lambda. I can push a commit that I think should solve this. I would do `LatticeElementFinderDevice d_lattice_element_finder`, and add an internal flag to `LatticeElementFinderDevice` signaling whether it has been initialized.

EZoni (approved):

PR #5711, opened 2025-02-26
Add Python function to extract particles scraped during the last step.

RemiLehe:
This adds a function that extracts the particles that were scraped at the current timestep. 

This is useful in callback functions, where we often want to re-inject particles that hit the boundary, and therefore need to select the ones that were scraped at the current timestep.

This also avoids calling `clear_buffer`, which potentially interferes with the `BoundaryScrapingDiagnostic`

### DIFF
--- a/Python/pywarpx/particle_containers.py
+++ b/Python/pywarpx/particle_containers.py
@@ -866,7 +866,8 @@ class ParticleBoundaryBufferWrapper(object):
     def get_particle_scraped_this_step(self, species_name, boundary, comp_name, level):
         """
         This returns a list of numpy or cupy arrays containing the particle array data
-        for a species that has been scraped by a specific simulation boundary.
+        for particles that have been scraped at the current timestep, 
+        for a specific species and simulation boundary.
 
         The data for the arrays is a view of the underlying boundary buffer in WarpX ;
         writing to these arrays will therefore also modify the underlying boundary buffer.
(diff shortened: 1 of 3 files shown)


Python/pywarpx/particle_containers.py
```diff
@@ -863,6 +863,50 @@ def get_particle_boundary_buffer(self, species_name, boundary, comp_name, level)
             raise RuntimeError("Name %s not found" % comp_name)
         return data_array
 
+    def get_particle_scraped_this_step(self, species_name, boundary, comp_name, level):
+        """
+        This returns a list of numpy or cupy arrays containing the particle array data
+        for a species that has been scraped by a specific simulation boundary.
```
RemiLehe:
```suggestion
        for particles that have been scraped at the current timestep, 
        for a specific species and simulation boundary.
```

Python/pywarpx/particle_containers.py
```diff
@@ -863,6 +863,51 @@ def get_particle_boundary_buffer(self, species_name, boundary, comp_name, level)
...
+        )
+        step_scraped_array = self.get_particle_boundary_buffer(
+            species_name, boundary, "stepScraped", level
+        )
+
+        # Select on the particles from the previous step
+        data_array_over_previous_step = []
+        for data, step in zip(data_array, step_scraped_array):
+            data_array_over_previous_step.append(data[step == current_step])
+        return data_array_over_previous_step
```
RemiLehe:
```suggestion
        data_array_this_step = []
        for data, step in zip(data_array, step_scraped_array):
            data_array_this_step.append(data[step == current_step])
        return data_array_this_step
```

RevathiJambunathan (approved):
Thank you @RemiLehe for this PR!!!
I tested it and it works!

dpgrote (approved):
This looks good, thanks!
A minor change would be to have the level default to 0 which is likely the most common case.

PR #5715, opened 2025-02-27
[WIP] Add secondary electron emission

budjensen:
Adding a simple model of secondary electron emission, as per my message in https://github.com/ECP-WarpX/WarpX/discussions/5708. More work is still be done.

All recommendations and any feedback is appreciated!


budjensen:
I closed my initial discussion on SEE, but copied my original comment below:

> I would like to add secondary electron emission to WarpX. Could you point me in the best way to get started on this?
> 
> I would like to build the feature so that in the future that we can incorporate both energy and angular dependence to the secondary emission coefficient, but for speed I'd prefer to implement a basic version which accepts a coefficient, $\gamma_\mathrm{SEE}$, for each species. 
> 
> SEE emission would be determined by drawing a random number $R\in[0,1]$ and adding a secondary electron if $R>\gamma_\mathrm{SEE}$ (If $\gamma_\mathrm{SEE} > 1$, floor($\gamma_\mathrm{SEE})$ particles would be added before comparing R and $\gamma_\mathrm{SEE}$). All SEE electrons would have a pre-specified energy and be isotropically scattered in velocity.

So far, I have added a flag `do_SEE_flag` to the function `WarpXParticleContainer::ApplyBoundaryConditions` which initiates the function `do_SEE` when a particle crosses a boundary with absorbing boundary conditions. `do_SEE` calculates the number of electrons to emit, and then draws their ejection angle(s) from an isotropic distribution.

I've now realized that I need to inject electrons, which seems simple enough when inside the electron particle container instance. Is there a way to do this from a pc for a different species?

PR #5716, opened 2025-02-28
Extending preconditioners to handle a diagonal operator for E-field

debog:
- Extended the `CurlCurlMLMGPC` interface to work with non-scalar `beta` coefficient. It can now be a `MultiFab` array (same size/layout as E-Fields) with one component representing a diagonal matrix/operator on the E-field.

- Implemented a `JacobiPC` preconditioner to solve the system `\sigma E = b` using the Jacobi iterations. For now, `\sigma` is a diagonal matrix (stored as a `MultiFab` array (same size/layout as E-Fields) with one component. The solution is set as `E = b / \sigma`, so the iterations are not yet implemented. The complete Jacobi method will be implemented once non-diagonal elements of `\sigma` are available.

The preconditioners fetch a pointer to the matrix operator (`beta` or `\sigma`) from the implicit time integrator object.


Source/NonlinearSolvers/JacobiPC.H
```diff
@@ -0,0 +1,246 @@
...
+        a_b.getArrayVecType()==warpx::fields::FieldType::Efield_fp,
+        "JacobiPC::Apply() - a_b must be Efield_fp type");
+
+    // Get the data vectors
+    auto& b_mfarrvec = a_b.getArrayVec();
+    auto& x_mfarrvec = a_x.getArrayVec();
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+        ((b_mfarrvec.size() == m_num_amr_levels) && (x_mfarrvec.size() == m_num_amr_levels)),
+        "Error in JacobiPC::Apply() - mismatch in number of levels." );
+
```
JustinRayAngus: This method doesn't currently handle the situation where m_bcoefs is not set.

```suggestion
    // Do nothing if m_mcoefs is not set
    if (m_bcoefs == nullptr) {
        a_x.Copy(a_b);
        return;
    }

debog: @JustinRayAngus "Do nothing" means not changing the value of `a_x`; setting it to `a_b` implies an identity matrix when `m_bcoefs` is not set.

Which one do you prefer? I prefer the former.

JustinRayAngus: "Do nothing" is the wrong comment. This function is called from JacobianFunctionMF::precond(). It looks to me like the correct thing to do is the latter, which is consistent with what precond() does if a preconditioner is not used.

JustinRayAngus (approved):
Looks good. Thanks Debo!

PR #5718, opened 2025-02-28
WarpX class: sort_particles_for_deposition and sort_idx_type no longer static

lucafedeli88:
This PR turns the public static variables of the WarpX class `sort_particles_for_deposition` and `sort_idx_type` into private member variables. The final goal is to avoid static variables in the WarpX class.

### DIFF
--- a/Source/Particles/MultiParticleContainer.H
+++ b/Source/Particles/MultiParticleContainer.H
@@ -211,7 +211,10 @@ public:
 
     void WriteHeader (std::ostream& os) const;
 
-    void SortParticlesByBin (amrex::IntVect bin_size);
+    void SortParticlesByBin (
+        const amrex::IntVect& bin_size,
+        bool sort_particles_for_deposition,
+        const amrex::IntVect& sort_idx_type);
 
     void Redistribute ();
 
--- a/Source/WarpX.H
+++ b/Source/WarpX.H
@@ -352,11 +352,6 @@ public:
     static utils::parser::IntervalsParser sort_intervals;
     static amrex::IntVect sort_bin_size;
 
-    //! If true, particles will be sorted in the order x -> y -> z -> ppc for faster deposition
-    static bool sort_particles_for_deposition;
-    //! Specifies the type of grid used for the above sorting, i.e. cell-centered, nodal, or mixed
-    static amrex::IntVect sort_idx_type;
-
     static bool do_multi_J;
     static int do_multi_J_n_depositions;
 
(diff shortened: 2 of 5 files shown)


Source/WarpX.H
```diff
@@ -1489,6 +1484,18 @@ private:
...
+    //! If true, particles will be sorted in the order x -> y -> z -> ppc for faster deposition
+#if defined(AMREX_USE_CUDA)
+    bool sort_particles_for_deposition = true;
+#else
+    bool sort_particles_for_deposition = false;
+#endif
+
+    //! Specifies the type of grid used for the above sorting, i.e. cell-centered, nodal, or mixed
+    amrex::IntVect sort_idx_type = amrex::IntVect(AMREX_D_DECL(0,0,0));
+
```
EZoni: Should we prefix these private variables with `m_` since they are still members of a class?

lucafedeli88: Yes, I think that you are right.

lucafedeli88: Done, @EZoni !

Source/Particles/MultiParticleContainer.H
```diff
@@ -211,7 +211,10 @@ public:
 
     void WriteHeader (std::ostream& os) const;
 
-    void SortParticlesByBin (amrex::IntVect bin_size);
+    void SortParticlesByBin (
+        const amrex::IntVect& bin_size,
+        bool sort_particles_for_deposition,
```
EZoni, line 216: Should this be `const` like in the .cpp file, or is that something that clang-tidy does not like for some reason (I can't remember)?
```suggestion
        const bool sort_particles_for_deposition,
```

lucafedeli88: You can put `const` in the header file as well (the code will compile and I am almost sure that `clang-tidy` won't complain). However, from the point of view of someone calling the function, it doesn't change anything, since the variable is passed by value. The fact that you don't change the value of `sort_particles_for_deposition` inside the function can be considered an implementation detail. That's why I put `const` only in the cpp file.  We can, of course, decide to use a different convention! There's some debate on this topic on stackoverflow, e.g. [here](https://stackoverflow.com/questions/1554750/is-const-necessary-in-function-parameters-when-passing-by-value)

EZoni: Okay, that's fine by me. I understand that it does not make any difference, but I thought it could potentially increase "transparency", especially if it doesn't have negative side effects. In some sense, I was asking myself "why not?" instead of "why?". But I'm okay with leaving it only in the implementation.

EZoni (approved):

PR #5726, opened 2025-03-04
Synchronize after copying in 2D reduced diagnostics

aeriforme:
Add `Gpu::streamSynchronize();` after copying `TableData` from either host-to-device and device-to-host in the `ParticleHistogram2D` and `DifferentialLuminosity2D` reduced diagnostics. 

We have observed unstable behaviors with the `DifferentialLuminosity2D` diagnostic and @RemiLehe suggested this could be a reason. 

Marked as bug, unless otherwise recommended.

### DIFF
--- a/Source/Diagnostics/ReducedDiags/ParticleHistogram2D.cpp
+++ b/Source/Diagnostics/ReducedDiags/ParticleHistogram2D.cpp
@@ -149,7 +149,7 @@ void ParticleHistogram2D::ComputeDiags (int step)
     }
 
     d_data_2D.copy(m_h_data_2D);
-    Gpu::streamSynchronize();
+    // Gpu::streamSynchronize() is not needed, because there is a sync in WarpXParIter constructor later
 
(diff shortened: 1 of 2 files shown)


Source/Diagnostics/ReducedDiags/ParticleHistogram2D.cpp
```diff
@@ -149,6 +149,8 @@ void ParticleHistogram2D::ComputeDiags (int step)
     }
 
     d_data_2D.copy(m_h_data_2D);
+    Gpu::streamSynchronize();
```
WeiqunZhang: This sync is not needed, because there is a sync in WarpXParIter constructor later.

RemiLehe:
```suggestion
    // Gpu::streamSynchronize() is not needed, because there is a sync in WarpXParIter constructor later
```

RemiLehe (approved):

PR #5739, opened 2025-03-07
[WIP] Apply clang-format to WarpX.H, WarpX.cpp

EZoni:
Following up on #5687, this PR applies the current `clang-format` style, as is, to two more files, WarpX.H and WarpX.cpp. The code changes were implemented automatically by running `pre-commit run -a` from the root directory.

This should be enough to showcase most of the style changes that this `clang-format` configuration (LLVM's default, with few minor adjustments introduced in #5687) produces.

The idea would be for reviewers  to briefly go through the changes and raise objections, if any, which we could try to mitigate by adjusting the clang-format configuration further.

Of course it's still possible to object the whole idea of automatic style formatting, too.

### DIFF
--- a/Source/WarpX.H
+++ b/Source/WarpX.H
@@ -19,31 +19,31 @@
 #include "EmbeddedBoundary/WarpXFaceInfoBox_fwd.H"
 #include "FieldSolver/ElectrostaticSolvers/ElectrostaticSolver_fwd.H"
 #include "FieldSolver/FiniteDifferenceSolver/FiniteDifferenceSolver_fwd.H"
-#include "FieldSolver/FiniteDifferenceSolver/MacroscopicProperties/MacroscopicProperties_fwd.H"
 #include "FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel_fwd.H"
+#include "FieldSolver/FiniteDifferenceSolver/MacroscopicProperties/MacroscopicProperties_fwd.H"
 #include "Filter/NCIGodfreyFilter_fwd.H"
-#include "Initialization/ExternalField_fwd.H"
-#include "Particles/ParticleBoundaryBuffer_fwd.H"
-#include "Particles/MultiParticleContainer_fwd.H"
-#include "Particles/WarpXParticleContainer_fwd.H"
 #include "Fluids/MultiFluidContainer_fwd.H"
 #include "Fluids/WarpXFluidContainer_fwd.H"
+#include "Initialization/ExternalField_fwd.H"
+#include "Particles/MultiParticleContainer_fwd.H"
+#include "Particles/ParticleBoundaryBuffer_fwd.H"
+#include "Particles/WarpXParticleContainer_fwd.H"
 
 #ifdef WARPX_USE_FFT
-#   ifdef WARPX_DIM_RZ
-#       include "FieldSolver/SpectralSolver/SpectralSolverRZ_fwd.H"
-#       include "BoundaryConditions/PML_RZ_fwd.H"
-#   else
-#       include "FieldSolver/SpectralSolver/SpectralSolver_fwd.H"
-#   endif
+#ifdef WARPX_DIM_RZ
+#include "BoundaryConditions/PML_RZ_fwd.H"
+#include "FieldSolver/SpectralSolver/SpectralSolverRZ_fwd.H"
+#else
+#include "FieldSolver/SpectralSolver/SpectralSolver_fwd.H"
+#endif
 #endif
 #include "AcceleratorLattice/AcceleratorLattice.H"
 #include "Evolve/WarpXDtType.H"
 #include "Evolve/WarpXPushType.H"
-#include "Fields.H"
-#include "FieldSolver/MagnetostaticSolver/MagnetostaticSolver.H"
 #include "FieldSolver/ImplicitSolvers/ImplicitSolver.H"
 #include "FieldSolver/ImplicitSolvers/WarpXSolverVec.H"
+#include "FieldSolver/MagnetostaticSolver/MagnetostaticSolver.H"
+#include "Fields.H"
 #include "Filter/BilinearFilter.H"
 #include "Parallelization/GuardCellManager.H"
 #include "Utils/Parser/IntervalsParser.H"
@@ -58,7 +58,7 @@
 #include <AMReX_Array.H>
 #include <AMReX_Config.H>
 #ifdef AMREX_USE_EB
-#   include <AMReX_EBFabFactory.H>
+#include <AMReX_EBFabFactory.H>
 #endif
 #include <AMReX_GpuContainers.H>
 #include <AMReX_IntVect.H>
@@ -70,8 +70,8 @@
 #include <AMReX_Vector.H>
 #include <AMReX_VisMF.H>
 
-#include <AMReX_BaseFwd.H>
 #include <AMReX_AmrCoreFwd.H>
+#include <AMReX_BaseFwd.H>
 
 #include <array>
 #include <iostream>
@@ -82,45 +82,47 @@
 #include <string>
 #include <vector>
 
-class WARPX_EXPORT WarpX
-    : public amrex::AmrCore
-{
-public:
+class WARPX_EXPORT WarpX : public amrex::AmrCore {
+  public:
     static WarpX& GetInstance ();
 
     static void ResetInstance ();
 
     /**
      * \brief
-     * This method has to be called at the end of the simulation. It deletes the WarpX instance.
+     * This method has to be called at the end of the simulation. It deletes the
+     * WarpX instance.
      */
-    static void Finalize();
+    static void Finalize ();
 
     /** Destructor */
     ~WarpX () override;
 
     /** Copy constructor */
-    WarpX ( WarpX const &) = delete;
+    WarpX (WarpX const&) = delete;
     /** Copy operator */
-    WarpX& operator= ( WarpX const & ) = delete;
+    WarpX& operator=(WarpX const&) = delete;
 
     /** Move constructor */
-    WarpX ( WarpX && ) = delete;
+    WarpX (WarpX&&) = delete;
     /** Move operator */
-    WarpX& operator= ( WarpX && ) = delete;
+    WarpX& operator=(WarpX&&) = delete;
 
-    static std::string Version (); //!< Version of WarpX executable
+    static std::string Version ();       //!< Version of WarpX executable
     static std::string PicsarVersion (); //!< Version of PICSAR dependency
 
-    [[nodiscard]] int Verbose () const { return verbose; }
+    [[nodiscard]] int
+    Verbose () const {
+        return verbose;
+    }
 
-    [[nodiscard]] const amrex::Array<FieldBoundaryType,AMREX_SPACEDIM>& GetFieldBoundaryLo () const
-    {
+    [[nodiscard]] const amrex::Array<FieldBoundaryType, AMREX_SPACEDIM>&
+    GetFieldBoundaryLo () const {
         return field_boundary_lo;
     }
 
-    [[nodiscard]] const amrex::Array<FieldBoundaryType,AMREX_SPACEDIM>& GetFieldBoundaryHi () const
-    {
+    [[nodiscard]] const amrex::Array<FieldBoundaryType, AMREX_SPACEDIM>&
+    GetFieldBoundaryHi () const {
         return field_boundary_hi;
     }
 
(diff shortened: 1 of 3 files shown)


Source/WarpX.H
```diff
@@ -19,31 +19,31 @@
...
-#       include "BoundaryConditions/PML_RZ_fwd.H"
-#   else
-#       include "FieldSolver/SpectralSolver/SpectralSolver_fwd.H"
-#   endif
+#ifdef WARPX_DIM_RZ
+#include "BoundaryConditions/PML_RZ_fwd.H"
+#include "FieldSolver/SpectralSolver/SpectralSolverRZ_fwd.H"
+#else
+#include "FieldSolver/SpectralSolver/SpectralSolver_fwd.H"
+#endif
```
lucafedeli88, line 38: I am not completely sure about this style change. It seems to me that indentation was quite useful to make the code more readable here.

Source/WarpX.H
```diff
@@ -82,45 +82,47 @@
 #include <string>
 #include <vector>
 
-class WARPX_EXPORT WarpX
-    : public amrex::AmrCore
-{
-public:
+class WARPX_EXPORT WarpX : public amrex::AmrCore {
```
lucafedeli88, line 85: I am not completely sure about this style change. 
I personally prefer the style:

```
void function_name ()
{
}

class MyClass
{
};
```
for functions and classes, and 
```
if (/*..*/) {
// ...
}
```
for `if`, `while`, `for`...

Source/WarpX.H
```diff
@@ -82,45 +82,47 @@
...
 
     /** Move constructor */
-    WarpX ( WarpX && ) = delete;
+    WarpX (WarpX&&) = delete;
     /** Move operator */
-    WarpX& operator= ( WarpX && ) = delete;
+    WarpX& operator=(WarpX&&) = delete;
 
-    static std::string Version (); //!< Version of WarpX executable
+    static std::string Version ();       //!< Version of WarpX executable
```
lucafedeli88, line 111: Sometimes we use spaces to align comments and make them more readable. I don't have a strong opinion on this specific change, but we may want to discuss it.

Source/WarpX.H
```diff
@@ -136,112 +138,160 @@ public:
     //
     // Functions used by implicit solvers
     //
-    void ImplicitPreRHSOp ( amrex::Real cur_time,
-                            amrex::Real a_full_dt,
-                            int         a_nl_iter,
-                            bool        a_from_jacobian );
+    void ImplicitPreRHSOp (amrex::Real cur_time, amrex::Real a_full_dt,
```
lucafedeli88, line 141: Sometimes we use newlines in function declarations strategically to improve readability, e.g. :

```
void my_function (
    double Ex, double Ey, double Ez,
    double Bx, double By, double Bz,
    int i, int j, int k,
    bool flag);
```
I think that we may want to discuss if and how we would like `clang-format` to change this for us.

Source/WarpX.H
```diff
@@ -489,22 +577,28 @@ public:
     void UpdateInjectionPosition (amrex::Real dt);
 
     void ResetProbDomain (const amrex::RealBox& rb);
-    void EvolveE (         amrex::Real dt, amrex::Real start_time);
+    void EvolveE (amrex::Real dt, amrex::Real start_time);
```
lucafedeli88, line 580: I am not completely sure about this change. I think that the original version was slightly more readable thanks to the extra spaces.

Source/WarpX.H
```diff
@@ -861,28 +1059,33 @@ public:
      * \param[in] fy_parser parser function to initialize y-field
      * \param[in] fz_parser parser function to initialize z-field
      * \param[in] lev level of the Multifabs that is initialized
-     * \param[in] patch_type PatchType on which the field is initialized (fine or coarse)
-     * \param[in] eb_update_field flag indicating which gridpoints should be modified by this functions
-     * \param[in] use_eb_flags (default:true) flag indicating if eb points should be excluded or not
+     * \param[in] patch_type PatchType on which the field is initialized (fine
```
lucafedeli88, line 1062: While I like the idea of stetting a limit for the number of columns used for docstrings, I think that we may need an extra indentation here for the description of the input parameters.

lucafedeli88: Like:
```
\param[in] patch_type PatchType on which the field is initialized (fine
                    or coarse)
```

Source/WarpX.cpp
```diff
@@ -186,50 +187,54 @@ amrex::IntVect m_rho_nodal_flag;
...
-    {
-        constexpr auto is_pml = [](const FieldBoundaryType fbt) {return (fbt == FieldBoundaryType::PML);};
-        const auto is_any_pml =
-            std::any_of(field_boundary_lo.begin(), field_boundary_lo.end(), is_pml) ||
-            std::any_of(field_boundary_hi.begin(), field_boundary_hi.end(), is_pml);
-        return is_any_pml;
-    }
+namespace {
+
+[[nodiscard]] bool
```
lucafedeli88, line 192: I know that it is not recommended, but I actually like the indentation inside a namespace. We may want to discuss this.

Source/WarpX.cpp
```diff
@@ -2027,58 +2153,42 @@ WarpX::ClearLevel (int lev)
...
-        nox_fft, noy_fft, noz_fft,
-        NCIGodfreyFilter::m_stencil_width,
-        electromagnetic_solver_id,
-        maxLevel(),
-        WarpX::m_v_galilean,
-        WarpX::m_v_comoving,
-        m_safe_guard_cells,
-        WarpX::do_multi_J,
-        WarpX::fft_do_time_averaging,
+        dt[lev], dx, m_do_subcycling, WarpX::use_fdtd_nci_corr, grid_type,
```
lucafedeli88, line 2170: Here I think that the original version was more readable. The main issue, however, is probably having methods with more than ~10 parameters!

lucafedeli88 (commented):
Thanks a lot, @EZoni, for working on this!
 
I generally like the changes. 

My main doubts concern:

- indentation with pre-processor directives: I think that they are useful for readability!
- opening curly brace on a new line for classes and functions  : I personally like this style since it marks a difference with respect to curly braces after `if()`, `while()`...
- automatic new lines for variables in function declarations: I think that sometimes we use newlines in function declaration to express ideas (e.g., some variables are related) and to improve readability. I am not sure that the proposed style changes are more readable than the original version.

See more detailed comments below.

Source/WarpX.H
dpgrote: I would agree - the indentation makes it more readable.

Source/WarpX.H
dpgrote: I prefer the original. Putting it all on one line makes it harder to read. With the original, I can just glance at the code and see what it is doing. With it all on one line, I have to stare at it a moment to parse it out.

Source/WarpX.H
```diff
@@ -82,45 +82,47 @@
 #include <string>
 #include <vector>
 
-class WARPX_EXPORT WarpX
-    : public amrex::AmrCore
-{
-public:
+class WARPX_EXPORT WarpX : public amrex::AmrCore {
+  public:
```
dpgrote, line 86: Maybe it's just what I'm used to, but I prefer no indentation on the `public`, 'private', `protected' lines.

Source/WarpX.H
```diff
@@ -82,45 +82,47 @@
...
      */
-    static void Finalize();
+    static void Finalize ();
 
     /** Destructor */
     ~WarpX () override;
 
     /** Copy constructor */
-    WarpX ( WarpX const &) = delete;
+    WarpX (WarpX const&) = delete;
```
dpgrote, line 102: I approve of this standard. I've never liked the extra spaces before the arguments (and at the end of the list).

Source/WarpX.H
dpgrote: I would agree here, that the layout of the arguments should be up to the developer to allow nice grouping.

Source/WarpX.H
```diff
@@ -136,112 +138,160 @@ public:
...
-    MultiParticleContainer& GetPartContainer () { return *mypc; }
-    MultiFluidContainer& GetFluidContainer () { return *myfl; }
-    ElectrostaticSolver& GetElectrostaticSolver () {return *m_electrostatic_solver;}
-    [[nodiscard]] HybridPICModel * get_pointer_HybridPICModel () const { return m_hybrid_pic_model.get(); }
-    MultiDiagnostics& GetMultiDiags () {return *multi_diags;}
-    ParticleBoundaryBuffer& GetParticleBoundaryBuffer () { return *m_particle_boundary_buffer; }
-    amrex::Vector<std::array< std::unique_ptr<amrex::iMultiFab>,3 > >& GetEBUpdateEFlag() { return m_eb_update_E; }
-    amrex::Vector<std::array< std::unique_ptr<amrex::iMultiFab>,3 > >& GetEBUpdateBFlag() { return m_eb_update_B; }
-    amrex::Vector< std::unique_ptr<amrex::iMultiFab> > const & GetEBReduceParticleShapeFlag() const { return m_eb_reduce_particle_shape; }
+    void SetElectricFieldAndApplyBCs (const WarpXSolverVec& a_E,
```
dpgrote, line 145: Presumably these changes are being made because of the line length limit. I've always preferred longer lines which, to me, makes this type of code easier to read and more compact. What is the maximum line length?

Source/WarpX.cpp
dpgrote: I would agree that it is better to use indentation inside a name space. This makes it readily clear that the routines defined are in a defined space.

Source/WarpX.H
EZoni: The LLVM configuration file sets 
```
ColumnLimit:     80
```
so it's probably 80 at the moment.

Source/WarpX.cpp
EZoni: I think it should be possible to guarantee this by replacing LLVM's
```
NamespaceIndentation: None
```
with 
```
NamespaceIndentation: All
```

EZoni:
@lucafedeli88 @dpgrote 

I will try to go through each of your comments and see if there is a corresponding option that can be set to match your suggestions.

However, I'm starting to think that we might need to choose a strategy to move forward.

Should we discuss in the Technical Committee every formatting change that receives an objection? I am afraid that this might not be interesting for some. On the other hand, I or others might not agree with your or others' preferences, in which case a vote would be the only way to resolve, so I'm not sure.

Source/WarpX.H
EZoni: This might be controlled by `IndentAccessModifiers`?
Screenshot from 2025-03-11 17-12-13

Source/WarpX.H
EZoni: I am not sure if it is possible to customize this in a way that is systematic, beyond wrapping the relevant lines in a "don't format" wrap.

Source/WarpX.H
EZoni: This is probably conflicting with https://github.com/BLAST-WarpX/warpx/pull/5739#discussion_r1989950188. Meaning, if we like one and don't like the other, or viceversa, there might be no way to make this systematic.

Source/WarpX.cpp
EZoni: Here too I am not sure if there is a way to make this systematic, as one version might look more readable in one place, the other version in another place. I personally like the Ruff linting we use in Python, which enforces one argument per line, except for very short lines, because it makes reviewing code changes easier. 

These might be our only options:
Screenshot from 2025-03-11 17-22-20

Source/WarpX.H
EZoni: See https://github.com/BLAST-WarpX/warpx/pull/5739#discussion_r1990343527.

Source/WarpX.H
EZoni: This might be controlled by the following option:
Screenshot from 2025-03-11 17-28-24

So you both would prefer `BeforeHash`, correct?

dpgrote: Yes, `BeforeHash`.

Source/WarpX.H
dpgrote: Can the formatting of the routine arguments be turned off so that there is no standard? In this case, considering the wide variety of argument lists, there is no single standard that works well in all cases.

lucafedeli88:
> @lucafedeli88 @dpgrote
> 
> I will try to go through each of your comments and see if there is a corresponding option that can be set to match your suggestions.
> 
> However, I'm starting to think that we might need to choose a strategy to move forward.
> 
> Should we discuss in the Technical Committee every formatting change that receives an objection? I am afraid that this might not be interesting for some. On the other hand, I or others might not agree with your or others' preferences, in which case a vote would be the only way to resolve, so I'm not sure.

I think that having a vote in the Technical Committee is the best strategy. Sure, discussions about formatting style are usually not the most exciting ones, but I think that we can discuss and vote on all the changes proposed above in a limited amount of time (e.g. , 1 week).

EZoni:
@lucafedeli88

> I think that having a vote in the Technical Committee is the best strategy. Sure, discussions about formatting style are usually not the most exciting ones, but I think that we can discuss and vote on all the changes proposed above in a limited amount of time (e.g. , 1 week). 

The configuration file .clang-format has about 270 lines of format rules and this PR alone, which tackles only WarpX.H and WarpX.cpp, has `+2,603 −1,984` changed lines count. When you say "discuss and vote on all the changes proposed above", do you mean all the format rules encoded in the 270 lines of .clang-format or just the rules that @dpgrote, you and I discussed in the comments above?

PR #5740, opened 2025-03-07
Make labels with math text raw strings since they behave better

dpgrote:
This fixes one analysis file. There are probably others with the same issue. Note that using raw strings is only needed when the strings include back slashes, but for consistency all labels with math text are made raw.


lucafedeli88 (approved):
Thanks, @dpgrote

PR #5744, opened 2025-03-10
HybridPICModel : pass m_fields as an argument to InitData() to reduce reliance on WarpX::GetInstance()

lucafedeli88:
This PR is a small step towards reducing the reliance on ` WarpX::GetInstance()` to get access to member variables of the WarpX class.

### DIFF
--- a/Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
+++ b/Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
@@ -152,7 +152,7 @@ void HybridPICModel::AllocateLevelMFs (
 #endif
 }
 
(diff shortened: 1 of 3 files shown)


Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -183,19 +183,19 @@ void HybridPICModel::InitData ()
         m_external_current_has_time_dependence += J_ext_symbols.count("t");
     }
 
-    auto & warpx = WarpX::GetInstance();
+    auto& warpx = WarpX::GetInstance();
```
lucafedeli88, line 186: We still need this because it is used somewhere below.

roelof-groenewald (approved):

PR #5746, opened 2025-03-10
AcceleratorLattice: avoid using WarpX::GetInstance()

lucafedeli88:
This PR removes all the occurrences of `WarpX::GetInstance()` from the source code under `AcceleratorLattice`.


dpgrote (approved):
Thanks!

PR #5750, opened 2025-03-10
Check if implicit diagnostics file exist before writing the header

dpgrote:
This is a small fix to prevent the implicit diagnostics file from being overwritten during a restart.


JustinRayAngus (changes requested):
Hi Dave,

Just need to do the same thing in Picard.H and then we can merge it.

Thanks!

JustinRayAngus (approved):

PR #5752, opened 2025-03-11
Diagnostics: reduce reliance on WarpX::GetInstance()

lucafedeli88:
Reduce reliance on `WarpX::GetInstance()` in diagnostics by passing more parameters as arguments. 

What do you think about passing a group of parameters as a struct (** updated ** also according to @ax3l  suggestion)?

```C++
/** A struct containing the moving window parameters required to initialize a diagnostic */
struct InitDiagnosticsMovingWindowParameters
{
    std::optional<int> dir;                    /** The moving window direction */
    std::optional<amrex::Real> x;              /** The moving window position */
    std::optional<amrex::Real> cell_size_lev0; /** The cell size in the moving window direction (at level 0)*/
    std::optional<amrex::Real> prob_lo_lev0;   /** The lo end of the problem domain in the moving window direction (at level 0)*/
};

/** A struct containing the parameters required to initialize a diagnostic */
struct InitDiagnosticsParameters
{
    std::optional<int> finest_level;     /** The finest level */
    std::optional<int> max_level;        /** The max level */

    std::optional<InitDiagnosticsMovingWindowParameters> moving_window; /** Moving window parameters */
};
```

I think that it is useful especially for `WarpXIO.cpp` since several function calls requiring these parameters are made.
I am really sorry that C++17 doesn't have label initialization for structs like C .

⚠️⚠️ ⚠️ Note that `FlushFormatSensei` requires a pointer to the base class of the WarpX object, which is now passed as an argument `auto* p_warpx_mesh = dynamic_cast<amrex::AmrMesh*>(this);` . Before, this was obtained by `const` casting a `const` pointer obtained from  `WarpX::GetInstance()`. The refactored code has the added benefit that it removes this `const_cast` (usage of `const_cast` is often discouraged because it could lead to Undefined Behavior issues, see: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es50-dont-cast-away-const).


Source/Diagnostics/Diagnostics.H
```diff
@@ -22,7 +22,26 @@
...
+/** A struct containing the parameters required to initialize a diagnostics */
+struct InitDiagnosticsParameters
+{
+    int finest_level;
+    int max_level;
+
+    int do_moving_window;
+    int moving_window_dir;
+    amrex::Real moving_window_x;
+    amrex::Real cell_size_lev0_mwdir;
```
ax3l: As a purist, one could make them `std::optional` to ensure no UB can occur :nerd_face:

ax3l (dismissed):

Source/Diagnostics/Diagnostics.cpp
```diff
@@ -509,27 +504,26 @@ Diagnostics::InitData ()
...
 
     // For restart, move the m_lo and m_hi of the diag consistent with the
     // current moving_window location
-    if (WarpX::do_moving_window) {
-        const int moving_dir = WarpX::moving_window_dir;
+    if (params.do_moving_window) {
+        const int moving_dir = params.moving_window_dir;
         const amrex::Real displacement =
-            warpx.getmoving_window_x() - warpx.Geom(0).ProbLo(moving_dir);
+            params.moving_window_x - warpx.Geom(0).ProbLo(moving_dir);
```
ax3l: @lucafedeli88 I rebased your PR but in the meantime this extra term came in here.

I think you need to store `warpx.Geom(0).ProbLo(moving_dir)` now as well in `InitDiagnosticsParameters`?

Source/Diagnostics/Diagnostics.H
lucafedeli88: good idea, @ax3l ! That is what I am doing now.

Source/Diagnostics/Diagnostics.cpp
lucafedeli88: Yes, you are right! I've just changed it!

lucafedeli88:
I've also added some documentation

lucafedeli88:
And I've changed the structure to separate the parameters related to the moving window

lucafedeli88:
@ax3l , do you think it's ok now ?

lucafedeli88:
Oh... I see conflicts now ! I will address them as soon as possible

lucafedeli88:
Ping, @ax3l  ;-)

EZoni:
I was assigned to this PR a long time ago, apologies for letting it slip through. I just merged `development` (no conflicts, no changes) to check the latest status of the CI checks. I will also review it again to see if we can get it across the finish line.

Source/Diagnostics/Diagnostics.cpp
```diff
@@ -443,7 +441,7 @@ Diagnostics::InitData (const MultiParticleContainer& mpc)
         // This includes full diagnostics and BTD as well as cell-center functors for BTD.
         // Note that the cell-centered data for BTD is computed for all levels and hence
         // the corresponding functor is also initialized for all the levels
-        for (int lev = 0; lev <= finest_level; ++lev) {
+        for (int lev = 0; lev <= params.finest_level; ++lev) {
```
EZoni, line 444: Shouldn't this be `params.finest_level.value()` since `finest_level` is a `std::optional<int>`? We seem to use `finest_level.value()` elsewhere, too.

```suggestion
        for (int lev = 0; lev <= params.finest_level.value(); ++lev) {
```

Source/Diagnostics/Diagnostics.H
```diff
@@ -17,11 +17,37 @@
...
+    std::optional<amrex::Real> x;              /** The moving window position */
+    std::optional<amrex::Real> cell_size_lev0; /** The cell size in the moving window direction (at level 0)*/
+    std::optional<amrex::Real> prob_lo_lev0;   /** The lo end of the problem domain in the moving window direction (at level 0)*/
+};
+
+/** A struct containing the parameters required to initialize a diagnostic */
+struct InitDiagnosticsParameters
+{
+    std::optional<int> finest_level;     /** The finest level */
+    std::optional<int> max_level;        /** The max level */
```
EZoni, line 46: These are declared optional, but usually accessed with `.value()` unconditionally. Are these truly optional?

Source/Diagnostics/Diagnostics.H
```diff
@@ -17,11 +17,37 @@
...
+    TimeAveraged
+};
+
+/** A struct containing the moving window parameters required to initialize a diagnostic */
+struct InitDiagnosticsMovingWindowParameters
+{
+    std::optional<int> dir;                    /** The moving window direction */
+    std::optional<amrex::Real> x;              /** The moving window position */
+    std::optional<amrex::Real> cell_size_lev0; /** The cell size in the moving window direction (at level 0)*/
+    std::optional<amrex::Real> prob_lo_lev0;   /** The lo end of the problem domain in the moving window direction (at level 0)*/
```
EZoni, line 39: Similar question as for `finest_level` and `max_level`, are these truly optional if `moving_window` is present?

EZoni, line 39: Also, these docstrings, as is, are not interpreted correctly by Doxygen (at least, some versions).

When I build the documentation locally, the syntax `/** docstring */` produces

<p align="center">
Image
</p>

while the syntax `/**< docstring */` produces 

<p align="center">
Image
</p>

I learned this recently when I added an algorithm selector:
https://github.com/BLAST-WarpX/warpx/blob/ead820f7b1ebb52d056a282b0791f9af55654889/Source/Utils/WarpXAlgorithmSelection.H#L190-L198

Therefore, I suggest:

```suggestion
    std::optional<int> dir;                    /**< The moving window direction */
    std::optional<amrex::Real> x;              /**< The moving window position */
    std::optional<amrex::Real> cell_size_lev0; /**< The cell size in the moving window direction (at level 0) */
    std::optional<amrex::Real> prob_lo_lev0;   /**< The lo end of the problem domain in the moving window direction (at level 0) */
```

EZoni (changes requested):

Source/Diagnostics/Diagnostics.H
```diff
@@ -72,9 +98,17 @@ public:
      */
     virtual void Flush (int i_buffer, bool force_flush) = 0;
 
-    /** Initialize pointers to main fields and allocate output multifab m_mf_output. */
-    void InitData (const MultiParticleContainer& mpc);
-    void InitDataBeforeRestart ();
+    /** \brief * Initialize pointers to main fields and allocate output multifab m_mf_output.
```
EZoni, line 101: Typo:

```suggestion
    /** \brief Initialize pointers to main fields and allocate output multifab m_mf_output.
```

Source/Initialization/WarpXInitData.cpp
```diff
@@ -991,7 +991,21 @@ WarpX::AddExternalFields (int const lev)
...
+        InitDiagnosticsParameters{finestLevel(), maxLevel(), std::nullopt};
+    if (do_moving_window){
+        init_diag_params.moving_window =
+            InitDiagnosticsMovingWindowParameters{
+                moving_window_dir, moving_window_x,
+                Geom(0).CellSize(moving_window_dir),
+                Geom(0).ProbLo(moving_window_dir)};
+    }
+
+    auto* p_warpx_mesh = dynamic_cast<amrex::AmrMesh*>(this);
```
EZoni, line 1006: This is duplicated in `InitFromCheckpoint` (239-250):
```cpp
        // Parameters used to initialize diagnostics
        auto init_diag_params =
            InitDiagnosticsParameters{finestLevel(), maxLevel(), std::nullopt};
        if (do_moving_window){
            init_diag_params.moving_window =
                InitDiagnosticsMovingWindowParameters{
                    moving_window_dir, moving_window_x,
                    Geom(0).CellSize(moving_window_dir),
                    Geom(0).ProbLo(moving_window_dir)};
        }

        auto* p_warpx_mesh = dynamic_cast<amrex::AmrMesh*>(this);
```

Do you think it is possible and useful to extract this into a small private helper, e.g., `WarpX::makeInitDiagnosticsParameters()`?

EZoni (changes requested):

PR #5753, opened 2025-03-11
Fix a runtime parameter issue in time-averaged diagnostics

### LINKED ISSUE #5751: Constant `clight` not recognized
## Description
I'm not sure if this is a bug or a misunderstanding. I defined a constant in the input file
`my_constants.omega0 = 2 * pi * clight / lambda0`
based on the [input guide](https://warpx.readthedocs.io/en/latest/usage/parameters.html#warpx-constants), which says that some universal constants are available for parsing.
However, the code crashed at start complaining that `clight` wasn't defined. Am I doing something wrong?
The specific error message is
`amrex::Error::15444::ParmParse: failed to parse 2*pi*clight/lambda0 due to unknown symbol clight !!!`
This is based on WarpX `25.03`.

WeiqunZhang:
Use `utils::parser::queryWithParser` instead of vanilla `amrex::Parser`, because latter does not know WarpX's constants unless it's informed.

Fix #5751

### DIFF
--- a/Source/Diagnostics/FullDiagnostics.cpp
+++ b/Source/Diagnostics/FullDiagnostics.cpp
@@ -16,6 +16,7 @@
 #include "FlushFormats/FlushFormat.H"
 #include "Particles/MultiParticleContainer.H"
 #include "Utils/Algorithms/IsIn.H"
+#include "Utils/Parser/ParserUtils.H"
 #include "Utils/TextMsg.H"
 #include "Utils/WarpXAlgorithmSelection.H"
 #include "WarpX.H"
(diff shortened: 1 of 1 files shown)


Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -155,9 +156,8 @@ FullDiagnostics::ReadParameters ()
         const bool averaging_period_steps_specified = pp_diag_name.query(
                 "average_period_steps", m_average_period_steps
         );
-        const bool averaging_period_time_specified = pp_diag_name.queryWithParser(
-                "average_period_time", m_average_period_time
-        );
+        const bool averaging_period_time_specified = utils::parser::queryWithParser
```
ax3l, line 158: Temporary work-around until we have #5254

ax3l (approved):

PR #5754, opened 2025-03-11
Fix parser function string in MCCCollision

### LINKED ISSUE #5741: Unclear behavior from string expression for background density with MCCCollision
Specifying a non-uniform background gas density for a background MCC ionization calculation via the background_density argument does not seem to produce the expected behavior.

I specified the density as a function (`"1.0e18 * exp(-(z)**2/(2*(30e-6)**2))"`), which seems to be correctly communicated to the input deck. This should produce ionization localized near the central z-plane, but the resulting simulation indicates ionization occurs roughly uniformly along the path of the electron beam. The particle weights are all equivalent.

Image

I link the input file ( [PICMI_custom_dens_ionization.py](https://gist.github.com/ncook882/a625cb842ad10fee70edc6679a49e4ee#file-picmi_custom_dens_ionization-py)) and corresponding ionization file ([test_ionization.dat](https://gist.github.com/ncook882/a625cb842ad10fee70edc6679a49e4ee#file-test_ionization-dat)). Can be run on a single CPU or via MPI.

## System information
WarpX 25.02 on Linux Fedora 36. Tested on both MPI (16 processes) and a single CPU.

## Steps taken so far
The issue was also reported in the [Slack channel](https://ecpwarpx.slack.com/archives/CAR7QN5HQ/p1741295819766119).

WeiqunZhang:
We should use `utils::parser::Store_parserString` instead of `amrex::Parser::get` to obtain the parser function string because the latter treats spaces as separators for different strings.

Fix #5741

### DIFF
--- a/Source/Particles/Collision/BackgroundMCC/BackgroundMCCCollision.cpp
+++ b/Source/Particles/Collision/BackgroundMCC/BackgroundMCCCollision.cpp
@@ -40,7 +40,7 @@ BackgroundMCCCollision::BackgroundMCCCollision (std::string const& collision_nam
     }
     else {
         std::string background_density_str;
-        pp_collision_name.get("background_density(x,y,z,t)", background_density_str);
+        utils::parser::Store_parserString(pp_collision_name, "background_density(x,y,z,t)", background_density_str);
         m_background_density_parser =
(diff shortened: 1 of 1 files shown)


roelof-groenewald (approved):
Thanks @WeiqunZhang!

PR #5761, opened 2025-03-12
Add automated test for ion impact ionization

RemiLehe:
TODO: update analysis script and recreate checksums


EZoni:
Just a side comment, as far as creating the checksum file is concerned, I think this could be a good opportunity to verify that the steps in our how-to documentation (https://warpx.readthedocs.io/en/latest/developers/how_to_test.html#how-to-add-automated-tests) work properly, in particular step 7. below (I'm not sure if developers actually do this, instead of copying existing checksums and then overwriting them):

Screenshot from 2025-03-12 10-04-18

roelof-groenewald:
@EZoni I was able to reset the benchmarks using the instructions you linked 🎉 . 
I guess we'll see if the checksum values from my local system matches what the Azure systems get.

@RemiLehe I modified the scattering logic to avoid the `nan` values from before (at least for the test you created). Could you please check if the update works for your actual use case?

Btw, I checked that the electron impact ionization test still match theory (as close as it did before), and it looks good:
ionization_dsmc_density_Te

roelof-groenewald:
> TODO: update analysis script and recreate checksums

@RemiLehe are you planning on updating the analysis script to get the theoretical ion impact ionization rate and the effect on the ion temperature?

Examples/Tests/ionization_dsmc/inputs_base_3d
```diff
@@ -21,7 +22,7 @@ warpx.const_dt = 1e-09
 geometry.dims = 3
 geometry.prob_hi = 0.1 0.1 0.1
 geometry.prob_lo = 0 0 0
-amr.max_grid_size = 8
+amr.max_grid_size = 4
```
roelof-groenewald, line 28: I noticed the test runs on 2 procs but with `max_grid_size=8` there was only 1 tile, so parallel execution was not really tested.

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -412,52 +412,58 @@ public:
...
+                    // is done as follows:
+                    // Three numbers are generated that satisfy the triangle
+                    // inequality. These numbers will be scaled to give the
+                    // momentum for each particle (satisfying the triangle
+                    // inequality ensures that the momentum vectors can be
+                    // arranged such that the net linear momentum is zero).
+                    // Product 2 is definitely an ion, so we first choose its
+                    // proportion to be n_2 = min(E2 / E_coll, 0.5 * R), where
+                    // R is a random number between 0 and 1.
+                    // The second and third numbers (n_0 & n_1) are then found
```
```suggestion
                    // The other two numbers (n_0 & n_1) are then found
```

roelof-groenewald:
The DSMC ionization tests require https://github.com/BLAST-WarpX/warpx-data/pull/6 to be merged to add the necessary cross-section files.

Now using the hydrogen cross-sections and having turned off the field solver in the test, we get pretty close match to the theoretical model:
image

RemiLehe:
Relaunching the CI tests, now that https://github.com/BLAST-WarpX/warpx-data/pull/6 is merged.

Examples/Tests/ionization_dsmc/CMakeLists.txt
```diff
@@ -2,11 +2,21 @@
...
     "analysis_default_regression.py --path diags/diag1000250"  # checksum
     OFF  # dependency
 )
+
+add_warpx_test(
+    test_3d_ionization_ion_dsmc  # name
+    3  # dims
+    2  # nprocs
+    inputs_test_3d_ionization_ion_dsmc  # inputs
+    OFF # "analysis_ionization_dsmc_3d.py"  # analysis
```
RemiLehe, line 19: Could we turn the automated test on here?

roelof-groenewald: The analysis would have to be modified substantially. The current test implicitly assumes that the neutral is much heavier than the impacting particle - a good assumption for electron impact but not for ion impact. So I'd say we should merge this as is and then possibly add a similar but modified analysis for the ion-impact case in a future PR.

roelof-groenewald (approved):

PR #5763, opened 2025-03-12
Calculate global debye length for Coulomb collisions

dpgrote:
This adds the option to calculate the Debye length including all species in the simulation when doing Coulomb collisions (and the Coulomb log is being calculated). This is turned on by setting the new flag `<collision_name>.use_global_debye_length = 1`.


ax3l:
Triggering CI

ax3l:
ping @JustinRayAngus do you like to review this one? :)

JustinRayAngus:
I will review. Probably good for another reviewer as well.

JustinRayAngus (commented):
This all looks good to me Dave.

Only question I have is have you done some basic sanity checks such as print out the density, temperature, and Debye length you get from the calculations here and compare this with what is expected?

Source/Particles/Collision/BinaryCollision/Coulomb/PairWiseCoulombCollisionFunc.H
```diff
@@ -53,15 +53,23 @@ public:
...
         amrex::ParticleReal CoulombLog = -1.0_prt;
         utils::parser::queryWithParser(
             pp_collision_name, "CoulombLog", CoulombLog);
         m_CoulombLog = CoulombLog;
-
         m_exe.m_CoulombLog = m_CoulombLog;
-        if (m_CoulombLog<0.0) { m_exe.m_computeSpeciesTemperatures = true; }
+
+        if (m_CoulombLog<0.0 && !m_use_global_debye_length) {
+            m_exe.m_computeSpeciesTemperatures = true;
```
JustinRayAngus, line 68: :+1:

dpgrote: Yes, I did do some checks and all of the quantities look reasonable. I also ran a plasma test case comparing with and without this feature and the results were consistent. I could add a CI test, reusing one of the existing plasma tests with this turned on.

JustinRayAngus:
@ax3l @EZoni @RemiLehe 

This PR looks good to me. Does anyone else want to take a pass at it before merging?

Docs/source/usage/parameters.rst
```diff
@@ -2088,6 +2088,13 @@ Details about the collision models can be found in the :ref:`theory section <mul
     a Coulomb logarithm will be computed automatically according to the algorithm in
     :cite:t:`param-PerezPOP2012`.
 
+* ``<collision_name>.use_global_debye_length`` (`bool`) optional
+    Only for ``pairwisecoulomb``. When set, the Debye length used in the Coulomb log
+    is calculated including all species in the simulation. The lengths are combined
+    using the square root of one over the sum of one over the sqaures of the Debye lengths
+    of each species. By default, this is turned off. Note that if ``<collision_name>.CoulombLog``
+    is specified, this Debye length is not used.
```
EZoni: Wouldn't it be easier to write a formula (or pseudo-formula) rather than explaining in words as `The lengths are combined using the square root of one over the sum of one over the sqaures of the Debye lengths of each species`? If you prefer the word explanation, note the typo `sqaures`.

dpgrote: In this case, I prefer the written description - it seems more concise. Writing a formula would take up more space and would require the extra text defining terms and subscripts.

EZoni (commented):
Thanks for the PR, @dpgrote!

One question that applies to most changes. When we have something like
```cpp
amrex::MultiFab & global_debye_length = *warpx.m_fields.get(warpx::fields::FieldType::global_debye_length, lev);
```
or similar, should we be using `ablastr::fields::ScalarField` as a type, for consistency with the fact that we use these aliases for most fields recorded in the MultiFab register, since it was introduced?

dpgrote:
> Thanks for the PR, @dpgrote!
> 
> One question that applies to most changes. When we have something like
> 
> ```c++
> amrex::MultiFab & global_debye_length = *warpx.m_fields.get(warpx::fields::FieldType::global_debye_length, lev);
> ```
> 
> or similar, should we be using `ablastr::fields::ScalarField` as a type, for consistency with the fact that we use these aliases for most fields recorded in the MultiFab register, since it was introduced?

@ezoni Good question. Grepping through the source, most places that use `fields.get` use the `amrex::MultiFav &` type. Only a few used `ScalarField`. My preference would be to stick with `MultiFab` since that is more transparent and it is very unlikely that that type would ever change.

JustinRayAngus (approved):
Thanks Dave. This looks good to me.

PR #5764, opened 2025-03-13
Doc: More Nsight Hints

ax3l:
Add more hints and links how to work with Nsight tools.


ax3l:
"Force-merging" because CI is skipped and will never pass, FYI @EZoni :)

PR #5766, opened 2025-03-14
Add more profiling instrumentation in the collision kernel

RemiLehe:


roelof-groenewald (approved):
As discussed in the developer meeting we can merge this as the extra profiling should not add significant overhead (some synchronization, but that is not expected to be a major slowdown).

PR #5767, opened 2025-03-14
Refactor ReadBCParams

lucafedeli88:
I would like to propose to refactor the `ReadBCParams` function (now defined in `WarpXUtil`), with the aim of:
- avoid manipulating directly the static variables of the WarpX class
- re-organize what the function does into smaller functions that only  do one thing

Specifically, the changes are:

- create a new `/FieldBoundaries.H/cpp` file in `BoundaryConditions`, that for the moment contains only one pure function to parse the field boundary types from the inputfile (the idea is to extend this file later):  
```
std::pair<
        amrex::Array<FieldBoundaryType, AMREX_SPACEDIM>,
        amrex::Array<FieldBoundaryType, AMREX_SPACEDIM>
    >
    warpx::boundary_conditions::parse_field_boundaries ()
```

- move the part of `ReadBCParams`  that sets `pp_geometry.addarr("is_periodic", geom_periodicity);` according to boundary conditions to `set_periodicity_according_to_boundary_types` in `Initialization/WarpXAMReXInit.cpp` (we usually place here methods that change AMReX default parameters  during the initialization).

-  move the part of  `ReadBCParams` that parses the particle boundary conditions to a new pure function  `parse_particle_boundaries` in `Particles/ParticleBoundaries.H/cpp` (note that an array indicating if field boundary conditions are periodic is passed as arguments for consistency checks and for assigning default values):
```
    std::pair<
        amrex::Array<ParticleBoundaryType, AMREX_SPACEDIM>,
        amrex::Array<ParticleBoundaryType, AMREX_SPACEDIM>
    >
    parse_particle_boundaries (
        const amrex::Array<bool, AMREX_SPACEDIM>& is_field_boundary_periodic);
```

- consistency checks in `Particles/ParticleBoundaries.cpp`  and  `BoundaryConditions/FieldBoundaries.cpp` are moved to helper functions inside a `details` namespace

- a helper function `get_periodicity_array` is provided for field boundary conditions in order to get an array specifying if boundary conditions are periodic in a given direction.

- initialize `field_boundary_lo`, `field_boundary_hi`,  `particle_boundary_lo`, `particle_boundary_hi` inside `WarpX::MakeWarpX`

- Move consistency checks involving field solver to `WarpX::ReadParameters` (a further improvement would be to break `ReadParameters` into smaller pieces, but this will require several dedicated PRs). 

- remove `void ReadBCParams ()` from `WarpXUtil.H/cpp`


Source/BoundaryConditions/FieldBoundaries.H
```diff
@@ -0,0 +1,30 @@
...
+#include <AMReX_Array.H>
+#include <AMReX_Config.H>
+
+#include <utility>
+
+namespace warpx::boundary_conditions
+{
+    /**
+     * \brief Reads the field boundary conditions from
+     * the inputfile
```
EZoni: This can probably fit on one line?
```suggestion
     * \brief Reads the field boundary conditions from the input file
```

Source/Particles/ParticleBoundaries.H
```diff
@@ -10,8 +10,34 @@
...
+#include <AMReX_Config.H>
 #include <AMReX_Parser.H>
 
+#include <utility>
+
+namespace warpx::particles
+{
+    /**
+     * \brief Reads the particle boundary conditions from
+     * the inputfile
```
EZoni: This can probably fit on one line as well?
```suggestion
     * \brief Reads the particle boundary conditions from the input file
```

Source/Particles/ParticleBoundaries.cpp
```diff
@@ -8,6 +8,61 @@
...
+                        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                            (particle_boundary_lo[idim] == ParticleBoundaryType::Periodic) &&
+                            (particle_boundary_hi[idim] == ParticleBoundaryType::Periodic),
+                            "field and particle boundary must be periodic in both lo and hi");
+                    } else {
+                        // set particle boundary to periodic
+                        particle_boundary_lo[idim] = ParticleBoundaryType::Periodic;
+                        particle_boundary_hi[idim] = ParticleBoundaryType::Periodic;
+                    }
+                }
```
EZoni: I wonder if it would be clearer to have `parse_particle_boundaries` take no arguments and skip this check (and default set), and define a separate function, say, `check_periodicity` or something along these lines, that takes both the field and particle boundary tuples, and performs the periodicity check.

This separate function would be then called both in WarpX.cpp and Initialization/WarpXAMReXInit.cpp (from within `set_periodicity_according_to_boundary_types`).

This would seem clearer to me, from a functional point of view, but it's just a suggestion. Let me know what you think or if you see any drawbacks.

EZoni (commented):
Thanks for this PR, @lucafedeli88! I left a few minor comments.

Source/Particles/ParticleBoundaries.cpp
EZoni: @lucafedeli88

I think we are close to finalizing this PR, as soon as we resolve the last comment here. Let me know what you think!

lucafedeli88: Thanks for your feedback @EZoni , and sorry for my late answer. 
I would like to propose a slightly different solution, since field boundary conditions are required to select the default value of the boundary conditions for particles (i.e., they are not just used for consistency checks): 

- consistency checks are moved to helper functions hidden in the implementation file, for clarity
- a new helper function generates a "periodicity array" for field boundary conditions (true = periodic, false = otherwise )
- this bool array is now the argument of the function that reads particles' boundary conditions. 

What do you thinks about this solution ?

EZoni: Thanks, @lucafedeli88. The solution sounds good to me, although seeing it implemented would make it easier to fully understand how much it simplify/"complexify" things.

EZoni: Oh I just noticed that you already pushed more commits to refactor things! Will review the changes again.

Source/WarpX.cpp
```diff
@@ -283,8 +283,12 @@ void WarpX::MakeWarpX ()
 
     std::tie(field_boundary_lo, field_boundary_hi) =
         warpx::boundary_conditions::parse_field_boundaries();
+
+    const auto is_field_boundary_periodic =
+        warpx::boundary_conditions::get_periodicity_array (field_boundary_lo, field_boundary_hi);
```
```suggestion
        warpx::boundary_conditions::get_periodicity_array(field_boundary_lo, field_boundary_hi);
```

Source/Initialization/WarpXAMReXInit.cpp
```diff
@@ -118,8 +118,12 @@ namespace {
 
         const auto [field_boundary_lo, field_boundary_hi] =
             warpx::boundary_conditions::parse_field_boundaries();
+
+        const auto is_field_boundary_periodic =
+            warpx::boundary_conditions::get_periodicity_array (field_boundary_lo, field_boundary_hi);
```
EZoni:
```suggestion
            warpx::boundary_conditions::get_periodicity_array(field_boundary_lo, field_boundary_hi);
```

EZoni (changes requested):
Thanks, the latest changes look good to me. I left only two requests to fix a white space in a couple of function calls.

EZoni (approved):

PR #5768, opened 2025-03-15
adding infrastructure for computing the mass matrices to be used by the implicit EM solvers

JustinRayAngus:
This PR adds the basic infrastructure for computing the mass matrices needed by the implicit EM solvers to efficiently take time steps larger than the plasma period.

The mass matrices are deposited to the grid from the particles in the same fashion as the current is deposited to the grid. The relationship between dJ and dE is written as 
dJx = MassMatrices_xx * dEx + MassMatrices_xy *  dEy + MassMatrices_xz * dEz
dJy = MassMatrices_yx * dEx + MassMatrices_yy * dEy + MassMatrices_yz * dEz
dJz = MassMatrices_zx * dEx + MassMatrices_zy * dEy + MassMatrices_zz * dEz

At each grid point, each element in the expressions above is a sum over all local grid points depending on the particle shape factor. **For this PR, we are only computing the diagonal elements of the diagonal mass matrices**:
dJx(i,j,k) = **MassMatrices_xx(i,j,k)** * dEx(i,j,k)
dJy(i,j,k) = **MassMatrices_yy(i,j,k)** * dEy(i,j,k)
dJz(i,j,k) = **MassMatrices_zz(i,j,k)** * dEz(i,j,k)

The diagonal elements are used in the preconditioner for capturing plasma response.

The mass matrices are described in
Justin Ray Angus, et. al., An implicit particle code with exact energy and charge conservation for electromagnetic studies of dense plasmas, Journal of Computational Physics, Volume 491, 2023, 112383, https://doi.org/10.1016/j.jcp.2023.112383.

dane_gmres_planar_pinch_longer

This figure illustrates the effectiveness of including the plasma response in the preconditioner. These results are from 1D planar pinch simulations as described in the above reference. The parameters are such that \omega_{pe}*\Delta_t ranges from 18 at t = 0 to 33 for t > 15 ns where the density is compressed. The CFL number is 4.2.

The number of GMRES iterations per Newton iteration increases as the plasma density increases when not using a preconditioner. Using the curl curl preconditioner without the plasma response actually result in more iterations. This is not totally unexpected since the stiffness of the system is dominated by the plasma rather than light waves. When the diagonal mass matrices are included in the curl curl preconditioner, the number of iterations decreases and becomes relatively insensitive to the change in density. This behavior is similar to that for the same conditions as reported in the above reference.

The number of GMRES iterations can be reduced further by including the first layer of off-diagonal elements of the mass matrices in the PC. That will be done in a future PR.


Source/FieldSolver/ImplicitSolvers/WarpXImplicitOps.cpp
```diff
@@ -61,11 +62,85 @@ WarpX::ImplicitPreRHSOp ( amrex::Real  a_cur_time,
...
+    // Copy mass matrices elements used for the preconditioner
+    for (int lev = 0; lev <= finest_level; ++lev) {
+
+        const int ncomp0 = m_fields.get(FieldType::MassMatrices_PC, Direction{0}, lev)->nComp();
+        const int ncomp1 = m_fields.get(FieldType::MassMatrices_PC, Direction{1}, lev)->nComp();
+        const int ncomp2 = m_fields.get(FieldType::MassMatrices_PC, Direction{2}, lev)->nComp();
+
+        ablastr::fields::VectorField MM = m_fields.get_alldirs(FieldType::MassMatrices, lev);
+        ablastr::fields::VectorField MM_PC = m_fields.get_alldirs(FieldType::MassMatrices_PC, lev);
+        amrex::MultiFab::Copy(*MM_PC[0], *MM[0], 0, 0, ncomp0, MM[0]->nGrowVect());
```
dpgrote: It looks like the `ncomp0` etc temporaries are not needed. This could be
```
amrex::MultiFab::Copy(*MM_PC[0], *MM[0], 0, 0, MM[0]->nComp(), MM[0]->nGrowVect());
```

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
```diff
@@ -84,3 +86,38 @@ Array<LinOpBCType,AMREX_SPACEDIM> ImplicitSolver::convertFieldBCToLinOpBC (const
...
+        m_WarpX->m_fields.alloc_init(FieldType::MassMatrices, Direction{1}, lev, ba_Ey, dm, 1, ngb, 0.0_rt);
+        m_WarpX->m_fields.alloc_init(FieldType::MassMatrices, Direction{2}, lev, ba_Ez, dm, 1, ngb, 0.0_rt);
+        m_WarpX->m_fields.alloc_init(FieldType::MassMatrices_PC, Direction{0}, lev, ba_Ex, dm, 1, ngb, 0.0_rt);
+        m_WarpX->m_fields.alloc_init(FieldType::MassMatrices_PC, Direction{1}, lev, ba_Ey, dm, 1, ngb, 0.0_rt);
+        m_WarpX->m_fields.alloc_init(FieldType::MassMatrices_PC, Direction{2}, lev, ba_Ez, dm, 1, ngb, 0.0_rt);
+    }
+
+    // Set the pointer to mass matrix MultiFab
+    for (int lev = 0; lev < m_num_amr_levels; ++lev) {
+        m_mmpc_mfarrvec.push_back(m_WarpX->m_fields.get_alldirs(FieldType::MassMatrices_PC, 0));
```
dpgrote, line 120: Rather than saving references to the MultiFabs, it would be better to get them from the registry each time they are used. The saved references would, for example, become invalid after a load balance.

Source/FieldSolver/ImplicitSolvers/WarpXImplicitOps.cpp
```diff
@@ -61,11 +62,85 @@ WarpX::ImplicitPreRHSOp ( amrex::Real  a_cur_time,
...
+
+    using ablastr::fields::Direction;
+    using warpx::fields::FieldType;
+
+    // Scale mass matrices used by preconditioner by c^2*mu0*theta*dt
+    // Note: This should be done after Sync/communication has been called
+
+    const amrex::Real pc_factor = PhysConst::c * PhysConst::c * PhysConst::mu0 * a_theta_dt;
+
+    for (int lev = 0; lev <= finest_level; ++lev) {
```
dpgrote: For clarity, I would combine these two loops and use a loop over the direction.
```
    const int diag_comp = 0;
        for (int idir = 0 ; idir < 3 ; idir++) {
            MultiFab * mass_matrix = m_fields.get(FieldType::MassMatrices_PC, Direction{idir}, lev);
            mass_matrix->mult(pc_factor, 0, mass_matrix->nComp());
            mass_matrix->plus(1.0_rt, diag_comp, 1, 0);
        }
    }
```

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
JustinRayAngus: I believe this setup is related to interacting with the curl curl solver in AMReX. @debog should comment on this.

Source/FieldSolver/ImplicitSolvers/WarpXImplicitOps.cpp
JustinRayAngus: I wrote this explicitly separate because we were hoping to have this done elsewhere in the curl curl solver. I don't think that is going to happen now, so I can combine the two as you suggest.

Source/Particles/Deposition/MassMatricesDeposition.H
```diff
@@ -0,0 +1,1041 @@
...
+    amrex::Array4<amrex::Real> const& jz_arr = jz_fab.array();
+    amrex::IntVect const jx_type = jx_fab.box().type();
+    amrex::IntVect const jy_type = jy_fab.box().type();
+    amrex::IntVect const jz_type = jz_fab.box().type();
+
+    // Loop over particles and deposit into jx_fab, jy_fab and jz_fab
+    amrex::ParallelFor(
+            np_to_deposit,
+            [=] AMREX_GPU_DEVICE (long ip) {
+            amrex::ParticleReal xp_nph, yp_nph, zp_nph;
```
dpgrote, line 367: Could this code setting up the `fp` values be put into a separate kernel routine since it looks like it is repeated in the two deposition routines?

dpgrote:
Could you add a CI test for this? Presumably you have test cases that you're using already.

Source/FieldSolver/ImplicitSolvers/WarpXImplicitOps.cpp
JustinRayAngus: Done.

Source/Particles/Deposition/MassMatricesDeposition.H
JustinRayAngus: Done

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
JustinRayAngus: Discussed off line. Leaving as is for now.

Source/FieldSolver/ImplicitSolvers/WarpXImplicitOps.cpp
JustinRayAngus: Done.

JustinRayAngus:
The 1D CI test I added is failing during step 1 with message:
Erroneous arithmetic operation
See Backtrace.1.0 file for details

@EZoni @dpgrote @RemiLehe  I have ran the CI test locally and it works just fine. It even works with debug build. Perhaps there are other options for the build that I need to consider? Where can I find the exact build flags used in the CI tests?

dpgrote:
> The 1D CI test I added is failing during step 1 with message: Erroneous arithmetic operation See Backtrace.1.0 file for details
> 
> @EZoni @dpgrote @RemiLehe I have ran the CI test locally and it works just fine. It even works with debug build. Perhaps there are other options for the build that I need to consider? Where can I find the exact build flags used in the CI tests?

Did you run the test using `ctest`? That would use all of the correct options. Also, what machine did you test it on? Sometimes running on a different machine will help find errors.

JustinRayAngus:
> > The 1D CI test I added is failing during step 1 with message: Erroneous arithmetic operation See Backtrace.1.0 file for details
> > @EZoni @dpgrote @RemiLehe I have ran the CI test locally and it works just fine. It even works with debug build. Perhaps there are other options for the build that I need to consider? Where can I find the exact build flags used in the CI tests?
> 
> Did you run the test using `ctest`? That would use all of the correct options. Also, what machine did you test it on? Sometimes running on a different machine will help find errors.

I ran using ctest, but with my own build directory.
ctest --test-dir /usr/workspace/angus1/codes/WarpX/build_1D_MM -R test_1d_pl

EZoni:
@JustinRayAngus 

Sorry for the late reply, as for the exact build flags used to build the CI tests, they can be found in https://github.com/BLAST-WarpX/warpx/blob/development/.azure-pipelines.yml. In particular,
- https://github.com/BLAST-WarpX/warpx/blob/0013f2f2eeeb0743871e481a2e0462b5b0fe11bc/.azure-pipelines.yml#L32-L49
- https://github.com/BLAST-WarpX/warpx/blob/0013f2f2eeeb0743871e481a2e0462b5b0fe11bc/.azure-pipelines.yml#L133-L134

If you configure, e.g., with
```bash
cmake -S . -B build -DAMReX_ASSERTIONS=ON -DAMReX_TESTING=ON -DWarpX_TEST_CLEANUP=OFF -DWarpX_TEST_FPETRAP=ON -DWarpX_BACKTRACE_INFO=ON -DWarpX_FFT=ON -DWarpX_PYTHON=ON -DWarpX_DIMS=1
```
you should be building in the same way as we do in our Azure pipelines. Note that I'm suggesting that you build with `-DWarpX_TEST_CLEANUP=OFF` here, otherwise your test directory is purged after the test has run.

JustinRayAngus:
> @JustinRayAngus
> 
> Sorry for the late reply, as for the exact build flags used to build the CI tests, they can be found in https://github.com/BLAST-WarpX/warpx/blob/development/.azure-pipelines.yml. In particular,
> 
> * https://github.com/BLAST-WarpX/warpx/blob/0013f2f2eeeb0743871e481a2e0462b5b0fe11bc/.azure-pipelines.yml#L32-L49
> * https://github.com/BLAST-WarpX/warpx/blob/0013f2f2eeeb0743871e481a2e0462b5b0fe11bc/.azure-pipelines.yml#L133-L134
> 
> If you configure, e.g., with
> 
> ```shell
> cmake -S . -B build -DAMReX_ASSERTIONS=ON -DAMReX_TESTING=ON -DWarpX_TEST_CLEANUP=OFF -DWarpX_TEST_FPETRAP=ON -DWarpX_BACKTRACE_INFO=ON -DWarpX_FFT=ON -DWarpX_PYTHON=ON -DWarpX_DIMS=1
> ```
> 
> you should be building in the same way as we do in our Azure pipelines. Note that I'm suggesting that you build with `-DWarpX_TEST_CLEANUP=OFF` here, otherwise your test directory is purged after the test has run.

Thanks for the exact config command. Running the ctest locally with this build does produce the same error I see on azure, but If I just run the input deck without doing ctest it runs just fine. I'm not sure how to proceed from here.

EZoni:
I will try to run it locally from your branch with the extra `-DCMAKE_BUILD_TYPE=Debug` and see if I can gather more insight, but, even without this flag, thanks to `-DWarpX_BACKTRACE_INFO=ON` I would expect that you can find some backtrace information in `build/bin/test_1d_planar_pinch_withPC`, where `build` is your build directory. Do you see any backtrace there?

JustinRayAngus:
> I will try to run it locally from your branch with the extra `-DCMAKE_BUILD_TYPE=Debug` and see if I can gather more insight, but, even without this flag, thanks to `-DWarpX_BACKTRACE_INFO=ON` I would expect that you can find some backtrace information in `build/bin/test_1d_planar_pinch_withPC`, where `build` is your build directory. Do you see any backtrace there?

@WeiqunZhang  Seems that the issue occurs in the curl curl PC in AMReX at this function: MLMG::mgVcycle_down::0

JustinRayAngus:
@EZoni 

One thing I still don't understand is why the failures only occur when I run using ctest? What is different about running with ctests vs running normally with srun?

EZoni:
There should be virtually no difference if you configure and build with the same flags and run with the same number of processes. However, we do set a few more runtime parameters, once for all tests, which you may not have in your individual input file.

This is the command we actually execute when we run each test:
https://github.com/BLAST-WarpX/warpx/blob/0013f2f2eeeb0743871e481a2e0462b5b0fe11bc/Examples/CMakeLists.txt#L167-L173

The runtime parameters are set here:
https://github.com/BLAST-WarpX/warpx/blob/0013f2f2eeeb0743871e481a2e0462b5b0fe11bc/Examples/CMakeLists.txt#L145-L160

It is possible that the erroneous arithmetic operation is caught by one of these runtime parameters.

JustinRayAngus:
@EZoni 

This is the flag that reproduces the error when running locally:
amrex.fpe_trap_invalid = 1

EZoni:
> This is the flag that reproduces the error when running locally: `amrex.fpe_trap_invalid = 1`

@WeiqunZhang 
`amrex.fpe_trap_invalid = 1` usually catches `NaN` and similar things, right? I wonder if the code is accessing data that has not been initialized or assigned properly.

JustinRayAngus:
@EZoni @WeiqunZhang @dpgrote 

It seems there are 

> > This is the flag that reproduces the error when running locally: `amrex.fpe_trap_invalid = 1`
> 
> @WeiqunZhang `amrex.fpe_trap_invalid = 1` usually catches `NaN` and similar things, right? I wonder if the code is accessing data that has not been initialized or assigned properly.

Is there a way to print where the NAN is? It would be helpful to know if its at/near the simulation boundary.

JustinRayAngus:
The unique thing about this CI test is that it uses a MultiFab for the beta coefficient of the curl curl solver in AMReX. I forced this coefficient using setVal() to be 1.0, but it did not fix the problem. 

@WeiqunZhang do you have a different test problem that uses the MultiFab for the beta coefficient that you could try? The geometry is 1D.

EZoni:
By the way, I think `-DAMReX_TESTING=ON` is the one that sets initial MultiFab data to `NaN`:
https://github.com/AMReX-Codes/amrex/blob/5e764e55c3cf549b97f73c2c1f3b64047a8b4e10/Docs/sphinx_documentation/source/BuildingAMReX.rst?plain=1#L496-L497

My guess would be that some of the initial data remains `NaN` and is accessed at some point. May be data in the ghost cells that isn't filled, data that should not be accessed at all, or something along these lines. Does the erroneous arithmetic operation also occurr if you run with one grid and one MPI process?

JustinRayAngus:
> By the way, I think `-DAMReX_TESTING=ON` is the one that sets initial MultiFab data to `NaN`: https://github.com/AMReX-Codes/amrex/blob/5e764e55c3cf549b97f73c2c1f3b64047a8b4e10/Docs/sphinx_documentation/source/BuildingAMReX.rst?plain=1#L496-L497
> 
> My guess would be that some of the initial data remains `NaN` and is accessed at some point. May be data in the ghost cells that isn't filled, data that should not be accessed at all, or something along these lines. Does the erroneous arithmetic operation also occurr if you run with one grid and one MPI process?

Using one grid and one MPI process did not fix the error.

However, using one grid with on MPI process AND not using -DAMReX_TESTING=ON seems to fix the problem. With this flag OFF, it stills fails in the same way with 2 grids on 2 procs.

WeiqunZhang:
There is a bug in the 1d curl curl solver. It's fixed in https://github.com/AMReX-Codes/amrex/pull/4407.

JustinRayAngus:
> There is a bug in the 1d curl curl solver. It's fixed in [AMReX-Codes/amrex#4407](https://github.com/AMReX-Codes/amrex/pull/4407).

Thanks Weiqun for looking into this. I have verified that this fixes the CI test problem.

EZoni:
Glad you fixed the bug. Once https://github.com/AMReX-Codes/amrex/pull/4407 is merged, I can fetch it in #5814 and then you can rebase this PR once #5814 is merged.

JustinRayAngus:
@dpgrote CI test is now implemented. I think I've address each of your comments now.

dpgrote (approved):
This looks good to me! I think it's ready to be merged. Did you want to get comments from anyone else, since this is a somewhat large PR?

JustinRayAngus:
> This looks good to me! I think it's ready to be merged. Did you want to get comments from anyone else, since this is a somewhat large PR?

It is large, but I think it is rather straight-forward and basic. There will be plenty of time for feedback during the next steps, which should be doable in smaller PRs.

PR #5770, opened 2025-03-17
MacroscopicEvolveE: pass macroscopic_solver_algo to MacroscopicEvolveE as a parameter instead of using WarpX::macroscopic_solver_algo

lucafedeli88:
With this PR  `macroscopic_solver_algo` is passed to  `MacroscopicEvolveE` as an argument, instead of directly accessing the static variable of the WarpX class `WarpX::macroscopic_solver_algo` .

Note that after this change `macroscopic_solver_algo` does not need to be static. Therefore, it is renamed `m_macroscopic_solver_algo` and it is changed into a private member variable of the WarpX class.


aeriforme (approved):
Looks good to me. 
Do I understand correctly that the main advantage of this is removing 
`#include "WarpX.H"`?

lucafedeli88:
Thanks for the review, @aeriforme ! There are two advantages in my opinion: 
- not including the massive `WarpX.H` header, as you mentioned, is an advantage. On a single file it won't have much impact, but if we remove it from many files this could speed up the compilation a little bit.
- reading/writing parameters via static variables of another class is a potentially confusing design choice, because from the interface of a function you can't see which parameters affect how the function work. I am slowly trying to remove occurrences of this pattern from the code.

PR #5776, opened 2025-03-17
direct gather routine for single vector field

JustinRayAngus:
A direct gather routine for B is used in doGatherPicnic() (only used by implicit solver) and will also be used in the mass matrix deposit routine in PR #5768. Making a separate routine for a direct Gather of the B-field will reduce code duplication.

This PR adds a direct gather routine for a single vector field, which can be used for E or B and accounts for the galerkin_interpolation flag in parallel and perpendicular directions.

@RemiLehe This PR also seems to overlap with PR #5673.


dpgrote:
This routine seems to duplicate what is in `doGatherShapeN` (except for the Galerkin factor) which is doing direct deposition. It looks like a single kernel could be written that gathers the three field components that would be called separately for E and B. It would need parallel and perpendicular Galerkin factors for the E and B.

JustinRayAngus:
> This routine seems to duplicate what is in `doGatherShapeN` (except for the Galerkin factor) which is doing direct deposition. It looks like a single kernel could be written that gathers the three field components that would be called separately for E and B. It would need parallel and perpendicular Galerkin factors for the E and B.

I noticed that too. Originally, this routine was optimized for the implicit solver where a Yee grid is assumed. Now that I have made the routine more generalized (and less optimized), it is now basically the same as doGatherShapeN for a single vector field without the galerkin_interpolation flag. There is no need to use that flag for B. What do you suggest we do?

JustinRayAngus:
@dpgrote I've generalized the routine for an arbitrary vector field F with separate depos_order template parameters for parallel and perpendicular gathers.

JustinRayAngus:
FYI. The file diff is very misleading now. I added the function doGatherVectorFieldShapeN() above doGatherShapeN() (which is required) and then modified the latter to call the former. However, the file diff recognizes this as many separate changes to the doGatherShapeN() routine rather than one green block for doGatherVectorFieldShapeN() and then another few red/green blocks showing the modifications to doGatherShapeN().

Source/Particles/Gather/FieldGather.H
```diff
@@ -16,413 +16,305 @@
...
+ * \param dinv                    3D cell size inverse
+ * \param xyzmin                  The lower bounds of the domain
+ * \param lo                      Index lower bounds of domain
+ * \param n_rz_azimuthal_modes    Number of azimuthal modes when using RZ geometry
  */
-template <int depos_order, int galerkin_interpolation>
+template <int depos_order_perp, int depos_order_para>
 AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
-void doGatherShapeN ([[maybe_unused]] const amrex::ParticleReal xp,
+void doGatherVectorFieldShapeN (
```
dpgrote: To be more precise, can this name be changed to `doDirectGatherVectorFieldShapeN`. And the same for `doGatherShapeN`, change to `doDirectGatherShapeN`.

Source/Particles/Gather/FieldGather.H
```diff
@@ -16,413 +16,305 @@
...
-#else // defined(WARPX_DIM_3D)
-    // Gather field on particle Exp from field on grid ex_arr
-    for (int iz=0; iz<=depos_order; iz++){
-        for (int iy=0; iy<=depos_order; iy++){
-            for (int ix=0; ix<= depos_order - galerkin_interpolation; ix++){
-                Exp += sx_ex[ix]*sy_ex[iy]*sz_ex[iz]*
-                    ex_arr(lo.x+j_ex+ix, lo.y+k_ex+iy, lo.z+l_ex+iz);
+    // Gather vector field F
+    amrex::Real weight;
+    for (int i=0; i<=depos_order_para; i++) {
```
dpgrote: For optimization, the loop ordering should be swapped, with `i` the innermost loop. The same for all of the nested loops.

Source/Particles/Gather/FieldGather.H
```diff
@@ -16,413 +16,305 @@
...
+                                  - Fz_arr(lo.x+ip_Fz+i, lo.y+kp_Fz+k, 0, 2*imode)*xy.imag());
+                Fzp += weight_Fz*dFz;
+                xy = xy*xy0;
             }
+#endif
         }
     }
+
+#if defined(WARPX_DIM_RZ)
+    // Convert Fxp and Fyp (which are actually Fr and Fth) to Fx and Fy
```
dpgrote: This routine is adding to previously calculated fields, so for RZ, this transformation needs to be done on only the fields calculated in this routine. The fields need to be accumulated into temporaries and the transform done on those before adding them to the total fields. You can see this in the original `doGatherShapeN. Also, note that there the XZ and RZ are fully separated.

Source/Particles/Gather/FieldGather.H
```diff
@@ -16,413 +16,305 @@
...
-    }
-    if ((ex_type[0] == NODE) || (by_type[0] == NODE) || (bz_type[0] == NODE)) {
-        j_node_v = compute_shape_factor_galerkin(sx_node_galerkin, x);
-    }
-    if ((ex_type[0] == CELL) || (by_type[0] == CELL) || (bz_type[0] == CELL)) {
-        j_cell_v = compute_shape_factor_galerkin(sx_cell_galerkin, x - 0.5_rt);
+    // Compute particle position in grid units
+#if defined(WARPX_DIM_RZ)
+    amrex::Real const rp = std::sqrt(xp*xp + yp*yp);
+    amrex::Real costheta, sintheta;
```
dpgrote: In other places, this bit of code is written like this, which is cleaner and allows use of `const`.
```
    amrex::Real const costheta = (rp > 0. ? xp/rp : 1.);;
    amrex::Real const sintheta = (rp > 0. ? yp/rp : 0.);;
```

JustinRayAngus: There are actually many places in Gather.H where the former is used. I agree that this way is cleaner. Will change the other in a different PR.

Source/Particles/Gather/FieldGather.H
JustinRayAngus: I didn't know this. The Villasenor deposition routine also uses the bad ordering. Should that be changed as well? Separate PR of course.

Source/Particles/Gather/FieldGather.H
JustinRayAngus: I'm ok with renaming the new function, but I'm hesitant to change doGatherShapeN. Seems there are a lot of functions that start with that, so you could argue that all of them should change as well. That seems like something for a separate PR.

Also, why have ShapeN in the function names? I feel like it adds no value here.

Source/Particles/Gather/FieldGather.H
JustinRayAngus: I think this routine does what you describe. It is also identical to how it is done in doGatherShapeNEsirkepovStencilImplicit() and in doGatherPicnicShapeN().

In either case, I made a small change that I think makes it more clear.

JustinRayAngus: Discussed offline. Fixed now.

JustinRayAngus:
Documenting recent commits. To reduce code duplication, we originally changed doGatherShapeN() to call the new doDirectGatherSingleField() once for E and once for B. @dpgrote and myself did some timing tests on CPU and GPU (Lassen and Perlmutter) using the 3D uniform_plasma input deck. There was no noticeable difference on CPU, but there was a 36-50% increase in the GatherAndPush timings on GPU. The increase seems to be because the shape factors in each direction are now computed twice.

In order to not affect timing associated with GatherAndPush, doGatherShapeN() is restored to its original form and the new routine is only called by the implicit routines as needed. However, this means that there is a fair amount of duplicate code.

Source/Particles/Gather/FieldGather.H
```diff
@@ -15,6 +15,264 @@
...
+                                  - Fz_arr(lo.x+ip_Fz+i, lo.y+kp_Fz+k, 0, 2*imode)*xy.imag());
+                Fzp += weight_Fz*dFz;
+            }
+        }
+        xy = xy*xy0;
+
+    }
+
+    // Convert Frp and Fthp to Fxp and Fyp
+    Fxp = costheta*Frp  - sintheta*Fthp;
```
dpgrote: These should be `+=` to be consistent with the fields being accumulated.

Source/Particles/Gather/FieldGather.H
```diff
@@ -495,22 +753,10 @@ void doGatherShapeNEsirkepovStencilImplicit (
 #if defined(WARPX_DIM_RZ)
     amrex::Real const xp_new = xp_np1;
     amrex::Real const yp_new = yp_np1;
-    amrex::Real const xp_mid = xp_nph;
```
dpgrote, line 498: The changes in this routine are unrelated to the main topic of the PR so should be reverted to keep the PR minimal.

Source/Particles/Gather/FieldGather.H
```diff
@@ -15,6 +15,264 @@
 
 #include <AMReX.H>
 
+/**
```
dpgrote, line 18: Since this is a somewhat special purpose routine, maybe move this further down in the file, perhaps to just before `doGatherPicnicShapeN`.

Source/Particles/Gather/FieldGather.H
JustinRayAngus: Since it is just a small change and makes things consistent with the cleaner implementation of the new routine, I would prefer to keep it.

Source/Particles/Gather/FieldGather.H
JustinRayAngus: If we decide to use the new routine in doGatherShapeN, then it has to be before that routine. If I move it and someone wants to test out the timings again by calling it from doGatherShapeN, then they would have to move it back to where it is now.

dpgrote (approved):
Looks good to me!

PR #5780, opened 2025-03-18
Fix FieldPoyntingFlux diagnostic for GPU

dpgrote:
The FieldPoyntingFlux reduced diagnostic uses `amrex::ReduceOps` to calculate the diagnostic. The loops are structured so that the calculation is only done on the boundaries, by limiting the boxes to only include the boundary cells. For any blocks in the interior, the box will be empty. The `amrex::ReduceOps` on an empty box works Ok on the CPU but crashes on the GPU. The solution is to only call the reduction when the box is not empty.


JustinRayAngus (approved):
Thanks Dave!

PR #5785, opened 2025-03-20
Tests: `amrex.the_arena_init_size=0`

ax3l:
`amrex.the_arena_init_size=0` is default on CPU but not GPU. In order to run tests in parallel on the same GPU, we need to change the default so the first test does not allocate 3/4th of the memory of the GPU (to speed up later allocations in a re-allocated heap).

This currently only applies to app tests, but not yet to the Python CTests. Need to add an extra option for those, because we cannot inject CLI options. This will be a follow-up, because we need to inject a few vars (debugger breaking, init size, warning thresholds) and this might need a WarpX-Python specific test env var or something of that sort.


EZoni (approved):
Thanks, @ax3l. Looks good to me.

@WeiqunZhang 
I will wait for your approval and then merge.

WeiqunZhang (approved):

EZoni:
Weiqun approved as well, I just merged `development` in (since this was a few weeks old) and enabled auto-merge.

PR #5789, opened 2025-03-21
HankelTransform : remove unnecessary #include "WarpX.H"

lucafedeli88:
This PR removes an unnecessary include.


dpgrote (approved):
Thanks for the fix!

PR #5809, opened 2025-04-02
Fix bad memory access in 1D and 2D in VandB deposition

dpgrote:
The `x` and/or `y` particle arrays were being accessed when running 1D and 2D where they are set to null pointers. The fix is to add the appropriate macros around the accesses. This was causing seg faults when running the test case on my Mac.

### DIFF
--- a/Source/Particles/Deposition/CurrentDeposition.H
+++ b/Source/Particles/Deposition/CurrentDeposition.H
@@ -1974,9 +1974,9 @@ void doVillasenorDepositionShapeNExplicit (const GetParticlePosition<PIdx>& GetP
  * \param n_rz_azimuthal_modes  Number of azimuthal modes when using RZ geometry.
  */
 template <int depos_order>
-void doVillasenorDepositionShapeNImplicit ([[maybe_unused]]const amrex::ParticleReal * const xp_n,
-                                           [[maybe_unused]]const amrex::ParticleReal * const yp_n,
-                                           [[maybe_unused]]const amrex::ParticleReal * const zp_n,
+void doVillasenorDepositionShapeNImplicit ([[maybe_unused]]const amrex::ParticleReal * const xp_n_data,
+                                           [[maybe_unused]]const amrex::ParticleReal * const yp_n_data,
+                                           [[maybe_unused]]const amrex::ParticleReal * const zp_n_data,
                                            const GetParticlePosition<PIdx>& GetPosition,
                                            const amrex::ParticleReal * const wp,
                                            [[maybe_unused]]const amrex::ParticleReal * const uxp_n,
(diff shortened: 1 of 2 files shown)


Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -2034,11 +2034,23 @@ void doVillasenorDepositionShapeNImplicit ([[maybe_unused]]const amrex::Particle
...
+#if (AMREX_SPACEDIM >= 2)
+            amrex::ParticleReal const xp_n = xp_n_data[ip];
+#else
+            amrex::ParticleReal const xp_n = 0._prt;
+#endif
+#if defined(WARPX_DIM_3D) || defined(WARPX_DIM_RZ)
+            amrex::ParticleReal const yp_n = yp_n_data[ip];
+#else
+            amrex::ParticleReal const yp_n = 0._prt;
+#endif
```
lucafedeli88, line 2046: I am wondering if something like this could be more readable. What do you think?

```suggestion
            amrex::ParticleReal const xp_n = AMREX_D_PICK(0._prt, xp_n_data[ip], xp_n_data[ip]);
            amrex::ParticleReal const yp_n = AMREX_D_PICK(0._prt, 0._prt,               yp_n_data[ip]);
```

lucafedeli88 (approved):
Thanks a lot, @dpgrote .
I approve these changes. I left an optional suggestion that may be considered to possibly improve readability.

@dpgrote , you have mentioned that you discovered this issue while running an example. Do you think that it would be useful to add this case to the Undefined Behavior sanitizer CI test ( inside `clang_sanitizers.yml`) ?

Source/Particles/Deposition/CurrentDeposition.H
dpgrote: Yes, I can add this to the sanitizer. This is in the `build_UB_sanitizer` job?

I've been wanting to find a use for `AMREX_D_PICK`, but unfortunately, the `yp_n` version won't work since the value is needed for RZ, which is 2D, but not for XZ, which is also 2D.

lucafedeli88: ah, yes, you are right!  Your solution is then the best one. 

If you want to add this case to the sanitizer CI test, you have to modify the file `clang_sanitizers.yml` , in particular the section:

```
    - name: run with UB sanitizer
      run: |

        export OMP_NUM_THREADS=2

        #MPI implementations often leak memory
        export "ASAN_OPTIONS=detect_leaks=0"

        mpirun -n 2 ./build/bin/warpx.rz Examples/Physics_applications/laser_acceleration/inputs_base_rz
        mpirun -n 2 ./build/bin/warpx.1d Examples/Physics_applications/laser_acceleration/inputs_base_1d
        mpirun -n 2 ./build/bin/warpx.2d Examples/Physics_applications/laser_acceleration/inputs_base_2d
        mpirun -n 2 ./build/bin/warpx.3d Examples/Physics_applications/laser_acceleration/inputs_base_3d
```

For the moment, we use the UB sanitizer only for the `laser_acceleration` example (1D, 2D, RZ, 3D), which uses only a fraction of the features of WarpX. Running these examples require only few minutes, while the compilation of WarpX may take ~ 1h. Therefore, we can add many more test cases with a negligible impact on the total runtime of this test.

PR #5811, opened 2025-04-03
Add feature running variance deposition

clarkse:
is a single pass algorithm that can compute current and variance using accumulation buffers to mitigate requiring a second pass through the particle list when computing variance. It is used to get temperature deposition with matching shape functions as the current density multifabs for applying relaxation algorithms in the hybrid solver that depend on computing ion moments.


clarkse:
@dpgrote I wanted to sync up with you on this as it augments your 2 pass variance update with a possible single pass routine for online temperature calculations. I wanted to get your thoughts on this. There is an opportunity here to integrate it with the current deposition as well to mitigate any runtime penalties, however I think the time of deposition may not match.

I am looking for some algorithm review, we are currently finishing out testing on it. I need to add a few more input checks, docs, and figure out how to integrate with the existing code in a way that makes sense.

Source/Particles/Deposition/VarianceAccumulationBuffer.cpp
```diff
@@ -0,0 +1,212 @@
...
+                        }
+
+                        // Apply Normalization in any owned cell
+                        if (w_arr(i,j,k) > 0.0_rt) {
+                            amrex::Real denom = (w_arr(i,j,k) - w2_arr(i,j,k)/w_arr(i,j,k) );
+                            if (denom > 10._rt * std::numeric_limits<amrex::Real>::epsilon()) {
+                                variance_arr(i,j,k) /= (w_arr(i,j,k) - w2_arr(i,j,k)/w_arr(i,j,k) );
+                            } else {
+                                // This occurs when only a single sample is within a bin
+                                // In this case shoudl be zero variance
```
dpgrote: Typo "should"

Source/Particles/Deposition/VarianceAccumulationBuffer.cpp
```diff
@@ -0,0 +1,212 @@
...
+        warpx.m_fields.alloc_init("w2_" + species_name, Direction{1}, lev, bay, dmy, ncomps, ngy, 0.0_rt);
+        warpx.m_fields.alloc_init("w2_" + species_name, Direction{2}, lev, baz, dmz, ncomps, ngz, 0.0_rt);
+
+        warpx.m_fields.alloc_init("vbar_" + species_name, Direction{0}, lev, bax, dmx, ncomps, ngx, 0.0_rt);
+        warpx.m_fields.alloc_init("vbar_" + species_name, Direction{1}, lev, bay, dmy, ncomps, ngy, 0.0_rt);
+        warpx.m_fields.alloc_init("vbar_" + species_name, Direction{2}, lev, baz, dmz, ncomps, ngz, 0.0_rt);
+    }
+
+    // Grab references to these arrays to store in this data structure
+    this->w = warpx.m_fields.get_mr_levels_alldirs("w_" + species_name, warpx.finestLevel());
```
dpgrote: It would be better to fetch the MFs from the registry each time they are needed rather than saving references. The main issues is that during a load balance, these saved references would become invalid. Here, you would want to save only the name of the MF.

Source/Particles/Deposition/VarianceAccumulationBuffer.H
```diff
@@ -0,0 +1,35 @@
...
+
+#include <ablastr/fields/MultiFabRegister.H>
+
+namespace warpx::particles::deposition {
+
+class VarianceAccumulationBuffer {
+    public:
+
+    // Accumulation vector fields, aliased to fields in register
+    ablastr::fields::MultiLevelVectorField w;
```
dpgrote: WarpX uses the standard that class attributes have the prefix "m_".

Source/Particles/Deposition/VarianceAccumulationBuffer.cpp
```diff
@@ -0,0 +1,212 @@
...
+using namespace warpx::particles::deposition;
+
+VarianceAccumulationBuffer::VarianceAccumulationBuffer (ablastr::fields::MultiLevelVectorField const& T_vf, std::string const& species_name )
+{
+    using ablastr::fields::Direction;
+    auto & warpx = WarpX::GetInstance();
+
+    const int ncomps = 1;
+
+    for (int lev = 0; lev <= warpx.finestLevel(); ++lev) {
```
dpgrote, line 31: It looks likes this could be simplified by using a loop over the dimensions,
```
for (int idir=0 ; idir < 3 ; idir++) {
    amrex::BoxArray const& ba = T_vf[lev][Direction{idir}]->boxArray();
    amrex::DistributionMapping const& dm = T_vf[lev][Direction{idir}]->DistributionMap();
    amrex::IntVect const& ng = T_vf[lev][Direction{idir}]->nGrowVect();
    warpx.m_fields.alloc_init("w_" + species_name, Direction{idir}, lev, ba, dm, ncomps, ng, 0.0_rt);
    ...
}
```

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -425,6 +429,51 @@ PhysicalParticleContainer::PhysicalParticleContainer (AmrCore* amr_core, int isp
...
+    if (do_temperature_deposition) {
+        using ablastr::fields::Direction;
+
+        auto& warpx = WarpX::GetInstance();
+        ablastr::fields::MultiLevelVectorField J_vf =
+            warpx.m_fields.get_mr_levels_alldirs(warpx::fields::FieldType::current_fp, warpx.finestLevel());
+
+        const std::string T_field_name = "T_" + species_name;
+
+        for (int lev = 0; lev <= warpx.finestLevel(); ++lev) {
```
dpgrote, line 453: This could also be simplified using a loop over dimensions.

Source/Particles/Deposition/VarianceAccumulationBuffer.cpp
clarkse: I have updated the VarianceAccumulationBuffer to store the speciens name and added a getter method to retrieve the MF pointers from the buffer object that grabs a fresh reference from the MF Register.

clarkse:
@dpgrote I have addressed your comments, fixed the behavior of the cross term to avoid underflow, and added a few more lines to the docs. I also renamed the NGP method for diagnostics, as we still want that around for outputting temperatures when the online accumulation is not required. I still need to clean up the cylinder compression CI test, but I think this should be close to being ready for merge. I will update the CI tests and get benchmarks straightened out if you think it is good to go.

Docs/source/usage/parameters.rst
```diff
@@ -1502,6 +1502,10 @@ Particle initialization
     Resampling is performed everytime the number of macroparticles per cell of the species
     averaged over the whole simulation domain exceeds this parameter.
 
+* ``<species>.do_temperature_deposition`` (`boolean`) optional (default `false`)
+    When running with Ohm's Law Hybrid Solver, this will enable temperature deposition
+    in each dimension with a matched shape function and filtering used for current deposition.
+    This is required when using the electron energy solver with electron-ion temperature relaxation.
```
dpgrote, line 1517: When the electron energy solver is being used, is `do_temperature_deposition` turned on automatically? I don't see that here.

Source/Diagnostics/Diagnostics.cpp
```diff
@@ -308,6 +308,28 @@ Diagnostics::BaseReadParameters ()
                 + ".fields_to_plot does not match any species"
             );
         }
+
+        // Check if m_varnames contains a string of the form T_<species_name>
+        if (var.rfind("Tx_", 0) == 0 || var.rfind("Ty_", 0) == 0 || var.rfind("Tz_", 0) == 0) {
```
dpgrote, line 316: Should this be setting `do_temperature_deposition` for the species? This would allow this diagnostic to be used without the hybrid solver.

Source/Particles/Deposition/VarianceAccumulationBuffer.cpp
```diff
@@ -0,0 +1,210 @@
...
+        for (int idir = 0; idir < 3; ++idir) {
+            amrex::BoxArray const& ba = T_vf[lev][Direction{idir}]->boxArray();
+            amrex::DistributionMapping const& dm = T_vf[lev][Direction{idir}]->DistributionMap();
+            amrex::IntVect const& ng = T_vf[lev][Direction{idir}]->nGrowVect();
+
```
dpgrote, line 36: Maybe use names that are more specific, something like "variance_w".

dpgrote:
This is looking good to me, though I have a couple of minor comments. Maybe get a review from someone else also since this is a fairly big PR.

Docs/source/usage/parameters.rst
clarkse: We are still working to debug and test that PR internally. I can add a reference to it when we submit the PR to the WarpX main repo. This would be in preparation for the electron energy solver. I figured I would forget to update this comment later.

Source/Diagnostics/Diagnostics.cpp
clarkse: I guess we could automatically turn it on, but I wasn't sure if I should just remove the isotropic temperature diagnostic and force component wise outputs. I opted to leave the existing isotropic 2 pass scheme in to not change the default behavior of using the temperature diagnostic. I could change the temperature diagnostic functor for isotropic T_<species> to run this algorithm and then add the components in quadrature, but based on the comments previously about keeping memory footprint down I thought that having the existing working code that you wrote would be fine. I am up for doing whatever makes more sense. I can see from a code cleanliness perspective that removing the 2 pass algorithm would keep things a bit more trim, however I did not want to change the interface here. Perhaps a topic for Tuesday morning?

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -63,7 +64,12 @@ void WarpX::HybridPICEvolveFields ()
     // Perform charge deposition in component 0 of rho_fp at t_{n+1}.
     mypc->DepositCharge(m_fields.get_mr_levels(FieldType::rho_fp, finest_level), 0._rt);
     // Perform current deposition at t_{n+1/2}.
-    mypc->DepositCurrent(m_fields.get_mr_levels_alldirs(FieldType::current_fp, finest_level), dt[0], -0.5_rt * dt[0]);
+    mypc->DepositCurrent(m_fields.get_mr_levels_alldirs(FieldType::current_fp, finest_level),
+        dt[0], -0.5_rt * dt[0]);
+
+    // TODO: Perhaps add flag here for when using temperature accumulation in Hybrid
```
roelof-groenewald: Please do add the flag.
I realize `DepositTemperatures()` will return without really doing anything if all species have `pc->getTemperatureDepositionFlag() == false`, but since the temperature can be output as a diagnostic this could lead to inefficiency. If someone does not want the heat equation coupled to the Ohm solver but would like to have the individual temperature components as a diagnostic, this would result in lots of unnecessary depositions.

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -85,8 +91,21 @@ void WarpX::HybridPICEvolveFields ()
     // for the hybrid-PIC solver since current values are interpolated to
     // a nodal grid
     for (int lev = 0; lev <= finest_level; ++lev) {
+        ablastr::utils::communication::FillBoundary(
+            *m_fields.get(FieldType::rho_fp, lev),
+            m_fields.get(FieldType::rho_fp, lev)->nGrowVect(),
+            WarpX::do_single_precision_comms,
+            Geom(lev).periodicity(),
+            true
+        );
```
roelof-groenewald: Was this `FillBoundary` call added due to an observed bug? If so, could you please open a separate PR for it with a description of the symptom?

I'm asking since `FillBoundary` takes a substantial percentage of runtime currently, so if this is not known to fix a bug we should not add it.

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -85,8 +91,21 @@ void WarpX::HybridPICEvolveFields ()
...
         for (int idim = 0; idim < 3; ++idim) {
-            m_fields.get(FieldType::current_fp, Direction{idim}, lev)->FillBoundary(Geom(lev).periodicity());
+            ablastr::utils::communication::FillBoundary(
```
roelof-groenewald: What is the reason for this change?

Source/Particles/MultiParticleContainer.H
```diff
@@ -167,6 +167,21 @@ public:
     DepositCurrent (ablastr::fields::MultiLevelVectorField const& J,
                     amrex::Real dt, amrex::Real relative_time);
 
+    /**
+     * \brief Deposit temperatures to species MFs. This is done for each
+     *        species and can be used in the future for a single pass update of rho, J, and T.
+     *
+     * \param[in,out] fields Multifab field register to grab Temeprature MFs
```
roelof-groenewald: Typo on "Temperature".

Source/Particles/MultiParticleContainer.H
```diff
@@ -167,6 +167,21 @@ public:
...
+     * \param[in] dt Time step for particle level
+     * \param[in] relative_time Time at which to deposit velocities, rel
+     *                          current positions of the partic
```
roelof-groenewald: ~the partic~

Source/Particles/PhysicalParticleContainer.H
```diff
@@ -395,6 +415,11 @@ protected:
     // A flag to enable saving of the previous timestep positions
     bool m_save_previous_position = false;
 
+    // Flag controlling whether or not variance and then temperatures are computed per species
+    bool do_temperature_deposition = false;
```
roelof-groenewald: I guess to stick with convention this should be `m_do_temperature_deposition`.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -3296,3 +3335,252 @@ PhysicalParticleContainer::getPairGenerationFilterFunc ()
...
+
+/* \brief Temperature Deposition for thread thread_num
+ * \param pti         Particle iterator
+ * \param wp          Array of particle weights
+ * \param uxp uyp uzp Array of particle momenta
+ * \param ion_lev     Pointer to array of particle ionization level. This is
+                      required to have the charge of each macroparticle
+                      since q is a scalar. For non-ionizable species,
+                      ion_lev is a null pointer.
+ * \param Tx Ty Tz    Full array of current density
```
roelof-groenewald: `Full array of temperature components`

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -3296,3 +3335,252 @@ PhysicalParticleContainer::getPairGenerationFilterFunc ()
...
+ * \param offset      Index of first particle for which current is deposited
+ * \param np_to_deposit Number of particles for which current is deposited.
```
roelof-groenewald: "current" -> "temperature" throughout this function.

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -1357,7 +1360,7 @@ WarpXParticleContainer::GetChargeDensity (int lev, bool local)
  * \param lev         Level of box that contains particles
  */
 void
-WarpXParticleContainer::DepositTemperature (amrex::MultiFab* temperature, const int lev)
+WarpXParticleContainer::DepositIsotropicNGPTemperature (amrex::MultiFab* temperature, const int lev)
```
roelof-groenewald: This is maybe a bit nit-picky, but I would say "total" or "average" temperature is better than "isotropic", since isotropic, to me, implies the temperature **is** the same in all components, rather than it being a single value collecting the temperature in all directions. Certainly not a big deal though.

roelof-groenewald (commented):
Thanks for this PR. Overall looks good! I just left a few initial comments. I still need to go through the changes in the `Deposition` folder in detail as well.

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
clarkse: This is intended to be used as a tightly coupled deposition loop for use with the electron energy equation. This can easily be adapted to run similarly to the NGP Total temperature at diagnostic intervals, but it wasn't really the purpose of the PR. I can change a few things around to do this, but I was hoping leaving the existing functionality would be fine for now. I can get back to this a bit later to clean it up for diagnostic use, but it may be a little bit before I can clean up and test that use case.

Source/Particles/Deposition/TemperatureDeposition.H
```diff
@@ -0,0 +1,463 @@
...
+    const amrex::Array4<amrex::Real> & wy_arr,
+    const amrex::Array4<amrex::Real> & wz_arr,
+    const amrex::Array4<amrex::Real> & w2x_arr,
+    const amrex::Array4<amrex::Real> & w2y_arr,
+    const amrex::Array4<amrex::Real> & w2z_arr,
+    const amrex::Array4<amrex::Real> & vxbar_arr,
+    const amrex::Array4<amrex::Real> & vybar_arr,
+    const amrex::Array4<amrex::Real> & vzbar_arr
+)
+{
```
dpgrote, line 59: Looking at this closely, unfortunately this doesn't look thread safe. The RHS depends on the LHS, but other threads maybe contributing to the quantities so they become inconsistent. For example, between setting `wx_old` and `wxsum`, other threads may also be adding to `wx_arr` so the `old` and `sum` are not consistent. And the same for other quantities below.

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
clarkse: I added the fillboundary with a nodal sync because I observed in the temperature deposition cases that the corner cells can have reduced statistics when not summed properly or synchronized. This ensures that the nodal value on the high end of domain and the corners is updated with the value of the owner box. I was seeing cases where the values looked right, but in w_sum only half or a quarter of the values were present. I updated rho to behave the same way.

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
clarkse: same as above for rho.

Source/Particles/Deposition/TemperatureDeposition.H
clarkse: So I have been thinking about this a bit and I will ultimately use a blocking semaphore or mutex to control this over multiple operations. I need to figure out the best way to do it in the AMReX sense using atomic_ops. Likely will need a slight change in AMReX to make it work appropriately.

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
roelof-groenewald: I wonder if that isn't a separate issue. There are `FillBoundary` calls which only sets guard cell values and there are `WarpXSumGuardCells` calls which adds values deposited in tiles' guard cells to the valid cells in neighboring tiles. It sounds like what you are describing is a need to call `WarpXSumGuardCells` for the `w_sum` multifab.

It seems to me that calling `FillBoundary` should not be necessary for `rho_fp` since rho is nodal and therefore all interpolations to other grids are onto grids contained by the rho grid, i.e., rho never needs to access guard cells during the solve step and therefore it does not need accurate guard cell values (what FillBoundary would do). Of course if there is some filtering applied this would not be true anymore since the filter might reach onto guard cells, but if that is the case the `FillBoundary` call should come before `SyncCurrentAndRho` (and in that case there should maybe be a larger discussion since I believe the same issue would occur in the Maxwell solver).

If you still think this should be added, could you please open a new PR with just these `FillBoundary` changes? I think that would be better since the changes are not specific to the new feature from this PR, and I'm curious to see if any of the benchmark values for the hybrid tests change.

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
roelof-groenewald: I guess this question was more about whether there is a specific reason to call `ablastr::utils::communication::FillBoundary(...)` rather than using `mf->FillBoundary(...)` as is currently done?

clarkse: I guess I was transitioning over to using the WarpX wrappers to make sure that the "single precision comms" is respected and the wrapper has the flag for doing the nodal sync. I am mainly trying to make the boundary fills be done in the same way for each MF.

Source/Particles/Deposition/VarianceAccumulationBuffer.cpp
clarkse: This has been updated to prepend "variance_buffer_" to the field name for the register.

Examples/Tests/ohm_solver_cylinder_compression/inputs_test_3d_ohm_solver_cylinder_compression_picmi.py
roelof-groenewald, line 1: This test takes ~160 seconds to run on Azure. Do you think it would be possible to make it faster without losing the value of testing? Perhaps fewer timesteps or lower resolution?

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -260,8 +265,21 @@ void WarpX::HybridPICDepositRhoAndJ ()
     // for the hybrid-PIC solver since current values are interpolated to
     // a nodal grid
     for (int lev = 0; lev <= finest_level; ++lev) {
+        ablastr::utils::communication::FillBoundary(
+            *m_fields.get(FieldType::rho_fp, lev),
+            m_fields.get(FieldType::rho_fp, lev)->nGrowVect(),
+            WarpX::do_single_precision_comms,
+            Geom(lev).periodicity(),
+            true
+        );
```
roelof-groenewald, line 274: I don't think this is necessary since `rho` is nodal. This would just add extra communication which is already limiting the hybrid-solver's speed in some cases. Can we remove it?

Source/Particles/MultiParticleContainer.H
```diff
@@ -169,6 +169,20 @@ public:
     DepositCurrent (ablastr::fields::MultiLevelVectorField const& J,
                     amrex::Real dt, amrex::Real relative_time);
 
+    /**
+     * \brief Deposit temperature to species MFs. This is done for each
+     *        species and can be used in the future for a single pass update of rho, J, and T.
+     *
+     * \param[in,out] fields Multifab field register to grab Temperature MFs
+     * \param[in] dt Time step for particle level
```
roelof-groenewald: `dt` is not actually in the argument list.

Source/Particles/Deposition/TemperatureDeposition.H
```diff
@@ -0,0 +1,493 @@
...
+    using namespace amrex::literals;
+
+    constexpr int NODE = amrex::IndexType::NODE;
+    constexpr int CELL = amrex::IndexType::CELL;
+
+    // --- Compute shape factors
+    Compute_shape_factor< depos_order > const compute_shape_factor;
+#if !defined(WARPX_DIM_1D_Z)
+    // x direction
+    // Get particle position after 1/2 push back in position
```
roelof-groenewald: This comment seems to be left over from something else? The 1/2 push back is not necessarily happening (depending on `relative_time`).

Source/Particles/Deposition/TemperatureDeposition.H
```diff
@@ -0,0 +1,493 @@
...
+
+            // --- Get particle quantities
+            const amrex::ParticleReal gaminv = 1.0_rt/std::sqrt(1.0_rt + uxp[ip]*uxp[ip]*clightsq
+                                                        + uyp[ip]*uyp[ip]*clightsq
+                                                        + uzp[ip]*uzp[ip]*clightsq);
+            const amrex::ParticleReal vx  = uxp[ip]*gaminv;
+            const amrex::ParticleReal vy  = uyp[ip]*gaminv;
+            const amrex::ParticleReal vz  = uzp[ip]*gaminv;
+
+            amrex::ParticleReal wq  = q*wp[ip];
```
roelof-groenewald, line 472: It doesn't have to be in this PR since I know this is modeled after the existing deposition routines, but it would be nice to be able to deposit neutral species quanties too, i.e., not weight the macroparticles by charge before deposition, but rather multiply density or velocity fields by charge afterwards when charge or current densities are required.
I guess since this file is specifically called `TemperatureDeposition` it could be good to already change this in the current PR...

Source/Particles/Deposition/VarianceAccumulationBuffer.cpp
```diff
@@ -0,0 +1,208 @@
...
+                            // Update Mean and Variance
+                            vbar_arr(i,j,k) = (vsum_a + w_b*v_b)/w_arr(i,j,k);
+                            variance_arr(i,j,k) +=  variance_old_arr(i,j,k) + dwa*dwb/w_arr(i,j,k);
+                        }
+
+                        // Apply Normalization in any owned cell
+                        if (w_arr(i,j,k) > 0.0_rt) {
+                            amrex::Real denom = (w_arr(i,j,k) - w2_arr(i,j,k)/w_arr(i,j,k) );
+                            if (denom > 10._rt * std::numeric_limits<amrex::Real>::epsilon()) {
+                                variance_arr(i,j,k) /= (w_arr(i,j,k) - w2_arr(i,j,k)/w_arr(i,j,k) );
```
roelof-groenewald:
```suggestion
                                variance_arr(i,j,k) /= denom;
```

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -3456,3 +3495,254 @@ PhysicalParticleContainer::getPairGenerationFilterFunc ()
...
+    // or when depositing inside the level (depos_lev=lev)
+    Box tilebox;
+    if (lev == depos_lev) {
+        tilebox = pti.tilebox();
+    } else {
+        const IntVect& ref_ratio = WarpX::RefRatio(depos_lev);
+        tilebox = amrex::coarsen(pti.tilebox(),ref_ratio);
+    }
+
+    // ng_J.diagShift(1);
```
roelof-groenewald:
```suggestion
```

Source/Particles/WarpXParticleContainer.H
```diff
@@ -334,9 +336,43 @@ public:
...
                                 PushType push_type);
-
-    void DepositTemperature (amrex::MultiFab* temperature, const int lev);
-    std::unique_ptr<amrex::MultiFab> GetTemperature (int lev);
+    /**
+     * \brief Accumulate velocity moments for matched temperature deposition using particle shapes
+     *
+     * \param[in,out] T_vf vector field of Temperatures (one three-dimensional array of pointers
+     *                to MultiFabs per mesh refinement level)
+     * \param[in] dt Time step for particle level
```
roelof-groenewald: `dt` not in argument list.

roelof-groenewald (commented):
It would be nice to make an inline comment pointing to a reference where the online algorithm is described. Perhaps https://doi.org/10.1145%2F359146.359153?

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
roelof-groenewald: Oh, sorry I just reread your comment above and realized we already addressed this.

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -240,6 +241,10 @@ void WarpX::HybridPICDepositRhoAndJ ()
     // Perform current deposition at t_{n-1/2}.
     mypc->DepositCurrent(m_fields.get_mr_levels_alldirs(FieldType::current_fp, finest_level), dt[0], -0.5_rt * dt[0]);
 
+    // TODO: Perhaps add flag here for when using temperature accumulation in Hybrid
+    // Perform Temperature Deposition at time t_{n+1}
```
roelof-groenewald: Should this be? 
```suggestion
    // Perform Temperature Deposition at time t_{n}
```
Or is the comment on line 241 above wrong?

roelof-groenewald: I see the comment above the current deposition is wrong...

Source/Particles/Deposition/TemperatureDeposition.H
clarkse: I pondered this one a bit regarding rewriting the entire deposition loop to at least deposit density and temperature at the same time but since they can be deposited onto different grids and at different time staggerings I thought that it would be too cumbersome to catch all of the edge cases. I do want to revisit it at some point though since we could collapse 3 particle loops into one here for density, current, and temperature. I do think that it could easily be adapted to be able to calculate neutral temperatures though, but perhaps for another PR for that since it would require linking it up with the appropriate solvers, etc. I would like to punt on this to keep this PR from growing in scope further.

Examples/Tests/ohm_solver_cylinder_compression/inputs_test_3d_ohm_solver_cylinder_compression_picmi.py
clarkse: I am reducing NPPC, NX, NY, and NZ by a factor of 2 to reduce the time as well as only output one diagnostic interval.

Source/Particles/Deposition/TemperatureDeposition.H
roelof-groenewald: Sounds good!

Source/Particles/Deposition/VarianceAccumulationBuffer.cpp
```diff
@@ -0,0 +1,208 @@
...
+            WarpXSumGuardCells(w2mf_old, w2mf, periodicity, w2mf.nGrowVect(), 0, 1);
+            WarpXSumGuardCells(variancemf_old, variancemf, periodicity, variancemf.nGrowVect(), 0, 1);
+
+            amrex::Gpu::streamSynchronize();
+
+            // Since we need to do a weighted combination subtract off the old data
+            amrex::MultiFab::Subtract(wmf_old, wmf, 0, 0, 1, wmf_old.nGrowVect());
+            amrex::MultiFab::Subtract(w2mf_old, w2mf, 0, 0, 1, w2mf_old.nGrowVect());
+            amrex::MultiFab::Subtract(vbarmf_old, vbarmf, 0, 0, 1, vbarmf_old.nGrowVect());
+            amrex::MultiFab::Subtract(variancemf_old, variancemf, 0, 0, 1, variancemf_old.nGrowVect());
```
roelof-groenewald: I'm still just trying to understand what is happening here.

Am I right that `wmf_old`, `w2mf_old`, etc. at this point only holds the values deposited in the overlapping cells from neighboring tiles? I'm not following what is happening in the kernel below.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -3456,3 +3495,253 @@ PhysicalParticleContainer::getPairGenerationFilterFunc ()
...
+        for (int idim = 0; idim < 3; ++idim) {
+            // Multiplies internal cells to convert from variance to temperature
+            T_vf[lev][Direction{idim}]->mult(Tnorm, 0, 1);
+
+            // Synchronize Ghost cells after normalization.
+            ablastr::utils::communication::FillBoundary(
+                *T_vf[lev][Direction{idim}],
+                WarpX::do_single_precision_comms,
+                WarpX::GetInstance().Geom(lev).periodicity(),
+                true);
```
roelof-groenewald: I wonder if it wouldn't be cleaner to add the `FillBoundary` call at the end of `SynchronizeBoundaryAndNormalizeVariance`, so that the MF that comes out of that function is properly populated. Then the normalization done here could be done in each tiles guard cells as well and since it is a local operation it would not require re-communicating the guard values. What do you think?

Source/Particles/PhysicalParticleContainer.H
```diff
@@ -361,6 +363,23 @@ public:
         return GetVecOfPtrs(m_user_real_attrib_parser);
     }
 
+    // By default this returns false, can be overridden in PysicalParticleContainer
```
roelof-groenewald:
```suggestion
    // By default this returns false; overridden here to return runtime value
```

Source/Particles/PhysicalParticleContainer.cpp
clarkse: I moved the normalization and filtering code into the function so a strange MF in an odd state doesn't come out of the routine.

Source/Particles/Deposition/VarianceAccumulationBuffer.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+            // Do a local copy of the weight sum to vbarmf_old
+            amrex::MultiFab::Copy(vbarmf_old, wmf, 0, 0, 1, vbarmf_old.nGrowVect());
+            amrex::Gpu::streamSynchronize();
+
+            // Do a local multiply of the weight sum by vbar so we end up with w_a * vbar_a
+            // This is the first part of teh vbar update, which is being computed as
+            // vbar_ab = (wsum_a * vbar_a + wsum_b * vbar_b) / wsum_ab,  where the owner cell
+            // is a and b would be contributions from overlapping non-owner cells.
+            // This can support tile corners wehre more than 2 boxes overlap as well since the
```
roelof-groenewald:
```suggestion
            // This is the first part of the vbar update, which is being computed as
            // vbar_ab = (wsum_a * vbar_a + wsum_b * vbar_b) / wsum_ab,  where the owner cell
            // is a and b would be contributions from overlapping non-owner cells.
            // This can support tile corners where more than 2 boxes overlap as well since the
```

Source/Particles/Deposition/VarianceAccumulationBuffer.cpp
```diff
@@ -0,0 +1,257 @@
...
+            // weights are combined prior to summantion via the Guard Cell summation.
```
roelof-groenewald:
```suggestion
            // weights are combined prior to summation via the Guard Cell summation.
```

roelof-groenewald (approved):
Thanks for this addition! I just left two comments for minor typo fixes.

Source/Particles/Deposition/VarianceAccumulationBuffer.cpp
clarkse: Hopefully the additional comments help with unraveling the algorithm and the gymnastics that were done.

Source/Particles/Deposition/TemperatureDeposition.H
clarkse: The thread safety was resolved with this commit https://github.com/BLAST-WarpX/warpx/pull/5811/commits/30fb26b29c89d58042377599acbcb9717ea1659d

PR #5812, opened 2025-04-04
Initial partition based particle density profile

WeiqunZhang:
This adds the option to do initial partition based on particle density profile so that none of the Boxes has too many particles.

Two new runtime parameters `warpx.init_partition_species` and `warpx.init_partition_efficiency` are introduced. If `warpx.init_partition_species` is set (e.g., to `electron`), we will query `electron.profile`, `electron.parse_density_function` and `electron.density_min`. If all these parameters are present, we will try to divide the original BoxArray until no Box has more particles than the threshold defined as the total number of particles divided by the number of processes and then multiplied by `warpx.init_partition_efficiency`.


n01r:
Thanks a lot, @WeiqunZhang! I'll give this a try! :tada:

Source/Initialization/WarpXInitData.cpp
```diff
@@ -245,6 +245,95 @@ WarpX::PostProcessBaseGrids (BoxArray& ba0) const
...
+        ParmParse pp_warpx("warpx");
+        std::string species;
+        pp_warpx.query("init_partition_species", species);
+        pp_warpx.query("init_partition_efficiency", efficiency_target);
+        if ( ! species.empty()) {
+            ParmParse pp(species);
+            std::string profile;
+            pp.query("profile", profile);
+            if (profile == "parse_density_function") {
+                if (pp.queryline("density_function(x,y,z)", density_function) &&
```
n01r: I am getting a compile time error:
```
/global/homes/m/mgarten/src/warpx/Source/Initialization/WarpXInitData.cpp(264): error: class "amrex::ParmParse" has no member "queryline"
                  if (pp.queryline("density_function(x,y,z)", density_function) &&
                         ^

/global/homes/m/mgarten/src/warpx/Source/Initialization/WarpXInitData.cpp(282): error: The enclosing parent function ("PostProcessBaseGrids") for an extended __device__ lambda cannot have private or protected access within its class
                                                    (int b, int i, int j, int k)
                                                    ^

2 errors detected in the compilation of "/global/homes/m/mgarten/src/warpx/Source/Initialization/WarpXInitData.cpp".
gmake[2]: *** [CMakeFiles/lib_2d.dir/build.make:2042: CMakeFiles/lib_2d.dir/Source/Initialization/WarpXInitData.cpp.o] Error 2
gmake[2]: *** Waiting for unfinished jobs....
gmake[1]: *** [CMakeFiles/Makefile2:3674: CMakeFiles/lib_2d.dir/all] Error 2
```

In `Source/Utils/SpeciesUtils.cpp` we are using 
```
std::string str_density_function;
utils::parser::Store_parserString(pp_species, source_name, "density_function(x,y,z)", str_density_function);  
// Construct InjectorDensity with InjectorDensityParser.
density_parser = std::make_unique<amrex::Parser>(
    utils::parser::makeParser(str_density_function,{"x","y","z"})); 
```

WeiqunZhang: Looks like it's because it uses an old version of amrex.

WeiqunZhang: Re: the second error, a fix is pushed. Needed to make PostProcessBaseGrids a public function for CUDA.

Also pushed a change to avoid a new amrex function.

n01r: Thanks a lot! :)

n01r:
Hi @WeiqunZhang,

The result looks really good!
At step 0, we already have 56.7% load balance efficiency.
image

At step 100, after the first load-balancing iteration, we went up to 88.4%.
image

There are these artifacts still in there, I think (if they don't come from my plotting), where we get extremely long boxes vertically. Do you see where they might come from?

WeiqunZhang:
I don't see any reason that we would get extremely long boxes.

Source/Initialization/WarpXInitData.cpp
```diff
@@ -245,6 +245,96 @@ WarpX::PostProcessBaseGrids (BoxArray& ba0) const
...
+
+    bool init_partition_species = false;
+    std::string density_function;
+    Real density_min=0;
+    Real efficiency_target = 0.7;
+    if (ParallelDescriptor::NProcs() > 1)
+    {
+        ParmParse pp_warpx("warpx");
+        std::string species;
+        pp_warpx.query("init_partition_species", species);
```
RemiLehe: We could remove this input parameter and instead always loop over all species.

RemiLehe: Instead we use a boolean flag `split_high_density_boxes`?

RemiLehe: `split_high_density_boxes_threshold` with a default of `1.1` (so that it does not split boxes if particles are uniformly distributed)

Source/Initialization/WarpXInitData.cpp
```diff
@@ -245,6 +245,96 @@ WarpX::PostProcessBaseGrids (BoxArray& ba0) const
...
+        pp_warpx.query("init_partition_efficiency", efficiency_target);
+        if ( ! species.empty()) {
+            ParmParse pp(species);
+            std::string profile;
+            pp.query("profile", profile);
+            if (profile == "parse_density_function") {
```
RemiLehe: We should add warning here, for species that don't have a `parse_density_function`, to let the user know that they were ignored.

Source/Initialization/WarpXInitData.cpp
```diff
@@ -245,6 +245,96 @@ WarpX::PostProcessBaseGrids (BoxArray& ba0) const
         AMREX_D_TERM(},},})
         ba0 = BoxArray(std::move(bl));
     }
+
```
RemiLehe, line 310: Could you add some more comments @WeiqunZhang

n01r:
@WeiqunZhang, is it possible to add a "fake" density profile of a species that we won't actually initialize? Can we use this to anticipate the load later in the simulation?

ax3l:
@WeiqunZhang Thank you for this PR!

As a next step to get KISMET and INCITE sims like those to run, I think what we need is a functionality to dynamically split boxes during simulation runtime, if too many particles move into it.

Merging boxes again is another challenge for later, we did something along these lines in I/O a while back where we can lend an algorithm from: https://doi.org/10.1109/TPDS.2021.3100784

ax3l:
@n01r Can you pls share on Slack or email the inputs files for the scenario benchmarked here?

I think this PR makes perfect sense and we now want to work on a follow-up where we dynamically split the boxes behind the target once particles start moving into them :)

Docs/source/usage/parameters.rst
```diff
@@ -823,6 +823,19 @@ Distribution across MPI ranks and parallelization
     initialization by avoiding putting neighboring boxes on the same
     process.
 
+* ``warpx.split_high_density_boxes`` (`bool`) optional (default: false)
```
ax3l, line 826: We could keep the name like this, even though it is only init.

And later on we use the same name fro dynamic splitting (during init _and_ at runtime).

What do you think @WeiqunZhang ? Or should we do `_init` named options for now?

n01r:
> @n01r Can you pls share on Slack or email the inputs files for the scenario benchmarked here?
> 

A copy of the corresponding directory with the input deck is here (NERSC):
```/global/cfs/cdirs/m4546/mgarten/projects/kismet/2025/043_256_node_test_Weiqun_partitioning/```

Alternatively, here's a GDrive link that will only work for LBL-internal folks
https://drive.google.com/drive/folders/1KL6uOKMUCsljRcCRw620VCZV5a8ETniW?usp=sharing

File: Docs/source/usage/parameters.rst (hunk 5 above)
Comment (WeiqunZhang) [reply Docs/source/usage/parameters.rst:826]: It sounds good to keep the name.

**Review by ax3l — APPROVED** [review]

**Comment by ax3l** [comment]
I think I found a bug in a smaller example (#6107) #6158 , which might help us understand these large boxes we saw above, too.

File: Source/Initialization/WarpXInitData.cpp [hunk 6]
```diff
@@ -307,6 +307,143 @@ WarpX::PostProcessBaseGrids (BoxArray& ba0) const
...
+            ParmParse pp_spec(species);
+            pp_spec.query("profile", profile);
+            bool split_using_this_species = false;
+            if (profile == "parse_density_function" &&
+                pp_spec.queryline("density_function(x,y,z)", density_function))
+            {
+                split_using_this_species = true;
+                utils::parser::queryWithParser(pp_spec, "density_min", density_min);
+                std::vector<int> nppc_v(3,1);
+                utils::parser::getArrWithParser(pp_spec, "num_particles_per_cell_each_dim", nppc_v);
```
ax3l, line 349: This needs to be generalized to also support `num_particles_per_cell` for `injection_style=NRandomPerCell`.

(This is not related to #6107 -- the example there is using `num_particles_per_cell_each_dim`)

titoiride: Uh, that may be the reason why I haven't seen any effects yet. I'm indeed using `injection_style = NRandomPerCell` with `num_particles_per_cell`

Docs/source/usage/parameters.rst
ax3l: Opened tracking for dynamic splitting in #6158

Source/Initialization/WarpXInitData.cpp
ax3l: Adding support for this in #6159 now

PR #5813, opened 2025-04-05
Fix new bug in particle BTD

atmyers:
This fixes a new issue in the back-transformed diagnostics for particles that was introduced in #5571. The problem affected runs where not all particle species were written into the diagnostic. In that case, the index of the particle buffer and the index of the species in the `MultiParticleContainer` are not the same.


RemiLehe (approved):

EZoni:
Thanks for the fix, @atmyers! Do you think there is any chance to update any of the existing CI tests that use BTD so that this case is covered as well?

PR #5682, opened 2025-02-19
Complete Implementation of PSATD JRhom Algorithm

oshapoval:
This PR should replace older attempts and/or incorporate changes from older PRs.

**Old user interface:**
`algo.maxwell_solver = psatd`
`warpx.do_multi_J = 1`
`warpx.do_multi_J_n_depositions = 2`
`psatd.J_in_time = linear`
`psatd.rho_in_time = linear`

**New user interface:**
`algo.maxwell_solver = psatd`
`psatd.JRhom = "LL2"`

Here's the list of older PRs, in order from most recent to oldest:
- #5647
- #5623
- #5229
- #5088 
- #4263 
- #4261
- #3506

## Notes
Renamed
- `JInTime` to `TimeDependencyJ` (`AMREX_ENUM`)
- `RhoInTime` to `TimeDependencyRho` (`AMREX_ENUM`)
- `J_in_time` to `time_dependency_J` (internal variable)
- `rho_in_time` to `time_dependency_rho` (internal variable)

## To-do
- ~Merge `OneStep_nosub` and `OneStep_psatd_JRhom` for PSATD case~
- [x] Adapt PICMI to new user interface
  - [x] Add new parameter `psatd.JRhom`
  - [x] Remove obsolete parameters
- [x] Clean up code, update documentation

PR #5820, opened 2025-04-08
Remove buggy synchronization at end of steps

### LINKED ISSUE #5727: Move velocity and position synchronization to appropriate place in PIC loop
This is a follow-up issue from #1751, to fix the comment at https://github.com/ECP-WarpX/WarpX/blob/b858e36c24dab324f4fd951b62540ec76843ea76/Source/Evolve/WarpXEvolve.cpp#L248

See also https://github.com/ECP-WarpX/WarpX/pull/1751#discussion_r1978274734.

dpgrote:
This PR does cleanup, removing a potential bug that happens when the velocity is synchronized with position in the final diagnostic. The particle boundary conditions had not been applied so the field gather could be out of bounds.

Further cleanup is done, renaming Synchronize to SynchronizeVelocityWithPosition, and having it check is_synchronized, since the synchronization may have already been done.

Note that the factor `1e-3` in the code change
```cpp
bool const end_of_step_loop = (step == numsteps_max - 1) || (cur_time + dt[0] >= stop_time - 1.e-3*dt[0]);
```
is a somewhat arbitrary factor to allow for round off error in the time. If the `cur_time + dt[0]` is almost but not quite equal to `stop_time`, then that's close enough. Comment copied here from https://github.com/BLAST-WarpX/warpx/pull/5820#discussion_r2076467690 for better visibility.


RemiLehe:
@dpgrote Thanks for this PR! Would you be able to merge `development` into this PR and update the checksums?

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -79,39 +79,36 @@ namespace
...
-            lev,
-            0.5_rt*dt[lev],
-            *m_fields.get(FieldType::Efield_aux, Direction{0}, lev),
-            *m_fields.get(FieldType::Efield_aux, Direction{1}, lev),
-            *m_fields.get(FieldType::Efield_aux, Direction{2}, lev),
-            *m_fields.get(FieldType::Bfield_aux, Direction{0}, lev),
-            *m_fields.get(FieldType::Bfield_aux, Direction{1}, lev),
-            *m_fields.get(FieldType::Bfield_aux, Direction{2}, lev)
-        );
+    if (!is_synchronized) {
```
EZoni: This may be one of those examples where making it clear that `is_synchronized` is a member variable of the `WarpX` class, through the prefix `m_`, would help not get lost while reading the code. For a minute I was trying to understand where `is_synchronized` is defined in this function. We could rename it according to the WarpX style rules in a separate PR.

EZoni:
Two big PRs are being merged as we speak:
- #3630 
- #5682 

Since those PRs introduce new CI tests, I would recommend that we merge `development` here, after those PRs are merged, before resetting the checksums in this PR.

aeriforme:
I could use this fix after #3630 was merged.
I have some simulations (that use that PR)  where the last time step fails, not matter which step it is. 
I merged this branch and it works fine.

EZoni:
After the last merge of `development`  the checksums need a new update. Once that's done, I think this PR would be ready to approve and merge. Let's aim at having this merged before the hackathon of May 12-14.

EZoni:
Thanks for all the updates, @dpgrote. I think `test_3d_effective_potential_electrostatic_picmi` is the only remaining test that needs a checksum update. It could be nice to merge this before the release update in #5865.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -351,12 +337,13 @@ WarpX::Evolve (int numsteps)
         }
 
         bool const do_diagnostic = (multi_diags->DoComputeAndPack(step) || reduced_diags->DoDiags(step));
+        bool const end_of_step_loop = (step == numsteps_max - 1) || (cur_time + dt[0] >= stop_time - 1.e-3*dt[0]);
```
EZoni, line 340: Just a question to clarify, where does the `1.e-3` factor come from here?

Examples/Tests/electrostatic_sphere/analysis_electrostatic_sphere.py
```diff
@@ -184,7 +184,11 @@ def return_energies(iteration):
     print("Checking conservation of energy")
     Ek_i, Ep_i = return_energies(0)
     Ek_f, Ep_f = return_energies(30)
+    print(f"Ek_i + Ep_i = {Ek_i} + {Ep_i} = {Ek_i + Ep_i}")
+    print(f"Ek_f + Ep_f = {Ek_f} + {Ep_f} = {Ek_f + Ep_f}")
+    print(f"(Ek_i + Ep_i) - (Ek_f + Ep_f) = {(Ek_i + Ep_i) - (Ek_f + Ep_f)}")
+    print(f"Energy change tolerance = 0.0032*(Ek_i + Ep_i) = {0.0032 * (Ek_i + Ep_i)}")
     assert Ep_f < 0.7 * Ep_i  # Check that potential energy changes significantly
-    assert abs((Ek_i + Ep_i) - (Ek_f + Ep_f)) < 0.003 * (
+    assert abs((Ek_i + Ep_i) - (Ek_f + Ep_f)) < 0.0032 * (
```
dpgrote, line 192: I don't know why the error is worse here than before. Note that this has `warpx.synchronize_velocity_for_diagnostics` turned on. With it off, the error is a factor of several worse.

Source/Evolve/WarpXEvolve.cpp
dpgrote: It's a somewhat arbitrary factor to allow for round off error in the time. If the `cur_time + dt` is almost but not quite equal to `stop_time`, then that's close enough.

EZoni (approved):

PR #5825, opened 2025-04-10
WarnManager.cpp : display informations of warning causing warpx to stop

grobertdautun:
This PR adds informations about the warning that causes warpx to abort when a warning threshold is set. 

If the user doesn't explicitely ask for `always_warn_immediately`, a warning above the abort threshold will abort warpx without ever displaying which warning caused it. I changed the message to include the warning in this case.

Example with `warpx.abort_on_warning_threshold = low` and a particle diagnostics without position :
```
0::Assertion `msg_priority < abort_priority' failed, file "/home/gabriel/applications/warpx/WarpX/Source/ablastr/warn_manager/WarnManager.cpp", line 101, Msg: 
### ERROR   : A warning with priority 'low' has been raised:
#            [Diagnostics] diag1.electron.variables contains no particle
#            positions!
 !!!
SIGABRT
```


lucafedeli88 (approved):
Thanks @grobertdautun , I think that this is a useful feature.

Source/ablastr/warn_manager/WarnManager.cpp
```diff
@@ -93,12 +93,14 @@ void WarnManager::RecordWarning(
         } else if(m_abort_on_warning_threshold == WarnPriority::medium) {
             abort_priority = abl_msg_logger::Priority::medium;
         }
+        std::string abort_msg = "A warning with priority '"
+            + abl_msg_logger::PriorityToString(abort_priority)
+            + "' has been raised";
+        abort_msg += ((m_always_warn_immediately) ? "." : (" :\n[" + topic + "] " + text));
```
EZoni: Thanks for the PR! 

Could you post an example of message with the new implementation, either here or in the PR description?

Is the white space between `raised` and (potentially) `:` intentional or should we remove it?
```suggestion
        std::string abort_msg = "A warning with priority '"
            + abl_msg_logger::PriorityToString(abort_priority)
            + "' has been raised";
        abort_msg += ((m_always_warn_immediately) ? "." : (":\n[" + topic + "] " + text));
```

grobertdautun: I added an example in the description and removed the white space (which was intentional but is a french thing)

EZoni: Ah, I see! Thank you for the updates.

EZoni (approved):

PR #5826, opened 2025-04-11
CI tests: re-enable clang thread sanitizer test

lucafedeli88:
This PR re-enables the thread sanitizer CI test.
The issues that we experienced (https://github.com/BLAST-WarpX/warpx/pull/5492) were ultimately caused by the fact that thread sanitizer is not able to automatically load `libarcher` under some Linux distributions like Ubuntu, and  `libarcher` is required to avoid false positives with OpenMP ! The fix is  adding this line  ( see https://hpc-wiki.info/hpc/ThreadSanitizer ):
```
export OMP_TOOL_LIBRARIES=/usr/lib/llvm-17/lib/libarcher.so
``` 
 In addition to this change, the PR adds some comments. Moreover, WarpX is now compiled always with EB support, since this is now the default and since this (slightly) reduces the time required to run this CI test.

### DIFF
--- a/.github/workflows/clang_sanitizers.yml
+++ b/.github/workflows/clang_sanitizers.yml
@@ -118,24 +118,9 @@ jobs:
           -DWarpX_QED=ON                    \
           -DWarpX_QED_TABLE_GEN=ON          \
           -DWarpX_OPENPMD=ON                \
-          -DWarpX_EB=OFF                    \
           -DWarpX_PRECISION=DOUBLE          \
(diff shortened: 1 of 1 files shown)


EZoni (approved):
Thanks a lot, @lucafedeli88!

Could you add a note in the PR description explaining what steps/logic you followed to find out that the problem was related to the Archer library? Just to leave it as an insight note for posterity, so that one can learn the debugging rationale when looking at this PR.

PR #5828, opened 2025-04-11
ablastr constants : add variable templates to allow for flexibility in using single/double precision

lucafedeli88:
With this PR the constants in ablastr are first defined using variable templates and then specialized for `amrex::Real`, e.g., for mathematical constants:

```C++
    /** Mathematical constants */
    namespace math
    {
        //! ratio of a circle's circumference to its diameter (variable template)
        template<typename T>
        constexpr auto pi_v = static_cast<T>(3.14159265358979323846);

        //! https://tauday.com/tau-manifesto (variable template)
        template<typename T>
        constexpr auto tau_v = static_cast<T>(2.0 * pi_v<double>);

        //! ratio of a circle's circumference to its diameter
        constexpr auto pi = pi_v<amrex::Real>;

        //! https://tauday.com/tau-manifesto
        constexpr auto tau = tau_v<amrex::Real>;

    } // namespace math

```

This is done with the intent of allowing the use of `double` precision constants even if the code is compiled in single precision (or vice versa if you really want it !), e.g. :

```C++
const auto a = 2_rt * math::pi; // a is an amrex::Real

const auto b = 2 * math::pi_v<double>; // b is a double !

const auto c = 2 * math::pi_v<float>; // c is a float !
```

I will need this feature to import the QED module from PICSAR into WarpX. I've started migrating the code in  https://github.com/BLAST-WarpX/warpx/pull/5677 ,but the PR is becoming too big, so I decided to split it into smaller chunks.

The new feature is now used in the template function `Algorithms::KineticEnergy` . This function is conceived to be able to force double precision in some cases regardless of how WarpX has been compiled. Therefore, using `c` in double precision in these cases is coherent with its goal.

Note that I've changed the variables from `static constexpr` to `constexpr` because, to my understanding, `static` does not make any difference in this context. 

## Previous discussions

With this PR the constants in ablastr are first defined using variable templates and then specialized for `amrex::Real`, e.g., for mathematical constants:

```C++
    namespace math
    {
        namespace variable_templates
        {

            template<typename T>
            constexpr auto pi = static_cast<T>(3.14159265358979323846);

            template<typename T>
            constexpr auto tau = static_cast<T>(2.0 * pi<double>);

        } 

        constexpr auto pi = variable_templates::pi<amrex::Real>;

        constexpr auto tau = variable_templates::tau<amrex::Real>;

    } // namespace math

```

This is done with the intent of allowing the use of `double` precision constants even if the code is compiled in single precision (or vice versa if you really want it !), e.g. :

```C++
using namespace ablastr;
namespace math_vt = math::variable_templates;

const auto a = 2_rt * math::pi; // a is an amrex::Real

const auto b = 2 * math_vt::pi<double>; // b is a double !
```

I will need this feature to import the QED module from PICSAR into WarpX. I've started migrating the code in  https://github.com/BLAST-WarpX/warpx/pull/5677 ,but the PR is becoming too big, so I decided to split it into smaller chunks.

The new feature is now used in the template function `Algorithms::KineticEnergy` . This function is conceived to be able to force double precision in some cases regardless of how WarpX has been compiled. Therefore, using `c` in double precision in these cases is coherent with its goal.

In case you don't like the long namespace name `variable_template`, an alternative would be to call it `vt`. 

Note that I've changed the variables from `static constexpr` to `constexpr` because, to my understanding, `static` does not make any difference in this context. 

### Updates
- Following @dpgrote 's suggestion, I've changed `variable_templates` to `const_templates`
- Following @aeriforme 's suggestion, I've added `c2`, `invc` and `inv_c2` constants, and I have tested them by modifying a couple of source files.
- Following @WeiqunZhang  's suggestion, I am now using `_v` suffix to indicate variable templates, as done in C++20 https://en.cppreference.com/w/cpp/numeric/constants.html


Source/ablastr/constant.H
```diff
@@ -17,13 +17,24 @@ namespace ablastr::constant
     /** Mathematical constants */
     namespace math
     {
-        using namespace amrex::literals;
+        namespace variable_templates
```
dpgrote: Since this is for constants, perhaps call it `constant_templates`?

lucafedeli88: Sorry for not having replied earlier, @dpgrote !
Would `const_templates` work according to you?

dpgrote: Thanks, it's better, but `const` has a specific meaning in C++ so that could cause confusion.

dpgrote:
Would there be much penalty if the constants were all explicitly declared double? This might generate warnings of explicit down casting when single precision is being used, but would there be a performance penalty?

Source/ablastr/constant.H
lucafedeli88: I think that you are right, @dpgrote . I've changed the name to `constant_templates`, as you originally suggested.

lucafedeli88:
> Would there be much penalty if the constants were all explicitly declared double? This might generate warnings of explicit down casting when single precision is being used, but would there be a performance penalty?

The main issue I see is that, due to our extensive use of the `auto` keyword, inserting just one constant in double precision may silently transform whole blocks of calculations to double precision. I am afraid that this might have an impact on performances. What I like of the approach followed in this PR is that everything works as before unless you specifically want your constants with a given precision. The verbosity of `constant_templates` is a feature: you only use it when you really need it!

Source/ablastr/constant.H
```diff
@@ -38,54 +49,158 @@ namespace ablastr::constant
...
-        using namespace amrex::literals;
+        namespace constant_templates
+        {
+            //! vacuum speed of light [m/s] (variable template)
+            template<typename T>
+            constexpr auto c = static_cast<T>(299'792'458);
+
+            //! vacuum permittivity: dielectric permittivity of vacuum [F/m] (variable template)
+            template<typename T>
+            constexpr auto ep0 = static_cast<T>(8.8541878188e-12);
```
aeriforme: Can we call it `epsilon0` or, my favorite, `epsilon_0`?

Source/ablastr/constant.H
```diff
@@ -38,54 +49,158 @@ namespace ablastr::constant
...
+            template<typename T>
+            constexpr auto alpha = static_cast<T>(0.0072973525643330135);
+
+            //! classical electron radius = 1./(4*pi*ep0) * q_e*q_e/(m_e*c*c) [m] (variable template)
+            template<typename T>
+            constexpr auto r_e = static_cast<T>(2.8179403205e-15);
+
+            //! xi times c2 = xi*c*c. This should be usable for single precision instead of xi; very close to smallest float32 number possible (1.2e-38) (variable template)
+            template<typename T>
+            constexpr auto xi_c2 = static_cast<T>(1.1728865075613984e-35);
```
aeriforme: What is `xi`? :thinking:

Source/ablastr/constant.H
```diff
@@ -38,54 +49,158 @@ namespace ablastr::constant
...
+
+        } // namespace constant_templates
 
         //! vacuum speed of light [m/s]
-        static constexpr auto c = 299'792'458._rt;
+        constexpr auto c = constant_templates::c<amrex::Real>;
+
         //! vacuum permittivity: dielectric permittivity of vacuum [F/m]
-        static constexpr auto ep0 = 8.8541878188e-12_rt;
+        constexpr auto ep0 = constant_templates::ep0<amrex::Real>;
```
aeriforme: Same as above

aeriforme:
We make massive and inconsistent use of $c^{-2}, c^{-1}, c^2$ and similar.
Should we add them here, so that we don't have to re-calculate them every time?

lucafedeli88:
> We make massive and inconsistent use of c − 2 , c − 1 , c 2 and similar. Should we add them here, so that we don't have to re-calculate them every time?

We often use `constexpr` when we define these constants in the code, so this should not result into an overhead.
However, we can consider adding these constants.

lucafedeli88:
> We make massive and inconsistent use of c − 2 , c − 1 , c 2 and similar. Should we add them here, so that we don't have to re-calculate them every time?

@aeriforme , I've added `c2`, `invc`, and `invc2` constants, and I have tested them by modifying a couple of source files. If we like the new constants we can use them everywhere in WarpX, maybe by making the change with another PR.

lucafedeli88:
ping, @aeriforme  ;-)

Source/ablastr/constant.H
```diff
@@ -38,54 +49,179 @@ namespace ablastr::constant
...
+        constexpr auto alpha = constant_templates::alpha<amrex::Real>;
+
         //! classical electron radius = 1./(4*pi*ep0) * q_e*q_e/(m_e*c*c) [m]
-        static constexpr auto r_e = 2.8179403205e-15_rt;
+        constexpr auto r_e = constant_templates::r_e<amrex::Real>;
+
         //! xi: nonlinearity parameter of Heisenberg-Euler effective theory = (2.*alpha*alpha*ep0*ep0*hbar*hbar*hbar)/(45.*m_e*m_e*m_e*m_e*c*c*c*c*c)
-        static constexpr double xi = 1.3050122383827227e-52;
+        constexpr double xi = 1.3050122383827227e-52;
+
```
EZoni, line 180: Maybe we could take this opportunity to rewrite the docstring from
```cpp
//! xi: nonlinearity parameter of Heisenberg-Euler effective theory = (2.*alpha*alpha*ep0*ep0*hbar*hbar*hbar)/(45.*m_e*m_e*m_e*m_e*c*c*c*c*c)
```
to
```cpp
//! xi: nonlinearity parameter of Heisenberg-Euler effective theory = (2.*alpha**2*ep0**2*hbar**3)/(45.*m_e**4*c**5)
```

Source/ablastr/constant.H
```diff
@@ -17,13 +17,24 @@ namespace ablastr::constant
     /** Mathematical constants */
     namespace math
     {
-        using namespace amrex::literals;
+        namespace constant_templates
```
EZoni: I'm wondering whether we could drop the `templates` part in the name...

Obviously it does describe what this is, as far as the implementation is concerned, but from the perspective of this being used in the code I wonder if `PhysConst::constant_templates::invc2<amrex::Real>` is really "needed", as opposed to, say, simply `PhysConst::constants::invc2<amrex::Real>`?

EZoni (commented):
Thanks for the PR, @lucafedeli88! I left only a couple of minor comments.

Source/ablastr/constant.H
lucafedeli88: @EZoni , what about an abbreviation, such as `tmpl`  (constant is already in `PhysConst`):
e.g. 
`PhysConst::tmpl::invc2<amrex::Real>`

?

EZoni: Hm, I think I'm a bit confused. Can we try to summarize all namespaces we have currently in `development`?

I'm seeing
https://github.com/BLAST-WarpX/warpx/blob/b4009ed9944ef4262140d1afb498573823c24352/Source/Utils/WarpXConst.H#L14-L22

and also
https://github.com/BLAST-WarpX/warpx/blob/b4009ed9944ef4262140d1afb498573823c24352/Source/ablastr/constant.H#L15-L89

So, we can do, e.g.,
```cpp
PhysConst::c
```
but we cannot do
```cpp
PhysConst::constant::c
```
right?

Where does the conflict with, say, `PhysConst::constant::c<amrex::Real>` (although I would use `constants`) come from? When I say "conflict" I refer to what you mentioned about `constant` being already in `PhysConst`.

Is the argument that `PhysConst::constant_templates::c<amrex::Real>` is more intuitive than `PhysConst::constant::c<amrex::Real>` because with the latter one doesn't immediately foresee that they should pass the template parameter?

EZoni: @lucafedeli88 

I think we are close to finalizing this PR, as long as we resolve the latest questions raised in https://github.com/BLAST-WarpX/warpx/pull/5828#discussion_r2140801417.

Source/ablastr/constant.H
lucafedeli88: Done ! Thanks for the suggestion, @EZoni .

Source/ablastr/constant.H
lucafedeli88: Sorry for not having replied earlier, @EZoni !

If you have the nested namespace structure that we have in ablastr and WarpX you can do - and can't do - the following :

```
namespace ablastr::constant 
{ 
    namespace SI
    {
        constexpr double  c = 299792458.0;
    }
}

 namespace PhysConst 
 { 
     using namespace ablastr::constant::SI; 
 } 

void test ()
{
    auto ttA = ablastr::constant::SI::c;

    auto ttB = PhysConst::c;

    // ILLEGAL
    // auto ttC = c;
    // auto ttC = SI::c;
    // auto ttC = constant::SI::c;
    // auto ttC = PhysConst::SI::c;
    // auto ttC = PhysConst::constant::SI::c;
    // auto ttC = PhysConst::ablastr::constant::SI::c;
}

```

What you suggest can work if we do the following :

```
namespace ablastr::constant 
{ 
    namespace SI
    {
        namespace constants 
        {
            template <typename T>
            constexpr T c = static_cast<T>(299792458.0);
        }

        constexpr double  c = constants::c<double>;

    }
}

 namespace PhysConst 
 { 
     using namespace ablastr::constant::SI; 
 } 

void test ()
{
    auto ttA = PhysConst::c;

    auto ttB = PhysConst::constants::c<float>;
}
```

However, having `ablastr::constant::SI::constants::c<T>` disturbs me, since we confusingly repeat almost the same name in the nested namespace hierarchy. That's why I suggest something like : 

 `ablastr::constant::SI::tmpl::c<T>` 
 
 for templated constants . 
 
 What we can't do is having a template and a non-template constant with the same name in the same namespace.
 
 
 I also considered setting a default template value, but we need to use the suffix `<>` to use it, which is not very practical ! 
 ```
 template <typename T = double>
constexpr T c = static_cast<T>(299792458.0);

auto tt = c<>;
```

**Comment by WeiqunZhang** [comment]
Why not following the C++ standard's approach? Once we start using C++20, it will also be easier to switch to std::numbers.
```
double a = PhysConst::c;
float b = PhysConst:c_v<float>;
```
https://en.cppreference.com/w/cpp/numeric/constants.html

**Comment by EZoni** [comment]
@lucafedeli88 

@WeiqunZhang's suggestion seems good to me, what do you think?

**Comment by lucafedeli88** [comment]
It's a very good idea ! Sorry for my late answer, I've been away for vacation and for a conference, but I'll resume working on this PR very soon.

**Comment by lucafedeli88** [comment]
@EZoni , I've just implemented @WeiqunZhang 's suggestion .

**Comment by lucafedeli88** [comment]
Ping, @EZoni ;-) !

File: Source/ablastr/constant.H
Comment (EZoni) [inline Source/ablastr/constant.H:1]: Thanks for the latest changes, @lucafedeli88!

I would suggest that we rename the following constants (`old`: `new` syntax):
- `q_e_v`: `qe_v`
- `m_e_v`: `me_v`
- `m_p_v`: `mp_v`
- `m_u_v`: `mu_v`
- `r_e_v`: `re_v`

and maybe also the ones with the type specified, i.e., the ones without the `_v` - though this may be more optional.

Personally I find the new names above a bit more human-friendly than the current names. What do you think? 

If you strongly prefer the current names, I would be fine with leaving things as they are, but I would like to know the opinion of other reviewers.

Comment (EZoni) [reply Source/ablastr/constant.H:1]: However, I do see that some of these may be problematic, e.g., `mu_v` gets easily confused with $\mu$ ... :unamused:

Comment (EZoni) [reply Source/ablastr/constant.H:1]: If we keep the `_` separator, which is probably the best thing to do to avoid confusion, then maybe we could use it also for the $\frac{1}{c}$ and $\frac{1}{c^2}$ constants, so `inv_c` instead of `invc` and `inv_c2` instead of `invc2` - I'd find these more readable as well.

Comment (dpgrote) [reply Source/ablastr/constant.H:1]: A suggestion I would make is to use the names from scipy.constants. It would be helpful is the equations were written using the same constant names, making it easier to convert the equations between the c++ and Python.

Comment (EZoni) [reply Source/ablastr/constant.H:1]: In that case we would have:
- `e` (neither `q_e` nor `qe`).
- `m_e` (not `me`).
- `m_p` (not `mp`).
- `u` (neither `m_u` nor `mu`).
- Not sure about the electron radius (I found only `scipy.constants.physical_constants['classical electron radius']`, but no shortcut). Maybe this one could stay `re` or `r_e_v`?

Comment (lucafedeli88) [reply Source/ablastr/constant.H:1]: I don't have any strong opinion about these names. However, since changing them would require changes everywhere in WarpX, I would suggest discussing the topic in the next developer's meeting.

Comment (EZoni) [reply Source/ablastr/constant.H:1]: Okay, I added the topic in next week's agenda.

**Review by aeriforme — APPROVED** [review]

File: Source/ablastr/constant.H [hunk 7]
```diff
@@ -34,58 +41,182 @@ namespace ablastr::constant
...
-        static constexpr auto m_p = 1.67262192595e-27_rt;
+        constexpr auto m_p = m_p_v<amrex::Real>;
+
         //! dalton: unified atomic mass unit [kg]
-        static constexpr auto m_u = 1.66053906892e-27_rt;
+        constexpr auto m_u = m_u_v<amrex::Real>;
 
         //! reduced Planck Constant = h / tau [J*s]
-        static constexpr auto hbar = 1.0545718176461565e-34_rt;
+        constexpr auto hbar = hbar_v<amrex::Real>;
```
aeriforme, line 170: Is it possible that `hbar` is not exposed to the input file?

Source/ablastr/constant.H
```diff
@@ -34,58 +41,182 @@ namespace ablastr::constant
...
+        //! xi: nonlinearity parameter of Heisenberg-Euler effective theory = (2.*alpha**2*ep0**2*hbar**3)/(45.*m_e**4*c**5)
+        constexpr double xi = 1.3050122383827227e-52;
+
         //! xi times c2 = xi*c*c. This should be usable for single precision instead of xi; very close to smallest float32 number possible (1.2e-38)
-        static constexpr auto xi_c2 = 1.1728865075613984e-35_rt;
+        constexpr auto xi_c2 = xi_c2_v<amrex::Real>;
+
+        //! Schwinger field [V/m] (variable template)
+        template<typename T>
+        constexpr auto schwinger_field = schwinger_field_v<amrex::Real>;
```
aeriforme, line 186: Should we specify that it's E field?

Source/ablastr/constant.H
```diff
@@ -34,58 +41,182 @@ namespace ablastr::constant
...
-        //! The value is calculated from the expression. This differs
-        //! slightly from the CODATA2022 value 0.0072973525643.
-        static constexpr auto alpha = 0.0072973525643330135_rt;
+        constexpr auto alpha = alpha_v<amrex::Real>;
+
         //! classical electron radius = 1./(4*pi*ep0) * q_e*q_e/(m_e*c*c) [m]
-        static constexpr auto r_e = 2.8179403205e-15_rt;
-        //! xi: nonlinearity parameter of Heisenberg-Euler effective theory = (2.*alpha*alpha*ep0*ep0*hbar*hbar*hbar)/(45.*m_e*m_e*m_e*m_e*c*c*c*c*c)
-        static constexpr double xi = 1.3050122383827227e-52;
+        constexpr auto r_e = r_e_v<amrex::Real>;
```
aeriforme, line 176: Should we add the Compton wavelength?

Source/ablastr/constant.H
EZoni: I agree. SciPy's name is `epsilon_0` - I would use the same.

Source/ablastr/constant.H
EZoni: I think this is related to what we call "hybrid QED Maxwell solver", which was introduced in https://github.com/BLAST-WarpX/warpx/pull/417, back in 2020.

Our inputs documentation for `warpx.use_hybrid_QED` and `warpx.quantum_xi` (under "Numerics and algorithms" / "Additional parameters") cites [(Grismayer, Thomas, et al. "Quantum electrodynamics vacuum polarization solver." New Journal of Physics 23.9 (2021): 095005)](https://iopscience.iop.org/article/10.1088/1367-2630/ac2004), where they call $\xi$ the "nonlinearity coupling parameter".

ax3l (approved):
LGTM, thank you all!

@lucafedeli88 please address the three doc string comments by @aeriforme and then feel free to self-merge.

EZoni:
Note that there is also https://github.com/BLAST-WarpX/warpx/pull/5828#discussion_r2087573096 that isn't just about docstrings. We already changed other names in this PR, e.g., `invc` and `invc2`, so we could change `ep0` too, as suggested by @aeriforme - to be discussed.

Source/ablastr/constant.H
lucafedeli88: Yes, it's not something that was introduced in this PR. However, we should probably add some more details in the documentation.

Source/ablastr/constant.H
lucafedeli88: Yes, it is not exposed at the moment. We probably should do that! It's a good idea for another PR, I think.

Source/ablastr/constant.H
lucafedeli88: Yes, that's a good remark. I've updated the docstring.

Source/ablastr/constant.H
lucafedeli88: I think that we can. If it is already used in WarpX or if you plan to use it we should! However, I think that it is better if it is done in a different PR for clarity.

lucafedeli88:
> Note that there is also [#5828 (comment)](https://github.com/BLAST-WarpX/warpx/pull/5828#discussion_r2087573096) that isn't just about docstrings. We already changed other names in this PR, e.g., `invc` and `invc2`, so we could change `ep0` too, as suggested by @aeriforme - to be discussed.

I'll open a poll

PR #5830, opened 2025-04-11
[WIP] Loop over particles rather than cells to compute density and temperature

atmyers:


RemiLehe:
Thanks for this PR!

I wonder if the changes made in https://github.com/BLAST-WarpX/warpx/pull/5763 also need a similar update.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -655,79 +692,163 @@ public:
...
+            amrex::ParticleReal* AMREX_RESTRICT vs1_in_each_cell = vs1_vec.dataPtr();
+            amrex::ParticleReal* AMREX_RESTRICT vs2_in_each_cell = vs2_vec.dataPtr();
 
-            // Loop over cells
             WARPX_PROFILE_VAR("BinaryCollision::doCollisionsWithinTile::findDensityTemperatures", prof_findDensityTemperatures);
-            amrex::ParallelForRNG( n_cells,
-                [=] AMREX_GPU_DEVICE (int i_cell, amrex::RandomEngine const& engine) noexcept
+            WARPX_PROFILE_VAR("BinaryCollision::doCollisionsWithinTile::findDensityTemperatures::atomics", prof_findDensityTemperatures1);
+        // Loop over particles and compute density and temperature
+            amrex::ParallelFor( std::max(np1, np2),
```
dpgrote, line 758: This is clever. If you want to maximize the parallelism, could this be `amrex::ParallelFor( np1 + np2,` and change the `if` statements below accordingly?

atmyers: That's a good idea - I'll try that.

atmyers:
The tests are currently failing, but I think this is just a consequence of the random number sequence being different. For example, the `test_3d_ionization_electron_dsmc` problem gives:

ionization_dsmc_density_Te

using the current code. But if someone more familiar with the physics could take a look and confirm, that would be great.

atmyers:
Closing this as it has been superseded by #5981

PR #5838, opened 2025-04-16
Enable Collocated grid Solve for Ohm's Law Hybrid solver

clarkse:
Adding second derivative nodal operators. Enabling Hybrid Solver to use collocated nodal Algorithm.


Source/FieldSolver/FiniteDifferenceSolver/FiniteDifferenceAlgorithms/CartesianNodalAlgorithm.H
```diff
@@ -137,6 +156,25 @@ struct CartesianNodalAlgorithm {
         // For CartesianNodalAlgorithm, UpwardDy and DownwardDy are equivalent
     }
 
+    /**
+     * Perform derivative along y on a nodal grid, from a cell-centered field `F`*/
```
roelof-groenewald: This comment is outdated, right?

Source/FieldSolver/FiniteDifferenceSolver/FiniteDifferenceAlgorithms/CartesianNodalAlgorithm.H
```diff
@@ -172,6 +210,26 @@ struct CartesianNodalAlgorithm {
         // For CartesianNodalAlgorithm, UpwardDz and DownwardDz are equivalent
     }
 
+
+    /**
+     * Perform second derivative along z on a cell-centered field `F`*/
```
roelof-groenewald: This one too?
```suggestion
     * Perform second derivative along z on a nodal grid, from a nodal field `F`*/
```

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -197,25 +197,25 @@ void HybridPICModel::InitData (const ablastr::fields::MultiFabRegister& fields)
...
+//         Bx_stag == IntVect(1,0) && By_stag == IntVect(0,0) && Bz_stag == IntVect(0,1) &&
+// #elif defined(WARPX_DIM_3D)
+//         Ex_stag == IntVect(0,1,1) && Ey_stag == IntVect(1,0,1) && Ez_stag == IntVect(1,1,0) &&
+//         Bx_stag == IntVect(1,0,0) && By_stag == IntVect(0,1,0) && Bz_stag == IntVect(0,0,1) &&
+// #endif
+//         Jx_stag == Ex_stag && Jy_stag == Ey_stag && Jz_stag == Ez_stag
+//     );
+//     WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+//         appropriate_grids,
+//         "Ohm's law E-solve only works with staggered (Yee) grids.");
```
roelof-groenewald: Seems like I could have saved myself a lot of trouble here before by just checking `if (WarpX::GetInstance().grid_type == GridType::Staggered)` hehe

roelof-groenewald (commented):
Wow! I'm actually also surprised by how few line changes this required. I guess if it wasn't for the hyper-resistivity term that needs the second derivatives the changes would have been really minimal.

One thing though, I do think you now need the `FillBoundary` calls from your other PR since $\nabla P_e$ will require a guard cell value for `rho`, to get the derivative at the tile edge, right?

clarkse:
> Wow! I'm actually also surprised by how few line changes this required. I guess if it wasn't for the hyper-resistivity term that needs the second derivatives the changes would have been really minimal.
> 
> One thing though, I do think you now need the `FillBoundary` calls from your other PR since ∇ P e will require a guard cell value for `rho`, to get the derivative at the tile edge, right?

Thats a good point. I tested this without electron pressure as a first go. I think we can also make it a touch faster by avoiding the interpolation calls for nodal grids.

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
```diff
@@ -171,31 +188,59 @@ ProjectionDivCleaner::solve ()
 
     for (int ilev = 0; ilev < m_levels; ++ilev)
     {
-        MLPoisson mlpoisson({geom[ilev]}, {ba[ilev]}, {dmap[ilev]}, info);
-
-        mlpoisson.setMaxOrder(m_linop_maxorder);
-        mlpoisson.setDomainBC(lobc, hibc);
+        if (m_grid_type == GridType::Collocated) {
```
roelof-groenewald: These parts can be simplified a bit by setting `linop` to an uninitialized pointer for a linear solver. That way only the initialization of the linear solver has to be in the `if` block and the rest of the code does not have to be duplicated.

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
```diff
@@ -288,34 +359,53 @@ ProjectionDivCleaner::correctBfield ()
...
-            },
-            [=] AMREX_GPU_DEVICE (int i, int j, int k)
-            {
-                By_arr(i,j,k) += CartesianYeeAlgorithm::DownwardDy(sol_arr, coefs_y, n_coefs_y, i, j, k);
-            },
-            [=] AMREX_GPU_DEVICE (int i, int j, int k)
-            {
-                Bz_arr(i,j,k) += CartesianYeeAlgorithm::DownwardDz(sol_arr, coefs_z, n_coefs_z, i, j, k);
-            });
+            if (m_grid_type == GridType::Staggered) {
```
roelof-groenewald: Perhaps this function can also be simpified by templating it on the finite difference algorithm?

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
```diff
@@ -288,34 +359,53 @@ ProjectionDivCleaner::correctBfield ()
...
+            }
 #endif
         }
         // Synchronize the ghost cells, do halo exchange
         ablastr::utils::communication::FillBoundary(*Bx,
                                                     Bx->nGrowVect(),
                                                     WarpX::do_single_precision_comms,
-                                                    geom[ilev].periodicity());
+                                                    geom[ilev].periodicity(),
+                                                    true);
```
roelof-groenewald, line 379: Do I understand correctly that this `true` is just making the `FillBoundary` call blocking (to synchronize threads)? If so it would make sense to only make the `Bz` call synchronize, no?

roelof-groenewald:
> I think we can also make it a touch faster by avoiding the interpolation calls for nodal grids.

I guess the best way would be to just return early in `Interp` if the target and source staggering matches. That way we can continue to easily support both staggered and colocated grids in the Ohm solver.

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
roelof-groenewald: If it helps, this can be used as an example of how the above is done elsewhere:
https://github.com/BLAST-WarpX/warpx/blob/480c53de3d95649a9774c76b764678b5ff735d61/Source/ablastr/fields/PoissonSolver.H#L309

clarkse: I tried this, but since the common inherited linear operator is the base MLLinOp class and the interface is different between cell centered and nodal operators, it would require reworking the class structure and virtual function structure of MLMG and the linear operator classes as a whole. I think having a few duplicated lines of code is OK here, even though it irks me a bit.

clarkse: OK, I ended up setting up a private templated member function to remove duplicated code.

Source/Initialization/DivCleaner/ProjectionDivCleaner.cpp
clarkse: the true is for nodal synchronization, it is still asynchronous.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/ExternalVectorPotential.cpp
```diff
@@ -210,6 +211,14 @@ ExternalVectorPotential::InitData ()
 
         amrex::Gpu::streamSynchronize();
 
+        if (m_do_clean_divA && warpx.grid_type == GridType::Collocated) {
```
roelof-groenewald: This is only called during the initialization, right? If so, why not also do the divergence cleaning when the grid type is staggered?
I guess in principle, this is only needed when A has time dependence. Could the check be for that instead?

roelof-groenewald (commented):
Looks good! I just had the one comment and it looks like there are some CI failures to address before we can merge.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/ExternalVectorPotential.cpp
clarkse: I only am doing it now when non-staggered because the staggered version is set up to clean on face-centered staggering and the A field is an edge-centered field. The solver for a staggered B field needs a cell-centered operator to work. In the collocated version I use a nodal solver, which I would need for the face-centered case.

I think this would need a bit more testing for the cleaning of A when staggered. I will modify it and give it a shot.

roelof-groenewald (approved):
Looks good!

clarkse:
Adding PR to AMReX to fix assertion error.

https://github.com/AMReX-Codes/amrex/pull/4488

PR #5839, opened 2025-04-16
Implement inverse bremsstrahlung

dpgrote:
This PR implements inverse bremsstrahlung, the absorption of photon during the collision of electrons and ions. This is done in a Monte Carlo scheme where photon particles are absorbed at a rate that depends on a background of electrons and ions.

The energy and momentum absorbed from the photons is distributed among the electrons so that the quantities are exactly conserved.

A CI test can be added once PR #5537 is merged, allowing a case with forward and inverse Bremsstrahlung.


JustinRayAngus:
FYI. This PR is high on my list. I've been waiting for 1) Getting the implicit solver fully working in axisymmetric geometries (PR #6738) and 2) Having the ability to load particle distribution from a grid file (PR #6716). Then I can setup and use the NIF implosion simulation to verify this works well for this important application.

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -2021,40 +2082,247 @@ WarpXParticleContainer::DepositTotalNGPTemperature (amrex::MultiFab* temperature
...
-        amrex::Array4<amrex::Real> const& sum_array = sum_mf.array(mfi);
-        amrex::Array4<amrex::Real> const& temp_array = temperature->array(mfi);
+        amrex::Array4<amrex::Real> const& N_array = particle_number.array(mfi);
+        amrex::Array4<amrex::Real> const& temp_array = temperature.array(mfi);
         amrex::ParallelFor(box,
             [=] AMREX_GPU_DEVICE (int i, int j, int k) {
-                if (sum_array(i,j,k,0) > 0) {
-                    const amrex::Real invsum = 1._rt/sum_array(i,j,k,0);
+                if (N_array(i,j,k) > 0) {
+                    const amrex::Real invsum = 1._rt/N_array(i,j,k);
```
JustinRayAngus: Based on our recent discussion, should we use the `?` formalism here?

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -2021,40 +2082,247 @@ WarpXParticleContainer::DepositTotalNGPTemperature (amrex::MultiFab* temperature
...
+#endif
+
+        amrex::Array4<amrex::Real> const& num_array = particle_number.array(mfi);
+        amrex::Array4<amrex::Real> const& temp_array = temperature.array(mfi);
+        amrex::Array4<amrex::Real> const& debye_array = debye_length->array(mfi);
+
+        amrex::ParallelFor(box,
+            [=] AMREX_GPU_DEVICE (int i, int j, int k) {
+
+                amrex::Real const N = num_array(i,j,k);  // number of particles
```
JustinRayAngus, line 2150: Return when N <= 0.?

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -2021,40 +2082,247 @@ WarpXParticleContainer::DepositTotalNGPTemperature (amrex::MultiFab* temperature
...
+                // Nanbu 1998: gab^2 = 3*Ta/ma + 3*Tb/mb + |Ua - Ub|^2
+                amrex::Real const g12sq = 3.0_rt*VTe*VTe + 3.0_rt*VTi*VTi
+                                          + std::pow(uxe_array(i,j,k) - uxi_array(i,j,k), 2._rt)
+                                          + std::pow(uye_array(i,j,k) - uyi_array(i,j,k), 2._rt)
+                                          + std::pow(uze_array(i,j,k) - uzi_array(i,j,k), 2._rt);
+                amrex::Real const g12sq_norm = g12sq/(PhysConst::c*PhysConst::c);
+                amrex::Real constexpr b0_factor = PhysConst::q_e/(PhysConst::c*PhysConst::c)/
+                                                  (2.0_rt*MathConst::pi*PhysConst::epsilon_0*PhysConst::m_e)*PhysConst::q_e; // [m]
+                amrex::Real const mu = PhysConst::m_e*rimass/(PhysConst::m_e + rimass);
+                amrex::Real const b0 = b0_factor*Zi/(mu*g12sq_norm + 2.0_rt*EF/(PhysConst::m_e*PhysConst::c*PhysConst::c)); // [m]
```
JustinRayAngus: `PhysConst::c2`. @EZoni recently went through and made this clean up in PR #6615

Source/Particles/Collision/InverseBremsstrahlung/InverseBremsstrahlung.cpp
```diff
@@ -0,0 +1,340 @@
...
+}
+
+void
+InverseBremsstrahlung::doCollisions (amrex::Real /*cur_time*/, amrex::Real dt, MultiParticleContainer* mypc)
+{
+    ABLASTR_PROFILE("InverseBremsstrahlung::doCollisions()");
+    using namespace amrex::literals;
+
+    auto& photons = mypc->GetParticleContainerFromName(m_species_names[0]);
+    auto& electrons = mypc->GetParticleContainerFromName(m_species_names[1]);
```
JustinRayAngus, line 48: Are there any checks that the specified species are photons and electrons?

Source/Particles/Collision/InverseBremsstrahlung/InverseBremsstrahlung.cpp
```diff
@@ -0,0 +1,340 @@
...
+        int const ri = i_cell;
+        // rr is radius at the cell center
+        amrex::ParticleReal const rr = rmin + (ri + 0.5_prt)*dr;
+        return 4.0_prt*static_cast<amrex::ParticleReal>(MathConst::pi)*rr*rr;
+#else
+        // No factor is needed for Cartesian
+        amrex::ignore_unused(i_cell);
+        return 1.0_prt;
+#endif
+    };
```
JustinRayAngus, line 160: Just a suggestion to double check this. This same function lives in BinaryCollision.H, and was previously used erroneously (see PR #6711).

There is also at least one line different here compared to what lives in BinaryCollision.H. Here you have:
`int const ri = i_cell % nr;`

In BinaryCollisions.H it reads:
`int const ri = (i_cell - i_cell%nz)/nz;`

These two may be equivalent, but I think it would be good to be consistent. At some point the volume_factor should be made a function that lives at a higher level and can be called from both places.

Source/Particles/Collision/InverseBremsstrahlung/InverseBremsstrahlung.cpp
```diff
@@ -0,0 +1,340 @@
...
+        // Distribute the remaining energy among the electrons without conserving momentum
+        // Note that this only works if the particle energy is nonzero
+        amrex::ParallelFor( np_electrons,
+            [=] AMREX_GPU_DEVICE (int ie) noexcept
+            {
+                const int i_cell = bins_electrons_ptr[ie];
+                amrex::ParticleReal const dKE = KE_in_each_cell[i_cell]/w_sum_electrons_in_each_cell[i_cell];
+                if (dKE == 0.) { return; }
+
+                amrex::ParticleReal constexpr c2 = PhysConst::c * PhysConst::c;
```
JustinRayAngus: See PR #6615

Source/Particles/Collision/InverseBremsstrahlung/InverseBremsstrahlung.H
```diff
@@ -0,0 +1,69 @@
...
+     * @param electrons the electrons particle container
+     *
+     */
+    void doInverseBremsstrahlungWithinTile (amrex::Real dt, int lev, amrex::MFIter const& mfi,
+                                            WarpXParticleContainer& photons,
+                                            WarpXParticleContainer& electrons) const;
+
+private:
+
+    amrex::ParticleReal m_energy_fraction = 0.05;
```
JustinRayAngus, line 65: Does it mention in the Docs that this parameter can be parsed from input?

JustinRayAngus (approved):
Thanks Dave. I left several minor comments.

Source/Particles/WarpXParticleContainer.cpp
dpgrote: I changed it to return when N is zero.

Source/Particles/WarpXParticleContainer.cpp
dpgrote: Yes. I was already working on cleaning up this section some.

Source/Particles/Collision/InverseBremsstrahlung/InverseBremsstrahlung.cpp
dpgrote: It checks that the first species is photons, but there's no check of the second.

JustinRayAngus: That is fine.

Source/Particles/Collision/InverseBremsstrahlung/InverseBremsstrahlung.cpp
dpgrote: That line `ri = i_cell % nr` was incorrect since it was using the wrong ordering of the cells. But that has now been changed in AMReX so now it is the correct expression and is consistent with what is now in BinaryCollision.H.

Source/Particles/Collision/InverseBremsstrahlung/InverseBremsstrahlung.H
dpgrote: Yes, is it mentioned in the docs, but I'm updating the text to be more clear. Would you prefer a different default value, something smaller? Presumably the value should be roughly 1/(number of particles per cell) so that the energy is distributed evenly? Should the electrons be shuffled here so that the electrons are not paired up always the same way?

JustinRayAngus: The default value is fine.

Note that the default use of this parameter is to modify the energy in the center-of-mass frame for randomly paired electrons. The intent is not to spread the energy evenly over all particles. Ideally it will finish before going through all pairs.

Yes, the electrons should be randomly paired for energy restoration. That is what I do in PICNIC.

If the center-of-mass energy-correction scheme fails, then we do the standard energy-correction that is not strictly momentum conserving. In PICNIC I additionally use the `energy_fraction` parameter to prevent the energy correction for a given electron being too small:
`Real dgamma = std::max(m_energy_fraction*gp2m1,dgamma0*1.001);`

Here, `dgamma` is the change in gamma for a specific electron with `gp2m1 = gamma -1`. `dgamma0` is the common value that would be used if the correction is applied to all electrons. I forget exactly what the factor of `1.001` is for, but I think it was to ensure all the energy gets absorbed in 1 pass over all the electrons. It's basically needed because of round-off-error.

dpgrote: @JustinRayAngus What I implemented for when the scheme fails is to distribute the energy evenly among all of the electrons, which in your case would be setting `dgamma = dgamma0`. This has the advantage of being a loop over particles instead of grid cells, which is substantially faster on a GPU.

Your method would typically give more energy to electrons with higher energy. The same could be done by weighting the calculation of `dgamma0` by the energy. This would keep the loop over particles.

PR #5840, opened 2025-04-17
CI: Remove Obsolete `geometry.is_periodic` from Input File

EZoni:
@oshapoval pointed out that we have one input file where `geometry.is_periodic` is set, even though we comment in the source code that this parameter is used only internally and should not be set by the user:
https://github.com/BLAST-WarpX/warpx/blob/2e895d5bcbceacd64617660c48aae12351d88536/Source/Utils/WarpXUtil.cpp#L386-L394

This PR removes the parameter from the input file, to avoid confusion for developers and users who look at examples.


oshapoval:
Thanks for this PR @EZoni.  Looks good to me.

RemiLehe (approved):

PR #5846, opened 2025-04-23
Exposing mf register

RevathiJambunathan:
This PR exposes the multifab register so that when we create a local copy using pyamrex interface in a python script, 
1. the multifab gets allocated and initialized
2. this allows the multifab to also be redistributed during load balancing
3. when finalize is called, the multifab memory is cleared in a clean way

This PR https://github.com/AMReX-Codes/pyamrex/pull/439) adds wrappers to ba.convert, ba.surroundingNodes, mf.copy 

so now , we could add a new "test_mf" with desired nodality
```
        reg.alloc_init("test_nf",0,warpx.boxArray(0).convert(amrex.IntVect(1,1)),warpx.DistributionMap(0),
                       1,amrex.IntVect(1,1),0.0,True,True)
```

Screenshot 2025-05-14 at 10 58 19 AM


Source/Python/MultiFabRegister.cpp
```diff
@@ -56,6 +57,7 @@ void init_MultiFabRegister (py::module & m)
                  bool,
                  bool
              >(&MultiFabRegister::alloc_init<std::string>),
+             py::return_value_policy::reference_internal,
              py::arg("name"),
              py::arg("dir"),
              py::arg("level"),
```
RemiLehe, line 63: We should probably also add `return_value_policy::reference_internal` for the other methods that return a `MultiFab`, including e.g. `alias_init`

Examples/Physics_applications/spacecraft_charging/inputs_test_rz_spacecraft_charging_picmi.py
```diff
@@ -308,4 +322,5 @@ def compute_actual_charge_on_spacecraft():
 installafterInitEsolve(spc.save_normalized_vacuum_Efields)
 installafterEsolve(spc.correct_space_charge_fields)
 
-sim.step(max_steps)
+sim.step(2)
+# sim.step(max_steps)
```
RemiLehe:
```suggestion
sim.step(max_steps)
```

Examples/Physics_applications/spacecraft_charging/inputs_test_rz_spacecraft_charging_picmi.py
```diff
@@ -71,12 +71,25 @@ def save_normalized_vacuum_Efields(
         self,
     ):
         warpx = sim.extension.warpx
-
+        amrex = sim.externsion.amrex
```
EZoni: CI failure caused by this typo:
```suggestion
        amrex = sim.extension.amrex
```

Examples/Physics_applications/spacecraft_charging/inputs_test_rz_spacecraft_charging_picmi.py
```diff
@@ -70,19 +70,59 @@ def correct_space_charge_fields(self, q=None):
...
+        reg = warpx.multifab_register()
+
+        self.phi = warpx.multifab("phi_fp", 0)
+        self.Er = warpx.multifab("Efield_fp", dir=self.dir_r, level=0)
+        self.Ez = warpx.multifab("Efield_fp", dir=self.dir_z, level=0)
+        # allocate and initialize normalized fields
+        self.normalized_Er = reg.alloc_init(
+            "normalized_Er",
+            0,
+            warpx.boxArray(0).convert(amrex.IntVect(0, 1)),
```
RemiLehe, line 89: Could we do `Er.box_array()`, or something similar, here?

RemiLehe (approved):

PR #5848, opened 2025-04-24
Check that species involved in linear Breit-Wheeler are photons

RemiLehe:
When using the linear Breit-Wheeler package, we check that the product species are electron and positrons, but we did not yet check that the reactant are photons.

This PR fixes the issue.


Source/Particles/Collision/BinaryCollision/BinaryCollisionUtils.cpp
```diff
@@ -11,6 +11,7 @@
 #include "Particles/WarpXParticleContainer.H"
 
 #include <AMReX_ParmParse.H>
+#include <AMReX_Parser.H>
```
RemiLehe:
```suggestion
```

aeriforme (approved):

PR #5849, opened 2025-04-24
Reduce duplication in linear Breit-Wheeler by reusing code from fusion, DSMC

RemiLehe:
We can reduce duplication in linear Breit-Wheeler module (collision between two photons, producing an electron-positron pair) by reusing the function `get_collision_parameters`, which is already used in the fusion and DSMC modules.

In order to do so, I had to slightly rewrite the function `get_collision_type` so that it is also valid for massless particles. In particular, this involved passing the momentum `p` to the function, rather than the normalized momentum `u`.

I also added some details about the derivation of some of the equations in the code.

### DIFF
--- a/Source/Particles/Collision/BinaryCollision/BinaryCollisionUtils.H
+++ b/Source/Particles/Collision/BinaryCollision/BinaryCollisionUtils.H
@@ -65,7 +65,6 @@ namespace BinaryCollisionUtils{
 
         constexpr double c = PhysConst::c;
         constexpr double c_sq = PhysConst::c * PhysConst::c;
-        constexpr double inv_csq = 1.0 / c_sq;
 
         // Cast input parameters to double before computing collision properties
         // This is needed to avoid errors when using single-precision particles
--- a/Source/Particles/Collision/BinaryCollision/BinaryCollisionUtils.H
+++ b/Source/Particles/Collision/BinaryCollision/BinaryCollisionUtils.H
@@ -113,7 +113,7 @@ namespace BinaryCollisionUtils{
         // p^2 = E^2/4c^2 - (m_1^2 + m_2^2)c^2/2 + (m_1^2-m_2^2)^2 c^6/4E^2
         double p_star_sq;
         if (m1_dbl + m2_dbl == 0) {
-            p_star_sq = 0.5 * E_star * inv_c;
+            p_star_sq = powi<2>( 0.5 * E_star * inv_c );
         } else {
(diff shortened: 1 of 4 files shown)


Source/Particles/Collision/BinaryCollision/BinaryCollisionUtils.H
```diff
@@ -118,20 +108,20 @@ namespace BinaryCollisionUtils{
...
         // relative velocity in the center of mass frame, cast back to chosen precision
-        v_rel_COM = static_cast<amrex::ParticleReal>(std::sqrt(p_star_sq) * (1.0/(m1_dbl*g1_star) + 1.0/(m2_dbl*g2_star)));
+        v_rel_COM = PhysConst::c * static_cast<amrex::ParticleReal>(std::sqrt(p_star_sq) * c * (1.0/E1_star + 1.0/E2_star));
 
         // Cross sections and relative velocity are computed in the center of mass frame.
         // On the other hand, the particle densities (weight over volume) in the lab frame are used.
-        // To take this disrepancy into account, it is needed to multiply the
+        // To take this discrepancy into account, it is needed to multiply the
         // collision probability by the ratio between the Lorentz factors in the
         // COM frame and the Lorentz factors in the lab frame (see
```
RemiLehe: This comment should be updated to avoid mentioning Lorentz factors.

Source/Particles/Collision/BinaryCollision/LinearBreitWheeler/SingleLinearBreitWheelerCollisionEvent.H
```diff
@@ -66,71 +66,26 @@ void SingleLinearBreitWheelerCollisionEvent (const amrex::ParticleReal& u1x, con
                                        const amrex::ParticleReal& probability_target_value,
                                        const amrex::RandomEngine& engine)
 {
-    // General notations in this function:
-    //     x_sq denotes the square of x
-    //     x_star denotes the value of x in the center of mass frame
+    amrex::ParticleReal E_coll, v_coll, lab_to_COM_factor;
+
+    constexpr amrex::ParticleReal me = PhysConst::m_e;
```
RemiLehe, line 73:
```suggestion
    // By convention in WarpX, the momentum of photons is normalized by m_e
    // i.e. `p = m_e u`.
    constexpr amrex::ParticleReal me = PhysConst::m_e;
```

aeriforme (approved):

PR #5853, opened 2025-04-25
Unify user-input names between nuclear fusion and linear Breit-Wheeler

aeriforme:
This PR removes the tag/word `fusion` from three user inputs: 
* `fusion_multiplier` becomes `event_multiplier`
* `fusion_probability_threshold` becomes `probability_threshold`
* `fusion_probability_target_value` becomes `probability_target_value`

This makes the user interfaces of the nuclear fusion and linear Breit-Wheeler modules match.


RemiLehe (changes requested):
Thanks for this PR!
Could you also update [this function](https://github.com/BLAST-WarpX/warpx/blob/development/Source/WarpX.cpp#L1820) so that people that used the previous syntax receive a clear error message, telling them how to update to the new syntax.

RemiLehe (approved):

PR #5863, opened 2025-05-02
Fix bug with double counting of guard cell values in effective potential ES solver

### LINKED ISSUE #5855: Domain decomposition artifact shows up with `EffectivePotential` ES solver
## Description
We're seeing artifacts of the domain decomposition when using the semi-implicit Poisson solver.
See 
- https://github.com/BLAST-WarpX/warpx/blob/development/Source/FieldSolver/ElectrostaticSolvers/EffectivePotentialES.cpp
- https://github.com/BLAST-WarpX/warpx/blob/development/Source/ablastr/fields/EffectivePotentialPoissonSolver.H

The image below shows charge density for two simulations where the domain consisted of just one tile versus an identical simulation where the domain was decomposed into 4 tiles:

Image

The same issue does **not** come occur with the explicit ES solver. 

The above case uses an embedded boundary (in a 3d simulation), so both the explicit solver and the semi-implicit solver uses `amrex::MLEBNodeFDLaplacian`. But only the semi-implicit case uses a spatially varying `sigma` (see https://github.com/BLAST-WarpX/warpx/blob/480c53de3d95649a9774c76b764678b5ff735d61/Source/FieldSolver/ElectrostaticSolvers/EffectivePotentialES.cpp#L93).

It would appear that somewhere we are missing a call to fill the guard cells. @WeiqunZhang do you have any idea where this problem might be coming from?

roelof-groenewald:
As described in the title this fixes a bug where we were adding guard cell values to valid cells twice in the collection of the spatially dependent prefactor in the Poisson equation.

Fixes #5855.


roelof-groenewald:
I checked the CI test case and the plasma diffusion results are still very close to the theoretical expectation (dashed grey lines):
image

I'll therefore just reset the benchmark values.

WeiqunZhang (approved):

PR #5868, opened 2025-05-07
DSMC ion impact ionization: ionization products are created with collinear velocities

RemiLehe:
# Overview

This changes the kinematics of collisional ionization (i.e. how to choose the velocities of the product particles after a collisional ionization event: the ionized particle and the resulting electron).

The kinematic model now depends on the nature of the incident particle (i.e. the first reactant / non-target species). In practice, this is decided based on its mass: if the incident species is heavier than twice the electron mass, it is treated as **ion impact ionization**; otherwise as **electron impact ionization**.

## Ion impact ionization (new, collinear model)

For ion impact ionization, we use a new model:
- Just like [Perez et al.](https://pubs.aip.org/aip/pop/article/19/8/083104/108747/Improved-modeling-of-relativistic-collisions-and) as well as [Stephen Diorio's thesis](https://deepblue.lib.umich.edu/items/c799120e-7d66-4e75-8564-d30bd76c2fa6), we neglect angular scattering, and assume that the velocities of the product species are aligned with that of the incident particle, in the center-of-mass frame. (Many thanks to @peterparsons2 for pointing out these references. For completeness, @JustinRayAngus also flagged these papers as relevant: [Koura, Phys. Fluids 29, 3509–3511 (1986)](https://pubs.aip.org/aip/pfl/article/29/11/3509/804023/Null-collision-technique-in-the-direct-simulation), [Vahedi, CPC, 1995](https://www.sciencedirect.com/science/article/pii/001046559400171W))
- Unlike [Perez et al.](https://pubs.aip.org/aip/pop/article/19/8/083104/108747/Improved-modeling-of-relativistic-collisions-and) as well as [Stephen Diorio's thesis](https://deepblue.lib.umich.edu/items/c799120e-7d66-4e75-8564-d30bd76c2fa6), we ensure that every single collision conserves energy and momentum exactly. This is done by using energy-momentum conservation, along with assumption that the resulting particles have the same velocity vector.

## Electron impact ionization (original model)

For electron impact ionization, we retain the original model, in which the momenta of the three products are distributed by sampling a random point on an ellipse (so that the net momentum can be made zero while conserving energy).

# Possible follow-up PRs

- [ ] Unify ion and electron impact ionization, as well as the MCC and DSMC models: remove angular scattering and make sure that the distribution of energy for the incident and resulting electrons is such that the Turner benchmark passes.
- [ ] For ion impact ionization: check that we recover the results of the Bethe formula
- [ ] Reproduce the tests of [Perez et al](https://pubs.aip.org/aip/pop/article/19/8/083104/108747/Improved-modeling-of-relativistic-collisions-and)

In discussions with @roelof-groenewald, it also seemed that a reasonable assumption would be that the velocity of the atom to be ionized does not get modified, so that conservation of momentum and energy would determine the properties of the ionizing particle and the ionized electron.


roelof-groenewald:
We discussed using the following distribution for the outgoing angle:

image

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -456,93 +456,38 @@ public:
...
+                // have equal velocity and that their velocity vector is aligned with that of the target particle
+
+                // With these assumptions, conservation of momentum in the center of mass frame allows to express the momenta
+                // of the products, as a function of the outcoming momentum of the non-target particle:
+                // p_slot2 = - p_non_target_out * mass_slot2 / m1
+                // p_slot3 = - p_non_target_out * mass_slot3 / m1
+                // (with m1 = mass_slot2 + mass_slot3, by conservation of mass)
+
+                // Amplitude of the momentum of the non-target particle, obtained from conservation of energy
+                const amrex::ParticleReal p_non_target_out = std::sqrt( 2.0_prt * E_out / (1.0_prt / m0 + 1.0_prt / m1 ) );
```
roelof-groenewald: I wonder if here and on line 484 if it wouldn't be better to do:

```suggestion
                const amrex::ParticleReal p_non_target_out = std::sqrt( 2.0_prt * E_out / (1.0_prt / m0 + 1.0_prt / (mass_slot2 + mass_slot3) ) );
```

It is true that technically `m1 == mass_slot2 + mass_slot3` should be true, but that is assigning burden to the user which might not be obeyed. For example in my simulations I usually set the mass of hydrogen ions equal to the mass of hydrogen atoms.

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -456,93 +456,38 @@ public:
...
+
+                // Reduce the velocity of the non-target particle, to reflect the loss of energy due to ionization
+                const amrex::ParticleReal reduce_factor = p_non_target_out / p_non_target_in;
+                ux0 *= reduce_factor;
+                uy0 *= reduce_factor;
+                uz0 *= reduce_factor;
+                // Obtain the other velocity vectors by conservation of momentum
+                const amrex::ParticleReal mass_ratio = m0 / m1;
```
```suggestion
                const amrex::ParticleReal mass_ratio = m0 / (mass_slot2 + mass_slot3);
```

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -456,93 +456,143 @@ public:
...
+                    const amrex::ParticleReal E1 = (
+                        0.5_prt * m0 * (ux0*ux0 + uy0*uy0 + uz0*uz0) / PhysConst::q_e
+                    );
+                    const amrex::ParticleReal E2 = (
+                        0.5_prt * m1 * (ux3*ux3 + uy3*uy3 + uz3*uz3) / PhysConst::q_e
+                    );
+                    const amrex::ParticleReal E_coll = E1 + E2;
+
+                    // subtract the energy cost for ionization
+                    const amrex::ParticleReal E_out = (E_coll - reaction_energy) * PhysConst::q_e;
```
roelof-groenewald: Your way of calculating the remaining energy is mathematically equivalent to this, so the `E_out` calculation can be moved out of the `if` statement (and I think using your way is cleaner than this way).

roelof-groenewald: I will make that change in a follow-up PR so that we don't hold this one up any longer.

roelof-groenewald (approved):

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -456,93 +456,139 @@ public:
...
+                    // have equal velocity and that their velocity vector is aligned with that of the target particle
+
+                    // With these assumptions, conservation of momentum in the center of mass frame allows to express the momenta
+                    // of the products, as a function of the outcoming momentum of the non-target particle:
+                    // p2 = - p_non_target_out * mass_slot2 / m1
+                    // p3 = - p_non_target_out * mass_slot3 / m1
+                    // (with m1 = mass_slot2 + mass_slot3, by conservation of mass)
+
+                    // Amplitude of the momentum of the non-target particle, obtained from conservation of energy
+                    const amrex::ParticleReal p_non_target_out = std::sqrt( 2.0_prt * E_out / (1.0_prt / m0 + 1.0_prt / (mass_slot2 + mass_slot3) ) );
```
RemiLehe, line 486: The change from ` 1.0_prt / m1` to `1.0_prt / (mass_slot2 + mass_slot3)` caused modifications in the checksum, because `m1` is indeed different from `mass_slot2 + mass_slot3` for the `test_3d_ionization_ion_dsmc`:
```
ions.mass = m_p
electrons.mass = m_e
neutrals.mass = m_p
```

PR #5869, opened 2025-05-07
PML_RZ.cpp: don't use WarpX::GetInstance()

lucafedeli88:
This PR removes the occurrences of `WarpX::GetInstance()` from `PML_RZ.cpp`. This is achieved by passing `m_fields` and the spectral solver object as function parameters. 

The PR also moves the code of the function `PML_RZ::PushPMLPSATDSinglePatchRZ`  inside `PML_RZ::PushPSATD` in order to avoid an unnecessary intermediate layer (without `WarpX::GetInstance()`, `PML_RZ::PushPSATD` would have directly called `PML_RZ::PushPMLPSATDSinglePatchRZ`).


dpgrote (approved):
Looks good to me! Thanks!

PR #5871, opened 2025-05-07
Pass Efield as optional argument to `computePhi`

roelof-groenewald:
This PR slightly cleans up the code but mostly sets the stage for an upcoming PR (#5872) that will enable calculation of the electrostatic potential during diagnostic output for non-ES solvers.


RemiLehe (changes requested):
Thanks for this PR!
Could you update the docstrings of the various versions of the `computePhi` function, so that the docstrings mention that new argument, and the fact that it is used only with EB.

RemiLehe (approved):

PR #5872, opened 2025-05-07
Allow `phi` output with non-ES solvers

roelof-groenewald:
This PR adds `PhiFunctor` to the diagnostic suite. In the case that a non-electrostatic solver is used, the `PhiFunctor` will perform a Poisson solve using $\rho = \epsilon_0\nabla\cdot E$ as the charge density.
This allows the electrostatic potential to be output as part of the diagnostics, even for electromagnetic solvers.

Note that I opted for $\epsilon_0\nabla\cdot E$ as the charge density rather than depositing the charge from simulation particles since the Ohm's law model does not have an electron species, and therefore $\epsilon_0\nabla\cdot E$ is the only way to get $\rho$.


Source/FieldSolver/ElectrostaticSolvers/RelativisticExplicitES.cpp
```diff
@@ -25,9 +25,10 @@
     }
     prepare_field_solve |= m_poisson_boundary_handler->m_boundary_potential_specified;
 
-    if (prepare_field_solve) {
-        m_poisson_boundary_handler->DefinePhiBCs(warpx.Geom(0));
-    }
+    // if (prepare_field_solve) {
+    //     m_poisson_boundary_handler->DefinePhiBCs(warpx.Geom(0));
+    // }
```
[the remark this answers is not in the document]
roelof-groenewald: I'm not sure about a good way around this. Currently, `DefinePhiBCs` is not called with EM solvers unless a boundary potential is set. But we need it initialized if `phi` is in the diagnostic list - should we add a check of the requested diagnostics here?

Regression/Checksum/benchmarks_json/test_2d_ohm_solver_magnetic_reconnection_picmi.json
```diff
@@ -6,6 +6,7 @@
     "Ex": 55751949.54321176,
     "Ey": 74312405.49508089,
     "Ez": 310308772.614497
```
roelof-groenewald:
```suggestion
    "Ez": 310308772.614497,
```

Regression/Checksum/benchmarks_json/test_2d_ohm_solver_magnetic_reconnection_picmi.json
```diff
@@ -5,7 +5,8 @@
     "Bz": 7.428600985388779,
     "Ex": 55751949.54321176,
     "Ey": 74312405.49508089,
-    "Ez": 310308772.614497
+    "Ez": 310308772.614497,
+    "phi": 13485153.403965198
```
roelof-groenewald:
```suggestion
    "phi": 13485129.404149422
```

Regression/Checksum/benchmarks_json/test_2d_ohm_solver_magnetic_reconnection_picmi.json
```diff
@@ -6,6 +6,7 @@
     "Ex": 55760103.16960256,
     "Ey": 74340311.8150126,
     "Ez": 307873828.1999537
```
roelof-groenewald:
```suggestion
    "Ez": 307873828.1999537,
```

Source/Diagnostics/ComputeDiagFunctors/PhiFunctor.cpp
```diff
@@ -0,0 +1,80 @@
...
+        // define solver BCs in case they have not been defined yet
+        es_solver.m_poisson_boundary_handler->DefinePhiBCs(warpx.Geom(0));
+        const std::array<amrex::Real, 3> beta = {0.0};
+        es_solver.computePhi(
+            amrex::GetVecOfPtrs(rho_vec), amrex::GetVecOfPtrs(phi_vec),
+            beta, es_solver.self_fields_required_precision,
+            es_solver.self_fields_absolute_tolerance,
+            es_solver.self_fields_max_iters, es_solver.self_fields_verbosity,
+            false
+        );
```
EZoni: I would write this as 
```cpp
        es_solver.computePhi(
            amrex::GetVecOfPtrs(rho_vec), amrex::GetVecOfPtrs(phi_vec),
            beta, es_solver.self_fields_required_precision,
            es_solver.self_fields_absolute_tolerance,
            es_solver.self_fields_max_iters, es_solver.self_fields_verbosity,
            /*is_igf_2d=*/false
        );
```
or at least with an inline comment
```cpp
        es_solver.computePhi(
            amrex::GetVecOfPtrs(rho_vec), amrex::GetVecOfPtrs(phi_vec),
            beta, es_solver.self_fields_required_precision,
            es_solver.self_fields_absolute_tolerance,
            es_solver.self_fields_max_iters, es_solver.self_fields_verbosity,
            false  // is_igf_2d
        );
```
but up to you, I'm fine if you prefer to keep it as is.

Source/Diagnostics/ComputeDiagFunctors/PhiFunctor.cpp
```diff
@@ -0,0 +1,80 @@
...
+        // allocate fields for charge and potential
+        amrex::Vector<std::unique_ptr<amrex::MultiFab>> rho_vec(num_levels);
+        amrex::Vector<std::unique_ptr<amrex::MultiFab>> phi_vec(num_levels);
+
+        // calculate the total charge density as the divergence of the E-field
+        amrex::BoxArray ba = warpx.boxArray(m_lev);
+        ba.surroundingNodes();
+        rho_vec[0] = std::make_unique<amrex::MultiFab>(ba, warpx.DistributionMap(m_lev), 1, 1);
+        rho_vec[0]->setVal(0.);
+        warpx.ComputeDivE(*rho_vec[0], m_lev);
```
EZoni, line 54: This `ComputeDivE` works only for the FDTD solver, correct? Do we need to add a warning in case the user requests this diagnostic while using the PSATD solver?

Source/Diagnostics/ComputeDiagFunctors/PhiFunctor.cpp
```diff
@@ -0,0 +1,80 @@
...
+    {
+        const int num_levels = 1;
+        // allocate fields for charge and potential
+        amrex::Vector<std::unique_ptr<amrex::MultiFab>> rho_vec(num_levels);
+        amrex::Vector<std::unique_ptr<amrex::MultiFab>> phi_vec(num_levels);
+
+        // calculate the total charge density as the divergence of the E-field
+        amrex::BoxArray ba = warpx.boxArray(m_lev);
+        ba.surroundingNodes();
+        rho_vec[0] = std::make_unique<amrex::MultiFab>(ba, warpx.DistributionMap(m_lev), 1, 1);
```
EZoni: Similarly to the comment about the `computePhi` call, it may be worth using temporary integers with variable names that explain what they are.

In this case, for example, even an IDE (or at least my IDE) doesn't help too well, because of the `std::make_unique` wrapping. 

I think the first `1` would be the `ncomp` argument and the second `1` would be the `ngrow` argument in this case.

So maybe either this
```cpp
        int const ncomp = 1;
        int const ngrow = 1;
        rho_vec[0] = std::make_unique<amrex::MultiFab>(ba, warpx.DistributionMap(m_lev), ncomp, ngrow);
```
or this
```cpp
        rho_vec[0] = std::make_unique<amrex::MultiFab>(ba, warpx.DistributionMap(m_lev), /*ncomp=*/1, /*ngrow=*/1);
```

What do you think?

EZoni (commented):
@roelof-groenewald 

Thanks and sorry for the delay in reviewing this. It looks good to me, I left only a few minor comments. 

I'll ask @dpgrote to take a look as well, just to see if there's anything I overlooked. I will follow up to avoid stalling.

Source/Diagnostics/ComputeDiagFunctors/PhiFunctor.cpp
roelof-groenewald: It should also work for PSATD as long as WarpX was compiled with FFT support. The `ComputeDivE` function handles the different solver cases:
```
void
WarpX::ComputeDivE(amrex::MultiFab& divE, const int lev)
{
    if ( WarpX::electromagnetic_solver_id == ElectromagneticSolverAlgo::PSATD ) {
#ifdef WARPX_USE_FFT
        const ablastr::fields::VectorField Efield_aux_lev = m_fields.get_alldirs(FieldType::Efield_aux, lev);
        spectral_solver_fp[lev]->ComputeSpectralDivE(lev, Efield_aux_lev, divE);
#else
        WARPX_ABORT_WITH_MESSAGE(
            "ComputeDivE: PSATD requested but not compiled");
#endif
    } else {
        const ablastr::fields::VectorField Efield_aux_lev = m_fields.get_alldirs(FieldType::Efield_aux, lev);
        m_fdtd_solver_fp[lev]->ComputeDivE(Efield_aux_lev, divE);
    }
}
```

Source/Diagnostics/ComputeDiagFunctors/PhiFunctor.cpp
roelof-groenewald: Good point! I like the second approach.

EZoni:
I think the Read the Docs CI build fails because it lacks the fix in #6151. I cannot force-merge PRs, so I'm afraid you would need to merge `development` into this branch if I will be tasked with merging the PR eventually.

EZoni (approved):

EZoni (approved):

PR #5873, opened 2025-05-08
Fix DSMC typo for non-product producing processes

RemiLehe:
The DSMC kernel performs several types of collisions:
- non-product producing collisions: elastic scattering, charge exchange, etc.
- product producing collisions: ionization

The first types of collisions (non-product producing collisions) still creates new macroparticles (because the high-weight macroparticle in a collision can be split into a lower weight "scattered" particles, and another "unscattered" particle).

However, these non-product producing collisions only involve the first 2 "products".

### DIFF
--- a/Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
+++ b/Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
@@ -88,7 +88,7 @@ public:
         );
 
         amrex::Vector<int> num_added_vec(m_num_product_species, 0);
-        for (int i = 0; i < m_num_product_species; i++)
+        for (int i = 0; i < 2; i++)
         {
             // Record the number of non product producing events lead to new
             // particles for species1 and 2. Only 1 particle is created for


Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -88,7 +88,7 @@ public:
         );
 
         amrex::Vector<int> num_added_vec(m_num_product_species, 0);
-        for (int i = 0; i < m_num_product_species; i++)
+        for (int i = 0; i < 2; i++)
```
RemiLehe, line 91: @roelof-groenewald Does DSMC support passing only one species in `<collision_name>.species` (for intra-species scattering)? If yes, this should be generalized...

roelof-groenewald: I don't believe that has been tried (at least not by me). I guess generally the plasma simulations would you Coulomb collisions for intra-species scattering, but I could imagine a case where neutral-neutral elastic scattering might be included to capture thermalization.
I guess we can deal with that in a follow-up PR.

roelof-groenewald (approved):
Thanks for fixing!

PR #5876, opened 2025-05-08
Adding python bindings and a helper setter/getter object in python to…

clarkse:
… expose the density floor and substep parameters that can be adjusted at runtime.

### DIFF
--- a/Python/pywarpx/HybridPICModel.py
+++ b/Python/pywarpx/HybridPICModel.py
@@ -7,20 +7,6 @@
 #
 # License: BSD-3-Clause-LBNL
 
-
-try:
-    import cupy as cp
-except ImportError:
-    cp = None
-
-try:
-    from mpi4py import MPI as mpi
-
-    comm_world = mpi.COMM_WORLD
-    npes = comm_world.Get_size()
-except ImportError:
-    npes = 1
-
 from ._libwarpx import libwarpx
 from .Bucket import Bucket
 
--- a/Python/pywarpx/HybridPICModel.py
+++ b/Python/pywarpx/HybridPICModel.py
@@ -1,12 +1,62 @@
-# Copyright 2023 The WarpX Community
+# Copyright 2023-2025 The WarpX Community
 #
 # This file is part of WarpX.
 #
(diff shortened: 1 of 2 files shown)


roelof-groenewald (commented):
The `Python/WarpX.cpp` changes look good to me. I just have a question on how you envision the `HybridPICModel` class to be used? The usual approach with these extensions is that there would be a simulation object, `sim = picmi.Simulation()`, and then the hooks would be called with
```
sim.extension.warpx.set_hybrid_pic_substeps(substeps)
```
With the current implementation, are you envisioning creating another object, `HybridPICModel()`, to call the hooks with? What is the benefit of having this second object?

clarkse:
> The 'Python/WarpX.cpp`changes look good to me. I just have a question on how you envision the`HybridPICModel`class to be used? The usual approach with these extensions is that there would be a simulation object,`sim = picmi.Simulation()`, and then the hooks would be called with
> 
> ```
> sim.extension.warpx.set_hybrid_pic_substeps(substeps)
> ```
> 
> With the current implementation, are you envisioning creating another object, `HybridPICModel()`, to call the hooks with? What is the benefit of having this second object?

I just added the object in picmi as a helper setter/getter object to make the callbacks easier syntax wise. It is just a wrapper object like the MF wrapper that simplfies calling a bit to not have to remember odd syntax. With the setters and getters in a callback all one needs to do is instantiate the HybridPicModel object and it is mirroring aspects of the data structure in WarpX. 

For example in a callback or between steps in a PICMI script one can do:

from pywarpx.HybridPICModel import HybridPicModel

hpm = HybridPICModel()
hpm.substeps = 1000  

I just though it was Pythonic and handy. Not necessary though, just two ways to modify the values. The direct call to the warpx object works too.

roelof-groenewald:
> I just though it was Pythonic and handy. Not necessary though, just two ways to modify the values. The direct call to the warpx object works too.

Thanks for clarifying. Sounds good! I guess we could start to do for other things too - for example adding a similar class to `Boundary.py` that would have hooks to change the boundary or EB potentials, but that is a discussion for a different time.

roelof-groenewald (approved):

PR #5883, opened 2025-05-12
Update reference to new paper

RemiLehe:


EZoni:
Could we wait for #5889 to be merged and then rebase here, so that we can test if the CI checks are effectively skipped, allowing all TC members to merge PRs like this one (which isn't the case now)?

EZoni:
@WeiqunZhang 

Trying #5889 here and it seems like it still doesn't work for our `clang-tidy` checks, probably because they are added automatically through a matrix rather than manually as separate jobs in the workflow file. In fact we see `clang-tidy / clang-tidy-${{ matrix.dim }}D` as skipped check in the list...

I think I had resolved this problem with my alternative implementation in #5469, I will try to see if I find another workaround for the implementation that we just merged.

EZoni (approved):

PR #5886, opened 2025-05-12
Scale the charge/current in RZ geometry when folding high-radius guard cells into domain

RemiLehe:
This PR scales the current density J and the charge density rho, in RZ geometry, when wrapping the fields in the high-radius guard cells back into the valid domain.

More specifically, WarpX treats the charge/current in the following way:
- Deposit charge/current
- Convert to charge/current density by dividing by the cell volume
- Wrap charge/current density from guard cells outside the domain back into the valid domain
In order to preserve the total charge/current associated with any given particle, the ratio of cell volumes (between the guard cells and the valid cell into which it is wrapped) needs to be applied.

As part of the same PR, we also remove a scaling of E and B in the guard cell. This scaling has no strong theoretical justification, and it has been observed that energy conservation is better without this scaling.

This isolates some of the changes in https://github.com/BLAST-WarpX/warpx/pull/5189/ that are not specific to 1D cylindrical or spherical.


dpgrote (approved):
Looks good. Thanks!

PR #5888, opened 2025-05-12
CI: Remove Hard-Coded Path to Checksum Files

EZoni:
The hard-coded path to the checksum files in the default regression analysis script,
```py
sys.path.insert(1, "../../../../warpx/Regression/Checksum/")
```
may cause issues when running tests locally through `ctest`, due to the `warpx` directory name.

To-do:
- [x] Test locally
  - [x] by @EZoni 
  - [x] by @dpgrote (who reported the issue with the `warpx` name first) 
- [x] Test with CI
- [x] Test with IDE
  - [x] by @RemiLehe

### DIFF
--- a/Examples/Tests/laser_injection_from_file/inputs_test_2d_laser_injection_from_binary_file_prepare.py
+++ b/Examples/Tests/laser_injection_from_file/inputs_test_2d_laser_injection_from_binary_file_prepare.py
@@ -12,7 +12,6 @@
 # injection of a laser pulse from an external binary file:
 # - Generate an input binary file with a gaussian laser pulse.
 
-import sys
 
 import matplotlib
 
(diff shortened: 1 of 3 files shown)


EZoni:
@dpgrote 

This PR may fix the issue you had locally - please let me know if ctest runs successfully (after a new build from scratch).

EZoni:
@dpgrote 

Would you have a chance to check if this fixes the issue you had reported during the last hackathon?

dpgrote:
> @dpgrote
> 
> Would you have a chance to check if this fixes the issue you had reported during the last hackathon?

Yes, this does fix the problem. It can now find the benchmark files.

dpgrote (approved):
This fixes the issues I was having. Thanks!

PR #5889, opened 2025-05-13
CI: Fix Skipped Required Workflows

EZoni:
This implement AMReX's solution to handle skipped but required CI workflows properly, as in https://github.com/AMReX-Codes/amrex/pull/4197.

This is necessary due to the limitation described in https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks.

PR #5469 implements an alternative solution based on a in-house script, but I figured it may be easier to implement the same solution as in AMReX for maintenance purposes.

The new preliminary jobs are named `Analyze / Check changes`. For example, they show up as `CUDA / Analyze / Check changes`, `macOS / Analyze / Check changes`, etc.

This will allow all TC members to merge PRs where required workflows are skipped (e.g., PRs that change only documentation files), instead of relying on the WarpX admin team only.


.github/workflows/check_changes.yml
```diff
@@ -0,0 +1,31 @@
...
+      - uses: dorny/paths-filter@v3
+        id: changes
+        with:
+          filters: |
+            docs:
+              - 'Docs/**'
+              - '**.rst'
+            others:
+              - '!Docs/**'
+              - '!**.rst'
```
EZoni: The `**.rst` filter is not present in the AMReX implementation, hopefully this is the right syntax to handle multiple filters. Not entirely intuitive why one needs to specify also the `others` filters...

.github/workflows/clang_sanitizers.yml
```diff
@@ -79,7 +81,7 @@ jobs:
   build_thread_sanitizer:
     name: Clang thread sanitizer
     runs-on: ubuntu-24.04
-    if: github.event.pull_request.draft == false
+    if: ${{ github.event.pull_request.draft == false && needs.check_changes.outputs.has_non_docs_changes == 'true' }}
```
EZoni, line 86:
```suggestion
    needs: check_changes
    if: ${{ github.event.pull_request.draft == false && needs.check_changes.outputs.has_non_docs_changes == 'true' }}
```

.github/workflows/check_changes.yml
WeiqunZhang: I think the filter rules are combined by `or` by default. So `has_non_docs_changes` means no changes in either `Docs/` or `*.rst`. The logic for `has_non_docs_changes` seems incorrect. Suppose I *only* change GOVERANCE.rst. Because there are no changes to `Docs/`, `has_non_docs_changes` will be set to true.

WeiqunZhang: I think we should do this.
```
diff --git a/.github/workflows/check_changes.yml b/.github/workflows/check_changes.yml
index f526f162d..af2c2e62b 100644
--- a/.github/workflows/check_changes.yml
+++ b/.github/workflows/check_changes.yml
@@ -20,13 +20,10 @@ jobs:
         id: changes
         with:
           filters: |
-            docs:
-              - 'Docs/**'
-              - '**.rst'
-            others:
+            non_docs:
               - '!Docs/**'
               - '!**.rst'
+          predicate-quantifier: 'every'
       - id: set-output
         run: |
-          echo "has_docs_changes=${{ steps.changes.outputs.docs }}" >> $GITHUB_OUTPUT
-          echo "has_non_docs_changes=${{ steps.changes.outputs.others }}" >> $GITHUB_OUTPUT
+          echo "has_non_docs_changes=${{ steps.changes.outputs.non_docs }}" >> $GITHUB_OUTPUT
```

WeiqunZhang: We should also remove this because we don't need it.
```
      has_docs_changes:
        value: ${{ jobs.check.outputs.has_docs_changes }}
```

WeiqunZhang:
https://github.com/AMReX-Codes/amrex/pull/4474

I tested that amrex PR on my own fork. It works as expected.

EZoni:
Thanks, @WeiqunZhang! I updated the PR, feel free to approve if everything looks good to you.

WeiqunZhang (approved):

PR #5890, opened 2025-05-14
[WIP] Treat charge exchange as a product-producing process

RemiLehe:
The aim is to allow charge exchange between different species, e.g.:
$H^{+} + H_2 \rightarrow H + H_2^{+}$

In the above case, $H^+$ and $H_2$ are the **colliding** species, $H$ and $H_2^+$ are the **product** species.

TODO:
- [ ] Add documentation
- [ ] When charge exchange is active: prevent any other scattering processes (to avoid issues/complications with allocation)
- [ ] Check that charge exchange within the same species (e.g. $H^{+} + H \rightarrow H + H^{+}$) still works? (This is apparently not yet tested.)


Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.cpp
```diff
@@ -17,50 +17,62 @@ SplitAndScatterFunc::SplitAndScatterFunc (const std::string& collision_name,
...
-        // DSMC process with products)
-        amrex::Vector<std::string> product_species;
-        pp_collision_name.queryarr("product_species", product_species);
+        // Check if the scattering processes include reactions that produce macroparticles in new species
+        // (i.e. not in the incident species list), i.e. if it contains ionization or charge exchange
+        amrex::Vector<std::string> scattering_processes;
+        pp_collision_name.queryarr("scattering_processes", scattering_processes);
+        const bool reaction_produces_new_species = std::any_of(scattering_processes.begin(), scattering_processes.end(), [](const std::string& process) {
+            return process == "ionization" || process == "charge_exchange";
+        });
```
RemiLehe, line 26: We should think about which processes should be in the constructor of DSMCFunc vs SplitAndScatterFunc

RemiLehe:
Superseded by https://github.com/BLAST-WarpX/warpx/pull/5908

PR #5893, opened 2025-05-15
PML_RZ: don't use warpx static variables directly

lucafedeli88:
This PR avoids accessing directly `WarpX::do_single_precision_comms` inside `PML_RZ.cpp` . Instead, this variable is passes as an argument.


RemiLehe (approved):
Thanks for this PR!

PR #5895, opened 2025-05-16
Remove superfluous if condition in impact ionization

RemiLehe:
This `if` condition is already enclosed in an existing `if` condition:
https://github.com/BLAST-WarpX/warpx/blob/development/Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H#L322

### DIFF
--- a/Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
+++ b/Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
@@ -392,99 +392,94 @@ public:
                 uy_p2 -= uCOM_y;
                 uz_p2 -= uCOM_z;
 
-                if (mask[i] == int(ScatteringProcessType::IONIZATION)) {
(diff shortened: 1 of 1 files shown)


roelof-groenewald (commented):
I put the double `if` statements with the idea that this would make it clear where other product producing processes could be added (like inter-species charge exchange, excitation, etc.). 
You are planning on adding inter-species charge exchange soon, right? Do you think you will keep this approach to implementing the different product producing process or do you have something else in mind?

RemiLehe:
Yes, good point.
However, for inter-species charge exchange, we actually did not reuse that part of the code:
https://github.com/BLAST-WarpX/warpx/pull/5890/files#diff-030f5798f49da4a99874f43a9ea695c84e3817501eae939861b46aea30bdf69cR313
It makes me think that the code included in **outer** `if` condition for ionization is also very specific to ionization.

roelof-groenewald (approved):

PR #5896, opened 2025-05-16
[WIP] Impose that the target species in DSMC impact ionization is always the second species

RemiLehe:
The code in `SplitAndScatterFunc.cpp` is somewhat confusing because:
- This line seems to indicate that we do not know which of the two colliding species is the target species: https://github.com/BLAST-WarpX/warpx/blob/development/Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.cpp#L41
- While this line seems to indicate that we do know which one it is:
https://github.com/BLAST-WarpX/warpx/blob/development/Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.cpp#L50

In reality, these lines are consistent, because the first part refers to the list of products as entered by the user, while the second part refers to the order in the WarpX internal list of products, which has been modified by this line:
https://github.com/BLAST-WarpX/warpx/blob/2e242f8aa2c39914df9c06420b70f862992953c7/Source/Particles/Collision/BinaryCollision/BinaryCollision.H#L121
However, it is difficult to know this from just reading the code.

Instead I would suggest that we always impose that the second species that the user passes is the target species (the one that gets ionized). In this case, the user also does not need to pass `ionization_target_species` as a separate argument.

TODO: 
- [ ] Add checks that the ionized species is consistent with the products
- [ ] Impose that the user enters the electron species as the first species + check the species type in the code.
- [ ] Consolidate checks in only one file

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -2080,6 +2080,7 @@ Details about the collision models can be found in the :ref:`theory section <mul
 * ``<collision_name>.species`` (`strings`)
     If using ``dsmc``, ``pairwisecoulomb`` or ``nuclearfusion``, this should be the name(s) of the species,
     between which the collision will be considered. (Provide only one name for intra-species collisions.)
+    When using ``dsmc`` with impact ionization, the second species is the one that gets ionized.
     If using ``background_mcc`` or ``background_stopping`` type this should be the name of the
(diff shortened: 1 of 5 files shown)


RemiLehe:
This PR is rendered largely obsolete by #5910.

PR #5897, opened 2025-05-16
Move WarpX::ErrorEst from Utils/WarpXTagging.cpp to WarpX.cpp

lucafedeli88:
With this PR, I would like to propose to move the `WarpX::ErrorEst` method from `Utils/WarpXTagging.cpp` to `WarpX.cpp`, and to eliminate the `Utils/WarpXTagging.cpp` file.

The rationale is that `Utils` is currently a very miscellaneous folder, containing both auxiliary functions and implementations of essential core methods (as in this case). I think that we should gradually move most of the content of the `Utils` folder somewhere else (e.g., the more general utility functions into `ablastr`, `SpeciesUtils` into `Particles`...).


EZoni (approved):
Looks good to me.

PR #5905, opened 2025-05-19
remove erroneous use of is_tangent_to_bndy in SetRhoOrJfieldFromPEC()

JustinRayAngus:
SetRhoOrJfieldFromPEC(), which is applied for both PEC and PMC boundaries, does two things. First, the rho/J deposited to the ghost region is reflected back into the valid cells with the appropriate sign. Second, the guard cells are filled with values consistent with the BCs same as the electric and magnetic fields. How to do the second step depends solely on the is_tangent_to_bndy flag. This flag is correct for PEC, but it is incorrect for PMC. The correct thing to use is the psign flag.

This PR uses the psign flag for the second step instead of the is_tangent_to_bndy flag, making this flag obsolete.


Source/BoundaryConditions/WarpX_PEC.cpp
```diff
@@ -394,52 +394,26 @@ namespace
...
-#endif
-                    }
-#endif
-                    if (tangent_to_bndy[idim]) {
-                        field(iv_mirror, n) = -rscale*field(ijk_vec, n);
-                    } else {
-                        field(iv_mirror, n) = rscale*field(ijk_vec, n);
-                    }
+                    // 2) Updated guard cells with the appropriate image J/rho
+                    field(iv_mirror,n) = inv_rscale * psign[idim][iside] * field(ijk_vec,n);
```
dpgrote: What happens in corners? Would part 1 need to happen from both sides of the corner before part 2 is done?

JustinRayAngus: Good catch. I will restore original part here.

dpgrote (approved):

JustinRayAngus:
@RemiLehe You assigned yourself to this PR. Anything holding up a merge? I would recommend this PR gets merged before PR #5909

PR #5906, opened 2025-05-20
Bugfix: Hybrid-pic restart causing inconsistencies

roelof-groenewald:
I have observed that restarting with the hybrid-PIC solver is not smooth (leads to subtle jumps in various quantities). This PR ensures a smooth evolution with regards to the field-solver (having consistent values for the previous timesteps' fields after a restart).


roelof-groenewald:
The changes made in this PR have the following consequences beyond restarts:
1. The initial current deposition is now (appropriately) done at t = -1/2 dt
2. The initial charge and current deposition is filtered if `use_filter = true`

The above consequences necessitates an update of the CI test checksum values.

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -60,35 +60,9 @@ void WarpX::HybridPICEvolveFields ()
...
-    // SyncCurrent does not include a call to FillBoundary, but it is needed
-    // for the hybrid-PIC solver since current values are interpolated to
-    // a nodal grid
-    for (int lev = 0; lev <= finest_level; ++lev) {
-        for (int idim = 0; idim < 3; ++idim) {
-            m_fields.get(FieldType::current_fp, Direction{idim}, lev)->FillBoundary(Geom(lev).periodicity());
-        }
-    }
+    // Perform charge deposition in component 0 of rho_fp at t_{n+1} and
+    // current deposition at t_{n+1/2}.
```
```suggestion
    // Perform charge deposition at t_{n+1} and current deposition at t_{n+1/2}.
```

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -257,34 +231,74 @@ void WarpX::HybridPICEvolveFields ()
...
+    if (restart_chkfile.empty()) {
+        // This is not a restart, so the rho_fp and current_fp multifabs are
+        // still empty.
+        HybridPICDepositRhoAndJ();
+    }
 
+    // Copy the rho_fp values to rho_fp_temp and the current_fp values to
+    // current_fp_temp, since the "temp" multifabs are meant to store the
+    // particle and current densities from the previous step during the field
+    // solve routine and are therefore needed when the first field solve is
```
```suggestion
    // solve routine and are needed when the first field solve is
```

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -257,34 +230,74 @@ void WarpX::HybridPICEvolveFields ()
     }
 }
 
+void WarpX::HybridPICDepositRhoAndJ ()
```
clarkse, line 233: Should this be in the WarpX class? I think this could be abstracted a bit to be a separate class that is friend's with the WarpX class to modularize it a bit. Not a deal breaker though since the main loops in the Field solver are in WarpX currently. It would be a bit of a refactor and could take place in another PR, however this could be a time to discuss architecture a bit.

clarkse (changes requested):
I think the changes look good. I am glad that the checkpoint restarting is being fixed. I do think we should have a CI test that exercises this code branch though for better code coverage.

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
roelof-groenewald: That sounds good as a larger, follow-up refactor to the overall structure.

RemiLehe (approved):

Issue #5901, opened 2025-05-19
exponential growth of electric field energy in CI test


JustinRayAngus:
I ran the 2d_particle_thermal_boundary CI test for 15 times longer and I observed an exponential increase in the electric field energy. See the image below.

The input deck is here:
Examples/Tests/particle_thermal_boundary/inputs_test_2d_particle_thermal_boundary

This test is a thermal e-C12 plasma with a temperature of 2 keV in a 2D box. The field BCs are PML at all boundaries. The particle BCs are thermal at all boundaries.

![Image](https://github.com/user-attachments/assets/480ba099-28f4-4a08-b710-e7ca5864c80b)

The fact that the magnetic field energy does not grow suggests that this problem may be related to lack of charge conservation. Below is a summary of the charge conservation from this test at the last time step. Charge is not conserved near the boundaires.

![Image](https://github.com/user-attachments/assets/166fe612-1263-4f4d-877c-47c1d0e607be)

WarpX uses filtering by default. If filtering is turned off, the system goes unstable faster and the charge density at an early time looks as follows

![Image](https://github.com/user-attachments/assets/d4706351-8517-4f17-874c-019f38a2ee0c)

These simulations use 4 boxes on 4 processors. There is a diagnostic issue at the box corners that touch the boundaries. Ignoring that, there seems to still be a lack of charge conservation in the first cell near the boundary.

JustinRayAngus:
I've looked into how rho/J are set at the boundaries and I found several bugs. Below is a comparison of the electric field energy for this problem obtained from the development branch and using PR #5909 that address the boundary bugs. The electric field energy still blows up eventually, but the growth is slower with the bugs fixed. As @RemiLehe mentioned during the weekly meeting, there still seems to be an issue with charges near PML boundaries.

![Image](https://github.com/user-attachments/assets/1ca7405b-7804-4f10-b132-d3a300be048b)

The boundary bugs in the development branch are that rho/J deposited into the guard region is not appropriately reflected back into the domain at domain corners that touch multiple PMC/PEC boundaries.

With these fixes, the charge density (with filtering on) looks as follows:

![Image](https://github.com/user-attachments/assets/740b3162-87fc-449e-a3b3-e6f54351bd5f)

[...]

PR #5909, opened 2025-05-22
refactored how to reflect/set Rho and J at PEC/PMC boundaries.

JustinRayAngus:
There is a bug in the routines that reflect the charge and current density (Rho/J) deposited to the guard regions back into the valid domain at conducting (PEC) and symmetry (PMC) boundaries. Domain corners that touch multiple PMC/PEC boundaries are not handled properly. This results in a lack of charge and energy conservation for simulations with such boundaries in 2 and 3 dimensions.

Note that Rho/J are commonly treated using the PMC boundary even when the fields are treated using a different boundary conditions, such as PML.

The fix is to extend the box where the reflection is done to include the guard cells transverse to the boundary. This requires somewhat of a large refactoring of the PEC::ApplyReflectiveBoundarytoRhofield() and PEC:ApplyReflectiveBoundarytoJfield() routines. However, these refactored routine also fix an additional bug that has to do with the charge density diagnostic at all box corners touching a PMC/PEC boundary not being set correctly. Additionally, the physical corners of Rho and J are also set now, while they are not set using the original routines.

The test problem for this bug is a thermal uniform e-C12 plasma in a 2D box with 4 PMC boundaries. The exactly energy conserving implicit EM method is used to rigorously test charge and energy conservation.

Charge conservation results using the development branch for this test are shown below. The lack of charge conservation is isolated to the corner regions of the domain.

developmentBranch_PMC_BCs

Results obtained using this PR, which preserves charge everywhere, are shown below

reflectBranch_PMC_BCs

Energy is also not conserved for this problem with the development branch. Energy is conserved to machine precision with this PR. See below.

netEnergy_develop_vs_reflect

This PR also improves the long time behavior of the 2d_particle_thermal_boundary CI test #5901 .

Remaining tasks to do:
- [x] Add CI test for this problem.
- [x] Verify that it works in 3D
- [x] Verify that it works on GPU
- [x] Do timing comparisons on GPU

Timing comparisons on GPU:
The new routine potentially has more kernel launches that could increase the cost of WarpX::ApplyJfieldBoundary() on GPUs. The original method has 3 kernel launches (1 for each component of J) independent of the number of boundaries that are PEC/PMC (assuming at least one boundary is). The new routine does 2 kernel launches (1 for reflect and 1 for set) for each direction that has a PEC or PMC boundary.

Timing comparisons from a worst-case scenario test for a 3D problem with all 8 boundaries set to PMC are given below. The simulation is a uniform thermal e-C12 plasma. The simulations are performed using 1 gpu on Perlmutter.
•	Grid: 16 x 16 x 16 = 4096
•	Particles per cell per species: 4 x 4 x 4 = 64
•	Time steps: 400

Exclusive per step timings for WarpX::ApplyJfieldBoundary() (results averaged over 5 identical tests):
•	Development branch with 1 box: 3.6220e-05
•	Development branch with 8 boxes: 7.596e-05 
•	This PR with 1 box: 5.039e-05
•	This PR with 8 boxes: 1.3547e-04

Summary: Per step cost of WarpX::ApplyJfieldBoundary() is between 1.5X and 2X more expensive using this PR for this worse-case scenario problem. However, the cost for this routine is only about 0.5% to 1.0% of the total simulation cost. The increase in cost is most likely associated with the increase in the number of kernel launches. However, this is somewhat of an unfair comparison because the original routine has several bugs associated with it while the new routine does not.


Examples/Tests/ohm_solver_em_modes/analysis_rz.py
```diff
@@ -179,5 +179,5 @@ def process(it):
     amps = np.abs(F_kw[2, 1, len(kz) // 2 - 2 : len(kz) // 2 + 2])
     print("Amplitude sample: ", repr(amps))
     assert np.allclose(
-        amps, np.array([80.99103401, 27.98511857, 215.63733823, 24.81072749])
+        amps, np.array([80.63370256, 28.62192717, 214.81372568, 22.33889693])
```
JustinRayAngus: @roelof-groenewald 

This is the only CI test where an analysis failed. The changes to the checksum values for this test all have relative error of order 1.0e-6 compared to the previous values. 

What exactly is the np.allclose() test checking? Do the new values look reasonable? Could you perhaps run this CI test using this PR and verify the results are as expected?

roelof-groenewald: > This is the only CI test where an analysis failed. The changes to the checksum values for this test all have relative error of order 1.0e-6 compared to the previous values.
> 
> What exactly is the np.allclose() test checking? Do the new values look reasonable? Could you perhaps run this CI test using this PR and verify the results are as expected?

Hi @JustinRayAngus. That part of the test is checking whether the amplitudes for specific normal modes match what was calculated when the test was created. The exact amplitudes are very sensitive to simulation evolution since the normal modes are excited through PIC noise. I therefore wouldn't be worried about the magnitude of these changes. I'd be happy to run the full test again to confirm that the correct physics result is still obtained, but I won't be able to do that until late June. Don't worry about waiting for that rerun before merging this PR.

dpgrote: Hi @roelof-groenewald I've had to reset these numbers several times in recent PRs due to various small changes. Since these numbers depend on the numerics including round off, does this check provide any additional validation of the simulation beyond what is done by the CI benchmark values? If not, could this check be removed since it requires an extra step to update it in addition to updating the benchmark values?

Source/BoundaryConditions/WarpX_PEC.cpp
```diff
@@ -341,111 +341,148 @@ namespace
...
-                    amrex::Real rscale = 1._rt;
+            if (!is_reflective[iside]) { continue; }
+
+            // Get the mirror guard cell index
+            amrex::IntVect ijk_mirror = ijk_vec;
+            ijk_mirror[idim] = mirrorfac[iside] - ijk_vec[idim];
+
+            // Update the cell if the mirror guard cell exists
+            if (ijk_vec == ijk_mirror && psign[iside] == -1) {
+                field(ijk_vec,n) = 0._rt;
```
dpgrote: Perhaps add a comment here that when odd, the value on the boundary is set to zero since it is subtracted from itself.

Source/BoundaryConditions/WarpX_PEC.cpp
```diff
@@ -341,111 +341,148 @@ namespace
...
+
+            if (!is_reflective[iside]) { continue; }
+
+            // Get the mirror guard cell index
+            amrex::IntVect ijk_mirror = ijk_vec;
+            ijk_mirror[idim] = mirrorfac[iside] - ijk_vec[idim];
+
+            // Update the cell if the mirror guard cell exists
+            if (ijk_vec == ijk_mirror && psign[iside] == -1) {
+                field(ijk_mirror,n) = 0.0;
```
dpgrote, line 462: Is this clause needed since it is also done in `ReflectJorRho`?

JustinRayAngus: I think so. While currently it is only called after ReflectJorRho(), it could be called independently. So I think it's a good idea to not assume anything here.

Source/BoundaryConditions/WarpX_PEC.cpp
```diff
@@ -686,82 +729,147 @@ PEC::ApplyReflectiveBoundarytoRhofield (
...
+            if ( !is_reflective[idim][0] && !is_reflective[idim][1] ) { continue; }
+
+            if ( (node_box.smallEnd()[idim] != domain_lo[idim]) &&
+                 (node_box.bigEnd()[idim] != domain_hi[idim]) ) { continue; }
+
+            // Get Rho box and grow to include guard cells in directions transverse
+            // to this boundary. This is required to correctly reflect Rho at domain
+            // corners that touch multiple PEC/PMC boundaries.
+            amrex::Box rho_box = amrex::convert(mfi.validbox(),rho_nodal);
+            for (int jdim = 0; jdim < AMREX_SPACEDIM; ++jdim) {
```
dpgrote, line 796: I think a bit of work could be saved here by changing this to
`for (int jdim = idim+1; jdim < AMREX_SPACEDIM; ++jdim) {`
And the `if (jdim==idim) { continue; }` would not be needed.

Source/BoundaryConditions/WarpX_PEC.cpp
```diff
@@ -686,82 +729,147 @@ PEC::ApplyReflectiveBoundarytoRhofield (
...
+        for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
+
+            if ( !is_reflective[idim][0] && !is_reflective[idim][1] ) { continue; }
+
+            if ( (node_box.smallEnd()[idim] != domain_lo[idim]) &&
+                 (node_box.bigEnd()[idim] != domain_hi[idim]) ) { continue; }
+
+            // Get Rho box and grow to include guard cells in transverse dirs
+            amrex::Box rho_box = amrex::convert(mfi.validbox(),rho_nodal);
+            for (int jdim = 0; jdim < AMREX_SPACEDIM; ++jdim) {
```
dpgrote, line 828: And similarly here, this could be `for (int jdim = 0; jdim < idim; ++jdim) {`.

Source/BoundaryConditions/WarpX_PEC.cpp
```diff
@@ -774,187 +882,201 @@ PEC::ApplyReflectiveBoundarytoJfield(
...
+            if ( (node_box.smallEnd()[idim] != domain_lo[idim]) &&
+                 (node_box.bigEnd()[idim] != domain_hi[idim]) ) { continue; }
+
+            // Get J boxes and grow to include guard cells in directions transverse
+            // to this boundary. This is required to correctly reflect J at domain
+            // corners that touch multiple PEC/PMC boundaries.
+            amrex::Box Jx_box = amrex::convert(mfi.validbox(),Jx_nodal);
+            amrex::Box Jy_box = amrex::convert(mfi.validbox(),Jy_nodal);
+            amrex::Box Jz_box = amrex::convert(mfi.validbox(),Jz_nodal);
+            for (int jdim = 0; jdim < AMREX_SPACEDIM; ++jdim) {
```
dpgrote, line 981: The same change could be done here as for `rho`.

Source/BoundaryConditions/WarpX_PEC.cpp
```diff
@@ -774,187 +882,201 @@ PEC::ApplyReflectiveBoundarytoJfield(
...
+            if ( (node_box.smallEnd()[idim] != domain_lo[idim]) &&
+                 (node_box.bigEnd()[idim] != domain_hi[idim]) ) { continue; }
 
-        // Get the multifab box including ghost cells
-        Box const& fabbox = mfi.fabbox();
+            // Get J boxes and grow to include guard cells in transverse dirs
+            amrex::Box Jx_box = amrex::convert(mfi.validbox(),Jx_nodal);
+            amrex::Box Jy_box = amrex::convert(mfi.validbox(),Jy_nodal);
+            amrex::Box Jz_box = amrex::convert(mfi.validbox(),Jz_nodal);
+            for (int jdim = 0; jdim < AMREX_SPACEDIM; ++jdim) {
```
dpgrote, line 1035: And here also.

Source/BoundaryConditions/WarpX_PEC.cpp
JustinRayAngus: I'm not sure I follow this change suggestion. For D = 2 and idim = 1, the suggested for loop would start at jdim = 2 and would not proceed since 2 = AMREX_SPACEDIM. How would this grow the box in the idim=0 direction?

Source/BoundaryConditions/WarpX_PEC.cpp
JustinRayAngus: Same confusion here. For 2D and idim = 0, this for loop looks to me like it would not do anything.

Source/BoundaryConditions/WarpX_PEC.cpp
dpgrote: Is there a case where one of these routines would be called without the other?

Source/BoundaryConditions/WarpX_PEC.cpp
dpgrote: When idim == 1, I don't think the box needs to grow in dim 0, since that would update values in the guard cells that would then be overwritten in the call to SetJorRho. This is assuming that both routines would always be called together.

Source/BoundaryConditions/WarpX_PEC.cpp
dpgrote: Here, when idim == 0, there's no need to do the copy from the dim = 1 guard cell since it hasn't been set yet, since that happens when idim == 1.

JustinRayAngus: I understand this now. However, it seems to assume that all the boundaries are set in this routine. That is not the case. For example, what if the system has PMC or PEC boundaries in X, but is periodic in Y and Z? In that case, you would want to grow in Y and Z here before setting in X.

Source/BoundaryConditions/WarpX_PEC.cpp
JustinRayAngus: I do understand this logic, but there is an issue with doing it this way. Let's discuss offline.

Source/BoundaryConditions/WarpX_PEC.cpp
JustinRayAngus: If it wasn't for the scaling with r for axisymmetric geometries (which will need to be removed/relocated in a future PR if filtering is to be used in such geometries), then this routine is just a generic method for setting guard cells for even or odd functions. In principle, it could be applied to any field that has those symmetries. I can't foresee all possible scenarios where one might want to call this function, but it seems like a source for future confusion or a future bug to make the behavior of this routine dependent on ReflectJorRho being called first.

Source/BoundaryConditions/WarpX_PEC.cpp
JustinRayAngus: The issue is that the suggested change assumes that all BCs are filled in this routine, which is not generally true. BCs for other directions can be set before or after the call here.

JustinRayAngus:
@EZoni The CI test failing here doesn't seem to be related to the PR. How do you re-run failed tests?

EZoni:
@JustinRayAngus 

If you click on the three dots `...` next to the failing check in the summary of CI checks at the bottom of the PR page, it brings you to the following page:
Screenshot from 2025-06-13 14-19-46

There you can click "Re-run failed jobs" and re-run only the checks that failed.

If you don't see the "Re-run jobs" selector, it may mean that you don't have permissions to do that, which I think we could change with a decision in one of the committees.

My first guess would be that you do have the right access to see that selector.

RemiLehe:
Thanks for this PR!

It seems that one of the CI tests is failing because of an unused Parameter:
```
END REGION WarpX::Evolve()
Unused ParmParse Variables:
  [TOP]::implicit_evolve.use_mass_matrices(nvals = 1)  :: [true]

  what():  ERROR: unused ParmParse variables.
SIGABRT
```
Would you be able to fix this?

JustinRayAngus:
@EZoni I'm experience the same test failing again that seemingly is disconnected from this PR. test_rz_langmuir_multi_psatd_JRhom_LL4 fails with an erroneous arithmetic operation at STEP 1. When that happened previously, I reran the RZ CI tests and it passed. I've re-ran it once already this time and it still fails. I've ran this tests locally and I don't see any issues. I've tried running with ctests, but it seems that there is something messed up with that for me because every tests that I run for all geometries fail.

JustinRayAngus:
> @EZoni I'm experience the same test failing again that seemingly is disconnected from this PR. test_rz_langmuir_multi_psatd_JRhom_LL4 fails with an erroneous arithmetic operation at STEP 1. When that happened previously, I reran the RZ CI tests and it passed. I've re-ran it once already this time and it still fails. I've ran this tests locally and I don't see any issues. I've tried running with ctests, but it seems that there is something messed up with that for me because every tests that I run for all geometries fail.

@EZoni @RemiLehe @ax3l @dpgrote @WeiqunZhang @atmyers I can confirm that test_rz_langmuir_multi_psatd_JRhom_LL4 passes when running ctest locally. It may require someone with more knowledge than me to figure out why it sometimes fails on this PR with an `Erroneous arithmetic operation` error at STEP 1.

EZoni:
@JustinRayAngus 

Do the backtraces in the "Logs" step (after "Test") of the CI pipeline tell us anything about where the erroneous arithmetic operation is occurring? I think we compile with flags good enough to produce meaningful backtraces, so hopefully there should be some information there.

EZoni:
@JustinRayAngus 

Also note that this is not specific to your PR, see, e.g., https://dev.azure.com/BLAST-WarpX/WarpX/_build/results?buildId=1212&view=results.

EZoni:
From our CDash dashboard, it seems like it started failing occasionally sometime towards the end of July:

Screenshot from 2025-07-11 11-29-55

JustinRayAngus:
> @JustinRayAngus
> 
> Do the backtraces in the "Logs" step (after "Test") of the CI pipeline tell us anything about where the erroneous arithmetic operation is occurring? I think we compile with flags good enough to produce meaningful backtraces, so hopefully there should be some information there.

Where do I find the Backtrace file?

JustinRayAngus:
> From our CDash dashboard, it seems like it started failing occasionally sometime towards the end of July:
> 
> Screenshot from 2025-07-11 11-29-55

Ah. Good to know.

WeiqunZhang:
I think test_rz_langmuir_multi_psatd_JRhom_LL4 is fixed in #6017.

Source/BoundaryConditions/WarpX_PEC.cpp
```diff
@@ -763,187 +879,201 @@ PEC::ApplyReflectiveBoundarytoJfield(
...
+
     }
 
-    // Each current component is handled separately below, starting with Jx.
-    grown_domain_box.convert(Jx_nodal);
 #ifdef AMREX_USE_OMP
 #pragma omp parallel if (amrex::Gpu::notInLaunchRegion())
 #endif
-    for (amrex::MFIter mfi(*Jx); mfi.isValid(); ++mfi) {
+    for (amrex::MFIter mfi(*Jx, false); mfi.isValid(); ++mfi) {
```
RemiLehe, line 955: As a follow-up PR: could you add a comment that explains why you had to add `false` here?

RemiLehe (approved):

PR #5910, opened 2025-05-24
Simplify logic in the DSMC code

RemiLehe:
This PR simplifies the logic in the DSMC code, esp. in the case where the same species is both in the colliding species and in the products of an ionization reaction, such as in the following case:
$e^- + H \rightarrow 2 e^- + H^+$
`species = electron H`
`product_species = electron Hplus`

- Previously: we were careful to list this species only once in `m_product_species` (e.g., `m_product_species = electron H Hplus`), but this resulted in complicated logic to make sure that the array allocation and offsets were properly calculated.
- With this PR: the species can be listed several times in `m_product_species` (e.g., `m_product_species = electron H electron Hplus`). The corresponding logic is more streamlined (see comments in the code).

In addition, the new code forces the first species in `product_species` to be an electron in the case where ionization is used, as this also simplifies some of the logic. 
This check is more strict that the previous `if` condition (which was used to swap indices internally so that the electron species is indeed the first product): instead of simply checking for the sign of the charge, we now explicitly check the charge and mass of the species.

This will likely raise error messages in existing user code, but the error message should hopefully be clear enough for the user to make the required change.

### DIFF
--- a/Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
+++ b/Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
@@ -70,10 +70,11 @@ DSMCFunc::DSMCFunc (
             // during the scattering operation).
             amrex::Vector<std::string> product_species_names;
             pp_collision_name.getarr("product_species", product_species_names);
+            // Check that the charge and mass of the species is consistent with an electron species
             auto& species1 = mypc->GetParticleContainerFromName(product_species_names[0]);
-            if(species1.AmIA<PhysicalSpecies::electron>() == false) {
-                amrex::Abort("The first species in " + collision_name + ".product_species must be an electron, " +
-                "and must be marked as such with species_type=electron.");
+            if( abs( species1.getCharge() - PhysConst::q_e ) > 1e-6*PhysConst::q_e   ||
+                abs( species1.getMass() - PhysConst::m_e ) > 1e-6*PhysConst::m_e ) {
+                amrex::Abort("The first species in " + collision_name + ".product_species must be an electron.");
             }
 
             // TODO: add a check that the ionization species has the same mass
--- a/Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
+++ b/Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
@@ -70,10 +70,9 @@ DSMCFunc::DSMCFunc (
             // during the scattering operation).
             amrex::Vector<std::string> product_species_names;
             pp_collision_name.getarr("product_species", product_species_names);
-            // Check that the charge and mass of the species is consistent with an electron species
+            // Check that the charge is consistent with an electron species
             auto& species1 = mypc->GetParticleContainerFromName(product_species_names[0]);
-            if( std::abs(  species1.getCharge() + PhysConst::q_e ) > 1e-6*PhysConst::q_e   ||
-                std::abs( species1.getMass() - PhysConst::m_e ) > 1e-6*PhysConst::m_e ) {
+            if( species1.getCharge() < 0._prt ) {
                 amrex::Abort("The first species in " + collision_name + ".product_species must be an electron.");
             }
 
--- a/Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
+++ b/Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
@@ -56,16 +56,27 @@ DSMCFunc::DSMCFunc (
         WARPX_ALWAYS_ASSERT_WITH_MESSAGE(process.type() != ScatteringProcessType::INVALID,
                                         "Cannot add an unknown scattering process type");
 
-        // Only one ionization process is currently supported as part of a given
-        // collision set.
         if (process.type() == ScatteringProcessType::IONIZATION) {
-            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
-                !ionization_flag,
-                "DSMC only supports a single ionization process"
-            );
+
+            // Only one ionization process is currently supported as part of a given
+            // collision set.
+            if (ionization_flag) {
+                amrex::Abort("Multiple ionization processes were specified in " + collision_name +
+                ".scattering_processes, but DSMC only supports a single ionization process.");
+            }
             ionization_flag = true;
 
-            // And add a check that the ionization species has the same mass
+            // Ensure that the first product species is always an electron (which is assumed
+            // during the scattering operation).
+            amrex::Vector<std::string> product_species_names;
+            pp_collision_name.getarr("product_species", product_species_names);
+            auto& species1 = mypc->GetParticleContainerFromName(product_species_names[0]);
+            if(species1.AmIA<PhysicalSpecies::electron>() == false) {
+                amrex::Abort("The first species in " + collision_name + ".product_species must be an electron, " +
+                "and must be marked as such with species_type=electron.");
+            }
+
+            // TODO: add a check that the ionization species has the same mass
             // (and a positive charge), compared to the target species
         }
         m_scattering_processes.push_back(std::move(process));
--- a/Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
+++ b/Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
@@ -72,7 +72,7 @@ DSMCFunc::DSMCFunc (
             pp_collision_name.getarr("product_species", product_species_names);
             // Check that the charge is consistent with an electron species
             auto& species1 = mypc->GetParticleContainerFromName(product_species_names[0]);
-            if( species1.getCharge() < 0._prt ) {
+            if( species1.getCharge() >= 0._prt ) {
                 amrex::Abort("The first species in " + collision_name + ".product_species must be an electron.");
             }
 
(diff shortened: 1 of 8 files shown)


Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
```diff
@@ -56,16 +56,27 @@ DSMCFunc::DSMCFunc (
...
-                !ionization_flag,
-                "DSMC only supports a single ionization process"
-            );
+
+            // Only one ionization process is currently supported as part of a given
+            // collision set.
+            if (ionization_flag) {
+                amrex::Abort("Multiple ionization processes were specified in " + collision_name +
+                ".scattering_processes, but DSMC only supports a single ionization process.");
+            }
```
RemiLehe, line 66: The above makes the error message more explicit ; it is independent/orthogonal to the rest of this PR.

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -58,7 +58,7 @@ public:
         const amrex::ParticleReal m1, const amrex::ParticleReal m2,
         const amrex::Vector<amrex::ParticleReal>& /*products_mass*/,
         const index_type* AMREX_RESTRICT mask,
-        const amrex::Vector<index_type>& products_np,
+        amrex::Vector<index_type>& products_np,
```
RemiLehe, line 61: This function now modifies `products_np` internally.

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -329,31 +300,31 @@ public:
                 soa_products_data[0].m_rdata[PIdx::w][species1_index] = p_pair_reaction_weight[i];
 
                 // create a copy of the first product species at the location of species 2
-                const auto product1_index = ioniz_product1_offset + with_product_p_offsets[i];
-                copy_species1[ioniz_product1_list_index](soa_products_data[ioniz_product1_list_index], soa_2, static_cast<int>(p_pair_indices_2[i]),
+                const auto product1_index = products_np_data[2] + with_product_p_offsets[i];
+                copy_species1[2](soa_products_data[2], soa_2, static_cast<int>(p_pair_indices_2[i]),
```
RemiLehe, line 304: Btw, @roelof-groenewald this is orthogonal to this PR, but should this be `copy_species2` instead of `copy_species1`?
```suggestion
                copy_species2[2](soa_products_data[2], soa_2, static_cast<int>(p_pair_indices_2[i]),
```

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -329,31 +300,31 @@ public:
...
                                 static_cast<int>(product1_index), engine);
                 // Set the weight of the new particle to p_pair_reaction_weight[i]
-                soa_products_data[ioniz_product1_list_index].m_rdata[PIdx::w][product1_index] = p_pair_reaction_weight[i];
+                soa_products_data[2].m_rdata[PIdx::w][product1_index] = p_pair_reaction_weight[i];
 
                 // create a copy of the other product species at the location of species 2
-                const auto product2_index = ioniz_product2_offset + with_product_p_offsets[i];
-                copy_species1[ioniz_product2_list_index](soa_products_data[ioniz_product2_list_index], soa_2, static_cast<int>(p_pair_indices_2[i]),
+                const auto product2_index = products_np_data[3] + with_product_p_offsets[i];
+                copy_species1[3](soa_products_data[3], soa_2, static_cast<int>(p_pair_indices_2[i]),
```
RemiLehe, line 311: Same question here
```suggestion
                copy_species2[3](soa_products_data[3], soa_2, static_cast<int>(p_pair_indices_2[i]),
```

Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
```diff
@@ -56,16 +56,28 @@ DSMCFunc::DSMCFunc (
...
             ionization_flag = true;
 
-            // And add a check that the ionization species has the same mass
+            // Ensure that the first product species is always an electron (which is assumed
+            // during the scattering operation).
+            amrex::Vector<std::string> product_species_names;
+            pp_collision_name.getarr("product_species", product_species_names);
+            // Check that the charge and mass of the species is consistent with an electron species
+            auto& species1 = mypc->GetParticleContainerFromName(product_species_names[0]);
+            if( std::abs(  species1.getCharge() - PhysConst::q_e ) > 1e-6*PhysConst::q_e   ||
```
RemiLehe:
```suggestion
            if( std::abs(  species1.getCharge() + PhysConst::q_e ) > 1e-6*PhysConst::q_e   ||
```

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.cpp
```diff
@@ -28,35 +28,11 @@ SplitAndScatterFunc::SplitAndScatterFunc (const std::string& collision_name,
         // if ionization is one of the processes, check if one of the colliding
         // species is also used as a product species
```
JustinRayAngus: This statement suggests that below one is checking if one of the colliding species is also a product species. It's not clear to me that such a check is being performed.

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.cpp
```diff
@@ -28,35 +28,11 @@ SplitAndScatterFunc::SplitAndScatterFunc (const std::string& collision_name,
...
-                m_num_products_host.push_back(1); // the non-target species
-                m_num_products_host.push_back(0); // the target species
-                m_num_products_host.push_back(1); // first product species
-                m_num_products_host.push_back(1); // second product species
-            }
+            m_num_product_species = 4;
+            m_num_products_host.push_back(1); // the non-target species
+            m_num_products_host.push_back(0); // the target species
+            m_num_products_host.push_back(1); // first product species
+            m_num_products_host.push_back(1); // second product species
```
JustinRayAngus, line 33: I probably need to look at this part of the code in more detail, but I find the m_num_products_host and m_num_products variables confusing. In the .H file it says that the m_num_products_hosts vector of size m_num_product_species stores how many particles of a given species are produced by a collision event. It makes sense to me that there would be 1 for the first and second product species. But why is there 1 for the non-target species?

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.cpp
RemiLehe:
```suggestion
```

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.cpp
roelof-groenewald: This is because the non-target species is split into two particles. The "new" particle will have the same weight as the target species, while the original particle will keep the remaining weight (which could be 0 - in which case the particle will be removed).

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
roelof-groenewald: The reason there are two vectors (`copy_species1` and `copy_species2`) is because in the fusion module products are placed at both the colliding particles' positions. In this module the new particles are placed at the "target species" (species1) positions, so in all cases `copy_species1` is used. Does that make sense? I'd be happy to explain that in more detail if needed.

Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
```diff
@@ -56,16 +56,28 @@ DSMCFunc::DSMCFunc (
...
 
-            // And add a check that the ionization species has the same mass
+            // Ensure that the first product species is always an electron (which is assumed
+            // during the scattering operation).
+            amrex::Vector<std::string> product_species_names;
+            pp_collision_name.getarr("product_species", product_species_names);
+            // Check that the charge and mass of the species is consistent with an electron species
+            auto& species1 = mypc->GetParticleContainerFromName(product_species_names[0]);
+            if( std::abs(  species1.getCharge() + PhysConst::q_e ) > 1e-6*PhysConst::q_e   ||
+                std::abs( species1.getMass() - PhysConst::m_e ) > 1e-6*PhysConst::m_e ) {
```
roelof-groenewald: These strict tests might lead to some issues. Often "enhanced electron mass" is used as a numerical tool to reduce the electron plasma and/or cyclotron frequencies. With these tests the code would not allow anything other than "real" electrons to be used with the DSMC routine. That is why I originally only checked for negative charge, but I also understand that is not necessarily strict enough (since negative ions could be used). Can we strike a middle ground somehow?

roelof-groenewald (changes requested):
This is great! Thanks for the significant simplification in logic. I just have the comment about the strictness of checking for the first product species being an electron.

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
RemiLehe: OK, thanks for clarifying.

RemiLehe: Hm, but actually, shouldn't the target species be `species2`, precisely?

Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
RemiLehe: OK, sounds good. For now, I will replace this with something that checks for negative charge.

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
roelof-groenewald: Ahh, yes, that is true. This should change then, I agree.

roelof-groenewald (approved):
Looks good! Thanks!

PR #5914, opened 2025-05-27
CI: Update PETSc Workflows

EZoni:
This removes the two PETSc workflows labeled as "debug", which I think differed from the other ones by
- the `--with-debugging=1` flag for the PETSc configuration, and
- the `-DCMAKE_BUILD_TYPE=Debug` flag for the WarpX configuration.

Note that `CXXFLAGS: "-Werror"` was set for all workflows and is maintained for the workflows that are kept.

Note also that none of the workflows is running WarpX at this point - this isn't changed in this PR.

@RemiLehe @debog @atmyers 
We can still choose to keep only the "debug" ones instead - please feel free to comment if you have new opinions on this.

To-do:
- [x] Run on `ubuntu-24.04` instead of `ubuntu-latest` (we usually fix the Ubuntu version for these complex workflows)
- [ ] Flag workflows as required in the repository's settings (must be done by contributors with admin/owner access)

### DIFF
--- a/.github/workflows/petsc.yml
+++ b/.github/workflows/petsc.yml
@@ -1,4 +1,4 @@
-name: PETSc
+name: 🐧 PETSc
 
 on:
   push:
@@ -18,7 +18,7 @@ jobs:
 
   test_petsc_gcc:
     name: GCC
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     needs: check_changes
     if: ${{ github.event.pull_request.draft == false && needs.check_changes.outputs.has_non_docs_changes == 'true' }}
     env:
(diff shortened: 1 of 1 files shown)


debog:
Thanks, @EZoni. Can you retain the `--with-debugging=1` for PETSc? At some point, I will try to add running WarpX for an example in this test, and I want to run with PETSc's debug build.

EZoni:
@debog 

> Thanks, @EZoni. Can you retain the --with-debugging=1 for PETSc? At some point, I will try to add running WarpX for an example in this test, and I want to run with PETSc's debug build.

Got it, done in [6478179](https://github.com/BLAST-WarpX/warpx/pull/5914/commits/64781798264f84d81638edab47ab7c54e620dd86).

RemiLehe (approved):

PR #5922, opened 2025-05-30
For FieldEnergy, use validbox instead of tilebox for the volume factor

dpgrote:
The FieldEnergy diagnostic sums the field energy over all grid cells. Along directions that are nodal, the loop includes the nodes that are at the lower and upper end of the block, so those nodes get counted twice. The volume factor is set to ½ to account for this. However, when tiling is done, the upper end of the tile is not included except when it is also the upper end of the valid box. But the factor of ½ was being applied at lower and upper end of the tile. The fix is to only apply the ½ to the lower and upper end of the valid box. This PR fixes this.


JustinRayAngus (approved):
Thanks for this fix, Dave!

PR #5925, opened 2025-06-02
Add new field diagnostic that indicates whether a given cell is covered by the EB or not

RemiLehe:
As described in the title, this adds a new field, which can then be used to visualized where the eb is located.

For instance, in the case of `electrostatic_sphere_eb`:
Screenshot 2025-06-02 at 4 59 24 PM


Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -626,6 +632,7 @@ FullDiagnostics::AddRZModesToDiags (int lev)
             lev, m_crse_ratio, false, ncomp_multimodefab));
         AddRZModesToOutputNames(std::string("divE"), ncomp_multimodefab);
     }
+    // eb_covered
```
```suggestion
```

WeiqunZhang (approved):

PR #5866, opened 2025-05-06
Implement particle_max_grid_crossings

dpgrote:
This PR implements the optional input parameter `particles.max_grid_crossings` that allows letting particles violate the Courant condition of one grid crossing per step, which can be needed when using the implicit scheme (with large time steps).

PR #5929, opened 2025-06-03
Infrastructure for using full mass matrices in the implicit solvers

JustinRayAngus:
This PR adds the infrastructure for using the full mass matrices in the implicit solvers. This is a follow-on to PR #5768, where only the diagonal terms of the diagonal matrices were computed and used. With this new PR, the user has the option to compute the full mass matrices which can then be used to compute the linearized current density for the Jacobian calculation needed during the GMRES iterations of JFNK, bypassing the need to do particle based operations.

The setup is such that the user can use the full mass matrices for computing J during the Jacobian calculation in GMRES, and/or use the diagonal terms for the preconditioner. How to utilize the mass matrices is controlled by the following flags in the input deck:

`implicit_evolve.use_mass_matrices_jacobian = true`
`implicit_evolve.use_mass_matrices_pc = true`

If the second flag is true and the first flag is false, then memory is only allocated for the mass matrix terms that are used by the PC (diagonal terms only for now).

The ability to use the full mass matrices as described above is only implemented for 1D and 2D Cartesian geometries and only for direct deposit. Other geometries and other deposition schemes are deferred to future PRs.

- [x] Merge PR #5866 to correctly set J and E/B guard cells.
- [x] Merge PR #5919 and verify that everything works for shape = 3 and shape = 4.

A CI test will be implemented in a future PR once the full mass matrices have been implemented for charge-conserving depositions.


Source/Parallelization/GuardCellManager.cpp
```diff
@@ -359,4 +359,10 @@ guardCellManager::Init (
             ng_MovingWindow[moving_window_dir] = 1;
         }
     }
+    for (int i = 0; i < AMREX_SPACEDIM; i++) {
+        ng_alloc_J[i] = nox+1;
+        ng_depos_J[i] = nox+1;
+        ng_alloc_Rho[i] = nox+1;
+        ng_alloc_EB[i] = nox+1;
+    }
```
JustinRayAngus: Just commenting here that these changes are a temporary work-around that will not be needed once PR #5866 is merged.

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.H
```diff
@@ -179,6 +198,35 @@ protected:
...
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                !m_use_mass_matrices,
+                "Using mass matrices is not setup for DIM = RCYLINDER!");
+#endif
+#if defined(WARPX_DIM_RSPHERE)
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                !m_use_mass_matrices,
+                "Using mass matrices is not setup for DIM = RSHERE!");
+#endif
+#if defined(WARPX_DIM_RCYL)
```
dpgrote: This should presumably be `WARPX_DIM_RZ`

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
```diff
@@ -89,6 +91,324 @@ Array<LinOpBCType,AMREX_SPACEDIM> ImplicitSolver::convertFieldBCToLinOpBC (const
...
+{
+
+    // Copy Efield_fp and current_fp to Efield_fp_save and current_fp_save
+    // Do this BEFORE call to SyncCurrentAndRho() so that BCs are set for J
+
+    using warpx::fields::FieldType;
+    for (int lev = 0; lev < m_num_amr_levels; ++lev) {
+        ablastr::fields::VectorField E = m_WarpX->m_fields.get_alldirs(FieldType::Efield_fp, lev);
+        ablastr::fields::VectorField E0 = m_WarpX->m_fields.get_alldirs(FieldType::Efield_fp_save, lev);
+        amrex::MultiFab::Copy(*E0[0], *E[0], 0, 0, 1, E[0]->nGrowVect());
```
dpgrote: For completeness, this could be `amrex::MultiFab::Copy(*E0[0], *E[0], 0, 0, E[0].nComp(), E[0]->nGrowVect());`

JustinRayAngus:
@dpgrote @EZoni 

There is something funny going on with 1D. All CI tests are failing because executable cannot be found. I have no trouble compiling and running CI tests locally. Error message below.

mpiexec was unable to launch the specified application as it could not access
or execute an executable:

Executable: /home/vsts/work/1/s/build/bin/warpx.1d.MPI.OMP.DP.PDP.OPMD.FFT.EB.QED
Node: fv-az933-910

Any idea of what is going on?

EZoni:
@JustinRayAngus 

This PR is not modifying anything related to CI, so it may be some sort of random glitch. I restarted the failed checks just to see if the issue persists. If that's the case, I will try to run those tests locally from your branch.

JustinRayAngus:
@EZoni I found a bug in one of the 1D implicit input decks related to changes made in this PR. I'm not sure why that would cause the problem mentioned previously, but we will see once the CI tests finish.

One thing I did notice is that the 1D build takes around 53 minutes, which is more than twice as long as the build for 2D, 3D, and RZ. I looked back at several PR's and the build times are similar. Any reason why the 1D build would be expected to build much slower?

JustinRayAngus:
@EZoni 

Found the issue. It was hard to find because it happens during the Build part of the 1D CI tests, which still gets a green checkmark even though it fails there.

EZoni:
@JustinRayAngus 

Got it, thanks. Very strange, though. If we look at that step in the workflow,
https://github.com/BLAST-WarpX/warpx/blob/b4009ed9944ef4262140d1afb498573823c24352/.azure-pipelines.yml#L134-L150
we set `set -o nounset errexit pipefail` precisely to avoid a situation where the step fails but it's not caught at the workflow level.

Do you remember what type of build error was occurring?

JustinRayAngus:
@EZoni 

There was an undefined variable being used, but only for DIM_RCYLINDER and DIM_RSPHERE. The error was inside the following compile flag ifdef lines:

#elif defined(WARPX_DIM_RCYLINDER) || defined(WARPX_DIM_RSPHERE)
#endif

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.H
```diff
@@ -153,6 +155,23 @@ protected:
...
+     * \brief Whether to use mass matrices in the preconditioner
+     */
+    bool m_use_mass_matrices_pc = false;
+
+    /**
+     * \brief Direction-dependent component number of mass matrices elements
+     */
+    amrex::IntVect m_ncomp_xx, m_ncomp_xy, m_ncomp_xz;
+    amrex::IntVect m_ncomp_yx, m_ncomp_yy, m_ncomp_yz;
+    amrex::IntVect m_ncomp_zx, m_ncomp_zy, m_ncomp_zz;
```
ax3l: Let's default init those variables here in-place, e.g. to zero, like the other control variables?

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
```diff
@@ -89,6 +91,325 @@ Array<LinOpBCType,AMREX_SPACEDIM> ImplicitSolver::convertFieldBCToLinOpBC (const
...
+        ablastr::fields::VectorField SZ = m_WarpX->m_fields.get_alldirs(FieldType::MassMatrices_Z, lev);
+
+        const amrex::IntVect Jx_nodal = J[0]->ixType().toIntVect();
+        const amrex::IntVect Jy_nodal = J[1]->ixType().toIntVect();
+        const amrex::IntVect Jz_nodal = J[2]->ixType().toIntVect();
+
+        // Compute the component offset in each direction (careful with staggering)
+        amrex::IntVect offset_xx, offset_xy, offset_xz;
+        amrex::IntVect offset_yx, offset_yy, offset_yz;
+        amrex::IntVect offset_zx, offset_zy, offset_zz;
```
ax3l, line 143: You could potentially default-init the offsets to zero, but it looks below like all components are assigned anyway.

ax3l (approved):
This looks pretty sound to me.

I added a small suggestion.

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
JustinRayAngus: Since they are defined below, I think I will keep this as is.

Source/Fields.H
```diff
@@ -82,7 +82,11 @@ namespace warpx::fields
         Efield_avg_cp,
         Bfield_avg_cp,
         B_old, /**< Stores the value of B at the beginning of the timestep, for the implicit solver */
-        MassMatrices,    // Stores full mass matrix elements for the implicit solver
+        Efield_fp_save,  // Efield_fp at beginning of Newton step used by the implicit solvers
+        current_fp_save, // current_fp at beginning of Newton step  used by the implicit solvers
```
JustinRayAngus, line 86: @ax3l @dpgrote I'm not married to the naming convention here. See ImplicitSolver::SaveEandJ() and ImplicitSolver::ComputeJfromMassMatrices() functions for usage of these variables

atmyers:
I tried this locally on my workstation GPU and everything appears to work as expected.

atmyers (approved):

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.H
```diff
@@ -153,6 +155,29 @@ protected:
...
+     */
+    amrex::IntVect m_ncomp_xx = amrex::IntVect::TheZeroVector();
+    amrex::IntVect m_ncomp_xy = amrex::IntVect::TheZeroVector();
+    amrex::IntVect m_ncomp_xz = amrex::IntVect::TheZeroVector();
+    amrex::IntVect m_ncomp_yx = amrex::IntVect::TheZeroVector();
+    amrex::IntVect m_ncomp_yy = amrex::IntVect::TheZeroVector();
+    amrex::IntVect m_ncomp_yz = amrex::IntVect::TheZeroVector();
+    amrex::IntVect m_ncomp_zx = amrex::IntVect::TheZeroVector();
+    amrex::IntVect m_ncomp_zy = amrex::IntVect::TheZeroVector();
+    amrex::IntVect m_ncomp_zz = amrex::IntVect::TheZeroVector();
```
WeiqunZhang, line 179: No need to change. But @JustinRayAngus mentioned in the past that `TheZeroVector()` is too long to type. If we want a short version, we could write `amrex::IntVect m_ncomp_zz{0}` or `amrex::IntVect m_ncomp_zz = amrex::IntVect{0}`.

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
```diff
@@ -89,6 +91,325 @@ Array<LinOpBCType,AMREX_SPACEDIM> ImplicitSolver::convertFieldBCToLinOpBC (const
...
+            amrex::Array4<const amrex::Real> const& Syx = SY[0]->array(mfi);
+            amrex::Array4<const amrex::Real> const& Syy = SY[1]->array(mfi);
+            amrex::Array4<const amrex::Real> const& Syz = SY[2]->array(mfi);
+
+            amrex::Array4<const amrex::Real> const& Szx = SZ[0]->array(mfi);
+            amrex::Array4<const amrex::Real> const& Szy = SZ[1]->array(mfi);
+            amrex::Array4<const amrex::Real> const& Szz = SZ[2]->array(mfi);
+
+            // Use grown boxes here with all guard cells
+            amrex::Box Jbx = amrex::convert(mfi.validbox(),J[0]->ixType().toIntVect());
```
WeiqunZhang, line 197: No need to change. Just FYI. `amrex::Box Jbx = amrex::convert(mfi.validbox(),J[0]->ixType());` also works.

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
```diff
@@ -89,6 +91,325 @@ Array<LinOpBCType,AMREX_SPACEDIM> ImplicitSolver::convertFieldBCToLinOpBC (const
...
+
+                // Compute Sxx*dEx
+                for (int dim=0; dim<AMREX_SPACEDIM; ++dim) {
+                    index_min[dim] = std::max(-offset_xx[dim],Ebx.smallEnd(dim)-idx[dim]);
+                    index_max[dim] = std::min(ncomp_xx[dim]-1-offset_xx[dim],Ebx.bigEnd(dim)-idx[dim]);
+                }
+                amrex::Real SxxdEx = 0.0;
+                for (int ii = index_min[0]; ii <= index_max[0]; ++ii) {
+                    for (int jj = index_min[1]; jj <= index_max[1]; ++jj) {
+                        for (int kk = index_min[2]; kk <= index_max[2]; ++kk) {
```
WeiqunZhang, line 236: It is probably more efficient, if the ii is the innermost loop.

WeiqunZhang (approved):

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
JustinRayAngus: @WeiqunZhang @ax3l I would like to test this at some point. @dpgrote tested the ordering of such loops for current deposition in the recent past and saw no difference.

PR #5930, opened 2025-06-03
Switch Windows Clang test to windows-2022

RemiLehe:
Windows-2019 is deprecated ; see:
https://github.com/actions/runner-images/issues/12045


RemiLehe:
The tests fail with SegFault...

.github/workflows/windows.yml
```diff
@@ -96,7 +96,7 @@ jobs:
     - name: Build & Install
       shell: cmd
       run: |
-        call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\vc\Auxiliary\build\vcvarsall.bat" x64
+        call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x64
```
RemiLehe: See https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md, which lists this path

.github/workflows/windows.yml
```diff
@@ -130,8 +128,9 @@ jobs:
     - name: Test
       shell: cmd
       run: |
-        call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\vc\Auxiliary\build\vcvarsall.bat" x64
+        call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\vc\Auxiliary\build\vcvarsall.bat" x64
         set "PATH=C:/Program Files (x86)/WarpX/bin/;%PATH%"
 
         python3 Examples/Tests/gaussian_beam/inputs_test_3d_gaussian_beam_picmi.py --diagformat=openpmd
+        cat Backtrace.0.0
```
RemiLehe:
```suggestion
```

ax3l:
Indeed, I also see something is off combining ClangCl and Python runners... Debugging in https://github.com/BLAST-ImpactX/impactx/pull/992

Interestingly, the pyAMReX builds work.

ax3l:
Ok, so a solution might be for now to build ClangCl in Debug.

What I observe is: Only the Python tests segfault the moment that we build in Release/RelWithDebInfo. I was not yet able to run a Windows debugger on it, and this might be environment specific to this runner because we do _not_ see similar issues with ClangCl on Conda-Forge (building Release).

What could happen in the specific environment here: it is possible that the Python build pulls in, in some way, a debug build and standard-lib, which is not compatible in Windows (debug-symbol and release stdlibs are ABI incompatible in Windows). So for now for testing, it is sufficient to just roll with it and build Debug to keep going. Running `ldd` to find the dynamic libs did not immediately show something weird...

We could ultimately run a debugger like WinDbg (see commits for details) with `tmate` interactively on this and find out more, but for now I would book this as environment issue.

EZoni (approved):
Thanks both for digging into this! :pray:

ax3l:
For future updates: I think we can set `Python3_FIND_ABI` on the command line for these CI runners to better fine-select with Python ABI flavor is picked up.
https://cmake.org/cmake/help/latest/module/FindPython3.html#hints

PR #5932, opened 2025-06-04
Fix energy and momentum conservation after pair wise Coulomb collisions

dpgrote:
Binary, pair wise, Coulomb collisions between two particles with differing weights does not exactly conserve energy and momentum. It would only be conserved on average over many collisions. This PR adds a fix to the momentum of the particles that restores the energy and momentum so that it is exactly conserved. This is done in each grid cell, so that the energy and momentum in each cell is conserved. This correction is done for both intra-species and inter-species collisions. The method is described here https://doi.org/10.1016/j.jcp.2025.113927.

The implementation is designed for pairwise Coulomb collisions, but is implemented in the binary collisions code. It does not account for the energy and momentum of newly created particles or deleted particles. Currently, it can only be activated when using pairwise Coulomb collisions.

A question remains about how to handle the loss of precision that happens during the energy correction. When the correction is small, the calculations will take the difference of two similar large numbers, which can lose precision. One solution is to use `long double`, but this does not work on GPUs (where the type is not available) and Macs (where the type is the same as double). On a Mac, in the CI test, the relative change in energy is ~1.e-11 compared to 1.e-14 in Linux. For now, the code uses `long double` when it is available, otherwise only `double`.

Todo:
- [x] Add documentation
- [x] Handle cases where the correction fails
- [x] Add CI test


RemiLehe:
Offline discussion: this correction does not need to be used for `fusion`, `dsmc`, `linear_breit_wheeler`, `linear_compton` since these algorithm conserve energy automatically ; we should raise a warning/error if a user tries to activate it in these case.

Source/Utils/ParticleUtils.H
```diff
@@ -170,6 +170,193 @@ namespace ParticleUtils {
...
+
+        // long double is used when available for the calculations below since
+        // extra precision is required when updating the energy since it
+        // involves taking differences between similar numbers, which can lose
+        // precision.
+#ifdef AMREX_USE_GPU
+#define LONGDOUBLE double
+#else
+#define LONGDOUBLE long double
+#endif
```
ax3l: Uh, this can be surprising on CPU/GPU comparisons :D

Source/Utils/ParticleUtils.H
```diff
@@ -170,6 +170,193 @@ namespace ParticleUtils {
...
+            const amrex::ParticleReal ux2 = uxp[ indices[i2] ];
+            const amrex::ParticleReal uy2 = uyp[ indices[i2] ];
+            const amrex::ParticleReal uz2 = uzp[ indices[i2] ];
+            const amrex::ParticleReal wpmp1 = mass*w[ indices[i1] ];
+            const amrex::ParticleReal wpmp2 = mass*w[ indices[i2] ];
+
+            const amrex::ParticleReal ux = ux1 - ux2;
+            const amrex::ParticleReal uy = uy1 - uy2;
+            const amrex::ParticleReal uz = uz1 - uz2;
+            const LONGDOUBLE dbetasq = (ux*ux + uy*uy + uz*uz)/c2;
```
ax3l: Phew, `long double` is a strong requirement on any platform (and is implemented in-consistently across platforms).

Is there a way to normalize or taylor data for those operations?

I am not opposed to merging this for now, but this something that we might need to document (user-facing, in terms of costs) and as an issue for follow-ups.

ax3l (commented):
I have not read the paper yet, but structurally speaking this PR looks sound to me.

There is an implementation detail using `long double`, which is concerning because it might lead to wrong/less precise values on GPU (uses `double` there, because GPUs do no support `long double`) but more importantly is generally really not performant for an operation that we perform for each particle.

That said, I am ok to merge it for now as long as this is documented as an issue an user-facing warning.

Source/Utils/ParticleUtils.H
JustinRayAngus: From a physics standpoint, energy conservation will be plenty good enough without using long double. It is only needed for rigorous machine-level conservation.

Docs/source/usage/parameters.rst
```diff
@@ -2227,6 +2227,36 @@ Details about the collision models can be found in the :ref:`theory section <mul
     Only for ``dsmc`` with impact ionization. This specifies which one of the
     colliding particles is ionized.
 
+* ``collisions.correct_energy_momentum`` (`bool`) optional (default 0)
+    For pairwisecoulomb collisions, whether to correct the energy and momentum after the collisions so that they are conserved.
+    In binary collisions, if the weights of the colliding particles are not the same, the collision does not
+    conserve energy and momentum. When this option is on, after the collisions, small modifications are made to the
+    particle mometum so that the energy and momentum are exactly conserved in each cell.
+
+* ``collisions.energy_fraction`` (`float`) optional (default 0.5)
```
JustinRayAngus: The default value for this parameter in PICNIC is 0.05. The idea is that it should be small to not affect scattering physics.

Source/Utils/ParticleUtils.H
ax3l: Then I would definitely use `amrex::ParticleReal` here.

Docs/source/usage/parameters.rst
```diff
@@ -2231,6 +2231,36 @@ Details about the collision models can be found in the :ref:`theory section <mul
     Only for ``dsmc`` with impact ionization. This specifies which one of the
     colliding particles is ionized.
 
+* ``collisions.correct_energy_momentum`` (`bool`) optional (default 0)
```
roelof-groenewald, line 2234: It would be good to add a reference to the paper describing the method to the documentation.

Docs/source/usage/parameters.rst
```diff
@@ -2231,6 +2231,36 @@ Details about the collision models can be found in the :ref:`theory section <mul
...
+    With a value greater than 1, it will distribute more of the correction to particles with higher weights.
+
+* ``<collision_name>.correct_energy_momentum`` (`bool`) optional
+    For pairwisecoulomb collisions, override the parameter ``collisions.correct_energy_momentum`` for the specific collision.
+
+* ``<collision_name>.energy_fraction`` (`float`) optional
+    For pairwisecoulomb collisions, override the parameter ``collisions.energy_fraction`` for the specific collision.
+
+* ``<collision_name>.energy_fraction_max`` (`float`) optional
+    For pairwisecoulomb collisions, override the parameter ``collisions.energy_fraction_ma`` for the specific collision.
```
```suggestion
    For pairwisecoulomb collisions, override the parameter ``collisions.energy_fraction_max`` for the specific collision.
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -432,8 +456,90 @@ public:
...
+                );
+            }
+
+            bool const correct_energy_momentum = m_correct_energy_momentum;
+            amrex::ParticleReal const beta_weight_exponent = m_beta_weight_exponent;
+
             // Loop over cells
             WARPX_PROFILE_VAR("BinaryCollision::doCollisionsWithinTile::computeDensityTemperatures", prof_computeDensityTemperatures);
+            if (correct_energy_momentum) {
+                amrex::ParallelFor( np1,
```
roelof-groenewald, line 536: Why did you choose to have this kernel loop over the particles rather than the cells as is done below this? If populating the local energy and momentum arrays were done in the same way the local density is obtained in the kernel below it would be easier to read the code in my opinion. (Alternatively, the local density and temperature calculations can be moved into this kernel).

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -674,8 +866,145 @@ public:
...
+                );
+            }
+
+            bool const correct_energy_momentum = m_correct_energy_momentum;
+            amrex::ParticleReal const beta_weight_exponent = m_beta_weight_exponent;
+
             // Loop over cells
             WARPX_PROFILE_VAR("BinaryCollision::doCollisionsWithinTile::findDensityTemperatures", prof_findDensityTemperatures);
+            if (correct_energy_momentum) {
+                amrex::ParallelFor( np1,
```
roelof-groenewald: Same comment here about looping over cells rather than particles. Here that would also allow the two kernels to be merged into one, which should be a nice performance optimization.

Source/Utils/ParticleUtils.H
```diff
@@ -170,6 +170,190 @@ namespace ParticleUtils {
...
+                if (energy_frac_eff > energy_fraction_max || loop_count > max_loop_count) {
+#ifndef AMREX_USE_GPU
+                    if (numCell > 50) {
+                        amrex::Print() << "Notice: energy_frac_eff = " << energy_frac_eff << "\n";
+                        amrex::Print() << "        on loop_count = " << loop_count << "\n";
+                        amrex::Print() << "        is larger than energy_frac_max = " << energy_fraction_max << "\n";
+                        amrex::Print() << "        num_parts    = " << numCell << "\n";
+                        amrex::Print() << "        deltaE     = " << deltaE << "\n";
+                    }
+#endif
```
roelof-groenewald: I'm guessing these print statements are just for debugging and will be removed, right?

Source/Utils/ParticleUtils.H
dpgrote: Done

Source/Utils/ParticleUtils.H
dpgrote: Switched to use only `ParticleReal`

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
dpgrote: It is faster to have the loop over particles. See PR #5981.

Source/Utils/ParticleUtils.H
dpgrote: Good question. If this situation arises, it would be good to have some notification since it means that this correction failed in a place that is unlikely. The printed info helps to understand why it failed. The alternative is to save some information and output it outside the kernel (this would work better on the GPU).

roelof-groenewald: That sounds good, then the `WarnManager` can be used too rather than regular print statements.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -447,25 +470,98 @@ public:
...
-        WARPX_PROFILE_VAR("BinaryCollision::doCollisionsWithinTile::findDensityTemperatures::atomics", prof_findDensityTemperatures_atomics);
+
+            if (m_correct_energy_momentum) {
+                WARPX_PROFILE_VAR("BinaryCollision::doCollisionsWithinTile::findMomentumEnergy", prof_findMomentumEnergy);
+                amrex::ParallelFor( np1,
+                    [=] AMREX_GPU_DEVICE (int i1) noexcept
+                    {
+                        u1x_before_ptr[i1] = u1x[i1];
+                        u1y_before_ptr[i1] = u1y[i1];
+                        u1z_before_ptr[i1] = u1z[i1];
```
roelof-groenewald: Just for my education: if `ux/y/z_n` exist as particle attributes, this will overwrite those values. Is it the case that the values in those arrays are already equal to `u1x/y/z`?

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -447,25 +470,98 @@ public:
...
+                }
+
+            }
+
+            amrex::ParticleReal const beta_weight_exponent = m_beta_weight_exponent;
+
             WARPX_PROFILE_VAR("BinaryCollision::doCollisionsWithinTile::findDensityTemperatures", prof_findDensityTemperatures);
-        WARPX_PROFILE_VAR("BinaryCollision::doCollisionsWithinTile::findDensityTemperatures::atomics", prof_findDensityTemperatures_atomics);
+
+            if (m_correct_energy_momentum) {
```
roelof-groenewald: I have a similar question as before now. Here there is a loop over particles inside `if (m_correct_energy_momentum)`, and then immediately afterwards there is another loop over the particles with two `if` blocks inside the kernel. Would it be more performant to merge these two kernels and just add a third `if` block to the existing kernel, or should the existing kernel rather be split into two each inside their respective `if` blocks? Let me know if the question is not clear.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
dpgrote: The `ux/y/z_n` are there for the implicit solvers which need to save the velocities at the start of the time steps. The arrays are only used during the implicit advance. Using them here won't interfere with the solver since the collisions are done separately from the solver. As it is now, the data is copied into the `ux/y/z_n` at the start of the solver, so this copy is needed here also to get the values after the advance.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
dpgrote: I understand the question, but don't know the answer. There's the tradeoff of extra kernel launches versus a more complex kernel. I can do some timings to see is there is a difference. @atmyers would you have any intuition?

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
roelof-groenewald: Cool! Thanks for clarifying.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
dpgrote: I did some tests on Lassen and the combined kernel was definitely faster, by around 40%, presumably because of the overhead of the kernel launching. I implemented the combined kernels. Note that I also added the `if` check around the `ParallelFor` checking if any of the flags are set.

roelof-groenewald: Great! That's good to know. Thanks for taking the time to do the tests.

Source/Utils/ParticleUtils.H
roelof-groenewald: I think this is the last thing to address before this PR is ready to be merged, right?

dpgrote: This is now cleaned up.

roelof-groenewald (approved):

PR #5935, opened 2025-06-06
Fix secondary ion emission test for multi-rank MPI runs

### LINKED ISSUE #5916: Unable to modify particle containers using MPI
Greeting, 

I am currently trying to implement secondary electron emission in a similar scheme to the provided example.
https://github.com/BLAST-WarpX/warpx/tree/development/Examples/Tests/secondary_ion_emission

It works as expected when using a single thread but hangs indefinitely when MPI parallelization is used. No error output (because it technically did not crash).
The issue occurs during the call of elect_pc.add_particles().

We tried testing the original file "inputs_test_rz_secondary_ion_emission_picmi.py". 
Same issue. No problem with 1 thread. Hangs indefinitely with MPI.

We have used warpX in a different context and MPI without issues so I do not believe it is related to the setup.

Are there special precautions that need to be taken when handling the high-level containers with MPI?
What is the best practice for working with them in MPI contexts?
Is rank/tile awareness included in these high-level interfaces?
Is the listed example supposed to work with MPI?

Any guidance on the issue is greatly appreciated.

RemiLehe:
This makes the `test_rz_secondary_ion_emission_picmi` test run on **2 MPI ranks** (previously 1), and fixes the parallel issues that prevented it from doing so.

Note: I would recommend reviewing with Github's "Hide Spaces" feature. Since much of the change is simply indentation.

Closes #5916

### Changes
- **Run the CI test on 2 MPI ranks** (`CMakeLists.txt`).
- **Fix the deadlock reported in #5916.** The callback used to call `add_particles` inside an `if n != 0:` guard, so a rank that scraped no ion called it a different number of times than the others. Since `add_particles` triggers a collective redistribution, this hung in parallel. The guard is removed so every rank calls `add_particles` exactly once per step (with empty arrays when nothing was scraped), and `unique_particles=True` makes each rank add only the electrons it generated locally.
- **Use a separate, reproducible random seed on each MPI rank**, so the secondary electrons emitted on different ranks are not drawn from the same random sequence. The seed is chosen so that the expected 2 secondary electrons are emitted.
- **Regenerate the checksum benchmark** for the 2-rank run (now using the per-mode openPMD RZ keys introduced on `development`).


Examples/Tests/secondary_ion_emission/analysis.py
```diff
@@ -23,36 +23,19 @@
...
-    print(f"Relative percentage error for y = {rel_err_y * 100:5.4f} %")
-    print(f"Relative percentage error for z = {rel_err_z * 100:5.4f} %")
-
-    assert (
-        (rel_err_x < tolerance) and (rel_err_y < tolerance) and (rel_err_z < tolerance)
-    ), "Test particle_boundary_interaction did not pass"
+    r = np.sqrt(x[i] ** 2 + y[i] ** 2 + z[i] ** 2)
+    print(f"Electron # {i}: r={r:5.5f}")
+    assert r - radius > 0, "Electron is not outside the sphere"
+    assert r - radius < tolerance, "Electron is not on the surface of the sphere"
```
RemiLehe:
```suggestion
    assert r - radius < tolerance, "Electron is not close to the surface of the sphere"
```

Examples/Tests/secondary_ion_emission/inputs_test_rz_secondary_ion_emission_picmi.py
```diff
@@ -182,85 +185,90 @@ def sigma_nascap(energy_kEv, delta_H, E_HMax):
...
+        x=xe,
+        y=ye,
+        z=ze,
+        ux=uxe,
+        uy=uye,
+        uz=uze,
+        w=we,
+        unique_particles=True,
+    )
+    # unique_particles=True: each MPI rank adds
```
RemiLehe: @dpgrote Did I understand the argument `unique_particles` correctly here?

Examples/Tests/secondary_ion_emission/inputs_test_rz_secondary_ion_emission_picmi.py
```diff
@@ -189,97 +193,100 @@ def to_numpy(arr):
 
 def secondary_emission():
     buffer = particle_containers.ParticleBoundaryBufferWrapper()  # boundary buffer
-    # STEP 1: extract the different parameters of the boundary buffer (normal, time, position)
```
RemiLehe, line 192: There was no `STEP 2` comment in the file...

EZoni:
I was assigned to this PR a long time ago, apologies for letting it slip through. I just merged `development` (no conflicts, no changes) to check the latest status of the CI checks. I will also review it again to see if we can get it across the finish line.

EZoni (approved):

PR #5936, opened 2025-06-06
Run sanitizer test in double precision

RemiLehe:
The sanitizer test include a run with the implicit solver (`inputs_test_2d_theta_implicit_jfnk_vandb`). 
However, as pointed out offline by @dpgrote, the implicit solver fails to converge to the required precision when run with single precision. This doesn't actually result in an error, but instead in very long runs (because the solver reaches the maximum number of iterations, when trying to converge) and in many warnings.

Therefore, it would make more sense to run this in double precision.

@lucafedeli88 It seems that this test was added in this PR: https://github.com/BLAST-WarpX/warpx/pull/4783. Do you remember if there was a good reason to use single precision here?

### DIFF
--- a/.github/workflows/clang_sanitizers.yml
+++ b/.github/workflows/clang_sanitizers.yml
@@ -57,9 +57,7 @@ jobs:
           -DWarpX_FFT=ON                    \
           -DWarpX_QED=ON                    \
           -DWarpX_QED_TABLE_GEN=ON          \
-          -DWarpX_OPENPMD=ON                \
-          -DWarpX_PRECISION=SINGLE          \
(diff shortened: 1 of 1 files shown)


lucafedeli88 (approved):
Hi @RemiLehe ,

I see... there is potentially a reason to compile in single precision, since using single precision may sometimes lead to arithmetic errors such as divisions by zero, which can be caught by the sanitizer. However, I don't think that we are currently checking for this specific issue.   Therefore, I am in favor of setting the precision to double here, and maybe add a new check to look for arithmetic errors in single precision in a separate PR.

PR #5937, opened 2025-06-08
Add CI test to read particles from openPMD

RemiLehe:
It seems that reading particles from an openPMD file is not being automatically tested in WarpX.

This PR adds a CI test that generates particles in Python and saves them with openPMD-api. The beam is then read in WarpX.


Regression/requirements.txt
```diff
@@ -5,6 +5,7 @@ matplotlib
 mpi4py
 numpy
 openpmd-api
+openpmd-validator
```
RemiLehe:
```suggestion
```

Examples/Tests/gaussian_beam/inputs_test_3d_focusing_gaussian_beam_from_openpmd_prepare.py
```diff
@@ -0,0 +1,118 @@
+"""
+This script prepares the inputs for the test_3d_focusing_gaussian_beam_from_openpmd.py script.
```
RemiLehe:
```suggestion
This script prepares the inputs for the test_3d_focusing_gaussian_beam_from_openpmd script.
```

Examples/Tests/gaussian_beam/inputs_test_3d_focusing_gaussian_beam_from_openpmd
```diff
@@ -0,0 +1,74 @@
...
+#################################
+algo.particle_shape = 3
+
+#################################
+########### PARTICLES ###########
+#################################
+particles.species_names = beam1
+
+beam1.injection_style = external_file
+beam1.injection_file = openpmd_generated_particles.h5
```
RemiLehe:
```suggestion
beam1.injection_file = ../test_3d_focusing_gaussian_beam_from_openpmd_prepare/openpmd_generated_particles.h5
```

ax3l (approved):
Looks great, thank you!

We could add an analysis script to be 100% sure we read what we expected, but this is already great to run the logic!

PR #5940, opened 2025-06-09
Use the correct copy functor for DSMC impact ionization

RemiLehe:
The code called the wrong copy functor in DSMC impact ionization: since the particles to be created are ionized from the target species (i.e. species2), we need to call the copy functor for species 2.


roelof-groenewald (approved):
Seeing as none of the checksums needed updating, it seems that this did not really change anything, but it also definitely doesn't hurt.

PR #5952, opened 2025-06-17
[WIP] Increase CI coverage of runtime real attribute handling

roelof-groenewald:
This increases the CI coverage of functionality to handle runtime particle attributes.

It also serves as a demonstration of a shortcoming in the way `AddNParticles` is written. Currently, runtime attributes are passed to `AddNParticles` but in an assumed order and only attributes beyond the number of attributes passed in are populated with default values (given the user defined expression for initial values). Due to this, the Python version (`add_particles`) forces all runtime attributes that are not passed to `add_particles` to be zero. A better approach would be to have `AddNParticles` accept a list of runtime attribute indices along with the vector of values. All runtime attributes not included in the list of indices would then be initialized with default values.

### DIFF
--- a/Examples/Tests/particle_boundary_scrape/inputs_test_3d_particle_scrape_picmi.py
+++ b/Examples/Tests/particle_boundary_scrape/inputs_test_3d_particle_scrape_picmi.py
@@ -50,6 +50,10 @@ electrons = picmi.Species(
     initial_distribution=uniform_plasma_elec,
     warpx_save_particles_at_xhi=1,
     warpx_save_particles_at_eb=1,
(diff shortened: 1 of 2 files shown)


dpgrote:
I would agree, that that would be a good change to `AddNParticles`. When `AddNParticles` was first written, there was only one attribute (that I remember) so this wasn't an issue. Now with the user definable attributes, updating the interface is a good idea.

BTW, I see other existing CI tests that use Python level user defined attributes, for example `test_3d_laser_acceleration_picmi`. Is this new test testing something not covered there?

PR #5954, opened 2025-06-17
Updating documentation to clarify that certain callbacks are not exec…

clarkse:
…uted at initialization.


dpgrote (approved):
This looks good, thanks!

PR #5956, opened 2025-06-18
Adding flags to field initializer objects in PICMI to control project…

clarkse:
…ion based div cleaner. Refactoring some of the variable names since the div cleaner is additionally applied to external varying A fields. Updated documentation to match refactor and added a CI test to exercise the projection div cleaner in 2D wth analytically loaded fields.


clarkse:
This PR is in response to https://github.com/BLAST-WarpX/warpx/discussions/5953

This is an API change, but it seems to be necessary since the scope of the divergence cleaner application has increased since the initial implementation. I think this should make it a bit clearer and differentiate itself from the diffusion/advection approach used for the PSATD solver, which had a very similar flag to this one for control.

roelof-groenewald:
These changes look good to me. I see there are two tests failing that are still using the old API.

Docs/source/usage/parameters.rst
```diff
@@ -2652,9 +2652,15 @@ Additional parameters
...
-    Whether to use projection method to scrub B field divergence in externally
-    loaded fields. This is automatically turned on if external B fields are loaded.
+* ``warpx.do_initial_div_cleaning`` (`0` or `1` ; default: 0)
+    Whether to use projection method to scrub A/B field divergence in externally
+    loaded fields. This is automatically turned on if external/initial B or time varying A fields are loaded.
+
+* ``projection_div_cleaner.rtol`` (`float`) optional (default `5e12` when double precision and `5e-5` for single precision)
+    Controls the relative tolerance when solving for the projected divergence of the field in the MLMG AMReX solver.
+
+* ``projection_div_cleaner.atol`` (`float`) optional (default `0`)
```
ax3l: Should these two parameters also have a `warpx.` prefix?

ax3l (approved):
LGTM! I would suggest to add the `warpx.` suffix to the projection cleaner options for consistency, feel free to merge afterwards.

Docs/source/usage/parameters.rst
clarkse: I parse them based on not having the warpx. prefix currently. I figured the WarpX class holds the flag for running the cleaner or not, then the cleaner object parses its own parameters from the input.

clarkse: OK fixed.

roelof-groenewald (approved):

PR #5963, opened 2025-06-20
used perform data write in all BTD cases

guj:
It so turns out in ADIOS, PerformDataWrite (PDW) works better in BTD. So let flatten step use BTD as well. 
chart-84


ax3l:
Hi @guj, for the benchmark you can, can you zip and upload the inputs/submission/output.txt file here?

guj:
[frontier.tar.gz](https://github.com/user-attachments/files/20871013/frontier.tar.gz)

guj:
> Hi @guj, for the benchmark you can, can you zip and upload the inputs/submission/output.txt file here?

uploaded the frontier part.

ax3l (approved):

PR #5968, opened 2025-06-24
SymPy > 1.12.1 yield wrong results in PSATD symbolic notebooks

EZoni:
I recently reused the SymPy notebooks that were added in #3456 and #4316, and I noticed that versions of SymPy higher than 1.12.1 exhibit buggy calculations and yield wrong results (e.g., wrong coefficients for the equations in #4316).

This is quite a puzzling bug, I will try to create a minimal reproducer to report the bug to the SymPy community.

In the meantime, this PR implements the following changes:
- Force `sympy<=1.12.1` in the notebooks, raise exception otherwise.
- Raise exceptions whenever the symbolic solutions cannot be verified.

Future work could include moving most of the SymPy code from the notebooks to reusable Python modules.


Tools/Algorithms/psatd_pml.ipynb
```diff
@@ -276,8 +284,7 @@
     "    diff = X_t.diff(t).diff(t).subs(t, tn).subs(om, c * knorm).expand() - d2X_dt2\n",
     "    diff.simplify()\n",
     "    if diff != zeros(DD, 1):\n",
-    "        print(rf\"Could Not Verify Time Integration for {str(X)}\")\n",
-    "        display(diff)\n",
+    "        raise ValueError(rf\"Could not verify time integration for {str(X)}\")\n",
```
EZoni, line 287: We now fail the CI check with this kind of exceptions, to make sure that the notebook continues to be correct.

ax3l (approved):
Wow! Thanks!

ax3l:
Seems all green, but you self-assigned.

Feel free to merge if ready :) Also ok to set to WIP if it needs more work

PR #5969, opened 2025-06-24
Implement particle suborbits for the implicit solver

dpgrote:
In the implicit solver, a Picard iteration is done for the particle advance. Occasionally, there will be a few particles that fail to converge (for example reaching a bistable state). A solution is to do sub-orbits during the advance, dividing the time step into smaller steps, with the Picard iteration done on each substep. This PR implements this scheme.

The PR is ready for merging accept for adding a CI test. A full test of this would require a full implicit simulation with large time steps (where the need for the sub-orbits arises). A CI test could be added at a later point when the implicit capability is fully implemented.

Merge first:
- [x] #6021
- [x]  #6116
- [x] #6120


JustinRayAngus:
FYI. I have several PRs related to mass matrices that should be merged prior to this PR. Including. PR #6021. After that one and two small follow ons, this will be highest priority.

EZoni:
> FYI. I have several PRs related to mass matrices that should be merged prior to this PR. Including. PR https://github.com/BLAST-WarpX/warpx/pull/6021. After that one and two small follow ons, this will be highest priority.

I edited the PR description with a checklist of PRs that must be merged first. That would be for you to edit/update with the correct PRs. This should make it easier to prevent this PR from being merged earlier by mistake (since the checklist shows up as tasks completed or pending, on GitHub).

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -0,0 +1,1104 @@
...
+#endif
+                             );
+
+        // check if particle did not converge
+        if (max_iterations > 1 && !convergence) {
+
+            if (nsuborbits) {
+                // Suborbits are required for this particle to converge.
+                // It will be handled later in a special loop with suborbiting.
+                nsuborbits[ip] = 2;
```
JustinRayAngus: If I understand this correctly, particles with more than one suborbit from previous Newton iterations will only potentially come here and the suborbits will be reset to 2. If that is true, then the push above should not happen for particles with nsuborbits > 1 and don't reset it to 2.

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -0,0 +1,1104 @@
...
+        amrex::ParticleReal uyp_n = uyp_n0;
+        amrex::ParticleReal uzp_n = uzp_n0;
+
+        amrex::ParticleReal xp = xp_n;
+        amrex::ParticleReal yp = yp_n;
+        amrex::ParticleReal zp = zp_n;
+
+        int isuborbit = 0;
+        bool doing_deposition = false;
+        while (isuborbit < nsuborbits[ip] and !doing_deposition) {
```
JustinRayAngus: I don't understand this logic. The loop will terminate as soon as doing_deposition flips to true. I don't see how a deposition will be performed. I see no reason for the second logic check here.

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -0,0 +1,1104 @@
...
+                const amrex::ParticleReal gamma_new = std::sqrt(1.0_prt + (uxp_new*uxp_new + uyp_new*uyp_new + uzp_new*uzp_new)*inv_c2);
+                const amrex::ParticleReal gaminv = 2.0_prt/(gamma_old + gamma_new);
+#else
+                // unused
+                const amrex::ParticleReal gaminv = 1.;
+#endif
+
+                if (deposit_mass_matrices) {
+                    const amrex::Real wq_invvol = wq*invvol/nsuborbits[ip];
+                    const amrex::Real rhop = wq*invvol*gaminv; // use normal wq here
```
JustinRayAngus: I was incorrect about this. I worked out the math and rhop should also use the weight / nsuborbits.

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -0,0 +1,1104 @@
...
+                                                                  Sxx_arr, Sxy_arr, Sxz_arr,
+                                                                  Syx_arr, Syy_arr, Syz_arr,
+                                                                  Szx_arr, Szy_arr, Szz_arr,
+                                                                  dt_suborbit, dinv, xyzmin, lo );
+                        }
+                    }
+
+                } else {
+
+                    wq /= nsuborbits[ip];
```
JustinRayAngus: This will change the particle weight for each suborbit.

JustinRayAngus (changes requested):
I left several comments. Things are still not working as expected for the 1D/2D planar pinch tests I'm doing.

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -0,0 +1,1104 @@
...
+                if (!convergence) {
+                    // particle did not converge
+                    // Increase the number of suborbits and start over
+                    nsuborbits[ip]++;
+                    isuborbit = 0;
+                } else {
+                    // Convergence was reached for all suborbits, now redo the loop
+                    // and do the deposition
+                    doing_deposition = true;
+                }
```
JustinRayAngus: Shouldn't isuborbit reset to zero in either case?

JustinRayAngus (changes requested):

Source/Particles/Pusher/ImplicitPushPX.cpp
JustinRayAngus: NM. It looks like lines 415-420 above take care of my concern here.

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -0,0 +1,1104 @@
...
+                                           fpxx, fpxy, fpxz,
+                                           fpyx, fpyy, fpyz,
+                                           fpzx, fpzy, fpzz);
+
+                    // The ignore_unused is needed so that the variables are not first-captured
+                    // in a constexpr-if context.
+                    amrex::ignore_unused(full_mass_matrices, max_crossings);
+                    amrex::ignore_unused(Jx_arr, Jy_arr, Jz_arr, invvol);
+                    amrex::ignore_unused(Sxx_arr, Sxy_arr, Sxz_arr, Syx_arr, Syy_arr, Syz_arr, Szx_arr, Szy_arr, Szz_arr);
+                    if constexpr (depos_order_control == one) {
```
JustinRayAngus, line 891: Isn't one a double? This seems to cause issues in the tests I'm doing. I think it should be 1.
`if constexpr (depos_order_control == 1) {`

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -0,0 +1,1104 @@
...
+                                                                  fpyx, fpyy, fpyz,
+                                                                  fpzx, fpzy, fpzz,
+                                                                  Jx_arr, Jy_arr, Jz_arr,
+                                                                  max_crossings,
+                                                                  Sxx_arr, Sxy_arr, Sxz_arr,
+                                                                  Syx_arr, Syy_arr, Syz_arr,
+                                                                  Szx_arr, Szy_arr, Szz_arr,
+                                                                  dt_suborbit, dinv, xyzmin, lo );
+                        }
+                    } else if constexpr (depos_order_control == two) {
```
JustinRayAngus, line 919: Same comment here.
`if constexpr (depos_order_control == 2) {`

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -0,0 +1,1104 @@
...
+                                                                  fpyx, fpyy, fpyz,
+                                                                  fpzx, fpzy, fpzz,
+                                                                  Jx_arr, Jy_arr, Jz_arr,
+                                                                  max_crossings,
+                                                                  Sxx_arr, Sxy_arr, Sxz_arr,
+                                                                  Syx_arr, Syy_arr, Syz_arr,
+                                                                  Szx_arr, Szy_arr, Szz_arr,
+                                                                  dt_suborbit, dinv, xyzmin, lo );
+                        }
+                    } else if constexpr (depos_order_control == three) {
```
JustinRayAngus, line 947: And here.
`if constexpr (depos_order_control == 3) {`

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -0,0 +1,1104 @@
...
+
+                } else {
+
+                    wq /= nsuborbits[ip];
+
+                    // Only CurrentDepositionAlgo::Villasenor is supported
+                    // The ignore_unused is needed so that the variables are not first-captured
+                    // in a constexpr-if context.
+                    amrex::ignore_unused(Jx_arr, Jy_arr, Jz_arr, invvol);
+                    if constexpr (depos_order_control == one) {
```
JustinRayAngus, line 1013: And here.
`if constexpr (depos_order_control == 1) {`

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -0,0 +1,1104 @@
...
+                                                            Jx_arr, Jy_arr, Jz_arr,
+                                                            dt_suborbit, dinv, xyzmin, lo, invvol, n_rz_azimuthal_modes);
+                    }
+                    else if constexpr (depos_order_control == two) {
+                        VillasenorDepositionShapeNKernel<2>(xp_old, yp_old, zp_old, xp_new, yp_new, zp_new, wq,
+                                                            uxp_nph, uyp_nph, uzp_nph, gaminv,
+                                                            Jx_arr, Jy_arr, Jz_arr,
+                                                            dt_suborbit, dinv, xyzmin, lo, invvol, n_rz_azimuthal_modes);
+                    }
+                    else if constexpr (depos_order_control == three) {
```
JustinRayAngus, line 1025: And here.
`if constexpr (depos_order_control == 3) {`

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -0,0 +1,1104 @@
...
+#include <AMReX_Scan.H>
+#include <AMReX_Utility.H>
+
+using namespace amrex::literals;
+
+namespace {
+
+    enum exteb_flags : int { no_exteb, has_exteb };
+    enum qed_flags : int { no_qed, has_qed };
+    enum depos_order_flags : int { one, two, three, four };
```
JustinRayAngus: I believe this translates to
one = 0
two = 1
three = 2
four = 3

Why is this needed? If it is not needed, then just delete this line. If it is needed, then I think it should be
`enum depos_order_flags : int { zero, one, two, three, four };`

Source/Particles/Pusher/ImplicitPushPX.cpp
JustinRayAngus: I think it would be more transparent to use 
`if constexpr (depos_order_control == 1) {`

Is it really necessary to use the enum defined above? (Which has a bug in it that was really hard to find.)

Source/Particles/Pusher/ImplicitPushPX.cpp
JustinRayAngus: Same here.
`if constexpr (depos_order_control == 2) {`

Source/Particles/Pusher/ImplicitPushPX.cpp
JustinRayAngus: And here.
`if constexpr (depos_order_control == 3) {`

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -0,0 +1,1104 @@
...
+                                                                  fpyx, fpyy, fpyz,
+                                                                  fpzx, fpzy, fpzz,
+                                                                  Jx_arr, Jy_arr, Jz_arr,
+                                                                  max_crossings,
+                                                                  Sxx_arr, Sxy_arr, Sxz_arr,
+                                                                  Syx_arr, Syy_arr, Syz_arr,
+                                                                  Szx_arr, Szy_arr, Szz_arr,
+                                                                  dt_suborbit, dinv, xyzmin, lo );
+                        }
+                    } else if constexpr (depos_order_control == four) {
```
JustinRayAngus: `if constexpr (depos_order_control == 4) {`

Source/Particles/Pusher/ImplicitPushPX.cpp
JustinRayAngus: `if constexpr (depos_order_control == 1) {`

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -0,0 +1,1104 @@
...
+                    // The ignore_unused is needed so that the variables are not first-captured
+                    // in a constexpr-if context.
+                    amrex::ignore_unused(Jx_arr, Jy_arr, Jz_arr, invvol);
+                    if constexpr (depos_order_control == one) {
+                        VillasenorDepositionShapeNKernel<1>(xp_old, yp_old, zp_old, xp_new, yp_new, zp_new, wq,
+                                                            uxp_nph, uyp_nph, uzp_nph, gaminv,
+                                                            Jx_arr, Jy_arr, Jz_arr,
+                                                            dt_suborbit, dinv, xyzmin, lo, invvol, n_rz_azimuthal_modes);
+                    }
+                    else if constexpr (depos_order_control == two) {
```
JustinRayAngus: `if constexpr (depos_order_control == 2) {`

Source/Particles/Pusher/ImplicitPushPX.cpp
JustinRayAngus: `if constexpr (depos_order_control == 3) {`

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -0,0 +1,1104 @@
...
+                                                            Jx_arr, Jy_arr, Jz_arr,
+                                                            dt_suborbit, dinv, xyzmin, lo, invvol, n_rz_azimuthal_modes);
+                    }
+                    else if constexpr (depos_order_control == three) {
+                        VillasenorDepositionShapeNKernel<3>(xp_old, yp_old, zp_old, xp_new, yp_new, zp_new, wq,
+                                                            uxp_nph, uyp_nph, uzp_nph, gaminv,
+                                                            Jx_arr, Jy_arr, Jz_arr,
+                                                            dt_suborbit, dinv, xyzmin, lo, invvol, n_rz_azimuthal_modes);
+                    }
+                    else if constexpr (depos_order_control == four) {
```
JustinRayAngus: `if constexpr (depos_order_control == 4) {`

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
```diff
@@ -28,6 +28,10 @@ void ImplicitSolver::CreateParticleAttributes () const
         pc->AddRealComp("ux_n", comm);
         pc->AddRealComp("uy_n", comm);
         pc->AddRealComp("uz_n", comm);
+
+        if (m_particle_suborbits) {
```
JustinRayAngus, line 31: Code crashes here on restart. The easy fix is to check using pc->HasiAttrib("nsuborbits") before adding the new attribute. But I think the better solution would be to make it such that this attribution is not written to the checkpoint file in the first place.

Source/Particles/Pusher/ImplicitPushPX.cpp
dpgrote: This is fixed.

Source/Particles/Pusher/ImplicitPushPX.cpp
dpgrote: Yes. This is fixed.

Source/Particles/Pusher/ImplicitPushPX.cpp
dpgrote: This is fixed. Now one = 1.

Source/Particles/Pusher/ImplicitPushPX.cpp
dpgrote: Fixed, now using a temporary.

Source/Particles/Pusher/ImplicitPushPX.cpp
dpgrote: With isuborbit now set to zero correctly, this will do one more loop.

Source/Particles/Pusher/ImplicitPushPX.cpp
dpgrote: Not sure. This is mirroring the pattern used for the exteb and qed. Presumably it should be fine to use `1` instead of `one` since depos_order_control will be an int (at least on CPU). Note I changed `one` to `order_one` to be more precise.

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
dpgrote: Not sure why the crash. When `nsuborbits` is setup, the `comm` flag is false which means that it won't be communicated and won't be written to checkpoints.

Source/Particles/Pusher/ImplicitPushPX.cpp
JustinRayAngus: Why is the check for !doing_deposition needed here? If I read this line correctly, this loop will terminate when doing_deposition is set to true, even if isuborbit < nsuborbits[ip].

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
JustinRayAngus: Good question. It seems that nsuborbits is still being written to the checkpoint file. Here is the error message:

--- INFO    : restart from checkpoint ./diags/checkpoint024914/
--- INFO    : Runtime int component nsuborbits was found in the checkpoint file,
             but it has not been added yet. Adding it now.
--- INFO    : Runtime int component nsuborbits was found in the checkpoint file,
             but it has not been added yet. Adding it now.
Defined DOF object for linear solves (total DOFs = 756).
  what():  AddIntComp: name 'nsuborbits' is already present in the SoA.
SIGABRT

JustinRayAngus: The problem seems to be an assumption that certain arrays are the same size:

Here is a print of relevant values from FlushFormatCheckpoint::CheckpointParticles():
 pc->NumRealComps() = 9
 pc->h_redistribute_real_comp.size() = 9
 pc->NumIntComps() = 1
 pc->h_redistribute_int_comp.size() = 3

Looks like the code assumes  pc->NumIntComps() =  pc->h_redistribute_int_comp.size()

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -0,0 +1,1105 @@
...
+
+                // Compute inverse Lorentz factor, the average of gamma at time levels n and n+1
+                // The uxp,uyp,uzp are the velocities at time level n+1/2
+                const amrex::ParticleReal gamma_old = std::sqrt(1.0_prt + (uxp_old*uxp_old + uyp_old*uyp_old + uzp_old*uzp_old)*inv_c2);
+                const amrex::ParticleReal gamma_new = std::sqrt(1.0_prt + (uxp_new*uxp_new + uyp_new*uyp_new + uzp_new*uzp_new)*inv_c2);
+                const amrex::ParticleReal gaminv = 2.0_prt/(gamma_old + gamma_new);
+#else
+                // unused
+                const amrex::ParticleReal gaminv = 1.;
+#endif
```
JustinRayAngus: gaminv is needed for mass matrices in 3D

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -126,8 +130,11 @@ namespace {
         bool convergence = false;
         for (int iter=0; iter < max_iterations; iter++) {
 
-            // Position is advanced using the time-centered velocity.
-            // A converged advance will have self-consistent position and velocity.
+            // Position advance starts from the position at the start of the step
+            // and uses the velocity at the start of the step on the first iteration.
```
JustinRayAngus: I think this line in the description is inaccurate. The first iteration does not use the velocity at time t_n.

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -452,18 +490,57 @@ PhysicalParticleContainer::ImplicitPushXP (WarpXParIter& pti,
...
+
     });
 
-    auto const num_unconverged_particles = *(unconverged_particles.copyToHost());
+    amrex::Gpu::synchronize();
+
+    // Setup for handling the unconverged particles. A list of their indices is
+    // gathered, their weights saved, and their weight set to zero (so they
+    // don't contribute to the current density).
+    num_unconverged_particles = *(unconverged_particles.copyToHost());
```
JustinRayAngus: I noticed you added a call to amrex::Gpu::synchronize() here. For my own understanding, why is that needed now when it was not before?

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -473,3 +550,570 @@ PhysicalParticleContainer::ImplicitPushXP (WarpXParIter& pti,
...
+                const amrex::ParticleReal gamma_new = std::sqrt(1.0_prt + (uxp_new*uxp_new + uyp_new*uyp_new + uzp_new*uzp_new)*inv_c2);
+                const amrex::ParticleReal gaminv = 2.0_prt/(gamma_old + gamma_new);
+#else
+                // unused
+                const amrex::ParticleReal gaminv = 1.;
+#endif
+
+                if (deposit_mass_matrices) {
+                    const amrex::Real wq_invvol = wq*invvol/nsuborbits[ip];
+                    const amrex::Real rhop = wq*invvol*gaminv/nsuborbits[ip]; // use normal wq here
```
JustinRayAngus: this line should be 
`const amrex::Real rhop = 2.0_rt*wq_invvol*gaminv;` // approximation when neglecting MM coupling terms

Note use of "wq_invvol" rather than "wq*invvol"

Source/Particles/Pusher/ImplicitPushPX.cpp
dpgrote: The synchronization is probably not needed here, assuming the the previous ParallelFor will complete before the PrefixSum is done. It was removed.

Source/Particles/Pusher/ImplicitPushPX.cpp
dpgrote: Now fixed.

Source/Particles/Pusher/ImplicitPushPX.cpp
dpgrote: We should discuss this. The ux, y, z passed in to this routine are the values at the end of the previous step. What does picnic do at the start of the iterations?

JustinRayAngus: **Without suborbits:**
The values of ux, uy, and uz passed in are the time-centered values at the current iteration m for the field solve, call it ux^(m), etc. I think it's the phrase "start of the step" that is misleading. To me, that means the start of the time step, which here is uxp_n.

**With suborbits:**
What you say is correct, since we don't save the suborbit values, ux, uy, uz at the start is the same as uxp_n, uyp_n, uzp_n from the previous suborbit.

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
JustinRayAngus: I no longer see this issue. No idea what it happened, but it works now so I'm not going to worry about it.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -662,7 +673,54 @@ PhysicalParticleContainer::Evolve (ablastr::fields::MultiFabRegister& fields,
...
+                        amrex::MultiFab * jy = fields.get(current_fp_string, Direction{1}, lev);
+                        amrex::MultiFab * jz = fields.get(current_fp_string, Direction{2}, lev);
+                        long const offset = 0;
+                        ImplicitPushXPSubOrbits(pti, fields, exfab, eyfab, ezfab,
+                                                bxfab, byfab, bzfab,
+                                                implicit_options,
+                                                Ex.nGrowVect(),
+                                                jx, jy, jz,
+                                                offset, lev, gather_lev, dt, ScaleFields(false), skip_deposition,
+                                                num_unconverged_particles_c, unconverged_indices, saved_weights);
```
JustinRayAngus: Bug here.

Should be num_unconverged_particles (without _c)

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -231,21 +238,28 @@ namespace {
...
+ * \param[in] offset the particle index offset for the particles to be pushed
+ * \param[in] np_to_push the number of particles to push
+ * \param[in] lev the refinement level
+ * \param[in] gather_lev the refinement level at which to do the field gather
+ * \param[in] dt the time step size
+ * \param[in] scaleFields allows scale factor to the fields (for rigid injection)
+ * \param[in/out] num_unconverged_particles number of unconverged particles that have already been flagged
+ * \param[in/out] unconverged_indices the list of indices of unconverged particles
+ * \param[in/out] saved_weights the saved weights of the unconverged particles
+ * \param[in] a_dt_type the push type (which part of the time step)
```
JustinRayAngus: You prefer the first word of the description to start with a lower-case letter?

JustinRayAngus:
This is odd. The restart issue is occurring again.

--- INFO    : restart from checkpoint ./diags/checkpoint030873/
--- INFO    : Runtime int component nsuborbits was found in the checkpoint file,
             but it has not been added yet. Adding it now.
--- INFO    : Runtime int component nsuborbits was found in the checkpoint file,
             but it has not been added yet. Adding it now.
Defined DOF object for linear solves (total DOFs = 756).
  what():  AddIntComp: name 'nsuborbits' is already present in the SoA.
  what():  AddIntComp: name 'nsuborbits' is already present in the SoA.
SIGABRT

JustinRayAngus:
> This is odd. The restart issue is occurring again.
> 
> --- INFO : restart from checkpoint ./diags/checkpoint030873/ --- INFO : Runtime int component nsuborbits was found in the checkpoint file, but it has not been added yet. Adding it now. --- INFO : Runtime int component nsuborbits was found in the checkpoint file, but it has not been added yet. Adding it now. Defined DOF object for linear solves (total DOFs = 756). terminate called after throwing an instance of 'std::runtime_error' what(): AddIntComp: name 'nsuborbits' is already present in the SoA. terminate called after throwing an instance of 'std::runtime_error' what(): AddIntComp: name 'nsuborbits' is already present in the SoA. SIGABRT

I gave you an input deck on dane. Run it for 100 steps, then do a restart.

JustinRayAngus:
> > This is odd. The restart issue is occurring again.
> > --- INFO : restart from checkpoint ./diags/checkpoint030873/ --- INFO : Runtime int component nsuborbits was found in the checkpoint file, but it has not been added yet. Adding it now. --- INFO : Runtime int component nsuborbits was found in the checkpoint file, but it has not been added yet. Adding it now. Defined DOF object for linear solves (total DOFs = 756). terminate called after throwing an instance of 'std::runtime_error' what(): AddIntComp: name 'nsuborbits' is already present in the SoA. terminate called after throwing an instance of 'std::runtime_error' what(): AddIntComp: name 'nsuborbits' is already present in the SoA. SIGABRT
> 
> I gave you an input deck on dane. Run it for 100 steps, then do a restart.

And here is the checkpoint Header:Version_Two_Dot_One_double
1
4
weight
momentum_x
momentum_y
momentum_z
1
nsuborbits
1

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
JustinRayAngus: I don't know why you didn't see this in your test. My guess is maybe the particle_suborbit flag in the input was not being parsed correctly (typo?). Here is what is in the checkpoint Header file

And here is the checkpoint Header:
Version_Two_Dot_One_double
1
4
weight
momentum_x
momentum_y
momentum_z
1
nsuborbits
1

JustinRayAngus: Being addressed with PR #6130

JustinRayAngus (approved):
Thanks Dave. This looks good.

EZoni:
@dpgrote 

#6136 fixed the CI macOS issue, please feel free to rebase on `development` to clear the CI checks here.

PR #5972, opened 2025-06-24
CI: improve paths exclusion filters

EZoni:
This improves the paths exclusion filters that we use in our CI workflows, from GitHub Actions to Azure pipelines:
- Anything inside `Docs/`
- Anything inside `Tools/machines/`
- Anything ending in `.rst`
- Anything ending in `.md`


ax3l (approved):
Ty! :)

PR #5973, opened 2025-06-24
CI: fix Open MPI bug in Azure pipelines

EZoni:
### Goal

Resolve the following error seen in our Azure DevOps pipelines:
```
There are not enough slots available in the system to satisfy the 2
slots that were requested by the application:

  /home/vsts/work/1/s/build/bin/warpx.3d.MPI.OMP.DP.PDP.OPMD.FFT.EB.QED

Either request fewer slots for your application, or make more slots
available for use.

A "slot" is the Open MPI term for an allocatable unit where we can
launch a process.  The number of slots available are defined by the
environment in which Open MPI processes are run:

  1. Hostfile, via "slots=N" clauses (N defaults to number of
     processor cores if not provided)
  2. The --host command line parameter, via a ":N" suffix on the
     hostname (N defaults to 1 if not provided)
  3. Resource manager (e.g., SLURM, PBS/Torque, LSF, etc.)
  4. If none of a hostfile, the --host command line parameter, or an
     RM is present, Open MPI defaults to the number of processor cores

In all the above cases, if you want Open MPI to default to the number
of hardware threads instead of the number of processor cores, use the
--use-hwthread-cpus option.

Alternatively, you can use the --oversubscribe option to ignore the
number of available slots when deciding the number of processes to
launch.
```

### Workaround

The workaround consists in adding, say, `--host localhost:2` to the mpirun/mpiexec command when running with 2 processes, following the suggestion of point 2. in the error message above. Please check the comments below to see how I have debugged this so far.

### Open questions

- Why doesn't point 4. above ("Open MPI defaults to the number of processor cores") apply here?
- Do you think this is an Open MPI 4.1.6 issue or an Azure DevOps pipeline environment issue?

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -157,6 +157,8 @@ jobs:
       echo "mpirun --do-not-launch --display-allocation:"
       warpx_exec=$(find /home/vsts/work/1/s/build/bin -type f -name 'warpx.*' -executable | head -n 1)
       mpirun --do-not-launch --display-allocation ${warpx_exec}
+      echo "Open MPI environment variables:"
+      env | grep "OMPI"
       # TODO remove debugging block............................................
       # set options
       set -o nounset errexit pipefail
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -154,8 +154,9 @@ jobs:
       echo "number of processing units: $(nproc)"
       echo "mpirun --version:"
       mpirun --version
-      echo "mpirun --display-allocation:"
-      mpirun --display-allocation
+      echo "mpirun --do-not-launch --display-allocation:"
+      warpx_exec=$(find /home/vsts/work/1/s/build/bin -type f -name 'warpx.*' -executable | head -n 1)
+      mpirun --do-not-launch --display-allocation ${warpx_exec}
       # TODO remove debugging block............................................
       # set options
       set -o nounset errexit pipefail
@@ -167,7 +168,7 @@ jobs:
           -D ExperimentalTest -D ExperimentalSubmit
       else
         # run tests (exclude pytest.AMReX when running Python tests)
-        ctest --test-dir build --output-on-failure -E AMReX
+        ctest --test-dir build --verbose -E AMReX
       fi
     displayName: 'Test'
 
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -150,16 +150,6 @@ jobs:
     displayName: 'Build'
 
   - bash: |
-      # TODO remove debugging block............................................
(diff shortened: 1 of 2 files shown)


EZoni:
This 
```bash
echo "number of processing units: $(nproc)"
echo "mpirun --version:"
mpirun --version
echo "mpirun --display-allocation:"
mpirun --display-allocation
```
gives this output:
```
number of processing units: 2
mpirun --version:
mpirun (Open MPI) 4.1.6

Report bugs to http://www.open-mpi.org/community/help/
mpirun --display-allocation:
--------------------------------------------------------------------------
No executable was specified on the mpirun command line.

Aborting.
--------------------------------------------------------------------------
```

so `echo $(nproc)` returns 2.

EZoni:
This (like the above, but corrected)
```bash
echo "number of processing units: $(nproc)"
echo "mpirun --version:"
mpirun --version
echo "mpirun --do-not-launch --display-allocation:"
warpx_exec=$(find /home/vsts/work/1/s/build/bin -type f -name 'warpx.*' -executable | head -n 1)
mpirun --do-not-launch --display-allocation ${warpx_exec}
```
gives this output:
```
number of processing units: 2
mpirun --version:
mpirun (Open MPI) 4.1.6

Report bugs to http://www.open-mpi.org/community/help/
mpirun --do-not-launch --display-allocation:
 Data for JOB [27373,1] offset 0 Total slots allocated 1

 ========================   JOB MAP   ========================

 Data for node: pkrvm2xkesd8i0z	Num slots: 1	Max slots: 0	Num procs: 1
 	Process OMPI jobid: [27373,1] App: 0 Process rank: 0 Bound: socket 0[core 0[hwt 0]]:[B/.]

 =============================================================
 ```

so `$(nproc)` returns 2, but the job map shows `Num slots: 1` and `Num procs: 1`.

EZoni:
Adding an `env | grep "OMPI"` to check any of the Open MPI environment variables listed at https://docs.open-mpi.org/en/v5.0.x/tuning-apps/environment-var.html... (although we're running Open MPI 4.1.6): no `OMPI` environment variables detected.

EZoni:
Adding `--host localhost:${nprocs}` to the mpirun/mpiexec command seems to resolve the issue.

roelof-groenewald (approved):
I vote we merge this as it unblocks open PRs. We can always revert this later when we find the root cause of the problem.

EZoni:
Thanks for the feedback, @roelof-groenewald. Okay, I will merge now to unblock the other PRs, but I will still invite feedback from the reviewers and possibly open a follow-up PR if a different bug fix turns out to be recommended.

PR #5974, opened 2025-06-24
RZ: GPU-Side Abort in FDTD

ax3l:
The FDTD code for collocated solves in RZ uses a host-side abort in device code. This breaks compiles on ROCm GPUs.


roelof-groenewald (approved):
Thanks for fixing! Actually just ran into this issue when trying to build with SYCL.

clarkse (commented):
Looks good to me. Should we unify how we handle aborts? Perhaps we can adjust the macro to work in each case rather than having a different abort/messaging interface for device vs. host code?

ax3l:
Yes, possibly! But to be honest, I think we should try to keep abort logic inside host-side code. Especially code like this here can be determined before starting a kernel.

PR #5979, opened 2025-06-25
Aurora setup

zippylab:
Initial configuration/dependency scripts and example job script for the Aurora system at ALCF, including associated documentation. I didn't create an issue for this, but will do so if there are other tasks going forward.


Docs/source/install/hpc/aurora.rst
```diff
@@ -0,0 +1,161 @@
...
+
+
+Introduction
+------------
+
+If you are new to this system, **please see the following resources**:
+
+* `ALCF user guide <https://docs.alcf.anl.gov/>`_
+* Batch system: `PBS <https://docs.alcf.anl.gov/running-jobs/>`_
+* `Filesystems <https://docs.alcf.anl.gov/data-management/filesystem-and-storage/>`_:
```
ax3l: Minor warning suppression: for named links in Sphinx it creates an internal reference and to suppress warnings we can use this to anonymize them (e.g., if the label `Filesystems` is used for more than one link throughout the whole documentation):
```suggestion
* `ALCF user guide <https://docs.alcf.anl.gov/>`__
* Batch system: `PBS <https://docs.alcf.anl.gov/running-jobs/>`__
* `Filesystems <https://docs.alcf.anl.gov/data-management/filesystem-and-storage/>`__:
```

ax3l (commented):
Thank you so much for the PR and following our existing doc style so carefully!

I added two colleagues with Aurora/ALCF access as well for review.

Tools/machines/aurora-alcf/aurora_warpx.profile.example
```diff
@@ -0,0 +1,29 @@
+# Set the project name
+export proj=""  # change me!
+
+# required dependencies
+module load cmake
+module load adios2/2.10.2-cpu
+module load hdf5/1.14.5
```
roelof-groenewald: Loading `hdf5` is repeated on line 13 below.
```suggestion
```

Tools/machines/aurora-alcf/aurora_warpx.profile.example
```diff
@@ -0,0 +1,29 @@
+# Set the project name
+export proj=""  # change me!
+
+# required dependencies
+module load cmake
+module load adios2/2.10.2-cpu
```
roelof-groenewald, line 6: Out of curiosity; I noticed there are two modules available for `adios2`, a "cpu" and "sycl" version. Why load the "cpu" version rather than the default?

Tools/machines/aurora-alcf/install_dependencies.sh
```diff
@@ -0,0 +1,92 @@
...
+python3 -m pip install --upgrade build
+python3 -m pip install --upgrade packaging
+python3 -m pip install --upgrade wheel
+python3 -m pip install --upgrade setuptools[core]
+python3 -m pip install --upgrade "cython>=3.0"
+python3 -m pip install --upgrade numpy
+python3 -m pip install --upgrade h5py
+python3 -m pip install --upgrade pandas
+python3 -m pip install --upgrade scipy
+# Is next line OK/needed for Aurora?
```
roelof-groenewald, line 81: That's a good question. I guess it depends on whether the `mpi4py` installation that comes from the standard `frameworks` Python was linked to a the same `mpicc` compiler with gpu-aware-mpi flags turned on. Then we could just specify `--system-site-packages` during the creation of the virtual environment, right? In which case the system `mpi4py` will be used.

We can leave it as is for now and do some tests over the next few weeks to see if one way or the other is preferable.

roelof-groenewald (commented):
Thank you very much for setting up these instructions, Tim!
I just have a question about the ADIOS2 module.

Tools/machines/aurora-alcf/aurora_warpx.profile.example
ax3l: Just a note: we do not (yet) use any GPU features of ADIOS2 in openPMD-api and thus WarpX. So it might not matter which module we pick, as long as both are compatible (i.e. built with the same compiler & mpi modules).

Tools/machines/aurora-alcf/install_dependencies.sh
ax3l: I am not familiar with the `frameworks` module, but if you use the default `cray-python` module on HPE systems, then it comes with numpy, scipy and mpi4py pre-built. But you can still re-compile and upgrade `mpi4py` of course, if you like.

e.g. https://cpe.ext.hpe.com/docs/24.11/release_notes/index.html

Tools/machines/aurora-alcf/aurora_warpx.profile.example
roelof-groenewald: That makes sense. Thanks @ax3l!

roelof-groenewald (approved):
Thanks again, Tim!

I'll merge this is now and we can adjust as needed with follow-up PRs when we get more experience running on Aurora.

Tools/machines/aurora-alcf/aurora_warpx.profile.example
zippylab: If WarpX is using ADIOS2 to initiate I/O (or other connections) directly from the GPUs, I believe then it would need the "sycl" version.

Tools/machines/aurora-alcf/install_dependencies.sh
zippylab: I believe the `mpi4py` via the `frameworks` module is the most likely one to be optimized for Aurora MPICH. Generally speaking, the `frameworks` module is the best starting point for Python and AI functionalities on Aurora. The ALCF data science team works with that.

PR #5981, opened 2025-06-26
Further optimizations for the binary collision module

atmyers:
This PR implements two optimizations, both of which contribute to improving the runtime of the binary collisions module. The first optimization is to use atomics to compute the density and temperature in each cell, similar to what we do in the current deposition. The second is to expose more parallelism in the step that does the Fisher-Yates shuffle.

The following inputs file:

```
max_step = 100
amr.n_cell= 128 128 128
amr.max_level = 0
amr.max_grid_size = 10000
geometry.dims=2
geometry.prob_lo = -20.e-6   -20.e-6   -20.e-6    # physical domain
geometry.prob_hi =  20.e-6    20.e-6    20.e-6

####### Boundary condition ######
boundary.field_lo = periodic periodic periodic
boundary.field_hi = periodic periodic periodic

warpx.verbose = 1
warpx.cfl = 1.0

# Order of particle shape factors
algo.particle_shape = 1

particles.species_names = electrons ions

electrons.species_type = electron
electrons.injection_style = "nrandompercell"
electrons.num_particles_per_cell=512
electrons.profile = constant
electrons.density = 1.e25  # number of electrons per m^3
electrons.ux_th  = 0.01 # uth the std of the (unitless) momentum
electrons.uy_th  = 0.01 # uth the std of the (unitless) momentum
electrons.uz_th  = 0.01 # uth the std of the (unitless) momentum

ions.species_type = hydrogen
ions.injection_style = "nrandompercell"
ions.num_particles_per_cell=512
ions.profile = constant
ions.density = 1.e25  # number of electrons per m^3
ions.momentum_distribution_type = "gaussian"
ions.ux_th  = 0.0001 # uth the std of the (unitless) momentum
ions.uy_th  = 0.0001 # uth the std of the (unitless) momentum
ions.uz_th  = 0.0001 # uth the std of the (unitless) momentum

# Collisions
collisions.collision_names = cei 
cei.species = electrons ions
cei.CoulombLog = 10.0
```

running on 1 MI300, gives:

```
MultiParticleContainer::doCollisions() 100   0.8964   0.8964   0.8964  30.60%
```

on dev and 

```
MultiParticleContainer::doCollisions() 100   0.6717   0.6717   0.6717  24.16%
```

on this branch.


Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -676,9 +676,10 @@ public:
 
             // Loop over cells
             WARPX_PROFILE_VAR("BinaryCollision::doCollisionsWithinTile::findDensityTemperatures", prof_findDensityTemperatures);
-            amrex::ParallelForRNG( n_cells,
-                [=] AMREX_GPU_DEVICE (int i_cell, amrex::RandomEngine const& engine) noexcept
+            amrex::ParallelForRNG( 2*n_cells,
+                [=] AMREX_GPU_DEVICE (int i, amrex::RandomEngine const& engine) noexcept
                 {
+            int i_cell = i/2;
```
AlexanderSinn:
```suggestion
            int i_cell = i < n_cells ? i : i - n_cells;
```

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -676,9 +676,10 @@ public:
 
             // Loop over cells
             WARPX_PROFILE_VAR("BinaryCollision::doCollisionsWithinTile::findDensityTemperatures", prof_findDensityTemperatures);
-            amrex::ParallelForRNG( n_cells,
-                [=] AMREX_GPU_DEVICE (int i_cell, amrex::RandomEngine const& engine) noexcept
+            amrex::ParallelForRNG( 2*n_cells,
+                [=] AMREX_GPU_DEVICE (int i, amrex::RandomEngine const& engine) noexcept
                 {
+            int i_cell = i < n_cells ? i : i - n_cells;
```
atmyers:
```suggestion
            int i_cell = n_cells / 2;
```

atmyers: This actually seems to be somewhat slower than the original ordering (0.6781s versus 0.5149s for the doCollisions() module).

AlexanderSinn: Doesn't this have to match with the `if (i < n_cells) {` checks? I would think `int i_cell = n_cells / 2;` produces incorrect results.

ax3l:
@atmyers do you like to post the theory plots (which match well) and reset the change sums?

clarkse:
How this is speeding up execution? I thought when doing branched execution like this in CUDA that the branches are executed sequentially because it is a SIMT execution. Is it along the lines of when all threads in a particular warp are one species or another then they can all coalesce into one of the execution branches and avoid one extra function call? I could use a few more comments inline to understand better what is happening in this code change.

atmyers:
> How this is speeding up execution? I thought when doing branched execution like this in CUDA that the branches are executed sequentially because it is a SIMT execution. Is it along the lines of when all threads in a particular warp are one species or another then they can all coalesce into one of the execution branches and avoid one extra function call? I could use a few more comments inline to understand better what is happening in this code change.

I think that in this case, the thread divergence is not too bad because most of the time threads in the same warp will take the same path. Adjacent threads are processing cells for the same species. I think the reason this speeds things up is that one thread per cell is not enough to keep the GPU busy for this kernel, so it's better to launch 2 threads per cell and process the two species simultaneously. I've tried to add some comments explaining this better.

atmyers:
The tests are currently failing, but I think this is just a consequence of the random number sequence being different. For example, the test_3d_ionization_electron_dsmc problem gives:

ionization_dsmc_density_Te

using the current code.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -348,9 +348,11 @@ public:
             // Extract low-level data
             auto const n_cells = static_cast<int>(bins_1.numBins());
             // - Species 1
+            auto np1 = ptile_1.numParticles();
             const auto soa_1 = ptile_1.getParticleTileData();
             index_type* AMREX_RESTRICT indices_1 = bins_1.permutationPtr();
             index_type const* AMREX_RESTRICT cell_offsets_1 = bins_1.offsetsPtr();
+            index_type const* AMREX_RESTRICT bins_1_ptr = bins_1.binsPtr();
```
clarkse, line 355: General comment on the ParallelFor loops below. Shouldn't we put a streamSynchronize after each one that computes a dependency? E.g. the calculation of p_n_pairs_in_each_cell comes after a ParallelFor loop that executes asynchronously.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -462,15 +466,63 @@ public:
...
+                        amrex::Gpu::Atomic::AddNoRet(&vz1_in_each_cell[bins_1_ptr[ip]], w1[ip]*u1z[ip]/gm);
+                        amrex::Gpu::Atomic::AddNoRet(&vs1_in_each_cell[bins_1_ptr[ip]], w1[ip]*us/gm/gm);
                     }
+                }
+            );
+            WARPX_PROFILE_VAR_STOP(prof_findDensityTemperatures_atomics);
+
+            // Finish temperature calculation
+        WARPX_PROFILE_VAR("BinaryCollision::doCollisionsWithinTile::findDensityTemperatures::finishTemperature", prof_findDensityTemperatures_finish);
+            amrex::ParallelFor( n_cells, [=] AMREX_GPU_DEVICE (int i_cell) noexcept
```
clarkse, line 484: Perhaps this can be addressed in a follow up PR, but PR#5811 and PR#6001 are adding species based component-wise temperature deposition with the particle shape functions in preparation for a hybrid ohm's law energy equation including interspecies temperature relaxation. I am planning to do some cleanup on it soon after #6001 is merged to integrate the diagnostics calculations and perhaps this should also be integrated. I can follow up with you on this as I will be digging around in the collisions area next and this could possibly be refactored somewhat.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -462,15 +466,63 @@ public:
...
+                        amrex::Gpu::Atomic::AddNoRet(&T1_in_each_cell[bins_1_ptr[ip]],  w1[ip]);
+                        amrex::Gpu::Atomic::AddNoRet(&vx1_in_each_cell[bins_1_ptr[ip]], w1[ip]*u1x[ip]/gm);
+                        amrex::Gpu::Atomic::AddNoRet(&vy1_in_each_cell[bins_1_ptr[ip]], w1[ip]*u1y[ip]/gm);
+                        amrex::Gpu::Atomic::AddNoRet(&vz1_in_each_cell[bins_1_ptr[ip]], w1[ip]*u1z[ip]/gm);
+                        amrex::Gpu::Atomic::AddNoRet(&vs1_in_each_cell[bins_1_ptr[ip]], w1[ip]*us/gm/gm);
                     }
+                }
+            );
+            WARPX_PROFILE_VAR_STOP(prof_findDensityTemperatures_atomics);
+
```
clarkse, line 481: streamSynchronize here?

clarkse (commented):
I think this looks reasonable for speeding this up. I do think we can streamline some of the moment calculations since this object shouldn't need to internally compute T, but as an existing optimization I think it is fine. I would just double check the results you posted against the same thing in development and overlay the curves.

atmyers:
> I would just double check the results you posted against the same thing in development and overlay the curves.

It's difficult to see the difference from the theory plots, but here are the relative errors for dev:

relative_error_density_Te

and this branch:

relative_error_density_Te

They look statistically the same to me. To get the analysis scripts to pass, I just had to bump the tolerance for this test and the proton-boron fusion one slightly. The remaining failures are all checksum-related and I think just due to the different random number sequence.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
atmyers: I agree it would be good to integrate the various temperature calculations as a follow-on.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
atmyers: Because the ParallelFors in `doCollisionsWithinTile` are all launched on the same GPU stream, it is guaranteed that that they will execute sequentially, without streamSynchronize calls.

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
atmyers: See comment above - I don't think we need to explicitly sync here.

dpgrote (approved):
Looks good to me! Thanks!

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
dpgrote, line 355: @atmyers This `bins_1_ptr` is very useful. I can use it in my PR #5932. Also, could this be used in the loop over independent pairs instead of the `bisect`?

Source/Particles/Collision/BinaryCollision/BinaryCollision.H
```diff
@@ -462,15 +466,63 @@ public:
...
+                    // The particles from species1 that are in the cell `i_cell` are
+                    // given by the `indices_1[cell_start_1:cell_stop_1]`
+                    index_type const cell_start_1 = cell_offsets_1[i_cell];
+                    index_type const cell_stop_1  = cell_offsets_1[i_cell+1];
+
+                    // Do not need if there is only one particle in the cell
+                    if ( cell_stop_1 - cell_start_1 <= 1 ) { return; }
+
+                    // finish temperature calculation if needed
+                    if (binary_collision_functor.m_computeSpeciesTemperatures) {
```
roelof-groenewald, line 495: Seeing as this kernel is only doing computation when `binary_collision_functor.m_computeSpeciesTemperatures == true`, the `if` statement could be put around the kernel as a further (small) performance improvement for non-Coulomb binary collisions. @atmyers do you think that is worth a follow-up PR?

roelof-groenewald:
I believe with this PR merged `ComputeTemperature.H` is not used anymore, right? Should I open a PR to remove it?
(https://github.com/BLAST-WarpX/warpx/blob/development/Source/Particles/Collision/BinaryCollision/Coulomb/ComputeTemperature.H)

atmyers:
> I believe with this PR merged `ComputeTemperature.H` is not used anymore, right? Should I open a PR to remove it? (https://github.com/BLAST-WarpX/warpx/blob/development/Source/Particles/Collision/BinaryCollision/Coulomb/ComputeTemperature.H)

Good point - I went ahead and opened one here: https://github.com/BLAST-WarpX/warpx/pull/6015

PR #5982, opened 2025-06-26
Tuolumne (LLNL): AMReX Memory Release

ax3l:
See https://github.com/AMReX-Codes/amrex/pull/4531#issuecomment-3014937899 and  https://github.com/AMReX-Codes/amrex/pull/4527

### DIFF
--- a/Docs/source/install/hpc/tuolumne.rst
+++ b/Docs/source/install/hpc/tuolumne.rst
@@ -201,6 +201,7 @@ Note that we append these non-default runtime options:
 
 * ``amrex.use_gpu_aware_mpi=1``: make use of fast APU to APU MPI communications
 * ``amrex.the_arena_init_size=1``: avoid overallocating memory that is *shared* on APUs between CPU & GPU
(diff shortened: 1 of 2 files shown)


ax3l:
Downside: this makes the code 2.67x slower :-o

ax3l:
Need to redo experiments with:
- pre-allocate a sensible number of the device memory with AMReX, e.g., 96GB of the 128GB per GPU
- somehow change the `amrex.the_arena_release_threshold` to a sensible number (?)

@WeiqunZhang ?

ax3l:
On a single-node (4 GPU) test, `amrex.the_arena_defragmentation=0` together with `amrex.the_arena_init_size=1` seems to also keep the memory growth at bay :)
Have not yet checked performance impact and memory overhead at scale.

update: see https://github.com/AMReX-Codes/amrex/pull/4531#issuecomment-3014937899

ax3l:
@WeiqunZhang probably a sensible setting for now. Maybe could even bump the default-allocation up so AMReX never re-allocates until it actually runs out of memory...

WeiqunZhang (approved):

PR #4760, opened 2024-03-09
Zenodo: Add Justin & Kale as Co-Authors

ax3l:
Adding @JustinRayAngus and @kale-j as a co-authors of WarpX. Thank you for your contributions! :sparkles:

Please help me to double-check your ORCiDs and spellings (e.g., inclusion of middle names, affiliations).

PR #5985, opened 2025-06-29
Fix moving-window subvolume field output

bnara:
At initialization of moving-window simulations, the diagnostic subvolume  is displaced along the direction of the moving window.  This is done to account for restarted simulations.  However, the displacement is calculated based on the trailing edge of the diagnostic subvolume rather than the simulation volume.  This causes a displacement error when the trailing edges of the simulation volume and the diagnostic subvolume do not coincide.

For example, I noticed this issue when I was doing a beam-plasma study with a moving-window along *z*.   When I tried to take an *xy* slice in the middle of the simulation volume,  the resulting *xy* slice was from the trailing edge of the simulation volume rather than the middle of the simulation volume.

This PR corrects the displacement calculation.


EZoni (commented):
Thanks for the PR, @bnara!

Would it be possible for you to post two plots that show the behavior before and after this change, just for more clarity?

bnara:
**Minimal example**

Taking the positive *z*-direction to be rightward and the positive *y*-direction to be upward, consider the following beam moving along the positive *z*-axis,

Periodically throughout a moving window simulation, I would like to record slices in the *yz*-plane, as well as several slices parallel to the *xy*-plane.  Currently, due to the error in the displacement calculation, WarpX places all of the *xy* slices at the **trailing edge** of the simulation volume.

**WarpX input file:**  [warpx.txt](https://github.com/user-attachments/files/21023194/warpx.txt)


- Moving-window simulation
- Creates truncated ellipsoidal beam
- Requests output of one *yz* slice and two *xy* slices: one at zeta=0um and another at zeta=-80um

**PR-5985**
Correctly outputs the two *xy* slices at their requested locations in the moving window.  


**warpx:development**
Incorrectly outputs both *xy* slices at the trailing edge of the simulation volume.

EZoni (approved):
@RemiLehe  @jlvay  @dpgrote 

Based on the description and the example provided here, this looks like a correct bug fix to me. Do you have any comments, objections, etc.?

EZoni:
@bnara 

Thanks again for the recent bug fixes! Would you like to open a PR to add yourself to the list of contributors that we have in https://github.com/BLAST-WarpX/warpx/blob/development/.zenodo.json? See, for example, PR [#4760](https://github.com/BLAST-WarpX/warpx/pull/4760).

bnara:
@EZoni
Thank you for the suggestion!  Yes, I will open a PR for the contributor list after I get a significant feature merged.  Fingers crossed for my `twiss` injection source (#5426), which is what we are using for our beam studies at UCLA.  In any case, thanks to you and the other developers for providing a quality codebase that is a pleasure to work with.

PR #5986, opened 2025-06-30
Allow rigid free-drift in transverse coordinates

bnara:
For `rigid_advance` = `false`, macroparticles are rigidly advanced each with their own velocity `vz`, ignoring any transverse velocity, until reaching `zinject_plane`.  In our beam-plasma studies at UCLA, we match a strongly focusing electron beam into a longitudinally-ramped plasma profile.  For a correct description, we need a true "rigid free-drift" in all spatial dimensions, including transverse.

 The following movie illustrates the difference (`zinject_plane` = 0),
<video src="https://github.com/user-attachments/assets/dfb6c5e0-5089-4433-b6e7-2f2b51bdbf4f" loop controls muted title="PR-RigidFreeDrift.mp4"></video>

- **Top:** Current `warpx:development` behavior
- **Bottom:** Proposed "rigid free-drift" behavior

This PR implements the correct "rigid free-drift" behavior for `rigid_advance` = `false`. 

If the current behavior is actually desired, I can also promote `rigid_advance` from a bool to an enum with three separate modes, preserving backward compatibility while clarifying semantics.

**Input used for movie:** [warpx.txt](https://github.com/user-attachments/files/20972363/warpx.txt) (also used the Twiss injection source from PR #5426)


dpgrote:
This PR looks good, thanks. Though, it should preserve the existing behavior. I would suggest a new input parameter to switch on and off the transverse part of the drift, perhaps `<species_name>.rigid_transverse_drift`, defaulting to false.

dpgrote (approved):
This looks good, thanks!

bnara:
Thanks!

If `rigid_advance` is a bool (`false`, `true`, `0`, or `1`), then the behavior is backward compatible.   Might want to eventually deprecate bool input.

PR #5992, opened 2025-07-02
Splitting: Remove `pc_tmp`

ax3l:
There was an other temporary pc that makes refactoring init logic hard (e.g., #5015)

Follow-up to #5571

- [x] remove `pc_tmp`
- [x] finalize particle splitting (`AddNParticles`) cc @atmyers @EZoni

### DIFF
--- a/Source/Particles/PhysicalParticleContainer.cpp
+++ b/Source/Particles/PhysicalParticleContainer.cpp
@@ -2323,8 +2323,7 @@ PhysicalParticleContainer::applyNCIFilter (
 void
 PhysicalParticleContainer::SplitParticles (int lev)
 {
-    auto& mypc = WarpX::GetInstance().GetPartContainer().GetParticleContainerFromName(this->species_name);
-    auto pctmp_split = mypc.make_alike<amrex::DefaultAllocator>();
+    PhysicalParticleContainer pctmp_split(&WarpX::GetInstance());
     RealVector psplit_x, psplit_y, psplit_z, psplit_w;
     RealVector psplit_ux, psplit_uy, psplit_uz;
     long np_split_to_add = 0;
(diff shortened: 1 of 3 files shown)


Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -2507,7 +2508,7 @@ PhysicalParticleContainer::SplitParticles (int lev)
     amrex::Vector<amrex::Vector<ParticleReal>> attr;
     attr.push_back(wp);
     const amrex::Vector<amrex::Vector<int>> attr_int;
-    pctmp_split.AddNParticles(lev,
+    pctmp_split.AddNParticles(lev,   // TODO
```
ax3l: @atmyers @EZoni can you please finalize this? I think using the appropriate particle add function here is all that is needed to complete this.

atmyers: See if you like this way of doing things. I think it makes sense to have this be a `PhysicalParticleContainer`, so we can just call the same `AddNParticles` method.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -224,6 +224,7 @@ PhysicalParticleContainer::PhysicalParticleContainer (AmrCore* amr_core, int isp
     : WarpXParticleContainer(amr_core, ispecies),
       species_name(name)
 {
+    amrex::Print() << "PhysicalParticleContainer::PhysicalParticleContainer\n";
```
atmyers:
```suggestion
```

Source/Particles/PhysicalParticleContainer.cpp
ax3l: Ah, that makes sense! Good idea.
```suggestion
    pctmp_split.AddNParticles(lev,
```

EZoni (approved):

ax3l:
Thanks Andrew :pray:

PR #5965, opened 2025-06-21
Tools: simplify workflows for commit, version updates

EZoni:
### Main Features
- Central configuration file (JSON) that stores version/commit information for WarpX, AMReX, pyAMReX, and PICSAR:
    ```json
    {
        "version_warpx": "25.06",
        "version_amrex": "25.06",
        "version_pyamrex": "25.06",
        "version_picsar": "25.04",
        "commit_amrex": "e59329031fdad45285ebe04fb7101e62f77a49d9",
        "commit_pyamrex": "c9cec5c0743af5f14d9ed674e4e80ab9fcf153e3",
        "commit_picsar": "0c329e66010267662a82219f7de7abbd231463f4"
    }
    ```
- CMake can parse the JSON file directly, Python can parse the JSON file through the package `json`, bash (inside a YAML file, e.g., the CUDA workflow file) can parse the JSON file through the package `jq`. No need to edit individual files anymore.
- The commit/version update scripts are greatly simplified and become more accessible to all developers:
  - They update only the JSON file.
  - They are merged into one shorter script.
  - They do not require complicated regex searches and edits.
- The three commit/version update scripts `updateAMReX.py`, `updatepyAMReX.py`, and `updatePICSAR.py`, in total ~400 lines of code, are replaced by the new script `update_dependencies.py`, in total ~100 lines of code, with this command line interface:
```
$ ./Tools/Release/update_dependencies.py --help
usage: update_dependencies.py [-h] [--amrex] [--pyamrex] [--picsar] [--warpx]

options:
  -h, --help  show this help message and exit
  --amrex     Update AMReX only
  --pyamrex   Update pyAMReX only
  --picsar    Update PICSAR only
  --warpx     Update WarpX only
```
- Example of output:
```
$ ./Tools/Release/update_dependencies.py

Updating AMReX...
- old commit: a21af25a11cafa1c61e7101b5b24112ecd8d0b44
- new commit: 0108935f6c3e0b153e86c1114caeacfbdc962717
Do you want to continue? [y/n] y
Updating commit...
- old version: 25.06
- new version: 25.06
Do you want to continue? [y/n] y
Updating version...

Updating pyAMReX...
- old commit: 5042df9cd6cbb9d0f0952fd47dd238357321e010
- new commit: 5042df9cd6cbb9d0f0952fd47dd238357321e010
Do you want to continue? [y/n] n
Skipping commit update...
- old version: 25.06
- new version: 25.06
Do you want to continue? [y/n] n
Skipping version update...

[...]

PR #5993, opened 2025-07-02
CMake: Fix Superbuilds

ax3l:
Use WarpX-specific and not CMake-superbuild-specific directory. Follow-up to #5965

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 # Parse WarpX version information
-file(READ "${CMAKE_SOURCE_DIR}/dependencies.json" dependencies_data)
+file(READ "${CMAKE_CURRENT_LIST_DIR}/dependencies.json" dependencies_data)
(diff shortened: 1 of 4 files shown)


EZoni:
Oh cool, thanks for catching that!

Do we need to do the same fix also in these other files?
- https://github.com/BLAST-WarpX/warpx/blob/a9ec6c30a8c7537f156cdf2dc2d3590a3cb549f3/cmake/dependencies/pyAMReX.cmake#L78-L81
- https://github.com/BLAST-WarpX/warpx/blob/a9ec6c30a8c7537f156cdf2dc2d3590a3cb549f3/cmake/dependencies/PICSAR.cmake#L113-L116

Maybe it's not "necessary", but would that be more robust?

EZoni (approved):

ax3l:
You are welcome!

Yes we do, I pushed this a minute ago.

PR #5994, opened 2025-07-02
Reduce kernel sizes

roelof-groenewald:
The SYCL compiler on Aurora complains that these two kernels are too large so here they are split into three separate kernels each.


roelof-groenewald:
I can confirm that these changes fixes the `Total size of kernel arguments exceeds limit! Total arguments size: 2232, limit: 2048`-type errors on Aurora.

WeiqunZhang (approved):

PR #6001, opened 2025-07-04
Fix temperature deposition thread safety

clarkse:
This PR addresses issues raised in the dev meeting where online updates for variance deposition are quite difficult to make thread safe. This removes the online single pass update and replaces it with the traditional single and double pass variance calculations. It currently is hardcoding the double pass option. A follow up PR will make it user selectable while also integrating the deposition into the diagnostics routines.


Source/Particles/Deposition/TemperatureDeposition.H
```diff
@@ -54,61 +56,81 @@ void varianceDepositionSubKernel(
...
+        // in either case. Will be non-demeaned in 1 pass
+        amrex::Real vxb = 0.;
+        amrex::Real vyb = 0.;
+        amrex::Real vzb = 0.;
+
+        // Normalization of accumulation arrays are not done until after all deposition, so nomralize
+        // here for double pass algorithm
+        if (type == TemperatureDepositionType::DOUBLE_PASS)
+        {
+            if (nx_arr(ixv[0], ixv[1], ixv[2], 0) > 0) {
```
dpgrote, line 106: Is `nx_arr` needed here? This should be checking if `wx_arr > 0`.

Source/Particles/Deposition/TemperatureDeposition.H
```diff
@@ -54,61 +56,81 @@ void varianceDepositionSubKernel(
...
+            // Update wv2_sum
+        amrex::Gpu::Atomic::AddNoRet(
+            &w2x_arr(ixv[0], ixv[1], ixv[2], 0), wqx_var*vxd*vxd);
+        amrex::Gpu::Atomic::AddNoRet(
+            &w2y_arr(iyv[0], iyv[1], iyv[2], 0), wqy_var*vyd*vyd);
+        amrex::Gpu::Atomic::AddNoRet(
+            &w2z_arr(izv[0], izv[1], izv[2], 0), wqz_var*vzd*vzd);
+    }
+
+    if (type == TemperatureDepositionType::SINGLE_PASS || pass == TemperatureDepositionPass::FIRST)
```
dpgrote: Is there a reason that this `if` block is not included in the one above at line 68?

Source/Particles/Deposition/VarianceAccumulationBuffer.cpp
```diff
@@ -37,21 +39,25 @@ VarianceAccumulationBuffer::VarianceAccumulationBuffer (ablastr::fields::MultiLe
...
 {
     using ablastr::fields::Direction;
     auto &warpx = WarpX::GetInstance();
 
     for (int lev = 0; lev <= warpx.finestLevel(); ++lev) {
         for (int idir = 0; idir < 3; ++idir) {
-            get("w", Direction{idir}, lev)->setVal(val);
-            get("w2", Direction{idir}, lev)->setVal(val);
-            get("vbar", Direction{idir}, lev)->setVal(val);
+            get("w", Direction{idir}, lev)->setVal(0._rt);
```
dpgrote: Using the shortened name here, `w`, could be confusing since this is actually getting `variance_buffer_w_`. I would prefer not using the `get` convenience function but writing out the full `warpx.m_fields.get` expression for clarity.

Source/Particles/Deposition/TemperatureDeposition.H
clarkse: wx_arr will be zero if nx is zero. I thought an integer comparison would be better so it doesn't need to be fuzzy. I.e. instead of using something like wx_arr > std::numeric_limits::epsilon(). nx is needed for the (n-1)/n normalization correction for sample variance, so I thought this would be more robust.

Source/Particles/Deposition/TemperatureDeposition.H
clarkse: No reason. I just fenced off those calls. Moving up since it makes it a bit cleaner.

Source/Particles/Deposition/VarianceAccumulationBuffer.cpp
clarkse: Changed.

Source/Particles/Deposition/TemperatureDeposition.H
clarkse: I guess the only issue would be if we are depositing zero weight particles, I don't think we will have that situation, but if you think it may be problematic, I can at least short circuit with n, then do a fuzzy comparison for the weight sum as well.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -3705,21 +3725,175 @@ PhysicalParticleContainer::AccumulateVelocitiesAndComputeTemperature (
...
+            bool single_pass = (depos_type == warpx::particles::deposition::TemperatureDepositionType::SINGLE_PASS);
+
+            // Update Mean and Variance values after running through weight deposition loop
+            amrex::ParallelFor(tbx, tby, tbz,
+                [=] AMREX_GPU_DEVICE (int i, int j, int k) {
+                    if (nx_arr(i,j,k) > 1) {
+                        amrex::Real sumw = wx_arr(i,j,k);
+                        amrex::Real sumwv = vxbar_arr(i,j,k);
+                        amrex::Real n = static_cast<amrex::Real>(nx_arr(i,j,k));
+                        amrex::Real norm = n/((n-1._rt)*sumw);
```
dpgrote, line 3834: Why do you use the `n/(n-1)` factor?

Source/Particles/Deposition/TemperatureDeposition.H
dpgrote: When comparing to zero, there shouldn't be any fuzziness since zero is an exact value. I don't think there is any need to compare against epsilon.

There would never be zero weight particles.

Source/Particles/PhysicalParticleContainer.cpp
clarkse: I am conforming to the algorithm in https://dl.acm.org/doi/pdf/10.1145/359146.359153 which uses sample variance normalization 1/(n-1) or Bessel's correction for normalization instead of population variance 1/n. I initially was using population variance, but transitioned it to the sample variance to stick to the algorithm in Bell. I think in terms of how the macro-particles represent samples of real particle distributions it makes sense to use.

I am thinking about it this way. Say we have some PDF and calculate its corresponding CDF, then randomly draw from that CDF to generate a set of macro-particles. In the case where they are not evolved, a computation of variance using Bessel's correction on the particle's velocities should be more accurate. This should still be the case statistically after evolution as well since we are evolving samples of distribution functions. Do you agree?

dpgrote: I've always used `1/n` - here, the mean is being calculated using all of the particles, the whole population, so the correction for using a sample is not needed.

clarkse: I guess the point is that the macro-particle population is intrinsically a sampled dataset. I have waffled back and forth on it a bit, but I think based on the thought experiment of how we load particles, by sampling a larger population, I would argue that we are not sampling the entire real population and we should have the correction included. I have plotted out both cases for each temperature component and computed the average temperatures of the corresponding temperature calculations after 1 small timestep. They particles are loaded in with 10 eV in each component. The sample variance calculation gets closer to the 10 eV that is expected.

image

clarkse: I tried this with a much lower PPC and the error is quite substantial when not using the n/(n-1). E.g. when using 5 PPC instead of 50 and doing the same thing the T_avg_pop ~ 9.42 eV and T_Avg_sample ~ 9.67. So the bias adjustment makes a difference in low sampling regions.

Source/Particles/Deposition/TemperatureDeposition.H
clarkse: I think that using the sample count is superior. There is a check when doing the finalization that requires at least 2 samples present to compute the variance. This helps remove a lot of the spurious issues in the 1-2 particle cells. I think that I have demonstrated the need for using Bessel's correction below and if I have the sample counts to compute that, then I have the counts available for these checks. I prefer discrete checks as I have seen weird rounding issues in the past and don't necessarily trust non-fuzzy floating point comparisons. Since it is an iMultiFab it has much lower memory requirements than the other arrays, so I don't really see an issue here.

Another situation where I see possible issues with rounding is when computing in the two pass scheme a mean and then squaring something that maybe has rounding error and the subtraction doesn't cancel exactly. Without the 2 particle check this value can be non-zero and small. When we enforce using 2 samples for the variance calculation then it is highly unlikely that something like that can occur. I have seen some of those artifacts previously and the discrete checks here and during normalization really seem to help a lot with it.

Source/Particles/Deposition/TemperatureDeposition.H
```diff
@@ -53,61 +55,78 @@ void varianceDepositionSubKernel(
...
+
+    if (type == TemperatureDepositionType::SINGLE_PASS || pass == TemperatureDepositionPass::SECOND)
+    {
+        // For a single pass run through, the vbar arrays will be zeroed, so this should work
+        // in either case. Will be non-demeaned in 1 pass
+        amrex::Real vxb = 0.;
+        amrex::Real vyb = 0.;
+        amrex::Real vzb = 0.;
+
+        // Normalization of accumulation arrays are not done until after all deposition, so nomralize
```
roelof-groenewald:
```suggestion
        // Normalization of accumulation arrays are not done until after all deposition, so normalize
```

Source/Particles/Deposition/TemperatureDeposition.H
```diff
@@ -116,10 +135,12 @@ void varianceDepositionSubKernel(
  * \param xp, yp, zp    The particle positions.
  * \param wq            The charge of the macroparticle
  * \param vx,vy,vz      The particle velocities
- * \param varx_arr,vary_arr,varz_arr Array4 of velocity variance, either full array or tile.
+ * \param nx_arr,ny_arr,nz_arr Array4 of velocity variance, either full array or tile.
```
roelof-groenewald: The doc-string here should be updated to reflect that `n*_arr` stores the number of samples.

Source/Particles/Deposition/TemperatureDeposition.H
```diff
@@ -116,10 +135,12 @@ void varianceDepositionSubKernel(
...
  * \param wx_arr,wy_arr,wz_arr Array4 of weight sums, either full array or tile.
  * \param w2x_arr,w2y_arr,w2z_arr Array4 of weight**2 sums, either full array or tile.
  * \param varx_type,vary_type,varz_type The grid types along each direction, either NODE or CELL
+ * \param type The type of deposition reoutine, either SINGLE_PASS or DOUBLE_PASS
```
roelof-groenewald:
```suggestion
 * \param type The type of deposition routine, either SINGLE_PASS or DOUBLE_PASS
```

Source/Particles/Deposition/TemperatureDeposition.H
```diff
@@ -380,9 +411,12 @@ void doVarianceDepositionShapeNKernel(
                          since q is a scalar. For non-ionizable species,
                          ion_lev is a null pointer.
  * \param varx_fab,vary_fab,varz_fab FArrayBox of variance, either full array or tile.
+ * \param nx_iab,ny_iab,nz_iab IArrayBox of sample count buffer, either full array or tile.
  * \param wx_fab,wy_fab,wz_fab FArrayBox for running weights buffer, either full array or tile.
  * \param w2x_fab,w2y_fab,w2z_fab FArrayBox for weights^2 buffer, either full array or tile.
  * \param vxbar_fab,vybar_fab,vzbar_fab FArrayBox of current density, either full array or tile.
+ * \param type The type of deposition reoutine, either SINGLE_PASS or DOUBLE_PASS
```
roelof-groenewald:
```suggestion
 * \param type The type of deposition routine, either SINGLE_PASS or DOUBLE_PASS
```

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -3674,13 +3684,23 @@ PhysicalParticleContainer::AccumulateVelocitiesAndComputeTemperature (
     const amrex::Real relative_time)
 {
     using ablastr::fields::Direction;
+    using warpx::particles::deposition::TemperatureDepositionType;
+    using warpx::particles::deposition::TemperatureDepositionPass;
+
+    // Todo: link this to inputs (hardcoded for the time being)
+    // Will fix this in a follow up PR.
```
roelof-groenewald, line 3684: Could you briefly summarize the reasons to support both single and double pass schemes?

Source/Particles/PhysicalParticleContainer.cpp
roelof-groenewald, line 3834: Out of curiosity, is this cast actually necessary? I would think the `norm = n/((n-1._rt)*sumw)` would implicitly perform the cast itself.

roelof-groenewald (commented):
I left a few minor comments.

I also had some renewed thoughts about the fact that (weight * charge) is being deposited and then the charge is normalized away - this will not really work for cases where the ionization level is used, no? In those cases not all particles will have the same charge and therefore dividing by species charge will not actually give you the total weight. Perhaps for now there can just be a check that if ionization level is used, the temperature cannot be used?

Source/Particles/PhysicalParticleContainer.cpp
clarkse: I want to benchmark the performance between the two algorithms for accuracy and speed. two-pass is going to be default, but may be able to squeeze some performance gains out if the numerics issues do not crop up in single pass for the calculations. It was fairly trivial to set up single and double pass so I just did it.

Source/Particles/PhysicalParticleContainer.cpp
clarkse: The cast is not necessary, but I prefer not to leave these things up to the compiler. I have seen dumb precision loss things in the past trying to rely on implicit casts in C/C++. I think it is more clear and will never lead to compilers acting in strange ways.

clarkse:
> I left a few minor comments.
> 
> 
> 
> I also had some renewed thoughts about the fact that (weight * charge) is being deposited and then the charge is normalized away - this will not really work for cases where the ionization level is used, no? In those cases not all particles will have the same charge and therefore dividing by species charge will not actually give you the total weight. Perhaps for now there can just be a check that if ionization level is used, the temperature cannot be used? 

Aren't the charge states species dependent? This does the deposition for each species independently. I have not used the  ionization levels, but I can assert it out of that is going to be problematic. I will revisit it.

Source/Particles/PhysicalParticleContainer.cpp
roelof-groenewald: Sounds good! Thanks for explaining.

roelof-groenewald:
> Aren't the charge states species dependent? This does the deposition for each species independently. I have not used the ionization levels, but I can assert it out of that is going to be problematic. I will revisit it.

The ionization level is particle specific - the idea with that is that you could have a species of some atom and then individual particles could be ionized to different levels (as opposed to having different species for each ionization level). So I think adding the `assert` would be good to be safe.

clarkse:
> > Aren't the charge states species dependent? This does the deposition for each species independently. I have not used the ionization levels, but I can assert it out of that is going to be problematic. I will revisit it.
> 
> The ionization level is particle specific - the idea with that is that you could have a species of some atom and then individual particles could be ionized to different levels (as opposed to having different species for each ionization level). So I think adding the `assert` would be good to be safe.

I am just dropping the ionization level as it would weigh highly ionized species particles more than the non-ionized particles, and it is easy enough to remove.

Source/Particles/Deposition/TemperatureDeposition.H
```diff
@@ -507,10 +497,7 @@ void doVarianceDepositionShapeN (const GetParticlePosition<PIdx>& GetPosition,
             const amrex::ParticleReal vy  = uyp[ip]*gaminv;
             const amrex::ParticleReal vz  = uzp[ip]*gaminv;
 
-            amrex::ParticleReal wq  = q*wp[ip];
-            if (do_ionization){
-                wq *= ion_lev[ip];
-            }
+            amrex::ParticleReal wq  = wp[ip];
```
roelof-groenewald: This is a super minor thing and can be addressed in a follow-up PR, but since `wq` is actually just the weight `wp` or just `w` would be less likely to be confusing. Similarly, the docstring in `doVarianceDepositionShapeNKernel` says "The charge of the macroparticle" but should be "The weight of the macroparticle".

roelof-groenewald (approved):

PR #3478, opened 2022-10-24
CI: oneAPI with `-O1`

ax3l:
oneAPI 2022.2.0 hangs for `-O2` and higher: #3442

Thx @WeiqunZhang for triaging this. Currently takes 1h+ to compile `CoarsenIO::interp` ([reproducer](https://github.com/WeiqunZhang/amrex-devtests/blob/main/dpcpp_compile_time/main.cpp)). Reported to Intel.

PR #6007, opened 2025-07-08
SYCL: Turn on AOT in CI and Update MKL

WeiqunZhang:
- AOT can help expose issues earlier.
- A header we use has been deprecated  by Intel.
- A big kernel has been split into three so that their kernel parameter size can be less than 2KB.

### DIFF
--- a/.github/workflows/dependencies/dpcpp.sh
+++ b/.github/workflows/dependencies/dpcpp.sh
@@ -29,16 +29,16 @@ df -h
 # https://github.com/BLAST-WarpX/warpx/pull/1566#issuecomment-790934878
 
 # try apt install up to five times, to avoid connection splits
-# FIXME install latest version of IntelLLVM, Intel MKL
-#       after conflicts with openPMD are resolved
 status=1
 for itry in {1..5}
 do
     sudo apt-get install -y --no-install-recommends \
         build-essential \
         cmake           \
-        intel-oneapi-compiler-dpcpp-cpp=2024.2.1-1079 \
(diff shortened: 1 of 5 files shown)


WeiqunZhang:
Please ignore my change to amrex commit. I just wanted to test the new Parser in the same PR. It will be reverted.

.github/workflows/dependencies/dpcpp.sh
```diff
@@ -29,16 +29,16 @@ df -h
 # https://github.com/BLAST-WarpX/warpx/pull/1566#issuecomment-790934878
 
 # try apt install up to five times, to avoid connection splits
-# FIXME install latest version of IntelLLVM, Intel MKL
-#       after conflicts with openPMD are resolved
```
WeiqunZhang, line 33: @ax3l I updated the code that uses MKL FFT in this PR. What were the conflicts with openPMD?

WeiqunZhang: Okay it was added in #5419. Since there is no error now, it is safe to remove it.

WeiqunZhang: We can also try to remove `-O1 -NDEBUG`, which was added in #3478 to an address oneAPI 2022.2.0 issue.

WeiqunZhang:
Do we know why this test test_rz_langmuir_multi_psatd_JRhom_LL4 failed? This PR could not be the reason.

roelof-groenewald (approved):
Thanks for adding the AOT flag and fixing the large kernel!

PR #5955, opened 2025-06-17
Energy-preserving explicit electrostatic PIC-MCC algorithm

EZoni:
Implement the electrostatic part of the energy-preserving PIC-MCC algorithm described in #5851 (i.e., explicit PIC coupled with Monte Carlo collisions).

TODO:
- [x] Rebase after #6116.
- [x] Clean up unnecessary style changes to minimize diff view.
- [x] Introduce an input parameter to control the new algorithm.
- [x] Address https://github.com/BLAST-WarpX/warpx/pull/5955#issuecomment-3268377726.
- [x] Update documentation.
- [x] Address https://github.com/BLAST-WarpX/warpx/pull/5955#discussion_r2341648574: #6174.
- [x] Address review comments.
- [x] Switch option on by default.

New entry in the documentation:
Screenshot from 2025-09-22 15-40-19

PR #6008, opened 2025-07-09
Rename `np_current` variable for more clarity

EZoni:
While working on #5955, I noticed that we have a local variable named `np_current`, which denotes the number of particles for which the charge or current are deposited. This is passed as an argument to the functions `DepositCharge`, `DepositCurrent`, and `DepositCurrentAndMassMatrices`.

I find it somewhat confusing or misleading to see `np_current` passed as an argument of, say, `DepositCharge`.

I think choosing a better name could make the code simpler to read for humans, possibly even less error prone.

Since the corresponding function argument is defined as `np_to_deposit` in all three functions (see below), I propose to rename the variable `np_to_deposit`, for better clarity.

- `DepositCharge`:
https://github.com/BLAST-WarpX/warpx/blob/9d887fd61fb54ddb1630c0ae85d461f47b966d76/Source/Particles/WarpXParticleContainer.H#L313-L322
- `DepositCurrent`:
https://github.com/BLAST-WarpX/warpx/blob/9d887fd61fb54ddb1630c0ae85d461f47b966d76/Source/Particles/WarpXParticleContainer.H#L324-L340
- `DepositCurrentAndMassMatrices`:
https://github.com/BLAST-WarpX/warpx/blob/9d887fd61fb54ddb1630c0ae85d461f47b966d76/Source/Particles/WarpXParticleContainer.H#L385-L410

### DIFF
--- a/Source/Particles/LaserParticleContainer.cpp
+++ b/Source/Particles/LaserParticleContainer.cpp
@@ -624,20 +624,20 @@ LaserParticleContainer::Evolve (ablastr::fields::MultiFabRegister& fields,
             amplitude_E.resize(np);
 
             // Determine whether particles will deposit on the fine or coarse level
-            long np_current = np;
+            long np_to_deposit = np;
             if (lev > 0 && m_deposit_on_main_grid && has_buffer) {
-                np_current = 0;
+                np_to_deposit = 0;
             }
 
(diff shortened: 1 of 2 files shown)


lucafedeli88 (approved):
Thanks for this PR, @EZoni !

PR #6009, opened 2025-07-09
Update dane scripts

dpgrote:
This updates the script for setting up WarpX on Dane. Here is a list of the changes:

- Use an older version of Python. The newer version, 3.12.2 breaks the build of Adios2 because of a reference to a too new version of GLIBCXX.
- Makes the software installation directory user settable, in the variable WARPX_SW_DIR
- Executes the deactivate function if it is defined in case a Python virtual environment was already activated.
- Removes the install of virtualenv in user space so that the installation is completely contained in the install directory.
- The install script uses its own location to find the requirements.txt file. This allows that user to download WarpX anywhere.
- Update the instructions in the doc accordingly

### DIFF
--- a/Docs/source/install/hpc/dane.rst
+++ b/Docs/source/install/hpc/dane.rst
@@ -26,13 +26,15 @@ Preparation
 
 Use the following commands to download the WarpX source code.
 Note that these commands and the shell scripts all assume the bash shell.
+This downloads WarpX into the workspace directory, which is recommended.
+WarpX can be downloaded elsewhere if that doesn't work with your directory structure, but note that the commands shown below refer to WarpX in the workspace directory.
 
 .. code-block:: bash
 
    git clone https://github.com/BLAST-WarpX/warpx.git /usr/workspace/${USER}/dane/src/warpx
 
-We use system software modules, add environment hints and further dependencies via the file ``$HOME/dane_warpx.profile``.
-Create it now:
+The system software modules, environment hints, and further dependencies are setup via the file ``$HOME/dane_warpx.profile`` which is copied from the WarpX source.
+Set it up now:
 
 .. code-block:: bash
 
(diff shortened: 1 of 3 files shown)


Tools/machines/dane-llnl/dane_warpx.profile.example
```diff
@@ -27,7 +27,8 @@ export LD_LIBRARY_PATH=${SW_DIR}/install/blaspp-2024.10.26/lib64:$LD_LIBRARY_PAT
 export LD_LIBRARY_PATH=${SW_DIR}/install/lapackpp-2024.10.26/lib64:$LD_LIBRARY_PATH
 
 # optional: for Python bindings
-module load python/3.12.2
+#module load python/3.12.2 # Note this version breaks the adios build
```
ax3l:
```suggestion
#module load python/3.12.2 # Note this version uses a too new GLIBCXX and breaks the adios build
```

Docs/source/install/hpc/dane.rst
```diff
@@ -58,13 +60,18 @@ Exit the ``vi`` editor with ``Esc`` and then type ``:wq`` (write & quit).
...
+   Now, and as the first step on future logins to Dane, activate these environment settings by executing the file:
 
    .. code-block:: bash
 
       source $HOME/dane_warpx.profile
 
-Finally, since Dane does not yet provide software modules for some of our dependencies, install them once:
+Finally, since Dane does not yet provide software modules for some of our dependencies, WarpX provides a script to install them.
+This is done executed now.
+They are by default installed in the workspace directory (which is recommended), but can be installed elsewhere by setting the environment variable ``WARPX_SW_DIR``.
```
ax3l, line 71: Probably sensible to generally rename from `SW_DIR` to `WARPX_SW_DIR` in all scripts...

ax3l (approved):
Thank you, looks great!

Docs/source/install/hpc/dane.rst
dpgrote: Yes, I had update dane_warpx.profile.example but hadn't yet committed the changes. I'll submit another PR. I didn't think that this would be merged so fast :)

ax3l: Fix in #6010

PR #6011, opened 2025-07-10
Add ohms law multiple species support

clarkse:
This PR adds multiple species support to the hybrid ohm's law solver in warpx. It includes holding multifabs for each ion species, and applying resistive drag to each species as a collision operator that is called in the particle pusher.


Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -238,11 +238,52 @@ void WarpX::HybridPICDepositRhoAndJ ()
 
     // Perform charge deposition in component 0 of rho_fp at current time.
     mypc->DepositCharge(m_fields.get_mr_levels(FieldType::rho_fp, finest_level), 0._rt);
+
     // Perform current deposition at t_{n-1/2}.
-    mypc->DepositCurrent(m_fields.get_mr_levels_alldirs(FieldType::current_fp, finest_level), dt[0], -0.5_rt * dt[0]);
+    // This is broken up into deposition across multiple species MultiFabs to preserve
+    // Species average velocities, whic is used in collision processes.
```
```suggestion
    // Species average velocities, which is used in collision processes.
```

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -259,7 +300,14 @@ void WarpX::HybridPICDepositRhoAndJ ()
     // Synchronize J and rho:
     // filter (if used), exchange guard cells, interpolate across MR levels
     // and apply boundary conditions
-    SyncCurrentAndRho();
+    SyncRho();
+
+    // Reflect charge and current density over PEC boundaries, if needed.
```
```suggestion
    // Reflect charge density over PEC boundaries, if needed.
```

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -259,7 +300,14 @@ void WarpX::HybridPICDepositRhoAndJ ()
     // Synchronize J and rho:
     // filter (if used), exchange guard cells, interpolate across MR levels
     // and apply boundary conditions
-    SyncCurrentAndRho();
+    SyncRho();
```
roelof-groenewald, line 302: The comment above this line should be changed since now it is only "synchronizing" rho (and not applying the boundary condition anymore).

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -281,6 +329,16 @@ void WarpX::HybridPICDepositRhoAndJ ()
...
+        for (const auto & spec : mypc->GetSpeciesNames())
+        {
+            ablastr::utils::communication::FillBoundary(
+                *m_fields.get("hybrid_current_fp_" + spec, Direction{idim}, lev),
+                m_fields.get("hybrid_current_fp_" + spec, Direction{idim}, lev)->nGrowVect(),
+                WarpX::do_single_precision_comms,
+                Geom(lev).periodicity(),
+                true
+            );
+        }
```
roelof-groenewald, line 340: This should be inside the for-loop over `idim` above.

Source/Particles/Collision/BackgroundStopping/HybridResistiveDrag.cpp
```diff
@@ -0,0 +1,176 @@
...
+        }
+    }
+}
+
+void HybridResistiveDrag::doHybridResistiveDragOnIonsWithinTile (WarpXParIter& pti, amrex::Real dt, amrex::Real t,
+                                                               amrex::ParticleReal species_mass, amrex::ParticleReal species_charge)
+{
+    using namespace amrex::literals;
+
+    // So that CUDA code gets its intrinsic, not the host-only C++ library version
```
roelof-groenewald: Since WarpX runs on other GPU's beside nvidia:
```suggestion
    // So that GPU code gets its intrinsic, not the host-only C++ library version
```

Source/Particles/MultiParticleContainer.H
```diff
@@ -169,6 +169,26 @@ public:
     DepositCurrent (ablastr::fields::MultiLevelVectorField const& J,
                     amrex::Real dt, amrex::Real relative_time);
 
+/**
+     * \brief Deposit current density for only a single species
+     *        This is useful for depoisting just one speceis to compute
```
roelof-groenewald:
```suggestion
     *        This is useful for depositing just one species to compute
```

Source/Particles/MultiParticleContainer.H
```diff
@@ -169,6 +169,26 @@ public:
...
+     *                to MultiFabs per mesh refinement level)
+     * \param[in] dt Time step for particle level
+     * \param[in] relative_time Time at which to deposit J, relative to the time of the
+     *                          current positions of the particles. When different than 0,
+     *                          the particle position will be temporarily modified to match
+     *                          the time of the deposition.
+     *  \param[in] species_name Name of the species to do current dposition
+     */
+    void
+    DepositSpeciesCurrent (ablastr::fields::MultiLevelVectorField const& J,
```
roelof-groenewald, line 186: What is the reason for having this in the `MultiParticleContainer` and passing the `species_name` rather than just putting it in the `PhysicalParticleContainer`?

Source/Particles/Collision/CollisionHandler.H
```diff
@@ -28,6 +28,8 @@ public:
     /* Perform all of the collisions */
     void doCollisions (int step, amrex::Real cur_time, amrex::Real dt, MultiParticleContainer* mypc);
 
+    void addCollision (const std::string & name, const std::string & type )
```
roelof-groenewald, line 31: I don't see this being used anywhere, is this perhaps left over from a different iteration?

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
```diff
@@ -238,11 +238,52 @@ void WarpX::HybridPICDepositRhoAndJ ()
...
+        // Retrieve species specific array for deposition
+        mypc->DepositSpeciesCurrent(J_species, dt[0], -0.5_rt * dt[0]);
+
+        // Filter and sychronize current for each species
+        SyncCurrent("hybrid_current_fp_" + spec);
+
+        for (int lev = 0; lev <= finest_level; ++lev)
+        {
+            ApplyJfieldBoundary(lev,
+                m_fields.get("hybrid_current_fp_" + spec, Direction{0}, lev),
```
roelof-groenewald: I think we can drop the `hybrid` at the start of the name, and then this overall functionality would be easy to extend to generally allow diagnostic output of the individual species currents, which is something I've been meaning to do for a while.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.H
```diff
@@ -184,6 +184,8 @@ public:
     /** Electron pressure scaling exponent */
     amrex::Real m_gamma = 5.0/3.0;
 
+    std::vector<std::string> m_ion_species_names;
```
roelof-groenewald, line 187: Is it necessary to keep the list of species as a member of the `HybridPICModel` class? It looks like it is only used once in the `AllocateLevelMFs` function and otherwise `mypc->GetSpeciesNames()` is used. Can `GetSpeciesNames()` just be used in `AllocateLevelMFs` as well?

roelof-groenewald (commented):
This looks to be coming together very nicely. I left a few preliminary comments for your consideration.
Thanks for working on this! It'll be a very nice extension to the current code capabilities.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -125,6 +128,30 @@ void HybridPICModel::AllocateLevelMFs (
         lev, amrex::convert(ba, jz_nodal_flag),
         dm, ncomps, ngJ, 0.0_rt);
 
+    // Allocate species specific multifabs to deposit individual currents
+    for (const auto & specname : m_ion_species_names)
```
roelof-groenewald, line 132: It would be good to add a check in here for species charge to not allocate all the extra space for neutral particles (and then similarly skip the deposition, synchronization, guard cell exchange and addition to J_total in `DepositRhoAndJ`).

Source/Particles/Collision/CollisionHandler.H
clarkse: Yes, this will be removed. I was toying with the idea of having the HybridPIC model itself instantiate the collision operators and adding to the collision queue. I realized though that the CollisionHandler queue is driven highly from the parameter parser structure and I didn't want to deviate from the modus operandi of how all the other collisions work in WarpX. So the current way will require having to declare a collision in the input deck. I am planning to instead handle constructing the parmparse parameters in the picmi object and then sticking to the existing convention in c++ input decks.

Source/FieldSolver/WarpXPushFieldsHybridPIC.cpp
clarkse: ok, that works for me. I can adjust the MF naming convention accordingly.

PR #6012, opened 2025-07-11
Minor refactoring of PIC loop extracted from PR 5955

EZoni:
### Overview

I'm extracting changes from #5955 to separate PRs, in order to make it easier to implement, debug, and review #5955.

For context, the final algorithm that we want to implement in #5955 is similar to what is shown in this diagram (except that particle injection will be after collisions):

Image

### Summary of code changes

- Move the function calls related to collisions inside the "one step" functions, for both the implicit and explicit loops, since in the explicit case collisions will be placed between separate parts of the particle push (which typically belongs to the "one step" functions).
- Add a `WarpX::OneStep` function that encapsulates all the "one step" functions for the explicit loop, similarly to the `OneStep` method of the class `ImplicitSolver`.
- Add a `DoCollisions` method to the class `ImplicitSolver`, which is in turn called from within the `OneStep` methods of each derived class.
- Restyle some function declarations listing one input argument per line, as I expect to introduce new input arguments to control the position and momentum push in 5955. This is also typically good practice to produce diffs that are easier to review on GitHub.


Source/Evolve/WarpXEvolve.cpp
```diff
@@ -216,58 +216,28 @@ WarpX::Evolve (int numsteps)
...
-        // Electromagnetic case: subcycling with one level of mesh refinement
-        else if (m_do_subcycling && (finest_level == 1))
-        {
-            OneStep_sub1(cur_time);
-        }
-        else
-        {
-            WARPX_ABORT_WITH_MESSAGE(
-                "do_subcycling = " + std::to_string(m_do_subcycling)
-                + " is an unsupported do_subcycling type.");
```
EZoni, line 270: I rewrote this if/else logic as:
```
- implicit:
    ...
- explicit:
    - electrostatic:
        ...
    - electromagnetic: 
        - without mesh refinement:
            - standard PIC loop:
                ...
            - JRhom PIC loop:
                ...
        - with mesh refinement:
            - without subcycling:
                ...
            - with subcycling:
                ...
```
instead of
```
- implicit:
    ...
- electrostatic:
    ...
- JRhom PIC loop:
    ...
- without mesh refinement and without subcycling:
    ...
- with mesh refinement (one level) and with subcycling:
    ...
```

Without this change, (implicit vs explicit), (electrostatic vs electromagnetic), (standard vs JRhom), (without vs with mesh refinement), etc., are basically all on equal footing in the if/else logic, which I think is quite counterintuitive and error prone.

For example, as is, the two `else if` blocks `else if (m_JRhom)` and `else if ( !m_do_subcycling || (finest_level == 0))` could not be reordered, otherwise the JRhom tests would fail, since the conditions checked in each statement are not mutually exclusive.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -1186,20 +1213,36 @@ WarpX::doQEDEvents ()
...
-        PushParticlesandDeposit(lev, cur_time, DtType::Full, skip_current,
-                                deposit_mass_matrices, push_type);
+        PushParticlesandDeposit(
+            lev,
+            cur_time,
+            DtType::Full,
+            skip_current,
+            deposit_mass_matrices,
+            push_type
+        );
```
EZoni, line 1229: This is an example of function call where I expect to add more input arguments in #5955, hence the restyling with one input argument per line.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -216,58 +216,28 @@ WarpX::Evolve (int numsteps)
...
-        }
-        else
-        {
-            WARPX_ABORT_WITH_MESSAGE(
-                "do_subcycling = " + std::to_string(m_do_subcycling)
-                + " is an unsupported do_subcycling type.");
+        // explicit solver
+        else {
+            // Maybe TODO: m_explicit_solver->OneStep(cur_time, dt[0], step);
+            OneStep(cur_time, dt[0], step);
```
EZoni: As is, we could avoid passing `cur_time`, `dt[0]`, and `step` as input arguments, since this `OneStep` is a member of the `WarpX` class. However, so far I have chosen to replicate the function interface of the `OneStep` method of the `ImplicitSolver` class, for symmetry.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -216,58 +216,28 @@ WarpX::Evolve (int numsteps)
...
         mypc->doQEDSchwinger();
 #endif
 
-        // Main PIC operation:
-        // gather fields, push particles, deposit sources, update fields
-
+        // callback called when particle injection happens, after the position
+        // advance and before deposition is called, allowing a user
+        // defined particle distribution to be injected each time step
         ExecutePythonCallback("particleinjection");
```
EZoni: Inline comment copied from https://github.com/BLAST-WarpX/warpx/blob/1ed220d77792c3618482e4ecb8781695094200ff/Python/pywarpx/callbacks.py#L47-L49

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -414,6 +384,63 @@ WarpX::Evolve (int numsteps)
...
+{
+    WARPX_PROFILE("WarpX::OneStep()");
+
+    // electrostatic solver or hybrid solver
+    if (electromagnetic_solver_id == ElectromagneticSolverAlgo::None ||
+        electromagnetic_solver_id == ElectromagneticSolverAlgo::HybridPIC) {
+        // perform collisions
+        ExecutePythonCallback("beforecollisions");
+        mypc->doCollisions(a_step, a_cur_time, a_dt);
+        ExecutePythonCallback("aftercollisions");
```
JustinRayAngus: This may be needed for PR's to come, but for this PR it seems that these three lines, which are duplicated in the else{} condition below could be pulled outside the if statement above.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -216,58 +216,28 @@ WarpX::Evolve (int numsteps)
...
-            OneStep_sub1(cur_time);
-        }
-        else
-        {
-            WARPX_ABORT_WITH_MESSAGE(
-                "do_subcycling = " + std::to_string(m_do_subcycling)
-                + " is an unsupported do_subcycling type.");
+        // explicit solver
+        else {
+            // Maybe TODO: m_explicit_solver->OneStep(cur_time, dt[0], step);
```
JustinRayAngus: 👍

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
```diff
@@ -583,7 +600,12 @@ void ImplicitSolver::PreRHSOp ( const amrex::Real  a_cur_time,
     }
     else {  // Conventional particle-suppressed JFNK
         bool deposit_mass_matrices = false;
-        m_WarpX->PushParticlesandDeposit(a_cur_time, skip_current, deposit_mass_matrices, push_type);
+        m_WarpX->PushParticlesandDeposit(
+            a_cur_time,
+            skip_current,
+            deposit_mass_matrices,
+            push_type
+        );
```
JustinRayAngus: This will need to be changed. As is, it will call collisions at each JNFK iteration. The collisions only need to occur once at the end of the time step.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -216,58 +216,28 @@ WarpX::Evolve (int numsteps)
...
         }
 
-        // Run multi-physics modules:
-        // ionization, Coulomb collisions, QED
+        // multi-physics: field ionization
         doFieldIonization();
 
-        ExecutePythonCallback("beforecollisions");
-        mypc->doCollisions( step, cur_time, dt[0] );
-        ExecutePythonCallback("aftercollisions");
```
JustinRayAngus, line 225: I think these three lines should be moved to their own function in WarpX. Then, and elsewhere in the code, such as right after m_implicit_solver->OneStep(cur_time, dt[0], step);, one can just call this routine. No need to modify the implicit routines to include call to collisions.

EZoni: Thanks, @JustinRayAngus! Will flag it as [WIP] again and work on fixing this.

EZoni: The reason why initially I put it in the implicit `OneStep` is that I was trying to have some symmetry between implicit and explicit, since in the explicit case it will be part of a `OneStep` function (as it will be in the middle of position or velocity push).

From what you're saying here and in https://github.com/BLAST-WarpX/warpx/pull/6012#discussion_r2226675247, it sounds like the various, concrete, implicit `OneStep` functions do not span the whole time step but rather each JFNK iteration, is that the case?

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
EZoni: I'm not sure if this comment was meant to go here. This was just a style change, I can revert it so that the PR is easier to parse through: [faa8a21](https://github.com/BLAST-WarpX/warpx/pull/6012/commits/faa8a219f2b96c4db25478efebd652b4fbafdbbb).

Source/Evolve/WarpXEvolve.cpp
EZoni: Yes, I expect these to look different in the end, but I'm okay with pulling the lines outside the `if` block for now: [023f6dc](https://github.com/BLAST-WarpX/warpx/pull/6012/commits/023f6dc4ea3121f7fab485b8e60112c231a01f18).

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
```diff
@@ -31,6 +31,18 @@ void ImplicitSolver::CreateParticleAttributes () const
...
+    amrex::Real cur_time,
+    amrex::Real dt
+) const
+{
+    // multi-physics: collisions
+    ExecutePythonCallback("beforecollisions");
+    m_WarpX->GetPartContainer().doCollisions(step, cur_time, dt);
+    ExecutePythonCallback("aftercollisions");
+}
+
```
JustinRayAngus: I'm not sure I see the logic in having this function defined in the implicit solver class. Collisions have nothing to do with the implicit solver. Why can't this function live in the WarpX class?

Source/FieldSolver/ImplicitSolvers/SemiImplicitEM.cpp
```diff
@@ -70,6 +70,8 @@ void SemiImplicitEM::OneStep ( amrex::Real  start_time,
 
     amrex::ignore_unused(a_step);
 
+    DoCollisions(a_step, start_time, a_dt);
+
```
JustinRayAngus: I'm not a fan of having this called from inside the Implicit solver, which has nothing to do with collisions. Logically it should be called before (or after) the call to OneStep() from the WarpX Evolve routine. I understand that you need to put this inside OneStep for the new way it is going to be done in the explicit solver, which breaks this logical separation of PIC and MCC. However, I don't see why that means this logical seperation needs to be broken elsewhere when it is not needed.

EZoni: Indeed, I don't think the separation needs to be broken in the implicit solver as well. It was just a suggestion, but it's definitely not needed. I can break the separation only in the explicit solver and leave the implicit solver as is.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -414,6 +377,68 @@ WarpX::Evolve (int numsteps)
...
+    amrex::Real a_dt,
+    int a_step
+)
+{
+    WARPX_PROFILE("WarpX::OneStepWithCollisions()");
+
+    // perform collisions
+    ExecutePythonCallback("beforecollisions");
+    mypc->doCollisions(a_step, a_cur_time, a_dt);
+    ExecutePythonCallback("aftercollisions");
```
EZoni: This will move around for the explicit solver changes in #5955.

Source/FieldSolver/ImplicitSolvers/SemiImplicitEM.cpp
EZoni: @JustinRayAngus 

I pushed some code to fix this, let me know if this looks better to you or if you would like to request more changes.

JustinRayAngus (approved):
Looks good to me now. Thanks!

roelof-groenewald (approved):
So just to make this explicit - functionally the only difference here is that now `ExecutePythonCallback("particleinjection");` happens before collisions, right?
This could change CI on external projects that use WarpX, so it would be worth highlighting this change in Slack or somewhere so people are aware.

EZoni:
@roelof-groenewald 

Yes, functionally that's the only change.

This PR was meant to take some of the code refactoring that doesn't change (too much) the way things work out of #5955, to make that PR smaller, more targeted, and easier to debug (since there is and will be a significant amount of debugging to do there).

There is also the change described in https://github.com/BLAST-WarpX/warpx/pull/6012#discussion_r2200742952, which doesn't change the way things work but does improve one of the first bits of code one sees when starting to look into the PIC loop implementation. This could have been a separate PR in itself - I acknowledge that.

If having particle injection before the collisions may be a problem (besides just CI), we should probably discuss it before merging this PR.

@jlvay @JustinRayAngus @oshapoval @RemiLehe @dpgrote @ax3l 
You authored the paper that describes the algorithm, do you have any opinion on whether particle injection should happen before or after collisions?

JustinRayAngus:
> @jlvay @JustinRayAngus @oshapoval @RemiLehe @dpgrote @ax3l
> You authored the paper that describes the algorithm, do you have any opinion on whether particle injection should happen before or after collisions?

I don't think injection was ever considered previously. It's a good question.

jlvay:
I would propose to put it after the collisions. This way, if the process that generates particles to be injected depends on the knowledge of the position and velocities right in the middle of the time step, this can be calculated or stored as needed. E.g., if a particle gets close to a wall, you would want the collision to occur before the injection of particles that result from the potential collision between the particle and the wall so as to know precisely if the particle hits the wall in the first place and, if so, at which position and with which velocity amplitude and direction.

EZoni:
Readding work-in-progress tag until @jlvay's suggestion in https://github.com/BLAST-WarpX/warpx/pull/6012#issuecomment-3152652309 is implemented and all CI checks run.

Note that he also suggested that we keep names simple and replace `OneStepWithCollisions` with simply `OneStep`.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -414,6 +372,71 @@ WarpX::Evolve (int numsteps)
...
+{
+    WARPX_PROFILE("WarpX::OneStep()");
+
+    // perform particle collisions
+    ExecutePythonCallback("beforecollisions");
+    mypc->doCollisions(a_step, a_cur_time, a_dt);
+    ExecutePythonCallback("aftercollisions");
+
+    // perform particle injection
+    ExecutePythonCallback("particleinjection");
```
EZoni, line 389: These will move around in #5955 when we split the push.

roelof-groenewald (approved):
Looks good!

PR #6013, opened 2025-07-11
adding total iteration count to the nonlinear diagnostic file.

JustinRayAngus:


dpgrote (approved):
This looks good. A detail is whether these iteration counts should be preserved across a restart, but that can be a later PR is needed.

JustinRayAngus:
> This looks good. A detail is whether these iteration counts should be preserved across a restart, but that can be a later PR is needed.

Yes, I thought about that. In fact, that is what I do in PICNIC. I agree that it can be deferred to a future PR. It may also not really be needed.

PR #6014, opened 2025-07-11
CI: fix workaround for Open MPI oversubscription issue

EZoni:
## Overview

@JustinRayAngus ran the test suite on a local computing cluster and encountered this error (reported via email):
```
/usr/bin/srun: unrecognized option '--host'
Try "srun --help" for more information
```

This PR tries to avoid this error by adopting a different workaround for the original oversubscription issue fixed in #5973.


JustinRayAngus:
I can verify that that this PR fixes the issue when running ctests locally on Dane.

EZoni:
Unfortunately it doesn't work for Azure, so we will need to find a different workaround.

Examples/CMakeLists.txt
```diff
@@ -12,6 +12,12 @@ if(WarpX_MPI)
         if(${MPIEXEC_HELP_TEXT} MATCHES "^.*allow-run-as-root.*$")
             set(MPI_ALLOW_ROOT --allow-run-as-root)
         endif()
+        if(${MPIEXEC_HELP_TEXT} MATCHES "^.*--host.*$")
+            # Avoid the following error seen with Open MPI 4.1.6 in the Azure pipelines:
+            # "There are not enough slots available in the system to satisfy the 2 slots
+            # that were requested by the application"
+            set(MPI_HOST_NUMPROC --host localhost:${nprocs})
+        endif()
```
EZoni: @ax3l @WeiqunZhang @atmyers 

Do you know any other way to check whether a given MPI executable supports a given option, `--host` in this case, to avoid passing it when it is not supported?

The solution I implemented here does not seem to work in this case, although we use it for other options (see code right above).

EZoni:
Current rationale with [1f78637](https://github.com/BLAST-WarpX/warpx/pull/6014/commits/1f7863739914e0ee6aac6ea23afb5cc343aa96f4):
- Azure (root, Open MPI 4.1.6): The root branch runs `--help`, captures `stderr` (as before), detects `--host` in the output → `MPIEXEC_SUPPORTS_HOST ON` → `--host localhost:${nprocs}` is set inside the function with the correct `nprocs`.
- Cluster (non-root, `srun`): The new `else`-branch runs `srun --help`, `--host` is not in the output → `MPIEXEC_SUPPORTS_HOST OFF` → `MPI_HOST_NUMPROC` is empty → no unrecognized `--host` option passed to `srun`.
- Local non-root with Open MPI: The `else`-branch detects `--host` in the `mpiexec help` output → `MPIEXEC_SUPPORTS_HOST ON` → flag is applied correctly.

EZoni:
> I can verify that that this PR fixes the issue when running ctests locally on Dane.

@JustinRayAngus 

Would it be possible for you to test the latest version of this branch on Dane?

EZoni:
@WeiqunZhang 

What do you think of the changes in this PR? It came up again while trying to fix the new GPU tests in #6980, so I think we should try to finalize it sooner or later.

WeiqunZhang:
I think we could use environment variables instead of `--host` to work around the OpenMP oversubscribe issue.
```
export OMPI_MCA_rmaps_base_oversubscribe=1
export PRTE_MCA_rmaps_default_mapping_policy=:oversubscribe
```

WeiqunZhang:
The first line is for OpenMPI 4 and the second for OpenMPI 5.

EZoni:
@WeiqunZhang 

I pushed a commit to try your suggestion that we avoid `--host` entirely. Let's see if the Azure pipelines pass.

EZoni:
@WeiqunZhang 

The Azure pipelines passed successfully, so I think we could go ahead with this new workaround, which will in turn avoid entirely the follow-up issue encountered by Justin. 

Please feel free to review again and I'll merge the PR once approved, so we can rebase #6980 and hopefully clear the GPU CI pipelines there.

WeiqunZhang (approved):

Examples/CMakeLists.txt
```diff
@@ -108,16 +108,10 @@ function(add_warpx_test
         return()
     endif()
 
-    # Avoid the following error seen with Open MPI 4.1.6 in the Azure pipelines:
-    # "There are not enough slots available in the system to satisfy the 2 slots
-    # that were requested by the application"
-    set(MPI_HOST_NUMPROC --host localhost:${nprocs})
```
ax3l, line 114: Oh, the env vars below are way better.
This old logic should not ship to users that are on more general systems.

Issue #5984, opened 2025-06-29
PhysicalParticleContainer.cpp takes a long time to compile


AlexanderSinn:
I compiled a clean build of WarpX with 50 threads using HIP clang and ninja (which is 25% faster than the default cmake) and generated a trace using [these](https://gist.github.com/jrhemstad/811d1e41af9685670241fa472170ef32) instructions. The `PhysicalParticleContainer.cpp` file accounts for a large chunk of the time due to taking the longest of all files and starting rather late (I suspect because it comes late in alphabetical order). I tried to profile the file individually using `-ftime-trace` however, this only traces the host compiler and not the offload GPU compiler, which takes most of the time.

It would be good to split some of the large kernels like push and AddPlasma out of `PhysicalParticleContainer.cpp` into separate new .cpp files to greatly reduce compile times when using many threads.

![Image](https://github.com/user-attachments/assets/b26d5c12-67a5-43b6-8345-d352de728255)

ax3l:
That makes perfect sense. I often compile on HPC with `-j 64` on exclusive nodes and this clearly is on the critical path.

From similar PRs, this looks like something that @lucafedeli88 might enjoy optimizing? :)

AlexanderSinn:
Yes I was also thinking of Luca.

Here is a trace of `PhysicalParticleContainer.cpp` (adding `export CXXFLAGS="-ftime-trace"`). It only accounts for 23 seconds of the 2min 26sec time since the GPU compile pass is missing.

![Image](https://github.com/user-attachments/assets/2c60eaa2-493f-4784-9ca7-96a96002f399)

And here is a trace for HiPACE++, which compiles much faster than WaprX.

![Image](https://github.com/user-attachments/assets/bdf4ea3b-49c0-47df-8c64-2fe006eeb07f)

dpgrote:
I can start this in my PR #5969 since I add another large routine to that file. I can move the implicit related routines to a new file.

ax3l:
Thank you, @dpgrote ! Discussed in person: in a separate PR: `AddPlasma`, `PushPX` and `ImplicitPushPx` (#5969) would be good to into individial `.cpp` files.

ax3l:
I will take on the openPMD include clean-up.

ax3l:
openPMD include optimized in #6016 :)

franzpoeschel:
> Yes I was also thinking of Luca.
> 
[...]

PR #6016, opened 2025-07-12
`PlasmaInjector.H`: Hide `openPMD.hpp`

ax3l:
Avoid the `openPMD/openPMD.hpp` include in `PlasmaInjector.H`, because the latter file is included in tons of translation units, and the former is an "expensive" (~1.5sec) include, in terms of compile-time #5984.

Use a PIMPL-like idom with `std::any` to still store the series as a member variable, but only pull in the type in the two specific translation units, `AddParticles.cpp` and `PlasmaInjector.cpp` that access the member.

All other openPMD includes look good and TU local.

### DIFF
--- a/Source/Initialization/PlasmaInjector.H
+++ b/Source/Initialization/PlasmaInjector.H
@@ -25,10 +25,7 @@
 
 #include <AMReX_BaseFwd.H>
 
-#ifdef WARPX_USE_OPENPMD
-#   include <openPMD/openPMD.hpp>
-#endif
-
+#include <any>
 #include <limits>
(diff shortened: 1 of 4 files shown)


AlexanderSinn:
Alternatively, you could also forward declare openPMD::Series, but then all the constructors of PlasmaInjector would need to be put in the cpp file.

ax3l:
That is true, usually depends a bit if I need a concrete type of the class that uses the member (e.g., in unique_ptr, etc.)

EZoni (approved):

PR #6018, opened 2025-07-13
Update perlmutter.rst

truewis:
source environment does not work, at least in some cases. Since we created environment using conda above, use conda to activate the environment.


ax3l (commented):
I usually intentionally use `source activate` over conda activate, but I think here it should be available either way.

Can you comment a bit more when the problem showed up for you and how it manifested (i.e. error messages/screenshots)?

ax3l (approved):

PR #6021, opened 2025-07-14
Add full mass matrices for charge-conserving Villasenor deposition

JustinRayAngus:
This is a follow-on PR to PR #5929, where the infrastructure for using the full mass matrices with the particle-suppressed JFNK nonlinear solver was implemented, but was only setup to be used with direct deposition. This PR adds the ability to use the full mass matrices with the charge-conserving Villasenor deposition.

Two CI tests that use the full mass matrices with Villasenor deposition are added. The tests are a dynamic pinch in 1D and 2D planar geometry.

- [x] Merge PR #5909 that fixes boundary issue needed to reveal exact charge conservation in the 2D planar pinch CI test.
- [x] Merge PR #6060 that fixes the Poynting flux diagnostic as needed to reveal exact energy conservation in the 2D planar pinch CI test.


Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
```diff
@@ -478,8 +481,70 @@ void ImplicitSolver::InitializeMassMatrices ()
...
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                shape==2,
+                "Mass Matrices for Jacobian with Villasenor deposition requires shape = 2.");
+#ifndef WARPX_DIM_3D
+            int max_crossings = ngJ[0] - 1;
+            AMREX_ASSERT(max_crossings>0);
+            AMREX_ASSERT(max_crossings==m_WarpX->particle_max_grid_crossings);
+#endif
+#if defined(WARPX_DIM_1D_Z)
+            m_ncomp_xx[0] = 5 + 2*max_crossings;
```
dpgrote: Can you add a comment explaining where the numbers 3, 4, and 5 come from, that they are from the shape factor and stencil?

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
```diff
@@ -478,8 +481,70 @@ void ImplicitSolver::InitializeMassMatrices ()
                 Nc_tot_zz *= m_ncomp_zz[dir];
             }
         }
+        else if (m_WarpX->current_deposition_algo == CurrentDepositionAlgo::Villasenor) {
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+                shape==2,
+                "Mass Matrices for Jacobian with Villasenor deposition requires shape = 2.");
+#ifndef WARPX_DIM_3D
+            int max_crossings = ngJ[0] - 1;
```
dpgrote: Is 3D not supported here?

Source/Particles/Deposition/MassMatricesDeposition.H
```diff
@@ -1114,6 +1167,147 @@ void doVillasenorJandSigmaDepositionKernel ( [[maybe_unused]] const amrex::Parti
 
     } // end loop over segments
 
+    if constexpr (full_mass_matrices) {
+
+    // Loop over segments and deposit full mass matrices
+    for (int ns=0; ns<num_segments; ns++) {
+
+        // Deposit Sxx, Sxz, and Sxy for this segment
```
dpgrote, line 1173: There is a fair amount of duplicated code here, both the outer and inner loops. Could a kernel be written that can be used for each case? Maybe it's not worth it, or could be a follow on PR for code cleaning.

Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -1154,48 +1154,70 @@ WarpXParticleContainer::DepositCurrentAndMassMatrices ( WarpXParIter& pti, const
...
-                Sxx_arr, Syy_arr, Szz_arr,
+                jx_arr, jy_arr, jz_arr, WarpX::particle_max_grid_crossings,
+                Sxx_arr, Sxy_arr, Sxz_arr,
+                Syx_arr, Syy_arr, Syz_arr,
+                Szx_arr, Szy_arr, Szz_arr,
                 Bx_arr, By_arr, Bz_arr, Bx_type, By_type, Bz_type,
                 np_to_deposit, dt, dinv, xyzmin, lo, qs, ms);
-        } else if (WarpX::nox == 2){
-            doVillasenorJandSigmaDeposition<2>(
+        } else if (WarpX::nox == 2 && Szz_nComp==1){
```
dpgrote: Testing on `Szz_nComp` is an indirect test. Could this instead directly check the input parameter `m_use_mass_matrices_jacobian` (it would have to be a local copy of it for the GPU)? This might be more clear about what is happening here.

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
JustinRayAngus: Yes, using mass matrices for the Jacobian calculation is not supported in 3D. It could be if needed in the future, but the thought is that the memory overhead for storing the mass matrices is too high in 3D. 

There is an assert for this in the .H file when the options are parsed.

The diagonal elements of the mass matrices can still be used for the PC in 3D.

Source/Particles/Deposition/MassMatricesDeposition.H
JustinRayAngus: Exactly what part of the coding are you referring to that is duplicated?

What are the different cases you are referring to?

Source/Particles/WarpXParticleContainer.cpp
JustinRayAngus: It is an indirect check, but it's sufficient. I modified this section to make the true/false full_mass_matrices template parameter more explicit.

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
JustinRayAngus: I have now extended the method to work for arbitrary shape.

Source/Particles/Deposition/MassMatricesDeposition.H
dpgrote: I'm referring to the code that sets the Sx[xyz], Sy[xyz], and Sz[xyz]. The loops for each of those look similar to each other, the two double loops over the depos_order, with of course small differences. If there are too many differences, then it wouldn't be worth worrying about.

Source/Particles/Deposition/MassMatricesDeposition.H
```diff
@@ -1049,8 +1078,16 @@ void doVillasenorJandSigmaDepositionKernel ( [[maybe_unused]] const amrex::Parti
         double sz_cell[depos_order] = {0.};
         double const x0_bar = (x0_new + x0_old)/2.0;
         double const z0_bar = (z0_new + z0_old)/2.0;
-        const int i0_cell = compute_shape_factor_cell( sx_cell, x0_bar-0.5 );
-        const int k0_cell = compute_shape_factor_cell( sz_cell, z0_bar-0.5 );
+        i0_cell[ns] = compute_shape_factor_cell( sx_cell, x0_bar-0.5 );
+        k0_cell[ns] = compute_shape_factor_cell( sz_cell, z0_bar-0.5 );
```
ax3l, line 1079: Already in the PR before and can be fixed in a follow-up PR: it would be good to write `double` as `amrex::Real` and literals as `0.5_rt`. We default to `double` anyway.

ax3l: (might just need an inline comment to mark as critical to be dp for now)

ax3l (approved):
Clarified my questions via our weekly meeting.

Depends on a PR by @dpgrote so I will let him do the final merge on CI passes.

dpgrote (approved):
Looks good!

Source/Particles/Deposition/MassMatricesDeposition.H
JustinRayAngus: Positions used for interp are kept double to "avoid bug in single precision". At least that is the comment given elsewhere in the deposition routines where this is done. I added the comment here for future reference.

PR #6033, opened 2025-07-21
Add `--release` option to script for update of dependencies

EZoni:
Added the new command line option `--release` to the script `update_dependencies.py` which automatically uses the version tag (e.g., `25.07`) instead of the commit hash to replace the commit field in the dependencies file.

Requested by @ax3l on Slack:

> Can we slightly update the version script to support to put tags into commit fields, when we have a tag?
I added this in the last commit. Whenever we can add a tag it is just more readable for this field, so I used tags instead of hashes for releases usually in the past. In between, of course we use hashes for weekly/intermediate updates.

@ax3l 
Does this implementation work for you?

### DIFF
--- a/Tools/Release/releasePR.py
+++ b/Tools/Release/releasePR.py
@@ -74,7 +74,11 @@ AMReX_version = f"{datetime.now().strftime('%y')}.{datetime.now().strftime('%m')
 answers = concat_answers(["y", AMReX_version, AMReX_version, "y"])
 
 process = subprocess.Popen(
-    [Path(REPO_DIR).joinpath("Tools/Release/update_dependencies.py"), "--amrex"],
+    [
+        Path(REPO_DIR).joinpath("Tools/Release/update_dependencies.py"),
+        "--amrex",
+        "--release",
+    ],
     stdin=subprocess.PIPE,
     stdout=subprocess.PIPE,
     stderr=subprocess.PIPE,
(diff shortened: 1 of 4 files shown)


Tools/Release/releasePR.py
```diff
@@ -174,11 +190,11 @@ def concat_answers(answers):
...
 # update dependencies
-./Tools/Release/update_dependencies.py --amrex
-./Tools/Release/update_dependencies.py --picsar # no changes, still {PICSAR_version}
-./Tools/Release/update_dependencies.py --pyamrex
+./Tools/Release/update_dependencies.py --amrex --release
+./Tools/Release/update_dependencies.py --picsar --release # no changes, still {PICSAR_version}
+./Tools/Release/update_dependencies.py --pyamrex --release
 # bump version number
-./Tools/Release/update_dependencies.py --warpx
+./Tools/Release/update_dependencies.py --warpx --release
```
ax3l, line 197: Can you again add a an extra line at the end of commit messages that list the tool command that was used? That makes it easier for others to follow and use the same script when they read our PRs and try to do the same.

Example: https://github.com/BLAST-WarpX/warpx/pull/5948

EZoni: This is what we had originally and I have not changed it so far:
https://github.com/BLAST-WarpX/warpx/blob/928d867485ac000faeb37ab25ffb0d4ae2272d59/Tools/Release/releasePR.py#L184

I thought this was enough. You mean you would like to change it by saying something like this instead?
```
This pull request was created with this script: `./Tools/Release/releasePR.py`
```

ax3l: Yeah I did something like the latter as a last line (besides the link), because it is easy to copy-paste next time.

ax3l (approved):
Approving already, minor suggestion.

PR #6037, opened 2025-07-23
feat: quartz boundary test, analysis, and skip divergence cleaner for quartz

Miracle-Ein:
- Add 3D quartz boundary condition test and analysis script
     - Skip divergence cleaner (ProjectionDivCleaner) for quartz boundaries in the code
     - Update README with test and verification instructions
     - The quartz boundary condition is implemented with custom logic, and its numerical accuracy requires further validation. Reviews and testing are welcome.


ax3l (changes requested):
Please remove the `diags/` folders from an PR. They are binary output data and should not be committed.

ax3l:
replaced by #6056 ?

PR #6038, opened 2025-07-25
Tools: improve automated dependencies update

EZoni:
### Overview

Follow up on #5965 and improve further the automated update of dependencies:
- The script chooses whether to update or skip the update automatically, user input (`[y/n]`) not requested anymore.
- The script fetches both the latest commit and the list of available version tags (to be used for a new release update).
- For dependencies, we use the latest available version tag (instead of the current date YY.MM), as not all dependencies publish regular monthly releases (e.g., PICSAR). The script takes care of filtering out old-format tags as well.

### To do

- [x] Rebase after #6033 

### Coming next

Try setting up a GitHub Actions workflow for fully automated weekly update and monthly release, using GitHub CLI as in https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/using-github-cli-in-workflows.

### DIFF
--- a/Tools/Release/update_dependencies.py
+++ b/Tools/Release/update_dependencies.py
@@ -100,7 +100,7 @@ def update(args):
             # current date version for the WarpX release update
             repo_version_tag = datetime.date.today().strftime("%y.%m")
         else:
-            # last available tag (index 0) for all other dependencies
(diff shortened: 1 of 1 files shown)


ax3l (approved):
Nice, appreciate it!

PR #6042, opened 2025-07-26
[WIP] [DO NOT MERGE] Test amrex PR with WarpX CI

AlexanderSinn:
Test to see if https://github.com/AMReX-Codes/amrex/pull/4529 passes WarpX CI


ax3l:
There is CI in AMReX to check if a PR breaks WarpX :)

AlexanderSinn:
That one just compiles. Here I was testing if the behavior of RedistributeCPU is correct.

PR #6047, opened 2025-07-29
Updating initialization of external vector potential based fields to …

clarkse:
…use t_new, since t_old is not set and t_new has not been incremented yet. Additionally added a check for finite fields.


roelof-groenewald (approved):

roelof-groenewald:
Looks like it just needs a few checksum updates.

PR #6052, opened 2025-07-29
Doxygen: No Python

ax3l:
Unbreak XML builds of Doxygen, following #3560.
Please do not read Python files, Doxygen. You are too eager.

See:
- https://github.com/BLAST-WarpX/warpx/pull/3560#issuecomment-3134161573

### DIFF
--- a/Docs/Doxyfile
+++ b/Docs/Doxyfile
@@ -2709,6 +2709,3 @@ GENERATE_LEGEND        = YES
 # The default value is: YES.
 
 DOT_CLEANUP            = YES
-
-# Overwrite value of WARN_AS_ERROR
-WARN_AS_ERROR = YES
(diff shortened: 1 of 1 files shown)


ax3l:
Oh, and there is more:
```
Gener/home/docs/checkouts/readthedocs.org/user_builds/warpx/checkouts/6052/Source/FieldSolver/FiniteDifferenceSolver/HybridPICSolveE.cpp:73: error: The following parameters of FiniteDifferenceSolver::CalculateCurrentAmpereCylindrical(ablastr::fields::VectorField &Jfield, ablastr::fields::VectorField const &Bfield, std::array< std::unique_ptr< amrex::iMultiFab >, 3 > const &eb_update_E, int lev) are not documented:
  parameter 'eb_update_E'
  parameter 'lev' (warning treated as error, aborting now)
```

roelof-groenewald (approved):
Thanks for fixing!

ax3l:
Merging to unbreak CI in ImpactX.

PR #6053, opened 2025-07-29
openPMD: Species Temperature Dimensionality

### LINKED ISSUE #6039: Temperature missing dimensions in output data
## Description
It looks to me that the temperature field outputted using, for instance `T_Electrons`, doesn't carry the `unit_dimension` information in openPMD.
It's a very minor issue, but just wanted to report it.

## How to reproduce

```
temperature = series.iterations[chosen_iteration].meshes["T_Electrons"]
print(temperature.unit_dimension)
[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
```

ax3l:
- [x] Properly add the meta data for thermodynamic temperature to the `T_<species>` field outputs.
- [x] Set `unitSI` to a non-1 value, because we store in eV.

Fix #6039

### DIFF
--- a/Source/Diagnostics/WarpXOpenPMD.cpp
+++ b/Source/Diagnostics/WarpXOpenPMD.cpp
@@ -371,6 +371,10 @@ namespace detail
                                           {openPMD::UnitDimension::I,  1},
                                           {openPMD::UnitDimension::T,  1},
                                   });
+        } else if (field_name.substr(0,2) == "T_"){ // temperature in eV
+            mesh.setUnitDimension({
+                                          {openPMD::UnitDimension::theta,  1},
+                                  });
         }
     }
 #endif // WARPX_USE_OPENPMD
--- a/Source/Diagnostics/WarpXOpenPMD.cpp
+++ b/Source/Diagnostics/WarpXOpenPMD.cpp
@@ -1471,6 +1471,9 @@ WarpXOpenPMDPlot::WriteOpenPMDFieldsAll ( //const std::string& filename,
                                         field_name,
                                         mf[lev],
                                         var_in_theta_mode );
+                        if (field_name.substr(0,2) == "T_") {
+                            mesh.setUnitSI(11604.5);  // temperature in eV -> K
+                        }
                     }
                 } else {
                     auto mesh = meshes[field_name];


roelof-groenewald (approved):

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -371,6 +371,10 @@ namespace detail
                                           {openPMD::UnitDimension::I,  1},
                                           {openPMD::UnitDimension::T,  1},
                                   });
+        } else if (field_name.substr(0,2) == "T_"){ // temperature in eV
+            mesh.setUnitDimension({
+                                          {openPMD::UnitDimension::theta,  1},
```
titoiride, line 376: Can I ask the rationale behind defining a special unit for temperature? If openPMD expresses quantities with respect to SI, shouldn't this be eV [energy]/[charge]-> [mass][length]^2[time]^-1[current]^-1?

titoiride: Oh sorry, I see where I'm confused: because temperature in eV is actually kT it should be a measure of energy, but the intended quantity is a temperature, so it's correct to use K, i.e. `openPMD::UnitDimension::theta`, as base unit, my bad.

titoiride (approved):

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -1467,6 +1471,9 @@ WarpXOpenPMDPlot::WriteOpenPMDFieldsAll ( //const std::string& filename,
                                         field_name,
                                         mf[lev],
                                         var_in_theta_mode );
+                        if (field_name.substr(0,2) == "T_") {
+                            mesh.setUnitSI(11604.5);  // temperature in eV -> K
```
ax3l, line 1475: @roelof-groenewald suggested a better solution: calculate directly in Kelvin, so plotfiles & openPMD files have the same numerical values when read (e.g., in checksums)

EZoni:
Just checking, what is the status of this PR? 

It's labeled as a bug fix, so I wonder if we want to finalize it before the September release. 

It may simply need a rebase and a checksums update - happy to do that and push to the branch if it helps.

roelof-groenewald:
It would be good to finalize. We decided at the end that if we want to add the openPMD "dimensionality" attribute, we should switch the temperature diagnostic to output in Kelvin (rather than eV), otherwise the output obtained from openPMD vs plotfile would not be the same. This would break existing analysis using the temperature diagnostic though, so we should communicate this clearly to users.
@dpgrote since you added the temperature diagnostic, do you have thoughts on the above?
(cc @clarkse)

EZoni:
Hi @dpgrote, do you have any feedback on this PR and/or the last comment https://github.com/BLAST-WarpX/warpx/pull/6053#issuecomment-3255569948 in particular?

dpgrote:
I don't have a strong opinion on this. Note that now the plotfile outputs the temperature in eV. Looking through the documentation, the temperature is used in both Kelvin and eV in various places. I don't know if it is important to be consistent, as long as it is documented. My only hesitation is that it is better to not change the code behavior that affects users without good reason.

PR #6056, opened 2025-07-31
[WIP] Quartz Boundary Conditions

Miracle-Ein:
Implement boundary conditions for Quartz.


warpx_used_inputs
```diff
@@ -0,0 +1,112 @@
+DistributionMapping.verbose = 0
```
ax3l, line 1: Please do not add this file, it is an output file.

Source/BoundaryConditions/Quartz.H
```diff
@@ -0,0 +1,61 @@
...
+    std::string m_str_mu_r_function;
+    std::string m_str_sigma_function;
+    amrex::Real m_ring_inner_radius = 0.05;
+    amrex::Real m_ring_outer_radius = 0.08;
+    amrex::Real m_ring_height = 0.02;
+    amrex::Real m_ring_center_x = 0.0;
+    amrex::Real m_ring_center_y = 0.0;
+    amrex::Real m_ring_bottom_z = 0.0;
+    std::unique_ptr<amrex::Parser> m_ring_geometry_parser;
+    std::string m_str_ring_geometry_function;
```
ax3l, line 58: The approach here seems to build a custom boundary (hard-codes a ring?) and apply the Quartz boundary conditions to it.

This is very specific and in the current form not well suited for mainline WarpX inclusion.

One way to make this suitable for mainline, we would need a more general approach here. For instance, we have a mechanism for embedded boundaries in WarpX, where one could apply more boundary conditions on. Embedded boundaries can have arbitrary shape and be used to describe the ring shape.

At the moment, we only use the embedded boundaries to set electrostatic potentials on them. One would need to first generalize them to accept further boundary conditions (like the ones here, applied to the domain boundary on the outer surface of the simulation), and then add the Quartz boundary conditions there, too.

PR #6057, opened 2025-07-31
CMake: Add `WarpX_FASTMATH`

ax3l:
... and `ABLASTR_FASTMATH` options.

Note: on Nvidia-GPU, we still auto-enable fast-math via `AMReX_CUDA_FASTMATH`.
We can unify that in a follow-up PR, have not discussed/decided this yet.

- [x] manually check the CUDA build is still doing the fast-math build (on device only!) by default, while all others don't
- [x] depends on https://github.com/AMReX-Codes/amrex/pull/4545
- [x] depends on https://github.com/AMReX-Codes/amrex/pull/4591 / #6059


cmake/dependencies/AMReX.cmake
```diff
@@ -53,6 +53,16 @@ macro(find_amrex)
...
         set(AMReX_SIMD "${WarpX_SIMD}" CACHE INTERNAL "")
 
+        if(WarpX_FASTMATH OR ABLASTR_FASTMATH)
+            set(AMReX_FASTMATH ON CACHE INTERNAL "")
+            # TODO: set consistently (default: ON)
+            # set(AMReX_CUDA_FASTMATH ON CACHE INTERNAL "")
+        else()
+            # TODO: set consistently (default: ON)
+            # set(AMReX_CUDA_FASTMATH ON CACHE INTERNAL "")
+            set(AMReX_FASTMATH OFF CACHE INTERNAL "")
```
WeiqunZhang, line 63: I thin setting AMReX_FASTMATH to OFF will change the default because it will override AMReX_CUDA_FASTMATH.

ax3l:
@WeiqunZhang @EZoni I did the check we discussed checking the local flags passed to the CUDA/CXX compiler and it looks good on my end (no defaults changed).

CMakeLists.txt
```diff
@@ -77,6 +77,7 @@ option(WarpX_LIB           "Build WarpX as a library"                   OFF)
 option(WarpX_MPI           "Multi-node support (message-passing)"       ON)
 option(WarpX_SIMD          "CPU SIMD Acceleration"                      OFF)
 option(WarpX_OPENPMD       "openPMD I/O (HDF5, ADIOS)"                  ON)
+option(WarpX_FASTMATH      "Enable fast-math optimizations"             OFF)  # TODO: set to ON by default!
```
EZoni: The `TODO` comment here may be outdated based on our discussion this morning, but up to you whether you want to keep it as is, change it, or remove it.

cmake/dependencies/AMReX.cmake
```diff
@@ -53,6 +53,16 @@ macro(find_amrex)
 
         set(AMReX_SIMD "${WarpX_SIMD}" CACHE INTERNAL "")
 
+        if(WarpX_FASTMATH OR ABLASTR_FASTMATH)
+            set(AMReX_FASTMATH ON CACHE INTERNAL "")
+            # TODO: set consistently (default: ON)
+            # set(AMReX_CUDA_FASTMATH ON CACHE INTERNAL "")
+        else()
+            # TODO: set consistently (default: ON)
+            # set(AMReX_CUDA_FASTMATH ON CACHE INTERNAL "")
```
EZoni, line 62: Just to double check, the CUDA change to do here would be to intentionally enable this even when both `WarpX_FASTMATH` and `ABLASTR_FASTMATH` are `OFF` (i.e., to intentionally avoid disabling it), do I understand this correctly?

CMakeLists.txt
ax3l: Thanks! Fixed

cmake/dependencies/AMReX.cmake
ax3l: So currently, this `AMReX_CUDA_FASTMATH` is always `ON`, no matter what we set for `AMReX_FASTMATH` or the new `WarpX_FASTMATH OR ABLASTR_FASTMATH`.

The TODO comment is to remember that we want to make this consistently in the future.

ax3l: We can, if we want, with this PR change this to disable fast-math on CUDA by default.

I suspect that some of the long-running KISMET sims on Lassen/Perlmutter might actually see heating from that...

EZoni (approved):

PR #6058, opened 2025-07-31
[WIP] Workaround for GCC warray-bounds warnings

WeiqunZhang:
These warnings are false positives. This PR provides a workaround.

### DIFF
--- a/Source/Diagnostics/BTDiagnostics.cpp
+++ b/Source/Diagnostics/BTDiagnostics.cpp
@@ -1342,7 +1342,7 @@ BTDiagnostics::InterleaveBufferAndSnapshotHeader ( const std::string& buffer_Hea
     amrex::IntVect box_lo(0);
     amrex::IntVect box_hi(1);
     // Update prob_lo with min of buffer and snapshot
-    for (int idim = 0; idim < snapshot_HeaderImpl.spaceDim(); ++idim) {
+    for (int idim = 0; idim < std::min(snapshot_HeaderImpl.spaceDim(),AMREX_SPACEDIM); ++idim) {
         const amrex::Real min_prob_lo = amrex::min(buffer_HeaderImpl.problo(idim),
                                              snapshot_HeaderImpl.problo(idim));
         const amrex::Real max_prob_hi = amrex::max(buffer_HeaderImpl.probhi(idim),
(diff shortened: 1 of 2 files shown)


EZoni:
Here are other solutions proposed on Slack:

- From @ax3l: Add `Wno-array-bounds` to the list here
https://github.com/BLAST-WarpX/warpx/blob/6ae81f30ca811df016e4b86ee84c20e86e555612/cmake/WarpXFunctions.cmake#L177-L208

- From @WeiqunZhang: Do something along the lines of
    ```cpp
    #if defined(__GNUC__) && !defined(__clang__)
    #pragma GCC diagnostic push
    #pragma GCC diagnostic ignored "-Warray-bounds"
    #endif
    // code with false positive
    #if defined(__GNUC__) && !defined(__clang__)
    #pragma GCC diagnostic pop
    #endif
    ```

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -1342,7 +1342,7 @@ BTDiagnostics::InterleaveBufferAndSnapshotHeader ( const std::string& buffer_Hea
     amrex::IntVect box_lo(0);
     amrex::IntVect box_hi(1);
     // Update prob_lo with min of buffer and snapshot
-    for (int idim = 0; idim < snapshot_HeaderImpl.spaceDim(); ++idim) {
+    for (int idim = 0; idim < std::min(snapshot_HeaderImpl.spaceDim(),AMREX_SPACEDIM); ++idim) {
```
ax3l, line 1345: Good solution, but a bit cryptic :)

Could be a separate variable before the loop with an inline comment why we do this...

Or we just extend the compiler flags...

ax3l (changes requested):
My preference would be to just add `-Wno-array-bounds` for GCC to our `-Wall -Wextra` for now.

EZoni:
@WeiqunZhang 

Do you have objections against @ax3l's preference for adding `-Wno-array-bounds` to our GCC flags for now, https://github.com/BLAST-WarpX/warpx/pull/6058#pullrequestreview-3089899154, e.g., you think we may miss too many real positives (instead of false positives)?

If you have objections, I can push a commit to apply the other suggestion from @ax3l, https://github.com/BLAST-WarpX/warpx/pull/6058#discussion_r2255474269, and merge this PR as is, postponing the decision on the GCC flags to another time.

WeiqunZhang:
I have no objects to adding `-Wno-array-bounds`.

EZoni:
Okay, thanks! I will open a separate PR to do that, so we keep this one clean. Will close it later if we merge the other one.

EZoni:
Replaced by #6088.

PR #6060, opened 2025-08-01
Update to use the correct stencils in FieldPoyntingFlux

dpgrote:
It was pointed out by @JustinRayAngus that the stencil used to calculate the Poynting flux was not consistent with the finite difference Yee staggering and was not calculating the exact energy flowing out of the grid. This was interfering the energy calculation when exact energy conservation is being checked. This PR fixes this by using the correct stencil. This complicated the code since other stencils are needed depending on the grid centering.

Note that the CI tests in PR #6021 will depend on this PR so a new CI test is not needed here.

### DIFF
--- a/Source/Diagnostics/ReducedDiags/FieldPoyntingFlux.H
+++ b/Source/Diagnostics/ReducedDiags/FieldPoyntingFlux.H
@@ -343,6 +343,10 @@ namespace Poynting {
 
                 constexpr int comp = 0;
 
+                // This is needed by the GPU compilers where captured variables cannot first appear
+                // in a constexpr-if context
+                amrex::ignore_unused(Ez_arr, By_arr, Ex_arr, Bz_arr, Ey_arr, Bx_arr);
+
                 amrex::Real const af = area_factor(i,j,k);
                 if constexpr (normal_dir == 0) {
                     return af*(T_Algo::EyBz(i,j,k,comp,Ey_arr,Bz_arr) - T_Algo::EzBy(i,j,k,comp,Ez_arr,By_arr));
--- a/Source/Diagnostics/ReducedDiags/FieldPoyntingFlux.H
+++ b/Source/Diagnostics/ReducedDiags/FieldPoyntingFlux.H
@@ -10,7 +10,358 @@
 
 #include "ReducedDiags.H"
 
+#include <AMReX_Array4.H>
+#include <AMReX_GpuControl.H>
+#include <AMReX_Box.H>
+#include <AMReX_REAL.H>
+#include <AMReX_Reduce.H>
+
 #include <string>
+
+using namespace amrex::literals;
+
+struct PoyntingStaggered {
+    // This requires the arrays to have Yee centering
+
+    AMREX_GPU_HOST_DEVICE AMREX_INLINE
+    static amrex::Real EyBx(int i, int j, int k, int comp,
+                            amrex::Array4<const amrex::Real> const & Ey_arr,
+                            amrex::Array4<const amrex::Real> const & Bx_arr) {
+#if defined(WARPX_DIM_3D)
+        amrex::Real EyBx_dn = Ey_arr(i,j,k,comp)*0.5_rt*(Bx_arr(i,j,k-1,comp) + Bx_arr(i,j,k,comp));
+        amrex::Real EyBx_up = Ey_arr(i+1,j,k,comp)*0.5_rt*(Bx_arr(i+1,j,k-1,comp) + Bx_arr(i+1,j,k,comp));
+        return 0.5_rt*(EyBx_dn + EyBx_up);
+#elif defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
+        amrex::Real EyBx_dn = Ey_arr(i,j,k,comp)*0.5_rt*(Bx_arr(i,j-1,k,comp) + Bx_arr(i,j,k,comp));
+        amrex::Real EyBx_up = Ey_arr(i+1,j,k,comp)*0.5_rt*(Bx_arr(i+1,j-1,k,comp) + Bx_arr(i+1,j,k,comp));
+        return 0.5_rt*(EyBx_dn + EyBx_up);
+#elif defined(WARPX_DIM_1D_Z)
+        return Ey_arr(i,j,k,comp)*0.5_rt*(Bx_arr(i-1,j,k,comp) + Bx_arr(i,j,k,comp));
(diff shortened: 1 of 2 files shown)


Source/Diagnostics/ReducedDiags/FieldPoyntingFlux.cpp
```diff
@@ -262,38 +253,46 @@ void FieldPoyntingFlux::ComputePoyntingFlux ()
...
+                        flux += Poynting::Kernel<0, PoyntingStaggered>(box, Ex_arr, Ey_arr, Ez_arr, Bx_arr, By_arr, Bz_arr, area_factor);
+                    }
+                    else if (normal_dir == 1) {
+                        flux += Poynting::Kernel<1, PoyntingStaggered>(box, Ex_arr, Ey_arr, Ez_arr, Bx_arr, By_arr, Bz_arr, area_factor);
+                    }
+                    else if (normal_dir == 2) {
+                        flux += Poynting::Kernel<2, PoyntingStaggered>(box, Ex_arr, Ey_arr, Ez_arr, Bx_arr, By_arr, Bz_arr, area_factor);
+                    }
+                }
+                else if (Ex.is_nodal()) {
```
JustinRayAngus: Maybe use warpx.grid_type == ablastr::utils::enums::GridType::Collocated && Ex.is_nodal() here for clarity. It will work as is, but it is possibly error prone. For 1D CYL and 1D SPH, Ex is centered, while Ey and Ez are nodal. For 1D Z, Ex and Ey are nodal, while Ez is centered.

Source/Diagnostics/ReducedDiags/FieldPoyntingFlux.cpp
```diff
@@ -262,38 +253,46 @@ void FieldPoyntingFlux::ComputePoyntingFlux ()
...
+                        flux += Poynting::Kernel<0, PoyntingNodal>(box, Ex_arr, Ey_arr, Ez_arr, Bx_arr, By_arr, Bz_arr, area_factor);
+                    }
+                    else if (normal_dir == 1) {
+                        flux += Poynting::Kernel<1, PoyntingNodal>(box, Ex_arr, Ey_arr, Ez_arr, Bx_arr, By_arr, Bz_arr, area_factor);
+                    }
+                    else if (normal_dir == 2) {
+                        flux += Poynting::Kernel<2, PoyntingNodal>(box, Ex_arr, Ey_arr, Ez_arr, Bx_arr, By_arr, Bz_arr, area_factor);
+                    }
+                }
+                else if (Ex.is_cell_centered()) {
```
JustinRayAngus: Same comment here. Consider using warpx.grid_type == ablastr::utils::enums::GridType::Collocated && Ex.is_cell_centered() for clarity.

Source/Diagnostics/ReducedDiags/FieldPoyntingFlux.H
```diff
@@ -10,8 +10,363 @@
...
+#include <AMReX_REAL.H>
+#include <AMReX_Reduce.H>
+
 #include <string>
 
+using namespace amrex::literals;
+
+struct PoyntingStaggered {
+    // This requires the arrays to have Yee centering
+
```
JustinRayAngus: I think it would be good to add a comment here with an explanation of the logic, which is to 1) interpolate Bj to Ei location, then 2) interpolate the product Ei*Bj to a location that is nodal in k-direction, but centered in i and j.

Source/Diagnostics/ReducedDiags/FieldPoyntingFlux.cpp
dpgrote: I had originally had it checking the `grid_type` but wanted it to be more flexible. It is only the 1D_Z that has Ex nodal when staggered - I overlooked that. I'll change it back.
I originally added `GridType::CollocatedCell` for the RZ psatd but that was breaking things. That would need to be a separate PR if something like that was added.

Source/Diagnostics/ReducedDiags/FieldPoyntingFlux.H
```diff
@@ -10,8 +10,369 @@
...
+    AMREX_GPU_HOST_DEVICE AMREX_INLINE
+    static amrex::Real EyBx(int i, int j, int k, int comp,
+                            amrex::Array4<const amrex::Real> const & Ey_arr,
+                            amrex::Array4<const amrex::Real> const & Bx_arr) {
+#if defined(WARPX_DIM_3D)
+        return 0.25_rt*(Ey_arr(i,j,k,comp)*Bx_arr(i,j,k,comp)
+                      + Ey_arr(i+1,j,k,comp)*Bx_arr(i+1,j,k,comp)
+                      + Ey_arr(i,j+1,k,comp)*Bx_arr(i,j+1,k,comp)
+                      + Ey_arr(i+1,j+1,k,comp)*Bx_arr(i+1,j+1,k,comp));
```
JustinRayAngus, line 226: Here is a suggestion to consider. You could combine PoyntingCell and PoyntingNodal into one routine called PoyntingCollocated. You would just need to pass a binary nodal flag (0 for cell centered and 1 for node centered) and then you could compute the product as

```C++
return 0.25_rt*(Ey_arr(i,j,k-1+nodal,comp)*Bx_arr(i,j,k-1+nodal,comp)
              + Ey_arr(i+nodal,j,k-1+nodal,comp)*Bx_arr(i+nodal,j,k-1+nodal,comp)
              + Ey_arr(i,j+nodal,k,comp)*Bx_arr(i,j+nodal,k,comp)
              + Ey_arr(i+nodal,j+nodal,k,comp)*Bx_arr(i+nodal,j+nodal,k,comp));
```

and similarly for the other components

dpgrote: I don't think this works. The cell centered averages over only the two nearby cell centers.

JustinRayAngus: That is correct, but for nodal = 0 in the expression above, the first two values are the same and the third and fourth values are same, and this give the same result computed using the two nearby cell center values.

ax3l:
Hi @dpgrote, is this one still WIP? It seems to be a prerequisite to merging #6021?

Source/Diagnostics/ReducedDiags/FieldPoyntingFlux.H
dpgrote: Ok, I see. I would rather leave this as it is now to avoid extra work.

JustinRayAngus (commented):
Thanks Dave!

JustinRayAngus (approved):

PR #6062, opened 2025-08-02
Load density from openPMD file

WeiqunZhang:
This adds a new capability of loading density from an openPMD file. It works for initialization and continuous injection. It supports 3d, 2d rz, 2d xz and 1d z coordinates. The openPMD data must have the C order for multi-dimensional arrays. This restriction could be relaxed in the future. Currently, it loads all the data from the file. In future, we could try to load data by chunks.

This PR has also refactored an existing function `ReadExternalFieldFromFile` in `SpeciesUtils.cpp`. The openPMD part of that function has been moved to two new classes, `ExternalFieldReader` and `GetExternalFieldFromFile`. This allows us to reuse the openPMD reader in both the loading of external fields and loading of density. `ExternalFieldReader` reads an openPMD file and stores the data in either `std::shared_ptr<double>` for CPU or `amrex::Gpu::DeviceVector<double>` for GPU. `GetExternalFieldFromFile` is a lightweight class for viewing the data stored in `ExternalFieldReader`, and it is suitable for GPU kernels.


WeiqunZhang:
This supersedes #4754.

WeiqunZhang:
I believe the 2d rz CI failure is because the current development has a bug. The CI passed before the this commit https://github.com/BLAST-WarpX/warpx/pull/6062/commits/bbc5ad556eed9f50c0cc1664287ad6d5681abf7a.
```
commit bbc5ad556eed9f50c0cc1664287ad6d5681abf7a
Author: Weiqun Zhang <weiqunzhang@lbl.gov>
Date:   Fri Aug 1 23:47:58 2025 -0700

    Fix a preexisting bug

diff --git a/Source/Initialization/ExternalField.H b/Source/Initialization/ExternalField.H
index ed0ef723e..0cc950790 100644
--- a/Source/Initialization/ExternalField.H
+++ b/Source/Initialization/ExternalField.H
@@ -98,10 +98,8 @@ struct GetExternalFieldFromFile
             ablastr::math::bilinear_interp<double>(
                 xx0, xx0+dx[0], xx1, xx1+dx[1],
                 table(i0  ,i1  ),
-//                table(i0  ,i1+1),
-//                table(i0+1,i1  ),
-                table(i0+1,i1  ), // This is wrong order
                 table(i0  ,i1+1),
+                table(i0+1,i1  ),
                 table(i0+1,i1+1),
                 pos[0], pos[1]));
```

Docs/source/usage/parameters.rst
```diff
@@ -1158,6 +1158,12 @@ Particle initialization
       ``electrons.density_function(x,y,z) = "n0+n0*x**2*1.e12"`` where ``n0`` is a
       user-defined constant, see above. WARNING: where ``density_function(x,y,z)`` is close to zero, particles will still be injected between ``xmin`` and ``xmax`` etc., with a null weight. This is undesirable because it results in useless computing. To avoid this, see option ``density_min`` below.
 
+    * ``fromfile``: The density profile is computed from data in an openPMD
+      file. It requires additional parameters
+      ``<species_name>.read_density_from_path`` and
+      ``<species_name>.density_name`` specifying the file name and the
```
RemiLehe: I think that we could simplify the interface by imposing a fixed name (e.g. `density`), so that the user does not need to pass `density_name`. This is already what we do for the external fields (i.e. the name in the openPMD file has to be `E`/`B`)

WeiqunZhang: Sure we can do that. Or we can make `density` to be the default if density_name is not provided. Note that the test file I have uses `rho`.

RemiLehe: OK. I updated the test so that it creates an openPMD file that has the field `density` instead of `rho`. 
For the sake of simplicity, I think it would make sense to remove the `density_name` argument entirely. We can always add it later if we find a need, but I think that, for now, imposing that the openPMD file has `density` is fine for all expected use cases.

WeiqunZhang: Done

Examples/Tests/load_density/inputs_test_3d_load_density
```diff
@@ -0,0 +1,52 @@
...
+############ PLASMA #############
+#################################
+particles.species_names = electrons
+
+electrons.species_type = electron
+electrons.injection_style = "NUniformPerCell"
+electrons.num_particles_per_cell_each_dim = 1 1 2
+electrons.profile = "fromfile"
+electrons.read_density_from_path = "example-density.h5"
+electrons.density_name = "density"
```
RemiLehe:
```suggestion
```

Source/Initialization/ExternalField.cpp
```diff
@@ -182,3 +188,100 @@ ExternalFieldParams::ExternalFieldParams(const amrex::ParmParse& pp_warpx)
...
+    const openPMD::Extent chunk_extent = extent;
+
+    m_FC_data_cpu = FC.loadChunk<double>(chunk_offset,chunk_extent);
+    series.flush();
+
+#if defined(AMREX_USE_GPU)
+    auto *FC_data_host = m_FC_data_cpu.get();
+    auto const total_extent = std::accumulate(chunk_extent.begin(), chunk_extent.end(),
+                                              1, std::multiplies<std::size_t>());
+    m_FC_data_gpu.resize(total_extent);
```
RemiLehe, line 255: Just checking my understanding here: this moves the entirety of the data contained on file to the GPU.

Source/Initialization/ExternalField.H
```diff
@@ -67,4 +71,97 @@ struct ExternalFieldParams
...
+        }
+#endif
+    }
+
+    amrex::RealVect dx;
+    amrex::RealVect offset;
+#if (AMREX_SPACEDIM == 1)
+    amrex::Table1D<double> table;
+#elif (AMREX_SPACEDIM == 2)
+    amrex::Table2D<double,amrex::Order::C> table;
```
RemiLehe, line 147: Is there documentation somewhere on this AMReX object?

RemiLehe (approved):

Source/Initialization/ExternalField.H
WeiqunZhang: There is a doxygen page for this. https://amrex-codes.github.io/amrex/doxygen/structamrex_1_1Table2D.html

Source/Initialization/ExternalField.cpp
WeiqunZhang: Yes

PR #6070, opened 2025-08-08
Docs: Hide tutorial chapter title from front page

EZoni:
I think this fixes the issue I reported in https://github.com/BLAST-WarpX/warpx/pull/6065#issuecomment-3161530836, provided that we agree on the fact that it is an issue. 

This way we treat the tutorial chapter like all other chapters and hide it from the front page. 

I tested this locally but let's double check that this works by inspecting the docs once the `docs/readthedocs.org:warpx` CI check runs for this PR.


aeriforme:
It works. My fault - sorry.

aeriforme (approved):

PR #6076, opened 2025-08-11
Several fixes for warnings from doxygen

dpgrote:
Somehow several errors slipped past the changes from PR #3560. Also, this updates the Doxyfile to remove warnings for it.

Side comments:
- It might be perhaps useful to set `TIMESTAMP         = YES` so that the timestamp is included on the generated pages.
- Perhaps set `WARN_AS_ERROR = FAIL_ON_WARNINGS` in `Doxyfile` since the requirement is already that there should be no errors, instead of the modification that is done in `.github/workflows/source/doxygen`. `FAIL_ON_WARNINGS` is better that `YES` so than it will show all of the errors and not just the first one.

### DIFF
--- a/Docs/Doxyfile
+++ b/Docs/Doxyfile
@@ -1347,7 +1347,7 @@ HTML_COLORSTYLE_GAMMA  = 80
 # The default value is: NO.
 # This tag requires that the tag GENERATE_HTML is set to YES.
 
-HTML_TIMESTAMP         = NO
+TIMESTAMP         = NO
 
 # If the HTML_DYNAMIC_MENUS tag is set to YES then the generated HTML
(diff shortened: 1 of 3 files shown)


EZoni:
I suspect that these were undetected because of the `#if defined(...)` - still not clear to me how Doxygen handles those.

EZoni:
This might be what controls preprocessing:
https://github.com/BLAST-WarpX/warpx/blob/d420fbf95b77c421d07f93f2a42ea4ff6e8c7c52/Docs/Doxyfile#L2234-L2238

We could try to set it to `NO` in a separate PR and see if it catches more warnings.

See #6077 for more updates on this.

EZoni (approved):

PR #6077, opened 2025-08-11
[WIP] Doxygen: Fix preprocessor directives

EZoni:
Follow up to #3560, prompted by the discussion in #6076. We may be missing some Doxygen warnings (which we treat as errors for CI purposes) due to preprocessor directives.

Test locally with, e.g.,
```bash
./.github/workflows/source/doxygen WARPX_DIM_1D_Z
```


EZoni:
Hm, this seems to run into the following warning that may be unavoidable, as it is related to a preprocessor directive that is in place for PISCAR (`#define PXRMP_WITH_GPU`):
```console
/home/runner/work/warpx/warpx/Source/Particles/ElementaryProcess/QEDInternals/QedWrapperCommons.H:31:
error: found documented #define PXRMP_WITH_GPU but ignoring it because ENABLE_PREPROCESSING is NO. (warning treated as error, aborting now)
```

EZoni:
This may be the right place to look at:
https://github.com/BLAST-WarpX/warpx/blob/c5b8a512cde594b052b3a4c3b3a0faaeffc9bd8b/Docs/Doxyfile#L2280-L2307

EZoni:
I think one chance we have is to run the analysis separately for each geometry and also wrap the docstring that are specific to a given geometry in ~~a preprocessor `#if defined ...` directive~~ Doxygen's own `\if` conditional - otherwise we will always run into false positive warnings that fail the analysis when treated as errors.

EZoni:
The
 ```bash
./.github/workflows/source/doxygen WARPX_DIM_1D_Z
```
still fails with
```
/home/runner/work/warpx/warpx/Source/BoundaryConditions/PEC_Insulator.H:76: error: 
argument 'set_F_y_lo' of command @param is not found in the argument list of PEC_Insulator::ApplyPEC_InsulatortoField(std::array< amrex::MultiFab *, 3 > field, amrex::Array< FieldBoundaryType, 1 > const &field_boundary_lo, amrex::Array< FieldBoundaryType, 1 > const &field_boundary_hi, amrex::IntVect const &ng_fieldgather, amrex::Geometry const &geom, int lev, PatchType patch_type, amrex::Vector< amrex::IntVect > const &ref_ratios, amrex::Real time, bool split_pml_field, bool E_like, bool set_F_z_lo, bool set_F_z_hi, std::unique_ptr< amrex::Parser > const &a_Fx_z_lo, std::unique_ptr< amrex::Parser > const &a_Fy_z_lo, std::unique_ptr< amrex::Parser > const &a_Fx_z_hi, std::unique_ptr< amrex::Parser > const &a_Fy_z_hi) 
(warning treated as error, aborting now)
```
even though I think it shouldn't fail there.

PR #5269, opened 2024-09-13
[WIP] Add prototype of code flow diagrams

jlvay:
Added "DocDiagram" comments in a few source files, as well as python and shell scripts in tools to generate the Mermaid diagram.

## To Do

- [ ] decide how to long-term generate & update/maintain

PR #6080, opened 2025-08-13
[WIP] Doxygen: enable call graphs

EZoni:
This is a prototype for an implementation of call graphs that, if refined, could replace #5269 (although this is a call graph, not a control flow graph).

When the scope of a function call is not fully qualified, it would be enough to add a call of the form 
```cpp
DOXY_GRAPH(MyClass::bar();)  // doesn't need to match the interface
my_class_ptr->bar(foo);
```
using the `DOXY_GRAPH` macro defined in a dedicated header file DoxyGraph.H, for the function call to be displayed correctly in the call graph generated by Doxygen.

In this PR we set by default
- `DOT_GRAPH_MAX_NODES = 200`: show up to 200 nodes. Graphs with a number of _direct_ children of the root node larger than 200 will not be shown at all. Holds for both call/caller graphs and include/included by graphs - the large max limit is needed by the include/included by graphs, to avoid triggering a warning/error when we run the CI check.
- `MAX_DOT_GRAPH_DEPTH = 1`: only nodes reachable from the root by following a path via at most 1 edge will be shown.

Example for `WarpX::OneStep` (instrumentation with `DOXY_GRAPH` not complete yet):

Screenshot from 2025-08-14 16-44-59

Each node can be clicked interactively leading to its own graphs. 

Explore it for review at https://warpx--6080.org.readthedocs.build/en/6080/_static/doxyhtml/index.html.


Source/Evolve/WarpXEvolve.cpp
```diff
@@ -382,6 +384,7 @@ void WarpX::OneStep (
 
     // perform particle collisions
     ExecutePythonCallback("beforecollisions");
+    DOXY_GRAPH(MultiParticleContainer::doCollisions();)
     mypc->doCollisions(a_step, a_cur_time, a_dt);
```
EZoni, line 388: Without the additional `DOXY_GRAPH` line, this call would not show up in the call graph automatically, because Doxygen's static analysis does not know what `mypc` is.

The scope of the call needs to be fully qualified, as in `MultiParticleContainer::doCollisions()`.

For the purpose of generating the call graph, the interface of the call passed to `DOXY_GRAPH` doesn't need to match the interface of the actual call - hence I drop the input arguments for brevity.

Source/Particles/Collision/CollisionHandler.cpp
```diff
@@ -95,12 +97,14 @@ void CollisionHandler::doCollisions ( int step, amrex::Real cur_time, amrex::Rea
...
     if (m_use_global_debye_length) {
+        DOXY_GRAPH(MultiParticleContainer::GenerateGlobalDebyeLength();)
         mypc->GenerateGlobalDebyeLength();
     }
 
     for (auto& collision : allcollisions) {
         int const ndt = collision->get_ndt();
         if ( step % ndt == 0 ) {
+            DOXY_GRAPH(CollisionBase::doCollisions();)
             collision->doCollisions(cur_time, dt*ndt, mypc);
```
EZoni: This is an example of an approximate graph, where the type of `collision` is determined at runtime and cannot be analyzed statically. 

What I do here is display the call from the base virtual method, assuming that the developer who looks at the graph understands that the actual method called will be the one from the runtime derived class.

EZoni:
@BLAST-WarpX/warpx-technical-committee 

If the majority thinks that this may be useful for developers, I could do more work on this PR and "instrument" (see PR description) a few major functions of interest to most (e.g., OneStep, PushParticlesandDeposit, Evolve, etc.).

Feel free to simply react with a :+1: or :-1: if you want to express a vote in favor or against without further comments, or feel free to comment if you want to elaborate more.

ax3l:
Yes, that is cool. Note that you need to install the "dot" tool from graphviz for this to work. This is not shipped with Doxygen by default (and one package manager might pull it automatically as dependency, another might not).

Possibly update the `.github/` and CI/doc conda/spack/... recipe files for this?

Source/Evolve/WarpXEvolve.cpp
ax3l: Oh nice, I did not know `DOXY_GRAPH` exists. I only knew the graph feature in doxygen can be used for inheritance plots.

I wonder if you can cleverly re-use our tiny-profiler defines for `DOXY_GRAPH`. That way we could instrument once, inside the called function?

Note that doxygen macro processing is a bit simplistic and maybe I am guiding you in the wrong direction.

EZoni: In fact `DOXY_GRAPH` does not exist :sweat_smile:, I created the macro to achieve the goal of making these function calls visible for Doxygen's call graphs with one-line changes.

I'm afraid that merging this with the tiny-profiler wouldn't work for call graphs (you click on a function and see what other functions it calls), but only possibly for caller graphs (you click on a function and see what other functions call it).

I think Doxygen goes through the list of function calls within a scope and skips the ones that can't be resolved, e.g., the ones where the function is called from a polymorphic object. In this context, I think the instrumentation needs to be in the same original scope, rather than hidden in the tiny-profiler macro in the scope of the called function.

EZoni:
Tagging this as "[WIP]" until we gather more feedback from the technical committee. Will try to request a review from everyone and/or bring it up in some other way.

PR #6088, opened 2025-08-14
Suppress GCC -Warray-bounds warnings

EZoni:
This implements the suggestion discussed in #6058 in order to suppress many false positives that occurr when building with GCC.


EZoni:
I'm rebuilding from scratch locally to test this, but I still see, e.g., 
```
[ 10%] Building CXX object CMakeFiles/lib_1d.dir/Source/Diagnostics/BTDiagnostics.cpp.o
In file included from /home/edoardo/src/warpx/build/_deps/fetchedamrex-src/Src/Base/AMReX_Box.H:9,
                 from /home/edoardo/src/warpx/build/_deps/fetchedamrex-src/Src/Base/AMReX_ParallelDescriptor.H:19,
                 from /home/edoardo/src/warpx/build/_deps/fetchedamrex-src/Src/Base/AMReX_Print.H:7,
                 from /home/edoardo/src/warpx/build/_deps/fetchedamrex-src/Src/Base/AMReX_GpuAllocators.H:5,
                 from /home/edoardo/src/warpx/Source/Particles/WarpXParticleContainer_fwd.H:11,
                 from /home/edoardo/src/warpx/Source/Diagnostics/ParticleDiag/ParticleDiag.H:6,
                 from /home/edoardo/src/warpx/Source/Diagnostics/Diagnostics.H:4,
                 from /home/edoardo/src/warpx/Source/Diagnostics/BTDiagnostics.H:10,
                 from /home/edoardo/src/warpx/Source/Diagnostics/BTDiagnostics.cpp:8:
In member function 'constexpr int& amrex::IntVectND<dim>::operator[](int) [with int dim = 1]',
    inlined from 'void BTDiagnostics::InterleaveBufferAndSnapshotHeader(const std::string&, const std::string&)' at /home/edoardo/src/warpx/Source/Diagnostics/BTDiagnostics.cpp:1353:20:
/home/edoardo/src/warpx/build/_deps/fetchedamrex-src/Src/Base/AMReX_IntVect.H:238:79: warning: array subscript 1 is above array bounds of 'int [1]' [-Warray-bounds=]
  238 |     int& operator[] (int i) noexcept { BL_ASSERT(i>=0 && i < dim); return vect[i]; }
      |                                                                           ~~~~^
/home/edoardo/src/warpx/build/_deps/fetchedamrex-src/Src/Base/AMReX_IntVect.H: In member function 'void BTDiagnostics::InterleaveBufferAndSnapshotHeader(const std::string&, const std::string&)':
/home/edoardo/src/warpx/build/_deps/fetchedamrex-src/Src/Base/AMReX_IntVect.H:786:9: note: while referencing 'amrex::IntVectND<1>::vect'
  786 |     int vect[dim] = {};
      |         ^~~~
```
so maybe I'm not disabling the warning in the right way.

EZoni:
I think the problem has something to do with the way the macro `set_cxx_warnings` is executed.

I'm not yet able to debug the macro easily - printing from there does not seem to work... :eyes:

Updates:
- :heavy_check_mark: `cmake -S . -B build -DCMAKE_EXPORT_COMPILE_COMMANDS=ON` works: the macro is called and executed correctly.
- :x: `cmake -S . -B build -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DWarpX_PYTHON=ON` doesn't work: the macro is called but not executed...?

Note that with `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON` one can check the compiler flags in the output file compiler_commands.json, created automatically in the build directory.

CMakeLists.txt
```diff
@@ -9,6 +9,11 @@ project(WarpX VERSION ${warpx_version})
 
 include(${WarpX_SOURCE_DIR}/cmake/WarpXFunctions.cmake)
 
+# Warnings ####################################################################
+#
+# Must call this before any target is created.
+set_cxx_warnings()
+
```
EZoni: Had to moved this here to fix the issues reported in https://github.com/BLAST-WarpX/warpx/pull/6088#issuecomment-3189571065, when we build with `-DWarpX_PYTHON=ON`. 

However, this affects the flags used to compile AMReX and other dependencies as well, which is not good - some CI builds that treat warnings as errors now fail.

EZoni: After some (re)search online, it seemed to me that calling CMake's `target_compile_options` is the correct way to append extra compile flags, instead of setting `CMAKE_CXX_FLAGS`.

At least this is the only way I managed to solve the problem reported in https://github.com/BLAST-WarpX/warpx/pull/6088#issuecomment-3189571065 and in the earlier comment of this very thread.

ax3l: yes, that is the right way :) Sorry, I created and copied this part of the code like a decade ago and never modernized it :D

cmake/WarpXFunctions.cmake
```diff
@@ -171,41 +171,36 @@ macro(set_default_build_type default_build_type)
     endif()
 endmacro()
 
-# Set CXX
-# Note: this is a bit legacy and one should use CMake TOOLCHAINS instead.
+# Set compile warnings
 #
-macro(set_cxx_warnings)
+function(set_compile_warnings tgt)
```
ax3l: We might define similar functions in pyAMReX/ImpactX, so avoiding a collision in superbuilds we can call this:
```suggestion
function(warpx_set_compile_warnings tgt)
```

CMakeLists.txt
```diff
@@ -841,6 +836,13 @@ if(BUILD_TESTING)
 endif()
 
 
+# Warnings ####################################################################
+#
+foreach(warpx_tgt IN LISTS _ALL_TARGETS)
+    set_compile_warnings(${warpx_tgt})
```
ax3l:
```suggestion
    warpx_set_compile_warnings(${warpx_tgt})
```

ax3l (commented):
This is fab, thank you for modernizing this!

Do you mind doing the same PRs to pyAMReX/ImpactX and maybe HiPACE++? :sweat_smile: I definitely reused this block generously :sweat_smile:

EZoni:
@ax3l 

Pushed the name change for superbuilds and verified locally that everything works. This should be ready to go.

cmake/WarpXFunctions.cmake
```diff
@@ -171,41 +171,36 @@ macro(set_default_build_type default_build_type)
...
         # set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=address")
         # set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -fsanitize=address")
         # set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} -fsanitize=address")
-
         # note: might still need a
         #   export LD_PRELOAD=libclang_rt.asan.so
         # or on Debian 9 with Clang 6.0
         #   export LD_PRELOAD=/usr/lib/llvm-6.0/lib/clang/6.0.0/lib/linux/libclang_rt.asan-x86_64.so:
         #                     /usr/lib/llvm-6.0/lib/clang/6.0.0/lib/linux/libclang_rt.ubsan_minimal-x86_64.so
         # at runtime when used with symbol-hidden code (e.g. pybind11 module)
```
EZoni: How much of these comments could we remove at this point?

cmake/WarpXFunctions.cmake
```diff
@@ -171,41 +171,36 @@ macro(set_default_build_type default_build_type)
...
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wpedantic -Wshadow -Woverloaded-virtual -Wunreachable-code")
+        target_compile_options(${tgt} PRIVATE -Wall -Wextra -Wpedantic -Wshadow -Woverloaded-virtual -Wunreachable-code -Wno-array-bounds)
     elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
         # Warning C4503: "decorated name length exceeded, name was truncated"
         # Symbols longer than 4096 chars are truncated (and hashed instead)
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -wd4503")
         # Yes, you should build against the same C++ runtime and with same
         # configuration (Debug/Release). MSVC does inconvenient choices for their
         # developers, so be it. (Our Windows-users use conda-forge builds, which
         # are consistent.)
```
EZoni: How much of these comments could we remove at this point?

cmake/WarpXFunctions.cmake
EZoni: @ax3l 

Do you think we should clean up these comments or would you prefer to merge the PR as is?

cmake/WarpXFunctions.cmake
EZoni: @ax3l 

Do you think we should clean up these comments or would you prefer to merge the PR as is?

EZoni:
@ax3l 

I finally cleaned up the inline comments. This PR is also ready to review/merge.

ax3l (approved):
Thank you, LGTM :)

PR #6089, opened 2025-08-14
Python: Precision & SIMD in `Config`

ax3l:
Extend our `Config` Python object:  Add particle precision properties in the Python API, similar to what we do in pyAMReX. This is useful for users to quickly find out what features are supported.

Also add details on SIMD status and size of SIMD lanes/registers.

Note: SIMD is not yet used in WarpX, but we will soon use it more (e.g., for the ES solver, Green's function).


roelof-groenewald (approved):
Looks good!

PR #6091, opened 2025-08-15
Fix bug for field probe + moving window

### LINKED ISSUE #6082: FieldProbe not working properly when moving window is enabled
Dear developers,

I am encountering an issue while testing WarpX simulations using the official example input file “Laser-Wakefield Acceleration of Electrons”:

* When the moving window (`warpx.do_moving_window=1`) is enabled, the FieldProbe diagnostics do not work properly. The probe data only shows the initial timestep.
* However, if I disable the moving window (`warpx.do_moving_window=0`), the probe data outputs correctly as expected.

I’m running the official WarpX codebase on Windows.

Has anyone experienced this behavior before? Is this a known limitation of WarpX’s current moving window implementation, or could I be missing some configuration?

Any advice or guidance would be greatly appreciated. Thanks!

dantargz:
### Issues being fixed
There are currently three issues with FieldProbe related to moving window:
1. FieldProbe filtering logic is intended to only get data for points within the problem domain; the implementation checks the initial configuration settings of the probe, which will become inaccurate over time if `do_moving_window_FP` is enabled, as the actual probe positions diverge from initial configured settings
2. if `end_moving_window_step` is set to default value (-1), FieldProbe sample points will not be moved with the window, but the intended meaning of the default value is to move indefinitely
3. After FieldProbe updates the position of its probe "particles", it does not call `Redistribute()` on the particle container; this can cause a memory access violation when reading field state, if particles cross between storage cells as a result of movement

### How the fixes work
Issue 1 is fixed by removing the `ProbeInDomain` check; the particle container used to store the probe particles will handle keeping the data inside of the domain. A helper class `ParticleInDomainChecker` has been added, that can check if a FieldProbe particle is still within the problem domain, and is checked via `AMREX_ASSERT` before accessing field data

Issue 2 is fixed by removing the current logic for checking if moving_window applied, and replacing it with the WarpX helper function `WarpX::moving_window_active` that makes the same check. The helper function correctly handles default vale of `end_moving_window_step`. 

Issue 3 is fixed by invoking `Redistribute()` on the FieldProbe particle container after particles are moved & before fields are read. `Redistribute()` cannot be called while there is a `MyParIter` iterating, so the main loop in `FieldProbe::ComputeDiags` is split.

### Testing
Added a few automated tests + a new analysis script. The new analysis script checks these characteristics:
* Verifies data columns are as expected
* Verifies that data is generated for every step (Fix #6082 / Bug 1)
* Verifies that coordinates are updated appropriately (Bug 2)
* Verifies integration still happens even when probe points are cut out of the problem domain

These scenarios are run:
* No moving window
* Problem domain has moving window, but field probe does not
* ProblemDomain + FieldProbe have moving window
* FieldProbe line definition has some points inside of the problem domain and some that lay outside of the problem domain
* FieldProbe points frequently change cells of internal AMReX data structures (repros bug 3 under ASAN when `Redistribute()` call is not made)

### Additional concerns
* There is already a call to `Redistribute()` in FieldProbe; it may be possible to perform a larger refactor, triggering FieldProbe to move its particles as part of `WarpX::MoveWindow`, in a way that happens ahead of the existing call to `Redistribute()`. I did not pursue that option because `WarpX::MoveWindow` happens every evolve step, and it is unclear if the performance cost of moving them every evolve step (vs once every `interval`) was worthwhile.
* The logic in `WarpX::moving_window_active` uses  `>=` start vs the logic in FieldProbe used to be `> start`, which may slightly change when the FieldProbe points move.

Fix #6082


ax3l:
@dantargz thank you for digging into the issue and proposing a fix!

Can you please double check that a simulation in which the probes run out of the simulation domain is still safe/not violating memory? You could check by doing a moving window simulation and disabling that the probe co-moves with the window.

I think the current implementation might be safe but has the bug that it does no interpolation to probes at all once a single particle of the probe leaves the domain.

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
```diff
@@ -346,32 +346,30 @@ void FieldProbe::LoadBalance ()
...
-    const amrex::Geometry& gm = warpx.Geom(lev);
-    const auto *const prob_lo = gm.ProbLo();
-    const auto *const prob_hi = gm.ProbHi();
+using ProblemDomainVec = amrex::GpuArray<amrex::Real, AMREX_SPACEDIM>;
 
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE bool
+ProbeInDomain (const ProblemDomainVec& prob_lo, const ProblemDomainVec& prob_hi,
+               [[maybe_unused]] amrex::ParticleReal xp,
+               [[maybe_unused]] amrex::ParticleReal yp,
+               [[maybe_unused]] amrex::ParticleReal zp) {
```
ax3l: You could make `ProbeInDomain` a functor that on `ComputeDiags` stores the current prob_lo/hi of the simulation.

The functor could be executed on GPU and compare a current particle position.

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
```diff
@@ -449,31 +452,33 @@ void FieldProbe::ComputeDiags (int step)
...
-                    amrex::ParticleReal xp, yp, zp;
-                    getPosition(ip, xp, yp, zp);
-                    if (temp_warpx_moving_window == 0)
-                    {
-                        setPosition(ip, xp+move_dist, yp, zp);
+            bool allInDomain = true;
+            bool* allInDomainPtr = &allInDomain;
+            const auto temp_warpx_moving_window = WarpX::moving_window_dir;
+
+            amrex::ParallelFor(np, [=] AMREX_GPU_DEVICE(long ip) {
```
ax3l, line 638: Let's remove the "all-in" check. Otherwise we will disable all probes if even just one of them is out of bounds.

Instead, go to the `amrex::ParallelFor( np, [=] AMREX_GPU_DEVICE` in line 518 and check for the individual particle. if it is out of bounds, assign zero values, otherwise interpolate as before.

dantargz: Will do. One thing I wanted to double check: 
```C++
const auto &arrEx = Ex[pti].array();
```
That is the next step after the all-in check. When I dig into the source for that, it seems like operator[] in this case is getting a `LocalIndex()` value from `pti`; I was not sure that using that to index warpx data was always valid if some of the probes were outside of the domain. I have not finished reading through the design docs, if there is part that you could point me to, to help me learn more about that / how to access the data safely, I would appreciate it.

This was my thought process wrt options for behavior, I lean towards option 2, but I am happy to defer to your opinion as you have much more experience with using this code.

1. Current behavior ("all-in") -- omit no data for the step when any particle is outside of the domain
2. Best effort -- read data for probes that are inside of the domain, and emit no data for probes that are outside of the domain
3. Default value -- read some default value for probes that are outside of the domain

Pros / cons for each option:
option 1 "all-in"
* (+) it is (almost?) impossible for any further analysis of output data to be done incorrectly -- main risk would be from not reading the "step" field
* (+) easiest to avoid memory access violation
* (-) it takes additional (potentially non-trivial) work for the user to configure their FieldProbe to always be inside the domain -- especially if they are e.g. using python interface

option 2 "best effort"
* (+) Makes the most intuitive sense
* (+) Gives the user the most possible data from each run
* (-) Doing analysis of any individual probe's data over time becomes complicated, especially if a moving window is used (can't rely on output index)

option 3 "default value"
* (+) Gives the user the most possible data from each run
* (-) Analysis of individual probe over time, or aggregates over multiple probe within a single step (e.g. interpolating over n adjacent probes) is error prone, because it might include a default 0 value
* (-) may create ambiguous output (user might not know when a pobe goes outside of the domain vs the probe observed a simulated state where there was truly no E/B)

dantargz: Additional con for option 1:
* (-) Does not perform integration when it should (as you indicated in your other comment)

dantargz:
@ax3l it looks like there are some automated tests in `Examples/Tests/reduced_diags`, but they are failing for me locally (running on CPU on mac):
```
$ ctest --test-dir build -j 4 -R reduced_diags
[...]
The following tests FAILED:
        196 - test_3d_reduced_diags.analysis (Failed)
        197 - test_3d_reduced_diags.checksum (Failed)
        200 - test_3d_reduced_diags_load_balance_costs_heuristic.checksum (Failed)
        203 - test_3d_reduced_diags_load_balance_costs_timers.checksum (Failed)
```
Do you recommend that I try to make those tests pass & add to them, or that I do something else for testing?

ax3l:
They are likely failing because you disable all probes now the moment one goes out-of-bounds.

Let us implement the correct solution for the bug we see first (per-particle check if in bounds) and then double-check which tests fail, plot their data, and update them.

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
ax3l: 2 is best. We could either set 0 or a silent NaN on probe particles that are out-of-bounds.

Instead of the position `i` in the output array of particles, you can always use our particle id (`cpuid`) attribute, which is unique for a particle in a species, to track a particle between time steps.

ax3l: The construct
```C++
const auto &arrEx = Ex[pti].array();
```
gives you the field box for `Ex` in which the current particle chunk of data (`pti`) resides. The call to `doGatherShapeN` then does the logic of offset-correction and interpolation between particle position x,y,z and indices in the local `arrEx`.

For the purpose here, you just need to go into
```C++
                // Interpolating to the probe positions for each particle
                amrex::ParallelFor( np, [=] AMREX_GPU_DEVICE (long ip)
                {
                    amrex::ParticleReal xp, yp, zp;
                    getPosition(ip, xp, yp, zp);
```

and add an if just right after `getPosition(ip, xp, yp, zp)` to continue with an interpolation or setting a default value.

dantargz: Okay, I think I have implemented as described, but something is wrong with the formatting, I am working on fixing it now.

EZoni: See naming conventions for "base" input files described in our documentation: https://warpx.readthedocs.io/en/latest/developers/how_to_test.html#naming-conventions-for-automated-tests.

Examples/Tests/field_probe/CMakeLists.txt
```diff
@@ -12,3 +12,33 @@ if(WarpX_EB)
...
+    test_3d_field_probe_line_partial_window
+    3  # dims
+    2  # nprocs
+    inputs_test_3d_field_probe_line_partial_window  # inputs
+    "analysis_field_probe_line.py --path diags/reducedfiles/FP.txt"  # analysis
+    OFF  # checksum
+    OFF  # dependency
+)
+
+add_warpx_test(
```
dantargz, line 86: TODO: add test that verifies behavior of `integrate = 1` when part of the probe is outside of the domain

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
dantargz: Thanks, I missed that. I will include the base test file name fix in my next push; holding on next push until I finish also fixing the precommit checks that I just got a failure on, to not spam the CI system.

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
```diff
@@ -346,34 +346,40 @@ void FieldProbe::LoadBalance ()
...
-
-    /*
-     * Determine if probe exists within simulation boundaries. During 2D simulations,
-     * y values will be set to 0 making it unnecessary to check. Generally, the second
-     * value in a position array will be the y value, but in the case of 2D, prob_lo[1]
-     * and prob_hi[1] refer to z. This is a result of warpx.Geom(lev).
-     */
+using ProblemDomainVec = amrex::GpuArray<amrex::Real, AMREX_SPACEDIM>;
+
+struct ParticleInDomainChecker {
```
dantargz: TODO @dantargz , this might be redundant with the particle container's internal implementation -- it looks like there is some mechanism already in place that removes particles that are outside of the problem domain, figure out how that works and how it interacts with FieldProbe

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
```diff
@@ -396,10 +406,16 @@ void FieldProbe::ComputeDiags (int step)
         amrex::Real const dt = WarpX::GetInstance().getdt(lev);
         // Calculates particle movement in moving window sims
         amrex::Real move_dist = 0.0;
+
+        // Keep in sync with WarpX::moving_window_active()
+        // TODO -- is this supposed to be offset by 1 vs that logic (starting one step after)?
```
dantargz: TODO @dantargz

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
dantargz: implemented as recommended.

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
dantargz: @ax3l  I needed to add a call to [Redistribute()](https://amrex-codes.github.io/amrex/docs_html/Particle.html#redistribute) after moving the particles anyway -- do you know if that will also eliminate any particles that are outside of the domain? If so, then perhaps I can remove this entirely, and rely on Redistribute() to guarantee all Particles have valid mesh data to query?

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
dantargz: I think this is resolved now

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
dantargz: Changed to an assert for now

dantargz:
@ax3l Okay, I just removed the WIP tag, I think this is ready for proper review.

titoiride:
I wasn't aware of this when I opened #6257. I'm curious, how do you handle the case with moving window then? Is the probe in the reference frame moving with the window, which means it gets shifted with it, or in the lab frame? I think both cases would have different use cases (I'm currently in the latter), although I see the argument that the former is of more general use.

EZoni:
I'll let @dantargz or @ax3l answer @titoiride's question:

> I wasn't aware of this when I opened https://github.com/BLAST-WarpX/warpx/issues/6257. I'm curious, how do you handle the case with moving window then? Is the probe in the reference frame moving with the window, which means it gets shifted with it, or in the lab frame? I think both cases would have different use cases (I'm currently in the latter), although I see the argument that the former is of more general use.

As a side note, I just wanted to observe that the linked issue #6082 and the user @smilingly1919 who had opened it seem to have misteriously disappeared from GitHub.

However, I found this track in my emails, in case it helps remind how everything started:

Screenshot from 2025-10-10 13-32-12

ax3l:
Hi all, sorry for the delay. I was on vacation and am still on reduced hours for some family support. I will unassign myself for now, to not indicate that I am leading the review right now.

dantargz:
> I'm curious, how do you handle the case with moving window then? Is the probe in the reference frame moving with the window, which means it gets shifted with it, or in the lab frame?

@titoiride , moving window behavior for field probes is intended to be controlled by `<reduced_diags_name>.do_moving_window_FP` https://warpx.readthedocs.io/en/latest/usage/parameters.html#reduced-diagnostics near the end of the section, search for `do_moving_window_FP`. Some caveats:
* If you want the field probe locations to move with the window, you also need to set `end_moving_window_step` (that is a bug, it would be fixed by this PR).
* The field probe locations do not move in the same part of the code that handles shifting the problem domain, and the field probe locations are moved only when the field probe values are sampled; that saves on perf (because we read field probe values at lower frequency than general simulation), but the inconsistency means that it is possible for field probes near the edge of the window to get dropped even if moving window is set up correctly

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
```diff
@@ -443,154 +452,158 @@ void FieldProbe::ComputeDiags (int step)
...
-                const auto &arrBy = By[pti].array();
-                const auto &arrBz = Bz[pti].array();
+                amrex::ParticleReal xp, yp, zp;
+                getPosition(ip, xp, yp, zp);
+
+                amrex::ParticleReal Exp = 0._prt, Eyp = 0._prt, Ezp = 0._prt;
+                amrex::ParticleReal Bxp = 0._prt, Byp = 0._prt, Bzp = 0._prt;
+
+                // Amrex particle container should guarantee that all particles are in domain
+                AMREX_ASSERT(inDomainChecker(xp, yp, zp));
```
ax3l: This assert will only be executed in debug mode.

Do you mean to write an `if` here?

```suggestion
                if (inDomainChecker(xp, yp, zp))
                {
                    // doGatherShapeN
                    // calc poynting vector
                }
```
and assign/add the default zeros in `else` to the particle.

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
```diff
@@ -396,10 +406,9 @@ void FieldProbe::ComputeDiags (int step)
         amrex::Real const dt = WarpX::GetInstance().getdt(lev);
         // Calculates particle movement in moving window sims
         amrex::Real move_dist = 0.0;
+
         bool const update_particles_moving_window =
-            do_moving_window_FP &&
-            step > WarpX::start_moving_window_step &&
-            step <= WarpX::end_moving_window_step;
+            do_moving_window_FP && WarpX::moving_window_active(step);
         if (update_particles_moving_window)
```
ax3l: The logic here as likely similar issues as we try to fix in #6399, namely that if the moving window starts or stops between the last `m_last_compute_step` it likely gets off.

We could move the `move_dist`  calc up, before we return in the `// Judge if the diags should be done` block, to somehow make this milder - that part then might be called every step and the `WarpX::moving_window_active(step);` will give us a correct indication if we are still moving - not duplicating any logic of the start/stop of the window.

ax3l (commented):
I think this looks already great!

Small idea about the moving window to be more robust and about the assert (replace it with an `if` that works in `Release` mode, too).

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
dantargz: When I did testing, I wasn't able to create a situation where there was a particle in the container that was outside of the problem domain. To improve my understanding, let me know if there is a straightforward way to create that situation, I can try to add a test that creates that situation.

If the particle is not in the problem domain, I thought we could not call doGatherShapeN (or at least wouldn't get reliable results). If `inDomainChecker(...)` is false, what values should I use for E field, B fields, and S vector?

Source/Diagnostics/ReducedDiags/FieldProbe.cpp
dantargz: Sorry, I have lost a lot of context for this change since the last time I touched it, so I am not sure I am groking the implications. I would prefer not to add additional complexity to this PR -- would committing this PR without adding fixes for stopping and restarting make the stopping & restarting issue worse? If not, would it be reasonable to treat that issue as separate and not blocking for this change?

ax3l:
I did a Codex GPT 5.6 sol review and the implementation still has a few issues left:

  ### Findings

  - [P1] Redistribute() is not a safe domain filter.
    At Source/Diagnostics/ReducedDiags/FieldProbe.cpp:480, out-of-domain
    non-periodic probes are permanently invalidated, while probes outside
    periodic dimensions are wrapped back into the domain. This means a static
    probe ahead of a moving window is deleted before the window reaches it, and
    an out-of-range periodic probe samples the wrong location. That follows
    directly from the pinned AMReX locator implementation
    (https://github.com/AMReX-Codes/amrex/blob/5ca2c5e2326770403418a63714b9db2f8
8ef1796/Src/Particle/AMReX_ParticleContainerI.H#L447-L460).

  - [P1] Co-moving probes can be deleted before they move.
    WarpX shifts the domain at Source/Evolve/WarpXEvolve.cpp:268, then calls
    reduced-diagnostic LoadBalance() at Source/Evolve/WarpXEvolve.cpp:349.
    FieldProbe immediately redistributes its still-unshifted particles at
    Source/Diagnostics/ReducedDiags/FieldProbe.cpp:346, so trailing-edge
    particles can be discarded before ComputeDiags() moves them. The new test
    misses this because its --resolution argument is parsed but never checked at
    Examples/Tests/field_probe/analysis_field_probe_line.py:217.

  - [P1] Empty or sparse probe IDs can corrupt memory.
    If all probes are removed, Source/Diagnostics/ReducedDiags/
    FieldProbe.cpp:664 reads m_data_out[0] from an empty vector. If clipping
    leaves non-contiguous IDs—easy with a partially clipped plane—Source/
    Diagnostics/ReducedDiags/FieldProbe.cpp:680 uses id-first_id as a dense
    vector index and can write beyond sorted_data. The line test only clips a
    contiguous suffix, so it does not expose either case.

  - [P2] Redistribution inside the AMR-level loop can move a probe twice.
    The added call at Source/Diagnostics/ReducedDiags/FieldProbe.cpp:480
    redistributes all levels. AMReX can promote a moved level-0 particle to
    level 1; the outer loop then reaches level 1 and moves that particle again.
    Movement should be completed across all levels, followed by one
    redistribution, followed by gathering.

  - [P2] The finite start/stop moving-window bug remains.
    Your existing thread at Source/Diagnostics/ReducedDiags/FieldProbe.cpp:411
    is still valid. WarpX calls MoveWindow(step+1), but FieldProbe queries
    moving_window_active(step). More importantly, multiplying the entire step -
    m_last_compute_step interval when only part of it was active over- or under-
    shifts sparse diagnostics. Tracking the actual geometry displacement would
    avoid duplicated start/stop logic.

  A robust direction would be to move co-moving probes from the actual
  Geom(0).ProbLo() displacement before FieldProbe redistribution, while
  retaining a separate logical registry for probes currently outside the domain.



I'll start on those and push here.

PR #6092, opened 2025-08-15
Add verbosity flag for diagnostics

### LINKED ISSUE #6031: Diagnostic write events no longer appear in stdout
I recently rebuilt WarpX and noticed that my diagnostic write events are no longer printing anything to stdout. Setting `verbose=1` in the `Simulation()` instance restores it, but also turns on undesired prints that occur every time step, ballooning the size of the stdout file. It seems that [this commit](https://github.com/BLAST-WarpX/warpx/commit/540268f2c05d46b90ac8f33718efa63b47687cae#diff-f03035e5d437bc992ba6a971c6af5a35dc78e7a2d20f8d01eb5071d84347c5d1) changed the diagnostic write event print behavior. I don't see any verbosity control parameter in e.g. the `ParticleDiagnostic` [class in picmi.py](https://github.com/BLAST-WarpX/warpx/blob/89a8034ee4e32cad7038e0f9bc1cbae6c878e6c1/Python/pywarpx/picmi.py#L3676).

How do I restore the print statements on diagnostic write events without turning on a huge amount of other prints?

roelof-groenewald:
Fixes #6031.

### DIFF
--- a/Examples/Tests/ohm_solver_em_modes/inputs_test_1d_ohm_solver_em_modes_picmi.py
+++ b/Examples/Tests/ohm_solver_em_modes/inputs_test_1d_ohm_solver_em_modes_picmi.py
@@ -266,7 +266,7 @@ class EMModes(object):
                 grid=self.grid,
                 period=self.total_steps,
                 data_list=["B", "E", "J_displacement"],
-                warpx_verbosity=0,
+                warpx_verbose=0,
                 # warpx_format = 'openpmd',
                 # warpx_openpmd_backend = 'h5'
             )
--- a/Examples/Tests/ohm_solver_em_modes/inputs_test_1d_ohm_solver_em_modes_picmi.py
+++ b/Examples/Tests/ohm_solver_em_modes/inputs_test_1d_ohm_solver_em_modes_picmi.py
@@ -266,6 +266,7 @@ class EMModes(object):
                 grid=self.grid,
                 period=self.total_steps,
                 data_list=["B", "E", "J_displacement"],
+                warpx_verbosity=0,
                 # warpx_format = 'openpmd',
                 # warpx_openpmd_backend = 'h5'
             )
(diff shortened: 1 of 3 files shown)


EZoni:
Thanks for the fix, @roelof-groenewald! 

I see we use both `verbose` and `verbosity`, wouldn't it be possible to use only one name everywhere (in which case I would choose `verbose` personally)? I'm afraid using two different, yet similar, names might lead to some confusion.

roelof-groenewald:
> I see we use both `verbose` and `verbosity`, wouldn't it be possible to use only one name everywhere (in which case I would choose `verbose` personally)? I'm afraid using two different, yet similar, names might lead to some confusion.

I also had that concern but didn't want to make this PR cause breaking changes to existing inputs. I tried to look for a pattern as best I can, but couldn't really spot one. My opinion is that `verbose` represents a boolean (on or off) whereas `verbosity` indicates a level (integer), which is why I opted for `verbosity`.

EZoni:
I understand that, and I would think the same ideally.

On the other hand, though, AMReX chose `verbose`: 
Screenshot from 2025-08-26 11-12-10

So I wonder if we should choose the same to streamline things as much as possible between AMReX and WarpX.

But regardless of whether we choose `verbose` or `verbosity` (I'm fine with both), I still think that exposing both to the user is redundant/not ideal.

I would be okay if the fix here requires a breaking change in the inputs. We can see what other reviewers think as well.

ax3l:
I think we should use use `verbose` as in ParmParse.

We can refactor the `verbosity` ones separately.

ax3l (approved):

jwestern:
Sorry for commenting after merge, but can't inputs be preserved by looking for either `verbose` or `verbosity` in the inputs and then writing it to `verbose` internally? I'm not asking to reopen this, I'm just curious what the problem with that would be. As a user I prioritize backwards-compatibility highly.

PR #6096, opened 2025-08-18
CMake: Fix Fast-Math with Pre-built AMReX

ax3l:
The fast-math flags were not added if AMReX was pre-built without fast-math. This fixes it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -293,6 +293,21 @@ foreach(D IN LISTS WarpX_DIMS)
     list(APPEND _ALL_TARGETS ablastr_${SD})
     add_library(WarpX::ablastr_${SD} ALIAS ablastr_${SD})
 
+    # Fast-Math
+    if(ABLASTR_FASTMATH)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -293,6 +293,21 @@ foreach(D IN LISTS WarpX_DIMS)
...
     add_library(WarpX::ablastr_${SD} ALIAS ablastr_${SD})
 
+    # Fast-Math
+    if(ABLASTR_FASTMATH)
+        target_link_libraries(ablastr_${SD} PUBLIC AMReX::Flags_FASTMATH)
+    endif()
+
+    # GCC < 9 need to manually link std::filesystem
+    if(NOT CXX_HAS_STD_FS)
+        if(CXX_HAS_STD_FS_WITH_LINK)
```
ax3l, line 303: This is another mini-bug that the PR fixes: if no WarpX lib is built but ABLASTR is used, the GCC<9 work-around was no applied.

EZoni (approved):

PR #6103, opened 2025-08-20
Update WarpX's high-level description

RemiLehe:
Update WarpX's high-level description to reflect changes that have happened over the last few years.

### DIFF
--- a/Docs/source/index.rst
+++ b/Docs/source/index.rst
@@ -3,15 +3,16 @@
 WarpX
 -----
 
-WarpX is an advanced **electromagnetic & electrostatic Particle-In-Cell** code.
+WarpX is an advanced **Particle-In-Cell** code.
 
 It supports many features including:
 
-    - Perfectly-Matched Layers (PML)
-    - Boosted-frame simulations
-    - Mesh refinement
+    - Multiple types of field solvers (incl. electromagnetic, electrostatic)
+    - Various grid geometries (Cartesian, cylindrical, spherical)
+    - Multi-physics packages (incl. ionization and atomic physics, Quantum Electrodynamics)
+    - Advanced numerical methods (incl. explicit and explicit PIC loop, mesh refinement, boosted-frame simulations)
 
-For details on the algorithms that WarpX implements, see the :ref:`theory section <theory>`.
+For details on these features, see the :ref:`theory section <theory>`.
 
 WarpX is a *highly-parallel and highly-optimized code*, which can run on GPUs and multi-core CPUs, and includes load balancing capabilities.
 WarpX scales to the world's largest supercomputers and was awarded the `2022 ACM Gordon Bell Prize <https://www.exascaleproject.org/ecp-supported-collaborative-teams-win-the-2022-acm-gordon-bell-prize-and-special-prize/>`__.
--- a/Docs/source/index.rst
+++ b/Docs/source/index.rst
@@ -8,7 +8,7 @@ WarpX is an advanced **Particle-In-Cell** code.
 It supports many features including:
 
     - Multiple types of field solvers (incl. electromagnetic, electrostatic)
-    - Various grid geometries (Cartesian, cylindrical, spherical)
+    - Various grid geometries (1D/2D/3D Cartesian, cylindrical, spherical)
     - Multi-physics packages (incl. ionization, atomic and collisional physics, quantum electrodynamics)
     - Advanced numerical methods (incl. explicit and implicit PIC loop, mesh refinement, boosted-frame simulations, embedded boundaries)
 
--- a/Docs/source/index.rst
+++ b/Docs/source/index.rst
@@ -15,7 +15,7 @@ It supports many features including:
 For details on these features, see the :ref:`theory section <theory>`.
 WarpX has been applied to a wide variety of science projects, see :ref:`highlights <highlights>`.
 
-WarpX is a *highly-parallel and highly-optimized code*:
+In addition, WarpX is a *highly-parallel and highly-optimized code*:
 
     - Can run on multi-core CPUs as well as NVIDIA, AMD or Intel GPUs
     - Scales to the world's largest supercomputers and includes load balancing capabilities. WarpX was awarded the `2022 ACM Gordon Bell Prize <https://www.exascaleproject.org/ecp-supported-collaborative-teams-win-the-2022-acm-gordon-bell-prize-and-special-prize/>`__.
--- a/Docs/source/index.rst
+++ b/Docs/source/index.rst
@@ -10,7 +10,7 @@ It supports many features including:
     - Multiple types of field solvers (incl. electromagnetic, electrostatic)
     - Various grid geometries (Cartesian, cylindrical, spherical)
     - Multi-physics packages (incl. ionization and atomic physics, quantum electrodynamics)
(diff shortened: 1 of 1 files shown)


Docs/source/index.rst
```diff
@@ -3,15 +3,16 @@
...
+WarpX is an advanced **Particle-In-Cell** code.
 
 It supports many features including:
 
-    - Perfectly-Matched Layers (PML)
-    - Boosted-frame simulations
-    - Mesh refinement
+    - Multiple types of field solvers (incl. electromagnetic, electrostatic)
+    - Various grid geometries (Cartesian, cylindrical, spherical)
+    - Multi-physics packages (incl. ionization and atomic physics, Quantum Electrodynamics)
```
RemiLehe:
```suggestion
    - Multi-physics packages (incl. ionization and atomic physics, quantum electrodynamics)
```

Docs/source/index.rst
```diff
@@ -3,15 +3,16 @@
...
 
 It supports many features including:
 
-    - Perfectly-Matched Layers (PML)
-    - Boosted-frame simulations
-    - Mesh refinement
+    - Multiple types of field solvers (incl. electromagnetic, electrostatic)
+    - Various grid geometries (Cartesian, cylindrical, spherical)
+    - Multi-physics packages (incl. ionization and atomic physics, quantum electrodynamics)
+    - Advanced numerical methods (incl. explicit and explicit PIC loop, mesh refinement, boosted-frame simulations)
```
RemiLehe:
```suggestion
    - Advanced numerical methods (incl. explicit and implicit PIC loop, mesh refinement, boosted-frame simulations)
```

Docs/source/index.rst
```diff
@@ -3,19 +3,23 @@
...
-WarpX is an advanced **electromagnetic & electrostatic Particle-In-Cell** code.
+WarpX is an advanced **Particle-In-Cell** code.
 
 It supports many features including:
 
-    - Perfectly-Matched Layers (PML)
-    - Boosted-frame simulations
-    - Mesh refinement
+    - Multiple types of field solvers (incl. electromagnetic, electrostatic)
+    - Various grid geometries (Cartesian, cylindrical, spherical)
```
RemiLehe:
```suggestion
    - Various grid geometries (1D/2D/3D Cartesian, cylindrical, spherical)
```

Docs/source/index.rst
```diff
@@ -3,19 +3,23 @@
...
 
-WarpX is an advanced **electromagnetic & electrostatic Particle-In-Cell** code.
+WarpX is an advanced **Particle-In-Cell** code.
 
 It supports many features including:
 
-    - Perfectly-Matched Layers (PML)
-    - Boosted-frame simulations
-    - Mesh refinement
+    - Multiple types of field solvers (incl. electromagnetic, electrostatic)
```
ax3l, line 10: In case this is clear enough we can list @roelof-groenewald :
```suggestion
    - Multiple types of field solvers (incl. electromagnetic, electrostatic, kinetic-fluid)
```

ax3l (approved):
Looks great, I let others take a look to finalize over the day.

Docs/source/index.rst
RemiLehe: Hm, I feel like `kinetic-fluid` is a bit vague. The proper way to describe the solver that we have in WarpX would be to have a phrasing that mentions Ohm's law (because the solver replaces one of the Maxwell equation with Ohm's law), which is not captured in the phrasing `kinetic-fluid`.

aeriforme (approved):

Docs/source/index.rst
RemiLehe: Actually: the same comment applies to this page:
https://warpx.readthedocs.io/en/latest/theory/kinetic_fluid_hybrid_model.html

RemiLehe: Technically: "kinetic-fluid" refers to how we treat the particles, not to the type of field equations that we use.

roelof-groenewald: I agree the "field solver" is more aptly described as "Ohm's law". But I would say we should overall not paint ourselves into a corner like this by saying "electromagnetic, electrostatic" since much of the versatility of WarpX comes from the zoo of solvers. I would say we should list the various types as, "Maxwell FDTD, Maxwell pseudo-spectral, Poisson, Ohm's law". And then in the future we can add "Darwin".

EZoni: @roelof-groenewald 

Are you saying that you would prefer `Multiple types of field solvers` rather than `Multiple types of field solvers (incl. electromagnetic, electrostatic)`?

I would agree with that, but I'm fine with what the majority prefers.

To elaborate even more, I would prefer a `Multiple types of field solvers` with `field solvers` direct link to the section of the docs where the field solvers are listed, rather than a subselection of solvers in parentheses, which implies a judgement call on what is more important and what is less important.

Docs/source/index.rst
```diff
@@ -3,19 +3,23 @@
...
+    - Various grid geometries (1D/2D/3D Cartesian, cylindrical, spherical)
+    - Multi-physics packages (incl. ionization, atomic and collisional physics, quantum electrodynamics)
```
EZoni: Just for transparency, I repost, and elaborate on, my Slack comment on
```
(incl. ionization, atomic and collisional physics, quantum electrodynamics)
```
pointing out that, in my opinion, whether or not the word "fusion" appears in the landing page of the WarpX docs may (or should?) be a bit of a strategic decision.

EZoni: Alternatively, we could at least point explicitly to the science highlights section, https://warpx.readthedocs.io/en/latest/highlights.html, where it's easier to see the various physics applications that the code can tackle, based on the publications record.

Docs/source/index.rst
roelof-groenewald: Linking to a different page with a list & description of each solver sounds good to me. Then that page could have a reference to the method paper that describes each solver (assuming such a paper was written).

Docs/source/index.rst
roelof-groenewald: I think it makes sense to mention `fusion` if we specifically call out `atomic`. 

It might make more sense to link to the highlights page in another paragraph, after the one saying WarpX is highly parallel. Even just adding
```
WarpX has been applied to a wide variety of science projects, see [highlights](https://warpx.readthedocs.io/en/latest/highlights.html).
```

Docs/source/index.rst
```diff
@@ -3,19 +3,24 @@
...
 
 It supports many features including:
 
-    - Perfectly-Matched Layers (PML)
-    - Boosted-frame simulations
-    - Mesh refinement
+    - Multiple types of field solvers (incl. electromagnetic, electrostatic)
+    - Various grid geometries (1D/2D/3D Cartesian, cylindrical, spherical)
+    - Multi-physics packages (incl. ionization, atomic, fusion and collisional physics, as well as quantum electrodynamics)
+    - Advanced numerical methods (incl. explicit and implicit PIC loop, mesh refinement, boosted-frame simulations, embedded boundaries)
```
RemiLehe:
```suggestion
    - Advanced numerical methods (incl. explicit and implicit time advance, mesh refinement, boosted-frame simulations, embedded boundaries)
```

ax3l (approved):

ax3l:
Merging the current iteration so we can use it for ongoing discussions in proposals.

PR #6110, opened 2025-08-23
Fix openPMD output of "eb_covered" in RZ geometry

RemiLehe:
Before, this PR, "eb_covered" was dumped as a 2Dcartesian field, when compiling WarpX in RZ geometry.

Adding `AddRZModesToOutputNames` affects the rest of the WarpX logic (esp. this line: https://github.com/BLAST-WarpX/warpx/blob/development/Source/Diagnostics/WarpXOpenPMD.cpp#L1445), so that "eb_covered" is written in RZ format (i.e. `thetaMode` in openPMD).


ax3l (approved):
Thanks for the fix!

Issue #5291, opened 2024-09-18
CTest follow-up: migrate CI tests from Azure to GitHub Actions


EZoni:
We could migrate our CI tests from Azure to GitHub Actions, which could lead to better caching thanks to recent ccache improvements by @WeiqunZhang.

Alternatively, we could check if we can make use of more control in Azure Pipelines for caches. 

There might be more limitations on parallelism, which we would need to assess once we try the migration.

PR #6115, opened 2025-08-26
CI: debug autoupdate workflows (GitHub Actions)

EZoni:
Debugging the weekly autoupdate workflow, ~trying `permissions: write-all` this time.~ 

Also in this PR:
- Renamed the workflows to avoid duplication of the "Autoupdate" label.
- Replaced `git push -u ...` with `git push -f ...` to force push when necessary.
- Added `--draft` flag to the `gh pr create` command to create a draft pull request and work around the CI issue.
- Added activity types to the `pull_request` trigger event for all regular workflows to include `ready_for_review`.

See https://github.com/BLAST-WarpX/warpx/pull/6115#issuecomment-3225928638 and references therein for more details on the last two points.

### DIFF
--- a/.github/workflows/clang_sanitizers.yml
+++ b/.github/workflows/clang_sanitizers.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - "development"
   pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
 
 concurrency:
   group: ${{ github.ref }}-${{ github.head_ref }}-clangsanitizers
--- a/.github/workflows/clang_tidy.yml
+++ b/.github/workflows/clang_tidy.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - "development"
   pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
 
 concurrency:
   group: ${{ github.ref }}-${{ github.head_ref }}-clangtidy
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -5,6 +5,7 @@ on:
     branches: [ "development" ]
   pull_request:
     branches: [ "development" ]
+    types: [opened, synchronize, reopened, ready_for_review]
 
 concurrency:
   group: ${{ github.ref }}-${{ github.head_ref }}-codeql
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - "development"
   pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
 
 concurrency:
   group: ${{ github.ref }}-${{ github.head_ref }}-cuda
--- a/.github/workflows/hip.yml
+++ b/.github/workflows/hip.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - "development"
   pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
 
 concurrency:
   group: ${{ github.ref }}-${{ github.head_ref }}-hip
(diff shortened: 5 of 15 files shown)


EZoni:
To address https://github.com/BLAST-WarpX/warpx/pull/6117#issuecomment-3225910507, we could set this in our regular workflows
```yaml
on:
  pull_request:
    types: [opened, synchronize, reopened, ready_for_review]
```
(noting that `opened`, `synchronize` and `reopened` are the default types) and add `--draft` to the `gh pr create` command in the autoupdate workflows.

References:
- https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs
- https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows
- https://cli.github.com/manual/gh_pr_create

.github/workflows/clang_sanitizers.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - "development"
   pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
```
EZoni, line 8: Note that `opened`, `synchronize`, and `reopened` are the default types if nothing is specified, according to the documentation:
> More than one activity type triggers this event. For information about each activity type, see [Webhook events and payloads](https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request). By default, a workflow only runs when a `pull_request` event's activity type is `opened`, `synchronize`, or `reopened`. To trigger workflows by different activity types, use the types keyword. For more information, see [Workflow syntax for GitHub Actions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onevent_nametypes).

So here I am just adding `ready_for_review` to work around the CI issue.

EZoni:
@ax3l 

I implemented the tentative workaround, so this is ready for review.

.github/workflows/clang_sanitizers.yml
ax3l: Makes sense! I always wondered why ready for review did not trigger CI :D

ax3l (approved):
Splendid :rocket: :sparkles:

EZoni:
Now the Azure DevOps pipeline is the one who is not happy... :unamused: 

Because we set `drafts: false` in our pipeline file, and Azure does not support the concept of "ready for review", like GitHub Actions does:
https://github.com/BLAST-WarpX/warpx/blob/4c0c95bc7bc8e9ac0da3a684b9be23f7a5199094/.azure-pipelines.yml#L16-L24

A complete list of options supported by Azure for the `pr` event can be found at https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/pr?view=azure-pipelines:
```yaml
pr:
  autoCancel: boolean # Whether to cancel running PR builds when a new commit lands in the branch. Default: true.
  branches: # Branch names to include or exclude for triggering a run.
    include: [ string ] # List of items to include.
    exclude: [ string ] # List of items to exclude.
  paths: # File paths to include or exclude for triggering a run.
    include: [ string ] # List of items to include.
    exclude: [ string ] # List of items to exclude.
  drafts: boolean # Whether to start a run when a draft PR is created. Default: true.
```

So, I think we have three options:
1. Set `drafts: true` in the Azure pipeline file.
2. Revert the draft change and choose another workaround, e.g., close and re-open the autoupdate PR.
3. Drop Azure DevOps and migrate all workflows to GitHub Actions :sweat_smile: (see also #5291).

I think I'm leaning towards option 1., but I'm fine with option 2. as well. Option 3. is more long-term.

PR #6116, opened 2025-08-26
created an ImplicitOptions struct to pass into Evolve().

JustinRayAngus:
In future PRs, the PhysicalParticleContainer::Evolve() routine will require more parameters specific to the implicit methods. To avoid bloating the list of parameter passed into Evolve(), this PR creates an ImplicitOptions structure to pass into Evolve() that owns all of the implicit options needed by the particle push and deposit routines.

This allows direct and dynamic control of the what goes on in the Evolve() routine from the ImplicitSolver class.


Source/FieldSolver/ImplicitSolvers/ImplicitSolver_fwd.H
```diff
@@ -0,0 +1,12 @@
+#ifndef _IMPLICIT_SOLVER_FWD_H_
+#define _IMPLICIT_SOLVER_FWD_H_
+
+struct ImplicitOptions {
+    int max_particle_iterations;
+    double particle_tolerance;
+    bool deposit_mass_matrices;
+    bool linear_stage_of_jfnk;
+    int nonlinear_iteration;
```
EZoni: Where is `nonlinear_iteration` used? I searched the code base but did not understand where it is used, once accessed through the `options` input argument.

Source/FieldSolver/ImplicitSolvers/ImplicitSolver_fwd.H
```diff
@@ -0,0 +1,12 @@
+#ifndef _IMPLICIT_SOLVER_FWD_H_
+#define _IMPLICIT_SOLVER_FWD_H_
+
+struct ImplicitOptions {
+    int max_particle_iterations;
+    double particle_tolerance;
```
EZoni: Is there a reason why this needs to be `double` here instead of `amrex::ParticleReal`?

I see that we have things like
```cpp
const amrex::ParticleReal particle_tolerance = implicit_options->particle_tolerance;
```
in PhysicalParticleContainer.cpp, so I was wondering if the types should match.

Source/WarpX.cpp
```diff
@@ -118,8 +118,6 @@ Vector<int> WarpX::boost_direction = {0,0,0};
 bool WarpX::compute_max_step_from_btd = false;
 Real WarpX::zmin_domain_boost_step_0 = 0._rt;
 
-int WarpX::max_particle_its_in_implicit_scheme = 21;
-ParticleReal WarpX::particle_tol_in_implicit_scheme = 1.e-10;
```
EZoni, line 122: The fact that these defaults were removed and not replicated in the new struct (if I understand correctly), means that this PR effectively changes the previous default behavior of the code when used in the implicit mode, correct?

Source/Particles/RigidInjectedParticleContainer.H
```diff
@@ -71,8 +72,7 @@ public:
                  amrex::Real dt,
                  DtType a_dt_type=DtType::Full,
                  bool skip_deposition=false,
-                 bool deposit_mass_matrices=false,
-                 PushType push_type=PushType::Explicit) override;
+                 const ImplicitOptions* implicit_options = nullptr) override;
```
EZoni: @WeiqunZhang @ax3l 

Here we use `const X* p`, while elsewhere in this PR (e.g., in Source/Particles/PhysicalParticleContainer.cpp above) we use `X const * p`.

From the FAQ from https://isocpp.org/wiki/faq/const-correctness#const-ptr-vs-ptr-const:
> - `const X* p` means “p points to an X that is const”: the X object can’t be changed [via p](https://isocpp.org/wiki/faq/const-correctness#ptr-to-const-aliasing).
> - `X* const p` means “p is a const pointer to an X that is non-const”: you can’t change the pointer p itself, but you can change the X object [via p](https://isocpp.org/wiki/faq/const-correctness#ptr-to-const-aliasing).
> - `const X* const p` means “p is a const pointer to an X that is const”: you can’t change the pointer p itself, nor can you change the X object [via p](https://isocpp.org/wiki/faq/const-correctness#ptr-to-const-aliasing)

I dont' see the one we use in Source/Particles/PhysicalParticleContainer.cpp, `X const * p`, in this list. I think it's equivalent to `const X* p`, right?

In this PR, what should we use where, and why?

If the two versions are equivalent, would it be beneficial for the average code reader to agree on one and stick to it?

Source/Particles/PhysicalParticleContainer.H
```diff
@@ -91,8 +93,7 @@ public:
      * \param dt time step by which particles are advanced
      * \param a_dt_type type of time step (used for sub-cycling)
      * \param skip_deposition Skip the charge and current deposition.
-     * \param deposit_mass_matrices whether to deposit the mass matrices (in the case of the implicit solver)
-     * \param push_type Type of particle push, explicit or implicit. Defaults to explicit
+     * \param implicit_options (in the case of the implicit solver)
```
EZoni: The part in parenthesis here will be the whole docstring for this parameter:
Screenshot from 2025-08-26 13-48-47

Personally I wouldn't put it all in parenthesis, but up to you - I just wanted to post what it looks like.

From https://warpx--6116.org.readthedocs.build/en/6116/_static/doxyhtml/index.html (PR build).

Source/FieldSolver/ImplicitSolvers/ImplicitSolver_fwd.H
JustinRayAngus: It is not used yet. It will be used. I figured it was ok to go ahead and add it to the implicit options structure. Similar to how it is passed into PreRHSOp, but not used.

Source/FieldSolver/ImplicitSolvers/ImplicitSolver_fwd.H
JustinRayAngus: Since it's just a number to compare with and not used in any calculation, I figured double would not matter. The benefit is that the _fwd.H file doesn't care about using AMReX. If it's against  the standard practice, then I can change it AMReX::ParticleReal.

Source/FieldSolver/ImplicitSolvers/ImplicitSolver_fwd.H
EZoni: It is okay, I just wanted to make sure I wasn't missing something here.

Source/FieldSolver/ImplicitSolvers/ImplicitSolver_fwd.H
EZoni: I'm not sure if we have a standard practice for this.

@WeiqunZhang @ax3l What do you think?

Source/WarpX.cpp
JustinRayAngus: These defaults were never used because they were always overwritten by the members that live in ImplicitSolver.H. So default behavior is the same.

Source/Particles/RigidInjectedParticleContainer.H
JustinRayAngus: Good question. 
As far as I know, the following two are the same:
const X* p;
X const* p;
They both declare X as the thing that is constant and can't be changed.

Source/Particles/PhysicalParticleContainer.H
JustinRayAngus: I fixed this.

Source/Particles/RigidInjectedParticleContainer.H
JustinRayAngus: We have discussed this previously on slack. The "west const" vs "east const" battle. 

I changed all instances to be east const, consistent with other uses in WarpX for pointers to a constant.

Source/FieldSolver/ImplicitSolvers/ImplicitSolver_fwd.H
ax3l: For consistency, I would include `<AMReX_REAL.H>` and use the usual types, but since you directly assign it later, which might implicitly cast, before any calcs, it truly does not matter besides style. Ok for me as is, too.

Source/FieldSolver/ImplicitSolvers/ImplicitSolver_fwd.H
ax3l: @JustinRayAngus can you please assign default values here, to ensure the struct values are always well-defined?
```suggestion
    int max_particle_iterations = 1;
    double particle_tolerance = WarpX::particle_tol_in_implicit_scheme;  // see note
    bool deposit_mass_matrices = false;
    bool linear_stage_of_jfnk = false;
    int nonlinear_iteration = 1;
```

Otherwise, if we forget to set a value in usage we will get UB.

`WarpX::particle_tol_in_implicit_scheme`  would need an include. Maybe go the other way: define the default value of `particle_tol_in_implicit_scheme` here and use it to assign the default in `WarpX.cpp`

Source/FieldSolver/ImplicitSolvers/ImplicitSolver_fwd.H
```diff
@@ -0,0 +1,12 @@
+#ifndef _IMPLICIT_SOLVER_FWD_H_
+#define _IMPLICIT_SOLVER_FWD_H_
```
ax3l: This is not really a "forward declaration" file of C++ classes.

Instead, it is a file definining `ImplicitOptions` whole.

I would thus call the file `ImplicitOptions.H`

ax3l (commented):
This makes total sense to me.

Small suggestions:
- the file name for the `ImplictOptions` struct should maybe be reconsidered, it is not a forward-declaration but just a standalone file.
- Consider using this file for defaults and assign/use these defaults by including your new file queries to `ParmParse` and `WarpX.cpp` when querying/storing defaults in member variables.

Source/FieldSolver/ImplicitSolvers/ImplicitSolver_fwd.H
ax3l: While you refactor this file you could also use `amrex::ParticleReal` as per the other comment, but works either way for me.

Source/FieldSolver/ImplicitSolvers/ImplicitSolver_fwd.H
JustinRayAngus: Done.

Source/FieldSolver/ImplicitSolvers/ImplicitOptions.H
```diff
@@ -0,0 +1,14 @@
+#ifndef _IMPLICIT_OPTIONS_H_
+#define _IMPLICIT_OPTIONS_H_
```
WeiqunZhang: We might want to change the name here. Identifies with a leading `_` followed a capital letter are reserved.

https://en.cppreference.com/w/cpp/keywords.html

> each identifier that begins with an underscore followed by an uppercase letter is always reserved

JustinRayAngus: So just remove the leading underscore?

WeiqunZhang: Right.

Source/FieldSolver/ImplicitSolvers/ImplicitOptions.H
```diff
@@ -0,0 +1,15 @@
+#ifndef IMPLICIT_OPTIONS_H_
+#define IMPLICIT_OPTIONS_H_
+
+#include <AMReX_Particles.H>
+using namespace amrex::literals;
+
+struct ImplicitOptions {
+    int max_particle_iterations = 21;
```
EZoni, line 8: Very curious about this default value... :eyes: I know it matches the default value we had in Source/WarpX.cpp, but why was it 21 in the first place?

Source/Particles/PhysicalParticleContainer.H
```diff
@@ -81,18 +83,17 @@ public:
...
-     * \param deposit_mass_matrices whether to deposit the mass matrices (in the case of the implicit solver)
-     * \param push_type Type of particle push, explicit or implicit. Defaults to explicit
+     * \param fields The WarpX field register
+     * \param lev Level on which particles are living
+     * \param current_fp_string Current (coarse or fine patch) identifier in fields
+     * \param t Current physical time
+     * \param dt Time Step by which particles are advanced
+     * \param a_dt_type Type of time step (used for sub-cycling)
+     * \param skip_deposition Skip the charge and current deposition
+     * \param implicit_options Structure with specific options for implicit methods
```
EZoni, line 96: Thank you for polishing the docstrings here!

EZoni (approved):

EZoni:
@ax3l @WeiqunZhang 

This looks good to me. Please feel free to re-review the changes you requested.

Source/FieldSolver/ImplicitSolvers/ImplicitOptions.H
JustinRayAngus: I don't have a scientific answer. This number is chosen based on experience.

EZoni: I asked because I have always seen multiples of 10 as default values for these kinds of parameters - but there is no reason for that either, per se, except that maybe it makes it easier to do some mental math.

JustinRayAngus: I like to add 1 for good luck :)

PR #6120, opened 2025-08-27
Implicit push cleanup

JustinRayAngus:
This PR refactors the ImplicitPushPX() function and is a peal-off PR from PR #5969. The purpose of this PR is to 1) reduce the size of PR #5969 and 2) help track down something in PR #5969 that is causing small changes with the implicit solvers even when the new suborbit model is not being used.

I have confirmed that the changes in this PR yield identical results with the development branch.


ax3l (approved):
LGTM :+1: 

Thanks for splitting the PRs for the investigation in the other one.

PR #6123, opened 2025-08-27
 use span to allocate mesh data by default

guj:
So we can aggregate (mainly for BTD) data for a later flush. 

introduced an optional but  related entry in the input file: `<diag>.buffer_flush_limit_btd` by default it is 5. 

New implementation of #3200 but so far only for fields/meshes. Particles will be a follow-up PR.


Source/Diagnostics/FlushFormats/FlushFormatOpenPMD.H
```diff
@@ -53,6 +53,7 @@ public:
 private:
     /** This is responsible for dumping to file */
     std::unique_ptr< WarpXOpenPMDPlot > m_OpenPMDPlotWriter;
+    int m_NumAggBTDBufferToFlush=5;
```
ax3l: - Docstring needed, please.
- Formatting (pls see code around it).
```suggestion
    /** Please add a descriptive doc string here, Junmin. */
    int m_NumAggBTDBufferToFlush = 5;
```

Source/Diagnostics/FlushFormats/FlushFormatOpenPMD.cpp
```diff
@@ -64,6 +64,9 @@ FlushFormatOpenPMD::FlushFormatOpenPMD (const std::string& diag_name)
             ablastr::warn_manager::WMRecordWarning("Diagnostics", warnMsg);
             encoding = openPMD::IterationEncoding::groupBased;
         }
+
+        pp_diag_name.query("buffer_flush_limit_btd", m_NumAggBTDBufferToFlush);
+        amrex::Print()<<" BTD: ForceFlushEvery: "<<m_NumAggBTDBufferToFlush<<" Buffers per snapshot "<<std::endl;
```
ax3l: We usually do not print this level of detail for all input parameters.
```suggestion
```

Source/Diagnostics/FlushFormats/FlushFormatOpenPMD.cpp
```diff
@@ -176,6 +179,9 @@ FlushFormatOpenPMD::WriteToFile (
     m_OpenPMDPlotWriter->WriteOpenPMDParticles(
         particle_diags, static_cast<amrex::Real>(time), use_pinned_pc, isBTD, isLastBTDFlush);
 
+     if (bufferID % m_NumAggBTDBufferToFlush == 0)
+        m_OpenPMDPlotWriter->ForceFlush(isBTD);
```
ax3l: Do we also need to force a flush at the end of the simulation or so?

Source/Diagnostics/WarpXOpenPMD.H
```diff
@@ -152,6 +152,7 @@ public:
   /** Return OpenPMD File type ("bp5", "bp4", "h5" or "json")*/
   std::string OpenPMDFileType () { return m_OpenPMDFileType; }
 
+  void ForceFlush(bool isBTD);
```
ax3l:
```suggestion
  /** Docstring needed please! See other methods that have one */
  void ForceFlush (bool isBTD);
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -407,13 +407,27 @@ WarpXOpenPMDPlot::~WarpXOpenPMDPlot ()
...
+ * If I/O is through ADIOS:
+ *   isBTD=true =>  PerformPut
+ *      this way we do not flush out every buffer in a snapshot,
+ *      (BTD uses few data ranks so this is costly for ADIOS collective functions)
+ *      Instead we aggregate a few buffers before calling ForceFlush(isBTD) to write out.
+ *      Note that SPAN is used to allocate CPU data in ADIOS.
+ *      The advantage is when SPAN is successful,  PerformPut takes no action.
+ *
+ *   isBTD=false => PDW
+ */
```
ax3l: This kind of doc string belongs in the `.H` file, not the `.cpp` file. Please move over.

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -407,13 +407,27 @@ WarpXOpenPMDPlot::~WarpXOpenPMDPlot ()
...
 void WarpXOpenPMDPlot::flushCurrent (bool isBTD) const
 {
-     WARPX_PROFILE("WarpXOpenPMDPlot::flushCurrent");
-
-     openPMD::Iteration currIteration = GetIteration(m_CurrentStep, isBTD);
-
-     currIteration.seriesFlush();
+    openPMD::Iteration currIteration = GetIteration(m_CurrentStep, isBTD);
+    if (isBTD) {
+        WARPX_PROFILE("WarpXOpenPMDPlot::flushCurrent-PP()");
```
ax3l: Profiler string is not self-explanatory. Suggesting this:
```suggestion
        WARPX_PROFILE("WarpXOpenPMDPlot::flushCurrent()::BTD");
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -407,13 +407,27 @@ WarpXOpenPMDPlot::~WarpXOpenPMDPlot ()
...
+        currIteration.seriesFlush(  "adios2.engine.preferred_flush_target = \"buffer\"" );
```
ax3l: Why the odd formatting? Cleanup:
```suggestion
        currIteration.seriesFlush("adios2.engine.preferred_flush_target = \"buffer\"");
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -666,19 +681,37 @@ for (const auto & particle_diag : particle_diags) {
         pc->getCharge(), pc->getMass(),
         isBTD, isLastBTDFlush);
     }
+}
+
+/*
+ * Flush a few BTD buffers in a snapshot
+ *    controlled by FlushFormatOpenPMD::m_NumAggBTDBufferToFlush (default to 5)
+ *    can be adjusted in the input file: <diag>.buffer_flush_limit_btd
+ */
```
ax3l: Doxygen doc string belong into `.H` files, not into `.cpp` files. Please move over.

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -666,19 +681,37 @@ for (const auto & particle_diag : particle_diags) {
...
+void
+WarpXOpenPMDPlot::ForceFlush(bool isBTD)
+{
+    if (!isBTD)
```
ax3l:
```suggestion
    WARPX_PROFILE("WarpXOpenPMDPlot::ForceFlush()");

    if (!isBTD)
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -666,19 +681,37 @@ for (const auto & particle_diag : particle_diags) {
...
+    const bool result = (m_Series->backend() == "ADIOS2") && (hasOption != std::string::npos);
 
-    if (flattenSteps)
+    if (result)
     {
-       // forcing new step so data from each btd batch in
-       // preferred_flush_target="buffer" can be flushed out
-       openPMD::Iteration currIteration = GetIteration(m_CurrentStep, isBTD);
-       currIteration.seriesFlush(R"(adios2.engine.preferred_flush_target = "new_step")");
+        WARPX_PROFILE("WarpXOpenPMDPlot::FlattenSteps()");
```
ax3l: String is not self-explanatory. Suggestion:
```suggestion
        WARPX_PROFILE("WarpXOpenPMDPlot::ForceFlush()::Buffer");
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -666,19 +681,37 @@ for (const auto & particle_diag : particle_diags) {
...
+        openPMD::Iteration currIteration = GetIteration(m_CurrentStep, isBTD);
+        currIteration.seriesFlush(R"(adios2.engine.preferred_flush_target = "new_step")");
     }
+    else
+      {
+        WARPX_PROFILE("WarpXOpenPMDPlot::PDW()");
```
ax3l: String is not self-explanatory. Suggestion:
```suggestion
        WARPX_PROFILE("WarpXOpenPMDPlot::ForceFlush()::Disk");
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -666,19 +681,37 @@ for (const auto & particle_diag : particle_diags) {
...
+ */
+void
+WarpXOpenPMDPlot::ForceFlush(bool isBTD)
+{
+    if (!isBTD)
+        return;
 
     auto hasOption = m_OpenPMDoptions.find("FlattenSteps");
-    const bool flattenSteps = isBTD && (m_Series->backend() == "ADIOS2") && (hasOption != std::string::npos);
+    const bool result = (m_Series->backend() == "ADIOS2") && (hasOption != std::string::npos);
```
ax3l: I am confused. What does that have to do with flatten steps?

Source/Diagnostics/FlushFormats/FlushFormatOpenPMD.cpp
guj: So one needs to check the unused parameter list to see whether this parameter is taken info effect.

Source/Diagnostics/FlushFormats/FlushFormatOpenPMD.H
guj: added in the next commit

Source/Diagnostics/FlushFormats/FlushFormatOpenPMD.cpp
guj: will remove in the next commit

Source/Diagnostics/WarpXOpenPMD.cpp
guj: done.

Source/Diagnostics/WarpXOpenPMD.H
guj: moved over from cpp

Source/Diagnostics/WarpXOpenPMD.cpp
guj: moved

Source/Diagnostics/WarpXOpenPMD.cpp
guj: it is only effective for BTD. what is why I moved down.

Source/Diagnostics/WarpXOpenPMD.cpp
guj: ok, added

Source/Diagnostics/WarpXOpenPMD.cpp
guj: ok removed space

Source/Diagnostics/WarpXOpenPMD.cpp
guj: what is wrong with "flatten steps"?  the new name "buffer" does not reflect the fact  that a new step is written.

Source/Diagnostics/WarpXOpenPMD.cpp
guj: OK, renamed.

Source/Diagnostics/WarpXOpenPMD.cpp
guj: so this is ForceFlush()to Disk with a newstep created (for checkpointing)

Source/Diagnostics/WarpXOpenPMD.cpp
guj: Because with flatte step, data is written to disk and a new step is created (for checkpointing). Otherwise, just data been written out.

Source/Diagnostics/FlushFormats/FlushFormatOpenPMD.cpp
guj: At the end, there is adios Endstep() being called from CloseStep(), take care of the last flushing.

ax3l: Ok, perfect.

Source/Diagnostics/WarpXOpenPMD.H
```diff
@@ -152,6 +152,13 @@ public:
   /** Return OpenPMD File type ("bp5", "bp4", "h5" or "json")*/
   std::string OpenPMDFileType () { return m_OpenPMDFileType; }
 
+  /** Flush a few BTD buffers in a snapshot
+   * @param[in] isBTD if the current diagnostic is BTD
+   * This function is controlled by the paramter
+   * FlushFormatOpenPMD::m_NumAggBTDBufferToFlush (default to 5),
+   * it can be adjusted in the input file: diag_name.buffer_flush_limit_btd
```
ax3l: Making this valid Doxygen:
```suggestion
   *
   * This function is controlled by the parameter
   * FlushFormatOpenPMD::m_NumAggBTDBufferToFlush (default to 5),
   * it can be adjusted in the input file: diag_name.buffer_flush_limit_btd
   *
   * @param[in] isBTD if the current diagnostic is BTD
```

Source/Diagnostics/WarpXOpenPMD.H
```diff
@@ -181,7 +188,11 @@ private:
    * @param[in] isBTD if the current diagnostic is BTD
    *
    * if isBTD=false, apply the default flush behaviour
-   * if isBTD=true,  advice to use ADIOS Put() instead of PDW for better performance.
+   *                 in ADIOS, the action will be PerformDataWrite
+   * if isBTD=true,  in ADIOS, the action will be PerformPut
+   *                 because no action is taken for the span tasks.
```
ax3l, line 201: Is this generally true? You only add spans for fields right now, what if particles are involved?

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -1509,12 +1526,22 @@ WarpXOpenPMDPlot::WriteOpenPMDFieldsAll ( //const std::string& filename,
...
+                        WARPX_PROFILE("WarpXOpenPMDPlot::WriteOpenPMDFields::D2H_Span()");
+                        auto dynamicMemoryView = mesh_comp.storeChunk<amrex::Real>(
+                                                                                   chunk_offset, chunk_size,
+                                                                                   [&local_box](size_t size) {
+                                                                                      (void) size;
+                                                                                      amrex::Print()<<" span failed \n";
+                                                                                      amrex::BaseFab<amrex::Real> foo(local_box, 1, amrex::The_Pinned_Arena());
+                                                                                      std::shared_ptr<amrex::Real> data_pinned(foo.release());
+                                                                                      return data_pinned;
+                                                                                  });
```
ax3l: Please align this only 4 spaces to the last line above it. This super deep indentation makes it hard to read.

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -1509,12 +1526,22 @@ WarpXOpenPMDPlot::WriteOpenPMDFieldsAll ( //const std::string& filename,
...
-                    // intentionally delayed until before we .flush(): amrex::Gpu::streamSynchronize();
-                    mesh_comp.storeChunk(data_pinned, chunk_offset, chunk_size);
-                } else
+                    {
+                        WARPX_PROFILE("WarpXOpenPMDPlot::WriteOpenPMDFields::D2H_Span()");
+                        auto dynamicMemoryView = mesh_comp.storeChunk<amrex::Real>(
+                                                                                   chunk_offset, chunk_size,
+                                                                                   [&local_box](size_t size) {
+                                                                                      (void) size;
+                                                                                      amrex::Print()<<" span failed \n";
```
ax3l: Why is there a `span failed` print here? Debug leftover?

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -666,19 +670,32 @@ for (const auto & particle_diag : particle_diags) {
...
+
+void
+WarpXOpenPMDPlot::ForceFlush(bool isBTD)
+{
+    if (!isBTD)
+        return;
 
     auto hasOption = m_OpenPMDoptions.find("FlattenSteps");
-    const bool flattenSteps = isBTD && (m_Series->backend() == "ADIOS2") && (hasOption != std::string::npos);
+    const bool result = (m_Series->backend() == "ADIOS2") && (hasOption != std::string::npos);
```
ax3l: Please keep calling this `flattenSteps`, more descriptive than `result`, especially in the `if` below
```suggestion
    const bool flattenSteps = (m_Series->backend() == "ADIOS2") && (hasOption != std::string::npos);
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -666,19 +670,32 @@ for (const auto & particle_diag : particle_diags) {
...
-       currIteration.seriesFlush(R"(adios2.engine.preferred_flush_target = "new_step")");
+        WARPX_PROFILE("WarpXOpenPMDPlot::FlattenSteps()");
+        openPMD::Iteration currIteration = GetIteration(m_CurrentStep, isBTD);
+        currIteration.seriesFlush(R"(adios2.engine.preferred_flush_target = "new_step")");
     }
+    else
+      {
+        WARPX_PROFILE("WarpXOpenPMDPlot::ForceFlush()::Disk()");
+        openPMD::Iteration currIteration = GetIteration(m_CurrentStep, isBTD);
+        currIteration.seriesFlush(R"(adios2.engine.preferred_flush_target = "disk")");
```
ax3l, line 694: Why do we need the `else` branch now? Isn't `disk` the default?

Source/Diagnostics/WarpXOpenPMD.cpp
ax3l: Maybe just keep the single `WarpXOpenPMDPlot::ForceFlush()` then in the function?

Source/Diagnostics/WarpXOpenPMD.H
```diff
@@ -152,6 +152,15 @@ public:
...
 
+  /** Flush a few BTD buffers in a snapshot
+   *
+   * This function is controlled by the parameter
+   * FlushFormatOpenPMD::m_NumAggBTDBufferToFlush (default to 5),
+   * it can be adjusted in the input file: diag_name.buffer_flush_limit_btd
+   *
+   * @param[in] isBTD if the current diagnostic is BTD
+   */
+  void ForceFlush (bool isBTD);
```
ax3l:
```suggestion
  /** Ensure BTD buffers are written to disk
   *
   * This function can be called to intermediately ensure ADIOS buffered "steps"
   * are written to disk, including valid metadata. This is needed to read partial
   * data while a simulation is running or to support restarting (the BTD diagnostics)
   * in WarpX, so it can continue to append to a partially written labframe station
   * after restart.
   */
  void FlushBTDtoDisk ();
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -666,19 +670,32 @@ for (const auto & particle_diag : particle_diags) {
...
-    if (flattenSteps)
+    if (result)
     {
-       // forcing new step so data from each btd batch in
-       // preferred_flush_target="buffer" can be flushed out
-       openPMD::Iteration currIteration = GetIteration(m_CurrentStep, isBTD);
-       currIteration.seriesFlush(R"(adios2.engine.preferred_flush_target = "new_step")");
+        WARPX_PROFILE("WarpXOpenPMDPlot::FlattenSteps()");
+        openPMD::Iteration currIteration = GetIteration(m_CurrentStep, isBTD);
+        currIteration.seriesFlush(R"(adios2.engine.preferred_flush_target = "new_step")");
```
ax3l, line 688: Need an inline comment like this (please adjust for correctness)
```suggestion
        // in ADIOS2, this calls EndStep, which triggers a write to disk,
        // even if buffered
        currIteration.seriesFlush(R"(adios2.engine.preferred_flush_target = "new_step")");
```

Source/Diagnostics/WarpXOpenPMD.H
guj: You are right BTD particles will be calling PerformPut.  Right now particles is not the I/O performance bottleneck for BTD.  It will happen when you have enormous amount of particles.   I will definitely revisit this issue with another pull request.

Source/Diagnostics/WarpXOpenPMD.cpp
guj: yes, will remove it.

Source/Diagnostics/WarpXOpenPMD.cpp
guj: used FlushBTDToDisk() instead of ForceFlush()

Source/Diagnostics/WarpXOpenPMD.cpp
guj: ok

Source/Diagnostics/WarpXOpenPMD.cpp
guj: OK

Source/Diagnostics/WarpXOpenPMD.cpp
guj: we explained in zoom

Source/Diagnostics/WarpXOpenPMD.cpp
guj: done

Source/Diagnostics/WarpXOpenPMD.H
guj: added

Source/Diagnostics/WarpXOpenPMD.cpp
ax3l: Not yet changed :)

Source/Diagnostics/FlushFormats/FlushFormatOpenPMD.cpp
```diff
@@ -64,6 +64,8 @@ FlushFormatOpenPMD::FlushFormatOpenPMD (const std::string& diag_name)
             ablastr::warn_manager::WMRecordWarning("Diagnostics", warnMsg);
             encoding = openPMD::IterationEncoding::groupBased;
         }
+
+        pp_diag_name.query("buffer_flush_limit_btd", m_NumAggBTDBufferToFlush);
```
ax3l, line 68: Please document in `Docs/source/usage/parameters.rst`

guj: just did.

Source/Diagnostics/WarpXOpenPMD.cpp
guj: Should be in there now.

Source/Diagnostics/WarpXOpenPMD.cpp
guj: is in

Docs/source/usage/parameters.rst
```diff
@@ -2914,6 +2914,9 @@ In-situ capabilities can be used by turning on Sensei or Ascent (provided they a
      ``variable based`` is an `experimental feature with ADIOS2 BP5 <https://openpmd-api.readthedocs.io/en/0.16.1/backends/adios2.html#experimental-new-adios2-schema>`__ that will replace ``g``.
      Default: ``f`` (full diagnostics)
 
+* ``<diag_name>.buffer_flush_limit_btd`` (`integer`; defaults to 5) optional, only read if ``<diag_name>.diag_type = BackTransformed``
+    This parameter is intended for ADIOS backend to group every N buffers (N is the value of this parameter) and then flush to disk.
```
ax3l, line 2932: Does it work with any ADIOS2 format or only with BP5? In the latter case, I would write:
```suggestion
    This parameter is intended for ADIOS ``bp5`` backend to group every N buffers (N is the value of this parameter) and then flush to disk.
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -409,11 +409,14 @@ WarpXOpenPMDPlot::~WarpXOpenPMDPlot ()
...
-     openPMD::Iteration currIteration = GetIteration(m_CurrentStep, isBTD);
-
-     currIteration.seriesFlush();
+    openPMD::Iteration currIteration = GetIteration(m_CurrentStep, isBTD);
+    if (isBTD) {
+        WARPX_PROFILE("WarpXOpenPMDPlot::flushCurrent()::BTD");
+        currIteration.seriesFlush("adios2.engine.preferred_flush_target = \"buffer\"");
+    } else {
+        WARPX_PROFILE("WarpXOpenPMDPlot::flushCurrent()");
+        currIteration.seriesFlush();
```
ax3l: In a follow-up PR, we should rename this function to `WarpXOpenPMDPlot::seriesFlush` to be more descriptive.

"Current" means electrical current in WarpX, and the entry `WarpXOpenPMDPlot::flushCurrent` in the table will confuse them. (Not new to this PR, just noticed now.)

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -666,19 +670,32 @@ for (const auto & particle_diag : particle_diags) {
...
+void
+WarpXOpenPMDPlot::FlushBTDToDisk()
+{
+    bool isBTD = true;
     auto hasOption = m_OpenPMDoptions.find("FlattenSteps");
-    const bool flattenSteps = isBTD && (m_Series->backend() == "ADIOS2") && (hasOption != std::string::npos);
+    const bool doFlattenSteps = (m_Series->backend() == "ADIOS2") && (hasOption != std::string::npos);
 
-    if (flattenSteps)
+    if (doFlattenSteps)
```
ax3l: The `do` prefix does not really add anything, I would have kept the old name for this `bool`.

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -1509,12 +1526,21 @@ WarpXOpenPMDPlot::WriteOpenPMDFieldsAll ( //const std::string& filename,
...
-                    amrex::Gpu::dtoh_memcpy_async(data_pinned.get(), fab.dataPtr(icomp), local_box.numPts()*sizeof(amrex::Real));
-                    // intentionally delayed until before we .flush(): amrex::Gpu::streamSynchronize();
-                    mesh_comp.storeChunk(data_pinned, chunk_offset, chunk_size);
-                } else
+                    {
+                        WARPX_PROFILE("WarpXOpenPMDPlot::WriteOpenPMDFields::D2H_Span()");
+                        auto dynamicMemoryView = mesh_comp.storeChunk<amrex::Real>(
+                             chunk_offset, chunk_size,
+                             [&local_box](size_t size) {
+                                  (void) size;
```
ax3l: Simpler:
```suggestion
                             [&local_box](size_t /* size */) {
```

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -1509,12 +1526,21 @@ WarpXOpenPMDPlot::WriteOpenPMDFieldsAll ( //const std::string& filename,
...
+                                  return data_pinned;
+                              });
+
+                        auto span = dynamicMemoryView.currentBuffer();
+                        amrex::Gpu::dtoh_memcpy_async(span.data(), fab.dataPtr(icomp), local_box.numPts()*sizeof(amrex::Real));
+                    }
+                }  else
 #endif
                 {
                     amrex::Real const *local_data = fab.dataPtr(icomp);
```
ax3l: The `else` branch here is for CPU runs. You can replace the `mesh_comp.storeChunkRaw` as well with the span API, so CPU users of WarpX also get the speedup from the span API.

Docs/source/usage/parameters.rst
guj: uh bp4 supports span, so it is ok too.

Source/Diagnostics/WarpXOpenPMD.cpp
guj: hmm, true.  openpmd already has "seriesFlush".  confusing to use the same function name,.

Source/Diagnostics/WarpXOpenPMD.cpp
guj: fine with me

Source/Diagnostics/WarpXOpenPMD.cpp
guj: ok

Source/Diagnostics/WarpXOpenPMD.cpp
guj: Right, adding it now.

Source/Diagnostics/WarpXOpenPMD.cpp
ax3l: I think the same name is ok, because this is essentially a `seriesFlush` wrapper in `WarpXOpenPMDPlot::`

Docs/source/usage/parameters.rst
ax3l: Oh, cool! Thanks for clarifying!

Source/Diagnostics/WarpXOpenPMD.cpp
guj: Maybe "flushSeries"?

ax3l: It is no problem to use the exact same name, it is even helpful here :)

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -1509,21 +1528,53 @@ WarpXOpenPMDPlot::WriteOpenPMDFieldsAll ( //const std::string& filename,
...
+                        auto dynamicMemoryView = mesh_comp.storeChunk<amrex::Real>(
+                             chunk_offset, chunk_size,
+                             [&local_box](size_t /* size */) {
+                                  amrex::BaseFab<amrex::Real> foo(local_box, 1, amrex::The_Pinned_Arena());
+                                  std::shared_ptr<amrex::Real> data_pinned(foo.release());
+                                  return data_pinned;
+                              });
+                        auto span = dynamicMemoryView.currentBuffer();
+                        amrex::Gpu::dtoh_memcpy_async(span.data(), fab.dataPtr(icomp), local_box.numPts()*sizeof(amrex::Real));
+                    } else
```
ax3l, line 1543: Do we need this `else` branch for `D2H`? I thought openPMD-api has a fallback implementation for HDF5...?

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -1509,21 +1528,53 @@ WarpXOpenPMDPlot::WriteOpenPMDFieldsAll ( //const std::string& filename,
...
+                           [&local_box](size_t /* size */) {
+                              amrex::BaseFab<amrex::Real> foo(local_box, 1);
+                              std::shared_ptr<amrex::Real> data_pinned(foo.release());
+                              return data_pinned;
+                            });
+
+                      auto span = dynamicMemoryView.currentBuffer();
+                      std::memcpy(span.data(), fab.dataPtr(icomp), local_box.numPts()*sizeof(amrex::Real));
+                    }
+                  else
```
ax3l, line 1569: Same as above: I thought openPMD-api has a fallback implementation for Span for backends that do not support it?

Source/Diagnostics/WarpXOpenPMD.cpp
guj: fine fine

Source/Diagnostics/WarpXOpenPMD.cpp
guj: Turns out no.  We do need it.

Source/Diagnostics/WarpXOpenPMD.cpp
guj: I added this because the first round of CPU span failed pull request (it uses HDF5)  so the fallback does not work with collecting a few buffers  (during a snapshot) and then flush out.  The fallback works if flushes out immediately.

Source/Diagnostics/WarpXOpenPMD.cpp
ax3l: That might be an issue, can you please open an issue in https://github.com/openPMD/openPMD-api/ ?

ax3l (approved):
Thank you!

ax3l:
@guj thank you! As follow-ups, can you please
- open an openPMD-api issue that the fallback (for HDF5) for spans does not work? (please add a standalone reproducer)
- add span support to particles (i.e., extending/updating the work in #3200)

Thank you! :pray:

Source/Diagnostics/WarpXOpenPMD.cpp
```diff
@@ -1509,26 +1528,58 @@ WarpXOpenPMDPlot::WriteOpenPMDFieldsAll ( //const std::string& filename,
...
-                    amrex::BaseFab<amrex::Real> foo(local_box, 1, amrex::The_Pinned_Arena());
-                    std::shared_ptr<amrex::Real> data_pinned(foo.release());
-                    amrex::Gpu::dtoh_memcpy_async(data_pinned.get(), fab.dataPtr(icomp), local_box.numPts()*sizeof(amrex::Real));
-                    // intentionally delayed until before we .flush(): amrex::Gpu::streamSynchronize();
-                    mesh_comp.storeChunk(data_pinned, chunk_offset, chunk_size);
-                } else
+                    if (hasADIOS)
+                    {
+                        WARPX_PROFILE("WarpXOpenPMDPlot::WriteOpenPMDFields::D2H_Span()");
+                        auto dynamicMemoryView = mesh_comp.storeChunk<amrex::Real>(
```
ax3l, line 1534: Oh no, when we have an un-equal number of blocks over all MPI ranks, this will not work because `storeChunk` is collective :-o https://github.com/openPMD/openPMD-api/pull/1794

Source/Diagnostics/WarpXOpenPMD.cpp
ax3l: @franzpoeschel do you understand why this is needed?

Source/Diagnostics/WarpXOpenPMD.cpp
ax3l: To double check: is `Iteration::open()` called before these blocks, so the ADIOS engine is definitely open? -- Yes
https://github.com/guj/WarpX/blob/d0bbf79e64beed82b64d29f85ec0a56ed0f5f087/Source/Diagnostics/WarpXOpenPMD.cpp#L1426-L1430

guj: > Oh no, when we have an un-equal number of blocks over all MPI ranks, this will not work because `storeChunk` is collective :-o [openPMD/openPMD-api#1794](https://github.com/openPMD/openPMD-api/pull/1794)

Only one rank is writing when using BTD.  Other ranks have no data.

ax3l: > Only one rank is writing when using BTD. Other ranks have no data.

Yes, small correction: Only a _small subset of ranks_ is writing when using BTD. This is, because BTD stations collect data in (transverse) slices that move through the simulation.

PR #6125, opened 2025-08-28
[WIP] DSMC: implement generic two-product reaction

RemiLehe:
# Overview

This replaces the DSMC `CHARGE_EXCHANGE` process by a `TWO_PRODUCT_REACTION` process, whereby:
- The reaction can have an associated energy cost (or gain) - while `CHARGE_EXCHANGE` assumed no energy cost/gain.
- The reactant and products can have any mass/charge - while `CHARGE_EXCHANGE` implicitly assumed that reactants and products had the same masses, and imposed that their charge is incremented/decremented by one during the reaction.

This change is backward compatible for users ; input scripts using `charge_exchange` will work, but will internally call the new `TWO_PRODUCT_REACTION` (with 0 energy cost/gain)

# Implementation

This leverages the function `TwoProductComputeProductMomenta` which is also used in the fusion code. This function has been modified to add an option to not scatter the products isotropically (which is used in DSMC)

As a result the calculation is valid for relativistic particles too.

TODO:
- [ ] Add documentation ; explain differences between charge-exchange and two-product ; explain scattering angle.
- [ ] Add CI test that check for energy and momentum conservation.
- [ ] Apply correct transformation for RZ
- [ ] Should we make a version of `TwoProduct...` that is non-relativistic?


Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
```diff
@@ -66,7 +67,7 @@ DSMCFunc::DSMCFunc (
             reaction_produces_new_species = true;
         }
 
-        if (process.type() == ScatteringProcessType::IONIZATION) {
+        if (scattering_process.find("excitation") != std::string::npos) {
```
RemiLehe:
```suggestion
        if (scattering_process.find("ionization") != std::string::npos) {
```

RemiLehe:
Superseded by #6250

PR #6127, opened 2025-08-28
FieldGather: Enable Lower D

ax3l:
Enable interpolating fields to particles in 1D or 2D even when running a 3D simulation. Needed for 2D and 2.5D space charge in ImpactX: https://github.com/BLAST-ImpactX/impactx/pull/909


EZoni (approved):
Approving even if my review wasn't requested here.

PR #6130, opened 2025-08-30
FlushFormat int comp writing

JustinRayAngus:
This PR fixes an indexing bug related to writing integer components of particles in WarpX. 

This PR fixes a bug in PR #5969, where an integer value is added to the particle container at runtime and an error occurs on restarts.

### DIFF
--- a/Source/Diagnostics/FlushFormats/FlushFormatCheckpoint.cpp
+++ b/Source/Diagnostics/FlushFormats/FlushFormatCheckpoint.cpp
@@ -237,7 +237,7 @@ FlushFormatCheckpoint::CheckpointParticles (
         write_int_comps.resize(pc->NumIntComps());
         //   note: inames and h_redistribute_int_comp are note the same size
         auto inames = pc->GetIntSoANames();
-        std::size_t const i0_redist = pc->h_redistribute_int_comp.size() - pc->NumIntComps();
+        std::size_t const i0_redist = pc->h_redistribute_int_comp.size() - inames.size();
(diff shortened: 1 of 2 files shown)


Source/Diagnostics/FlushFormats/FlushFormatCheckpoint.cpp
```diff
@@ -235,10 +235,12 @@ FlushFormatCheckpoint::CheckpointParticles (
         // and the int comps
         int_names.resize(pc->NumIntComps());
         write_int_comps.resize(pc->NumIntComps());
+        //   note: inames and h_redistribute_int_comp are note the same size
```
JustinRayAngus: @WeiqunZhang @atmyers are these Vectors supposed to have different sizes? I ask because the analogous Vectors for Reals are the same size.

atmyers: As you noticed, I believe there's an inconsistency in how this is handled in AMReX. The question is whether you want these vectors to include the *required* components or not (the particle positions for the real numbers, the ids for the ints). It looks we do not include them for the reals, but do for the integers.

dpgrote (approved):
This looks good to me! This will be safe if AMReX is updated so that the names list and redistribute lists are made to be consistent.

EZoni:
@JustinRayAngus 

#6136 fixed the CI macOS issue, please feel free to rebase on `development` to clear the CI checks here.

atmyers (approved):

PR #6136, opened 2025-09-02
CI: workaround for macOS issue

EZoni:
Workaround for https://github.com/actions/runner-images/issues/12912 to avoid the following error:
```
Error: cmake is already installed from local/pinned!
Please `brew uninstall cmake` first."
```

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -47,6 +47,7 @@ jobs:
         brew install pkg-config
         set -e
         brew tap openpmd/openpmd
+        brew uninstall cmake  # https://github.com/actions/runner-images/issues/12912
         brew install openpmd-api
     - name: install pip dependencies
       run: |
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -70,7 +70,7 @@ jobs:
         export CCACHE_SLOPPINESS=time_macros
         ccache -z
 
-        export CXXFLAGS="-Werror -Wno-error=pass-failed"
+        export CXXFLAGS="-Werror -Wno-error=pass-failed -Wno-error=c++23-attribute-extensions"
 
         cmake -S . -B build_dp         \
           -DCMAKE_VERBOSE_MAKEFILE=ON  \


EZoni:
Now that we manage to pass the brew dependencies installation step, we get this error:

```
/Users/runner/work/warpx/warpx/Source/Diagnostics/ComputeDiagFunctors/CellCenterFunctor.cpp:23:5: error: use of the 'assume' attribute is a C++23 extension [-Werror,-Wc++23-attribute-extensions]
   23 |     AMREX_ASSUME(m_mf_src != nullptr);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/runner/work/warpx/warpx/build_dp/_deps/fetchedamrex-src/Src/Base/AMReX_Extension.H:232:39: note: expanded from macro 'AMREX_ASSUME'
  232 | #   define AMREX_ASSUME(ASSUMPTION) [[assume(ASSUMPTION)]]
      |                                     ~~^~~~~~~~~~~~~~~~~~~~
1 error generated.
```

@WeiqunZhang

Do we need to support C++23 if we use `AMREX_ASSUME`? Why does this show up only now, though?

.github/workflows/macos.yml
```diff
@@ -47,6 +47,7 @@ jobs:
         brew install pkg-config
         set -e
         brew tap openpmd/openpmd
+        brew uninstall cmake  # https://github.com/actions/runner-images/issues/12912
         brew install openpmd-api
```
EZoni, line 51: Note that cmake is installed as a dependency of openpmd-api.

ax3l:
@EZoni as a quick fix, add `-Wno-c++23-attribute-extensions` to the CXXFLAGS for macOS CI.

We generally want to avoid relying on C++20/23 features at this point as mandatory, but the implementation here looks absolutely safe:
https://github.com/AMReX-Codes/amrex/blob/25.09/Src/Base/AMReX_Extension.H#L225-L245

ax3l:
@WeiqunZhang I saw the same noisy warning btw on Tuo (LLNL) with the AMD/Clang compilers there. We need to find a way to make this less noisy for general users.

.github/workflows/macos.yml
ax3l: ok. the macOS runners and brew are oscillating pre-installed stuff all the time. I sometimes have lines like `... || true` appended to guesstimates.

EZoni:
@ax3l 

Sounds good, I will add the flag to skip that class of warnings. I still don't understand why this shows up only now, given that we have had `AMREX_ASSUME` in the code for a while.

ax3l:
@EZoni That works, or we suppress it at the source @WeiqunZhang . Should we add add LLVM-specific warning suppressions in assume code of `AMReX_Extension.H`?

ax3l (approved):
LGTM, thanks for the fix to CI.

PR #6137, opened 2025-09-04
Update picmi.py

yanghf263:
The default value for gamma should be changed from 3/2 to 5/3 for the adiabatic index of an ideal gas.


dpgrote (approved):
Thanks for the fix! Note that the default value of gamma is 5/3 (as set in HybridPICModel/HybridPICModel.H). It is only the comment that is incorrect.

PR #6138, opened 2025-09-04
[WIP] Workaround for AMD w/ MPI bug on Tuolumne

EZoni:
Similar to the workaround already implemented at https://github.com/BLAST-WarpX/warpx/blob/d103a0e3446af6cb63e240e25904bdf9554671a7/Source/ablastr/parallelization/MPIInitHelpers.cpp#L51-L59

### DIFF
--- a/.github/workflows/dependencies/hip.sh
+++ b/.github/workflows/dependencies/hip.sh
@@ -46,7 +46,6 @@ sudo apt-get update
 sudo apt-get install -y --no-install-recommends \
     build-essential \
     gfortran        \
-    libhsa-runtime-dev \
     libhiredis-dev  \
     libnuma-dev     \
     libopenmpi-dev  \
--- a/.github/workflows/dependencies/hip.sh
+++ b/.github/workflows/dependencies/hip.sh
@@ -87,6 +87,6 @@ if [ ! -e /opt/rocm ] && [ -d /opt/rocm-${VERSION} ]; then
     sudo ln -s /opt/rocm-${VERSION} /opt/rocm
 fi
 
-# Ensure the HSA runtime library is found
-export LIBRARY_PATH=/opt/rocm/lib:${LIBRARY_PATH:-}
-export LD_LIBRARY_PATH=/opt/rocm/lib:${LD_LIBRARY_PATH:-}
+## Ensure the HSA runtime library is found
+#export LIBRARY_PATH=/opt/rocm/lib:${LIBRARY_PATH:-}
+#export LD_LIBRARY_PATH=/opt/rocm/lib:${LD_LIBRARY_PATH:-}
--- a/.github/workflows/dependencies/hip.sh
+++ b/.github/workflows/dependencies/hip.sh
@@ -86,7 +86,3 @@ export CEI_TMP="/tmp/cei"
 if [ ! -e /opt/rocm ] && [ -d /opt/rocm-${VERSION} ]; then
     sudo ln -s /opt/rocm-${VERSION} /opt/rocm
 fi
-
-## Ensure the HSA runtime library is found
-#export LIBRARY_PATH=/opt/rocm/lib:${LIBRARY_PATH:-}
-#export LD_LIBRARY_PATH=/opt/rocm/lib:${LD_LIBRARY_PATH:-}
--- a/.github/workflows/dependencies/hip.sh
+++ b/.github/workflows/dependencies/hip.sh
@@ -46,7 +46,7 @@ sudo apt-get update
 sudo apt-get install -y --no-install-recommends \
     build-essential \
     gfortran        \
-    hsa-runtime-dev \
+    libhsa-runtime-dev \
     libhiredis-dev  \
     libnuma-dev     \
     libopenmpi-dev  \
--- a/.github/workflows/dependencies/hip.sh
+++ b/.github/workflows/dependencies/hip.sh
@@ -81,3 +81,8 @@ sudo curl -L -o /usr/local/bin/cmake-easyinstall https://raw.githubusercontent.c
 sudo chmod a+x /usr/local/bin/cmake-easyinstall
 export CEI_SUDO="sudo"
 export CEI_TMP="/tmp/cei"
+
+# Point /opt/rocm to /opt/rocm-${VERSION} to find headers
+if [ ! -e /opt/rocm ] && [ -d /opt/rocm-${VERSION} ]; then
+    sudo ln -s /opt/rocm-${VERSION} /opt/rocm
+fi
--- a/.github/workflows/dependencies/hip.sh
+++ b/.github/workflows/dependencies/hip.sh
@@ -88,5 +88,5 @@ if [ ! -e /opt/rocm ] && [ -d /opt/rocm-${VERSION} ]; then
 fi
 
 # Ensure the HSA runtime library is found
-export LIBRARY_PATH=/opt/rocm/lib:$LIBRARY_PATH
-export LD_LIBRARY_PATH=/opt/rocm/lib:$LD_LIBRARY_PATH
+export LIBRARY_PATH=/opt/rocm/lib:${LIBRARY_PATH:-}
+export LD_LIBRARY_PATH=/opt/rocm/lib:${LD_LIBRARY_PATH:-}
--- a/.github/workflows/dependencies/hip.sh
+++ b/.github/workflows/dependencies/hip.sh
@@ -46,6 +46,7 @@ sudo apt-get update
 sudo apt-get install -y --no-install-recommends \
     build-essential \
     gfortran        \
+    hsa-runtime-dev \
     libhiredis-dev  \
     libnuma-dev     \
     libopenmpi-dev  \
--- a/.github/workflows/dependencies/hip.sh
+++ b/.github/workflows/dependencies/hip.sh
@@ -86,3 +86,7 @@ export CEI_TMP="/tmp/cei"
 if [ ! -e /opt/rocm ] && [ -d /opt/rocm-${VERSION} ]; then
     sudo ln -s /opt/rocm-${VERSION} /opt/rocm
 fi
+
+# Ensure the HSA runtime library is found
+export LIBRARY_PATH=/opt/rocm/lib:$LIBRARY_PATH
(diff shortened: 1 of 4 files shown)


Source/ablastr/parallelization/MPIInitHelpers.cpp
```diff
@@ -20,6 +20,11 @@
 #include <hip/hip_runtime.h>
 #endif
 
+// INC0509050: Handle MPI initialization/finalization errors on Tuolumne (LLNL)
+#if defined(AMREX_USE_HIP)
```
EZoni, line 24: Is this the right macro or should it be `AMREX_USE_ROCTX`?

Source/ablastr/parallelization/MPIInitHelpers.cpp
```diff
@@ -54,7 +59,16 @@ namespace ablastr::parallelization
...
+            throw std::runtime_error("hipInit failed. Did not proceed with MPI_Init_thread.");
+        }
+#endif
+
+        // INC0509050: Handle MPI initialization error on Tuolumne (LLNL)
+#if defined(AMREX_USE_HIP) && defined(AMREX_USE_MPI)
+        hsa_status_t status = hsa_init();
+        if (status != HSA_STATUS_SUCCESS) {
+            std::cerr << "hsa_init failed with error code " << status << "! Aborting now.\n";
+            throw std::runtime_error("hsa_init failed. Did not proceed with MPI_Init_thread.");
```
EZoni, line 71: Do we want to throw a runtime error or handle the exception differently?

Source/ablastr/parallelization/MPIInitHelpers.cpp
```diff
@@ -75,6 +89,15 @@ namespace ablastr::parallelization
     {
 #ifdef AMREX_USE_MPI
         MPI_Finalize();
+
+        // INC0509050: Handle MPI finalization error on Tuolumne (LLNL)
+#ifdef AMREX_USE_HIP
+        hsa_status_t status = hsa_shut_down();
+        if (status != HSA_STATUS_SUCCESS) {
+            std::cerr << "hsa_shut_down failed with error code " << status << "! Aborting now.\n";
+            throw std::runtime_error("hsa_shut_down failed after MPI_Finalize completed.");
```
EZoni, line 98: Do we want to throw a runtime error or handle the exception differently?

Source/ablastr/parallelization/MPIInitHelpers.cpp
```diff
@@ -54,7 +59,16 @@ namespace ablastr::parallelization
         hipError_t hip_ok = hipInit(0);
         if (hip_ok != hipSuccess) {
             std::cerr << "hipInit failed with error code " << hip_ok << "! Aborting now.\n";
-            throw std::runtime_error("hipInit failed. Did not proceeding with MPI_Init_thread.");
+            throw std::runtime_error("hipInit failed. Did not proceed with MPI_Init_thread.");
```
EZoni, line 62: Note that for the previous MPI initialization (on Crusher) error we do throw a runtime error.

Source/ablastr/parallelization/MPIInitHelpers.cpp
atmyers: This is the one we want.

.github/workflows/dependencies/hip.sh
```diff
@@ -46,6 +46,7 @@ sudo apt-get update
 sudo apt-get install -y --no-install-recommends \
     build-essential \
     gfortran        \
+    hsa-runtime-dev \
```
EZoni, line 49:
```suggestion
    libhsa-runtime-dev \
```

.github/workflows/hip.yml
```diff
@@ -52,6 +52,9 @@ jobs:
         export CXX=$(which clang++)
         export CC=$(which clang)
 
+        export LIBRARY_PATH=/opt/rocm/lib:${LIBRARY_PATH:-}
+        export LD_LIBRARY_PATH=/opt/rocm/lib:${LD_LIBRARY_PATH:-}
+
```
atmyers:
```suggestion
        export LDFLAGS="-lhsa-runtime64 ${LDFLAGS}"
```

.github/workflows/hip.yml
```diff
@@ -107,6 +110,9 @@ jobs:
...
```
atmyers:
```suggestion
        export LDFLAGS="-lhsa-runtime64 ${LDFLAGS}"
```

.github/workflows/dependencies/hip.sh
```diff
@@ -81,3 +81,12 @@ sudo curl -L -o /usr/local/bin/cmake-easyinstall https://raw.githubusercontent.c
...
 export CEI_TMP="/tmp/cei"
+
+# Point /opt/rocm to /opt/rocm-${VERSION} to find headers
+if [ ! -e /opt/rocm ] && [ -d /opt/rocm-${VERSION} ]; then
+    sudo ln -s /opt/rocm-${VERSION} /opt/rocm
+fi
+
+## Ensure the HSA runtime library is found
+#export LIBRARY_PATH=/opt/rocm/lib:${LIBRARY_PATH:-}
+#export LD_LIBRARY_PATH=/opt/rocm/lib:${LD_LIBRARY_PATH:-}
```
EZoni: To remove:
```suggestion
```

.github/workflows/hip.yml
```diff
@@ -52,6 +52,9 @@ jobs:
         export CXX=$(which clang++)
         export CC=$(which clang)
 
+        export LIBRARY_PATH=/opt/rocm/lib:${LIBRARY_PATH:-}
+        export LD_LIBRARY_PATH=/opt/rocm/lib:${LD_LIBRARY_PATH:-}
+        export LDFLAGS="-lhsa-runtime64 ${LDFLAGS}"
```
EZoni, line 57: @ax3l 

I already added the export of `LDFLAGS` (which is a HSA-specific change) to the Tuolumne profile file. Do you think we need to add the other two as well (`LIBRARY_PATH`, `LD_LIBRARY_PATH`)?

Asking because I still don't understand why exporting `LIBRARY_PATH` and `LD_LIBRARY_PATH` turned out to be needed now and not before - so not sure if it was just a CI environment issue here.

.github/workflows/dependencies/hip.sh
```diff
@@ -81,3 +81,8 @@ sudo curl -L -o /usr/local/bin/cmake-easyinstall https://raw.githubusercontent.c
 sudo chmod a+x /usr/local/bin/cmake-easyinstall
 export CEI_SUDO="sudo"
 export CEI_TMP="/tmp/cei"
+
+# Point /opt/rocm to /opt/rocm-${VERSION} to find headers
+if [ ! -e /opt/rocm ] && [ -d /opt/rocm-${VERSION} ]; then
+    sudo ln -s /opt/rocm-${VERSION} /opt/rocm
+fi
```
ax3l, line 88: This looks wrong to me, probably something to search with CMake for (HSA)?

.github/workflows/hip.yml
ax3l: This works as a quick-fix, but I think we need a CMake `find_package` call to be robust and portable (maybe in AMReX).

.github/workflows/dependencies/hip.sh
ax3l: There is probably a `hsa-runtime` package we can use (or similarly named):
https://rocm.docs.amd.com/projects/install-on-linux/en/latest/install/install-methods/package-manager/package-manager-ubuntu.html

EZoni: I agree I don't fully understand why we didn't need it before, but I can say that without this we couldn't find the hsa.h file when we try to `#include` it. This made us bypass that error somehow.

.github/workflows/hip.yml
ax3l: https://rocm.docs.amd.com/projects/ROCR-Runtime/en/latest/install/installation.html#building-against-the-runtime

.github/workflows/dependencies/hip.sh
ax3l: We need to use `find_package`, see https://rocm.docs.amd.com/projects/ROCR-Runtime/en/latest/install/installation.html#building-against-the-runtime

We did not directly use the runtime before.

`/opt/rocm` is hard-coded in a lot of ROCm logic, while via CMake we can use more general (i.e., module) installation paths.

ax3l:
I think this patch turns out to be superfluous: from an earlier MPI init bug on early Frontier, we already have `hipInit()` in the code to init the HIP (and HSA) runtime.
https://rocm.docs.amd.com/projects/HIP/en/docs-develop/how-to/hip_runtime_api/initialization.html

> # Initialization
>
> The initialization involves setting up the environment and resources needed for using GPUs. The following steps are covered with the initialization:
> 
> - Setting up the HIP runtime
> 
>     This includes reading the environment variables set during init, setting up the active or visible devices, loading necessary libraries, setting up internal buffers for memory copies or cooperative launches, initialize the compiler as well as HSA runtime and checks any errors due to lack of resources or no active devices.
> 
> - Querying and setting GPUs
> 
>     Identifying and querying the available GPU devices on the system.
> 
> - Setting up contexts
> 
>     Creating contexts for each GPU device, which are essential for managing resources and executing kernels. For further details, check the [context section](https://rocm.docs.amd.com/projects/HIP/en/docs-develop/how-to/hip_porting_driver_api.html#context-driver-api).
> 
> ## Initialize the HIP runtime
> 
> The HIP runtime is initialized automatically when the first HIP API call is made. However, you can explicitly initialize it using [hipInit()](https://rocm.docs.amd.com/projects/HIP/en/docs-develop/reference/hip_runtime_api/modules/initialization_and_version.html#_CPPv47hipInitj), to be able to control the timing of the initialization. The manual initialization can be useful to ensure that the GPU is initialized and ready, or to isolate GPU initialization time from other parts of your program.

EZoni:
Flagging this as "[WIP]" after @ax3l's latest comment and ready to close without merging if confirmed that this is superfluous.

PR #6139, opened 2025-09-05
Speedup MultiFab registry

dpgrote:
This PR is a possible solution to speed up the MultiFab registry, when getting MultiFabs.

Most MultiFabs in WarpX are specified by the enum FieldType, and when the MultiFab is fetched (using `get`), the enum is converted to a string since the MultiFab map keys are strings. This conversion is done using the `amrex::getEnumNameString` routine. That routine then internally calls `amrex::getEnumNameValuePairs` which generates a vector of the name, value pairs for whole enum type. Then `getEnumNameString` returns the one name requested. This means that every time a `get` is done, the whole name, value vector for the enum is generated.

The `FieldType` enum has 67 values, so creating that vector is somewhat expensive. For 2D and 3D, this time is not significant, but for 1D it can be. In a test case, this was taking up about half of the run time (since it was done almost 30 million times).

There are two relatively simple solutions. The first is to not use the `FieldType` enum, but always use strings for the MultiFabs. A second solution is to cache the names. This PR shows what that might look like.

This cache is not perfect. This assumes that the int values of each of the enums is unique. This is true in WarpX, but some other code could break this (by using more than one enum type for the MultiFabs for example).

### DIFF
--- a/Source/ablastr/fields/MultiFabRegister.H
+++ b/Source/ablastr/fields/MultiFabRegister.H
@@ -35,6 +35,8 @@ namespace
     template <typename T>
     struct is_castable_to_string<T, std::void_t<decltype(static_cast<std::string>(std::declval<T>()))>> : std::true_type {};
 
+    std::map<int, std::string> name_map;
+
     /** helper to either cast a string/char array to string to query an AMREX_ENUM */
     template<typename T>
     std::string getExtractedName (T name)
@@ -45,8 +47,19 @@ namespace
             return std::string(name);
         } else
         {
-            // user-defined AMREX_ENUM or compile error
-            return amrex::getEnumNameString(name);
+            // Check if the string name had already been generated.
+            // If so, return that, otherwise generate it.
+            // The generated names are cached in the map since the
+            // generation of the string name (getEnumNameString) can be relatively slow.
+            int k = static_cast<int>(name);
+            if (name_map.count(k) > 0) {
+                return name_map.at(k);
+            } else {
+                // user-defined AMREX_ENUM or compile error
+                std::string v = amrex::getEnumNameString(name);
+                name_map.emplace(k, v);
+                return v;
+            }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


ax3l:
Wow, good catch!!

Do we maybe want to implement a faster to-string routine in AMReX for `AMReX_Enum`, that only creates one string at a time, that we can use here?

RemiLehe:
Interesting!
Is there a way to not call `amrex::getEnumNameValuePairs` for every call to `amrex::getEnumNameString`. 
Could we call it only once at initialization?

ax3l:
This might achieve the same caching, but upstream :)
https://github.com/AMReX-Codes/amrex/pull/4643

AlexanderSinn:
Just curious, could the key type of the m_mf_register map be changed from string to the enum type (int)? Or is the access by string still more common?

dpgrote:
> Just curious, could the key type of the m_mf_register map be changed from string to the enum type (int)? Or is the access by string still more common?

Access using the enum and string are needed. In some cases the same MF is accessed with the enum and string so it needs to be consistent. There are a number of MFs where the name is generated dynamically (for instance to include the species name).

JustinRayAngus:
@ax3l While the recent commit to AMReX (https://github.com/AMReX-Codes/amrex/pull/4643) improves things, it doesn't seem to be as big of an improvement compared to the changes in this PR.

Using the 1D uniform plasma test that I have showed during KISMET meetings, here are the wall times:
Original: 398 s
with AMReX/4643: 197 s
with this PR: 179 s

I've ran the tests several times and get the same result - This PR is about 10% faster.

WeiqunZhang:
Could you try this? https://github.com/AMReX-Codes/amrex/pull/4654

JustinRayAngus:
> Could you try this? [AMReX-Codes/amrex#4654](https://github.com/AMReX-Codes/amrex/pull/4654)

That seems to work just as good as this PR.

ax3l:
Fixed upstream via https://github.com/AMReX-Codes/amrex/pull/4654
Will be pulled in with the next weekly AMReX update.

ax3l:
Thank you @dpgrote @JustinRayAngus and @WeiqunZhang for identifying, testing and fixing this!

ax3l:
@JustinRayAngus @dpgrote if you have the 1D benchmark you used available and could add it to the WarpX CI tests, I would be curious to also benchmark it at some point :)

PR #6140, opened 2025-09-05
Doc: Flux (Tuo@LLNL)

ax3l:
- [x] tested on Tuo

Document the Flux scheduler:
- add a cheat sheet & reference link
- add a job script for Tuo @ LLNL


Tools/machines/tuolumne-llnl/tuolumne_mi300a.flux
```diff
@@ -45,10 +51,14 @@ export OMP_NUM_THREADS=16
 # GPU-aware MPI optimizations
 GPU_AWARE_MPI="amrex.use_gpu_aware_mpi=1"
 
-# MPI parallel processes
+# APUs share memory with the host:
+# Do NOT pre-allocate a large heap in AMReX
+# RELEASE IMMEDIATELY in arena
+APU_SHARED_MEMORY="amrex.the_arena_init_size=1"  # amrex.the_arena_release_threshold=0"
```
ax3l: Gotta double check what we actually used now.

ax3l: This should go away, I think. Outdated.

Docs/source/install/batch/flux.rst
```diff
@@ -0,0 +1,45 @@
...
+* `interactive job <https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man1/flux-submit.html>`__:
+
+  * ``flux submit --time-limit=1:00:00 --nodes=1 --tasks-per-node=4 --cores-per-task=8``
+
+    * e.g. ``flux submit "hostname"``
+  * GPU allocation requires additional flags, e.g. ``--gpus-per-task=1``
+
+* details for my jobs:
+
+  * ``flux jobs -a`` all jobs under my user name
```
dpgrote: This might be confusing since this will include completed jobs as well as active jobs, which could be a long list. I would simplify this and only show ``flux jobs`` which shows only active jobs.

Docs/source/install/batch/flux.rst
```diff
@@ -0,0 +1,45 @@
...
+  * ``flux job info 12345 jobspec`` all details for job with <job id> ``12345``
```
dpgrote: It might be worth commenting that the job ids will be an alphanumeric string (instead of a number which is typical of other batch systems). I had to poke at it a bit to understand what the string of random characters was.

Docs/source/install/batch/flux.rst
ax3l: Good point!
```suggestion
  * ``flux jobs`` all jobs under my user name
```

Docs/source/install/batch/flux.rst
ax3l:
```suggestion
  * ``flux job info abc123 jobspec`` all details for job with <job id> ``abc123``
```

ax3l:
@dpgrote @debog ready to go! :)

dpgrote (approved):
Looks good!

EZoni:
I merged `development` to pull the fix in #6258 and make sure that the CI check for Read the Docs passes. I will enable auto-merge after my push so this gets merged, since @dpgrote already approved.

PR #6143, opened 2025-09-07
Add density floor and time filtering to effective potential solver

roelof-groenewald:
~Currently, the `sigma` MF used to dress the Poisson equation in the semi-implicit "effective potential" electrostatic solver uses the charge density **deposited** at the nodes, which requires an interpolation step to get the charge density at the cell-centers. This PR corrects the calculation of sigma by directly using the cell average number density via the `GetNumberDensity` function.~

The "effective potential" solver is updated to allow use of a density floor in the calculation of sigma and to use time filtering of sigma - this substantially reduces noise in the electrostatic potential.
The solver performance is also improved by only depositing the charge density once per step instead of twice as in the current `development` version.

The `GetNumberDensity` function is also exposed to Python since I used it to test the a prototype solver (1d) with the above mentioned change.


JustinRayAngus:
Does the particle shape factor not matter here?

JustinRayAngus:
I'm sure this is explained in the paper on the method, but I thought that the potential phi lives at cell nodes in ES PIC. Since your Poisson solver needs the density at cell centers, it seems that phi lives at cell centers for this method. Is that correct? If yes, why is it different than the conventional nodal centering of phi for ES PIC?

roelof-groenewald:
> I'm sure this is explained in the paper on the method, but I thought that the potential phi lives at cell nodes in ES PIC. Since your Poisson solver needs the density at cell centers, it seems that phi lives at cell centers for this method. Is that correct? If yes, why is it different than the conventional nodal centering of phi for ES PIC?

Thanks for the comments @JustinRayAngus. The potential still lives on the cell nodes. In this method we solve the modified Poisson equation:

$-\nabla\cdot\varepsilon_0\left( 1 + \sum_{p\in species}C_{SI}\frac{\omega_p^2\Delta t^2}{4} \right)\nabla\phi = -\nabla\cdot\varepsilon_0\sigma\nabla\phi = \rho$

and the way the linear solver for this form is implemented in AMReX, it uses a cell-centered $\sigma$.

> Does the particle shape factor not matter here?

That's why I marked it WIP. I'm going to discuss that with Dan Barnes this week. The same question applies to whether any filtering done on $rho$ should also be applied to $\sigma$.

dpgrote:
Another question, does the sigma need to be consistent with rho?

In Warp, when implementing this model, rho was deposited at the nodes, and then averaged as needed to calculate the sigmas with the correct centerings when doing the finite differences.

JustinRayAngus:
> Another question, does the sigma need to be consistent with rho?
> 
> 
> 
> In Warp, when implementing this model, rho was deposited at the nodes, and then averaged as needed to calculate the sigmas with the correct centerings when doing the finite differences.

This makes sense to me. I would think that you need sigma at the same location as \nabla\phi, which would be cell edges for a Yee grid. It's funny to me that the solver in AMReX wants sigma at cell centers.

roelof-groenewald:
@dpgrote and @JustinRayAngus this PR is now ready for review. I was experimenting a bit with different strategies of accumulating $sigma$ but it turned out using the shape function to gather the density resulted in the best energy conservation. Furthermore, we found that applying a floor on the density corresponding to roughly one macro-particle per cell when setting up sigma, significantly reduced the noise in the resulting electrostatic potential. I also modified the code to no longer perform the charge deposition twice, which helps with performance.

roelof-groenewald:
We've now tested these changes quite substantially and they are important to get robust evolution.
I think it would be good to merge this before it becomes too old and starts to require continuous rebasing.

@JustinRayAngus @dpgrote

Source/FieldSolver/ElectrostaticSolvers/EffectivePotentialES.H
```diff
@@ -63,12 +65,15 @@ public:
...
         ablastr::fields::MultiLevelVectorField const& efield,
-        amrex::MultiFab const& sigma,
+        std::unique_ptr<amrex::MultiFab> const& sigma,
         amrex::Real required_precision,
         amrex::Real absolute_tolerance,
         int max_iters,
         int verbosity
     ) const;
+
+    std::unique_ptr<amrex::MultiFab> m_sigma;
```
dpgrote: One more comment - could this MF be put in the multifab registry?

Source/FieldSolver/ElectrostaticSolvers/EffectivePotentialES.cpp
```diff
@@ -133,47 +135,68 @@ void EffectivePotentialES::ComputeSigma (MultiFab& sigma) const
...
+                    ))
                 );
                 // add species term to sigma:
                 // C_SI * w_p^2 * dt^2 / 4 = C_SI / 4 * q*rho/(m*eps0) * dt^2
-                sigma_arr(i, j, k, 0) += mult_factor_pc * rho_cc;
+                sigma_arr(i, j, k, 0) += time_filter_param * mult_factor_pc * rho_cc;
             });
         }
     }
+    m_sigma->plus(time_filter_param, 0);
```
JustinRayAngus: @roelof-groenewald I'm trying to follow how the time_filter_param is used. Can you explain this line here?

Reading above it looks like you have something like
`m_sigma = (1-time_filter_param)*sigma_old + time_filter_param*sigma_new`.

Why do you then add time_filter_param to m_sigma after this?

roelof-groenewald: Thanks for the question @JustinRayAngus. The reason is because the "dressing" for the Poisson equation is 1 + [plasma frequency term], i.e., the updated coefficient is:
$$\sigma^t = 1 + \sum_p C_{SI} \frac{\omega_p^2\Delta t^2}{4} \rightarrow f + f \sum_p C_{SI} \frac{\omega_p^2\Delta t^2}{4} + (1 - f)\times(\sigma^{t-1})$$

Does that clear it up?

JustinRayAngus (approved):
Looks good to me.

Issue #5283, opened 2024-09-17
Add the option to load a WarpX input file in a Python environment


RemiLehe:
When switching from using the WarpX executable to using Python/PICMI (e.g. in order to take advantage of the Python callback functions), users usually have to rewrite their input script **from scratch**. (i.e. switching from the input script syntax to the PICMI syntax)

This is often time consuming and error prone. Instead it could be useful to have the option to load an input script in Python, e.g.:
```python
sim = Simulation.from_input_script( './inputs_3d' )  # loads an existing input script
installcallback('afterstep', myplots)  # optional: add callback function
sim.step()  # runs the simulation
```

One could use a syntax similar to the one in ImpactX:
https://github.com/ECP-WarpX/impactx/blob/24.09/src/python/ImpactX.cpp#L79

which relies on:
https://github.com/AMReX-Codes/pyamrex/blob/development/src/Base/ParmParse.cpp#L33

See also:
https://github.com/AMReX-Codes/amrex/pull/4156

EZoni:
I have a feeling that https://github.com/AMReX-Codes/pyamrex/pull/429, if developed further, might play a role in this, no?

ax3l:
I actually already implemented this in ImpactX & pyAMReX a long time ago, and will try to give it a stab for WarpX today :)

This is the existing AMReX functionality it leverages:
- [ImpactX](https://github.com/BLAST-ImpactX/impactx/blob/25.08/src/python/ImpactX.cpp#L73-L76)
- [pyAMReX ParmParse addFile](https://pyamrex.readthedocs.io/en/latest/usage/api.html#amrex.space3d.ParmParse.addfile)
- [AMReX (fixes)](https://github.com/AMReX-Codes/amrex/pulls?q=is%3Apr+addFile+is%3Aclosed+author%3Aax3l)

ax3l:
First working example in #6154, but have to work on urgent proposals now.

dpgrote:
An alternative way of dealing with this situation is shown in PR #6149, where the input file is Python (without using PICMI) and is constructed very similarly to the standard C++ input file. Only relatively minor changes are needed to convert between the two.

ax3l:
Discussed today in the developer meeting:

We like both #6154 (documented for users) and #6149 (as a test for now, no user-facing advertising), follow-up ideas are:
[...]

PR #6149, opened 2025-09-09
Add example of basic python input file

dpgrote:
This PR adds an example using the lower level Python interface for WarpX. Here, the input parameters are set directly instead of by using the PICMI interface. This uses a fairly straightforward example to demonstrate the basic method. More advanced capabilities can be setup this way as well, for instance adding call backs to Python routines.

Related to #6149 and #5283


Examples/Tests/plasma_lens/inputs_test_3d_plasma_lens.py
```diff
@@ -0,0 +1,77 @@
...
+
+particles.E_ext_particle_init_style = "repeated_plasma_lens"
+particles.B_ext_particle_init_style = "repeated_plasma_lens"
+particles.repeated_plasma_lens_period = 0.5
+particles.repeated_plasma_lens_starts = 0.1, 0.11, 0.12, 0.13
+particles.repeated_plasma_lens_lengths = 0.1, 0.11, 0.12, 0.13
+particles.repeated_plasma_lens_strengths_E = 600000.0, 800000.0, 600000.0, 200000.0
+particles.repeated_plasma_lens_strengths_B = 0.0, 0.0, 0.0, 0.0
+
+# Diagnostics
```
dpgrote: Note that the complicated code below could be wrapped in a routine `Diagnostics.new_diagnostic`, similar to the `newspecies` routine above.

Examples/Tests/plasma_lens/inputs_test_3d_plasma_lens.py
```diff
@@ -0,0 +1,77 @@
...
+algo.particle_shape = 1
+warpx.cfl = 0.7
+
+vel_z = 0.5 * clight
+
+# particles
+particles.species_names = ("electrons",)
+
+electrons = Particles.newspecies("electrons")
+electrons.charge = -q_e
```
dpgrote: This will be slightly different behavior since the `-q_e` will be evaluated in the input file rather than in the C++ parameter parsing. This could also be
```electrons.charge = "-q_e"```
which would be evaluated in C++.

File: Examples/Tests/plasma_lens/inputs_test_3d_plasma_lens.py [hunk 3]
```diff
@@ -0,0 +1,77 @@
...
+# Maximum number of time steps
+max_step = 84
+
+# number of grid points
+amr.n_cell = 16, 16, 16
+
+amr.max_level = 0
+
+# Geometry
+geometry.dims = "3"
```
dpgrote: Because of the way this parameter is parsed, this must be a string.

Examples/Tests/plasma_lens/inputs_test_3d_plasma_lens_python.py
```diff
@@ -0,0 +1,77 @@
...
+
+particles.E_ext_particle_init_style = "repeated_plasma_lens"
+particles.B_ext_particle_init_style = "repeated_plasma_lens"
+particles.repeated_plasma_lens_period = 0.5
+particles.repeated_plasma_lens_starts = 0.1, 0.11, 0.12, 0.13
+particles.repeated_plasma_lens_lengths = 0.1, 0.11, 0.12, 0.13
+particles.repeated_plasma_lens_strengths_E = 600000.0, 800000.0, 600000.0, 200000.0
+particles.repeated_plasma_lens_strengths_B = 0.0, 0.0, 0.0, 0.0
+
+# Diagnostics
```
dpgrote, line 64: Note that the complicated code below could be wrapped in a routine Diagnostics.new_diagnostic, similar to the newspecies routine above.

Examples/Tests/plasma_lens/inputs_test_3d_plasma_lens_python.py
```diff
@@ -0,0 +1,77 @@
...
+algo.particle_shape = 1
+warpx.cfl = 0.7
+
+vel_z = 0.5 * clight
+
+# particles
+particles.species_names = ("electrons",)
+
+electrons = Particles.newspecies("electrons")
+electrons.charge = -q_e
```
dpgrote, line 45: This will be slightly different behavior since the -q_e will be evaluated in the input file rather than in the C++ parameter parsing. This could also be
```electrons.charge = "-q_e"```
which would be evaluated in C++.

File: Examples/Tests/plasma_lens/inputs_test_3d_plasma_lens_python.py [hunk 6]
```diff
@@ -0,0 +1,77 @@
...
+# Maximum number of time steps
+max_step = 84
+
+# number of grid points
+amr.n_cell = 16, 16, 16
+
+amr.max_level = 0
+
+# Geometry
+geometry.dims = "3"
```
dpgrote: Because of the way this parameter is parsed, this must be a string.

ax3l:
Discussed today in the developer meeting:

We like both and will merge both this and #6149 for now, follow-up ideas are:
- Generalize (and rename) the Bucket to support nested ParmParse options by default – remove derived classes
- With that, use the load file in 6154 to populate the nested ParmParse Bucket, so we can manipulate it.
- Enable to update ParmParse interactively through changes to Parmparse Bucket, i.e., in WarpX interactive use

With those updates, the logic here will be a little less rough (e.g., see `Diagnostics`) and can then be documented for users together with the example in #6149 (i.e., load file and then change parameters on "ParmParse Bucket")

Examples/Tests/plasma_lens/inputs_test_3d_plasma_lens_python.py
ax3l: Or we generalize the `Bucket` to support nested dicts by default.

Examples/Tests/plasma_lens/inputs_test_3d_plasma_lens_python.py
ax3l: We could generalize `libwarpx` to assume Cartesian, if an integer is passed. We already do this assumption/convention in the ParmParse:
```ini
geometry.dims = 1  # Cartesian
geometry.dims = 2  # Cartesian
geometry.dims = 3  # Cartesian
geometry.dims = RCYLINDER
geometry.dims = RSPHERE
geometry.dims = RZ
```

ax3l (approved):
LGTM :)

ax3l (approved):

PR #6152, opened 2025-09-09
Change y to z in WarpX 2D examples

RemiLehe:
In 2D, in WarpX, the variables are `x` and `z`. It is good to reflect this in the name of the Python variables used in the scripts.

Many thanks to [yanghf263](https://github.com/yanghf263) for pointing out this issue.


Examples/Tests/pass_mpi_communicator/inputs_test_2d_pass_mpi_comm_picmi.py
```diff
@@ -138,7 +138,7 @@
 # NOTE: Some tests are done in this input PICMI file. These tests
 # check that amrex initialized the correct amount of procs and
 # that the procs ranks in amrex are correct.
-# If any of these tests fail, the terminal will print that the
+# If anz of these tests fail, the terminal will print that the
```
RemiLehe:
```suggestion
# If any of these tests fail, the terminal will print that the
```

EZoni (approved):

PR #6153, opened 2025-09-10
UB sanitizer test : print stack trace for easier debugging

lucafedeli88:
By doing `export UBSAN_OPTIONS=print_stacktrace=1` we can print stack traces when running the Undefined Behavior sanitizer. This can considerably speed up debugging.

### DIFF
--- a/.github/workflows/clang_sanitizers.yml
+++ b/.github/workflows/clang_sanitizers.yml
@@ -145,6 +145,9 @@ jobs:
         # This option is required to avoid false positive reports from the OpenMP runtime
         export TSAN_OPTIONS='ignore_noninstrumented_modules=1'
 
+        # Print stacktrace for easier debugging
+        export UBSAN_OPTIONS=print_stacktrace=1
+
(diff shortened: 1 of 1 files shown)


.github/workflows/clang_sanitizers.yml
```diff
@@ -145,6 +145,9 @@ jobs:
         # This option is required to avoid false positive reports from the OpenMP runtime
         export TSAN_OPTIONS='ignore_noninstrumented_modules=1'
 
+        # Print stacktrace for easier debugging
+        export UBSAN_OPTIONS=print_stacktrace=1
+
```
EZoni: This is currently added to the `build_thread_sanitizer` job. Did you mean to add it to the `build_UB_sanitizer` job instead?

lucafedeli88: Oh, right! I am extremely sorry !

lucafedeli88: It has been corrected.

EZoni (approved):

PR #6154, opened 2025-09-10
Python: `load_inputs_file`

### LINKED ISSUE #5283: Add the option to load a WarpX input file in a Python environment

ax3l:
Provide a rapid transition path for users from inputs files to Python APIs. Close #5283

## To Do

- [x] first working example
- [x] generalize
- [x] add hooks in example
- [x] [document new API/workflow](https://warpx--6154.org.readthedocs.build/en/6154/usage/workflows/python_extend.html)


Source/Python/pyWarpX.cpp
```diff
@@ -125,6 +125,11 @@ PYBIND11_MODULE(PYWARPX_MODULE_NAME, m) {
     m.def("amrex_finalize", [] () { amrex::Finalize(); },
         "Close out the amrex related data");
 
+    m.def("load_inputs_file",
+        [](std::string const & filename) {
+            amrex::ParmParse::addfile(filename);
+        });
+
```
ax3l: Actually not using this right now.
```suggestion
```

Python/pywarpx/WarpX.py
```diff
@@ -107,6 +107,64 @@ def get_bucket(self, bucket_name):
             self._bucket_dict[bucket_name] = bucket
             return bucket
 
+    @staticmethod
+    def read_dims_from_file(filename, visited=None):
```
dpgrote, line 100: There is already Tools/Parser/input_file_parser.py. Could this be reused? (Though would have to be updated to handle FILE.)

.github/workflows/source/check_inputs.py
```diff
@@ -92,7 +93,10 @@
     print("      Test names must end with '_picmi' for PICMI tests.")
 if wrong_testinput:
     print("NOTE: Test input names must start with 'inputs_' followed by the test name.")
-    print("      Test input names must end with '.py' for PICMI tests.")
+    print("      Test input names must end with '_picmi.py' for PICMI tests.")
+    print(
+        "      Test input names must start with 'run_' and end with '.py' for non-PICMI Python tests."
```
dpgrote: Is there a reason to require the "run" prefix for non-PICMI Python input files?

Python/pywarpx/WarpX.py
ax3l: Oh thanks, I was not aware of this! :)

.github/workflows/source/check_inputs.py
ax3l: In ImpactX, I called all Python scripts `run_...py` because I [found it more descriptive](https://impactx.readthedocs.io/en/latest/usage/how_to_run.html#how-to-run).

I think the naming here with `inputs_...` (for Python scripts) is a little confusing, because we refer to AMReX "ParmParse" input files as `inputs files` quite often.

Either way, for this script here we need a new naming convention because it is not "PICMI", but the script / convention expects this so far. Open to any ideas :)

Python/pywarpx/WarpX.py
ax3l: Hm, is that one standalone or installed with WarpX?

There are a few updates that it needs with regards to parser generality and recursively includes. Cannot be easily used from here just now.

Examples/Physics_applications/laser_acceleration/inputs_test_3d_laser_acceleration_py.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python3
+
+from pywarpx import fields, particle_containers, warpx
+
+# Callback locations: https://warpx.readthedocs.io/en/latest/usage/workflows/python_extend.html#callback-locations
+from pywarpx.callbacks import callfromafterstep
+
+# Define simulation from inputs file
+sim = warpx
```
ax3l: While the implementation here in the `WarpX` class works for a first attempt, it is a bit weird to put in this class that derives from `Bucket` (for PICMI argv building). But less weird than putting it in the `Simulation` class.

One could go a little "lower level" and with that make access to `warpx = libwarpx.libwarpx_so` and the pybind11 classes (like actual `warpx = libwarpx.libwarpx_so.get_instance()`) less awkward, but maybe it's ok for now... Maybe I can add the `sim.extension` attribute also to the class here as a shortcut.

There is also a lot of logic I like to reuse from `_libwarpx.py` right now when it comes to shared lib loading.

Might be a general access/API improvement that we need to simplify (also for PICMI files).

Examples/Physics_applications/laser_acceleration/inputs_test_3d_laser_acceleration_py.py
```diff
@@ -0,0 +1,60 @@
...
+    # electrons: access (and potentially manipulate)
+    electrons_pc = warpx.multi_particle_container().get_particle_container_from_name(
+        "electrons"
+    )
+    print(f"    {electrons_pc}")
+
+    # electric field: access (and potentially manipulate)
+    multifab_register = warpx.multifab_register()
+    Direction = sim.extension.libwarpx_so.Direction
+    Ex_mf = multifab_register.get("Efield_fp", dir=Direction(0), level=0)
```
ax3l: With #6162 we can write
```suggestion
    Ex_mf = multifab_register.get("Efield_fp", dir=Direction.x, level=0)
```
and
```suggestion
    Ex_mf = multifab_register.get("Efield_fp", dir=Direction("x"), level=0)
```
and even
```suggestion
    Ex_mf = multifab_register.get("Efield_fp", dir="x", level=0)
```

Examples/Physics_applications/laser_acceleration/CMakeLists.txt
```diff
@@ -82,6 +82,16 @@ add_warpx_test(
...
 
+add_warpx_test(
+    test_3d_laser_acceleration_py  # name
+    3  # dims
+    2  # nprocs
+    inputs_test_3d_laser_acceleration_py.py  # inputs
+    OFF  # analysis
+    "analysis_default_regression.py --path diags/diag1000100"  # checksum
+    OFF  # dependency
+)
```
ax3l, line 93: I probably need to copy the inputs files (`inputs_test_3d_laser_acceleration` and the included base file) into the run directory:
```
  File "/home/vsts/work/1/s/Examples/Physics_applications/laser_acceleration/inputs_test_3d_laser_acceleration_py.py", line 10, in <module>
    sim.load_inputs_file("./inputs_test_3d_laser_acceleration")
    sim.load_inputs_file("./inputs_test_3d_laser_acceleration")
  File "/home/vsts/work/1/s/build/lib/site-packages/pywarpx/WarpX.py", line 167, in load_inputs_file
  File "/home/vsts/work/1/s/build/lib/site-packages/pywarpx/WarpX.py", line 167, in load_inputs_file
    raise RuntimeError(
RuntimeError: Error: Could not find the geometry.dims in the input file './inputs_test_3d_laser_acceleration' or its included files.
```

ax3l: Fixed (we do the same in ImpactX)

ax3l:
Oops, forgot to write a checksum file, I think
```
The following tests FAILED:
	476 - test_3d_laser_acceleration_py.checksum (Failed)
```

.github/workflows/source/check_inputs.py
```diff
@@ -92,7 +92,6 @@
     print("      Test names must end with '_picmi' for PICMI tests.")
 if wrong_testinput:
     print("NOTE: Test input names must start with 'inputs_' followed by the test name.")
```
ax3l, line 94:
```suggestion
    print("NOTE: Test input names must start with 'inputs_' followed by the test name.")
    print(" Test input names must end with '.py' for PICMI tests.")
```

Python/pywarpx/WarpX.py
ax3l: To be improved separately, as discussed in the developer meeting today.
Once we make this a `warpx-tools` package and auto-install it as a sub-package of `pywarpx` we can use it here! :)

Let's improve this part in a follow-up PR.

.github/workflows/source/check_inputs.py
```diff
@@ -92,7 +92,7 @@
...
-    print("      Test input names must end with '.py' for PICMI tests.")
+    print(" Test input names must end with '.py' for PICMI tests.")
```
ax3l:
```suggestion
    print("      Test input names must end with '.py' for PICMI tests.")
```

Examples/Physics_applications/laser_acceleration/inputs_test_3d_laser_acceleration_py.py
ax3l: updated :)

dpgrote (approved):
This looks good, thanks!

ax3l:
@dpgrote thanks! I rebased to remove the merge conflict we had from merging your PR first. Ready for merge now :)

PR #6155, opened 2025-09-11
Fix loading photons from openPMD file

aeriforme:
When injecting particles from an openPMD file using  `<species>.injection_style = external_file`, the momentum is normalized using the mass of the particles. This fails with photons. This PR changes the mass used to normalize the momentum of massless particles to `m_e` when injecting from an external file.

### DIFF
--- a/Source/Particles/ParticleCreation/AddParticles.cpp
+++ b/Source/Particles/ParticleCreation/AddParticles.cpp
@@ -642,7 +642,7 @@ PhysicalParticleContainer::AddPlasmaFromFile(PlasmaInjector & plasma_injector,
 
             if (plasma_injector.insideBounds(x, y, z)) {
 
-                amrex::Particle real const mass_eff = (mass > 0.0_prt) ? mass : PhysConst::m_e;
+                amrex::ParticleReal const mass_eff = (mass > 0.0_prt) ? mass : PhysConst::m_e;
                 amrex::ParticleReal const ux = ptr_ux.get()[i]*momentum_unit_x/mass_eff;
                 amrex::ParticleReal const uz = ptr_uz.get()[i]*momentum_unit_z/mass_eff;
                 amrex::ParticleReal uy = 0.0_prt;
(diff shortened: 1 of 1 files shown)


Source/Particles/ParticleCreation/AddParticles.cpp
```diff
@@ -641,11 +641,13 @@ PhysicalParticleContainer::AddPlasmaFromFile(PlasmaInjector & plasma_injector,
 #endif
 
             if (plasma_injector.insideBounds(x, y, z)) {
-                amrex::ParticleReal const ux = ptr_ux.get()[i]*momentum_unit_x/mass;
-                amrex::ParticleReal const uz = ptr_uz.get()[i]*momentum_unit_z/mass;
+
+                amrex::Particle real const mass_eff = (mass > 0.0_prt) ? mass : PhysConst::m_e;
```
EZoni:
```suggestion
                amrex::ParticleReal const mass_eff = (mass > 0.0_prt) ? mass : PhysConst::m_e;
```

Source/Particles/ParticleCreation/AddParticles.cpp
```diff
@@ -641,11 +641,13 @@ PhysicalParticleContainer::AddPlasmaFromFile(PlasmaInjector & plasma_injector,
 #endif
 
             if (plasma_injector.insideBounds(x, y, z)) {
-                amrex::ParticleReal const ux = ptr_ux.get()[i]*momentum_unit_x/mass;
-                amrex::ParticleReal const uz = ptr_uz.get()[i]*momentum_unit_z/mass;
+
+                amrex::ParticleReal const mass_eff = (mass > 0.0_prt) ? mass : PhysConst::m_e;
```
EZoni: I wonder if an inline comment would help, saying something about the fact that (for some reason that I don't remember) we use the electron mass for photons?

aeriforme: Yes, sure. Thing is, that's what we do throughout the whole code. The reason is that the processes that involve photons are QED processes that involve electrons or positrons. Other masses wouldn't make sense (e.g. proton mass). Maybe the question can be change into: why do we normalize the momentum? That I don't know for sure.

EZoni: Right, I think that's probably the correct question.

EZoni (approved):

PR #6156, opened 2025-09-11
[WIP] check that cell crossings does not exceed threshold

JustinRayAngus:
The implicit solvers permit time steps that violate the CFL condition for particles. Sometimes this results in a particle gathering or depositing from/to an invalid position that is outside the ghost region of the box, which results in a segfault with no information about why the segfault occurred.

This PR adds a check that the number of particle cell crossings in the Villasenor gather/deposit routines is within the number permitted based on the number of ghost cells. The code aborts when a violation occurs with an appropriate error message.


JustinRayAngus:
@atmyers mentioned during a meeting that amrex::numParticlesOutOfRange() should handle the same situation this PR is targeting. We believe it is not catching the out-of-bounds particles for the implicit solver because the particles are at the time-centered position. This may require a separate PR to change it such that the implicit particles own the position at step n+1 rather than n+1/2. For now, this PR is converted to a WIP.

JustinRayAngus:
> @atmyers mentioned during a meeting that amrex::numParticlesOutOfRange() should handle the same situation this PR is targeting. We believe it is not catching the out-of-bounds particles for the implicit solver because the particles are at the time-centered position. This may require a separate PR to change it such that the implicit particles own the position at step n+1 rather than n+1/2. For now, this PR is converted to a WIP.

Looks like amrex::numParticlesOutOfRange() is  only called inside AMREX_ASSERT_WITH_MESSAGE(), which I believe only works when compiled with debug. This PR offers a light way method for checking the particles are within range when using the Villasenor depositions using parameters that are already computed - no extra work outside comparing two int values.

One non-ideal aspect of this check is that it will abort if any particle has a range larger than that permitted by the ghost cells, even in situations where the particle is not near the box boundary and no segfault would occur otherwise. However, probably not that non-ideal because such particles will most likely be at a box boundary eventually.

JustinRayAngus:
@WeiqunZhang @dpgrote @ax3l Question remains on what to do for GPU, since AMReX::Abort() doesn't work on device.

dpgrote:
> @WeiqunZhang @dpgrote @ax3l Question remains on what to do for GPU, since AMReX::Abort() doesn't work on device.

You can do what is done in the implicit push for unconverged particles. Create a `Gpu::Buffer` to count the number of bad particles, and check it after the kernel is finished.

Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -1599,9 +1603,31 @@ void VillasenorDepositionShapeNKernel ([[maybe_unused]]amrex::ParticleReal const
...
-    // e.g., if (num_segments > 7) ...
+    // Abort if cell crossings are larger than permitted
+    if (cell_crossings_x > max_crossings) {
+        amrex::Gpu::Atomic::AddNoRet(out_of_range_particles_ptr,1);
+#if !defined(AMREX_USE_GPU)
+        amrex::Abort("Error: cell_crossings_x = " + std::to_string(cell_crossings_x) +
+                     " is greater than the maximum allowed = " +
+                     std::to_string(max_crossings));
+#endif
+    }
```
ax3l, line 1614: This looks quite expensive for hot loops in kernels.

We could consider either checking this only in debug mode or compiling (as a template parameter) to variants that we can select at runtime (one checked and one unchecked).

JustinRayAngus: Thanks for the feedback. This PR has gone somewhat stale. Weiqun had a suggestion above, but for now this is on the back burner. Many more important things to worry about.

JustinRayAngus:
Closing this PR. It was replaced by PR https://github.com/BLAST-WarpX/warpx/pull/7225

PR #6162, opened 2025-09-11
MF Register Direction: Descriptive Members

ax3l:
Simplify calls t other `MultiFabRegister` by adding descriptive member variables to `ablastr::fields::Direction` as well as string conversion. Fix internal values for `dir` to make sense in other dimensions than 3D, too.

Isolated from #6154

## To Do

- [x] improve implementation
- [x] update usage / call sites RZ/RCylindrical/RSpherical: `Direction{0} -> 'r'`, etc.
- [x] update usage / call sites Cartesian: `Direction{0} -> 'x'`, etc. where explicitly Cartesian code is written

### DIFF
--- a/Source/Diagnostics/BTDiagnostics.cpp
+++ b/Source/Diagnostics/BTDiagnostics.cpp
@@ -615,7 +615,7 @@ BTDiagnostics::UpdateVarnamesForRZopenPMD ()
     auto & warpx = WarpX::GetInstance();
     auto & fields = warpx.m_fields;
     using ablastr::fields::Direction;
-    const int ncomp_multimodefab = fields.get(FieldType::Efield_aux, Direction{0}, 0)->nComp();
+    const int ncomp_multimodefab = fields.get(FieldType::Efield_aux, Direction::r, 0)->nComp();
     const int ncomp = ncomp_multimodefab;
 
 
@@ -677,7 +677,7 @@ BTDiagnostics::InitializeFieldFunctorsRZopenPMD (int lev)
 
     auto & warpx = WarpX::GetInstance();
     auto & fields = warpx.m_fields;
-    const int ncomp_multimodefab = fields.get(FieldType::Efield_aux, Direction{0}, 0)->nComp();
+    const int ncomp_multimodefab = fields.get(FieldType::Efield_aux, Direction::r, 0)->nComp();
     const int ncomp = ncomp_multimodefab;
     // Clear any pre-existing vector to release stored data
     // This ensures that when domain is load-balanced, the functors point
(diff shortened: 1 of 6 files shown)


Source/ablastr/fields/MultiFabRegister.H
```diff
@@ -58,21 +58,116 @@ namespace ablastr::fields
...
+    inline constexpr Direction Direction::y = Direction(1);
+#endif
+
+#if defined(WARPX_DIM_1D_Z)
+    inline constexpr Direction Direction::z = Direction(0);
+#elif defined(WARPX_DIM_RZ) || defined(WARPX_DIM_XZ)
+    inline constexpr Direction Direction::z = Direction(1);
+#elif defined(WARPX_DIM_3D)
+    inline constexpr Direction Direction::z = Direction(2);
+#endif
```
ax3l: We need to make sure the values make sense for blocks like these (e.g., in RZ):
https://github.com/BLAST-WarpX/warpx/blob/25.09/Source/Diagnostics/BTDiagnostics.cpp#L583-L605

ax3l: Alternatively, if we want to store reconstructured x,y,z fields in the MF register, e.g., during RZ simulations, we could also plainly define
```
x=0
y=1
z=2
r=3
theta=4
phi=5
```
use those ints only internally...

Not entirely sure what we do with `m_all_dirs` (used in `..._alldirs` getters) for RZ/RCYLINDER/RSPHERE sims right now, too.
https://github.com/BLAST-WarpX/warpx/blob/25.09/Source/ablastr/fields/MultiFabRegister.H#L825-L826

ax3l:
@EZoni interesting observation: the 1D test fails to build but the Azure Pipelines `Build` step shows as green.
image


Let us check why. It is good to error out immediately so one does not see a confusing follow-up error on compile issues.

Source/Diagnostics/BTDiagnostics.cpp
```diff
@@ -582,23 +582,23 @@ BTDiagnostics::InitializeFieldFunctors (int lev)
         m_cell_center_functors.at(lev).size());
     for (int comp=0; comp<m_cell_center_functors_at_lev_size; comp++){
         if        ( m_cellcenter_varnames[comp] == "Ex" ){
-            m_cell_center_functors[lev][comp] = std::make_unique<CellCenterFunctor>(fields.get(FieldType::Efield_aux, Direction{0}, lev), lev, m_crse_ratio);
+            m_cell_center_functors[lev][comp] = std::make_unique<CellCenterFunctor>(fields.get(FieldType::Efield_aux, Direction::x, lev), lev, m_crse_ratio);
```
ax3l: Looks like the RCylinder code (accidentially?) compiles/uses this (very Cartesian-looking?) code section:
```
FAILED: [code=1] CMakeFiles/lib_rcylinder.dir/Source/Diagnostics/BTDiagnostics.cpp.o 
/usr/bin/ccache /usr/bin/c++ -DABLASTR_USE_FFT -DAMREX_SPACEDIM=1 -DAMReX_FFTW_OMP=1 -DWARPX_DIM_RCYLINDER -DWARPX_QED -DWARPX_STATIC_DEFINE -DWARPX_USE_FFT -DWARPX_USE_OPENPMD -DWarpX_FFTW_OMP=1 -Dlib_rsphere_EXPORTS -I/home/vsts/work/1/s/Source -I/home/vsts/work/1/s/build/Source -I/home/vsts/work/1/s/build/_deps/fetchedopenpmd-src/include -I/home/vsts/work/1/s/build/_deps/fetchedopenpmd-build/include -I/home/vsts/work/1/s/build/_deps/fetchedpicsar-src/multi_physics/QED/include -I/home/vsts/work/1/s/build/_deps/fetchedamrex-src/Tools/C_scripts -isystem /home/vsts/work/1/s/build/_deps/fetchedamrex-src/Src/Base -isystem /home/vsts/work/1/s/build/_deps/fetchedamrex-src/Src/Base/Parser -isystem /home/vsts/work/1/s/build/_deps/fetchedamrex-src/Src/Boundary -isystem /home/vsts/work/1/s/build/_deps/fetchedamrex-src/Src/AmrCore -isystem /home/vsts/work/1/s/build/_deps/fetchedamrex-src/Src/LinearSolvers -isystem /home/vsts/work/1/s/build/_deps/fetchedamrex-src/Src/LinearSolvers/MLMG -isystem /home/vsts/work/1/s/build/_deps/fetchedamrex-src/Src/Particle -isystem /home/vsts/work/1/s/build/_deps/fetchedamrex-src/Src/FFT -isystem /home/vsts/work/1/s/build/_deps/fetchedamrex-build -isystem /usr/lib/x86_64-linux-gnu/openmpi/include -isystem /usr/lib/x86_64-linux-gnu/openmpi/include/openmpi -g1 -Wall -Wextra -Wpedantic -Wshadow -Woverloaded-virtual -Wunreachable-code -O3 -DNDEBUG -std=c++17 -fPIC -fopenmp -finline-limit=43210 -MD -MT CMakeFiles/lib_rcylinder.dir/Source/Diagnostics/BTDiagnostics.cpp.o -MF CMakeFiles/lib_rcylinder.dir/Source/Diagnostics/BTDiagnostics.cpp.o.d -o CMakeFiles/lib_rcylinder.dir/Source/Diagnostics/BTDiagnostics.cpp.o -c /home/vsts/work/1/s/Source/Diagnostics/BTDiagnostics.cpp
/home/vsts/work/1/s/Source/Diagnostics/BTDiagnostics.cpp: In member function ‘virtual void BTDiagnostics::InitializeFieldFunctors(int)’:
/home/vsts/work/1/s/Source/Diagnostics/BTDiagnostics.cpp:585:130: error: ‘x’ is not a member of ‘ablastr::fields::Direction’
  585 |             m_cell_center_functors[lev][comp] = std::make_unique<CellCenterFunctor>(fields.get(FieldType::Efield_aux, Direction::x, lev), lev, m_crse_ratio);
      |                                                                                                                                  ^
```
will simply keep the general write style as before and revert this.

@dpgrote is this intentional or maybe lacking an `#ifdef`?

dpgrote (approved):
Good idea!

Source/ablastr/fields/MultiFabRegister.cpp
```diff
@@ -628,15 +628,11 @@ namespace ablastr::fields
...
-        //       in the future we might want to break this to "[dir=0/1/2]".
-        //       This will be a breaking change for (Python) users that rely on that string.
-        constexpr int x_in_ascii = 120;
-        std::string const component_name{char(x_in_ascii + dir.dir)};
+        // Add the suffix for the direction [dir=x] or [dir=y] or [dir=z] or [dir=r]
+        std::string const component_name = dir;
         return mf_name(
             name
-            .append("[")
+            .append("[dir=")
```
ax3l, line 635: @roelof-groenewald @RemiLehe @dpgrote just highlighting that this is a breaking change if someone uses the "raw old strings" still in Python instead of the APIs we defined a year ago. I do not think someone does and our wrappers are mostly used anyway, which do the right thing...

RemiLehe: @n01r I think that you may have some scripts that use this. Just flagging the breaking change here.

dpgrote: Should the use of the internal names be deprecated?

ax3l: Yes, ideally. This is mostly to make sure that `list()` looks understandable.

Do we use internal names still somewhere?

RemiLehe (approved):

Source/ablastr/fields/MultiFabRegister.cpp
dpgrote: It's not used anywhere within WarpX, but the Python interface is still there
```
        .def("multifab",
             [](WarpX & wx, std::string internal_name) {
```
There is a note that this is deprecated, but this should just be removed.

ax3l: Good point. Let me remove this in a follow-up PR, since this one is all green.

ax3l: Posted in #6180

PR #6165, opened 2025-09-12
Fix ablastr::utils::write_used_inputs_file

WeiqunZhang:
The intended behavior is to write only used inputs, but previously it also wrote unused inputs. This PR fixes it by using a new amrex function, amrex::ParmParse::prettyPrintUsedInputs.


EZoni:
@JustinRayAngus 

I requested your review here because I think you raised the issue in the first place, if I'm not mistaken. Just in case you want to have a look and/or test locally if this works for your case prior to merging the PR.

EZoni (approved):

JustinRayAngus (approved):
Great. Thanks.

PR #6166, opened 2025-09-12
[WIP] Adding support to collect memory usage from boxes to load balancing

jessdagostini:
This PR aims to add support for collecting memory usage of each box (considering `MultiFabs` and `Particles`) in the simulation. The goal is to use this information for the load balancing algorithms.

### DIFF
--- a/Source/ablastr/fields/MultiFabRegister.cpp
+++ b/Source/ablastr/fields/MultiFabRegister.cpp
@@ -366,6 +366,9 @@ namespace ablastr::fields
     {
         for (auto & element : m_mf_register ) {
             const MultiFabOwner & mf_owner = element.second;
+            if (mf_owner.is_alias()) {
+                continue;
+            }
             amrex::LayoutData<std::size_t> tmpmem(ba, dm);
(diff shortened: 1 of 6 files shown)


Source/ablastr/fields/MultiFabRegister.cpp
```diff
@@ -357,6 +357,23 @@ namespace ablastr::fields
...
 
+    void
+    MultiFabRegister::CapacityOfFabs (
+        amrex::LayoutData<std::size_t>& mem,
+        amrex::BoxArray const & ba,
+        amrex::DistributionMapping const & dm
+    ) const
+    {
+        for (auto & element : m_mf_register ) {
+            const MultiFabOwner & mf_owner = element.second;
```
ax3l, line 368: You could check if `mf_owner.is_alias()` is true. In case it is an alias, skip the field in this calculation to avoid double counting.

jessdagostini: Fixed on a11263fcc82dcbce48901de9d28c47d42844b6a0!

PR #6168, opened 2025-09-12
move detailed print for implicit solvers to base class.

JustinRayAngus:
This PR moves the print routine for the implicit solver options to the base implicit solver class. This reduces code duplication and adds print messages for some parameters that were previously missing for some of the derived classes.

PR #6169, opened 2025-09-12
add option to modify particle update on initial Newton step

JustinRayAngus:
The PR adds an option that can enhance the efficiency of the nonlinear solver in the implicit methods. At each iteration In the standard implementation of the PS-JNFK method, each particle is advanced iteratively using a Picard method until the position and velocity are self-consistent. The number of Picard iterations required to converge can be large on the first Newton iteration where the system is typically far from the converged solution. This PR adds an option to only do a single Picard iteration for each particle on the first Newton step. This has been found to improve the efficiency of the implicit time step by lowering the number of total Picard iterations (https://doi.org/10.1016/j.jcp.2023.112383).

Tasks:
- [x] merge PR #6168 
- [x] update CI benchmarks for implicit tests affected by this PR


Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -226,6 +226,17 @@ namespace {
 
         }
 
+        // The Charge-conserving deposits assume the change in position is consistent with
```
dpgrote: Instead of doing this separate push here, could the `else if` be added to the convergence check above
```
if (step_norm < particle_tolerance) {
    convergence = true;
    break;
} else if (iter+1 == max_iterations) {
    break;
}
```
The code would then need to set `options.max_particle_iterations = 2` instead of `1`.

Docs/source/usage/parameters.rst
```diff
@@ -152,6 +152,11 @@ Overall simulation parameters
           - ``implicit_evolve.particle_suborbits`` (`bool`, default: false)
           - ``implicit_evolve.print_unconverged_particle_details`` (`bool`, default: false)
 
+        - ``implicit_evolve.modify_initial_newton_step`` (`bool`, default: false).
```
dpgrote: Could a more specific name be used? Maybe "skip_initial_newton_step_iteration"?

JustinRayAngus: I agree the name is not great. 

Maybe "skip_picard_initial_newton_step"?

JustinRayAngus: ChatGPT thinks that is too long and suggests using `skip_picard_init`, which is still explicit on what's happening while being more concise. I think maybe a good name is `skip_particle_picard_init`, still kind of long, but more explicit.

Source/Particles/Pusher/ImplicitPushPX.cpp
JustinRayAngus: Another similar option would be to put the following lines after setPosition() in the for loop:
```
setPosition(ip, xp, yp, zp);
if (iter == max_iterations) {
    break;
}
```
and change the for loop to use `<=`:
`for (int iter=0; iter <= max_iterations; iter++)`.

However, I'm not sure either of these options offer any real advantage. I think the way it is now makes it more explicit as to why it is being done. Or, at least its placement outside the loop motivates a detailed comment about why it is needed.

dpgrote: My main concern is that this is duplicated code. Also, I would find it more clear what is happening seeing that break after the position advance, that the position advance always needs to happen so that the positions are consistent with the velocities.

I think this could be made more clear if the loop was changed to a `while (true)` loop so that the only exit points are right after the position advance.

JustinRayAngus: Sorry Dave, but I disagree about the duplicate code issue. It's just a few lines of code that explicitly call other functions, and the lines are wrapped inside an if (!convergence) condition, which (at least to me) makes it clear when and why these lines are needed.

It seems that we just have different style preferences here. I prefer the conditional check on convergence post for loop, which is called only once per particle, over an additional conditional check inside the for loop that is called once per iteration.

JustinRayAngus: If we really wanted to make things super optimized, we should probably switch the order the momentum and position push. As is, each particle does a position push at the end of the Picard iteration, and then during the next call to this function (on the next Newton step) it re-does the same position push unnecessarily at the beginning of the Picard iterations. The initial position push is there because experience tells me that works best on the first Newton step. So, if we switch the order of pushX and pushP, we would still want an extra call to pushX before the Picard iterations for newton iter = 0:
```
if (newton_iter==0) {
    UpdatePositionImplicit();
    setPosition();
}
```

If we decide to do this, it should be a separate PR, but I don't think it's needed because the position push is not a wall time bottleneck.

dpgrote: The check if iter < max_iterations is already being done by the for loop statement every iteration, at the start of the loop. With my suggestion using a while loop, the check would still be done once per iteration but at what seems to be the more appropriate place in the loop. But do what you think is appropriate.

JustinRayAngus: @dpgrote I reformatted the Picard loop to end with the position push.

I should comment that the new format is in the classical form for a Picard fixed-point iteration method for dxp:
-start with initial guess for dxp.
-do iterative loop to update dxp.

dpgrote (approved):

PR #6170, opened 2025-09-13
Implicit gamma inverse function

JustinRayAngus:
This PR creates a function for computing gamma inverse in the implicit schemes, reducing code duplication.


ax3l (approved):
Legit, thanks!

PR #6174, opened 2025-09-15
New SubcyclingHalf enum for subcycling algorithm

EZoni:
Implement @JustinRayAngus's suggestion from https://github.com/BLAST-WarpX/warpx/pull/5955#discussion_r2341648574:

> I looked a bit more carefully at the use of `DtType` in the development branch. It seems to me that it is really only needed because of the subcycling done in `OneStep_sub1` for `Evolve` on the fine level. Perhaps the simplest fix is to replace the `DtType` flag with something like `SubCycleStage` flag. The options could be `None` (default), `FirstHalf`, `SecondHalf`. This would at least be more transparent and not conflicting with your new usage of the `DtType` flag, which seems like a more appropriate use of `DtType` in my opinion.

The `DtType` enum is removed in this PR. It will be reintroduced in #5955, with the usage described in the comment above.

### Main changes

- Replaced `DtType` enum with `SubcyclingHalf` enum (default: `None`).
- Removed `SubcyclingHalf` argument from `EvolveG` (unused).
- Replaced `SubcyclingHalf` argument in `EvolveF` with an integer to select the component of rho.
- Renamed `rhocomp` as `rho_comp` everywhere (to make grepping and searching more effective).


Source/Evolve/WarpXEvolve.cpp
EZoni, line 1: Note that in this file there are instances where the enum is used in `OneStep_nosub`, so _without_ subcycling.

It is used to split the fields update (only B, and F, G) in two halves with the FDTD solver - can't remember why, at this point.

So maybe `SubcylingStage` isn't the best name either? Not sure. Still better than `DtType`, in my opinion.

JustinRayAngus: EvolveG(): The flag is never used.
EvolveB(): The flag is only used in WarpXFieldBoundaries.cpp for the SilverMueller BC.
EvolveF(): The flag is only used to set an index (rhocomp) to 0 or 1.

Seems like it might be easy to remove the need for this flag in EvolveG() and EvolveF().

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -505,12 +504,12 @@ WarpX::OneStep_nosub (Real cur_time)
             }
         }
     } else {
-        EvolveF(0.5_rt * dt[0], DtType::FirstHalf);
-        EvolveG(0.5_rt * dt[0], DtType::FirstHalf);
+        EvolveF(0.5_rt * dt[0], /*rho_comp=*/0);
```
EZoni, line 507: Personally I like the C++ syntax
```cpp
my_function(/*m=*/1, /*n=*/2);
```
because it mimics the Python keyword syntax
```py
my_function(m=1, n=2)
```

If you have objections I can use a temporary variable to store the integer instead.

I am generally quite opposed to simply writing
```cpp
my_function(1, 2);
```
in C++ because it's less self-explanatory (even though one can hover over the function's name and see the interface with an IDE). Especially for integer arguments.

Source/Evolve/WarpXEvolve.cpp
EZoni: @JustinRayAngus 

Thanks, I pushed some updates to improve things following your suggestions. Let me know what you think!

Source/Evolve/WarpXEvolve.cpp
JustinRayAngus: I like the C++ syntax.

Also, using rho_comp as the variable makes much more sense than DtType here.

JustinRayAngus:
This looks good, but I wonder if a more precise name could be used. `SubcyclingStage` to me makes me think of an arbitrary number of stages. What about `TwoCycleStage`?

EZoni:
We can do `TwoCycleStage`. I thought of `SubcyclingHalf` (or `SubcylceHalf`) as well, what do you think?

Personally I like the idea of keeping the root "subcycl-" in the variable name because it links immediately to the input parameter `warpx.do_subcycling`.

JustinRayAngus (approved):

PR #6179, opened 2025-09-16
`MANIFEST.in`: `dependencies.json`

ax3l:
Needed to build (CMake) Python source packages.


EZoni (approved):
Thanks! I didn't know that this file exists :sweat_smile:

PR #6180, opened 2025-09-16
Python: Remove Deprecated `multifab` Overload

ax3l:
This particular overload is deprecated for a while. Please use instead:
* `multifab('prefix', level=...)` for scalar fields
* `multifab('prefix', dir=..., level=...)` for vector field components


dpgrote (approved):
Thanks!

roelof-groenewald (approved):

PR #6184, opened 2025-09-18
Implementing sparse matrix representation of Jacobian and using PETSc preconditioners

debog:
Implemented a `MatrixPC` preconditioner that constructs the sparse matrix representing the Jacobian for the implicit/semi-implicit time integrators. Currently, this PC provides no native matrix solvers; it relies on PETSc's preconditioners and matrix solvers. In the PETSc interfaces for the JFNK algorithm, added the option of using matrix-based preconditioners.

In this PR, the matrix comprises the Maxwell's equation terms (discretized curl-curl) and the diagonal element of the mass matrix (for the current term).

Other changes:
+ Implemented a wrapper for PETSc's SNES module to solve nonlinear systems; it is an alternative to the native Newton solver.
+ Streamlined the PETSc interface by creating a `PETScSolver_impl` base class that contains everything common to the linear and nonlinear solvers. `KSP_impl` and `SNES_impl` now inherit from it.

Minor changes:
+ Moved PETSc initialization and finalizing calls to `main.cpp` to allow for the use of command line inputs.
+ Simplified initializing the nonlinear solver based on user input by removing `nlsolver_type_str` and directly using `AMREX_ENUM` types.

CI test will be added in a later PR.


Source/NonlinearSolvers/WarpX_PETSc.cpp
```diff
@@ -78,6 +92,62 @@ void copyVec( Vec& a_pvec, const VecType& a_wvec)
...
+
+    SNES_impl *context = (SNES_impl*) ctxt;
+    KSP lin_solver;
+    SNESGetKSP(a_solver, &lin_solver);
+    PC pc;
+    KSPGetPC(lin_solver, &pc);
+    PCType pctype;
+    PCGetType(pc, &pctype);
+
+    if (strcmp(pctype,PCNONE) && strcmp(pctype,PCSHELL)) {
```
JustinRayAngus, line 134: Can this ever be true?

debog: In C `string.h`, `strcmp(str1,str2)` returns `0` if `str1 == str2` and `1` if not! So this `if` condition reads "if `pctype` is neither `PCNONE` nor `PCSHELL`".

I can replace this with the following C++ equivalent:
```
if ((std::string(pctype) != std::string(PCNONE)) && (std::string(pctype) != std::string(PCSHELL))
```
if you think that is preferable.

JustinRayAngus: ?`strcmp(str1,str2)` returns `0` if `str1 == str2`? That's confusing to me. I would think that it would return `true` or `1` if `str1 == str2`, which is the same as if you used `str1 == str2`. 

Maybe use something like `strcmp(str1,str2)==0`? I won't be picky here. Do what you think reads the best.

debog: Yeah it's pretty counter-intuitive but intuitive once you know why (https://www.geeksforgeeks.org/c/strcmp-in-c/).

Source/NonlinearSolvers/MatrixPC.H
```diff
@@ -0,0 +1,583 @@
...
+    m_bcoefs = m_ops->GetMassMatricesCoeff();
+    if (m_bcoefs == nullptr) {
+        if (m_pc_mass_matrix_width > 0) {
+            std::stringstream warning_message;
+            warning_message << "pc_mass_matrix_width set as nonzero but unable to get mass matrices from operator.\n";
+            warning_message << "setting m_pc_mass_matrix_width = 0.\n";
+            ablastr::warn_manager::WMRecordWarning("MatrixPC", warning_message.str());
+        }
+        m_pc_mass_matrix_width = 0;
+    }
```
JustinRayAngus, line 290: Perhaps I don't fully understand this check, but it seems like this is catching a potential setup bug to ensure that `m_bcoefs` is set when using mass matrix terms in the PC, but even with `m_pc_mass_matrix_width = 0`, there would be an issue if `m_bcoefs == nullptr`, right?

Source/NonlinearSolvers/MatrixPC.H
```diff
@@ -0,0 +1,583 @@
...
+}
+
+template <class T, class Ops>
+void MatrixPC<T,Ops>::computePCMatrixNNZ()
+{
+    if (m_pc_diag_only) {
+        m_pc_mat_nnz = 1;
+    } else {
+#if defined(WARPX_DIM_1D_Z)
```
JustinRayAngus: While WarpX is not currently setup to use the mass matrix for non-planar geometries, I don't think we need to limit this part of the code to planar geometries.

`#if defined(WARPX_DIM_1D_Z) || defined(WARPX_DIM_RCYLINDER) || defined(WARPX_DIM_RSPHERE)`

Source/NonlinearSolvers/MatrixPC.H
```diff
@@ -0,0 +1,583 @@
...
+    pp.query("verbose", m_verbose);
+    pp.query("pc_diagonal_only", m_pc_diag_only);
+    pp.query("pc_mass_matrix_width", m_pc_mass_matrix_width);
+    if(m_pc_mass_matrix_width>0) {
+       pp.query("pc_mass_matrix_include_ij", m_pc_mass_matrix_include_ij);
+    }
+}
+
+template <class T, class Ops>
+void MatrixPC<T,Ops>::computePCMatrixNNZ()
```
JustinRayAngus: To improve readability, could we call this function `computePCMatrixNumNonZeroBands()` ?

Source/NonlinearSolvers/MatrixPC.H
```diff
@@ -0,0 +1,583 @@
...
+{
+    if (m_pc_diag_only) {
+        m_pc_mat_nnz = 1;
+    } else {
+#if defined(WARPX_DIM_1D_Z)
+        static int nbands_J_mass_matrix = 1 + 2*3*(m_pc_mass_matrix_width+1);
```
JustinRayAngus: I believe this is supposed to be the maximum number of non-zero elements per row. If that is the case, then I get the following for rows corresponding to the in-plane E (Ez for WarpX in 1D):
```
nbands_Jz_mm = 1 + 2*m_pc_mass_matrix_width (for zz)
                 + 2*m_pc_mass_matrix_width (for zx)
                 + 2*m_pc_mass_matrix_width (for zy)
             = 1 + 2*3*m_pc_mass_matrix_width
```
```
nbands_Jy_mm = 1 + 2*m_pc_mass_matrix_width (for yy)
               1 + 2*m_pc_mass_matrix_width (for yx)
                 + 2*m_pc_mass_matrix_width (for yz)
             = 2 + 2*3*m_pc_mass_matrix_width
```
`nbands_Jx_mm = nbands_Jy_mm`

  So, if this is supposed to be the maximum value, I get 
`static int nbands_J_mass_matrix = 2 + 2*3*(m_pc_mass_matrix_width)`

Source/NonlinearSolvers/MatrixPC.H
JustinRayAngus: Can you also add a comment here explaining what is being done.

Source/NonlinearSolvers/MatrixPC.H
JustinRayAngus: For 2D I get the max is set by the Jy term, which lives at cell nodes, to be:
```
nbands_Jy_mm = (1 + 2*m_pc_mass_matrix_width)^2 (for yy)
             + (1 + 2*m_pc_mass_matrix_width)*(2*m_pc_mass_matrix_width) (for yz)
             + (2*m_pc_mass_matrix_width)*(1 + 2*m_pc_mass_matrix_width) (for yx)
             = (1 + 2*m_pc_mass_matrix_width)*(1 + 6*m_pc_mass_matrix_width)
```

It may not matter in terms of functionality since the value used now is larger than this, but I think it helps improve readability.

JustinRayAngus: The expressions worked out above are for a Yee grid. It may be of interest to use the mass matrix on a colocated grid in the future, which gives an upper bound to the number of non-zero terms for a given width. 

Maybe it's best to remove the geometry dependence here and replace using something like the following:

```
static int nbands_J_mass_matrix =
    3 * ((AMREX_SPACEDIM == 1) ? (1 + 2*m_pc_mass_matrix_width) :
         (AMREX_SPACEDIM == 2) ? (1 + 2*m_pc_mass_matrix_width) * (1 + 2*m_pc_mass_matrix_width) :
         (1 + 2*m_pc_mass_matrix_width) * (1 + 2*m_pc_mass_matrix_width) * (1 + 2*m_pc_mass_matrix_width));
```

and add a comment about this being an upper bound for a given width if the variables are colocated on the grid.

Source/NonlinearSolvers/WarpX_PETSc.H
```diff
@@ -122,22 +154,106 @@ class KSP_impl
...
+        //! absolute tolerance
+        amrex::Real m_atol = 0.;
+        //! relative tolerance
+        amrex::Real m_rtol = 1e-6;
+        //! step tolerance
+        amrex::Real m_stol = 1e-16;
+        //! maximum iterations
+        int m_maxits= 100;
+        //! verbosity
+        bool m_verbose = true;
```
JustinRayAngus, line 288: m_verbose can be removed. Use the base class m_verbose, which is mutable.

Source/NonlinearSolvers/WarpX_PETSc.cpp
```diff
@@ -271,6 +478,283 @@ void KSP_impl::setVerbose(int a_v)
     }
 }
 
+SNES_impl::SNES_impl(const VecType& a_vec, TIType* a_op)
+{
+    BL_PROFILE("SNES_impl::SNES_impl()");
+    amrex::Print() << "SNES_impl: Initialized PETSc's SNES solver.\n";
+
+    const amrex::ParmParse pp_newton("newton");
+    pp_newton.query("verbose",             m_verbose);
```
JustinRayAngus, line 582: replace m_verbose with this->m_verbose, same as in Newton class.

Source/NonlinearSolvers/WarpX_PETSc.cpp
```diff
@@ -271,6 +478,283 @@ void KSP_impl::setVerbose(int a_v)
...
+    delete this->m_A;
+    delete this->m_P;
+    delete this->m_x;
+    delete this->m_b;
+    delete this->m_linop;
+}
+
+void SNES_impl::printParams () const
+{
+    amrex::Print()     << "SNES_impl verbose:             " << (m_verbose?"true":"false") << "\n";
```
JustinRayAngus, line 711: use `this->m_verbose`

Source/NonlinearSolvers/WarpX_PETSc.cpp
```diff
@@ -271,6 +478,283 @@ void KSP_impl::setVerbose(int a_v)
...
+    SNESConvergedReason reason;
+    SNESGetConvergedReason( m_snes->obj, &reason );
+    m_status = (int) reason;
+    SNESGetFunctionNorm(m_snes->obj, &m_norm);
+
+    const char* conv_reason;
+    SNESGetConvergedReasonString(m_snes->obj, &conv_reason);
+    amrex::Print() << "Newton (PETSc SNES): exited due to \""
+                   << conv_reason << "\" "
+                   << "(abs. norm=" << m_norm << ").\n";
```
JustinRayAngus: add space between `norm` and `=`

Source/NonlinearSolvers/WarpX_PETSc.cpp
```diff
@@ -271,6 +478,283 @@ void KSP_impl::setVerbose(int a_v)
...
+    SNESGetLinearSolveIterations(m_snes->obj, &m_niters_l);
+
+    SNESConvergedReason reason;
+    SNESGetConvergedReason( m_snes->obj, &reason );
+    m_status = (int) reason;
+    SNESGetFunctionNorm(m_snes->obj, &m_norm);
+
+    const char* conv_reason;
+    SNESGetConvergedReasonString(m_snes->obj, &conv_reason);
+    amrex::Print() << "Newton (PETSc SNES): exited due to \""
```
JustinRayAngus: put print statement inside `if (this->m_verbose) {}`

Source/NonlinearSolvers/WarpX_PETSc.cpp
```diff
@@ -271,6 +478,283 @@ void KSP_impl::setVerbose(int a_v)
...
+#endif
+        MatSetOption(this->m_P->obj, MAT_NEW_NONZERO_LOCATION_ERR, PETSC_FALSE);
+        MatSetOption(this->m_P->obj, MAT_NEW_NONZERO_ALLOCATION_ERR, PETSC_FALSE);
+        MatSetUp(this->m_P->obj);
+        SNESSetJacobian(m_snes->obj, this->m_A->obj, this->m_P->obj, JacobianFunction, this);
+    }
+
+    setTolerances(m_rtol, m_atol, m_maxits, m_rtol_l, m_atol_l, m_maxits_l);
+    setMaxIters(m_maxits, m_maxits_l);
+    if (m_verbose) {
```
JustinRayAngus, line 689: use `this->m_verbose`

Source/NonlinearSolvers/WarpX_PETSc.cpp
```diff
@@ -271,6 +478,283 @@ void KSP_impl::setVerbose(int a_v)
...
+}
+
+void SNES_impl::setVerbose(bool a_v)
+{
+    BL_PROFILE("SNES_impl::setVerbose()");
+    m_verbose = a_v;
+    if (a_v && isDefined()) {
+        SNESMonitorSet(m_snes->obj, printSNESResidual, NULL,NULL );
+    }
+}
```
JustinRayAngus, line 849: Is this function used?

Source/NonlinearSolvers/WarpX_PETSc.cpp
```diff
@@ -166,50 +317,94 @@ void KSP_impl::createObjects(const VecType& a_vec)
...
+#endif
+        MatSetOption(this->m_P->obj, MAT_NEW_NONZERO_LOCATION_ERR, PETSC_FALSE);
+        MatSetOption(this->m_P->obj, MAT_NEW_NONZERO_ALLOCATION_ERR, PETSC_FALSE);
+        MatSetUp(this->m_P->obj);
+        MatSetFromOptions(this->m_P->obj);
+        KSPSetOperators( m_ksp->obj, this->m_A->obj, this->m_P->obj );
     }
+    KSPSetTolerances( m_ksp->obj, m_rtol, m_atol, PETSC_CURRENT, m_maxits );
+    KSPSetNormType( m_ksp->obj, KSP_NORM_UNPRECONDITIONED );
+    if (m_verbose > 0) {
```
JustinRayAngus: use `this->m_verbose`?

Source/NonlinearSolvers/WarpX_PETSc.cpp
```diff
@@ -249,20 +444,32 @@ void KSP_impl::solve(VecType& a_Y, const VecType& a_R)
...
     KSPConvergedReason reason;
     KSPGetConvergedReason( m_ksp->obj, &reason );
     m_status = (int)reason;
     KSPGetResidualNorm( m_ksp->obj, &m_norm );
+
+    const char* conv_reason;
+    KSPGetConvergedReasonString(m_ksp->obj, &conv_reason);
+    amrex::Print() << "GMRES (PETSc KSP): exited due to \""
+                   << conv_reason << "\" "
+                   << "(abs. norm=" << m_norm << ").\n";
```
JustinRayAngus: put Print() statements inside `if (this->m_verbose) {}`

JustinRayAngus:
Here is some basics on symmetry(PMC) and conductor (PEC) boundaries, which may be a lot of work and we should probably save implementing this for a follow-on PR.

image

JustinRayAngus:
I'm looking at the PC_mask_E in PICNIC. Why is the number of components `2 + 2*SpaceDim`?
```
     m_PC_mask_E.define(grids,2+2*SpaceDim,ghostVect);
     for (int d=0; d < m_PC_mask_E.nComp(); d++) {
       SpaceUtils::setVal(m_PC_mask_E, d, 1.0);
     }
```

Source/NonlinearSolvers/WarpX_PETSc.H
debog: This is `SNES_impl`, not `PETScSNES` (`Source/NonlinearSolvers/PETScSNES_Wrapper.H`); it doesn't inherit from the base class `NonlinearSolver` (`PETScSNES` does).

Source/NonlinearSolvers/WarpX_PETSc.cpp
debog: See [this response](https://github.com/BLAST-WarpX/warpx/pull/6184#discussion_r2393046222).

Source/NonlinearSolvers/WarpX_PETSc.cpp
debog: This [this response](https://github.com/BLAST-WarpX/warpx/pull/6184#discussion_r2393046222).

Source/NonlinearSolvers/WarpX_PETSc.cpp
debog: This [this response](https://github.com/BLAST-WarpX/warpx/pull/6184#discussion_r2393046222).

Source/NonlinearSolvers/WarpX_PETSc.cpp
debog: It's called from `PETScSNES::Verbose()` (overriding `NonlinearSolver::Verbose()`).

Source/NonlinearSolvers/WarpX_PETSc.cpp
debog: This is `KSP_impl`, not `PETScKSP` (`Source/NonlinearSolvers/PETScKSP_Wrapper.H`); it doesn't inherit from the base class `LinearSolver` (`PETScKSP` does). At any rate, `LinearSolver` doesn't have `m_verbose`.

Source/NonlinearSolvers/WarpX_PETSc.H
JustinRayAngus: I see. So this m_verbose can be set from PETScSNES by calling Verbose().

I think it might be a good idea to add some description at the top of the WarpX_PETSc.H file to explain how it works with and differs from PETScSNES class.

Source/NonlinearSolvers/MatrixPC.H
```diff
@@ -0,0 +1,623 @@
...
+                        {
+                            auto cidx_g_rhs = (int) (dir == 0 ? dof_arr(i,j+1,k,1) : dof_arr(i+1,j,k,1));
+                            amrex::Real val = -alpha * dxi[dir==0?1:0] * dxi[dir==0?1:0];
+                            auto flag = MatrixPCUtils::insertOrAdd( cidx_g_rhs, val,
+                                                                    &c_indices_g_ptr[ridx_l*nnz_max],
+                                                                    &a_ij_ptr[ridx_l*nnz_max],
+                                                                    nnz_max, icol );
+                            if (!flag) { Gpu::Atomic::Max(nnz_actual_ptr, icol); }
+                        }
+                        {
```
JustinRayAngus: I verified the following four blocks. Would be useful to add the following comments here:                        
// The following four blocks are for
 // dir = 0: d/dx(dEz/dz) at Ex(i,j) with Ex centered in x and nodal in z
 // dir = 2: d/dz(dEx/dx) at Ez(i,j) with Ez centered in z and nodal in x

Source/NonlinearSolvers/MatrixPC.H
```diff
@@ -0,0 +1,623 @@
...
+                            auto cidx_g_rhs = (int) (dir == 0 ? dof_tdir_arr(i,j-1,k,1) : dof_tdir_arr(i-1,j,k,1));
+                            amrex::Real val = alpha * dxi[dir==0?0:1] * dxi[dir==0?1:0];
```
JustinRayAngus: Not necessary, but this line could be simplified to
`amrex::Real val = alpha * dxi[0] * dxi[1];`

and same for the next three blocks.

Source/NonlinearSolvers/WarpX_PETSc.H
debog: I added some comments in `WarpX_PETSc.H` as well as the wrapper files.

Source/NonlinearSolvers/MatrixPC.H
```diff
@@ -0,0 +1,649 @@
...
+    //                  (row-major format)
+    // where n is the local number of rows, and ncmax is the maximum number of non-zero
+    // elements per row.
+
+    BL_PROFILE("MatrixPC::Assemble()");
+    using namespace amrex;
+
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+        IsDefined(),
+        "MatrixPC::Update() called on undefined object" );
```
JustinRayAngus: MatrixPC::Assemble()

JustinRayAngus (commented):
I think it's time to remove the WIP

Source/NonlinearSolvers/MatrixPC.H
```diff
@@ -0,0 +1,649 @@
...
+                int tdir = -1;
+                if  (dir == 0) { tdir = 2; }
+                else if (dir == 2) { tdir = 0; }
+                else { tdir = 1; }
+                auto dof_tdir_arr = dofs_mfarrvec[lev][tdir]->const_array(mfi);
+#elif defined(WARPX_DIM_3D)
+                amrex::ignore_unused(dof_arr);
+                amrex::ignore_unused(dxi);
+                WARPX_ABORT_WITH_MESSAGE("MatrixPC<T,Ops>::Update() not yet implemented for WARPX_DIM_3D");
+#endif
```
JustinRayAngus, line 437: I think we want to add 3D for this PR.

Source/NonlinearSolvers/MatrixPC.H
```diff
@@ -0,0 +1,701 @@
...
+                                                                &a_ij_ptr[ridx_l*nnz_max],
+                                                                nnz_max, icol );
+                        if (!flag) { Gpu::Atomic::Max(nnz_actual_ptr, icol); }
+                    }
+
+#if defined(WARPX_DIM_1D_Z)
+                    if (dir != 2) {
+                        {
+                            auto cidx_g_rhs = (int) dof_arr(i,j,k,1);
+                            amrex::Real val = 2*alpha * dxi[0]*dxi[0];
```
JustinRayAngus: 2._rt 

Same for other places where 2 appears below

Source/main.cpp
```diff
@@ -15,9 +15,18 @@
...
+#endif
+
 int
 main (int argc, char* argv[]) {
     warpx::initialization::initialize_external_libraries(argc, argv);
+#ifdef AMREX_USE_PETSC
+    PETSC_COMM_WORLD = amrex::ParallelContext::CommunicatorSub();
+    PetscInitialize(&argc, &argv, nullptr, "WarpX with PETSc");
+    amrex::Print() << "Initialized PETSc.\n";
+#endif
```
WeiqunZhang: May the initialization and finalization could be moved into `initialize_external_libraries` and `finalize_external_libraries`.

Source/NonlinearSolvers/PETScSNES_Wrapper.H
```diff
@@ -0,0 +1,122 @@
...
+        PETScSNES () = default;
+        ~PETScSNES () override = default;
+
+        // Prohibit Move and Copy operations
+        PETScSNES(const PETScSNES&) = delete;
+        PETScSNES& operator=(const PETScSNES&) = delete;
+        PETScSNES(PETScSNES&&) noexcept = delete;
+        PETScSNES& operator=(PETScSNES&&) noexcept = delete;
+
+        AMREX_FORCE_INLINE
```
WeiqunZhang: There is no need to mark the functions in this class and the ones in PETScKSP_Wrapper.H `AMREX_FORCE_INLINE`. We don't want to force inline a function unless it's used in GPU and CPU kernels.

Source/NonlinearSolvers/PETScKSP_Wrapper.H
```diff
@@ -30,6 +30,10 @@
  *
  *    The Vec class must have basic math operators, such as Copy, +=, -=,
  *    increment(), linComb(), scale(), etc.. See WarpXSolverVec.H for an example.
+ *
+ *    PETScKSP is just a wrapper; the actual implementation is in KSP_impl in
+ *    Warpx_PETSc.H/cpp. This is to avoid including PETSc's header files here
+ *    and consequent conflicts between PETSc and WarpX variables.
```
WeiqunZhang, line 36: The virtual functions in this class should be marked with `override`.

Source/NonlinearSolvers/NonlinearSolverLibrary.H
```diff
@@ -3,15 +3,19 @@
 
 #include "PicardSolver.H"  // IWYU pragma: export
 #include "NewtonSolver.H"  // IWYU pragma: export
+#include "PETScSNES_Wrapper.H"  // IWYU pragma: export
+
+#include <AMReX_Enum.H>
```
WeiqunZhang, line 8: `AMReX_Enum.H` is already included two lines below.

Source/FieldSolver/ImplicitSolvers/WarpXSolverVec.H
```diff
@@ -315,6 +315,9 @@ public:
     // return the number of AMR levels
     inline auto numAMRLevels () const { return m_num_amr_levels; }
 
+    // return DOFs object pointer
+    inline const auto& getDOFsObject () const { return m_dofs; }
```
WeiqunZhang, line 319: This function, `getWarpX` and `numAMRLevels`  should be marked with `[[nodiscard]]` to avoid clang-tidy warning.

Source/FieldSolver/ImplicitSolvers/WarpXSolverDOF.cpp
```diff
@@ -57,9 +59,9 @@ void WarpXSolverDOF::Define ( WarpX* const        a_WarpX,
             for (int n = 0; n < 3; n++) {
                 auto ncomp = this_array[n]->nComp();
                 m_array[lev][n] = new amrex::MultiFab( this_array[n]->boxArray(),
-                                                                this_array[n]->DistributionMap(),
-                                                                2*ncomp, // {local, global} for each comp
-                                                                amrex::IntVect::TheUnitVector() );
+                                                       this_array[n]->DistributionMap(),
+                                                       2*ncomp, // {local, global} for each comp
+                                                       amrex::IntVect::TheUnitVector() );
```
WeiqunZhang: Just a minor comment. It might be clearer if we use `amrex::IntVect(1)` (or just `1`).

Source/FieldSolver/ImplicitSolvers/WarpXSolverDOF.cpp
```diff
@@ -171,5 +173,32 @@ void WarpXSolverDOF::Define ( WarpX* const        a_WarpX,
...
     }
 
+    if (m_array_type != FieldType::None) {
+        for (int lev = 0; lev < a_num_amr_levels; ++lev) {
+            const auto& geom = a_WarpX->Geom(lev);
+            for (int n = 0; n < 3; n++) {
+                m_array_lhs[lev][n] = new amrex::MultiFab( m_array[lev][n]->boxArray(),
+                                                           m_array[lev][n]->DistributionMap(),
+                                                           m_array[lev][n]->nComp(),
+                                                           amrex::IntVect::TheZeroVector() );
```
WeiqunZhang: Similarly, we could use `amrex::IntVect(0)` or just `0`. I mention this because Justin commented in the past that `amrex::IntVect::TheZeroVector` is too long.

Source/NonlinearSolvers/MatrixPC.H
```diff
@@ -0,0 +1,701 @@
...
+#include <AMReX_ParmParse.H>
+#include <AMReX_Array.H>
+#include <AMReX_Vector.H>
+#include <AMReX_MultiFab.H>
+
+namespace MatrixPCUtils
+{
+    template <typename T>
+    AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE
+    T sign(const T a_x) { return (a_x < 0 ? T(-1) : T(1)); }
```
WeiqunZhang: We don't need to have this function. There is a std function called `std::copysign" for this.

Source/NonlinearSolvers/MatrixPC.H
```diff
@@ -0,0 +1,701 @@
...
+                                 amrex::Gpu::DeviceVector<int>& a_num_nz,
+                                 amrex::Gpu::DeviceVector<int>& a_c_indices_g,
+                                 amrex::Gpu::DeviceVector<RT>& a_a_ij,
+                                 int& a_n, int& a_ncols_max ) override
+        {
+            a_n = m_ndofs_l;
+            a_ncols_max = m_pc_mat_nnz;
+
+            a_r_indices_g.resize(m_r_indices_g.size());
+            amrex::Gpu::copy( amrex::Gpu::deviceToDevice,
```
WeiqunZhang: We could use `amrex::Gpu::copyAsync` instead, and add `amrex::Gpu::streamSynchronize()` at the end of this function.

Source/NonlinearSolvers/MatrixPC.H
```diff
@@ -0,0 +1,701 @@
...
+                                                                        &a_ij_ptr[ridx_l*nnz_max],
+                                                                        nnz_max, icol );
+                                if (!flag) { Gpu::Atomic::Max(nnz_actual_ptr, icol); }
+                            }
+                        }
+                    }
+#endif
+                    num_nz_ptr[ridx_l] = icol;
+                });
+                Gpu::synchronize();
```
WeiqunZhang: I don't think we need to call Gpu::synchronize() here and in line 663.

Source/NonlinearSolvers/MatrixPC.H
```diff
@@ -0,0 +1,701 @@
...
+                    int icol = 0;
+                    auto ridx_l = (int) dof_lhs_arr(i,j,k,0);
+                    auto ridx_g = (int) dof_lhs_arr(i,j,k,1);
+                    AMREX_ALWAYS_ASSERT((ridx_l >= 0) && (ridx_l < ndofs_l));
+                    AMREX_ALWAYS_ASSERT((ridx_g >= 0) && (ridx_g < ndofs_g));
+
+                    r_indices_g_ptr[ridx_l] = ridx_g;
+
+                    {
+                        auto cidx_g_lhs = dof_lhs_arr(i,j,k,1);;
```
JustinRayAngus: two ;;

Source/NonlinearSolvers/MatrixPC.H
```diff
@@ -0,0 +1,701 @@
...
+                amrex::ignore_unused(dxi);
+                WARPX_ABORT_WITH_MESSAGE("MatrixPC<T,Ops>::Assemble encountered unknown dimensionality");
+#endif
+
+                // Add the Maxwell's piece
+                ParallelFor(bx, [=] AMREX_GPU_DEVICE (int i, int j, int k)
+                {
+                    int icol = 0;
+                    auto ridx_l = (int) dof_lhs_arr(i,j,k,0);
+                    auto ridx_g = (int) dof_lhs_arr(i,j,k,1);
```
JustinRayAngus: Small style comment. It's possible that auto is used here because clang tidy told you to do so, but if the type is something basic like bool, int, Real, then I prefer to explicitly use the type over auto because it improves readability.

Source/FieldSolver/ImplicitSolvers/WarpXSolverDOF.cpp
JustinRayAngus: It does sound like something I would say, but I really don't care either way.

Source/NonlinearSolvers/MatrixPC.H
debog: I prefer `auto`.

Source/FieldSolver/ImplicitSolvers/WarpXSolverDOF.cpp
```diff
@@ -171,5 +173,32 @@ void WarpXSolverDOF::Define ( WarpX* const        a_WarpX,
...
+    if (m_array_type != FieldType::None) {
+        for (int lev = 0; lev < a_num_amr_levels; ++lev) {
+            const auto& geom = a_WarpX->Geom(lev);
+            for (int n = 0; n < 3; n++) {
+                m_array_lhs[lev][n] = new amrex::MultiFab( m_array[lev][n]->boxArray(),
+                                                           m_array[lev][n]->DistributionMap(),
+                                                           m_array[lev][n]->nComp(),
+                                                           0 );
+                amrex::MultiFab::Copy(*m_array_lhs[lev][n], *m_array[lev][n], 0, 0, m_array[lev][n]->nComp(), 0);
+                m_array[lev][n]->FillBoundary(geom.periodicity());
```
debog, line 187: @JustinRayAngus Is this okay for shared faces/nodes? In PICNIC, the `FillBoundary` replaced the value in the "copy" with that from the "primary". Is that what's happening here too?

JustinRayAngus: I don't know. @ax3l  @RemiLehe @atmyers @WeiqunZhang. Does one of you know the answer to this?

WeiqunZhang: Could you explain you mean by replacing the value in the "copy" with that from the "primary"? Or what do you want to do?

JustinRayAngus: This is for nodal quantities on MPI rank (or box) boundaries where a valid point on the high end of one box is shared with a valid point on the low end of the neighboring box. In PICNIC we declare one side to be the "primary" and the other to be the "copy". Each time we do a communication we copy the "primary" to the "copy".

WeiqunZhang: If that is needed, you can call FillBoundaryAndSync. If the data are consistent, then you just need to call FillBoundary, which only modifies ghost cells/nodes/edges.

WeiqunZhang: If you only need to sync the shared nodes, but not fill ghost cells, you can call OverrideSync.

I am confuse by the code. It does something like
```
a = b;
b.(fillboundary | overridesync | fillboundaryandsync);
```
Should it be done before the assignment?

debog: @JustinRayAngus @WeiqunZhang Yeah, I looked at the AMReX documentation and replaced the `FillBoundary` by `FillBoundaryAndSync`.

@WeiqunZhang Regarding your question
```
a = b;
b.(fillboundary | overridesync | fillboundaryandsync);
```
This is intentional. I need `a` to contain the values (especially on the shared nodes) before they are synced. This is the global DOF object - each element of the MultiFab contains the global DOF (in a serialized vector) of that grid point. `a` contains the unique ones, `b` contains the "primary" ones.

Source/FieldSolver/ImplicitSolvers/WarpXSolverDOF.cpp
WeiqunZhang: This does not seem correct. It does not matter if ncomp is 1. So it's not the bug. But let's suppose ncomp is 2. In that case, dof_arr(i,j,k,0) and dof_arr(i,j,k,2) will have the same value because the RHS does not depend on `v`.

debog: I am surprised I did not fix this bug before!! Thanks!

Source/FieldSolver/ImplicitSolvers/WarpXSolverDOF.H
```diff
@@ -24,7 +24,9 @@ class WarpX;
 struct WarpXSolverDOF
 {
     ablastr::fields::MultiLevelVectorField m_array;
+    ablastr::fields::MultiLevelVectorField m_array_lhs;
     ablastr::fields::MultiLevelScalarField m_scalar;
+    ablastr::fields::MultiLevelScalarField m_scalar_lhs;
```
debog: @WeiqunZhang @JustinRayAngus & others: I guess this one of the reason I just used `MultiFabs` instead of `iMultiFabs` - I am not sure what to do here. Should I just define `m_array`, `m_scalar`, etc as follows?
```
Vector<std::array<MultiFab,3>> m_array;
```

Or should I create a copy of `ablastr/fields/MultiFabRegister.H/cpp` for `iMultiFabs` (and call it `ablastr/fields/iMultiFabRegister.H/cpp`?

WeiqunZhang: I would just use Vector<std::array<iMultiFab,3>> directly.

JustinRayAngus (approved):
Thanks Debo!

I'll let @WeiqunZhang do the final approve + merge.

debog:
No!! Don’t merge, something got broken in the recent changes, and the 2D preconditioner is not working!!

debog:
@WeiqunZhang @JustinRayAngus good to go!

Source/FieldSolver/ImplicitSolvers/WarpXSolverDOF.H
```diff
@@ -23,8 +23,10 @@ class WarpX;
  */
 struct WarpXSolverDOF
 {
-    ablastr::fields::MultiLevelVectorField m_array;
-    ablastr::fields::MultiLevelScalarField m_scalar;
+    amrex::Vector<std::array<amrex::iMultiFab*,3>> m_array;
+    amrex::Vector<std::array<amrex::iMultiFab*,3>> m_array_lhs;
+    amrex::Vector<amrex::iMultiFab*> m_scalar;
+    amrex::Vector<amrex::iMultiFab*> m_scalar_lhs;
```
WeiqunZhang: I think there is a memory leak here. I cannot find where the memory is freed. Even if it were freed somewhere, it would still be too error prone. We should use either `amrex::Vector<std::array<amrex::iMultiFab,3>>` or `amrex::Vector<std::array<std::unique_ptr<iMultiFab>,3>>`.

Note that there are a number of amrex helper functions that can be used to create `Vector<std::array<iMultiFab*,3>>`, when needed. https://github.com/AMReX-Codes/amrex/blob/de2510c13339a9f1e583da0a74eb419912335232/Src/Base/AMReX_Vector.H#L136

WeiqunZhang: I also noticed another memory leak WarpXSolverVec.
```
    ablastr::fields::MultiLevelVectorField m_array_vec;
    ablastr::fields::MultiLevelScalarField m_scalar_vec;
```
`MultiLevelVectorField` is `Vector<std::array<MulitFab*,3>>`. We should use
```
    Vector<std::array<std::unique_ptr<MultiFab>,3>> m_array_vec;
    Vector<std::array<std::unique_ptr<MultiFab>,3>> m_scalar_vec;
```

There are a number of typedefs in `ablastr/fields/MultiFabRegister.H` defined for convenience. But they are dangerous. Their names are not transparent, and unsafe by default. @ax3l @RemiLehe @EZoni

WeiqunZhang (changes requested):

Source/FieldSolver/ImplicitSolvers/WarpXSolverDOF.H
WeiqunZhang: Fix a typo in my comments above. I meant we should std::unique_ptr.

WeiqunZhang:
In WarpX_PETSc.H, there are
```
        //! Matrix (lin operator) object                                                                                                                                     
        MatObj* m_A = nullptr;
        //! Matrix (PC) object                                                                                                                                               
        MatObj* m_P = nullptr;
        //! solution vector                                                                                                                                                  
        VecObj* m_x = nullptr;
        //! right-hand-side vector                                                                                                                                           
        VecObj* m_b = nullptr;

        //! Linear operator object                                                                                                                                           
        LinOpType* m_linop = nullptr;
```
Currently, they are deleted in the destructor and there is no memory leak. However, we should try to make it more memory safe by using `std::unique_ptr` here.

Source/FieldSolver/ImplicitSolvers/WarpXSolverDOF.H
JustinRayAngus: @WeiqunZhang I believe you meant the following:
```
Vector<std::array<std::unique_ptr<MultiFab>,3>> m_array_vec;
Vector<std::unique_ptr<MultiFab>> m_scalar_vec;
```

In either case, I think changes to the WarpXSolverVec class should be made in a separate PR. Feel free to create such a PR and I can review it.

debog:
@WeiqunZhang I can't use `std::unique_ptr<Something>` here because the `Something` is forward-declared (thus incomplete type).

> In WarpX_PETSc.H, there are
> 
> ```
>         //! Matrix (lin operator) object                                                                                                                                     
>         MatObj* m_A = nullptr;
>         //! Matrix (PC) object                                                                                                                                               
>         MatObj* m_P = nullptr;
>         //! solution vector                                                                                                                                                  
>         VecObj* m_x = nullptr;
>         //! right-hand-side vector                                                                                                                                           
>         VecObj* m_b = nullptr;
> 
>         //! Linear operator object                                                                                                                                           
>         LinOpType* m_linop = nullptr;
> ```
> 
> Currently, they are deleted in the destructor and there is no memory leak. However, we should try to make it more memory safe by using `std::unique_ptr` here.

Source/FieldSolver/ImplicitSolvers/WarpXSolverDOF.H
debog: @WeiqunZhang I switched to using `std::unique_ptr` here (see [commit](https://github.com/BLAST-WarpX/warpx/pull/6184/commits/c6efa9152b5a21315054cef80140c2ef49d3eb53)). The code runs fine but at termination I get this error now:
```
pure virtual method called
terminate called without an active exception
```
Do you know what's happening?

WeiqunZhang:
> @WeiqunZhang I can't use `std::unique_ptr<Something>` here because the `Something` is forward-declared (thus incomplete type).

The declaration of `std::unique_ptr` does not require a complete type. You just need the constructor and destructor to be in a cpp file that has the complete type information.

Source/FieldSolver/ImplicitSolvers/WarpXSolverDOF.H
WeiqunZhang: I think there is a static variable somewhere that is causing this issue.

WeiqunZhang: It's because of this `static std::unique_ptr<WarpXSolverDOF> m_dofs;` in `WarpXSolverVec.H`. It seems that it should not be a static variable. But if it's absolutely necessary, we can fix it. Just let me know.

debog: It should be a `static` variable; the native solvers (Newton and GMRES) create several objects of type `WarpXSolverVec`; this shouldn't result in creating separate copies of `m_dofs`.

WeiqunZhang: I see. @JustinRayAngus @roelof-groenewald  This may have implication for using WarpXSolverVec for other solvers.

A simple fix while keeping it static could be
```
--- a/Source/FieldSolver/ImplicitSolvers/WarpXSolverVec.cpp
+++ b/Source/FieldSolver/ImplicitSolvers/WarpXSolverVec.cpp
@@ -105,6 +105,7 @@ void WarpXSolverVec::Define ( WarpX*  a_WarpX,
     if (m_dofs == nullptr) {
         m_dofs = std::make_unique<WarpXSolverDOF>();
         m_dofs->Define(m_WarpX, m_num_amr_levels, m_vector_type_name, m_scalar_type_name);
+        amrex::ExecOnFinalize([p=&m_dofs] () { p->reset(); });
     }
 
     m_is_defined = true;
```

JustinRayAngus:
@WeiqunZhang Are the recent recommended changes required for this PR? If there is actually a memory leak, then we should fix that, but if they are mainly just better coding practices then I think those could be addressed in a separate PR. Would be nice to finalize this PR so we can move forward.

WeiqunZhang:
The code is currently in a state that crashes at the end of the run. So we cannot merge it as is. We either have to revert the recent changes or fix it.

The petsc class does not have a memory leak. It's a best practice issue.

The other two are *currently* one-time leak. If we don't fix it, it will bite us in the future when we start to do load balance. It's okay to delay. But it's also not hard to fix. The bottom line is the current state of the PR is unmergeable.

WeiqunZhang:
> But it's also not hard to fix.

I made a one-line suggestion in a comment above.

Source/FieldSolver/ImplicitSolvers/WarpXSolverDOF.H
WeiqunZhang: ^ @JustinRayAngus I think this will make the CI pass.

debog:
@JustinRayAngus @WeiqunZhang Give me a day or two; I'll fix the code and make the other change that Weiqun suggested as well. These are not big changes.

Source/FieldSolver/ImplicitSolvers/WarpXSolverDOF.H
debog: Done!

debog:
Yups, that worked. Made this change as well.

> > @WeiqunZhang I can't use `std::unique_ptr<Something>` here because the `Something` is forward-declared (thus incomplete type).
> 
> The declaration of `std::unique_ptr` does not require a complete type. You just need the constructor and destructor to be in a cpp file that has the complete type information.

WeiqunZhang (approved):

PR #6187, opened 2025-09-18
Move PushType enum to WarpXAlgorithmSelection.H

EZoni:
In #5955 I noticed that the `PushType` enum is defined in its own separate header file. Is that necessary? If not, with this PR I propose that we move the enum to the WarpXAlgorithmSelection.H header file, together with most of the other enums for algorithm selection.


dpgrote (approved):
This looks good to me, a nice cleanup.

PR #6189, opened 2025-09-18
Add more dependencies to root JSON file

EZoni:
Add more dependencies to the JSON file in the root directory, to improve automation:
- [x] pybind11
- [x] picmistandard (including workaround for PICMI version update in the requirements.txt files)


dependencies.json
```diff
@@ -3,7 +3,11 @@
     "version_amrex": "25.09",
     "version_pyamrex": "25.09",
     "version_picsar": "25.06",
+    "version_pybind11": "3.0.0",
+    "version_picmi": "0.33.0",
     "commit_amrex": "a377c17c36a9885ca1ce4975aaee5a240d8eeaf1",
     "commit_pyamrex": "e73fc6081d1750d1558ae43f7904982535e93fb3",
-    "commit_picsar": "0c329e66010267662a82219f7de7abbd231463f4"
+    "commit_picsar": "0c329e66010267662a82219f7de7abbd231463f4",
+    "commit_pybind11": "ed5057ded698e305210269dafa57574ecf964483",
```
EZoni: Using commit https://github.com/pybind/pybind11/commit/ed5057ded698e305210269dafa57574ecf964483, which corresponds to [version 3.0.0](https://github.com/pybind/pybind11/releases/tag/v3.0.0)

Python/setup.py
```diff
@@ -64,11 +64,12 @@
 else:
     package_data = {}
 
-# Parse WarpX version information
+# Parse WarpX, PICMI versions information
 dependencies_file = "../dependencies.json"
 with open(dependencies_file, "r") as file:
     dependencies_data = json.load(file)
 warpx_version = dependencies_data.get("version_warpx")
+picmi_version = dependencies_data.get("version_picmi")
```
ax3l, line 72: I think PICMI's version is set in a couple locations:
- [x] here (`Python/setup.py`)
- [x] `requirements.txt`
- [x] `Docs/requirements.txt`

Python/setup.py
```diff
@@ -77,7 +78,7 @@
     package_dir={"pywarpx": "pywarpx"},
     description="""Wrapper of WarpX""",
     package_data=package_data,
-    install_requires=["numpy", "picmistandard==0.33.0", "periodictable"],
+    install_requires=["numpy", f"picmistandard=={picmi_version}", "periodictable"],
```
ax3l, line 81: I think the file here should maybe just use `requirements.txt` instead of documenting the dependencies again...?

Python/setup.py
ax3l: Oh, I see it below.

ax3l (approved):
This looks great to me, thank you!

cmake/dependencies/pybind11.cmake
```diff
@@ -52,7 +52,16 @@ option(WarpX_pybind11_internal "Download & build pybind11" ON)
...
     "Repository URI to pull and build pybind11 from if(WarpX_pybind11_internal)")
-set(WarpX_pybind11_branch "v3.0.1"
+
+# Parse pybind11 version and commit information
+file(READ "${WarpX_SOURCE_DIR}/dependencies.json" dependencies_data)
+string(JSON pybind11_version GET "${dependencies_data}" version_pybind11)
+string(JSON pybind11_commit GET "${dependencies_data}" commit_pybind11)
+
+# Strip "v" prefix from version for find_package
+string(REGEX REPLACE "^v" "" pybind11_version "${pybind11_version}")
```
EZoni, line 62: Bug fix copied from https://github.com/AMReX-Codes/pyamrex/pull/486.

EZoni:
I think there is one more problem to fix, both here and in pyAMReX.

For release updates we replace the commit hash with the release tag. However, I think this is buggy for dependencies that do not publish monthly releases regularly, e.g., pybind11 or even PICSAR. 

Replacing a commit hash from the latest weekly update with the latest available release tag may actually bring us backward in the commit history, if the latest available release tag is older than the latest commit.

I think we should be doing something different. I don't see major drawbacks in always keeping the version and the commit separate from each other, using the release tag only for the version (e.g., `version_pybind11`) and the commit hash only for the commit (e.g., `commit_pybind11`) in dependencies.json. And I think I would do so for all dependencies, to make things simpler.

EZoni:
@ax3l 

In view of the upcoming monthly release, this may be worth reviewing. In particular, the outstanding questions in https://github.com/BLAST-WarpX/warpx/pull/6189#issuecomment-3361923244.

ax3l:
> For release updates we replace the commit hash with the release tag. However, I think this is buggy for dependencies that do not publish monthly releases regularly, e.g., pybind11 or even PICSAR.

We generally try to publish our releases only on tags, because otherwise we make the live of downstream package maintainers (spack, conda, etc.) very hard.

For pybind11, I always use published releases.
For PICSAR, whenever a month has a change in it, we do a new tag before we publish WarpX.

Some tags might be from prior months in our dependencies, which is fine.

dependencies.json
```diff
@@ -3,7 +3,11 @@
...
     "version_picsar": "25.06",
+    "version_pybind11": "v3.0.1",
+    "version_picmi": "0.34.0",
     "commit_amrex": "c4fcda1fb2ea238c3e6fa273dc9cfcde54b4cc51",
     "commit_pyamrex": "5fc00b46b2a60b2c20b9067500c4b68f709ccb94",
-    "commit_picsar": "0c329e66010267662a82219f7de7abbd231463f4"
-}
\ No newline at end of file
+    "commit_picsar": "0c329e66010267662a82219f7de7abbd231463f4",
+    "commit_pybind11": "af796d0a99f0cbd9aebb10591257c41a56811cf6",
```
ax3l: How did you end up with this sha exactly?
pybind11 tag `v3.0.1` is commit `f5fbe867d2d26e4a0a9177a51f6e568868ad3dc8`

https://github.com/pybind/pybind11/releases/tag/v3.0.1

Tools/Release/update_dependencies.py
```diff
@@ -42,6 +42,22 @@ def update(args):
...
             "https://api.github.com/repos/ECP-WarpX/picsar/tags"
         )
+    if args.all or args.pybind11:
+        repo_dict["pybind11"] = {}
+        repo_dict["pybind11"]["commit"] = (
+            "https://api.github.com/repos/pybind/pybind11/commits/master"
+        )
+        repo_dict["pybind11"]["tags"] = (
+            "https://api.github.com/repos/pybind/pybind11/tags"
+        )
```
ax3l, line 52: Note that we do not do a weekly tracking of pybind11. We manually update when we want a newer release.

Most importantly, when we release WarpX we do _not_ depend on the pybind11 development branch (or any dependency as of their development branch).

ax3l (changes requested):
The logic here is off for at least pybind11.

Note that we do on release track a tag, not development. We only track (some) dependencies development weekly between releases.

EZoni:
I pushed a fix similar to https://github.com/AMReX-Codes/pyamrex/pull/517.

ax3l (approved):
Thank you! :)

PR #6190, opened 2025-09-18
Add Python hook to change `do_no_gather` and `do_not_deposit` value for particle container

roelof-groenewald:
This PR follows the same logic as already implemented for `do_not_push`.


dpgrote (approved):
Looks good, thanks!

PR #6199, opened 2025-09-23
Fix a bug in the 2d mode of the IGF solver

WeiqunZhang:
When there is only one node in the z-direction, the previous way of creating the domain for the IGF solver ends up with two nodes. That's not the desired behavior. Using the nodal BoxArray in MultiFab rho solves the issue.


ax3l (approved):
Thank you, that's great! :tada:

PR #6202, opened 2025-09-23
Implicit particle suborbits refactor.

JustinRayAngus:
This PR is a follow-up to PR #5969 where the ability to use particle suborbits in the implicit schemes was introduced. Currently, an approximation is used to compute the contribution of suborbit particles to the mass matrices that ignores terms that arise owing to coupling between suborbits. This approximation is not always a good one and has been observed to lead to an increase in total Newton iterations when being used for the Jacobian calculation, thereby increasing the wall time.

This PR refactors the suborbit implementation to do a direct particle-based calculation of J for suborbit particles during the linear stage of JFNK when the mass matrices are being used, mimicking the behavior in PICNIC. However, unlike PICNIC, the suborbit particles are still included in the preconditioner, where the approximation discussed above is sufficient.


Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -244,13 +243,121 @@ namespace {
...
+
+    auto num_previous = unconverged_indices.size();
+    unconverged_indices.resize(num_previous + num_unconverged_particles);
+    saved_weights.resize(num_previous + num_unconverged_particles);
+
+    long * unconverged_i = unconverged_indices.data() + num_previous;
+    amrex::ParticleReal * saved_w = saved_weights.data() + num_previous;
+
+    if (nsuborbits) {
+        // This looks for the unconverged particles, which had been flagged as invalid
```
dpgrote: This comment could be updated since the invalid flag is no longer used.

```suggestion
        // This looks for the particles that require suborbits to converge.
```

dpgrote (approved):
This looks good (except for the one minor comment).

PR #6203, opened 2025-09-24
[WIP] Accept external fields from file in F-order 

aeriforme:
This PR extends the external-field-loading feature controlled by the following input parameters:
```
warpx.B_ext_grid_init_style = read_from_file 
warpx.E_ext_grid_init_style = read_from_file 
warpx.read_fields_from_path = <path/to/openpmd/file>
```
Specifically, `read_fields_from_path` now accepts openPMD files where the data is stored in F column-major order (or, equivalently, C-order with axis labels = {z, y, x}).

This is useful because WarpX openPMD outputs are saved in C-order with axis labels = {z,y,x}. Therefore with this PR one can initialize an external field from a file that was obtained from another WarpX simulation.

TO DO:

- [ ] fix behavior at boundaries: the loaded field is incorrect at cells `0,1,N-2,N-1`
- [ ] check that it works for any grid/staggering choices


ax3l:
@aeriforme I think @WeiqunZhang might have added the F/C conditional transpose now already in #6221

Source/Initialization/ExternalField.cpp
```diff
@@ -225,19 +239,19 @@ ExternalFieldReader::ExternalFieldReader (std::string const& read_fields_from_pa
...
 
     // Load the first component if F_component is empty
     auto FC = F_component.empty() ? F.begin()->second : F[F_component];
     const auto extent = FC.getExtent();
 
+    const auto offset = F.gridGlobalOffset();
+    const auto position = FC.getAttribute("position").get<std::vector<double>>();
+    AMREX_D_TERM(m_external_field_view.offset[0] = amrex::Real(offset.at(0+do_swap) + position.at(0+do_swap) * d.at(0+do_swap));,
+                 m_external_field_view.offset[1] = amrex::Real(offset.at(1) + position.at(1) * d.at(1));,
+                 m_external_field_view.offset[2] = amrex::Real(offset.at(2-do_swap) + position.at(2-do_swap) * d.at(2-do_swap)););
```
aeriforme, line 254: Move this in a separate PR. 
CC @ax3l

PR #6209, opened 2025-09-25
Some cleaning of the #include directives

lucafedeli88:
This PR fixes some `#include` issues found with the `iwyu` tool (mostly unnecessary inclusions).
There are several other issues like these ones, but I prefer to address them in separate PRs in order to keep them small.

### DIFF
--- a/Source/Diagnostics/FlushFormats/FlushFormatInSitu.cpp
+++ b/Source/Diagnostics/FlushFormats/FlushFormatInSitu.cpp
@@ -1,6 +1,5 @@
 #include "FlushFormatInSitu.H"
 
-#include "WarpX.H"
 #include "Utils/TextMsg.H"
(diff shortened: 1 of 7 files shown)


EZoni (approved):
Thanks, @lucafedeli88! Looks good to me, I can merge once the CI checks have completed.

Question for a future PR, but related to this one: Do you think it would be useful to add a CI workflow that runs the IWYU tool regularly going forward?

lucafedeli88:
> Question for a future PR, but related to this one: Do you think it would be useful to add a CI workflow that runs the IWYU tool regularly going forward?

It would be great, but unfortunately I don't think that it can be transformed into a CI test right now. As far as I know, for a given `foo.cpp` file it would ask to include directly also the headers that are already included in `foo.h`. Moreover, it often recommends forward declarations of classes in header files, which we do only by using forward declaration headers.

PR #6211, opened 2025-09-25
DSMC: rename m_ioniz_product to m_product

RemiLehe:
This is in preparation for https://github.com/BLAST-WarpX/warpx/pull/6125/commits

The masses of the products will be used for the DSMC two-product reaction (that generalizes the charge exchange process), and not only for ionization. Therefore it makes sense to rename these variables.

### DIFF
--- a/Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
+++ b/Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
@@ -168,7 +168,7 @@ public:
         const int num_product_species = m_num_product_species;
         const auto ionization_energy = m_ionization_energy;
 
-        // Grab the masses of ionization products
+        // Grab the masses of reaction products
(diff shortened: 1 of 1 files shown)


Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -169,11 +169,11 @@ public:
         const auto ionization_energy = m_ionization_energy;
 
         // Grab the masses of ionization products
```
RemiLehe:
```suggestion
        // Grab the masses of reaction products
```

roelof-groenewald (approved):
Looks good. I realize now that perhaps `m_ioniz_product1` was not a great name since we often use `m_something` to denote a member variable. Perhaps using `mass_product1` would be better for that reason. I'll approve this as is though and leave it up to you if you want to change the names one more time.

RemiLehe:
Good idea. I will update the name of the masses here.

PR #6212, opened 2025-09-25
Binary collisions: move two-product initialization out of the fusion folder

RemiLehe:
This is in preparation for https://github.com/BLAST-WarpX/warpx/pull/6125

The two-product initialization function is more general than just the fusion application, and will be reused in DSMC. Therefore, this moves the function out of the the fusion folder and renames variables to make them non-specific to fusion.


roelof-groenewald (approved):
Looks good!

PR #6213, opened 2025-09-25
Shared deposition not implemented in RCYLINDER and RSPHERE geometries

EZoni:
I was compiling WarpX on Frontier in all geometries and ran into a build error with the RCYLINDER and RSPHERE geometries. This PR skips the shared deposition routines when building in these geometries for now.


Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -510,7 +510,10 @@ void doDepositionSharedShapeN (const GetParticlePosition<PIdx>& GetPosition,
     amrex::IntVect const jy_type = jy_fab.box().type();
     amrex::IntVect const jz_type = jz_fab.box().type();
 
+#if !defined(WARPX_DIM_RCYLINDER) && !defined(WARPX_DIM_RSPHERE)
```
dpgrote: I used this more direct test
```
#if defined(WARPX_ZINDEX)
```

Source/Particles/Deposition/CurrentDeposition.H
```diff
@@ -594,7 +597,11 @@ void doDepositionSharedShapeN (const GetParticlePosition<PIdx>& GetPosition,
             const unsigned int ip = permutation[ip_orig];
             depositComponent<depos_order>(GetPosition, wp, uxp, uyp, uzp, ion_lev, jx_buff, jx_type,
                                           relative_time, dinv, xyzmin, lo, q, n_rz_azimuthal_modes,
-                                          ip, zdir, NODE, CELL, 0);
+                                          ip,
+#if defined(WARPX_ZINDEX)
+                                          zdir,
+#endif
+                                          NODE, CELL, 0);
```
EZoni: I still don't love these kinds of `if` conditions in the middle of the list of a function's arguments, but I don't have in mind an alternative right now - and I think we do this in many places in the code.

EZoni:
@dpgrote 

I think the problem may be bigger than I thought here. 

I think the whole file Source/Particles/Deposition/SharedDepositionUtils.H hasn't been updated at all for 1D cylindrical and 1D spherical geometries.

I was simply putting guardrails to fix the compilation bugs, but maybe we should actually update the deposition routines to have the correct implementation for 1D cylindrical and 1D spherical geometry in this file as well? 

Do you have any guidance on this?

dpgrote:
> @dpgrote
> 
> I think the problem may be bigger than I thought here.
> 
> I think the whole file Source/Particles/Deposition/SharedDepositionUtils.H hasn't been updated at all for 1D cylindrical and 1D spherical geometries.
> 
> I was simply putting guardrails to fix the compilation bugs, but maybe we should actually update the deposition routines to have the correct implementation for 1D cylindrical and 1D spherical geometry in this file as well?
> 
> Do you have any guidance on this?

Good point. I think the thing to do now is not allow shared deposition with 1D sphere and cylinder. Maybe change the macro at the start of that routine to

```
#if (defined(AMREX_USE_HIP) || defined(AMREX_USE_CUDA)) && !(defined(WARPX_DIM_RCYLINDER) || defined(WARPX_DIM_RSPHERE))
```
And the same in WarpX.cpp where `do_shared_mem_current_deposition` is read in.

Alternative, updating `depositComponent` for 1D sphere and cylinder would not be much work. It would be mostly copying code from one of the other deposition routines, such as `doDepositionShapeNKernel`. Maybe do the above for now to get it working quickly, and a PR can be done later to implement it.

EZoni:
@dpgrote 

Thanks for the input. I pushed an update to skip the shared deposition routine in RCYLINDER and RSPHERE geometries. 

Does it look good to you? I just compiled on Frontier and it seems to build without errors.

I think it would be good to merge this before we merge #6239 for the October release.

RemiLehe (approved):

PR #6214, opened 2025-09-25
[WIP] Generalize two product reaction for photons

RemiLehe:
Merge https://github.com/BLAST-WarpX/warpx/pull/6125 first


Source/Particles/Collision/BinaryCollision/TwoProductUtil.H
```diff
@@ -115,39 +121,72 @@ namespace {
...
-        const amrex::ParticleReal vc_sq   = vcx*vcx + vcy*vcy + vcz*vcz;
-
-        // Convert momentum of first product to lab frame, using equation (13) of F. Perez et al.,
-        // Phys.Plasmas.19.083104 (2012)
+        // Convert momentum of first product to lab frame, using equation (13)
+        // of F. Perez et al., Phys.Plasmas.19.083104 (2012)
         if ( vc_sq > std::numeric_limits<amrex::ParticleReal>::min() )
         {
-            const amrex::ParticleReal gc = 1._prt / std::sqrt( 1._prt - vc_sq*inv_csq );
             const amrex::ParticleReal g_star = std::sqrt(1._prt + p_star_f_sq / (m1_out*m1_out*c_sq));
```
RemiLehe: We need to check that this line is still correct, when photons are involved.

Source/Particles/Collision/BinaryCollision/TwoProductUtil.H
```diff
@@ -63,6 +67,7 @@ namespace {
                             amrex::ParticleReal& u2z_out,
```
RemiLehe, line 67: Note, should we actually change the signature of this function, so that it uses `p` as input, instead of `u`?

PR #6217, opened 2025-09-26
LinearBreitWheelerCollisionFunc.H and LinearComptonCollisionFunc.H  : remove unnecessary #include of the WarpX header

lucafedeli88:
This PR removes a couple of unnecessary `#include` directives


RemiLehe (approved):
Thanks for this PR!

PR #6221, opened 2025-09-27
Load Density: Distributed Approach

WeiqunZhang:
Previously, the data are duplicated on all processes. For a very large file, this might be problematic. A distributed approach is developed in this PR.

For the initial particle creation, we load the data we need in a distributed way. Because we also know the particle BoxArray and DistributionMapping, we can then redistribute the data to where they are needed.

For the continuously particle injection during moving window, the data are duplicated on every process, but they are still distributed in time in the sense that we only load the data needed for the new region created by the moving window plus some extra caching so that we don't have to do disk I/O too frequently.

A caveat is that it might be difficult to determine the range of the data we need if the boosted frame is used and the boosted velocity is not a constant. In that case, one might encounter an out-of-bound error at run time. One can also disable the distribute approach with a run time parameter, `<species_name>.read_density_distributed`.

This also add support for Fortran order data and axes with the order of either x, y and z or z, y and x.


Source/Initialization/ExternalField.cpp
```diff
@@ -189,99 +189,375 @@ ExternalFieldParams::ExternalFieldParams(const amrex::ParmParse& pp_warpx)
...
+#if defined(WARPX_DIM_RZ)
+    // extent[0] is for theta
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(extent.size() == 3 && extent[0] == 1,
+                                     "External field reading is not implemented for more than one RZ mode (see #3829)");
+    if (xyz_order) {
+        m_size[0] = extent[1];
+        m_size[1] = extent[2];
+    } else {
+        m_size[0] = extent[2];
+        m_size[1] = extent[1]; // xxxxx Ask Axel if this is correct.
```
WeiqunZhang: Question for @ax3l

Source/Initialization/ExternalField.cpp
```diff
@@ -189,99 +189,375 @@ ExternalFieldParams::ExternalFieldParams(const amrex::ParmParse& pp_warpx)
...
+        m_mf = std::move(tmpmf);
+        m_fab.clear();
+        m_FC_data_cpu.reset();
+    }
+}
+
+void ExternalFieldReader::make_cache_box (amrex::RealBox const& pbox)
+{
+    m_cache_domain = pbox;
+    // xxxx Ask Remi. How general do we need to be about moving window
```
WeiqunZhang: Question for @RemiLehe

Source/Utils/SpeciesUtils.cpp
```diff
@@ -106,8 +107,10 @@ namespace SpeciesUtils {
                 density_parser->compile<3>()));
         } else if (rho_prof_s == "read_from_file") {
             std::string density_file;
+            bool distributed = false; // xxxxx Ask Remi. should the default be true or false?
```
WeiqunZhang: Question for @RemiLehe

Source/Initialization/InjectorDensity.cpp
```diff
@@ -75,14 +107,51 @@
...
-InjectorDensityFromFile::InjectorDensityFromFile (std::string const& a_file_name)
+InjectorDensityFromFile::InjectorDensityFromFile (std::string const& a_file_name,
+                                                  amrex::Geometry const& a_geom,
+                                                  bool a_distributed)
 {
-    m_external_field_reader = new ExternalFieldReader(a_file_name, "density", "");
-    m_external_field_view = m_external_field_reader->getView();
+    m_external_field_reader = new ExternalFieldReader
+        (a_file_name, "density", "", a_geom.ProbLoArray(), a_geom.CellSizeArray(),
+         amrex::convert(a_geom.Domain(),amrex::IntVect(1)), a_distributed);
```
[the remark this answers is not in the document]
WeiqunZhang: Intentional

Source/Initialization/ExternalField.cpp
```diff
@@ -189,99 +217,362 @@ ExternalFieldParams::ExternalFieldParams(const amrex::ParmParse& pp_warpx)
...
+            lo = b.smallEnd();
+            hi = b.  bigEnd();
+        } else {
+            has_load = false;
+        }
+    }
+
+    openPMD::Offset chunk_offset(extent.size(),0);
+    openPMD::Extent chunk_extent(extent.size(),1);
+#if defined(WARPX_DIM_RZ)
```
WeiqunZhang, line 362: @ax3l Is this correct? Should offset have two or three elements and what they should be in the rz mode?

aeriforme:
Does this PR replace and expand [this other PR](https://github.com/BLAST-WarpX/warpx/pull/6203)?

WeiqunZhang:
> Does this PR replace and expand [this other PR](https://github.com/BLAST-WarpX/warpx/pull/6203)?

It depends on how long this PR takes. This PR needs amrex's parallel copy function that assumes Fortran order. Thus, we need to do the transpose of the data.

Source/Utils/SpeciesUtils.cpp
```diff
@@ -106,8 +107,10 @@ namespace SpeciesUtils {
                 density_parser->compile<3>()));
         } else if (rho_prof_s == "read_from_file") {
             std::string density_file;
+            bool distributed = true; // xxxxx Ask Remi. should the default be true or false?
```
RemiLehe: Yes, I think that this should indeed be the default. Could you add a short description in `parameters.rst`?

Source/Initialization/InjectorDensity.H
```diff
@@ -122,10 +125,20 @@ private:
 // struct whose getDensity returns density from file.
 struct InjectorDensityFromFile
 {
-    InjectorDensityFromFile (std::string const& a_file_name);
+    InjectorDensityFromFile (std::string const& a_file_name, amrex::Geometry const& a_geom, bool a_distributed);
 
     void clear ();
 
+    void prepare (amrex::BoxArray const& grids,
```
RemiLehe, line 139: Could you add a docstring for each of the `prepare` functions?

Source/Initialization/ExternalField.cpp
RemiLehe: Yes, my understanding from `thetaMode` in openPMD is that this is indeed what we need here.

Source/Initialization/ExternalField.cpp
```diff
@@ -189,99 +190,363 @@ ExternalFieldParams::ExternalFieldParams(const amrex::ParmParse& pp_warpx)
...
     // Determine the chunk data that will be loaded.
-    // Now, the full range of data is loaded.
-    // Loading chunk data can speed up the process.
-    // Thus, `chunk_offset` and `chunk_extent` should be modified accordingly in another PR.
-    const openPMD::Offset chunk_offset(extent.size(), 0);
-    const openPMD::Extent chunk_extent = extent;
+    BoxArray grids;
+    DistributionMapping dmap;
+    bool has_load = true;
+    if (m_distributed && !m_moving_window) {
```
RemiLehe, line 341: Could you add a few comments here to explain
- why the behavior depends on `moving_window`
- what is the logic to determine which MPI rank gets which box

Source/Fluids/WarpXFluidContainer.cpp
```diff
@@ -174,20 +177,60 @@ void WarpXFluidContainer::InitData(
...
+        h_inj_rho->prepare(mf.boxArray(), mf.DistributionMap(), IntVect(0));
+    }
+
     // Extract grid geometry properties
     const auto dx = geom_lev.CellSizeArray();
     const auto problo = geom_lev.ProbLoArray();
     const amrex::Real clight = PhysConst::c;
 
+#ifdef AMREX_USE_OMP
+    std::unique_ptr<void,amrex::DataDeleter> inj_rho_data;
```
RemiLehe: could you add a comment explaining what this does?

Source/Fluids/WarpXFluidContainer.cpp
```diff
@@ -174,20 +177,60 @@ void WarpXFluidContainer::InitData(
...
+    }
+#endif
+
     // Loop through cells and initialize their value
 #ifdef AMREX_USE_OMP
 #pragma omp parallel if (amrex::Gpu::notInLaunchRegion())
 #endif
     for (MFIter mfi(*fields.get(name_mf_N, lev), TilingIfNotGPU()); mfi.isValid(); ++mfi)
     {
+        InjectorDensity* inj_rho = d_inj_rho;
```
RemiLehe, line 207: Could you add a comment explaining what this does?

Source/Fluids/WarpXFluidContainer.cpp
```diff
@@ -252,6 +295,10 @@ void WarpXFluidContainer::InitData(
 
             }
         );
+
+        if (h_inj_rho->distributed()) {
```
RemiLehe, line 276: Could you explain why this is necessary?

Source/Particles/ParticleCreation/AddParticles.cpp
```diff
@@ -738,13 +738,19 @@ PhysicalParticleContainer::AddPlasmaFromFile(PlasmaInjector & plasma_injector,
...
 void
-PhysicalParticleContainer::AddPlasma (PlasmaInjector const& plasma_injector, int lev, amrex::RealBox part_realbox)
+PhysicalParticleContainer::AddPlasma (PlasmaInjector& plasma_injector, int lev, amrex::RealBox part_realbox)
 {
     WARPX_PROFILE("PhysicalParticleContainer::AddPlasma()");
 
     // If no part_realbox is provided, initialize particles in the whole domain
     const Geometry& geom = Geom(lev);
-    if (!part_realbox.ok()) { part_realbox = geom.ProbDomain(); }
+    bool initial_injectcion;
```
RemiLehe: Could you fix the typo throughout this file?
```suggestion
    bool initial_injection;
```

RemiLehe: Also: could you a comment to explain what happens for `initial_injection=true` and `initial_injection=false` (my understanding is that `initial_injection=false` corresponds to the moving window).

Source/Initialization/ExternalField.cpp
WeiqunZhang: I assume the offset should be the same as extent in terms of which element is for which direction.

Source/Initialization/ExternalField.cpp
RemiLehe: Offline discussion: at this point the data is distributed in an arbitrary way.

Source/Initialization/ExternalField.cpp
```diff
@@ -189,99 +190,363 @@ ExternalFieldParams::ExternalFieldParams(const amrex::ParmParse& pp_warpx)
...
+            }
+            Box box(lo, hi);
+            if (box.isEmpty()) {
+                box = Box(IntVect(0),IntVect(0));
+            }
+            bl.push_back(box);
+        }
+        BoxArray ba(std::move(bl));
+        FabArray<BaseFab<double>> tmpmf(ba,dmap,1,0);
+        tmpmf.ParallelCopy(m_mf);
```
RemiLehe, line 487: Offline discussion: this copies from the arbitrary initially loaded distribution to the actual boxarray distribution.

Source/Initialization/InjectorDensity.H
```diff
@@ -122,10 +125,29 @@ private:
...
+                  amrex::IntVect const& ngrow);
+
+    // This function needs to be called before we do continuous injection of
+    // particles to inform the injector to load the data we need.
+    void prepare (amrex::RealBox const& pbox);
+
+    // For distributed data, this function prepares the injector for loading
+    // data in this Box. Here `li` is the local index for the Box in a
+    // particle container.
+    void prepare (int li);
```
RemiLehe, line 152: Offline discussion: this is called inside the MFIter, for the local box.

Source/Initialization/InjectorDensity.H
```diff
@@ -122,10 +125,29 @@ private:
...
-    InjectorDensityFromFile (std::string const& a_file_name);
+    InjectorDensityFromFile (std::string const& a_file_name, amrex::Geometry const& a_geom, bool a_distributed);
 
     void clear ();
 
+    // This function needs to be called before intializing the density
+    // profile, if the openPMD data are distributed. When the openPMD data
+    // are loaded in the constructor of this class, we do not know the
+    // paricle containers' BoxArray and DistributionMapping yet.
+    void prepare (amrex::BoxArray const& grids,
```
RemiLehe, line 139: Offline discussion: this is called outside the MFIter

Source/Fluids/WarpXFluidContainer.cpp
RemiLehe: We can try to think how to hide the complexity here in a separate function...

Source/Initialization/ExternalField.cpp
WeiqunZhang: Yes, we do need to support moving window in any direction.

Source/Initialization/ExternalField.cpp
ax3l: @WeiqunZhang sorry for the delay during my absence (and backlog).
In RZ, `gridGlobalOffset` and `gridSpacing` should have, as the `axisLabels` attribute, 2 values - one for R and one for Z.

https://github.com/openPMD/openPMD-standard/blob/1.1.0/STANDARD.md#mesh-based-records

The arrays in RZ are 3-dimensional, the additional dimension are the modes.

Source/Initialization/ExternalField.cpp
```diff
@@ -189,99 +190,368 @@ ExternalFieldParams::ExternalFieldParams(const amrex::ParmParse& pp_warpx)
...
-    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(F.getAttribute("dataOrder").get<std::string>() == "C",
-                                     "Reading from files with non-C dataOrder is not implemented");
-#endif
+    bool c_order = F.getAttribute("dataOrder").get<std::string>() == "C";
+    amrex::ignore_unused(c_order);
 
     auto axisLabels = F.getAttribute("axisLabels").get<std::vector<std::string>>();
     auto fileGeom = F.getAttribute("geometry").get<std::string>();
 
+    bool xyz_order = true;
```
ax3l, line 223: I assume with `xyz_order` you mean: do we need to transpose the data when we read it, or not. This is a good criteria to put in a bool.

In openPMD 1.X, we unfortunately make the test below a bit more complicated, because of the `dataOrder` attribute. In essence: the logic here is already correct if `dataOrder == "C"`.

If `dataOrder == "F"` in the file, then all array attributes would have to be interpreted in inverse order, starting with `axisLabels` and then following through for `gridSpacing` and `gridGlobalOffset`. (This is a superfluous, and we removed this extra possibility to inverse attributes in openPMD 2.*, which is in development. In the future, we just demand that all attributes are written as if the `dataOrder` was `"C"`.) The attribute `dataOrder` has no influence on the actual data, just the labeling.

Source/Initialization/ExternalField.cpp
```diff
@@ -189,99 +190,368 @@ ExternalFieldParams::ExternalFieldParams(const amrex::ParmParse& pp_warpx)
...
+
 #if defined(WARPX_DIM_3D)
     WARPX_ALWAYS_ASSERT_WITH_MESSAGE(fileGeom == "cartesian", "3D can only read from files with cartesian geometry");
-    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(axisLabels.at(0) == "x" && axisLabels.at(1) == "y" && axisLabels.at(2) == "z",
-                                     "3D expects axisLabels {x, y, z}");
+    if (axisLabels.at(0) == "x" && axisLabels.at(1) == "y" && axisLabels.at(2) == "z") {
+        xyz_order = true;
```
ax3l, line 228: The logic to add here would be:
```suggestion
        xyz_order = dataOrder == "C"? true : false;
```

Source/Initialization/ExternalField.cpp
```diff
@@ -189,99 +190,368 @@ ExternalFieldParams::ExternalFieldParams(const amrex::ParmParse& pp_warpx)
...
+    } else if (axisLabels.at(2) == "x" && axisLabels.at(1) == "y" && axisLabels.at(1) == "z") {
+        xyz_order = false;
```
ax3l, line 230: The logic to add here would be:
```suggestion
        xyz_order = dataOrder == "C"? false : true;
```

ax3l: and same for the ones below.

One can simplify the logic by reading and conditionally inverting the `axisLabels`, `gridSpacing` and `gridGlobalOffset` when reading from openPMD (based on `dataOrder`) and then keep the logic here in place as is (in dataOrder C).

Source/Initialization/ExternalField.cpp
WeiqunZhang: No, what I mean by xyz_order is the order of the axis labels, which has nothing to do with C or F order. In other words, is axis-0 x or z? Naturally, one might expect the order is x, followed by y and then by z. But WarpX's openPMD output's axis labels are
```
axisLabels{"x", "y", "z"};
return {axisLabels.rbegin(), axisLabels.rend()};
```
So the data in WarpX as amrex data container are treated as Fortran Array `a(x,y,z)`. Because Fortran array is column major, x is the unit stride direction. But the data in openPMD output of WarpX, they are C Arrays like `a[z][y][x]`, where x is still the unit stride direction, but axis 0 is z.

There are four possibilities with the combination of axis label order and array's row or major order. As for whether we need to transpose data, there are only two possibilities, yes or no. Because there is only yes or no regarding data transpose, supporting all 4 combinations does not add much complexities.

Source/Initialization/ExternalField.cpp
WeiqunZhang: I don't think this is correct. See my comments above.

Source/Initialization/ExternalField.cpp
WeiqunZhang: See my comments above.

Source/Fluids/WarpXFluidContainer.cpp
WeiqunZhang: done

Source/Initialization/ExternalField.cpp
```diff
@@ -189,99 +189,353 @@
...
+    }
+
+    amrex::Abort("xxxxx prepare(RealBox) todo");
+
+    // If it's not distributed, we will not need to do anything.
+    // The constructor's load_data should try to load everything!
+
+    if (!m_moving_window) {
+        // clear the data?
+    }
```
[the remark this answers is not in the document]
WeiqunZhang: Already addressed

WeiqunZhang:
I think this PR is ready now.

Source/Fluids/WarpXFluidContainer.cpp
```diff
@@ -173,21 +178,36 @@ void WarpXFluidContainer::InitData(
...
     const amrex::Real clight = PhysConst::c;
 
+    // Create local copies of pointers for GPU kernels
+    InjectorMomentum* inj_mom = d_inj_mom;
+
+    if (h_inj_rho && h_inj_rho->needPreparation()) {
+        auto get_zlab = [=] (amrex::Real z) -> amrex::Real
+        {
+            return gamma_boost*(z + beta_boost*clight*cur_time);
+        };
```
ax3l, line 193: Not needed for this PR, but this might need a more globally shared helper function?

Source/Initialization/ExternalField.H
```diff
@@ -144,24 +157,58 @@ struct ExternalFieldView
...
 #endif
+    amrex::IntVect global_size;
 };
 
 class ExternalFieldReader
 {
 public:
-    ExternalFieldReader (std::string const& read_fields_from_path,
-                         std::string const& F_name, std::string const& F_component);
+    ExternalFieldReader (std::string read_fields_from_path,
```
ax3l, line 196: Can you please add a doxygen block for the parameters here?
especially poblo and pdf and dombox are things that I would find myself searching for in the function body.

Source/Initialization/ExternalField.H
```diff
@@ -144,24 +157,58 @@ struct ExternalFieldView
...
+                         amrex::GpuArray<amrex::Real,AMREX_SPACEDIM> const& problo,
+                         amrex::GpuArray<amrex::Real,AMREX_SPACEDIM> const& pdx,
+                         amrex::Box const& dombox, bool distributed);
 
-    ExternalFieldView const& getView () const { return m_external_field_view; }
+    [[nodiscard]] ExternalFieldView getView (int li) const;
+
+    [[nodiscard]] ExternalFieldView getView () const;
+
+    void prepare (amrex::BoxArray const& grids,
```
ax3l, line 218: Can you please add a doxygen block for the parameters here?

Source/Initialization/ExternalField.H
```diff
@@ -144,24 +157,58 @@ struct ExternalFieldView
...
+                  amrex::DistributionMapping const& dmap,
+                  amrex::IntVect const& ngrow,
+                  std::function<amrex::Real(amrex::Real)> const& get_zlab = nullptr);
+
+    void prepare (amrex::RealBox const& pbox, int moving_dir, int moving_sign,
```
ax3l, line 231: Can you please add a doxygen block for the parameters here?

Source/Initialization/ExternalField.H
```diff
@@ -144,24 +157,58 @@ struct ExternalFieldView
...
+                  std::function<amrex::Real(amrex::Real)> const& get_zlab = nullptr);
+
+    [[nodiscard]] bool distributed () const { return m_distributed; }
 
 private:
+
+    void load_data (amrex::RealBox const& pbox);
```
ax3l, line 240: Can you please add a doxygen block for the parameters here? In which reference frame is pbox? lab frame or simulation frame?

Source/Initialization/ExternalField.H
```diff
@@ -144,24 +157,58 @@ struct ExternalFieldView
...
+    [[nodiscard]] ExternalFieldView make_view (amrex::BaseFab<double> const& fab) const;
+    void make_cache_box (amrex::RealBox const& pbox, int moving_dir, int moving_sign);
```
ax3l, line 245: Probably good to add parameter descriptions here, too.

Source/Initialization/ExternalField.H
```diff
@@ -144,24 +157,58 @@ struct ExternalFieldView
...
+
+    std::string m_file;
+    std::string m_name;
+    std::string m_component;
+    amrex::GpuArray<amrex::Real,AMREX_SPACEDIM> m_problo; // WarpX's initial prob lo
+    amrex::GpuArray<amrex::Real,AMREX_SPACEDIM> m_probdx; // WarpX's cell size
+    amrex::Box m_dombox;                                  // WarpX's initial domain
```
ax3l: This makes them doxygen:
```suggestion
    amrex::GpuArray<amrex::Real,AMREX_SPACEDIM> m_problo; //! WarpX's initial prob lo
    amrex::GpuArray<amrex::Real,AMREX_SPACEDIM> m_probdx; //! WarpX's cell size
    amrex::Box m_dombox;                                  //! WarpX's initial domain
```

Source/Initialization/ExternalField.H
```diff
@@ -144,24 +157,58 @@ struct ExternalFieldView
...
+    bool m_distributed = false;
+    bool m_moving_window = false;
+    amrex::RealVect m_dx;     // Field data's dx
+    amrex::RealVect m_offset; // Field data's physical location at (0,0,0)
+    amrex::IntVect m_size;    // Field data's size (i.e., number of elements)
+    amrex::RealBox m_domain;  // Field data's total physical domain
+    amrex::RealBox m_cache_domain; // Current cache's domain
```
ax3l, line 259: Please use `//! ...` for doxyen and also describe the last three member variables.

Source/Initialization/ExternalField.cpp
ax3l: Ok. I think this works for now, but I still think we need to check for `dataOrder`. If `dataOrder` in the file we read is not C, then we need to tread `axisLabels` and all other meta data as the exact inverse order to what the logic does now.
For that, I see below that `c_order` is actually applied all the time, it might be complete already.

I can generate an example file for this after the break, next year. This PR can go in before that.

ax3l (approved):
This looks great to me, thank you for the hard work!

The `ExternalField` interface class could be better documented, otherwise ready to go from my end.

Please merge at your own convenience after the doc update.

Source/Initialization/ExternalField.cpp
WeiqunZhang: Yes, we already have
```
  bool c_order = F.getAttribute("dataOrder").get<std::string>() == "C";
  ...
  if ((xyz_order && c_order) || (!xyz_order && !c_order)) {
      // transpose
  }
```

WeiqunZhang:
I have added doxygen documentation for all members of `ExternalFieldView` and `ExternalFieldReader`.

PR #6223, opened 2025-09-28
Fix potential recursive calls of WarpX constructor.

### LINKED ISSUE #6222: Recursive calls in creating WarpX
There seems to be an issue with how a WarpX object is created. In main.cpp, we call
```
auto& warpx = WarpX::GetInstance();
```
which in turn calls
```
    if (!m_instance) {
        MakeWarpX();
    }
```
which in turn calls
```
m_instance = new WarpX();
```

The issue is the constructor of WarpX contains
```
    if (do_fluid_species) {
        myfl = std::make_unique<MultiFluidContainer>(); 
    }
```
which in turn calls
```
std::make_unique<WarpXFluidContainer>(i, species_names[i]);
```
which in turn calls
```
    WarpX &warpx = WarpX::GetInstance();
```
`m_instance` at this time is still a nullptr. So we have an infinite loop.

I don't understand why this has not become an issue. Nevertheless, while I was working on density loading, a fluid container CI failed with recursive calls to WarpX:GetInstance.

Note that for particle containers, we pass `this` (a pointer to WarpX) to their constructor. We could fix the fluid container by passing `this` as well. But there might be other objects in WarpX that have a similar issue. Note that we used to have `m_instance = this;` as the first line of the WarpX constructor. It got deleted during refactoring. I think we should put it back.

WeiqunZhang:
WarpX's constructor may build objects that call WarpX::getInstance, which in turn build WarpX again if `m_instance` is nullptr. So we should assign `this` to `m_instance` at the beginning of the constructor to avoid the recursion.

### DIFF
--- a/Source/WarpX.cpp
+++ b/Source/WarpX.cpp
@@ -319,6 +319,8 @@ WarpX::Finalize()
 
 WarpX::WarpX ()
 {
+    m_instance = this;
+
     warpx::initialization::initialize_warning_manager();
 
     ReadParameters();
--- a/Source/WarpX.cpp
+++ b/Source/WarpX.cpp
@@ -319,7 +319,8 @@ WarpX::Finalize()
 
 WarpX::WarpX ()
 {
-    m_instance = this;
+    m_instance = this; // This guarantees that GetInstance() can be
+                       // indirectly used in WarpX constructor.
 
     warpx::initialization::initialize_warning_manager();
 


AlexanderSinn:
Note that with this, the issue could still happen if any of the subjects of WarpX that are not wrapped in a unique pointer call getInstance in their constructor, which executes before the WarpX constructor as part of the initializer list. I think in general it is better if it is not possible for getInstance to create a new WarpX instance.

lucafedeli88:
So, @AlexanderSinn , what you propose is providing a method like :

```
WarpX::Init( /* params */ );
```
which would be the only one able to create an instance and which would be called in `main.cpp`
and making sure that `GetInstance` fails if the instance does not yet exist ?

In general, I think that we should completely avoid `GetInstance()` as a mechanism to read or (arguably much worse!) write parameters. However, this goal will require a substantial amount of time. Meanwhile, a solution like what you propose seems reasonable.

AlexanderSinn:
Yes

lucafedeli88:
@WeiqunZhang , what do you think about the solution proposed by @AlexanderSinn  ?

WeiqunZhang:
I think it makes the run time error easier to understand, but it does not solve the issue if there is such a need to call member initializer. If we want to allow that, we could perhaps do
```
class WarpX {
    WarpX () : m_first_member(set_instance(this)) {}
    static bool set_instance (WarpX* p) { m_instance = p; return true; }
    bool m_first_member = false;
    FOO m_another_member; // FOO::FOO () { // call WarpX::getInstance() ...}
    ...
};
```

lucafedeli88:
Maybe we could add this to the next developers' meeting agenda.  I think that we should discourage the use of `WarpX::GetInstance()` and gradually remove its occurrences from the code, so I am leaning towards what @AlexanderSinn proposed, but I think that it's worth discussing it.

WeiqunZhang:
We have the following options.

1. One line change in this PR. We can then decide what to do in the future.
2. @AlexanderSinn's suggestion. I don't actually know what exactly is in Alex's mind. So either Alex or @lucafedeli88 needs to implement it. My understanding is it can print out an error message, but cannot avoid the error.
3. Adding a special first member variable to WarpX as I suggested in comments. This should be able to avoid runtime error. As for getting rid of GetInstance completely, I am neutral. There are pros and cons for both sides.
4. Getting rid of GetInstance. https://github.com/BLAST-WarpX/warpx/pull/6247

AlexanderSinn:
I was thinking of replacing the MakeWarpX call in GetInstance with a runtime assert, updating main() accordingly, and also setting m_instance earlier like Weiqun suggested. However, instead of using a member variable, make a special object from which WarpX inherits so that it is initialized even before AmrCore. See
https://github.com/Hi-PACE/hipace/pull/592/files
https://github.com/Hi-PACE/hipace/pull/595/files#diff-8625c4ae58ada08195644f43edcdcd7793377f08f6bd51f34579ee3dfb7c7a4f
https://github.com/Hi-PACE/hipace/pull/595/files#diff-d6da8eab3f297020476fada6e9220e320306284f9c256ca521887ccfb547f137
But removing GetInstance is also good.

ax3l:
Thanks for digging into this!

I think to remove `GetInstance` while WarpX is essentially still used as a singleton will not make the situation better, e.g., we will even store potentially dangling pointers and pass around raw pointers as in #6247 proposed. It does not help with compile-time nor make the usage safer, does it? I agree generally though, that WarpX should not be a singleton class, but a regular variable, to remove globals and potentially enable that we can have multiple WarpX and ImpactX and HiPACE++ simulations alive at the same time in the same process (i.e. Python workflows).

I think merging this PR and adding more asserts for wrong usage that Alex has in mind might be the a good first step.

ax3l (approved):
Great first step! :)

Let's continue next week in the dev meeting :)

PR #6231, opened 2025-09-30
Fix position update for flux of photons

EZoni:
### Background

While investigating a failure in the test `test_3d_linear_compton_bunch_laser` in #5955, we realized that the position update implemented in `UpdatePosition` was not correct for massless particles like photons. This was called from `PhysicalParticleContainer::AddPlasmaFlux` for the flux injection of photons.

@RemiLehe suggested that we correct the implementation in `UpdatePosition` and remove the photon counterpart `UpdatePositionPhoton`, which was only called from `PhotonParticleContainer::PushPX`.

### Changes

Take into account the mass of the particle in the position update equation:
- For massive particles: $\Delta x = \frac{u_x}{\gamma} \Delta t$.
- For massless particles: $\Delta x = c \frac{u_x}{|u|} \Delta t$.

Note that for now I kept the equations as they were. Alternatively, we could also write the equations as $\Delta x = c^2 \frac{p_x}{E} \Delta t$, with $E^2 = p^2 c^2 + m^2 c^4$, with $m=0$ but $p$ computed using the electron mass for photons.

### Validation

This is a snapshot of one photon species in the (x,z) plane from the test `test_3d_linear_compton_bunch_laser`, `development` branch on the left vs. this branch on the right:

<p align="center">
    
    
</p>

### To-do

- [x] Merge #6232 and rebase this branch.
- [x] Merge #6235 and rebase this branch.


Source/Particles/Pusher/UpdatePosition.H
```diff
@@ -22,29 +22,54 @@
...
-                    [[maybe_unused]] amrex::ParticleReal& y,
-                    [[maybe_unused]] amrex::ParticleReal& z,
-                    const amrex::ParticleReal ux, const amrex::ParticleReal uy, const amrex::ParticleReal uz,
-                    const amrex::Real dt )
+void UpdatePosition ([[maybe_unused]] amrex::ParticleReal& x,
+                     [[maybe_unused]] amrex::ParticleReal& y,
+                     [[maybe_unused]] amrex::ParticleReal& z,
+                     const amrex::ParticleReal ux, const amrex::ParticleReal uy, const amrex::ParticleReal uz,
+                     const amrex::Real dt,
+                     amrex::ParticleReal const mass)
```
EZoni, line 29: I could also try to make `mass` a template parameter and use `if constexpr` instead of simple `if` conditions. Let me know what you think.

EZoni:
The test `test_3d_photon_pusher` fails badly:
```
max(disc_pos) = 0.29289321881345365
tol_pos = 1e-14
max(disc_mom) = 1.7217529609352883e-16
tol_mom = 2.220446049250313e-16
```

The test is supposed to be checking if "photons initialized with different momenta and different initial directions propagate along straight lines at the speed of light".

However, I think the test fails due to the fact that in the input file we set photons with electron charge and mass (???), e.g.
https://github.com/BLAST-WarpX/warpx/blob/c13beb28cc6dcce000135ac6175cbd99d02335dd/Examples/Tests/photon_pusher/inputs_test_3d_photon_pusher#L25-L28

So our check on whether the mass is zero or not probably goes into the wrong branch of the code.

@RemiLehe @aeriforme I don't understand why we have that setup in the input file. Do you know?

RemiLehe (approved):

Source/Particles/Pusher/UpdatePosition.H
EZoni, line 1: I recommend reviewing the changes in this file in the "hide whitespace" mode.

PR #6232, opened 2025-10-01
Rename WarpXParticleContainer's member variable mass to m_mass

EZoni:
While working on the bug fix in #6231 I ran into the warning/error

> warning: capture host side class data member by this pointer in device or host device lambda function may result in invalid memory access if this pointer is not accessible on device side [-Wgpu-maybe-wrong-side]

which I had missed due to the fact that the WarpXParticleContainer's member variable `mass` is not prefixed with `m_`, as WarpX's naming conventions require (see https://warpx.readthedocs.io/en/latest/developers/contributing.html#style-and-conventions).

This PR implements the following changes:
- The WarpXParticleContainer's member variable is named `m_mass`.
- Any local copy for device lambda capture is named `mass` (rather than `m`, `ms`, `mass_local`, etc.).

I implement this changes separately in this PR to keep the bug fix in #6231 as minimal as possible.

### DIFF
--- a/Source/Particles/WarpXParticleContainer.H
+++ b/Source/Particles/WarpXParticleContainer.H
@@ -528,9 +528,8 @@ public:
         return m_do_back_transformed_particles;
     }
 
-    //amrex::Real getCharge () {return charge;}
     amrex::ParticleReal getCharge () const {return charge;}
-    //amrex::Real getMass () {return mass;}
+
     amrex::ParticleReal getMass () const {return m_mass;}
 
     int DoFieldIonization() const { return do_field_ionization; }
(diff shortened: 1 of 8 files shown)


Source/Particles/WarpXParticleContainer.H
```diff
@@ -531,7 +531,7 @@ public:
     //amrex::Real getCharge () {return charge;}
     amrex::ParticleReal getCharge () const {return charge;}
     //amrex::Real getMass () {return mass;}
```
lucafedeli88:
```suggestion
    //amrex::Real getMass () {return m_mass;}
```
I think that we should do that for consistency. However, I also think that this line should be removed! Is there a reason to keep this commented out method ?

lucafedeli88 (approved):
Thanks for this PR, @EZoni

Source/Particles/WarpXParticleContainer.H
EZoni: Thanks Luca, I had not noticed that. I removed the two methods that were commented out, `getMass` and `getCharge`. I don't see any reason why they should be kept there as is, especially given that there is no comment explaining what they are there for.

PR #6236, opened 2025-10-01
Add error message for non-supported scattering processes in DSMC

roelof-groenewald:
Excitation and forward scattering is not yet supported in DSMC but currently these processes do not result in sensible error messages.

Addresses https://github.com/BLAST-WarpX/warpx/discussions/6229.

### DIFF
--- a/Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
+++ b/Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
@@ -53,6 +53,10 @@ DSMCFunc::DSMCFunc (
 
         ScatteringProcess process(scattering_process, cross_section_file, energy);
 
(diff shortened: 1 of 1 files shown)


RemiLehe (approved):

RemiLehe:
Thanks for this PR!
For consistency, should we remove this line:
https://github.com/BLAST-WarpX/warpx/blob/development/Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H#L259

roelof-groenewald:
> For consistency, should we remove this line: https://github.com/BLAST-WarpX/warpx/blob/development/Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H#L259

I think we should keep it as it gives a nice place-holder for where forward scattering should be handled once it is implemented.

PR #6237, opened 2025-10-02
WarpX class: remove unnecessary getPMLRZ() method from WarpX class

lucafedeli88:
The `bool WarpX::getPMLRZ()` method is not really necessary, since it is a simple single-line function used only once from within the WarpX class (in `WarpXMovingWindow.cpp` ) . I think that we can remove it to simplify the WarpX class.


dpgrote (approved):
Looks good to me!

PR #6241, opened 2025-10-02
Update aurora.rst fixing a mistakenly pasted path from Perlmutter

Noerr:
Updated source script path on Aurora to be consistent with preceding Aurora instructions.  The errant reference to ${CFS} is clearly a duplication issue stemming from the Perlmutter instructions.


roelof-groenewald (approved):
Thanks for the fix!

PR #6242, opened 2025-10-03
Only allow Gaussian beam rotation in 3D and 2D XZ

aeriforme:
Do not rotate a Gaussian beam in geometries that are not 3D or 2D XZ.

Addresses [this comment](https://github.com/BLAST-WarpX/warpx/pull/4767#discussion_r2395494195).


EZoni (approved):
Thanks!

PR #6247, opened 2025-10-04
[WIP] Remove WarpX::GetInstance

WeiqunZhang:
We now pass a `WarpX*` to where we previously needed `WarpX::GetInstance()`. For classes, this means storing a pointer to `WarpX` as a member variable either in the class itself or in its base class. We use `WarpX*` rather than `WarpX const*` as the data type to avoid frequent use of `const_cast` in practice.


WeiqunZhang:
For backward compatibility on the python side, I didn't go all the way in by removing GetInstance completely. I don't have any opinions on how to handle python.

WeiqunZhang:
https://github.com/BLAST-WarpX/warpx/pull/6223#issuecomment-3379227836

I don't think passing around raw pointers in the PR is unsafe. The objects storing the raw pointer of WarpX are part of WarpX. If WarpX is gone, they are gone also. If we are willing to break picmi (e.g., https://github.com/BLAST-WarpX/warpx/blob/ba5ade8b23613d216c75ae18e8fdca46225eccbd/Python/pywarpx/_libwarpx.py#L145), then there will be no singleton. I think the new pywarpx stuff using pybind11 is fine, because we can initialize warpx directly in the main python function. The C++ part is almost there. It can be done with a few lines of code.

WeiqunZhang:
For example, I just removed the singleton completely in C++ and this is now the main.cpp
https://github.com/WeiqunZhang/WarpX/blob/45f54458c9a3633682b94ebb978057cb19abcd72/Source/main.cpp#L30

If we remove some static variables from WarpX, we can easily have multiple WarpX objects.

WeiqunZhang:
Will resolve the conflicts or future conflicts if necessary.

EZoni:
@WeiqunZhang 

My review is requested for this PR, but I forgot what conclusion we had reached on whether we want to go forward with the changes proposed here or not. Can you remind me?

WeiqunZhang:
Although I opened this PR, I actually like WarpX::GetInstance. My view is there is nothing wrong with WarpX::GetInstance given how WarpX is written and we should keep using it. I even like goto, so singleton is certainly acceptable to me. I submitted this because I think that's how we could do it if we wanted to get rid of the warpx singleton.

EZoni:
Thanks, @WeiqunZhang. Okay, given the context, I will label the PR as "[WIP]" for now.

PR #6250, opened 2025-10-07
DSMC: implement generic two-product reaction

oshapoval:
**Description edited by @RemiLehe**

# Overview

This replaces the DSMC `CHARGE_EXCHANGE` process by a `TWO_PRODUCT_REACTION` process, whereby:
- The reaction can have an associated energy cost (or gain) - while `CHARGE_EXCHANGE` assumed no energy cost/gain.
- The reactant and products can have any mass/charge - while `CHARGE_EXCHANGE` implicitly assumed that reactants and products had the same masses, and imposed that their charge is incremented/decremented by one during the reaction.

This change is backward compatible for users ; input scripts using `charge_exchange` will work, but will internally call the new `TWO_PRODUCT_REACTION` (with 0 energy cost/gain)

# Implementation

This leverages the function `TwoProductComputeProductMomenta` which is also used in the fusion code. This function has been modified to add an option to not scatter the products isotropically (which is used in DSMC)

As a result the calculation is valid for relativistic particles too.

# CI test

We added a test that is very similar to `test_1d_charge_exchange_dsmc` (i.e. a beam of H+ goes through an H2 gas background, leading to the charge exchange reaction `H+ + H2 -> H + H2+`).

However, while the beam energy in `test_1d_charge_exchange_dsmc` is 12 keV, the beam energy in the new test is only 5 eV. This means that the energy consumed by the charge exchange reaction (1.66 eV in this case) is not negligible, therefore justifying the need to use `twoproduct_reaction` here. The new test checks the final energy of the H and H2+ products, and the fact that it is consistent with the 1.66 eV consumed by the reaction.


Docs/source/usage/parameters.rst
```diff
@@ -2264,7 +2264,7 @@ Details about the collision models can be found in the :ref:`theory section <mul
 
 * ``<collision_name>.scattering_processes`` (`strings` separated by spaces)
     Only for ``dsmc`` and ``background_mcc``. The scattering processes that should be
-    included. Available options are ``elastic``, ``excitationX``, ``forward``, ``back``, and ``charge_exchange``
+    included. Available options are ``elastic``, ``excitationX``, ``forward``, ``back``, ``twoproduct_reaction`` and ``charge_exchange``
```
RemiLehe, line 2263: I couldn't find a better way to document this, within the current document format.
@roelof-groenewald Let me know if you have better suggestions.

RemiLehe: In particular, some detailed are left out:
- clarifying that `charge_exchange` is the same as `twoproduct_reaction`, but with 0 energy cost
- clarifying that `twoproduct_reaction` can represent any reaction, and has 0 scattering angle (the velocity of the first and second product are aligned with the velocity vector of the first and second incident particle respectively, in the center of mass frame)

Examples/Tests/ionization_dsmc/analysis_two_product_reaction_dsmc_1d.py
```diff
@@ -0,0 +1,57 @@
...
+    m_H2 / (m_H + m_H2)
+) * E0  # [eV] Total kinetic energy in the center-of-mass frame (before the reaction)
+# (e.i., E_com_before  = ( (m_H * m_H2 / (m_H + m_H2) ) * (v0-v_H2)**2 )/2)
+E_com_after = (
+    E_com_before + Q
+)  # [eV] Total kinetic energy in the center-of-mass frame (after the reaction)
+
+# Compute celocity of H in COM frame
+vH_com = np.sqrt(
+    1 / 2 * (E_com_after * eV_to_J) / (m_H * (1 + m_H / m_H2))
```
RemiLehe:
```suggestion
    2 * (E_com_after * eV_to_J) / (m_H * (1 + m_H / m_H2))
```

Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
```diff
@@ -70,7 +71,7 @@ DSMCFunc::DSMCFunc (
             reaction_produces_new_species = true;
         }
 
-        if (process.type() == ScatteringProcessType::IONIZATION) {
+        if (scattering_process.find("ionization") != std::string::npos) {
```
roelof-groenewald, line 74: Why is this change necessary? It seems to me that `process.type() == ScatteringProcessType::IONIZATION` is easier to understand.

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.cpp
```diff
@@ -40,17 +40,25 @@ SplitAndScatterFunc::SplitAndScatterFunc (const std::string& collision_name,
...
                 m_num_products_host.push_back(0); // the colliding species are consumed in the reaction
                 m_num_products_host.push_back(1); // first product species
                 m_num_products_host.push_back(1); // second product species
+
+                // get the reaction energy
+                if (std::find(scattering_processes.begin(), scattering_processes.end(), "two_product_reaction") != scattering_processes.end()) {
+                    pp_collision_name.get("two_product_reaction_energy", m_reaction_energy);
+                } else {
+                    m_reaction_energy = 0; // assume zero energy cost for charge exchange
+                }
```
roelof-groenewald: This could be simplified to the following since `m_reaction_energy` is initialized as `0`:
```suggestion
                // get the reaction energy, assuming zero energy for charge exchange
                pp_collision_name.query("two_product_reaction_energy", m_reaction_energy);
```

Source/Particles/Collision/BinaryCollision/NuclearFusion/ProtonBoronFusionInitializeMomentum.H
```diff
@@ -104,7 +104,9 @@ namespace {
             soa_2.m_rdata[PIdx::uz][idx_2], m2,
             ux_alpha1, uy_alpha1, uz_alpha1, static_cast<amrex::ParticleReal>(m_alpha),
             ux_Be, uy_Be, uz_Be, static_cast<amrex::ParticleReal>(m_beryllium),
-            E_fusion, engine);
+            E_fusion,
+            true, // isotropic scattering
```
roelof-groenewald: I like the way @EZoni handles the keywords. He does
```suggestion
            /*isotropic scattering=*/true,
```

Maybe we should agree on a WarpX standard format at some point.

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
```diff
@@ -279,21 +281,45 @@ public:
...
+                const auto reactant2_index = static_cast<int>(p_pair_indices_2[i]);
                 const auto product2_index = products_np_data[3] + with_product_p_offsets[i];
-                copy_species2[3](soa_products_data[3], soa_2, static_cast<int>(p_pair_indices_2[i]),
+                copy_species2[3](soa_products_data[3], soa_2, reactant2_index,
                                 static_cast<int>(product2_index), engine);
                 // Set the weight of the new particle to p_pair_reaction_weight[i]
                 soa_products_data[3].m_rdata[PIdx::w][product2_index] = p_pair_reaction_weight[i];
+
+                // Update the momenta of the products
+                TwoProductComputeProductMomenta(
```
roelof-groenewald, line 303: This function involves quite a few FLOPS, which are unnecessary if `reaction_energy == 0`. Could we wrap this in an `if` statement to only call `TwoProductComputeProductMomenta` if `reaction_energy != 0`? Or alternatively, could we return early in `TwoProductComputeProductMomenta` if `reaction_energy == 0`?

roelof-groenewald (commented):
Looks good to me. I just have one concern about unnecessary new calculations that could slow down the charge exchange collisions - see comments.

Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
RemiLehe: This was in order to be consistent with the `if` condition below:
```
else if (scattering_process.find("charge_exchange") != std::string::npos) 
```
(that one cannot `if (process.type() == ScatteringProcessType::CHARGE_EXCHANGE)` since `ScatteringProcessType::CHARGE_EXCHANGE` does not exist anymore)

Let me know if you think it is preferable to keep this consistency, or if we should instead revert to `ScatteringProcessType::IONIZATION`

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
RemiLehe: Good point!

I think that, in addition to treating the case `reaction_energy=0`, this function has also two additional benefits:

- It correctly treats the case `m1_in != m1_out` / `m2_in != m2_out`. (This was previously not done correctly, since the previous code simply copies `ux` from the incident species to the product species.) That being said, I agree it would be possible to treat this case correctly without calling this large function.
- It makes the function correct for the relativistic case, and has a natural extension for photons (https://github.com/BLAST-WarpX/warpx/pull/6214)

Maybe we can do some performance test to decide how to proceed.

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.cpp
RemiLehe: Agreed. Thanks for the suggestion!

Source/Particles/Collision/BinaryCollision/NuclearFusion/ProtonBoronFusionInitializeMomentum.H
RemiLehe: Thanks, I agree!

Source/Particles/Collision/BinaryCollision/DSMC/DSMCFunc.cpp
roelof-groenewald: Oh I see. That makes sense, and I think it is better then to keep the changes as is for consistency sake. Thanks for clarifying!

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
roelof-groenewald: Good points! I agree doing some performance comparisons for the case with `reaction_energy=0` and equal masses would be nice, since if we can convince ourselves that the extra FLOPS are not affecting performance much it would be nice to keep the changes as they are.

Source/Particles/Collision/BinaryCollision/NuclearFusion/ProtonBoronFusionInitializeMomentum.H
EZoni: Thanks for pointing this out, @roelof-groenewald. 

My idea would be to use the exact input argument name in the comment, rather than a comment itself. 

So if I were to suggest a convention, I would suggest in this case
```
/*isotropic_scattering=*/true,
```
since `isotropic_scattering` (with the `_`) is the exact input argument name. 

I think this can make searching easier, for example if one searches `isotropic_scattering`. 

It also mimics the Python keyword syntax, `f(x=1.0, y=2.0)`, where `x` and `y` match the input argument names.

Source/Particles/Collision/BinaryCollision/NuclearFusion/ProtonBoronFusionInitializeMomentum.H
```diff
@@ -104,7 +104,9 @@ namespace {
             soa_2.m_rdata[PIdx::uz][idx_2], m2,
             ux_alpha1, uy_alpha1, uz_alpha1, static_cast<amrex::ParticleReal>(m_alpha),
             ux_Be, uy_Be, uz_Be, static_cast<amrex::ParticleReal>(m_beryllium),
-            E_fusion, engine);
+            E_fusion,
+            /*isotropic scattering=*/true,
```
EZoni: Based on https://github.com/BLAST-WarpX/warpx/pull/6250#discussion_r2426857062:
```suggestion
            /*isotropic_scattering=*/true,
```

Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
oshapoval: Hi @roelof-groenewald.  Thank you for your detailed review. Here are the numbers from the performance comparison: 

- using `dev` branch:
```
Level 0: dt = 4.12209591e-09 ; dz = 0.00625
Grids Summary:
  Level 0   1 grids  32 cells  100 % of domain

ParticleContainer::RedistributeCPU()                                                    904     0.3584     0.3584     0.3584  30.89%
PhysicalParticleContainer::Evolve::GatherAndPush                                        792     0.2942     0.2942     0.2942  25.36%
WarpXParticleContainer::ApplyBoundaryConditions()                                       400     0.1136     0.1136     0.1136   9.79%
BinaryCollision::doCollisionsWithinTile::LoopOverCollisions                             100    0.09924    0.09924    0.09924   8.55%
PhysicalParticleContainer::Evolve()                                                     800    0.08362    0.08362    0.08362   7.21%
```
- using `feature` branch:
```
Level 0: dt = 4.12209591e-09 ; dz = 0.00625
Grids Summary:
  Level 0   1 grids  32 cells  100 % of domain

PhysicalParticleContainer::Evolve::GatherAndPush                                        397      0.245      0.245      0.245  27.29%
ParticleContainer::RedistributeCPU()                                                    504     0.1905     0.1905     0.1905  21.23%
WarpXParticleContainer::ApplyBoundaryConditions()                                       400      0.114      0.114      0.114  12.70%
BinaryCollision::doCollisionsWithinTile::LoopOverCollisions                             100     0.0867     0.0867     0.0867   9.66%
PhysicalParticleContainer::Evolve()                                                     400    0.04064    0.04064    0.04064   4.53%
```

roelof-groenewald: Thanks for checking that @oshapoval!

Source/Particles/Collision/BinaryCollision/NuclearFusion/TwoProductFusionInitializeMomentum.H
```diff
@@ -77,6 +77,7 @@ namespace {
             ux1_out, uy1_out, uz1_out, m1_out,
             ux2_out, uy2_out, uz2_out, m2_out,
             E_fusion,
+            true, // isotropic scattering
```
roelof-groenewald:
```suggestion
            /*isotropic_scattering=*/true,
```

roelof-groenewald (approved):

PR #6253, opened 2025-10-08
Rework the PECInsulator boundary conditions

dpgrote:
This reworks the PECInsulator boundary conditions to be more correct, specifically to fix conservation of energy. In the simplest form, the conditions for both PEC and insulator (on the upper x axis at L) are
```
E||(x > L) = 2*E||(L) - E||(2*L - x)
Ep(x > L) = Ep(2*L - x)
B||(x > L) = B||(2*L - x)
Bp(x > L) = 2*Bp(L) - Bp(2*L - x)
```
Where `||` and `p` are parallel and perpendicular to the boundary. However, the actual implementation is not so simple. For instance, `E||(L)` and `Bp(L)` are zero for PEC, but not for insulator. Also, depending on whether `E` or `B` is specified in the insulator, either `E||(L)` or `B||(L)` will be specified or be evolved (and so unchanged here) which means that in the extrapolations either the evolved value or the specified values needs to be used.


Source/BoundaryConditions/WarpX_PEC.cpp
```diff
@@ -403,7 +407,19 @@ namespace
...
+                } else {
+                    // For J-parallel in PECInsulator boundaries, sum the values in the guard
+                    // cells and add it to the J in the boundary (noting that J-parallel is nodal).
+                    int const isign = (iside == 0 ? -1 : +1);
+                    amrex::IntVect ijk_guard = ijk_vec;
+                    for (int ig = 0 ; ig < nguards ; ig++) {
+                        ijk_guard[idim] += isign;
+                        field(ijk_vec,n) += 2._rt*field(ijk_guard,n);
+                    }
+                }
```
JustinRayAngus: I'm having trouble following the logic here.

1) There are three logic checks going on for the first if condition. What situations does it not go here?
2) Should there be an rscale factor somewhere in the else {} condition?

Can you walk me through what happens here for the insulator part of the PEC_insulator? I'm trying to see that the folding in of the current is conservative. For each guard cell you subtract the value from the mirror location in the domain and then add twice the value to the boundary.  I only see adding twice the value to the mirror location inside the else{} statement.

JustinRayAngus: Would this logic work?
```              
if (ijk_vec == ijk_mirror && sum_on_boundary[iside] && psign[iside] < 0._rt) {
    // For J-parallel in PECInsulator boundaries, sum the values in the guard
    // cells and add it to the J in the boundary (noting that J-parallel is nodal).
    int const isign = (iside == 0 ? -1 : +1);
    amrex::IntVect ijk_guard = ijk_vec;
    for (int ig = 0 ; ig < nguards ; ig++) {
         ijk_guard[idim] += isign;        
         field(ijk_vec,n) += 2._rt*field(ijk_guard,n);
    } 
} else {
    field(ijk_vec,n) += rscale * psign[iside] * field(ijk_mirror,n);
}
```

JustinRayAngus: Discussed offline. There is only one specific set of situations where the sum is performed, and so the suggestion is to use an if condition with the specific combination of logical checks using && rather than than checking for not meeting each condition and using ||

JustinRayAngus (approved):
Thanks Dave. Looks good to me.

PR #6264, opened 2025-10-14
Docs: update LXPLUS (CERN) documentation

SeverinDiederichs:
This PR fixes the LXPLUS installation guide. While there is hardly any software installed on LXPLUS itself, the software is instead provided via the LCG software stack. 

Thus, instead of relying on Spack, an appropriate LCG view should be loaded, which provides all dependencies for WarpX. As a minor caveat, one must force picking the parallel HDF5 version over the serial one, as both are available in the LCG view. 

This drastically simplifies the build procedure on LXPLUS.


RemiLehe (approved):
Thanks a lot for updating the instructions 🚀 ⭐ ✨

PR #6265, opened 2025-10-14
Deprecate photon species list

lucafedeli88:
This PR deprecates `particles.photon_species`, since it is redundant (we can use species type to define a photon species).  
Specifically:
-  `particles.photon_species` is removed from the documentation
- `particles.photon_species` is removed from all the examples (in all the examples we were already using `species_type`)
- the new logic in `MultiParticleContainer.cpp` is that a photon species is initialized as such if `species_type` is set to `photon` or if it listed in the `particles.photon_species`
- If `particles.photon_species` is used a warning is generated stating that `photon_species` is deprecated
- An error message in `PhotonParticleContainer.cpp` is reworded to take into account the preferred way of defining a photon species. 
- Explicitly setting particles' mass to zero is now forbidden. This is potentially a breaking change, but in my opinion it is very unlikely to generate issues.


Docs/source/usage/parameters.rst
aeriforme, line 1: Cool cool!

Can you also add in the docs, at the ``<species_name>.mass`` entry, this:

must be > zero. For massless species (i.e. photons for now), use ``<species_name>.species_type``. 

?

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -169,6 +169,12 @@ PhysicalParticleContainer::PhysicalParticleContainer (AmrCore* amr_core, int isp
     const bool charge_is_specified = utils::parser::queryWithParser(pp_species_name, "charge", charge);
     const bool mass_is_specified = utils::parser::queryWithParser(pp_species_name, "mass", m_mass);
 
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE (
+        (!mass_is_specified) ||
+        (m_mass > 0.0),
```
aeriforme, line 174: Are we sure that there's no checks throughout the code that use the condition `if mass == 0`?

aeriforme: Also, why do we want to forbid `m=0`?

RemiLehe (approved):

RemiLehe:

Source/WarpX.cpp
```diff
@@ -2112,6 +2112,14 @@ WarpX::BackwardCompatibility ()
             ablastr::warn_manager::WarnPriority::low);
     }
 
+    if (pp_particles.contains("photon_species")){
+        ablastr::warn_manager::WMRecordWarning("Species",
+            "particles.photon_species is deprecated and may be removed in the future. "
+            "It is recommended to initialize photon particles by setting their "
+            "'species_type' to 'photon', instead.",
+            ablastr::warn_manager::WarnPriority::low);
+    }
```
EZoni, line 2121: I had a question here, but I didn't ask quickly enough. The warning says that the option "may be removed in the future". Does this mean that it hasn't been removed yet? If it has been removed, i.e., either we don't parse it or we don't use it even if we parse it, shouldn't the warning say that the option "has been removed" instead?

RemiLehe: My understanding is that it hasn't been removed yet. (i.e. the previous input script using `particles.photon_species` will still work, but raise the warning.)

EZoni: Okay, I don't know what policy we follow, but I would have just removed it at once and raised an error instead of a warning. The plan now would be that one of us remembers to actually remove it at some point in the future? Or do we just leave all old input options hidden but still functional usually? I don't remember our policy here.

aeriforme:
@lucafedeli88 @RemiLehe 

I think we should add in the docs that mass = 0 is not allowed and that photons must be defined using the species type (see my comment above)!

Also I think there may be at least one part of the code where we do "if mass=0" (see other comment above).

UPDATE: I forgot to submit the review, so you couldn't see my comments 🥲

aeriforme:
[Here we do a if mass==0](https://github.com/BLAST-WarpX/warpx/blob/development/Source/Particles/ParticleCreation/AddParticles.cpp#L699). Not sure if this breaks, but it would be nice to align on this.

Source/WarpX.cpp
lucafedeli88: I can confirm what @RemiLehe said: I was reluctant to introduce a breaking change, so with this PR only a low-priority warning is generated and the instruction is still parsed and used as before (with a minor difference in the case of explicitly setting the mass of a species to zero, which is now forbidden). I am open to discuss other solutions and I acknowledge that we don't have a clear protocol to deal with deprecated features. I think that, if we decide to introduce this possibility, we should remove deprecated features after ~ 3-6 months.

lucafedeli88:
> @lucafedeli88 @RemiLehe
> 
> I think we should add in the docs that mass = 0 is not allowed and that photons must be defined using the species type (see my comment above)!
> 
> Also I think there may be at least one part of the code where we do "if mass=0" (see other comment above).
> 
> UPDATE: I forgot to submit the review, so you couldn't see my comments 🥲

I agree. The documentation may need some additional info. I added the error when mass is set to zero because I've observed a segmentation fault if I don't set a species as a photon species and I set its mass to zero.

Source/Particles/PhysicalParticleContainer.cpp
lucafedeli88: Maybe,  but if we set the mass of a non-photon species  to zero we get segmentation faults at least in some cases. We could decide in principle to consider that a massless particle is always a photon and only raise an error if it has a non-zero charge...

Source/WarpX.cpp
```diff
@@ -2112,6 +2112,14 @@ WarpX::BackwardCompatibility ()
             ablastr::warn_manager::WarnPriority::low);
     }
 
+    if (pp_particles.contains("photon_species")){
+        ablastr::warn_manager::WMRecordWarning("Species",
+            "particles.photon_species is deprecated and may be removed in the future. "
```
ax3l, line 2117: I would phrase `may` to `will`

ax3l: Discussed today:
```
"particles.photon_species was deprecated in WarpX 25.11 and will be removed in the future. "
```

PR #6269, opened 2025-10-15
External Particle Fields: enable reading of multiple fields from file

SeverinDiederichs:
This PR adds the functionality to read in multiple external particle fields from an openPMD file. Inspired from the HybridSolver, a new helper class `ExternalParticleFields` is introduced, which holds the arrays of the meta data for the external E and B fields.

Then, the fields are read and written into the `m_fields` by using as many components as there are external E and B fields, respectively. This allows to read in fields from different files witth different time dependencies.

The previous syntax of reading in a single field is preserved and functional. Now, additionally, multiple fields can be read in as
```
            particles.B_ext_particle_init_style = read_from_file
            particles.B_ext_particle_fields = b1 b2
            particles.b1.read_fields_from_path = diags/Bfield_map1
            particles.b1.read_fields_B_dependency(t) = cos(omega*t + phase)
            particles.b2.read_fields_from_path = diags/Bfield_map2
            particles.b2.read_fields_B_dependency(t) = cos(2*omega*t + phase)
```

For the picmi interface, a dictionary has to be created per field, similarly to the HybridSolver:
```python
    applied_field = picmi.LoadAppliedField(
        load_E=False,
        load_B=True,
        B_external_fields={
            "b1": {
                "read_fields_from_path": "diags/Bfield_map1",
                "read_fields_B_dependency(t)": f"cos({omega}*t + {phase})",
            },
            "b2": {
                "read_fields_from_path": "diags/Bfield_map2",
                "read_fields_B_dependency(t)": f"cos(2*{omega}*t + {phase})",
            },
        },
    )
```

Note that the global `particles.read_fields_from_path` is ignored as soon as multiple fields are specified and a warning is issued. Furthermore, for each field, the path must be specified individually.

Additional CI tests are added with different time dependency, similar to the previously existing one that read only a single field.

This plot shows the beating from two field maps at different frequencies:
frequency_beating

Any feedback or suggestions are more than welcome!


SeverinDiederichs:
One more comment that might be open for discussion on the syntax:

Currently, the syntax reads like this:

```
particles.E_ext_particle_init_style = read_from_file
particles.E_ext_particle_fields = RF_E_comp
particles.RF_E_comp.read_fields_from_path = dynamic_RF_field_file
particles.RF_E_comp.read_fields_E_dependency(t) = cos(omega*t) # <- here it is read_fields_E_dependency
particles.B_ext_particle_init_style = read_from_file
particles.B_ext_particle_fields = RF_B_comp
particles.RF_B_comp.read_fields_from_path = dynamic_RF_field_file
particles.RF_B_comp.read_fields_B_dependency(t) = cos(omega*t + phase_B) # <- here it is read_fields_B_dependency
```
(note that this simple example would not require the multi-field read, it would only be required if another field is added)

The reason was to keep it close to the single field syntax, as there one had to provide the dependency per field:
```
particles.E_ext_particle_init_style = read_from_file
particles.B_ext_particle_init_style = read_from_file
particles.read_fields_from_path = dynamic_RF_field_file
particles.read_fields_E_dependency(t) = cos(omega*t)
particles.read_fields_B_dependency(t) = cos(omega*t + phase_B)
```

However, with the multiple fields, the time dependency is read in per field. Therefore, one could simplify:
`read_fields_E_dependency(t)` and `read_fields_B_dependency(t)` to `read_fields_time_dependency(t)`, as the additional clarification of E and B is redundant, as it has to be already clarified before what kind of field (E or B) it is. 

The advantage would be a simpler and clearer syntax for the multi-field read-in, the disadvantage would be a greater discrepancy (not exactly the same syntax) as for the single field read-in. 

No strong opinion on my side, but wanted to raise the point for discussion

Examples/Tests/load_external_field/CMakeLists.txt
```diff
@@ -93,3 +93,25 @@ add_warpx_test(
...
+
+add_warpx_test(
+  test_3d_load_external_field_particle_multi_time_picmi   # name
+  3                                                       # dims
+  1                                                       # nprocs
+  inputs_test_3d_load_external_field_particle_multi_time_picmi.py  # inputs
+  "analysis_time_scaling.py --pf0 diags/diag1000000 --pfN diags/diag1000300 --component Bz --expected_ratio 0.0 --rtol 1e-12 --min_abs 1e-12"
+  "analysis_default_regression.py --path diags/diag1000300"
+  OFF
+)
```
EZoni, line 117: As mentioned on Slack, I would still try to keep the comments at the end of each argument, for clarity - unless you think it hurts for some reason:
```suggestion

# Multi-field time dependency tests

add_warpx_test(
  test_3d_load_external_field_particle_multi_time     # name
  3                                                   # dims
  1                                                   # nprocs
  inputs_test_3d_load_external_field_particle_multi_time  # inputs
  "analysis_time_scaling.py --pf0 diags/diag1000000 --pfN diags/diag1000300 --component Bz --expected_ratio 0.0 --rtol 1e-12 --min_abs 1e-12"  # analysis
  "analysis_default_regression.py --path diags/diag1000300"  # checksum
  OFF
)

add_warpx_test(
  test_3d_load_external_field_particle_multi_time_picmi   # name
  3                                                       # dims
  1                                                       # nprocs
  inputs_test_3d_load_external_field_particle_multi_time_picmi.py  # inputs
  "analysis_time_scaling.py --pf0 diags/diag1000000 --pfN diags/diag1000300 --component Bz --expected_ratio 0.0 --rtol 1e-12 --min_abs 1e-12"  # analysis
  "analysis_default_regression.py --path diags/diag1000300"  # checksum
  OFF
)
```

SeverinDiederichs: Ah, yes, absolutely no problem to keep all lines commented! Thanks for the cleaning!

Examples/Tests/load_external_field/inputs_test_3d_load_external_field_particle_multi_time_picmi.py
```diff
@@ -0,0 +1,121 @@
...
+    initial_distribution=ion_dist,
+)
+
+# -------------------
+# Applied field (file)
+# -------------------
+applied_field = picmi.LoadAppliedField(
+    load_E=False,
+    load_B=True,
+    B_external_fields={
```
RemiLehe: Since this deviates from the PICMI standard and only WarpX would understand this syntax, our convention is that the argument should be prefixed with `warpx_`. Could you make the corresponding change throughout the source code, the docstrings, and the tests?
```suggestion
    warpx_B_external_fields={
```

Python/pywarpx/picmi.py
```diff
@@ -2391,13 +2391,53 @@ def applied_field_initialize_inputs(self):
...
+        warpx_B_time_function="cos(2*pi*2e6*t + pi/2)",
+    )
+    ```
+
+    **Multi-field mode (multiple independent field maps):**
+    ```python
+    applied_field = picmi.LoadAppliedField(
+        load_E=False,
+        load_B=True,
+        B_external_fields={
```
RemiLehe:
```suggestion
        warpx_B_external_fields={
```

Python/pywarpx/picmi.py
```diff
@@ -2406,62 +2446,157 @@ class LoadAppliedField(picmistandard.PICMI_LoadAppliedField):
...
-        Run the projection-based B-field divergence cleaner after loading.
+        If True, run the projection-based B-field divergence cleaner after loading.
 
     warpx_projection_div_cleaner_atol : float, optional
         Absolute tolerance for the divergence cleaner solve.
 
     warpx_projection_div_cleaner_rtol : float, optional
         Relative tolerance for the divergence cleaner solve.
+
+    B_external_fields : dict, optional
```
RemiLehe:
```suggestion
    warpx_B_external_fields : dict, optional
```

Python/pywarpx/picmi.py
```diff
@@ -2406,62 +2446,157 @@ class LoadAppliedField(picmistandard.PICMI_LoadAppliedField):
...
+        Dictionary specifying multiple magnetic field maps to load.
+        Each key is a unique field name, and each value is a dict containing:
+        - ``read_fields_from_path`` (str): Path to the field data.
+        - ``read_fields_B_dependency(t)`` (str, optional): Time scaling expression.
+
+    E_external_fields : dict, optional
```
RemiLehe:
```suggestion
    warpx_E_external_fields : dict, optional
```

Source/Particles/ExternalParticleFields.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+    // No E field names provided, initialize for single E field without name
+    if (m_E_field_metadata.empty()) {
+
+        // no E field names provided but read_from_file requested, force to read path
+        pp_particles.get("read_fields_from_path", single_field_path);
+
+        // initialize single E field meta data
+        ParticleFieldMetaData md;
+        md.name = "legacy_E";
```
RemiLehe: Maybe this would be a better name?
```suggestion
        md.name = "single_external_E_field";
```

Source/Particles/ExternalParticleFields.cpp
```diff
@@ -0,0 +1,167 @@
...
+        // No E field names provided, initialize for single E field without name
+        if (m_B_field_metadata.empty()) {
+
+            // no B field names provided but read_from_file requested, force to read path
+            pp_particles.get("read_fields_from_path", single_field_path);
+
+            // initialize single B field meta data
+
+            ParticleFieldMetaData md;
+            md.name = "legacy_B";
```
RemiLehe: Maybe this would be a better name?
```suggestion
            md.name = "single_external_B_field";
```

RemiLehe: Actually: do we really need to store `name` in the `ExternalParticleFields` class? What is it used for?

RemiLehe:
Thanks again for the PR, @SeverinDiederichs 

For the PICMI interface, I actually have an alternative suggestion. Do you think that repeated calls to `add_applied_field` would make sense? Like this:
```python
applied_field1 = picmi.LoadAppliedField(
    read_fields_from_path="../../../../openPMD-example-datasets/example-femm-3d.h5",
    load_E=False,
    load_B=True,
    warpx_B_time_function=f"cos({omega}*t + {phase})",
)

applied_field2 = picmi.LoadAppliedField(
    read_fields_from_path="../../../../openPMD-example-datasets/example-femm-3d.h5",
    load_E=False,
    load_B=True,
    warpx_B_time_function=f"cos(2*{omega}*t + {phase})",
)

sim.add_applied_field(applied_field1)
sim.add_applied_field(applied_field2)
```
This way, it would still conform to the official PICMI standard.

If you agree, please ignore my previous comments on the PICMI file. Also: if you agree with these changes, they could be implemented in a follow-up PR, so as to merge the present PR rapidly.

SeverinDiederichs:
Hi @RemiLehe, thank you for the suggestions!

1. Indeed, storing of the name was not necessary, I don't remember why I did it originally. I removed it and with that also the improved naming is not needed anymore.
2. The alternative suggestion for the PICMI interface works very well. It requires a little change in the back:

Because it has to under the hood model this behaviour:
```
            particles.B_ext_particle_init_style = read_from_file
            particles.B_ext_particle_fields = b1 b2
            particles.b1.read_fields_from_path = diags/Bfield_map1
            particles.b1.read_fields_B_dependency(t) = cos(omega*t + phase)
            particles.b2.read_fields_from_path = diags/Bfield_map2
            particles.b2.read_fields_B_dependency(t) = cos(2*omega*t + phase)
```
the code now auto-generates a new name whenever a new field is added (with incrementing a counter).

That way, your proposed setup of generating multiple external fields and just calling `add_applied_field` works.
I agree that this is the cleaner handling, as it is fully compliant with the PICMI interface.

Source/Particles/ExternalParticleFields.cpp
SeverinDiederichs: Indeed not needed, so I removed it.

RemiLehe (approved):

PR #6272, opened 2025-10-16
Move InitBorrowing and ShrinkBorrowing from WarpX class to anonymous namespace in WarpXFaceExtensions.cpp

lucafedeli88:
`InitBorrowing`  and `ShrinkBorrowing` are used only once in `WarpXFaceExtensions.cpp`. This PR moves them to anonymous namespace in that file in order to simplify the header of the WarpX class. Moreover, some refactoring is performed for `InitBorrowing` to make the function more compact.


RemiLehe (approved):
Thanks for cleaning this up!

PR #4867, opened 2024-04-17
[WIP] mcc in mid x push

oshapoval:
Do not merge, for testing purposes only.

PR #4866, opened 2024-04-17
[WIP] Splitting x- push

oshapoval:
Do not merge for now, for testing purposes.

PR #6275, opened 2025-10-17
[WIP] Energy-preserving explicit electromagnetic PIC-MCC algorithm

### LINKED ISSUE #5851: Implement Energy-Preserving PIC-MCC Loop
## Context and motivation
Different variations of the energy-preserving PIC-MCC loop described in [(Vay, Jean-Luc, et al. "Energy-preserving coupling of explicit particle-in-cell with Monte Carlo collisions." Physical Review E 111.2 (2025): 025306)](https://doi.org/10.1103/PhysRevE.111.025306) have been implemented in forked branches. Now we would like to implement one of those variations in the main branch of the code.

## Proposed feature
Here is a graph of the existing PIC-MCC loop, courtesy of @jlvay:

Image

And here is a graph of the proposed PIC-MCC loop, courtesy of @jlvay:

Image

## Alternative solutions
N/A

## Additional information
- List of branches of previous implementations:
  - the branch [mcc_in_mid_v_push_with_MCC_fix](https://github.com/oshapoval/WarpX/tree/mcc_in_mid_v_push_with_MCC_fix) with "mid-v-push" implementation (most up to date)
  - the branch [mcc_in_mid_x_push_2_depositions](https://github.com/BLAST-WarpX/warpx/compare/development...oshapoval:WarpX:mcc_in_mid_x_push_2_depositions) with "mid-x-push" implementation
- Being able to instrument the code so that graphs like the ones above can be generated, similarly to what was prototyped in #5269, would be useful to validate the implementation.
- Having a list of CI tests ready to be extended would be useful to validate the implementation.

EZoni:
### Overview

Follow up on #5955 and implement the electromagnetic part of the energy-preserving PIC-MCC algorithm described in [(Vay, Jean-Luc, et al. "Energy-preserving coupling of explicit particle-in-cell with Monte Carlo collisions." Physical Review E 111.2 (2025): 025306)](https://doi.org/10.1103/PhysRevE.111.025306) (i.e., explicit PIC coupled with Monte Carlo collisions).

### Notes

- Closes #5851.
- Production-ready implementation of #4866, #4867, #6302 (without diags change).
- Alternative algorithm for the one in #4726.

### To do

- [x] Add a new electromagnetic test.
- [x] Implement the new algorithm.
- [x] Test & debug:
  - [x] Fix GPU issue that seems to be resolved by passing `communicate=1` to `AddRealComp`.
  - [x] Fix charge deposition (deposit old $\rho$ in the first half, new $\rho$ in the second half).
  - [x] Check tests that fail and reset the checksums if needed.


EZoni:
Basically the goal here will be to have the analysis of the test `test_2d_collisions_split_position_push_electromagnetic` pass without breaking any other tests, with respect to this CI reference starting point: https://dev.azure.com/BLAST-WarpX/WarpX/_build/results?buildId=2275.

EZoni:
Discussed with @oshapoval offline:
- The momenta that are averaged should be the momentum at the end of the first position half-push and the momentum after the collisions.
- The relative time used for the current deposition should not need any correction.

Source/WarpX.cpp
```diff
@@ -346,7 +346,7 @@ WarpX::WarpX ()
     t_old.resize(nlevs_max, std::numeric_limits<Real>::lowest());
     dt.resize(nlevs_max, std::numeric_limits<Real>::max());
 
-    mypc = std::make_unique<MultiParticleContainer>(this);
+    mypc = std::make_unique<MultiParticleContainer>(this, m_collisions_split_position_push);
```
EZoni, line 350: If I keep the allocation of the average momentum runtime attributes in the constructor of `PhysicalParticleContainer`, then I have to pass the information in `m_collisions_split_position_push` down to the particle container classes.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -336,6 +339,14 @@ PhysicalParticleContainer::PhysicalParticleContainer (AmrCore* amr_core, int isp
...
         m_boundary_conditions.SetThermalVelocity(boundary_uth);
     }
+
+    // If the position push is split to perform collisions, add the
+    // average momentum components to deposit the current correctly
+    if (collisions_split_position_push) {
+        this->AddRealComp("ux_avg", /*communicate=*/0);
+        this->AddRealComp("uy_avg", /*communicate=*/0);
+        this->AddRealComp("uz_avg", /*communicate=*/0);
+    }
```
EZoni: To double check if this is the correct place to add the average momentum runtime attributes.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1466,6 +1496,20 @@ PhysicalParticleContainer::PushPX (WarpXParIter& pti,
...
+            uz_avg[ip] = uz[ip];
+        }
+        else { // The momentum was updated during collisions
+            ux_avg[ip] += ux[ip];
+            uy_avg[ip] += uy[ip];
+            uz_avg[ip] += uz[ip];
+            ux_avg[ip] *= 0.5_rt;
+            uy_avg[ip] *= 0.5_rt;
+            uz_avg[ip] *= 0.5_rt;
+        }
```
EZoni: To double check if the logic here is correct.

EZoni:
@RemiLehe @oshapoval 

I start requesting your review in case you have comments on the early draft. I don't think it's working correctly yet.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -441,34 +441,40 @@ void WarpX::OneStep (
...
+                    OneStep_nosub(a_cur_time, a_dt, a_step);
                 }
                 // JRhom PIC loop
                 else {
+                    // perform particle collisions
+                    ExecutePythonCallback("beforecollisions");
+                    mypc->doCollisions(a_step, a_cur_time, a_dt);
+                    ExecutePythonCallback("aftercollisions");
+
                     OneStep_JRhom(a_cur_time);
```
EZoni, line 457: Note that in this PR I won't attempt to make the new algorithm work with the JRhom method.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -441,34 +441,40 @@ void WarpX::OneStep (
...
                         finest_level == 1,
                         "Subcycling not implemented with more than 1 mesh refinement level"
                     );
+
+                    // perform particle collisions
+                    ExecutePythonCallback("beforecollisions");
+                    mypc->doCollisions(a_step, a_cur_time, a_dt);
+                    ExecutePythonCallback("aftercollisions");
+
                     OneStep_sub1(a_cur_time);
```
EZoni, line 478: Note that in this PR I won't attempt to make the new algorithm work with the MR+subcycling method.

Source/Particles/PhysicalParticleContainer.cpp
EZoni: Also note that passing `communicate=1` causes erroneous arithmetic operation somewhere in the OpenPMD diagnostics chain at the beginning of the test. Not sure what to make of this.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -465,6 +476,13 @@ PhysicalParticleContainer::Evolve (ablastr::fields::MultiFabRegister& fields,
         !do_not_deposit &&
         (position_push_type == PositionPushType::Full || position_push_type == PositionPushType::FirstHalf)
     );
+    // FIXME Check the logic here
+    bool const deposit_current = (
+        !skip_deposition &&
+        !do_not_deposit &&
+        !(implicit_options && implicit_options->evolve_suborbit_particles_only) &&
+        (position_push_type == PositionPushType::Full || position_push_type == PositionPushType::SecondHalf)
+    );
```
EZoni, line 500: Note that I added `do_not_deposit` here - it was _not_ taken into account in the previous `if` condition.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -664,8 +687,8 @@ PhysicalParticleContainer::Evolve (ablastr::fields::MultiFabRegister& fields,
 
                 WARPX_PROFILE_VAR_STOP(blp_fg);
 
-                // Current Deposition
-                if (!skip_deposition && !(implicit_options && implicit_options->evolve_suborbit_particles_only))
```
EZoni, line 679: This is the previous `if` condition where `do_not_deposit` was _not_ taken into account.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1327,6 +1346,11 @@ PhysicalParticleContainer::PushPX (WarpXParIter& pti,
     ParticleReal* const AMREX_RESTRICT uy = attribs[PIdx::uy].dataPtr() + offset;
     ParticleReal* const AMREX_RESTRICT uz = attribs[PIdx::uz].dataPtr() + offset;
 
+    // Average momentum
+    amrex::ParticleReal* const AMREX_RESTRICT ux_avg = pti.GetAttribs("ux_avg").dataPtr() + offset;
+    amrex::ParticleReal* const AMREX_RESTRICT uy_avg = pti.GetAttribs("uy_avg").dataPtr() + offset;
+    amrex::ParticleReal* const AMREX_RESTRICT uz_avg = pti.GetAttribs("uz_avg").dataPtr() + offset;
+
```
EZoni, line 1372: I'm copying the `+ offset` from the other definitions I see here, but I admit I don't remember what this is about.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -497,9 +516,9 @@ PhysicalParticleContainer::Evolve (ablastr::fields::MultiFabRegister& fields,
             // Extract particle data
             auto& attribs = pti.GetAttribs();
             auto&  wp = attribs[PIdx::w];
-            auto& uxp = attribs[PIdx::ux];
-            auto& uyp = attribs[PIdx::uy];
-            auto& uzp = attribs[PIdx::uz];
+            auto& uxp = (m_collisions_split_position_push) ? pti.GetAttribs("ux_avg") : attribs[PIdx::ux];
+            auto& uyp = (m_collisions_split_position_push) ? pti.GetAttribs("uy_avg") : attribs[PIdx::uy];
+            auto& uzp = (m_collisions_split_position_push) ? pti.GetAttribs("uz_avg") : attribs[PIdx::uz];
```
EZoni, line 535: This may break the electrostatic cas? Maybe we need a more sophisticated logic to decide which momentum to use here (e.g., use a `use_average_momentum` boolean defined in the constructor of the class, or something along these lines) - to be double checked.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -494,7 +504,29 @@ WarpX::OneStep_nosub (Real cur_time)
...
+    // push particles (half position and full momentum)
+    PushParticlesandDeposit(
+        a_cur_time,
+        /*skip_current=*/true,
+        PositionPushType::FirstHalf,
+        MomentumPushType::Full
+    );
+
+    // communicate particle data
+    mypc->Redistribute();
```
oshapoval: For better performance, reproduce the logic in `HandleParticlesAtBoundaries` to ensure only local communication.

Source/Particles/PhysicalParticleContainer.cpp
oshapoval: The averaged momentum should be used only in the current deposition, while `uxp`, `uyp`, and `uzp` may be used somewhere else too.

Source/Particles/PhysicalParticleContainer.cpp
EZoni: However, it is taken into account in 
https://github.com/BLAST-WarpX/warpx/blob/d5ee9f0d8cf6271d91be0197077957b8afa47e01/Source/Particles/WarpXParticleContainer.cpp#L403-L404
and in 
https://github.com/BLAST-WarpX/warpx/blob/d5ee9f0d8cf6271d91be0197077957b8afa47e01/Source/Particles/WarpXParticleContainer.cpp#L950-L951

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -494,7 +504,29 @@ WarpX::OneStep_nosub (Real cur_time)
...
+    mypc->doCollisions(a_step, a_cur_time, a_dt);
+    ExecutePythonCallback("aftercollisions");
+
+    // push particles (half position)
+    PushParticlesandDeposit(
+        a_cur_time,
+        /*skip_current=*/false,
+        PositionPushType::SecondHalf,
+        MomentumPushType::None
+    );
```
EZoni: This needs to take into account the value of the user input `m_collisions_split_position_push`.

Source/Particles/PhysicalParticleContainer.cpp
EZoni: Also, the average momentum attributes are not always added, so this can cause a runtime crash.

Source/Evolve/WarpXEvolve.cpp
EZoni: Resolved in [d16803f](https://github.com/BLAST-WarpX/warpx/pull/6275/commits/d16803f92ba2d47c222349b06e6a78ad66414108).

Source/Particles/PhysicalParticleContainer.cpp
EZoni: It seems like `uxp`, `uyp`, and `uzp` are indeed used only for current deposition within `Evolve` - so this shouldn't be a source of error.

Source/Particles/PhysicalParticleContainer.cpp
EZoni: To make things simpler and easier to debug, we now add the runtime components uncondionally for now: [d16803f](https://github.com/BLAST-WarpX/warpx/pull/6275/commits/d16803f92ba2d47c222349b06e6a78ad66414108).

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -494,7 +504,48 @@ WarpX::OneStep_nosub (Real cur_time)
     ExecutePythonCallback("particlescraper");
     ExecutePythonCallback("beforedeposition");
 
-    PushParticlesandDeposit(cur_time);
+    // with collisions placed in the middle of the position push and after the momentum push
+    if (m_collisions_split_position_push) {
+        // push particles (half position and full momentum)
+        PushParticlesandDeposit(
+            a_cur_time,
+            /*skip_current=*/true,
```
EZoni: Reflect new name after #6303:
```suggestion
            /*skip_deposition=*/true,
```

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -494,7 +504,48 @@ WarpX::OneStep_nosub (Real cur_time)
...
+
+        // perform particle collisions
+        ExecutePythonCallback("beforecollisions");
+        mypc->doCollisions(a_step, a_cur_time, a_dt);
+        ExecutePythonCallback("aftercollisions");
+
+        // push particles (half position)
+        PushParticlesandDeposit(
+            a_cur_time,
+            /*skip_current=*/false,
```
EZoni: Reflect new name after #6303:
```suggestion
            /*skip_deposition=*/false,
```

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -494,7 +504,48 @@ WarpX::OneStep_nosub (Real cur_time)
...
+    else {
+        // perform particle collisions
+        ExecutePythonCallback("beforecollisions");
+        mypc->doCollisions(a_step, a_cur_time, a_dt);
+        ExecutePythonCallback("aftercollisions");
+
+        // push particles (half position)
+        PushParticlesandDeposit(
+            a_cur_time,
+            /*skip_current=*/false,
```
EZoni: Reflect new name after #6303:
```suggestion
            /*skip_deposition=*/false,
```

EZoni: Resolved in [df7ffd7](https://github.com/BLAST-WarpX/warpx/pull/6275/commits/df7ffd784f989117779769f265dba79df38690d4).

Source/Evolve/WarpXEvolve.cpp
EZoni: Resolved in [df7ffd7](https://github.com/BLAST-WarpX/warpx/pull/6275/commits/df7ffd784f989117779769f265dba79df38690d4).

Source/Evolve/WarpXEvolve.cpp
EZoni: Resolved in [df7ffd7](https://github.com/BLAST-WarpX/warpx/pull/6275/commits/df7ffd784f989117779769f265dba79df38690d4).

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -336,6 +336,12 @@ PhysicalParticleContainer::PhysicalParticleContainer (AmrCore* amr_core, int isp
         utils::parser::getWithParser(pp_species_boundary,"u_th",boundary_uth);
         m_boundary_conditions.SetThermalVelocity(boundary_uth);
     }
+
+    // Add the average momentum components to deposit the current correctly
+    // if the position push is split to perform collisions
+    AddRealComp("ux_avg");
+    AddRealComp("uy_avg");
+    AddRealComp("uz_avg");
```
EZoni: This causes a crash for single-process runs, at the first diagnostic dump - to be debugged. 

Incidentally, passing `communicate=false` fixes the crash, though this isn't the proper fix probably.

EZoni: I think it has something to do with the number `7` hard-coded here,
https://github.com/BLAST-WarpX/warpx/blob/cac4478c2dbcf8763d496b75bac514b4ea09f256/Source/Diagnostics/ParticleDiag/ParticleDiag.H#L34
as well as things like what we do here,
https://github.com/BLAST-WarpX/warpx/blob/cac4478c2dbcf8763d496b75bac514b4ea09f256/Source/Diagnostics/WarpXOpenPMD.cpp#L633-L635

EZoni: My opinion is that we would not want the new runtime components in the diagnostics.

This seems to be achieved by passing `communicate=false` to `AddRealComp` - for reasons I don't understand.

However, passing `communicate=false` seems to be buggy with multi-process runs (observed on multi-process runs on GPU).

I'm probably just missing some particle data communication somewhere after adding the runtime components.

EZoni: @atmyers just commented on Slack that we probably want to:
- Pass `communicate=true` (default) - already done in [b517f65](https://github.com/BLAST-WarpX/warpx/pull/6275/commits/b517f65136d0f9daac63b30fe9a09e77a02d38d4).
- Add logic to turn off writing out the new components here: https://github.com/BLAST-WarpX/warpx/blob/cac4478c2dbcf8763d496b75bac514b4ea09f256/Source/Diagnostics/WarpXOpenPMD.cpp#L650

EZoni: Logic to turn off writing out the new components implemented in ~[0926df1](https://github.com/BLAST-WarpX/warpx/pull/6275/commits/0926df1d811db1171e85ac88cc8c55e39a541c80)~ [af99938](https://github.com/BLAST-WarpX/warpx/pull/6275/commits/af99938e60d5a782fb5761a6ee719505e0e2aa29).

@atmyers 
Please let me know if it looks reasonable to you.

EZoni: It seems like this wasn't enough to remove it from the output - I still see the components in the new checksum files, which are missing from the benchmark files, thus causing most CI tests to fail.

I did remove it from somewhere (because the code does not crash anymore), but not from all necessary places, I guess.

EZoni: Okay, I think I simply missed to port the fix to the plotfile implementation as well.

I kind of think that it would be nice to move the control for this at the level of `AddRealComp`, e.g., `AddRealComp("new_comp", communicate=true, write=false)`, rather than coupling data communication and data output into a single boolean control variable (`communicate` in this case).

EZoni: Hopefully resolved in [af99938](https://github.com/BLAST-WarpX/warpx/pull/6275/commits/af99938e60d5a782fb5761a6ee719505e0e2aa29).

EZoni:
The only CI test that crashes runtime after [648020d](https://github.com/BLAST-WarpX/warpx/pull/6275/commits/648020d4445392ff605eac877c1f8aabe4765739) is `test_1d_photoneutralization_dsmc`.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -336,6 +338,14 @@ PhysicalParticleContainer::PhysicalParticleContainer (AmrCore* amr_core, int isp
...
         m_boundary_conditions.SetThermalVelocity(boundary_uth);
     }
+
+    if (m_collisions_split_position_push) {
+        // Add the average momentum components to deposit the current correctly
+        // if the position push is split to perform collisions
+        AddRealComp("ux_avg");
+        AddRealComp("uy_avg");
+        AddRealComp("uz_avg");
+    }
```
EZoni, line 359: @atmyers 

Could you explain again where these components should be initialized, if I wanted to initialize them to zero for all particles and all containers?

ax3l: Check what is done for the QED runtime components, e.g., add your new names here:
https://github.com/BLAST-WarpX/warpx/blob/25.10/Source/Particles/ParticleCreation/DefaultInitialization.H#L148-L209

EZoni:
@atmyers 

I have some local code changes that I think do what we discussed, but I am stuck on one point.

I think I would have to pass the `ux_avg`, `uy_avg`, and `uz_avg` data pointers to the function `user_real_attrib_parserexec` here in order to default-initialize them, even though I don't really need to expose them to the user inputs, correct?
https://github.com/BLAST-WarpX/warpx/blob/5cdbc378fbb8043e80fd673727635c22ef49fdd4/Source/Particles/ParticleCreation/DefaultInitialization.H#L208-L238

If that's the case, here I see
```cpp
const amrex::ParserExecutor<7> user_real_attrib_parserexec = user_real_attrib_parser[ia]->compile<7>();
```
where `7` is the hard-coded number of components (positions, momenta, time).

In my case that number would need to be `10` for the three additional average momenta. However, this would be conditional, based on the value of a boolean input `collisions_split_push` (that I am passing as a new argument to `DefaultInitializeRuntimeAttributes` in my local changes).

I think the compiler is not happy if the template parameter (`7` in the current code) is not determined at compile time.

Does this mean that I'm forced to duplicate the whole code logic into a separate code block for my components...?

atmyers:
> @atmyers
> 
> I have some local code changes that I think do what we discussed, but I am stuck on one point.
> 
> I think I would have to pass the `ux_avg`, `uy_avg`, and `uz_avg` data pointers to the function `user_real_attrib_parserexec` here in order to default-initialize them, even though I don't really need to expose them to the user inputs, correct?
> 
> https://github.com/BLAST-WarpX/warpx/blob/5cdbc378fbb8043e80fd673727635c22ef49fdd4/Source/Particles/ParticleCreation/DefaultInitialization.H#L208-L238
> 
> If that's the case, here I see
> 
> ```c++
> const amrex::ParserExecutor<7> user_real_attrib_parserexec = user_real_attrib_parser[ia]->compile<7>();
> ```
> 
> where `7` is the hard-coded number of components (positions, momenta, time).
> 
> In my case that number would need to be `10` for the three additional average momenta. However, this would be conditional, based on the value of a boolean input `collisions_split_push` (that I am passing as a new argument to `DefaultInitializeRuntimeAttributes` in my local changes).
> 
> I think the compiler is not happy if the template parameter (`7` in the current code) is not determined at compile time.
> 
> Does this mean that I'm forced to duplicate the whole code logic into a separate code block for my components...?

I think that the hard-coded `7` refers to the number of *input* arguments to the function that initializes the particle attributes. Each attribute has it's own function, but it's assumed that these functions will only involve the particle's position, momenta, and time. So I don't think you need to change this.

EZoni:
Here's an example of output with print statements for debugging: [test_1d_photoneutralization_dsmc.txt](https://github.com/user-attachments/files/23490081/test_1d_photoneutralization_dsmc.txt).


This is the relevant diff showing where I'm printing:
```diff
diff --git a/Source/Particles/Collision/BinaryCollision/BinaryCollision.H b/Source/Particles/Collision/BinaryCollision/BinaryCollision.H
index 89d4a93e2..6f7889151 100644
--- a/Source/Particles/Collision/BinaryCollision/BinaryCollision.H
+++ b/Source/Particles/Collision/BinaryCollision/BinaryCollision.H
@@ -181,6 +181,7 @@ public:
         auto& species2 = mypc->GetParticleContainerFromName(m_species_names[1]);
 
         // In case of particle creation, create the necessary vectors
+        amrex::Print() << "Starting binary collisions between " << m_species_names[0] << " and " << m_species_names[1] << std::endl;
         const int n_product_species = m_product_species.size();
         amrex::Vector<WarpXParticleContainer*> product_species_vector;
         amrex::Vector<SmartCopyFactory> copy_factory_species1;
diff --git a/Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H b/Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
index e085d54d0..3a91dc696 100644
--- a/Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
+++ b/Source/Particles/Collision/BinaryCollision/DSMC/SplitAndScatterFunc.H
@@ -286,6 +286,7 @@ public:
                 // create a copy of the first product species at the location of species 1
                 const auto reactant1_index = static_cast<int>(p_pair_indices_1[i]);
                 const auto product1_index = products_np_data[2] + with_product_p_offsets[i];
+                amrex::Print() << "Creating " << pc_products[2]->getSpeciesTypeName() << " particle at index " << product1_index << " from reactant index " << reactant1_index << std::endl;
                 copy_species1[2](soa_products_data[2], soa_1, reactant1_index,
                                 static_cast<int>(product1_index), engine);
                 // Set the weight of the new particle to p_pair_reaction_weight[i]
@@ -294,6 +295,7 @@ public:
                 // create a copy of the other product species at the location of species 2
                 const auto reactant2_index = static_cast<int>(p_pair_indices_2[i]);
                 const auto product2_index = products_np_data[3] + with_product_p_offsets[i];
+                amrex::Print() << "Creating " << pc_products[3]->getSpeciesTypeName() << " particle at index " << product2_index << " from reactant index " << reactant2_index << std::endl;
                 copy_species2[3](soa_products_data[3], soa_2, reactant2_index,
                                 static_cast<int>(product2_index), engine);
                 // Set the weight of the new particle to p_pair_reaction_weight[i]
diff --git a/Source/Particles/ParticleCreation/SmartCopy.H b/Source/Particles/ParticleCreation/SmartCopy.H
index 6be363e63..6db08d172 100644
--- a/Source/Particles/ParticleCreation/SmartCopy.H
+++ b/Source/Particles/ParticleCreation/SmartCopy.H
@@ -50,9 +50,11 @@ struct SmartCopy
     {
         // initialize the real components
         for (int j = 0; j < DstData::NAR; ++j) {
+            amrex::Print() << "Initializing real component " << j << " for dst particle " << i_dst << " with policy " << static_cast<int>(m_policy_real[j]) << std::endl;
             dst.m_rdata[j][i_dst] = initializeRealValue(m_policy_real[j], engine);
         }
         for (int j = 0; j < dst.m_num_runtime_real; ++j) {
+            amrex::Print() << "Initializing runtime real component " << j << " for dst particle " << i_dst << " with policy " << static_cast<int>(m_policy_real[j+DstData::NAR]) << std::endl;
             dst.m_runtime_rdata[j][i_dst] = initializeRealValue(m_policy_real[j+DstData::NAR], engine);
         }
 
diff --git a/Source/Particles/PhysicalParticleContainer.cpp b/Source/Particles/PhysicalParticleContainer.cpp
index c4624c2eb..ae5fb6c36 100644
--- a/Source/Particles/PhysicalParticleContainer.cpp
+++ b/Source/Particles/PhysicalParticleContainer.cpp
@@ -1498,13 +1498,19 @@ PhysicalParticleContainer::PushPX (WarpXParIter& pti,
         if (collisions_split_position_push) {
             // Update average momentum
             if (momentum_push_type != MomentumPushType::None) {
+                amrex::Print() << "Setting ux_avg[" << ip << "] = ux[" << ip << "] for " << species_name << std::endl;
                 ux_avg[ip] = ux[ip];
+                amrex::Print() << "Setting uy_avg[" << ip << "] = uy[" << ip << "] for " << species_name << std::endl;
                 uy_avg[ip] = uy[ip];
+                amrex::Print() << "Setting uz_avg[" << ip << "] = uz[" << ip << "] for " << species_name << std::endl;
                 uz_avg[ip] = uz[ip];
             }
             else { // The momentum was updated during collisions
+                amrex::Print() << "Computing ux_avg[" << ip << "] += ux[" << ip << "] for " << species_name << std::endl;
                 ux_avg[ip] += ux[ip];
+                amrex::Print() << "Computing uy_avg[" << ip << "] += uy[" << ip << "] for " << species_name << std::endl;
                 uy_avg[ip] += uy[ip];
+                amrex::Print() << "Computing uz_avg[" << ip << "] += uz[" << ip << "] for " << species_name << std::endl;
                 uz_avg[ip] += uz[ip];
                 ux_avg[ip] *= 0.5_rt;
                 uy_avg[ip] *= 0.5_rt;
```

I also confirm that the erroneous arithmetic operation occurrs only in debug mode with FPE-trapping runtime options, which I believe confirms the hypothesis that this is due to invalid values of `ux_avg`.

EZoni:
@atmyers 

I was debugging this with @RemiLehe and things are still unclear here.

Here's the latest output with some ad hoc print statements: [test_1d_photoneutralization_dsmc.txt](https://github.com/user-attachments/files/23507870/test_1d_photoneutralization_dsmc.txt).

Currently, it seems like `ux_avg` cannot be accessed for the `electrons` created in the collisions.

Do you know if there's a way to print out the component names from `dst` or `dst.m_runtime_rdata` here (snippet from SmartCopy.H)?
```cpp
        // initialize the real components
        for (int j = 0; j < DstData::NAR; ++j) {
            amrex::Print() << "Initializing real component " << j << " for dst particle " << i_dst << " with policy " << static_cast<int>(m_policy_real[j]) << std::endl;
            dst.m_rdata[j][i_dst] = initializeRealValue(m_policy_real[j], engine);
        }
        for (int j = 0; j < dst.m_num_runtime_real; ++j) {
            amrex::Print() << "Initializing runtime real component " << j << " for dst particle " << i_dst << " with policy " << static_cast<int>(m_policy_real[j+DstData::NAR]) << std::endl;
            dst.m_runtime_rdata[j][i_dst] = initializeRealValue(m_policy_real[j+DstData::NAR], engine);
        }
```

Or do you know how to access and print the runtime components (e.g., `ux_avg`) from `soa_products_data` here (snippet from SplitAndScatterFunc.H)?
```cpp
                // create a copy of the first product species at the location of species 1
                const auto reactant1_index = static_cast<int>(p_pair_indices_1[i]);
                const auto product1_index = products_np_data[2] + with_product_p_offsets[i];
                amrex::Print() << "Creating " << pc_products[2]->getSpeciesTypeName() << " particle at index " << product1_index << " from reactant index " << reactant1_index << std::endl;
                copy_species1[2](soa_products_data[2], soa_1, reactant1_index,
                                static_cast<int>(product1_index), engine);
                soa_products_data[2].m_rdata[PIdx::w][product1_index] = p_pair_reaction_weight[i];

                // create a copy of the other product species at the location of species 2
                const auto reactant2_index = static_cast<int>(p_pair_indices_2[i]);
                const auto product2_index = products_np_data[3] + with_product_p_offsets[i];
                amrex::Print() << "Creating " << pc_products[3]->getSpeciesTypeName() << " particle at index " << product2_index << " from reactant index " << reactant2_index << std::endl;
                copy_species2[3](soa_products_data[3], soa_2, reactant2_index,
                                static_cast<int>(product2_index), engine);
                soa_products_data[3].m_rdata[PIdx::w][product2_index] = p_pair_reaction_weight[i];
```

EZoni:
I confirm that it does seem to have something to do with photons specifically. If I change the order of the parent species in the input file, the error occurs on the opposite product species, since the order of inheritance is fixed (and arbitrary). It is probably related to the fact that we set `photons.do_not_push = 1` in the input file, which I guess points back to the initialization issue.

@ax3l 
FYI, I tried running with GDB but wasn't able to catch the bug. For some reason the run from GDB goes through. Here's what I did:
```console
$ gdb --args ./build/bin/warpx.1d Examples/Tests/ionization_dsmc/inputs_test_1d_photoneutralization_dsmc
GNU gdb (Ubuntu 15.0.50.20240403-0ubuntu1) 15.0.50.20240403-git
Copyright (C) 2024 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./build/bin/warpx.1d...
$ (gdb) run
Starting program: /home/edoardo/src/warpx/build/bin/warpx.1d Examples/Tests/ionization_dsmc/inputs_test_1d_photoneutralization_dsmc

This GDB supports auto-downloading debuginfo from the following URLs:
  <https://debuginfod.ubuntu.com>
$ Enable debuginfod for this session? (y or [n]) y
...
...
...
AMReX (25.10-38-g3c1f1c78f605) finalized
[Thread 0x7fffef15e6c0 (LWP 52660) exited]
[Thread 0x7ffff09ff6c0 (LWP 52654) exited]
[Inferior 1 (process 52651) exited normally]
```
after adding these runtime parameters to the input file (to mimic our CI setup),
```diff
+amrex.fpe_trap_invalid = 1
+amrex.fpe_trap_overflow = 1
+amrex.fpe_trap_zero = 1
+amrex.signal_handling = 0
+amrex.throw_exception = 1
```
though I think the trapping flags are unused because we don't handle signals from AMReX intentionally.

EZoni:
Feedback from @WeiqunZhang shared on Slack:

`DefautlInitializeRuntimeAttributes` is supposed to initialize the resized runtime component vectors.

The problem seems to be its caller. It's called in `SplitAndScatterFunc`:
https://github.com/BLAST-WarpX/warpx/blob/cbe73680443ea633c24ba79f133e8d8b63310baf/Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H#L254-L267

The data pointers are in `tile_products`, but `pc_products` says there are zero real attributes. That's why the pointers' memory is not initialized. We could fix it in `DefaultInitializeRuntimeAttributes` by ignoring `pc_products`, but that might cause other issues.

That `pc_products` was assembled by the `doCollisions` function in BinaryCollision.H:
https://github.com/BLAST-WarpX/warpx/blob/cbe73680443ea633c24ba79f133e8d8b63310baf/Source/Particles/Collision/BinaryCollision/BinaryCollision.H#L192-L194

Mabye that `defineAllParticleTiles` does not know about runtime parameters. That's my guess.

So we either need to make that aware of runtime parameters, or we rewrite `DefaultInitizeRuntimeAttributes` to not trust `pc_products`. Or do the runtime components need to be added after `defineAllParticleTiles`?

Anyway, I think when real components are added, they are not added to the collision's production species. If they are not supposed to be there, we should not use `pc_products` to pass in the real comp meta-data information to the `DefaultInitializeRuntimeAttributes`.

EZoni:
With or without #6413, this is what I see from `PhysicalParticleContainer::Evolve` when called for the `electrons` species, which inherits from the `photons` species:
```
25: Starting PhysicalParticleContainer::Evolve for electrons
25: NumRealComps() = 8
25: NumRuntimeRealComps() = 3
25: NStructReal = 0
25: NArrayReal = 5
25: GetRealSoANames(): z
25: GetRealSoANames(): w
25: GetRealSoANames(): ux
25: GetRealSoANames(): uy
25: GetRealSoANames(): uz
25: GetRealSoANames(): ux_avg
25: GetRealSoANames(): uy_avg
25: GetRealSoANames(): uz_avg
25: Erroneous arithmetic operation
```

The erroneous arithmetic operation is due to NaNs in the runtime components, which points to initialization issues, as discussed already.

Here's the full output after a08f073, including print statements from `doCollisions` in BinaryCollision.H (where Weiqun pointed to in https://github.com/BLAST-WarpX/warpx/pull/6275#issuecomment-3553838185):
[test_1d_photoneutralization_dsmc.txt](https://github.com/user-attachments/files/23685551/test_1d_photoneutralization_dsmc.txt)

EZoni:
@WeiqunZhang @atmyers @ax3l @RemiLehe 

If you look at the output in https://github.com/BLAST-WarpX/warpx/pull/6275#issuecomment-3564815506, the number of components and the names seem corret to me. Isn't this inconsistent with what @WeiqunZhang had observed?

What I was not expecting is that this
https://github.com/BLAST-WarpX/warpx/blob/cbe73680443ea633c24ba79f133e8d8b63310baf/Source/Particles/Collision/BinaryCollision/BinaryCollision.H#L190-L202
loops over all species, both parent and product species...

EZoni:
I'm going to push the print statements I'm adding for debugging, to make it easier to follow my comments with output snippets.

EZoni:
@WeiqunZhang @atmyers @ax3l @RemiLehe 

For me something along the lines of 277fee4 (explicit initialization of the average momentum components to zero inside `DefaultInitializeRuntimeAttributes`, even though I thought it wasn't needed) fixes the runtime erroneous arithmetic operation in `test_1d_photoneutralization_dsmc`.

We should discuss if this makes sense or not.

Source/Particles/ParticleCreation/DefaultInitialization.H
```diff
@@ -149,6 +152,25 @@ void DefaultInitializeRuntimeAttributes (PTile& ptile,
...
+                                           attr_ptr[ip] = 0.0_rt;
+                                       });
+                // Otherwise (e.g., particle tile allocated in pinned memory), run on CPU
+                } else {
+                    for (int ip = start; ip < stop; ++ip) {
+                        attr_ptr[ip] = 0.0_rt;
+                    }
+                }
+            }
+
```
EZoni, line 177: @atmyers @WeiqunZhang 

Does this look reasonable to you?

Source/Evolve/WarpXEvolve.cpp
EZoni: Maybe this can be a follow-up optimization PR, since I think it'd require some refactoring of the code in `HandleParticlesAtBoundaries`. We should reference the first comment of this thread into a new issue to track this when the PR is merged.

Source/Particles/ParticleCreation/DefaultInitialization.H
WeiqunZhang: LGTM

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -493,7 +503,48 @@ WarpX::OneStep_nosub (Real cur_time)
...
+        PushParticlesandDeposit(
+            a_cur_time,
+            /*skip_deposition=*/true,
+            PositionPushType::FirstHalf,
+            MomentumPushType::Full
+        );
+
+        // perform essential particle house keeping at the boundaries
+        // (inject, communicate, scrape, sort, etc.)
+        HandleParticlesAtBoundaries(a_step, a_cur_time, /*num_moved=*/0);
```
EZoni, line 519: @RemiLehe @oshapoval 

To discuss whether it is safe to assume that the moving window has not moved by any cells at this point in the PIC loop.

EZoni:
@RemiLehe @oshapoval 

All CI tests passed here, despite the particle boundary change.

Source/Evolve/WarpXEvolve.cpp
EZoni: Resolved in [e26eb2d](https://github.com/BLAST-WarpX/warpx/pull/6275/commits/e26eb2ddcf4e420d8ae2b41e60157584c91477f1).

Source/Evolve/WarpXEvolve.cpp
EZoni: @oshapoval's latest test seems to confirm that it is safe to pass `num_moved=0` in this call.

EZoni:
This algorithm is now discontinued and replaced by a different algorithm implemented in #6530.

PR #6277, opened 2025-10-20
Fix Leonardo CINECA docs link

elehcim:


lucafedeli88 (approved):
Thanks for this fix, @elehcim  !

EZoni:
For some reason pre-commit.ci did not run (though I don't see anything pending on https://results.pre-commit.ci/), so merge is blocked. I can push an amended commit to retrigger the checks.

EZoni (approved):

EZoni:
If it doesn't work we need to look into the settings of pre-commit.ci, which I don't have access to.

PR #6281, opened 2025-10-20
Deprecation of photon species list: update documentation

lucafedeli88:
This PR implements some suggestions by @aeriforme regarding https://github.com/BLAST-WarpX/warpx/pull/6265 .
Now the documentation explicitly states that setting particles' mass to zero is forbidden (it has to be done via the `species_type` setting). It also states more clearly that a "photon species" is a species with `<species_name>.species_type` set to `photon` )

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -945,7 +945,7 @@ Particle initialization
 * ``<species_name>.mass`` (`float`) optional (default `NaN`)
     The mass of one `physical` particle of this species.
     If ``species_type`` is specified, the mass will be set to the physical value and ``mass`` is optional.
-    ``mass`` must be strictly positive. For massless species, use ``<species_name>.species_type``. The only allowed massless species type is ``photon``. 
+    ``mass`` must be strictly positive. For massless species, use ``<species_name>.species_type``. The only allowed massless species type is ``photon``.
 
 * ``<species_name>.xmin,ymin,zmin`` and ``<species_name>.xmax,ymax,zmax`` (`float`) optional (default unlimited)
     When ``<species_name>.xmin`` and ``<species_name>.xmax`` are set, they delimit the region within which particles are injected.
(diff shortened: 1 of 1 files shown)


Docs/source/usage/parameters.rst
```diff
@@ -1484,7 +1486,7 @@ Particle initialization
     Enables non-linear Breit-Wheeler process for this species.
     Breit-Wheeler lookup table should be either generated or loaded from disk to enable
     this process (see "Lookup tables for QED modules" section below).
-    `<species>` must be a photon species.
+    `<species>` must be a photon species (i.e., a species with ``<species_name>.species_type`` set to `photon`)
     **This feature requires to compile with QED=TRUE**
```
aeriforme:
```suggestion
    **This feature requires to compile with -DWarpX_QED=ON**

```

Docs/source/usage/parameters.rst
```diff
@@ -945,6 +945,8 @@ Particle initialization
 * ``<species_name>.mass`` (`float`) optional (default `NaN`)
     The mass of one `physical` particle of this species.
     If ``species_type`` is specified, the mass will be set to the physical value and ``mass`` is optional.
+    ``mass`` must be strictly positive. For massless species (i.e., photons for now), set ``<species_name>.species_type``
+    to the appropriate type.
```
aeriforme:
```suggestion
    ``mass`` must be strictly positive. For massless species, use ``<species_name>.species_type``. The only allowed massless species type is ``photon``. 
```

Docs/source/usage/parameters.rst
lucafedeli88: `QED=TRUE` is right for compilation relying on GNUMake ! Should we list both options ?

aeriforme: GNUMake is ["legacy", as per our documentation](https://warpx.readthedocs.io/en/latest/developers/gnumake.html). 
I think we want to focus on CMake, right @ax3l ?

EZoni: I believe it was flagged "legacy" in https://github.com/BLAST-WarpX/warpx/pull/1821, early 2021.

aeriforme: Thanks!

lucafedeli88: Ok, then let's replace it with the more modern version. I have never used GNUMake in the last 3-4 years indeed !

aeriforme (approved):

PR #6288, opened 2025-10-21
Python Control of OMP Threads

ax3l:
Sometimes one cannot set the env var `OMP_NUM_THREADS`, so Pythonic control is desired.

```py
# PICMI code ...

sim.initialize_inputs()
sim.initialize_warpx()
# or sim.initialize()
 
sim.extension.warpx.omp_threads = 2  # new!

# Advance simulation until last time step
sim.step(max_steps)
```

Same as in ImpactX (https://github.com/BLAST-ImpactX/impactx/pull/999), but hidden as usual in our sub-PICMI extension layer.

### DIFF
--- a/Source/Python/WarpX.cpp
+++ b/Source/Python/WarpX.cpp
@@ -55,6 +55,41 @@ namespace warpx {
     struct Config {};
 }
 
+namespace detail
+{
+    /** Helper Function for Property Getters
+     *
+     * This queries an amrex::ParmParse entry. This throws a
+     * std::runtime_error if the entry is not found.
+     *
+     * This handles the most common throw exception logic in WarpX instead of
+     * going over library boundaries via amrex::Abort().
+     *
+     * @tparam T type of the amrex::ParmParse entry
+     * @param prefix the prefix, e.g., "warpx" or "amr"
+     * @param name the actual key of the entry, e.g., "particle_shape"
+     * @return the queried value (or throws if not found)
+     */
+    template< typename T>
+    auto get_or_throw (std::string const & prefix, std::string const & name)
+    {
+        using V = std::decay_t<T>;
+        V value;
+
+        bool has_name = false;
+        // TODO: if array do queryarr
+        // has_name = amrex::ParmParse(prefix).queryarr(name.c_str(), value);
+        if constexpr (std::is_same_v<V, bool> || std::is_same_v<V, std::string>)
+            has_name = amrex::ParmParse(prefix).query(name.c_str(), value);
+        else
+            has_name = amrex::ParmParse(prefix).queryWithParser(name.c_str(), value);
+
+        if (!has_name)
+            throw std::runtime_error(prefix + "." + name + " is not set yet");
+        return value;
+    }
+}
+
 void init_WarpX (py::module& m)
 {
     using ablastr::fields::Direction;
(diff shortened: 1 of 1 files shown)


Source/Python/WarpX.cpp
```diff
@@ -90,6 +126,26 @@ void init_WarpX (py::module& m)
...
+                std::visit([&]( auto && omp_threads) {
+                    amrex::ParmParse pp_amrex("amrex");
+                    pp_amrex.add("omp_threads", omp_threads);
+
+                    // set the value if not "system" or "nosmt"
+                    if constexpr(std::is_same_v<std::decay_t<decltype(omp_threads)>, int>) {
+                        amrex::Print() << "Changing WarpX threads to N=" << omp_threads << "\n";
+                        amrex::OpenMP::set_num_threads(omp_threads);
+                    }
+                }, omp_threads_var);
```
ax3l, line 146: @lucafedeli88 in some cases, the setter is a bit more complex, e.g., calling an extra function to make sure an option goes into effect immediately. Should also go into the get/set param parse logic files we want to introduce.

Source/Python/WarpX.cpp
```diff
@@ -55,6 +56,41 @@ namespace warpx {
...
+        bool has_name = false;
+        // TODO: if array do queryarr
+        // has_name = amrex::ParmParse(prefix).queryarr(name.c_str(), value);
+        if constexpr (std::is_same_v<V, bool> || std::is_same_v<V, std::string>)
+            has_name = amrex::ParmParse(prefix).query(name.c_str(), value);
+        else
+            has_name = amrex::ParmParse(prefix).queryWithParser(name.c_str(), value);
+
+        if (!has_name)
+            throw std::runtime_error(prefix + "." + name + " is not set yet");
```
lucafedeli88:
```suggestion
        if constexpr (std::is_same_v<V, bool> || std::is_same_v<V, std::string>) {
            has_name = amrex::ParmParse(prefix).query(name.c_str(), value);
        }
        else {
            has_name = amrex::ParmParse(prefix).queryWithParser(name.c_str(), value);
        }

        if (!has_name) {
            throw std::runtime_error(prefix + "." + name + " is not set yet");
        }
```

I think that we enforce the `readability-braces-around-statements` rule in `clang-tidy` . Theoretically, it should have complained about these lines in CI ... I don't really know why it doesn't : I have to investigate!

Source/Python/WarpX.cpp
lucafedeli88: Yes, I agree! In this case the setter is very simple, but in other cases we will need to call extra functions to ensure that an option has the intended effect (e.g., switching an algorithm may require to change the number of guard cells). I am not sure about what is the best architecture to achieve this cleanly (e.g., does the setter call the appropriate function or do certain containers in WarpX "observe" a set of parameters in order to act as soon as they are changed).  However, as you said, in this case it is very simple!

lucafedeli88 (approved):
Thanks a lot for this PR, @ax3l ! I've left a minor suggestion concerning braces around statements, but the rest looks good to me!

Source/Python/WarpX.cpp
ax3l: I think what I wrote as a lambda here we write as a free standing C++ function, and among using it in C++ code and other getter/setters, use it also here (as a function pointer, as is common for pybind11)

PR #6293, opened 2025-10-23
Add Darwin solver 

roelof-groenewald:
This PR adds a semi-implicit Darwin field solver. 

## Overview of the model

In the Darwin approximation, the transverse displacement current is dropped from the Maxwell-Ampere equation, which removes light waves from the system while retaining the inductive (low-frequency magnetic) physics. In other words, the Darwin uses the regular Maxwell equations, except for the Maxwell-Ampere equation, which is replaced by:

$$\nabla\times\mathbf{B} = \mu_0\left(\mathbf{J} + \epsilon_0\frac{\partial\mathbf{E}_{irr}}{\partial t}\right)$$

where the electric field has been decomposed into its irrotational (i.e. curl-free) and solenoidal (i.e. divergence-free) part:

$$\mathbf{E} = \mathbf{E}_{irr} + \mathbf{E}_{sol} \qquad \mathbf{E}_{irr} = -\nabla\phi \qquad \mathbf{E}_{sol} = - \frac{\partial\mathbf{A}}{\partial t} \qquad \mathbf{B} = \nabla\times\mathbf{A}$$

using the Coulomb gauge $\nabla\cdot\mathbf{A} = 0$

## User interface

The solver is selected as an evolve scheme, on top of an electrostatic solver:

```
algo.evolve_scheme = semi_implicit_darwin
warpx.do_electrostatic = labframe    # required: the Darwin scheme adds the inductive
                                     # field on top of an electrostatic solve
algo.maxwell_solver = yee            # required (default)

amrex_gmres.relative_tolerance = 1.e-4   # magnetostatic (GMRES) solve controls
amrex_gmres.max_iterations = 1000
```

From PICMI:

```python
simulation.evolve_scheme = picmi.SemiImplicitDarwinEvolveScheme(
    linear_solver=picmi.GMRESLinearSolver(relative_tolerance=1e-4)
)
simulation.solver = picmi.ElectrostaticSolver(...)
```

Unlike a pure electrostatic run, setting this scheme does *not* disable the electromagnetic solver, since the magnetic field is still evolved. The new parameters are documented in `Docs/source/usage/parameters.rst`.

### Time staggering and updates during one timestep

| Quantity | Time level |
|---|---|
| position $\mathbf{x}$ | Updated $n \rightarrow n+1$ (integer) during a timestep |
| momentum $\mathbf{u}$ | Updated $n-1/2 \rightarrow n+1/2$ (half-integer) during a timestep |
| $\phi$, $\mathbf{E}_{irr} = -\nabla\phi$ | Computed from scratch at $n$ during a timestep |
| $\mathbf{A}$, $\mathbf{B}$ | Updated $n-1/2 \rightarrow n+1/2$ during a timestep |
| $\mathbf{E}_{sol}$ | Computed at $n$ from $\mathbf{A}$ |

### The discretized equations

In this scheme, the Maxwell equations

$$\nabla \cdot \mathbf{E} = \frac{\rho}{\epsilon_0} \qquad \nabla\times\mathbf{B} = \mu_0\left(\mathbf{J} + \epsilon_0\frac{\partial\mathbf{E}_{irr}}{\partial t}\right)$$

are rewritten with potentials (after taking the curl of the Maxwell-Ampere equation, which eliminates the $\mathbf{E}_{irr}$ terms):

$$ \nabla^2 \phi = -\frac{\rho}{\epsilon_0} \qquad -\nabla^2 (\nabla \times \mathbf{A}) = \mu_0\nabla\times\mathbf{J} \qquad (\mathrm{with}\;\nabla\cdot\mathbf{A} = 0)$$

Both equations are evaluated at $n$, and an auxiliary field $\mathbf{Z}^n$ is then introduced through $\mathbf{A}^{n+1/2}-\mathbf{A}^{n-1/2} \equiv \nabla\times\mathbf{Z}^n$, which guarantees by construction that $\mathrm{with}\;\nabla\cdot\mathbf{A} = 0$ when updating $\mathbf{A}$. Using $\mathbf{A}^n = \mathbf{A}^{n-1/2} + \nabla\times \mathbf{Z}^n/2$, the equations become:

$$ \nabla^2 \phi^n = -\frac{\rho^n}{\epsilon_0} \qquad \nabla^4\mathbf{Z}^n = 2\mu_0\nabla\times\mathbf{J}^n + 2\nabla^2\mathbf{B}^{n-1/2}$$

and the equation of motion is:

$$ m \frac{\mathbf{u}^{n+1/2} - \mathbf{u}^{n-1/2}}{\Delta t} = q(\mathbf{E}^n + q\mathbf{v}^n \times \mathbf{B}^{n-1/2}) $$

where $\mathbf{E}^n = -\nabla \phi^n - \frac{\nabla\times\mathbf{Z}^n}{\Delta t}$. (Note that the equation of motion uses $\mathbf{B}^{n}$ instead of $\mathbf{B}^{n-1/2}$, which is a deliberate approximation to keep the implicit system linear: see below.)

The system is therefore partially implicit:

- $\nabla^2 \phi^n = -\frac{\rho^n}{\epsilon_0}$ remains explicit: it can be solved directly at the beginning of a timestep (using the positions $x^n$ to deposit charge). 

- $\nabla^4\mathbf{Z}^n = 2\mu_0\nabla\times\mathbf{J}^n + 2\nabla^2\mathbf{B}^{n-1/2}$ is implicit: it requires to know $\mathbf{u}^{n+1/2}$ to compute $\mathbf{J}^n$ from the average of $\mathbf{u}^{n+1/2}$ and $\mathbf{u}^{n-1/2}$, which in turn requires to know $\mathbf{E}^n$ and therefore $\mathbf{Z}^n$.

### Solving the implicit system: Predictor/corrector, and where the mass matrix comes in

The Boris push is affine in the electric field, so writing $\boldsymbol{\Theta} = \frac{q\Delta t}{2m}\mathbf{B}^{n-1/2}$ for the usual Boris rotation vector, the velocity update splits *exactly* into a part driven by the (known) electrostatic field and a part driven by the (unknown) inductive field:

$$\mathbf{u}^{n+1/2} + \mathbf{u}^{n-1/2} = \mathbf{u}^{\ast} + \mathbf{u}^{\dagger}$$

$$\mathbf{u}^{\ast} = 2\,\frac{\mathbb{I} - \boldsymbol{\Theta}\times + \boldsymbol{\Theta}\boldsymbol{\Theta}^T}{1+\Theta^2}\left(\mathbf{u}^{n-1/2} + \frac{q\Delta t}{2m}\mathbf{E}_{irr}^n\right), \qquad \mathbf{u}^{\dagger} = 2\,\frac{\mathbb{I} - \boldsymbol{\Theta}\times + \boldsymbol{\Theta}\boldsymbol{\Theta}^T}{1+\Theta^2}\left(\frac{q\Delta t}{2m}\mathbf{E}_{sol}^n\right)$$

The predictor push computes $\mathbf{u}^{\ast}$, which is all that is knowable before the field solve. Taking the charge-weighted moments of the two terms and using $\mathbf{u}^n = (\mathbf{u}^{n+1/2}+\mathbf{u}^{n-1/2})/2$ gives the current at the integer time,

$$2\mathbf{J}^n = \mathbf{J}^{\ast} + \mathbf{J}^{\dagger}, \qquad \mathbf{J}^{\ast} = 2\hat{\mathbf{J}}^n = \sum_p q_p w_p \mathbf{u}_p^{\ast}\,S(\mathbf{x}_p^n)$$

**This is where the mass matrix enters.** Because $\mathbf{u}^{\dagger}$ is *linear* in $\mathbf{E}_{sol}^n$, the moment $\mathbf{J}^{\dagger}$ can be written in closed form as a linear operator acting on the still-unknown field, rather than requiring an iterative solve:

$$\mathbf{J}^{\dagger} = \frac{\Delta t}{\mu_0}\boldsymbol{\chi}\,\mathbf{E}_{sol}^n, \qquad \boldsymbol{\chi}_{ii'} = \sum_s\sum_{p\in s}\frac{\mu_0 q_p^2 w_p}{m_p}\,\frac{\mathbb{I} + \boldsymbol{\Theta}_p\boldsymbol{\Theta}_p^T - \boldsymbol{\Theta}_p\times}{1+\Theta_p^2}\,S_i(\mathbf{x}_p)S_{i'}(\mathbf{x}_p)$$

$\boldsymbol{\chi}$ is the mass matrix (susceptibility): a grid-point-to-grid-point tensor accumulated over particles in the same sweep as the current deposition. It is exactly the linear response of the deposited current to the field the solve is about to produce.

Finally, substituting $\mathbf{E}_{sol}^n = -\Delta\mathbf{A}^n/\Delta t = -\nabla\times\mathbf{Z}^n/\Delta t$ turns the response term into an operator on the unknown $\mathbf{Z}^n$,

$$\mathbf{J}^{\dagger} = -\frac{1}{\mu_0}\boldsymbol{\chi}\,\nabla\times\mathbf{Z}^n$$

so that $2\mu_0\nabla\times\mathbf{J}^n = 2\mu_0\nabla\times\hat{\mathbf{J}}^n - \nabla\times(\boldsymbol{\chi}\,\nabla\times\mathbf{Z}^n)$. Moving the response term to the left-hand side produces the $\nabla\times(\boldsymbol{\chi}\,\nabla\times\mathbf{Z}^n)$ term of the field equation above, leaving the system **linear** — hence "semi-implicit": one linear solve per step, with no Newton iteration. In the code, that term is evaluated by `ApplySusceptibility`, which applies the deposited mass matrices to $\nabla\times\mathbf{Z}$ with the appropriate $\mu_0/\Delta t$ scaling.

### `SemiImplicitDarwin::OneStep`

At the start of a step the fields hold $\mathbf{E} = -\nabla\phi^n$ (left over from the electrostatic solve at the end of the previous step) and $\mathbf{B}^{n-1/2}$; the particles hold $\mathbf{x}^n$ and $\mathbf{u}^{n-1/2}$. The step then proceeds as:

1. **Predictor (electrostatic) push.** `PushP` applies a full Boris push with the electrostatic field only, giving $\hat{\mathbf{u}}^{n+1/2}$. `PrepareVelocitiesForCurrentDeposition` then replaces `u` by the time-centred average $(\hat{\mathbf{u}}^{n+1/2} + \mathbf{u}^{n-1/2})/2$ and stashes $\hat{\mathbf{u}}^{n+1/2}$ in `u_n` for later.

2. **Current and mass-matrix deposition** (`AccumulateCurrentAndSusceptibility`). $\hat{\mathbf{J}}^n$ is deposited from those time-centred velocities, and $\boldsymbol{\chi}$ from the same `DepositMassMatrices` machinery already used by the implicit EM solvers.

3. **Magnetostatic solve.** `CalculateSourceVector` builds the right-hand side $2\mu_0\nabla\times\hat{\mathbf{J}}^n + 2\nabla^2\mathbf{B}^{n-1/2}$, and `amrex::GMRES` solves the fourth-order equation for $\mathbf{Z}^n$. The operator $\nabla^4 + \nabla\times(\boldsymbol{\chi}\,\nabla\times\,\cdot\,)$ is applied matrix-free by `ComputeRHS`.

4. **Inductive field.** `ComputeInductiveEfromdA` sets $\mathbf{E}_{sol}^n = -\Delta\mathbf{A}^n/\Delta t = -\nabla\times\mathbf{Z}^n/\Delta t$, overwriting `Efield_fp`, which from here on holds the inductive component only.

5. **Corrector (solenoidal) push.** The velocities are zeroed and `PushP` is called again with the inductive field alone, which isolates $\delta\mathbf{u}(\mathbf{E}_{sol}^n)$; `FinishVelocityUpdate` then adds back $\hat{\mathbf{u}}^{n+1/2}$ from `u_n` to form the complete $\mathbf{u}^{n+1/2}$. Positions are advanced to $\mathbf{x}^{n+1}$.

6. **Magnetic field update.** `EvolveB` advances $\mathbf{B}^{n-1/2}$ to $\mathbf{B}^{n+1/2}$ with $\partial\mathbf{B}/\partial t = -\nabla\times\mathbf{E}$. Since `Efield_fp` holds $-\nabla\times\mathbf{Z}^n/\Delta t$, this is equivalent to $\mathbf{A}^{n+1/2} = \mathbf{A}^{n-1/2} + \nabla\times\mathbf{Z}^n$ followed by $\mathbf{B} = \nabla\times\mathbf{A}$, without ever storing $\mathbf{A}$.

The electrostatic solve for $\phi^{n+1}$ is then performed by the existing electrostatic branch of `WarpX::Evolve`, which leaves `Efield_fp` holding $-\nabla\phi^{n+1}$ ready for the next step.

## Code structure

- **`Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.{H,cpp}`** — the new solver, an `ImplicitSolver` subclass. `OneStep` drives the sequence above; `ComputeRHS` evaluates the linear operator for GMRES.
- **`Source/NonlinearSolvers/LinearFunctionMF.H`** — matrix-free linear operator that lets `amrex::GMRES` call back into `ComputeRHS`.
- **Finite-difference operators** — `ComputeCurlB.cpp` (curl of a B-staggered field, output on E staggering), `ComputeLaplacian.cpp` (scalar and vector Laplacian, plus a single-pass vector bi-Laplacian), and the fourth-derivative stencils `Dxxxx`/`Dyyyy`/`Dzzzz`/`Dxxyy`/`Dyyzz`/`Dxxzz` in `CartesianYeeAlgorithm.H`.
- **`Source/Fields.H`** — new `dA_fp` vector field holding the vector-potential increment over a step.
- **`Source/Parallelization/WarpXComm.cpp`** — `WarpX::SyncMassMatrices()`, the boundary summation of the deposited mass matrices.
- **`Source/Diagnostics/ComputeDiagFunctors/DarwinEfieldFunctor.{H,cpp}`** — since `Efield_fp` holds only the electrostatic component at diagnostic time, this functor reconstructs the full field as `E_es + (-dA/dt)` for output; wired up in `FullDiagnostics.cpp`.
- **`Source/WarpX.cpp` / `WarpXEvolve.cpp`** — the new `Semi_Implicit_Darwin` evolve scheme, its input validation, and the separation of the hybrid-PIC branch from the electrostatic branch of the PIC loop (Darwin needs the electrostatic solve without the accompanying B-field reset).
- **`Python/pywarpx/picmi.py`** — `SemiImplicitDarwinEvolveScheme`.

## Tests

The existing `Examples/Tests/ohm_solver_em_modes` directory has been generalised into `Examples/Tests/magnetized_plasma_modes`, with a single input script that runs the same magnetized-plasma EM-mode setup with either solver (`--darwin` / `--ohm`). Two new tests were added there:

- `test_1d_darwin_solver_em_modes_picmi`
- `test_2d_darwin_solver_em_modes_es_picmi` (Darwin combined with the effective-potential electrostatic solver)

The implementation was verified by reproducing the dispersion of left- and right-hand circularly polarized Alfvén waves propagating parallel to an applied magnetic field (see comment further below).


roelof-groenewald:
At present the GMRES solver converges very slowly as tested on the above 1d problem. Here is an example output from a random timestep:
```
GMRES: iter = 0, residual = 23.35809154, 1 (rel.)
GMRES: iter = 1, residual = 9.174234898, 0.3927647464 (rel.)
calling ComputeRHS...
GMRES: iter = 2, residual = 6.244034034, 0.2673178168 (rel.)
calling ComputeRHS...
GMRES: iter = 3, residual = 5.343642676, 0.228770517 (rel.)
calling ComputeRHS...
GMRES: iter = 4, residual = 5.272462275, 0.2257231618 (rel.)
calling ComputeRHS...
GMRES: iter = 5, residual = 5.271985566, 0.2257027531 (rel.)
calling ComputeRHS...
GMRES: iter = 6, residual = 5.271329889, 0.2256746824 (rel.)
calling ComputeRHS...
GMRES: iter = 7, residual = 5.270875834, 0.2256552435 (rel.)
calling ComputeRHS...
GMRES: iter = 8, residual = 5.270736453, 0.2256492764 (rel.)
calling ComputeRHS...
GMRES: iter = 9, residual = 5.270618621, 0.2256442318 (rel.)
calling ComputeRHS...
GMRES: iter = 10, residual = 5.269625198, 0.2256017017 (rel.)
calling ComputeRHS...
GMRES: iter = 11, residual = 5.249355966, 0.224733941 (rel.)
calling ComputeRHS...
GMRES: iter = 12, residual = 5.09618176, 0.2181762903 (rel.)
```

JustinRayAngus:
How stiff is your problem? What is omega_pe*dt?

Could be many issues. But one thing you can try is to use a small timestep like for explicit methods and then you should see convergence in 3-5 iterations.

roelof-groenewald:
Thanks @JustinRayAngus! I'll see how it does with smaller timesteps. 
In this case omega-pe*dt was 10, but it's unclear to me that it is the appropriate stiffness measure in this problem since the electrostatic solve (which would usually constrain dt relative to omega_pe) is done separately.

roelof-groenewald:
The model is now very nearly fully inside WarpX; the only remaining step executed from Python is to apply periodic boundaries to the mass-matrices via:

```
        #################################################################
        # Fix susceptibility for periodic boundaries
        #################################################################

        S_xx = fields.MultiFabWrapper(mf_name="MassMatrices_X", idir=0, level=0)
        S_xy = fields.MultiFabWrapper(mf_name="MassMatrices_X", idir=1, level=0)

        S_xx[0, 0] += S_xx[-1, 0]
        S_xx[-1, 2] += S_xx[0, 2]
        S_xx[0, 1] += S_xx[-1, 1]
        S_xx[-1, 1] = S_xx[0, 1]

        S_xy[0, 0] += S_xy[-1, 0]
        S_xy[-1, 2] += S_xy[0, 2]
        S_xy[0, 1] += S_xy[-1, 1]
        S_xy[-1, 1] = S_xy[0, 1]

        S_yx = fields.MultiFabWrapper(mf_name="MassMatrices_Y", idir=0, level=0)
        S_yy = fields.MultiFabWrapper(mf_name="MassMatrices_Y", idir=1, level=0)

        S_yx[0, 0] += S_yx[-1, 0]
        S_yx[-1, 2] += S_yx[0, 2]
        S_yx[0, 1] += S_yx[-1, 1]
        S_yx[-1, 1] = S_yx[0, 1]

        S_yy[0, 0] += S_yy[-1, 0]
        S_yy[-1, 2] += S_yy[0, 2]
        S_yy[0, 1] += S_yy[-1, 1]
        S_yy[-1, 1] = S_yy[0, 1]
```

Examples/Tests/darwin_solver_EM_modes/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+# Add tests (alphabetical order) ##############################################
+#
+
+add_warpx_test(
+    test_1d_darwin_solver_em_modes_picmi  # name
+    1  # dims
+    1  # nprocs
```
roelof-groenewald: There is some remaining bug in the implementation - if I run with more than 1 processor (or more than 1 tile) the MS solver does not converge.

roelof-groenewald:
The GMRES solver does not converge very well in 1d (in 2d it does not converge at all).

However, energy conservation (in the 1d) case improves substantially with increased number of GMRES iterations:
image
("superLU" shows the result when directly solving the MS equation via Python)

As suggested by @JustinRayAngus, the GMRES convergence is substantially improved at smaller timestep, which also results in better energy conservation:
image

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,863 @@
...
+        const auto& ba_phi = m_WarpX->m_fields.get(FieldType::phi_fp, lev)->boxArray();
+        m_WarpX->m_fields.alloc_init(FieldType::xi_fp, lev, ba_phi, dm, 1, amrex::IntVect(AMREX_D_DECL(1,1,1)), 0.0_rt );
+    }
+
+    // Initialize a projection divergence cleaner for the current density
+    // m_div_cleaner("current_fp", true);
+
+    // Define WarpXSolverVec instances for the MS equation solution (dA) and
+    // source
+    m_dA.Define( m_WarpX, "dA_fp", "xi_fp" );
```
JustinRayAngus: I think you had mentioned previously that xi_fp is solved independently of dA_fp. Is that still the case? If so, then this may not be proper setup for the WarpXSolverVector. Is the idea to use GMRES to solve for both dA and xi? Or just solve for dA?

roelof-groenewald: Thanks for the comment, @JustinRayAngus. In this implementation, `xi` is a scalar field used to divergence clean the predicted current coming from the susceptibility (mass matrices). So ideally it would be solved simultaneously with `dA`. Basically, there's an additional set of equations along with the magnetostatic field equations to force the Coulomb gauge.
The electrostatic potential is solved separately at a later time.

JustinRayAngus: Thanks for clearing that up. Makes sense to me now.

WeiqunZhang:
When I run the test in debug mode, I get
```
-----------------------------------------------------------
--------- SEMI IMPLICIT DARWIN SOLVER PARAMETERS ----------
-----------------------------------------------------------
Linear solver (amrex_gmres) verbose:            2
Linear solver (amrex_gmres) restart length:     30
Linear solver (amrex_gmres) max iterations:     32
Linear solver (amrex_gmres) relative tolerance: 0.0001
Linear solver (amrex_gmres) absolute tolerance: 0
-----------------------------------------------------------

For full input parameters, see the file: warpx_used_inputs

--- INFO    : Writing plotfile diags/field_diag000000
STEP 1 starts ...
amrex::Abort::0:: (257,0,0,0) is out of bound (-1:256,0:0,0:0,0:0) !!!
```

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,863 @@
...
+    BL_PROFILE("SemiImplicitDarwin::CalculateSourceVector()");
+
+    const int lev = 0;
+
+    // Zero out existing source values
+    m_source.zero();
+
+    // Divergence clean J
+    // Initialize a projection divergence cleaner for the current density
+    warpx::initialization::ProjectionDivCleaner m_div_cleaner("current_fp", true);
```
WeiqunZhang: This causes out-of-bound errors. `ProjectionDivCleaner` assumes that index type is B type, not the current type.

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,863 @@
...
+            const amrex::IntVect ncomp_xy = m_ncomp_xy;
+            const amrex::IntVect ncomp_xz = m_ncomp_xz;
+            const amrex::IntVect ncomp_yx = m_ncomp_yx;
+            const amrex::IntVect ncomp_yy = m_ncomp_yy;
+            const amrex::IntVect ncomp_yz = m_ncomp_yz;
+            const amrex::IntVect ncomp_zx = m_ncomp_zx;
+            const amrex::IntVect ncomp_zy = m_ncomp_zy;
+            const amrex::IntVect ncomp_zz = m_ncomp_zz;
+
+            amrex::ParallelFor(
```
WeiqunZhang: There is out-of-bound errors here too.

Source/FieldSolver/FiniteDifferenceSolver/ComputeGradient.cpp
```diff
@@ -0,0 +1,181 @@
...
+        auto const n_coefs_y = static_cast<int>(m_stencil_coefs_y.size());
+        Real const * const AMREX_RESTRICT coefs_z = m_stencil_coefs_z.dataPtr();
+        auto const n_coefs_z = static_cast<int>(m_stencil_coefs_z.size());
+
+        // Extract tileboxes for which to loop
+        Box const& tbx  = mfi.tilebox(out_field[0]->ixType().toIntVect());
+        Box const& tby  = mfi.tilebox(out_field[1]->ixType().toIntVect());
+        Box const& tbz  = mfi.tilebox(out_field[2]->ixType().toIntVect());
+
+        // Calculate the vector Laplacian of the input field (G)
```
roelof-groenewald:
```suggestion
        // Calculate the gradient of the input field (G)
```

roelof-groenewald:
The underlying MS equation being solved in the Darwin model has been changed. Now instead of solving two second-order equations, a single fourth-order equation is solved. The out-of-bounds issue pointed out by @WeiqunZhang has also been fixed, which also solved the previously discussed difficulty of the GMRES solver to converge. Now using relative tolerance of 1e-6 works fine (takes only a few hundred GMRES iterations), and the model's energy conservation is excellent (results for 1d test case):
image

roelof-groenewald:
@JustinRayAngus I believe this PR is finally ready. When you have a chance it would be great if you could look over the way I couple to the existing implicit functionality to confirm that I'm doing it in the intended manner. Thanks in advance!

Python/pywarpx/picmi.py
```diff
@@ -2782,12 +2808,13 @@ def applied_field_initialize_inputs(self):
...
-            pywarpx.warpx.add_new_group_attr(
-                "projection_div_cleaner", "rtol", self.div_cleaner_rtol
-            )
+            if self.do_initial_div_cleaning:
+                pywarpx.warpx.add_new_group_attr(
+                    "projection_div_cleaner", "atol", self.div_cleaner_atol
+                )
+                pywarpx.warpx.add_new_group_attr(
+                    "projection_div_cleaner", "rtol", self.div_cleaner_rtol
+                )
```
roelof-groenewald: This could be a separate PR.

Examples/Tests/darwin_solver_EM_modes/analysis.py
```diff
@@ -0,0 +1,323 @@
+#!/usr/bin/env python3
```
RemiLehe, line 1: My understanding (which might be wrong) is that `darwin_solver_EM_modes` and `ohm_solver_EM_modes` test the same physics (i.e. plasma modes affected by a background magnetic field).

If that is the case, to reduce duplication, I would suggest consolidating them into a single folder `magnetized_plasma_modes`, and run both the Darwin and Ohm solver with a set of consolidated script:

- Consolidated `analysis.py`: it would have a flag (e.g. `--analyze_darwin_sim` or `--analyze_ohm_sim`) and would share most of the analysis. Of course some of the analysis would be specific to one of the two solvers (e.g. plotting/checking the electron cyclotron mode, which, in my understanding, should only exist with the Darwin solver)

- Consolidated `inputs_test_.... .py`: again with a flag (e.g. `--darwin` or `--ohm`) since my impression is that most of the script is common between these two solvers.

Docs/source/developers/particles.rst
```diff
@@ -83,7 +83,7 @@ Main functions
```
RemiLehe, line 85: Could you add documentation for the new parameters in `parameters.rst`?
Do I understand correctly that the user needs to set `evolve_scheme = semi_implicit_darwin` and (separately) `do_electrostatic` do a non-`none` value, as well as `maxwell_solver = yee`?

Maybe it would also be useful to add a summary of the algorithm in https://warpx.readthedocs.io/en/latest/theory/intro.html#models-algorithms

Source/Utils/WarpXAlgorithmSelection.H
```diff
@@ -171,30 +172,26 @@ AMREX_ENUM(PushType,
...
 /** \brief For binary collision algorithms, the strategy to determine
  *  the scattering angle in the center of mass frame */
-AMREX_ENUM(
-    ScatteringAngleModel,
-    Forward,    /**< Scattering angle is zero, i.e., products have the same direction as the incident particle in the center of mass frame */
-    Backward,   /**< Scattering angle is pi, i.e., products have the opposite direction of the incident particle in the center of mass frame */
-    Isotropic,  /**< Scattering angle is random, i.e., products are scattered isotropically in the center of mass frame */
-    Default = Isotropic  /**< Isotropic */
-);
+AMREX_ENUM(ScatteringAngleModel,
```
RemiLehe, line 189: Could you revert the change in indentation here?

Source/Utils/WarpXAlgorithmSelection.H
```diff
@@ -171,30 +172,26 @@ AMREX_ENUM(PushType,
...
 
 /** \brief For advanced collision algorithms that split the particle push in substeps */
-AMREX_ENUM(
-    MomentumPushType,
-    Full,
-    FirstHalf,
-    SecondHalf,
-    Default = Full
-);
+AMREX_ENUM(MomentumPushType,
```
RemiLehe, line 181: Could you revert the change in indentation here?

Source/Utils/WarpXAlgorithmSelection.H
```diff
@@ -171,30 +172,26 @@ AMREX_ENUM(PushType,
...
 /** \brief For advanced collision algorithms that split the particle push in substeps */
-AMREX_ENUM(
-    PositionPushType,
-    None,
-    Full,
-    Default = Full
-);
+AMREX_ENUM(PositionPushType,
+           Full,
+           FirstHalf,
```
RemiLehe: It seems that `FirstHalf/SecondHalf` are never used here (unless I am mistaken). 
Could you revert the change?

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -286,43 +287,54 @@ WarpX::Evolve (int numsteps)
             ExecutePythonCallback("aftercollisions");
         }
 
-        // Field solve step for electrostatic or hybrid-PIC solvers
-        if( electrostatic_solver_id != ElectrostaticSolverAlgo::None ||
-            electromagnetic_solver_id == ElectromagneticSolverAlgo::HybridPIC )
+        // Field solve step for electrostatic, hybrid-PIC, or Darwin solvers
```
RemiLehe:
```suggestion
        // Field solve step for electrostatic or Darwin solvers
```

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -286,43 +287,54 @@ WarpX::Evolve (int numsteps)
...
+
+            // The external fields are added back on to the fine patch fields
+            // (which are overwritten by electrostatic / magnetostatic solvers),
+            // unless Darwin is used in which case the "external" fields function
+            // as initial conditions (like for the electromagnetic solvers)
+            // The net fields are then the sum of the field solutions and any
+            // external field.
+            if (evolve_scheme != EvolveScheme::Semi_Implicit_Darwin) {
                 for (int lev = 0; lev <= max_level; ++lev) {
                     AddExternalFields(lev);
```
RemiLehe, line 323: Maybe to make the logic more generic, we should split this function into:
- `AddExternalEFields`
- `AddExternalBFields`
and then have:
```
// If the fields have been reset (rather than updated) as part of the field solve,
// add back the external fields
if (reset_E_field) AddExternalEFields(lev);
if (reset_B_field) AddExternalBFields(lev);
```

RemiLehe: This (along with the modified interface `ComputeSpaceChargeField( reset_E_field, reset_B_field );`) could be a small, separate PR. 

In that case, we could even have (as part of this small PR) `reset_B_field = false` for the `labframe` and `labframe-effective-potential`, which would save a bit of computation in these cases.

Source/WarpX.cpp
```diff
@@ -3630,7 +3645,7 @@ WarpX::getFieldDotMaskPointer ( FieldType field_type, int lev, ablastr::fields::
             ::SetDotMask( Afield_dotMask[lev][dir], m_fields.get("vector_potential_fp", dir, lev), periodicity);
             return Afield_dotMask[lev][dir].get();
         case FieldType::phi_fp :
-            ::SetDotMask( phi_dotMask[lev], m_fields.get("phi_fp", dir, lev), periodicity);
+            ::SetDotMask( phi_dotMask[lev], m_fields.get("phi_fp", lev), periodicity);
```
RemiLehe, line 3689: This change seems unrelated to the PR, is that correct?
If yes, should it be a separate PR?

Source/NonlinearSolvers/CurlCurlMLMGPC.H
```diff
@@ -13,6 +13,7 @@
 #include "Preconditioner.H"
 
 #include <ablastr/fields/MultiFabRegister.H>
+#include <ablastr/warn_manager/WarnManager.H>
```
RemiLehe, line 16:
```suggestion
```
Could you instead `include` the `WarnManager` directly in the `.cpp` file that calls the warning?

Source/NonlinearSolvers/MatrixPC.H
```diff
@@ -13,6 +13,7 @@
 #include "Preconditioner.H"
 
 #include <ablastr/fields/MultiFabRegister.H>
+#include <ablastr/warn_manager/WarnManager.H>
```
RemiLehe, line 16:
```suggestion
```
Could you instead include this directly in the `.cpp` file that calls the warning?

Source/Initialization/WarpXInitData.cpp
```diff
@@ -900,8 +900,9 @@ WarpX::InitData ()
              has_boundary_potential)
             && WarpX::electromagnetic_solver_id != ElectromagneticSolverAlgo::HybridPIC)
         {
-            bool const reset_fields = false; // Do not erase previous user-specified values on the grid
-            ComputeSpaceChargeField(reset_fields);
+            bool const reset_E_field = false; // Do not erase previous user-specified values on the grid
+            bool const reset_B_field = false; // Do not erase previous user-specified values on the grid
+            ComputeSpaceChargeField(reset_E_field, reset_B_field);
```
RemiLehe, line 907: As discussed in `WarpXEvolve.cpp`, this could be a separate PR.

Source/FieldSolver/WarpXSolveFieldsES.cpp
```diff
@@ -13,18 +13,20 @@
 
 #include <ablastr/profiler/ProfilerWrapper.H>
 
-void WarpX::ComputeSpaceChargeField (bool const reset_fields)
+void WarpX::ComputeSpaceChargeField (bool const reset_E_field, bool const reset_B_field)
```
RemiLehe, line 16: As discussed in `WarpXEvolve.cpp`, this could be a separate PR.

Python/pywarpx/picmi.py
RemiLehe: Sounds good.

Could you also include the `nonlinear_solver.linear_solver = "petsc_ksp"` in that same PR? The PR would be about minor fixes to `picmi.py`

Source/Diagnostics/ComputeDiagFunctors/DarwinEfieldFunctor.cpp
```diff
@@ -0,0 +1,54 @@
...
+        m_mf_Efield != nullptr && m_mf_dA != nullptr,
+        "m_mf_Efield/m_mf_dA can't be nullptr.");
+
+    auto& warpx = WarpX::GetInstance();
+
+    // SemiImplicitDarwin::OneStep() leaves Efield_fp (and therefore Efield_aux,
+    // which is derived from it) holding only the electrostatic component plus
+    // any external field at this point in the step; the inductive component
+    // (E = -dA/dt) is recovered from dA_fp, which
+    // SemiImplicitDarwin::UpdateEandAfromdA() computes once per step and
```
RemiLehe:
```suggestion
    // SemiImplicitDarwin::UpdateEfromdA() computes once per step and
```

Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -869,7 +870,18 @@ FullDiagnostics::InitializeFieldFunctors (int lev)
     for (int comp=0; comp<nvar; comp++){
         for (int idir=0; idir < 3; idir++) {
             if        ( m_varnames[comp] == "E"+field_names[idir] ){
-                m_all_field_functors[lev][comp] = std::make_unique<CellCenterFunctor>(warpx.m_fields.get(FieldType::Efield_aux, Direction{idir}, lev), lev, m_crse_ratio);
+                if (warpx.evolve_scheme == EvolveScheme::Semi_Implicit_Darwin) {
+                    // Efield_aux (like Efield_fp, which it aliases at this level)
+                    // only holds the electrostatic E-field at this point in the
+                    // step; recover the full field using dA_fp (see
+                    // DarwinEfieldFunctor and SemiImplicitDarwin::UpdateEandAfromdA).
```
```suggestion
                    // DarwinEfieldFunctor and SemiImplicitDarwin::UpdateEfromdA).
```

RemiLehe (changes requested):
Thanks for this very exciting PR!

I took a first, very high-level pass, without diving into the details of the solver yet, with some suggestions. Let me know if they make sense.

Source/Utils/WarpXAlgorithmSelection.H
roelof-groenewald: I changed these indentations to match the rest of the file's standard. It didn't seem like something worth a separate PR, but also worth fixing. Should I open a separate PR or just leave it to be inconsistent?

Source/WarpX.cpp
roelof-groenewald: It's a bug fix that I happened to spot during the previous version of the Darwin model. `phi_fp` is a scalar, so it doesn't have a direction. I guess nothing in the code actually uses `phi_dotMask`. I can move this to a new PR. Or we can remove it completely?

Source/Evolve/WarpXEvolve.cpp
roelof-groenewald: Opened in #7066 for the new flags. I'll keep the `AddExternalFields` split for yet another PR since that is not strictly needed for the current purposes yet (although would be a nice, cheap performance boost).

Examples/Tests/darwin_solver_EM_modes/analysis.py
roelof-groenewald: Great suggestion! This change has now been made.

Source/NonlinearSolvers/MatrixPC.H
roelof-groenewald: There isn't actually a corresponding `.cpp` file. The warning is called on line 312 of this file. But I can move it to a different PR.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -153,6 +153,7 @@ WarpX::Evolve (int numsteps)
     ABLASTR_PROFILE("WarpX::Evolve()");
 
     using ablastr::fields::Direction;
+    using warpx::fields::FieldType;
```
roelof-groenewald:
```suggestion
```

Examples/Tests/magnetized_plasma_modes/inputs_test_em_modes_picmi.py
```diff
@@ -24,71 +24,134 @@
...
+    hybrid) solver. The analysis script (in this same directory) analyzes the
+    output field data for EM modes. This input is based on the EM modes tests
+    as described by Munoz et al. (2018) and tests done by Scott Nicks at TAE
+    Technologies.
+    """
 
-    def __init__(self, test, dim, B_dir, verbose, use_rkf45):
+    # Darwin solver parameters
+    DARWIN_PARAMS = dict(
+        B0=0.15,  # Initial magnetic field strength (T)
```
RemiLehe, line 59: Is intentional that `B0`, `beta` and `vA_over_c` are slightly different for these two tests? Or could they be made similar?

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -288,47 +289,54 @@ WarpX::Evolve (int numsteps)
...
+                // Call Magnetostatic Solver to solve for the vector potential A and compute the
+                // B field.  Time varying A contribution to E field is neglected.
+                // This is currently a lab frame calculation.
+                ComputeMagnetostaticField();
+            }
+
+            // The external fields are added back on to the fine patch fields
+            // (which are overwritten by electrostatic / magnetostatic solvers),
+            // unless Darwin is used in which case the "external" fields function
+            // as initial conditions (like for the electromagnetic solvers)
```
RemiLehe: The `in which case` clause seems incomplete. Also, maybe this comment can generally be made slightly more concise.

.clang-tidy
```diff
@@ -10,6 +10,7 @@ Checks: '
         -cert-err58-cpp,
         -cert-int09-c,
     clang-analyzer-*,
+        -clang-analyzer-optin.core.EnumCastOutOfRange,
```
RemiLehe, line 13: Is this change intentional? Should it be a separate PR?

Source/FieldSolver/FiniteDifferenceSolver/FiniteDifferenceAlgorithms/CartesianYeeAlgorithm.H
```diff
@@ -244,6 +244,175 @@ struct CartesianYeeAlgorithm {
...
+        amrex::Real const * const coefs_x, int const /*n_coefs_x*/,
+        int const i, int const j, int const k, int const ncomp=0 ) {
+
+        using namespace amrex;
+#if (defined WARPX_DIM_1D_Z)
+        amrex::ignore_unused(F, coefs_x, i, j, k, ncomp);
+        return 0._rt; // 1D Cartesian: derivative along x is 0
+#else
+        amrex::Real const inv_dx2 = coefs_x[0]*coefs_x[0];
+        amrex::Real const inv_dx4 = inv_dx2*inv_dx2;
```
RemiLehe: Could we do
```suggestion
        amrex::Real const inv_dx4 = amrex::Math::powi<4>(coefs_x[0]);
```
(and similar changes throughout this file)

Source/FieldSolver/FiniteDifferenceSolver/ComputeCurlB.cpp
```diff
@@ -0,0 +1,203 @@
...
+#   include "FiniteDifferenceAlgorithms/CartesianYeeAlgorithm.H"
+#   include "FiniteDifferenceAlgorithms/CartesianNodalAlgorithm.H"
+#endif
+
+#include "Utils/TextMsg.H"
+#include "WarpX.H"
+
+using namespace amrex;
+
+void FiniteDifferenceSolver::ComputeCurlB (
```
RemiLehe, line 25: This function is called `ComputeCurlB` (and the documentation discusses computing the curl of B and overwriting E) but this notation is maybe confusing because, in WarpX, it is only ever used to compute the curl of Z (and overwriting dA) ; should the name be updated to `ComputeCurlZ`?

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,969 @@
+/* Copyright 2025 The WarpX Community
+ *
+ * This file is part of WarpX.
+ *
+ * Authors: Roelof Groenewald (TAE Technologies)
```
RemiLehe:
```suggestion
/* Copyright 2026 The WarpX Community
 *
 * This file is part of WarpX.
 *
 * Authors: Roelof Groenewald (Realta Fusion)
```

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,969 @@
...
+
+    // Parse implicit solver parameters
+    m_use_mass_matrices = true;
+    m_use_mass_matrices_pc = false;
+    m_use_mass_matrices_jacobian = true;
+    m_nlsolver_type = NonlinearSolverType::none;
+    m_max_particle_iterations = 1;
+
+    // Get the linear solver input parameters
+    const amrex::ParmParse pp_l(amrex::getEnumNameString(m_linear_solver_type));
```
RemiLehe, line 65: My understanding is that this parses input blocks of the type
```
amrex_gmres.verbose_int = 2
amrex_gmres.restart_length = 30
amrex_gmres.max_iterations = 1000
amrex_gmres.relative_tolerance = 1.e-4
amrex_gmres.absolute_tolerance = 0.0
```
If this is correct, it would be good to have a way to document it.

This applies also to the Newton solver, so this potentially be a separate PR.

RemiLehe, line 65:
```suggestion
    const amrex::ParmParse pp_l("amrex_gmres");
```
Given that only this solver is supported, I would also remove `m_linear_solver_type` from `SemiImplicitSolver.H` for now (we can always reintroduce it if other linear solvers are supported).

RemiLehe: Or otherwise, if we want to keep `m_linear_solver_type`, then I would add an `if` condition around
```
m_linear_solver = std::make_unique<AMReXGMRES<WarpXSolverVec,LinearFunctionMF<WarpXSolverVec,SemiImplicitDarwin>>>();
```
below.

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,969 @@
...
+    pp_l.query("verbose_int",         m_linsol_verbose_int);
+    pp_l.query("restart_length",      m_linsol_restart_length);
+    pp_l.query("absolute_tolerance",  m_linsol_atol);
+    pp_l.query("relative_tolerance",  m_linsol_rtol);
+    pp_l.query("max_iterations",      m_linsol_maxits);
+
+    // Define the linear function - Note we could use JacobianFunctionMF if we
```
RemiLehe: How difficult would it be to as part of this PR?
Would it allow us to remove the new code for `LinearFunctionMF`?

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,969 @@
...
+    m_WarpX->SaveParticlesAtImplicitStepStart();
+
+    // Push particle velocities with E_fp (which currently just contains -∇ phi since
+    // the E-field was cleared during the last Poisson solve)
+    for (int lev = 0; lev <= finest_level; ++lev)
+    {
+        m_WarpX->GetPartContainer().PushP(
+            lev,
+            m_dt,
+            *m_WarpX->m_fields.get(FieldType::Efield_fp, Direction{0}, lev),
```
RemiLehe, line 138: Should this be `Efield_aux`, in general?

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,969 @@
...
+            amrex::elemwiseMax(Zvec[lev][0]->nGrowVect(), amrex::IntVect(2));
+        m_Zscratch_x.define(Zvec[lev][0]->boxArray(), Zvec[lev][0]->DistributionMap(),
+                            Zvec[lev][0]->nComp(), biharmonic_ng);
+        m_Zscratch_y.define(Zvec[lev][1]->boxArray(), Zvec[lev][1]->DistributionMap(),
+                            Zvec[lev][1]->nComp(), biharmonic_ng);
+        m_Zscratch_z.define(Zvec[lev][2]->boxArray(), Zvec[lev][2]->DistributionMap(),
+                            Zvec[lev][2]->nComp(), biharmonic_ng);
+    }
+
+    // Parse implicit solver parameters
```
RemiLehe:
```suggestion
    // Set parameters used by `InitializeMassMatrices`
```

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,969 @@
...
+    m_use_mass_matrices = true;
+    m_use_mass_matrices_pc = false;
+    m_use_mass_matrices_jacobian = true;
+    m_nlsolver_type = NonlinearSolverType::none;
+    m_max_particle_iterations = 1;
```
RemiLehe: My understanding is that these variables are not even used, so in principle, we don't even need to set them, is that correct?

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,969 @@
...
+            *m_WarpX->m_fields.get(FieldType::Bfield_fp, Direction{0}, lev),
+            *m_WarpX->m_fields.get(FieldType::Bfield_fp, Direction{1}, lev),
+            *m_WarpX->m_fields.get(FieldType::Bfield_fp, Direction{2}, lev),
+            MomentumPushType::Full
+        );
+    }
+
+    // Prepare current deposition by setting particle velocities to twice the
+    // t = n velocity values (with just the ES acceleration applied for the
+    // advanced velocity)
```
RemiLehe: My understanding is that `twice the t = n velocity values` is inaccurate here, since this does:
```
ux[ip] = 0.5_prt*(ux_es + ux_n[ip]);
```

In addition, I think it would be good if the comment also mentions that this updates the saved velocity, i.e.
```
ux_n[ip] = ux_es;
```

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,969 @@
...
+    PrepareCurrentDeposition();
```
RemiLehe: Maybe this should be rename `PrepareVelocitiesForCurrentDeposition`, since this changes the particle velocities (not the current, which is what someone might read from the name of the function)

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,969 @@
...
+    // AMReX's GMRES::getStatus() returns 0 on convergence and a positive
+    // value (e.g. 1 if the iteration count was exceeded) otherwise. Map
+    // that onto the negative-means-failure convention used by the caller.
+    const int exit_status = (m_linear_solver->getStatus() == 0) ? 0 : -1;
+    if (exit_status < 0) {
+        return exit_status;
+    }
+
+    // Update E to E = -dA/dt (B is updated after the corrector push below)
+    UpdateEfromdA(a_step);
```
RemiLehe: Maybe this could be renamed `ComputeInductiveEfromdA`

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,969 @@
...
+
+    // AMReX's GMRES::getStatus() returns 0 on convergence and a positive
+    // value (e.g. 1 if the iteration count was exceeded) otherwise. Map
+    // that onto the negative-means-failure convention used by the caller.
+    const int exit_status = (m_linear_solver->getStatus() == 0) ? 0 : -1;
+    if (exit_status < 0) {
+        return exit_status;
+    }
+
+    // Update E to E = -dA/dt (B is updated after the corrector push below)
```
RemiLehe:
```suggestion
    // Set E to E = -dA/dt (B is updated after the corrector push below)
```

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,969 @@
...
+    // separate ComputeVectorLaplacian calls (with an intermediate boundary
+    // fill in between) introduces a parasitic, sign-alternating mode in Z's
+    // nodal component, since each application's guard cells are filled
+    // independently rather than being derived from a single consistent
+    // wide-stencil read of Z.
+    m_WarpX->get_pointer_fdtd_solver_fp(lev)->ComputeVectorBiLaplacian(
+        rhs_vec[lev], Zscratch, m_WarpX->GetEBUpdateBFlag()[lev], lev
+    );
+
+    // Calculate dA = curl(Z) (ComputeCurlB resets dA_fp to zero internally).
```
RemiLehe: The term "to zero" is maybe misleading here.
```suggestion
    // Calculate dA = curl(Z)
```

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,969 @@
...
+    // Accumulate current* and susceptibility (mass matrices)
+    AccumulateCurrentAndSusceptibility();
+
+    // Python callback insertion
+    ExecutePythonCallback("afterdeposition");
+
+    // Populate the source vector
+    CalculateSourceVector();
+
+    // Solve MS equation
```
RemiLehe: In the comment: could you show the full equation that we are solving here?

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,969 @@
...
+    return exit_status;
+}
+
+void SemiImplicitDarwin::ComputeRHS ( WarpXSolverVec& a_RHS,
+                                      const WarpXSolverVec& a_Z,
+                                      [[maybe_unused]] amrex::Real start_time,
+                                      [[maybe_unused]] int a_nl_iter,
+                                      [[maybe_unused]] bool a_from_jacobian )
+{
+    BL_PROFILE("SemiImplicitDarwin::ComputeRHS()");
```
RemiLehe: At the beginning of this function, could you show the full equation that we are solving, and which part of that is the RHS?

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,969 @@
...
+        );
+    }
+
+    // Prepare current deposition by setting particle velocities to twice the
+    // t = n velocity values (with just the ES acceleration applied for the
+    // advanced velocity)
+    PrepareCurrentDeposition();
+
+    // Accumulate current* and susceptibility (mass matrices)
+    AccumulateCurrentAndSusceptibility();
```
RemiLehe: Could you replace the term "susceptibility" with "mass matrix" throughout the code? 
Right now, both terms are used (e.g. `InitializeMassMatrices` is called for the Darwin solver), which might lead to confusion.

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,969 @@
...
+            *m_WarpX->m_fields.get(FieldType::Efield_fp, Direction{2}, lev),
+            *m_WarpX->m_fields.get(FieldType::Bfield_fp, Direction{0}, lev),
+            *m_WarpX->m_fields.get(FieldType::Bfield_fp, Direction{1}, lev),
+            *m_WarpX->m_fields.get(FieldType::Bfield_fp, Direction{2}, lev),
+            MomentumPushType::Full
+        );
+    }
+
+    // Update particle velocities to include acceleration from both
+    // electrostatic and inductive electric field components
```
RemiLehe, line 202: There is a little dance throughout the code here, where we put different quantities into `ux`, `uy`, `uz` and `ux_n`, `uy_n`, `uz_n`. I wonder if that could be made clearer in the comments. (But I don't have a very good suggestion.)

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,969 @@
...
+                });
+            }
+        }
+    }
+}
+
+void SemiImplicitDarwin::AccumulateCurrentAndSusceptibility ()
+{
+    /*
+        Note: The functionality here deposits current to the Yee grid (and
```
RemiLehe: Is this comment still relevant/accurate? Could you update it?

.clang-tidy
roelof-groenewald: It is needed for this PR to pass the `clang-tidy` tests.

Source/FieldSolver/FiniteDifferenceSolver/ComputeCurlB.cpp
roelof-groenewald: I wanted to call it `ComputeCurlB` to make it obvious that it operates on a field that matches the staggering of B and produces a field that matches the staggering of E. If we call it `ComputeCurlZ`, one would have to remember (or look up every time) what the staggering of `Z` is.

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
roelof-groenewald: I'll remove this comment. `LinearFunctionMF` allows enough compute savings per GMRES iteration to justify the extra code.

Examples/Tests/magnetized_plasma_modes/inputs_test_em_modes_picmi.py
roelof-groenewald: The difference in parameters mostly came from wanting a smaller domain for the Darwin solver (which is slower than the hybrid solver). And since the cell size for Darwin is relative to the Debye length (rather than the ion skin-depth), I needed to adjust the physical parameters in order to have the relatively small number of Debye lengths of the domain, fit sufficient ion skin-depths to still see the mode dispersion. We can play with this now that the Darwin solver is working better, but maybe we should do that after the preconditioner is merged (which will help make the run go a lot faster).

Examples/Tests/magnetized_plasma_modes/inputs_test_em_modes_picmi.py
```diff
@@ -24,71 +24,134 @@
...
-        self.diag_steps = int(1.0 / 20 / self.DT)
+            self.total_steps = 50 if self.solver == "darwin" else 250
+
+        if self.solver == "darwin":
+            self.diag_steps = 3
+        else:
+            # output diagnostics 20 times per cyclotron period
+            self.diag_steps = int(1.0 / 20 / self.DT)
+
+        if self.solver == "darwin":
```
RemiLehe, line 150: The reference to SIPIC (in the comment and in `sipic_factor` might be confusing here)

Examples/Tests/magnetized_plasma_modes/inputs_test_em_modes_picmi.py
```diff
@@ -24,71 +24,134 @@
...
+            # calculate SIPIC modified plasma quantities
+            sipic_factor = np.sqrt(1.0 + self.C_SI * (self.w_pe * self.dt) ** 2 / 4.0)
+            self.lambda_e_SI = self.lambda_e * sipic_factor
```
RemiLehe: I don't think that this is being used.
```suggestion
```

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,667 @@
...
+        accumulates the susceptibility to a staggered grid). The prototype
+        Darwin solver does the depositions to nodal grids!
+    */
```
RemiLehe: This seems to reference the Python-based prototype implementation. If this is correct, this should be removed.
```suggestion
```

Docs/source/usage/parameters.rst
```diff
@@ -367,6 +367,35 @@ Overall simulation parameters
...
       non-periodic behavior of the individual blocks.
 
+    * ``semi_implicit_darwin``: Use the semi-implicit Darwin field solver.
+
+      This solver advances the electrostatic (longitudinal) field together with the inductive
+      (magnetostatic Darwin) field, thereby retaining low-frequency magnetic effects while
+      filtering out light waves. The magnetostatic Darwin equation is solved for an auxiliary
+      vector potential using the AMReX GMRES linear solver; the resulting field is used to
+      obtain the inductive :math:`\textbf{E}`-field and to update the magnetic field.
+
```
```suggestion
      filtering out light waves. 

```

Source/NonlinearSolvers/NonlinearSolverLibrary.H
```diff
@@ -13,7 +13,8 @@
 AMREX_ENUM (NonlinearSolverType,
     picard,
     newton,
-    petsc_snes
+    petsc_snes,
+    none
```
RemiLehe: I don't think that this change is actually needed.
```suggestion
    petsc_snes
```

Source/NonlinearSolvers/LinearFunctionMF.H
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * \brief This is a linear function class for computing the action of a
+ *  linear operator on a vector using a matrix-free method.
+ *  This class has all of the required functions to be used as the
+ *  linear operator template parameter in AMReX_GMRES.
+ */
+
+template <class T, class Ops>
+class LinearFunctionMF : public LinearFunction<T,Ops>
```
RemiLehe: `LinearFunctionMF` looks like a mostly empty shell on top of `LinearFunction`. 
I think that other developers might then wonder what to use `LinearFunctionMF` for, versus what to use `LinearFunction` for.

What do you think about instead making `LinearFunctionMF` specific to the Darwin solver (and renaming it as e.g. `DarwinLinearOperator`). That way, we could also specialize some of the code directly in this file (i.e. implement more of Darwin-specific logic in `precond` and `apply`, instead of calling very generic function like `ComputeRHS`).

If other developers need a generic linear operator, I feel like that they could subclass `LinearFunction` directly (I don't think that would gain that much from having an additional, generic `LinearFunctionMF`)

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,667 @@
...
+    // for the advanced velocity), and the advanced velocity is saved to u_n
+    PrepareVelocitiesForCurrentDeposition();
+
+    // Accumulate current* and susceptibility (mass matrices)
+    AccumulateCurrentAndSusceptibility();
+
+    // Python callback insertion
+    ExecutePythonCallback("afterdeposition");
+
+    // Populate the source vector
```
RemiLehe, line 159:
```suggestion
    // Populate the source vector
    // i.e. fill m_source with `2 * laplacian(B) + 2 * mu_0 curl(J)`
```

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,667 @@
...
+    CalculateSourceVector();
+
+    // Solve MS equation:
```
RemiLehe: We refer to this equation as the "magnetostatic" equation, but this seems slightly misleading to me. Because this equation capture the physics of induction, it is not purely magnetostatic.

Is "magnetostatic" the standard term used in the literature (in which case it would make sense to keep it), or could we try to find a better name. (e.g. "magneto-inductive field equation"?)

roelof-groenewald: We could do either - funnily enough both are actually used in the literature. The opening sentence for a 2014 paper states "Dispersion relations from the Darwin (a.k.a., magnetoinductive or magnetostatic) model ...".

roelof-groenewald: In an early (1977) paper the Darwin model is referred to only as "magnetostatic" - https://www.sciencedirect.com/science/article/abs/pii/0021999177900961

Source/FieldSolver/ImplicitSolvers/SemiImplicitDarwin.cpp
```diff
@@ -0,0 +1,557 @@
...
+    using warpx::fields::FieldType;
+
+    const int lev = 0;
+
+    // Deposit the current density from all species, using the time-centered
+    // particle velocities as appropriate for the implicit push. This also
+    // resets the current MultiFabs before depositing.
+    m_WarpX->GetPartContainer().DepositCurrent(
+        m_WarpX->m_fields.get_mr_levels_alldirs(FieldType::current_fp, lev),
+        m_dt, 0.0_rt, PushType::Implicit);
```
RemiLehe, line 305: I am not sure I understand why `PushType::Implicit` is needed here. It seems that one of the main difference is that it will call `GetImplicitGammaInverse`.

My understanding is that `GetImplicitGammaInverse` is important to get **exact** energy conservation **for implicit EM PIC**. In particular, my understanding is that `GetImplicitGammaInverse` contains a formula that makes the system exactly energy conserving, **under the assumption** that `u` and `E` are defined at the same time (i.e. not staggered). This is the case for the implicit EM PIC which solves (assuming $B=0$ for simplicity):

$$ m\frac{u^{n+1} - u^n}{\Delta t} = q\frac{E^{n+1}+E^n}{2} \qquad \frac{E^{n+1}-E^{n}}{\Delta t} = \frac{j^{n+1/2}}{\epsilon_0}$$

and `GetImplicitGammaInverse` produces the $$j^{n+1/2}$$ that makes this system exactly energy conserving, i.e.
$$j^{n+1/2}\propto \frac{u^{n+1}-u^n}{\gamma^{n+1} - \gamma^n}$$

However, for Darwin, `u` and `E` are **not** defined at the same time but are instead staggered (i.e. we advance $E^n$ but $u^{n+1/2}$). In that context, I am not sure that using `GetImplicitGammaInverse` is justified (compared to the corresponding explicit formula, which simply uses the `gamma` that corresponds to the current `u`).

Accordingly, I would argue in favor of using `PushType::Explicit` here instead of `PushType::Implicit`

RemiLehe: Hm, after more thought the mass matrix deposition code below uses the implicit formula for gamma anyway (no choice, unlike the current deposition) ...

RemiLehe (approved):

roelof-groenewald:
Thanks for all the clean-up and improvements @RemiLehe!

PR #6294, opened 2025-10-23
Make PIdx struct more readable

lucafedeli88:
This PR makes the definition of `PIdx` more readable and more compact.


ax3l (approved):
Looks legit, thank you!

PR #6310, opened 2025-10-27
Fix single precision issue in DSMC ionization

roelof-groenewald:
The DSMC ionization routine currently produces incorrect results when compiling in single precision. Here is the `development` result for the `test_3d_ionization_electron_dsmc` benchmark in single precision:
image

This PR fixes the energy evolution by calculating scattered velocities using double precision. Here is the same test run with this branch:
image


RemiLehe (approved):
Thanks a lot for identifying and fixing this issue! ✨

PR #6314, opened 2025-10-29
PICMI: WarpX OMP Threads

ax3l:
Add `warpx_omp_threads` support to `PICMI.Simulation`. Follow-up to #6288

I renamed this from the internal `amrex_` prefix, because it seems more sensible to call it `warpx_`, but please let me know if you prefer to keep it 1:1 like ParmParse (as we do for GPU-aware MPI and other options).


RemiLehe:
Thanks for this PR! Could you try to use it in one of the PICMI CI tests, to make sure that this is tested?

RemiLehe (changes requested):
Thanks! Could you add a CI test?

dpgrote:
"warpx_" is the correct prefix to use for the input parameter so that it is properly handled by PICMI.

PR #6315, opened 2025-10-29
WarpX Remake: Keep MultiFab Pointers for EBs

ax3l:
We want to keep the `MultiFab` pointers the same during load balancing (`WarpX::RemakeLevel`).

In interactive usage, users might store a reference/handle to our `MultiFab` classes directly between simulation steps. We want to make sure these references stay valid.

Using `std::swap` and `std::move` helps us to move the new data underneath the existing pointer to `MultiFab` instead of replacing it with a new pointer altogether during load balancing, where we recreate `MultiFab` instances.

This fixes the load balancing of EB fields to stay stable, too. With this fix (and the already working handles handed out by `MutliFabRegister` for most of our fields), we can avoid our high level field wrapper and transparently forward all AMReX methods and properties to users, without another level of indirection.

### DIFF
--- a/Source/Parallelization/WarpXRegrid.cpp
+++ b/Source/Parallelization/WarpXRegrid.cpp
@@ -183,7 +183,7 @@ WarpX::RemakeLevel (int lev, Real /*time*/, const BoxArray& ba, const Distributi
         const IntVect& ng = mf->nGrowVect();
         auto pmf = std::remove_reference_t<decltype(mf)>{};
         AllocInitMultiFab(pmf, mf->boxArray(), dm, mf->nComp(), ng, lev, mf->tags()[0]);
-        std::swap(mf, pmf);
+        *mf = std::move(*pmf);
     };
 
     bool const eb_enabled = EB::enabled();
--- a/Source/Parallelization/WarpXRegrid.cpp
+++ b/Source/Parallelization/WarpXRegrid.cpp
@@ -183,7 +183,7 @@ WarpX::RemakeLevel (int lev, Real /*time*/, const BoxArray& ba, const Distributi
         const IntVect& ng = mf->nGrowVect();
         auto pmf = std::remove_reference_t<decltype(mf)>{};
         AllocInitMultiFab(pmf, mf->boxArray(), dm, mf->nComp(), ng, lev, mf->tags()[0]);
-        mf = std::move(pmf);
+        std::swap(mf, pmf);
     };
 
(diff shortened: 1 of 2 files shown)


WeiqunZhang:
I don't think this will make any difference. Do we have a test showing it does not work?

WeiqunZhang:
We do need to fix `RemakeMultiFab`. But we don't need to change MultiFab Register. I believe the comment in the documentation is mostly obsolete.

Source/ablastr/fields/MultiFabRegister.cpp
```diff
@@ -273,7 +273,7 @@ namespace ablastr::fields
                 }
 
                 // replace old MultiFab with new one, deallocate old one
-                mf_owner.m_mf = std::move(new_mf);
+                std::swap(mf_owner.m_mf, new_mf);
```
WeiqunZhang:
```suggestion
                mf_owner.m_mf = std::move(new_mf);
```
Revert the change.

Source/Parallelization/WarpXRegrid.cpp
```diff
@@ -183,7 +183,7 @@ WarpX::RemakeLevel (int lev, Real /*time*/, const BoxArray& ba, const Distributi
         const IntVect& ng = mf->nGrowVect();
         auto pmf = std::remove_reference_t<decltype(mf)>{};
         AllocInitMultiFab(pmf, mf->boxArray(), dm, mf->nComp(), ng, lev, mf->tags()[0]);
-        mf = std::move(pmf);
+        std::swap(mf, pmf);
```
WeiqunZhang:
```suggestion
        *mf = std::move(*pmf);
```
We want to keep `MultiFab*` stable, not `std::unique_ptr<MultiFab>*`.

ax3l:
@dpgrote did you have a simple test for this? Otherwise we can design one.

ax3l:
@RemiLehe wrote a test in https://github.com/RemiLehe/WarpX/commit/588bff5f75da81783f9600e9525bc141cce9b2e2 and it turns out that we fixed the issue by accident already :D

When we introduced the `MultiFabRegister` in #5230 in last year's hackathon, we made the references we hand out persistent over load balance @dpgrote :tada: 

RemakeMultiFab still has the old style and needs a fix. (For testing, we will need to expose `eb_e_update` to Python.)

WeiqunZhang (approved):

PR #6316, opened 2025-10-30
Adastra.rst: disable openMP support for QED lookup table generation

grobertdautun:
Compilation with openMP support for QED lookup tables generation crashes on the ADASTA HPC. This PR disables it

### DIFF
--- a/Docs/source/install/hpc/adastra.rst
+++ b/Docs/source/install/hpc/adastra.rst
@@ -100,7 +100,7 @@ Use the following :ref:`cmake commands <building-cmake>` to compile the applicat
    cd $SHAREDHOMEDIR/src/warpx
    rm -rf build_adastra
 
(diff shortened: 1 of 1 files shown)


lucafedeli88 (approved):
Thanks @grobertdautun . I would just suggest adding a sentence stating that compiling with this option = ON or AUTO on Adastra does not work, so for table generation we recommend using the QED tools.

lucafedeli88:
> Thanks @grobertdautun . I would just suggest adding a sentence stating that compiling with this option = ON or AUTO on Adastra does not work, so for table generation we recommend using the QED tools.

I would write a comment like:
```
# Enabling openMP support for QED lookup tables generation in WarpX while compiling WarpX for GPUs on Adastra
# does not work. It is recommended to generate QED lookup tables locally (e.g., with the standalone tool) and then
# transfer them to Adastra. The usage of the standalone tool for QED lookup tables generation is documented in the
# usage/workflows section of the documentation.
```

PR #6318, opened 2025-10-30
Update runtime names to access particles from Python

RemiLehe:
This ensures that, when the user tries to get the `r` position of a particle from Python, they do not need to type `"x"` instead of `"r"`.

### DIFF
--- a/Source/Particles/WarpXParticleContainer.H
+++ b/Source/Particles/WarpXParticleContainer.H
@@ -71,13 +71,13 @@ struct PIdx
     static constexpr auto names = {"x", "z", "w", "ux", "uy", "uz"};
 #elif defined(WARPX_DIM_RZ)
     enum                          { x ,  z ,  w ,  ux ,  uy ,  uz ,  theta,        nattribs};
-    static constexpr auto names = {"x", "z", "w", "ux", "uy", "uz", "theta"};
+    static constexpr auto names = {"r", "z", "w", "ux", "uy", "uz", "theta"};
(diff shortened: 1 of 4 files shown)


Source/Particles/WarpXParticleContainer.H
```diff
@@ -71,13 +71,13 @@ struct PIdx
     static constexpr auto names = {"x", "z", "w", "ux", "uy", "uz"};
 #elif defined(WARPX_DIM_RZ)
     enum                          { x ,  z ,  w ,  ux ,  uy ,  uz ,  theta,        nattribs};
-    static constexpr auto names = {"x", "z", "w", "ux", "uy", "uz", "theta"};
+    static constexpr auto names = {"r", "z", "w", "ux", "uy", "uz", "theta"};
 #elif defined(WARPX_DIM_RCYLINDER)
     enum                          { x ,  w ,  ux ,  uy ,  uz ,  theta,             nattribs};
-    static constexpr auto names = {"x", "w", "ux", "uy", "uz", "theta"};
+    static constexpr auto names = {"r", "w", "ux", "uy", "uz", "theta"};
```
RemiLehe, line 77: This might controversial/confusing. However, I think that it does make the user interface more intuitive, and is therefore overall beneficial.

Changing the `enum` itself from `x` to `r` is much more work:
https://github.com/BLAST-WarpX/warpx/pull/4667

ax3l (approved):
LGTM, ship it :D

PR #6320, opened 2025-10-30
Remove particles.photon_species inputs left over

EZoni:
The `particles.photon_species` input parameter is deprecated since #6265.

On the other hand, I think this is an example of how deprecated parameters can still propagate in the code due to the fact that they still function (the one in `inputs_test_1d_photoneutralization_dsmc` was merged two days ago) - which was my concern behind https://github.com/BLAST-WarpX/warpx/pull/6265#discussion_r2440528311.


lucafedeli88 (approved):
Thanks for this PR, @EZoni .
I see your comment. You are right. A workaround would be to introduce another warning level alongside "low", "medium", "high" : "deprecated", and have an option like
`warpx.abort_on_warning_threshold=deprecated` switched on by default in all the tests.

PR #6322, opened 2025-10-30
Spacecraft Charging Example: Get Field Dir String

ax3l:
Use the MultiFabRegister using its string interface for the component.

I used named arguments for clarity, but using the argument names is optional.


RemiLehe (approved):
Very nice! Thanks for adding this.

PR #6325, opened 2025-10-31
Fix default tolerance in restart test

RemiLehe:
The current script does not let the user choose which tolerance to use, in the case of restart tests.

### DIFF
--- a/Examples/analysis_default_regression.py
+++ b/Examples/analysis_default_regression.py
@@ -15,14 +15,8 @@ def main(args):
     # parse test name from test directory
     test_name = os.path.split(os.getcwd())[1]
     if "_restart" in test_name:
-        rtol_restart = 1e-12
-        print(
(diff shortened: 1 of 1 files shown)


Examples/analysis_default_regression.py
```diff
@@ -45,12 +39,19 @@ def main(args):
         type=str,
     )
     # add arguments: relative tolerance
+    # Identify whether the test is a restart test,
+    # if it is, use a 1e-12 tolerance for the restart checksum analysis
+    test_name = os.path.split(os.getcwd())[1]
+    if "_restart" in test_name:
+        default_tolerance = 1e-12
+    else:
+        default_tolerance = 1e-9
```
EZoni: This could be a one-liner:
```suggestion
    default_tolerance = 1e-12 if "_restart" in test_name else 1e-9
```

EZoni (approved):

PR #6326, opened 2025-11-01
Remove mentions of ARTEMIS in the examples

RemiLehe:
ARTEMIS is a different code than WarpX.

Therefore listing ARTEMIS references/examples is not directly relevant, and might be confusing to users (for the same reaons, we do not list HiPACE++ or ImpactX references/examples ; these connections could be made on the BLAST website instead: https://blast.lbl.gov/)


Docs/source/highlights.rst
```diff
@@ -194,28 +194,6 @@ Scientific works in astrophysical plasma modeling.
    The Astrophysical Journal **952** 8, 2023.
    `DOI:10.3847/1538-4357/acd75b <https://doi.org/10.3847/1538-4357/acd75b>`__
```
RemiLehe, line 196: In the future, we could reintroduce a section "Microelectronics / plasma processing" and including some of the work done with WarpX under the AMMTO project.

ax3l (changes requested):
Hm, I think the Science Highlights is a semi-advertising page and I would keep it there.

I also think the examples section does not hurt. Let's discuss.

ARTEMIS is truly 95% WarpX, forked for modifications.

EZoni (dismissed):

ax3l (approved):
LGTM, as discussed! :)

Keeping in highlights because of close relationship/fork.

PR #6336, opened 2025-11-03
Consistently use language:none for literal include in the documentation

RemiLehe:
This is a follow-up of #6333 and makes these changes consistent across the documentation.


dpgrote (approved):
Thanks!

PR #6341, opened 2025-11-04
Update high-level page of the theory documentation

RemiLehe:
This updates the overview of the theory page, to make it less specific to plasma-based accelerators:
Screenshot 2025-11-04 at 6 14 11 AM


Docs/source/theory/intro.rst
```diff
@@ -66,6 +49,14 @@ Species Representations
    kinetic_particles
    cold_fluid_model
 
+Boundary Conditions
```
RemiLehe, line 52: It seems to me that it would make more sense to first discuss the species representation, and later the boundary conditions.

roelof-groenewald (approved):
Looks good to me! Thanks for the improvement!

PR #6342, opened 2025-11-04
Update title of the hybrid PIC section to mention Ohm's law

RemiLehe:
Since the hybrid PIC model is documented as a field solver, I think it makes sense to explicitly mention the field equation that it solves in the title.

### DIFF
--- a/Docs/source/theory/kinetic_fluid_hybrid_model.rst
+++ b/Docs/source/theory/kinetic_fluid_hybrid_model.rst
@@ -1,7 +1,7 @@
 .. _theory-kinetic-fluid-hybrid-model:
 
-Kinetic-fluid Hybrid Model
-==========================
+Hybrid Model based on Maxwell-Ampere coupled with Ohm's law
+===========================================================
 
 Many problems in plasma physics fall in a class where both electron kinetics and electromagnetic waves do not
 play a critical role in the solution. Examples of such situations include the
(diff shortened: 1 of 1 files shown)


Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -1,7 +1,7 @@
 .. _theory-kinetic-fluid-hybrid-model:
 
-Kinetic-fluid Hybrid Model
-==========================
+Hybrid Model based on Maxwell-Ampere coupled with Ohm's law
+===========================================================
```
roelof-groenewald: Yeah, I agree with you that mentioning the field equations would be good.
I wonder, though, if we should change this as follows since the "hybrid" part is meant to signify that it is part fluid part kinetic (so saying hybrid without "kinetic-fluid" is maybe not clear)
```suggestion
Maxwell-Ampere coupled with Ohm's law
=====================================
```

Docs/source/theory/kinetic_fluid_hybrid_model.rst
```diff
@@ -1,7 +1,7 @@
 .. _theory-kinetic-fluid-hybrid-model:
 
-Kinetic-fluid Hybrid Model
-==========================
+Hybrid Model based on Maxwell-Ampere coupled with Ohm's law
```
RemiLehe:
```suggestion
Maxwell-Ampere coupled with Ohm's law (a.k.a. "hybrid PIC")
```

roelof-groenewald (approved):

PR #6344, opened 2025-11-04
Python: Implement `mesh` on `MultiFab`

ax3l:
Move from `MultiFabWrapper` to `MultiFab`. Will automatically be forwarded still to the `MultiFabWrapper`.

### DIFF
--- a/Python/pywarpx/extensions/MultiFabRegister.py
+++ b/Python/pywarpx/extensions/MultiFabRegister.py
@@ -1,0 +1,48 @@
+"""
+This file is part of WarpX
+
+Copyright 2025 WarpX community
+Authors: Axel Huebl
+License: BSD-3-Clause-LBNL
+"""
+
+
+def get(self, name, dir=None, level=None):
+    """Return a MultiFab that is part of a vector/tensor field.
+
+    This throws a runtime error if the requested field is not present.
+
+    Parameters
+    ----------
+    self : MultiFabRegister
+        The MultiFabRegister.
+    name : str
+        The name of the field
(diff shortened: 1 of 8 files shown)


Python/pywarpx/extensions/MultiFabRegister.py
```diff
@@ -0,0 +1,48 @@
...
+        Required for vector/tensor fields, omitted for scalar fields.
+    level : int
+        The MR level.
+
+    Returns
+    -------
+    MultiFab
+        A non-owning reference to the MultiFab (field)
+    """
+    if level is None:
```
dpgrote, line 32: Can you do the same thing here with level that was done for the particle container? Have it default to 0 if there is no mesh refinement, otherwise have it be required.

ax3l: Sounds good. Challenge here is that we need to go back to the WarpX object to find out what the highest MR level is... (the MF register itself does not know, necessarily...)

ax3l: Maybe let us not add this here, but in the `sim.fields` method?

dpgrote: The `sim.fields` returns the multifab register. It doesn't have any arguments.

ax3l: Ah, agreed. Can we still make it a follow-up PR when we add `sim.fields`?

RemiLehe (approved):

PR #6337, opened 2025-11-03
Docs: Complete hybrid solver parameter docs

n01r:
- Add HybridPICSolver to Python PICMI docs via autoclass
- Document all 16 parameters with units and variable dependencies
- Add constraints: required params, best practices, limitations
- Fix typos and parameter name inconsistency

PR #6346, opened 2025-11-04
Fix HybridPICSolver docs

n01r:
Fixes some indentation warnings and errors by replacing the more extensive code block with indentations in the docstring with a sentence that explains the same content.


n01r:
Fixes the warnings introduced in #6337

roelof-groenewald (approved):

PR #6348, opened 2025-11-05
Docs: Fix :alt: figure text again

EZoni:
Follow up to #5386.

@ax3l pointed out that the :alt: text should always provide a concise description of the image for digital accessibility, even if the caption is also always accessible.


RemiLehe (approved):

ax3l (approved):
Thank you, that looks good :)

PR #6354, opened 2025-11-05
Add cupy support for picmi CIs with callback functions

oshapoval:
To fix (separate PR?): during the shutdown, several CI tests still crash: `amrex::Abort::1::CUDA error 709 in file /global/homes/o/oshapova/src/warpx/build/_deps/fetchedamrex-src/Src/Base/AMReX_GpuDevice.cpp line 691: context is destroyed !!!`
Crashing tests: 

- `/Examples/Tests/ohm_solver_ion_Landau_damping/inputs_test_2d_ohm_solver_landau_damping_picmi.py`
- `/Examples/Physics_applications/capacitive_discharge/inputs_test_2d_background_mcc_picmi.py`
- `/Examples/Physics_applications/capacitive_discharge/inputs_base_1d_picmi.py`
- `/Examples /Physics_applications/spacecraft_charging/inputs_test_rz_spacecraft_charging_picmi.py`
- `/Examples/Tests/ohm_solver_cylinder_compression/inputs_test_3d_ohm_solver_cylinder_compression_picmi.py`
- `/Examples/Tests/projection_div_cleaner/inputs_test_3d_projection_div_cleaner_callback_picmi.py`


Examples/Tests/ohm_solver_magnetic_reconnection/inputs_test_2d_ohm_solver_magnetic_reconnection_picmi.py
```diff
@@ -306,15 +322,25 @@ def check_fields(self):
...
             return
 
-        rho = simulation.fields.get("rho_fp", level=0)[...]
-        Jiy = simulation.fields.get("current_fp", dir="y", level=0)[...] / self.J0
-        Jy = (
+        get_xp()
+
+        rho = to_host(simulation.fields.get("rho_fp", level=0)[...])
+
+        Jiy = to_host(simulation.fields.get("current_fp", dir="y", level=0)[...])
```
ax3l: Do this instead and stay on device:
```suggestion
        rho = simulation.fields.get("rho_fp", level=0)

        Jiy = simulation.fields.get("current_fp", dir="y", level=0).mult(1 / self.J0)
```
https://pyamrex.readthedocs.io/en/25.11/usage/api.html#amrex.space3d.MultiFab.mult

Examples/Tests/particle_boundary_interaction/inputs_test_rz_particle_boundary_interaction_picmi.py
```diff
@@ -112,14 +113,22 @@
...
     else:
-        return np.concatenate(list_of_arrays)
+        return xp.concatenate(list_of_arrays)
+
+
+def to_numpy(arr):
+    if hasattr(arr, "get"):
+        return arr.get()
+    else:
+        return np.asarray(arr)
```
ax3l:
```suggestion
        return arr
```

Examples/Tests/particle_data_python/inputs_test_2d_particle_attr_access_picmi.py
```diff
@@ -104,35 +105,48 @@
...
-    newPid = 5.0
+    nps = nps_per_rank(my_id)
+    x = xp.linspace(0.005, 0.025, nps)
+    y = xp.zeros(nps)
+    z = xp.linspace(0.005, 0.025, nps)
+    ux = xp.random.normal(loc=0, scale=1e3, size=nps)
+    uy = xp.random.normal(loc=0, scale=1e3, size=nps)
+    uz = xp.random.normal(loc=0, scale=1e3, size=nps)
+    w = xp.ones(nps) * 2.0
+    newPid = xp.ones(nps) * 5.0
```
ax3l: Use `np.` here because `electrons.add_particles` anyway only accepts numpy.

Examples/Tests/restart/inputs_test_2d_runtime_components_picmi.py
```diff
@@ -104,25 +106,43 @@
...
+        return arr.get()
+    else:
+        return np.asarray(arr)
+
 
-    electrons.add_particles(x=x, y=y, z=z, ux=ux, uy=uy, uz=uz, w=w, newPid=newPid)
+def add_particles():
+    rank = mpi.COMM_WORLD.Get_rank()
+    nps = 10 if rank == 0 else 0
+    x = xp.linspace(0.005, 0.025, nps)
```
ax3l: As above, keep np because of add_particles

Examples/Tests/secondary_ion_emission/inputs_test_rz_secondary_ion_emission_picmi.py
```diff
@@ -142,44 +142,51 @@
...
+        0.0,
+    )
     return sigma_nascap
 
 
+def to_numpy(arr):
+    if hasattr(arr, "get"):
+        return arr.get()
+    else:
+        return np.asarray(arr)
```
ax3l:
```suggestion
        return arr
```

Examples/Tests/ohm_solver_magnetic_reconnection/inputs_test_2d_ohm_solver_magnetic_reconnection_picmi.py
```diff
@@ -306,12 +312,16 @@ def check_fields(self):
         if not (step == 1 or step % self.diag_steps == 0):
             return
 
-        rho = simulation.fields.get("rho_fp", level=0)[...]
+        get_xp()
+
```
ax3l, line 316: `xp` is not used below. Is this a work-around or a leftover?

Examples/Tests/particle_data_python/inputs_test_2d_particle_attr_access_picmi.py
```diff
@@ -132,7 +146,7 @@ def add_particles():
         uz=uz,
         w=w,
         newPid=newPid,
-        unique_particles=args.unique,
+        unique_particles=True,
```
ax3l, line 149: Did you intentionally remove `args.unique` here?
```suggestion
        unique_particles=args.unique,
```

It is still used in the `parser` above and potentially passed by the test runtime arguments.

Examples/Tests/restart/inputs_test_2d_runtime_components_picmi.py
```diff
@@ -151,3 +171,6 @@ def add_particles():
 ##########################
 
 sim.step(1)
+
+if not sim.amr_restart:
+    del electrons
```
ax3l, line 176: Why not always delete it?
```suggestion
del electrons
```

Examples/Tests/ohm_solver_magnetic_reconnection/inputs_test_2d_ohm_solver_magnetic_reconnection_picmi.py
oshapoval: Yes, it's a work-arounds, otherwise same shutdown error `context is destroyed !!!
SIGABRT`

Examples/Tests/particle_data_python/inputs_test_2d_particle_attr_access_picmi.py
oshapoval: I intentionally removed it so that all added particles are treated as unique, which simplifies the subsequent check on both cpu/gpu:
`assert electrons.size == total_expected_electrons_size`,
where 
`total_expected_electrons_size = sum(
    nps_per_rank(rank) * (max_steps - 1) for rank in range(num_ranks)
)`.

Examples/Tests/restart/inputs_test_2d_runtime_components_picmi.py
oshapoval: If I just have `del electrons` all the tests pass indeed, but pre-commit complains `Undefined name electrons`.

ax3l: Oh, interesting! Then leave as is for now of course.

Examples/Tests/ohm_solver_magnetic_reconnection/inputs_test_2d_ohm_solver_magnetic_reconnection_picmi.py
ax3l: Ah, got it.

ax3l (approved):
Thank you, Olga! :rocket: :sparkles:

Examples/Tests/particle_data_python/inputs_test_2d_particle_attr_access_picmi.py
```diff
@@ -104,24 +105,37 @@
...
     y = np.zeros(nps)
     z = np.linspace(0.005, 0.025, nps)
     ux = np.random.normal(loc=0, scale=1e3, size=nps)
     uy = np.random.normal(loc=0, scale=1e3, size=nps)
     uz = np.random.normal(loc=0, scale=1e3, size=nps)
     w = np.ones(nps) * 2.0
-    newPid = 5.0
+    newPid = np.ones(nps) * 5.0
 
     electrons.add_particles(
```
EZoni, line 140: @ax3l @oshapoval 

I still think that the `electrons` of `electrons.add_particles(...)` is not defined within the scope of the function `add_particles` (defined on line 129 above), which I think is what pre-commit was complaining about (discussed briefly off GitHub with @oshapoval).

It looks to me like it should be either declared as `global` at the beginning of the function (i.e., on line 130) or passed as an input argument to the function, no?

PR #6357, opened 2025-11-05
Python: Remove PC Wrapper in Usage

ax3l:
Modernize examples to use the new `sim.particles` access to `MultiParticleContainer` directly.

### DIFF
--- a/Examples/Physics_applications/capacitive_discharge/inputs_base_1d_picmi.py
+++ b/Examples/Physics_applications/capacitive_discharge/inputs_base_1d_picmi.py
@@ -11,7 +11,7 @@ import numpy as np
 from scipy.sparse import csc_matrix
 from scipy.sparse import linalg as sla
 
-from pywarpx import callbacks, libwarpx, particle_containers, picmi
+from pywarpx import callbacks, libwarpx, picmi
 from pywarpx.LoadThirdParty import load_cupy
 
 constants = picmi.constants
@@ -410,13 +410,11 @@ class CapacitiveDischargeExample(object):
             return
 
         if not hasattr(self, "neutral_cont"):
-            self.neutral_cont = particle_containers.ParticleContainerWrapper(
-                self.neutrals.name
-            )
+            self.neutral_particles = self.sim.particles.get(self.neutrals.name)
 
-        ux_arrays = self.neutral_cont.uxp
-        uy_arrays = self.neutral_cont.uyp
-        uz_arrays = self.neutral_cont.uzp
+        ux_arrays = self.neutral_particles.uxp
+        uy_arrays = self.neutral_particles.uyp
+        uz_arrays = self.neutral_particles.uzp
 
         xp, _ = load_cupy()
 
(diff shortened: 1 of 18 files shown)


Examples/Tests/ohm_solver_ion_beam_instability/inputs_test_1d_ohm_solver_ion_beam_picmi.py
```diff
@@ -347,8 +343,8 @@ def text_diag(self):
 
         status_dict = {
             "step": step,
-            "nplive beam ions": self.ion_container_wrapper.nps,
-            "nplive ions": self.beam_ion_container_wrapper.nps,
+            "nplive beam ions": self.ion_container.size,
+            "nplive ions": self.beam_ion_container.size,
```
ax3l: @roelof-groenewald I realize now that the original code already mismatches the dict keys to the values on the RHS. Shall I fix on the go?
```suggestion
            "nplive ions": self.ion_container.size,
            "nplive beam ions": self.beam_ion_container.size,
```

dpgrote (approved):
This looks great! Thanks!

ax3l:
Will push small fixes related to counting examples (and https://github.com/AMReX-Codes/pyamrex/pull/503)

ax3l:
Note: not updating docs here as we work on a separate update of the Python extension page w/ @RemiLehe for another PR.

Examples/Physics_applications/capacitive_discharge/inputs_base_1d_picmi.py
```diff
@@ -410,13 +410,11 @@ def rethermalize_neutrals(self):
             return
 
         if not hasattr(self, "neutral_cont"):
```
roelof-groenewald:
```suggestion
        if not hasattr(self, "neutral_particles"):
```

ax3l: Thanks! Overlooked :)

roelof-groenewald (approved):
These are very nice improvements!

ax3l:
oops, will update the last three 1D tests that fail:

	 87 - test_1d_ohm_solver_ion_beam_picmi.run (Failed)    slow
	104 - test_1d_background_mcc_picmi.run (Failed)
	108 - test_1d_dsmc_picmi.run (Failed)

Examples/Physics_applications/capacitive_discharge/inputs_base_1d_picmi.py
```diff
@@ -409,36 +409,32 @@ def rethermalize_neutrals(self):
...
+            pti["uz"][:] = xp.array(vel_std * self.rng.normal(size=nps))
 
     def _get_rho_ions(self):
         # deposit the ion density in rho_fp
-        he_ions_wrapper = particle_containers.ParticleContainerWrapper("he_ions")
-        he_ions_wrapper.deposit_charge_density(level=0)
+        self.rho.set_val(0.0)
+        he_ions = self.sim.particles.get("he_ions")
+        he_ions.deposit_charge(self.rho, lev=0)
+        libwarpx.warpx.sync_rho()
```
ax3l, line 429: I think we could move `deposit_charge` from `ParticleContainerWrapper` straight to `sim.` (otherwise we can also add it on `WarpXParticleContainer`)

roelof-groenewald: That sounds good. It would also be good to fold the `libwarpx.warpx.sync_rho()` call into the `deposit_charge` function.

PR #6358, opened 2025-11-05
Add succinct explanation of hybrid-PIC model

roelof-groenewald:
Add top section description of the actual equations solved in the hybrid-PIC scheme:

image


RemiLehe (approved):
Thanks a lot!

PR #6360, opened 2025-11-05
Document how to extend simulations via Python (first step): add overview section

RemiLehe:
Add an overview section that motivates the Python extensions:
Screenshot 2025-11-05 at 3 39 07 PM


ax3l (approved):
Splendid

PR #6362, opened 2025-11-05
Add an alias warpx.step for warpx.evolve

dpgrote:
This adds the alias warp.step to be consistent with Simulation.step. This is to avoid confusion since in various examples warpx and Simulation are used interchangeably - the method for running the simulation should be the same.

Should the warpx.evolve just be removed (i.e. renamed "step")?

### DIFF
--- a/Docs/source/usage/workflows/python_extend.rst
+++ b/Docs/source/usage/workflows/python_extend.rst
@@ -39,7 +39,7 @@ Extend a Simulation with Python
          # register callbacks ...
 
          # advance simulation until the last time step
-         sim.evolve()
+         sim.step()
 
       .. dropdown:: Full Example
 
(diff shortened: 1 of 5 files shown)


ax3l (approved):
Good idea! :)

EZoni:
> Should the `warpx.evolve` just be removed (i.e., renamed `step`)?

I would vote for removing it, if the two do exactly the same thing (which is the case now).

ax3l:
It's quite a central, often used function. What we usually then do:

Let's print a warning that it is deprecated in the old one, so people are getting annoyed and migrate away from it to the new call.

Then in a year or two we remove it for good.

dpgrote:
PR #6380 makes this change, making warpx.evolve obsolete.

PR #6364, opened 2025-11-06
Document how to extent simulations with Python (second step): explain how to run

RemiLehe:
This improves the next section of the page:

Screenshot 2025-11-05 at 5 04 41 PM


ax3l (approved):
Good start :)

PR #6367, opened 2025-11-06
GMRES for Poisson Equation

ajnonaka:
GMRES implementation of poisson solver, enabled with `warpx.poisson_solver = gmres`.

Instead of pure multigrid via the AMReX MLMG class, this uses an MLMG-preconditioned GMRES algorithm to solve the Poisson equation.  More documentation forthcoming.


dpgrote:
I've tried this out and it works quite nicely. It's fully functional for my needs, in 2D, with Dirichlet and Neumann boundaries, and embedded boundaries. Comparing it to multigrid, it is consistently faster.

On a Mac with the small grid (80x200), the multigrid takes ~0.047 s per solve, the gmres 0.028 s per solve. (Though by comparison, the multigrid solver in Warp takes 0.0024 s per solve.)

On AMD GPU for a larger grid (360x800), the multigrid takes 0.058 s and gmres 0.048 s per solve, not a big difference.

On parallel CPU (256 processors) with the larger grid, multigrid taks 0.032 s, and gmres 0.029 s, also not a big difference. (For reference, the multigrid solver in Warp takes 0.01 s per solve.)

JustinRayAngus:
What is (if any) the precondition matrix solver?

dpgrote:
> What is (if any) the precondition matrix solver?

It says that it does 1 MLMG cycle.

ax3l:
@ajnonaka can you merge `development` again and check why the 1D test fails? I somehow have no permission to push to this branch.

Source/ablastr/fields/PoissonSolver.H
```diff
@@ -437,12 +445,47 @@ computePhi (
...
+                    warpx.m_fields.get(warpx::fields::FieldType::Efield_fp, Direction{0}, lev),
+                    warpx.m_fields.get(warpx::fields::FieldType::Efield_fp, Direction{2}, lev)
+                        }
+#elif defined(WARPX_DIM_3D)
+                amrex::Array<amrex::MultiFab *, 3>{
+                    warpx.m_fields.get(warpx::fields::FieldType::Efield_fp, Direction{0}, lev),
+                    warpx.m_fields.get(warpx::fields::FieldType::Efield_fp, Direction{1}, lev),
+                    warpx.m_fields.get(warpx::fields::FieldType::Efield_fp, Direction{2}, lev)
+                        }
+#endif
```
JustinRayAngus, line 480: @dpgrote Is electrostatic mode in WarpX not setup for RCYLINDER and RSPHERE geometries?

Source/ablastr/fields/PoissonSolver.H
```diff
@@ -416,8 +418,14 @@ computePhi (
...
         // Solve Poisson equation at lev
-        mlmg.solve( {phi[lev]}, {rho[lev]},
-                     relative_tolerance, absolute_tolerance );
+        if (WarpX::poisson_solver_id == PoissonSolverAlgo::Multigrid) {
+            mlmg.solve( {phi[lev]}, {rho[lev]},
+                        relative_tolerance, absolute_tolerance );
+        } else {
+            amrex::GMRESMLMG gmsolve(mlmg);
+            gmsolve.solve( *phi[lev], *rho[lev], relative_tolerance, absolute_tolerance );
+            linop->postSolve({phi[lev]});
```
JustinRayAngus, line 428: What is the need for `linop->postSolve() `here ? Why is it not needed when using mlmg?

Source/ablastr/fields/PoissonSolver.H
JustinRayAngus, line 409: Right now it doesn't look like this PR is setup to change additional parameters used by GMRES. I think it would be a good ideal to make the linear solver a member variable defined once on initialization, then you could have control over more options. For example, you may want to manually set the restart length:
`m_linear_solver->setRestartLength( m_linsol_restart_length );`

See `Source/NonlinearSolvers/NewtonSolver.H` for an example of what I'm referring to.

Source/ablastr/fields/PoissonSolver.H
dpgrote: @JustinRayAngus Those geometries are not supported in electrostatic. Also, this block of code is for EB which is also not supported.

Source/ablastr/fields/PoissonSolver.H
ajnonaka: MLMG calls postSolve within, which sets the value of phi in covered cells to the Dirichlet value.  GMRES initially left phi in covered cells at zero; I added this call so the plotfiles would match up in the covered cells.

Source/ablastr/fields/PoissonSolver.H
ajnonaka: Reva wants to discuss at the next dev meeting; the MLMG option has no exposure either.

Source/ablastr/fields/PoissonSolver.H
```diff
@@ -416,8 +418,14 @@ computePhi (
...
 
         // Solve Poisson equation at lev
-        mlmg.solve( {phi[lev]}, {rho[lev]},
-                     relative_tolerance, absolute_tolerance );
+        if (WarpX::poisson_solver_id == PoissonSolverAlgo::Multigrid) {
+            mlmg.solve( {phi[lev]}, {rho[lev]},
+                        relative_tolerance, absolute_tolerance );
+        } else {
+            amrex::GMRESMLMG gmsolve(mlmg);
+            gmsolve.solve( *phi[lev], *rho[lev], relative_tolerance, absolute_tolerance );
```
ax3l, line 427: @RemiLehe commented: can we also handle `max_iters` here? Otherwise it is silently ignored if user-set

RemiLehe:
Superseded by https://github.com/BLAST-WarpX/warpx/pull/7171

PR #6368, opened 2025-11-07
Improve documentation for callback function (third step): data access

RemiLehe:
This reorganizes the section on data access into separate pages (the current page was getting too long), and updates the documentation to describe the new interface that was implemented during the November 2025 hackathon.

The new structure is visible here in particular.
Screenshot 2026-01-25 at 7 41 24 AM

Before reviewing the code changes, I would recommend comparing the existing Sphinx documentation [here](https://warpx.readthedocs.io/en/latest/usage/workflows/python_extend.html) with the new proposed documentation [here](https://warpx--6368.org.readthedocs.build/en/6368/).

Next steps:
- Should we remove the previous Python interfaces? (e.g. `MultiFabWrapper`, `ParticleContainerWrapper`)


Examples/Tests/particle_boundary_interaction/inputs_test_rz_particle_boundary_interaction_picmi.py
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-# @Eya Dammak supervised by @Remi Lehe, 2024
```
RemiLehe, line 2: We usually do not add the name of the authors in the inputs scripts.

Docs/source/usage/workflows/python_field_data.rst
```diff
@@ -0,0 +1,207 @@
...
+            x_coords = Ex.mesh("x")
+            y_coords = Ex.mesh("y")
+            z_coords = Ex.mesh("z")
+
+        The method accepts a direction string (``"x"``, ``"y"``, ``"z"`` in 3D; ``"r"``, ``"z"`` in RZ geometry)
+        and an optional ``include_ghosts`` parameter (default ``False``) to include ghost cell coordinates.
+        The returned array contains the physical coordinates of the mesh points along the specified direction,
+        properly accounting for the field's cell-centered or face-centered staggering.
+
+        .. note::
```
ax3l: This file is very nice.

Two suggestions:
- instead of a `note`, use `warning` or similar that stands out
- maybe move this box further up
- for a tab, this global index tab is surprisingly long compared to the other two (a bit unhandy with scrolling & popping content below it)

Docs/source/usage/workflows/python_portable.rst
```diff
@@ -0,0 +1,24 @@
...
+Depending on whether WarpX is running with GPU support or not, these arrays are stored either on CPU or GPU.
+Working with those arrays requires a Python package that operates on CPU (e.g. `numpy <https://numpy.org/doc/stable/>`__) or GPU (e.g. `cupy <https://docs.cupy.dev/en/stable/>`__).
+Note that ``numpy`` and ``cupy`` have almost identical syntax, making it easy to write portable code that is not specific to CPU or GPU.
+In order to do so, one needs a functionality that will automatically detect whether WarpX runs on CPU or GPU and import the package ``numpy`` or ``cupy`` accordingly.
+This functionality is provided by the function :func:`load_cupy`, which can be used as shown below.
+
+.. code-block:: python
+
+      from pywarpx.LoadThirdParty import load_cupy
+      xp, _ = load_cupy()
```
ax3l: In the example here, you could be explicit if you want to:
```suggestion
      xp, status = load_cupy()

      # optional: print a warning if an issue occurs when loading cupy
      if status is not None:
          print(status)
```

ax3l: The above is just an idea.

The rationale here is that we have a relatively large set of working numpy fallbacks and explicit copies, even with GPU runs. Thus `load_cupy()` does not necessarily fail immediately when it suspects GPUs are in the play. For read-only GPU interaction, pyAMReX/WarpX have enough understanding to work directly with `numpy` if cupy is hard-to-get (for some reason, like being on AMD or Intel GPUs or having trouble to compile cupy for some reason on a system).

But maybe the other read-only APIs we show are sufficient then.

ax3l (approved):
This is great.

Minor ideas inline, but adjust as you see fit.

PR #6369, opened 2025-11-07
[WIP] Docs: Rename inputs sections

EZoni:
I would like to propose an audacious change, inspired by a comment from @aeriforme during the documentation hackathon, and rename the following sections:
- "Inputs: Parameter List": "Raw Inputs"
- "Inputs: PICMI Python Script": "PICMI Inputs"

Before this PR:
Screenshot from 2025-11-06 17-01-48

After this PR:
Screenshot from 2025-11-06 17-02-49

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -1,7 +1,7 @@
 .. _running-cpp-parameters:
 
-Inputs: Parameter List
-======================
+Raw Inputs
+==========
 
 This documents on how to use WarpX with an inputs file (e.g., ``warpx.3d input_3d``).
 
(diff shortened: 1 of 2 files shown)


ax3l (changes requested):
Oh, I thought we converged on the naming in the first screenshot during the hackathon :sweat_smile: 

I don't think I like the naming proposed here :sweat_smile: 
Let's discuss in dev meeting? :)

My main concern is that `raw` is not as descriptive as Parameter List.
My secondary concern is that I liked the prefixes being identical, putting them structurally on the same level by eye (vs. suffixes that hang around)

EZoni:
I think we can just vote, I'm not sure we need to discuss. I'll try to poll the TC either here or on Slack sometime soon!

PR #3657, opened 2023-01-31
ADIOS2: Fix BTD Particle Resize w/ Empty Ranks

ax3l:
For ADIOS2 BP4 and BP5 particle BTD writes, setting an update of a variable's (openPMD record's) shape is not sufficient to drop all meta-data on disk. In situations where the last write to a BTD stage only adds further particles from a few ranks, we need to pad with zero-block writes so ADIOS2 `Put` gets called.

Backend details:
- BP4 (as of ADIOS 2.8): last MPI rank's `Put` meta-data wins
- BP5 (as of ADIOS 2.8): everyone has to write an empty block

## Tests

6 MPI ranks w/ CUDA on Summit  with read-back as in https://github.com/ECP-WarpX/WarpX/issues/3389

- Inputs: [inputs.txt](https://github.com/ECP-WarpX/WarpX/files/10552060/inputs.6.txt)
- Run: `jsrun -r 6 -a 1 -g 1 -c 7 -l GPU-CPU -d packed -b rs --smpiargs="-gpu" ./warpx.RZ.MPI.CUDA.DP.PDP.OPMD.PSATD.QED inputs.txt`
- Test command: `python3 -c "from openpmd_viewer import OpenPMDTimeSeries; ts = OpenPMDTimeSeries('./diags/diag_btd/'); ts.get_particle(['x'], iteration=1)"`

Results:
- [x] reproduced on `development`
- [x] same test, with PR applied, passed
- [ ] same test, with PR applied, and with compression operators (blosc) enabled, passed
  - [x] ADIOS 2.7.1: fails during write
  - [x] ADIOS 2.8.1: fails during write
  - [x] ADIOS 2.8.3 w/ c-blosc 1.21.0: fails during write https://github.com/ornladios/ADIOS2/issues/3460
  - [x] ADIOS `master` as of `98bef0ee88838c034832a1e280dfd555841462dc` w/ c-blosc 2.6.1: also fails

Requires this patch in openPMD-viewer: https://github.com/openPMD/openPMD-viewer/pull/355

## Related Issues

- fix #3389
- https://github.com/ECP-WarpX/WarpX/pull/1898#discussion_r745008290
- https://github.com/ornladios/ADIOS2/issues/3455
- https://github.com/openPMD/openPMD-api/issues/1147

PR #6377, opened 2025-11-10
BTD Resize ADIOS for `int`

### LINKED ISSUE #6376: BTD Issue in ADIOS2
In the latest `development`, Chris (PhD student at BELLA) gets the following error with a BTD simulation of WarpX:
```
and it runs for ~900 steps and then I get an error about the back transformed diagnostics:

STEP 902 starts ...
--- INFO    : Writing buffer 3 of 16 to snapshot 2 to openPMD BTD diags/diag1
amrex::Abort::0::
### ERROR   : WarpX openPMD ADIOS2 work-around has unknown dtype: INT
#            (/global/homes/c/cmccombs/src/warpx/Source/Diagnostics/WarpXOpenPMD.cpp:845)
 !!!
SIGABRT

it runs fine on 1 gpu
```

attn @guj , could be related to our recent changes in #6123

ax3l:
Fix #6376

Follow-up to #3657

### DIFF
--- a/Source/Diagnostics/WarpXOpenPMD.cpp
+++ b/Source/Diagnostics/WarpXOpenPMD.cpp
@@ -824,6 +824,11 @@ WarpXOpenPMDPlot::DumpToFile (ParticleContainer* pc,
                         comp.storeChunk(empty_data, {uint64_t(0)}, {uint64_t(0)});
                         break;
                     }
+                    case openPMD::Datatype::INT : {
+                        auto empty_data = std::make_shared<int>();
+                        comp.storeChunk(empty_data, {uint64_t(0)}, {uint64_t(0)});
+                        break;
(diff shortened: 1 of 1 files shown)


ax3l:
@EZoni I am CC-ing your for review because Remi is on PTO for part of this week. He reported the issue to me on Slack.

EZoni (approved):

guj (approved):
Thanks for the catch.  Do we need to include other likely types as well?

ax3l:
I was wondering if we need to add unsigned long ints, yes.

I think this happened here because we allow user-specified `int` attributes (programmatic and in input). So at least for now, there is no likely other type appearing.

We should generally revisit if the work-around is still needed that had receives this patch.

PR #6381, opened 2025-11-11
refactor direct deposit for Mass Matrices.

JustinRayAngus:
This PR makes a few modifications to the direct mass matrix deposition routine.

1) The switching between doing a full mass matrix deposition and just depositing the diagonal elements of the diagonal mass matrices for the preconditioner (PC) is now controlled with a template parameter just as it is in the villasenor deposition. This removes an if check inside hot loops.

2) The full mass matrices are deposited using the product of shape factors for current and field interpolation (`shape_J x shape_E`). In other words, for each particle location where current J is deposited, the code loops over all field interpolation points E (the same locations used for J) and deposits the corresponding mass matrix contribution with `shape_J x shape_E`. In the current development branch, when only the diagonal elements of the diagonal mass matrices are used for the PC, the deposition still uses `shape_J x shape_E`, which is non-conservative. With this PR, when operating in this way, the deposition instead uses only the current deposition shape factor (`shape_J`). This makes the deposition conservative, consistent with the J deposition scheme. Mathematically, this is equivalent to depositing to all E-location for each J-location (as is done for the full mass matrices) and summing the contributions. @RemiLehe

### DIFF
--- a/Source/Particles/Deposition/MassMatricesDeposition.H
+++ b/Source/Particles/Deposition/MassMatricesDeposition.H
@@ -90,16 +90,14 @@ void setMassMatricesKernels ( const amrex::ParticleReal qs,
 
 /**
  * \brief Kernel for the direct deposition of J and S (mass matrices) for thread thread_num
- * \tparam depos_order deposition order
+ * \tparam depos_order             deposition order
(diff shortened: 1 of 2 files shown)


roelof-groenewald (approved):
This way is definitely more readable which is a nice benefit on top of the performance improvement :tada:

PR #6383, opened 2025-11-12
Curl2 matrix BC masks

JustinRayAngus:
Add masks to implement the boundary conditions in the curl curl matrix used for the PETSc preconditioner.

- [x] Add comments explaining the masking logic
- [x] Add CI test


JustinRayAngus:
@EZoni I've added a new 2D CI tests that uses PETSc, but it doesn't work because the PETSc flag is not defined when building for the CI tests. What should we do about this?

Source/FieldSolver/ImplicitSolvers/ThetaImplicitEM.cpp
```diff
@@ -175,3 +178,226 @@ void ThetaImplicitEM::FinishFieldUpdate ( amrex::Real end_time )
...
+#elif defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
+        const int ncomps_Ex = 3;
+        const int ncomps_Ey = 4;
+        const int ncomps_Ez = 3;
+#elif defined(WARPX_DIM_3D)
+        const int ncomps_Ex = 6;
+        const int ncomps_Ey = 6;
+        const int ncomps_Ez = 6;
+#endif
+        m_curl2_BC_mask[lev][0] = std::make_unique<amrex::MultiFab>(ba_Ex, dm, ncomps_Ex, ghosts);
```
dpgrote: Could these be put in multifab register?

Source/FieldSolver/ImplicitSolvers/ThetaImplicitEM.cpp
```diff
@@ -175,3 +178,231 @@
     m_WarpX->FinishMagneticFieldAndApplyBCs( B_old, m_theta, end_time );
 
 }
+
+const amrex::MultiFab* ThetaImplicitEM::GetCurl2BCmask (const int lev, const int field_dir) const
+{
+    using ablastr::fields::Direction;
+    const amrex::MultiFab* mask = m_WarpX->m_fields.get(FieldType::curl2_BC_mask, Direction{field_dir}, lev);
+    //return m_WarpX->m_fields.get(FieldType::curl2_BC_mask, Direction{field_dir}, lev);
```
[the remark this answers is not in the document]
JustinRayAngus: @ax3l The commented out code here is what I originally tried, but caused the code to crash on the first call here with an out-of-memory error. Any idea why the following does not work?
`return m_WarpX->m_fields.get(FieldType::curl2_BC_mask, Direction{field_dir}, lev);`

Examples/Tests/implicit/CMakeLists.txt
```diff
@@ -31,6 +31,16 @@ add_warpx_test(
     OFF  # dependency
 )
 
+add_warpx_test(
```
ax3l: Let's add a `if(WarpX_PETSC)` around this test

JustinRayAngus:
[ ] https://github.com/BLAST-WarpX/warpx/blob/development/.azure-pipelines.yml add https://packages.ubuntu.com/search?keywords=libpetsc&searchon=names&suite=questing&section=all & flags

@ax3l It's not clear to me what I'm supposed to do here. Can you elaborate?

Here is how I would normally build with CMake when using PETSc:
`cmake -S . -B build_2D_petsc -DWarpX_DIMS=2 -DAMReX_PETSC=YES -DPETSC_DIR=${PETSC_DIR} -DPETSC_ARCH=${PETSC_ARCH}`

EZoni:
@JustinRayAngus

Some info on the CI configuration topic. 

This is the step where we configure and compile WarpX before running the CI tests in the Azure pipelines:
https://github.com/BLAST-WarpX/warpx/blob/83cfdd028a5aba612f1d03cb0cf22c432f0f98b8/.azure-pipelines.yml#L142-L158

You see that the configuration relies on the environment variables `AMReX_CMAKE_FLAGS`, `WARPX_CMAKE_FLAGS`, and `WARPX_TEST_FLAGS`. 

`AMReX_CMAKE_FLAGS` and `WARPX_TEST_FLAGS` are set right above the `cmake` command, while `WARPX_CMAKE_FLAGS` is set here, depending on which pipeline we are running:
https://github.com/BLAST-WarpX/warpx/blob/83cfdd028a5aba612f1d03cb0cf22c432f0f98b8/.azure-pipelines.yml#L39-L53

In your case it looks like the extra build options are
- `-DAMReX_PETSC=YES` 
- `-DPETSC_DIR=${PETSC_DIR}`
- `-DPETSC_ARCH=${PETSC_ARCH}`

I would maybe add them to `AMReX_CMAKE_FLAGS`, since at least the first one is indeed an AMReX flag, so here:
https://github.com/BLAST-WarpX/warpx/blob/83cfdd028a5aba612f1d03cb0cf22c432f0f98b8/.azure-pipelines.yml#L147-L149

The outstanding question for me is on the environment variables `PETSC_DIR` and `PETSC_ARCH` (so the right hand side of the build options `-DPETSC_DIR` and `-DPETSC_ARCH`). I'm not sure how those are supposed to be set. How do you set them locally?

@ax3l Please feel free to chime in and comment/correct my suggestion.

Examples/Tests/implicit/CMakeLists.txt
```diff
@@ -31,6 +31,18 @@ add_warpx_test(
     OFF  # dependency
 )
 
+if(WarpX_PETSC)
```
ax3l: Wait, is there even a `WarpX_PETSC` CMake option yet?

ax3l: This should work for now:
```suggestion
if(AMReX_PETSC)
```

Do we use PETSC in WarpX at all or just _through_ AMReX?

JustinRayAngus:
@EZoni Thanks for the explanation. I agree that the AMReX_PETSC flag should go in the AMReX_CMAKE_FLAGS. The other flags would be set depending on where PETSc is configured and compiled. It is still not clear to me how to accomplish that. Perhaps @ax3l @WeiqunZhang @debog can answer that.

WeiqunZhang:
@JustinRayAngus Could you add this to your PR? It's unrelated to this PR. But it's just a one-line change that is needed for PETSc in debug mode.
```
--- a/Source/NonlinearSolvers/WarpX_PETSc.cpp
+++ b/Source/NonlinearSolvers/WarpX_PETSc.cpp
@@ -341,6 +341,7 @@ void PETScSolver_impl::applyPC( VecType& a_F,
 //! Assemble preconditioner matrix
 int PETScSolver_impl::assemblePCMatrix(LinOpType* const a_linop)
 {
+    PetscFunctionBeginUser;
     BL_PROFILE("PETScSolver_impl::assemblePCMatrix()");
 
     AMREX_ALWAYS_ASSERT(m_is_defined);
```

JustinRayAngus:
> PetscFunctionBeginUser;

Sure. I believe there is also a PetscFunctionEndUser function. Should I add that too?

WeiqunZhang:
> > PetscFunctionBeginUser;
> 
> Sure. I believe there is also a PetscFunctionEndUser function. Should I add that too?

No. The issue is there is a PetscFunctionEndUser already without Begin.

EZoni:
@JustinRayAngus 

What is the status of this PR? Would you like to ping specific reviewers? I'm not a reviewer here, but the PR came up while filtering the `KISMET` label.

JustinRayAngus:
> @JustinRayAngus
> 
> What is the status of this PR? Would you like to ping specific reviewers? I'm not a reviewer here, but the PR came up while filtering the `KISMET` label.

@EZoni 

This PR is ready to go and it is a bottleneck holding up additional PRs related to the implicit solver. @WeiqunZhang would be a good reviewer for this.

Examples/Tests/implicit/analysis_petsc_matrix.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python3
+
+# This file is part of WarpX.
+#
+# License: BSD-3-Clause-LBNL
+#
+# This is a script that analyses the simulation results from
+# and script `inputs_test_2d_curl_curl_petsc_pc`.
```
EZoni: Something you can patch if you have more changes to make:
```suggestion
# the input file `inputs_test_2d_curl_curl_petsc_pc`.
```

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
```diff
@@ -52,15 +52,15 @@ const Array<FieldBoundaryType,AMREX_SPACEDIM>& ImplicitSolver::GetFieldBoundaryH
 
 Array<LinOpBCType,AMREX_SPACEDIM> ImplicitSolver::GetLinOpBCLo () const
 {
-    return convertFieldBCToLinOpBC(m_WarpX->GetFieldBoundaryLo());
+    return convertFieldBCToLinOpBC(m_WarpX->GetFieldBoundaryLo(),0);
```
EZoni: This could make the call more readable:


```suggestion
    return convertFieldBCToLinOpBC(m_WarpX->GetFieldBoundaryLo(), /*bdry_side=*/0);
```

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
```diff
@@ -52,15 +52,15 @@ const Array<FieldBoundaryType,AMREX_SPACEDIM>& ImplicitSolver::GetFieldBoundaryH
...
 }
 
 Array<LinOpBCType,AMREX_SPACEDIM> ImplicitSolver::GetLinOpBCHi () const
 {
-    return convertFieldBCToLinOpBC(m_WarpX->GetFieldBoundaryHi());
+    return convertFieldBCToLinOpBC(m_WarpX->GetFieldBoundaryHi(),1);
```
EZoni: This could make the call more readable:

```suggestion
    return convertFieldBCToLinOpBC(m_WarpX->GetFieldBoundaryHi(), /*bdry_side=*/1);
```

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
```diff
@@ -82,10 +82,15 @@ Array<LinOpBCType,AMREX_SPACEDIM> ImplicitSolver::convertFieldBCToLinOpBC (const
...
-                ablastr::warn_manager::WarnPriority::medium);
-            lbc[i] = LinOpBCType::symmetry;
+            const int voltage_driven = m_WarpX->GetPECInsulator_IsESet(i,bdry_side);
+            if (voltage_driven) { // Dirichlet for E
+                lbc[i] = LinOpBCType::Dirichlet;
+            } else { // Dirichlet for B
+                ablastr::warn_manager::WMRecordWarning("Implicit solver with current-driven PECInsulator",
+                    "in the Curl-Curl preconditioner. Neumann boundary will be used since the full boundary is not yet implemented.",
+                    ablastr::warn_manager::WarnPriority::medium);
+                lbc[i] = LinOpBCType::symmetry;
```
EZoni, line 92: So `LinOpBCType::symmetry` means Neumann boundary conditions currently?

Source/FieldSolver/ImplicitSolvers/ThetaImplicitEM.cpp
```diff
@@ -175,3 +178,245 @@ void ThetaImplicitEM::FinishFieldUpdate ( amrex::Real end_time )
...
+        const int ncomps_Ez = 0;
+#elif defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
+        const int ncomps_Ex = 3;
+        const int ncomps_Ey = 4;
+        const int ncomps_Ez = 3;
+#elif defined(WARPX_DIM_3D)
+        const int ncomps_Ex = 6;
+        const int ncomps_Ey = 6;
+        const int ncomps_Ez = 6;
+#endif
```
EZoni: Does this need special treatment for the 1D cylindrical and spherical geometry?

Source/FieldSolver/ImplicitSolvers/ThetaImplicitEM.cpp
```diff
@@ -175,3 +178,245 @@ void ThetaImplicitEM::FinishFieldUpdate ( amrex::Real end_time )
...
+
+        for (int bdry_dir = 0; bdry_dir < AMREX_SPACEDIM; ++bdry_dir) {
+
+            if (bc_type_lo[bdry_dir] == FieldBoundaryType::Periodic) { continue; }
+
+            for (int field_dir = 0; field_dir < 3; ++field_dir) {
+
+#if AMREX_SPACEDIM == 3
+                if (field_dir == bdry_dir) { continue; }
+                const int tdir1 = field_dir + 1 % 3; // next direction after field_dir
```
EZoni: If the `3` here is in fact `AMREX_SPACEDIM`, would it be useful to write it that way?

```suggestion
                const int tdir1 = field_dir + 1 % AMREX_SPACEDIM; // next direction after field_dir
```

Source/NonlinearSolvers/MatrixPC.H
EZoni, line 1: Would it be useful to use an auxiliary variable for the fourth component of the array `BC_mask_Edir_arr`, wherever it is changed in this file? I see the values currently passed are: `0`, `1`, `2`, `jdir+1`, `comp_shift+1`, `comp_shift+2`. I wonder if giving these a common name could make the code more readable.

EZoni (commented):
@JustinRayAngus 

I didn't check the algorithm because I would need to first learn about it, but I skimmed through the code changes and left a few easy comments or questions.

Source/FieldSolver/ImplicitSolvers/ThetaImplicitEM.cpp
JustinRayAngus: Technically, yes. My strategy for now has been to get everything working for planar geometries, then go back and make it work for non-planar geometries. The code will complain right now if trying to use PETSc in non-planar geometries.

In either case, the change needed here for 1D spherical and 1D cylindrical is easy to do and I have implemented it.

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
JustinRayAngus: I think this question stems from an inconsistent naming convention between WarpX and AMReX. For electromagnetic (EM), Neumann in WarpX is actually a Symmetry boundary. Neumann implies zero gradient (or specified gradient), but for EM it is zero gradient for the electric field parallel to the boundary, but zero value for fields normal to the boundary.

To be specific, if you specify Neumann (or PMC) as the boundary condition in WarpX and you are using an electromagnetic solver, then a symmetry BC is used for the E and B fields.

Source/NonlinearSolvers/MatrixPC.H
JustinRayAngus: Perhaps.

See the detailed description of the mask components given at the beginning of the following routine:
`ThetaImplicitEM::InitializeCurlCurlBCMasks()`

EZoni:
The CI test `test_2d_python_wrappers_picmi` will be fixed with the next weekly update, so no need to worry about it.

Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp
EZoni: Understood, so it fits within the overall issue of WarpX boundary conditions, which I thought we were tracking somewhere, although I can't find a single tracking issue right now.

Source/NonlinearSolvers/MatrixPC.H
EZoni: So basically the first three indices are the usual spatial indices and the fourth index is a mask index that tracks which derivative of which component of the operator that mask is for, correct?

WeiqunZhang:
AMReX has a class Orientation that might be useful here. https://amrex-codes.github.io/amrex/doxygen/AMReX__Orientation_8H.html

You can do things like
```
for (OrientationIter oit; oit; oit++) {
    Orientation ori = oit(); // this contains both direction and side information
    int cdir = ori.coordDir(); // 0, 1 or 2
    Side side = ori.faceDir(); // 0 or 1, i.e., lo or hi
    if (ori.isLow()) {.. } // lo face. there is also isHig
    int idx = ori; // 0, 1, 2, ..., 5 in 3d.
    Box bbox = amrex::bdryNode(box, ori);
}

for (int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
    Orientation lo_face(idim, Orientation::low);
    Orientation hi_face(idim, Orentation::high);
    Orientation lo_face_again(idim, 0);
    Orientation hi_face_again(idim, 1);
}
```

Source/NonlinearSolvers/MatrixPC.H
JustinRayAngus: Yes. Perhaps it is easiest just to add a comment in this section to refer the code reader to the description given in InitializeCurlCurlBCMasks().

JustinRayAngus:
@WeiqunZhang 

Thanks for the comment. I'm imagining that you are suggesting it to potentially use for the component index of `BC_mask_Edir_arr` in response to the comment by @EZoni above? This looks very useful for many situations, but I'm not sure it is needed here. Take the 1D case for example for second derivative operator where the stencil is (1 -2 1). The first component is the mask for the co-located term (-2) of the operator and the second component is the mask for the nearest neighbor term (1), and we only ever need two components because we only need a mask for the side the reaches out to the boundary. It is not clear to me how I could make use of an `Orientation` object to make things simpler here.

JustinRayAngus:
Screenshot 2026-01-21 at 2 30 35 PM
??

@WeiqunZhang @EZoni @ax3l 

This pull request was closed?

I think I may have accidentally done something wrong...

WeiqunZhang:
My comment was not about `BC_mask_Edir_arr`. You don't need to use it. I only suggested as a minor cleanup. For example,
```
// Create a edge box that only contains locations right on the boundary
amrex::Box bdry_box = edge_box;
if (bdry_side == 0) { bdry_box.setBig(bdry_dir,domain_lo[bdry_dir]); }
if (bdry_side == 1) { bdry_box.setSmall(bdry_dir,domain_hi[bdry_dir]); }
```
could be
```
amrex::Box bdry_box = amrex::bdryNode(edge_box, Orientation(bdry_dir,bdry_side));
```
Or you don't even need to construct Orientation explicitly if you use Orientationiter instead of looping over directions and then sides with nested for-loops.

Or this
```
int GetPECInsulator_IsESet ( const int  bdry_dir, const int  bdry_side ) const;
```
could be
```
int GetPECInsulator_IsESet ( Orientation ori ) const;
```
Since it's just a single argument, it's less likely to get it wrong like swapping the parameters.

Again, you don't need to do it. Just a minor suggestion.

WeiqunZhang (approved):

JustinRayAngus:
> My comment was not about `BC_mask_Edir_arr`. You don't need to use it. I only suggested as a minor cleanup. For example,
> 
> ```
> // Create a edge box that only contains locations right on the boundary
> amrex::Box bdry_box = edge_box;
> if (bdry_side == 0) { bdry_box.setBig(bdry_dir,domain_lo[bdry_dir]); }
> if (bdry_side == 1) { bdry_box.setSmall(bdry_dir,domain_hi[bdry_dir]); }
> ```
> 
> could be
> 
> ```
> amrex::Box bdry_box = amrex::bdryNode(edge_box, Orientation(bdry_dir,bdry_side));
> ```
> 
> Or you don't even need to construct Orientation explicitly if you use Orientationiter instead of looping over directions and then sides with nested for-loops.
> 
> Or this
> 
> ```
> int GetPECInsulator_IsESet ( const int  bdry_dir, const int  bdry_side ) const;
> ```
> 
> could be
> 
> ```
> int GetPECInsulator_IsESet ( Orientation ori ) const;
> ```
> 
> Since it's just a single argument, it's less likely to get it wrong like swapping the parameters.
> 
> Again, you don't need to do it. Just a minor suggestion.

Ah, I understand. That is useful. For sake of moving forward, I think it's best to save this for a future PR. There will be follow on PR's to make this work with axisymmetric geometries. I can refactor to use the Orientation object in that PR.

PR #6388, opened 2025-11-13
Add "watchdog" script on Frontier runs

### LINKED ISSUE #5584: HPC: Monitoring Progress & Aborting
In unsupervised (or: _ideally they should be unsupervised, but I am checking the status every other minute_) runs such as batched HPC execution, it is preferable that the operator does not have to be synchronously monitoring a running job.

In real-world loads of HPC systems (changing loads on networks, filesystems, OoM scenarios, changing software, etc.), it is not uncommon that a batched job starts up outside of regular working hours and in some cases, causes a hang until walltime. In some cases, this can be costly.

We should establish a mechanism (in job scripts) that programatically monitors progress / health of a simulation and if a configurable timeout is reached, aborts the simulation, first with sigterm (for backtrace generation) and then sigkill.

## Possible Implementations

A very simple implementation could be to write some kind of status (e.g., the current time) into a file (e.g., from the I/O processor) every time step. In the batch job, a single polling process could check the time difference.

File-based I/O is of course far from ideal, e.g., due to sync, load, short time steps, or for I/O-free runs (e.g., optimization). Better might be to have a port open for health queries (could be later reused to query things like memory usage per MPI process, load, etc.) or to react on a POSIX signal and print something on a specific channel (e.g., stderr), like `dd` does.

titoiride:
In this PR we add a script that checks periodically if the code execution on Frontier is proceeding as expected.
In case the code is hanging, which is verified by checking the output file timestamp, the execution is killed to save computational time.

Compared to previous attempts, this script has a double timer: a frequent check and an output modification check. The check is ran frequently (every `check_interval` seconds) and checks that the output file has been modified not more than `timeout_sec` seconds before. Differentiating the timers, compared to just running a single check, prevents the script to stall a simulation that, for instance, finishes correctly between two checks. This way, if the simulation ends between two `timeout_sec` checks, the script will wait at most `check_interval` before closing.

Close #5584

### DIFF
--- a/Tools/machines/frontier-olcf/submit.sh
+++ b/Tools/machines/frontier-olcf/submit.sh
@@ -74,7 +74,7 @@ check_interval=200  # check every 200 seconds
 while kill -0 "$srun_pid" 2>/dev/null
 do
     sleep ${check_interval}
-    
+
     # Check if output file has been modified recently
     if [[ -f ${WARPX_OUTPUT} ]]; then
         file_mtime=$(stat -c %Y ${WARPX_OUTPUT})
--- a/Tools/machines/frontier-olcf/submit.sh
+++ b/Tools/machines/frontier-olcf/submit.sh
@@ -45,7 +45,7 @@ export ROCFFT_RTC_CACHE_PATH=/dev/null
 export OMP_NUM_THREADS=1
 export WARPX_NMPI_PER_NODE=8
 export TOTAL_NMPI=$(( ${SLURM_JOB_NUM_NODES} * ${WARPX_NMPI_PER_NODE} ))
-export WARPX_OUTPUT=output.txt
+export WARPX_OUTPUT=output_${SLURM_JOB_ID}.txt
 
 # For large-scale simulations, uncomment this line and manually set the lfs striping
 # Incorrect striping could result in very slow file writing and simulations hanging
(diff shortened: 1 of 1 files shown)


Tools/machines/frontier-olcf/submit.sh
```diff
@@ -45,5 +45,61 @@ export ROCFFT_RTC_CACHE_PATH=/dev/null
 export OMP_NUM_THREADS=1
 export WARPX_NMPI_PER_NODE=8
 export TOTAL_NMPI=$(( ${SLURM_JOB_NUM_NODES} * ${WARPX_NMPI_PER_NODE} ))
-srun -N${SLURM_JOB_NUM_NODES} -n${TOTAL_NMPI} --ntasks-per-node=${WARPX_NMPI_PER_NODE} \
-    ./warpx inputs > output.txt
+export WARPX_OUTPUT=output.txt
```
n01r: Maybe let's include the job id to avoid overwriting on restarts?
```suggestion
export WARPX_OUTPUT=output_${SLURM_JOB_ID}.txt
```

n01r (approved):

Tools/machines/frontier-olcf/submit.sh
titoiride: I mean, I guess that's up to the user if they need restart? This is supposed to be a sample and I don't want it to be too complicated, but also this is not necessarily an issue to add

PR #6393, opened 2025-11-14
Array PEC Warning in nvcc

ax3l:
Fix a warning with `nvcc` in CUDA 12.4 (Perlmutter) for initialization.

This can be written w/o a for loop, so this is ok to silence the compiler diagnostics (which is misbehaving).


dpgrote (approved):
Looks good to me

Issue #6392, opened 2025-11-14
Diagnostics: Geometry Wrong After Restart


ax3l:
In a large scale case of @titoiride, the checkpoint file has the wrong upper bound of the domain in `checkpoint*/WarpXHeader` lines 10 & 11.

The simulation uses:
- load balancing
- a moving window that starts and stops
- checkpoint-restart: after the 2nd restart it gets the bounds in the WarpX header wrong
- the issue appears with a restart after the moving window stopped

We still work on a reproducer.

ax3l:
One thing we want to update is the logic for the upper bound here in write:

https://github.com/BLAST-WarpX/warpx/blob/25.11/Source/Diagnostics/FullDiagnostics.cpp#L761-L774

Instead of taking the first box from `ba = diag_ba.maxSize()` we should use `ba.minimalBox()` as we do below:
https://github.com/BLAST-WarpX/warpx/blob/25.11/Source/Diagnostics/FullDiagnostics.cpp#L802

Submitting a PR for this in #6394, which should help to avoid corrrupted checkpoints with load balancing.

ax3l:
We have a reproducer now (file is still a bit with sim details, on Slack).

```
export OMP_NUM_THREADS=4

$ ./warpx.3d minimal.input

# still right: (size in z is 1um)
$ head -n 12 checkpoint000150/WarpXHeader
...
-4.9999999999999998e-07 -4.9999999999999998e-07 -3.2812499999999993e-07 
4.9999999999999998e-07 4.9999999999999998e-07 6.718749999999995e-07 
...

$ ./warpx.3d minimal.input amr.restart=checkpoint000100

# wrong: size in z is now 0.375um)
$ head -n 12 checkpoint000150/WarpXHeader
...
-4.9999999999999998e-07 -4.9999999999999998e-07 1.4062499999999969e-07 
4.9999999999999998e-07 4.9999999999999998e-07 5.1562499999999919e-07 
...
```

ax3l:
- If I set `warpx.end_moving_window_step = 0` (i.e. never) the checkpoints after restart stay correct.
- If I set `warpx.start_moving_window_step = 0` and keep `warpx.end_moving_window_step = 30` the checkpoints after restart has WRONG meta data.

So the issue is likely with checkpoint-restart after `warpx.end_moving_window_step`.

titoiride:
Not sure about that, I had wrong checkpoint boundaries even during the moving window motion, before it stopped

[...]

PR #1241, opened 2020-08-10
Defining geom object for diags

RevathiJambunathan:
In this PR, geometry objects are defined for diagnostics such that the user-defined physical co-ordinates for the diagnostics are taken into account. 

Without this addition, for a slice input :
```
amr.n_cell =  64 32 32
amr.max_grid_size = 32
amr.blocking_factor = 16
amr.max_level = 0
geometry.coord_sys   = 0                  # 0: Cartesian
geometry.is_periodic = 1     1     1      # Is periodic?
geometry.prob_lo     = -20.e-6   -20.e-6   -20.e-6    # physical domain
geometry.prob_hi     =  20.e-6    20.e-6    20.e-6

#################################
############ NUMERICS ###########
#################################
warpx.serialize_ics = 1
warpx.verbose = 1
warpx.cfl = 1.0

#################################
############ PLASMA #############
#################################
particles.species_names = electrons

electrons.species_type = electron
electrons.injection_style = "NUniformPerCell"
electrons.num_particles_per_cell_each_dim = 1 1 2
electrons.profile = constant
electrons.density = 1.e25  # number of electrons per m^3
electrons.momentum_distribution_type = "gaussian"
electrons.ux_th  = 0.01 # uth the std of the (unitless) momentum
electrons.uy_th  = 0.01 # uth the std of the (unitless) momentum
electrons.uz_th  = 0.01 # uth the std of the (unitless) momentum


diagnostics.diags_names = slice
slice.period = 4
slice.diag_type = Full
slice.diag_lo = -20.e-6 0.0 -20.e-6
slice.diag_hi =  20.e-6 0.0  20.e-6
```
When visualizing the slice plotfile in yt using 
```
ds = yt.load( plotfile )
ds.domain_dimensions
```
resulted in [64,32,32] instead of reduced [64,1,32]

Now, after defining the geometry object in this PR and using it to writeplotfiles, 
we get [64,1,32], where, the dimension is reduced in y to two cells.

@MaxThevenet thank you for using this feature and letting me know about the error.

PR #1085, opened 2020-06-11
New BTD class layout and member functions - Part 1

RevathiJambunathan:
This is the first PR in the series "reformat back-transformed Diagnostics" 

This PR includes the following changes to diagnostics.

- Created a new Back-transformed diagnostics class::BTDiagnostics in Source/Diagnostics/BTDiagnostics.H 

- A new functor "class::BackTransformFunctor" is included in Source/Diagnostics/ComputeDiagFunctor/BackTransformFunctor.H. The main objective of this functor is to compute the core part of BTD implementation, where, cell-centered data is sliced for a given snapshot and back-transformed to lab-frame.

- A lot of the implementation in the Diagnostics class was specific to FullDiagnostics. So I have moved functions relevant only to FullDiagnostics to class::FullDiagnostics.

- Now, the layout looks like this :
  * Base class::Diagnostics
  * Two derived classes: class::FullDiangotics and class::BTDiagnostics
  * Member variables, such as, vector of multifabs storing output buffers on multiple levels, which are common to both derived classes are in the Diagnostics class.
   * Similarly, Functions common to both dervied classes are also in the Base class.
   * In the new BTDiagnostics class, member variables required to construct the class and build the back-transformed data is included.
   * The functions to initialize these vectors/varaibles are currently empty with comments describing the main objective of the that function or steps to be included in the function. (These functions will be "filled" with implementation in subsequent PRs)

addition to the input file to perform Back-Transformed diagnostics

```
diagnostics.diags_names = diag1 diag2

diag1.period = 2
diag1.diag_type = Full

diag2.diag_type = BackTransformed
diag2.do_back_transformed_fields = 1
diag2.num_snapshots_lab = 4
diag2.dz_snapshots_lab = 0.001
diag2.fields_to_plot = Ex Ey Ez By rho
```

PR #6394, opened 2025-11-14
Checkpoint/Plotfile: Robust Bounds Calc

ax3l:
The calculation of the bounds in plotfiles (incl. checkpoints) appears fishy. It relies on the order of the boxes in the box array, which is not always pointing to a box on the lower/upper bound of the simulation when load balancing is used.

This can corrupt checkpoints (plotfiles) when writing, by a) filtering out valid particles & fields and b) remembering the wrong simulation meta-data for the simulation geometry of fields/particles.

Related to #6392

Co-authored-by: @atmyers 
Co-authored-by: @RemiLehe 
Co-authored-by: @titoiride 

Introduced in #1085 (refined in #1241)

## Follow-up

`BTDiagnostics.cpp` contains similar assumptions/logic it seems and needs a similar fix.

### DIFF
--- a/Source/Diagnostics/FullDiagnostics.cpp
+++ b/Source/Diagnostics/FullDiagnostics.cpp
@@ -764,13 +764,16 @@ FullDiagnostics::InitializeBufferData (int i_buffer, int lev, bool restart ) {
         // Coarsen and refine so that the new BoxArray is coarsenable.
         ba.coarsen(m_crse_ratio).refine(m_crse_ratio);
 
+        // Box covering the extent of the user-defined diagnostic domain
+        amrex::Box const domain = ba.minimalBox();
+
         // Update the physical co-ordinates m_lo and m_hi using the final index values
         // from the coarsenable, cell-centered BoxArray, ba.
         for ( int idim = 0; idim < AMREX_SPACEDIM; ++idim) {
             diag_dom.setLo( idim, warpx.Geom(lev).ProbLo(idim) +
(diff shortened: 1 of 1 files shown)


titoiride:
Is this ready to be tested at full scale or do you need to make some more changes?

ax3l:
I tested it on my minimal reproducer and it does not fix the issue we observe. Nonetheless, I would change the logic here because it does not look safe under load balancing.

Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -764,13 +764,16 @@ FullDiagnostics::InitializeBufferData (int i_buffer, int lev, bool restart ) {
         // Coarsen and refine so that the new BoxArray is coarsenable.
         ba.coarsen(m_crse_ratio).refine(m_crse_ratio);
 
+        // Box covering the extent of the user-defined diagnostic domain
+        amrex::Box const domain = ba.minimalBox();
```
WeiqunZhang:
```suggestion
        amrex::Box domain = diag_box;
        domain.coarsen(m_crse_ratio).refine(m_crse_ratio);
```
It's cheaper to do it this way. `ba.minimalBox()` needs to loop over all boxes in the boxarray.

WeiqunZhang:
I agree with the change. Although the existing code does not produce incorrect results, it should be avoided. It has nothing to do with load balancing. But it depends on how amrex does `BoxArray::maxSize`. If the implementation in amrex changes in the future, it might break.

atmyers (approved):

PR #6397, opened 2025-11-16
[Do not merge] Temporary fix for checkpoints

RemiLehe:
This addresses #6392 but cannot be merged as is, since it breaks some of the WarpX functionalities.


RemiLehe:
Hm, the fact that the CI tests are passing is worrisome!

PR #6399, opened 2025-11-16
Diagnostics: Fix Restart with Start/Stop Moving Window

### LINKED ISSUE #6392: Diagnostics: Geometry Wrong After Restart

### LINKED ISSUE #6944: Checkpoints and restarts fail with any moving window
Simulation setup:
A minimal moving window, without boosted frame sumulations.

After restart the simulation box seems to be weird.  I can't get it to work.

The simulation box shrinks after the restart. 

Attached are 2 simulation outputs: Just before restart at iteration 160, and another at iteration 166 restarted from the checkpoint generated at iter 160.

Image

Image

Image

Attached is a minimal example with a moving window in the z direction (the window always moves). Also attached are the headers copied from the checkpoint folder.
[CheckpointRestart.zip](https://github.com/user-attachments/files/28911035/CheckpointRestart.zip)

If i look at the checkpoint headers before the restart (the z box size is 500 um) After the restart the box size in the header is suddenly smaller.
Iteration 160 box length = 500 um
Restart from checkpoint = Box length is less than 500 um (also the number of cells! as shown in the picture above), but the job info shows the correct grid.

It seems like saving checkpoints is working well, but it's when it restarts that something seems odd.

This issue seems to be different, already pointed out in (but im not sure if they are completely related) [without a moving window the restart seems fine as pointed out below]

https://github.com/BLAST-WarpX/warpx/issues/6392 
https://github.com/BLAST-WarpX/warpx/pull/6399

Attached is the header:: 
```

Initializing AMReX (26.05)...
MPI initialized with 4 MPI processes
MPI initialized with thread support level 3
Initializing CUDA...
CUDA initialized with 4 devices.
AMReX (26.05) initialized

PICSAR (26.05)
WarpX (26.05)

                                   ___
    __        __             ___  /  /
    \ \      / /_ _ _ __ _ __\  \/  /
     \ \ /\ / / _` | '__| '_ \\    /
      \ V  V / (_| | |  | |_) /    \
       \_/\_/ \__,_|_|  | .__/__/\  \
                        |_|       \__\

--- INFO    : restart from checkpoint
             /home/tata/sim_data/BeamFocusing/02_BeamFocusing_checkpoint/checkpoints/000160
Level 0: dt = 9.111772335e-16 ; dx = 2.734375e-07 ; dy = 5.46875e-07 ; dz = 9.765625e-07

Grids Summary:
  Level 0   16 grids  268435456 cells  100 % of domain
            smallest grid: 128 x 256 x 512  biggest grid: 128 x 256 x 512

-------------------------------------------------------------------------------
--------------------------- MAIN EM PIC PARAMETERS ----------------------------
-------------------------------------------------------------------------------
Precision:            | DOUBLE
Particle precision:   | DOUBLE
Geometry:             | 3D (XYZ)
Operation mode:       | Electromagnetic
                      | - vacuum
-------------------------------------------------------------------------------
Current Deposition:   | Esirkepov
Guard cells           | - ng_alloc_J  = (8,8,8)
 (allocated for J)    |
Particle Pusher:      | Boris
Charge Deposition:    | standard
Field Gathering:      | energy-conserving
Particle Shape Factor:| 3
-------------------------------------------------------------------------------
Maxwell Solver:       | PSATD
                      | - staggered grid
Spectral order:       | - psatd.nox = 16
                      | - psatd.noy = 16
                      | - psatd.noz = 16
Guard cells           | - ng_alloc_EB = (8,8,8)
 (allocated for E/B)  |
-------------------------------------------------------------------------------
Moving window:        |    ON
                      |  - moving_window_dir = z
                      |  - moving_window_v = 299792458
-------------------------------------------------------------------------------
For full input parameters, see the file: warpx_used_inputs

STEP 161 starts ...


**** WARNINGS ******************************************************************
* GLOBAL warning list  after  [ FIRST STEP ]
*
* --> [!  ] [PML] [raised 4 times]
*     Using PSATD together with PML may lead to instabilities if the plasma
*     touches the PML region. It is recommended to leave enough empty space
*     between the plasma boundary and the PML region.
*     @ Raised by: ALL
*
********************************************************************************

STEP 161 ends. TIME = 1.466995346e-13 DT = 9.111772335e-16
Evolve time = 0.45069618 s; This step = 0.45069618 s; Avg. per step = 0.45069618 s

STEP 162 starts ...
STEP 162 ends. TIME = 1.476107118e-13 DT = 9.111772335e-16
Evolve time = 0.736446512 s; This step = 0.285750332 s; Avg. per step = 0.368223256 s

STEP 163 starts ...
```

I have tested out on a number of versions i seem to be facing a similar issue, the size after restart seems odd
25.11 ,25.12 , 26.01 , 26.02, 26.04, 26.05

All seem to show similar behavior, maybe I'm wrong somewhere.

### LINKED ISSUE #7217: Restart error when using a coarsened field diagnostic
## Description
Simulation restart from a checkpoint fails when there is a coarsened field diagnostic (see error message below).

## Expected behavior
The simulation runs successfully on the first go, and saves checkpoint files. I expected the simulation to restart from the saved checkpoint files successfully.

## How to reproduce
I have included a minimal working example (MWE) to reproduce the error. Please run [`input_restart`](https://github.com/user-attachments/files/31563099/input_restart.txt) first, which will save some diagnostics and a checkpoint. Then run [`input_checkpoint`](https://github.com/user-attachments/files/31563101/input_checkpoint.txt) to attempt the restart and reproduce the error.

The error only happens if the simulation runs long enough (restarting from iteration 500 in the MWE does not lead to the error), which I suspect has something to do with the time at which the plasma electrons are introduced.

## System information
Please check all relevant boxes and provide details.

- Operating system (name and version):
  - [x] Linux: e.g., Ubuntu 22.04 LTS
  - [ ] macOS: e.g., macOS Monterey 12.4
  - [ ] Windows: e.g., Windows 11 Pro
- Version of WarpX: 26.04 (26.04-64-g6b32fecc73d3)
- Installation method:
  - [ ] Conda
  - [ ] Spack
  - [ ] PyPI
  - [ ] Brew
  - [x] From source with CMake
  - [x] Module system on an HPC cluster
- Other dependencies: no
- Computational resources:
  - [x] MPI: 1 MPI processes
  - [x] OpenMP: 128 OpenMP threads
  - [x] CPU: 128 CPUs
  - [ ] GPU:

The HPC cluster this was run on (ARCHER2) is not yet supported in the WarpX documentation.

## Steps taken so far

- I have reduced the original input file to a MWE that can reproduce the problem
- I have tried to set the `diag_lo` and `diag_hi` parameters explicitly for the coarsened diagnostic
- I have tried to switch the order in which the checkpoint and coarsened diagnostics are defined in the input file

I have not tried debugging the code.

## Additional information

Error message:
```
0::Assertion `m_crse_ratio[idim]==1' failed, file "/work/e985/e985/moreira/software/src/warpx/Source/Diagnostics/FullDiagnostics.cpp", line 750, Msg: 
### ERROR   : coarsening ratio in reduced dimension must be 1.
 !!!
SIGABRT
MPICH ERROR [Rank 0] [job id ] [Fri Aug 28 17:51:03 2026] [ln01] - Abort(6) (rank 0 in comm 496): application called MPI_Abort(comm=0x84000000, 6) - process 0
```
I have included the log and the Backtrace file in the attachment.

[Backtrace0.0.txt](https://github.com/user-attachments/files/31563102/Backtrace0.0.txt)
[input_checkpoint.txt](https://github.com/user-attachments/files/31563101/input_checkpoint.txt)
[input_restart.txt](https://github.com/user-attachments/files/31563099/input_restart.txt)
[log.txt](https://github.com/user-attachments/files/31563100/log.txt)

ax3l:
After restarts, all diagnostics lo/hi bounds were not properly restored. This was specifically the case if the moving window had a non-zero start step and/or a stop step before the checkpoint step. After a restart, this causes new checkpoints and diagnostics to become corrupted, as the wrong spatial data gets filtered.

This fixes it. Existing checkpoints (from simulations that started from 0) are still readable with this fix.

Fix #6392
Fix #6944
Fix #7217
Close #6400 (included)

## To Do

- [x] debug
- [x] semi-vibe debug (Cursor)
- [x] semi-vibe fix
- [x] review & clean up

## Cleanup Notice

This bug shows how risky it is to duplicate moving-window logic at multiple places. The moving/shift logic should go into functions and they should be re-used. This PR makes this anti-pattern even worse by doubling down.

A follow-up or additional commit should deduplicate the logic to make it safer: #6400

## Known Limitations

As before, this still cannot account for adaptive timesteps.

As a follow-up, one could restore displacement from checkpoint geometry relative to the original simulation geometry, preserving each diagnostic's offset.


ax3l:
@titoiride can you potentially test this PR with your data, too? :) Please restart from a checkpoint that was written from a simulation that ran from the beginning (not a checkpoint created by a restarted simulation).

RemiLehe (commented):
Thanks a for this PR!

In addition to adapting the code to the start/end of the moving window, it seems that this PR is making more changes. Was this intentional?

For instance, the previous code used `warpx.getmoving_window_x()` to infer the current position of the moving window, while the new code relies instead on `current_step` to infer the current position.

In principle,  `warpx.getmoving_window_x()` should be able to work with starting/stopping moving window. To avoid duplicating code (as you pointed out), should we remove the function `warpx.getmoving_window_x()` (since it is not used anymore) or should we try to fix it and introduce it again?

Source/Diagnostics/Diagnostics.cpp
```diff
@@ -529,15 +529,41 @@ Diagnostics::InitBaseData ()
...
-    // current moving_window location
-    if (WarpX::do_moving_window) {
+    // current moving_window location. Account for start_moving_window_step and
+    // end_moving_window_step to only shift based on steps where the window was active.
+    if (WarpX::do_moving_window && warpx.getistep(0) > 0) {
         const int moving_dir = WarpX::moving_window_dir;
-        const amrex::Real displacement =
-            warpx.getmoving_window_x() - warpx.Geom(0).ProbLo(moving_dir);
-        const int shift_num_base = static_cast<int>
-            (displacement / warpx.Geom(0).CellSize(moving_dir));
```
ax3l, line 538: The issue here is subtle.

In `FullDiagnostics::MovingWindowAndGalileanDomainShift` this shift is only done for steps where `if (WarpX::moving_window_active(step+1))` an accumulates the `m_lo`/`m_hi` on truncated integer (cell) locations.

The implementation here omitted these details and thus introduces a drift to the real dimensions on restart.

EZoni: Just for context, some of this had been changed recently in https://github.com/BLAST-WarpX/warpx/pull/5985. More information on what bug this was fixing in the PR description.

ax3l: attn @bnara : can you double-check this PR does not introduce any issues with your existing moving-window workflows? This should improve the logic when moving windows start later than step 0 or stop at a certain step.

ax3l:
@RemiLehe 

> In principle, warpx.getmoving_window_x() should be able to work with starting/stopping moving window. To avoid duplicating code (as you pointed out), should we remove the function warpx.getmoving_window_x() (since it is not used anymore) or should we try to fix it and introduce it again?

I was a bit puzzled as well why I could not find a solution with `warpx.getmoving_window_x`. Looking at the moving window implementation in WarpX as of today, I think it has a general flaw that the diags prob domain can shift over time compared to the simulation geometry (stored among others in `getmoving_window_x`) because of the step-wise (accumulative) integer rounding to align with cells:
https://github.com/BLAST-WarpX/warpx/blob/25.11/Source/Diagnostics/FullDiagnostics.cpp#L1007-L1008

I think this can be fixed by fully removing the double book-keeping in `FullDiagnostics::MovingWindowAndGalileanDomainShift` but that will render existing restart points unusable.

In other words, `WarpX::MoveWindow` should be the only source of truth, but the `FullDiags` does its own tracking and actually is not doing it well: compared to
https://github.com/BLAST-WarpX/warpx/blob/25.11/Source/Utils/WarpXMovingWindow.cpp#L356-L376
it cumulative looses rounds on the order of a cells size, every time it updates.

While the over all moving window does the same:
https://github.com/BLAST-WarpX/warpx/blob/25.11/Source/Utils/WarpXMovingWindow.cpp#L392
I think this causes an issue in the situation where the moving window is not active 100% of the sim steps.

atmyers:
Note - the original fix was correct for default settings but had an off-by-one error in the case that moving window didn't start moving at the beginning of simulation. This has been corrected. This PR now adds several new tests that cover corner cases, and includes the de-duplication logic in PR #6400.

This is now ready for review.

ax3l:
Thank you @atmyers , this looks great to me!!

Source/Diagnostics/Diagnostics.cpp
ax3l: @bnara I took a look at the changes in #5985 that already landed and I think the latest version of this PR here preserves your earlier fix and more. Otherwise, please let @atmyers and me know if you see any regressions from your fix in there.

Examples/Tests/restart/inputs_base_3d_mw
EZoni, line 1: The three new `inputs_base_3d_mw_*` files seem to differ only in a few moving-window lines. Maybe we could try using yet another `base` input file for these (either an existing one, if available, or a new one) and put the moving-window overrides in the `inputs_test_*` files. This could remove several lines of copy-paste that may drift.

ax3l: Great point, consolidated.

Examples/Tests/restart/inputs_test_3d_acceleration_mw_start_stop
```diff
@@ -0,0 +1,10 @@
+FILE = inputs_base_3d_mw
+
+# The window is only active for steps 2-3 (start=2, end=4 is exclusive),
+# i.e. it has already stopped by the time the checkpoint at step 5 is
+# written. This is the case that exposed the restart bug: with the window
```
ax3l, line 5:
```suggestion
# written. This is the case that exposed a restart bug: with the window
```

atmyers (approved):

PR #6400, opened 2025-11-16
[WIP] Deduplicate Moving Window Logic

ax3l:
The math of a variable-window and variable-start/stop moving window is duplicated partly over the code base. This caused bugs. Refactoring to avoid bugs in the future.

- [x] vibe refactor
- [ ] rebase after #6399
- [ ] clean up

### DIFF
--- a/Source/Diagnostics/Diagnostics.cpp
+++ b/Source/Diagnostics/Diagnostics.cpp
@@ -529,15 +529,41 @@ Diagnostics::InitBaseData ()
     m_all_field_functors.resize( nmax_lev );
 
     // For restart, move the m_lo and m_hi of the diag consistent with the
-    // current moving_window location
-    if (WarpX::do_moving_window) {
+    // current moving_window location. Account for start_moving_window_step and
+    // end_moving_window_step to only shift based on steps where the window was active.
+    if (WarpX::do_moving_window && warpx.getistep(0) > 0) {
         const int moving_dir = WarpX::moving_window_dir;
-        const amrex::Real displacement =
-            warpx.getmoving_window_x() - warpx.Geom(0).ProbLo(moving_dir);
-        const int shift_num_base = static_cast<int>
-            (displacement / warpx.Geom(0).CellSize(moving_dir));
-        m_lo[moving_dir] += shift_num_base * warpx.Geom(0).CellSize(moving_dir);
-        m_hi[moving_dir] += shift_num_base * warpx.Geom(0).CellSize(moving_dir);
+        const int current_step = warpx.getistep(0);
+
+        // Calculate how many steps the moving window has been active up to current_step
+        // This accounts for start_moving_window_step and end_moving_window_step
+        int active_steps = 0;
+        const int start_step = WarpX::start_moving_window_step;
+        const int end_step = (WarpX::end_moving_window_step < 0) ?
+            current_step : std::min(WarpX::end_moving_window_step, current_step);
+
+        if (end_step >= start_step && current_step >= start_step) {
+            // Count steps from start_step to min(end_step, current_step - 1)
+            // Note: current_step is the checkpoint step, but the replay loop goes from 0 to current_step - 1
+            // So we need to count steps up to current_step - 1, not current_step
+            const int last_replayed_step = current_step - 1;
+            const int effective_end_step = std::min(end_step, last_replayed_step);
+            if (effective_end_step >= start_step) {
+                active_steps = effective_end_step - start_step + 1;
+            }
+        }
+
+        if (active_steps > 0) {
+            // Calculate the shift based on the number of active steps
+            const amrex::Real shift_per_step = (WarpX::moving_window_v - WarpX::beta_boost * PhysConst::c)
+                                             / (1._rt - WarpX::moving_window_v * WarpX::beta_boost / PhysConst::c)
+                                             * warpx.getdt(0);
+            const amrex::Real total_shift = active_steps * shift_per_step;
+            const int shift_num_base = static_cast<int>
+                (total_shift / warpx.Geom(0).CellSize(moving_dir));
+            m_lo[moving_dir] += shift_num_base * warpx.Geom(0).CellSize(moving_dir);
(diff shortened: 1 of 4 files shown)


Source/Diagnostics/Diagnostics.cpp
```diff
@@ -529,15 +529,24 @@ Diagnostics::InitBaseData ()
...
+
+        // Note: current_step is the checkpoint step, but the replay loop goes from 0 to current_step - 1
+        // So we need to calculate the shift for steps 0 to current_step - 1
+        const int last_replayed_step = current_step - 1;
+        const amrex::Real total_shift = WarpX::CalculateMovingWindowShift(0, last_replayed_step, warpx.getdt(0));
+
+        if (total_shift > 0.0_rt) {
+            // Use the unified helper function to calculate physical shift aligned to cell boundaries
+            const amrex::Real physical_shift = WarpX::CalculateMovingWindowAlignedShift(
+                total_shift, warpx.Geom(0).CellSize(moving_dir));
```
ax3l, line 546: While we refactor this logic... Does restart (in development) even work with `m_galilean_shift` being active?

I would like to share even more logic that resides in `FullDiagnostics::MovingWindowAndGalileanDomainShift` here...

Source/Diagnostics/Diagnostics.cpp
```diff
@@ -529,15 +529,24 @@ Diagnostics::InitBaseData ()
     m_all_field_functors.resize( nmax_lev );
 
     // For restart, move the m_lo and m_hi of the diag consistent with the
-    // current moving_window location
-    if (WarpX::do_moving_window) {
+    // current moving_window location. Account for start_moving_window_step and
+    // end_moving_window_step to only shift based on steps where the window was active.
+    if (WarpX::do_moving_window && warpx.getistep(0) > 0) {
```
RemiLehe, line 534: Why was `warpx.getistep(0) > 0` added here? 
Do we expect negative step number in WarpX?

Source/Diagnostics/Diagnostics.cpp
```diff
@@ -529,15 +529,24 @@ Diagnostics::InitBaseData ()
...
         const int moving_dir = WarpX::moving_window_dir;
-        const amrex::Real displacement =
-            warpx.getmoving_window_x() - warpx.Geom(0).ProbLo(moving_dir);
-        const int shift_num_base = static_cast<int>
-            (displacement / warpx.Geom(0).CellSize(moving_dir));
-        m_lo[moving_dir] += shift_num_base * warpx.Geom(0).CellSize(moving_dir);
-        m_hi[moving_dir] += shift_num_base * warpx.Geom(0).CellSize(moving_dir);
+        const int current_step = warpx.getistep(0);
+
+        // Note: current_step is the checkpoint step, but the replay loop goes from 0 to current_step - 1
```
RemiLehe, line 538: What does "replay loop" mean here? Do you mean "restart"?

Source/Diagnostics/Diagnostics.cpp
```diff
@@ -529,15 +529,24 @@ Diagnostics::InitBaseData ()
...
+        // So we need to calculate the shift for steps 0 to current_step - 1
+        const int last_replayed_step = current_step - 1;
+        const amrex::Real total_shift = WarpX::CalculateMovingWindowShift(0, last_replayed_step, warpx.getdt(0));
+
+        if (total_shift > 0.0_rt) {
+            // Use the unified helper function to calculate physical shift aligned to cell boundaries
```
RemiLehe, line 544:
```suggestion
            // Calculate physical shift aligned to cell boundaries
```

Source/Diagnostics/FullDiagnostics.cpp
```diff
@@ -1004,16 +1004,17 @@ FullDiagnostics::MovingWindowAndGalileanDomainShift (int step)
         const amrex::Real* cur_lo = m_geom_output[0][0].ProbLo();
         const amrex::Real* cur_hi = m_geom_output[0][0].ProbHi();
         const amrex::Real* geom_dx = m_geom_output[0][0].CellSize();
-        const auto num_shift_base = static_cast<int>((moving_window_x - cur_lo[moving_dir])
-                                              / geom_dx[moving_dir]);
+        // Use the unified helper function to calculate physical shift aligned to cell boundaries
```
RemiLehe, line 1007:
```suggestion
        // Calculate physical shift aligned to cell boundaries
```

Source/Utils/WarpXMovingWindow.cpp
```diff
@@ -353,6 +353,55 @@ namespace
 
 }
 
+amrex::Real
+WarpX::CalculateMovingWindowShiftPerStep (amrex::Real dt)
+{
+    // Calculate the shift per step (same formula as used in MoveWindow)
```
RemiLehe, line 359:
```suggestion
    // Calculate the shift per step
```
The formula does not exist anymore in `MoveWindow`, so this comment is confusing.
(Instead we now call `CalculateMovingWindowShiftPerStep` in `MoveWindow` ; but I don't think comments should point out where the function is called - or if they should, then we also need to say that this function is called in `MovingWindowAndGalileanDomainShift`

Source/WarpX.H
```diff
@@ -756,6 +756,41 @@ public:
         bool const step_after_start = (step >= start_moving_window_step);
         return do_moving_window && step_before_end && step_after_start;
     }
+    /** Calculate the shift per step for the moving window
+     *
+     * This function calculates how much the moving window shifts in one timestep.
+     * This is the same calculation used in MoveWindow.
```
RemiLehe, line 762: This function **is** used in `MoveWindow`, so it is confusing to say that this is the "same" calculation used in MoveWindow.

Source/Diagnostics/Diagnostics.cpp
ax3l: Added for only catching restarts, this PR is still wip as an idea to refactor. Forgot the title

EZoni:
As pointed out by @atmyers in https://github.com/BLAST-WarpX/warpx/pull/6399#issuecomment-5547047657, the changes in this PR are now included in #6399, so this PR should be closed once #6399 is merged.

PR #6404, opened 2025-11-17
Add crabwaist

pkicsiny:
Added crab waist transformation at the IP.

This is a transformation of the y and px particle coordinates. The formulas derive from the sextupole Hamiltonian of the "crab-sextupole" that is used to implement the crab waist scheme at the FCC-ee. The transformation is applied in the `AddParticles.cpp` file in the lab frame, before the beam rotation.
Added documentation of the user input flags.
Added unittest input and analysis script.

Unit test and analysis script added.


Examples/Tests/gaussian_beam/analysis_crabwaist_no_focusing.py
```diff
@@ -0,0 +1,286 @@
...
+#
+# This file is part of WarpX.
+#
+# License: BSD-3-Clause-LBNL
+
+import numpy as np
+import matplotlib.pyplot as plt
+from openpmd_viewer import OpenPMDTimeSeries
+import os
+from scipy.constants import alpha, c, eV, m_e, pi
```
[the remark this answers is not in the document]
pkicsiny: resolved

pkicsiny:
fixed unittests. there are 2 runs and one analysis script that compares output of the 2 warpx runs.

pkicsiny:
fixed ifdef guard that broke some tests

Source/Initialization/PlasmaInjector.cpp
```diff
@@ -255,11 +255,27 @@ void PlasmaInjector::setupGaussianBeam (amrex::ParmParse const& pp_species)
     const bool focusing_is_specified = pp_species.contains("focal_distance");
     utils::parser::queryWithParser(pp_species, source_name, "do_gaussian_beam_rotation", do_rotation);
     utils::parser::queryWithParser(pp_species, source_name, "do_gaussian_beam_rotation_momenta", do_rotation_momenta);
+    utils::parser::queryWithParser(pp_species, source_name, "do_crabwaist", do_crabwaist);
```
aeriforme:
```suggestion
    utils::parser::queryWithParser(pp_species, source_name, "do_gaussian_beam_crabwaist", do_crabwaist);
```

Examples/Tests/gaussian_beam/inputs_test_3d_crabwaist_on_no_focusing
aeriforme, line 1: This input file is almost a copy of `inputs_test_3d_crabwaist_off_no_focusing`, correct?
To avoid too much duplication, could you import  one input file (`off`) into the other (`on`)? 
The `on` input file would become something like this: 
```
FILE=inputs_test_3d_crabwaist_off_no_focusing
beam1.do_crabwaist = 1
beam1.crabwaist_strength = 0.4
beam2.do_crabwaist = 1
beam2.crabwaist_strength = 0.4
```

See [here for reference](https://github.com/BLAST-WarpX/warpx/blob/development/Examples/Tests/electrostatic_sphere/inputs_test_3d_electrostatic_sphere_lab_frame) how we do it in another example.

Examples/Tests/gaussian_beam/analysis_crabwaist_no_focusing.py
```diff
@@ -0,0 +1,476 @@
...
+assert np.all(x_pos_pos_idx == (x_pos_cw_1 > 0))
+assert np.all(uy_ele_cw_0 == uy_ele_cw_1)
+assert np.all(uy_pos_cw_0 == uy_pos_cw_1)
+assert np.all(x_ele_cw_0 == x_ele_cw_1)
+assert np.all(x_pos_cw_0 == x_pos_cw_1)
+assert np.all(z_ele_cw_0 == z_ele_cw_1)
+assert np.all(z_pos_cw_0 == z_pos_cw_1)
+
+####################
+# plot phase space #
```
aeriforme: Could you: 
- save the plots to file
- produce the plots conditionally, by introducing a flag like `do_plots` and default it to `False`
?

That way we don't make the CI test do useless computations.

Examples/Tests/gaussian_beam/analysis_crabwaist_no_focusing.py
```diff
@@ -0,0 +1,476 @@
...
+assert np.all(z_pos_cw_1_init == z_pos_cw_0_init)
+
+assert np.allclose(
+    uz_ele_cw_1_init / gamma, uz_ele_cw_0_init / gamma, rtol=1e-14, atol=1e-14
+)
+assert np.allclose(
+    uz_pos_cw_1_init / gamma, uz_pos_cw_0_init / gamma, rtol=1e-14, atol=1e-14
+)
+
+# cw was applied on these
```
aeriforme:
```suggestion
# check the variables that are actually modified by crab waist
```

Examples/Tests/gaussian_beam/analysis_crabwaist_no_focusing.py
```diff
@@ -0,0 +1,476 @@
...
+ax[0].legend()
+ax[1].legend()
+
+fig.suptitle(r"CW on backtransformed ($CW^{-1}R^{-1}x$)")
+
+#########
+# tests #
+#########
+
+# these are trivial
```
aeriforme:
```suggestion
# sanity check: these variables should not change upon crab waist
```

Source/Initialization/PlasmaInjector.cpp
```diff
@@ -255,11 +255,27 @@ void PlasmaInjector::setupGaussianBeam (amrex::ParmParse const& pp_species)
...
         utils::parser::queryWithParser(pp_species, source_name, "gaussian_beam_rotation_angle", rotation_angle);
         utils::parser::getArrWithParser(pp_species, source_name, "gaussian_beam_rotation_axis", rotation_axis, 0, 3);
+
+        // crabwaist only if beams collide with a nonzero crossing angle in the x-z plane
+        if(do_crabwaist){
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE(rotation_angle!=0,
+                "Error: Invalid value for rotation_angle. It must not be 0!");
+            WARPX_ALWAYS_ASSERT_WITH_MESSAGE((rotation_axis[0]==0 && rotation_axis[1]==1 && rotation_axis[2]==0),
+                "Error: Invalid value for rotation_axis. It must be [0,1,0]!");
+            ablastr::warn_manager::WMRecordWarning("Species", "Currently crab waist is supported in y dimension only.\n");
```
aeriforme: Is this necessary? The assert above would handle it already, no?

Source/Particles/ParticleCreation/AddParticles.cpp
```diff
@@ -469,7 +472,15 @@ PhysicalParticleContainer::AddGaussianBeam (PlasmaInjector const& plasma_injecto
                 x = x - (v_x - v_dot_n*n_x) * t;
 #endif
             }
-#if defined(WARPX_DIM_3D) || defined(WARPX_DIM_XZ)
+#if defined(WARPX_DIM_3D)
+        // crabwaist before rotation
+            // Lorentz boost currently only possible along z direction: x and z coords swapped if boost on, not if I only rotate
```
aeriforme: Do we need this comment here? 

What happens if boost is on then?

Docs/source/usage/parameters.rst
```diff
@@ -1019,6 +1019,20 @@ Particle initialization
...
+      .. math::
+
+          \alpha &= -\frac{k}{\tan(2\varphi)},
+
+          p_{x,CW} &= p_x - \frac{1}{2}\alpha p_y^2,
+
+          y_{CW} &= y + \alpha x p_y.
+
+    It is applied on the coordinates in the lab frame before the rotation. Currently supports vertical crab waist only. See (:cite:t:`param-Zobov2010`) for more details.
+
```
aeriforme, line 1532: Could you: 
- clarify that this is only possible in `y`
- clarify what happens if boost is on
?

aeriforme (changes requested):

Examples/Tests/gaussian_beam/analysis_crabwaist_no_focusing.py
pkicsiny: added `do_plots` flag

Source/Initialization/PlasmaInjector.cpp
pkicsiny: removed warning and expanded the assert error message

Source/Initialization/PlasmaInjector.cpp
pkicsiny: changed all occurrences of the input flag from `do_crabwaist ` to `do_gaussian_beam_crabwaist `

Examples/Tests/gaussian_beam/inputs_test_3d_crabwaist_on_no_focusing
pkicsiny: simplified the crabwaist on input.

Docs/source/usage/parameters.rst
pkicsiny: - the way the crab waist is designed is to change px and y because y is the flat direction. One could also design a crab waist transformation that affects x and py but real world colliders are typically flat in y.
- the crab waist is applied before the boost and the inverse crab waist is applied after the inverse boost. The crab waist is technically part of the arc lattice not the beambeam. The effect of the crab waist at the IP is mimicked by this operation. The crab waist only makes sense if we have a crossing angle and not without it so using it together with the boost is the default.

Source/Particles/ParticleCreation/AddParticles.cpp
pkicsiny: If the boost is on we would have to then swap x and z so wherever there is x in the crab waist code we have to put z. But the boost happens after the crab waist and the inverse boost happens before the inverse crab waist so we dont need to modify anything.

PR #6408, opened 2025-11-19
Update convergence criterion for MLMG

RemiLehe:
In https://github.com/BLAST-WarpX/warpx/pull/2410, the convergence criterion of the MLMG solver was modified so as to use the norm of `b` (whenever `|b|` is not zero), to evaluate the relative error in the Poisson equation (written as `Ax = b`) and decide when to stop the iterations.

However, there are cases where this it not suitable. A typical case is when the user fixes a large value of the potential at the boundaries (or on the embedded boundary), and also has a very low-charge beam in the simulation. In this case, the residual is dominated by the potential created by the boundaries, but the norm of `b` (which depends on the charge density of the beam) is very low, resulting in a large `resid/bnorm`. In some cases, this large `resid/bnorm` can even trigger the error `MLMG failing so lets stop here` because of this check in AMREX:
https://github.com/AMReX-Codes/amrex/blob/development/Src/LinearSolvers/MLMG/AMReX_MLMG.H#L539

Instead, with this PR, we always use the maximum of the norm of `b` and of the initial residual `resid0` to evaluate the relative error in the Poisson equation - instead of using directly the maximum of the norm of `b`.

I also checked that this fixes the issue mentioned here: https://github.com/BLAST-WarpX/warpx/issues/6161#issuecomment-3424497174

The changes in the checksum is because of the change in convergence criterion.


Source/ablastr/fields/PoissonSolver.H
```diff
@@ -274,6 +270,9 @@ computePhi (
                               geom[lev].CellSize(2)/std::sqrt(1._rt-beta_solver[2]*beta_solver[2]))};
 
 #if (defined(ABLASTR_USE_FFT) && defined(WARPX_DIM_3D))
+        // determine if rho is zero everywhere
+        const amrex::Real max_norm_b = getMaxNormRho(
+            amrex::GetVecOfConstPtrs(rho), finest_level, absolute_tolerance);
```
RemiLehe, line 275: I move the calculation of `max_norm_b` here, because it is still used by the FFT solver.

WeiqunZhang (approved):

roelof-groenewald (approved):

PR #6270, opened 2025-10-15
Tuolumne (LLNL): CPU-Only, HDF5, PETSC, Signal Handling

ax3l:
Add CPU-only instructions for Tuolumne at LLNL.
This is mostly for development, because this is not using the GPU-part of the APU.

For both GPU and CPU builds: Add support for a self-built HDF5 (system module went missing), self-built PETSC, and document signal handling.

[Preview.](https://warpx--6270.org.readthedocs.build/en/6270/install/hpc/tuolumne.html)

- [x] draft
- [x] install: tested
- [x] runtime: tested
- [x] fix many-thread issue with TLS
- [x] add HDF5
- [x] add PETSC
- [x] document signal usage `flux run --signal=SIGUSR1@120s --exclusive ...` (only works if in `flux run ...` command, not the script comments)

PR #6410, opened 2025-11-20
Fix path to petsc directory in tuolumne install scripts

eebasso:
Use the absolute path to 'petsc' directory in tuolumne install scripts. The relative path to 'petsc' fails when the current working directory is not SRC_DIR.

Follow-up to #6270


ax3l (approved):
Oh definitely, thanks for the fix!

PR #6409, opened 2025-11-20
[WIP] Fix DefaultInitializeRuntimeAttributes

EZoni:
### Overview

Coming soon...

### To do

- [x] Particle component names
- [ ] Particle user attributes?

PR #6413, opened 2025-11-21
Forward SoA Names to Temporary Tiles

ax3l:
Ensure temporary particle tiles have access to SoA names, by passing in the names from their referenced PC.

Related to #6409


ax3l:
@EZoni can you rebase #6275 against this and see if it fixes your issues?

If not, let us print the number of runtime real components in the photon species before your crash and the vector of real names again.

EZoni:
> @EZoni can you rebase #6275 against this and see if it fixes your issues? If not, let us print the number of runtime real components in the photon species before your crash and the vector of real names again.

It doesn't seem to fix the issue with the test `test_1d_photoneutralization_dsmc` in #6275, but we should probably merge this PR anyways if it makes things more robust. I will report more debugging info on the test directly in #6275.

ax3l:
Agreed. Feel free to merge.

atmyers (approved):

PR #6414, opened 2025-11-21
Fix: NERSC PM Interactive

ax3l:
Fix the `getNode` and `runNode` alias in our profile to support multiple GPUs well.

Fix #6382

PR #6416, opened 2025-11-21
Remove project number left over in Perlmutter profile

EZoni:
Follow up on #6414 and remove a project number that was left over in the Perlmutter profile. Noticed by @titoiride in https://github.com/BLAST-WarpX/warpx/pull/6414/files#r2551120689.


titoiride (approved):

ax3l (approved):
Thx!

PR #6418, opened 2025-11-24
Fix the periodic 1D Poisson solver

dpgrote:
There were problems with the implementation of the tridiagonal solver for phi with periodic boundaries. This PR replaces it with a more standard solver as given in the Wikipedia page https://en.wikipedia.org/wiki/Tridiagonal_matrix_algorithm#Variants.

### DIFF
--- a/Examples/Tests/energy_conserving_thermal_plasma/CMakeLists.txt
+++ b/Examples/Tests/energy_conserving_thermal_plasma/CMakeLists.txt
@@ -10,3 +10,13 @@ add_warpx_test(
     "analysis_default_regression.py --path diags/diag1000500"  # checksum
     OFF  # dependency
 )
+
+add_warpx_test(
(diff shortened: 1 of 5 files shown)


EZoni:
I have no objections to the fix per se, but isn't this (a tridiagonal matrix algorithm) something basic enough to belong better to a numerical library like AMReX rather than a simulation code like WarpX?

dpgrote:
> I have no objections to the fix per se, but isn't this (a tridiagonal matrix algorithm) something basic enough to belong better to a numerical library like AMReX rather than a simulation code like WarpX?

I would agree. @WeiqunZhang 
Maybe merge this now with this note to move it to AMReX later.

roelof-groenewald (approved):
Thanks for the update @dpgrote. I agree we should merge this in order to have the development version of WarpX give correct results, but it would be good to move the solver to AMReX eventually.

PR #6423, opened 2025-11-26
Documentation for Adastra (CINES, France): change installation directory to WORKDIR & clean scripts

lucafedeli88:
This PR concerns the documentation for the Adastra supercomputer.

The main point of this PR is to change the installation directory from `$SHAREDHOMEDIR` to `$WORKDIR` .
On Adastra the quota for the `$HOME` directory is shared among all the group members of a given project and it is quite stringent, so having individual installations in the `$HOME` folder may be challenging. Therefore, the proposed solution was to have a unique installation in  `$SHAREDHOMEDIR` for all the project members. However, this introduces a significant constraint for the users. For this reason, this PR changes the default installation directory to  `$WORKDIR`. 

In addition,  the installation of additional modules in the python environment is proposed (notably `jupyter` and `lasy`), and some cleaning of the installation scripts is performed.


ax3l (approved):
Thank you, this is a great update! :)

PR #6424, opened 2025-11-26
Fix PEC_insulators

dpgrote:
The vectors holding the parsers were not being setup correctly, leading to out of bounds access. The fix is to add empty parsers for the fields that are normal to the boundaries so that the vectors have the correct length (the number of dimensions).

Also, comments were cleaned up for clarity.


JustinRayAngus (approved):
Looks good. Thanks Dave.

PR #6385, opened 2025-11-13
Docs: inputs section cleanup - QED

aeriforme:
QED input parameters are moved to a dedicated subsection. Split from #6355.

PR #6427, opened 2025-11-27
Integrate picsar_qed in ablastr : move in the Schwinger process

lucafedeli88:
### Summary
This PR is the second step of a long-overdue process: the integration of picsar_qed into ablastr. Specifically, this PR moves the Schwinger pair production process into `ablastr/sfqed` and makes the necessary changes in the rest of the code . In addition, some details on the Schwinger process are added to the documentation.

The code is adapted from PICSAR QED, with minor changes.

### Main changes:

- Calculations for the Schwinger process are performed using `amrex::Real` and `amrex::RealVectND<3>` types.

- `Source/ablastr/sfqed/SchwingerProcess.H` : this new source file contains the methods that compute the pair production rate

- `Source/ablastr/sfqed/UnitConversion.H` : QED-related calculations are performed internally using the Heaviside-Lorentz units. This new file contains helper functions to convert quantities between these units and SI units.

- `Source/Particles/ElementaryProcess/QEDInternals/SchwingerProcessWrapper.H` : this file is no-longer needed. The only one method that was implemented there is now  in  `ablastr/sfqed/SchwingerProcess.H` .

- `Source/Particles/ElementaryProcess/QEDSchwingerProcess.H` : in this file a simple name change to call the new method in `ablastr` instead of the one in the wrapper is performed.

- `Source/Particles/MultiParticleContainer.H/cpp` : the Schwinger process no longer requires the PICSAR QED library. Therefore, the relevant code is moved out of the `#ifdef WARPX_QED ... #endif` sections.

- `Source/Evolve\WarpXEvolve.cpp` : same as above

- `Docs/source/usage/parameters.rst` : the documentation is updated to reflect that the Schwinger process no longer requires the PICSAR QED library.

- `Docs/source/theory/multiphysics/qed.rst` : some details on the physics of the Schwinger process are added to the documentation.

- `Docs/source/refs.bib` a reference to a paper cited in `Docs/source/theory/multiphysics/qed.rst` is added.

- `Regression/Checksum/benchmarks_json/test_3d_qed_schwinger_4.json` : I had to reset the checksum, but the change is really tiny (and the python script passes) .


Source/Evolve/WarpXEvolve.cpp
```diff
@@ -218,11 +218,12 @@ WarpX::Evolve (int numsteps)
         // multi-physics: field ionization
         doFieldIonization();
 
-#ifdef WARPX_QED
         // multi-physics: QED effects
+#ifdef WARPX_QED
         doQEDEvents();
-        mypc->doQEDSchwinger();
 #endif
+        mypc->doQEDSchwinger();
```
aeriforme, line 224: Why this? 
I think it makes sense that the Schwinger effect can only be used when compiling with QED ON, no?
Even if it's not technically essential...

Source/Particles/MultiParticleContainer.H
aeriforme, line 1: Same as above, I don't see why we would prefer this.

Source/Particles/MultiParticleContainer.cpp
```diff
@@ -1903,4 +1731,174 @@ void MultiParticleContainer::CheckQEDProductSpecies()
 
 }
 
+
+void
+MultiParticleContainer::doQEDSchwinger ()
```
aeriforme, line 1702: Why doesn't this code belong to ablastr?

Source/ablastr/CMakeLists.txt
```diff
@@ -4,5 +4,6 @@ add_subdirectory(fields)
 add_subdirectory(parallelization)
 #add_subdirectory(particles)
 #add_subdirectory(profiler)
+add_subdirectory(qed)
```
aeriforme: Can we maybe call it `sfqed`?

Source/Particles/MultiParticleContainer.cpp
aeriforme: Also, shall we add doxygen?

aeriforme (commented):
Merging [PR6385](https://github.com/BLAST-WarpX/warpx/pull/6385) first could be helpful

Source/ablastr/sfqed/SchwingerProcess.H
```diff
@@ -0,0 +1,176 @@
...
+            constexpr const auto coeff = static_cast<amrex::Real>(
+                amrex::Math::powi<2>(
+                    (heaviside_lorentz_elementary_charge<double>*heaviside_lorentz_schwinger_field<double>)/
+                    (2.0*ablastr::constant::math::pi_v<double>)));
+
+            constexpr auto rate_conv =
+                convert_from_to<unit_system::heaviside_lorentz, UnitSystem, quantity::rate>(1.0_rt);
+
+            constexpr auto pi = ablastr::constant::math::pi;
+            if(epsi != 0.0_rt && eta != 0.0_rt){
```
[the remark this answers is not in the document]
lucafedeli88: I think that this is OK, we want to avoid specifically a division by zero by replacing the function with its limit when epsilon or eta tend to zero.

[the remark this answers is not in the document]
lucafedeli88: See comment above.

Source/ablastr/sfqed/SchwingerProcess.H
```diff
@@ -0,0 +1,176 @@
...
+                return coeff*rate_conv*epsi*eta*std::exp(-pi/epsi)/std::tanh(pi*eta/epsi);
+            }
+            else if(epsi == 0.0_rt){
```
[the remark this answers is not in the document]
lucafedeli88: See comment above.

Source/Evolve/WarpXEvolve.cpp
lucafedeli88: We have to discuss this. 
In principle, if we integrate picsar QED into ablastr, we could enable QED effects at no cost (no additional libraries). We could have a compilation flag only for lookup tables generation (this feature requires the boost library).
Since I see this PR as a step towards the full integration of picsar QED in ablastr it seemed reasonable to me  to start putting stuff that can be used without picsar out of the `#ifdef` regions. However, this change can be certainly delayed till the end of the integration.

Source/ablastr/CMakeLists.txt
lucafedeli88: yes, that's a good idea. I'll do that.

Source/Particles/MultiParticleContainer.cpp
lucafedeli88: I think that it is a bit too WarpX - specific to be in ablastr. Would you suggest moving the whole function to ablastr?  Adding a doxygen comment is certainly a good idea.

Source/Evolve/WarpXEvolve.cpp
aeriforme: Ahh I see! We shall discuss at the developers meeting then.
I'll add it to the agenda.

Source/ablastr/CMakeLists.txt
lucafedeli88: done

Source/Particles/MultiParticleContainer.cpp
lucafedeli88: Actually,  I've noticed that the doxygen statement is in the `MultiParticleContainer.H` header file:

```
    /** If Schwinger process is activated, this function is called at every
     * timestep in Evolve and is used to create Schwinger electron-positron pairs.
     * Within this function we loop over all cells to calculate the number of
     * created physical pairs. If this number is higher than 0, we create a single
     * particle per species in this cell, with a weight corresponding to the number of physical
     * particles.
     */
    void doQEDSchwinger ();
```

Looking again at the code, I still think that it is a bit too warpx-specific to be in ablastr...

Source/ablastr/sfqed/SchwingerProcess.H
```diff
@@ -0,0 +1,176 @@
...
+            constexpr const auto coeff = static_cast<amrex::Real>(
+                amrex::Math::powi<2>(
+                    (heaviside_lorentz_elementary_charge<double>*heaviside_lorentz_schwinger_field<double>)/
+                    (2.0*ablastr::constant::math::pi_v<double>)));
+
+            constexpr auto rate_conv =
+                convert_from_to<unit_system::heaviside_lorentz, UnitSystem, quantity::rate>(1.0_rt);
+
+            constexpr auto pi = ablastr::constant::math::pi;
+            if(epsi != 0.0_rt && eta != 0.0_rt){
```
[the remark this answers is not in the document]
lucafedeli88: I think it's OK here. It protects against a corner case replacing an expression that would give a NaN with a valid expression obtained by computing the limits `eta -> 0` or `epsi -> 0`.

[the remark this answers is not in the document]
lucafedeli88: same as above

Source/ablastr/sfqed/SchwingerProcess.H
```diff
@@ -0,0 +1,176 @@
...
+                return coeff*rate_conv*epsi*eta*std::exp(-pi/epsi)/std::tanh(pi*eta/epsi);
+            }
+            else if(epsi == 0.0_rt){
```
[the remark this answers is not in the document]
lucafedeli88: same as above

Source/ablastr/sfqed/SchwingerProcess.H
```diff
@@ -0,0 +1,176 @@
...
+            constexpr auto one_over_schwinger = static_cast<amrex::Real>(
+                1.0/heaviside_lorentz_schwinger_field<double>);
+
+            const auto em_e_2 = em_e.dotProduct(em_e);
+            const auto em_b_2 = em_b.dotProduct(em_b);
+
+            const auto ff = (em_e_2 - em_b_2)*0.5_rt;
+            const auto gg = em_e.dotProduct(em_b);
+
+            const auto inner = std::sqrt(ff*ff+ gg*gg);
```
lucafedeli88:
```suggestion
            const auto inner = std::sqrt(ff*ff + gg*gg);
```

lucafedeli88:
Sorry for the long wait time, @aeriforme ! I had time to finalize this PR only now. It should be ready to be re-reviewed. 
As far as getting rid of the QED compilation flag is concerned (once the migration of the QED library into ablastr will be completed), it should not have any substantial performance impact. For the pushpx kernel we already compile the two specialized versions with and without QED effects.

ax3l:
@lucafedeli88 #6385 was merged and might have a created a small conflict. Do you like to update the PR? :)

PR #6428, opened 2025-11-30
allow dt update based on particle CFL for theta implicit solver.

JustinRayAngus:
This PR adds the ability to dynamically limit the time step based on the particle CFL when using the theta implicit EM solver, which treats light waves implicitly. This would be a one line PR if it wasn't for one subtlety described below.

The call to `UpdateDtFromParticleSpeeds()` in the` WarpX::Evolve()` routine occurs prior to the call to `OneStep()`. For many of the time-advance methods, including the implicit ones, `OneStep()` first performs collisions and then the PIC advance. However, to be meaningful, the call to `UpdateDtFromParticleSpeeds()` should occur **just before** the PIC advance. This PR swaps the order of calls inside `OneStep()` for the implicit methods to do the PIC advance before collisions. This ordering is actually more common in my experience, and I wonder if the ordering should be swapped for other time-advance methods as well. @roelof-groenewald @ax3l @RemiLehe


Source/Evolve/WarpXEvolve.cpp
```diff
@@ -192,7 +192,8 @@ WarpX::Evolve (int numsteps)
 
         // Update timestep for electrostatic solver if a constant dt is not provided
         // This first synchronizes the position and velocity before setting the new timestep
-        if (electromagnetic_solver_id == ElectromagneticSolverAlgo::None &&
+        if ((electromagnetic_solver_id == ElectromagneticSolverAlgo::None ||
+            evolve_scheme == EvolveScheme::ThetaImplicitEM) &&
             !m_const_dt.has_value() && m_dt_update_interval.contains(step+1)) {
```
JustinRayAngus: is `!m_const_dt.has_value()` needed here?

EZoni: This is the full code:
https://github.com/BLAST-WarpX/warpx/blob/2fd719c953ff281b50f9f4b823703e66c4aa37d1/Source/Evolve/WarpXEvolve.cpp#L193-L202

Based on the comments and the names of the functions it seems to me that it is correct as is. It seems that we want to update the time step by calling `UpdateDtFromParticleSpeeds` only if the time step is not required to be constant (i.e., only if `m_const_dt.has_value()` returns `false`).

Why would we allow to update the time step if the user requested a constant time step in the inputs file?

This was added by @archermarx in #5176, if I'm not mistaken.

EZoni:
As far as CI goes, I see only two tests whose checksums need to be reset:
- New checksums file `test_1d_theta_implicit_planar_pinch.json`:
```
{
  "lev=0": {
    "Bx": 2.347554229130745,
    "By": 2.254227700454954,
    "Bz": 0.0,
    "Ex": 551393126.2574284,
    "Ey": 491617943.23209137,
    "Ez": 714658615.6303692,
    "divE": 7016314161711.369,
    "jx": 18692954464.699497,
    "jy": 20177259050.167336,
    "jz": 12721426909.097504,
    "rho": 62.123763388017494
  },
  "deuterium": {
    "particle_momentum_x": 4.035760123452947e-19,
    "particle_momentum_y": 4.004963598310352e-19,
    "particle_momentum_z": 4.0782810816965124e-19,
    "particle_position_x": 163.6336698244632,
    "particle_weight": 1.4999875e+21
  },
  "electrons": {
    "particle_momentum_x": 6.6030232424328065e-21,
    "particle_momentum_y": 6.60813358729526e-21,
    "particle_momentum_z": 6.559430730541378e-21,
    "particle_position_x": 163.63367200208899,
    "particle_weight": 1.4999875e+21
  }
}
```
- New checksums file `test_2d_theta_implicit_planar_pinch.json`:
```
{
  "lev=0": {
    "Bx": 9.275768517413653,
    "By": 19.472307474063484,
    "Bz": 11.610979590557633,
    "Ex": 4177113295.3698635,
    "Ey": 2879051080.583498,
    "Ez": 4181787135.040885,
    "divE": 47250533638439.05,
    "jx": 91271135865.5671,
    "jy": 115116958556.87042,
    "jz": 142255859953.87067,
    "rho": 418.1507147637307
  },
  "deuterium": {
    "particle_momentum_x": 3.2203338208705e-18,
    "particle_momentum_y": 3.2127841653787054e-18,
    "particle_momentum_z": 3.2284163892421025e-18,
    "particle_position_x": 1309.3091292829426,
    "particle_position_y": 96.00796866051637,
    "particle_weight": 1.6501375e+18
  },
  "electrons": {
    "particle_momentum_x": 5.297572837196199e-20,
    "particle_momentum_y": 5.298600055672404e-20,
    "particle_momentum_z": 5.276908853465337e-20,
    "particle_position_x": 1309.3090771126394,
    "particle_position_y": 96.00571854694559,
    "particle_weight": 1.6501375e+18
  }
}
```

Source/Evolve/WarpXEvolve.cpp
JustinRayAngus: _Why would we allow to update the time step if the user requested a constant time step in the inputs file?_

- I guess it's just a matter or what takes priority assuming both are defined in the input deck. If the m_dt_update_interval is defined, then it could take priority and not care if you have const_dt defined too.

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -192,7 +192,8 @@ WarpX::Evolve (int numsteps)
 
         // Update timestep for electrostatic solver if a constant dt is not provided
         // This first synchronizes the position and velocity before setting the new timestep
-        if (electromagnetic_solver_id == ElectromagneticSolverAlgo::None &&
+        if ((electromagnetic_solver_id == ElectromagneticSolverAlgo::None ||
+            evolve_scheme == EvolveScheme::ThetaImplicitEM) &&
```
lucafedeli88: This condition on three lines may be a bit difficult to read!
Would it be possible to do something like:
```
const bool variable_with_meaningful_name = (electromagnetic_solver_id == ElectromagneticSolverAlgo::None ||
            evolve_scheme == EvolveScheme::ThetaImplicitEM);
if (variable_with_meaningful_name && !m_const_dt.has_value() && m_dt_update_interval.contains(step+1))            
```
?

Source/Evolve/WarpXEvolve.cpp
EZoni: Ah, I didn't understand that they can both be specified in the inputs. If they're mutually exclusive then maybe we should raise an error if a user specifies both and instruct the user to choose what they want, a constant time step or a time step that can be updated. I don't have in mind the use cases, so my suggestion may be off.

JustinRayAngus:
@EZoni I added an assert to ensure that a user doesn't define both `warpx.const_dt` and `warpx.dt_update_interval` at runtime. @lucafedeli88 I added an assert to make sure that warpx.dt_update_interval is only specified for ES and theta-implicit EM simulations. Both of these asserts help simplify the logic for updating dt in the Evolve loop.

EZoni (approved):
Thanks, @JustinRayAngus! Looks good to me.

ax3l (approved):
LGTM, thank you!

PR #6435, opened 2025-12-03
For implicit, set ng_fieldgather = ng_alloc_EB

dpgrote:
For the implicit solver, the values in the guard cells need to be consistent for the various grid operations.

### DIFF
--- a/Source/Parallelization/GuardCellManager.cpp
+++ b/Source/Parallelization/GuardCellManager.cpp
@@ -375,6 +375,14 @@ guardCellManager::Init (
         // for the field solve too.
         ng_FieldGather = ng_FieldGather.max(ng_FieldSolver);
 
+        if (evolve_scheme == EvolveScheme::ThetaImplicitEM ||
+            evolve_scheme == EvolveScheme::SemiImplicitEM ||
+            evolve_scheme == EvolveScheme::StrangImplicitSpectralEM) {
+            // For these implicit schemes, the number of ghost cells
+            // needs to be consistent for EB, J, and the field gather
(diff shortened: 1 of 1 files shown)


Source/Parallelization/GuardCellManager.cpp
```diff
@@ -375,6 +375,14 @@ guardCellManager::Init (
...
         ng_FieldGather = ng_FieldGather.max(ng_FieldSolver);
 
+        if (evolve_scheme == EvolveScheme::ThetaImplicitEM ||
+            evolve_scheme == EvolveScheme::SemiImplicitEM ||
+            evolve_scheme == EvolveScheme::StrangImplicitSpectralEM) {
+            // For these implicit schemes, the number of ghost cells
+            // needs to be consistent for EB, J, and the field gather
+            ng_FieldGather = ng_alloc_EB;
+        }
+
```
JustinRayAngus: What do you think about combining this block with lines 325-331 and move them to after line 389? It would keep the implicit alterations to fewer blocks.

```
if (evolve_scheme == EvolveScheme::ThetaImplicitEM ||
    evolve_scheme == EvolveScheme::SemiImplicitEM ||
    evolve_scheme == EvolveScheme::StrangImplicitSpectralEM) {
    // For these implicit schemes, the number of ghost cells
    // for EB gather must be consistent with those for J.
    ng_alloc_EB.max( ng_alloc_J );
    ng_FieldGather.max( ng_alloc_J );
}
```

JustinRayAngus (approved):

PR #6436, opened 2025-12-04
For PEC_insulator, fix the source on the boundary

dpgrote:
For PEC_insulator, the source (rho and J-parallel) need to be zeroed out on the boundary in the conductor region. Without this, the linear solvers for implicit can fail to converge.

This PR implements two versions of this to compare them.


JustinRayAngus (approved):
Thanks!

JustinRayAngus:
I was just re-reading the PR description and I realize that we never implemented the ability to zero out Rho on the boundary. Can you add that?

PR #6438, opened 2025-12-07
Poisson Solver: Synchronize nodal data before solve

### LINKED ISSUE #6425: MLMG solver fails if particles initialization is `NUniformPerCell`, not if `NRandomPerCell`
## Description
In a simple 3D simulation of a uniform plasma, the electrostatic MLMG solver doesn't converge when switching particle initialization from `NRandomPerCell` to `NUniformPerCell`.
The error is: `MLMG: Bottom solve failed.`

## How to reproduce
- WarpX inputs files: [inputs_mlmg_issue.txt](https://github.com/user-attachments/files/23778383/inputs_mlmg_issue.txt)

## System information
- Perlmutter, GPU partition, on either 1 or 4 GPUs. 
- WarpX version: 25.10-113-g607b091a8295

## Additional information
- Terminal output when converging (i.e. using `NRandomPerCell`): [output_success.txt](https://github.com/user-attachments/files/23778451/output_success.txt)
- Terminal output when failing (i.e. using `NUniformPerCell`): [output_fail.txt](https://github.com/user-attachments/files/23778450/output_fail.txt)

CC: @RemiLehe

WeiqunZhang:
The nodal density in `LabFrameExplicitES::ComputeSpaceChargeField` was not synchronized on shared nodes. Thus we need to call OverrideSync before passing it to `MLMG::solve`.

Close #6425

### DIFF
--- a/Regression/Checksum/benchmarks_json/test_2d_collisions_split_position_push_electrostatic.json
+++ b/Regression/Checksum/benchmarks_json/test_2d_collisions_split_position_push_electrostatic.json
@@ -1,32 +1,32 @@
 {
   "electrons": {
-    "particle_momentum_x": 4.9920916268383415e-20,
-    "particle_momentum_y": 4.9932646157443774e-20,
-    "particle_momentum_z": 4.9787314890781723e-20,
-    "particle_position_x": 0.00043583088230209717,
+    "particle_position_x": 0.00043512967509139156,
     "particle_position_y": 0.0,
-    "particle_position_z": 0.00043475482465095205,
+    "particle_position_z": 0.00043538071799171445,
+    "particle_momentum_x": 4.952035070313062e-20,
+    "particle_momentum_y": 5.040983554665706e-20,
+    "particle_momentum_z": 4.98106129752681e-20,
     "particle_weight": 2823958725036869.5
   },
   "lev=0": {
     "Bx": 0.0,
     "By": 0.0,
     "Bz": 0.0,
-    "Ex": 20406461006491.938,
+    "Ex": 19244627289974.41,
     "Ey": 0.0,
-    "Ez": 20535265033355.64,
-    "jx": 7.047452932261251e+18,
-    "jy": 7.081966241139403e+18,
-    "jz": 7.50640321897593e+18,
-    "rho": 70664244105.11731
+    "Ez": 16678270172012.957,
+    "jx": 7.681633328331679e+18,
+    "jy": 6.262294172346471e+18,
+    "jz": 9.108171958470861e+18,
+    "rho": 61049608867.81595
   },
   "protons": {
-    "particle_momentum_x": 2.1468903064935752e-18,
-    "particle_momentum_y": 2.150458744659307e-18,
-    "particle_momentum_z": 2.12572635061564e-18,
-    "particle_position_x": 0.00043574422696149545,
+    "particle_position_x": 0.00043611710145463297,
     "particle_position_y": 0.0,
-    "particle_position_z": 0.000435118394748523,
+    "particle_position_z": 0.00043527724613341977,
+    "particle_momentum_x": 2.133089681191217e-18,
+    "particle_momentum_y": 2.1593836378778734e-18,
+    "particle_momentum_z": 2.1385918368539213e-18,
     "particle_weight": 2823958725036869.5
   }
 }
--- a/Source/ablastr/fields/PoissonSolver.H
+++ b/Source/ablastr/fields/PoissonSolver.H
@@ -412,6 +412,8 @@ computePhi (
             mlmg.setFinalFillBC(true);
         }
 
+        rho[lev]->OverrideSync(geom[lev].periodicity());
+
         // Solve Poisson equation at lev
         mlmg.solve( {phi[lev]}, {rho[lev]},
                      relative_tolerance, absolute_tolerance );
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -147,7 +147,9 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: Dependencies
-      run: .github/workflows/dependencies/nvhpc.sh 25.1
+      run: |
+        .github/workflows/dependencies/ubuntu_free_disk_space.sh
+        .github/workflows/dependencies/nvhpc.sh 25.1
     - name: CCache Cache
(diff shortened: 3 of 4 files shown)


Regression/Checksum/benchmarks_json/test_2d_collisions_split_position_push_electrostatic.json
```diff
@@ -1,32 +1,32 @@
 {
   "electrons": {
-    "particle_momentum_x": 4.9920916268383415e-20,
-    "particle_momentum_y": 4.9932646157443774e-20,
-    "particle_momentum_z": 4.9787314890781723e-20,
-    "particle_position_x": 0.00043583088230209717,
+    "particle_position_x": 0.00043512967509139156,
```
ax3l, line 3: @EZoni @WeiqunZhang just pinging here to sync on the test and its tolerance/coarseness, because the value changes are surprisingly large.

RemiLehe:
@WeiqunZhang Do you understand why one of the checksum files changed some much?
How did you reset this file: on your local computer or by using the values displayed in the CI runner?

Source/ablastr/fields/PoissonSolver.H
```diff
@@ -412,6 +412,8 @@ computePhi (
             mlmg.setFinalFillBC(true);
         }
 
+        rho[lev]->OverrideSync(geom[lev].periodicity());
```
aeriforme, line 415: Can anybody explain what this does? 
What does "sync" mean here? 
Why is there `.periodicity()`?

Regression/Checksum/benchmarks_json/test_2d_collisions_split_position_push_electrostatic.json
EZoni: Note that this test uses the new algorithm added in #5955.

WeiqunZhang:
I reset it by using the values in the CI.

As for the big checksum change, I think the difference starting from roundoff can get amplified a lot because the solver is iterative.

Source/ablastr/fields/PoissonSolver.H
WeiqunZhang: Suppose there is a valid point that is shared by 4 boxes. Charge deposition puts `a`, `b`, `c`, and `d` to the 4 points in the 4 boxes. We then call `SumBoundary` to add the 4 values together and store the values in the 4 boxes. Box 0 might be the result of `a + b + c + d`, whereas Box 1 might be the result of `b + a + c + d`. So the data are out of sync. `OverrideSync` tries to fix it by using one value (e.g. the one from Box 0) to override other values.  Why `periodicity()`? Because nodal points can be shared across periodic boundaries.

EZoni:
> As for the big checksum change, I think the difference starting from roundoff can get amplified a lot because the solver is iterative.

I think part of what seems puzzling is that this happened only for one test out of probably several that use the Poisson solver.

WeiqunZhang:
If the roundoff error does not change the number of iterations in the linear solver, the difference might be small. But if the number of iterations changes, the difference will be the solver's tolerance level. If random number generation is involved, small roundoff errors can also result in big differences.

Source/ablastr/fields/PoissonSolver.H
aeriforme: Thank you the explanation!

EZoni: I also see this here and have two questions:

Screenshot from 2025-12-12 09-22-20

<br>
<br>

1. Is the code you added in the PR, namely
```cpp
rho[lev]->OverrideSync(geom[lev].periodicity());
```
then equivalent to the following (plus/minus some `*`...)?
```cpp
auto mask = amrex::OwnerMask(*rho[lev], geom[lev].periodicity());
rho[lev]->OverrideSync(*mask, geom[lev].periodicity());
```
Are these two different ways to do the exact same thing or is there any subtle difference?

<br>
<br>

2. Should we add your example ("Suppose there is a valid point...") to the docstring of `OverrideSync` in https://github.com/AMReX-Codes/amrex/blob/bb5b55299976aedbf92a24b1ca5fabf8e6fca213/Src/Base/AMReX_MultiFab.H#L708-L709,
```cpp
    //! Sync up nodal data with owners overriding non-owners
    void OverrideSync (const iMultiFab& msk, const Periodicity& period = Periodicity::NonPeriodic());
```
which is probably where an IDE points to when one is reading through the code and trying to understand it? I would be happy to open a PR to do that, if you (both) think it could be useful.

WeiqunZhang: Thanks for the questions. The end results of the two approaches are the same. But one should prefer `rho[lev]->OverrideSync(geom[lev].periodicity());` version to `rho[lev]->OverrideSync(*mask, geom[lev].periodicity());`. The former has better performance, whereas the latter gives the user the flexibility of providing their own mask. In the documentation, we should mention both. The performance boost of the first version was a recent development. If you can help improving the documentation, that would be awesome! Thank you in advance!

EZoni: AMReX PR: https://github.com/AMReX-Codes/amrex/pull/4849.

aeriforme (approved):
I tested this with my original failing simulation, and it works. 
Therefore, I approve.

Regression/Checksum/benchmarks_json/test_2d_collisions_split_position_push_electrostatic.json
```diff
@@ -1,32 +1,32 @@
...
-    "particle_position_x": 0.00043574422696149545,
+    "particle_position_x": 0.00043611710145463297,
     "particle_position_y": 0.0,
-    "particle_position_z": 0.000435118394748523,
+    "particle_position_z": 0.00043527724613341977,
+    "particle_momentum_x": 2.133089681191217e-18,
+    "particle_momentum_y": 2.1593836378778734e-18,
+    "particle_momentum_z": 2.1385918368539213e-18,
     "particle_weight": 2823958725036869.5
   }
```
RemiLehe, line 31: In this test, electrons and protons were initialized with:
```
protons.density = n0
protons.num_particles_per_cell_each_dim = 8 8

electrons.density = n0
electrons.num_particles_per_cell_each_dim = 8 8
```
So it is similar to https://github.com/BLAST-WarpX/warpx/issues/6425

PR #6443, opened 2025-12-09
Add documentation for implicit attributes

dpgrote:
This adds to the documentation the particle attributes that are created when the implicit solver is used.

Note that no specific default value is needed for the attributes since they will always be set before being used.


JustinRayAngus (approved):
Looks good

PR #6445, opened 2025-12-09
Free disk space in nvhpc CI

WeiqunZhang:
The script is borrowed from AMReX. For the nvhpc CI using ubuntu-latest, the free space before was 18G, it becomes 23G after the cleanup.

### DIFF
--- a/.github/workflows/cuda.yml
+++ b/.github/workflows/cuda.yml
@@ -147,7 +147,9 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: Dependencies
-      run: .github/workflows/dependencies/nvhpc.sh 25.1
+      run: |
+        .github/workflows/dependencies/ubuntu_free_disk_space.sh
+        .github/workflows/dependencies/nvhpc.sh 25.1
     - name: CCache Cache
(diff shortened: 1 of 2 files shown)


EZoni (approved):

EZoni:
@ax3l 

This looks good to me. Any comments or objections on your end? Since the code is borrowed from AMReX, I would merge it as is to maintain consistency.

.github/workflows/cuda.yml
```diff
@@ -147,7 +147,9 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: Dependencies
-      run: .github/workflows/dependencies/nvhpc.sh 25.1
+      run: |
+        .github/workflows/dependencies/ubuntu_free_disk_space.sh
+        .github/workflows/dependencies/nvhpc.sh 25.1
```
ax3l, line 152: We can also use this one, which is relatively complete and fast:

https://github.com/BLAST-ImpactX/impactx/blob/25.11/.github/workflows/cuda.yml#L22-L31

https://github.com/ax3l/free-disk-space

ax3l (approved):

PR #6450, opened 2025-12-09
CMake: No Warn for No-MPI Tests

ax3l:
Our strategy for CMake is to enable tests that work in the current configuration, and silently not add other tests.

When building w/o MPI, we currently fill the terminal with warnings. This removes those.


ax3l:
The "clean" thing we can do is mark those tests as "skipped" instead of not adding them in the first place, but this is good enough for now.

EZoni (approved):

EZoni:
@ax3l 

Feel free to merge #6445 and rebase here if you want to fix the CUDA / NVHPC CI issue.

PR #6451, opened 2025-12-11
Add option to initialize a Gaussian beam using total number of particles instead of total charge 

aeriforme:
A Gaussian beam can now be initialized using the input parameter `<species>.npart_real` instead of `<species>.q_tot`.
Only one of the two should be specified by the user:
* if `npart_real` is specified, the weights are calculated as `npart_real/npart` (note that, in the user interface, we use `npart` to mean the number of macroparticles, which I think is a bit meh)
* if `q_tot` is specified, the weights are calculated as `q_tot / (npart * charge)`, as it was before this PR. 

This is useful for neutral species.


Source/Initialization/PlasmaInjector.cpp
```diff
@@ -238,14 +238,26 @@ void PlasmaInjector::setupGaussianBeam (amrex::ParmParse const& pp_species)
     utils::parser::queryWithParser(pp_species, source_name, "x_cut", x_cut);
     utils::parser::queryWithParser(pp_species, source_name, "y_cut", y_cut);
     utils::parser::queryWithParser(pp_species, source_name, "z_cut", z_cut);
-    utils::parser::getWithParser(pp_species, source_name, "q_tot", q_tot);
+
+    const bool q_tot_is_specified = pp_species.contains("q_tot");
+    const bool N_tot_is_specified = pp_species.contains("npart_real");
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE( (q_tot_is_specified || N_tot_is_specified) && !(q_tot_is_specified && N_tot_is_specified),
```
dpgrote: The check can be simplified to `q_tot_is_specified != N_tot_is_specified`

dpgrote (approved):
This looks good, with one minor comment.

Issue #6158, opened 2025-09-11
Dynamic Splitting of Boxes


ax3l:
After #5812, our primary goal is to enable splitting boxes dynamically as simulations progress (and later on, if needed, merge some again).

I created a small test case (CPU, laptop-size) that we can use to experiment with dynamic splitting of boxes: https://drive.google.com/drive/u/0/folders/1_jyUBo13_5JpmbU_keB2BrEUB2MTtqqP

This is very representative of cases in KISMET (modeled after thrust 3, but applies similarly to the hot spot physics and our current CY25 INCITE).

ax3l:
Update: the original test gets too hot and spreads the particle distribution too equally after a few 100 steps.

Idea to modify:
- Slowly moving window or bulk mean velocity
- Turn off laser and initial temperature to keep stationary distribution (that moves)
- Put target at high z end and move it over the sim box
- Expect: merge boxes and split new ones as window goes over plasma

I will upload a revised test

PR #6452, opened 2025-12-12
Custom weights for initial DistributionMapping

WeiqunZhang:
By default, amrex uses box volumes as weights when distributing boxes. But this would somewhat defeat the purpose of splitting high density boxes until the next load balance is performed. Thus, when splitting box is on, we are going to assume that the boxes have equal weights.

For a test based on #6158, this improves the initial load balance efficiency from 0.26 to 0.63 when running on 8 processes.

The figures below show the distribution mapping without and with this PR.

Screenshot from 2025-12-11 16-07-56

### DIFF
--- a/Source/WarpX.cpp
+++ b/Source/WarpX.cpp
@@ -3517,8 +3517,10 @@ amrex::DistributionMapping
 WarpX::MakeDistributionMap (int lev, amrex::BoxArray const& ba)
 {
     bool roundrobin_sfc = false;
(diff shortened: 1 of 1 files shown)


WeiqunZhang:
Screenshot from 2025-12-11 16-10-13

This is the number of particles per cell in that test.

ax3l (approved):
Thank you, this looks great!!

PR #6456, opened 2025-12-12
Fix undefined variables in PICMI test scripts

oshapoval:
This follow-up PR (initial https://github.com/BLAST-WarpX/warpx/pull/6354) makes electrons a global variable and removes previously added conditional deletion of electrons to prevent pre-commit check failures.


EZoni:
@oshapoval 

There are some conflicts that need to be resolved, otherwise the pre-commit CI check cannot run properly.

EZoni (approved):

PR #6453, opened 2025-12-12
[WIP] Fix bug in energy-preserving explicit electrostatic PIC-MCC algorithm

EZoni:
This may fix a bug that @kale-j observed in an electromagnetic simulation based on #6275, and that was troubleshooted by @oshapoval, @RemiLehe, and myself.

This PR takes care of the bug fix for the electrostatic case, originally implemented in #5955. The bug fix for the electromagnetic case is included in #6275.

This change may break a number of CI tests, so I flag the PR as "[WIP]" until the CI checks pass.

PR #6458, opened 2025-12-15
collisions should be after call to HandleParticlesAtBoundaries()

JustinRayAngus:
This PR moves the collisions for the implicit solver to the proper place in the Evolve loop. This fixes a bug introduced in PR #6428, where the collisions are moved from just before to just after the PIC advance. (See https://github.com/BLAST-WarpX/warpx/issues/6449). However, with this change I noticed odd behavior in simulation outputs at processor boundaries late in time for highly collisional simulations. The fix is that the collisions should be placed after the particle communication and boundary handling is performed, which occurs later in the Evolve loop.


EZoni (approved):
Looks good to me, I think this is similar to the fix we had to implement in #6275 and #6453.

RemiLehe (approved):

PR #6467, opened 2025-12-29
check for out of bounds particles with cropping in suborbits.

JustinRayAngus:
Particle suborbits were added in PR #5969. Suborbits are needed for improved accuracy of particle orbits when using large time steps. Particle orbit cropping at physical boundaries was added in PR #5649. This is used to achieve rigorous charge conservation for outflow particles at physical boundaries (conductor and insulator).

A bug occurs when the initial position of a suborbit is out of bounds (past the physical boundary) where cropping is being done. In this scenario, the particle will deposit current when it should not be doing so, breaking energy and charge conservation. This PR fixes this bug by checking for particles with the initial position out of bounds in the suborbit routine and neglecting deposition for these suborbits.

### DIFF
--- a/Source/Particles/Pusher/ImplicitPushPX.cpp
+++ b/Source/Particles/Pusher/ImplicitPushPX.cpp
@@ -151,6 +151,8 @@ namespace {
         zp = zp_n + dzp;
         setPosition(ip, xp, yp, zp);
 
+        // Propogate ballistically if the suborbit starts out of bounds, avoiding
+        // field gather and the possiblity that the particle orbit re-enters the domain.
         if (this_suborbit_out_of_bounds) { return true; }
 
         bool convergence = false;
--- a/Source/Particles/Pusher/ImplicitPushPX.cpp
+++ b/Source/Particles/Pusher/ImplicitPushPX.cpp
@@ -61,6 +61,7 @@ namespace {
         int const & ip,
         amrex::Real const & dt,
         SetParticlePosition<PIdx> const &  setPosition,
+        bool const this_suborbit_out_of_bounds,
         amrex::ParticleReal & xp,
         amrex::ParticleReal & yp,
         amrex::ParticleReal & zp,
@@ -149,6 +150,8 @@ namespace {
         yp = yp_n + dyp;
         zp = zp_n + dzp;
         setPosition(ip, xp, yp, zp);
+
+        if (this_suborbit_out_of_bounds) { return true; }
 
         bool convergence = false;
         for (int iter=0; iter < max_iterations; iter++) {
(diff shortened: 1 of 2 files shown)


Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -150,6 +151,8 @@ namespace {
         zp = zp_n + dzp;
         setPosition(ip, xp, yp, zp);
```
dpgrote, line 153: Add a comment

```suggestion
// If a suborbit starts out of bounds, propagate it ballistically (to avoid field gathers in the guard cells)
// and so it cannot be turned back into the domain.
```

Source/Utils/ParticleUtils.H
```diff
@@ -311,6 +311,57 @@ namespace ParticleUtils {
...
     }
 
+    /* \brief Check if the particle position is out of bounds at a boundary
+     *        where orbit cropping is being performed.
+     * \param[in] xp, yp, zp      The particle position
+     * \param[in] dinv            3D cell size inverse
+     * \param[in] xyzmin          Physical lower bounds of domain
+     * \param[in] domain_double   The domain boundary indicies cast to double values
+     * \param[in] do_cropping     Whether to crop particle orbits at domain boundaries
+     */
```
dpgrote, line 322:
```suggestion
     * \return whether position is out of bounds
     */
```

Source/Utils/ParticleUtils.H
```diff
@@ -311,6 +311,57 @@ namespace ParticleUtils {
...
+        xp_grid[0] = (xp - xyzmin.x)*dinv.x;
+        xp_grid[1] = (yp - xyzmin.y)*dinv.y;
+        xp_grid[2] = (zp - xyzmin.z)*dinv.z;
+#elif defined(WARPX_DIM_XZ)
+        xp_grid[0] = (xp - xyzmin.x)*dinv.x;
+        xp_grid[1] = (zp - xyzmin.z)*dinv.z;
+#elif defined(WARPX_DIM_RSPHERE)
+        const amrex::Real rp = std::sqrt(xp*xp + yp*yp + zp*zp);
+        xp_grid[0] = (rp - xyzmin.x)*dinv.x;
+#endif
```
dpgrote: This could be `#elif`

Source/Utils/ParticleUtils.H
```diff
@@ -311,6 +311,57 @@ namespace ParticleUtils {
...
+#if defined(WARPX_DIM_RZ) || defined(WARPX_DIM_RCYLINDER)
+        const amrex::Real rp = std::sqrt(xp*xp + yp*yp);
+        xp_grid[0] = (rp - xyzmin.x)*dinv.x;
+#if defined(WARPX_DIM_RZ)
+        xp_grid[1] = (zp - xyzmin.z)*dinv.z;
+#endif
+#endif
+        for (int dim = 0; dim < AMREX_SPACEDIM; dim++) {
+            if (do_cropping[dim][0] && xp_grid[dim] <= domain_double[dim][0]) {
```
dpgrote, line 357: This could be simplified
`out_of_bounds |= (do_cropping[dim][0] && xp_grid[dim] <= domain_double[dim][0]);`
instead of using the `if` statement.

JustinRayAngus: If it's all the same, I prefer using if statements here.

dpgrote (approved):
Looks good to me!

JustinRayAngus:
@dpgrote I found the bug that made the suborbit results on the deposition pass different from those on the first pass. I've committed that fix here. Unfortunately, it doesn't seem to be the culprit for the non-converged Newton solver that I'm seeing on GPUs.

PR #6469, opened 2026-01-05
Doc: AMD GPU Debugging

ax3l:
https://rocm.docs.amd.com/projects/HIP/en/docs-6.0.0/how_to_guides/debugging.html#summary-of-environment-variables-in-hip

### DIFF
--- a/Docs/source/usage/workflows/debugging.rst
+++ b/Docs/source/usage/workflows/debugging.rst
@@ -37,6 +37,7 @@ Try the following steps to debug a simulation:
 #. If the problem looks like a random memory might be used, try initializing memory with signaling Not-a-Number (NaN) values through the runtime option ``fab.init_snan = 1``.
    Further useful runtime options are ``amrex.fpe_trap_invalid``, ``amrex.fpe_trap_zero`` and ``amrex.fpe_trap_overflow`` (see details in the AMReX link below).
 #. On Nvidia GPUs, if you suspect the problem might be a race condition due to a missing host / device synchronization, set the environment variable ``export CUDA_LAUNCH_BLOCKING=1`` and rerun.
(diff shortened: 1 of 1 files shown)


Docs/source/usage/workflows/debugging.rst
```diff
@@ -37,6 +37,7 @@ Try the following steps to debug a simulation:
 #. If the problem looks like a random memory might be used, try initializing memory with signaling Not-a-Number (NaN) values through the runtime option ``fab.init_snan = 1``.
    Further useful runtime options are ``amrex.fpe_trap_invalid``, ``amrex.fpe_trap_zero`` and ``amrex.fpe_trap_overflow`` (see details in the AMReX link below).
 #. On Nvidia GPUs, if you suspect the problem might be a race condition due to a missing host / device synchronization, set the environment variable ``export CUDA_LAUNCH_BLOCKING=1`` and rerun.
+   On AMD GPUs, the equivalent variable is ``export AMD_SERIALIZE_KERNEL=3``.
```
WeiqunZhang: In my opinion, we should recommend using `export HIP_LAUNCH_BLOCKING=1`, because it's easier to remember.
```suggestion
   On AMD GPUs, the equivalent variable is ``export HIP_LAUNCH_BLOCKING=1``.
```

ax3l: I picked it because `AMD_SERIALIZE_KERNEL=3` covers [more cases](https://rocm.docs.amd.com/projects/HIP/en/docs-6.0.0/how_to_guides/debugging.html#summary-of-environment-variables-in-hip) (*Wait for completion before and after enqueue.*)

In practice, I think it might not make a big difference besides corner cases... But I think the latter is safer for host access to running kernels memory...?

WeiqunZhang (approved):

PR #6102, opened 2025-08-19
WarpX Citation

ax3l:
Rework the WarpX recommended citation as discussed in the Tue, Aug 19 developer meeting.

[Preview here.](https://warpx--6102.org.readthedocs.build/en/6102/acknowledge_us.html)

- [x] Zenodo: add new co-authors to 
- [x] Zenodo: remove "contributors" (not co-authors) that had minimal contributions
- [x] finalize more in the style of https://openpmd-api.readthedocs.io/en/0.16.1/citation.html#openpmd-api w/ link to homepage. two sections, each with a tab "copy bibtex" and a tab "copy plain text":
    - [x] One that is "how to credit generally WarpX (2018)" via the general Zenodo link
    - [x] One that is "how to reproducibly cite a specific version of WarpX (2025) via the version-specific Zenodo link

PR #6473, opened 2026-01-06
Fix Zenodo JSON Schema

ax3l:
Fixed via validation with https://www.npmjs.com/package/zenodraft

Broken since #6102

Not too easy to validate automatically: https://github.com/zenodo/zenodo-rdm/issues/1205

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -270,34 +270,12 @@
     "related_identifiers": [
     {
       "identifier": "https://warpx.readthedocs.io",
-      "relation_type": {
-        "id": "isdocumentedby",
-        "title": {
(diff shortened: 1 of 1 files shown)


EZoni (approved):

EZoni:
I still see `25.11` at https://zenodo.org/records/17539393 after merging this PR, so we may need more fixes. Do you happen to have access to any log that may show the actual issue, besides the validator linked above?

ax3l:
We only get new releases via our Zenodo webhook once we draft the new release.

The 25.11 and 26.01 releases are kinda "lost" now.

PR #6481, opened 2026-01-09
Small bugfix in particle container wrapper

roelof-groenewald:
`get_particle_container_from_name()` no longer exists, instead it is now just called `get`.
This functionality is deprecated anyway but still worth fixing until we remove `particle_containers.py` completely.

### DIFF
--- a/Docs/source/usage/workflows/python_extend.rst
+++ b/Docs/source/usage/workflows/python_extend.rst
@@ -301,7 +301,7 @@ Particles
 
        # data access
        multi_pc = warpx.multi_particle_container()
-       pc = multi_pc.get_particle_container_from_name("electrons")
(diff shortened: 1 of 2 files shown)


EZoni:
Should we update also this code copied in the documentation?

https://github.com/BLAST-WarpX/warpx/blob/6818ad4eb78ad622d30e601205a60fd7c10675fc/Docs/source/usage/workflows/python_extend.rst?plain=1#L304

RemiLehe:
@EZoni Agreed. More generally, the whole Python documentation is being refactored here, as a follow-up to the documentathon:
https://github.com/BLAST-WarpX/warpx/pull/6368

RemiLehe (approved):

PR #6490, opened 2026-01-13
Fix Cursor issue when running Python through CTest

RemiLehe:
When asking Cursor to run `CTest` tests, it seems that it is unable to find the correct Python (in my case the `conda` one) and uses the system Python instead, thus not finding `numpy` and `yt`:

Screenshot 2026-01-13 at 2 26 28 PM

Strangely, this error only happens on MacOS. (I don't get this error when running Cursor on Linux.)
Additionally, when I run the same command **myself directly in the Terminal**, I don't get the error (even on MacOS).

The changes made to the CMakeList.txt fix the issue, but were suggested by a coding assistant and I do not fully understand them. They should be simplified/updated before merging this PR.

- [x] follow-up PR: also add same change in ImpactX https://github.com/BLAST-ImpactX/impactx/pull/1267

### DIFF
--- a/Examples/CMakeLists.txt
+++ b/Examples/CMakeLists.txt
@@ -190,27 +190,13 @@ function(add_warpx_test
     set_property(TEST ${name}.run APPEND PROPERTY ENVIRONMENT "OMP_NUM_THREADS=1")
 
     # Always use Python_EXECUTABLE for analysis/checksum scripts to ensure conda Python is used
-    if(analysis OR checksum)
+    if(python OR analysis OR checksum)
+        # Try to find Python if not already set
+        # This is more robust in virtual envs (conda, venv) than
+        # relying only on /usr/bin/env python3
         if(NOT Python_EXECUTABLE)
-            # Try to find Python if not already set
-            find_package(Python3 COMPONENTS Interpreter QUIET)
-            if(Python3_EXECUTABLE)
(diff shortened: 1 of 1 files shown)


Examples/CMakeLists.txt
```diff
@@ -189,10 +189,14 @@ function(add_warpx_test
...
-        set(THIS_Python_SCRIPT_EXE ${Python_EXECUTABLE})
-    else()
-        set(THIS_Python_SCRIPT_EXE "")
+    # Always use Python_EXECUTABLE for analysis/checksum scripts to ensure conda Python is used
+    if(python OR analysis OR checksum)
+        # Try to find Python if not already set
+        # This is more robust in virtual envs (conda, venv) than
+        # relying only on /usr/bin/env python3
+        if(NOT Python_EXECUTABLE)
+            find_package(Python COMPONENTS Interpreter)
```
ax3l: Could even make this error when it cannot find one:
```suggestion
            find_package(Python COMPONENTS Interpreter REQUIRED)
```

ax3l:
@RemiLehe can you try it now again? I cleaned it up and make it more sensible/robust.

RemiLehe:
Thanks a lot! 🚀 It works for me!

ax3l (approved):

Issue #6487, opened 2026-01-13
Performance of WarpX in serial problems


dpgrote:
It has been reported multiple times that that WarpX can sometimes be slower than expected for problems when run of serial. There are multiple possible reasons for this, including vectorization on CPU, excessive kernel launches (on GPU), and relatedly, too little work per kernel launch. This issue provides a place to discuss this topic.

As a start, results from a test case of a plasma in a box using 2D electrostatic is shown. This is a somewhat small test case, with a 64x64 grid and 400 particles per cell. This table shows the timings of Warp and WarpX (on a Mac M3, single CPU, single thread) and WarpX (one HIP GPU, with warpx.do_device_synchronize = 0).

```
                 Warp   WarpX   WarpX
                 CPU    CPU     HIP
Total            2.8    11.3    1.2
GatherAndPush    1.7    5.8     0.002
fetche3d         1.03
RedistributeCPU         2.2     0.03
boundaries       0.33   1.9     0.016
deposition       0.33   0.76    0.17
computephi       0.07   0.13    0.9
```
The thing to note is that it is not one thing that dominates the time, but the timing difference arises in multiple places. Specifically though, on the GPU, the Poisson solve takes a significantly larger fraction of the time. The input files are posted below.

dpgrote:
Here is the link to the Warp code, https://bitbucket.org/berkeleylab/warp/src/master.
Here are the input files:

[input_warp.py](https://github.com/user-attachments/files/24598423/input_warp.py)

[input_warpx.txt](https://github.com/user-attachments/files/24598474/input_warpx.txt)

AlexanderSinn:
When running on GPU, make sure to set `warpx.do_device_synchronize = 0`.

dpgrote:
> When running on GPU, make sure to set `warpx.do_device_synchronize = 0`.

Thanks - that shaves off about 1 s. I've updated the table above.

AlexanderSinn:
Note that some TinyProfiler readings might not be accurate with that setting.
Another simple optimization is in `warpx\Source\ablastr\fields\PoissonSolver.H` after line 406 add `mlmg.setNoGpuSync(true);`.
I forgot that this was not done by default.
Default:
<img width="1023" height="1151" alt="Image" src="https://github.com/user-attachments/assets/21b0998e-c006-42ef-950d-6a4b763052f2" />
NoGpuSync:

[...]

PR #6491, opened 2026-01-13
Use NoGpuSync for MLMG in computePhi

AlexanderSinn:
This gives a significant performance improvement when running small problem sizes in GPU.
See #6487

### DIFF
--- a/Source/ablastr/fields/PoissonSolver.H
+++ b/Source/ablastr/fields/PoissonSolver.H
@@ -404,6 +404,7 @@ computePhi (
         mlmg.setVerbose(verbosity);
         mlmg.setMaxIter(max_iters);
         mlmg.setConvergenceNormType(amrex::MLMGNormType::greater);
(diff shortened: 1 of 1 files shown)


RemiLehe:
@WeiqunZhang Should this be the default for MLMG in AMReX?

WeiqunZhang:
We probably should change the default in amrex. I will run some tests.

WeiqunZhang (approved):

WeiqunZhang:
We might need more testing. The reason we did not have it as default was because there might be some temporary memory allocation that requires explicit sync.

ax3l:
Is this safe to merge for WarpX though? LGTM otherwise, but did not yet test on GPU.

ax3l (approved):

ax3l:
Yeah, sounds like it is safe for us.

ax3l:
attn @bathmatt

PR #6415, opened 2025-11-21
WarpXSolverVec: Remove duplicates

WeiqunZhang:
This appears necessary for PETSc to work.

PR #6498, opened 2026-01-15
fix bug in MatrixPC for 1D.

JustinRayAngus:
This bug, which prevents using the matrix PC in 1D, was accidentally introduced in PR #6415.


dpgrote (approved):
Looks good.

PR #6501, opened 2026-01-16
Dynamically Split/Merge High Density Boxes for Load Balancing

### LINKED ISSUE #6158: Dynamic Splitting of Boxes

WeiqunZhang:
Close #6158 

Previously we have introduced Box splitting to improve load balancing. This behavior is controlled by three parameters:
```ini
    warpx.split_high_density_boxes=false
    warpx.split_high_density_boxes_threshold=1.1
    warpx.split_high_density_boxes_min_box_size=8
```

In this PR, we have added supported for merging Boxes as well. Merging is controlled by the same runtime parameters. A merge is performed when the merged Boxes have a load below the threshold and a size below the max grid size.

### DIFF
--- a/Source/Parallelization/WarpXRegrid.cpp
+++ b/Source/Parallelization/WarpXRegrid.cpp
@@ -54,6 +54,213 @@
 
 using namespace amrex;
 
+std::pair<BoxArray, DistributionMapping>
+WarpX::LoadBalanceMakeNewLayout (int lev)
+{
+#ifndef AMREX_USE_MPI
+    amrex::ignore_unused(this, lev);
+    return std::make_pair(amrex::BoxArray{},amrex::DistributionMapping{});
+#else
+
+    auto comm = ParallelContext::CommunicatorSub();
+    int myproc = ParallelContext::MyProcSub();
+    int nprocs = ParallelContext::NProcsSub();
+
+    auto const& cst = *costs[lev];
+    auto const& ba = cst.boxArray();
+    auto const& dm = cst.DistributionMap();
+    auto const& mgs = maxGridSize(lev);
+
+    BoxList newbl{};
+    BoxArray newba{};
+    DistributionMapping newdm{};
+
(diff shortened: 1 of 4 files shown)


ax3l:
> A merge is performed when the merged Boxes have a load below the threshold and a size below the max grid size.

@WeiqunZhang do you think this can lead to a lot of flip-flopping? I wonder if we want to make split/merge more "sticky" by defining a range, below the range we merge, above we split. Could be done later as well.

ax3l:
> Previously we have introduced Box splitting to improve load balancing.

@WeiqunZhang maybe I missed it, just checking: before we only had splitting for initial load balancing at t=0 #5812 #6159

Do I see in the diff correctly that this PR works on _dynamic_ splitting & merging during the sim runtime? :star_struck:

WeiqunZhang:
Yes, it's dynamics during the entire run.

WeiqunZhang:
> > A merge is performed when the merged Boxes have a load below the threshold and a size below the max grid size.
> 
> @WeiqunZhang do you think this can lead to a lot of flip-flopping? I wonder if we want to make split/merge more "sticky" by defining a range, below the range we merge, above we split. Could be done later as well.

The merge threshold is 0.9 of the split threshold. So it can tolerate some fluctuation. If this turns out to be an issue, we can adjust it. Right now that number is hardwired. If you prefer, we can turn it into a runtime parameter. But having too many parameters can also be confusing.

PR #6503, opened 2026-01-16
Implement particle-absorbing boundary

atmyers:
Phase-space diagram of the electrons from a laser-plasma interaction test similar to the one in [Miller et. al. 2021](https://doi.org/10.1063/5.0065232) approximately 0.72 picoseconds after the leading edge of the laser pulse interacts with the plasma: 

phase_space

Same, but with a thermalizing region in the last 5 microns of the simulation domain that applies to particles with normalized momenta over 1.0 and draws new velocities from a Gaussian distribution with temperature parameter theta = 1.0:

phase_space

- [x] add lightweight test


atmyers:
You turn this on by adding the following to your inputs file:

```
particle_thermalizer.normal = "z"
particle_thermalizer.start = 145e-6
particle_thermalizer.end = 150e-6
particle_thermalizer.momentum_threshold = 1.0
particle_thermalizer.theta = 1.0
```

atmyers:
Another test, you can see a line (in 2D) of particles entering a region with the thermalizer and being slowed down:

before_Slice_z_Ey

middle_Slice_z_Ey

after_Slice_z_Ey

atmyers:
The stopping probability as a function of distance for a thermalizing region start at 100 microns and ending at 150:

pstop

Docs/source/usage/parameters.rst
```diff
@@ -653,6 +653,63 @@ additionally define the electric potential at the embedded boundary with an anal
...
+the boundary can lead to a non-physical return current of hot electrons that can have an
+effect on the plasma instabilities and laser-plasma interaction under study.
+
+To mitigate, WarpX implements a particle thermalizing region that reduces the flux of particles
+leaving the simulation domain that leads to the non-physical build-up of electric fields at the boundary. The
+method used is similar to that of `Miller et al. (Phys. Plasmas 28, 112702 (2021)) <https://doi.org/10.1063/5.0065232>`__.
+
+The user specifies a region in which particles will be thermalized, a normal direction, a temperature, and a
+momentum threshold. Inside the thermalizing region, the probability that a particle will be affected increases
+from 0 to 1 as $(1/(1-x))**(1/4)$. Particles that are affected have their momenta in the normal direction thermalized
```
```suggestion
from 0 to 1 as :math:`\frac{1}{1-x}^{1/4}`. Particles that are affected have their momenta in the normal direction thermalized
```

Source/Particles/ParticleThermalizer/ParticleThermalizer.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+void ParticleThermalizer::applyThermalizer(WarpXParticleContainer &pc)
+{
+    for (int lev = 0; lev < pc.numLevels(); ++lev) {
+        const auto& geom = pc.Geom(lev);
+        const auto& dx = geom.CellSizeArray();
+        const auto& problo = geom.ProbLoArray();
+        int dir = static_cast<int>(m_normal);
+
+        // TODO - Handle non-3D
```
ax3l: @atmyers do you want to add a runtime exception for now? Looking at the code below, I think only RZ, RSPHERE and RCYLINDER are missing?

ax3l (approved):
This looks great to me!

I think only a runtime assert for RZ (2D) and the two 1D Rsphere/cylinder modes are missing.

Source/Particles/ParticleThermalizer/ParticleThermalizer.cpp
atmyers: I should remove this comment. 

Currently, it will abort on initialization if run in those geometries (see [here](https://github.com/BLAST-WarpX/warpx/pull/6503/changes/BASE..8fcb5f9bb21c31224bd76838d208c62df1460df1#diff-5832be3660f900ee182dd656fabe9a4f25eb28c8b55edfd0a28c23191217cee9R53) in the constructor). So the code will never make it to this point. Would you rather throw an exception than Abort?

atmyers:
```suggestion
```

Examples/Tests/particle_absorbing_boundary/analysis.py
```diff
@@ -0,0 +1,22 @@
...
+from openpmd_viewer import OpenPMDTimeSeries
+
+ts = OpenPMDTimeSeries("./diags/reducedfiles/PhaseSpaceElectrons")
+it = ts.iterations
+data, info = ts.get_field(field="data", iteration=8000, plot=False)
+
+# This correspoonds to the region of phase space within 50 microns of the boundary
+# and with normalized momenta between -5 and -1. Without the thermalizer the total
+# weight of particles in this region is > 1e22
+assert data[250:317, 667:].sum() < 3.2e20
```
EZoni: Do you think there is a way to rewrite these numbers, `250`, `317`, and `667`, as results of operations on quantities that someone who would change the test in the future could relate to, such as the `50` microns?

Examples/Tests/particle_absorbing_boundary/inputs_full_size
EZoni, line 1: I'm a little surprised that our check
https://github.com/BLAST-WarpX/warpx/blob/843bd1796bd381e89ecb24cf85e8ef2c815c65b7/.github/workflows/source.yml#L31-L32
does not complain that this input file is not tested and/or does not follow the required naming convention. Do you understand why?

atmyers: `check_inputs.py` looks like it only complains if the filename contains certain strings, like `test_1d_`.

EZoni: Hm, that's probably not super optimal, because the goal of the check was to avoid having an input file that gets outdated without us noticing. 

I think this full-size input file will run the same risk, because the input syntax will eventually evolve and we will have to remember to update a file that does not cause any problems in any automated pipeline.

Not sure what @ax3l and @RemiLehe think, but I though we usually avoid including such files, unless I'm mistaken. What is the purpose of including the full-size input file per se? We could also simply link it in this PR, if its aim is reproducibility of the feature testing process followed in this PR or along those lines.

ax3l: Ah cool!

What we usually do is either:
- add inline comments on the full, converged size for select parameters
- create a new file that imports the first and sets the full-size values

The second bullet still have the issue of not checking the file if it gets outdated, but at least has less parameters not checked by CI.

atmyers: Ah okay. I combined added the full-size parameters to the tested file as comments.

Examples/Tests/particle_absorbing_boundary/analysis.py
EZoni, line 1: I think we could apply this patch to read some of the parameters directly from `warpx_used_inputs` rather than duplicating them manually in the analysis script:

```diff
diff --git a/Examples/Tests/particle_absorbing_boundary/analysis.py b/Examples/Tests/particle_absorbing_boundary/analysis.py
index f52c9255d..aa500d01a 100755
--- a/Examples/Tests/particle_absorbing_boundary/analysis.py
+++ b/Examples/Tests/particle_absorbing_boundary/analysis.py
@@ -9,26 +9,30 @@
 #
 # This script analyzes the phase-space plot from the 1D particle absorbing boundary test, ensuring that there are not too many fast-moving particles with negative velocities near the boundary
 
+import sys
+
 import numpy as np
 from openpmd_viewer import OpenPMDTimeSeries
 
+sys.path.append("../../../Tools/Parser/")
+from input_file_parser import parse_input_file
+
 ts = OpenPMDTimeSeries("./diags/reducedfiles/PhaseSpaceElectrons")
 it = ts.iterations
 data, info = ts.get_field(field="data", iteration=8000, plot=False)
 
 # We check the total weight of particles in the region of phase space with z
-# between 0 and 50 microns and uz between -5 and -1. If you change the bounds
-# or number of points in the PhaseSpaceElectrons diagnostic in the inputs file,
-# you need to update the calculations below.
-nz = 1000
-zmin = -100
-zmax = 50
-nuz = 1000
-uzmin = -20
-uzmax = 40
-
-reg_lo_z = 0
-reg_hi_z = 50
+# between 0 and 50 microns and uz between -5 and -1.
+input_dict = parse_input_file("./warpx_used_inputs")
+nz   = int(input_dict["PhaseSpaceElectrons.bin_number_abs"][0])
+zmin = float(input_dict["PhaseSpaceElectrons.bin_min_abs"][0])
+zmax = float(input_dict["PhaseSpaceElectrons.bin_max_abs"][0])
+nuz  = int(input_dict["PhaseSpaceElectrons.bin_number_ord"][0])
+uzmin = float(input_dict["PhaseSpaceElectrons.bin_min_ord"][0])
+uzmax = float(input_dict["PhaseSpaceElectrons.bin_max_ord"][0])
+
+reg_lo_z = 0.0
+reg_hi_z = 50.e-6
 reg_lo_uz = -5
 reg_hi_uz = -1
```

Note the additional `1e-6` factor which I think should be applied for consistency with the input file. Is that right?

I wasn't able to test locally because the test fails for me due to some openPMD backend issue.

EZoni: Applied in [90250ac](https://github.com/BLAST-WarpX/warpx/pull/6503/commits/90250acb33122d70bc435ede88af64dfc47a3100).

Examples/Tests/particle_absorbing_boundary/inputs_full_size
EZoni: Applied in [aa0e3a1](https://github.com/BLAST-WarpX/warpx/pull/6503/commits/aa0e3a119bd742bdaae50a3a6204efaaf03b573a).

Examples/Tests/particle_absorbing_boundary/analysis.py
EZoni: Applied in [175bb1d](https://github.com/BLAST-WarpX/warpx/pull/6503/commits/175bb1d2d7c4016316449c57f5f262d141818b53).

EZoni (approved):

ax3l (approved):

PR #6510, opened 2026-01-21
Docs: inputs section cleanup - parser and constants

EZoni:
Move and refine paragraph about math parser and constants. Split from #6355.

PR #6511, opened 2026-01-22
Docs: inputs section cleanup - #6510 follow-up

EZoni:
Follow up on #6510 and fix a few issues:
- ~Fix the names of the constants to match the names in https://github.com/BLAST-WarpX/warpx/blob/development/Source/ablastr/constant.H.~
- Annotate the units of all dimensional constants (not just some of them).
- Fix broken references to the renamed `Time intervals` subsection.
- Write "pi" as $\pi$ in the table.

<p align="center">
  Screenshot from 2026-01-26 16-45-44
</p>


Docs/source/usage/parameters.rst
```diff
@@ -32,17 +32,17 @@ WarpX constants
...
+``q_e``         elementary charge (C)
+``m_e``         electron mass (kg)
+``m_p``         proton mass (kg)
+``m_u``         unified atomic mass unit (kg)
+``epsilon_0``   vacuum permittivity (F/m)
+``mu0``         vacuum permeability (H/m)
+``c``           vacuum speed of light (m/s)
+``kb``          Boltzmann's constant (J/K)
+``pi``          mathematical constant :math:`\pi`
+==============  ==================================
```
EZoni: @aeriforme 

Could you please double-check that I fixed the names correctly? I'm surprised `epsilon0` and `clight` were wrong, maybe I'm missing something.

EZoni: Just for reference, this is the new table rendered:

Screenshot from 2026-01-21 17-29-57

aeriforme: They are not wrong. 
The names exposed to the parser have not been modified, [see this](https://github.com/BLAST-WarpX/warpx/blob/development/Source/Initialization/WarpXAMReXInit.cpp#L146).

We could add the new names on top of the old ones, just to avoid breaking everybody's input files.
We should also add at least `hbar` to the list.

Docs/source/usage/parameters.rst
```diff
@@ -32,17 +32,17 @@ WarpX constants
...
-``kb``       Boltzmann's constant (J/K)
-``pi``       math constant pi
-============ ===================
+==============  ==================================
+``q_e``         elementary charge (C)
+``m_e``         electron mass (kg)
+``m_p``         proton mass (kg)
+``m_u``         unified atomic mass unit (kg)
+``epsilon_0``   vacuum permittivity (F/m)
+``mu0``         vacuum permeability (H/m)
```
aeriforme:
```suggestion
``mu_0``         vacuum permeability (H/m)
```

Docs/source/usage/parameters.rst
EZoni: Ah okay, I knew I was missing something. This PR is only about the documentation, so I would leave source code changes aside. I will revert the changes for `epsilon0` and `clight`. This said, I do think that it is not super developer-friendly to name the same constant differently in the frontend and the backend.

Docs/source/usage/parameters.rst
EZoni: From the link you sent above it seems that `mu0` was already correct though, no?

https://github.com/BLAST-WarpX/warpx/blob/6654828c50e1b0d434cd44592aebd9725b69bd15/Source/Initialization/WarpXAMReXInit.cpp#L146-L169

lucafedeli88 (approved):
Thanks for this cleanup, @EZoni .
I agree with you and @aeriforme that allowing also epsilon_0 would be nice (but in a different PR).

Docs/source/usage/parameters.rst
aeriforme: Ah yes, yes, that would have been to be consistent with your proposed changes.
As discuss, we shall leave that for another PR.

aeriforme (approved):

PR #6514, opened 2026-01-22
Adding theta implicit hybrid pic method

prkkumar-he:
This PR implements theta-implicit hybrid-PIC evolve scheme. This is a fully implicit algorithm where both the fields and particles are treated implicitly.

 
 The magnetic field evolves via Faraday's law:
` Bg^{n+1} = Bg^n - dt*curl(Eg^{n+theta})`
 
The electric field is determined algebraically from the generalized Ohm's law.
 
 Particles are advanced in the same way as the theta implicit EM scheme
 
 *  The key difference from fully electromagnetic implicit solvers is that E is computed from Ohm's law at each
 nonlinear iteration. The nonlinear solver finds the self-consistent `E^{n+theta}`
 such that:
```
E^{n+theta} = Ohm(B^{n+theta}, Ji^{n+1/2}, rho^{n+1/2})
where B^{n+theta} depends on E^{n+theta} through Faraday's law.
```
 

Test problem based on the explicit EM mode test case (https://github.com/BLAST-WarpX/warpx/blob/development/Examples/Tests/ohm_solver_em_modes/inputs_test_1d_ohm_solver_em_modes_picmi.py)

1D EM mode test case with Newton solver with GMRES without preconditioning using  (Examples/Tests/ohm_solver_em_modes/implicit/inputs_test_1d_ohm_solver_em_modes_implicit_picmi.py) is shown below.

image

Perpendicular EM modes:

image

Cylindrical EM mode:
image

Magnetic Reconnection test case works as well, but is not included in this PR. 

https://github.com/user-attachments/assets/863dbe65-358e-4ed5-8e7b-c53ef4396fce

Support for time varying external vector potential will be included in an upcoming PR.


prkkumar-he:
Increased dt progressively from `2e-11` to `5e-10`, it still converges to the correct results. Nonlinear solver takes 8-10 iterations for dt = 5e-10 (vs 3 iterations for dt = 2e-11s). Explicit evolve scheme becomes unstable for this dt.


tried `jacobian.pc_type = none` and `jacobian.pc_type = pc_jacobi`. Both give similar convergence

roelof-groenewald:
Awesome PR! Thanks @prkkumar-he!
This is a really cool example of the benefit of modular functionality in WarpX where pieces can be assembled in different ways to add a very useful new model without code duplication.

Source/FieldSolver/ImplicitSolvers/ThetaImplicitHybrid.H
```diff
@@ -0,0 +1,127 @@
...
+ *  Particles are advanced as:
+ *  xp^{n+1} = xp^n + dt*up^{n+1/2}/(0.5*(gammap^n + gammap^{n+1}))
+ *  up^{n+1} = up^n + dt*qp/mp*(Ep^{n+theta} + up^{n+1/2}/gammap^{n+1/2} x Bp^{n+theta})
+ *
+ *  where f^{n+theta} = (1.0-theta)*f^{n} + theta*f^{n+1} with 0.5 <= theta <= 1.0
+ *
+ *  The key difference from fully electromagnetic implicit solvers is that E is computed from Ohm's law at each
+ *  nonlinear iteration. The nonlinear solver finds the self-consistent E^{n+theta}
+ *  such that:
+ *     E^{n+theta} = Ohm(B^{n+theta}, Ji^{n+1/2}, rho^{n+1/2})
```
JustinRayAngus: Would be good to explicitly write the ohms law here.

prkkumar-he: Thanks! I updated the docstring

clarkse-he:
Heads-up from a local rebase of this PR onto current `development` (b3e232fd6): there is an interaction with #6589 (merged 2026-03-06, after this PR's last update) that makes the RZ test fail hard on rebase.

**Mechanism.** Since #6589, `WarpX::PushParticlesandDeposit` skips the radial-geometry inverse-volume scaling block whenever it is called with `implicit_options != nullptr` (`Source/Evolve/WarpXEvolve.cpp`, the `if (!skip_deposition && !implicit_options)` gate). `ImplicitSolver::PreRHSOp` compensates — but **for the current density only**. Charge density deposited during implicit residual evaluations is therefore never volume-scaled in RZ/RCYLINDER/RSPHERE. The merged EM implicit solvers never read `rho` in their residuals, so this stayed invisible. This PR's Ohm's law divides by `rho`, so in RZ the first residual evaluation is wrong by an r-dependent ring-volume factor and Newton diverges immediately:

```
Newton: iteration =   0, norm = 1.44374e+06 (abs.), 1.00000e+00 (rel.)
Newton: iteration =   5, norm = 1.70077e+08 (abs.), 1.17803e+02 (rel.)
Newton: exiting at iteration =   5. SOLVER DIVERGED!
```

(`test_rz_ohm_solver_em_modes_implicit_picmi`, step 0; the 1D test is unaffected since there is no volume scaling in 1D.)

**Fix that restores the pre-#6589 semantics** (scale once per residual evaluation, after `CumulateJ`, alongside J) in `ImplicitSolver::PreRHSOp`:

```cpp
#if defined(WARPX_DIM_RZ) || defined(WARPX_DIM_RCYLINDER) || defined(WARPX_DIM_RSPHERE)
    for (int lev = 0; lev < m_num_amr_levels; ++lev) {
        ablastr::fields::VectorField J = m_WarpX->m_fields.get_alldirs(FieldType::current_fp, lev);
        m_WarpX->ApplyInverseVolumeScalingToCurrentDensity(J[0], J[1], J[2], lev);
        if (m_WarpX->m_fields.has(FieldType::rho_fp, lev)) {
            m_WarpX->ApplyInverseVolumeScalingToChargeDensity(
                m_WarpX->m_fields.get(FieldType::rho_fp, lev), lev);
        }
    }
#endif
```

The component/mode decoder in `ApplyInverseVolumeScalingToChargeDensity` handles the two time-level components of the hybrid `rho_fp` correctly, so one call scales both.

With this patch the RZ run converges and completes. Note that on rebase the RZ checksums and the pinned amplitude samples in `analysis_rz.py` need regeneration regardless: the RZ state has drifted ~3e-4 (final ion momenta) relative to the February baselines from post-February RZ changes (e.g. the `HybridPICSolveE` origin/axis fix in #6598), which reshuffles the near-noise spectral bins the analysis pins. The 1D baselines reproduce exactly.

Since unscaled `rho` also affects RZ rho output for the EM implicit solvers (diagnostics), the fix is arguably worth a small standalone PR against `development` independent of this one — happy to open it.

JustinRayAngus:
@clarkse-he re- Since unscaled rho also affects RZ rho output for the EM implicit solvers (diagnostics), the fix is arguably worth a small standalone PR against development independent of this one — happy to open it.

Yes, I'm aware of this issue. Since it doesn't affect the EM solvers, we kindly left it for you guys to fix :). Please open a separate PR.

clarkse-he:
Update: opened #7050 with the standalone fix for the rho volume-scaling issue described above (go-ahead from our side to split it out). Once #7050 merges, rebasing this PR onto development resolves the RZ divergence — the RZ test here then serves as its regression coverage. cc @prkkumar-he

PR #6515, opened 2026-01-23
[WIP] Implement mcc_in_mid_x_push with 2 depositions

oshapoval:
This PR presents the `mid-x-push` with 2 depositions implementation (collisions applied in the middle of the particle position push) with a new interface implemented in PR:  https://github.com/BLAST-WarpX/warpx/pull/6275 .

`mid-x-push` algorithm  is described in [DOI: 10.1103/PhysRevE.111.02530](https://doi.org/10.1103/PhysRevE.111.025306).


EZoni:
If this is built on #6275, ideally we would merge #6275 first and then rebase here to expose the code changes more easily and keep infrastructure changes separate from algorithm changes, etc.

EZoni:
Also, @oshapoval or @RemiLehe, could you comment here on the double charge/counting issue, so that it stays in the history? Based on the offline discussions we had so far, I haven't fully understood yet why there is an issue with double charge/counting.

oshapoval:
> Also, @oshapoval or @RemiLehe, could you comment here on the double charge/counting issue, so that it stays in the history? Based on the offline discussions we had so far, I haven't fully understood yet why there is an issue with double charge/counting.

@EZoni We need to divide by 2 to avoid double counting of `J`: when we deposit `Q` over `dt`, the current is proportional to `~ W_{ijk} Q / dt`. If we deposit twice over `dt/2` with the same `Q` and sum the contributions, we end up with `~ (W^{(1)}_{ijk} 2Q/dt + W^{(2)}_{ijk} 2Q/dt) = 2Q/dt (W^{(1)}_{ijk} + W^{(2)}_{ijk})`.

PR #6518, opened 2026-01-24
Change mass of pre-defined elements in `species_type`: subtract mass of the electrons

RemiLehe:
In WarpX, when species are initialized with the name of an element as `species_type` (e.g. `species_type=oxygen`), the species will correspond to fully ionized ions. (For example, a species initialized with `oxygen` will have charge 8+.) 

For consistency, the mass of that species should be the mass of the nucleus **only**. This PR accomplishes this by systematically subtracting the mass of the electrons to the pre-defined mass of various elements.

As a consequence, `alpha` and `helium4` are now equivalent. In addition, `protium` was removed as it is somewhat of an outlier.

In addition, this PR also includes changes made by @JustinRayAngus in `ParticleCreationFunc.H` (as part of #6519) to automatically compute the fusion energy yield from the (now accurate) mass of the reactants:
```
                    const amrex::ParticleReal mass_before = m1 + m2;
                    const amrex::ParticleReal mass_after = products_mass_data[0] + products_mass_data[1];
                    const amrex::ParticleReal fusion_energy = (mass_before - mass_after)*PhysConst::c2;
```

PR #6519, opened 2026-01-24
fusion reaction modifications

JustinRayAngus:
Currently, the product species of D + D => T + p fusion are required to be hydrogen3 and hydrogen1. hydrogen3 is the same as tritium, but hydrogen1 is not a proton. This PR updates the product species check to include a proton rather than hydrogen1. 

This PR also simplifies the hard-wired fusion energy of the reaction to be the difference in mass of the initial and product species. When all species are ions with mass computed as M = A*m_u - Z*m_e with the atomic mass numbers A given in SpeciesPhysicalProperties.cpp, this results in the following reaction energies:
```
D + T   => He4 + n: Q = 17.58929696 MeV
D + D   => T + p:   Q = 4.032653948 MeV
D + D   => He3 + n: Q = 3.26891111 MeV
D + He3 => He4 + p: Q = 18.35303980 MeV
p + B11 => 3He4:    Q = 8.68212502 MeV
```

- [x] Add check that fusion energy computed based on mass difference is close to expected value.

This PR also fixes a bug where the Particles from a D + He3 => He4 + p were never created.

### DIFF
--- a/Examples/Tests/nuclear_fusion/analysis_two_product_fusion.py
+++ b/Examples/Tests/nuclear_fusion/analysis_two_product_fusion.py
@@ -65,7 +65,7 @@ if re.search("tritium", warpx_used_inputs):
     reactant_species = ["deuterium", "tritium"]
     product_species = ["helium4", "neutron"]
     ntests = 2
-    E_fusion = 17.58929696 * MeV_to_Joule  # Energy released during the fusion reaction
+    E_fusion = 17.58929696 * MeV_to_Joule
 else:
     # else, this is the D+D test
     reaction_type = "DD"
--- a/Examples/Tests/nuclear_fusion/analysis_two_product_fusion.py
+++ b/Examples/Tests/nuclear_fusion/analysis_two_product_fusion.py
@@ -65,21 +65,21 @@ if re.search("tritium", warpx_used_inputs):
     reactant_species = ["deuterium", "tritium"]
     product_species = ["helium4", "neutron"]
     ntests = 2
-    E_fusion = 17.5893 * MeV_to_Joule  # Energy released during the fusion reaction
+    E_fusion = 17.58929696 * MeV_to_Joule  # Energy released during the fusion reaction
 else:
     # else, this is the D+D test
     reaction_type = "DD"
     reactant_species = ["deuterium", "hydrogen2"]
     product_species = ["helium3", "neutron"]
     ntests = 1
-    E_fusion = 3.268911e6 * MeV_to_Joule
+    E_fusion = 3.26891111e6 * MeV_to_Joule
 
 mass = {
-    "deuterium": 2.01410177812 * scc.m_u,
-    "hydrogen2": 2.01410177812 * scc.m_u,
-    "tritium": 3.0160492779 * scc.m_u,
-    "helium3": 3.016029 * scc.m_u,
-    "helium4": 4.00260325413 * scc.m_u,
+    "deuterium": 2.01410177812 * scc.m_u - scc.m_e,
+    "hydrogen2": 2.01410177812 * scc.m_u - scc.m_e,
+    "tritium": 3.0160492779 * scc.m_u - scc.m_e,
+    "helium3": 3.016029 * scc.m_u - 2*scc.m_e,
+    "helium4": 4.00260325413 * scc.m_u - 2*scc.m_e,
     "neutron": 1.0013784193052508 * scc.m_p,
 }
 m_reduced = np.prod([mass[s] for s in reactant_species]) / np.sum(
--- a/Examples/Tests/nuclear_fusion/analysis_two_product_fusion.py
+++ b/Examples/Tests/nuclear_fusion/analysis_two_product_fusion.py
@@ -78,8 +78,8 @@ mass = {
     "deuterium": 2.01410177812 * scc.m_u - scc.m_e,
     "hydrogen2": 2.01410177812 * scc.m_u - scc.m_e,
     "tritium": 3.0160492779 * scc.m_u - scc.m_e,
-    "helium3": 3.016029 * scc.m_u - 2*scc.m_e,
-    "helium4": 4.00260325413 * scc.m_u - 2*scc.m_e,
+    "helium3": 3.016029 * scc.m_u - 2 * scc.m_e,
+    "helium4": 4.00260325413 * scc.m_u - 2 * scc.m_e,
     "neutron": 1.0013784193052508 * scc.m_p,
 }
 m_reduced = np.prod([mass[s] for s in reactant_species]) / np.sum(
--- a/Examples/Tests/nuclear_fusion/inputs_test_3d_deuterium_deuterium_fusion
+++ b/Examples/Tests/nuclear_fusion/inputs_test_3d_deuterium_deuterium_fusion
@@ -30,11 +30,14 @@ algo.particle_shape = 1
 ############ PLASMA #############
 #################################
 particles.species_names = deuterium_1 hydrogen2_1 helium3_1 neutron_1
-my_constants.m_deuterium = 2.01410177812*m_u
+my_constants.m_deuterium = 2.01410177812*m_u - m_e
+my_constants.m_helium3 = 3.01602932243*m_u - 2*m_e
 my_constants.keV_to_J = 1.e3*q_e
 my_constants.Energy_step = 22. * keV_to_J
 
 deuterium_1.species_type = deuterium
+deuterium_1.mass = m_deuterium
+deuterium_1.charge = q_e
 deuterium_1.injection_style = "NRandomPerCell"
 deuterium_1.num_particles_per_cell = 10000
 deuterium_1.profile = constant
@@ -47,6 +50,8 @@ deuterium_1.do_not_push = 1
 deuterium_1.do_not_deposit = 1
 
 hydrogen2_1.species_type = deuterium
+hydrogen2_1.mass = m_deuterium
+hydrogen2_1.charge = q_e
 hydrogen2_1.injection_style = "NRandomPerCell"
 hydrogen2_1.num_particles_per_cell = 10000
 hydrogen2_1.profile = constant
@@ -59,6 +64,8 @@ hydrogen2_1.do_not_push = 1
 hydrogen2_1.do_not_deposit = 1
 
 helium3_1.species_type = helium3
+helium3_1.mass = m_helium3
+helium3_1.charge = 2*q_e
 helium3_1.do_not_push = 1
 helium3_1.do_not_deposit = 1
 
--- a/Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H
+++ b/Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H
@@ -167,6 +167,7 @@ public:
         const int t_num_product_species = m_num_product_species;
         const int* AMREX_RESTRICT p_num_products_device = m_num_products_device.data();
         const CollisionType t_collision_type = m_collision_type;
+        const amrex::ParticleReal t_reaction_energy = m_reaction_energy;
 
         amrex::ParallelForRNG(n_total_pairs,
         [=] AMREX_GPU_DEVICE (int i, amrex::RandomEngine const& engine) noexcept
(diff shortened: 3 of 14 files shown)


EZoni:
@JustinRayAngus @RemiLehe 

1. What is the status of this PR?
2. What is its relation with #6518, if any?
3. Should we finalize this/them before we make further changes to the fusion modules?

JustinRayAngus:
@EZoni 

There are only two changes made in this PR. One of the changes was to permit using a proton (rather than hydrogen1) for the product the `D + D ==> T + p` fusion reaction. It looks like @RemiLehe has this change also present in PR #6518. That seems like a bigger PR and may take more time before it can be merged.

The second change in this PR is to compute the fusion Q based on the difference in the mass of the initial and product species.  Right now hydrogen1 is a neutral species and the mass is not the same as a proton. This means that the hard-coded fusion reaction Q is inconsistent with the change in mass of the reaction. In general, since the user has the ability to change the mass of species of a specified type, I think this is a safer practice.

One additional change could be to save the fusion Q as a member variable rather than recomputing it each time the collision module is called. This would also allow one to implement an assert that (or warning if) the fusion Q associated with the mass difference is within some tolerance of the expected fusion Q for the reactions based on the exact known species mass, which I have listed in the PR description.

Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H
```diff
@@ -210,16 +210,9 @@ public:
...
-                    }
-                    else if (t_collision_type == CollisionType::DeuteriumDeuteriumToProtonTritiumFusion) {
-                        fusion_energy = 4.032667e6_prt * PhysConst::q_e;
-                    }
-                    else if (t_collision_type == CollisionType::DeuteriumDeuteriumToNeutronHeliumFusion) {
-                        fusion_energy = 3.268911e6_prt * PhysConst::q_e;
-                    }
+                    const amrex::ParticleReal mass_before = m1 + m2;
+                    const amrex::ParticleReal mass_after = products_mass_data[0] + products_mass_data[1];
+                    const amrex::ParticleReal fusion_energy = (mass_before - mass_after)*PhysConst::c*PhysConst::c;
```
EZoni: I think we could use `c2` directly, if you want:

```suggestion
                    const amrex::ParticleReal fusion_energy = (mass_before - mass_after)*PhysConst::c2;
```

Reference:
https://github.com/BLAST-WarpX/warpx/blob/5fc99bb8014dd4a9def9902c9286c62e11989e6b/Source/ablastr/constant.H#L213-L214

Examples/Tests/nuclear_fusion/analysis_two_product_fusion.py
```diff
@@ -65,21 +65,21 @@
     reactant_species = ["deuterium", "tritium"]
     product_species = ["helium4", "neutron"]
     ntests = 2
-    E_fusion = 17.5893 * MeV_to_Joule  # Energy released during the fusion reaction
+    E_fusion = 17.58929696 * MeV_to_Joule  # Energy released during the fusion reaction
```
EZoni: I feel like the inline comment here is a bit redundant. I'm not opposed to removing it, but it was there before, so I'm not opposed to keeping it either. If we keep it, maybe we could add it to line 75 as well, for consistency.

Examples/Tests/nuclear_fusion/inputs_test_3d_deuterium_deuterium_fusion
```diff
@@ -30,11 +30,14 @@ algo.particle_shape = 1
 ############ PLASMA #############
 #################################
 particles.species_names = deuterium_1 hydrogen2_1 helium3_1 neutron_1
-my_constants.m_deuterium = 2.01410177812*m_u
+my_constants.m_deuterium = 2.01410177812*m_u - m_e
```
EZoni, line 33: I may have missed parts of the discussion on this topic over time, but would it be useful to eventually store some of these atomic masses (e.g., 2.01410177812 for Deuterium) as constants accessible directly from WarpX, for the case where users do not want to modify the mass of a specified species type? As opposed to copying and pasting the mass in every input file where it's used.

EZoni (commented):
I left a couple of comments or questions, but this looks good to me overall. 

@RemiLehe 

Would you have objections against merging this before #6518? Or, what was your opinion on how the two PRs should be coordinated?

Examples/Tests/nuclear_fusion/inputs_test_3d_deuterium_deuterium_fusion
JustinRayAngus: This would not be needed after Remi's PR is merged where the electron mass is subtracted off by default.

Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H
EZoni: FYI, I just opened #6615 to apply this change uniformly in the code.

Examples/Tests/nuclear_fusion/inputs_test_3d_deuterium_deuterium_fusion
EZoni: Ah okay, understood. That makes sense.

EZoni (approved):

JustinRayAngus:
@EZoni @RemiLehe 
It seems that several of the fusion reactions are treated differently than the DD and DT channels. We should discuss. Converting this PR to a WIP.

JustinRayAngus:
@EZoni @RemiLehe 

I committed the changes to make the fusion energy a member variable,  allowing for direct check of the fusion energy with what is expected based on the species mass. I've added a m_reaction_energy member to the ParticleCreationFunc class. This value is only computed and used for the fusion reactions, but in principle it could also be computed and used by other reactions as well.

I also added the missing check for `CollisionType::DeuteriumHeliumToProtonHeliumFusion` (was missing in 2 places.) I haven't tried it to make sure it actually works though.

This took a bit more work than I had anticipated. Happy to discuss if it will help move things along.

RemiLehe:
Thanks for updating the PR @JustinRayAngus 
However, it looks like the tests are now failing with:
```
  what():  Assertion `energy_error < energy_tolerance' failed, file "/home/vsts/work/1/s/Source/Particles/Collision/BinaryCollision/BinaryCollisionUtils.cpp", line 235, Msg: 
### ERROR   : ERROR: D + T -> He4 + n fusion energy error is too large. Make
#            sure species mass are set properly.

SIGABRT
  what():  Assertion `energy_error < energy_tolerance' failed, file "/home/vsts/work/1/s/Source/Particles/Collision/BinaryCollision/BinaryCollisionUtils.cpp", line 235, Msg: 
### ERROR   : ERROR: D + T -> He4 + n fusion energy error is too large. Make
#            sure species mass are set properly.

SIGABRT

```

JustinRayAngus:
> Thanks for updating the PR @JustinRayAngus However, it looks like the tests are now failing with:
> 
> ```
> terminate called after throwing an instance of 'std::runtime_error'
>   what():  Assertion `energy_error < energy_tolerance' failed, file "/home/vsts/work/1/s/Source/Particles/Collision/BinaryCollision/BinaryCollisionUtils.cpp", line 235, Msg: 
> ### ERROR   : ERROR: D + T -> He4 + n fusion energy error is too large. Make
> #            sure species mass are set properly.
> 
> SIGABRT
>   what():  Assertion `energy_error < energy_tolerance' failed, file "/home/vsts/work/1/s/Source/Particles/Collision/BinaryCollision/BinaryCollisionUtils.cpp", line 235, Msg: 
> ### ERROR   : ERROR: D + T -> He4 + n fusion energy error is too large. Make
> #            sure species mass are set properly.
> 
> SIGABRT
> ```

That's what I get for not testing it myself before pushing.

Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H
```diff
@@ -167,6 +167,7 @@ public:
         const int t_num_product_species = m_num_product_species;
         const int* AMREX_RESTRICT p_num_products_device = m_num_products_device.data();
         const CollisionType t_collision_type = m_collision_type;
+        const amrex::ParticleReal t_reaction_energy = m_reaction_energy;
```
RemiLehe, line 170: The `t_` in the `t_num_product_species` and `t_collision_type` is because these are used as template parameters. I don't think that this is the case of `reaction_energy`.
```suggestion
        const amrex::ParticleReal reaction_energy = m_reaction_energy;
```

Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H
```diff
@@ -208,24 +209,15 @@ public:
...
-                    else if (t_collision_type == CollisionType::DeuteriumDeuteriumToNeutronHeliumFusion) {
-                        fusion_energy = 3.268911e6_prt * PhysConst::q_e;
-                    }
                     TwoProductFusionInitializeMomentum(soa_1, soa_2,
                         soa_products_data[0], soa_products_data[1],
                         p_pair_indices_1[i], p_pair_indices_2[i],
                         products_np_data[0] + 2*p_offsets[i]*p_num_products_device[0],
                         products_np_data[1] + 2*p_offsets[i]*p_num_products_device[1],
-                        m1, m2, products_mass_data[0], products_mass_data[1], fusion_energy, engine);
+                        m1, m2, products_mass_data[0], products_mass_data[1], t_reaction_energy, engine);
```
RemiLehe, line 220:
```suggestion
                        m1, m2, products_mass_data[0], products_mass_data[1], reaction_energy, engine);
```

Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H
RemiLehe, line 208: It seems that `m_reaction_energy` is calculated for the proton-boron fusion too, but not used.

Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H
```diff
@@ -268,6 +260,7 @@ private:
     // for device) which is necessary with GPUs but redundant on CPU.
     amrex::Gpu::DeviceVector<int> m_num_products_device;
     amrex::Gpu::HostVector<int> m_num_products_host;
+    amrex::ParticleReal m_reaction_energy = 0.0;
```
RemiLehe, line 263: The fact that it is stored in this general class, but then only used for fusion, is I think potentially misleading. (Other reactions like impact ionization in DSMC and linear Breit-Wheeler also have a non-zero reaction energy, but that energy is stored somewhere else.) Maybe we should discuss this a bit more.

RemiLehe (commented):
Thanks for this PR @JustinRayAngus I think that some of the latest changes on this PR require some discussion as mentioned above. 

On the other hand, I think that https://github.com/BLAST-WarpX/warpx/pull/6518 is now ready to go, and includes some of the key changes of #6519 (before the additional changes that require more discussion). 

While I originally suggested to merge #6519 first and then #6518, I now think that we should maybe do it the other way around. What do you think @JustinRayAngus? Would you be OK with merging #6518 now?

JustinRayAngus:
> Thanks for this PR @JustinRayAngus I think that some of the latest changes on this PR require some discussion as mentioned above. 
> 
> On the other hand, I think that https://github.com/BLAST-WarpX/warpx/pull/6518 is now ready to go, and includes some of the key changes of #6519 (before the additional changes that require more discussion). 
> 
> While I originally suggested to merge #6519 first and then #6518, I now think that we should maybe do it the other way around. What do you think @JustinRayAngus? Would you be OK with merging #6518 now?

Do you want me to literally merge #6518 now?  I'm fine doing so.

Source/Particles/Collision/BinaryCollision/ParticleCreationFunc.H
JustinRayAngus: I agree.

The logical place for it to live is in NuclearCollisionFunc, but the problem is that members of an instance of this class are not accessible from where the new particles are created in ParticleCreationFunc. This is probably why the reaction energies were hard-wired inside there in the first place (which was error prone).

I don't think I understand the code design for handling particle scattering/creation. It doesn't seem like either parts have access to the member variables of the collision class. I understand the desire to limit duplicate code, but different collision types require different things.

Any suggestions here? Perhaps there is a simple way to let m_fusion_energy live in NuclearCollisionFunc and to still have it accessible where needed that I'm not seeing.

Happy to discuss offline if needed.

JustinRayAngus:
@EZoni @RemiLehe 

Some of this PR has been implemented in PR #6518.

The rest of this PR has been split into two separate PRs: #6623 and #6624.

Once those are merged, I will delete this PR.

PR #6520, opened 2026-01-24
Mass matrices pc width part 1

JustinRayAngus:
Add off-diagonal terms of the diagonal mass matrices to the preconditioner.


dpgrote (approved):
This looks good to me.

PR #6526, opened 2026-01-27
bug fix for MM deposition in 1D.

JustinRayAngus:
This PR fixes a bug in the 1D Villasenor deposition routine for mass matrices. The number of indices for the mass matrices currently has only three values specified, with the intended component number the third value. This causes crash with amrex.fpe_trap_invalid = 1 because the code is interpreting the component number as the third spatial dimension index and setting the component number to 0. Using debug build, I get the following error message:
amrex::Abort::0:: (-1,0,3,0) is out of bound (-2:10,0:0,0:0,0:6) !!!

This error prevented me from running the implicit/inputs_test_1d_theta_implicit_pinch test locally. I have no idea why the crash didn't occur when running the CI tests on azure.

With or without this fix, if I set amrex.fpe_trap_invalid = 0, the CI test runs fine and the results look as expected.


dpgrote (approved):
Looks good

PR #6528, opened 2026-01-27
Docs: Remove Summit & Lassen (ppc64le / CORAL aquisition)

lucafedeli88:
Hi,

I've noticed that our documentation contains installation instructions for machines that have been decommissioned, in some cases more than one year ago.

This PR removes in particular the installation instructions and the machine files for the Summit supercomputer, since the machine exhaled its last FLOP in November 2024. 

A valid alternative would be to keep these files in a `legacy` folder. 

What I suggest is basically to avoid cluttering the documentation by mixing old and new machines. What do you think?


EZoni:
Thanks, @lucafedeli88! Personally, I agree that the documentation about HPC clusters that can no longer be used should not be exposed to users. On the other hand, given the amount of valuable know-how that goes into setting up WarpX at HPC clusters (in the forms of troubleshooting, IT tickets, etc.), I would think that it may be worth keeping those instructions in a `legacy` directory that is available  to developers for future reference but not exposed to users, as you suggested. This could come in handy in situations where a new HPC cluster shares common features with an old one, and similar.

ax3l:
I agree, luckily all PPC64le machines we had are now decommissioned (Lassen/Summit), so we can remove them for good for now. Rest is in RTD and git history if they resurface.

Because of the major changes (PPC64 -> AMD64, Nvidia -> AMD, GPU -> APU), there is no direct transition/ticket path between Summit and Frontier, Lassen and Tuo, so we can remove them for good.

ax3l (approved):
Time toooooo, say goooodbyyyeeee :sob: 
:fireworks: :us: :saluting_face:

PR #6532, opened 2026-01-28
Fix warning message for mass and injection file

Noerr:
correcting copy-paste error in warning output where charge was errantly mentioned instead of mass.


RemiLehe (approved):

RemiLehe:
Thanks a lot for fixing this!

PR #6533, opened 2026-01-28
Save lost particle time

dpgrote:
When particles are saved when they are lost at the boundary, this PR adds the time at which they were lost, `timeScraped`, to the saved attributes (in addition to the already saved `deltaTimeScraped`). When running electrostatic simulations, the time step size can be variable, so only knowing the step number and delta time when scraped is not enough information to recover the time when a particle is lost.

As an aside, there is no mechanism that saves the time step as it varies (except as printed out to the terminal). A future PR that saves this would be useful.


JustinRayAngus:
Would this make `deltaTimeScraped` obsolete?

dpgrote:
> Would this make `deltaTimeScraped` obsolete?

Yes, mostly, but there are several CI cases that use `deltaTimeScraped` and I don't know if there are users that now depend on it. Note though that with a variable time step, the `deltaTimeScraped` could not be recovered from `timeScraped` (without knowing the time step sizes).

JustinRayAngus (approved):

RemiLehe (approved):

PR #6539, opened 2026-01-30
VirtualPhotonCreation.H: add checks for qed_virtual_photons_min_energy and qed_virtual_photons_multiplier

lucafedeli88:
Yesterday https://github.com/BLAST-WarpX/warpx/pull/6538 was merged despite @aeriforme 's comment that `qed_virtual_photons_min_energy` must be strictly positive, since it goes into a log!

This PR attempts a temporary fix by introducing checks to ensure that both `qed_virtual_photons_min_energy` and `qed_virtual_photons_multiplier` are strictly positive (I believe that also `qed_virtual_photons_multiplier` must be strictly positive, maybe even close to an integer).

As @aeriforme commented in https://github.com/BLAST-WarpX/warpx/pull/6538, a better approach would be to choose sensible default values.

### DIFF
--- a/Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
+++ b/Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
@@ -71,12 +71,19 @@ void GenerateVirtualPhotons (MultiParticleContainer* mypc){
         // Minimum allowed energy of the virtual photons
         amrex::Real vphoton_min_energy = 0.0_rt;
         utils::parser::getWithParser(pp_species_name, "qed_virtual_photons_min_energy", vphoton_min_energy);
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+            vphoton_min_energy > 0.0_rt,
+            "qed_virtual_photons_min_energy must be strictly positive.");
 
         // Sampling factor (a.k.a. multiplier):
         // the number of virtual photons generated is multiplied by this factor,
         // the weight of each virtual photon is divided by this factor
         amrex::Real sampling_factor = 0.0_rt;
         utils::parser::getWithParser(pp_species_name, "qed_virtual_photons_multiplier", sampling_factor);
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+            sampling_factor > 0.0_rt,
+            "qed_virtual_photons_multiplier must be strictly positive.");
+
 
         amrex::Real const alpha_over_pi = PhysConst::alpha / MathConst::pi;
         amrex::Real const inv_c2 = 1._rt / (PhysConst::c * PhysConst::c);


Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -71,12 +71,19 @@ void GenerateVirtualPhotons (MultiParticleContainer* mypc){
...
         amrex::Real vphoton_min_energy = 0.0_rt;
         utils::parser::getWithParser(pp_species_name, "qed_virtual_photons_min_energy", vphoton_min_energy);
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+            vphoton_min_energy > 0.0_rt,
+            "qed_virtual_photons_min_energy must be strictly positive.");
 
         // Sampling factor (a.k.a. multiplier):
         // the number of virtual photons generated is multiplied by this factor,
         // the weight of each virtual photon is divided by this factor
         amrex::Real sampling_factor = 0.0_rt;
```
aeriforme, line 81:
```suggestion
        amrex::Real sampling_factor = 1.0_rt;
```

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -71,12 +71,19 @@ void GenerateVirtualPhotons (MultiParticleContainer* mypc){
...
         utils::parser::getWithParser(pp_species_name, "qed_virtual_photons_multiplier", sampling_factor);
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+            sampling_factor > 0.0_rt,
+            "qed_virtual_photons_multiplier must be strictly positive.");
+
```
aeriforme, line 86:
```suggestion
```

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -71,12 +71,19 @@ void GenerateVirtualPhotons (MultiParticleContainer* mypc){
         // Minimum allowed energy of the virtual photons
         amrex::Real vphoton_min_energy = 0.0_rt;
```
aeriforme, line 72:
```suggestion
        amrex::Real vphoton_min_energy = 1.e-6_rt * PhysConst::m_e * PhysConst::c2;
```

Just to avoid initializing it to zero and, immediately after, warning that it cannot be zero.

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
```diff
@@ -71,12 +71,19 @@ void GenerateVirtualPhotons (MultiParticleContainer* mypc){
...
         utils::parser::getWithParser(pp_species_name, "qed_virtual_photons_min_energy", vphoton_min_energy);
+        WARPX_ALWAYS_ASSERT_WITH_MESSAGE(
+            vphoton_min_energy > 0.0_rt,
+            "qed_virtual_photons_min_energy must be strictly positive.");
```
aeriforme, line 77: I have a question: how expensive is this? This function, if used, is called for all particles for all timesteps. I was wondering if it can become a bottleneck.

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
lucafedeli88: I think that if we use `getWithParser` (instead of `queryWithParser`) this default value will never be read (that's why I chose `0.0`). If you prefer a default value I think that we have to change the way the parameter is read. I also think that the default value should appear in the documentation.

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
lucafedeli88: Do you mean that it's OK if the `sampling_factor` is negative ?

lucafedeli88: or do you prefer not to test for performance reasons ?

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
lucafedeli88: I think that it's called only for all the species at each timestep :  it's outside the loop on boxes and particles .  This means that it should be OK for performances.

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
lucafedeli88: A default value makes sense, but I think that we have to change how we read the user input. What do you think?

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
aeriforme: I just mean that this should be checked once, not here where is checked multiple times. Not sure how to do it though, i.e. where to move this code.``

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
aeriforme: Ah interesting, I didn't know that. 
We can keep it like this, meaning no default, it won't be read and we just do it to satisfy the whining compiler.
However, if we do it, we might as well suggest a sensible number instead of a number that would break things. Or not?

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
aeriforme: I didn't know that. For this number the default is a bit trickier...Maybe we keep it like this for the same reason as above?

Source/Particles/Collision/BinaryCollision/VirtualPhotonCreation.H
aeriforme: Ah right right

PR #6545, opened 2026-02-02
Use finestLevel() and maxLevel() more consistently

lucafedeli88:
Hi! I would like to finalize a work that @tmsclark  (a.k.a. @tmsclark2 ) and I started a long time ago with this PR: https://github.com/BLAST-WarpX/warpx/pull/2948 . The PR concerns the possibility of removing the fine patch at a given time step. WarpX has changed a lot since then and we'll probably need to start from scratch (what do you think, @tmsclark2  ?).

A requirement to make https://github.com/BLAST-WarpX/warpx/pull/2948 work is to properly use `maxLevel()` and `finestLevel()` . Now, they are used interchangeably, because they always provide the same result. However, they are conceptually different : `maxLevel()` should not change during the simulation, while `finestLevel()`  could vary between 0 and `maxLevel()`  . 

In this PR I've tried replacing occurrences of   `maxLevel()` where in principle we should have used `finestLevel()` . However, since this is sometimes tricky, I would like to ask for help to check that these changes make sense. 

Once this is done, I would like to finalize the work in https://github.com/BLAST-WarpX/warpx/pull/2948 (possibly with the collaboration of @tmsclark2 if he's interested and has enough time). Finally adding the feature (as `experimental`) to the main branch of WarpX should not take too long.


tmsclark2:
Hi Luca,
I will definitely take a look and help you with this pull request. It would be helpful to have it.

tmsclark2 (approved):
The modifications look good to me.

PR #6548, opened 2026-02-04
[WIP] Fix Gaussian beam focusing when 

aeriforme:
Fix focusing of a Gaussian beam when `do_gaussian_beam_rotation = 1` but  `do_gaussian_beam_rotation_momenta = 0`. 
In this case, the desired behavior is that the bulk momentum remains unrotated, not rotating the total momentum. 

TO DO:
- [ ] change the name of the input parameter `do_gaussian_beam_rotation_momenta` to `do_gaussian_beam_rotation_bulk_momenta`
- [ ] add a CI test with beam rotation + focusing 

Thanks @pkicsiny


aeriforme:
I'm closing this because it's wrong. I went on a bug-fix-frenzy that was only in my imagination.

PR #6556, opened 2026-02-06
Hybrid PIC: change definition of number of substeps

RemiLehe:
The new definition matches the user intuition more closely: if the user sets `substeps=10`, then there will be 10 field updates (each using the RK4 algorithm) each using $\Delta t/10$ within one timestep.

(As opposed to the current behavior: 20 field updates with $\Delta t/20$.)

### DIFF
--- a/Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
+++ b/Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
@@ -32,8 +32,12 @@ void HybridPICModel::ReadParameters ()
     const ParmParse pp_hybrid("hybrid_pic_model");
 
     // The B-field update is subcycled to improve stability - the number
-    // of sub steps can be specified by the user (defaults to 50).
+    // of sub steps can be specified by the user.
     utils::parser::queryWithParser(pp_hybrid, "substeps", m_substeps);
+    if (m_substeps % 2 != 0) {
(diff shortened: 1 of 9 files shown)


Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -32,8 +32,12 @@ void HybridPICModel::ReadParameters ()
     const ParmParse pp_hybrid("hybrid_pic_model");
 
     // The B-field update is subcycled to improve stability - the number
-    // of sub steps can be specified by the user (defaults to 50).
+    // of sub steps can be specified by the user.
     utils::parser::queryWithParser(pp_hybrid, "substeps", m_substeps);
+    if (m_substeps % 2 != 0) {
+        Abort("hybrid_pic_model.substeps must be divisible by 2. "
+              "The value " + std::to_string(m_substeps) + " is not valid.");
+    }
```
roelof-groenewald, line 46: Instead of aborting, could we just add 1 to the substeps and print a warning? That would avoid the nuisance of a job waiting in the queue for a long time just to fail for such a small thing. And it covers the case where plasma parameters are used to calculate the needed substeps (which might round to an odd number).

roelof-groenewald:
Thanks for this PR! I think we will need to update the CI tests to have double the number of substeps to avoid checksum resets.

roelof-groenewald (approved):
Thanks for updating this!

PR #6558, opened 2026-02-06
add pulsed decay collision module.

JustinRayAngus:
This PR adds a new collision class that can be used to dynamically convert a parent species into two product species with a user-specified decay rate. Mathematically, it solves the following equations:
```math
\begin{align}
\frac{\partial n_A}{\partial t} &= +\nu(t) \cdot n_1(t) = \frac{\partial n_B}{\partial t}\\
\frac{\partial n_1}{\partial t} &= -\nu(t) \cdot n_1(t)
\end{align}
```
where $n_A$ and $n_B$ are the product species densities, $n_1$ is the parent species density, and $\nu(t)$ is the decay rate. In general, $\nu$ can also be a function of space. 

This can be used, for example, to represent ionization of a neutral gas by an external laser.

Collisions are done on a cell by cell basis as follows:
1) The number of product particles to create in a given time step is computed per cell.
2) Both product species particles are created with position matching a randomly chosen parent species particle in the cell. Their velocities are set to the random parent particle's velocity plus a direction- and species-dependent thermal velocity obtained from a user-defined normal distribution.
4) The weight of the new product particle pair is subtracted from the parent species.

Tasks:
- [x] get spatial coordinates to pass into collision rate parser
- [x] add charge and mass checks for the parent species and the product species
- [x] add documentation
- [x] add CI test

The figure below shows results of using this module in WarpX on both CPU and GPU, illustrating that the output matches a simple 0D model for the specified ionization rate.
PulsedIonization_forPR
0D model:
```matlab
shift = 150.0e-9;
tau = 20.0e-9;
n10 = 2.2e17; 
dt = 0.001e-9;
time = 0:dt:250e-9;
Amplitude = 5.975e7;
decay_rate = Amplitude*exp(-((time-shift)/tau).^2/2);
n1 = zeros(size(time)); n1(1) = n10;
for it = 2:length(time)
    n1(it) = n1(it-1)*exp(-decay_rate(it)*dt);
end
nI = n10 - n1;
```

### DIFF
--- a/Docs/source/theory/multiphysics/collisions.rst
+++ b/Docs/source/theory/multiphysics/collisions.rst
@@ -55,7 +55,7 @@ Once a particle is selected for a specific collision process, that process deter
 .. _multiphysics-collisions-pulseddecay:
 
 Pulsed Decay
----------------------------------------
+------------
 
 This collision module can be used to have a parent species decay into two product
 species with a user-defined decay rate. Mathematically, it solves the following
--- a/Docs/source/theory/multiphysics/collisions.rst
+++ b/Docs/source/theory/multiphysics/collisions.rst
@@ -52,6 +52,28 @@ for all scattering processes are evaluated at the energy as calculated above.
 
 Once a particle is selected for a specific collision process, that process determines how the particle is scattered as outlined below.
 
+.. _multiphysics-collisions-pulseddecay:
+
+Pulsed Decay
+---------------------------------------
+
+This collision module can be used to have a parent species decay into two product
+species with a user-defined decay rate. Mathematically, it solves the following
+rate equations on a cell-by-cell basis:
+
+    .. math::
+
+       \begin{aligned}
+        \ n_1 &\Rightarrow n_A + n_B, \\
+        \frac{dn_1}{dt} &= -\nu(t)n_1, \\
+        \frac{dn_A}{dt} &= +\nu(t)n_1 = \frac{dn_B}{dt}, \\
+       \end{aligned}
+where `n_1` is the parent species density, `n_A` and `n_B` are the product species densities,
+and `\nu` is the user-specified decay rate. Using the parser, `\nu = \nu(x,y,z,t)` can be specified
+as a function of space and time.
+
+This can be used, for example, to represent ionization of a parent species by an externally applied laser pulse.
+
 .. _multiphysics-collisions-dsmc:
 
 Direct Simulation Monte Carlo (DSMC)
--- a/Docs/source/theory/multiphysics/collisions.rst
+++ b/Docs/source/theory/multiphysics/collisions.rst
@@ -67,6 +67,7 @@ rate equations on a cell-by-cell basis:
         \frac{dn_1}{dt} &= -\nu(t)n_1, \\
         \frac{dn_A}{dt} &= +\nu(t)n_1 = \frac{dn_B}{dt}, \\
        \end{aligned}
+
 where :math:`n_1` is the parent species density, :math:`n_A` and :math:`n_B` are the product species densities,
 and :math:`\nu(x,y,z,t)` is the user-specified decay rate.
 
--- a/Docs/source/theory/multiphysics/collisions.rst
+++ b/Docs/source/theory/multiphysics/collisions.rst
@@ -64,13 +64,11 @@ rate equations on a cell-by-cell basis:
     .. math::
 
        \begin{aligned}
-        \ n_1 &\Rightarrow n_A + n_B, \\
         \frac{dn_1}{dt} &= -\nu(t)n_1, \\
         \frac{dn_A}{dt} &= +\nu(t)n_1 = \frac{dn_B}{dt}, \\
        \end{aligned}
-where `n_1` is the parent species density, `n_A` and `n_B` are the product species densities,
-and `\nu` is the user-specified decay rate. Using the parser, `\nu = \nu(x,y,z,t)` can be specified
-as a function of space and time.
+where :math:`n_1` is the parent species density, :math:`n_A` and :math:`n_B` are the product species densities,
+and :math:`\nu(x,y,z,t)` is the user-specified decay rate.
 
 This can be used, for example, to represent ionization of a parent species by an externally applied laser pulse.
 
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -3046,11 +3046,10 @@ Details about the collision models can be found in the :ref:`theory section <mul
     Only for ``dsmc`` with impact ionization. This specifies which one of the
     colliding particles is ionized.
 
-.. pp:param:: <collision_name>.decay_rate
-    :type: `float`
+.. pp:param:: <collision_name>.decay_rate(x,y,z,t)
+    :type: `string`
 
     The parent species decay rate (only for ``pulsed_decay``).
-    Can also provide ``<collision_name>.pulsed_decay(x,y,z,t)`` using the parser for spatial and temporal variation.
 
 .. pp:param:: <collision_name>.fixed_product_weight
     :type: `float`
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -2806,6 +2806,8 @@ Details about the collision models can be found in the :ref:`theory section <mul
     - ``background_mcc`` for collisions between particles and a neutral background.
       This is a relativistic Monte Carlo treatment for particles colliding
       with a neutral background gas. See :ref:`MCC section <multiphysics-collisions-mcc>`.
+    - ``pulsed_decay`` for decay of a parent species into two product species with a user-defined decay rate.
+      See :ref:`Pulsed Decay section <multiphysics-collisions-pulseddecay>`.
     - ``background_stopping`` for slowing of ions due to collisions with electrons or ions.
       This implements the approximate formulae as derived in Introduction to Plasma Physics,
       from Goldston and Rutherford, section 14.2.
@@ -2835,7 +2837,8 @@ Details about the collision models can be found in the :ref:`theory section <mul
     With ``bremsstrahlung``, the electron species must be given first, followed by the target species.
     If using ``background_mcc`` or ``background_stopping`` type this should be the name of the
     species for which collisions with a background will be included.
-    In this case, only one species name should be given.
+    If using ``pulsed_decay`` type this should be the name of the parent species.
+    In these three cases, only one species name should be given.
     If using ``linear_breit_wheeler`` these should be two photon species.
     If using ``linear_compton``, these should be two species: first, a photon species, and second, a lepton species, in this exact order.
 
@@ -2850,6 +2853,7 @@ Details about the collision models can be found in the :ref:`theory section <mul
     If using ``linear_breit_wheeler`` these should be two species: one of electrons and one of positrons.
     If using ``bremsstrahlung``, the product species must be of type photon.
     If using ``linear_compton``, these should be two species: first, a photon species, and second, a lepton species, in this exact order.
+    If using ``pulsed_decay``, the sum of the product species charges and mass must equal those of the parent species.
 
 .. pp:param:: <collision_name>.ndt_supercycle
     :type: ``int``
@@ -3041,6 +3045,30 @@ Details about the collision models can be found in the :ref:`theory section <mul
 
     Only for ``dsmc`` with impact ionization. This specifies which one of the
     colliding particles is ionized.
+
+.. pp:param:: <collision_name>.decay_rate
+    :type: `float`
+
+    The parent species decay rate (only for ``pulsed_decay``).
+    Can also provide ``<collision_name>.pulsed_decay(x,y,z,t)`` using the parser for spatial and temporal variation.
+
+.. pp:param:: <collision_name>.fixed_product_weight
+    :type: `float`
+
+    Fixed particle weight of product species (only for ``pulsed_decay``).
+    Can be estimated as :math:`n_{\text{target}}dV/N_{ppc}`, where :math:`n_{\text{target}}` is the target density of the product species, :math:`dV` is the cell volume, and :math:`N_{ppc}` is the target number of particle per cell for each product species.
+
+.. pp:param:: <collision_name>.productA_temperature_eV
+    :type: `float array, size 3`
+
+    Direction-dependent temperature used to assign a random thermal velocity to product species A (only for ``pulsed_decay``).
+    Example: ``<collision_name>.productA_temperature_eV = 0.1 0.2 0.3``.
+
+.. pp:param:: <collision_name>.productB_temperature_eV
+    :type: `float array, size 3`
+
+    Direction-dependent temperature used to assign a random thermal velocity to product species B (only for ``pulsed_decay``).
+    Example: ``<collision_name>.productB_temperature_eV = 0.1 0.2 0.3``.
 
 .. pp:param:: <collision_name>.Z
     :type: ``integer``
--- a/Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
+++ b/Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
@@ -214,17 +214,18 @@ PulsedDecay::doCollisions (amrex::Real cur_time, amrex::Real dt, MultiParticleCo
                     // Get physical coordinates at cell center
                     const amrex::IntVect iv = box.atOffset(i_cell);
                     amrex::XDim3 xyz_cc = {0.0_rt, 0.0_rt, 0.0_rt};
+                    const amrex::Real half = 0.5_rt;
 #if   defined(WARPX_DIM_1D_Z)
-                    xyz_cc.z = xyzmin.z + (iv[0] - lo.x + 0.5_rt)*dx[0];
+                    xyz_cc.z = xyzmin.z + (iv[0] - lo.x + half)*dx[0];
 #elif defined(WARPX_DIM_RCYLINDER) || defined(WARPX_DIM_RSPHERE)
-                    xyz_cc.x = xyzmin.x + (iv[0] - lo.x + 0.5_rt)*dx[0];
+                    xyz_cc.x = xyzmin.x + (iv[0] - lo.x + half)*dx[0];
 #elif defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
-                    xyz_cc.x = xyzmin.x + (iv[0] - lo.x + 0.5_rt)*dx[0];
-                    xyz_cc.z = xyzmin.z + (iv[1] - lo.y + 0.5_rt)*dx[1];
+                    xyz_cc.x = xyzmin.x + (iv[0] - lo.x + half)*dx[0];
+                    xyz_cc.z = xyzmin.z + (iv[1] - lo.y + half)*dx[1];
 #elif defined(WARPX_DIM_3D)
-                    xyz_cc.x = xyzmin.x + (iv[0] - lo.x + 0.5_rt)*dx[0];
-                    xyz_cc.y = xyzmin.y + (iv[1] - lo.y + 0.5_rt)*dx[1];
-                    xyz_cc.z = xyzmin.z + (iv[2] - lo.z + 0.5_rt)*dx[2];
+                    xyz_cc.x = xyzmin.x + (iv[0] - lo.x + half)*dx[0];
+                    xyz_cc.y = xyzmin.y + (iv[1] - lo.y + half)*dx[1];
+                    xyz_cc.z = xyzmin.z + (iv[2] - lo.z + half)*dx[2];
 #endif
 
                     // Compute total weight of products to create in this cell
--- a/Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
+++ b/Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
@@ -297,20 +297,20 @@ PulsedDecay::doCollisions (amrex::Real cur_time, amrex::Real dt, MultiParticleCo
                         const index_type ip_A = old_npA + new_idx;
                         const index_type ip_B = old_npB + new_idx;
 
-                        // Get a random particle from species 1 in this cell
-                        const index_type k = static_cast<index_type>(amrex::Random(engine) * amrex::Real(num_in_cell));
-                        index_type idx = cell_start_1 + k;
+                        // Get a random particle index from species 1 in this cell
+                        index_type k = static_cast<index_type>(amrex::Random(engine) * amrex::Real(num_in_cell));
 
                         // Probe until a valid particle is found (should always be at least one)
-                        index_type ip = -1;
+                        index_type ip  = -1;
                         for (index_type t = 0; t < num_in_cell; ++t) {
-                            const index_type cand = indices_1[idx];
-                            if (idcpu1[cand] != amrex::ParticleIdCpus::Invalid) {
-                                ip = cand;
+                            const index_type cand_k = (k + t) % num_in_cell;
+                            const index_type idx = cell_start_1 + cand_k;
+                            const index_type cand_ip = indices_1[idx];
+                            if (idcpu1[cand_ip] != amrex::ParticleIdCpus::Invalid) {
+                                k  = cand_k;
+                                ip = cand_ip;
                                 break;
                             }
-                            ++idx;
-                            if (idx == cell_stop_1) { idx = cell_start_1; }
                         }
                         AMREX_IF_ON_DEVICE((
                             AMREX_DEVICE_ASSERT(ip >= 0);
@@ -346,23 +346,20 @@ PulsedDecay::doCollisions (amrex::Real cur_time, amrex::Real dt, MultiParticleCo
                         wB[ip_B] = wpAB;
 
                         // Remove product weight from species 1
-                        amrex::ParticleReal wp_remove = amrex::min(wpAB, w1[ip]);
-                        BinaryCollisionUtils::remove_weight_from_colliding_particle(
-                            w1[ip], idcpu1[ip], wp_remove);
-                        amrex::ParticleReal wp_remaining = wpAB - wp_remove;
-                        if (wp_remaining > 0.0_prt) {
-                            for (index_type t = 0; t < num_in_cell; ++t) {
-                                ++idx;
-                                if (idx == cell_stop_1) { idx = cell_start_1; }
-                                const index_type ip2 = indices_1[idx];
-                                if (idcpu1[ip2] != amrex::ParticleIdCpus::Invalid) {
-                                    wp_remove = amrex::min(wp_remaining, w1[ip2]);
-                                    BinaryCollisionUtils::remove_weight_from_colliding_particle(
-                                        w1[ip2], idcpu1[ip2], wp_remove);
-                                    wp_remaining -= wp_remove;
-                                }
-                                if (wp_remaining <= 0.0_prt) { break; }
-                            }
+                        amrex::ParticleReal wp_remaining = wpAB;
+                        for (index_type t = 0; t < num_in_cell; ++t) {
+                            const index_type k2 = (k + t) % num_in_cell;
+                            const index_type idx2 = cell_start_1 + k2;
+                            const index_type ip2 = indices_1[idx2];
+
+                            if (idcpu1[ip2] == amrex::ParticleIdCpus::Invalid) { continue; }
+
+                            const amrex::ParticleReal wp_remove = amrex::min(wp_remaining, w1[ip2]);
+                            BinaryCollisionUtils::remove_weight_from_colliding_particle(
+                                w1[ip2], idcpu1[ip2], wp_remove);
+
+                            wp_remaining -= wp_remove;
+                            if (wp_remaining <= 0.0_prt) { break; }
                         }
 
                     }
--- a/Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
+++ b/Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
@@ -32,6 +32,8 @@ PulsedDecay::PulsedDecay (std::string const& collision_name, MultiParticleContai
     WARPX_ALWAYS_ASSERT_WITH_MESSAGE(m_species_names.size() == 1,
         "PulsedDecay: pulsed_decay must have exactly one species.");
 
+    using namespace amrex::literals;
+
     const amrex::ParmParse pp_collision_name(collision_name);
 
     // Get the product species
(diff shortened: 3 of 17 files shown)


Source/Particles/Collision/PulsedIonization/PulsedIonization.cpp
```diff
@@ -0,0 +1,377 @@
...
+    auto& product_species_0 = mypc->GetParticleContainerFromName(m_product_species[0]);
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(product_species_0.AmIA<PhysicalSpecies::electron>(),
+        "PulsedIonization: The first product species must be an electron");
+
+    // Get the product electron and ion particle weight
+    pp_collision_name.get("fixed_product_weight", m_fixed_product_weight);
+
+    // Parse the direction-dependent electron temperature
+    amrex::Vector<amrex::ParticleReal> Te_tmp;
+    pp_collision_name.getarr("electron_temperature_eV", Te_tmp);
```
dpgrote: This could be generalized by allowing the specification of the temperature for both product species. Then, the above check would not be needed since the first product species would not need to be electrons.

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE( m_product_species.size() == 2,
+        "PulsedIonization: product_species size must be equal to two");
+
+    auto& product_species_0 = mypc->GetParticleContainerFromName(m_product_species[0]);
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE(product_species_0.AmIA<PhysicalSpecies::electron>(),
+        "PulsedIonization: The first product species must be an electron");
+
+    // Get the product electron and ion particle weight
+    pp_collision_name.get("fixed_product_weight", m_fixed_product_weight);
```
dpgrote, line 72: Instead of a fixed weight, maybe this could be a fraction of the initial weight of the species. That might be more natural since it would be independent of the gridding and other numeric quantities.

Source/Particles/Collision/PulsedIonization/PulsedIonization.cpp
```diff
@@ -0,0 +1,377 @@
...
+    pp_collision_name.getarr("electron_drift_velocity", Vd_tmp);
+
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE( Vd_tmp.size() == 3,
+        "PulsedIonizationFunc: electron_drift_velocity must have exactly 3 values");
+
+    m_electron_drift_velocity = {Vd_tmp[0], Vd_tmp[1], Vd_tmp[2]};
+
+    // Parse the ionization rate
+    amrex::ParticleReal ionization_rate = 0;
+    if (utils::parser::queryWithParser(pp_collision_name, "ionization_rate", ionization_rate)) {
```
dpgrote: This simplified input parameter is probably not needed since it doesn't provide any extra functionality.

Source/Particles/Collision/PulsedIonization/PulsedIonization.cpp
```diff
@@ -0,0 +1,377 @@
...
+                    const amrex::ParticleReal wtot1 = wtot1_in_each_cell[i_cell];
+                    if (wtot1 == 0.0_prt) { return; }
+
+                    // Compute total weight of products to create in this cell
+                    const amrex::ParticleReal nu_izn = nu_func(0.,0.,0.,cur_time);
+                    const amrex::ParticleReal total_product = wtot1*(1.0 - std::exp(-nu_izn*dt));
+
+                    // Compute number of products macro particles to create in this cell
+                    const amrex::ParticleReal num_expected = total_product/fixed_product_weight;
+                    int num_macro_particles = static_cast<int>(std::floor(num_expected));
```
dpgrote: I usually write this kind of thing as
```
int num_macro_particles = static_cast<int>(std::floor(num_expected + amrex::Random(engine)));
```
Which does the same thing as the `if` block below.

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
```diff
@@ -0,0 +1,377 @@
...
+                    // Compute number of products macro particles to create in this cell
+                    const amrex::ParticleReal num_expected = total_product/fixed_product_weight;
+                    int num_macro_particles = static_cast<int>(std::floor(num_expected));
+                    const amrex::Real rand = amrex::Random(engine);
+                    if (rand < (num_expected - num_macro_particles)) {
+                        num_macro_particles++;
+                    }
+
+                    // Do not permit total product weight to exceed wtot1
+                    if (num_macro_particles*fixed_product_weight > wtot1) {
```
dpgrote, line 245: This check is probably not necessary. It could be handled below in the loops. Since you already take the weight from multiple particles if there is not enough weight left in the particle, a check could be added there that if there is no more weight left, then that last new particle would get whatever is left over.

Source/Particles/Collision/PulsedIonization/PulsedIonization.cpp
```diff
@@ -0,0 +1,377 @@
...
+                                amrex::Abort("Error in PulsedIonization: valid species 1 particle not found!");
+                            }
+                        ))
+
+                        // Create product particles at position of particle from species 1
+                        CopyIonF(ion, soa_1, ip, static_cast<int>(i_ion), engine);
+                        CopyEleF(ele, soa_1, ip, static_cast<int>(i_ele), engine);
+
+                        // Set electron velocity from specified normal distribution
+                        uxe[i_ele] = electron_Vdrift_x;
```
dpgrote: Is there a reason why the drift velocity is being overwritten? Why wouldn't it be the same as the particle being ionized?

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
```diff
@@ -0,0 +1,377 @@
...
+                        const index_type new_idx = start + j;
+                        const index_type i_ele = old_np_ele + new_idx;
+                        const index_type i_ion = old_np_ion + new_idx;
+
+                        // Get a random particle from species 1 in this cell
+                        const index_type k = static_cast<index_type>(amrex::Random(engine) * amrex::Real(num_in_cell));
+                        index_type idx = cell_start_1 + k;
+
+                        // Probe until a valid particle is found (should always be at least one)
+                        index_type ip = -1;
```
dpgrote: I think this loop and the one below could be combined and simplified.
```
amrex::Real wp_remaining = wpEI;
for (index_type t = 0; t < num_in_cell; ++t) {
    index_type ip = cell_start1 + (k + t) % num_in_cell;
    if (w1[ip] > 0.) {
        idx = ip;
        amrex::Real wp_remove = min(wp_remaining, w1[ip]);
        BinaryCollisionUtils::remove_weight_from_colliding_particle(w1[ip], idcpu1[ip], wp_remaining);
        wp_remaining -= wp_remove;
        if (wp_remaining <= 0.) { break; }
    }
}
if (wpEI - wp_remaining <= 0) { return; }
ion.m_rdata[PIdx::w][i_ion] = wpEI - wp_remaining;
ele.m_rdata[PIdx::w][i_ele] = wpEI - wp_remaining;
CopyIonF(ion, soa_1, ip, static_cast<int>(i_ion), engine);
CopyEleF(ele, soa_1, ip, static_cast<int>(i_ele), engine);
```
Though, special handling would be needed to invalidate any product particles that might be left over after the weight of the species is used up.

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
JustinRayAngus: Thanks for the suggestion. I'm not a fan of specifying the weight either for exactly the reasons you mentioned. However, I'm not sure doing what you suggest is without uninteded consequences. I imagine that I would need to compute the average weight of the parent species in a cell at the start of the simulation. In that case, then the weight could 1) vary in space and 2) be different on a restart.

Source/Particles/Collision/PulsedIonization/PulsedIonization.cpp
JustinRayAngus: I agree. I copied this from BackgroundMCCCollsions.cpp where they do the same thing.

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
JustinRayAngus: _if there is no more weight left, then that last new particle would get whatever is left over._ 

I'm not a fan of doing things that can introduce variable weight particles in special cases when you are expecting fixed weight particles.

Source/Particles/Collision/PulsedIonization/PulsedIonization.cpp
JustinRayAngus: The drift velocity is a mean property of all the particles. This value is not computed, so I don't have it.

Perhaps you are suggesting to compute the mean velocity of the parent species, and then initialize the electron velocity from a normal distribution with that drift?

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
dpgrote: Ok, fair point. But this could leave a little bit of the species that couldn't be ionized, or would be thrown away, leaving a small error.

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
dpgrote: Yes, I see the issue. Having the fixed weight may not be too bad since it could easily be an expression anyway.

Source/Particles/Collision/PulsedIonization/PulsedIonization.cpp
dpgrote: I was thinking that the velocity would be obtained from the particle being ionized, the same as the velocity of the new ion (then plus the thermal velocity).

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
JustinRayAngus: That's true, and I did think about that. I consider that to be within the expected error associated with finite particle weights, which diminishes as the number of particle per cell (~1/weight) increases.

Source/Particles/Collision/PulsedIonization/PulsedIonization.cpp
JustinRayAngus: I see. Yea, I think that would work.

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
JustinRayAngus: While we are on this topic - Even if you set the weight of the product to be an integer multiple of the parent particle weights, you can often end up with a super small weight parent particle in the end owing to round off error. In PICNIC, I have ran into this issue with electron-impact ionization. The fix I did was to round the particle weights to float precision on initialization, which is controlled by a user input parameter.

Source/Particles/Collision/PulsedIonization/PulsedIonization.cpp
JustinRayAngus: Yup. I did this.

Now that it is general, the name PulsedIonization makes no sense. I have changed it to PulsedDecay.

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
```diff
@@ -0,0 +1,408 @@
...
+                        ))
+
+                        // Create product particles at position of particle from species 1
+                        CopyAF(*soaA_ptr, soa_1, ip, static_cast<int>(ip_A), engine);
+                        CopyBF(*soaB_ptr, soa_1, ip, static_cast<int>(ip_B), engine);
+
+                        // Set product species A particle velocity to parent particle velocity plus thermal
+                        uAx[ip_A] = ux1[ip];
+                        uAy[ip_A] = uy1[ip];
+                        uAz[ip_A] = uz1[ip];
```
JustinRayAngus: @dpgrote I think CopyAF() above does this for me? If so, I can remove these lines.

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
```diff
@@ -0,0 +1,408 @@
...
+            // Initialize the user runtime components
+            if (total_new > 0) {
+                amrex::GpuArray<index_type, 2> products_np{old_npA, old_npB};
+                amrex::GpuArray<ParticleTileType*, 2> tile_products{&ptileA, &ptileB};
+                amrex::GpuArray<WarpXParticleContainer*, 2> pc_products{&productA, &productB};
+                for (int i = 0; i < 2; i++) {
+                    const auto start_index = int(products_np[i]);
+                    const auto stop_index  = int(products_np[i] + total_new);
+                    ParticleCreation::DefaultInitializeRuntimeAttributes(*tile_products[i],
+                        0, 0,
```
JustinRayAngus: @dpgrote I feel like DefaultInitializeRuntimeAttributes() could be called with much fewer arguments. It seems like you can just pass pc_products rather than lots of arguments obtained from pc_products. This identical chunk of code is called from 8 places in WarpX.

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
```diff
@@ -0,0 +1,407 @@
...
+    pp_collision_name.get("fixed_product_weight", m_fixed_product_weight);
+
+    // Parse the direction-dependent temperature for product species A
+    amrex::Vector<amrex::ParticleReal> TA_tmp;
+    pp_collision_name.getarr("productA_temperature_eV", TA_tmp);
+
+    WARPX_ALWAYS_ASSERT_WITH_MESSAGE( TA_tmp.size() == 3,
+        "PulsedDecayFunc: productA_temperature_eV must have exactly 3 values");
+
+    for (int i = 0; i < 3; ++i) {
```
dpgrote, line 81: This is another way of doing this that is a little cleaner (hopefully this compiles)
```
auto TA_tmp_min = std::min_element(TA_tmp.begin(), TA_tmp.end());
WARPX_ALWAYS_ASSERT_WITH_MESSAGE(*TA_tmp_min >= 0.,
           "PulsedDecayFunc: productA_temperature_eV must be greater than or equal to zero");
```

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
dpgrote: A check could be added to reject particles with a weight < 1.

Either way of doing this is Ok (having a lighter weight particle, or throwing away some weight), but it should be mentioned in the comments and/or the documentation.

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
dpgrote: Yes, correct.

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
dpgrote: I agree. The use of this routine could be much simplified. This would be a separate PR.

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
JustinRayAngus: I like keeping them separate, but I did clean up both of these loops.

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
```diff
@@ -0,0 +1,398 @@
...
+                        uBy[ip_B] += VtB_y*RandomNormal(0_prt, 1.0_prt, engine);
+                        uBz[ip_B] += VtB_z*RandomNormal(0_prt, 1.0_prt, engine);
+
+                        // Set the weight of the product particles
+                        wA[ip_A] = wpAB;
+                        wB[ip_B] = wpAB;
+
+                        // Remove product weight from species 1
+                        amrex::ParticleReal wp_remaining = wpAB;
+                        for (index_type t = 0; t < num_in_cell; ++t) {
```
dpgrote, line 368: This is fine. I was just trying to avoid duplicated code, calculating the indices and checking for valid particle.

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
```diff
@@ -0,0 +1,398 @@
...
+            const index_type* AMREX_RESTRICT nprod_cell_ptr = num_products_vec.dataPtr();
+
+            index_type const* AMREX_RESTRICT cell_offsets_1 = bins_1.offsetsPtr();
+            index_type const* AMREX_RESTRICT indices_1 = bins_1.permutationPtr();
+
+            uint64_t* AMREX_RESTRICT idcpu1 = soa_1.m_idcpu;
+
+            amrex::ParticleReal const* AMREX_RESTRICT ux1  = soa_1.m_rdata[PIdx::ux];
+            amrex::ParticleReal const* AMREX_RESTRICT uy1  = soa_1.m_rdata[PIdx::uy];
+            amrex::ParticleReal const* AMREX_RESTRICT uz1  = soa_1.m_rdata[PIdx::uz];
```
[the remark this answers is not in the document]
JustinRayAngus: Fixed.

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
```diff
@@ -0,0 +1,398 @@
...
+            const index_type* AMREX_RESTRICT offsets_ptr = offsets.dataPtr();
+            const index_type* AMREX_RESTRICT nprod_cell_ptr = num_products_vec.dataPtr();
+
+            index_type const* AMREX_RESTRICT cell_offsets_1 = bins_1.offsetsPtr();
+            index_type const* AMREX_RESTRICT indices_1 = bins_1.permutationPtr();
+
+            uint64_t* AMREX_RESTRICT idcpu1 = soa_1.m_idcpu;
+
+            amrex::ParticleReal const* AMREX_RESTRICT ux1  = soa_1.m_rdata[PIdx::ux];
+            amrex::ParticleReal const* AMREX_RESTRICT uy1  = soa_1.m_rdata[PIdx::uy];
```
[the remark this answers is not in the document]
JustinRayAngus: Fixed

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+            const index_type* AMREX_RESTRICT offsets_ptr = offsets.dataPtr();
+            const index_type* AMREX_RESTRICT nprod_cell_ptr = num_products_vec.dataPtr();
+
+            index_type const* AMREX_RESTRICT cell_offsets_1 = bins_1.offsetsPtr();
+            index_type const* AMREX_RESTRICT indices_1 = bins_1.permutationPtr();
+
+            uint64_t* AMREX_RESTRICT idcpu1 = soa_1.m_idcpu;
+
+            amrex::ParticleReal const* AMREX_RESTRICT ux1  = soa_1.m_rdata[PIdx::ux];
```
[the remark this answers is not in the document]
JustinRayAngus: Fixed

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
JustinRayAngus: Thanks for the suggestion, but it's functionally the same and so I think I'll keep it as is.

Docs/source/theory/multiphysics/collisions.rst
```diff
@@ -52,6 +52,27 @@ for all scattering processes are evaluated at the energy as calculated above.
 
 Once a particle is selected for a specific collision process, that process determines how the particle is scattered as outlined below.
 
+.. _multiphysics-collisions-pulseddecay:
+
+Pulsed Decay
+---------------------------------------
```
ax3l: Just RST alignment/warning:
```suggestion
Pulsed Decay
------------
```

Docs/source/usage/parameters.rst
```diff
@@ -2373,6 +2377,21 @@ Details about the collision models can be found in the :ref:`theory section <mul
...
+* ``<collision_name>.fixed_product_weight`` (`float`)
+    Fixed particle weight of product species (only for ``pulsed_decay``).
+    Can be estimated as :math:`n_{\text{target}}dV/N_{ppc}`, where :math:`n_{\text{target}}` is the target density of the product species, :math:`dV` is the cell volume, and :math:`N_{ppc}` is the target number of particle per cell for each product species.
+
+* ``<collision_name>.productA_temperature_eV`` (`float array, size 3`)
+    Direction-dependent temperature used to assign a random thermal velocity to product species A (only for ``pulsed_decay``).
+    Example: ``<collision_name>.productA_temperature_eV = 0.1 0.2 0.3``.
+
+* ``<collision_name>.productB_temperature_eV`` (`float array, size 3`)
+    Direction-dependent temperature used to assign a random thermal velocity to product species B (only for ``pulsed_decay``).
```
ax3l: Maybe it is clearer for user input if we grouped `pulse_decay` options like this under a prefix?
```ini
<collision_name>.pulsed_decay.fixed_product_weight
<collision_name>.pulsed_decay.productA_temperature_eV
<collision_name>.pulsed_decay.productB_temperature_eV
```

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
```diff
@@ -0,0 +1,406 @@
...
+
+#include "Particles/Collision/BinaryCollision/BinaryCollisionUtils.H"
+#include "Particles/ParticleCreation/SmartCopy.H"
+#include "Utils/Parser/ParserUtils.H"
+#include "Utils/TextMsg.H"
+#include "Utils/ParticleUtils.H"
+#include "WarpX.H"
+
+#include <ablastr/profiler/ProfilerWrapper.H>
+#include <AMReX_ParmParse.H>
```
ax3l, line 21: Used below:
```suggestion

#include <AMReX_Math.H>
#include <AMReX_ParmParse.H>
```

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
```diff
@@ -0,0 +1,406 @@
...
+#include <AMReX_REAL.H>
+#include <AMReX_Vector.H>
+#include <AMReX_ParticleTile.H>
+
+#include <string>
```
ax3l, line 27: Used below:
```suggestion
#include <cmath>
#include <string>
```

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
```diff
@@ -0,0 +1,406 @@
...
+    for (int lev = 0; lev <= flvl; ++lev) {
+
+        // Enable tiling
+        amrex::MFItInfo info;
+        if (amrex::Gpu::notInLaunchRegion()) { info.EnableTiling(species1.tile_size); }
+
+#ifdef AMREX_USE_OMP
+        info.SetDynamic(true);
+#pragma omp parallel if (amrex::Gpu::notInLaunchRegion())
+#endif
```
ax3l, line 176: @atmyers don't we set these things globally in WarpX anyway?

ax3l (commented):
That looks very clean to me, nice PR!

I added a few inline suggestions, including grouping the input options a bit more for structure/clarity.

Docs/source/usage/parameters.rst
JustinRayAngus: @ax3l Are you suggesting to also add the prefix to how it is parsed from the input deck? If not, then there could be a confusing mismatch between the docs and the input deck.

On this note, I'm not happy with how the docs read for the collision parameters. I added the parameters for this new module by just following suite with what is already there. A Refactor of the docs for collisions could be a future PR (or a task for AI).

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
JustinRayAngus: These lines were copied from `BinaryCollision.H`

Docs/source/usage/parameters.rst
JustinRayAngus: @ax3l

ax3l: I am confused on your response :sweat_smile: 

I suggest:
`<collision_name>.productB_temperature_eV` becomes
`<collision_name>.pulsed_decay.productB_temperature_eV`
etc.

So it is clear when reading the collision what the `productB_temperature_eV` is used for: it is a property of the `pulsed_decay` type of collision.

I suggest to rename them.

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
ax3l: Ok, but still looks like a code smell. let's check with @atmyers , does not have to hold up the PR though.

atmyers: `warpx.do_dynamic_scheduling = 1`, which is on by default, turns on dynamic scheduling for all `WarpXParIter` loops, but that won't affect loops like the one in `BinaryCollision.H`, which uses `MFIter`. Although maybe we *should* be using a `WarpXParIter` there...

ax3l: Ok good, let us look for this pattern after this PR.

Docs/source/usage/parameters.rst
JustinRayAngus: Is there precedent for this? Are there other collision model specific parameters that are parsed this way?

ax3l: Not yet done in collisions: I just saw we already have 7-8 other collision types with unstructured input by now...

So let's do this in a follow-up and try to give them some structure.

ax3l (approved):

Docs/source/usage/parameters.rst
```diff
@@ -3042,6 +3046,30 @@ Details about the collision models can be found in the :ref:`theory section <mul
     Only for ``dsmc`` with impact ionization. This specifies which one of the
     colliding particles is ionized.
```
dpgrote, line 3048: This should be updated. ParmParse is only called with the full parameter `pulsed_decay(x,y,z,t)`.

dpgrote (approved):
Looks good!

JustinRayAngus (commented):
@ax3l @WeiqunZhang @dpgrote 

Is the mapping here to cell indices consistent? It doesn't seem to work on the GPU.

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
```diff
@@ -0,0 +1,411 @@
...
+            amrex::ParticleReal* AMREX_RESTRICT wtot1_in_each_cell = wtot1_vec.dataPtr();
+            amrex::ParticleReal* AMREX_RESTRICT w1 = soa_1.m_rdata[PIdx::w];
+
+            amrex::ParallelFor( np1,
+                [=] AMREX_GPU_DEVICE (int ip) noexcept
+                {
+                    amrex::Gpu::Atomic::AddNoRet(&wtot1_in_each_cell[bins_1_ptr[ip]],
+                                                 w1[ip]);
+                }
+            );
```
JustinRayAngus, line 201: @RemiLehe @EZoni @ax3l @WeiqunZhang @dpgrote 

Is the mapping to cell indices consistent here? It doesn't seem to work on GPU.
`ParticleBins bins_1 = ParticleUtils::findParticlesInEachCell( geom_lev, mfi, ptile_1 );` and
`amrex::Gpu::Atomic::AddNoRet(&wtot1_in_each_cell[bins_1_ptr[ip]], w1[ip]);`

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
```diff
@@ -0,0 +1,411 @@
...
+#elif defined(WARPX_DIM_RCYLINDER) || defined(WARPX_DIM_RSPHERE)
+                    xyz_cc.x = xyzmin.x + (iv[0] - lo.x + half)*dx[0];
+#elif defined(WARPX_DIM_XZ) || defined(WARPX_DIM_RZ)
+                    xyz_cc.x = xyzmin.x + (iv[0] - lo.x + half)*dx[0];
+                    xyz_cc.z = xyzmin.z + (iv[1] - lo.y + half)*dx[1];
+#elif defined(WARPX_DIM_3D)
+                    xyz_cc.x = xyzmin.x + (iv[0] - lo.x + half)*dx[0];
+                    xyz_cc.y = xyzmin.y + (iv[1] - lo.y + half)*dx[1];
+                    xyz_cc.z = xyzmin.z + (iv[2] - lo.z + half)*dx[2];
+#endif
```
JustinRayAngus, line 234: @RemiLehe @EZoni @ax @WeiqunZhang @dpgrote 

When I compute the coordinates here on GPU they are not what I expect. That is, they do not correspond the the expected physical locations with non-zero wtot1.

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
atmyers: The mapping looks right, but is it possible that we need to skip invalid particles here?

JustinRayAngus: @atmyers we could add to skip invalid, but would that cause the bug I'm seeing on GPU described in my other comment?  The computed physical location of cells with nonzero wtot1 is incorrect on GPU. Works on CPU

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
JustinRayAngus: @atmyers @WeiqunZhang @ax3l @dpgrote @RemiLehe @EZoni 

I found the issue.` ParticleBins bins_1 = ParticleUtils::findParticlesInEachCell( geom_lev, mfi, ptile_1 );` uses DenseBins, which flattens in 2D (for example) like
`i_cell = ix * nz + iz`

but `box.atOffset(i_cell);` assumes
`i_cell = ix + nx * iz`

Is there a different version of `atOffset(i_cell)` that does this? Or do I need to create my own function?

Turns out it was also incorrect on CPU, but because it was using tiling (smaller boxes) the error was not nearly as exaggerated as it was on GPU.

JustinRayAngus: See PR #6827

Source/Particles/Collision/PulsedDecay/PulsedDecay.cpp
JustinRayAngus: See PR #6827

PR #6560, opened 2026-02-07
Replace missing Summit postprocessing documentation

RemiLehe:
The Summit post-processing documentation was removed in https://github.com/BLAST-WarpX/warpx/pull/6528. However, it was still being referenced in other parts of the documentation, leading to:
```
WARNING: undefined label: 'post-processing-summit'
```

This PR copies the post-processing documentation that was removed in https://github.com/BLAST-WarpX/warpx/pull/6528 into the current Frontier documentation, and fixes the corresponding references.


lucafedeli88 (approved):
oh, I see... thanks for quickly fixing this, @RemiLehe !
Maybe we should check for undefined references in CI...

PR #6562, opened 2026-02-09
[WIP] build documentation in CI

lucafedeli88:


EZoni:
@lucafedeli88 

Thanks for thinking about this!

We already build the documentation directly through Read the Docs, which handles the `docs/readthedocs.org:warpx` CI workflow.

The builds and logs are available at https://app.readthedocs.org/projects/warpx/. However, only a few maintainers have access. 

Below is an example of the build for this PR, where you have access to raw logs, debug info, etc. 

So in my opinion we don't need an extra CI workflow to build the documentation, unless I'm missing something.

We could discuss with @ax3l who else could/should be given access to the Read the Docs project.

Screenshot from 2026-02-09 09-04-23

RemiLehe:
Superseded by https://github.com/BLAST-WarpX/warpx/pull/6569

PR #6530, opened 2026-01-27
[WIP] New mcc-v-push

oshapoval:
This PR presents the ES/EM `mid-v-push` implementation (with collisions applied in the middle of the particle momentum push), using the new interface introduced in PR #6275.

`mid-v-push` algorithm is described in [DOI: 10.1103/PhysRevE.111.02530](https://doi.org/10.1103/PhysRevE.111.025306).

This PR will be closed, as there are now two separate PRs with the ES  (https://github.com/BLAST-WarpX/warpx/pull/6572) and EM (https://github.com/BLAST-WarpX/warpx/pull/6581) `mid-v-push` implementations.

PR #6565, opened 2026-02-09
Make checksum for `test_3d_linear_compton_bunch_laser` less brittle

RemiLehe:
The test `test_3d_linear_compton_bunch_laser` involves a photon bunch passing through the simulation box, as seen in these images from different timesteps:
Screenshot 2026-02-09 at 11 53 59 AM
Screenshot 2026-02-09 at 11 54 14 AM
Screenshot 2026-02-09 at 11 54 35 AM

The end of the simulation correspond **closely** to the time at which the photon bunch has exited the simulation box. 

However, because of this, there is sometimes still one or two photons in the box at the very last timestep. It seems that this happens randomly (depending on the seed) and is related to the fact that the particles injected with flux injection are emitted with random delays **within one timestep**. For this reason, the very last few photons of the bunch can be emitted with just the right random delays for them to still be in the simulation box at the last timestep. (This is confirmed by the current value of the sum of photon weights in the checksum: `Benchmark: [photon1,particle_weight] 9.398496240601504e+15` which, after verification, turns out to be the weight of a **single photon macroparticle**.)

For this reason, the checksum is very brittle: small unrelated changes in the code (e.g. drawing new random numbers in another part of the code) can change the random delay with which the photons are emitted, and thus can change whether there are still photons in the simulation box at the end of the simulation.

This PR avoids this issue by taking one more timestep, to make sure that the simulation box is completely cleared of photons at the time when the checksum is computed.

### DIFF
--- a/Examples/Tests/linear_compton/inputs_test_3d_linear_compton_bunch_laser
+++ b/Examples/Tests/linear_compton/inputs_test_3d_linear_compton_bunch_laser
@@ -21,7 +21,7 @@ my_constants.sigma_kn_max = 6.65e-29 # Maximum value of Klein-Nishina cross-sect
 my_constants.v_rel = 2 *clight
 my_constants.probability_target_value = 0.05
 my_constants.dt =  probability_target_value / (sigma_kn_max * dens * v_rel )
-my_constants.N_step = floor(T / dt)
+my_constants.N_step = floor(T / dt) + 1
 # space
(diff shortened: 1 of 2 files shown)


EZoni:
@aeriforme 
 
FYI, just for context, this came up while debugging this test in #6530, which in turn is related to previous PRs where the same was happening as well, e.g., #5955 and #6275.

EZoni (approved):

PR #6567, opened 2026-02-09
Improve performance of hybrid-PIC algorithm

roelof-groenewald:
A couple changes have been made to improve performance of the hybrid-PIC algorithm (focussed on the field solver).
1. GPU kernels were merged in the update of the B-field from Runge-Kutta sums.
2. ~~The `FillBoundary` call for the E-field used to update the B-field has been removed since local tests showed this did not affect results.~~ -> moved to #6575 
3. Unnecessary `setVal(0)` calls were removed for MFs that are fully overwritten anyway.

A quick test on a single V100 GPU (with only 1 tile in the simulation) showed performance improvement of ~20% due to the kernel merging.
A larger test on 8 A100 GPUs (on Polaris) also showed ~20% performance improvement, going from ~3.37 steps/second before this PR to ~4.06 steps/second with this PR.

### DIFF
--- a/Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
+++ b/Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
@@ -518,16 +518,52 @@ void HybridPICModel::BfieldEvolveRK (
     );
 
     // The Bfield is now given by:
-    // B_new = B_old + 0.5 * dt * K0 + 0.5 * dt * [-curl x E(B_old + 0.5 * dt * K1)]
-    //       = B_old + 0.5 * dt * K0 + 0.5 * dt * K1
-    for (int ii = 0; ii < 3; ii++)
-    {
-        // Subtract 0.5 * dt * K0 from the Bfield for each direction, to get
-        // B_new = B_old + 0.5 * dt * K1.
-        MultiFab::Subtract(*Bfield[lev][ii], K[ii], 0, 0, 1, ng);
-        // Extract 0.5 * dt * K1 for each direction into index 1 of K.
-        MultiFab::LinComb(
-            K[ii], 1._rt, *Bfield[lev][ii], 0, -1._rt, B_old[ii], 0, 1, 1, ng
+    //   B_new = B_old + 0.5 * dt * K0 + 0.5 * dt * [-curl x E(B_old + 0.5 * dt * K1)]
+    //         = B_old + 0.5 * dt * K0 + 0.5 * dt * K1
+    //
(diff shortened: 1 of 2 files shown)


Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -519,16 +518,52 @@ void HybridPICModel::BfieldEvolveRK (
...
-    // B_new = B_old + 0.5 * dt * K0 + 0.5 * dt * [-curl x E(B_old + 0.5 * dt * K1)]
-    //       = B_old + 0.5 * dt * K0 + 0.5 * dt * K1
-    for (int ii = 0; ii < 3; ii++)
-    {
-        // Subtract 0.5 * dt * K0 from the Bfield for each direction, to get
-        // B_new = B_old + 0.5 * dt * K1.
-        MultiFab::Subtract(*Bfield[lev][ii], K[ii], 0, 0, 1, ng);
-        // Extract 0.5 * dt * K1 for each direction into index 1 of K.
-        MultiFab::LinComb(
-            K[ii], 1._rt, *Bfield[lev][ii], 0, -1._rt, B_old[ii], 0, 1, 1, ng
```
roelof-groenewald, line 531: Merging these kernels did not have a significant performance improvement in my tests, so an argument could be made that we should leave these as is for code succinctness.

Source/FieldSolver/FiniteDifferenceSolver/HybridPICModel/HybridPICModel.cpp
```diff
@@ -597,7 +662,10 @@ void HybridPICModel::FieldPush (
     CalculatePlasmaCurrent(Bfield, eb_update_E);
     // Calculate the E-field from Ohm's law
     HybridPICSolveE(Efield, Jfield, Bfield, rhofield, eb_update_E, true);
-    warpx.FillBoundaryE(ng, nodal_sync);
+    // Call FillBoundary if a collocated grid is used
+    if (Bz_IndexType[0] == Ez_IndexType[0]) {
+        warpx.FillBoundaryE(ng, nodal_sync);
+    }
```
roelof-groenewald: @clarkse I'm not sure if the 3d cylindrical test failed with the previous commit (where the `FillBoundary` call was completely removed) because of the external field handling or because `FillBoundary` is needed for a collocated grid. I'm assuming it is due to the collocated grids since the RZ compression test (which also uses external fields) passed on the previous commit. What are your thoughts?

RemiLehe (approved):

RemiLehe:
Thanks for this PR!

PR #6569, opened 2026-02-10
Fail RTD doc on warning

RemiLehe:
Follow-up to https://github.com/BLAST-WarpX/warpx/pull/6560: as suggested by @lucafedeli88, it would be good to check for warnings from the documentation in CI

### DIFF
--- a/Docs/source/theory/intro.rst
+++ b/Docs/source/theory/intro.rst
@@ -3,7 +3,7 @@
 Overview
 ========
 
-.. _theory-pi:
+.. _theory-pic:
 
 WarpX simulates the **self-consistent** evolution of **particle species** (e.g., electrons, ions, etc.) in the presence of **electric and magnetic fields**.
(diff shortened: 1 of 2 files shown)


Docs/source/theory/intro.rst
```diff
@@ -3,7 +3,7 @@
 Overview
 ========
 
-.. _theory-pic:
+.. _theory-pi:
```
RemiLehe: Intentional typo, which should result in a warning. I will remove it.

EZoni:
It looks like it worked:

This is the beginning of the log:
Screenshot from 2026-02-10 09-36-05

And this is the end of the log:
Screenshot from 2026-02-10 09-57-19

EZoni:
Debugging might not be straightforward for every PR author depending on access to these details, but that's fine - we can delegate that to the maintainers.

Docs/source/theory/intro.rst
RemiLehe:
```suggestion
.. _theory-pic:
```

EZoni (approved):

PR #6571, opened 2026-02-10
Hybrid-PIC optimization - `CalculatePlasmaCurrent`

roelof-groenewald:
This PR improves the performance of the `CalculatePlasmaCurrent` function of the hybrid-PIC algorithm (called 4 times in each RK step). Specifically, adding zeros to the current MF is avoided if no external current is used (kernel launch reduction) and, more importantly, a `FillBoundary` call is avoided by calculating the plasma current in 1 layer of ghost cells as well as in internal cells.


roelof-groenewald:
I confirmed that the magnetic reconnection test still produces the correct benchmark results by running the full test on 4 A100 GPUs (on Perlmutter).

The magnetic reconnection rate as a function of time:
image

Animation of reconnection (compare to version on [docs](https://warpx.readthedocs.io/en/latest/usage/examples/ohm_solver_magnetic_reconnection/README.html) from before this PR):

https://github.com/user-attachments/assets/1bf7724b-8f27-49f6-909c-841deda329bd

roelof-groenewald:
@clarkse @clarkse-he I had to update the benchmark values for the RZ compression test with this PR as well. Do you think you could rerun that test to make sure that it is still giving correct results? The updates to the checksum values were pretty small, so I think it is probably just floating point arithmetic differences.

clarkse:
> @clarkse @clarkse-he I had to update the benchmark values for the RZ compression test with this PR as well. Do you think you could rerun that test to make sure that it is still giving correct results? The updates to the checksum values were pretty small, so I think it is probably just floating point arithmetic differences.

I will try to run this case tomorrow and let you know.

clarkse-he:
> > @clarkse @clarkse-he I had to update the benchmark values for the RZ compression test with this PR as well. Do you think you could rerun that test to make sure that it is still giving correct results? The updates to the checksum values were pretty small, so I think it is probably just floating point arithmetic differences.
> 
> I will try to run this case tomorrow and let you know.

I found an issue with RZ hyper-resistivity application that could be driving some of this problem. I submitted a PR https://github.com/BLAST-WarpX/warpx/pull/6598 that will address this issue. I will do some further testing on it and then try merging it locally with this PR to make sure the checksums don't change.

RemiLehe (approved):

PR #6572, opened 2026-02-10
Implement collisions in the middle of the momentum push for electrostatic PIC

oshapoval:
This PR implements only the ES `mid-v-push` (=collisions are placed in the middle of the momentum push) and replaces current ES `mid-x-push` (=collisions are placed in the middle of the position push) previously implemented in https://github.com/BLAST-WarpX/warpx/pull/5955.  

This PR  is one (ES) part of the previously opened PR: https://github.com/BLAST-WarpX/warpx/pull/6530.


EZoni:
Note to everyone that this reverts some of the changes in https://github.com/BLAST-WarpX/warpx/pull/5955 and implements a different algorithm.

Docs/source/usage/parameters.rst
```diff
@@ -2377,8 +2377,8 @@ Details about the collision models can be found in the :ref:`theory section <mul
     For pairwisecoulomb collisions, when correcting the energy and momentum conservation, this parameter controls the exponent used on the particle weight when distributing the momentum correction.
     With a value greater than 1, it will distribute more of the correction to particles with higher weights.
 
-* ``collisions.split_position_push`` (``bool``, optional, default = 1)
-    If true, collisions are performed in the middle of the position push, which is split into two substeps.
+* ``collisions.split_momentum_push`` (``bool``, optional, default = 1)
+    If true, collisions are performed in the middle of the momentum push, which is split into two substeps.
     This improves energy conservation, as demonstrated in (`Vay et al., Phys. Rev. E 111, 2025 <https://doi.org/10.1103/PhysRevE.111.025306>`__).
     This is only implemented for the explicit evolve scheme and is not available for the implicit evolve schemes.
     It is also not available with embedded boundaries.
```
RemiLehe, line 2419: In a follow-up PR, we can remove this comment, and turn on splitting for the tests in which #5955 explicitly disabled splitting: [here](https://github.com/BLAST-WarpX/warpx/pull/5955/changes#diff-44a60b5193b010954027e556ffa076e2cf7f5e426d56d5c510809a538e40d4a6), [here](https://github.com/BLAST-WarpX/warpx/pull/5955/changes#diff-d7c758badee35f4f64c8481f4cafb202108860cd13dcbb2d765865f7a6b2998e) and [here](https://github.com/BLAST-WarpX/warpx/pull/5955/changes#diff-802250e78ee19e1fcfec13b00d193423ffaa161af4aa76e18891cfa2247dd435)

Regression/Checksum/benchmarks_json/test_2d_collision_xz.json
```diff
@@ -1,26 +1,25 @@
...
     "Bz": 0.0,
     "Ex": 0.0,
     "Ey": 0.0,
     "Ez": 0.0
+  },
+  "ion": {
+    "particle_momentum_x": 2.4933059126793735e-19,
+    "particle_momentum_y": 2.2823553148843995e-19,
+    "particle_momentum_z": 2.2829588242547043e-19,
+    "particle_position_x": 2662166.5033456455,    "particle_position_y": 2651752.9690234065,
```
RemiLehe:
```suggestion
    "particle_position_x": 2662166.5033456455,
    "particle_position_y": 2651752.9690234065,
```

Source/WarpX.cpp
```diff
@@ -1887,37 +1887,37 @@ WarpX::ReadParameters ()
...
-    // Set the default value of m_collisions_split_position_push
-    m_collisions_split_position_push = false;
+    // Set the default value of m_collisions_split_momentum_push
+    m_collisions_split_momentum_push = false;
     const amrex::ParmParse pp_collisions("collisions");
     amrex::Vector<std::string> collision_names;
     pp_collisions.queryarr("collision_names", collision_names);
     bool const collisions = (static_cast<int>(collision_names.size()) == 0) ? false : true;
     if (collisions) {
         if (evolve_scheme == EvolveScheme::Explicit && !EB::enabled()) {
```
RemiLehe, line 1906: In the above-mentioned follow-up PR that enables splitting for the EB, we could remove `&& !EB::enabled()`

Source/Particles/RigidInjectedParticleContainer.cpp
```diff
@@ -440,15 +439,15 @@ RigidInjectedParticleContainer::PushP (int lev, Real dt,
...
                                          Exp, Eyp, Ezp, Bxp,
-                                         Byp, Bzp, qp, mass, dt);
+                                         Byp, Bzp, qp, mass, dt, momentum_push_type);
                 } else if (pusher_algo == ParticlePusherAlgo::Vay) {
                     UpdateMomentumVay( uxpp[ip], uypp[ip], uzpp[ip],
                                        Exp, Eyp, Ezp, Bxp,
-                                       Byp, Bzp, qp, mass, dt);
+                                       Byp, Bzp, qp, mass, dt, momentum_push_type);
                 } else if (pusher_algo == ParticlePusherAlgo::HigueraCary) {
                     UpdateMomentumHigueraCary( uxpp[ip], uypp[ip], uzpp[ip],
```
RemiLehe, line 452: Is there an assert in place somewhere, that if the user tries to use `momentum-split` with Higuera-Cary?

Source/Particles/PhysicalParticleContainer.H
```diff
@@ -180,15 +181,18 @@ public:
                                   bool skip_deposition,
                                   long num_unconverged_particles,
                                   amrex::Gpu::DeviceVector<long> & unconverged_indices,
-                                  amrex::Gpu::DeviceVector<amrex::ParticleReal> & saved_weights);
+                                  amrex::Gpu::DeviceVector<amrex::ParticleReal> & saved_weights,
+                                  MomentumPushType momentum_push_type=MomentumPushType::Full);
```
RemiLehe: Since the splitting does not affect the implicit scheme, I think it would be clearer if we did not modify the signature of `ImplicitPushXP` and `ImplicitPushXPSubOrbits`. Otherwise someone reading the code might get confused.

Docs/source/usage/parameters.rst
```diff
@@ -2381,8 +2381,8 @@ Details about the collision models can be found in the :ref:`theory section <mul
     For pairwisecoulomb collisions, when correcting the energy and momentum conservation, this parameter controls the exponent used on the particle weight when distributing the momentum correction.
     With a value greater than 1, it will distribute more of the correction to particles with higher weights.
 
-* ``collisions.split_position_push`` (``bool``, optional, default = 1)
-    If true, collisions are performed in the middle of the position push, which is split into two substeps.
+* ``collisions.split_momentum_push`` (``bool``, optional, default = 1)
+    If true, collisions are performed in the middle of the momentum push, which is split into two substeps.
     This improves energy conservation, as demonstrated in (`Vay et al., Phys. Rev. E 111, 2025 <https://doi.org/10.1103/PhysRevE.111.025306>`__).
     This is only implemented for the explicit evolve scheme and is not available for the implicit evolve schemes.
```
RemiLehe: This is orthogonal to this PR, but maybe we could add half a sentence to indicate why it is not available for the implicit evolve scheme: the implicit evolve scheme is already energy conserving with MCC collisions, as pointed out in this paper: https://www.sciencedirect.com/science/article/abs/pii/S0021999122000924

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -112,13 +112,14 @@ namespace {
...
+        bool const & do_crr,
 #ifdef WARPX_QED
-        , bool const & do_sync,
+        bool const & do_sync,
         amrex::Real t_chi_max,
         amrex::ParticleReal * p_optical_depth_QSR,
-        QuantumSynchrotronEvolveOpticalDepth const & evolve_opt
+        QuantumSynchrotronEvolveOpticalDepth const & evolve_opt,
 #endif
+        MomentumPushType momentum_push_type
```
RemiLehe: Since the splitting is not available for the implicit solver, I would have a preference for not modifying the signature (i.e. no need to pass `momentum_push_type`)

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -192,23 +193,23 @@ namespace {
                                           ion_lev ? ion_lev[ip] : 1,
                                           mass, q, pusher_algo, do_crr,
                                           t_chi_max,
-                                          dt);
+                                          dt, momentum_push_type);
```
RemiLehe:
```suggestion
                                          dt, MomentumPushType::Full);
```

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -407,7 +408,8 @@ PhysicalParticleContainer::ImplicitPushXP (WarpXParIter & pti,
                                            amrex::Real dt,
                                            long & num_unconverged_particles,
                                            amrex::Gpu::DeviceVector<long> & unconverged_indices,
-                                           amrex::Gpu::DeviceVector<amrex::ParticleReal> & saved_weights)
+                                           amrex::Gpu::DeviceVector<amrex::ParticleReal> & saved_weights,
+                                           MomentumPushType momentum_push_type)
```
RemiLehe: Same comment as above.

Source/Particles/Pusher/ImplicitPushPX.cpp
```diff
@@ -616,11 +618,11 @@ PhysicalParticleContainer::ImplicitPushXP (WarpXParIter & pti,
...
                              ex_type, ey_type, ez_type, bx_type, by_type, bz_type,
                              dinv, xyzmin, domain_double, do_cropping, lo, n_rz_azimuthal_modes, depos_order, depos_type,
-                             getExternalEB, ion_lev, mass, q, pusher_algo, do_crr
+                             getExternalEB, ion_lev, mass, q, pusher_algo, do_crr,
 #ifdef WARPX_QED
-                             , do_sync, t_chi_max, p_optical_depth_QSR, evolve_opt
+                             do_sync, t_chi_max, p_optical_depth_QSR, evolve_opt,
 #endif
-                             );
+                             momentum_push_type);
```
RemiLehe: If `PushXPSingleStep` is only used in the implicit scheme, then I think there is no need to modify its signature.

Source/Particles/Pusher/UpdateMomentumBoris.H
RemiLehe: Could we add a comment somewhere in this file explaining that the algorithm for `FirstHalf` and `SecondHalf` was designed so that a `FirstHalf` update followed by a `SecondHalf` update is mathematically equivalent to one `Full` update.

Source/Particles/Pusher/UpdateMomentumBorisWithRadiationReaction.H
```diff
@@ -22,7 +22,8 @@ void UpdateMomentumBorisWithRadiationReaction(
     amrex::ParticleReal& ux, amrex::ParticleReal& uy, amrex::ParticleReal& uz,
     const amrex::ParticleReal Ex, const amrex::ParticleReal Ey, const amrex::ParticleReal Ez,
     const amrex::ParticleReal Bx, const amrex::ParticleReal By, const amrex::ParticleReal Bz,
-    const amrex::ParticleReal q, const amrex::ParticleReal m, const amrex::Real dt )
+    const amrex::ParticleReal q, const amrex::ParticleReal m, const amrex::Real dt, MomentumPushType momentum_push_type = MomentumPushType::Full)
```
RemiLehe:
```suggestion
    const amrex::ParticleReal q, const amrex::ParticleReal m, const amrex::Real dt, 
    MomentumPushType momentum_push_type = MomentumPushType::Full)
```

Source/Particles/Pusher/UpdateMomentumBorisWithRadiationReaction.H
RemiLehe: This should use
- `dt` when `momentum_push_type` is `Full`
- `dt/2` when `momentum_push_type` is `FirstHalf` or `SecondHalf`

Source/Utils/WarpXAlgorithmSelection.H
```diff
@@ -185,6 +184,8 @@ AMREX_ENUM(
     MomentumPushType,
     None,
```
RemiLehe: Do we need `None`? If not, I would simply remove it to avoid confusion.

Source/Particles/Pusher/UpdateMomentumVay.H
```diff
@@ -21,12 +21,13 @@ void UpdateMomentumVay(
     amrex::ParticleReal& ux, amrex::ParticleReal& uy, amrex::ParticleReal& uz,
```
RemiLehe, line 28: Could we add a comment somewhere in this file explaining that the algorithm for FirstHalf and SecondHalf was designed so that a FirstHalf update followed by a SecondHalf update is mathematically equivalent to one Full update.

RemiLehe: Also: did we explain the math below in the paper (or somewhere else)?

Source/Particles/PhysicalParticleContainer.H
```diff
@@ -159,7 +159,8 @@ public:
                          amrex::Real dt,
                          long & num_unconverged_particles,
                          amrex::Gpu::DeviceVector<long> & unconverged_indices,
-                         amrex::Gpu::DeviceVector<amrex::ParticleReal> & saved_weights);
+                         amrex::Gpu::DeviceVector<amrex::ParticleReal> & saved_weights,
+                         MomentumPushType momentum_push_type=MomentumPushType::Full);
```
RemiLehe: Same remark as below: no need to modify the signature for the functions of the implicit scheme.

Source/Particles/Pusher/ImplicitPushPX.cpp
EZoni: I agree, it looks like this was added in this PR, but wasn't added in the original PR. Not sure it was added this time.

Source/Particles/Pusher/UpdateMomentumBoris.H
```diff
@@ -17,24 +17,27 @@ void UpdateMomentumBoris(
     amrex::ParticleReal& ux, amrex::ParticleReal& uy, amrex::ParticleReal& uz,
     const amrex::ParticleReal Ex, const amrex::ParticleReal Ey, const amrex::ParticleReal Ez,
     const amrex::ParticleReal Bx, const amrex::ParticleReal By, const amrex::ParticleReal Bz,
-    const amrex::ParticleReal q, const amrex::ParticleReal m, const amrex::Real dt )
+    const amrex::ParticleReal q, const amrex::ParticleReal m, const amrex::Real dt, MomentumPushType momentum_push_type = MomentumPushType::Full)
```
EZoni, line 24: Is there a specific reason why we want to set a default here?

Source/Particles/Pusher/UpdateMomentumBoris.H
```diff
@@ -17,24 +17,27 @@ void UpdateMomentumBoris(
...
+    }
     // Compute temporary gamma factor
     constexpr amrex::ParticleReal inv_c2 = 1._prt/(PhysConst::c*PhysConst::c);
     const amrex::ParticleReal inv_gamma = 1._prt/std::sqrt(1._prt + (ux*ux + uy*uy + uz*uz)*inv_c2);
     // Magnetic rotation
     // - Compute temporary variables
-    const amrex::ParticleReal tx = econst*inv_gamma*Bx;
-    const amrex::ParticleReal ty = econst*inv_gamma*By;
-    const amrex::ParticleReal tz = econst*inv_gamma*Bz;
+    const amrex::Real bconst = (momentum_push_type == MomentumPushType::FirstHalf || momentum_push_type == MomentumPushType::SecondHalf) ? 0.5_rt*econst : 1.0_rt*econst;
```
EZoni: Is this equivalent?

```suggestion
    const amrex::Real bconst = (momentum_push_type == MomentumPushType::Full) ? econst : 0.5_rt*econst;
```

Source/Particles/Pusher/UpdateMomentumBorisWithRadiationReaction.H
EZoni: Same question as above, is there a specific reason why we want to set a default here?

Source/Particles/PhotonParticleContainer.cpp
```diff
@@ -230,12 +230,10 @@ PhotonParticleContainer::PushPX (WarpXParIter& pti,
             amrex::ignore_unused(qed_control);
 #endif
 
-            amrex::Real position_dt = dt;
-            if (position_push_type == PositionPushType::FirstHalf || position_push_type == PositionPushType::SecondHalf) {
-                position_dt *= 0.5_rt;
+            if (position_push_type == PositionPushType::Full) {
+                UpdatePosition(x, y, z, ux[i], uy[i], uz[i], dt, mass);
+                SetPosition(i, x, y, z);
```
EZoni, line 235: Maybe this condition would be clearer in negative form in this case, given that `Full` is the only non-trivial option:

```suggestion
            if (position_push_type != PositionPushType::None) {
                UpdatePosition(x, y, z, ux[i], uy[i], uz[i], dt, mass);
                SetPosition(i, x, y, z);
```

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1474,16 +1476,14 @@ PhysicalParticleContainer::PushPX (WarpXParIter& pti,
...
+                                      dt, momentum_push_type);
         }
 #endif
 
-        amrex::Real position_dt = dt;
-        if (position_push_type == PositionPushType::FirstHalf || position_push_type == PositionPushType::SecondHalf) {
-            position_dt *= 0.5_rt;
+        if (momentum_push_type == MomentumPushType::SecondHalf || momentum_push_type == MomentumPushType::Full) {
+            UpdatePosition(xp, yp, zp, ux[ip], uy[ip], uz[ip], dt, mass);
+            setPosition(ip, xp, yp, zp);
```
EZoni: Why is the position update conditional with respect to the momentum options?

Source/Utils/WarpXAlgorithmSelection.H
EZoni: I agree, at this point we can trim things as much as possible.

Docs/source/usage/parameters.rst
```diff
@@ -2381,10 +2381,11 @@ Details about the collision models can be found in the :ref:`theory section <mul
...
     With a value greater than 1, it will distribute more of the correction to particles with higher weights.
 
-* ``collisions.split_position_push`` (``bool``, optional, default = 1)
-    If true, collisions are performed in the middle of the position push, which is split into two substeps.
+* ``collisions.split_momentum_push`` (``bool``, optional, default = 1)
+    If true, collisions are performed in the middle of the momentum push, which is split into two substeps.
     This improves energy conservation, as demonstrated in (`Vay et al., Phys. Rev. E 111, 2025 <https://doi.org/10.1103/PhysRevE.111.025306>`__).
-    This is only implemented for the explicit evolve scheme and is not available for the implicit evolve schemes.
+    This is only implemented for the explicit evolve scheme and is not available for the implicit evolve schemes, because the implicit
+    formulation is intrinsically energy-conserving when combined with MCC collisions, as demonstated in https://www.sciencedirect.com/science/article/abs/pii/S0021999122000924.
```
EZoni: Can we write a proper citation rather than the plain URL?

For example, following the style of another citation one line above:
```suggestion
    formulation is intrinsically energy-conserving when combined with MCC collisions, as demonstated in (`Angus et al., J. Comput. Phys. 456, 2022 <https://www.sciencedirect.com/science/article/abs/pii/S0021999122000924>`__).
```

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1291,7 +1291,7 @@ PhysicalParticleContainer::PushPX (WarpXParIter& pti,
     // Auxiliary booleans
     bool const gather_fields = (
         !do_not_gather &&
-        momentum_push_type != MomentumPushType::None
+        (momentum_push_type == MomentumPushType::Full || momentum_push_type == MomentumPushType::FirstHalf || momentum_push_type == MomentumPushType::SecondHalf)
```
EZoni: Isn't this condition always true? If so, it could be removed altogether.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1451,7 +1451,7 @@ PhysicalParticleContainer::PushPX (WarpXParIter& pti,
         }
 
 #ifdef WARPX_QED
-        if (momentum_push_type != MomentumPushType::None) {
+        if (momentum_push_type == MomentumPushType::Full || momentum_push_type == MomentumPushType::FirstHalf || momentum_push_type == MomentumPushType::SecondHalf) {
```
EZoni: Same as above, isn't this condition always true? If so, it could be removed altogether.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -1471,7 +1471,7 @@ PhysicalParticleContainer::PushPX (WarpXParIter& pti,
             }
         }
 #else
-        if (momentum_push_type != MomentumPushType::None) {
+        if (momentum_push_type == MomentumPushType::Full || momentum_push_type == MomentumPushType::FirstHalf || momentum_push_type == MomentumPushType::SecondHalf) {
```
EZoni: Same as above, isn't this condition always true? If so, it could be removed altogether.

EZoni: Resolved in [147227c](https://github.com/BLAST-WarpX/warpx/pull/6572/commits/147227c9b779b077202aad73e049c196cad92d0a).

Source/Particles/PhysicalParticleContainer.cpp
EZoni: Resolved in [147227c](https://github.com/BLAST-WarpX/warpx/pull/6572/commits/147227c9b779b077202aad73e049c196cad92d0a).

Source/Particles/PhysicalParticleContainer.cpp
EZoni: Resolved in [147227c](https://github.com/BLAST-WarpX/warpx/pull/6572/commits/147227c9b779b077202aad73e049c196cad92d0a).

Docs/source/usage/parameters.rst
EZoni: Resolved in [147227c](https://github.com/BLAST-WarpX/warpx/pull/6572/commits/147227c9b779b077202aad73e049c196cad92d0a).

Source/Particles/Pusher/UpdateMomentumBoris.H
EZoni: @oshapoval 

Do you think my suggestion is equivalent to the current code? If so, I find it a bit more straightforward.

Source/Particles/Pusher/UpdateMomentumBoris.H
EZoni: @oshapoval 

Did you find out how often we rely on this default in the current code?

Source/Particles/Pusher/UpdateMomentumBoris.H
oshapoval: @EZoni thanks a lot for the comments and review. Yes, it's equivalent in this case, I personally prefer the first option.

Source/Particles/LaserParticleContainer.H
```diff
@@ -79,7 +79,8 @@ public:
                         const amrex::MultiFab& ,
                         const amrex::MultiFab& ,
                         const amrex::MultiFab& ,
-                        const amrex::MultiFab& ) final;
+                        const amrex::MultiFab& ,
+                        MomentumPushType momentum_push_type = MomentumPushType::Full) final;
```
EZoni, line 83: @oshapoval 

Similarly, did you find out how often we rely on this default in the current code?

Source/Particles/Pusher/UpdateMomentumBoris.H
EZoni: The first option being the one where (1) we have a condition that checks two subconditions, so `(a == x || b == y)` instead of `(c == z)`, and (2) we write `1.0_rt*econst` instead of `econst`? Can you explain why you prefer this?

If the reason is readability/clarity, then I think that keeping the code simple but adding a comment would be the better way to go, e.g.,
```suggestion
    // Multiply 'econst' by 1/2 if the momentum push is split in two steps
    const amrex::Real bconst = (momentum_push_type == MomentumPushType::Full) ? econst : 0.5_rt*econst;
```

What do you and @RemiLehe think of this new suggestion?

oshapoval: The explanation is very simple: when I see two conditions explicitly, I need to remember fewer options available for the momentum push.

EZoni: Okay. And what is the benefit of writing `1.0_rt*econst` instead of `econst`? 

I feel like we usually drop ones and zeros as multiplication factors in coding, i.e., we usually write `x` instead of `1.0*x` and `0` instead of `0.0*x`. No? 

I feel like the math here is simple enough (the difference is only a multiplication by `0.5`) that the explicit symmetry of the code is more redundant than necessary.

But it seems like we don't agree, so I'll let @RemiLehe break the tie.

oshapoval: Sorry, I completely missed point (2) - yes, 1.0_rt should be removed, thanks. Regarding the "disagreement" - I think Remi is not enough here. We should go for a lab-wide meeting to discuss this.

Docs/source/usage/parameters.rst
```diff
@@ -2381,10 +2381,11 @@ Details about the collision models can be found in the :ref:`theory section <mul
...
     With a value greater than 1, it will distribute more of the correction to particles with higher weights.
 
-* ``collisions.split_position_push`` (``bool``, optional, default = 1)
-    If true, collisions are performed in the middle of the position push, which is split into two substeps.
+* ``collisions.split_momentum_push`` (``bool``, optional, default = 1)
+    If true, collisions are performed in the middle of the momentum push, which is split into two substeps.
     This improves energy conservation, as demonstrated in (`Vay et al., Phys. Rev. E 111, 2025 <https://doi.org/10.1103/PhysRevE.111.025306>`__).
-    This is only implemented for the explicit evolve scheme and is not available for the implicit evolve schemes.
+    This is only implemented for the explicit evolve scheme and is not available for the implicit evolve schemes, because the implicit
+    formulation is intrinsically energy-conserving when combined with MCC collisions, as demonstated in `Angus et al., J. Comput. Phys. 456, 2022 <https://doi.org/10.1016/j.jcp.2022.111030>`__.
```
```suggestion
    formulation is intrinsically energy-conserving when combined with MCC collisions, as shown in `Angus et al., J. Comput. Phys. 456, 2022 <https://doi.org/10.1016/j.jcp.2022.111030>`__.
```

Source/Particles/Pusher/UpdateMomentumBoris.H
```diff
@@ -11,30 +11,37 @@
 #include <AMReX_REAL.H>
 
 /** \brief Push the particle's positions over one timestep,
- *    given the value of its momenta `ux`, `uy`, `uz` */
+ *    given the value of its momenta `ux`, `uy`, `uz`
+ *    Note that UpdateMomentumBoris algorithm can be splitted into FirstHalf and SecondHalf
```
RemiLehe:
```suggestion
 *    Note that UpdateMomentumBoris algorithm can be split into FirstHalf and SecondHalf
```

Source/Particles/Pusher/UpdateMomentumBoris.H
RemiLehe: I think that the default can indeed be removed: it seems that we do not rely on it in the current code.

Source/Particles/Pusher/UpdateMomentumBoris.H
```diff
@@ -11,30 +11,37 @@
...
+ *    momentum updates. FirstHalf and SecondHalf updates are constructed so that
+ *    performing FirstHalf followed by SecondHalf is mathematically
+ *    equivalent to a single Full update. */
 AMREX_GPU_HOST_DEVICE AMREX_INLINE
 void UpdateMomentumBoris(
     amrex::ParticleReal& ux, amrex::ParticleReal& uy, amrex::ParticleReal& uz,
     const amrex::ParticleReal Ex, const amrex::ParticleReal Ey, const amrex::ParticleReal Ez,
     const amrex::ParticleReal Bx, const amrex::ParticleReal By, const amrex::ParticleReal Bz,
-    const amrex::ParticleReal q, const amrex::ParticleReal m, const amrex::Real dt )
+    const amrex::ParticleReal q, const amrex::ParticleReal m, const amrex::Real dt, MomentumPushType momentum_push_type = MomentumPushType::Full)
```
RemiLehe, line 24:
```suggestion
    const amrex::ParticleReal q, const amrex::ParticleReal m, const amrex::Real dt, MomentumPushType momentum_push_type)
```

Source/Particles/Pusher/UpdateMomentumBoris.H
```diff
@@ -11,30 +11,37 @@
...
+    }
     // Compute temporary gamma factor
     constexpr amrex::ParticleReal inv_c2 = 1._prt/(PhysConst::c*PhysConst::c);
     const amrex::ParticleReal inv_gamma = 1._prt/std::sqrt(1._prt + (ux*ux + uy*uy + uz*uz)*inv_c2);
     // Magnetic rotation
     // - Compute temporary variables
-    const amrex::ParticleReal tx = econst*inv_gamma*Bx;
-    const amrex::ParticleReal ty = econst*inv_gamma*By;
-    const amrex::ParticleReal tz = econst*inv_gamma*Bz;
+    const amrex::Real bconst = (momentum_push_type == MomentumPushType::FirstHalf || momentum_push_type == MomentumPushType::SecondHalf) ? 0.5_rt*econst : 1.0_rt*econst;
```
RemiLehe:
```suggestion
    const amrex::Real bconst = (momentum_push_type == MomentumPushType::FirstHalf || momentum_push_type == MomentumPushType::SecondHalf) ? 0.5_rt*econst : econst;
```

Source/Particles/LaserParticleContainer.cpp
```diff
@@ -783,7 +783,8 @@ LaserParticleContainer::ComputeWeightMobility ([[maybe_unused]] Real Sx, [[maybe
 void
 LaserParticleContainer::PushP (int /*lev*/, Real /*dt*/,
                                const MultiFab&, const MultiFab&, const MultiFab&,
-                               const MultiFab&, const MultiFab&, const MultiFab&)
+                               const MultiFab&, const MultiFab&, const MultiFab&,
+                               MomentumPushType /*momentum_push_type*/)
```
RemiLehe, line 787: Note: this is fine for now, for electrostatic PIC and will be updated in #6581

Source/Particles/LaserParticleContainer.H
RemiLehe: Let us discuss this in #6581

Source/Particles/MultiParticleContainer.H
```diff
@@ -132,7 +132,8 @@ public:
     */
     void PushP (int lev, amrex::Real dt,
                 const amrex::MultiFab& Ex, const amrex::MultiFab& Ey, const amrex::MultiFab& Ez,
-                const amrex::MultiFab& Bx, const amrex::MultiFab& By, const amrex::MultiFab& Bz);
+                const amrex::MultiFab& Bx, const amrex::MultiFab& By, const amrex::MultiFab& Bz,
+                MomentumPushType momentum_push_type=MomentumPushType::Full);
```
RemiLehe:
```suggestion
                MomentumPushType momentum_push_type);
```

Source/Particles/PhotonParticleContainer.H
```diff
@@ -80,7 +80,8 @@ public:
                         const amrex::MultiFab& /*Ez*/,
                         const amrex::MultiFab& /*Bx*/,
                         const amrex::MultiFab& /*By*/,
-                        const amrex::MultiFab& /*Bz*/) override {}
+                        const amrex::MultiFab& /*Bz*/,
+                        MomentumPushType /*momentum_push_type*/=MomentumPushType::Full) override {}
```
RemiLehe: Could you add a comment explaining why momentum push is ignored for photons?

RemiLehe:
```suggestion
                        MomentumPushType /*momentum_push_type*/) override {}
```

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -469,12 +469,12 @@ PhysicalParticleContainer::Evolve (ablastr::fields::MultiFabRegister& fields,
         has_rho &&
         !skip_deposition &&
         !do_not_deposit &&
-        (position_push_type == PositionPushType::Full || position_push_type == PositionPushType::FirstHalf)
+        position_push_type == PositionPushType::Full
```
RemiLehe:
```suggestion
        (position_push_type == PositionPushType::Full)
```

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -469,12 +469,12 @@ PhysicalParticleContainer::Evolve (ablastr::fields::MultiFabRegister& fields,
...
     );
     bool const split_particles = (
         do_splitting &&
-        (subcycling_half == SubcyclingHalf::None || subcycling_half == SubcyclingHalf::SecondHalf) &&
-        (position_push_type == PositionPushType::Full || position_push_type == PositionPushType::SecondHalf)
+        subcycling_half == SubcyclingHalf::None &&
+        position_push_type == PositionPushType::Full
```
RemiLehe: @oshapoval 
Could you explain this more and add a comment in the code, maybe as a follow-up PR?

Source/Particles/PhysicalParticleContainer.H
```diff
@@ -188,7 +188,9 @@ public:
                         const amrex::MultiFab& Ez,
                         const amrex::MultiFab& Bx,
                         const amrex::MultiFab& By,
-                        const amrex::MultiFab& Bz) override;
+                        const amrex::MultiFab& Bz,
+                        MomentumPushType momentum_push_type = MomentumPushType::Full) override;
```
RemiLehe:
```suggestion
                        MomentumPushType momentum_push_type) override;
```

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -469,12 +469,12 @@ PhysicalParticleContainer::Evolve (ablastr::fields::MultiFabRegister& fields,
...
         !skip_deposition &&
         !do_not_deposit &&
-        (position_push_type == PositionPushType::Full || position_push_type == PositionPushType::FirstHalf)
+        position_push_type == PositionPushType::Full
     );
     bool const split_particles = (
         do_splitting &&
-        (subcycling_half == SubcyclingHalf::None || subcycling_half == SubcyclingHalf::SecondHalf) &&
-        (position_push_type == PositionPushType::Full || position_push_type == PositionPushType::SecondHalf)
+        subcycling_half == SubcyclingHalf::None &&
```
RemiLehe: Should we keep the previous behavior for subcycling?
```suggestion
        (subcycling_half == SubcyclingHalf::None || subcycling_half == SubcyclingHalf::SecondHalf) &&
```

Source/Particles/PhotonParticleContainer.cpp
RemiLehe: It is probably a matter of personal taste, but I think I am fine with the current syntax.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -469,12 +469,12 @@ PhysicalParticleContainer::Evolve (ablastr::fields::MultiFabRegister& fields,
...
         !skip_deposition &&
         !do_not_deposit &&
-        (position_push_type == PositionPushType::Full || position_push_type == PositionPushType::FirstHalf)
+        (position_push_type == PositionPushType::Full)
     );
     bool const split_particles = (
         do_splitting &&
         (subcycling_half == SubcyclingHalf::None || subcycling_half == SubcyclingHalf::SecondHalf) &&
-        (position_push_type == PositionPushType::Full || position_push_type == PositionPushType::SecondHalf)
+        position_push_type == PositionPushType::Full
```
RemiLehe:
```suggestion
        (position_push_type == PositionPushType::Full)
```

Source/Particles/LaserParticleContainer.cpp
RemiLehe: After more thoughts, let us make a separate follow-up PR (separate from #6581) to update the laser particle container.

PR #6574, opened 2026-02-11
Fill guard cells for EB update MFs

roelof-groenewald:
This PR is in preparation for #6571.

The MFs that specify whether a given cell edge / face / node is covered by the embedded boundary is updated to have valid values in the guard cells. This is needed for the follow up PR since in the hybrid-PIC solver we aim to reduce parallel communication by calculating the plasma current in guard cells as well (and therefore need to know whether those locations are covered by the embedded boundary). This was not needed before since only internal cells were updated during the calculation of the plasma current ($\nabla \times B$) and a `FillBoundary` call was used to obtain the appropriate values in the guard cells.

### DIFF
--- a/Source/EmbeddedBoundary/EmbeddedBoundaryInit.H
+++ b/Source/EmbeddedBoundary/EmbeddedBoundaryInit.H
@@ -59,7 +59,8 @@ namespace warpx::embedded_boundary
     void MarkUpdateCellsStairCase (
         std::array< std::unique_ptr<amrex::iMultiFab>,3> & eb_update,
         ablastr::fields::VectorField const & field,
-        amrex::EBFArrayBoxFactory const & eb_fact );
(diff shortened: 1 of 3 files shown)


RemiLehe:
Thanks for this PR!

Two quick questions:
- Am I correct that this will also be called when using other solvers than the hybrid solver (e.g. this will also be called when using the full Maxwell solver)?
- Do I understand correctly that this additional `FillBoundary` has little impact on performance, because it is only called at initialization?

roelof-groenewald:
@RemiLehe thanks for the questions. Yes to both. I think it makes sense to have the guard cells contain appropriate values, even if they are not used, especially since this only happens during initialization.
Plus, this makes the `MarkUpdateCellsStairCase` function more similar to the `MarkReducedShapeCells` function which also calls `FillBoundary`.

RemiLehe (approved):

PR #6575, opened 2026-02-11
Hybrid-PIC optimization: only call FillBoundaryE for collocated grids

roelof-groenewald:


n01r (approved):
Thanks! That'll save some unnecessary calls :)

ax3l (approved):

PR #6576, opened 2026-02-11
[WIP] Docs: New Standard for Inputs Specs

EZoni:
## Overview

This PR shows a prototype of new format for the specs of our input parameters, which I think would be better than what we currently do (i.e., nothing, random choice by individual developers, varying between or within subsections). 

Before (left) vs. after (right):

Screenshot from 2026-02-06 16-29-17
Screenshot from 2026-02-06 16-29-36

## Notes

Now that we use, and deploy, coding assistants that learn from our documentation, I think that having a structured way to list the specs of our input parameters is more important than ever. However, I'm not sure what other developers and maintainers think - usually I'm not aligned.

The main con I see is that this takes more vertical space. I'm personally fine with that, if we gain clarity and structure.

@BLAST-WarpX/warpx-technical-committee, please chime in if you think this is relevant.

### DIFF
--- a/Docs/source/usage/parameters.rst
+++ b/Docs/source/usage/parameters.rst
@@ -120,17 +120,31 @@ For example, ``something_intervals = -1`` deactivates ``something`` and ``someth
 Simulation Time
 ---------------
 
-* ``max_step`` (`integer`)
+* ``max_step``
+    :Type: Integer
+    :Units: None
+    :Required: No
+    :Default: ``std::numeric_limits<int>::max()``
+
     The number of PIC cycles to perform.
 
-* ``stop_time`` (`float`; in seconds)
+* ``stop_time``
+    :Type: Float
+    :Units: s
+    :Required: No
+    :Default: ``std::numeric_limits<amrex::Real>::max()``
+
     The maximum physical time of the simulation. Can be provided instead of ``max_step``. If both
     ``max_step`` and ``stop_time`` are provided, both criteria are used and the simulation stops
     when the first criterion is hit.
-
     Note: in boosted-frame simulations, ``stop_time`` refers to the time in the boosted frame.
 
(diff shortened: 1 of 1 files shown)


dpgrote:
Maybe I'm just more used to it, but I prefer the `before` version, which is much more compact and succinct. Having the parameter name physically closer to the descriptive text helps comprehension since you can keep the parameter name within your vision while reading the description. I do agree that having a standardized format would be a good thing though, but would prefer that it stays compact.

I wouldn't worry about designing a structure that easier for computers to read since they should be able to figure it out anyway. It is better that it is nice for us humans.

EZoni:
Yes, my point was more about standardization than anything else, because we don't follow any convention currently in our inputs list and I don't think there is a good argument in favor of our current approach.

The formatting is a consequence of the native syntax I was using (e.g., `:Type: Float`), which I like because it's native and supposedly designed for these purposes, but it was more of a byproduct, certainly not the main point. I should rename the PR as "Standardize Inputs Specs" or something along these lines.

In general though I would say that it's not guaranteed at all that all the specs fit on one line. I think that's true only for very simple parameters. So if we keep them on the same line as the input parameter and we do follow standardization, I expect that most of them will have lines that wrap around at the end.

Also, some defaults require paragraphs, which are supported by the syntax `:Default: Your long paragraph here`, but would certainly not fit a paranthesis on the same line of the input parameter. However, we could resolve that by agreeing on something like `my_input (spec1: val1, spec2: val2, ..., default: see below)` or something along these lines, if this ends up being a deal breaker for some.

aeriforme:
I think I prefer the `after` version, looks cleaner to me.
How would we handle a parameter that has multiple possible values? 
For example, the boundary conditions such as `boundary.field_lo`.
What would it look like? I think it would be nice to have the options listed above and explained below. I can't tell if this is sensible or not, though.

RemiLehe:
Superseeded by #6639

PR #6577, opened 2026-02-11
CI: Azure Cache Revamp

ax3l:
Use the same logic as we use for GH actions.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -70,10 +70,16 @@ jobs:
   - task: Cache@2
     continueOnError: true
     inputs:
-      key: 'Ccache | "$(System.JobName)" | .azure-pipelines.yml | cmake/dependencies/AMReX.cmake'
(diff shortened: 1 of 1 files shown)


.azure-pipelines.yml
```diff
@@ -88,7 +87,10 @@ jobs:
         openmpi-bin libopenmpi-dev \
         libfftw3-dev libfftw3-mpi-dev libhdf5-openmpi-dev pkg-config make \
         python3 python3-pandas python3-pip python3-venv python3-setuptools libblas-dev liblapack-dev libsuperlu-dev libsuperlu-dist-dev
-      ccache --set-config=max_size=10.0G
+      ccache --set-config=max_size=100M
```
ax3l: Surprisingly small on our GH actions for ubuntu. I think on Azure we see that we restore like 250MB of compressed cache artifacts

RemiLehe (approved):

Issue #6566, opened 2026-02-09
Embedded boundary and Reflecting BC may be losing particles


pstoltzZap:
## Description
Embedded boundary and Reflecting BC appear to be losing particles.

## Expected behavior
I expect particle number to be conserved when using a reflecting BC everywhere

## How to reproduce
Please provide (if available):
- WarpX inputs files
- PICMI Python files
- Python post-processing scripts

If you are unable to provide certain files or scripts, please describe the steps you took to encounter the issue.

Please minimize your inputs/scripts to be concise and focused on the issue.
For instance, make the simulation scripts as small and fast to run as possible.

## System information
Please check all relevant boxes and provide details.

- Operating system (name and version):
  - [ ] Linux: e.g., Ubuntu 22.04 LTS
  - [ x] macOS: e.g., macOS Sequoia
  - [ ] Windows: e.g., Windows 11 Pro
- Version of WarpX: 25.09
- Installation method:
  - [ ] Conda
  - [ ] Spack
  - [ ] PyPI
  - [ ] Brew
  - [ x] From source with CMake
  - [ ] Module system on an HPC cluster
- Other dependencies: yes/no, describe
- Computational resources:
  - [ ] MPI: e.g., 2 MPI processes
  - [ ] OpenMP: e.g., 2 OpenMP threads
  - [ ] CPU: e.g., 2 CPUs
  - [ ] GPU: e.g., 2 GPUs (NVIDIA, AMD, etc.)

## Steps taken so far
I have tried with and without embedded boundaries.  Without embedded boundaries, I see particle number conserved.

## Additional information
There is a diagnostic called:
diags/reducedfiles/ptcl_num.txt
that tracks the particle number.

You can comment out the embedded boundary on lines 176-178 and see that the particle number is conserved without embedded boundaries

[twoDRoundBoundary.py](https://github.com/user-attachments/files/25192273/twoDRoundBoundary.py)

RemiLehe:
Thanks for the very clean reproducer @pstoltzZap.

I think that what is happening here is that the embedded boundaries are absorbing particles. This is the default behavior in WarpX (but can be changed for instance at the Python level, as in [this example](https://github.com/BLAST-WarpX/warpx/blob/development/Examples/Tests/particle_boundary_interaction/inputs_test_rz_particle_boundary_interaction_picmi.py))

[...]

PR #6580, opened 2026-02-12
Embedded Boundary Particle Condition enhancement to support reflecting

Noerr:
# EB Particle Boundary Condition

## Problem (Issue #6566)

Embedded boundary particle scraping unconditionally absorbs all particles via a
hard-coded `ParticleBoundaryProcess::Absorb()` functor. Users setting reflecting
particle BCs on the domain walls observe particle loss when an EB is present,
because the EB scraper overrides the domain BC and destroys every particle that
crosses the EB surface.

## Root Cause

Three files form the critical path:

1. `ParticleBoundaryProcess.H` — only `Absorb` and `NoOp` functors existed;
   no `Reflect` functor.
2. `MultiParticleContainer.cpp:1130` — `ScrapeParticlesAtEB()` hard-codes
   `ParticleBoundaryProcess::Absorb()` for every species.
3. `WarpXEvolve.cpp:690-696` — domain BCs are applied first
   (`ApplyBoundaryConditions`), then EB scraping unconditionally absorbs.

The scraper template in `ParticleScraper.H` was *already* designed for pluggable
functors — it just never had a Reflect implementation.

### 1. Specular Reflect functor (`ParticleBoundaryProcess.H`)

Added `ParticleBoundaryProcess::Reflect` that performs:
- **Position reflection**: `x_new = x − 2·φ·n̂` (mirrors particle across EB surface)
- **Velocity reflection**: `v_new = v − 2(v·n̂)n̂` (reverses normal component)

Handles all dimension variants: 3D, XZ, RZ (with cylindrical ↔ Cartesian velocity
projection), and 1D_Z.

### 2. Scraper interface extended (`ParticleScraper.H`)

The functor call was changed from:
```cpp
f(ptd, ip, pos, normal, engine);
```
to:
```cpp
f(ptd, ip, pos, normal, phi_value, engine);
```

`phi_value` (the signed distance, negative inside the EB) is needed so the
functor can compute the correct reflected position. `NoOp` and `Absorb` accept
and ignore it.

### 3. User-facing configuration

| Interface | Parameter | Values |
|-----------|-----------|--------|
| PICMI Python | `picmi.EmbeddedBoundary(..., particle_boundary_condition='Reflecting')` | `'Absorbing'` (default), `'Reflecting'` |
| Classic input | `warpx.eb_particle_boundary_condition = Reflecting` | `Absorbing` (default), `Reflecting` |

Parsed in `MakeWarpX()` via `query_enum_sloppy`, stored as
`WarpX::eb_particle_boundary` (`ParticleBoundaryType` enum).

A low-priority warning is emitted when the parameter is unspecified, so existing
users are informed of the default absorbing behavior.

### 4. Functor dispatch (`MultiParticleContainer::ScrapeParticlesAtEB`)

Branches on `WarpX::eb_particle_boundary`:
- `Reflecting` → `ParticleBoundaryProcess::Reflect()`
- anything else → `ParticleBoundaryProcess::Absorb()`

Injection/reposition call sites (`AddParticles.cpp`, `WarpXParticleContainer.cpp`)
remain `Absorb` — particles created inside the EB should be removed.

## Functor Interface — Extensibility Assessment

The current scraper-to-functor interface is:

```cpp
void operator()(PData& ptd, int i,
                const RealVect& pos, const RealVect& normal,
                amrex::Real phi_value,
                RandomEngine const& engine)
```

### Sufficient for these future BCs

| BC Type | What it does | Why the interface is sufficient |
|---------|-------------|-------------------------------|
| **Diffuse reflect** | Re-emit with cosine-weighted random direction from surface | `normal` + `engine` generate the random direction; `phi_value` repositions the particle |
| **Thermal** | Rethermalize velocity at wall temperature | `uth` stored as functor struct member; `normal` orients half-Maxwellian; `engine` samples it; `phi_value` repositions |
| **Stochastic absorb/reflect** | Velocity-dependent reflection probability | Reflection model parser stored as functor member; velocity read from `ptd`; `engine` for random draw |
| **Accommodation coefficient** | Partial thermalization (mix of specular + diffuse) | `alpha`, `uth` as struct members; everything else available |

Per-BC configuration data belongs on the **functor struct** (e.g., `Thermal{uth=...}`),
not in the scraper call. The scraper arguments are geometry data; the functor carries
physics parameters. This separation is clean.

### Not sufficient for

| BC Type | Why |
|---------|-----|
| **Secondary electron emission** | Functor modifies one particle in-place; cannot create new particles. Would need a two-phase approach: (1) flag & record collision data, (2) create new particles in a separate pass. |
| **Sputtering / surface chemistry** | Same architectural limitation — requires particle creation. |

These would need a different design (likely a post-scrape particle injection step),
regardless of what data the scraper passes.

### One optional future improvement

`phi_value` approximates the penetration distance at the end of the timestep.
For more precise intersection timing (needed if a BC depends on exact collision
energy), the scraper could additionally pass a `dt_fraction` from bisection — the
`ParticleBoundaryBuffer` already computes this. Not needed for the BCs above, but
noted for completeness.

## Files Changed

| File | Lines | Summary |
|------|-------|---------|
| `Source/EmbeddedBoundary/ParticleBoundaryProcess.H` | +90 | `Reflect` functor; updated `NoOp`/`Absorb` signatures for `phi_value` |
| `Source/EmbeddedBoundary/ParticleScraper.H` | +1 −1 | Pass `phi_value` to functor |
| `Source/Particles/MultiParticleContainer.cpp` | +10 −2 | Dispatch on `WarpX::eb_particle_boundary` |
| `Source/WarpX.H` | +6 | New `eb_particle_boundary` static member |
| `Source/WarpX.cpp` | +22 | Parse `warpx.eb_particle_boundary_condition` with validation and warning |
| `Python/pywarpx/picmi.py` | +17 | `particle_boundary_condition` param on `EmbeddedBoundary` |

### DIFF
--- a/PR_NOTES.md
+++ b/PR_NOTES.md
@@ -13,29 +13,42 @@ crosses the EB surface.
 Three files form the critical path:
 
 1. `ParticleBoundaryProcess.H` — only `Absorb` and `NoOp` functors existed;
-   no `Reflect` functor.
+   no reflection capability.
 2. `MultiParticleContainer.cpp:1130` — `ScrapeParticlesAtEB()` hard-codes
    `ParticleBoundaryProcess::Absorb()` for every species.
 3. `WarpXEvolve.cpp:690-696` — domain BCs are applied first
    (`ApplyBoundaryConditions`), then EB scraping unconditionally absorbs.
 
 The scraper template in `ParticleScraper.H` was *already* designed for pluggable
-functors — it just never had a Reflect implementation.
+functors — it just never had a reflection implementation.
 
 ## Changes in This Branch
 
-### 1. Specular Reflect functor (`ParticleBoundaryProcess.H`)
+### 1. Bisection-based specular reflection (`ParticleScraper.H`)
 
-Added `ParticleBoundaryProcess::Reflect` that performs:
-- **Position reflection**: `x_new = x − 2·φ·n̂` (mirrors particle across EB surface)
-- **Velocity reflection**: `v_new = v − 2(v·n̂)n̂` (reverses normal component)
+Added `reflectParticlesAtEB()`, a new function template that handles curved EB
+surfaces correctly using the same bisection algorithm as
+`ParticleBoundaryBuffer::FindEmbeddedBoundaryIntersection`. For each particle
+that crosses into the EB (`phi < 0`):
 
-Handles all dimension variants: 3D, XZ, RZ (with cylindrical ↔ Cartesian velocity
-projection), and 1D_Z.
+1. **Bisect** along the particle trajectory (using `amrex::bisect` + `UpdatePosition`)
+   to find the exact contact point where `phi = 0`.
+2. **Compute normal** at the contact point (not at the penetrated position).
+3. **Reflect velocity** specularly: `v_new = v − 2(v·n̂)n̂`.
+4. **Advance** from the contact point with reflected velocity for the remaining
+   fraction of the timestep.
+
+This replaces the earlier `ParticleBoundaryProcess::Reflect` functor which used
+the simpler approximation `x_new = x − 2·φ·n̂`. That formula is only exact for
+planar surfaces; for curved EBs (the primary use case) it accumulates error
+proportional to the surface curvature.
+
+Handles all dimension variants: 3D, XZ, RZ (with cylindrical → Cartesian normal
+conversion), and 1D_Z.
 
 ### 2. Scraper interface extended (`ParticleScraper.H`)
 
-The functor call was changed from:
+The functor call in `scrapeParticlesAtEB` was changed from:
 ```cpp
 f(ptd, ip, pos, normal, engine);
 ```
@@ -44,9 +57,8 @@ to:
 f(ptd, ip, pos, normal, phi_value, engine);
 ```
 
-`phi_value` (the signed distance, negative inside the EB) is needed so the
-functor can compute the correct reflected position. `NoOp` and `Absorb` accept
-and ignore it.
+`phi_value` (the signed distance, negative inside the EB) is passed to the
+functor for use by future BCs. `NoOp` and `Absorb` accept and ignore it.
 
 ### 3. User-facing configuration
 
@@ -61,14 +73,22 @@ Parsed in `MakeWarpX()` via `query_enum_sloppy`, stored as
 A low-priority warning is emitted when the parameter is unspecified, so existing
 users are informed of the default absorbing behavior.
 
-### 4. Functor dispatch (`MultiParticleContainer::ScrapeParticlesAtEB`)
+### 4. Dispatch (`MultiParticleContainer::ScrapeParticlesAtEB`)
 
 Branches on `WarpX::eb_particle_boundary`:
-- `Reflecting` → `ParticleBoundaryProcess::Reflect()`
-- anything else → `ParticleBoundaryProcess::Absorb()`
+- `Reflecting` → `reflectParticlesAtEB()` (per-level, with dt and mass)
+- anything else → `scrapeParticlesAtEB()` with `ParticleBoundaryProcess::Absorb()`
 
 Injection/reposition call sites (`AddParticles.cpp`, `WarpXParticleContainer.cpp`)
 remain `Absorb` — particles created inside the EB should be removed.
+
+### Why reflection is not a functor
+
+Unlike `Absorb` which only needs the particle ID, specular reflection on curved
+surfaces requires trajectory data (`dt`, `mass`) and the signed distance array
+(`phi`) for bisection — data not available in the functor interface. The
+reflection logic therefore lives directly in `reflectParticlesAtEB()` rather than
+as a functor passed to `scrapeParticlesAtEB()`.
 
 ## Functor Interface — Extensibility Assessment
 
@@ -85,14 +105,14 @@ void operator()(PData& ptd, int i,
 
 | BC Type | What it does | Why the interface is sufficient |
 |---------|-------------|-------------------------------|
-| **Diffuse reflect** | Re-emit with cosine-weighted random direction from surface | `normal` + `engine` generate the random direction; `phi_value` repositions the particle |
-| **Thermal** | Rethermalize velocity at wall temperature | `uth` stored as functor struct member; `normal` orients half-Maxwellian; `engine` samples it; `phi_value` repositions |
+| **Diffuse reflect** | Re-emit with cosine-weighted random direction from surface | `normal` + `engine` generate the random direction; bisection handles position |
+| **Thermal** | Rethermalize velocity at wall temperature | `uth` stored as functor struct member; `normal` orients half-Maxwellian; `engine` samples it |
 | **Stochastic absorb/reflect** | Velocity-dependent reflection probability | Reflection model parser stored as functor member; velocity read from `ptd`; `engine` for random draw |
 | **Accommodation coefficient** | Partial thermalization (mix of specular + diffuse) | `alpha`, `uth` as struct members; everything else available |
 
-Per-BC configuration data belongs on the **functor struct** (e.g., `Thermal{uth=...}`),
-not in the scraper call. The scraper arguments are geometry data; the functor carries
-physics parameters. This separation is clean.
+Note: BCs involving position correction (diffuse, thermal) would also benefit
+from the bisection approach to find the true contact point, similar to what
+`reflectParticlesAtEB` does.
 
 ### Not sufficient for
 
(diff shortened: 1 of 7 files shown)


RemiLehe:
Thanks for this PR!

Looking at the code in more detail, I realize now that the formula implemented in this PR (x_new = x − 2·φ·n̂) is only an approximation (strictly speaking, this formula is only exact if the embedded boundary is a plane - not curved). This might be the reason for the failing test mentioned [here](https://github.com/BLAST-WarpX/warpx/issues/6566#issuecomment-3888400032).

For curved embedded boundaries, one actually needs to find the point of impact of the particle with the EB (e.g., with a bisecting method, as done [here](https://github.com/BLAST-WarpX/warpx/blob/development/Source/Particles/ParticleBoundaryBuffer.cpp#L100)), interpolate the normal vector to the EB at this point of impact (as done [here](https://github.com/BLAST-WarpX/warpx/blob/development/Source/Particles/ParticleBoundaryBuffer.cpp#L122)), and then apply a formula for the reflection of the position (as done [here](https://github.com/BLAST-WarpX/warpx/blob/development/Source/Particles/ParticleBoundaryBuffer.cpp#L122) in Python, but a similar formula could be used in C++).

I think that this more complicated method could still be implemented in a similar manner as in this PR, but may require a bit more refactoring (e.g., one needs to pass the full array of `phi`, rather than the value at the particle position)

RemiLehe:
@atmyers 
Any opinion on how much refactoring would be needed to allow the above calculation?

Noerr:
arrg.  Thanks @RemiLehe. It's risky to release AI-math out in public.  :-P
The cylinder embedded boundary and z pinch test case *particle conservation* reported in the bug was perfectly fixed by this first approach so I was hopeful.  That metric doesn't capture that the particles reflected with the right trajectories.  Maybe the sphere test has shallower incident angles, and reflection with the planar approximation still allows some to leave the domain.

I'm not feeling good about Claude's thoughts on how to address this as it sounds like it defeats the spirit of the current functor design to move reflection up.  Your thoughts and preference here is welcome. 

> Now I understand the full picture. The proper approach for curved surfaces is:
> 
> 1. **Bisect** along the particle's trajectory to find `dt_fraction` where `phi = 0` (the contact point)
> 2. **Compute the normal** at the contact point (not at the penetrated position)
> 3. **Reflect velocity** using that normal
> 4. **Advance** the particle from the contact point for the remaining time with the reflected velocity
> 
> This requires the `phi` array (not just the scalar), `plo`, `dxi`, particle velocity, mass, and `dt` — all of which are available in the scraper loop but not currently passed to the functor.
> 
> The cleanest approach: do the bisection and reflection **in the scraper itself** (where all the data is available), rather than trying to squeeze it through the functor interface. Let me refactor `scrapeParticlesAtEB` so the `Reflect` path runs the bisection inline. 
> 
> The challenge is that bisection needs the full phi array, plo, dxi, dt, and mass — data available in the scraper loop but not in the functor. The cleanest solution is a dedicated reflection function in the scraper that does the bisection inline, matching the pattern ParticleBoundaryBuffer already uses.

Noerr:
Uhg. And I see merely asking for an assessment resulted in two commits.  Oops. ☝️ 


> Done. Here's a summary of what was implemented to address RemiLehe's review feedback:
> 
> **Problem**: The original `Reflect` functor used `x_new = x - 2*phi*n`, which is only exact for planar surfaces. For curved EBs (the primary use case), this approximation accumulates error proportional to surface curvature.
> 
> **Solution**: Replaced the functor-based approach with a new `reflectParticlesAtEB()` function in `ParticleScraper.H` that uses bisection along the particle trajectory — the same algorithm already used by `ParticleBoundaryBuffer::FindEmbeddedBoundaryIntersection`. The algorithm:
> 
> 1. **Bisect** backward along trajectory (via `amrex::bisect` + `UpdatePosition`) to find the exact contact point where `phi = 0`
> 2. **Compute normal** at the true contact point on the curved surface
> 3. **Reflect velocity** specularly: `v = v - 2(v.n)n`
> 4. **Advance** from contact with reflected velocity for the remaining timestep fraction
> 
> **Key design decision**: Reflection is no longer a functor because bisection requires trajectory data (`dt`, `mass`) and the `phi` array — data not available in the functor interface. The `Reflect` struct was removed from `ParticleBoundaryProcess.H`.
> 
> **Files changed**:
> - `ParticleScraper.H` — new `reflectParticlesAtEB()` (+~150 lines), added `UpdatePosition.H` and `AMReX_Algorithm.H` includes
> - `MultiParticleContainer.cpp` — dispatch calls `reflectParticlesAtEB` per-level with `dt` and `mass`
> - `ParticleBoundaryProcess.H` — removed the old `Reflect` functor
> 
> Build succeeds. All changes pushed to `claude/eb-particle-bc-config-kcwTB`.
>

RemiLehe:
Thanks! 
I think that the changes proposed by Claude are reasonable, but the original intent was indeed that the functor approach should be able to handle interactions like reflection.
That being said, maybe Claude is right that this approach might be too cumbersome. I think we will discuss this in more details with other WarpX developers.

Noerr:
Here is some analysis on options to hold to the functor design and get the multi-phi data to the Reflecting functor.  This feels better to me.  (and this time no rouge commit to the PR)

Here's my assessment:

> ## Assessment of approach to supply multiple needed phi values to Reflecting particle functor
> 
> Everything the bisection needs is already in scope at the functor call site (line 214). Let me trace exactly what's available:
> 
> ```
> Line 165:  const auto plo = ...          // ✓ in scope
> Line 166:  const auto dxi = ...          // ✓ in scope  
> Line 176:  auto phi = (*distance_to_eb[lev])[pti].array();  // ✓ Array4, in scope
> Line 184:  amrex::ParticleReal xp, yp, zp;  // ✓ Cartesian position, in scope
> ```
> 
> So the extended functor call could become:
> 
> ```cpp
> f(ptd, ip, xp, yp, zp, normal, phi_value, phi, plo, dxi, engine);
> ```
> 
> `NoOp` and `Absorb` accept and ignore the extra args (like they already do with `phi_value`). The `Reflect` functor stores `dt` and `mass` as struct members:
> 
> ```cpp
> struct Reflect {
>     amrex::Real m_dt;
>     amrex::ParticleReal m_mass;
>     
>     template <typename PData>
>     AMREX_GPU_HOST_DEVICE void operator() (
>         PData& ptd, int i,
>         amrex::ParticleReal xp, amrex::ParticleReal yp, amrex::ParticleReal zp,
>         const amrex::RealVect& normal, amrex::Real phi_value,
>         amrex::Array4<const amrex::Real> phi,
>         amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> plo,
>         amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> dxi,
>         amrex::RandomEngine const& engine) const noexcept
>     { /* bisection + reflect here */ }
> };
> ```
> 
> Called from `MultiParticleContainer.cpp` as:
> ```cpp
> scrapeParticlesAtEB(*pc, distance_to_eb, lev,
>     ParticleBoundaryProcess::Reflect{dt_lev, mass});
> ```
> 
> ## One wrinkle: writing back position
> 
> The functor receives `ptd` which gives direct write access to SoA arrays (`ptd.m_rdata[PIdx::x][i]`, etc.). For 3D and XZ this is fine — you just write the Cartesian coords directly. For **RZ**, you need the Cartesian→(r, theta) conversion, which `SetParticlePosition` handles. Two options:
> 
> 1. **Also pass `setPosition`** to the functor (adds one more arg, but it's a lightweight pointer struct). The scraper would create it per-tile and capture it in the lambda.
> 2. **Do the RZ conversion manually** inside the functor (3 lines: `r = sqrt(x^2+y^2)`, `theta = atan2(y,x)`, write r/theta/z).
> 
> Option 2 is self-contained and avoids yet another parameter. The existing `FindEmbeddedBoundaryIntersection` does exactly this — manual RZ writeback at line 151-153.
> 
> ## Alternatively: bundle into a context struct
> 
> If the parameter list feels too wide, you could bundle the geometry data:
> 
> ```cpp
> struct EBScrapeContext {
>     amrex::ParticleReal xp, yp, zp;
>     amrex::RealVect normal;
>     amrex::Real phi_value;
>     amrex::Array4<const amrex::Real> phi;
>     amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> plo, dxi;
> };
> // functor call: f(ptd, ip, ctx, engine)
> ```
> 
> This is a bigger interface change but future-proofs cleanly — adding more context data later doesn't change the functor signature again.
> 
> ## Tradeoffs vs. the separate-function approach
> 
> | | Functor approach | Separate `reflectParticlesAtEB` |
> |---|---|---|
> | **Single entry point** | Yes — one `scrapeParticlesAtEB` for all BCs | Two parallel functions to maintain |
> | **Interface width** | Wide (10+ args, or context struct) | N/A — logic is inline |
> | **Future BCs (diffuse, thermal)** | Same functor pattern; also get phi/plo/dxi | Would each need their own function |
> | **Code duplication** | None — scraper loop written once | Outer loop (tile iteration, phi check) duplicated |
> | **Existing call sites** | Must update `NoOp`/`Absorb` signatures everywhere | Untouched |
> 
> The "existing call sites" point is worth checking — `NoOp` and `Absorb` are used in 4 places total (the EB scraper plus particle injection/repositioning). They'd all need the updated signature.
> 
> **My take**: The functor approach is architecturally cleaner and more consistent with the existing codebase design. The context-struct variant is the best version of it — one signature change now, extensible later without further churn. The separate-function approach I implemented was expedient but duplicates the scraper loop structure.
>

RemiLehe:
@Noerr: After more discussion with @atmyers we agreed that the functor approach proposed by Claude makes sense. For now, we don't need to bundle the geometry data in a `EBScrapeContext` struct. Also, doing the RZ conversion manually (Option 2) is probably preferable.

Could you ask Claude to update the PR accordingly?

RemiLehe:
I started working on it in #6588

Noerr:
Closing as implemented by https://github.com/BLAST-WarpX/warpx/pull/6588  thanks.

PR #6582, opened 2026-02-14
Control Magnetostatic Solver Precision

n01r:
Add `magnetostatic_solver_*` input parameters for independent control of the magnetostatic Poisson solver precision and iteration settings. As a default we fall back to the labframe electrostatic solver configuration. 
This PR removes hard-coded values that were set before.


roelof-groenewald (approved):
Looks good! Thanks!

n01r:
@agargone also successfully run-tested this yesterday :)

PR #4508, opened 2023-12-13
Fix default initialization of runtime attributes.

atmyers:
Address Issue #4483

PR #6583, opened 2026-02-14
simplify DefaultInitializeRuntimeAttribute() parameters

JustinRayAngus:
This PR simplifies the parameters passed to DefaultInitializeRuntimeAttributes().

### DIFF
--- a/Source/Particles/Collision/BinaryCollision/Bremsstrahlung/PhotonCreationFunc.H
+++ b/Source/Particles/Collision/BinaryCollision/Bremsstrahlung/PhotonCreationFunc.H
@@ -272,7 +272,7 @@ public:
             const auto start_index = int(products_np[i]);
             const auto stop_index  = int(products_np[i] + num_added_vec[i]);
             ParticleCreation::DefaultInitializeRuntimeAttributes(*tile_products[i],
-                                      *pc_products[i], 0, 0, start_index, stop_index);
+                                                 *pc_products[i], start_index, stop_index);
         }
 
         amrex::Gpu::synchronize();
--- a/Source/Particles/Collision/BinaryCollision/Bremsstrahlung/PhotonCreationFunc.H
+++ b/Source/Particles/Collision/BinaryCollision/Bremsstrahlung/PhotonCreationFunc.H
@@ -271,8 +271,8 @@ public:
         {
             const auto start_index = int(products_np[i]);
             const auto stop_index  = int(products_np[i] + num_added_vec[i]);
-            BinaryCollisionUtils::DefaultInitializeRuntimeAttributes(tile_products[i],
-                                      pc_products[i], start_index, stop_index);
+            ParticleCreation::InitializeRuntimeAttributesFromPC(*tile_products[i],
+                                      *pc_products[i], start_index, stop_index);
         }
 
         amrex::Gpu::synchronize();
--- a/Source/Particles/Collision/BinaryCollision/Bremsstrahlung/PhotonCreationFunc.H
+++ b/Source/Particles/Collision/BinaryCollision/Bremsstrahlung/PhotonCreationFunc.H
@@ -271,8 +271,8 @@ public:
         {
             const auto start_index = int(products_np[i]);
             const auto stop_index  = int(products_np[i] + num_added_vec[i]);
-            ParticleCreation::InitializeRuntimeAttributesFromPC(*tile_products[i],
-                                      *pc_products[i], start_index, stop_index);
+            ParticleCreation::DefaultInitializeRuntimeAttributes(*tile_products[i],
+                                      *pc_products[i], 0, 0, start_index, stop_index);
         }
 
         amrex::Gpu::synchronize();
--- a/Source/Particles/Collision/BinaryCollision/Bremsstrahlung/PhotonCreationFunc.H
+++ b/Source/Particles/Collision/BinaryCollision/Bremsstrahlung/PhotonCreationFunc.H
@@ -271,20 +271,8 @@ public:
         {
             const auto start_index = int(products_np[i]);
             const auto stop_index  = int(products_np[i] + num_added_vec[i]);
-            ParticleCreation::DefaultInitializeRuntimeAttributes(*tile_products[i],
-                                       0, 0,
-                                       pc_products[i]->getUserRealAttribs(), pc_products[i]->getUserIntAttribs(),
-                                       pc_products[i]->GetRealSoANames(), pc_products[i]->GetIntSoANames(),
-                                       pc_products[i]->getUserRealAttribParser(),
-                                       pc_products[i]->getUserIntAttribParser(),
-#ifdef WARPX_QED
-                                       false, // do not initialize QED quantities, since they were initialized
-                                              // when calling the SmartCopy functors
-                                       pc_products[i]->get_breit_wheeler_engine_ptr(),
-                                       pc_products[i]->get_quantum_sync_engine_ptr(),
-#endif
-                                       pc_products[i]->getIonizationInitialLevel(),
-                                       start_index, stop_index);
+            BinaryCollisionUtils::DefaultInitializeRuntimeAttributes(tile_products[i],
+                                      pc_products[i], start_index, stop_index);
         }
 
         amrex::Gpu::synchronize();
--- a/Source/Particles/ParticleCreation/DefaultInitialization.H
+++ b/Source/Particles/ParticleCreation/DefaultInitialization.H
@@ -95,22 +95,15 @@ namespace ParticleCreation {
      * attributes, which have been in principle externally set elsewhere.
      *
      * @tparam PTile The type of the particle tile to operate on (e.g. could use different allocators)
+     * @tparam DstPC the dst particle container type
      * @param[inout] ptile the tile in which attributes are initialized
+     * @param[in] pc Particle container
      * @param[in] n_external_attr_real The number of real attributes that have been externally set.
      * These are NOT initialized by this function.
      * @param[in] n_external_attr_int The number of integer attributes that have been externally set.
      * These are NOT initialized by this function.
-     * @param[in] user_real_attribs The names of the real components for this particle tile
-     * @param[in] user_int_attribs The names of the int components for this particle tile
-     * @param[in] particle_comps particle component names for real comps
-     * @param[in] particle_icomps particle component names for int comps
-     * @param[in] user_real_attrib_parser the parser functions used to initialize the user real components
-     * @param[in] user_int_attrib_parser the parser functions used to initialize the user int components
      * @param[in] do_qed_comps whether to initialize the qed components (these are usually handled by
      * SmartCopy, but NOT when adding particles in AddNParticles)
-     * @param[in] p_bw_engine the engine to use for setting the breit-wheeler component for QED
-     * @param[in] p_qs_engine the engine to use for setting the quantum synchrotron component for QED
-     * @param[in] ionization_initial_level the ionization level particles created should start at
      * @param[in] start the index to start initializing particles
      * @param[in] stop the index to stop initializing particles
      */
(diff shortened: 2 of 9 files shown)


dpgrote:
There is already a wrapper for `DefaultInitializeRuntimeAttributes` in `Source/Particles/PhysicalParticleContainer.cpp`. Can this be reused? It looks like that only difference is the value passed in for the first QED parameter.

JustinRayAngus:
@dpgrote This could possibly be re-used. There are actually 3 arguments that are different for the wrapper in `Source/Particles/PhysicalParticleContainer.cpp`. In addition to the QED boolean flag, there is also `n_external_attr_real` and `n_external_attr_int`. I wonder though if this routine can be simplified to remove the need for all three of these input parameters. 

The wrapper you mention is only called from `WarpXParticleContainer::AddNParticles()`. It looks like the first `n_external_attr_real` and `n_external_attr_int` are set immediately before the call to `DefaultInitializeRuntimeAttributes()`. These values are passed to that routine to tell it to skip initializing them since they are already initialized. Seems to me that these flags can just be removed and let them be initialized inside `DefaultInitializeRuntimeAttributes()`. Perhaps there is some reason that I don't understand for why they need to be initialized before? @EZoni @RemiLehe @ax3l. 

As for the QED flag, this also doesn't seem like something that needs to exist in the lower-level routines that call this function. I guess the question is why is it always set to `false` when called from the collision routines? Or, why is it only set to `true` when called from `AddParticles()`? Perhaps this flag can always be `true`, and then it too can be removed.

If the above three parameters can be removed. Then I don't see the need for a wrapper to exist it all. Just change the base routine to take a reference to the particle container as an input parameter.

Source/Particles/Collision/BinaryCollision/Bremsstrahlung/PhotonCreationFunc.H
```diff
@@ -272,19 +272,7 @@ public:
...
-                                       pc_products[i]->getUserIntAttribParser(),
-#ifdef WARPX_QED
-                                       false, // do not initialize QED quantities, since they were initialized
-                                              // when calling the SmartCopy functors
-                                       pc_products[i]->get_breit_wheeler_engine_ptr(),
-                                       pc_products[i]->get_quantum_sync_engine_ptr(),
-#endif
-                                       pc_products[i]->getIonizationInitialLevel(),
-                                       start_index, stop_index);
+                                      *pc_products[i], 0, 0, start_index, stop_index);
```
EZoni: Since we're at it, we could also use a more reasonable indentation and a more explicit syntax, here and in other similar calls, but fine with me if you want to keep style changes out of this PR.

For example, here we could write:
```cpp
ParticleCreation::DefaultInitializeRuntimeAttributes(
    *tile_products[i],
    *pc_products[i], 
    /*n_external_attr_real=*/0,
    /*n_external_attr_int=*/0,
    start_index,
    stop_index
);
```

P.S. I wish we had managed to handle formatting automatically, so we wouldn't have to make comments like this in every PR, but we never agreed on how to finalize that work (e.g., #5739)...

Source/Particles/ParticleCreation/DefaultInitialization.H
```diff
@@ -95,45 +95,44 @@ namespace ParticleCreation {
      * attributes, which have been in principle externally set elsewhere.
      *
      * @tparam PTile The type of the particle tile to operate on (e.g. could use different allocators)
+     * @tparam DstPC the dst particle container type
```
EZoni: We could also write "destination" instead of "dst" in the docstring, to make it less trivial:
```suggestion
     * @tparam DstPC the destination particle container type
```

Source/Particles/ParticleCreation/DefaultInitialization.H
```diff
@@ -95,45 +95,44 @@ namespace ParticleCreation {
...
      * @param[inout] ptile the tile in which attributes are initialized
+     * @param[in] pc Particle container
```
EZoni:
```suggestion
     * @param[in] pc the destination particle container
```

Source/Particles/ParticleCreation/FilterCopyTransform.H
```diff
@@ -306,8 +271,8 @@ Index filterCopyTransformParticles (DstPC& pc1, DstPC& pc2, DstTile& dst1, DstTi
  * mask as input.
  *
  * \tparam N number of particles created in the dst(s) for each filtered src particle
- * \tparam DstTile the dst particle tile type
- * \tparam SrcTile the src particle tile type
+ * \tparam DstPC the dst particle container type
+ * \tparam SrcPC the src particle container type
```
EZoni: Similarly to my comment above:

```suggestion
 * \tparam DstPC the destination particle container type
 * \tparam SrcPC the source particle container type
```

Though I admit I don't see any `SrcPC` in the template parameters below:
```cpp
template <int N, typename DstPC, typename DstTile, typename SrcTile, typename Index,
          typename PredFunc, typename TransFunc, typename CopyFunc1, typename CopyFunc2>
Index filterCopyTransformParticles (DstPC& pc1, DstPC& pc2, DstTile& dst1, DstTile& dst2, SrcTile& src,
                                    Index dst1_index, Index dst2_index,
                                    PredFunc&& filter, CopyFunc1&& copy1, CopyFunc2&& copy2,
                                    TransFunc&& transform) noexcept
```

EZoni (changes requested):
Thanks for the PR, @JustinRayAngus. I left a few minor comments on the code changes. Haven't yet looked into the question of what to do with the other wrapper.

Source/Particles/PhysicalParticleContainer.cpp
```diff
@@ -418,19 +418,14 @@ PhysicalParticleContainer::DefaultInitializeRuntimeAttributes (
...
-                                       m_user_real_attribs, m_user_int_attribs,
-                                       GetRealSoANames(), GetIntSoANames(),
-                                       amrex::GetVecOfPtrs(m_user_real_attrib_parser),
-                                       amrex::GetVecOfPtrs(m_user_int_attrib_parser),
+                                       0, pinned_tile.numParticles()
 #ifdef WARPX_QED
-                                       true,
-                                       m_shr_p_bw_engine.get(),
-                                       m_shr_p_qs_engine.get(),
+                                       , true );
```
EZoni: One more suggestion:

```suggestion
                                       , /*do_qed_comps=*/true );
```

Source/Particles/Collision/BinaryCollision/Bremsstrahlung/PhotonCreationFunc.H
JustinRayAngus: One of the tasks is to potentially remove these arguments. I'll wait for the verdict on that. If they are to be retained, then I will implement your suggestions.

Source/Particles/ParticleCreation/FilterCopyTransform.H
JustinRayAngus: It look like `SrcPC` is not used. I will remove that.

JustinRayAngus: Looks like there are a lot of missing params in this particular file. I don't think it is the intent of this PR to clean that file up. I think I will keep changes in this file to a minimum.

EZoni:
@dpgrote 

What @JustinRayAngus is saying in https://github.com/BLAST-WarpX/warpx/pull/6583#issuecomment-3911090375 seems reasonable to me (and could be done in a follow-up PR). Does it make sense to you?

JustinRayAngus:
> @dpgrote
> 
> What @JustinRayAngus is saying in [#6583 (comment)](https://github.com/BLAST-WarpX/warpx/pull/6583#issuecomment-3911090375) seems reasonable to me (and could be done in a follow-up PR). Does it make sense to you?

I think I've already refactored this PR to do what I suggested in this comment. The only remaining questions are those that I put in the task list in the PR description.

EZoni:
@JustinRayAngus 

Understood, thanks. I'm also confused about `n_external_attr_real` and `n_external_attr_int` honestly. It's not clear to me how they're set and/or used in practice.

I think they might have been introduced in #4508. @atmyers, can you help us understand why these variables are necessary and how they are set "externally" in practice? Do you think there could be a way to count those numbers from other available data and remove them from the interface of the function that @JustinRayAngus is refactoring here?

Is this maybe a case where the real attributes are set externally? 
https://github.com/BLAST-WarpX/warpx/blob/b69ffb74d594c7490b01b8eb7b497f0db325568a/Examples/Tests/boundaries/inputs_test_3d_particle_boundaries#L43

Are the `n_external_attr_real` and `n_external_attr_int` the same as the `nattr_real` and `nattr_int` here? 
https://github.com/BLAST-WarpX/warpx/blob/b69ffb74d594c7490b01b8eb7b497f0db325568a/Source/Particles/WarpXParticleContainer.cpp#L182-L194
If so, I would name those variables in the same way to help developers trace this better.

P.S. See @JustinRayAngus's [comment](https://github.com/BLAST-WarpX/warpx/pull/6583#issuecomment-3911090375) above for more context.

JustinRayAngus:
@EZoni @atmyers 

I think the name of these parameters makes them confusing. They are only needed if you want to _skip_ setting the first `n_external_attr` components inside `DefaultInitilizeRuntimeAttributes()`. The loop over components starts at `n_external_attr`. These values passed into `DefaultInitilizeRuntimeAttributes()` are both zero at every place in the code except for the call from the function `AddParticles()`, where the first `n_external_attr` are set immediately before the call to `DefaultInitilizeRuntimeAttributes()`.

What I'm suggesting is the following:

1. Remove the setting of the first `n_external_attr_real` and `n_external_attr_int` from `AddParticles()` just before the call to `DefaultInitilizeRuntimeAttributes()`.
2. Remove `n_external_attr_real` and `n_external_attr_int` params from `DefaultInitilizeRuntimeAttributes()`.
3. Always start the loop for setting the real and int params in `DefaultInitilizeRuntimeAttributes()` at zero.

atmyers:
Those parameters predate #4508, I think they were added in [this](https://github.com/BLAST-WarpX/warpx/commit/b4686616d884e11f4a624989d370b3c032d0c1c4) commit.  I *think* the explanation is that, when adding particles from Python, users have the option to either pass in the data values for particle attributes or allow them to be default initialized. So even though this is never used in the C++ code, I think it would break some Python scripts to assume these are always zero.

EZoni:
@JustinRayAngus 

Based on @atmyers's comment, I think we should leave the two parameters untouched for now. What do you think?

Source/Particles/Collision/BinaryCollision/Bremsstrahlung/PhotonCreationFunc.H
JustinRayAngus: I've made these arguments 0 by default removing the need to pass `0, 0`, which was done in many places, while retaining the ability to pass them as needed by `AddParticles()`.

atmyers (approved):

JustinRayAngus:
> @JustinRayAngus
> 
> Understood, thanks. I'm also confused about `n_external_attr_real` and `n_external_attr_int` honestly. It's not clear to me how they're set and/or used in practice.
> 
> I think they might have been introduced in #4508. @atmyers, can you help us understand why these variables are necessary and how they are set "externally" in practice? Do you think there could be a way to count those numbers from other available data and remove them from the interface of the function that @JustinRayAngus is refactoring here?
> 
> Is this maybe a case where the real attributes are set externally?
> 
> https://github.com/BLAST-WarpX/warpx/blob/b69ffb74d594c7490b01b8eb7b497f0db325568a/Examples/Tests/boundaries/inputs_test_3d_particle_boundaries#L43
> 
> Are the `n_external_attr_real` and `n_external_attr_int` the same as the `nattr_real` and `nattr_int` here?
> 
> https://github.com/BLAST-WarpX/warpx/blob/b69ffb74d594c7490b01b8eb7b497f0db325568a/Source/Particles/WarpXParticleContainer.cpp#L182-L194
> 
> 
> If so, I would name those variables in the same way to help developers trace this better.
> P.S. See @JustinRayAngus's [comment](https://github.com/BLAST-WarpX/warpx/pull/6583#issuecomment-3911090375) above for more context.

@EZoni 
Looks like `nattr_int` is the same as `n_external_attr_int`, but `n_external_attr_real` is set to `natter_real - 1`. So they are not necessarily intended to be the same in general. For `DefaultInitializeRuntimeAttribute()`, these parameters are meant to specify the first so many number of attributes of each type that have already been set externally and therefore do not need to be set inside this function.

EZoni (approved):

PR #6588, opened 2026-02-17
Implement reflection from embedded boundaries

RemiLehe:
# Overview

This PR adds the ability to **specularly reflect** particles off the surface of an embedded boundary (EB), instead of the current behavior which unconditionally absorbs them.

# Changes

### New user parameter

A new input parameter `boundary.particle_eb` selects the particle boundary condition at the embedded boundary:

- `Absorbing` (default): particles that reach the EB are deleted (previous, unchanged behavior).
- `Reflecting`: particles that reach the EB are specularly reflected back into the domain.

The interface is consistent with the [existing](https://warpx.readthedocs.io/en/latest/usage/parameters.html#boundary.particle_lo-hi) `boundary.particle_lo` / `boundary.particle_hi` (which also have `absorbing` and `reflecting` as possible options).

### New `Reflect` functor

A `Reflect` struct is added to `ParticleBoundaryProcess.H`, alongside the existing `NoOp` and `Absorb`. For each particle that has crossed into the EB it:

1. Bisects along the particle trajectory to find the contact point on the EB surface (where the signed distance is zero);
2. Computes the surface normal at that contact point;
3. Applies specular velocity reflection on the momentum;
4. Advances the particle from the contact point with the reflected velocity for the remaining fraction of the timestep.

The scraper functor interface in `ParticleScraper.H` is extended accordingly: functors now receive the `SetParticlePosition` setter, the Cartesian particle position, the signed-distance array, and the geometry data (`plo`/`dxi`). `NoOp` and `Absorb` accept and ignore the additional arguments.

### Small refactor of `interp_normal`

`DistanceToEB::interp_normal` now returns the unit normal in **3D Cartesian coordinates** (`GpuArray<ParticleReal, 3>`) rather than a reduced-dimension `RealVect`, converting from the reduced-dimension coordinates in which it is computed. This makes it directly usable for the velocity reflection (which is done in 3D Cartesian) and for recording the surface normal in `ParticleBoundaryBuffer`, and handles all dimensionalities (3D, XZ, RZ, RCYLINDER, RSPHERE, 1D_Z).

### New test

Adds `test_rz_particle_boundary_interaction_reflecting`, whose physical configuration is **identical** to [`test_rz_particle_boundary_interaction_picmi`](https://github.com/BLAST-WarpX/warpx/blob/development/Examples/Tests/particle_boundary_interaction/inputs_test_rz_particle_boundary_interaction_picmi.py) (which uses PICMI and a Python callback for the reflection) and shares the same CI analysis script, but exercises the reflection directly through the WarpX executable.


Noerr:
This looks good @RemiLehe , thanks!  I'm just coming back from a week off.  Peter and I will aim to test this soon, but probably no feedback until early next week.

pstoltzZap:
I need to update my comment below.  I ran that test still including the mirror reflection callback...🙄...so obviously it passed.

When I updated the input file to comment out the callback but include the 
```
simulation.embedded_boundary = picmi.EmbeddedBoundary(
            implicit_function="(x**2+z**2-R_w**2)", R_w=0.9*self.r0,
            particle_boundary_condition='Reflecting',
        )
```
syntax, the test fails, but not catastrophically.  The particle survives (which is not true if absorbing is on), but the position errors are larger:
```
pstoltz@bugs:~/WARPX_RUNS/particle_boundary_interaction$ python analysis.py diags/diag1/
NUMERICAL coordinates of the point of contact:
x=0.03308, y=0.00000, z=-0.20299


ANALYTICAL coordinates of the point of contact:
x=0.03532, y=0.00000, z=-0.20531


Relative percentage error for x = 6.3461 %
Relative percentage error for z = 1.1276 %
```
Sorry for the confusion


>I built and ran the particle_boundary_interaction test on our local cluster, and it passed.
>
>pstoltz:~/WARPX_RUNS/particle_boundary_interaction$ python analysis.py diags/diag1/
>NUMERICAL coordinates of the point of contact:
>x=0.03532, y=0.00000, z=-0.20531
>
>
>ANALYTICAL coordinates of the point of contact:
>x=0.03532, y=0.00000, z=-0.20531
>
>
>Relative percentage error for x = 0.0001 %
>Relative percentage error for z = 0.0006 %
>pstoltz:

Noerr:
Hi @RemiLehe ,   we've managed to get turned around a few times whether we're utilizing the Reflecting boundary or not.  Here's a verbose analysis and recommended tweak to ensure the embedded particle boundary condition is always written to `warpx_used_inputs`.  

Would you also consider making the warning more stern that the default (absorbing) behavior will be deprecated some time in the future?



## Problem

PR #6588 introduces `warpx.eb_particle_boundary_condition` parsed via
`query_enum_sloppy` inside a `contains()` guard.  When the user **does not**
specify the parameter (the common case for existing simulations), the default
`Absorbing` value is set in C++ but **never written back to ParmParse**.

This means the default is invisible in `warpx_used_inputs`, breaking the
project's convention that every active parameter — whether user-supplied or
defaulted — appears in that file for reproducibility.

## Current code (lines 292–315 of `Source/WarpX.cpp` on the PR branch)

```cpp
if (EB::enabled()) {
    amrex::ParmParse const pp_warpx("warpx");
    if (pp_warpx.contains("eb_particle_boundary_condition")) {
        pp_warpx.query_enum_sloppy(
            "eb_particle_boundary_condition", eb_particle_boundary, "-_");
            eb_particle_boundary == ParticleBoundaryType::Absorbing ||
            eb_particle_boundary == ParticleBoundaryType::Reflecting,
            "warpx.eb_particle_boundary_condition must be Absorbing or Reflecting");
    } else {
        eb_particle_boundary = ParticleBoundaryType::Absorbing;
            "EmbeddedBoundary",
            "warpx.eb_particle_boundary_condition is not specified. ...",
    }
}
```

### Why `query_enum_sloppy` alone isn't enough

`query_enum_sloppy` (AMReX `ParmParse.H:1474`) internally calls `query()`,
which marks a parameter as "used" — but **only if the parameter exists**.
Unlike `queryAdd()`, it performs no database insertion when the key is absent.

| Method             | Marks used? | Adds default to ParmParse? | Appears in `warpx_used_inputs`? |
|--------------------|:-----------:|:--------------------------:|:-------------------------------:|
| `queryAdd`         | yes         | **yes**                    | always                          |
| `query_enum_sloppy`| yes         | no                         | only if user-specified          |

The `contains()` guard makes the bifurcation explicit, but the `else` branch
writes the default only to the C++ variable — ParmParse never learns about it.

## Recommended fix

Drop the `contains()` guard, use the return value of `query_enum_sloppy`
directly, and `add()` the string default in the else branch (manual
`queryAdd`-equivalent for enums):

```cpp
if (EB::enabled()) {
    amrex::ParmParse pp_warpx("warpx");  // non-const: add() needed
    bool const user_specified = pp_warpx.query_enum_sloppy(
        "eb_particle_boundary_condition", eb_particle_boundary, "-_");
        eb_particle_boundary == ParticleBoundaryType::Absorbing ||
        eb_particle_boundary == ParticleBoundaryType::Reflecting,
        "warpx.eb_particle_boundary_condition must be Absorbing or Reflecting");
    if (!user_specified) {
        // Write default into ParmParse so it appears in warpx_used_inputs
        pp_warpx.add("eb_particle_boundary_condition",
                      std::string("Absorbing"));
            "EmbeddedBoundary",
            "warpx.eb_particle_boundary_condition is not specified. "
            "Defaulting to Absorbing. Particles that cross the embedded "
            "boundary surface will be removed. Set to Reflecting to "
            "reflect particles instead.",
    }
}
```

### What changes

1. **`pp_warpx` drops `const`** — needed for the `add()` call.
2. **`contains()` guard removed** — `query_enum_sloppy` already returns 0
   when the key is absent, so the guard was redundant.
3. **`pp_warpx.add(...)` in the else branch** — ensures `Absorbing` appears
   in `warpx_used_inputs` even when the user doesn't specify it.
4. **Validation moves outside the branches** — the assert applies regardless
   of source (user input or default), keeping the check in one place.

### Why not just use `queryAdd`?

AMReX has no `queryAddEnumSloppy` helper.  `queryAdd` works on raw types
(`int`, `bool`, `std::string`) but not on `AMREX_ENUM` types with the
case-insensitive/ignore-chars matching that `query_enum_sloppy` provides.
The manual `query_enum_sloppy` + `add()` pattern is the idiomatic workaround.

### Precedent in the codebase

The `queryAdd` pattern is used ~19 times in WarpX (e.g., `WarpXAMReXInit.cpp`
lines 40–170, `WarpXInitData.cpp` lines 314–376).  The manual equivalent
(`query` + conditional `add`) is used for EB geometry in `WarpXInitEB.cpp:89`.

Noerr:
Hi @RemiLehe,  I have been experiencing multiple failure modes since using this PR reflecting boundary patch on both NVIDIA and AMD GPU.  Tricky to triage and special conditions are needed to trigger the failure.   Ultimately, I have determined that an extra Redistribute() particles is needed after particles reflect.

The fix is simple one-line in `Source/Evolve/WarpXEvolve.cpp`

```cpp
if (EB::enabled()) {
    mypc->ScrapeParticlesAtEB(...);
    m_particle_boundary_buffer->gatherParticlesFromEmbeddedBoundaries(...);
    if (eb_particle_boundary == ParticleBoundaryType::Reflecting) {
        mypc->Redistribute();    // <-- this is missing. Bug.
    } else { // <-- totally optional conditional, but reflecting does not invalidate particles.  deleteInvalid is a NoOp.
        mypc->deleteInvalidParticles();
    }
}
```

I'm going to paste in the more extensive and verbose report next
BTW, https://github.com/BLAST-WarpX/warpx/pull/6588 has been merged.  I'd like to keep moving this PR forward.  The impact of the reflecting embedded boundary feature has been terrific for our Z pinch sims.

Noerr:
# Reflecting EB Crash: Root Cause, Fix, and Confirmation

## Date: 2026-03-17

## Status: FIX CONFIRMED

The one-line `Redistribute()` fix has been validated:
- **Job 50205711**: Reflecting EB + `CUDA_LAUNCH_BLOCKING=1` + diags every step → **192+ steps, all diagnostics passed**
- **Job 50206823**: Reflecting EB + normal async CUDA + diags every step → **190+ steps, all diagnostics passed**

Both tests used the same case (`hg250923018_patched_reflecting_debug`) that previously crashed deterministically at step 2.

## Root Cause

After EB reflection, particles have new positions that may be in a different tile or MPI rank. But the code never calls `Redistribute()` after reflection, so particles remain stranded on the wrong rank. Subsequent field operations (deposition, FillBoundary) access invalid memory because particles index into arrays they don't belong to.

## Evidence

### 1. Failure requires all three conditions simultaneously
- Reflecting EB (absorbing removes particles — no position inconsistency)
- T_ions diagnostic enabled (triggers `SynchronizeVelocityWithPosition` → `FillBoundaryB`)
- High tangential flow velocity (large tangential velocities at EB surface → large post-reflection displacement)

### 2. Absorbing EB passes cleanly
- Job 50204399: Changed only `eb_particle_boundary` from Reflecting to Absorbing
- Passed 9+ steps with diagnostic writes at every step — no crash
- Absorbing marks particles invalid; `deleteInvalidParticles()` removes them locally

### 3. Crash is in FillBoundary, not T_ions deposition
- `CUDA_LAUNCH_BLOCKING=1` (job 50160342) pinpointed the faulting kernel:
  ```
  WarpX::Evolve::step
    → SynchronizeVelocityWithPosition
      → FillBoundaryB
        → FillBoundary_finish
          → unpack_recv_buffer_gpu   ← actual faulting kernel
  ```
- This is a field synchronization operation, not particle deposition
- `SynchronizeVelocityWithPosition` likely only runs on diagnostic steps, explaining the diagnostic-step correlation

### 4. compute-sanitizer memcheck found zero memory violations
- Job 50155207: 32 ranks, all clean (only benign `cuPointerGetAttribute` false positives)
- This is consistent with particles writing into *valid but wrong* memory (another rank's arrays) rather than unmapped pages

### 5. Failure threshold is step 2
- Step 0 diagnostic: always passes (no particles pushed yet)
- Step 1 diagnostic: always passes (particles pushed once, may not have reflected far enough)
- Step 2 diagnostic: crashes (job 50157336, un-instrumented, diags every 1 step)
- Only 1-2 push-reflect cycles needed to move particles across tile boundaries

### 6. Crash rank is non-deterministic
- Varies across runs — consistent with stochastic particle positions after reflection depending on floating-point non-determinism

## The Code Gap

**File:** `Source/Evolve/WarpXEvolve.cpp`, function `HandleParticlesAtBoundaries()`, lines 654-704

Current execution order:
```
Line 658:  ApplyBoundaryConditions()
Line 659:  gatherParticlesFromDomainBoundaries()
Line 665:  Redistribute()                          ← particles moved to correct rank/tile
Line 689:  if (EB::enabled()) {
Line 691:      ScrapeParticlesAtEB()                ← Reflect modifies positions IN-PLACE
Line 692:      gatherParticlesFromEmbeddedBoundaries()
Line 694:      deleteInvalidParticles()             ← local only, no MPI
Line 695:  }                                        ← NO Redistribute() after reflection!
```

The Reflect functor (`ParticleBoundaryProcess.H:70-190`) does:
1. Bisect to find contact point on EB surface
2. Compute surface normal
3. Specular reflection: `v_new = v - 2*(v·n)*n`
4. **Advance particle from contact point with reflected velocity for remaining time fraction**

Step 4 gives the particle a brand-new position. For grazing-angle reflections with high tangential velocity, this position can be far from the original cell — potentially in a different tile or MPI domain.

For comparison, the absorbing path (`ParticleBoundaryProcess::Absorb`) just sets the particle ID to invalid. `deleteInvalidParticles()` is sufficient because no position update occurs.

## Planned Fix

Add `Redistribute()` after the EB block when reflecting boundaries are active:

```cpp
// interact the particles with EB walls (if present)
if (EB::enabled()) {
    using warpx::fields::FieldType;
    mypc->ScrapeParticlesAtEB(m_fields.get_mr_levels(FieldType::distance_to_eb, finest_level));
    m_particle_boundary_buffer->gatherParticlesFromEmbeddedBoundaries(
        *mypc, m_fields.get_mr_levels(FieldType::distance_to_eb, finest_level), cur_time);
    mypc->deleteInvalidParticles();
    if (WarpX::eb_particle_boundary == ParticleBoundaryType::Reflecting) {
        mypc->Redistribute();   // Reflected particles may have crossed tile/rank boundaries
    }
}
```

**Why conditional on Reflecting:** Absorbing doesn't change positions, so `Redistribute()` would be wasted MPI overhead. Only reflecting creates new positions that need redistribution.

**Precedent:** The domain boundary reflection path already follows this pattern. In `HandleParticlesAtBoundaries()`:
- Line 658: `ApplyBoundaryConditions()` — handles domain boundary reflections (modifies positions)
- Line 659: `gatherParticlesFromDomainBoundaries()`
- Line 665: `Redistribute()` — correctly placed after domain boundary reflection

The EB block (lines 688-698) was added by the reflecting EB PR but omitted the `Redistribute()`. The PR likely modeled the code after the absorbing EB path (where `deleteInvalidParticles()` suffices) rather than the reflecting domain boundary path (which requires `Redistribute()`).

**Location:** `Source/Evolve/WarpXEvolve.cpp`, insert after line 694 (after `deleteInvalidParticles()`).

## All Test Jobs

| Job | Config | Result |
|-----|--------|--------|
| 50152794 | memcheck, diags@251 | ~1 step in 13 min; killed (sanitizer blocked signal) |
| 50155207 | memcheck, diags@3 | Crashed step 3. Zero memory violations (75,701 false positives only) |
| 50157336 | Un-instrumented, diags@1 | Crashed step 2. Steps 0-1 diags passed. |
| 50160342 | CUDA_LAUNCH_BLOCKING=1, diags@1 | Crashed step 2. Backtrace: `unpack_recv_buffer_gpu` in `FillBoundaryB` via `SynchronizeVelocityWithPosition` |
| 50157837 | racecheck, diags@3 | Crashed step 2. Sanitizer logs empty (killed before flush) |
| 50204399 | Absorbing EB, diags@1 | **Passed 9+ steps cleanly** |
| 50205137 | **Redistribute() fix**, CUDA_LAUNCH_BLOCKING=1, diags@1 | Failed — .so not installed into venv (stale binary) |
| 50205711 | **Redistribute() fix (installed)**, CUDA_LAUNCH_BLOCKING=1, diags@1 | **PASSED 192+ steps, all diags clean** |
| 50206823 | **Redistribute() fix (installed)**, async CUDA, diags@1 | **PASSED 190+ steps, all diags clean** |

## Upstream Action

This is a one-line fix for the reflecting EB PR. The `Redistribute()` call is necessary because the Reflect functor (`ParticleBoundaryProcess.H:70-190`) advances reflected particles from the contact point with reflected velocity for the remaining time fraction, giving them a brand-new position that can be in a different tile or MPI domain.

The absorbing path never needed this because `Absorb` just marks particles invalid — `deleteInvalidParticles()` handles cleanup locally without any position change.

RemiLehe:
Thanks a lot for pointing this out @Noerr! I just updated the PR accordingly.

Regression/Checksum/benchmarks_json/test_rz_particle_boundary_interaction_reflecting.json
```diff
@@ -0,0 +1,18 @@
...
+    "particle_weight": 1.0
+  },
+  "lev=0": {
+    "Er": 1.1876790141078713e-06,
+    "Ez": 1.5875032633308516e-06,
+    "phi": 2.98382958848146e-07,
+    "rho": 8.111193584817117e-16,
+    "rho_electrons": 8.111193584817117e-16
+  }
+}
```
RemiLehe, line 18:
```suggestion
{
  "electrons": {
    "particle_momentum_x": 7.086217697811126e-21,
    "particle_momentum_y": 0.0,
    "particle_momentum_z": 7.315162903043373e-21,
    "particle_position_x": 0.03308210419274677,
    "particle_position_y": 0.0,
    "particle_position_z": 0.20299146074286406,
    "particle_weight": 1.0
  },
  "lev=0": {
    "Er_0_real": 1.2100397143934173e-06,
    "Ez_0_real": 1.7200192263636746e-06,
    "phi_0_real": 3.105257562882442e-07,
    "rho_0_real": 8.110719371837059e-16,
    "rho_electrons_0_real": 8.110719371837059e-16
  }
}
```

Docs/source/usage/parameters.rst
```diff
@@ -1004,6 +1004,19 @@ additionally define the electric potential at the embedded boundary with an anal
...
+    :default: ``Absorbing``
+    :optional:
+
+    The boundary condition applied to the particles when they reach the surface of the embedded boundary. Options are:
+
+    * ``Absorbing``: Particles that reach the embedded boundary are deleted. This is the default behavior.
+
+    * ``Reflecting``: Particles that reach the embedded boundary are specularly reflected back into the simulation domain,
+      i.e. the component of the velocity normal to the surface is reversed while the tangential component is unchanged.
+      The exact point of contact on the (possibly curved) surface is found by bisection, so that the reflection is accurate even for curved boundaries.
```
```suggestion
    * ``Reflecting``: Particles that reach the embedded boundary are specularly reflected back into the simulation domain
```

Source/Evolve/WarpXEvolve.cpp
```diff
@@ -800,8 +800,15 @@ void WarpX::HandleParticlesAtBoundaries (int step, amrex::Real cur_time, int num
...
-        // Remove particles that have been flagged to be scraped
-        mypc->deleteInvalidParticles();
+        if (eb_particle_boundary == ParticleBoundaryType::Absorbing) {
+            // If particles are simply absorbed, no need for a full Redistribute.
+            // Instead: simply delete the absorbed particles
+            mypc->deleteInvalidParticles();
+        } else {
+            // For other particle boundary conditions (e.g. reflecting),
+            // particles can move to a different sub-domain, so we need a full Redistribute
+            mypc->Redistribute();
```
RemiLehe, line 810: This could be expensive. As a follow-up PR, one could use a local Redistribute, similarly to https://github.com/BLAST-WarpX/warpx/pull/6988

Docs/source/usage/parameters.rst
```diff
@@ -1004,6 +1004,17 @@ additionally define the electric potential at the embedded boundary with an anal
     inside the embedded boundary. For this reason, it is important to define
     this function in such a way that it is constant inside the embedded boundary.
 
+.. pp:param:: boundary.particle_eb
```
RemiLehe, line 1007: This is consistent with the existing syntax for the domain boundary: `boundary.particle_lo/boundary.particle_hi`, with options `absorbing`, `reflecting`, etc.
https://warpx.readthedocs.io/en/latest/usage/parameters.html#boundary.particle_lo-hi

Source/EmbeddedBoundary/ParticleScraper.H
```diff
@@ -185,32 +193,11 @@ scrapeParticlesAtEB (PC& pc, ablastr::fields::MultiLevelScalarField const& dista
...
+                amrex::Real const distance_to_eb_value = ablastr::particles::interp_field_nodal(i, j, k, W, distance_to_eb_arr);
 
-                if (phi_value < 0.0)
+                if (distance_to_eb_value < 0.0)
                 {
-                    int ic, jc, kc; // Cell-centered indices
-                    amrex::Real Wc[AMREX_SPACEDIM][2]; // Cell-centered weights
-                    ablastr::particles::compute_weights<amrex::IndexType::CELL>(
-                        xp, yp, zp, plo, dxi, ic, jc, kc, Wc);
-                    amrex::RealVect normal = DistanceToEB::interp_normal(i, j, k, W, ic, jc, kc, Wc, phi, dxi);
```
RemiLehe: In the previous code, we were computing the normal vector but never using it. 

This was here in preparation for the reflective boundary condition (i.e. what this PR adds). But when implementing this PR, we realized that this computation of `normal` cannot be actually used because:
- We first need to find the position of the intersection point (i.e. the intersection between the particle trajectory and the EB)
- We can then compute normal vector **at the intersection point**

Source/EmbeddedBoundary/ParticleScraper.H
```diff
@@ -171,7 +179,7 @@ scrapeParticlesAtEB (PC& pc, ablastr::fields::MultiLevelScalarField const& dista
             auto& tile = pti.GetParticleTile();
             auto ptd = tile.getParticleTileData();
             const auto np = tile.numParticles();
-            auto phi = (*distance_to_eb[lev])[pti].array();  // signed distance function
+            auto distance_to_eb_arr = (*distance_to_eb[lev])[pti].array();  // signed distance function
```
RemiLehe, line 192: This is a bit orthogonal to PR, but makes naming more transparent ; "phi" is reminiscent of the electrostatic potential and could thus be misleading here.

Source/EmbeddedBoundary/DistanceToEB.H
```diff
@@ -28,10 +28,13 @@ amrex::Real dot_product (const amrex::RealVect& a, const amrex::RealVect& b) noe
 
 // This function calculates the unit normal vector to the embedded boundary at
 // the particle position (xp, yp, zp), by interpolating the signed distance
-// function distance_to_eb. plo/dxi are the geometry data; the returned vector
-// is normalized to unit length.
+// function distance_to_eb. plo/dxi are the geometry data. The normal is
+// returned in 3D Cartesian coordinates (as is the particle velocity), by
+// converting from the reduced-dimension coordinates in which it is computed,
+// and is normalized to unit length.
```
RemiLehe:
```suggestion
// always returned in 3D Cartesian coordinates.
```

Source/EmbeddedBoundary/DistanceToEB.H
```diff
@@ -134,12 +137,42 @@ amrex::RealVect interp_normal (amrex::ParticleReal xp, amrex::ParticleReal yp, a
 
 #endif
 
-    // Normalize to unit length
+    // Normalize to unit length (in the reduced-dimension coordinates)
```
RemiLehe:
```suggestion
    // Normalize to unit length
```

Source/EmbeddedBoundary/DistanceToEB.H
```diff
@@ -134,12 +137,42 @@ amrex::RealVect interp_normal (amrex::ParticleReal xp, amrex::ParticleReal yp, a
...
     amrex::Real const inv_norm = 1.0_rt / std::sqrt(dot_product(normal, normal));
     AMREX_D_DECL(normal[0] *= inv_norm,
                  normal[1] *= inv_norm,
                  normal[2] *= inv_norm);
-    return normal;
+
+    // Express the unit normal in 3D Cartesian coordinates, using the Cartesian
+    // position (xp, yp, zp) to orient the radial / azimuthal components. This
+    // conversion preserves the unit length; out-of-plane components are zero.
```
RemiLehe:
```suggestion
    // position (xp, yp, zp) to orient the radial / azimuthal components.
```

Source/EmbeddedBoundary/ParticleBoundaryProcess.H
```diff
@@ -28,12 +42,96 @@ struct Absorb {
...
     }
 };
+
+/** \brief Specular reflection off the embedded boundary using bisection.
+ *
+ *  For each particle that has crossed into the EB (distance_to_eb < 0), this functor:
+ *   - Bisects along the particle trajectory to find the contact point on
+ *     the EB surface (where distance_to_eb = 0).
+ *   - Computes the surface normal at the exact contact point.
+ *   - Performs specular velocity reflection: v_new = v - 2*(v . n)*n.
```
RemiLehe:
```suggestion
 *   - Performs specular reflection on the momentum: u_new = u - 2*(u . n)*n.
```

Source/EmbeddedBoundary/ParticleScraper.H
```diff
@@ -185,25 +203,11 @@ scrapeParticlesAtEB (PC& pc, ablastr::fields::MultiLevelScalarField const& dista
                 amrex::Real W[AMREX_SPACEDIM][2];
                 ablastr::particles::compute_weights<amrex::IndexType::NODE>(
                     xp, yp, zp, plo, dxi, i, j, k, W);
-                amrex::Real const phi_value = ablastr::particles::interp_field_nodal(i, j, k, W, phi);
+                amrex::Real const distance_to_eb_value = ablastr::particles::interp_field_nodal(i, j, k, W, distance_to_eb_arr);
 
-                if (phi_value < 0.0)
+                if (distance_to_eb_value < 0.0)
                 {
-                    amrex::RealVect pos;
```
RemiLehe, line 192: No need to convert the Cartesian coordinates `xp`, `yp`, `zp` to the reduced-dimension `pos`: the functor `f` below actually needs the fully Cartesian coordinates.

Source/EmbeddedBoundary/DistanceToEB.H
```diff
@@ -28,10 +28,11 @@ amrex::Real dot_product (const amrex::RealVect& a, const amrex::RealVect& b) noe
...
 // This function calculates the unit normal vector to the embedded boundary at
 // the particle position (xp, yp, zp), by interpolating the signed distance
-// function distance_to_eb. plo/dxi are the geometry data; the returned vector
-// is normalized to unit length.
+// function distance_to_eb. plo/dxi are the geometry data. The normal is
+// always returned in 3D Cartesian coordinates.
 AMREX_GPU_HOST_DEVICE AMREX_INLINE
-amrex::RealVect interp_normal (amrex::ParticleReal xp, amrex::ParticleReal yp, amrex::ParticleReal zp,
+amrex::GpuArray<amrex::ParticleReal, 3>
+interp_normal (amrex::ParticleReal xp, amrex::ParticleReal yp, amrex::ParticleReal zp,
```
RemiLehe, line 35: As part of this PR, the signature of `interp_normal` has been changed. This function is called in two places now:
- in the particle scraper
- in the reflection functor (added as part of this PR)

In both cases, we need the normal vector in Cartesian coordinates, so including the conversion to Cartesian coordinates **as part of this PR** reduces duplication.

Source/EmbeddedBoundary/ParticleBoundaryProcess.H
```diff
@@ -28,12 +42,96 @@ struct Absorb {
...
+                     amrex::Array4<const amrex::Real> const& /*distance_to_eb*/,
+                     amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> const& /*plo*/,
+                     amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> const& /*dxi*/,
                      amrex::RandomEngine const& /*engine*/) const noexcept
     {
         amrex::ParticleIDWrapper{ptd.m_idcpu[i]}.make_invalid();
     }
 };
+
+/** \brief Specular reflection off the embedded boundary using bisection.
```
RemiLehe:
```suggestion
/** \brief Specular reflection off the embedded boundary.
```

Source/EmbeddedBoundary/ParticleBoundaryProcess.H
```diff
@@ -28,12 +42,96 @@ struct Absorb {
...
+        amrex::ParticleReal const u_dot_n = ux*n3d_x + uy*n3d_y + uz*n3d_z;
+        amrex::ParticleReal const ux_refl = ux - 2.0_prt * u_dot_n * n3d_x;
+        amrex::ParticleReal const uy_refl = uy - 2.0_prt * u_dot_n * n3d_y;
+        amrex::ParticleReal const uz_refl = uz - 2.0_prt * u_dot_n * n3d_z;
+        // Store reflected velocity
+        ptd.m_rdata[PIdx::ux][i] = ux_refl;
+        ptd.m_rdata[PIdx::uy][i] = uy_refl;
+        ptd.m_rdata[PIdx::uz][i] = uz_refl;
+
+        // Advance from contact point with reflected velocity
```
RemiLehe:
```suggestion
        // Advance from the contact point, with the reflected velocity,
```

Source/Particles/MultiParticleContainer.cpp
```diff
@@ -1246,8 +1246,20 @@ void MultiParticleContainer::CheckIonizationProductSpecies()
 void MultiParticleContainer::ScrapeParticlesAtEB (
     ablastr::fields::MultiLevelScalarField const& distance_to_eb)
 {
-    for (auto& pc : allcontainers) {
-        scrapeParticlesAtEB(*pc, distance_to_eb, ParticleBoundaryProcess::Absorb());
+    if (WarpX::eb_particle_boundary == ParticleBoundaryType::Reflecting) {
+        auto& warpx = WarpX::GetInstance();
+        for (auto& pc : allcontainers) {
+            amrex::ParticleReal const mass = pc->getMass();
+            for (int lev = 0; lev <= pc->finestLevel(); ++lev) {
```
RemiLehe, line 1253: Loop over levels is needed because `dt_lev` (used by the reflection functor) depends on the level..

Source/Particles/ParticleBoundaryBuffer.cpp
```diff
@@ -119,56 +119,33 @@ struct FindEmbeddedBoundaryIntersection {
         amrex::ParticleReal x_temp=xp, y_temp=yp, z_temp=zp;
         UpdatePosition(x_temp, y_temp, z_temp, ux, uy, uz, -dt_fraction*m_dt, m_mass);
 
-        // record the components of the normal on the destination
-        amrex::RealVect const normal = DistanceToEB::interp_normal(x_temp, y_temp, z_temp, plo, dxi, phiarr);
+        // Compute the surface normal (in 3D Cartesian coordinates) at the intersection point
+        auto const n3d = DistanceToEB::interp_normal(x_temp, y_temp, z_temp, plo, dxi, phiarr);
```
RemiLehe, line 123: Using the new version of `interp_normal` (which automatically returns a Cartesian vector) reduces some of the code below.

Source/Particles/ParticleBoundaryBuffer.cpp
```diff
@@ -119,56 +119,33 @@ struct FindEmbeddedBoundaryIntersection {
...
-        amrex::ignore_unused(y_temp);
-        //save normal components
-        dst.m_runtime_rdata[m_normal_index][dst_i] = normal[0];
-        dst.m_runtime_rdata[m_normal_index+1][dst_i] = 0.0;
-        dst.m_runtime_rdata[m_normal_index+2][dst_i] = normal[1];
 #elif (defined WARPX_DIM_RZ)
         dst.m_rdata[PIdx::r][dst_i] = std::sqrt(x_temp*x_temp + y_temp*y_temp);
         dst.m_rdata[PIdx::z][dst_i] = z_temp;
         dst.m_rdata[PIdx::theta][dst_i] = std::atan2(y_temp, x_temp);
-        //save normal components
```
RemiLehe, line 145: Not needed anymore because `interp_normal` now returns the normal in Cartesian coordinates.

Source/WarpX.H
```diff
@@ -218,6 +218,12 @@ public:
      */
     static inline amrex::Array<ParticleBoundaryType,AMREX_SPACEDIM> particle_boundary_hi;
 
+    /** Particle boundary condition at embedded boundaries.
+     *  Only Absorbing and Reflecting are supported.
```
RemiLehe:
```suggestion
```

Source/WarpX.H
```diff
@@ -218,6 +218,12 @@ public:
...
+     *  Default is Absorbing for backward compatibility.
```
RemiLehe:
```suggestion
     *  Default is Absorbing
```

atmyers:
Hi @RemiLehe 

Just a couple of comments:

1. What is the intended interaction between `boundary.particle_eb=Reflecting` and `species_name.save_particles_at_eb=1`? I think as it stands the reflected particles will not be copied to the buffer.
2. Should `eb_particle_boundary` be a species-specific option, rather than a single, shared state across all particle species?

RemiLehe:
@atmyers Thanks for your review!

> What is the intended interaction between `boundary.particle_eb=Reflecting` and `species_name.save_particles_at_eb=1`? I think as it stands the reflected particles will not be copied to the buffer.

Yes, I think that this is the intended behavior: only store the particles that were actually absorbed at the boundary. This is also consistent with the behavior at the **simulation box boundary**, where `save_particles_at_xlo/ylo/zlo/xhi/yhi/zhi` only stores particles for absorbing boundaries (but not for reflecting, periodic or thermal boundary). I will add a commit to clarify.

> Should eb_particle_boundary be a species-specific option, rather than a single, shared state across all particle species?

Yes, I think we want to eventually have a way to (optionally) specify this on a per-species basis, but I would do this in a follow-up PR, esp. because it would make sense, as part of the same PR, to have a consistent way to specify the [behavior at the **simulation box boundaries**](https://warpx.readthedocs.io/en/latest/usage/parameters.html#boundary.particle_lo-hi) on a per-species basis too. (see also https://github.com/BLAST-WarpX/warpx/pull/7008#issuecomment-4974339153)

atmyers (approved):

PR #6589, opened 2026-02-17
Separate current and mass matrix depositions

JustinRayAngus:
This PR separates the deposition of the current density (**J**) and the mass matrices (**MM**) for the implicit solvers.

When using JFNK, a deposition of **J** must be computed on the final nonlinear iteration to compute the residual. However,  deposition of **MM** on the final iteration is not needed since no linear solve will be performed. In the current development branch, **J** and **MM** are deposited together to avoid recomputing particle shape factors. While this reduces some duplicate work, it also results in an unnecessary (and expensive) final deposition of **MM**.

By separating the two operations, **MM** can be deposited only when required, improving the efficiency of the implicit solver.

This PR also enables future optimizations, such as:
- Lagging the deposition of **MM**
- Using lower-order shape factors for **MM** than for **J**

The latter may be beneficial when **MM** are only used as a preconditioner.


Source/Particles/WarpXParticleContainer.cpp
```diff
@@ -1205,202 +1175,206 @@ WarpXParticleContainer::DepositCurrentAndMassMatrices ( WarpXParIter& pti, const
...
-                uxp.dataPtr() + offset, uyp.dataPtr() + offset, uzp.dataPtr() + offset,
-                jx_fab, jy_fab, jz_fab,
-                Sxx_arr, Sxy_arr, Sxz_arr,
-                Syx_arr, Syy_arr, Syz_arr,
-                Szx_arr, Szy_arr, Szz_arr,
-                Bx_arr, By_arr, Bz_arr, Bx_type, By_type, Bz_type,
-                np_to_deposit, dt, dinv, xyzmin, lo, qs, mass);
+
+        // Ji type is same as Sii type
+        amrex::IntVect const jx_type = Sxx_fab.box().type();
```
dpgrote: Maybe change these to `Sxx_type` etc.

dpgrote (approved):
This looks good to me.

JustinRayAngus:
@RemiLehe 

I would like to merge this PR sometime this week. If you have time to take a pass at it soon that would be great.

WeiqunZhang:
Comments from Codex.

• Moving mass-matrix deposition into a new hook broke PETSc-based implicit solves because that backend never calls the new PreLinearSolve routine, so the mass matrices and related preconditioners stay stale. This regression prevents PETSc users from using the mass-matrix Jacobian or PC options, so the patch is not correct.

  Review comment:

  - [P1] PETSc backend never updates mass matrices after refactor — /home/wqzhang/mygitrepo/warpx/Source/NonlinearSolvers/WarpX_PETSc.cpp:784-841 Mass‑matrix deposition and preconditioner setup were moved out of PreRHSOp and into the new ImplicitSolver::PreLinearSolve (see Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp lines 758‑778), and the AMReX Newton solver now calls this hook before every linear solve. However, the PETSc path (SNES_impl::solve/computeRHS in Source/NonlinearSolvers/WarpX_PETSc.cpp lines 784‑841) still never invokes m_ops->PreLinearSolve, so in PETSc runs the MassMatrices_* and MassMatrices_PC MultiFabs are never refreshed anymore. As soon as a user enables implicit_evolve.use_mass_matrices_jacobian or use_mass_matrices_pc together with warpx.use_petsc = 1, the Jacobian and preconditioner will be built from stale (effectively zero) mass matrices and the implicit solve will converge incorrectly or not at all. The PETSc backend needs to call PreLinearSolve before each KSP solve, just like the Newton solver does.

JustinRayAngus:
> Comments from Codex.
> 
> • Moving mass-matrix deposition into a new hook broke PETSc-based implicit solves because that backend never calls the new PreLinearSolve routine, so the mass matrices and related preconditioners stay stale. This regression prevents PETSc users from using the mass-matrix Jacobian or PC options, so the patch is not correct.
> 
> Review comment:
> 
> * [P1] PETSc backend never updates mass matrices after refactor — /home/wqzhang/mygitrepo/warpx/Source/NonlinearSolvers/WarpX_PETSc.cpp:784-841 Mass‑matrix deposition and preconditioner setup were moved out of PreRHSOp and into the new ImplicitSolver::PreLinearSolve (see Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp lines 758‑778), and the AMReX Newton solver now calls this hook before every linear solve. However, the PETSc path (SNES_impl::solve/computeRHS in Source/NonlinearSolvers/WarpX_PETSc.cpp lines 784‑841) still never invokes m_ops->PreLinearSolve, so in PETSc runs the MassMatrices_* and MassMatrices_PC MultiFabs are never refreshed anymore. As soon as a user enables implicit_evolve.use_mass_matrices_jacobian or use_mass_matrices_pc together with warpx.use_petsc = 1, the Jacobian and preconditioner will be built from stale (effectively zero) mass matrices and the implicit solve will converge incorrectly or not at all. The PETSc backend needs to call PreLinearSolve before each KSP solve, just like the Newton solver does.

AI?

Yes. I'm aware of this. Right now things don't work using PETSc's SNES nonlinear solver anyway except for the special case of periodic BCs (@debog). I have an idea for how to make this PR work with PETSc.

JustinRayAngus:
@WeiqunZhang 

I fixed the issue with PETSc SNES. Can you have Codex do another review? I'm curious what it says.

WeiqunZhang:
After working for 5 minutes, Codex says,

• Calling PreLinearSolve unconditionally for the PETSc solver now overwrites the saved base electric field with the perturbed fields used in each GMRES Jacobian evaluation, which breaks the mass-matrix-based Jacobian computation. This regression makes the PETSc implicit solver incorrect when mass matrices are enabled.

  Review comment:

  - [P1] Avoid calling PreLinearSolve during PETSc Jacobian evaluations — Source/FieldSolver/ImplicitSolvers/ImplicitSolver.cpp:784-842
    At the end of ImplicitSolver::PreRHSOp you now call PreLinearSolve on every invocation when m_nlsolver_type == petsc_snes (lines 837‑842), regardless of whether the call originates from the nonlinear residual (a_from_jacobian=false) or from a PETSc GMRES Jacobian-vector product (a_from_jacobian=true). PreLinearSolve (lines 758‑778) unconditionally recomputes the mass matrices and calls SaveE(), which copies the current Efield_fp into Efield_fp_save. During GMRES the fields are perturbed for the finite-difference Jacobian, so this overwrites the stored base field with the perturbed one. The next Jacobian application then sees E - Efield_fp_save ≈ 0 in ComputeJfromMassMatrices (lines 128‑172), so the mass-matrix Jacobian contributes nothing and the linear solve effectively breaks. PreLinearSolve must only run once per nonlinear residual evaluation (i.e., when a_from_jacobian is false) so that the saved field and mass matrices continue to represent the base Newton state.

WeiqunZhang:
Codex has something to say again.

• Non‑suborbit currents are never inverse‑volume scaled in cylindrical geometries, so every implicit RZ/RCYL/RSPHERE run now produces incorrect currents and Jacobians.

  Review comment:

  - [P1] Scale non‑suborbit current before CumulateJ — /home/wqzhang/mygitrepo/warpx/Source/Evolve/WarpXEvolve.cpp:1315-1324
    Non‑suborbit particles now deposit all of their current into FieldType::current_fp_non_suborbit (PhysicalParticleContainer.cpp l.682‑689), and ImplicitSolver::CumulateJ() simply adds that MultiFab onto current_fp and also uses it as J0 inside ComputeJfromMassMatrices(). In cylindrical geometries these raw deposits must be multiplied by the inverse cell volume (Verboncoeur factor), but WarpX::PushParticlesandDeposit() still calls ApplyInverseVolumeScalingToCurrentDensity only for current_fp (and current_buf) as shown here. Because the new container is never scaled, every implicit RZ/RCYL/RSPHERE run now adds unscaled current_fp_non_suborbit data to the scaled current_fp, so the total current and the mass‑matrix Jacobian are wrong. Please apply the same inverse‑volume scaling to current_fp_non_suborbit before it is consumed by CumulateJ()/ComputeJfromMassMatrices().

JustinRayAngus:
> Codex has something to say again.
> 
> • Non‑suborbit currents are never inverse‑volume scaled in cylindrical geometries, so every implicit RZ/RCYL/RSPHERE run now produces incorrect currents and Jacobians.
> 
> Review comment:
> 
> * [P1] Scale non‑suborbit current before CumulateJ — /home/wqzhang/mygitrepo/warpx/Source/Evolve/WarpXEvolve.cpp:1315-1324
>   Non‑suborbit particles now deposit all of their current into FieldType::current_fp_non_suborbit (PhysicalParticleContainer.cpp l.682‑689), and ImplicitSolver::CumulateJ() simply adds that MultiFab onto current_fp and also uses it as J0 inside ComputeJfromMassMatrices(). In cylindrical geometries these raw deposits must be multiplied by the inverse cell volume (Verboncoeur factor), but WarpX::PushParticlesandDeposit() still calls ApplyInverseVolumeScalingToCurrentDensity only for current_fp (and current_buf) as shown here. Because the new container is never scaled, every implicit RZ/RCYL/RSPHERE run now adds unscaled current_fp_non_suborbit data to the scaled current_fp, so the total current and the mass‑matrix Jacobian are wrong. Please apply the same inverse‑volume scaling to current_fp_non_suborbit before it is consumed by CumulateJ()/ComputeJfromMassMatrices().

@WeiqunZhang 
This makes sense. I fixed it. The implicit methods don't fully work with non-planar geometries right now. It is high on my priority list to fix that.

JustinRayAngus:
@WeiqunZhang Can you run CodeX again?

WeiqunZhang:
I will run Codex again. For things that are not fully implemented, having asserts will reduce the warnings from AIs.

JustinRayAngus:
> I will run Codex again. For things that are not fully implemented, having asserts will reduce the warnings from AIs.

Any examples? There are asserts if you try to use the mass matrices with curvilinear geometries that will prevent you from doing so.

WeiqunZhang:
Maybe it's because it is not in the place it is looking at. For example, if function f1 has certain assumptions and the assertion is in function f2. From our point of view, if f1 is only called by f2, that is fine. But AIs might say f1 is a public function that could be called by other functions even if not now, maybe in the future.

WeiqunZhang:
Codex now says,

• The changes introduce a regression where fluid species are no longer evolved during implicit solves and cylindrical/spherical implicit runs now skip the inverse-volume scaling for buffer currents and charge densities, both of which yield incorrect physics.

  Full review comments:

  - [P2] Keep fluid species evolution when implicit solver is active — /home/wqzhang/mygitrepo/warpx/Source/Evolve/WarpXEvolve.cpp:1315-1349
    The new guard if (!skip_deposition && !implicit_options) now wraps the fluid advance (myfl->Evolve). Because implicit_options is non-null for every particle push performed by the implicit solver, this condition is never satisfied in implicit runs, so fluid species are no longer pushed or have their current deposited whenever the implicit solver is enabled. Any simulation that combines the implicit solver with fluid species will therefore leave the fluid component frozen and gives incorrect fields. The fluid advance should only be gated by skip_deposition, not by whether implicit options are supplied.
  - [P1] Scale RZ buffer currents/charge even when using implicit solver — /home/wqzhang/mygitrepo/warpx/Source/Evolve/WarpXEvolve.cpp:1315-1335
    The same !implicit_options guard now skips the inverse-volume scaling of current_fp, current_buf, rho_fp, and rho_buf for cylindrical/spherical geometries whenever the implicit solver is in use. While ImplicitSolver::PreRHSOp added a replacement scaling step, it only rescales current_fp (lines 831‑835 of ImplicitSolver.cpp), so the buffer currents and charge densities remain unscaled when SyncCurrentAndRho is called. Every implicit run in RZ/RCYL/RSP grids will therefore propagate unphysical currents/charges across coarse–fine boundaries and guard cells. The scaling of current_buf and rho needs to occur regardless of implicit_options, either here or in the new implicit path.

WeiqunZhang:
I have been using gpt-5.3 high. I am also going to try a different model.

WeiqunZhang:
codex with gpt-5.4 xhigh says

 In curvilinear geometries the implicit solver now skips the inverse-volume scaling for charge density and coarse-patch currents, which breaks any RZ/RCYL/RSPHERE run. This is a correctness regression introduced by the patch.
  
  Review comment:
  
  - [P1] Scale rho/current in RZ even with implicit solver — Source/Evolve/WarpXEvolve.cpp:1315-1344
    The PushParticlesandDeposit block that applies ApplyInverseVolumeScalingToCurrentDensity/ApplyInverseVolumeScalingToChargeDensity is now guarded by !implicit_options (lines 1315‑1344), so whenever the implicit solver is active the RZ/RCYL/RSPHERE Jacobian scaling is skipped entirely. ImplicitSolver::PreRHSOp only compensates by scaling the fine–grid current_fp (added block in the same patch) and never touches the coarse‐patch currents (current_buf) or any of the deposited charge densities (rho_fp/rho_buf). As a result, every implicit run in cylindrical/spherical coordinates integrates unscaled charge and coarse‑patch currents, yielding incorrect field solves and refinement syncs. The fix is to continue applying the inverse‑volume scaling for both current_fp/current_buf and rho_fp/rho_buf even when implicit_options is non‑null.
 - [P2] Frozen do_not_push species now contribute mass matrices — Source/Particles/PhysicalParticleContainer.cpp:571, Source/Particles/PhysicalParticleContainer.cpp:804. The old MM deposit lived under the !do_not_push path; after the refactor the standalone DepositMassMatrices() runs for every PhysicalParticleContainer, so a charged frozen species still feeds the Jacobian/PC even though its current is never advanced/deposited.
  - [P2] MM deposition no longer follows the filtered gather fields — Source/Particles/PhysicalParticleContainer.cpp:522, Source/Particles/PhysicalParticleContainer.cpp:804. The implicit push can use applyNCIFilter, but the new standalone MM pass always samples raw Bfield_aux; with use_fdtd_nci_corr=1, the MM linearization is built from a different particle orbit than the residual.
  - Assumption: if you only care about single-level, no-NCI, no-do_not_push implicit runs, I don’t see another blocker beyond the inverse-volume issue.

JustinRayAngus:
> codex with gpt-5.4 xhigh says
> 
> In curvilinear geometries the implicit solver now skips the inverse-volume scaling for charge density and coarse-patch currents, which breaks any RZ/RCYL/RSPHERE run. This is a correctness regression introduced by the patch.
> 
> Review comment:
> 
> * [P1] Scale rho/current in RZ even with implicit solver — Source/Evolve/WarpXEvolve.cpp:1315-1344
>   The PushParticlesandDeposit block that applies ApplyInverseVolumeScalingToCurrentDensity/ApplyInverseVolumeScalingToChargeDensity is now guarded by !implicit_options (lines 1315‑1344), so whenever the implicit solver is active the RZ/RCYL/RSPHERE Jacobian scaling is skipped entirely. ImplicitSolver::PreRHSOp only compensates by scaling the fine–grid current_fp (added block in the same patch) and never touches the coarse‐patch currents (current_buf) or any of the deposited charge densities (rho_fp/rho_buf). As a result, every implicit run in cylindrical/spherical coordinates integrates unscaled charge and coarse‑patch currents, yielding incorrect field solves and refinement syncs. The fix is to continue applying the inverse‑volume scaling for both current_fp/current_buf and rho_fp/rho_buf even when implicit_options is non‑null.
> * [P2] Frozen do_not_push species now contribute mass matrices — Source/Particles/PhysicalParticleContainer.cpp:571, Source/Particles/PhysicalParticleContainer.cpp:804. The old MM deposit lived under the !do_not_push path; after the refactor the standalone DepositMassMatrices() runs for every PhysicalParticleContainer, so a charged frozen species still feeds the Jacobian/PC even though its current is never advanced/deposited.
> * [P2] MM deposition no longer follows the filtered gather fields — Source/Particles/PhysicalParticleContainer.cpp:522, Source/Particles/PhysicalParticleContainer.cpp:804. The implicit push can use applyNCIFilter, but the new standalone MM pass always samples raw Bfield_aux; with use_fdtd_nci_corr=1, the MM linearization is built from a different particle orbit than the residual.
> * Assumption: if you only care about single-level, no-NCI, no-do_not_push implicit runs, I don’t see another blocker beyond the inverse-volume issue.

The coarse patch and rho are not used for the implicit solvers.

JustinRayAngus:
@WeiqunZhang @RemiLehe 

I'm not familiar with the fdtd_nci_corr filter. I see that it is related to numerical Cherenkov and that it is only to be used with the Esirkepov deposition. I'm going to put in an assert that this flag is false for the implicit solvers. If there is nothing else from CodeX (or humans), then this PR should be wrapped up.

WeiqunZhang (approved):

PR #6591, opened 2026-02-18
Docs: update instructions to compile/use WarpX on Adastra (CINES)

lucafedeli88:
This PR updates the instructions to compile and use WarpX on the Adastra supercomputer (CINES, France).

The main changes are the following:

- AMD compiler instead of Cray compiler is used by default for performance reasons
- mpich v9.0.1 , cpe v25.09, and rocm v6.4.3 are now used
- boost is installed manually from source (the version available on Adastra is not compatible with the aforementioned modules)
- blosc v2.23 instead of blosc v1.21.1 is used for compatibility reasons with ADIOS2 v2.10.2
- workarounds in the submission script have been mostly eliminated, since they do not seem to be necessary

### DIFF
--- a/Tools/machines/adastra-cines/submit.sh
+++ b/Tools/machines/adastra-cines/submit.sh
@@ -10,42 +10,19 @@
 module purge
 
 # A CrayPE environment version
-module load cpe/24.07
+module load cpe/25.09
 # An architecture
 module load craype-accel-amd-gfx90a craype-x86-trento
 # A compiler to target the architecture
-module load PrgEnv-cray
-# Some architecture related libraries and tools
-module load develop
-module load CCE-GPU-4.0.0
-# AMD related libraries
-module load rocm/6.1.2
-module load amd-mixed/6.1.2
-# note
-# cray-mpich versions 8.1.28 and 8.1.30 have known issues
-# that cause node memory increase over time which leads
-# to slowdown and out-of-memory crashes.
-module load cray-mpich/8.1.26
+module load rocm/6.4.3
+module load PrgEnv-amd
+# The MPI library
+module load cray-mpich/9.0.1
 
 date
 module list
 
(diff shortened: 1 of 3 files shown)


phsikorski-source (approved):
I tested this on Adastra. Working well.

Tools/machines/adastra-cines/submit.sh
```diff
@@ -10,43 +10,20 @@
...
-module load cray-mpich/8.1.26
+module load rocm/6.4.3
+module load PrgEnv-amd
+# The MPI library
+module load cray-mpich/9.0.1
 
 date
 module list
 
 export MPICH_GPU_SUPPORT_ENABLED=1
```
grobertdautun, line 26: I encounter a library error while running and @Pierre-Bartoli found how to load it manually, we could add something like this in either submit.sh or the profile
`# note`
`# if you encounter the "error while loading shared libraries: libucp.so.0" `
`# uncomment the following line to manually find the library`
`export LD_LIBRARY_PATH=/usr/diags/ucx/lib/:$LD_LIBRARY_PATH`

grobertdautun (commented):
Compilation works fine for me however I encounter an error when running warpx regarding a library that cannot be found. I proposed a solution in the submit.sh comments but it can go in the profile or in the documentation known issues

EZoni:
@lucafedeli88 @grobertdautun 

Does this PR need a decision on https://github.com/BLAST-WarpX/warpx/pull/6591#discussion_r2952308057 or is it ready to review and merge?

lucafedeli88:
@EZoni , today @phsikorski-source tried to compile again WarpX using these instructions and he is able to run simulations without adding paths to `LD_LIBRARY_PATH`. So, I think that the issue that @grobertdautun observed was just a temporary glitch of the machine.

EZoni (approved):

